PR #228, opened 2017-08-30
Add start/end tokens to tokenizers, instead of in readers

### LINKED ISSUE #191: Add ability to append or prepend tokens to the `Tokenizers`
In three places now we've implemented appending and/or prepending tokens to what goes into a `TextField` (a null token for the SNLI reader, a stop token for SQuAD, and sentence boundary tokens for the language model reader).  This should just be a basic functionality of the tokenizer.

matt-gardner:
Fixes #191.  This makes it a lot easier to add things like start and end characters, which otherwise would be very difficult to implement.  I also added the ability to change the padding index for characters when you're using byte encoding (MattP's code using 261 for the padding index, e.g., instead of 0, because 0 could be a byte in your byte stream*).

*Actually, I don't think that 0 is a valid byte in UTF-8, but we'll ignore that for now...

### DIFF
--- a/allennlp/data/tokenizers/character_tokenizer.py
+++ b/allennlp/data/tokenizers/character_tokenizer.py
@@ -40,6 +40,8 @@ class CharacterTokenizer(Tokenizer):
         self._byte_encoding = byte_encoding
         self._lowercase_characters = lowercase_characters
         self._start_tokens = start_tokens or []
+        # We reverse the tokens here because we're going to insert them with `insert(0)` later;
+        # this makes sure they show up in the right order.
         self._start_tokens.reverse()
         self._end_tokens = end_tokens or []
 
--- a/allennlp/data/tokenizers/character_tokenizer.py
+++ b/allennlp/data/tokenizers/character_tokenizer.py
@@ -29,22 +29,48 @@ class CharacterTokenizer(Tokenizer):
         If ``True``, we will lowercase all of the characters in the text before doing any other
         operation.  You probably do not want to do this, as character vocabularies are generally
         not very large to begin with, but it's an option if you really want it.
+    start_tokens : ``List[str]``, optional
+        If given, these tokens will be added to the beginning of every string we tokenize.  If
+        using byte encoding, this should actually be a ``List[int]``, not a ``List[str]``.
+    end_tokens : ``List[str]``, optional
+        If given, these tokens will be added to the end of every string we tokenize.  If using byte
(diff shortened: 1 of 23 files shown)


matt-gardner:
Sorry, @DeNeutoy, this isn't ready yet.  Don't bother looking at it.  I'll let you know when it's ready.

matt-gardner:
Ok @DeNeutoy, this is now ready to be looked at.

allennlp/data/tokenizers/character_tokenizer.py
```diff
@@ -17,34 +17,56 @@ class CharacterTokenizer(Tokenizer):
...
-        self.lowercase_characters = lowercase_characters
+    def __init__(self,
+                 byte_encoding: str = None,
+                 lowercase_characters: bool = False,
+                 start_tokens: List[str] = None,
+                 end_tokens: List[str] = None) -> None:
+        self._byte_encoding = byte_encoding
+        self._lowercase_characters = lowercase_characters
+        self._start_tokens = start_tokens or []
+        self._start_tokens.reverse()
```
DeNeutoy: What does this do?

allennlp/models/bidaf.py
```diff
@@ -331,18 +330,12 @@ def predict_span(self, question: TextField, passage: TextField) -> Dict[str, Any
...
-        span_start_logits = output_dict["span_start_logits"]
-        span_end_logits = output_dict["span_end_logits"]
-        best_span = self._get_best_span(span_start_logits, span_end_logits)
-        best_span = tuple(best_span.data.squeeze(0).numpy())
-        best_span_str = ' '.join(passage.tokens[best_span[0]:best_span[1]])
-
+        # Here we're just removing the batch dimension and converting things to numpy arrays /
+        # tuples instead of pytorch variables.
         return {
                 "span_start_probs": output_dict["span_start_probs"].data.squeeze(0).numpy(),
```
DeNeutoy: Do these need `.cpu()` to allow prediction on the GPU?

allennlp/data/tokenizers/character_tokenizer.py
matt-gardner: It's because of the `tokens.insert(0, start_token)` - you need to insert them in the reverse order so that they end up right.  I'll add a comment explaining this.

allennlp/models/bidaf.py
matt-gardner: Ah, yes, good catch, thanks.

PR #230, opened 2017-08-30
first iteration on srl viz

joelgrus:
screen shot 2017-08-30 at 2 49 57 pm


schmmd:
OK... so feel free to ignore any of my design advice.  I find the red and blue quite hard to look at.  I would choose softer colors.  I would also consider highlighting the words as well as the ARG/V stuff.  Feel free to ignore my design advice if you think it's bad--I'm not a designer...

PR #234, opened 2017-08-31
Highlight answer in paragraph for MC.

schmmd:


schmmd:
image

Doesn't look that great yet, but I'm going to see if Aaron can look at it and also look at the SRL "scrolling" issue which is pretty bad.

allennlp/common/util.py
```diff
@@ -9,7 +9,10 @@
 
 def sanitize(x: Any) -> Any:  # pylint: disable=invalid-name
     """
-    Sanitize x so that it can be JSON serialized
+    Sanitize x so that it can be JSON serialized.
+
+    This turn PyTorch and Numpy types into basic Python types so they
```
joelgrus: turns

also, I would get rid of the first sentence, it's redundant now

allennlp/service/static/index.html
```diff
@@ -125,13 +125,21 @@
...
     return (
       <div className="model__content">
         <div className="form__field">
           <label>Answer</label>
-          <div className="model__content__summary">{ JSON.stringify(rawOutput, null, "  ") }</div>
+          <div className="model__content__summary">{ answer }</div>
+          <div>
+            <span>{head}</span>
+            <span><strong>{answer}</strong></span>
+            <span>{tail}</span>
```
joelgrus: rather than doing it this way, I would add classes

```
<div className="passage">
  <span className="passage-head"> ...
  <span className="passage-answer"> ...
  <span className="passage-tail"> ...
</div>
```

and then use CSS to format them, that way it's much easier to play around with different formattings

joelgrus: and also I think that's just better practice

PR #237, opened 2017-08-31
move optimizer creation inside of Trainer.from_params

### LINKED ISSUE #205: Move the call to `model.cuda()` to before optimizer creation
Because of [this issue](https://github.com/pytorch/pytorch/issues/2021#issuecomment-325204922).  The optimizer might have state that's initialized from the model parameters, and needs to be on the right device.

This means we should either:
1. Have `cuda_device` be a top-level key in the experiment config, so we can move the model over in `commands.train()` before constructing the optimizer.
2. Move the optimizer creation and the call to `model.cuda()` into `Trainer.from_params()`.

I could go either way.  Calling `model.cuda()` inside of `from_params()` in the second option is a little bit more logic than we like to have in those methods, but not much.  The optimizer conceptually seems like it's part of the trainer, so having the optimizer params inside of the trainer params makes sense.

DeNeutoy:
Fixes #205.  This will allow me to use adagrad for the decomposable attention model.


allennlp/training/trainer.py
```diff
@@ -339,25 +340,30 @@ def _restore_checkpoint(self) -> int:
...
         patience = params.pop("patience", 2)
         validation_metric = params.pop("validation_metric", "-loss")
         num_epochs = params.pop("num_epochs", 20)
         cuda_device = params.pop("cuda_device", -1)
         grad_norm = params.pop("grad_norm", None)
         grad_clipping = params.pop("grad_clipping", None)
         lr_scheduler_params = params.pop("learning_rate_scheduler", None)
+
+        if cuda_device >= 0:
+            model = model.cuda(cuda_device)
```
matt-gardner: This is now duplicated in the init.  Would it make sense to remove it from there?  Not sure...

matt-gardner:
Fixes #205.

matt-gardner:
Hmm, I gave that extra comment because when I put that in the description once, it automatically linked it, like it does with commit messages.  Might not work with comments, though...

allennlp/training/trainer.py
DeNeutoy: Yeah, struggled a bit with that - if you aren't building from params, this would then have different behaviour, but equally, whether we want that behaviour there or not I don't know.

DeNeutoy:
Huh cool, I didn't know you could do that.

Issue #244, opened 2017-08-31
Allow config file overrides in `evaluate`


matt-gardner:
And other places where we load models.  Actually, because this is going through to `load_archive`, that method itself has to allow overrides, and any entry point that reaches `load_archive` needs a way to pass in overrides.

schmmd:
@matt-gardner what do you mean exactly by `evaluate`?  `allennlp/run evaluate` or a particular method?  Why do you need the ability to override here?

matt-gardner:
Yeah, I meant the command.  See #246 for the original motivation for this.  Not really sure how necessary this is anymore...  (The whole [`_sanitize_config()`](https://github.com/allenai/allennlp/blob/83d0253cd1bcba18b9fe31961146a8e81c0eae8c/allennlp/models/archival.py#L52-L66) method can probably be removed at this point, actually.)  I haven't recently run into a situation where I really wished I had this.  I know @OyvindTafjord ran into it a lot more frequently, though.

schmmd:
My WIP change https://github.com/allenai/allennlp/pull/402 support overriding fields, but does not support removing fields.

matt-gardner:
But that parameter was removed a long time ago, so that particular use case is no longer present.  You could also override it with `None` to effectively remove it (which means you should probably do something like our `replace_none` method in your override code).

schmmd:
The original motivation for this issue is moot.

matt-gardner:
I just talked with Oyvind, and he brought up a few cases where being able to change test-time configuration is helpful, and is currently a huge pain.

- Changing the batch size, because, e.g., the test data has some really long sentences
- Trying hacks with padding, like what Oyvind did to the character CNN in BiDAF

These cases are pretty rare, though.  We may want to open this issue again at some point in the future, but for now there's not enough motivation for it.

PR #246, opened 2017-08-31
remove evalution_json_file key from config during load_archive

joelgrus:
this may be too aggressive (I can imagine wanting to use the evaluation_json_file in `run evaluate`, if that's the case I could check if it exists or pass through a flag not to delete it, but this was simplest)

(also, I know the config file in the test is sort of nonsensical)

### DIFF
--- a/allennlp/models/archival.py
+++ b/allennlp/models/archival.py
@@ -49,6 +49,16 @@ def archive_model(serialization_dir: str,
         archive.add(os.path.join(serialization_dir, "vocabulary"),
                     arcname="vocabulary")
 
+def _sanitize_config(config: Params) -> None:
+    """
+    There are some elements of the model config that we need to get rid of
+    when we load from archive, as they refer to paths on the training machine
+    that are unlikely to exist on the un-archiving machine. To be extra-safe
+    we just remove them from the config.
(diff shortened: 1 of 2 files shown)


allennlp/models/archival.py
```diff
@@ -49,6 +49,16 @@ def archive_model(serialization_dir: str,
...
 
+def _sanitize_config(config: Params) -> None:
+    """
+    There are some elements of the model config that we need to get rid of
+    when we load from archive, as they refer to paths on the training machine
+    that are unlikely to exist on the un-archiving machine. To be extra-safe
+    we just remove them from the config.
+    """
+    # TODO(joelgrus): come up with a better solution for this
+    config.get("model", {}).pop('evaluation_json_file', None)
```
matt-gardner: Yeah, can you check if this exists before removing it?  I definitely want this to be there when running `evaluate` - that's how I knew that my model archive worked.

Other than that you might say in the docstring that this is a temporary fix until we have a better solution for https://github.com/allenai/allennlp/issues/244.

matt-gardner: Maybe also log a warning when you remove it.

PR #247, opened 2017-08-31
Simple LSTM wrapper with better default initialization

OyvindTafjord:
Not urgent, but this is an attempt at a mechanism for providing better default initialization for, say, LSTMs (and also suppress training of the superfluous `bias_hh` variables). I considered adding extra arguments to this new class so that it's easier to change the details, but that would be inconsistent with the other paradigm of parameter regexes global to the model. Maybe later though we'll also have module-specific initialization settings more generally, along the lines of Keras etc.

### DIFF
--- a/allennlp/modules/default_lstm.py
+++ b/allennlp/modules/default_lstm.py
@@ -1,0 +1,36 @@
+import re
+from overrides import overrides
+import torch
+
+from allennlp.nn.initializers import block_orthogonal
+
+
+class DefaultLstm(torch.nn.LSTM):
+    """
+    A wrapper of the Pytorch LSTM which provides a good default initialization
+    of weights and biases, and disables training of the superfluous bias_hh parameters.
+
+    Input parameters are identical to Pytorch LSTM.
+    """
+    @overrides
+    def reset_parameters(self):
+        hidden_size = self.hidden_size
+        for param_name, param in self.named_parameters():
+            if re.search("bias_hh", param_name):
+                torch.nn.init.constant(param, 0)
+                # This parameter is superfluous, so we should not update it in training
+                param.requires_grad = False
+            elif re.search("bias_ih", param_name):
+                bias = param.data
+                bias_size = bias.size(0)
+                # The forget get bias is between fractions 1/4 and 1/2 of the vector
+                start, end = bias_size//4, bias_size//2
+                bias.fill_(0)
+                bias[start:end].fill_(1)
+            elif re.search("weight_ih", param_name):
+                # Initialize each block matrix separately
+                for i in range(0, 4):
+                    block = param.data[i*hidden_size:(i+1)*hidden_size]
+                    torch.nn.init.xavier_uniform(block)
+            elif re.search("weight_hh", param_name):
+                block_orthogonal(param, [hidden_size, hidden_size])
--- a/allennlp/modules/initialized_lstm.py
+++ b/allennlp/modules/initialized_lstm.py
@@ -24,6 +24,7 @@ class InitializedLstm(torch.nn.LSTM):
                 param.requires_grad = False
             elif param_name.startswith("bias_ih"):
                 bias = param.data
+                bias.fill_(0)
                 # The forget get bias is in the second hidden_size block
                 bias[hidden_size:2*hidden_size].fill_(1)
             elif param_name.startswith("weight_ih"):
--- a/allennlp/modules/initialized_lstm.py
+++ b/allennlp/modules/initialized_lstm.py
@@ -1,14 +1,16 @@
-import re
 from overrides import overrides
 import torch
 
 from allennlp.nn.initializers import block_orthogonal
 
 
-class DefaultLstm(torch.nn.LSTM):
+class InitializedLstm(torch.nn.LSTM):
     """
     A wrapper of the Pytorch LSTM which provides a good default initialization
     of weights and biases, and disables training of the superfluous bias_hh parameters.
+
+    Warning: Parameters with `param.requires_grad` set to `False` should not be passed
+    to the optimizer.
 
     Input parameters are identical to Pytorch LSTM.
     """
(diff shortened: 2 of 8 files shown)


allennlp/modules/default_lstm.py
```diff
@@ -0,0 +1,36 @@
...
+    A wrapper of the Pytorch LSTM which provides a good default initialization
+    of weights and biases, and disables training of the superfluous bias_hh parameters.
+
+    Input parameters are identical to Pytorch LSTM.
+    """
+    @overrides
+    def reset_parameters(self):
+        hidden_size = self.hidden_size
+        for param_name, param in self.named_parameters():
+            if re.search("bias_hh", param_name):
```
matt-gardner: You shouldn't have to do a search here; you should get the name exactly, and can just match on that (`if name == "biah_hh":`).  Because you're calling `self.named_parameters()`, there's no nesting of names, or anything - you'll just get `"bias_hh"`, etc., as the names.

matt-gardner: In fact, you can just remove the whole loop over `self.named_parameters()`, and just directly initialize the things you want to initialize.  It's a tiny bit less robust to future changes in pytorch, but I think that's a feature, not a bug - if pytorch changes their LSTMs enough to break this, we will want to do things differently, too.

OyvindTafjord: There are the different layers and directions though, e.g., here's a full list for a 2-layer bidirectional one: `['weight_hh_l1_reverse', 'bias_hh_l1', 'bias_ih_l0', 'weight_ih_l1', 'weight_hh_l0', 'bias_ih_l1_reverse', 'bias_ih_l0_reverse', 'bias_hh_l1_reverse', 'weight_ih_l0', 'bias_hh_l0', 'weight_ih_l1_reverse', 'bias_ih_l1', 'weight_ih_l0_reverse', 'weight_hh_l0_reverse', 'weight_hh_l1', 'bias_hh_l0_reverse']`. I can generate that list, assuming nothing changes (and as you say, if it does, it might be a feature to find out), but this seemed cleaner.

OyvindTafjord:
Yes, not thrilled with the `DefaultLstm` name either, and thought of just using `Lstm` or `InitializedLstm`. It depends a bit on what tasks we think such wrappers might take on (e.g., in Keras there are various regularizers/constraints/activations at this level as well).

allennlp/modules/default_lstm.py
matt-gardner: Oh, yep, sorry, my bad.  You're right.

matt-gardner:
I like `InitializedLstm` the most out of the options that I've seen.

allennlp/modules/default_lstm.py
joelgrus: nit: it looks like in every case you're really just checking `param_name.startswith()`? if so, I feel like it's more readable that way.

allennlp/modules/default_lstm.py
```diff
@@ -0,0 +1,36 @@
...
+                torch.nn.init.constant(param, 0)
+                # This parameter is superfluous, so we should not update it in training
+                param.requires_grad = False
+            elif re.search("bias_ih", param_name):
+                bias = param.data
+                bias_size = bias.size(0)
+                # The forget get bias is between fractions 1/4 and 1/2 of the vector
+                start, end = bias_size//4, bias_size//2
+                bias.fill_(0)
+                bias[start:end].fill_(1)
```
DeNeutoy: can't it just be: `bias[hidden_size: 2 * hidden_size].fill_(1)`?

allennlp/modules/default_lstm.py
```diff
@@ -0,0 +1,36 @@
...
+    Input parameters are identical to Pytorch LSTM.
+    """
+    @overrides
+    def reset_parameters(self):
+        hidden_size = self.hidden_size
+        for param_name, param in self.named_parameters():
+            if re.search("bias_hh", param_name):
+                torch.nn.init.constant(param, 0)
+                # This parameter is superfluous, so we should not update it in training
+                param.requires_grad = False
```
DeNeutoy: I don't think this is worth the confusion for an initial user of having to do the `parameters = [p for p in self.model.parameters() if p.requires_grad]` trick were they to try to use this LSTM with their own training code. It's also pretty harmless.

allennlp/modules/default_lstm.py
```diff
@@ -0,0 +1,36 @@
...
+            elif re.search("bias_ih", param_name):
+                bias = param.data
+                bias_size = bias.size(0)
+                # The forget get bias is between fractions 1/4 and 1/2 of the vector
```
DeNeutoy: "forget gate"

allennlp/modules/default_lstm.py
OyvindTafjord: Sure, I'll change that

allennlp/modules/default_lstm.py
OyvindTafjord: Yes, that's clearer

allennlp/modules/default_lstm.py
OyvindTafjord: Right, that might be confusing, I don't have a good sense of that. I'm not sure how harmless or not it is. It seems unnecessary to me to have hundreds of completely flat directions in your parameter space, although it's a small enough fraction of the total parameters that it practically might not matter much. Maybe I can make it a boolean flag which is off by default to avoid this annoyance with parameters (note we do this `requires_grad` filtering by default in our training code since we also have things like word embeddings which maybe be fixed, so it's probably general good practice to do this). We could also add a method to our `Model` class to get trainable parameters which does this filtering for you.

OyvindTafjord:
OK, will change to `InitializedLstm` and make the other small suggested changes. I'm interested in opinions on whether we should, by default, disable gradients from the superfluous bias parameters or not.

matt-gardner:
I vote for disabling it by default, with a warning in the class docstring, so no one stumbles into the error.  If they're using our code to get the `InitializedLstm` `from_params`, then they are probably using our training code too, which does the right thing.  If they're not using our framework, they will (hopefully) have read the docstring for `InitializedLstm`.

OyvindTafjord:
I have made the discussed changes, but I propose just closing this PR for now and rather consider this pattern down the line if we decide more generally to add more custom initializations. Reasons:
   * Performance on BiDAF+SQuAD actually dropped (to EM 67.3 from EM 68.7 or so) with this change
   * For consistency it should be done more systematically across other modules, like GRU etc
   * We may decide on a different approach to initialization configuration

PR #251, opened 2017-09-01
Update kubernetes-webdemo configuration.

schmmd:
* 4 replicas.
* Explicit image version.


schmmd:
I `kubectl apply -f`ed this too.

schmmd:
@joelgrus I thought it would be helpful to keep the demo versions in a separate namespace.  We'll potentially be releasing the library and what we deploy to the demo quite separately.  I'd like to reserve `allennlp/allennlp` for clear,  versioned releases.  Does that make sense?  Happy if we come up with a better strategy.

joelgrus:
makes sense

PR #252, opened 2017-09-01
add cache to web server

joelgrus:
I'm not thrilled with this code, but it would take a lot more work to make it really good.

it can be disabled with an environment variable, in case things go awry.

### DIFF
--- a/allennlp/service/server_sanic.py
+++ b/allennlp/service/server_sanic.py
@@ -41,7 +41,11 @@ def make_app() -> Sanic:
     app.static('/', './allennlp/service/static/index.html')
     app.predictors = {}
 
-    cache_size = int(CACHE_SIZE)
+    try:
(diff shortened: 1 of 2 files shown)


allennlp/service/server_sanic.py
```diff
@@ -38,11 +57,27 @@ def make_app() -> Sanic:
...
+                prediction = model.predict_json(data)
         except KeyError as err:
             raise ServerError("Required JSON field not found: " + err.args[0], status_code=400)
 
+        post_hits = _caching_prediction.cache_info().hits  # pylint: disable=no-value-for-parameter
+
+        if post_hits > pre_hits:
+            # Cache hit, so insert an artifical pause
+            logger.info("cache hit: %s %s", model_name, json.dumps(data))
+            await asyncio.sleep(0.25)
```
schmmd: he he he

PR #258, opened 2017-09-04
Centralize creation of TokenIndexer dictionary

### LINKED ISSUE #181: Simplify / centralize `TokenIndexer.from_params()`
We have blocks like this in several places: https://github.com/allenai/allennlp/blob/166809c563f4d30342725fed524cb6fed00fe5bd/allennlp/data/dataset_readers/snli.py#L78-L85

These should all be put in one spot, probably something like `TokenIndexer.dict_from_params` (not thrilled with that name, but something similar).

matt-gardner:
Fixes #181.  I'm not thrilled with the method name I used here, but that's easy to change.

### DIFF
--- a/allennlp/data/dataset_readers/language_modeling.py
+++ b/allennlp/data/dataset_readers/language_modeling.py
@@ -96,24 +96,9 @@ class LanguageModelingReader(DatasetReader):
 
     @classmethod
     def from_params(cls, params: Params) -> 'LanguageModelingReader':
-        """
-        Parameters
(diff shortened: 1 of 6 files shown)


matt-gardner:
I think some kind of `TokenIndexer.something` is preferred, just so it's obvious to the caller what class is taking care of constructing the dependency.  No strong preference between what's here and `TokenIndexers.from_params`, except this is a bit simpler.  @DeNeutoy, any preference?

PR #259, opened 2017-09-04
Upgrade to python 3.6

### LINKED ISSUE #254: Upgrade to python 3.6
This will let us use variable type annotations, and remove all of the `unused-imports` in the code.

I don't _think_ there are any big issues with just changing the python version in our images and build settings, so I'm labeling this as easy, but it's possible there is some library we're using that's not compatible and it will end up being hard.

matt-gardner:
This lets us put type annotations on variables directly, instead of in comments, so we can get rid of all of the places we had to disable pylint's `unused-imports` check.

Fixes #254.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,8 +34,8 @@ RUN echo 'export PATH=/opt/conda/bin:$PATH' > /etc/profile.d/conda.sh && \
     /bin/bash ~/miniconda.sh -b -p /opt/conda && \
     rm ~/miniconda.sh
 
-# Use python 3.5
-RUN conda install python=3.5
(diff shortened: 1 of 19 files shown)


matt-gardner:
@schmmd - not sure why the docker cloud stuff is failing.  Installing the same package using travis works just fine.

schmmd:
My guess presently is that it's due to intermittent request failures for downloading dependencies.  I tried building the Docker image locally and it worked, so I restarted the build in Docker Cloud.

schmmd:
Seems like I guessed well.  PyTorch in particular has intermittent issues downloading.

PR #260, opened 2017-09-04
unify how Vocabs are created

### LINKED ISSUE #133: Figure out the right way to instantiate Vocabulary `from_params`
We need this so we can configure the vocabulary properly from an experiment config.  We need a method that takes a `Dataset` and a `Params` and instantiates the object.  Not sure what the most sane way is to do this - should we just only allow certain ways of constructing the vocab?

DeNeutoy:
Fixes #133

### DIFF
--- a/allennlp/data/vocabulary.py
+++ b/allennlp/data/vocabulary.py
@@ -288,10 +288,13 @@ class Vocabulary:
                           non_padded_namespaces=non_padded_namespaces)
 
     @classmethod
-    def from_params(cls, params: Params, dataset: "Dataset" = None):
+    def from_params(cls, params: Params, dataset=None):
         """
-        There are two possible ways to build a vocabulary: using
-        :func:`Vocabulary.from_dataset`, or
+        There are two possible ways to build a vocabulary from a
+        pre-existing dataset, using :func:`Vocabulary.from_dataset`, or
+        from a pre-saved vocabulary, using :func:`Vocabulary.from_files`.
+        This method wraps both of these options, allowing their specification
+        from a ``Params`` object, generated from a JSON configuration file.
 
         Parameters
         ----------
@@ -307,15 +310,15 @@ class Vocabulary:
         vocabulary_directory = params.pop("vocabulary_directory", None)
         if not vocabulary_directory and not dataset:
             raise ConfigurationError("You must provide either a Params object containing a "
-                                     "vocab directory path or a Dataset to build a vocabulary from.")
+                                     "vocab_directory key or a Dataset to build a vocabulary from.")
         if vocabulary_directory and dataset:
-            logger.warn("from_params was passed both a vocabulary_directory and a Dataset. This behaviour "
-                        "is undefined, so backing off to building the Vocabulary from the Dataset.")
+            logger.warning("from_params was passed both a vocabulary_directory and a Dataset. This "
+                           "behaviour is not well defined, so backing off to using the vocabulary "
+                           "defined at %s.", vocabulary_directory)
 
-        if vocabulary_directory and not dataset:
+        if vocabulary_directory:
             params.assert_empty("Vocabulary - from files")
             return Vocabulary.from_files(vocabulary_directory)
-
         return Vocabulary.from_dataset(dataset, **params.as_dict())
 
     def add_token_to_namespace(self, token: str, namespace: str = 'tokens') -> int:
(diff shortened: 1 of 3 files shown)


allennlp/data/vocabulary.py
```diff
@@ -285,6 +287,40 @@ def from_dataset(cls,
...
+            raise ConfigurationError("You must provide either a Params object containing a "
+                                     "vocab_directory key or a Dataset to build a vocabulary from.")
+        if vocabulary_directory and dataset:
+            logger.warning("from_params was passed both a vocabulary_directory and a Dataset. This "
+                           "behaviour is not well defined, so backing off to using the vocabulary "
+                           "defined at %s.", vocabulary_directory)
+
+        if vocabulary_directory:
+            params.assert_empty("Vocabulary - from files")
+            return Vocabulary.from_files(vocabulary_directory)
```
matt-gardner: I don't think there's a reason to create this `from_files` here - you can just call `Vocabulary.from_files` if that's what you want.

allennlp/data/vocabulary.py
```diff
@@ -285,6 +287,40 @@ def from_dataset(cls,
...
+        return Vocabulary.from_dataset(dataset, **params.as_dict())
```
matt-gardner: We've typically popped things off explicitly, except when passing the parameters off to pytorch functions, where we're wrapping things with different APIs and _can't_ explicitly pop off parameters.  We give better error messages when things are popped off, and you can call `params.assert_empty()`.

allennlp/data/vocabulary.py
DeNeutoy: I don't understand? This is required to be able to specify a pre-made vocab when using `allennlp/run.py train`, because otherwise we have to hard code in `Vocabulary.from_dataset` in `train`.

matt-gardner: Do we need to specify a pre-made vocab?

DeNeutoy: I'd imagine this would be how we do inference with a different set of glove vectors - use a different pre-specified vocab, and then load pretrained embeddings given this new vocabulary. We can't do that at the moment because of the way we disregard pretrained embeddings when we load models, but if that was a little more sophisticated, I could easily see this being how it works. Also, having this would speed things up - write a vocabulary once for the first run and then specify it in all other runs.

matt-gardner: If the goal is to make this work with different glove vectors, I think it'd be smarter to wait on this change until we have a whole solution for doing that - it's likely that actually trying to get that use case to work will want a different solution than what you have here.

Also, I think we can come up with a better way to handle caching pre-processed inputs that doesn't require modifying parameter files.  Maybe...

...

Ok, after thinking about this some more, I think it's fine to do it this way.  It seems reasonable to want to do this.  The logic in here is a little more complicated than I'd like, but I don't have a better option.

allennlp/data/vocabulary.py
```diff
@@ -285,6 +287,48 @@ def from_dataset(cls,
...
+        params: Params, required.
+        dataset: Dataset, optional.
+            If ``params`` doesn't contain a ``vocabulary_directory`` key,
+            the ``Vocabulary`` can be built directly from a ``Dataset``.
+
+        Returns
+        -------
+        A ``Vocabulary``.
+        """
+        vocabulary_directory = params.pop("vocabulary_directory", None)
```
matt-gardner: Remember, this will be under `{"vocabulary": {}}`, so you don't need to say "vocabulary" again.  I'd have this key be something like `from_directory`, or `from_files` instead.  It makes it a little more obvious what's going to happen when you give that key, too.

allennlp/data/vocabulary.py
```diff
@@ -285,6 +287,48 @@ def from_dataset(cls,
...
+        if not vocabulary_directory and not dataset:
+            raise ConfigurationError("You must provide either a Params object containing a "
+                                     "vocab_directory key or a Dataset to build a vocabulary from.")
+        if vocabulary_directory and dataset:
+            logger.warning("from_params was passed both a vocabulary_directory and a Dataset. This "
```
matt-gardner: This will happen every time you pass this flag during training, because the training code passes the dataset.  I'd instead have the warning say something like "Constructing vocabulary from files instead of from the dataset", and doing it as `logger.info` should be fine.

allennlp/data/vocabulary.py
DeNeutoy: sure.

PR #263, opened 2017-09-05
Demo UI - Reintroducing working state to submit button

aaronsarnat:
This PR:

- Disables the Run buttons when global `outputState` is set to `working`
- Removes an unused const (junk DNA from when i was building the UI)


aaronsarnat:
@schmmd this fixes the issue where the Run button isn't disabling when the app is in a `working` state. It was a simple fix. Just needed to pass the `outputState` prop down to the model input components. Everything else was already wired up.

PR #265, opened 2017-09-05
use broadcasting in viterbi

DeNeutoy:
Not sure how I missed this before...


allennlp/nn/util.py
```diff
@@ -220,9 +220,8 @@ def viterbi_decode(tag_sequence: torch.Tensor, transition_matrix: torch.Tensor):
 
     # Evaluate the scores for all possible paths.
     for timestep in range(1, sequence_length):
-        # TODO(Mark): Use broadcasting here once Pytorch 0.2 is released.
-        tiled_path_scores = path_scores[timestep - 1].expand_as(transition_matrix).transpose(0, 1)
```
matt-gardner: You left out the transpose - do you still need that?  I'm guessing not, because you're doing `unsqueeze(-1)` instead of `unsqueeze(0)`, just checking to be sure.

DeNeutoy: Yup this is the correct way around, thanks for checking though.

PR #268, opened 2017-09-05
api docs for the modules class

joelgrus:
there's a small amount of duplication, the docstring for the `text_field_embedders` module is basically a snippet from the docstring of `TextFieldEmbedder`, couldn't think of anything better to do. other than that should be pretty straightforward.

### DIFF
--- a/allennlp/modules/text_field_embedders/__init__.py
+++ b/allennlp/modules/text_field_embedders/__init__.py
@@ -1,7 +1,6 @@
 """
 A :class:`~allennlp.modules.text_field_embedders.text_field_embedder.TextFieldEmbedder`
-is a ``Module`` that takes as input the
-``dict`` of NumPy arrays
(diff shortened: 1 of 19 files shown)


allennlp/modules/text_field_embedders/__init__.py
```diff
@@ -1,2 +1,10 @@
+"""
+A :class:`~allennlp.modules.text_field_embedders.text_field_embedder.TextFieldEmbedder`
+is a ``Module`` that takes as input the
```
DeNeutoy: weird line breaks here?

joelgrus: fixed

joelgrus:
this is it for the API docs (for launch, anyway), mostly all I have left is cleaning up and arranging the tutorials, and I think I have that covered. you can write more tutorials if you want 😃

PR #270, opened 2017-09-05
Add some documentation around releasing.

schmmd:


schmmd:
@joelgrus take a look? I'd love your input.

RELEASE.md
```diff
@@ -0,0 +1,36 @@
...
+## Updating the web demo
+
+The web demo is deployed by sending a docker image to our kubernetes cluster.  When creating a new release, you should try the application locally before deploying it.
+
+```
+docker pull allennlp/allennlp:latest
+docker run -p 8000:8000 -it allennlp/allennlp:latest allennlp/run serve
+```
+
+After testing, you can cut a new release of the web demo.  Versioning for the web demo is separate for the library, because typically we will have many more releases for our web demo than our library.  Web demo releases are annotated by date in `yyyy.mm.dd-i` format, where i is a release number that starts at 0 and goes up for each release on a given day.
```
joelgrus: the `-i` looks weird to me, is that a docker convention? `yyyy.mm.dd.i` feels more natural?

schmmd: @joelgrus no it is not a Docker convention.  I'm happy to use another convention.  How about `yyyy-mm-dd-i`.  Or can you think of something that makes more sense than dates?

joelgrus: dates is fine, mostly I just don't like the mix of dots and dashes

PR #274, opened 2017-09-05
Add GA tag to web demo.

schmmd:


matt-gardner:
How will this behave if someone downloads our code and runs the demo?

schmmd:
I don't think it will aggregate the data after some basic testing.  I do register the URL for which to track using GA itself.  But I've already asked @codeviking the very same question, who certainly knows a whole lot more than I do.

schmmd:
I added a conditional to check the hostname before sending data to GA.

PR #275, opened 2017-09-06
Get rid of evaluation_json_file in model 

OyvindTafjord:
This adds `answer_texts` to the metadata for SQuAD data so we can calculate official EM/F1 without the separate `evaluation_json_file` parameter. This has the added benefit of fixing `allennlp/run evaluate` on other validation sets (modulo vocabulary expansion which is a separate matter), as well as providing EM/F1 scores on training data as well. There was no noticeable slowdown when I checked a few epochs on a 10% sample.

Note that I left in a `params.pop('evaluation_json_file', None)` on the model loading for now, so we don't immediately invalidate the model files we have out there (for the demo etc).

Also tacked on here is the removal of `--` in the beaker script which is necessary for the latest version of beaker.


allennlp/data/dataset_readers/squad.py
```diff
@@ -197,6 +200,8 @@ def text_to_instance(self,  # type: ignore
                 }
         if question_id:
             metadata['question_id'] = question_id
+        if answer_texts:
+            metadata['answer_texts'] = answer_texts
```
matt-gardner: Can you add this to the docstring above, where we say what metadata is saved?

allennlp/data/dataset_readers/squad.py
```diff
@@ -197,6 +200,8 @@ def text_to_instance(self,  # type: ignore
                 }
         if question_id:
```
matt-gardner: Looks like we can remove the question id now.

allennlp/models/bidaf.py
```diff
@@ -350,7 +329,7 @@ def from_params(cls, vocab: Vocabulary, params: Params) -> 'BidirectionalAttenti
         span_end_encoder = Seq2SeqEncoder.from_params(params.pop("span_end_encoder"))
         initializer = InitializerApplicator.from_params(params.pop("initializer", []))
         dropout = params.pop('dropout', 0.2)
-        evaluation_json_file = params.pop('evaluation_json_file', None)
+        params.pop('evaluation_json_file', None)  # Temporary, for backwards compatibility
```
matt-gardner: Maybe assign this to something and give a deprecation warning if it's not `None`?  And add a TODO to remove this once the saved models are updated?

allennlp/data/dataset_readers/squad.py
OyvindTafjord: I think it's nice to keep it there, e.g., I'm imagining using a version of this to dump individual scores per question into a file and then having easy access to the ID is handy.

allennlp/models/bidaf.py
OyvindTafjord: OK, do we have an existing pattern for deprecation warnings yet?

allennlp/data/dataset_readers/squad.py
matt-gardner: Ok, fair enough.

allennlp/models/bidaf.py
matt-gardner: No, we don't.

OyvindTafjord:
Found one earlier deprecation warning in `allennlp/run` so followed that pattern, not that it's easy to see that one line in the stream of stuff coming up when loading, but at least it's there. Added the docstring stuff.

matt-gardner:
Great.  And, sorry, I just merged a PR that will make you change this just a little bit.  I created a `MetadataField` that makes this metadata business a _little_ bit cleaner.

PR #276, opened 2017-09-06
Fix ordering of text field embedders

OyvindTafjord:
This fixes the order of `tokens` vs `token_characters` when concatenated. This used to rely on dictionary ordering which changed from Python 3.5 to 3.6. This will fix it to the order we had in 3.5 (`token_characters` before `tokens`) and should have the earlier models behaving well again.

FYI @schmmd @matt-gardner

### DIFF
--- a/allennlp/modules/text_field_embedders/basic_text_field_embedder.py
+++ b/allennlp/modules/text_field_embedders/basic_text_field_embedder.py
@@ -40,7 +40,9 @@ class BasicTextFieldEmbedder(TextFieldEmbedder):
                                                             str(text_field_input.keys()))
             raise ConfigurationError(message)
         embedded_representations = []
-        for key, tensor in text_field_input.items():
+        keys = sorted(text_field_input.keys())
+        for key in keys:
+            tensor = text_field_input[key]
             embedder = self._token_embedders[key]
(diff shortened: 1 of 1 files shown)


allennlp/modules/text_field_embedders/basic_text_field_embedder.py
```diff
@@ -40,7 +40,9 @@ def forward(self, text_field_input: Dict[str, torch.Tensor]) -> torch.Tensor:
                                                             str(text_field_input.keys()))
             raise ConfigurationError(message)
         embedded_representations = []
-        for key, tensor in text_field_input.items():
+        keys = sorted(text_field_input.keys())
```
joelgrus: why not just `for key, tensor in sorted(text_field_input.items())` ?

OyvindTafjord: Yes, that's probably also good, mine was a variant from earlier when we wanted to dictate a particular order (tokens before token_characters to match a different model). Maybe this is slightly clearer, that we want to order the keys in a particular order.

joelgrus: I'm cool either way

PR #277, opened 2017-09-06
add sniff test

joelgrus:
these slow down the build a non-trivial amount, but they would have caught the 3.5 -> 3.6 issue (and should catch other such issues in the future).

they should be robust in the sense that improving our models shouldn't break them but regressing our models should. but I'm open to feedback if that doesn't seem like the case


matt-gardner:
We could consider putting the pylint, mypy, and doc checks all into the same job, which would probably take about as long as the current (before this PR) job does.  Then you could add another job that just does these tests.

tests/models/sniff_test.py
```diff
@@ -0,0 +1,84 @@
+# pylint: disable=line-too-long,bad-whitespace,no-self-use
```
matt-gardner: Not super thrilled with the long lines or disabling bad whitespace.  I assume you just copy-pasted from the demo, or something?

tests/models/sniff_test.py
```diff
@@ -0,0 +1,84 @@
...
+    def test_config(self):
+        assert set(DEFAULT_MODELS.keys()) == {
+                'machine-comprehension',
+                'semantic-role-labeling',
+                'textual-entailment'
+        }
+
+
+    def test_machine_comprehension(self):
+        predictor = Predictor.from_archive(
```
matt-gardner: It might be worth modifying the [cache directories](https://github.com/allenai/allennlp/blob/e8a4a9c997a646618465ba42388fff7731ead34b/.travis.yml#L4-L9) in the Travis config to include our file cache dir.  It would speed this up a little.

tests/models/sniff_test.py
joelgrus: I can fix the whitespace, the long lines will be really hard, they're mostly just really long strings. I would have to really go out of my way to get them under 115

PR #291, opened 2017-09-11
Add DockerHub badge.

schmmd:


schmmd:
Any idea why we have 51 layers?  That seems like a lot, given our Dockerfile.

joelgrus:
maybe it's counting 29 from cuda ?

PR #293, opened 2017-09-11
refactor demo

joelgrus:
* move everything from `allennlp/service/static/` to `demo/`
* factor all the javascript out of the `index.html` into actual modules
* configure bundling / transpiling / minification so that it doesn't happen at runtime / happens at all
* cleaned up a bunch of js stuff that the linter complained about
* point `server_sanic` to the `demo/build` directory
* update the Dockerfile to install npm and build the demo
* update the getting started tutorial to reflect these changes

in the short term the biggest win is that the demo pages are much faster, as they're not using babel to transpile on the fly. in the long term this will also allow us to write unit tests for our js and will make it easier to make changes / additions to the demo.

the primary downside of this change is that now if you install from source you have to manually generate the demo files before you can `run serve`. I don't think this is a huge deal.

(pretty thrilled about a net -21k lines PR)


cendrillon:
> the primary downside of this change is that now if you install from source you have to manually generate the demo files before you can run serve

Could you please update the docs on how to do this? The demo seems broken now, at least following the documentation gives:

Starting a sanic server on port 8000.
2017-10-05 11:04:16,044 - ERROR - allennlp.service.server_sanic - app directory /Users/.../allennlp/allennlp/service/../../demo/build does not exist, aborting

PR #294, opened 2017-09-11
Use the runtime cuda images rather than devel.

schmmd:
This reduces our Docker image size from 3.7 GB to 2.9 GB.


schmmd:
I still need to run a test, to verify that we don't require the devel cuda images.

schmmd:
```
allennlp/run evaluate --archive_file https://s3-us-west-2.amazonaws.com/allennlp/models/bidaf-model-2017.08.31.tar.gz --evaluation_data_file dev-v1.1.json --cuda_device 0
```

The previous command runs just fine in a Docker image (`schmmd/allennlp:cuda-release-2`) built from this PR.  I'm pretty confident this PR is safe, but if someone has another GPU job they'd like me to try, please send it my way.

PR #297, opened 2017-09-12
log requests

joelgrus:


allennlp/service/server_sanic.py
```diff
@@ -73,6 +73,7 @@ def _caching_prediction(model: Predictor, data: str) -> JsonDict:
             raise ServerError("unknown model: {}".format(model_name), status_code=400)
 
         data = req.json
+        logger.info("model: %s inputs: %s", model_name, json.dumps(data))
```
schmmd: @joelgrus why not log the answer as well?

joelgrus: I thought about that, the main reason is that the answer is potentially very big (lots of arrays, etc...). secondary is that (in theory) the answer is a deterministic function of the request

schmmd: Good call re: the answer is potentially very big--but that just makes me think we need to figure out how to log the core part of the answer.  I'm happy to approve this PR, but we'll at least want an open ticket around logging the answer.

PR #305, opened 2017-09-13
Ignore empty lines in predict jsonl.

schmmd:


allennlp/commands/predict.py
```diff
@@ -51,14 +51,15 @@ def get_predictor(args: argparse.Namespace) -> Predictor:
...
     for line in input_file:
-        data = json.loads(line)
-        result = predictor.predict_json(data)
-        output = json.dumps(result)
-
-        if print_to_console:
-            print(output)
-        if output_file:
-            output_file.write(output + "\n")
+        if line.strip() != "":
```
joelgrus: a little bit more idiomatic would just be

```
if not line.isspace():
```

schmmd: That's much better.  I'll try it out.

PR #312, opened 2017-09-14
Add TokenIndexers for POS tags and dependency labels

### LINKED ISSUE #287: New TokenIndexer: POS tags
This is really common in parsing work, and in some SRL / other tagging.  Also really easy to add, once the tokenizer is reworked to return a token object (#286).  The corresponding `TokenEmbedder` is just an `Embedding` that uses a different vocab namespace.

matt-gardner:
Fixes #287.

### DIFF
--- a/allennlp/data/token_indexers/dep_label_indexer.py
+++ b/allennlp/data/token_indexers/dep_label_indexer.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Dict, List
+from typing import Dict, List, Set
 
 from overrides import overrides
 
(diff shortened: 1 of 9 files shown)


allennlp/data/token_indexers/dep_label_indexer.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    # pylint: disable=no-self-use
+    def __init__(self, namespace: str = 'dep_labels') -> None:
+        self.namespace = namespace
+
+    @overrides
+    def count_vocab_items(self, token: Token, counter: Dict[str, Dict[str, int]]):
+        if not token.dep_:
+            logger.warning("Token had no dependency label: %s", token.text)
+        counter[self.namespace][token.dep_] += 1
```
DeNeutoy: This will break the vocabulary creation if we count `None` elements, because when we add the token to the namespace, we typecheck that it is a string. I think there should be an `else` case here.

allennlp/data/token_indexers/pos_tag_indexer.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @overrides
+    def count_vocab_items(self, token: Token, counter: Dict[str, Dict[str, int]]):
+        if self._coarse_tags:
+            tag = token.pos_
+        else:
+            tag = token.tag_
+        if not tag:
+            logger.warning("Token had no POS tag: %s", token.text)
+        counter[self._namespace][tag] += 1
```
DeNeutoy: Same as above.

allennlp/data/token_indexers/dep_label_indexer.py
DeNeutoy: Also if this can happen, we should throw an error in case the user tries to use a non-padded namespace. Hmmm. Although that would make it impossible to use for prediction. Maybe not then.

matt-gardner: Yeah, good catch on the `None`.  For the padded namespace issue, this will actually be non-padded by default ("pos_tags" ends in "tags").  This is probably ok, because presumably you have a small POS tag set (or dependency label set), and you should see all of them at training time.

PR #313, opened 2017-09-14
[Not for merging] Custom LSTMs

DeNeutoy:
I am going to split this up into 3-ish PRs and there is still a fair bit of work to do, but I wanted some opinions before I started:

This PR has three parts:
- Removing the unnecessary packing and padding from the `AugmentedLSTM`. This had the side effect of having to remove the ability to use it directly, as a layer, as it doesn't support the `Seq2SeqEncoder` API without accepting `PackedSequences`. Turns out it didn't speed things up very much - do we still want to make this change? It's probably going to be superseded by other changes (like Nick's implementation), so maybe not. 
- Adding Nick's CUDA LSTM.
- Adding the SRU code. 

The main thing I wanted some advice on was how we should deal with these custom extensions. It's going to be pretty tricky to keep the CUDA code compatible with everything (I already spent about 2 hours yesterday trying to get Nick's kernels to compile for different GPU architectures) and inevitably people using the library will use them. However, it does mean that we should know apriori which GPUs the custom stuff will work with. 

The SRU code uses a couple of python libraries (cupy and pynvrtc) which wrap the CUDA runtime compilation library, which means we don't have to generate any binaries. However, it does mean we have C code in a python string which is compiled at runtime, which is kind of scary. However it did work, so :man_shrugging:. I think the way Nick has implemented the wrapper to the kernel is more sustainable and I think I should be able to convert the SRU kernel to have the same style. However, i'm very open to opinion here.

I might ask one of the XNOR guys to see if they have any advice on this front later, as they probably have more experience.


matt-gardner:
No strong opinion on this stuff from me - CUDA code and C extensions to python are not something I've dealt with.  I would say if we're replacing the `AugmentedLstm`, I probably wouldn't worry about fixing it up.

PR #317, opened 2017-09-14
Letters+digits word splitter

OyvindTafjord:
A basic, unopinionated word splitter which keeps runs of (unicode) letters and runs of digits together, anything else is treated as single-character "words". Showed marginal improvement (compared to the default spacy English tokenizer) in BiDAF on SQuAD.


OyvindTafjord:
Added a couple of tests

PR #318, opened 2017-09-14
fix epoch bug

### LINKED ISSUE #292: Inconsistent epoch numbering
It's a bit confusing that we display `epoch + 1` in the logger during training, while `epoch` is the number saved in the `model_state_epoch_xx.th` files. More problematic is that when restoring from a checkpoint, the epoch gets set to the number of the loaded file, which then is promptly overwritten at the end of the first epoch of training, which seems like a bug.

DeNeutoy:
Fixes #292

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -321,7 +321,8 @@ class Trainer:
         Returns
         -------
         epoch: int
-            The epoch at which to resume training.
+            The epoch at which to resume training, which should be one after the epoch
+            in the saved training state.
         """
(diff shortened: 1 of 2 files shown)


OyvindTafjord:
That seems good, although the more serious issue (in my mind) mentioned in https://github.com/allenai/allennlp/issues/292 is when restoring from a checkpoint, the counter is set to the counter for the checkpoint, thus overwriting it. One fix to this is to change `epoch_counter = self._restore_checkpoint()` to `epoch_counter = self._restore_checkpoint() + 1`

Issue #323, opened 2017-09-15
Add a way to configure predictor name in `predict` and `serve` commands


rilut:
Hi, I created a new predictor and already registered it with `@Predictor.register`, but how do I select the predictor?

CMIIW, I see the `predictor_name` parameter on `Predictor.from_archive` but I think it never gets called from the configuration/argumentparser. I see it only falls back to `DEFAULT_PREDICTORS` if `None`.

Additional info:
- I see my predictor class doesn't have `_model` and `_dataset_reader`.
- I've created my own dataset reader successfully

Also, which is the best channel to ask a question? Thank you very much.

matt-gardner:
For now, asking questions here is fine.  We're thinking of recommending stack overflow or something similar for questions like this in the future, but we aren't set up to monitor those questions yet.

I need a little more context to answer this question, though.  How are you wanting to use the predictor you've written?  Are you running `python -m allennlp.run serve`?

If so, then yes, we're not set up yet to have this be configurable (FYI @joelgrus).  The easiest hacky workaround for now is to modify [this dictionary](https://github.com/allenai/allennlp/blob/e125a490b71b21e914af01e70e9b00b165d64dcd/allennlp/service/predictors/predictor.py#L8-L13) in a script, then call the serve command yourself in a script.

joelgrus:
this might answer your question: https://github.com/allenai/allennlp/blob/master/tutorials/getting_started/configuration.md

Because a class doesn't get registered until it's loaded, any code that uses
`BaseClass.by_name('subclass_name')` must have already imported the code for `Subclass`.
In particular, this means that once you start creating your own named models and helper classes,
the included `allennlp.run` command will not be aware of them. However, `allennlp.run` is simply
a wrapper around the `allennlp.commands.main` function,
which means you just need to create your own script
that imports all of your custom classes and then calls `allennlp.commands.main()`.

matt-gardner:
@joelgrus, the issue is that even if you do import it correctly, you still can't use our server code, because specifying the predictor name doesn't actually happen anywhere in our code path: 

https://github.com/allenai/allennlp/blob/e125a490b71b21e914af01e70e9b00b165d64dcd/allennlp/service/server_sanic.py#L34

https://github.com/allenai/allennlp/blob/e125a490b71b21e914af01e70e9b00b165d64dcd/allennlp/service/predictors/predictor.py#L51

[...]

PR #329, opened 2017-09-15
make default models and default predictors configurable

joelgrus:
this addresses https://github.com/allenai/allennlp/issues/323 and is a better design anyway

here `DEFAULT_MODELS` and `DEFAULT_PREDICTORS` are defined with `allennlp.commands.main()` which now takes optional dicts of overrides.

after this change, if someone implements their own predictors, they can just specify them when they call `main()`

still needs tests


joelgrus:
hold off please, this is broken

joelgrus:
ok, should be fixed and ready to look at now

PR #331, opened 2017-09-15
Update BIDAF to use Oyvind's best model.

schmmd:
Our current BIDAF model underperforms the statistics we have on our site.  This model does slightly worse than our stated scores (68.3 vs 68.7), however it outperforms the scores reported in the BIDAF paper.

em: 0.6833491012298959
f1: 0.7780870748211246


schmmd:
When we merge this, I'll update the scores reported on the website and I will also get test scores on SQuAD.

PR #334, opened 2017-09-18
Custom extensions

DeNeutoy:
This is the first step in getting Nick's LSTM into allennlp. At the moment, it functions well if you have a GPU - however, wrapping it such that it is seamless across devices is going to require a little effort - [this rnn backend of pytorch](https://github.com/pytorch/pytorch/blob/master/torch/backends/cudnn/rnn.py) is dedicated to making this work properly and it looks non-trival. There's also naturally not a huge amount of documentation as it's meant to be a backend.

One question - should I actually add the compiled extension, or leave it? It's probably easier to have it in the repo, but I wasn't sure.


DeNeutoy:
Codecov looks bad but travis won't run the test included in this PR, because it requires GPUs.

joelgrus:
seems fine to me

soumith:
The RNN backend code is a bit overtly complicated, dont use that as a reference. It's one of those things that organically grew into that state.

Usually, if you wrote a cffi extension, just calling that with our `device` or `device_of` contexts should be sufficient:

```python
with torch.cuda.device_of(weight):
    backend.C.my_lstm_kernel(....)
```

Do let us know if you have questions in this direction.

PR #335, opened 2017-09-19
Move combine_tensor methods into a common location

matt-gardner:
So that I can use them in other places (particularly in a multi-headed attention encoder).

I also added an option for scaling the dot product similarity.

### DIFF
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -2,7 +2,7 @@
 Assorted utilities for working with neural networks in AllenNLP.
 """
 
-from typing import Dict, Optional, Union
+from typing import Dict, List, Optional, Union
 
 import numpy
 import torch
@@ -389,6 +389,7 @@ def replace_masked_values(tensor: Variable, mask: Variable, replace_with: float)
     values_to_add = replace_with * one_minus_mask
     return tensor * mask + values_to_add
 
+
 def device_mapping(cuda_device: int):
     """
     In order to `torch.load()` a GPU-trained model onto a CPU (or specific GPU),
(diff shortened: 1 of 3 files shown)


allennlp/nn/util.py
```diff
@@ -402,9 +403,99 @@ def inner_device_mapping(storage: torch.Storage, location) -> torch.Storage:  #
...
+
+    This function also accepts ``x`` and ``y`` in place of ``1`` and ``2`` in the combination
+    string.
+    """
+    combination = combination.replace('x', '1').replace('y', '2')
+    combinations = combination.split(',')
+    combined_tensor = _get_combination(combinations[0], tensors)
+    for piece in combinations[1:]:
+        to_concatenate = _get_combination(piece, tensors)
+        combined_tensor = torch.cat([combined_tensor, to_concatenate], dim=-1)
```
joelgrus: is it inefficient to concatenate tensors one at a time rather than all at once?  (I don't know)

allennlp/nn/util.py
```diff
@@ -402,9 +403,99 @@ def inner_device_mapping(storage: torch.Storage, location) -> torch.Storage:  #
...
+    combination = combination.replace('x', '1').replace('y', '2')
+    return sum([_get_combination_dim(piece, tensor_dims) for piece in combination.split(',')])
+
+
+def _get_combination_dim(combination: str, tensor_dims: List[int]) -> int:
+    if combination.isdigit():
+        index = int(combination) - 1
+        return tensor_dims[index]
+    else:
+        if len(combination) != 3:
```
joelgrus: this seems to assume only single digits, maybe that's a good assumption in practice, but it seems dangerous to rely on it

allennlp/nn/util.py
matt-gardner: Yeah, that's a good point.  I don't know why I did it this way originally...  I'll fix it.

allennlp/nn/util.py
matt-gardner: Concatenating ten or more tensors at a time seems pretty implausible.  I'll raise an error if you try to use it with more than is supported - that's easier than making the logic robust to multiple digits, and we'll almost certainly never need more than 9.

Issue #338, opened 2017-09-20
Vocabulary token roundtrip bug


OyvindTafjord:
On a new dataset, I ran into the tokens `'\x0b'` (vertical tab) and, weirdly, `'\x0b '` (with a space behind) which don't roundtrip well when read back from `tokens.txt` (each get read from file as two separate tokens, messing up the indexing).

matt-gardner:
I think we could handle this similarly to how we handle newlines: https://github.com/allenai/allennlp/blob/c6bc7fde370e54312d9fabdeec66818240b6e783/allennlp/data/vocabulary.py#L261

We'd probably want to generalize this into some `sanitize` and `desanitize` methods, with a list of characters to replace and things to replace them with.  Seems pretty easy.

OyvindTafjord:
FWIW, here's the list of character codes (<= 65535) which don't roundtrip properly with the current code: `11, 12, 28, 29, 30, 133, 8232, 8233` (in addition to the `10` already "handled"). Kinda seems like an unexpected "feature" of `.readlines()` (or upstream `.splitlines()`) that this happens?
```
>>> "foo\nbar\x0bxx\x0cyy".splitlines()
['foo', 'bar', 'xx', 'yy']
```

OyvindTafjord:
Simpler fix might be to just read the whole file and do `.split('\n')` only.

matt-peters:
How are these tokens with whitespace passing through the tokenizer anyway?  Most tokenizers will remove them.

OyvindTafjord:
Yes, spacy seems to have some interesting interpretations of these (I'm particularly intrigued by the '\x0b ' one with a space in it!):
```
>>> import spacy
>>> spacy_model = spacy.load("en")
>>> [x.text for x in list(spacy_model("foo bar aa\x0bbb cc\x0b dd"))]
['foo', 'bar', 'aa', '\x0b', 'bb', 'cc', '\x0b ', 'dd']
```

matt-peters:
Ah yes -- this is a hidden "feature" of spacy.  In order to reconstruct the character offsets (I think) it will include tokens that are all spaces.  There is an attribute on the token object `is_space` I usually use to eliminate them.

For example:
```
tokens = [token.text for token in spacy_model(text) if not token.is_space]
```

OyvindTafjord:
[...]

PR #344, opened 2017-09-20
Fix misreading of newline-like vocab tokens

OyvindTafjord:
As discussed in https://github.com/allenai/allennlp/issues/338, simple fix for bug in reading newline-like tokens.


OyvindTafjord:
Per discussion in the issue, I added a simple filter to the space word splitter, to remove whitespace tokens (we still want to be robust wrt these appearing in vocab files though).

PR #346, opened 2017-09-21
REFERENCE: Bidaf + Self Attention

chrisc36:
Reimplementation of our refined BiDAF model w/self attention. Currently the best I was able to get on SQuAD was around 79.4 F1 as seen on Tensorboard. The original implementation reached 80.7 F1 on the dev set.

Some known differences:
- My attempt using Adadelta failed badly, it actually started doing worse on the train set over time. Adadelta was fair bit better in my version so that might be important
- No exponential moving averages
- No length constraints when choosing the best span

The self-attention part causes OOMs, so I implemented a more bidaf-specific memory efficient version.

(This is still needs more work, just creating the pull request for reference.)

### DIFF
--- a/allennlp/models/bidaf_self_attention.py
+++ b/allennlp/models/bidaf_self_attention.py
@@ -1,0 +1,303 @@
+import logging
+from typing import Any, Dict, List
+
+import torch
+import torch.nn.functional as F
+from torch.autograd import Variable
+from torch.nn.functional import nll_loss
+
+from allennlp.common import Params, squad_eval
+from allennlp.data import Vocabulary
+from allennlp.models.model import Model
+from allennlp.modules import Seq2SeqEncoder, TimeDistributed, TextFieldEmbedder
+from allennlp.modules.tri_linear_attention import TriLinearAttention
+from allennlp.modules.variational_dropout import VariationalDropout
+from allennlp.nn import InitializerApplicator, util
+from allennlp.training.metrics import Average, BooleanAccuracy, CategoricalAccuracy
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+@Model.register("bidaf-self-atten")
+class BidafPlusSelfAttention(Model):
+
+    def __init__(self, vocab: Vocabulary,
+                 text_field_embedder: TextFieldEmbedder,
+                 phrase_layer: Seq2SeqEncoder,
+                 residual_encoder: Seq2SeqEncoder,
+                 span_start_encoder: Seq2SeqEncoder,
+                 span_end_encoder: Seq2SeqEncoder,
+                 initializer: InitializerApplicator,
+                 dropout: float = 0.2,
+                 mask_lstms: bool = True) -> None:
+        super(BidafPlusSelfAttention, self).__init__(vocab)
+
+        self._text_field_embedder = text_field_embedder
+        self._phrase_layer = phrase_layer
+        self._matrix_attention = TriLinearAttention(200)
+
+        self._merge_atten = TimeDistributed(torch.nn.Linear(200 * 4, 200))
+
+        self._residual_encoder = residual_encoder
+        self._self_atten = TriLinearAttention(200)
+        self._merge_self_atten = torch.nn.Linear(200 * 3, 200)
+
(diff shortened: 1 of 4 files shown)


valdersoul:
Hi any update here, or is there any chance to have the original implementation available?

matt-gardner:
I have been planning on getting this merged, but keep having other things come up.  We're writing a paper that includes this work, for instance, and that's higher priority.  This doesn't need that much work, though, so when we finally get time to look at this, it should be quick.

matt-gardner:
The original implementation (in tensorflow) will likely also be released, once the paper is submitted somewhere.  That repo is currently private.

valdersoul:
Thanks for your reply. Good luck to your paper!

matt-gardner:
And @chrisc36, I was assuming I'd be the one to clean this up, which is why I haven't really done anything with it yet (and we need a bunch more to get this working for TriviaQA, and the open setting, etc., and I was working on that...).  If I'm assuming wrong there, and you want to do it, let me know, and I'll review this.

guotong1988:
Does this code improve the results?

allennlp/modules/tri_linear_attention.py
```diff
@@ -0,0 +1,56 @@
...
+
+        weighted_x = matrix_1 * self._dot_weights  # still (batch, time1, dim)
+
+        matrix_2_t = torch.transpose(matrix_2, 1, 2)  # -> (batch, dim, time2)
+
+        # Batch multiplication,
+        # (batch, time1, dim), (batch, dim, time2) -> (batch, time1, time2)
+        dot_factors = torch.matmul(weighted_x, matrix_2_t)
+
+        # broadcasting will correctly repeat the x/y factors as needed
```
guotong1988: `# broadcasting will correctly repeat the x/y factors as needed`
What does this line mean？

matt-gardner: @guotong1988, reading this is probably a good start: https://docs.scipy.org/doc/numpy-1.13.0/user/basics.broadcasting.html

guotong1988:
@chrisc36 `The original implementation` means the code of original bidaf without self-attention?

chrisc36:
Nope, the original implementation is the same thing, but in tensorflow. It just became public:

github.com/allenai/document-qa

murphp15:
@chrisc36  I would love to help you with this PR. 
I can write tests for it. 
Do you have suggestions of what kind of tests should be written?

experiment_config/bidaf_self_atten.json
```diff
@@ -0,0 +1,99 @@
...
+        "type": "characters",
+        "character_tokenizer": {
+          "byte_encoding": "utf-8",
+          "start_tokens": [259],
+          "end_tokens": [260]
+        }
+      }
+    }
+  },
+  "train_data_path": "/home/chrisc/data/squad/train-v1.1.json",
```
murphp15: Would you be able to check these files into git and I will start writing tests.

allennlp/models/bidaf_self_attention.py
```diff
@@ -0,0 +1,303 @@
...
+        final_merged_passage = F.relu(self._merge_atten(final_merged_passage))
+
+        residual_layer = self._dropout(self._residual_encoder(self._dropout(final_merged_passage), passage_mask))
+        self_atten_matrix = self._self_atten(residual_layer, residual_layer)
+
+        mask = passage_mask.resize(batch_size, passage_length, 1) * passage_mask.resize(batch_size, 1, passage_length)
+
+        # torch.eye does not have a gpu implementation, so we are forced to use the cpu one and .cuda()
+        # Not sure if this matters for performance
+        self_mask = Variable(torch.eye(passage_length, passage_length).cuda()).resize(1, passage_length, passage_length)
```
murphp15: This will not work on machines without gpus, which is annoying for developers without gpus on their local machine.

murphp15:
Cool, best of luck on the paper! I'm going to sync down #804 and try train it etc.

schmmd:
I'm closing this because it's been un-active for a few months.  It will still be available for reference, just not in our list of open PRs.

PR #347, opened 2017-09-22
implement regularization

joelgrus:
things of note:

* moved `allennlp.training.regularizers` to `allennlp.nn.regularizers` to avoid a circular imports problem
* changed `RegularizerApplicator` to take a list of pairs (like `InitializerApplicator` does, previously it took a dict)
* added regularization + initialization methods on the base model class (lmk if you don't like the names of these)
* took all the `InitializerApplicator`s out of the model subclass constructors, and instantiated them in `Model.from_params`
* similarly, instantiated the `RegularizerApplicator` in `Model.from_params`
* modified `trainer.Train` to apply the regularization penalty

I added some simple tests for the regularization. As discussed on slack, the more end-to-end tests will have to wait until after we refactor the `Trainer`. I added a TODO for those.


allennlp/models/model.py
```diff
@@ -147,7 +182,18 @@ def get_metrics(self, reset: bool = False) -> Dict[str, float]:
     @classmethod
     def from_params(cls, vocab: Vocabulary, params: Params) -> 'Model':
         choice = params.pop_choice("type", cls.list_available())
-        return cls.by_name(choice).from_params(vocab, params)
+        regularizer_params = params.pop("regularizer", None)
+        initializer_params = params.pop("initializer", None)
+
+        model = cls.by_name(choice).from_params(vocab, params)
+
+        if regularizer_params is not None:
```
DeNeutoy: I'm very unsure that this is a good solution. This makes building a model `from_params` substantially different to building one using the constructor. As soon as this happens once, we will keep being tempted to add things that are useful for our specific training regime to the `from_params` constuctor, but make it effectively impossible to use a model in allennlp directly. 

I would vote every time for adding 4 lines to a new model constructor (two parameters, one to call super and one to initialize the model), which is explicit in how it works and explicit in how it breaks if you don't use it correctly. E.g we could log a warning in the base model constructor if you don't pass an initializer.

joelgrus: yeah, I think I am coming around to this point of view

allennlp/models/model.py
```diff
@@ -147,7 +182,18 @@ def get_metrics(self, reset: bool = False) -> Dict[str, float]:
...
+            model.add_regularizer(RegularizerApplicator.from_params(regularizer_params))
+        if initializer_params is not None:
+            model.add_initializer(InitializerApplicator.from_params(initializer_params))
+            model.initialize()
```
matt-gardner: I don't think this goes here.  The trouble is there's no good spot to put it.  I want it to be in the constructor, but it has to be called _after_ all submodules have been assigned to `self`, while pytorch requires that the super constructor be called _before_ any submodules have been assigned to `self`...

matt-gardner: Passing the initializer to the super class constructor doesn't solve the issue, either, because the modules wouldn't have been assigned...

joelgrus: we could put it in the trainer and call it if you're not starting from a checkpoint

allennlp/models/model.py
matt-gardner: I don't have strong opinions on this at this point; I can see pros and cons of doing it both ways.  I'd like to avoid required boilerplate, though.  So, here's a proposed solution that hopefully will make Mark happy:

Because of how regularization interacts with the `Trainer`, the superclass _has_ to know about it, so it needs to get it somehow.  I'm fine with passing it explicitly into the constructor, and having it be `None` by default, because it's not typical for models to use regularization.  A subclass can just totally ignore regularization, and that's fine, so there's no required extra boilerplate.  You just need to add some code if you want to try out regularization.  Not a huge deal.

Also, we should try to make the default initialization of everything we use reasonable, so that it's not typical to need an initializer.  Then most models can just ignore initialization also, and not have anything extra needed there.  I know I recommended putting the initialization in the super class, but seeing that it requires you to actually _run_ the initializer outside of the constructor makes me think that was a bad suggestion.  Here again we can just put this back how it was, with the model adding the ability to do initialization if it wants.  It's not that much boilerplate, and it's not required for a new model.

allennlp/models/semantic_role_labeler.py
```diff
@@ -209,13 +214,18 @@ def from_params(cls, vocab: Vocabulary, params: Params) -> 'SemanticRoleLabeler'
         text_field_embedder = TextFieldEmbedder.from_params(vocab, embedder_params)
         stacked_encoder = Seq2SeqEncoder.from_params(params.pop("stacked_encoder"))
         binary_feature_dim = params.pop("binary_feature_dim")
-        initializer = InitializerApplicator.from_params(params.pop("initializer", []))
```
DeNeutoy: Can we leave the initializer being initialized to empty rather than it being `None`? This way, it will log the default initializations of all the modules, which is important so you can notice if they are inappropriate.

DeNeutoy: This can be in the constructor actually : `initializer: InitializerApplicator = InitializerApplicator(None)`

joelgrus: done

joelgrus:
ok, I think this is ready for another look. high points

* initializer can't be None, default value is an empty InitializerApplicator
* base `Model` knows nothing about initialization
* instead every (current) `Model` subclass explicitly calls `initializer(self)` at the end of its constructor

* `model.get_regularization_penalty()` returns 0 (rather than None) when there's no regularization
* I ran into a lot of challenges about using variables where pytorch wanted tensors and vice versa, I think I handled it correctly, but this is something I keep running into trouble with
* I added a couple of tests to make sure that regularization and training are doing the right things

allennlp/models/model.py
```diff
@@ -49,13 +49,13 @@ def __init__(self,
...
+    def get_regularization_penalty(self) -> torch.autograd.Variable:
         """
-        Get the regularization penalty for the model. Will be ``None``
-        if there's no regularizer.
+        Computes the regularization penalty for the model.
+        Returns a zero tensor if the model was not configured to use regularization.
         """
         if self._regularizer is None:
-            return None
+            return torch.autograd.Variable(torch.Tensor([0.]))
```
DeNeutoy: This breaks if the model is on the GPU. I think you can just return `0.0` instead?

joelgrus: done

PR #349, opened 2017-09-22
Additional requirements needed for passing pytest

pdasigi:


matt-gardner:
These are in `requirements_test.txt`.

pdasigi:
Oh, you are right! I did not realize that because my shell (Fish) did not recognize "true" (with quotes) in INSTALL_TEST_REQUIREMENTS="true", and did not install test requirements. Closing this.

PR #350, opened 2017-09-22
[WIP] Encoder-decoder model

pdasigi:


matt-gardner:
I think this should be a new `Field`.  A labeling on an existing field is conceptually very different from a self-standing sequence of labels.  Yes, in implementation they are similar, but they are conceptually different enough that we should have separate objects for them.

pdasigi:
I see your point. Giving them both the same name could be confusing. Would it make sense to have a superclass that both these `Field`s inherit from? May be the parent class could be called `SequenceLabelField` and the two subclasses `TagSequenceField` and `DecoderSequenceField`?

matt-gardner:
I don't think we need the hierarchy, just a separate class with similar code should be fine.  Not sure what to call it, though - @DeNeutoy, any thoughts?  This is for decoder outputs in a seq2seq model.  It's a sequence of strings of arbitrary length.  Also not totally sure how to handle padding in the decoder - you just do it like everything else, and make sure the loss doesn't include padding output tokens?

pdasigi:
I suggested a hierarchy mainly because most of the implementation can be
shared.

matt-gardner:
I know, and it's tempting to share everything that you can.  The trouble is going down that route can lead you to bad places, like we ended up with the memory network implementation in deep_qa.  I think it's sensible to just have these classes be self-standing; there isn't that much code involved.  And it leaves us flexible in case we need to handle sequence labels differently from decoder outputs in the future.

DeNeutoy:
I think I agree with matt - we can re-name what we have as a `ConditionalSequenceLabelField` and have what you are suggesting as `SequenceLabelField`. 

Also, i'm not sure this is even required - there doesn't seem to be a reason why you can't use a `TextField` as a target? You'd have to add on `<END>` tokens in the tokeniser, but we already do that.

pdasigi:
Okay, I agree as well. But why not `TagSequenceField` for the current
class? And I'm thinking of decoded sequences that are not necessarily text.
They're grammar rules for what I'm trying to do now.

DeNeutoy:
Sure - can't you have a `TextField` with a `SingleIdTokenIndexer` which uses a namespace which is all the possible grammar rules? There's no requirement that the `TextField` tokens need to be text - it's more a way to support multiple different representations of the same token sequence, but you don't _have_ to use this feature.

pdasigi:
Hmm.. I think the argument made by Matt applies to using `TextField` for this as well. It might be cleaner to keep them separate, especially because `TextField` can be used in ways that could break the intended functionality here.

matt-gardner:
So, we've kind of run into this problem before: https://github.com/allenai/allennlp/blob/0b5a2ed6a226a70865d5e1ce045c91b46cf5b1b5/allennlp/data/dataset_readers/language_modeling.py#L55-L65

That's a little bit different, because it's language modeling, and there's one output for every input.  Maybe you could argue that we should have used a `SequenceLabelField` for that one, instead of a `TextField`, but I'm not sure...  I think a `TextField` isn't a bad use for this - it's an arbitrary-length sequence of strings.

Yeah, I think I'm with Mark here - using a `TextField` for the decoder output seems pretty reasonable.

pdasigi:
I see that a `TextField` takes a sequence of `Token`s. I'm still trying to understand this, but it looks like there is a semantic difference between tokens and labels, rightly so. For example, the label vocabulary does not need an OOV token, and it looks like `Vocabulary` is taking care of that based on the namespace's id.

We can definitely use `TextField` for output sequences when the output is clearly text (like summarization), but in the semantic parsing case, the output sequence does not need OOVs, and is more "label-like". Does that make sense? Alteast to make that distinction, it seems to me that we need a class that's different from `TextField` here.

DeNeutoy:
Ah, yes. That makes things more difficult - in this case, you want a padding token, but not an OOV token. That is annoying. Ok, maybe we do need the two types of `Field` then. @matt-gardner do you agree?

matt-gardner:
Whether the vocabulary has padding and OOV tokens or not is configurable, and thus really orthogonal to this question - just change the namespace of the token indexer to something that is not padded by default, or add a non-padded namespace.  Not being able to separate padding tokens from OOV tokens, and just have padding but no OOV, is indeed a little annoying, but also orthogonal to the question - that's handled by the `Vocabulary`, not by the `Field`, and you couldn't easily write _any_ `Field` code to make that happen; you'd need to change `Vocabulary` code.

So the only remaining objection is that `TextField` takes `Token` objects instead of strings.  Yeah, that's a little bit different, but I don't think it's a blocking issue.

What I'm saying now might seem a little inconsistent with what I said at the beginning of this PR.  Here's why I think these situations are different: in the first case, you were proposing _changing_ a `Field's` code to make it semantically ambiguous in a way that's pretty jarring to me.  In this case, I'm suggesting _using_ an existing `Field` to do something it was already designed to do, just in a constrained setting.  The caller has to be sure that the `Field` is constructed in a way that's consistent with its expectations, but other than that, the `Field` is just working as intended.  So this seems fine to me, and not really worth adding another abstraction that is just a restricted version of an abstraction we already have.

pdasigi:
Yes, the difference between the two situations is clear, and I agree about OOVs and padding being restricted to `Vocabulary`'s functionality.

My primary objection was about using the `Token` class for a label, because the former has attributes like pos, offset etc. While this is indeed not blocking, I think it could be confusing in the long run. Would it make sense to define a separate `Label` class, so that there's a fundamental difference between tokens and labels, and redefine (and rename) `TextField` such that it can take a sequence of either tokens or labels?

I don't know how big that change would be since I'm not very familiar with the repository yet. So I'll leave it up to you, and wouldn't strongly mind if you decide to go ahead with what Matt just suggested.

matt-gardner:
I guess, as we originally discussed this yesterday, I'm also fine with having a `Field` that is a sequence of labels.  It's just essentially a constrained version of a `TextField`, and there's some pretty severe naming collision with `SequenceLabelField` that has to be addressed.  I don't have a super strong opinion on which one is right here.

matt-gardner:
Actually, opinion: talking about this is slowing us down, when we really should just be implementing the decoder.  I'd probably start by just using a `TextField`, because it does right now what we need it to do, and add a TODO in the code to think about making it a new type of field later, if we want to do that.

pdasigi:
Okay, I'm with you on the "slowing us down" part. :)
I'll just close this PR, since we don't need any changes to the fields, and open a new PR with the decoder using a `TextField`.

PR #352, opened 2017-09-22
Entropy

DeNeutoy:
Adds entropy as a metric. 

I don't understand why I need to ignore the type in this PR for the `__call__` signature - 
This `Average` metric doesn't need to....
https://github.com/allenai/allennlp/blob/master/allennlp/training/metrics/average.py#L19

I also added a method on the base metric class to unpack the data from variables, because we do it in all of the metrics.


allennlp/training/metrics/boolean_accuracy.py
```diff
@@ -38,12 +37,7 @@ def __call__(self,
         # If you actually passed in Variables here instead of Tensors, this will be a huge memory
```
matt-gardner: This comment is in `unwrap_variables` - you probably don't need it here anymore.

allennlp/training/metrics/metric.py
```diff
@@ -46,3 +46,13 @@ def from_params(cls, params: Params, vocab: Optional[Vocabulary] = None):
         if vocab:
             params["vocabulary"] = vocab
         return cls.by_name(metric_type)(**params.as_dict())  # type: ignore
+
+    @staticmethod
+    def unwrap_variables(*tensors):
```
matt-gardner: Nit: should this be `unwrap_tensors` instead of `unwrap_variables`?

PR #353, opened 2017-09-22
big refactoring of Trainer

### LINKED ISSUE #204: Decompose `Trainer.train()` into smaller methods
It's grown to be a ~150 line method that's pretty hard to reason about.  It's a bit tricky to decompose, though, because of all of the dependencies between different parts of the method, but we should be able to pull out a bunch of it into separate methods.

joelgrus:
* broke `Trainer.train()` up into a lot of different functions
* got rid of all the `None` checks by writing wrapper functions that no-op in those cases
* created `TensorboardWriter` abstraction for the same reason

let me know if you don't like any of the design choices here

Fixes #204

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -161,9 +161,9 @@ class Trainer:
                 if parameter.requires_grad:
                     parameter.register_hook(clip_function)
 
-    def _clip_gradients(self) -> None:
+    def _rescale_gradients(self) -> None:
         """
-        Performs gradient clipping. Is a no-op if gradient clipping is not enabled.
+        Performs gradient rescaling. Is a no-op if gradient rescaling is not enabled.
         """
         if self._grad_norm:
             clip_grad_norm(self._model.parameters(), self._grad_norm)
@@ -229,8 +229,8 @@ class Trainer:
             # .cpu() is a no-op if you aren't using GPUs.
             train_loss += loss.data.cpu().numpy()
 
-            # Clip gradients (if enabled)
-            self._clip_gradients()
+            # Rescale gradients
+            self._rescale_gradients()
 
             # Advance the optimizer
             self._optimizer.step()
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -154,7 +154,7 @@ class Trainer:
 
     def _enable_gradient_clipping(self) -> None:
         if self._grad_clipping is not None:
-            # Pylint is unable to tell that we're in the case that _glar_clipping is not None...
+            # Pylint is unable to tell that we're in the case that _grad_clipping is not None...
             # pylint: disable=invalid-unary-operand-type
             clip_function = lambda grad: grad.clamp(-self._grad_clipping, self._grad_clipping)
             for parameter in self._model.parameters():
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -366,10 +366,12 @@ class Trainer:
 
         return False  # should not stop
 
-    def _next_epoch(self, epoch: int, metrics: dict,
-                    validation_metric_per_epoch: List[float]) -> bool:
+    def _should_stop(self, epoch: int, metrics: dict,
+                     validation_metric_per_epoch: List[float]) -> bool:
         """
-        Called at the end of each epoch. Different code paths if there's a validation set or not
+        Called at the end of each epoch.
+        Handles all end-of-epoch logic, and returns True if we should stop early.
+        Different code paths if there's a validation set or not.
         """
         if self._validation_dataset is not None:
             return self._next_epoch_validation(epoch, metrics, validation_metric_per_epoch)
@@ -388,16 +390,13 @@ class Trainer:
 
         logger.info("Beginning training.")
         validation_metric_per_epoch: List[float] = []
-        should_stop = False
 
         for epoch in range(epoch_counter, self._num_epochs):
             # Train one epoch and compute metrics.
             metrics = self._train_epoch(epoch)
 
             # Validate and move on to the next epoch
-            should_stop = self._next_epoch(epoch, metrics, validation_metric_per_epoch)
-
-            if should_stop:
+            if self._should_stop(epoch, metrics, validation_metric_per_epoch):
                 break
 
     def _forward(self, batch: dict, for_training: bool) -> dict:
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -248,6 +248,7 @@ class Trainer:
                                                            param.grad.data.std(),
                                                            batch_num_total)
                 self._tensorboard.add_train_scalar("loss/loss_train", metrics["loss"], batch_num_total)
+                self._metrics_to_tensorboard(epoch, metrics)
 
             # Log progress in no-tqdm case
             if self._no_tqdm and time.time() - self._last_log > self._log_interval:
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -40,11 +40,11 @@ class TensorboardWriter:
         self._train_log = train_log
         self._validation_log = validation_log
 
-    def add_train_scalar(self, name: str, value: float, global_step=None) -> None:
+    def add_train_scalar(self, name: str, value: float, global_step: int) -> None:
         if self._train_log is not None:
             self._train_log.add_scalar(name, value, global_step)
 
-    def add_validation_scalar(self, name: str, value: float, global_step=None) -> None:
+    def add_validation_scalar(self, name: str, value: float, global_step: int) -> None:
         if self._validation_log is not None:
             self._validation_log.add_scalar(name, value, global_step)
 
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -31,6 +31,24 @@ from allennlp.training.optimizers import Optimizer
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 
+class TensorboardWriter:
+    """
+    Wraps a pair of ``SummaryWriter`` instances but is a no-op if they're ``None``.
+    Allows Tensorboard logging without always checking for Nones first.
+    """
+    def __init__(self, train_log: SummaryWriter = None, validation_log: SummaryWriter = None) -> None:
+        self._train_log = train_log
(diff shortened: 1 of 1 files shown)


allennlp/training/trainer.py
```diff
@@ -125,18 +143,16 @@ def __init__(self,
...
-            if any(["model_state_epoch_" in x
-                    for x in os.listdir(self._serialization_dir)]):
-                logger.info("Loading model from checkpoint.")
-                epoch_counter = self._restore_checkpoint()
+        self._last_log = 0.0  # time of last logging
 
+        if serialization_dir is not None:
+            train_log = SummaryWriter(os.path.join(serialization_dir, "log", "train"))
+            validation_log = SummaryWriter(os.path.join(serialization_dir, "log", "validation"))
+            self._writer = TensorboardWriter(train_log, validation_log)
```
matt-gardner: I think this name would be better if it had a clear reference to tensorboard in it.  E.g., I'd prefer `_tensorboard` to `_writer`.  `_writer` could be writing lots of different things, but `_tensorboard` is pretty obvious.

allennlp/training/trainer.py
```diff
@@ -125,18 +143,16 @@ def __init__(self,
...
+        else:
+            self._writer = TensorboardWriter()
+
+    def _enable_gradient_clipping(self) -> None:
         if self._grad_clipping is not None:
             # Pylint is unable to tell that we're in the case that _glad_clipping is not None...
```
matt-gardner: Just noticed this comment says "_glad_clipping" instead of "_grad_clipping"...  Care to fix my typo?

allennlp/training/trainer.py
```diff
@@ -145,136 +161,243 @@ def train(self) -> None:
...
+        """
+        Performs gradient rescaling. Is a no-op if gradient rescaling is not enabled.
+        """
+        if self._grad_norm:
+            clip_grad_norm(self._model.parameters(), self._grad_norm)
+
+    def _compute_loss(self, batch: torch.Tensor, for_training: bool) -> torch.Tensor:
+        """
+        Does a forward pass on the given batch and returns the ``loss`` value in the result.
+        Also performs backpropagation if ``for_training`` is ``True``.
```
matt-gardner: If this is going to perform backprop, I think it should say so in its name: `_compute_and_backprop_loss`.  It's pretty unintuitive to expect a `_compute_loss` function to also do backprop.

matt-gardner: Actually, why put backprop in this function at all?  Why not just leave it in `_train_epoch`?  Seems cleaner.

allennlp/training/trainer.py
```diff
@@ -145,136 +161,243 @@ def train(self) -> None:
...
+            # Advance the optimizer
+            self._optimizer.step()
+
+            # Update the description with the latest metrics
+            metrics = self._get_metrics_for_batch(train_loss, batch_num)
+            description = self._description_from_metrics(metrics)
+            train_generator_tqdm.set_description(description)
+
+            # Log parameter values to Tensorboard
+            batch_num_total = num_training_batches * epoch + batch_num
```
matt-gardner: Can this block be pulled out into its own function, too?

allennlp/training/trainer.py
```diff
@@ -145,136 +161,243 @@ def train(self) -> None:
...
+            self._optimizer.zero_grad()
+
+            # Compute loss for batch and backpropagate
+            loss = self._compute_loss(batch, for_training=True)
+
+            # Make sure Variable is on the cpu before converting to numpy.
+            # .cpu() is a no-op if you aren't using GPUs.
+            train_loss += loss.data.cpu().numpy()
+
+            # Rescale gradients
```
matt-gardner: Most of these comments are just annotating a single line now, where the function being called already has a descriptive name.  I think you can remove these comments.

allennlp/training/trainer.py
matt-gardner: Also, doesn't look like all of the metrics are getting logged - shouldn't that be happening?

allennlp/training/trainer.py
```diff
@@ -145,136 +161,243 @@ def train(self) -> None:
...
+                                                      batch_num_total)
+                self._writer.add_train_scalar("loss/loss_train", metrics["loss"], batch_num_total)
+
+            # Log progress in no-tqdm case
+            if self._no_tqdm and time.time() - self._last_log > self._log_interval:
+                logger.info("Batch %d/%d: %s", batch_num, num_training_batches, description)
+                self._last_log = time.time()
+
+        # Return the metrics (and reset them)
+        return self._get_metrics_for_batch(train_loss, batch_num, reset=True)
```
matt-gardner: This is getting called twice, looks like.

allennlp/training/trainer.py
```diff
@@ -145,136 +161,243 @@ def train(self) -> None:
...
+            loss = self._compute_loss(batch, for_training=False)
+            val_loss += loss.data.cpu().numpy()
+
+            # Update the description with the latest metrics
+            val_metrics = self._get_metrics_for_batch(val_loss, batch_num)
+            description = self._description_from_metrics(val_metrics)
+            val_generator_tqdm.set_description(description)
+
+            # Log progress in the no-tqdm case
+            if self._no_tqdm and time.time() - self._last_log > self._log_interval:
```
matt-gardner: I think this was lost in `_train_epoch`.

matt-gardner: You could take this block and the two lines above (`description = ...; ...tqdm.set_description()`) and put them into a function that gets called by both train and validation, too, I think.

allennlp/training/trainer.py
```diff
@@ -145,136 +161,243 @@ def train(self) -> None:
...
+                logger.info("Batch %d/%d: %s", batch_num, num_validation_batches, description)
+                self._last_log = time.time()
+
+        # Log parameter values to Tensorboard
+        val_metrics = self._get_metrics_for_batch(val_loss, batch_num, reset=True)
+        message_template = "Training %s : %3f    Validation %s : %3f "
+        for name, value in metrics.items():
```
matt-gardner: I'd _probably_ be fine with just having the `_train_epoch` function log the train metrics, and not have to pass them to this function.  If we do pass them here, I'd definitely at least call them `train_metrics` instead of just `metrics`.

allennlp/training/trainer.py
```diff
@@ -145,136 +161,243 @@ def train(self) -> None:
...
+            logger.info(message_template, name, value, name, val_metrics[name])
+            self._writer.add_train_scalar(name, value, epoch)
+            self._writer.add_validation_scalar(name, val_metrics[name], epoch)
+
+        # Check validation metric for early stopping
+        this_epoch_val_metric = val_metrics[self._validation_metric]
```
matt-gardner: Can this be its own function?

allennlp/training/trainer.py
```diff
@@ -145,136 +161,243 @@ def train(self) -> None:
...
-                                  torch.optim.lr_scheduler.ReduceLROnPlateau):
-                        raise ConfigurationError("The reduce_on_plateau learning rate scheduler requires "
-                                                 "a validation metric to compute the schedule and therefore "
-                                                 "must be used with a validation dataset.")
-                    self._learning_rate_scheduler.step(epoch)
+            # Train one epoch and compute metrics.
+            metrics = self._train_epoch(epoch)
+
+            # Validate and move on to the next epoch
+            if self._should_stop(epoch, metrics, validation_metric_per_epoch):
```
matt-gardner: I feel like this one went a bit too far - I'd probably put an explicit check for validation data here, and a call to `validate` if so, otherwise just save a checkpoint, then put the learning rate scheduler update stuff after that.  Having the `next_epoch{_no}_validation` seems a little bit much to me.

allennlp/training/trainer.py
joelgrus: done

allennlp/training/trainer.py
joelgrus: done

allennlp/training/trainer.py
joelgrus: done

allennlp/training/trainer.py
joelgrus: done

allennlp/training/trainer.py
joelgrus: are you sure? the previous call is per-batch inside the loop, this is at the end of the loop to reset the metrics. or am I missing one?

allennlp/training/trainer.py
joelgrus: I can add that

allennlp/training/trainer.py
matt-gardner: Ah, yeah, you're right.  I think the name referring to `for_batch` threw me off.  Not sure how to make the name better, though...

allennlp/training/trainer.py
joelgrus: done

allennlp/training/trainer.py
joelgrus: I changed the name

allennlp/training/trainer.py
joelgrus: I renamed to make this clearer

joelgrus:
ok, I think this is ready for another look. I addressed most of the comments, there were a couple of "maybe factor this out" that I couldn't figure out a non-messy way to do, but I think it looks pretty good now.

allennlp/training/trainer.py
```diff
@@ -31,6 +31,24 @@
...
+    """
+    def __init__(self, train_log: SummaryWriter = None, validation_log: SummaryWriter = None) -> None:
+        self._train_log = train_log
+        self._validation_log = validation_log
+
+    def add_train_scalar(self, name: str, value: float, global_step=None) -> None:
+        if self._train_log is not None:
+            self._train_log.add_scalar(name, value, global_step)
+
+    def add_validation_scalar(self, name: str, value: float, global_step=None) -> None:
```
DeNeutoy: Why is global_step optional?

joelgrus: I'm sure I must have had a good reason for it, but I couldn't tell you what it is now.

joelgrus: I'll make it non-optional.

PR #354, opened 2017-09-26
Initial TriviaQA dataset reader

matt-gardner:
This sets up a basic TriviaQA dataset reader, and refactors the SquadReader a little bit to share some components.  I'll add a bit more functionality to it in later PRs (e.g., being smarter about picking the paragraph, returning all valid answer spans instead of picking one, etc.).  @chrisc36, if you wouldn't mind looking over this to make sure I didn't miss something obvious, that'd be great.


allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def pick_paragraphs(self,
+                        evidence_files: List[List[str]],
+                        question: str = None,
+                        answer_texts: List[str] = None) -> List[str]:
+        """
+        Given a list of evidence documents, return a list of paragraphs to use as training
+        examples.  Each paragraph returned will be made into one training example.
+
+        To aid in picking the best paragraph, you can also optionally pass the question text or the
```
DeNeutoy: This seems out of date - the arguments are unused. Seems like you can drop the pylint annotations make the method static and remove these arguments.

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
```diff
@@ -0,0 +1,162 @@
...
+        answer strings.  Note, though, that if you actually use the answer strings for picking the
+        paragraph on the dev or test sets, that's likely cheating, depending on how you've defined
+        the task.
+        """
+        # pylint: disable=no-self-use,unused-argument
+        paragraphs = []
+        for evidence_file in evidence_files:
+            whole_document = ' '.join(evidence_file)
+            tokens = whole_document.split(' ')
+            paragraph = ' '.join(tokens[:400])
```
DeNeutoy: Is this standard? Seems like it should be parameterised.

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
```diff
@@ -0,0 +1,162 @@
...
+            paragraphs.append(paragraph)
+        return paragraphs
+
+    @overrides
+    def text_to_instance(self,  # type: ignore
+                         question_text: str,
+                         passage_text: str,
+                         answer_texts: List[str] = None,
+                         question_tokens: List[Token] = None) -> Instance:
```
DeNeutoy: This type annotation should be `Optional[Instance]`.

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
matt-gardner: It's actually forward-looking.  I'm going to add in some options in the next PR that actually use these arguments.

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
matt-gardner: It will be.  This section is going to go into `util`, and there will be other options, where some parameters to `TriviaQaReader` determine which paragraph-picking method you actually use.

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
matt-gardner: Thanks, good point.

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
DeNeutoy: 🔮

matt-gardner:
@mandarjoshi90, FYI.

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
matt-gardner: This made me realize that making this optional was actually a bad decision - I moved the skipping logic to outside of this function, so this always returns an `Instance`.

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
```diff
@@ -0,0 +1,171 @@
...
+                    evidence_file = base_tarball.extractfile(os.path.join("evidence", "web", filename))
+                    evidence_files.append([line.decode('utf-8') for line in evidence_file.readlines()])
+            else:
+                for result in question_json['EntityPages']:
+                    filename = result['Filename']
+                    evidence_file = base_tarball.extractfile(os.path.join("evidence", "wikipedia", filename))
+                    evidence_files.append([line.decode('utf-8') for line in evidence_file.readlines()])
+
+            answer_json = question_json['Answer']
+            answer_texts = answer_json['NormalizedAliases'] + answer_json.get('HumanAnswers', [])
```
chrisc36: The human answers are stored in the normalized form. The official evaluation script runs `normalize_answer` on them before evaluation. To avoid mixing un-normalized and normalized answers I would pre-normalize these answers as well.

allennlp/data/dataset_readers/reading_comprehension/util.py
```diff
@@ -63,3 +75,121 @@ def char_span_to_token_span(token_offsets: List[Tuple[int, int]],
...
+                if answer_tokens[answer_index] == token:
+                    answer_index += 1
+                    span_end += 1
+                elif token in IGNORED_TOKENS:
+                    span_end += 1
+                else:
+                    break
+            if num_answer_tokens == answer_index:
+                spans.append((span_start, span_end))
+    return spans
```
chrisc36: You might want to remove duplicate spans here, I would hesitate to assert the its impossible for two different answer strings to "trigger" the same span.

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
matt-gardner: Thanks, good point.

allennlp/data/dataset_readers/reading_comprehension/util.py
matt-gardner: The logic that follows actually takes care of this.  It was built for the SQuAD dev set, which can have duplicate annotations, so it picks the most frequent one.  That means we'd bias toward picking spans which work for multiple aliases, which is probably a good thing if it ever happens.  But I'll leave a note here about that, so you don't have to be as familiar with the rest of the code when reading and modifying this piece.

PR #355, opened 2017-09-26
don't log metrics to console in the middle of an epoch

joelgrus:
based on a suggestion I added intermittent "log metrics to tensorflow" in the middle of an epoch. but that method had log statements in it, which means that you get metrics logged to the console in the middle of an epoch, which looks bad and mucks with tqdm and everything.

anyway, I split out the log statements into a separate `_metrics_to_console` function, and I only call it at the end of an epoch. (the metrics still are part of the tqdm description)

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -285,7 +285,7 @@ class Trainer:
             if val_metrics:
                 self._tensorboard.add_validation_scalar(name, val_metrics[name], epoch)
 
-    def _metrics_to_console(self,
+    def _metrics_to_console(self,                               # pylint: disable=no-self-use
                             train_metrics: dict,
                             val_metrics: dict = None) -> None:
         """
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -277,16 +277,28 @@ class Trainer:
                                 epoch: int,
                                 train_metrics: dict,
                                 val_metrics: dict = None) -> None:
+        """
+        send all of the train metrics (and validation metrics, if provided) to tensorboard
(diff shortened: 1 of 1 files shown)


allennlp/training/trainer.py
```diff
@@ -277,16 +277,28 @@ def _metrics_to_tensorboard(self,
                                 epoch: int,
                                 train_metrics: dict,
                                 val_metrics: dict = None) -> None:
+        """
+        send all of the train metrics (and validation metrics, if provided) to tensorboard
```
matt-gardner: Capitalization and periods.

allennlp/training/trainer.py
```diff
@@ -277,16 +277,28 @@ def _metrics_to_tensorboard(self,
...
+        """
+        for name, value in train_metrics.items():
+            self._tensorboard.add_train_scalar(name, value, epoch)
+            if val_metrics:
+                self._tensorboard.add_validation_scalar(name, val_metrics[name], epoch)
+
+    def _metrics_to_console(self,                               # pylint: disable=no-self-use
```
matt-gardner: Spacing seems odd here; we've typically just been doing two spaces after the end of the line.

allennlp/training/trainer.py
```diff
@@ -277,16 +277,28 @@ def _metrics_to_tensorboard(self,
...
+                            train_metrics: dict,
+                            val_metrics: dict = None) -> None:
+        """
+        log all the train metrics (and validation metrics, if provided) to the console
```
matt-gardner: Formatting here, too.

PR #364, opened 2017-10-02
Initial simplified e2e-coref implementation.

kentonl:
Here is a slightly simplified implementation of the coreference resolution system from here: https://arxiv.org/abs/1707.07045. This version achieves ~63 dev F1, compared to the reported ~67 dev F1.

There are a few differences that make the results worse than in the paper.
(1) This only considers up to 100 antecedents (rather than 250) due to memory constraints. The original implementation dynamically and randomly generated slices of documents that are too long. I'm not sure if that is currently supported by AllenNLP.
(2) This does not apply recurrent variational dropout.
(3) Does not enforce the non-crossing constraint between predicted mentions. This probably requires writing a custom kernel to do the greedy pruning. It's also not important unless you need to use the official CoNLL evaluation scripts, which implicitly require this constraint due to their formatting.
(4) Does not use genre and speaker metadata (unrealistic in many applications anyways...).
(5) Does not ensemble the models.


DeNeutoy:
Hi Kenton, this looks great, thanks for making the effort. Would you mind:

- Adding some documentation (take a look at the other models for formatting etc).
- Add some tests (for the model, inheriting from `AllennlpModelTestCase` should make this easy for the model. For the dataset reader, feel free to add anything you need in `tests/fixtures`. For the utils you have added, if you could add them to `tests/nn/nn_utils_test.py`, that would be great). This would make your PR easier to review. 
- General point concerning style - if you could take a look at one of the other model implementations and draw from there, that would be helpful - particularly having uniformly formatted shape comments is good (above the line, using round brackets and full variable names: 
```
# shape: (batch_size, ..., embedding_size)
embeddings = self.get_embeddings(...)
```).
- Possibly break up the forward pass of your model into a few more functions, as it is pretty large. 

Let us/me know if you need any help with any of this. Also, just FYI, we normally won't review PRs until you request a review directly, but given this was your first PR to the library I figured you might not expect that (although if this isn't finished yet, sorry for jumping in too early!).

Thanks!

File: allennlp/models/coreference_resolution/coref.py [hunk 1]
```diff
@@ -0,0 +1,244 @@
...
+    def _bucket_distance(distances):
+        # Buckets: [0, 1, 2, 3, 4, 5-7, 8-15, 16-31, 32-63, 64+].
+        logspace_idx = (distances.float().log()/math.log(2)).floor().long() + 3
+        use_identity = (distances <= 4).long()
+        combined_idx = use_identity * distances + (1 + (-1 * use_identity)) * logspace_idx
+        return combined_idx.clamp(0, 9)
+
+    def forward(self,  # type: ignore
+                text: Dict[str, torch.LongTensor],
+                span_starts: torch.IntTensor,
```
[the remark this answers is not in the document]
matt-gardner: Looking at your loss function, too, it doesn't look like span start and end are part of the loss.  So they are just considered input?  In a demo, we'd have to run some mention detector on the text first, before inputting to this?  Or am I missing something?

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,244 @@
...
+        flat_span_text_emb = span_text_emb.view(-1,
+                                                self._max_span_width,
+                                                span_text_emb.size(-1)) # [b * num_spans, max_span_width, emb]
+        flat_attended_text_emb = flat_span_head_weights.bmm(flat_span_text_emb) # [b * num_spans, 1, emb]
+        attended_text_emb = flat_attended_text_emb.view(batch_size, num_spans, -1) # [b, num_spans, emb]
+        span_emb = torch.cat([start_emb, end_emb, span_size_emb, attended_text_emb], -1) # [b, num_spans, emb]
+
+        # Compute mention scores.
+        mention_scores = self._mention_scorer(self._mention_feedforward(span_emb)) # [b, num_spans, 1]
+        mention_scores += span_mask.log() # [b, num_spans, 1]
```
matt-gardner: I _think_ `mention_scores *= span_mask` does what you need, and is faster than taking a log.  Well, maybe not, if the scorer outputs negative values...  Ok.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,244 @@
...
+
+        # Prune based on mention scores.
+        k = int(math.floor(self._spans_per_word * text_len)) # []
+        _, top_span_idx = mention_scores.topk(k, 1) # [b, k, 1]
+        top_span_idx, _ = torch.sort(top_span_idx, 1) # [b, k, 1]
+        top_span_idx = top_span_idx.squeeze(-1) # [b, k]
+        flat_top_span_idx = util.flatten_batched_indices(top_span_idx, num_spans) # [b * k]
+
+        # Select tensors relating to the top spans.
+        top_span_mask = util.batched_index_select(span_mask, top_span_idx, flat_top_span_idx) # [b, k, 1]
```
matt-gardner: Is this assuming you could pick something in the top k that was masked?

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,244 @@
...
+        top_span_emb = util.batched_index_select(span_emb, top_span_idx, flat_top_span_idx) # [b, k, emb]
+        top_span_labels = util.batched_index_select(span_labels.unsqueeze(-1),
+                                                    top_span_idx,
+                                                    flat_top_span_idx) # [b, k, 1]
+
+        # Compute indices for antecedent spans to consider.
+        max_ant = min(self._max_antecedents, k) # []
+        target_idx = util.get_indices(k, is_cuda).unsqueeze(1) # [k, 1]
+        ant_offsets = (util.get_indices(max_ant, is_cuda) + 1).unsqueeze(0) # [1, max_ant]
+        raw_ant_idx = target_idx - ant_offsets # [k, max_ant]
```
matt-gardner: This is another place where some comments would be particularly helpful.  You're essentially creating a triangular matrix like:

```
0 0 0 0 ...
1 0 0 0 ...
1 2 0 0 ...
1 2 3 0 ...
...
```

Right?

Since you do this in a few places, it's probably worth just making this a two-argument function in `nn.util`, with width and height as arguments (and probably whether the diagonal itself should be included).  And if that function has a good name, no comments needed here.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,244 @@
...
+                                                   ant_distance_emb.size(-1)) # [b, k, max_ant, emb]
+        pairwise_emb = torch.cat([target_emb,
+                                  ant_emb,
+                                  similarity_emb,
+                                  ant_distance_emb], -1) # [b, k, max_ant, emb]
+        ant_scores = self._antecedent_scorer(self._antecedent_feedforward(pairwise_emb)) # [b, k, max_ant, 1]
+        ant_scores = ant_scores.squeeze(-1) # [b, k, max_ant]
+        ant_scores += target_mention_scores + ant_mention_scores # [b, k, max_ant]
+        ant_scores += ant_mask # [b, k, max_ant]
+        dummy_scores = util.get_zeros((batch_size, k, 1), is_cuda) # [b, k, 1]
```
matt-gardner: What are the dummy scores for?  No antecedent?

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,244 @@
...
+                                                   max_ant,
+                                                   ant_distance_emb.size(-1)) # [b, k, max_ant, emb]
+        pairwise_emb = torch.cat([target_emb,
+                                  ant_emb,
+                                  similarity_emb,
+                                  ant_distance_emb], -1) # [b, k, max_ant, emb]
+        ant_scores = self._antecedent_scorer(self._antecedent_feedforward(pairwise_emb)) # [b, k, max_ant, 1]
+        ant_scores = ant_scores.squeeze(-1) # [b, k, max_ant]
+        ant_scores += target_mention_scores + ant_mention_scores # [b, k, max_ant]
+        ant_scores += ant_mask # [b, k, max_ant]
```
matt-gardner: Calling this a mask confused me - we've typically used `mask` to refer to the binary tensor, not the log of the binary tensor.

allennlp/nn/util.py
```diff
@@ -499,3 +499,47 @@ def _get_combination_dim(combination: str, tensor_dims: List[int]) -> int:
...
+    return max_score + (stable_vec.exp().sum(dim, keepdim=keepdim)).log()
+
+def get_zeros(shape, is_cuda):
+    if is_cuda:
+        zeros = torch.cuda.FloatTensor(*shape).fill_(0)
+    else:
+        zeros = torch.zeros(*shape).float()
+    return Variable(zeros, requires_grad=False)
+
+def get_indices(size, is_cuda):
```
matt-gardner: `get_indices` seems like too generic a name for this - it's not clear enough exactly what this does from the name.  Consider that above you're using `indices` to refer to a random assortment of integers, where this returns an `arange` variable.  Maybe something like `get_range_vector`? (`vector` here so there's room for something like `get_range_matrix` for the 2D version I recommended adding.) @DeNeutoy, any ideas?

allennlp/training/metrics/conll_scores.py
```diff
@@ -0,0 +1,155 @@
...
+from collections import Counter
+import numpy as np
+from sklearn.utils.linear_assignment_ import linear_assignment
+
+from overrides import overrides
+
+from allennlp.training.metrics.metric import Metric
+
+@Metric.register("conll_scores")
+class ConllScores(Metric):
```
matt-gardner: There are lots of CoNLL datasets, each with their own metric.  This should be something like `CorefScores`, or `ConllCorefScores` if you want to keep the reference to the CoNLL coref metrics (as opposed to, e.g., plain b-cubed).

allennlp/training/metrics/conll_scores.py
```diff
@@ -0,0 +1,155 @@
...
+    def get_metric(self, reset: bool = False) -> Tuple[float, float, float]:
+        metrics = (lambda e: e.get_precision(), lambda e: e.get_recall(), lambda e: e.get_f1())
+        precision, recall, f1_score = tuple(sum(m(e) for e in self.scorers) / len(self.scorers) for m in metrics)
+        if reset:
+            self.reset()
+        return precision, recall, f1_score
+
+    @overrides
+    def reset(self):
+        self.scorers = [Scorer(m) for m in (Scorer.muc, Scorer.b_cubed, Scorer.ceafe)]
```
matt-gardner: I'm surprised pylint isn't complaining about `self.scorers` being defined outside of `__init__`.

allennlp/training/metrics/mention_recall.py
```diff
@@ -0,0 +1,33 @@
...
+        self._num_gold_mentions = 0
+        self._num_recalled_mentions = 0
+
+    @overrides
+    def __call__(self, batched_top_spans, batched_metadata):
+        for top_spans, metadata in zip(batched_top_spans.data.tolist(), batched_metadata):
+            gold_mentions = set(m for c in metadata["clusters"] for m in c)
+            top_spans = set(tuple(s) for s in top_spans)
+            self._num_gold_mentions += len(gold_mentions)
+            self._num_recalled_mentions += len(gold_mentions & top_spans)
```
matt-gardner: It's not clear to me what exactly is getting compared in this `&` call.  Maybe types on the inputs would help?  Or just some docstrings?

scripts/compile_coref_data.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+dlx() {
```
matt-gardner: `download_and_extract`?

allennlp/data/fields/label_field.py
```diff
@@ -78,4 +78,4 @@ def as_array(self, padding_lengths: Dict[str, int]) -> numpy.ndarray:  # pylint:
 
     @overrides
     def empty_field(self):
-        return LabelField(0, self._label_namespace)
+        return LabelField(-1, self._label_namespace)
```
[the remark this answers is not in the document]
kentonl: ^

kentonl:
Yep, I'll fix up the cosmetics issues and add some tests. 

Regarding Matt's questions: 
(1) One problem with the model is that for large documents, the GPU will not have enough memory. One way to ensure things fit is to take random chunks of the document. Suppose you have a document x with n words: [x1, x2, ..., xn]. On the first iteration, you might use only x[5:5 + c], on the second iteration, you might use only x[22:22 + c]. This makes sure you eventually all of your data.
(2) Correct, variational dropout is only at the context layer.
(3) The tricky part is that this constraint enforcing happens in the middle of the computation graph, since it decides which span gets pruned. You can technically do this on pure Python relatively simply, but it would be quite slow. In reality, this doesn't affect the performance, but it is needed to use the official evaluation metrics. 
(4) Yes this is quite simple to do, I just haven't gotten to it yet.
(5) Sorry, the way I presented it was confusing. The lack of ensembling isn't contributing to the gap that I mentioned. It just something that's missing.

allennlp/data/fields/index_field.py
```diff
@@ -44,4 +44,4 @@ def as_array(self, padding_lengths: Dict[str, int]) -> numpy.array:  # pylint: d
 
     @overrides
     def empty_field(self):
-        return IndexField(None, None)
+        return IndexField(-1, None)
```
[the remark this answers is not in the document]
kentonl: In addition to the benefit of enabling masking, it also avoids silent errors due to forgetting to handle padding

allennlp/data/dataset_readers/coreference_resolution/conll.py
```diff
@@ -0,0 +1,192 @@
...
+
+    @staticmethod
+    def normalize_word(word):
+        if word == "/." or word == "/?":
+            return word[1:]
+        else:
+            return word
+
+    def handle_line(self, line, document_state):
+        begin_document_match = self._begin_doc_regex.match(line)
```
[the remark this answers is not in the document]
kentonl: Yep, this can be avoided. Though this isn't the slow preprocessing that I was referring to. It's the stuff that's in the scripts directory that reconciles ontonotes with the conll data. This code is reasonably quick to run (less than a minute).

allennlp/models/coreference_resolution/coref.py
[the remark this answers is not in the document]
kentonl: It's jointly inferring and learning mention detection and coref, which is why they are inseparable. The span starts and span ends are implicitly part of the loss, since the span representations depend on them.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,244 @@
...
+        head_idx = F.relu(raw_head_idx.float()).long() # [b, num_spans, max_span_width]
+        flat_head_idx = util.flatten_batched_indices(head_idx,
+                                                     text_len) # [b * num_spans * max_span_width]
+        span_text_emb = util.batched_index_select(text_emb,
+                                                  head_idx,
+                                                  flat_head_idx) # [b, num_spans, max_span_width, emb]
+        span_head_scores = util.batched_index_select(head_scores,
+                                                     head_idx,
+                                                     flat_head_idx).squeeze(-1) # [b, num_spans, max_span_width]
+        span_head_scores += head_mask.float().log() # [b, num_spans, max_span_width]
```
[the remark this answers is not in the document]
kentonl: IMO, this is a lot cleaner and faster than what's in masked_softmax. Masking in logspace only requires a single elementwise multiply. In practice there are no numerical stability issues. I was actually going to add a pull request that would fix this in bidaf...

allennlp/models/coreference_resolution/coref.py
kentonl: It could, but it would also conveniently end up masking the overall score, since the mention score is a factor in the antecedent score :)

allennlp/models/coreference_resolution/coref.py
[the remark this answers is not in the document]
matt-gardner: Ok, the confusion was due to me not looking closely enough at how `span_starts` and `span_ends` are created. You just enumerate all possible spans. It's just the labels that should be optional, then.

allennlp/models/coreference_resolution/coref.py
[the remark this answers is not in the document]
matt-gardner: Would it make sense to just fix our `masked_softmax`? I never thought of doing it like this, but it seems reasonable as long as it doesn't give NaNs (I'm still a little surprised it works, because you're taking `log(0)`...).

The benefit of using the function is that it improves readability of the code - you just have one line with a very transparent function call, instead of four lines of less-transparent logic.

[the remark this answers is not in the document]
kentonl: Not sure an extra abstraction is necessary here. Is `util.mask_softmax(scores, mask)` really more readable than `F.softmax(scores + mask.log())`? There are also a lot of cases where you'd want to reuse the logged masks or the masked scores, and chances are you'll end up with inconsistent usage of this API for efficiency reasons.

allennlp/data/dataset_readers/coreference_resolution/conll.py
```diff
@@ -0,0 +1,192 @@
...
+
+            if coref != "-":
+                for segment in coref.split("|"):
+                    if segment[0] == "(":
+                        if segment[-1] == ")":
+                            cluster_id = int(segment[1:-1])
+                            document_state.clusters[cluster_id].append((word_index, word_index))
+                        else:
+                            cluster_id = int(segment[1:])
+                            document_state.coref_stacks[cluster_id].append(word_index)
```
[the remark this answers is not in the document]
kentonl: Each cluster id specifies a set of spans (a set of starts and ends), so I'm not sure how to get away without a stack.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,244 @@
...
+        self._mention_recall = MentionRecall()
+        self._conll_scores = ConllScores()
+        if lexical_dropout > 0:
+            self._lexical_dropout = torch.nn.Dropout(p=lexical_dropout)
+        else:
+            self._lexical_dropout = lambda x: x
+        initializer(self)
+
+    @staticmethod
+    def _bucket_distance(distances):
```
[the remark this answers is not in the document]
kentonl: I'ved moved this to `nn.util`.

allennlp/models/coreference_resolution/coref.py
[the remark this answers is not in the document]
kentonl: Fixed.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,244 @@
...
+                                                     head_idx,
+                                                     flat_head_idx).squeeze(-1) # [b, num_spans, max_span_width]
+        span_head_scores += head_mask.float().log() # [b, num_spans, max_span_width]
+        flat_span_head_scores = span_head_scores.view(-1, self._max_span_width) # [b * num_spans, max_span_width]
+        flat_span_head_weights = F.softmax(flat_span_head_scores) # [b * num_spans, max_span_width]
+        flat_span_head_weights = flat_span_head_weights.unsqueeze(1) # [b * num_spans, 1, max_span_width]
+        flat_span_text_emb = span_text_emb.view(-1,
+                                                self._max_span_width,
+                                                span_text_emb.size(-1)) # [b * num_spans, max_span_width, emb]
+        flat_attended_text_emb = flat_span_head_weights.bmm(flat_span_text_emb) # [b * num_spans, 1, emb]
```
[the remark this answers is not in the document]
kentonl: In `weighted_sum` there is a broadcasted element-wise multiplication and then a summation. That is literally what a matrix multiply does. By using torch.bmm instead, you can avoid creating that enormous intermediate tensor.

allennlp/models/coreference_resolution/coref.py
kentonl: Fixed.

allennlp/nn/util.py
kentonl: Fixed.

allennlp/training/metrics/conll_scores.py
kentonl: Fixed.

allennlp/training/metrics/conll_scores.py
kentonl: Fixed, since I'm assuming the implication is that it's better to duplicate the call in `reset` and `__init__`.

allennlp/training/metrics/mention_recall.py
kentonl: It's comparing pairs of indices in the form of tuples. I've added some types.

scripts/compile_coref_data.sh
kentonl: Fixed.

allennlp/data/dataset_readers/coreference_resolution/conll.py
```diff
@@ -0,0 +1,192 @@
...
+                                                     clusters)
+                    instances.append(instance)
+                    document_state = DocumentState()
+        if not instances:
+            raise ConfigurationError("No instances were read from the given filepath {}. "
+                                     "Is the path correct?".format(file_path))
+        return Dataset(instances)
+
+    @overrides
+    def text_to_instance(self,  # type: ignore
```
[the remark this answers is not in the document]
kentonl: Fixed.

allennlp/training/metrics/conll_scores.py
matt-gardner: I wasn't trying to be passive-aggressive with that, I was just surprised (and still am) that pylint wasn't doing what I expected there.  I don't have a strong preference in this case.  Had pylint complained, I probably would have just done `self.scorers = []; self.reset()` in the constructor, to avoid the duplication, but what you have is fine.

allennlp/data/dataset_readers/coreference_resolution/conll.py
```diff
@@ -0,0 +1,197 @@
...
+        self._begin_doc_regex = re.compile(r"#begin document \((.*)\); part (\d+)")
+
+    @overrides
+    def read(self, file_path: str):
+        # if `file_path` is a URL, redirect to the cache
+        file_path = cached_path(file_path)
+
+        logger.info("Reading file at %s", file_path)
+        instances = []
+        with open(file_path) as dataset_file:
```
DeNeutoy: Are we missing an outer loop over documents in the dataset here? Seems like this just handles a single file.

allennlp/data/dataset_readers/coreference_resolution/conll.py
```diff
@@ -0,0 +1,197 @@
...
+from allennlp.data.instance import Instance
+from allennlp.data.tokenizers import Token
+from allennlp.data.token_indexers import SingleIdTokenIndexer, TokenIndexer
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+@DatasetReader.register("coref")
+class ConllCorefReader(DatasetReader):
+    def __init__(self,
+                 max_span_width: int,
```
DeNeutoy: A docstring for `max_span_width` would be good.

allennlp/data/dataset_readers/coreference_resolution/conll.py
```diff
@@ -0,0 +1,197 @@
...
+                    document_state = DocumentState()
+        if not instances:
+            raise ConfigurationError("No instances were read from the given filepath {}. "
+                                     "Is the path correct?".format(file_path))
+        return Dataset(instances)
+
+    @overrides
+    def text_to_instance(self,  # type: ignore
+                         sentences: List[List[str]],
+                         clusters: List[List[Tuple[int, int]]] = None) -> Instance:
```
DeNeutoy: This is a complex enough type that a short docstring would be useful, e.g are the spans inclusive/exclusive.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,411 @@
...
+class CoreferenceResolver(Model):
+    """
+    This ``Model`` implements the coreference resolution model described "End-to-end Neural
+    Coreference Resolution"
+    <https://www.semanticscholar.org/paper/End-to-end-Neural-Coreference-Resolution-Lee-He/3f2114893dc44eacac951f148fbff142ca200e83>
+    by Lee et al., 2017.
+    The basic outline of this model is to get an embedded representation of each span in the
+    document. These span representation are scored and use to prune away spans that are unlikely
+    to occur in a coreference cluster. For the remaining spans, the model decides which antecedent
+    span (if any) they are coreference with. The resulting coreference links, after applying,
```
DeNeutoy: "they are coreference" -> "they are coreferent"

DeNeutoy: "after applying, transitivity" -> "after applying transitivity, "

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,411 @@
...
+        self._max_antecedents = max_antecedents
+        self._mention_recall = MentionRecall()
+        self._conll_coref_scores = ConllCorefScores()
+        if lexical_dropout > 0:
+            self._lexical_dropout = torch.nn.Dropout(p=lexical_dropout)
+        else:
+            self._lexical_dropout = lambda x: x
+        initializer(self)
+
+    def _compute_head_attention(self, head_scores, text_emb, span_ends, span_size):
```
DeNeutoy: Nitty stuff now: Could you:
- Add types and docstrings to these private functions.
- use `embedding` rather than `emb` and `index` rather than `idx`

allennlp/training/metrics/conll_coref_scores.py
```diff
@@ -0,0 +1,155 @@
...
+                mention_to_predicted[predicted_antecedent] = predicted_cluster
+            mention = tuple(top_spans[i])
+            predicted_clusters[predicted_cluster].append(mention)
+            mention_to_predicted[mention] = predicted_cluster
+        predicted_clusters = [tuple(pc) for pc in predicted_clusters]
+        mention_to_predicted = {m:predicted_clusters[i] for m, i in mention_to_predicted.items()}
+        return predicted_clusters, mention_to_predicted
+
+
+class Scorer(object):
```
DeNeutoy: I know you just borrowed this class, but it would be great if you could clean up the variable naming etc. For some stuff I can guess what it means, but I actually don't know what `ceafe`, `muc` etc mean.

allennlp/data/dataset_readers/coreference_resolution/conll.py
DeNeutoy: Oh, maybe your bash script compiles it into a single file. My bad.

allennlp/training/metrics/conll_coref_scores.py
matt-gardner: ceafe and muc are names of coref metrics.  I don't think we need to change those names, but referencing them in the class docstring is probably warranted.

allennlp/data/dataset_readers/coreference_resolution/conll.py
kentonl: Added docstring.

allennlp/data/dataset_readers/coreference_resolution/conll.py
kentonl: Added docstring.

allennlp/training/metrics/conll_coref_scores.py
kentonl: Added reference of the various metrics to docstrings.

kentonl:
Thanks Mark and Matt for the code reviews! I've addressed most of the comments, but the tests are still missing. I'll add those soon...

DeNeutoy:
@kentonl - I've been making a few usage improvements this morning - are you happy for me to just push them to this branch?

kentonl:
Yes, please feel free to push any changes.

DeNeutoy:
@kentonl - i've just added a test for the coref model. One issue we have here is that we have a test which checks that predictions from the model are equal if you do them in batches vs individually, which is not currently true for the coref model, because the number of spans you consider is computed based on the size of the largest element in the batch, which means we consider more spans for shorter sentences if they are batched with longer examples.

kentonl:
Thanks for adding those! I could see arguments for both sides. On one hand, it would great if the model is agnostic to batching. On the other hand, if you're batching anyways, masking those extra spans is wasting computation. I've found that pruning less always produces strictly better results.

allennlp/data/dataset_readers/coreference_resolution/conll.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def complete_sentence(self) -> None:
+        self.sentences.append(self.sentence_buffer)
+        self.sentence_buffer = []
+
+    def add_word(self, word) -> None:
+        self.sentence_buffer.append(word)
+        self.num_total_words += 1
+
+    def finalize(self) -> List[Tuple]:
```
DeNeutoy: @kentonl - this method has stumped me a bit. Why do we ever need to merge clusters with different ids? Do we ever get exactly overlapping spans with different ids in the data?

kentonl: If the dataset were perfect, this would not be needed. Unfortunately, the same mention can appear in multiple clusters. This happens exactly once in the training set, once in the dev set, and never in the test set.

DeNeutoy: I can imagine the deep sense of rage you must have felt when you discovered that. Thanks!

DeNeutoy:
@kentonl, @matt-gardner  I'm merging into a new `coref-wip` branch - this PR is getting a bit unwieldy to look at, and the majority of this is looking great.

PR #366, opened 2017-10-02
Basic Encoder decoder model

pdasigi:


allennlp/data/dataset_readers/encoder_decoder.py
```diff
@@ -0,0 +1,82 @@
...
+                line = line.strip("\n")
+
+                if not line:
+                    continue
+
+                line_parts = line.split('\t')
+                if len(line_parts) != 2:
+                    raise ConfigurationError("Invalid line format: %s (line number %d)" % (line, line_num + 1))
+                source_sequence, target_sequence = line_parts
+                target_sequence = "%s %s %s" % (START_SYMBOL, target_sequence, END_SYMBOL)
```
matt-gardner: The target sequence probably doesn't need a start symbol, does it?  And you can get the behavior you want by adding an `end_token` in the tokenizer:

https://github.com/allenai/allennlp/blob/3b809fee6b37c6b99dd1fb5867b13e775e1f4b4d/allennlp/data/tokenizers/word_tokenizer.py#L36-L37

Probably just adding an assert in the constructor that tokenizer's `end_tokens` isn't empty is good enough.

pdasigi: I think we need the start token because at test time that is the input at the first time step to the decoder. To keep things consistent, we need it while training as well.

Good to know about passing start and end tokens to the tokenizer. Will change that.

matt-gardner: Hmm...  Alternatively, you could just have a learned initial hidden state.  I guess having a start token with a learned embedding is _almost_ the same thing.  And this makes the API around decoding more consistent and cleaner.  Ok, yeah, you're right.  I'd still probably also learn the initial hidden state, but that's a separate issue.

pdasigi: The initial hidden state of the decoder is supposed to be the final hidden state from the encoder, right?

pdasigi: And regarding passing start and end symbols to the tokenizer, I'm using the same tokenizer for both source and target sequences, and we do not need start and end symbols in the source sequences. Would it make sense to change the tokenizer such that the `tokenize` method accepts start and end symbols, instead of the constructor?

.gitignore
```diff
@@ -2,3 +2,6 @@
 .cache/
 .coverage
 doc/_build/
+*~
+*.swp
```
DeNeutoy: FYI: `set noswapfile` in your .vimrc makes these never occur, they are so annoying

matt-gardner: That also makes it so vim can't recover if it crashes for whatever reason (e.g., you're using vim in an ssh session, and you lose your connection).  I have never seen vim create a backup file (`*~`) in my ~10 years of using it, so I just spent a few minutes trying to figure out _why_, and I couldn't.  Clearly vim can create them in some circumstances, but something in my vim setup must prevent it somehow.  For some discussion of this, see, e.g., [here](https://stackoverflow.com/a/750692/1467533) and [here](https://stackoverflow.com/a/607475/1467533).

But, more applicable to this change in this PR, these lines are specific to your development environment, and you should not pollute the shared `.gitignore` with lines specific to your IDE.  These should go in a [global `.gitignore` file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

allennlp/data/dataset_readers/encoder_decoder.py
matt-gardner: Yes, you're right on the hidden state.  Ignore my comment there.

Re: the tokenizer, for the semantic parser you almost certainly want different tokenizers for the input and the output, because the output is probably just space-separated, and trying to split it as if it were English would be wrong.  So, if your goal is to make a general piece of code that can be used for any encoder/decoder model (though I still think that may be trying to do too much), you'd want to have separate tokenizers, maybe defaulting the output tokenizer to be the same as the input tokenizer.

DeNeutoy: I think this should be `input_sequence + DECODE_START_SYMBOL` and `target_sequence + END_SYMBOL` - I would say this is more standard. But Pradeep's method is equally valid, just maybe less standard. Also you are right that the hidden state is passed across from the encoder to the decoder.

DeNeutoy: Sorry, ignore the above - there needs to be two target fields, one for the input to the decoder during training, which should be `DECODE_START + target_sequence` and one for the actual targets, which should be `target_sequence + END_SYMBOL`. Both of these should be optional model inputs.

pdasigi: Matt, I agree that having separate tokenizers is a good idea. But I think the target tokenizer should default to a word tokenizer with start and end symbols, not the source tokenizer because even in the simplest case source and target should be tokenized differently.

Mark, how would adding a `DECODE_START_SYMBOL` at the end of the source sequence work? What would be the first input to the decoder then? I thought the whole point was to assign a symbol that can always be used as the first input to the decoder.

.gitignore
pdasigi: Noted. I agree that adding these to a global `.gitignore` is the right thing to do.

allennlp/data/dataset_readers/encoder_decoder.py
DeNeutoy: Sorry, see my next comment.

pdasigi: Hmm.. sounds like you're talking about targets and predictions. Is that the case? If so, what does it mean to have predictions as inputs to the model?

DeNeutoy: No - at training time, the model has 2 inputs: the encoder_input and the decoder_input. Separately, there are targets, which are:

```
decoder_input = [START_TOKEN] + [words]
decoder_target = decoder_input[1:] + [END_TOKEN]
```
I meant to make both of these inputs to `forward` optional, because at test time, your input will be just the input_sequence and the decoder_input will be `[START_TOKEN]`. Does that make sense?

pdasigi: Ah! I see what you mean. I understand the difference between decoder inputs and targets. But do we need them both as inputs to forward? Can we not just take the whole sequence [START_TOKEN] + [words] + [END_TOKEN] as the input and compute the two sequences? This should work as long as loss is computed against the right subset of the sequence, right?
(Thanks for bringing this up, though. I realized that my loss computation is not one-off. I should fix it. :))

DeNeutoy: Yeah, that would work. My only thought regarding this would be that separating them might make doing things like incorporating teacher forcing easier, because you could just have an `if decoder_input is None: ` block which would sample from the model during training rather than using the `decoder_input`. If you don't separate them, then you'll have to pass a flag, but I guess that's a really trivial detail. Proceed as you wish 🚛

pdasigi: Yes, I was going to do it with a flag. But I think the arguments to `forward` is an orthogonal issue because even while doing scheduled sampling, the method needs to get both the `decoder_inputs` and `predicted_outputs` and make a stochastic choice between the two.

In general, with these models, I cannot think of a case where `decoder_inputs = whole_sequence[:-1]` and `decoder_target = whole_sequence[1:]` does not hold true, and thus I do not think we should have them as separate inputs.

matt-gardner: Another argument for not splitting them: the direction we're going is to take a list of correct outputs, and doing constrained decoding (with different constraints at training and test time).  You don't have a per-token loss, just a loss at the end, and so you don't really want to think about things in just a "predict-the-next-token" sense.

I think the split decoder input/output makes it easy to implement the traditional seq2seq models efficiently, which is why Mark was suggesting that.  But our end goal is a bit different.

allennlp/data/dataset_readers/encoder_decoder.py
```diff
@@ -0,0 +1,91 @@
...
+        Indexers used to define output (target side) token representations.
+    """
+    def __init__(self, source_tokenizer: Tokenizer = None, target_tokenizer: Tokenizer = None,
+                 source_token_indexers: Dict[str, TokenIndexer] = None,
+                 target_token_indexers: Dict[str, TokenIndexer] = None) -> None:
+        self._source_tokenizer = source_tokenizer or WordTokenizer()
+        self._target_tokenizer = target_tokenizer or WordTokenizer(start_tokens=[START_SYMBOL],
+                                                                   end_tokens=[END_SYMBOL])
+        self._source_token_indexers = source_token_indexers or {"tokens": SingleIdTokenIndexer("source_tokens")}
+        self._target_token_indexers = target_token_indexers or {"tokens": SingleIdTokenIndexer("target_tokens")}
```
pdasigi: @matt-gardner I believe you were unsure about the way I used namespaces here. Do you have better suggestions?
Once that is resolved, do you think this PR ready to be merged?

matt-gardner: I think it _might_ make sense to have the `"source_tokens"` and `"target_tokens"` strings be configurable.  But now that I see that the token indexers themselves are configurable, you might not need to do this.

I haven't looked at the code in detail yet, because it was marked [WIP], and you didn't request a review.  I'll review it today.

pdasigi: Yeah, when I submitted this PR, I didn't have concrete plans for what's going in it. But you recently said that it may be a good idea to merge this and open a new PR for the constrained stuff, and I agree. So yeah, please review it when you get a chance.

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,274 @@
...
+            input_weights = self._decoder_attention(last_decoder_output, encoder_outputs)
+            # (batch_size, enc_output_dim)
+            attended_input = weighted_sum(encoder_outputs, input_weights)
+            # (batch_size, enc_output_dim + enc_embedding_dim)
+            return torch.cat((attended_input, embedded_input), -1)
+        else:
+            return embedded_input
+
+    def _predict_step(self, # pylint: disable=no-self-use
+                      target_scores: torch.LongTensor) -> Dict[str, torch.LongTensor]:
```
matt-gardner: Not sure the dictionary return type is buying you anything here over a simple tuple.

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,274 @@
...
+        """
+        predicted_indices = output_dict["predictions"]
+        if not isinstance(predicted_indices, numpy.ndarray):
+            predicted_indices = predicted_indices.cpu().numpy()
+        all_predicted_tokens = []
+        for indices in predicted_indices:
+            indices = list(indices)
+            # Collect indices till the first end_symbol
+            indices = indices[:indices.index(self._end_index)]
+            predicted_tokens = [self.vocab.get_token_from_index(x, namespace="output_tokens")
```
matt-gardner: This string doesn't match what's above.  Looks like you need a test for this.  And it should just be `self._target_namespace`, anyway.

tests/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,12 @@
+# pylint: disable=invalid-name
+from allennlp.common.testing import ModelTestCase
+
+
+class EncoderDecoderTest(ModelTestCase):
```
matt-gardner: This test isn't getting run, probably because the filename doesn't contain `test`.  Also, can you add a test for the dataset reader?

allennlp/data/dataset_readers/encoder_decoder.py
```diff
@@ -0,0 +1,91 @@
...
+@DatasetReader.register("encoder_decoder")
+class EncoderDecoderDatasetReader(DatasetReader):
+    """
+    Read a tsv file containing paired sequences, and create a dataset for an ``EncoderDecoder`` model.
+
+    Parameters
+    ----------
+    source_tokenizer : ``Tokenizer``, optional (default=``WordTokenizer()``)
+        Tokenizer to use to split the input sequences into words or other kinds of tokens.
+    target_tokenizer : ``Tokenizer``, optional (default=``WordTokenizer(start_tokens=[START_SYMBOL],
```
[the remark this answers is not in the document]
pdasigi: But we need start and end tokens on the target side, and not on the source side.

allennlp/data/dataset_readers/encoder_decoder.py
```diff
@@ -0,0 +1,91 @@
...
+                                                                        end_tokens=[END_SYMBOL])``)
+        Tokenizer to use to split the output sequences (during training) into words or other kinds of tokens.
+    source_token_indexers : ``Dict[str, TokenIndexer]``, optional
+                             (default=``{"tokens": SingleIdTokenizer("input_tokens")}``)
+        Indexers used to define input (source side) token representations.
+    target_token_indexers : ``Dict[str, TokenIndexer]``, optional
+                            (default=``{"tokens": SingleIdTokenizer("output_tokens")}``)
```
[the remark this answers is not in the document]
pdasigi: I don't think this is related to the training issue. We have two embeddings, one on the source side an another on the target side, and the vocabularies are not shared.

allennlp/data/dataset_readers/encoder_decoder.py
```diff
@@ -0,0 +1,91 @@
...
+        Indexers used to define output (target side) token representations.
+    """
+    def __init__(self, source_tokenizer: Tokenizer = None, target_tokenizer: Tokenizer = None,
+                 source_token_indexers: Dict[str, TokenIndexer] = None,
+                 target_token_indexers: Dict[str, TokenIndexer] = None) -> None:
+        self._source_tokenizer = source_tokenizer or WordTokenizer()
+        self._target_tokenizer = target_tokenizer or WordTokenizer(start_tokens=[START_SYMBOL],
```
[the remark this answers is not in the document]
pdasigi: Yes, that makes sense.

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,274 @@
...
+        self._start_index = self.vocab.get_token_index(START_SYMBOL, "target_tokens")
+        self._end_index = self.vocab.get_token_index(END_SYMBOL, "target_tokens")
+        self._num_classes = self.vocab.get_vocab_size("target_tokens")
+        # Decoder output dim needs to be the same as the encoder output dim since we initialize the hidden
+        # state of the decoder with that of the final hidden states of the encoder. Also, if we're using
+        # attention with ``DotProductSimilarity``, this is needed.
+        self._decoder_output_dim = self._encoder.get_output_dim()
+        # TODO (pradeep): target_embedding_dim need not be the same as the source embedding dim.
+        target_embedding_dim = self._source_embedder.get_output_dim()
+        self._target_embedder = Embedding(self._num_classes, target_embedding_dim)
```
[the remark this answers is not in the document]
pdasigi: Yes, there should be a way to let both the embeddings share weights. As a default choice, I think keeping them separate is fine because I cannot think of many cases where they should obviously be shared (summarization maybe?).
Also, I wouldn't say my toy example didn't work well. It went to a decently low NLL, and when I looked at the decoded sequences, most of them were correctly copied. Yes, it was not exactly a copy task, because it was essentially mapping symbols from the source side to a *different* set of symbols on the target side, but I am not sure if sharing the embeddings would have necessarily made it easier.

allennlp/data/dataset_readers/encoder_decoder.py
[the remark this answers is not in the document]
matt-gardner: Yes, which is why we add the start and end tokens separately, as I said in a different comment.

allennlp/models/encoder_decoders/encoder_decoder.py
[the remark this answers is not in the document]
matt-gardner: Ok, sounds good.

allennlp/data/dataset_readers/encoder_decoder.py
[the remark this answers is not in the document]
pdasigi: Yes, agreed. Saw that comment after I responded to this one.

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,275 @@
...
+            # input token from timestep i + 1. That is, the comparison between logits and targets should be
+            # one-off.
+            # Consider a single example where the target has 3 words, and padding is to 7 tokens. The complete
+            #   sequence would correspond to <S> w1 w2 w3 <E> <P> <P>
+            #   and the mask would be         1  1  1  1   1   0   0
+            #   and let the logits be         l1 l2 l3 l4  l5  l6  l7
+            # We actually need to compare:
+            #   the sequence w1 w2 w3 <E> <P> <P>
+            #   with masks   1  1  1   1   0   0
+            #   against      l1 l2 l3  l4  l5  l6
```
DeNeutoy: This should be `l2 l3  l4  l5  l6 l7`, I think?

matt-gardner: No, this is correct.  You really want to have four lines here, where the new line shows the decoder _inputs_.  This would be more clear.  The first column is then: `input: <S>, target: w1, logit: l1, mask: 1`.

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,274 @@
...
+            num_decoding_steps = self._max_decoding_steps
+        # TODO (pradeep): Define a DecoderState class?
+        decoder_hidden = final_encoder_output
+        with torch.cuda.device_of(decoder_hidden):
+            decoder_context = Variable(torch.zeros(1, self._decoder_output_dim))
+        last_predictions = None
+        step_logits = []
+        step_probabilities = []
+        step_predictions = []
+        for timestep in range(num_decoding_steps):
```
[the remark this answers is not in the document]
pdasigi: That could break the loss computation during training since it might make the lengths of logits and relevant targets different.

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,274 @@
...
+            if self.training and all(torch.rand(1) >= self._scheduled_sampling_ratio):
```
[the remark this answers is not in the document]
pdasigi: Hmm.. we'll need to gather appropriate rows from predictions and targets and reconstruct a batch. Not sure if we'll actually gain anything by doing that.

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,274 @@
...
+            Dict with indices of target side tokens, of shape ``(batch_size, target_sequence_length)``
+            in ``tokens`` field.
+        """
+        # (batch_size, input_seq_length, enc_output_dim)
+        encoder_outputs = self._get_encoder_outputs(source_tokens)
+        final_encoder_output = encoder_outputs[:, -1]  # (batch_size, enc_output_dim)
+        if target_tokens:
+            targets = target_tokens["tokens"]
+            target_sequence_length = targets.size()[1]
+            num_decoding_steps = target_sequence_length
```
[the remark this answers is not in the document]
pdasigi: Good point!

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,274 @@
...
+                                                                 (decoder_hidden, decoder_context))
+            # (batch_size, num_classes)
+            output_projections = self._output_projection_layer(decoder_hidden)
+            # (batch_size, 1, num_classes)
+            step_logits.append(output_projections.unsqueeze(1))
+            decoder_predictions = self._predict_step(output_projections)
+            step_probabilities.append(decoder_predictions["class_probabilities"].unsqueeze(1))
+            last_predictions = decoder_predictions["predicted_classes"]
+            # (batch_size, 1)
+            step_predictions.append(last_predictions.view(-1, 1))
```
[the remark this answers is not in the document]
pdasigi: Right. But I assume you meant `unsqueeze`.

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,274 @@
...
+        return output_dict
+
+    def _get_encoder_outputs(self, source_tokens: Dict[str, torch.LongTensor]) -> torch.LongTensor:
+        embedded_input = self._source_embedder(source_tokens)
+        source_mask = get_text_field_mask(source_tokens)
+        encoded_text = self._encoder(embedded_input, source_mask)
+        return encoded_text
+
+    def _prepare_decode_step_input(self, input_indices: torch.LongTensor,
+                                   last_decoder_output: torch.LongTensor = None,
```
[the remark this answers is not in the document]
pdasigi: Yes, that's better.

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,274 @@
...
+        source_tokens : Dict[str, torch.LongTensor]
+           The output of ``TextField.as_array()`` applied on the input ``TextField``. This will be passed
+           through a ``TextFieldEmbedder`` and then through an encoder.
+        target_tokens : Dict[str, torch.LongTensor], optional (default = None)
+            Dict with indices of target side tokens, of shape ``(batch_size, target_sequence_length)``
+            in ``tokens`` field.
+        """
+        # (batch_size, input_seq_length, enc_output_dim)
+        encoder_outputs = self._get_encoder_outputs(source_tokens)
+        final_encoder_output = encoder_outputs[:, -1]  # (batch_size, enc_output_dim)
```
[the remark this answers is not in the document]
DeNeutoy: I can't remember off the top of my head - if it is wrong, the fix is to gather along the second dimension after getting the lengths from the mask.

allennlp/models/encoder_decoders/encoder_decoder.py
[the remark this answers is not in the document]
matt-gardner: You could just use the last value of `timestep` to index `targets`, like `targets[:, 1:timestep]`, or something.  It's not a huge deal, though.  This loop will look pretty different with a constrained decoder, anyway.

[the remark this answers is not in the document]
pdasigi: We shouldn't ignore parts of the actual targets in loss computation, right? Say the model's prediction is 2 incorrect tokens followed by an end symbol. It should be penalized differently in a case where the target sequence is 2 tokens long vs. a case where it is 10 tokens long.

allennlp/data/dataset_readers/encoder_decoder.py
```diff
@@ -0,0 +1,91 @@
...
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+START_SYMBOL = "@@START@@"
+END_SYMBOL = "@@END@@"
+
+@DatasetReader.register("encoder_decoder")
+class EncoderDecoderDatasetReader(DatasetReader):
+    """
+    Read a tsv file containing paired sequences, and create a dataset for an ``EncoderDecoder`` model.
```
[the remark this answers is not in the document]
pdasigi: Done

pdasigi:
@matt-gardner I think this pretty much addresses all your earlier comments. The major changes are using default namespaces, and additional tests.

allennlp/models/encoder_decoders/encoder_decoder.py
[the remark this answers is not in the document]
matt-gardner: You don't need a gather, you can do it reasonably efficiently by doing something like:

```python
sampling_mask = torch.rand(batch_size) >= self._scheduled_sampling_ratio
forcing_input = targets[:, timestep]
sampling_input = ...  # what's currently in the `else` clause
input_choices = sampling_mask * forcing_input + (1 - sampling_mask)  * sampling_input
```

This would do a better job of actually mixing according to the sampling schedule you set.

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,320 @@
...
+            else:
+                if timestep == 0:
+                    # For the first timestep, when we do not have targets, we input start symbols.
+                    # (batch_size,)
+                    with torch.cuda.device_of(final_encoder_output):
+                        input_choices = Variable(torch.LongTensor([self._start_index]).expand_as(
+                                final_encoder_output[:, 0]))
+                else:
+                    input_choices = last_predictions
+            decoder_input = self._prepare_decode_step_input(input_choices, decoder_hidden,
```
matt-gardner: What do you think of `_get_decoder_input()` as a function name?

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,320 @@
...
+            output_projections = self._output_projection_layer(decoder_hidden)
+            # list of (batch_size, 1, num_classes)
+            step_logits.append(output_projections.unsqueeze(1))
+            decoder_predictions = self._predict_step(output_projections)
+            step_probabilities.append(decoder_predictions["class_probabilities"].unsqueeze(1))
+            last_predictions = decoder_predictions["predicted_classes"]
+            # (batch_size, 1)
+            step_predictions.append(last_predictions.unsqueeze(1))
+        # step_logits is a list containing tensors of shape (batch_size, 1, num_classes)
+        # This is (batch_size, num_decoding_steps-1, num_classes)
```
matt-gardner: Isn't it `(batch_size, num_decoding_steps, num_classes)`?  Where'd the `-1` come from?

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,320 @@
...
+                    input_choices = last_predictions
+            decoder_input = self._prepare_decode_step_input(input_choices, decoder_hidden,
+                                                            encoder_outputs, source_mask)
+            decoder_hidden, decoder_context = self._decoder_cell(decoder_input,
+                                                                 (decoder_hidden, decoder_context))
+            # (batch_size, num_classes)
+            output_projections = self._output_projection_layer(decoder_hidden)
+            # list of (batch_size, 1, num_classes)
+            step_logits.append(output_projections.unsqueeze(1))
+            decoder_predictions = self._predict_step(output_projections)
```
matt-gardner: `_predict_step` is a softmax to get probabilities, and a max to get predictions, and you have a method call with a complex dictionary and a long docstring where really you just need two lines here.  I know you did that for extensibility later, but I don't think you can do constrained decoding by overriding this method.  I'd probably just put the two lines in here, and remove the `_predict_step` function entirely.

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,320 @@
...
+        # appropriate comparison.
+        # Consider a single example where the target has 3 words, and padding is to 7 tokens. The complete
+        #   sequence would correspond to <S> w1 w2 w3 <E> <P> <P>
+        #   and the mask would be         1  1  1  1   1   0   0
+        #   and let the logits be         l1 l2 l3 l4  l5  l6
+        # We actually need to compare:
+        #   the sequence w1 w2 w3 <E> <P> <P>
+        #   with masks   1  1  1   1   0   0
+        #   against      l1 l2 l3  l4  l5  l6
+        relevant_targets = targets[:, 1:].contiguous()  # (batch_size, num_decoding_steps - 1)
```
matt-gardner: These `-1`s are wrong.

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,320 @@
...
+        """
+        Takes logits (unnormalized outputs from the decoder) of size (batch_size, num_decoding_steps - 1,
+        num_classes), target indices of size (batch_size, num_decoding_steps) and corresponding masks of size
+        (batch_size, num_decoding) steps and computes cross entropy loss while taking the mask into account.
+
+        The length of ``targets`` is expected to be greater than that of ``logits`` because the decoder does
+        not need to compute the output corresponding to the last timestep of ``targets``. This method aligns
+        the inputs appropriately to compute the loss.
+        """
+        # `targets` currently contains START + target_sequence + END, and logits correspond to the outputs
```
matt-gardner: I'd probably integrate this comment into the docstring, now that you have a separate method for this.  You've got some duplication between the docstring and the comment.

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,320 @@
...
+        #   against      l1 l2 l3  l4  l5  l6
+        relevant_targets = targets[:, 1:].contiguous()  # (batch_size, num_decoding_steps - 1)
+        relevant_mask = target_mask[:, 1:].contiguous()  # (batch_size, num_decoding_steps - 1)
+        loss = sequence_cross_entropy_with_logits(logits, relevant_targets, relevant_mask)
+        return loss
+
+    @overrides
+    def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
+        """
+        Trims the output predictions to the first end symbol, replaces indics with corresponding tokens,
```
matt-gardner: The name of this method clashes with the "decoder" of `EncoderDecoder`.  You implemented this correctly, but I think a reader could easily get confused about what this method does, given its name.  It'd be helpful to have a comment in the docstring that says something like "`Model.decode` is for doing inference or finalizing predictions after a call to `Model.forward`, and is separate from the notion of "decoding" in our `EncoderDecoder` model.  That decoder lives in `forward`; this `decode` method just cleans up the output and converts it to strings."

allennlp/models/encoder_decoders/encoder_decoder.py
```diff
@@ -0,0 +1,320 @@
...
+        target_namespace = params.pop("target_namespace", "tokens")
+        # If no attention function is specified, we should not use attention, not attention with default similarity
+        # function.
+        attention_function_type = params.pop("attention_function", None)
+        if attention_function_type is not None:
+            attention_function = SimilarityFunction.from_params(attention_function_type)
+        else:
+            attention_function = None
+        scheduled_sampling_ratio = params.pop("scheduled_sampling_ratio", 0.0)
+        return cls(vocab, source_embedder=source_embedder,
```
matt-gardner: I'd recommend one argument per line here.

allennlp/data/dataset_readers/encoder_decoder.py
```diff
@@ -0,0 +1,105 @@
...
+    def __init__(self,
+                 source_tokenizer: Tokenizer = None,
+                 target_tokenizer: Tokenizer = None,
+                 source_token_indexers: Dict[str, TokenIndexer] = None,
+                 target_token_indexers: Dict[str, TokenIndexer] = None) -> None:
+        self._source_tokenizer = source_tokenizer or WordTokenizer()
+        self._target_tokenizer = target_tokenizer or self._source_tokenizer
+        self._source_token_indexers = source_token_indexers or {"tokens": SingleIdTokenIndexer()}
+        self._target_token_indexers = target_token_indexers or self._source_token_indexers
+        self._token_indexers = {"source": self._source_token_indexers,
```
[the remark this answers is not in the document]
pdasigi: `ModelTestCase.setUpModel` expects this field to be set: https://github.com/allenai/allennlp/blob/master/allennlp/common/testing/model_test_case.py#L30

Also, I thought it would be nice to keep things consistent across dataset readers.

allennlp/models/encoder_decoders/encoder_decoder.py
pdasigi: The way I'm thinking about this is that our semantic parsing plan involves two additional things: decoding with constraints, which is done by the `_predict_step` taking a set of constraints for the next prediction and doing masked softmax appropriately, and a `DecoderState` which maintains a queue of non-terminals, chooses one at each timestep and passes the appropriate constraints to `_predict_step`. That way, the first part can be achieved by overriding this method.

allennlp/models/encoder_decoders/encoder_decoder.py
pdasigi: Yes, you're right. I'll fix it.

allennlp/data/dataset_readers/encoder_decoder.py
```diff
@@ -0,0 +1,105 @@
...
+            tokenized_target.append(Token(END_SYMBOL))
+            target_field = TextField(tokenized_target, self._target_token_indexers)
+            return Instance({"source_tokens": source_field, "target_tokens": target_field})
+        else:
+            return Instance({'source_tokens': source_field})
+
+    @classmethod
+    def from_params(cls, params: Params) -> 'EncoderDecoderDatasetReader':
+        source_tokenizer_type = params.pop('source_tokenizer', None)
+        source_tokenizer = None if source_tokenizer_type is None else Tokenizer.from_params(source_tokenizer_type)
```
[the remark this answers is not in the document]
pdasigi: I see that there are two kinds of defaults. When a choice is not specified in the json file, the classes processing those params have two options: either make a default choice themselves, or having the class relevant to the choice (like `Tokenizer`) make a default choice. I think what the `Tokenizer` is doing is fine. But it would be nice to have a cleaner way of doing the former as well.

allennlp/models/encoder_decoders/encoder_decoder.py
pdasigi: Hmm.. `decoder_input` does not distinguish the embedded vector from the indices (not that the current name does a much better job). Also, this is just for one timestep.

pdasigi:
Thanks for the review. I addressed all the documentation related comments. Please see my comment above for `_predict_step`. I may make some changes related to this in my next PR with a `DecoderState` anyway.
And regarding scheduled sampling, I'm not sure if one of per-batch vs per-example sampling is better than the other. the paper doesn't clearly say which one they did. I tried both on the copy dataset, and did not see a significant difference in validation error, although the error in per-example variant varied much more across epochs (not sure if this trend would generalize though). I think I'll keep it at the batch level for now.

PR #370, opened 2017-10-03
allow empty sequences as inputs to wrapped seq2vec encoders

joelgrus:
* I feel like a number of these steps are clunky, is there a cleaner way to accomplish the same?
* should we do something similar for seq2seq encoders?


joelgrus:
added seq2seq version, can you take another look

PR #374, opened 2017-10-05
[WIP / experimental] type stubs for pytorch + fixing Variable/Tensor conclusion

joelgrus:
this was an interesting exercise, but it ended up being kind of a mess, and I'm not convinced I want to merge it. some thoughts:

* this is not a full (or even wholly correct) set of stubs, it's more of the minimal set I needed to get `allennlp` to type-check
* in particular, that `Module.forward` can take any arguments whatsoever is a real challenge
* I know it's on the pytorch roadmap to unify `Tensor` and `Variable`, I suspect that all of this will be easier / cleaner once that happens
* I needed to use a handful of `cast`s, which I hate (some of them I could have eliminated if we didn't do the `p>0` check for dropout, which we don't *have* to do, dropout with `p=0` seems to do an unnecessary copy but is mostly written to no-op)
* there are a handful of TODOs where I used really bad type signatures
* I had to rewrite `arrays_to_variables` to make it more type-friendly. (I didn't have to, but otherwise I needed to add a ton of `cast`s all over the place)
* some of the changes lose information, e.g. `LongTensor` -> `Variable`
* I learned a fair bit of new mypy stuff that I never knew before!


allennlp/common/testing/model_test_case.py
```diff
@@ -57,7 +57,7 @@ def ensure_model_can_train_save_and_load(self, param_file: str):
         loaded_dataset = reader.read(params['validation_data_path'])
         loaded_dataset.index_instances(loaded_model.vocab)
         loaded_batch_arrays = next(iterator(loaded_dataset, shuffle=False))
-        loaded_batch = arrays_to_variables(loaded_batch_arrays, for_training=False)
+        loaded_batch = arrays_to_variables(loaded_batch_arrays, for_training=True)
```
matt-gardner: Why this change?

malmaud:
Variables and tensors have now merged on pytorch master, FWIW.

PR #375, opened 2017-10-05
add optional test data to train

DeNeutoy:
It was annoying to add the test vocab to the vocabulary creation, so I made it possible to read the test data like we do for the other datasets. Given that we load it in, I figured it would make sense to call `evaluate` on it if a user passed it.

### DIFF
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -158,14 +158,20 @@ def train_model(params: Params, serialization_dir: str) -> Model:
                                   train_data,
                                   validation_data,
                                   trainer_params)
+
+    evaluate_on_test = params.pop("evaluate_on_test", False)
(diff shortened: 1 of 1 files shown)


matt-gardner:
1. I don't think it's our job to encode in our framework a particular way of doing science.  It's up to the scientists to do the right thing.  But if we don't have a reasonable way of evaluating a final model on test data, there's a problem.

2. If there are words in the test vocabulary that don't appear in the training data, their embeddings will not be trained, so it isn't really possible to "cheat" with this.  What it _does_ make easy, though, is including test words that appear in glove in your embedding layer, or assign them unique random vectors, if that's what you want to do.  Both of these are perfectly reasonable strategies that are hard without this PR.

PR #380, opened 2017-10-09
don't pull cuda runtime twice

### LINKED ISSUE #372: allennlp docker image contains 2 versions of cuda runtime libraries and cudnn
e.g.
```
root@d5c40706fdb9:/stage# find / -name libcudart"*" | xargs ls -lrt
-rw-r--r-- 1 root root 415432 Jan 26  2017 /usr/local/cuda-8.0/targets/x86_64-linux/lib/libcudart.so.8.0.61
lrwxrwxrwx 1 root root     19 Jan 26  2017 /usr/local/cuda-8.0/targets/x86_64-linux/lib/libcudart.so.8.0 -> libcudart.so.8.0.61
lrwxrwxrwx 1 root root     16 Jan 26  2017 /usr/local/cuda-8.0/targets/x86_64-linux/lib/libcudart.so -> libcudart.so.8.0
-rwxr-xr-x 1 root root 424600 Oct  2 20:17 /opt/conda/lib/python3.6/site-packages/torch/lib/libcudart-5d6d23a3.so.8.0.61
```
When pytorch is installed via pip, it pulls its own cuda runtime and cudnn,  there's no need to start with cuda - cudnn base docker image, see e.g. runtime Dockerfile in pytorch repo https://github.com/pytorch/pytorch/blob/master/tools/docker/Dockerfile_runtime

DeNeutoy:
fixes #372 and shaves about 0.8GB off our image size 👍.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,12 @@
-FROM nvidia/cuda:8.0-cudnn5-runtime
+FROM ubuntu:16.04
 
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
-ENV PATH /opt/conda/bin:$PATH
+# Add nvidia and conda binaries to path.
+ENV PATH /usr/local/nvidia/bin/:/opt/conda/bin:$PATH
 ENV PYTHONHASHSEED 2157
 
-# Override Nvidia's default LD paths, since they're misconfigured in the base image.
-ENV LD_LIBRARY_PATH /usr/local/cuda-8.0/lib64:$LD_LIBRARY_PATH
+ENV LD_LIBRARY_PATH /usr/local/nvidia/lib64:$LD_LIBRARY_PATH
 
 WORKDIR /stage
 
@@ -56,6 +56,7 @@ COPY tests/ tests/
 COPY pytest.ini pytest.ini
 COPY scripts/ scripts/
 COPY tutorials/ tutorials/
+COPY experiment_config experiment_config/
 
 # Run tests to verify the Docker build
 RUN PYTHONDONTWRITEBYTECODE=1 pytest


schmmd:
@DeNeutoy this works on a GPU machine?  It would be fantastic if we could use ubuntu as a base image.  If we can really do this, once this is merged I'll look into using a Python base image and removing `conda`.

schmmd:
@DeNeutoy this works on a GPU machine?  It would be fantastic if we could use ubuntu as a base image.  If we can really do this, once this is merged I'll look into using a Python base image and removing `conda`.

schmmd:
I just read #372--fantastic!

DeNeutoy:
Yes, I have trained models with an image generated from this dockerfile - it definitely works.

schmmd:
👍 👍 

Fantastic to get off the nvidia base image.

@splittingfield @ckarenz FYI in case you can similarly simplify.

DeNeutoy:
I doubt this is broadly applicable because tensorflow doesn't install cuda libraries for you (the last time I used tensorflow, anyway).

schmmd:
Sure--that's why we have this configuration in the first place.  But more and more Aristo-ites are moving over to PyTorch and AllenNLP as far as I know.  I'd hate for a dependency on the nvidia Docker images to propagate...

schmmd:
All checks passed so I went ahead and merged this.  I'm going to try to simplify our setup further now that we no longer depend on an nvidia base image.

Dockerfile
```diff
@@ -1,12 +1,12 @@
...
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
-ENV PATH /opt/conda/bin:$PATH
+# Add nvidia and conda binaries to path.
+ENV PATH /usr/local/nvidia/bin/:/opt/conda/bin:$PATH
 ENV PYTHONHASHSEED 2157
 
-# Override Nvidia's default LD paths, since they're misconfigured in the base image.
-ENV LD_LIBRARY_PATH /usr/local/cuda-8.0/lib64:$LD_LIBRARY_PATH
+ENV LD_LIBRARY_PATH /usr/local/nvidia/lib64:$LD_LIBRARY_PATH
```
ngimel: Here LD_LIBRARY_PATH is likely empty, and pre-pending to empty library path is a bad practice, you might end up inadvertently adding your working dir to your $LD_LIBRARY_PATH, see https://joshumax.github.io/general/2017/06/08/how-torch-broke-ls.html or https://patchwork.ozlabs.org/patch/413215/

DeNeutoy: Thanks a lot for reporting originally and this review 👍

schmmd: It is empty.  I will remove the reference to `$LD_LIBRARY_PATH`.  Thanks!

PR #385, opened 2017-10-09
don't append to empty LD_LIBRARY_PATH

DeNeutoy:


schmmd:
I've already done the same in a parallel PR ;-)

schmmd:
Already merged in a parallel PR.

PR #389, opened 2017-10-09
Use a Python base image.

schmmd:
* Remove conda.
* Use Debian instead of Ubuntu.

This trims another 200 MB from our Docker file.

```
~/h/a/allennlp (python-docker $%) docker images                                                                                                                                                                                     (allennlp) 
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
allennlp/allennlp   python-jesse        74f6e1cbc41e        36 minutes ago      2.45GB
allennlp/allennlp   latest              547826cf4c97        About an hour ago   2.64GB
python              3.6.3-jessie        be512ebcbac9        5 days ago          690MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,8 @@
-FROM ubuntu:16.04
+FROM python:3.6.3-jessie
 
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
-# Add nvidia and conda binaries to path.
-ENV PATH /usr/local/nvidia/bin/:/opt/conda/bin:$PATH
+ENV PATH /usr/local/nvidia/bin/:$PATH
 ENV PYTHONHASHSEED 2157
 
 ENV LD_LIBRARY_PATH /usr/local/nvidia/lib64:$LD_LIBRARY_PATH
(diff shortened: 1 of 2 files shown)


schmmd:
@DeNeutoy @joelgrus any objections?

I still need to try this image on a GPU.  @DeNeutoy what test do you run to validate Docker images on GPU machines?  I would probably just run an evaluation with GPUs enabled.

joelgrus:
as long as it works, I have no objections 😀

DeNeutoy:
@schmmd I just train one of the models for a few epochs and check it's using the GPU. You can do this via kube if you change all the absolute paths from beaker to `/net/efs/aristo/dlfa/...`.

2 questions:
- Why debian instead of ubuntu (not a problem I just don't know)?
- My concern with not using the miniconda installation route is that we have 2 ways to install with python (directly via docker and indirectly via conda if a user isn't using docker), which might create more work for us.

schmmd:
@DeNeutoy -- Debian is smaller than Ubuntu by a 100 or 200 MB and has everything we need.  Python doesn't build a base image with Ubuntu.  I believe it's discouraged because of how large the Ubuntu base image is.

You have a reasonable concern about having multiple ways to install Python.  I favor this change regardless because:

* This simplifies our Docker build and makes it easier to troubleshoot issues.
* This allows us to pin the exact version of Python.
* We've had problems with `conda` in our Docker build in the past.
* `conda` and Docker are redundant technologies.
* Our Docker build is smaller by 200 MB.

schmmd:
I would also add that I don't trust conda enough to say that our two solutions (using conda on Mac desktop and using conda on a Ubuntu Docker image) are consistent enough to avoid problems.

schmmd:
Evaluations work and use a GPU.

PR #392, opened 2017-10-10
Add caching to docker build process

### LINKED ISSUE #306: docker build should cache the latest models
right now they have to be downloaded every time you start up a docker container

grahamannett:
I saw the issue for caching latest model: #306 and thought it seemed pretty simple and hadn't been done so added ARG to dockerfile which would allow something like:  
`docker build --build-arg CACHE_MODELS=true --tag allennlp/allennlp . ` 
to download/cache models during build otherwise it doesn't do anything.  Possible to incorporate it elsewhere since the script file being super simple might be a bit superfluous. 

Also not sure if this is worth documenting somewhere else.

Fixes #306


schmmd:
@grahamannett thanks!  Would you be able to rename your script from `docker_cache.sh` to `cache_models.sh`?  Then I'm happy to approve.

grahamannett:
@schmmd sure

schmmd:
Thanks @grahamannett !

PR #395, opened 2017-10-11
Rename EncoderDecoder to SimpleSeq2Seq

matt-gardner:
This is so that anyone wanting to build a sequence-to-sequence model in AllenNLP has an obvious simple starting place.  I also copied the `EncoderDecoder` model into a new `WikiTablesSemanticParser` model, which we'll be modifying as we add in various components that the original parser had.  So this effectively takes what was `EncoderDecoder` and makes two classes, one that is simpler (I made some very small changes to simplify the code slightly, and added some comments in the main docstring), and one that will soon get a whole lot more complicated.


allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -11,53 +11,57 @@
...
+    logical form is generated by a `type-constrained`, `transition-based` parser.  This is a
+    re-implementation of the model used for the paper `Neural Semantic Parsing with Type
+    Constraints for Semi-Structured Tables
+    <https://www.semanticscholar.org/paper/Neural-Semantic-Parsing-with-Type-Constraints-for-Krishnamurthy-Dasigi/8c6f58ed0ebf379858c0bbe02c53ee51b3eb398a>`_,
+    by Jayant Krishnamurthy, Pradeep Dasigi, and Matt Gardner (EMNLP 2017).
 
-    In this implementation, the decoder uses the encoder's outputs in two ways. The hidden state of the decoder
-    is intialized with the output from the final time-step of the encoder, and when using attention, a weighted
-    average of the outputs from the encoder is concatenated to the inputs of the decoder at every timestep.
+    WORK STILL IN PROGRESS.  This is just copying the SimpleSeq2Seq model for now, and we'll
```
pdasigi: Since this is work in progress, do we want this in the master branch?

matt-gardner: Yeah, good point, I'll split it out into a separate branch.

matt-gardner:
Ok, @pdasigi, does it look ok now?

PR #398, opened 2017-10-11
Use bmm to do an efficient weighted sum in cases where it applies

matt-gardner:
Following points made by @kentonl in #364.

I also tried to switch `masked_softmax` to just use `vector += mask.log(); return F.softmax(vector)`.  I ran into trouble there, because it doesn't handle cases where the mask is zero everywhere (or zero for all entries in a row in the matrix, e.g., because a whole row is padded), which is important for some of our uses.  There didn't seem to be an easy way to fix that without just doing what we're already doing.  Maybe I'll figure something out, but I'm not optimistic.


DeNeutoy:
Yeah that case seems non-trivial to fix - I guess you could check for rows of the mask which are all zero before taking the log, but then what we do in that case is still a little shaky. 

LGTM - Looks like a floating point error in the docker container builds..... I guess just merge?

PR #400, opened 2017-10-12
Move cache_models.sh to a Python script.

schmmd:


scripts/cache_models.py
```diff
@@ -0,0 +1,18 @@
...
+
+import os
+import sys
+
+sys.path.insert(0, os.path.dirname(os.path.abspath(os.path.join(__file__, os.pardir))))
+from allennlp.commands import DEFAULT_MODELS
+from allennlp.common.file_utils import cached_path
+
+value = os.environ.get('CACHE_MODELS')
+if isinstance(value, str) and value.lower() == "true":
```
joelgrus: it's a bad smell to use `isinstance` when you don't have to. maybe instead

```
cache_models = os.environ.get('CACHE_MODELS', 'false').lower() == 'true'
if cache_models:
  ...
```

scripts/cache_models.py
```diff
@@ -0,0 +1,18 @@
...
+    urls = DEFAULT_MODELS.values()
+    print("CACHE_MODELS is '%s'.  Downloading %i models." % (value, len(urls)))
+    for i, url in enumerate(urls):
```
joelgrus: if you're going to print them out, you might as well print the key/model-names too

```
for i, (model_name, url) in enumerate(DEFAULT_MODELS.items()):
  ...
```

scripts/cache_models.py
schmmd: Good call, thank you!

schmmd:
@joelgrus look good now?

scripts/cache_models.py
```diff
@@ -0,0 +1,18 @@
...
+
+value = os.environ.get('CACHE_MODELS', 'false')
+if value.lower() == "true":
+    models = DEFAULT_MODELS.items()
+    print("CACHE_MODELS is '%s'.  Downloading %i models." % (value, len(models)))
+    for i, (model, url) in enumerate(models):
+        print("Downloading '%s' model from %s" % (model, url))
+        print("Saved at %s" % cached_path(url))
+else:
+    print("CACHE_MODELS is '%s'.  Not caching models." % value)
```
joelgrus: I mean, this is sort of misleading, since you're printing "false" even if it's not set. I'm not sure we care though.

schmmd: @joelgrus, if it's not set then I print `CACHE_MODELS is ''`.

PR #401, opened 2017-10-12
Remove param method from logger, and just use info.

schmmd:


schmmd:
@DeNeutoy is there a reason we want params logged at a level above WARN?  It seems easiest to just log them at info--then they can be turned off as the user expects rather than understanding our custom logging level of PARAMS.

DeNeutoy:
Not logging parameters means that you can't recreate the exact configuration used because default arguments hard-coded in the library might change - this way, the parameters are always logged, even when you want to ignore warnings. I remember @matt-gardner being pretty keen for this, so I guess he should have a say.

schmmd:
Was the original motivation behind making sure the parameters were recorded when training a model?

DeNeutoy:
Yeah exactly.

matt-gardner:
This was for reproducibility.  Before the days of beaker, it was really easy to lose track of what experiment used what parameters, and so logging all of them seemed like a good idea.  I still think we should log all of them, including default values, though I'm not opposed to switching it to just `info`.  The benefit of a separate logging level is that it's theoretically easy to grep the log file for parameter lines, and with some simple shell commands you could create a hocon file that would work to reproduce the experiment.  We could get the same thing by prepending something like `PARAM: ` to the info statement, or by looking at the class that logged the message.  The drawback of the separate logging level (in addition to what @schmmd mentioned) is that it required us to pin sphinx, because the newer version of sphinx uses python loggers in a way that treats our `PARAM` level as an error, and so `html-strict` fails to build.

schmmd:
Thanks for the context.  At the least, I want to add a helpful comment for the extra logging level.

But from Matt's most recent comment, I'm going to push a commit that keeps parameters greppable, but moves them to the `info` log level.

schmmd:
Actually, it seems easier to just `git grep allennlp.common.params`.

PR #402, opened 2017-10-12
Support overrides for params during prediction and evaluation.

schmmd:
When complete, this should fix https://github.com/allenai/allennlp/issues/244.

### DIFF
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -66,7 +66,7 @@ def _train_model_from_args(args: argparse.Namespace):
     train_model_from_file(args.param_path, args.serialization_dir, args.overrides)
 
 
-def train_model_from_file(parameter_filename: str, serialization_dir: str, overrides: str = "") -> Model:
+def train_model_from_file(parameter_filename: str, serialization_dir: str, overrides: str = None) -> Model:
     """
     A wrapper around :func:`train_model` which loads the params from a file.
 
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -53,7 +53,7 @@ def add_subparser(parser: argparse._SubParsersAction) -> argparse.ArgumentParser
                            help='directory in which to save the model and its logs')
     subparser.add_argument('-o', '--overrides',
                            type=str,
-                           help='a sequence of overrides to the training configuration')
+                           help='a HOCON structure used to override the experiment configuration')
     subparser.set_defaults(func=_train_model_from_args)
 
     return subparser
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -53,6 +53,7 @@ def add_subparser(parser: argparse._SubParsersAction) -> argparse.ArgumentParser
                            help='directory in which to save the model and its logs')
     subparser.add_argument('-o', '--overrides',
                            type=str,
+                           default="",
                            help='a HOCON structure used to override the experiment configuration')
     subparser.set_defaults(func=_train_model_from_args)
 
@@ -66,7 +67,7 @@ def _train_model_from_args(args: argparse.Namespace):
     train_model_from_file(args.param_path, args.serialization_dir, args.overrides)
 
 
-def train_model_from_file(parameter_filename: str, serialization_dir: str, overrides: str = None) -> Model:
+def train_model_from_file(parameter_filename: str, serialization_dir: str, overrides: str = "") -> Model:
     """
     A wrapper around :func:`train_model` which loads the params from a file.
 
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -51,6 +51,10 @@ def add_subparser(parser: argparse._SubParsersAction) -> argparse.ArgumentParser
                            type=str,
                            required=True,
                            help='directory in which to save the model and its logs')
+    subparser.add_argument('-o', '--overrides',
+                           type=str,
+                           nargs='+',
+                           help='a sequence of overrides to the training configuration')
     subparser.set_defaults(func=_train_model_from_args)
 
     return subparser
@@ -60,10 +64,10 @@ def _train_model_from_args(args: argparse.Namespace):
     """
     Just converts from an ``argparse.Namespace`` object to string paths.
     """
-    train_model_from_file(args.param_path, args.serialization_dir)
+    train_model_from_file(args.param_path, args.serialization_dir, args.overrides)
 
 
-def train_model_from_file(parameter_filename: str, serialization_dir: str) -> Model:
+def train_model_from_file(parameter_filename: str, serialization_dir: str, overrides: List[str] = []) -> Model:
     """
     A wrapper around :func:`train_model` which loads the params from a file.
 
@@ -74,11 +78,12 @@ def train_model_from_file(parameter_filename: str, serialization_dir: str) -> Mo
     serialization_dir: str, required
         The directory in which to save results and logs.
     """
+
     # We need the python hashseed to be set if we're training a model
     ensure_pythonhashseed_set()
 
     # Load the experiment config from a file and pass it to ``train_model``.
-    params = Params.from_file(parameter_filename)
+    params = Params.from_file(parameter_filename, overrides)
     return train_model(params, serialization_dir)
 
 
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -53,7 +53,6 @@ def add_subparser(parser: argparse._SubParsersAction) -> argparse.ArgumentParser
                            help='directory in which to save the model and its logs')
     subparser.add_argument('-o', '--overrides',
                            type=str,
-                           nargs='+',
                            help='a sequence of overrides to the training configuration')
     subparser.set_defaults(func=_train_model_from_args)
 
@@ -67,7 +66,7 @@ def _train_model_from_args(args: argparse.Namespace):
     train_model_from_file(args.param_path, args.serialization_dir, args.overrides)
 
 
-def train_model_from_file(parameter_filename: str, serialization_dir: str, overrides: List[str] = []) -> Model:
+def train_model_from_file(parameter_filename: str, serialization_dir: str, overrides: str = "") -> Model:
     """
     A wrapper around :func:`train_model` which loads the params from a file.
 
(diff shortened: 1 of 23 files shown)


schmmd:
@joelgrus I'd appreciate an early look.  I wouldn't have chosen the CLI syntax as my first choice, but it was the easiest thing to do with argparse (why fight it?).

```
python allennlp/run.py train -s output experiment_config/bidaf.json --overrides '"train_data_path": "foo"'
```

Alternatively, instead of constructing a HOCON string and parsing that, I could use pyhocon's `from_dict`, which would pretty surely give more sensible errors with malformed input.  I'll poke around argparse a bit more to see whether the following is possible.

```
python allennlp/run.py train -s output experiment_config/bidaf.json --override train_data_path=' "foo"'
```

allennlp/common/params.py
```diff
@@ -203,14 +188,18 @@ def _check_is_dict(self, new_history, value):
...
-    def from_file(params_file: str) -> 'Params':
+    def from_file(params_file: str, overrides: List[str] = []) -> 'Params':
         """
         Load a `Params` object from a configuration file.
         """
         # redirect to cache, if necessary
         params_file = cached_path(params_file)
 
-        param_dict = pyhocon.ConfigFactory.parse_file(params_file)
+        override_hocon = "{" + ",".join(overrides) + "}"
```
joelgrus: yeah, I don't like this at all, let's try to think of a safer/cleaner way to do this

schmmd: What don't you like about it?  I can figure out how to use argparse and `from_dict`.

joelgrus: something more like

```
override_hocon = pyhocon.ConfigTree()
for override in overrides:
  key, value = ... # whatever you need to do here
  override_hocon.put(key, value)
```

joelgrus: I feel much better about something like this than about string.join -> parseString

schmmd: I can use `from_dict`, which will do about the same thing.  I will need to add my own parsing of the CLI arguments from argparse to separate them into key and value.  I'll use the following syntax.

```
python allennlp/run.py train -s output experiment_config/bidaf.json --overrides 'train_data_pat="foo"' 'validation_data_path="bar"'
```

As a disadvantage, this syntax is less flexible and requires more documentation.  Previously, the syntax was just "add HOCON overrides".

matt-gardner: When you're testing this, make sure you can override things that are nested a few levels in the parameters.

joelgrus: I don't care which format you use, my point was more just that glomming a bunch of stuff into a string and then parsing it smells bad to me, compared to parsing the things individually and doing something with them. for example, if one of them is malformed, you get a much better error message if you're trying to understand each one.

schmmd: @matt-gardner I'll be sure to.  I have not written any tests yet.

@joelgrus thanks--I agree we'll get much better error messages with another approach.  Thanks--I just wanted to understand your precise concern.

schmmd: @matt-gardner formal tests to come, but `dataset_reader.token_indexers.token_characters.character_tokenizer.end_tokens: [999]` overrode fine.

matt-gardner: Great, I was just concerned that a naive dictionary assignment wouldn't work for this, and wanted to be sure it was on your radar.  Glad you figured it out.

schmmd:
@joelgrus I'm still a fan of having the `--overrides` method for additional flexibility but here's a solution that seems even easier.

`env VALIDATION_DATA_PATH="dev-v1.1.json" TRAIN_DATA_PATH="train-v1.1.json" python allennlp/run.py train -s path/to/output experiment_config/bidaf.json`

The same environment variables can be set in https://github.com/schmmd/allennlp/blob/2524d67b6e3a66c0e236daec389d366d76f459e4/scripts/ai2-internal/run_with_beaker.sh.

joelgrus:
that's good for overriding a few specific values, but it's not flexible for overriding arbitrary ones?

schmmd:
Exactly (hence the additional utility of `--overrides`).

schmmd:
Retraining BiDAF.

```
env VALIDATION_DATA_PATH="https://s3-us-west-2.amazonaws.com/allennlp/datasets/mc/squad-dev-v1.1.json" TRAIN_DATA_PATH="https://s3-us-west-2.amazonaws.com/allennlp/datasets/mc/squad-train-v1.1.json" GLOVE_PATH="https://s3-us-west-2.amazonaws.com/allennlp/datasets/glove/glove.6B.100d.txt" python allennlp/run.py train -s path/to/output experiment_config/bidaf.json
```

matt-gardner:
I thought the intent was to change the config files to use s3 urls.  So those kinds of overrides aren't necessary.  Right?

I think any `--overrides` flag should be a _very_ rare use case, basically only when things are different at training vs. evaluation time for some odd reason.  Otherwise you should just modify the config file.

schmmd:
I don't understand the question.  I agree `--overrides` should be rare.  I started down the path of adding that functionality in response to https://github.com/allenai/allennlp/issues/244, only to find that functionality wasn't needed.

matt-gardner:
Yeah, sorry we had an outdated issue (in fairness, it _is_ marked as "low priority").  I think this _might_ be necessary at some point, but I don't think it's a good way to specify arguments for `train` - the issue mentioned overrides only for `evaluate`.

schmmd:
This is now ready for review.  This PR adds a `--overrides` flag which takes a HOCON blob.  That HOCON is used with the experiment configuration as a fallback, effectively overriding entries in it.  This satisfies what Oyvind was looking for--a way to override batch size when running evaluations.

allennlp/common/params.py
```diff
@@ -188,15 +188,21 @@ def _check_is_dict(self, new_history, value):
...
         Load a `Params` object from a configuration file.
         """
         # redirect to cache, if necessary
         params_file = cached_path(params_file)
 
-        param_dict = pyhocon.ConfigFactory.parse_file(params_file)
-        return Params(param_dict)
+        file_dict = pyhocon.ConfigFactory.parse_file(params_file)
+
+        if params_overrides != None:
```
joelgrus: idiomatic is to use `is not None`

joelgrus: but I feel like this method could be much cleaner if the default value for `params_overrides` was the empty string.

then you don't need a conditional at all, you can just do

```
original = pyhocon.ConfigFactory.parse_file(params_file)
overrides = pyhocon.ConfigFactory.parse_string(params_overrides)
return Params(overrides.with_fallback(original))
```

much simpler

allennlp/models/archival.py
```diff
@@ -49,7 +49,7 @@ def archive_model(serialization_dir: str,
         archive.add(os.path.join(serialization_dir, "vocabulary"),
                     arcname="vocabulary")
 
-def load_archive(archive_file: str, cuda_device: int = -1) -> Archive:
+def load_archive(archive_file: str, cuda_device: int = -1, overrides: str = None) -> Archive:
```
joelgrus: in which case use an empty string here too

allennlp/commands/train.py
```diff
@@ -51,6 +51,9 @@ def add_subparser(parser: argparse._SubParsersAction) -> argparse.ArgumentParser
                            type=str,
                            required=True,
                            help='directory in which to save the model and its logs')
+    subparser.add_argument('-o', '--overrides',
```
matt-gardner: I think this is great for `evaluate` and `predict`, but I'm conflicted about allowing this for `train`.  I kind of think this shouldn't even be an option, because it encourages you to do the wrong thing.

tests/common/params_test.py
```diff
@@ -14,3 +14,15 @@ def test_load_from_file(self):
 
         model_params = params.pop("model")
         assert model_params.pop("type") == "bidaf"
+
+    def test_overrides(self):
+        filename = 'tests/fixtures/bidaf/experiment.json'
+        params = Params.from_file(filename, '{ "train_data_path": "FOO", "model": { "type": "BAR" }, model.text_field_embedder.tokens.type: "BAZ" }') #pylint: disable=line-too-long
```
matt-gardner: Instead of disabling pylint, can you just move the string to its own line, so it's isn't too long?

allennlp/common/params.py
schmmd: Funny--you're right.  I'm misremembering `pyhocon.ConfigFactory.parse_string("")` throwing an exception.

tests/common/params_test.py
joelgrus: whitespace is implicitly string concatenation, so you should also be able to do something like

```
params = Params.from_file(filename, 
                          '{ "train_data_path": "FOO",'
                          '  "model": { "type": "BAR" },' 
                          '  model.text_field_embedder.tokens.type: "BAZ" }')')
```

if you find that tolerable

allennlp/commands/train.py
schmmd: I understand your point, but I'd like to leave it in:

* It's available for the other options.
* The code is already written.
* We may come up with a use case.

matt-gardner: Using this parameter during training actually breaks things.  The way that we archive models is by copying the config file that was used for training, and using it when loading the model.  The overrides that you passed to `train` will not be present, and thus your model will be broken.

Incidentally, this also means that using environment variables _also_ breaks things.  We really should just have one way to specify parameters to `train`, not ways to override a file that you can just directly edit.

schmmd: This doesn't seem to be a problem.  We save from the model from what we have in memory rather than copying the original file.

https://github.com/allenai/allennlp/blob/master/allennlp/commands/train.py#L113

Here's a snippet to verify this behavior.  It also substitutes variables with environment variables in the file it saves.

```
import json
import os
from copy import deepcopy

filename = 'tests/fixtures/bidaf/experiment.json'
params = Params.from_file(filename, '{ foo: $PATH }')
serialization_params = deepcopy(params).as_dict(quiet=True)
with open("model_params_test.json", "w") as param_file:
    json.dump(serialization_params, param_file, indent=4)
```

matt-gardner: Well, that's better than I thought.  I'm still not a fan of having this as an option for training; having the same functionality in more than one place is confusing and error prone.

schmmd:
@joelgrus does this look good now?  I've addressed your comments.

@matt-gardner let's follow up in person tomorrow re: overrides and training.

joelgrus:
you addressed my concerns, but I mostly agree with @matt-gardner about not allowing overrides for training

schmmd:
Well, if you both don't think it should be there then it shouldn't be!  Removed from train @joelgrus @matt-gardner .

PR #405, opened 2017-10-13
Mark test_model_can_train_save_and_load as flaky.

schmmd:


tests/models/decomposable_attention_test.py
```diff
@@ -22,6 +22,7 @@ def test_forward_pass_runs_correctly(self):
         output_dict = self.model.forward(**training_arrays)
         assert_almost_equal(numpy.sum(output_dict["label_probs"][0].data.numpy(), -1), 1, decimal=6)
 
+    @flaky
```
matt-gardner: Have you actually seen this one fail?  It's much smaller than BiDAF.  But I guess it does have weighted sums that use `bmm`...

schmmd: I think I've only seen the BiDAF one fail, but I'm not sure.

matt-gardner: It doesn't hurt anything to have the decorator there - if it passes the first time, it doesn't get re-run.  So it's fine.

PR #411, opened 2017-10-16
Make sure IndexField padding passes valid arguments

tusharkhot:
I am using the same logic as `LabelField` and the `coref-wip` branch to deal with padding. However the documentation of `IndexField` allows for the index to be `Optional`. Should I instead change the constructor to allow for `None` to be a valid index argument ?

### DIFF
--- a/allennlp/data/fields/index_field.py
+++ b/allennlp/data/fields/index_field.py
@@ -44,4 +44,4 @@ class IndexField(Field[numpy.ndarray]):
 
     @overrides
     def empty_field(self):
-        return IndexField(None, None)
+        return IndexField(-1, None)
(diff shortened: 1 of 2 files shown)


matt-gardner:
I think we should remove the `Optional` type and say in the docstring that a value of `-1` represents no label for padding purposes.

tusharkhot:
@matt-gardner Thanks! I was just about to make that change.

tusharkhot:
Is there anything else that you would recommend changing ? I don't have the permission to merge this PR, so someone from AllenNLP would have to merge it.

matt-gardner:
Nope, I approved it, and was waiting for tests to run before actually merging it.  Just got back, so merging it now.

matt-gardner:
Thanks for doing this Tushar!

PR #412, opened 2017-10-17
Add Coref shell to service

DeNeutoy:
Again, a PR into `coref-wip`.
- Refactors some comments in the coref model to be formatted more nicely.
- Adds a `decode` method to the coref model, which returns lists of lists of span indices corresponding to clusters, along with a test.
- Adds a coref `Predictor` and coresponding tests.
- Adds the coref fixture as a dummy model in the service.


allennlp/nn/util.py
```diff
@@ -507,16 +507,16 @@ def flatten_and_batch_shift_indices(indices: torch.Tensor,
...
-    .. code-block:: python
-        indices = torch.ones([2,3]).long()
-        # Sequence length of the target tensor.
-        sequence_length = 10
-        shifted_indices = flatten_and_batch_shift_indices(indices, sequence_length)
-        # Indices into the second element in the batch are correctly shifted
-        # to take into account that the target tensor will be flattened before
-        # the indices are applied.
-        assert shifted_indices == [1, 1, 1, 11, 11, 11]
+    ```
```
matt-gardner: Are you sure this works?  I think the `code-block` thing you had is right, it just maybe didn't like the `example:`?

allennlp/models/coreference_resolution/coref.py
```diff
@@ -552,6 +566,66 @@ def forward(self,  # type: ignore
...
+
+            clusters: Dict[int, List[Tuple[int, int]]] = {}
+            for span, antecedent_index in zip(all_spans, span_antecedents):
+                if antecedent_index != -1:
+                    # Find the right cluster to update with this span.
+                    # We might have referred to a span which in turn
+                    # refers to some previous span, so here we trace back
+                    # through the list to find the "root" of the cluster,
+                    # and only update that set.
+                    cluster_index_to_update = antecedent_index
```
matt-gardner: What you have is fine, but this might be slightly simpler and need less explanation if you do:

```python
cluster = clusters.setdefault(antecedent_index, set([]))
clusters[current_span_index] = cluster
cluster.add(antecedent_span)
cluster.add(current_span)
```

This way you're just making the `clusters[span]` object refer to the same set for all coreferent indices, and you always add both the antecedent span and the current span, because it's a set.

Hmm, this might break the call to `.values()` at the end, though.  I guess you could add those to a set, too, then return the result as a list...

allennlp/service/predictors/coref.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@Predictor.register("coreference-resolution")
+class CorefPredictor(Predictor):
+    """
+    Wrapper for the :class:`~allennlp.models.coreference_resolution.CoreferenceResolver` model.
+    """
+    def __init__(self, model: Model, dataset_reader: DatasetReader) -> None:
+        super().__init__(model, dataset_reader)
+        self._spacy = spacy.load("en", ner=False, tagger=False, vectors=False)
```
matt-gardner: There's some code in the tokenizer to make sure we only load spacy once.  Can you use that here?

allennlp/service/predictors/coref.py
```diff
@@ -0,0 +1,61 @@
...
+    def _json_to_instance(self, json: JsonDict) -> Instance:
+        # We're overriding `predict_json` directly, so we don't need this.  But I'd rather have a
+        # useless stub here then make the base class throw a RuntimeError instead of a
+        # NotImplementedError - the checking on the base class is worth it.
+        raise RuntimeError("this should never be called")
+
+    @overrides
+    def predict_json(self, inputs: JsonDict, cuda_device: int = -1) -> JsonDict:
+        """
+        Expects JSON that looks like ``{"document": "..."}``
```
matt-gardner: I think `"[document text]"` would be more clear than `"..."`.

allennlp/service/predictors/coref.py
```diff
@@ -0,0 +1,61 @@
...
+        and returns JSON that looks like:
+
+        .. code-block:: js
+
+            {
+            "document": [...]
```
matt-gardner: What is this `[...]`?

allennlp/service/predictors/coref.py
matt-gardner: Also adding a comment about how you can't just use one of our tokenizers, because you need sentence splitting also, would be good.

allennlp/service/predictors/coref.py
```diff
@@ -0,0 +1,61 @@
...
+class CorefPredictor(Predictor):
+    """
+    Wrapper for the :class:`~allennlp.models.coreference_resolution.CoreferenceResolver` model.
+    """
+    def __init__(self, model: Model, dataset_reader: DatasetReader) -> None:
+        super().__init__(model, dataset_reader)
+        self._spacy = spacy.load("en", ner=False, tagger=False, vectors=False)
+
+    def _json_to_instance(self, json: JsonDict) -> Instance:
+        # We're overriding `predict_json` directly, so we don't need this.  But I'd rather have a
```
matt-gardner: Can you just use this?  It looks like all `predict_json` is doing is converting the json to an instance, then doing what `super.predict_json` does.

tests/models/coreference_resolution/coref_test.py
```diff
@@ -13,6 +15,36 @@ def setUp(self):
     def test_coref_model_can_train_save_and_load(self):
         self.ensure_model_can_train_save_and_load(self.param_file)
 
-    @flaky
+    @pytest.mark.skip
     def test_batch_predictions_are_consistent(self):
+        # TODO(Mark): Make this pass by doing batch witchcraft in the model using .eval.
```
matt-gardner: Not sure that particular task is worth the effort at this point.

tests/models/coreference_resolution/coref_test.py
```diff
@@ -13,6 +15,36 @@ def setUp(self):
...
+                        "top_spans": spans,
+                        "predicted_antecedents": antecedents
+                    }
+        output = self.model.decode(output_dict)
+
+        clusters = output["clusters"][0]
+        gold1 = [(1, 2), (3, 4), (14, 56)]
+        gold2 = [(5, 6), (17, 80)]
+        assert gold1 in clusters
+        clusters.remove(gold1)
```
matt-gardner: Probably slightly cleaner to do `assert len(clusters) == 2`, and then the `assert gold in clusters` tests, instead of doing removes like this.

tests/service/predictors/coref_test.py
```diff
@@ -0,0 +1,32 @@
...
+        assert document == ['This', 'is', 'a', 'single', 'string',
+                            'document', 'about', 'a', 'test', '.', 'Sometimes',
+                            'it', 'contains', 'coreferent', 'parts', '.']
+
+        clusters = result["clusters"]
+        assert isinstance(clusters, list)
+
+        for cluster in clusters:
+            # Spans should be integer indices.
+            assert isinstance(cluster[0], int)
```
matt-gardner: I'm confused that this passes.  Shouldn't you get a list of clusters, where each cluster is a list of mentions?  There should be another nesting here, I think:

```python
for cluster in clusters:
    for mention in cluster:
        assert isinstance(mention[0], int)
        ...
```

allennlp/models/coreference_resolution/coref.py
```diff
@@ -552,6 +566,66 @@ def forward(self,  # type: ignore
...
 
+    def decode(self, output_dict: Dict[str, torch.Tensor]):
+        """
+        Converts the list of spans and predicted antecedent indices into clusters
+        of spans for each element in the batch.
+
+        Parameters
+        ----------
+        output_dict : ``Dict[str, torch.Tensor]``, required.
+            The result of calling :func:`~forward` on an instance or batch of instances.
```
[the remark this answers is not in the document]
DeNeutoy: Ah, I didn't know that, I thought it was some kind of relative path. Thanks!

allennlp/models/coreference_resolution/coref.py
DeNeutoy: Yeah I originally started out with a `defaultdict(set)`, but then I thought this way was more explicit about how we have to remember to add the root span - i've changed it back.

tests/service/predictors/coref_test.py
DeNeutoy: Yeah thanks, I always forget about the `output = output[0]` in `forward_on_instance`.

allennlp/service/predictors/coref.py
DeNeutoy: We do that for the tokenisers because we might load many instances of the class though right? Doesn't seem applicable here - the service only ever loads a single `Predictor` per model, and `predict` can't really cache spacy in memory because it's a command line tool and it only predicts on a single model at a time.

allennlp/service/predictors/coref.py
DeNeutoy: I want to return the tokenised document - I don't think we want to do this by reconstructing the document from the indexed version in the model, but we could change the signature of `_json_to_instance` to be `Instance, Optional[JsonDict]`, so that you could pass things you don't want to be modified through to the output. I'm happy with just using the predict_json, but happy to make that change if you prefer.

allennlp/service/predictors/coref.py
matt-gardner: I mean this: https://github.com/allenai/allennlp/blob/766911d6397c137248d4dab1f2d77a4c2f74a7be/allennlp/data/tokenizers/word_splitter.py#L181-L193

The issue is that loading spacy is expensive, and it has frequently caused us to run into memory errors on Travis when running tests, if we're not careful about only loading spacy once.  It might be worth putting that block above into a `common.util`, so you can call it here.

allennlp/service/predictors/coref.py
matt-gardner: Hmm, yeah, what you have is fine.  The fact that we need to do this, both here and for the SRL model, makes me wonder if we should change the API a bit.  But that's an issue for a separate PR.

PR #417, opened 2017-10-19
Update model configurations for easy training.

### LINKED ISSUE #397: Make training files universal, cut dependency on Beaker
Presently the training files have hard coded paths on mounted Beaker datasets.

https://github.com/allenai/allennlp/blob/master/experiment_config/bidaf.json

This makes our experiments un-trainable to people outside AI2 without using Beaker.  I think we should instead use paths to folders in S3.  This way, people external to AI2 can easily retrain our models and training inside Beaker.

This also would help us make our [`run_with_beaker.sh`](https://github.com/allenai/allennlp/blob/master/scripts/ai2-internal/run_with_beaker.sh) script less BiDAF-specific.  Presently it mounts in squad and glove, which is not needed for all training jobs.

Very concretely, for BiDAF I propose uploading the datasets to our s3 bucket.

* `/squad/train-v1.1.json` -> `s3://allennlp/datasets/squad/train-v1.1.json`
* `/squad/dev-v1.1.json` -> `s3://allennlp/datasets/squad/dev-v1.1.json`
* `/glove/glove.6B.100d.txt.gz` -> `s3://allennlp/datasets/glove/glove.6B.100d.txt.gz`

And then similarly updating those paths in [the experiment config](https://github.com/allenai/allennlp/blob/master/experiment_config/bidaf.json
).

### LINKED ISSUE #410: Glove file must be a local path
We should support http/https paths in addition to local paths, as we do for other fields in experiment configurations.

https://github.com/allenai/allennlp/blob/master/experiment_config/bidaf.json#L26
https://github.com/allenai/allennlp/blob/master/allennlp/modules/token_embedders/embedding.py#L210

schmmd:
This changes the paths in model configurations from artificial paths to http URLs, which allows new users to easily use these configurations to retrain models.

Variables are used in the SRL configuration files because we are not able to supply the training data for that model.  If the SRL configuration is run, the user will immediately get an error that will lead them to updating those values.

Fixes #410.
Fixes #397.


Dockerfile
```diff
@@ -46,7 +46,7 @@ COPY tests/ tests/
 COPY pytest.ini pytest.ini
 COPY scripts/ scripts/
 COPY tutorials/ tutorials/
-COPY experiment_config experiment_config/
+COPY model_config model_config/
```
matt-gardner: What do you think of `training_config` here?  I think it's a little more clear.

allennlp/training/trainer.py
```diff
@@ -258,6 +259,8 @@ def _train_epoch(self, epoch: int) -> dict:
                 logger.info("Batch %d/%d: %s", batch_num, num_training_batches, description)
                 self._last_log = time.time()
 
+        time.strftime("Duration: %H:%M:%S", time.gmtime(elapsed_time))
```
matt-gardner: `tqdm` gives you durations already, so this should only be printed in the `no_tqdm` case.

allennlp/training/trainer.py
```diff
@@ -213,7 +215,6 @@ def _train_epoch(self, epoch: int) -> dict:
         self._last_log = time.time()
         batch_num = 0
 
-        logger.info("Training")
```
matt-gardner: I would keep this here, as we have progress bars for training and validation separately, and this makes it more obvious which one is which.

model_config/bidaf.json
```diff
@@ -16,14 +16,14 @@
       }
     }
   },
-  "train_data_path": "/squad/train-v1.1.json",
-  "validation_data_path": "/squad/dev-v1.1.json",
+  "train_data_path": "https://s3-us-west-2.amazonaws.com/allennlp/datasets/mc/squad-train-v1.1.json",
```
matt-gardner: I think it'd be nicer to store things on S3 like `/datasets/mc/squad/[file]`, instead of `/datasets/mc/squad-[file]`, as the latter could get pretty cluttered as we add more machine comprehension datasets.

model_config/decomposable_attention.json
```diff
@@ -11,15 +11,15 @@
       "end_tokens": ["@@NULL@@"]
     }
   },
-  "train_data_path": "/snli/snli_1.0_train.jsonl",
-  "validation_data_path": "/snli/snli_1.0_dev.jsonl",
+  "train_data_path": "https://s3-us-west-2.amazonaws.com/allennlp/datasets/te/snli_1.0_train.jsonl",
+  "validation_data_path": "https://s3-us-west-2.amazonaws.com/allennlp/datasets/te/squad-dev-v1.1.jsonl",
```
matt-gardner: Looks like a copy-paste error here.

model_config/decomposable_attention.json
```diff
@@ -11,15 +11,15 @@
       "end_tokens": ["@@NULL@@"]
     }
   },
-  "train_data_path": "/snli/snli_1.0_train.jsonl",
-  "validation_data_path": "/snli/snli_1.0_dev.jsonl",
+  "train_data_path": "https://s3-us-west-2.amazonaws.com/allennlp/datasets/te/snli_1.0_train.jsonl",
```
matt-gardner: Same comment about `snli/` vs. `snli_`.

Dockerfile
schmmd: I don't care.  That works for me.

allennlp/training/trainer.py
schmmd: I'll revert this commit--it wasn't supposed to be in the CR.

model_config/bidaf.json
schmmd: @matt-gardner I'm going to put them in `/datasets/squad` instead.  I think this is better than what I had originally and will be easier to manage than organizing datasets by task additionally.

schmmd: @matt-gardner I'm going to keep the dataset name on the file in addition.  Although it's redundant, it makes it easier to understand downloaded files.

matt-gardner: Sounds good to me.

training_config/decomposable_attention.json
```diff
@@ -11,15 +11,15 @@
...
+  "train_data_path": "https://s3-us-west-2.amazonaws.com/allennlp/datasets/te/snli_1.0_train.jsonl",
+  "validation_data_path": "https://s3-us-west-2.amazonaws.com/allennlp/datasets/te/snli_1.0_dev.jsonl",
   "model": {
     "type": "decomposable_attention",
     "text_field_embedder": {
       "tokens": {
         "type": "embedding",
         "projection_dim": 200,
-        "pretrained_file": "/glove/glove.6B.300d.txt.gz",
+        "pretrained_file": "https://s3-us-west-2.amazonaws.com/allennlp/datasets/glove/glove.6B.100d.txt.gz",
```
matt-gardner: This is the line that caused your problem.  You need 300d vectors here.  Sorry I missed this on the first pass.

schmmd: Oops, thank you!

PR #421, opened 2017-10-19
Add timing per-epoch and time remaining estimate.

schmmd:


allennlp/training/trainer.py
```diff
@@ -394,6 +396,15 @@ def train(self) -> None:
             self._metrics_to_console(train_metrics, val_metrics)
             self._update_learning_rate(epoch, val_metric=this_epoch_val_metric)
 
+            epoch_elapsed_time = time.time() - epoch_start_time
+            logger.info("Epoch duration: " + time.strftime("%H:%M:%S", time.gmtime(epoch_elapsed_time)))
```
schmmd: @matt-gardner you suggested putting this behind a conditional, but I like printing this even when TQDM is enabled.  It provides the duration for training plus validation, and is much easier (for me at least) to find and read.  Otherwise, I frequently mistakenly read the duration of validation for the duration of training, and vice-versa.

PR #423, opened 2017-10-19
Add inspect_cache.py script (#406)

### LINKED ISSUE #406: Get descriptions of cached urls
Currently it's hard to know what has been cached, say if I want to delete some old stuff, or I want to invalidate a cache because I know the content has changed:
```
$ ls -l ~/.allennlp/datasets/
total 180464
-rw-r--r--  1 tafjord  staff  46217074 Sep  6 10:33 aHR0cHM6Ly9zMy11cy13ZXN0LTIuYW1hem9uYXdzLmNvbS9hbGxlbm5scC9tb2RlbHMvYmlkYWYtbW9kZWwtMjAxNy4wOC4zMS50YXIuZ3o=."4688241e6d52191372d0d8dcdc97e2ae-6"
-rw-r--r--  1 tafjord  staff  46175392 Sep 20 17:58 aHR0cHM6Ly9zMy11cy13ZXN0LTIuYW1hem9uYXdzLmNvbS9hbGxlbm5scC9tb2RlbHMvYmlkYWYtbW9kZWwtMjAxNy4wOS4xNS1jaGFycGFkLnRhci5neg==."bc6cc08ed491c6ac831bf363f7312fc1-6"
```
I don't know if this can be made a little less obscure (e.g., make these folders with a metadata file), but a workaround is to manually base64 decode them, e.g.,:
```
$ for f in `ls -1 ~/.allennlp/datasets/`; do echo $f | base64 --decode; echo ''; done
https://s3-us-west-2.amazonaws.com/allennlp/models/bidaf-model-2017.08.31.tar.gz
https://s3-us-west-2.amazonaws.com/allennlp/models/bidaf-model-2017.09.15-charpad.tar.gz
```

rHuggler:
This scripts goes through all dataset cached files and prints the file name and its decoded URL, making it easier to identify those files.

Fixes #406.


schmmd:
LGTM.  Assigning Joel, as he has much more expertise with our cache.

scripts/inspect_cache.py
```diff
@@ -0,0 +1,22 @@
...
+
+try:
+    cached_files = os.listdir(DATASET_CACHE)
+    if not cached_files:
+        print('No cached datasets found.')
+
+    for filename in cached_files:
+        url, _ = filename_to_url(filename)
+        print('Filename: %s' % filename)
+        print('Url: %s' % url)
```
joelgrus: I would probably print the ETag as well, so that people aren't confused if there are multiple versions of a file cached

scripts/inspect_cache.py
```diff
@@ -0,0 +1,22 @@
...
+        print()
+except FileNotFoundError:
+    print('Error: Dataset cache directory not found. Does it exist?')
```
joelgrus: in this case I would also print out what the missing directory is

rHuggler:
I wasn't sure whether if I was supposed to amend my previous commit or just create another commit and ended messing up my original commit (sorry for that), so I had to push with --force to my fork. But now it's updated with your suggestions @joelgrus!

scripts/inspect_cache.py
```diff
@@ -0,0 +1,24 @@
+import os
+import base64
+import sys
+
+sys.path.insert(0, os.path.dirname(os.path.abspath(os.path.join(__file__, os.pardir))))
+from allennlp.common.file_utils import filename_to_url
+
+CACHE_ROOT = os.path.expanduser(os.path.join('~', '.allennlp'))
+DATASET_CACHE = os.path.join(CACHE_ROOT, "datasets")
```
joelgrus: I think you should import both of these constants from `file_utils`, in case we decide to change them at some point

scripts/inspect_cache.py
```diff
@@ -0,0 +1,24 @@
...
+
+    for filename in cached_files:
+        url, etag = filename_to_url(filename)
+        print('Filename: %s' % filename)
+        print('Url: %s' % url)
+        print('ETag: %s' % etag)
+        print()
+except FileNotFoundError:
+    print('Dataset cache directory not found: %s' % str(DATASET_CACHE))
+    print('Does it exist?')
```
joelgrus: after thinking about this more, if no files have ever been cached, the directory won't exist, so I think we should display the same "no cached datasets found" message as if the directory is empty. we shouldn't warn about not existing at all, since that will happen if you've never downloaded anything

scripts/inspect_cache.py
rHuggler: Done :)

joelgrus:
thanks for the changes, looks good

PR #426, opened 2017-10-20
Fix patience

nafitzgerald:
This fixes early-stopping / patience in the trainer. Patience should be defined as "Stop if you haven't improved the metric in N epochs", but the previous definition was "Stop if the current score is worse than the worse in N epochs". This lead to cases where training would stop even if the metric had increased recently.

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -265,12 +265,11 @@ class Trainer:
         uses patience and the validation metric to determine if training should stop early
         """
         if len(prev_metrics) > self._patience:
-            # Is the worst validation performance in past self._patience
-            # epochs is better than current value?
+            # Is the best score in the past N epochs worse than the best score overall?
             if self._validation_metric_decreases:
-                return max(prev_metrics[-self._patience:]) < latest_metric
+                return min(prev_metrics[-self._patience:]) > min(prev_metrics)
             else:
-                return min(prev_metrics[-self._patience:]) > latest_metric
+                return max(prev_metrics[-self._patience:]) < max(prev_metrics)
 
         return False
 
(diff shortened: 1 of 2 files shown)


matt-gardner:
Can you add a test for `_should_stop_early` [here](https://github.com/allenai/allennlp/blob/master/tests/training/trainer_test.py)?

allennlp/training/trainer.py
```diff
@@ -260,17 +260,16 @@ def _train_epoch(self, epoch: int) -> dict:
 
         return self._get_metrics(train_loss, batch_num, reset=True)
 
-    def _should_stop_early(self, latest_metric: float, prev_metrics: List[float]) -> bool:
+    def _should_stop_early(self, prev_metrics: List[float]) -> bool:
```
matt-gardner: Given that this list now has the latest metric, `prev_metrics` doesn't make too much sense.  Maybe something like `metric_history`?  `metric_list`?  I don't have a great suggestion for a name here...

tests/training/trainer_test.py
```diff
@@ -61,6 +61,25 @@ def test_trainer_can_resume_training(self):
...
         new_trainer.train()
 
+    def test_should_stop_early_positive(self):
+        new_trainer = Trainer(self.model, self.optimizer,
+                              self.iterator, self.dataset,
+                              validation_dataset=self.dataset,
+                              num_epochs=3, serialization_dir=self.TEST_DIR,
+                              patience=5, validation_metric="+test")
+        assert new_trainer._should_stop_early([.5, .3, .2, .1, .4, .4])
+        assert not new_trainer._should_stop_early([.3, .3, .1, .2, .5, .3])
```
matt-gardner: Can you switch the .1 and the last .3 here, so that we can isolate the problem?  In other words, the best way to fix an issue is to first write a test that fails with current logic, show that it fails, then fix the code, and show that fixing the code fixes the test.  I think the current logic will pass this test, so you haven't isolated the issue yet.

matt-gardner: Oh, I might have this backwards...

matt-gardner: The "positive" and "negative" names could be better: `test_should_stop_early_with_increasing_metric` would be immediately obvious, and I wouldn't have had to figure out which case you were testing.

tests/training/trainer_test.py
```diff
@@ -61,6 +61,25 @@ def test_trainer_can_resume_training(self):
...
+
+    def test_should_stop_early_negative(self):
+        new_trainer = Trainer(self.model, self.optimizer,
+                              self.iterator, self.dataset,
+                              validation_dataset=self.dataset,
+                              num_epochs=3, serialization_dir=self.TEST_DIR,
+                              patience=5, validation_metric="-test")
+        assert new_trainer._should_stop_early([.02, .3, .2, .1, .4, .4])
+        assert not new_trainer._should_stop_early([.3, .3, .2, .1, .4, .4])
```
matt-gardner: Same comment here about making this test actually fail without the change that you made.

Issue #425, opened 2017-10-20
Training decomposable attention nondeterministically hangs


schmmd:
The hang occurs when reading SNLI instances--before training.

```
# allennlp/run train training_config/decomposable_attention.json -s output_path
2017-10-20 15:55:22,834 - WARNING - root - run is deprecated, please use run.py
2017-10-20 15:55:22,890 - INFO - allennlp.common.params - random_seed = 13370
2017-10-20 15:55:22,890 - INFO - allennlp.common.params - numpy_seed = 1337
2017-10-20 15:55:22,890 - INFO - allennlp.common.params - pytorch_seed = 133
2017-10-20 15:55:22,891 - INFO - allennlp.common.checks - Pytorch version: 0.2.0_3
2017-10-20 15:55:22,893 - INFO - allennlp.common.params - dataset_reader.type = snli
2017-10-20 15:55:22,893 - INFO - allennlp.common.params - dataset_reader.tokenizer.type = word
2017-10-20 15:55:22,894 - INFO - allennlp.common.params - dataset_reader.tokenizer.word_splitter.type = spacy
2017-10-20 15:55:22,894 - INFO - allennlp.common.params - dataset_reader.tokenizer.word_splitter.language = en
2017-10-20 15:55:22,894 - INFO - allennlp.common.params - dataset_reader.tokenizer.word_splitter.pos_tags = False
2017-10-20 15:55:22,894 - INFO - allennlp.common.params - dataset_reader.tokenizer.word_splitter.parse = False
2017-10-20 15:55:22,894 - INFO - allennlp.common.params - dataset_reader.tokenizer.word_splitter.ner = False
2017-10-20 15:55:24,977 - INFO - allennlp.common.params - dataset_reader.tokenizer.word_filter.type = pass_through
2017-10-20 15:55:24,978 - INFO - allennlp.common.params - dataset_reader.tokenizer.word_stemmer.type = pass_through
2017-10-20 15:55:24,978 - INFO - allennlp.common.params - dataset_reader.tokenizer.start_tokens = None
2017-10-20 15:55:24,978 - INFO - allennlp.common.params - dataset_reader.tokenizer.end_tokens = ['@@NULL@@']
2017-10-20 15:55:24,979 - INFO - allennlp.common.params - dataset_reader.token_indexers.tokens.type = single_id
2017-10-20 15:55:24,979 - INFO - allennlp.common.params - dataset_reader.token_indexers.tokens.namespace = tokens
[...]

PR #427, opened 2017-10-20
Close TQDM progress

schmmd:
Fixes https://github.com/allenai/allennlp/issues/425


schmmd:
While this resolve my repro case, it's not clear it's a complete fix.

matt-gardner:
What do you think is still missing?  The exception that you saw on the other issue?  Yeah, that seems different.

PR #430, opened 2017-10-21
Add NER tag token indexer

matt-gardner:
I also changed one of the test configuration fixtures to actually use this, so we have an example to point to when people ask.

### DIFF
--- a/tests/data/token_indexers/ner_tag_indexer_test.py
+++ b/tests/data/token_indexers/ner_tag_indexer_test.py
@@ -12,7 +12,7 @@ class TestNerTagIndexer(AllenNlpTestCase):
         super(TestNerTagIndexer, self).setUp()
         self.tokenizer = SpacyWordSplitter(ner=True)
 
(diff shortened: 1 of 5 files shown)


tests/data/token_indexers/ner_tag_indexer_test.py
```diff
@@ -0,0 +1,43 @@
...
+from allennlp.data.token_indexers import NerTagIndexer
+from allennlp.data.tokenizers.word_splitter import SpacyWordSplitter
+
+
+class TestNerTagIndexer(AllenNlpTestCase):
+    def setUp(self):
+        super(TestNerTagIndexer, self).setUp()
+        self.tokenizer = SpacyWordSplitter(ner=True)
+
+    def test_count_vocab_items_uses_pos_tags(self):
```
DeNeutoy: `pos` -> `ner`, and below.

matt-gardner: Right, thanks for catching that.

PR #434, opened 2017-10-21
Wikitables type declaration and action sequence generation

pdasigi:
These are helper classes that will be used by the upcoming `WikitablesDatasetReader`. I've added an initial `TypeDeclaration` class, a `World` representation containing methods for reading logical forms and converting them into action sequences, and a `TableKnowledgeGraph` class that maintains entity neighborhood information. The logical form to action sequence conversion logic works for simple logical forms (see `TestWorldRepresentation`). I still need to add more types to make it work for all of Lambda-DCS.

@matt-gardner This is still WIP, but feel free to look at the API and make comments.


allennlp/data/dataset_readers/wikitables/table.py
```diff
@@ -0,0 +1,95 @@
...
+        string = re.sub("[‐‑–—]", "-", string)
+        string = re.sub("[\\u2E00-\\uFFFF]", "", string)
+        string = re.sub("\\s+", " ", string)
+        # Canonicalization rules from Sempre
+        string = re.sub("[^\\w]", "_", string)
+        string = re.sub("_+", "_", string)
+        string = re.sub("_$", "", string)
+        return string.lower()
+
+    def get_cell_neighbors(self,
```
matt-gardner: What's the `name_mapping` for?  The caller could just as easily do `graph.get_cell_neighbors(name_mapping[cell])`.  That'd be preferable to making this API less clean.  Same with getting column neighbors.

allennlp/data/dataset_readers/wikitables/type_declaration.py
```diff
@@ -0,0 +1,73 @@
+"""
+Defines all the types in the WikitablesQuestions domain. We exploit the type logic in ``nltk.sem.logic``
+here. This module also contains two helper classes that add some functionality on top of NLTK's logic module.
+"""
+from overrides import overrides
+
+from nltk.sem.logic import BasicType, ComplexType, EntityType, ANY_TYPE
+
+
+class BasicTypeFactory(BasicType):
```
matt-gardner: If you really want to do something like this, I think it's sufficient to just do this:

```python
class NamedBasicType(BasicType):
  def __init__(self, name):
    self.name = name
  def __str__(self) -> str:
    return self.name.lower()
```
And then you call `ROW_TYPE = NamedBasicType("ROW")` below.

But, given that you only do this twice, I think it'd just be cleaner to create two actual subclasses of `BasicType`, `RowType` and `DateNumType`, and then you'd match the nltk API better.

allennlp/data/dataset_readers/wikitables/type_declaration.py
```diff
@@ -0,0 +1,73 @@
...
+        reversed_second = ComplexType(other.second.second, other.second.first)
+        first = other.first.resolve(reversed_second)
+        reversed_first = ComplexType(first.second, first.first)
+        second = other.second.resolve(reversed_first)
+        if first and second:
+            return ReverseType(first, second)
+        return None
+
+
+class WikitablesTypeDeclaration:
```
matt-gardner: Does having this additional level of nesting actually help you with anything?  You could just have all of the items below as constants in this file.

allennlp/data/dataset_readers/wikitables/type_declaration.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    Defines the types in the WikitablesQuestions domain. This type system uses ``nltk.sem.logic.Type``.
+    """
+    CELL_TYPE = EntityType()
+    ROW_TYPE = BasicTypeFactory.define_type("ROW")
+    # TODO: Merging dates and nums. Can define a hierarchy instead.
```
matt-gardner: [Always format `TODOs` like `TODO(pradeep):`.](http://google.github.io/styleguide/cppguide.html#TODO_Comments)

allennlp/data/dataset_readers/wikitables/world.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def process_sempre_forms(self, sempre_forms: List[str]) -> List[Expression]:
+        """
+        Processes logical forms in Sempre format (that are either gold annotations or output from DPD)
+        and converts them into ``Expression`` objects in NLTK. The conversion involves mapping names to
+        follow NLTK's variable naming conventions.
+        """
+        expressions = []
+        for sempre_form in sempre_forms:
+            parsed_lisp = OneOrMore(nestedExpr()).parseString(sempre_form).asList()
```
matt-gardner: I think bare functions are better done like `pyparsing.nestedExpr()` - it's more obvious where the function is coming from, instead of wondering where `nestedExpr()` is defined.

allennlp/data/dataset_readers/wikitables/world.py
```diff
@@ -0,0 +1,111 @@
...
+                if is_leaf:
+                    mapped_names.append(self._map_name(element))
+                else:
+                    mapped_names.append(self._process_nested_expression(element))
+        return "(%s %s)" % (mapped_names[0], " ".join("(%s)" % name for name in mapped_names[1:]))
+
+    def _map_name(self, sempre_name: str):
+        """
+        Takes the name of a predicate or a constant as used by Sempre, maps it to a unique string such that
+        NLTK processes it appropriately. This is needed because NLTK has a naming convention for variables:
```
matt-gardner: Does NLTK have processing logic that switches on this naming convention?

allennlp/data/dataset_readers/wikitables/world.py
```diff
@@ -0,0 +1,111 @@
...
+                self.type_signature[translated_name] = WTD.COLUMN_TYPE
+            elif sempre_name.startswith("fb:cell"):
+                # Cell name.
+                translated_name = "cell:%s" % sempre_name.split(".")[-1]
+                self.type_signature[translated_name] = WTD.CELL_TYPE
+            self.name_mapping[sempre_name] = translated_name
+        return self.name_mapping[sempre_name]
+
+    @staticmethod
+    def get_action_sequence(expression: Expression) -> List[str]:
```
matt-gardner: This function feels like it belongs with the type declarations, not in this module.

tests/data/dataset_readers/wikitables/world_representation_test.py
```diff
@@ -0,0 +1,20 @@
...
+from allennlp.data.dataset_readers.wikitables import World
+from allennlp.common.testing import AllenNlpTestCase
+
+
+class TestWorldRepresentation(AllenNlpTestCase):
+    def test_world_processes_sempre_forms_correctly(self):
+        world = World("tests/fixtures/data/wikitables/sample_table.tsv")
+        sempre_form = "((reverse fb:row.row.year) (fb:row.row.league fb:cell.usl_a_league))"
+        expression = world.process_sempre_forms([sempre_form])[0]
+        assert str(expression) == "R(C0,C1(cell:usl_a_league))"
```
matt-gardner: This looks wrong to me.  Shouldn't it be `R(C0)(C1(cell:usl_a_league)`?

allennlp/data/dataset_readers/wikitables/world.py
pdasigi: https://github.com/nltk/nltk/blob/develop/nltk/sem/logic.py#L1817-L1848

tests/data/dataset_readers/wikitables/world_representation_test.py
pdasigi: NLTK processes it as `R` (reverse) taking two arguments, a function and its argument. This is how it stores it internally.

allennlp/data/dataset_readers/wikitables/world.py
pdasigi: I think it should be here. It is the world that defines the action sequences, not the type declaration.

allennlp/data/dataset_readers/wikitables/table.py
pdasigi: I agree. Changed this.

allennlp/data/dataset_readers/wikitables/type_declaration.py
pdasigi: Agreed. Changed it to `NamedBasicType` for now. I may wither split number and date or define a hierarchy for them later.

allennlp/data/dataset_readers/wikitables/type_declaration.py
pdasigi: You're right. Changed this.

allennlp/data/dataset_readers/wikitables/type_declaration.py
pdasigi: Done.

allennlp/data/dataset_readers/wikitables/world.py
pdasigi: Done.

allennlp/data/dataset_readers/wikitables/table.py
```diff
@@ -0,0 +1,95 @@
...
+        column names.
+    """
+    def __init__(self,
+                 column_neighbors: Dict[str, List[str]],
+                 cell_neighbors: Dict[str, List[str]]):
+        self._column_neighbors = column_neighbors
+        self._cell_neighbors = cell_neighbors
+
+    @classmethod
+    def read_table_from_tsv(cls, table_filename: str) -> 'Table':
```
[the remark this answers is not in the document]
pdasigi: Done.

pdasigi:
Type inference is finally done. I made sure there are no resolution errors on a random sample of 1000 logical forms from the DPD outputs on all questions in the training data. Next up, `WikitablesDatasetReader`.

allennlp/data/dataset_readers/wikitables/world.py
```diff
@@ -0,0 +1,143 @@
...
+        ----------
+        sempre_name : str
+            Token from Sempre's logical form.
+        """
+        if sempre_name not in self.name_mapping:
+            if sempre_name.startswith("fb:row.row"):
+                # Column name
+                translated_name = "C%d" % self._column_counter
+                self._column_counter += 1
+                self.type_signature[translated_name] = COLUMN_TYPE
```
matt-gardner: It sure looks like you could use that three-argument function I mentioned earlier here, too.

allennlp/data/dataset_readers/wikitables/world.py
```diff
@@ -0,0 +1,143 @@
...
+            mapped_names = [self._map_name(name) for name in nested_expression]
+        else:
+            mapped_names = []
+            for element, is_leaf in zip(nested_expression, elements_are_leaves):
+                if is_leaf:
+                    mapped_names.append(self._map_name(element))
+                else:
+                    mapped_names.append(self._process_nested_expression(element))
+        if mapped_names[0] == "\\":
+            # This means the predicate is lambda. NLTK wants the variable name to not be within parantheses.
```
matt-gardner: Aren't you adding parentheses here?  Is there an extra "not" in this comment?

tests/data/dataset_readers/wikitables/world_representation_test.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def test_world_returns_correct_actions_with_reverse(self):
+        table_kg = TableKnowledgeGraph.read_from_file("tests/fixtures/data/wikitables/sample_table.tsv")
+        world = World(table_kg)
+        sempre_form = "((reverse fb:row.row.year) (fb:row.row.league fb:cell.usl_a_league))"
+        expression = world.process_sempre_forms([sempre_form])[0]
+        actions = world.get_action_sequence(expression)
+        target_action_sequence = ['e -> [<r,e>, r]', '<r,e> -> [<<#1,#2>,<#2,#1>>, <e,r>]',
```
matt-gardner: Shouldn't the first action be the type of the overall expression?

allennlp/data/dataset_readers/wikitables/type_declaration.py
```diff
@@ -0,0 +1,447 @@
...
+"""
+from overrides import overrides
+
+from nltk.sem.logic import ApplicationExpression, ConstantExpression, Variable, LogicParser
+from nltk.sem.logic import BasicType, ComplexType, EntityType, ANY_TYPE
+
+
+class NamedBasicType(BasicType):
+    """
+    A ``BasicType`` that also takes the name of the type as an argument to its constructor.
```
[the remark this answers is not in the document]
pdasigi: Good point. I put that in the docstring and added a test for this as well.

allennlp/data/dataset_readers/wikitables/type_declaration.py
```diff
@@ -0,0 +1,447 @@
...
+        For example, say this type is of the function variable F in F(cell), and we start out with <?, d> (that
+        is, it takes any type and returns d). Now we have already resolved `cell` to be of type `e`. Then
+        ``resolve`` gets called with other = <e, ?>, because we know F is a function that took a constant of
+        type `e`. When we resolve <e, ?> against <?, d>, there will not be a contradiction, because any type
+        can be successfully resolved against ?. Finally we return <e, d> as the resolved type.
+
+        As a counter example, if we are trying to resolve <?, d> against <?, e>, the resolution fails, and in
+        that case, this method returns ``None``.
+
+        Note that resolution may be unidirectional because of ?, and so in the subclasses of this type, we
```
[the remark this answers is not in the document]
pdasigi: I meant that successful resolution does not imply equality of types. I changed that part to clarify.

allennlp/data/dataset_readers/wikitables/type_declaration.py
```diff
@@ -0,0 +1,447 @@
...
+        """
+        raise NotImplementedError
+
+    @overrides
+    def __eq__(self, other):
+        return self.__class__ == other.__class__
+
+    @overrides
+    def matches(self, other):
+        return self == other or self == ANY_TYPE or other == ANY_TYPE
```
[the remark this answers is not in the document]
pdasigi: The equality can return true. `self == ANY_TYPE` would be `True` iff the argument and return types of `self` are `ANY_TYPE`s.
Added a comment.

allennlp/data/dataset_readers/wikitables/type_declaration.py
```diff
@@ -0,0 +1,447 @@
...
+
+    @overrides
+    def __str__(self):
+        if self == ANY_TYPE:
```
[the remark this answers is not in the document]
pdasigi: Same as above. This is because `ANY_TYPE.first = ANY_TYPE` and `ANY_TYPE.second = ANY_TYPE` by definition, and `ComplexType` checks the two fields in its overridden `_eq_` method.
The point of this comparison here is to return `?` (`ANY_TYPE._str_()`) if the type remains unresolved.

allennlp/data/dataset_readers/wikitables/type_declaration.py
```diff
@@ -0,0 +1,447 @@
...
+        <<r,?>, <e,?>>  :   <<r,e>, <e,r>>
+        <<r,?>, ?>      :   <<r,?>, <?,r>>>
+        <<r,?>, <?,e>>  :   None  (causes resolution failure)
+    """
+    @property
+    def _signature(self):
+        return "<<#1,#2>,<#2,#1>>"
+
+    @overrides
+    def resolve(self, other):
```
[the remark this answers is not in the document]
pdasigi: Since its signature is `<<#1,#2>,<#2,#1>>` no information about types in `self` is relevant. All that matters is that `other.fiirst` resolves against the reverse of `other.second` and vice versa.

allennlp/data/dataset_readers/wikitables/type_declaration.py
```diff
@@ -0,0 +1,447 @@
...
+        try:
+            # This is the first #1 in the type signature above.
+            selector_function_type = resolved_second.second.first
+            # This is the second #1 in the type signature above.
+            quant_function_argument_type = resolved_second.second.second.first.second
+            # This is the third #1 in the type signature above.
+            return_type = resolved_second.second.second.second
+
+            # All three types above should resolve against each other.
+            resolved_first_ph = selector_function_type.resolve(quant_function_argument_type)
```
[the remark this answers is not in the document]
pdasigi: Placeholder. Variable name was already long, so had to shorten it. I'll add a comment.

allennlp/data/dataset_readers/wikitables/type_declaration.py
```diff
@@ -0,0 +1,447 @@
...
+# index
+ROW_INDEX_TYPE = ComplexType(DATE_NUM_TYPE, ROW_TYPE)
+# count
+COUNT_TYPE = CountType(ANY_TYPE, DATE_NUM_TYPE)
+# and, or
+CONJUNCTION_TYPE = ConjunctionType(ANY_TYPE, ANY_TYPE)
+# argmax, argmin
+ARG_EXTREME_TYPE = ArgExtremeType(ANY_TYPE, ANY_TYPE)
+
+COMMON_NAME_MAPPING = {"reverse": "R",
```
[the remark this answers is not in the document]
pdasigi: Good point. Thanks.

allennlp/data/dataset_readers/wikitables/world.py
```diff
@@ -0,0 +1,143 @@
+import re
+
+from typing import List
+import pyparsing
+
+from nltk.sem.logic import Expression, LambdaExpression
+
+from allennlp.data.dataset_readers.wikitables.type_declaration import COLUMN_TYPE, CELL_TYPE, PART_TYPE
```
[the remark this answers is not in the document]
pdasigi: Done.

allennlp/data/dataset_readers/wikitables/world.py
pdasigi: I'm adding parentheses only after the variable (as opposed to what's happening in else). I'll make the comment clearer.

tests/data/dataset_readers/wikitables/world_representation_test.py
pdasigi: Hmm.. yeah. That is a good idea. I'll add that to the action sequence.

PR #435, opened 2017-10-21
Coref demo

DeNeutoy:


demo/src/components/CorefComponent.js
```diff
@@ -0,0 +1,203 @@
...
+    },
+    {
+      document: "Kerbal Space Program (KSP) is a space flight simulation video game developed and published by Squad for Microsoft Windows, OS X, Linux, PlayStation 4, Xbox One, with a Wii U version that was supposed to be released at a later date. The developers have stated that the gaming landscape has changed since that announcement and more details will be released soon. In the game, players direct a nascent space program, staffed and crewed by humanoid aliens known as \"Kerbals\". The game features a realistic orbital physics engine, allowing for various real-life orbital maneuvers such as Hohmann transfer orbits and bi-elliptic transfer orbits.",
+    }
+];
+
+class CorefInput extends React.Component {
+constructor() {
```
schmmd: Please fix your indentation inside `CorefInput`.

demo/src/components/CorefComponent.js
```diff
@@ -0,0 +1,203 @@
...
+    super();
+    this.state = {
+    corefDocumentValue: "",
+    };
+    this.handleListChange = this.handleListChange.bind(this);
+    this.handleDocumentChange = this.handleDocumentChange.bind(this);
+}
+
+handleListChange(e) {
+    if (e.target.value !== "") {
```
schmmd: I'm not sure whether it's better to use `!==` or `!=`, but I would choose the latter.

demo/src/components/CorefComponent.js
```diff
@@ -0,0 +1,203 @@
...
+        body: JSON.stringify(payload)
+      }).then((response) => {
+        return response.json();
+      }).then((json) => {
+        this.setState({clusters: json["clusters"]});
+        this.setState({document: json["document"]});
+        this.setState({outputState: "received"});
+      }).catch((error) => {
+        this.setState({outputState: "error"});
+        throw error; // todo(michaels): is this right?
```
schmmd: @codeviking someone had this question at some point, and now it's been copy-pasted through our 4 examples.  Is this a reasonable thing to do, or should we somehow log our JavaScript exception without throwing it?

demo/src/components/CorefComponent.js
matt-gardner: I think conventional JS wisdom is to use `!==`: https://stackoverflow.com/a/359509/1467533.

demo/src/components/CorefComponent.js
codeviking: In this case I'd replace it with `console.error(error)`.  The thrown exception can't actually be caught, as the exception is thrown within an asynchronous block.   In other words, this wouldn't work.

```
try {
  runCorefModel(...)
} catch(err) {
  // You won't catch it here.
}
```

If you wanted the invoker to be able to handle the exception I'd simply return the result of `fetch` (so that the implementor can make a decision around control flow).  That said it doesn't seem like you need that flexibility here, so I'd just dump it to the console and call it good :).

schmmd: What are the negative consequences of uncaught Javascript exception @codeviking ?

codeviking: On the client, none.  It'll just propagate to a `console.log` ultimately.

On the server it'll cause the program to crash, which isn't great if it's an HTTP server :).

DeNeutoy:
@matt-gardner Can you take a look at the non-JS bits of this now? It's basically just correcting the decoding and adding a few comments. I also fixed the Spacy issue.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -27,7 +26,7 @@ class CoreferenceResolver(Model):
     <https://www.semanticscholar.org/paper/End-to-end-Neural-Coreference-Resolution-Lee-He/3f2114893dc44eacac951f148fbff142ca200e83>
     by Lee et al., 2017.
     The basic outline of this model is to get an embedded representation of each span in the
-    document. These span representation are scored and use to prune away spans that are unlikely
+    document. These span representations are scored and use to prune away spans that are unlikely
```
matt-gardner: `s/and use/and used/`

allennlp/models/coreference_resolution/coref.py
```diff
@@ -586,40 +602,49 @@ def decode(self, output_dict: Dict[str, torch.Tensor]):
             which are in turn comprised of a list of (start, end) inclusive spans into the
             original document.
         """
-        top_spans = output_dict["top_spans"].data.cpu()
-        predicted_antecedents = output_dict["predicted_antecedents"].data.cpu()
+        batch_top_spans = output_dict["top_spans"].data.cpu()
```
matt-gardner: Shape comments here would be helpful.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -586,40 +602,49 @@ def decode(self, output_dict: Dict[str, torch.Tensor]):
...
+            clusters: List[List[Tuple[int, int]]] = []
+
+            for i, (span, predicted_antecedent) in enumerate(zip(top_spans, predicted_antecedents)):
+                if predicted_antecedent < 0:
+                    # We don't care about spans which are
+                    # not co-referent with anything.
+                    continue
+
+                # Find the right cluster to update with this span.
+                # To do this, we find the row in ``antecedent_indices``
```
matt-gardner: I'm having a hard time following what you're saying here, probably because there's no clear description above of what "antecedent_indices" _is_.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -508,6 +507,17 @@ def forward(self,  # type: ignore
...
         max_antecedents = min(self._max_antecedents, num_spans_to_keep)
 
+        # This next section can be confusing - the returned tensors do _not_ have a batch size
+        # dimension. This is because the antecedent indices are agnostic to the batch size -
+        # it is a function only of the number of spans we are considering (i.e which survived the
+        # pruning stage) and the max_antecedents that we will consider. We then _look up_
+        # these indices in top_span_embeddings to get the embeddings for them for each
+        # instance in the batch. This tensor represents the indices which we will be generating
+        # distributions over - for each span we kept, we will generate a distribution over
+        # the maximum number of anteceedents we choose to consider. As these are just the
```
matt-gardner: `s/anteceedents/antecedents/`

allennlp/models/coreference_resolution/coref.py
matt-gardner: Also saying _what_ is in here, not just the shape, would be nice.  I _think_ it's this: `batch_top_spans` has shape `(batch_size, num_spans, 2)`, where it holds, for each span, the start and end index of the span.  `batch_predicted_antecedents` has shape `(batch_size, num_spans)`, and each element is an index into `num_spans`, saying what the span's antecedent is.  `antecedent_indices` is...?

allennlp/models/coreference_resolution/coref.py
```diff
@@ -602,8 +603,19 @@ def decode(self, output_dict: Dict[str, torch.Tensor]):
...
+        # A tensor of shape (batch_size, num_spans_to_keep, 2), representing
+        # the start and end indices of each span.
         batch_top_spans = output_dict["top_spans"].data.cpu()
+
+        # A tensor of shape (batch_size, num_spans_to_keep) representing, for each span,
+        # the index into ``antecedent_indices`` which specifies the antecedent span. Additionally,
+        # the index can be -1, specifying that the span has no predicted antecedent.
         batch_predicted_antecedents = output_dict["predicted_antecedents"].data.cpu()
+
+        # A tensor of shape (num_spans_to_keep, max_antecedents), representing the indices
```
matt-gardner: Possible modification: A tensor of shape `(num_spans_to_keep, max_antecedents)`, representing _which_ antecedents were considered for each span.  Each element is an index between 0 and `num_spans_to_keep`, so if `antecedent_indices[i, j] = k` for some `j`, it means that span `i` considered span `k` as a possible antecedent.

allennlp/training/metrics/conll_coref_scores.py
```diff
@@ -48,29 +48,38 @@ def get_gold_clusters(gold_clusters):
...
+            # Must be a previous span.
             assert i > predicted_index
-            predicted_antecedent = tuple(top_spans[predicted_index])
-            if predicted_antecedent in mention_to_predicted:
-                predicted_cluster = mention_to_predicted[predicted_antecedent]
+            antecedent_span = tuple(top_spans[predicted_index])
+
+            # Check if we've seen the span before.
+            if antecedent_span in mention_to_predicted.keys():
+                predicted_cluster_id: int = mention_to_predicted[antecedent_span]
```
matt-gardner: Given the type annotation above, you probably don't need this one here?

allennlp/training/metrics/conll_coref_scores.py
```diff
@@ -48,29 +48,38 @@ def get_gold_clusters(gold_clusters):
 
     @staticmethod
     def get_predicted_clusters(top_spans, antecedent_indices, predicted_antecedents):
-        mention_to_predicted = {}
-        predicted_clusters = []
+        mention_to_predicted: Dict[Tuple[int, int], int] = {}
```
matt-gardner: I read this at first as "mention to predict", not a mapping from mentions to predicted cluster ids.  I think something like `mention_clusters` or `predicted_mention_clusters` would be fine as a name here.

allennlp/training/metrics/conll_coref_scores.py
```diff
@@ -48,29 +48,38 @@ def get_gold_clusters(gold_clusters):
...
-            predicted_clusters[predicted_cluster].append(mention)
-            mention_to_predicted[mention] = predicted_cluster
-        predicted_clusters = [tuple(pc) for pc in predicted_clusters]
-        mention_to_predicted = {m:predicted_clusters[i] for m, i in mention_to_predicted.items()}
-        return predicted_clusters, mention_to_predicted
+            clusters[predicted_cluster_id].append(mention)
+            mention_to_predicted[mention] = predicted_cluster_id
+
+        # finalise the spans and clusters.
+        clusters = [tuple(pc) for pc in clusters]
```
matt-gardner: Why `pc` here, instead of just `c`?  Oh, "predicted cluster"...  Just using `c` would be less confusing, I think.

allennlp/training/metrics/conll_coref_scores.py
```diff
@@ -48,29 +48,38 @@ def get_gold_clusters(gold_clusters):
...
+        mention_to_predicted = {m:clusters[i] for m, i in mention_to_predicted.items()}
```
matt-gardner: This line is pretty dense; I think it could either use a comment or some re-writing to be less dense.

tests/models/coreference_resolution/coref_test.py
```diff
@@ -23,21 +23,30 @@ def test_decode(self):
                                   [5, 6],
                                   [14, 56],
                                   [17, 80]])
-        spans = Variable(spans.unsqueeze(0))
 
-        # Indices into ``spans`` indicating that the
-        # two mentions are co-referent.
-        antecedents = torch.LongTensor([-1, 0, -1, -1, 1, 3])
-        antecedents = Variable(antecedents.unsqueeze(0))
+        antecedent_indices = torch.LongTensor([[0, 0, 0, 0, 0, 0],
```
matt-gardner: Are these 1-indexed?

DeNeutoy: I think so - it's an edge case because the model will always learn to predict that the first mention can't have an antecedent (i.e it always predicts -1), so I don't exactly know what the behaviour is here. @kentonl?

kentonl: Yes, they are zero-indexed. Even though the scores are prepended with zeros due to the dummy antecedent, [this line](https://github.com/DeNeutoy/allennlp/blob/c2289bf210186dcaf6ffe68e447a77446b70baa3/allennlp/models/coreference_resolution/coref.py#L558) shifts all the indices back, and the dummy antecedent has an index of -1.

PR #436, opened 2017-10-22
add ability to provide evidence to viterbi_decode

DeNeutoy:


allennlp/nn/util.py
```diff
@@ -184,7 +187,9 @@ def masked_log_softmax(vector, mask):
     return torch.nn.functional.log_softmax(vector)
 
 
-def viterbi_decode(tag_sequence: torch.Tensor, transition_matrix: torch.Tensor):
+def viterbi_decode(tag_sequence: torch.Tensor,
+                   transition_matrix: torch.Tensor,
+                   tag_evidence: Optional[List[int]] = None):
```
matt-gardner: When I saw the title of this PR in the email, I wondered what "evidence" was.  If you had instead called this "observed_tags", I would have immediately known what you were doing.  I _think_ "observed" is the more standard term for what you're doing here.  Your docstring below seems to also agree with me.

allennlp/nn/util.py
```diff
@@ -207,17 +220,47 @@ def viterbi_decode(tag_sequence: torch.Tensor, transition_matrix: torch.Tensor):
...
+    else:
+        tag_evidence = [-1 for _ in range(sequence_length)]
+
     path_scores = []
     path_indices = []
-    path_scores.append(tag_sequence[0, :])
+
+    if tag_evidence[0] != -1:
+        one_hot = torch.zeros(num_tags)
+        one_hot[tag_evidence[0]] = 100000.
```
matt-gardner: I think it'd be better to define some constants, `VERY_LARGE`, and `VERY_NEGATIVE`, in `common.util` or `nn.util`, and use those instead of hard-coding these values throughout these functions.  I _thought_ we had already done this, but I can't find those constants at the moment.

tests/nn/util_test.py
```diff
@@ -437,6 +437,30 @@ def test_viterbi_decode(self):
         assert indices == [3, 2, 1]
         assert value.numpy() == 18
 
+        # Test that providing evidence results in paths containing specified tags.
+        sequence_predictions = torch.FloatTensor([[0, 0, 0, 4, 4],
```
matt-gardner: Seeing 4s here and the "4th" tag, along with the name "predictions", got me a little confused.  I think it'd be a bit more readable if this were named `sequence_logits`, and if you used a number that is not 4 or 5 for the logit values, so there's less room for garden path readings of the comment below.

tests/nn/util_test.py
```diff
@@ -437,6 +437,30 @@ def test_viterbi_decode(self):
...
+                                                  [0, 0, 0, 4, 4],
+                                                  [0, 0, 0, 4, 4],
+                                                  [0, 0, 0, 4, 4]])
+        # The 5th tag has a penalty for appearing sequentially
+        # or for transitioning to the 4th tag, making the best
+        # path to take the 4th tag for every label.
+        transition_matrix = torch.zeros([5, 5])
+        transition_matrix[4, 4] = -10
+        transition_matrix[4, 3] = -2
+        # The 1st, 4th and 5th tags are observed - they should be equal to
```
matt-gardner: I'm pretty sure you're using "4th" and "5th" here to refer to sequence indices, while above (and below!) they refer to tag indices.  Maybe always referring to tags by their values would make this easier to read?  I.e., above, "Tag 4 has a penalty..."

allennlp/nn/util.py
DeNeutoy: I was caught in a bit of a dilemma here, because we need to make these observed potential functions to be larger than any transition potential to any other tag, but at the same time, it's not very helpful if you are actually using the path scores for some purpose to have them unnecessarily inflated, for instance if you don't have any transition constraints. I couldn't come up with a good way to get around this, do you have any ideas?

matt-gardner: The caller has the logits and the transition matrix, so they can recover the original path scores themselves if they really want them.  Or, alternatively, if you use a known constant, they could subtract/add the value of that constant in the appropriate places to recover the scores, assuming the constant wasn't large enough that it washed out the original score.

PR #440, opened 2017-10-23
sphinx coverage

joelgrus:
most of this is just sphinx junk. the interesting parts are

* in [doc/conf.py](https://github.com/allenai/allennlp/pull/440/files#diff-684a0ceb2878d48b19612ca360e277ef) I enabled the sphinx-coverage plugin, which allows you to run "sphinx coverage", which is useful only for telling you what functions you missed in the python modules that are covered. In particular, it's not helpful for telling you that you forgot to add your new module to the docs.

* [scripts/check_docs.py](https://github.com/allenai/allennlp/pull/440/files#diff-803091b6e53da6bf7a2e907676fbd417) is a somewhat hacky script to tell you if you have any modules in your code that are not mentioned in the docs and vice versa. This is not perfect, but it's a lot better than what we have now. I also added a Travis option to run it, and I'll turn that on after I merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,7 @@ env:
     - COVERAGE="true" RUN_TESTS="true"
     - RUN_PYLINT="true"
     - RUN_MYPY="true"
-    - BUILD_DOCS="true"
-    - CHECK_DOCS="true"
+    - BUILD_DOCS="true" CHECK_DOCS="true"
 notifications:
   email: false
 addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ env:
     - RUN_PYLINT="true"
     - RUN_MYPY="true"
     - BUILD_DOCS="true"
+    - CHECK_DOCS="true"
 notifications:
   email: false
 addons:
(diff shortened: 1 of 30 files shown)


scripts/check_docs.py
```diff
@@ -0,0 +1,70 @@
...
+        # no docs at top level
+        'allennlp',
+
+        # no docs for custom extensions, which aren't even in python
+        'allennlp.custom_extensions',
+        'allennlp.custom_extensions._ext',
+        'allennlp.custom_extensions._ext.highway_lstm_layer',
+        'allennlp.custom_extensions.build',
+
+        # TODO(joelgrus) figure out how to make these docs build, the cffi part
```
matt-gardner: Missing colon, and it'd be better if all comments here were capitalized and had periods.

.travis.yml
```diff
@@ -13,6 +13,7 @@ env:
     - RUN_PYLINT="true"
     - RUN_MYPY="true"
     - BUILD_DOCS="true"
+    - CHECK_DOCS="true"
```
matt-gardner: I'd put this on the same line as `BUILD_DOCS`, so you both build the docs and check the docs in the same build.  We're allowed four concurrent builds with travis, but five will slow things down considerably.

PR #441, opened 2017-10-23
Increase number of attempts for BiDAF training test, which is very flaky

matt-gardner:


matt-gardner:
I think this way is a little better.  That test is shared, and it's only the really deep models that are really flaky.  Making it so you can configure the precision per model is more work.

DeNeutoy:
Ah good point.

PR #444, opened 2017-10-25
Enable CORS so dev is possible using npm.

schmmd:
This change allows fast development.  Specifically, it allows the developer to run the backend using `allennlp/run serve` on port 8000 and the frontend using `npm start` on port 3000.  This way, npm watches for file changes and quickly updates the browser, with a non-production javascript build which is easier to debug.

For production deployments, we can continue to use Sanic to serve our API and our built assets.

A downside to this approach is that, when using `allennlp/run serve`, the application *must* be served on port 8000.  Otherwise the Javascript components won't be able to communicate back to the server.  I believe the "correct" approach to setting this up would be to have well-configured nginx configurations--one which serves static assets and the api and another which forwards the assets built by `npm start` and the api to the same origin.

I think this is the right approach for the short-term, but if we get help we can move to a different solution.


schmmd:
@codeviking happy to hear your thoughts.  You're welcome to review our setup any time.

allennlp/service/server_sanic.py
```diff
@@ -32,6 +33,7 @@ def run(port: int, workers: int,
     print("Starting a sanic server on port {}.".format(port))
 
     app = make_app(static_dir)
+    cors = CORS(app, resources={r"/predict/*": {"origins": "localhost"}})
```
codeviking: > This package has a simple philosophy, when you want to enable CORS, you wish to enable it for all use cases on a domain. This means no mucking around with different allowed headers, methods, etc. By default, submission of cookies across domains is disabled due to the security implications, please see the documentation for how to enable credential’ed requests, and please make sure you add some sort of CSRF protection before doing so!

That's from `sanic_cors`.

Sweet!

allennlp/service/server_sanic.py
```diff
@@ -69,9 +71,13 @@ def _caching_prediction(model: Predictor, data: str) -> JsonDict:
         """
         return model.predict_json(json.loads(data))
 
-    @app.route('/predict/<model_name>', methods=['POST'])
+    @app.route('/predict/<model_name>', methods=['POST', 'OPTIONS'])
     async def predict(req: request.Request, model_name: str) -> response.HTTPResponse:  # pylint: disable=unused-variable
         """make a prediction using the specified model and return the results"""
+
+        if req.method == "OPTIONS":
```
DeNeutoy: What is this doing?

joelgrus:
we should probably add a `log.warning` to `allennlp/commands/serve.py` if you specify a port other than 8000 ("warning: the demo requires port 8000") or something

allennlp/service/server_sanic.py
schmmd: First off, CORS is pretty weird overall.  It doesn't prevent anything outright--it just makes it a bit harder to do.  This keeps bad javascript coders from hacking too easily.

As I understand it, when making a cross-origin request the browser first sends an OPTION request to retrieve CORS headers.  I need to handle this request specifically, because the rest of the method expects a POST body and sends the results to AllenNLP.

PR #446, opened 2017-10-26
add Subcommand abstraction + overrides

joelgrus:
this addresses @OyvindTafjord 's need to have a custom evaluate command without having to rewrite his own copy of `main()`, hopefully it's useful beyond that as well.


joelgrus:
hooray, my "check docs" check failed (as it should have)

PR #448, opened 2017-10-26
add class variable for logging

DeNeutoy:
@tusharkhot FYI this is what we went with - using a `LabelField` is fine, and now with this logging change you shouldn't get all the annoying warnings.


tusharkhot:
Thanks!! 👍

allennlp/data/fields/label_field.py
```diff
@@ -35,18 +36,26 @@ class LabelField(Field[numpy.ndarray]):
...
+
     def __init__(self,
                  label: Union[str, int],
                  label_namespace: str = 'labels',
                  skip_indexing: bool = False) -> None:
         self.label = label
         self._label_namespace = label_namespace
         self._label_id = None
-        if not (self._label_namespace.endswith("labels") or self._label_namespace.endswith("tags")):
+        if not (self._label_namespace.endswith("labels") or self._label_namespace.endswith(
```
matt-gardner: This is a pretty messy line.  I'd recommend simplifying this to:

```python
def __init__(self, ...):
  if self._should_warn_for_namespace(label_namespace):
    ...

def _should_warn_for_namespace(self, label_namespace):
  if label_namespace.endswith('labels') or label_namespace.endswith('tags'):
    if label_namespace not in self._already_warned_namespaces:
      logger.warning("...")
      self._already_warned_namespaces.add(label_namespace)
```

This has the benefit of also moving the multi-line warning statement out of the constructor, which will make this class easier to read.

PR #451, opened 2017-10-30
permalinks

joelgrus:
ok, I think this is ready for review.

--

**on the backend:**

allennlp/service/db.py contains code for interacting with postgres
allennlp/service/permalink.py contains helpers for converting between ids and slugs

allennlp/service/server_sanic.py now has a `permadata/xyz` POST endpoint that returns the permanent data (inputs + outputs + model name) for the permalink slug. also, when you call the `predict` endpoint it will (if a database is configured) write the data to the database and return the permalink slug as part of the response

also, now that the model names (and permalink slugs) are part of the URL, I had to add new routes for those that just return index.html. when we add new models, we'll have to update those routes too.

**on the frontend:**

I ended up having to rewrite large chunks of the front-end. Now when you go to the demo it will redirect you to

`/semantic-role-labeler/`

and the header links will correspondingly take you to 

`/textual-entailment/` and `/machine-comprehension/`

Once you run an example, in addition to getting the results the URL will change to

`/semantic-role-labeler/xyz`

which is the permalink for the result, can be copied, pasted, shared, and so on.

If you go directly to one of those permalink results, you'll get the same experience, but the inputs and outputs will be prepopulated with whatever was saved to the database.

--

IF you run the demo without a database (which is determined by four environment variables), then it will work without permalinks. When you get a prediction the URL won't change but will just stay `/semantic-role-labeler` or whatever.


allennlp/service/server_sanic.py
```diff
@@ -74,10 +77,49 @@ def _caching_prediction(model: Predictor, data: str) -> JsonDict:
...
+        # If we don't have a database configured, there are no permalinks.
+        if demo_db is None:
+            raise ServerError('Permalinks are not enabled', 400)
+
+        # Convert the provided slug to an integer id.
+        slug = req.json["slug"]
+        perma_id = slug_to_int(slug)
+        if perma_id is None:
+            # Malformed slug
+            raise ServerError('Unrecognized permalink', 400)
```
schmmd: Nit: I'd add the slug to the error message. `'Unrecognized permalink: ' + slug`

allennlp/service/server_sanic.py
```diff
@@ -74,10 +77,49 @@ def _caching_prediction(model: Predictor, data: str) -> JsonDict:
...
+        try:
+            permadata = demo_db.get_result(perma_id)
+        except Exception:
+            # TODO(joelgrus) more specific exception type
+            # Error talking to database
+            raise ServerError('Database trouble', 500)
+
+        if permadata is None:
+            # No data found, invalid id?
+            raise ServerError('Unrecognized permalink', 400)
```
schmmd: `'Unrecognized permalink: ' + perma_id`

kubernetes-webdemo.yaml
```diff
@@ -18,6 +18,27 @@ spec:
           # See
           # https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
           # for documentation on the resources section.
+          env:
+            - name: DEMO_POSTGRES_HOST
```
schmmd: How do you enter these secrets into Kubernetes?  We should probably document this--although I'm fine starting with documenting the web demo in general.

joelgrus: I haven't tried yet, but when I do I'm going to follow this:

https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually

joelgrus:
added Sam to review the react stuff

demo/public/css/main.css
```diff
@@ -848,7 +848,7 @@ header {
   height: .375em;
   background: #63a7d4;
   display: block;
-  bottom: 0;
+  bottom: -20px;
```
codeviking: I generally try to avoid negative offsets, as it usually means I'm doing something wrong in the first place (and that things are even less likely to work across browsers).

That said, I don't think you're the original author of the CSS, so if it works (and you've done some light x-browser testing to verify), this is likely a-ok :).

demo/src/App.js
```diff
@@ -1,48 +1,117 @@
...
+
+    // Our components will be using history.push to change the location,
+    // and they will be attaching any `requestData` and `responseData` updates
+    // to the location object. That means we need to listen for location changes
+    // and update our state accordingly.
+    props.history.listen((location, action) => {
+      const { state } = location;
+      if (state) {
+        const { requestData, responseData } = state;
+        this.setState({requestData: requestData, responseData: responseData})
```
codeviking: You should be able to do just:

```
this.setState({ requestData, responseData })
```

Save a few 'chars.

demo/src/App.js
```diff
@@ -1,48 +1,117 @@
...
+      // Make an ajax call to get the permadata,
+      // and then use it to update the state.
+      fetch('http://localhost:8000/permadata', {
+        method: 'POST',
+        headers: {
+          'Accept': 'application/json',
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({"slug": slug})
+      }).then(function(response) {
```
codeviking: This could be an arrow function too, `response => response.json()`.

demo/src/App.js
```diff
@@ -1,48 +1,117 @@
...
+        return response.json();
+      }).then((json) => {
+        const { requestData, responseData } = json;
+        this.setState({requestData: requestData, responseData: responseData});
```
codeviking: Same as above, you can omit the property names if the local name is the same.

```
this.setState({ requestData, responseData })
```

demo/src/App.js
```diff
@@ -1,48 +1,117 @@
...
+      }).catch((error) => {
+        this.setState({outputState: "error"});
+        throw error; // todo(michaels): is this right?
```
codeviking: Since you've got a UI state conveying the error, I would just do `console.error(error)`.

joelgrus:
the main reason is that the inputs could (in theory) be arbitrarily long

demo/public/css/main.css
joelgrus: this was the only way I could get it to work! (the original code just used manual anchors, after I changed it to react-router <Link>s the formatting broke)

kubernetes-webdemo.yaml
joelgrus: ok, I did this, check out

```
kubectl get secret allennlp-demo-postgres --namespace allennlp -o yaml
```

allennlp/service/db.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def get_result(self, perma_id: int) -> Permadata:
+        """
+        Gets the result from the database with the given id.
+        Returns ``None`` if no such result.
+        """
+        raise NotImplementedError
+
+    @classmethod
+    def from_environment(cls) -> Optional['DemoDatabase']:
```
schmmd: Docstring?

allennlp/service/server_sanic.py
```diff
@@ -135,4 +193,18 @@ def _caching_prediction(model: Predictor, data: str) -> JsonDict:
         """list the available models"""
         return response.json({"models": list(app.predictors.keys())})
 
+    # As a SPA, we need to return index.html for /model-name and /model-name/permalink
```
schmmd: @codeviking I remember you cautioning against SPAs at one point.  Do any of your concerns apply here?

schmmd:
@joelgrus looks great!  Should we add scripts to create the database in case we need to recreate it?

allennlp/service/server_sanic.py
codeviking: Your demo is a bit closer to something that's SPA appropriate.  S2 less so, y'all are good to go :)

PR #453, opened 2017-10-31
add batch prediction to model and predictor

DeNeutoy:
- Adds batch prediction to the `Model` and `Predictor`. 
- I left adding batch prediction to the sanic server because that seems fairly separate, although easily possible.

Currently, `Model.forward_on_batch` returns a dict like {"key": [list of predictions for batch]}, but I could imagine it might be better to have [{"key": batch1prediction}, {"key": batch2prediction}] as then the output from a single element in the batch would match the output of `Model.forward_on_instance`. If people have strong opinions either way i'll change it.

@lucylw FYI - this is what you need to have the stream prediction from JSON you wanted.

### DIFF
--- a/allennlp/models/model.py
+++ b/allennlp/models/model.py
@@ -3,13 +3,13 @@
 an AllenNLP model.
 """
 
-from typing import Dict, Union
+from typing import Dict, Union, List
 import os
 import logging
 
 from allennlp.common.params import Params
 from allennlp.common.registrable import Registrable
-from allennlp.data import Instance, Vocabulary
+from allennlp.data import Instance, Vocabulary, Dataset
 from allennlp.nn.util import arrays_to_variables, device_mapping
 from allennlp.nn.regularizers import RegularizerApplicator
 
@@ -21,6 +21,7 @@ logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 # When training a model, many sets of weights are saved. By default we want to
 # save/load this set of weights.
 _DEFAULT_WEIGHTS = "best.th"
+
 
 class Model(torch.nn.Module, Registrable):
     """
@@ -116,7 +117,32 @@ class Model(torch.nn.Module, Registrable):
         outputs = self.decode(self.forward(**model_input))
 
         for name, output in list(outputs.items()):
+            # We are predicting on a single instance and we added a batch
+            # dimension, so here we remove it.
             output = output[0]
+            if isinstance(output, torch.autograd.Variable):
+                output = output.data.cpu().numpy()
+            outputs[name] = output
+        return outputs
+
+    def forward_on_batch(self, instances: List[Instance],
+                         cuda_device: int) -> Dict[str, numpy.ndarray]:
+        """
+        Takes a list of  :class:`~allennlp.data.instance.Instance`s, converts that text into
+        arrays using this model's :class:`Vocabulary`, passes those arrays through
+        :func:`self.forward()` and :func:`self.decode()` (which by default does nothing)
+        and returns the result.  Before returning the result, we convert any
+        ``torch.autograd.Variables`` or ``torch.Tensors`` into numpy arrays.
+        """
+
+        dataset = Dataset(instances)
+        dataset.index_instances(self.vocab)
+        model_input = arrays_to_variables(dataset.as_array_dict(),
+                                          cuda_device=cuda_device,
+                                          for_training=False)
+        outputs = self.decode(self.forward(**model_input))
+
+        for name, output in list(outputs.items()):
             if isinstance(output, torch.autograd.Variable):
                 output = output.data.cpu().numpy()
             outputs[name] = output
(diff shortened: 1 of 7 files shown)


schmmd:
@DeNeutoy is the impact of this faster bulk predictions?

DeNeutoy:
Yes, considerably so, depending on the model. I'll add it to the predict CLI, I forgot about that part.

allennlp/models/model.py
```diff
@@ -116,12 +117,37 @@ def forward_on_instance(self, instance: Instance, cuda_device: int) -> Dict[str,
...
         for name, output in list(outputs.items()):
+            # We are predicting on a single instance and we added a batch
+            # dimension, so here we remove it.
             output = output[0]
             if isinstance(output, torch.autograd.Variable):
                 output = output.data.cpu().numpy()
             outputs[name] = output
         return outputs
 
+    def forward_on_batch(self, instances: List[Instance],
```
matt-gardner: `forward(inputs)` already operates on batches, so this name is potentially confusing.  I think something like `forward_on_instances` would be better, or `forward_on_instance_batch` if you really want to keep the word "batch" in there.

matt-gardner: Also, do you have strong feelings about putting the first argument on the same line as `self`?  I kind of think it should go on the next line.

allennlp/models/model.py
```diff
@@ -116,12 +117,37 @@ def forward_on_instance(self, instance: Instance, cuda_device: int) -> Dict[str,
...
+                         cuda_device: int) -> Dict[str, numpy.ndarray]:
+        """
+        Takes a list of  :class:`~allennlp.data.instance.Instance`s, converts that text into
+        arrays using this model's :class:`Vocabulary`, passes those arrays through
+        :func:`self.forward()` and :func:`self.decode()` (which by default does nothing)
+        and returns the result.  Before returning the result, we convert any
+        ``torch.autograd.Variables`` or ``torch.Tensors`` into numpy arrays.
```
matt-gardner: You could note in here the similarity between this method and `forward_on_instance`, that this will actually perform batched computation on the GPU.

allennlp/service/predictors/predictor.py
```diff
@@ -25,6 +26,25 @@ def _json_to_instance(self, json: JsonDict) -> Instance:
         """
         raise NotImplementedError
 
+    def predict_batch_json(self, inputs: List[JsonDict], cuda_device: int = -1) -> JsonDict:
+        instances = self._batch_json_to_instance(inputs)
+        outputs = self._model.forward_on_batch(instances, cuda_device)
+        return sanitize(outputs)
+
+    def _batch_json_to_instance(self, json: List[JsonDict]) -> List[Instance]:
```
matt-gardner: `_batch_json_to_instances`, not `to_instance`.

allennlp/service/predictors/semantic_role_labeler.py
```diff
@@ -48,6 +48,10 @@ def _json_to_instance(self, json: JsonDict) -> Instance:
         raise RuntimeError("this should never be called")
 
     @overrides
+    def _batch_json_to_instance(self, json: List[JsonDict]) -> List[Instance]:
+        raise NotImplementedError("The SRL Predictor does not currently support batch prediction.")
```
matt-gardner: This seems like a pretty good argument for just having this use `_json_to_instance` - what's the reason we can't do that?

tests/service/predictors/decomposable_attention_test.py
```diff
@@ -39,3 +39,41 @@ def test_uses_named_inputs(self):
         sumexps = sum(exps)
         for e, p in zip(exps, label_probs):
             assert e / sumexps == approx(p)
+
+    def test_batch_prediction(self):
```
matt-gardner: Not really complaining here, but this was more work than you needed.  Because the functionality is in the base class, you really only need one of these models to have the test for batch prediction.  It could even be in a `predictor_test.py` file, using an arbitrary model for the test.

joelgrus:
is it worth having separate `forward_on_instance` and `forward_on_instances` vs just having the latter and feeding a "batch" of one instance into it?

allennlp/service/predictors/semantic_role_labeler.py
DeNeutoy: We want to have the sentence and verbs in the output, but we don't want to have to extract them from the model.

matt-gardner:
@joelgrus, no strong opinion there. What you suggest is a simpler API, but puts a _little_ more work on the caller. I could go either way.

joelgrus:
aren't the primary callers just `command.predict` and `command.serve`

DeNeutoy:
There is some complexity around handling the batch dimension both of the inputs and outputs and I think it's cleaner if its separate for now, although I don't hold this opinion particularly strongly. Also removing `forward_on_instance` would not be backwards compatible - don't know if we care about that yet?

Issue #428, opened 2017-10-21
Embedding POS/NER/arbitrary features


EgorLakomkin:
First of all, I want to apologize if this question is not suitable for an issue tracker (but I could not find any other place).

I am trying to incorporate POS-tag and NER-type features (one-hot encoded vectors mainly) to the word&character embeddings and I wanted to ask what is the best way to do this using this library?

Thank you in advance!

matt-gardner:
Using POS tags is currently easy, but we don't have NER tags currently available for this.  It was easy to add, so I just submitted #430 that adds it.  That PR also modifies one of the test configuration files to show how to use POS tag and NER embeddings as additional components of the word representations.  Just look at the modifications to `tests/fixtures/encoder_decoder/simple_seq2seq/experiment.json` in that PR, and make similar modifications to your configuration file.  Let me know if you have any questions.

EgorLakomkin:
Thank you very much, I will take a look at your PR. 

I am still curious how it is possible to incorporate arbitrary features (e.g. binary flag if token is capitalized,  token frequence etc). As I understood TextFieldEmbedders operate on a list of indices, but somehow it would be nice to map  Token -> feature vector directly?

matt-gardner:
For this, there are two options.

The quickest option is to do something like what the SRL model does to incorporate the verb indicator.  The `DatasetReader` adds a `SequenceLabelField` with binary values indicating whether each token is the verb to find arguments for: https://github.com/allenai/allennlp/blob/ed37843d528b0902d73e4d5dde931812fd65835d/allennlp/data/dataset_readers/semantic_role_labeling.py#L277

The model then takes that field, embeds it, and concatenates it manually to the word embedding:
https://github.com/allenai/allennlp/blob/ed37843d528b0902d73e4d5dde931812fd65835d/allennlp/models/semantic_role_labeler.py#L119-L122

You could do this for each binary feature you want, but just skip the feature embedding.

If you have a lot of binary features you want to add, though, this could get annoying.  So the second option is to write your own `TokenIndexer` that converts word tokens into feature vectors using whatever feature extractors you want to use, and a `TokenEmbedder` that just passes through the given feature vector.  If you want to go this route, I'm happy to look at a work-in-progress PR to let you know if you're doing it the right way.

EgorLakomkin:
Great, thank you for the very detailed answer. I will try to make a PR for a second option.

EgorLakomkin:
[...]

PR #454, opened 2017-11-01
adding token indexer for custom feature extraction + bypass embedder

EgorLakomkin:
This PR contains few things that I found very useful while experimenting with allennlp library for a QA system development. I mentioned them here #428 with the ideas and help from @matt-gardner .
Mainly the idea is to introduce token indexer that is able to return a feature vector (just a numpy array) for a given token using a provided feature extraction function. For example, if one would need to add a feature like if the token start with lowercase or uppercase letter. In addition to such a token indexer, an embedder (BypassEmbedder) that just propagates the input keeping it untouched.

If it is within the scope of the library I would be happy to polish it to get it merged.

### DIFF
--- a/allennlp/data/token_indexers/token_feature_indexer.py
+++ b/allennlp/data/token_indexers/token_feature_indexer.py
@@ -1,0 +1,54 @@
+import logging
+from typing import Dict, List, Set
+import numpy as np
+from allennlp.common.checks import ConfigurationError
+from overrides import overrides
+
+from allennlp.common.util import pad_sequence_to_length
+from allennlp.common import Params
+from allennlp.data.vocabulary import Vocabulary
+from allennlp.data.tokenizers.token import Token
+from allennlp.data.token_indexers.token_indexer import TokenIndexer
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+class FeatureIndexer(TokenIndexer):
+
+    def __init__(self, feature_extractor, default_value) -> None:
+        self.feature_extractor = feature_extractor
+        self.default_value = default_value
+        self.feature_dim = self.default_value.shape
+
+    @overrides
+    def token_to_indices(self, token: Token, vocabulary: Vocabulary = None):
+        features = self.feature_extractor(token)
+        if not isinstance(features, np.ndarray):
+            raise ConfigurationError('Feature function for FeatureIndexer should return numpy array')
+        if features.shape != self.feature_dim:
+            raise ConfigurationError("feature function should return " \
+            "vectors of dimensionality {}, but got {}".format(self.feature_dim, features.shape))
+        return features
+
+    @overrides
+    def count_vocab_items(self, token, counter):
+        return
+
+    @overrides
+    def get_padding_token(self):
+        return []
+
+    @classmethod
+    def from_params(cls, params: Params):
+        return
+
+    @overrides
+    def get_padding_lengths(self, token):
+        return {}
+
+    @overrides
+    def pad_token_sequence(self,
+                           tokens,
+                           desired_num_tokens,
+                           padding_lengths):
+        return pad_sequence_to_length(tokens, desired_num_tokens,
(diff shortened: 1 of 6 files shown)


tests/data/token_indexers/token_feature_indexer.py
```diff
@@ -0,0 +1,30 @@
...
+def lowercase_feature_extractor(token: Token):
+    if token.text[0].isupper():
+        return np.array([1.0], dtype=np.float32)
+    else:
+        return np.array([0.0], dtype=np.float32)
+
+class TestPosTagIndexer(AllenNlpTestCase):
+    def setUp(self):
+        super(TestPosTagIndexer, self).setUp()
+        self.tokenizer = SpacyWordSplitter(pos_tags=True)
```
matt-gardner: You're not using POS tags in this test, so I wouldn't load spacy using them.  Just the default spacy is fine.

tests/data/token_indexers/token_feature_indexer.py
```diff
@@ -0,0 +1,30 @@
...
+from allennlp.data.tokenizers.word_splitter import SpacyWordSplitter
+import numpy as np
+
+def lowercase_feature_extractor(token: Token):
+    if token.text[0].isupper():
+        return np.array([1.0], dtype=np.float32)
+    else:
+        return np.array([0.0], dtype=np.float32)
+
+class TestPosTagIndexer(AllenNlpTestCase):
```
matt-gardner: Change the name of the test to match what you're testing.

tests/data/token_indexers/token_feature_indexer.py
```diff
@@ -0,0 +1,30 @@
...
+    def setUp(self):
+        super(TestPosTagIndexer, self).setUp()
+        self.tokenizer = SpacyWordSplitter(pos_tags=True)
+
+
+
+    def test_count_extract_features(self):
```
matt-gardner: This name here is wrong.

tests/data/token_indexers/token_feature_indexer.py
```diff
@@ -0,0 +1,30 @@
+# pylint: disable=no-self-use,invalid-name
```
matt-gardner: The name of this file needs to include the word "test" for pytest to actually discover that it contains tests.

tests/modules/text_field_embedders/custom_feature_embedder.py
```diff
@@ -0,0 +1,50 @@
...
+
+def lowercase_feature_extractor(token: Token):
+    if token.text[0].isupper():
+        return np.array([1.0], dtype=np.float32)
+    else:
+        return np.array([0.0], dtype=np.float32)
+
+
+
+class TestCustomFeatureFieldEmbedder(AllenNlpTestCase):
```
matt-gardner: This should be `TestBypassEmbedder(AllenNlpTestCase)`, and the name of the file should match.

matt-gardner: Ah, I see that you have a separate file for that.  I'd probably just put these tests in that file.  Though, actually, these tests are probably unnecessary, as this is just an integration test on all of the components of the featurized token indexer / embedder.  We have plenty of integration tests for this pipeline, so as long as the components are tested well enough, an additional integration test isn't necessary.

allennlp/data/token_indexers/token_feature_indexer.py
```diff
@@ -0,0 +1,54 @@
...
+    @overrides
+    def count_vocab_items(self, token, counter):
+        return
+
+    @overrides
+    def get_padding_token(self):
+        return []
+
+    @classmethod
+    def from_params(cls, params: Params):
```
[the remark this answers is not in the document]
joelgrus: If you think it's worth having in the core library, we can merge it. Otherwise, one of my Q4 deliverables is to come up with a strategy for handling contributed models and so on, I can accelerate that if needed.

PR #458, opened 2017-11-02
Wikitables dataset reader and a knowledge graph field

pdasigi:
This PR contains the actual wikitables dataset reader, which creates a dataset with instances that contain three fields: `utterance`, `context` and `action_sequences`. This should be enough to train a bare bones semantic parser without entity linking.

This PR also includes a `KnowledgeGraphField` that is used for the `context`, but the field is not very useful at this point. To use it for entity linking, we still need token indexers for entities, and an entity embedder, which I will include in a different PR.


allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
```diff
@@ -0,0 +1,141 @@
...
+    def _make_action_sequence_field(action_sequence: List[str]) -> ListField:
+        action_sequence.insert(0, START_SYMBOL)
+        action_sequence.append(END_SYMBOL)
+        return ListField([LabelField(action, label_namespace='actions') for action in action_sequence])
+
+    @staticmethod
+    def _parse_line_as_lisp(lisp_string: str) -> Dict[str, LispValueType]:
+        parsed_info = {}
+        input_nested_list = pyparsing.OneOrMore(pyparsing.nestedExpr()).parseString(lisp_string).asList()[0]
+        for key_value in input_nested_list:
```
matt-gardner: Can you do `for key, value in input_nested_list`?

allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
```diff
@@ -0,0 +1,141 @@
...
+            elif key_value[0] == "targetValue":
+                parsed_info["targetValue"] = [x[1] for x in key_value[1][1:]]
+        assert all([x in parsed_info for x in ["id", "utterance", "context"]]), "Invalid format"
+        return parsed_info
+
+    @classmethod
+    def from_params(cls, params: Params) -> 'WikitablesDatasetReader':
+        tables_directory = params.pop('tables_directory')
+        dpd_output_directory = params.pop('dpd_output_directory', None)
+        utterance_tokenizer_type = params.pop('utterance_tokenizer', None)
```
matt-gardner: You can just do this: https://github.com/allenai/allennlp/blob/c4d80d0b68d575e831cef2e249b844db15b5e228/allennlp/data/dataset_readers/reading_comprehension/triviaqa.py#L166-L167

allennlp/data/fields/knowlege_graph_field.py
```diff
@@ -0,0 +1,58 @@
...
+        for entity in self._knowledge_graph.get_all_entities():
+            indexed_entity = {indexer_name: indexer.token_to_indices(entity, vocab)
+                              for indexer_name, indexer in self._token_indexers.items()}
+            entity_arrays[entity] = indexed_entity
+        self._indexed_entities = entity_arrays
+
+    @overrides
+    def get_padding_lengths(self) -> Dict[str, int]:
+        # pylint: disable=no-self-use
+        return {}
```
matt-gardner: I'm not confident this is correct.  Can you add some tests for this `Field` to show the intended use?

allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
```diff
@@ -0,0 +1,141 @@
...
+                    continue
+                instances.append(self.text_to_instance(line, True))
+        if not instances:
+            raise ConfigurationError("No instances read!")
+        return Dataset(instances)
+
+    @overrides
+    def text_to_instance(self,  # type: ignore
+                         input_lisp_string: str,
+                         for_train: bool = False) -> Instance:
```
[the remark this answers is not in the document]
pdasigi: Done.

allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
```diff
@@ -0,0 +1,141 @@
...
+        # pylint: disable=arguments-differ
+        parsed_info = self._parse_line_as_lisp(input_lisp_string)
+        tokenized_utterance = self._utterance_tokenizer.tokenize(parsed_info["utterance"])
+        utterance_field = TextField(tokenized_utterance, self._utterance_token_indexers)
+        table_filename = "%s/%s" % (self._tables_directory, parsed_info["context"])
+        # We want the TSV file, but the ``*.examples`` files typically point to CSV.
+        table_filename = table_filename.replace(".csv", ".tsv")
+        table_knowledge_graph = TableKnowledgeGraph.read_from_file(table_filename)
```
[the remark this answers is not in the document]
pdasigi: Done. I added a `TableKnowledgeGraph.read_from_json` method, and made the table related input to this method be either a string (for filenames) or a dict (for actual table content).

allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
pdasigi: Done.

allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
pdasigi: Done.

allennlp/data/fields/knowlege_graph_field.py
pdasigi: You're right. This is incorrect. I was going to add a TODO to implement this correctly after I have the relevant entity token indexers ready, but forgot. I implemented this method now, but can add tests only after I implement the two entity indexers (name and type), and I plan to do that in the next PR with entity embedding and linking. Is that fine? I added a TODO for this.

allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
```diff
@@ -0,0 +1,171 @@
...
+                sempre_forms = [line.strip().decode('utf-8') for line in gzip.open(dpd_output_filename)]
+                instances.append(self.text_to_instance(utterance, table_filename, sempre_forms))
+        if not instances:
+            raise ConfigurationError("No instances read!")
+        return Dataset(instances)
+
+    @overrides
+    def text_to_instance(self,  # type: ignore
+                         utterance: str,
+                         table_info: TableInfoType,
```
matt-gardner: You can just do `Union[str, JsonDict]` here, instead of creating a type that's only used once.  I'd recommend the same for your `LispValueType`.

pdasigi: Done.

PR #459, opened 2017-11-03
Bug fixes in loss computation, speed improvements from profiling

matt-gardner:
I did some profiling to see where the bottlenecks were, and I fixed some of them.  This got me a 2-2.5x speedup.  I don't think there's much more gain to be had apart from with batching.

I also went over the computation we're doing in detail with @DeNeutoy, because I was seeing some odd behavior trying to get this to learn.  We discovered that we weren't using the current input at timestep t as part of the attention over the question, and that we weren't doing the output projection as nicely as we could, either.  I fixed those issues here, and tried to have the model learn my simple word segmentation problem.  I made it even simpler, with just one corrupted example, but I think I'm asking the model to learn too much from too small a dataset.  It clearly is able to learn something, but after adding the first space it gets pretty confused.  I'm going to call this good enough - it looks like the model works as intended, and with a larger dataset with easier actions, we should be good.

Next up: seeing if we can do batched computation, both on the decoder states, and on the input questions.

### DIFF
--- a/allennlp/models/encoder_decoders/wikitables_semantic_parser.py
+++ b/allennlp/models/encoder_decoders/wikitables_semantic_parser.py
@@ -1,4 +1,4 @@
-from typing import Dict, Generator, List, Set
+from typing import Dict, Generator, List, Set, Tuple
 
 from overrides import overrides
 
@@ -70,16 +70,16 @@ class WikiTablesSemanticParser(Model):
                  action_embedding_dim: int,
                  attention_function: SimilarityFunction) -> None:
         super(WikiTablesSemanticParser, self).__init__(vocab)
-        self._source_embedder = question_embedder
+        self._question_embedder = question_embedder
         self._encoder = encoder
         self._decoder_trainer = decoder_trainer
         self._beam_search = decoder_beam_search
         self._max_decoding_steps = max_decoding_steps
         self._action_namespace = action_namespace
 
-        num_actions = self.vocab.get_vocab_size(self._action_namespace)
         self._start_index = self.vocab.get_token_index(START_SYMBOL, self._action_namespace)
         self._end_index = self.vocab.get_token_index(END_SYMBOL, self._action_namespace)
+        num_actions = self.vocab.get_vocab_size(self._action_namespace)
         self._decoder_step = WikiTablesDecoderStep(self._encoder.get_output_dim(),
                                                    action_embedding_dim,
                                                    num_actions,
@@ -105,15 +105,15 @@ class WikiTablesSemanticParser(Model):
            work, and will be stripped in this method with a ``squeeze``.
         """
         # (batch_size, input_sequence_length, encoder_output_dim)
-        embedded_input = self._source_embedder(question)
-        source_mask = util.get_text_field_mask(question)
+        embedded_input = self._question_embedder(question)
+        question_mask = util.get_text_field_mask(question)
         batch_size = embedded_input.size(0)
         if batch_size != 1:
             raise ConfigurationError("This implementation does not currently handle batched "
                                      "inputs.  Please set the batch size of your dataset reader "
                                      "to 1.")
 
-        encoder_outputs = self._encoder(embedded_input, source_mask)
+        encoder_outputs = self._encoder(embedded_input, question_mask)
 
         final_encoder_output = encoder_outputs[:, -1]  # (batch_size, encoder_output_dim)
         decoder_context = Variable(encoder_outputs.data.new(batch_size, self._encoder.get_output_dim())
@@ -128,11 +128,12 @@ class WikiTablesSemanticParser(Model):
             target_mask = None
 
         initial_score = Variable(embedded_input.data.new(1).fill_(0))
+        initial_state = (final_encoder_output, decoder_context, None)
         initial_state = WikiTablesDecoderState([],
                                                initial_score,
                                                encoder_outputs=encoder_outputs,
-                                               encoder_output_mask=source_mask.float(),
-                                               hidden_state=(final_encoder_output, decoder_context),
+                                               encoder_output_mask=question_mask.float(),
+                                               hidden_state=initial_state,
                                                end_index=self._end_index)
         if self.training:
             return self._decoder_trainer.decode(initial_state,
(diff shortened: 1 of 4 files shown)


allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -270,36 +277,52 @@ def take_step(self,
...
+                          state: WikiTablesDecoderState,
+                          log_probs: Variable,
+                          new_hidden_state: Tuple[Variable, Variable],
+                          allowed_actions: Set[int]) -> Generator[WikiTablesDecoderState, None, None]:
+        sorted_log_probs, sorted_actions = log_probs.sort(dim=-1, descending=True)
+        sorted_actions = sorted_actions.data.cpu().numpy().tolist()
         valid_actions = state.get_valid_actions()
-        for _, log_prob, action in sorted_log_probs:
+        for i in range(sorted_log_probs.size(1)):
+            action = sorted_actions[0][i]
```
DeNeutoy: This is `sorted_actions[0][i]` just because we have a batch size of 1, right? Could use a comment.

allennlp/nn/decoding/beam_search.py
```diff
@@ -35,10 +35,13 @@ def search(self,
...
                             finished_states.append(next_state)
                         next_states.append(next_state)
+            # TODO(mattg): do this sort on the GPU, not on the CPU, and copy once.
             to_sort = [(-state.score.data[0], state) for state in next_states]
             to_sort.sort(key=lambda x: x[0])
             states = [state[1] for state in to_sort[:self._beam_size]]
-        finished_to_sort = [(-state.score.data[0], state) for state in finished_states]
+        # The time this one takes is pretty negligible, no particular need to optimize this yet.
+        # Maybe with a larger beam size...
+        finished_to_sort = [(-state.score.data[0] / len(state.action_history), state) for state in finished_states]
```
DeNeutoy: I think this length normalisation is something that should happen in the `MaxMarginalLikelihood` decoder trainer as well?

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -270,36 +277,52 @@ def take_step(self,
...
-                                              state.encoder_outputs,
-                                              state.encoder_output_mask)
-        # (batch_size, encoder_output_dim)
-        attended_input = util.weighted_sum(state.encoder_outputs, input_weights)
-        # (batch_size, encoder_output_dim + target_embedding_dim)
-        decoder_input = torch.cat((attended_input, embedded_input), -1)
+
+        if attended_question is None:
+            # Unless this is the first timestep, we've already computed an attention over the input
+            # given the current hidden state.  If it's the first timestep, we compute an initial
```
pdasigi: This comment could be clearer. I didn't get `attended_question` being `None` implies that this the first time step from this comment. How about replacing the first sentence with "This means this is the first time step."?

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -128,11 +128,12 @@ def forward(self,  # type: ignore
             target_mask = None
 
         initial_score = Variable(embedded_input.data.new(1).fill_(0))
+        initial_state = (final_encoder_output, decoder_context, None)
```
pdasigi: I think it would be cleaner to compute the initial `attended_question` here itself instead of having the conditional in `WikitablesDecoderStep.take_step`.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -250,18 +251,24 @@ def __init__(self,
...
-        self._decoder_input_dim = encoder_output_dim + action_embedding_dim
+        output_dim = encoder_output_dim
+        input_dim = output_dim
+        # Our decoder input will be the concatenation of the decoder hidden state and the previous
+        # action embedding, and we'll project that down to the decoder's `input_dim`, which we
+        # arbitrarily set to be the same as `output_dim`.
+        self._input_projection_layer = Linear(output_dim + action_embedding_dim, input_dim)
+        # Before making a prediction, we'll compute an attention over the input given our updated
+        # hidden state.  Then we concatenate that with the decoder state and project to
+        # `num_actions` to make a prediction.
```
pdasigi: Why concatenate the attention over input before projection? That is not a common practice, is it?

allennlp/models/encoder_decoders/wikitables_semantic_parser.py

allennlp/nn/decoding/beam_search.py
matt-gardner: We talked about this in person, but to add my response here: `MaxMarginalLikelihood` doesn't have anything that would need this, but perhaps the `DecoderStep` should set the score to be normalized by number of steps.  It turns out that this isn't necessary if you're guaranteed to have every state be at the same timestep, which is currently the case, but I'll add a note and a TODO about this.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
matt-gardner: Yes, that would be nice, but it breaks the abstractions we have here - you would need to know the parameters of the `DecoderStep` in order to do that in the parser's `forward` method directly.  I think having the clean separation of parameters and logic between the parser and the decoding step is worth having the slight ugliness on the first timestep.

matt-gardner: Well...  I suppose we could have a method that's `compute_question_attention`, or something, given a hidden state, and have `forward` call that method for the first iteration.  That'd probably work, and be cleaner.  I'll do that.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
matt-gardner: We had several long conversations about this in other places, but the relevant outcome of those conversations is that these equations are slightly different from what most papers I've seen have done, but they are now closer to current best practices than what we had before, and the difference between what's here and what's in other papers is this has an extra use of the attention from the previous timestep.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
matt-gardner: I'll fix this by adding a method and calling it in `forward`, as mentioned above.

matt-gardner:
The docs build is still failing, but Pradeep is fixing that in his PR, and this is merging to a branch, not to master, so I'm just going to merge this.

PR #460, opened 2017-11-03
More sanitize cases

OyvindTafjord:
I ran into a case where I needed to sanitize `Variable` and `LongTensor` (not caught by `Tensor`). I could get rid of the former in my output, but it shouldn't hurt to cover this case too?

### DIFF
--- a/allennlp/common/util.py
+++ b/allennlp/common/util.py
@@ -14,7 +14,7 @@ from allennlp.common.params import Params
 
 JsonDict = Dict[str, Any]  # pylint: disable=invalid-name
 
-def sanitize(x: Any) -> Any:  # pylint: disable=invalid-name
+def sanitize(x: Any) -> Any:  # pylint: disable=invalid-name,too-many-return-statements
     """
     Sanitize turns PyTorch and Numpy types into basic Python types so they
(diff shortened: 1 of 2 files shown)


allennlp/common/util.py
```diff
@@ -14,15 +14,17 @@
...
     Sanitize turns PyTorch and Numpy types into basic Python types so they
     can be serialized into JSON.
     """
     if isinstance(x, (str, float, int, bool)):
         # x is already serializable
         return x
-    elif isinstance(x, torch.Tensor):
+    elif isinstance(x, torch.autograd.Variable):
+        return sanitize(x.data)
+    elif isinstance(x, (torch.Tensor, torch.LongTensor)):
```
joelgrus: it might be good to just use `torch._TensorBase` here, b/c at some point someone will probably try to sanitize a `ByteTensor` or something and break it again.

OyvindTafjord:
Sounds good to use `torch._TensorBase`, unless it's considered unwise to rely on `_xxx` stuff in general?

OyvindTafjord:
OK, made the change to `_TensorBase`, added a test for some security in case this changes.

PR #461, opened 2017-11-03
List field padding fix

DeNeutoy:
- Fixes some padding logic in `TextFields` which assumed the number of tokens would be non-zero.
- Fixes padding of `ListFields[SequenceFields]` for non-`TextField` `SequenceFields` by making the `empty_field()` method respect the requirements of the sub field in the case of `IndexFields` and `SequenceLabelFields`. 
- Allows proper padding of `ListField[ListField[SequenceField]]`, where the first element of a batch may be empty. 
- Adds tests for the above cases.


tests/data/fields/list_field_test.py
```diff
@@ -19,6 +19,9 @@ def setUp(self):
         self.vocab.add_token_to_namespace("n", 'characters')
         self.vocab.add_token_to_namespace("t", 'characters')
         self.vocab.add_token_to_namespace("c", 'characters')
+        for label in ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k']:
+            self.vocab.add_token_to_namespace(label, 'labels')
+        self.vocab.add_token_to_namespace("c", 'characters')
```
matt-gardner: Looks like a duplicate line here.

tests/data/fields/list_field_test.py
```diff
@@ -29,13 +32,47 @@ def setUp(self):
...
+        list_field = ListField([self.field1, self.field2, self.empty_text_field])
+        list_field.index(self.vocab)
+        array_dict = list_field.as_array(list_field.get_padding_lengths())
+
+        numpy.testing.assert_array_equal(array_dict["words"], numpy.array([[2, 3, 4, 5, 0],
+                                                                           [2, 3, 4, 1, 5],
+                                                                           [0, 0, 0, 0, 0]]))
+
+    def test_list_field_can_handle_empty_index_fields(self):
+
```
matt-gardner: Extra lines here and above.

tests/data/fields/list_field_test.py
```diff
@@ -91,3 +140,35 @@ def test_as_array_can_handle_multiple_token_indexers(self):
                                                                             [1, 4, 1, 5, 1, 3, 1, 0, 0],
                                                                             [2, 3, 4, 5, 3, 4, 6, 3, 0],
                                                                             [0, 0, 0, 0, 0, 0, 0, 0, 0]]))
+
+    def test_as_array_can_handle_multiple_token_indexers_and_empty_fields(self):
+        # pylint: disable=protected-access
+        self.field1._token_indexers = self.words_and_characters_indexer
```
matt-gardner: Slight nit on this name - this is actually a dictionary with two indexers.  Could be confusing to someone trying to learn about the library by looking at test cases.

PR #462, opened 2017-11-03
Add sanic-cors to match requirements.txt

OyvindTafjord:
The setup.py file needs to be kept in sync with requirements.txt for pip install of allennlp (from latest) to work.


schmmd:
@OyvindTafjord sorry!

chiphuyen:
So the new version of sanic-cors seems to require sanic >= 0.7.0, so having sanic==0.6.0 will make it impossible to install via pip3, as we keep running into this problem:
```
sanic-plugins-framework 0.5.2.dev20180201 has requirement sanic>=0.7.0, but you'll have sanic 0.6.0 which is incompatible.
sanic-cors 0.9.4 has requirement sanic>=0.7.0, but you'll have sanic 0.6.0 which is incompatible.
```

Does anybody know how to get over this? Thanks a lot!

joelgrus:
what version of allennlp are you using? we removed the sanic dependency a long time ago.

PR #465, opened 2017-11-06
Coref Model

DeNeutoy:


DeNeutoy:
Ok, this is ready for review. You've already reviewed large chunks of this PR in several different places, so hopefully this final one won't be too arduous.

allennlp/data/dataset_readers/coreference_resolution/conll.py
```diff
@@ -0,0 +1,284 @@
...
+        return cls(token_indexers=token_indexers, max_span_width=max_span_width)
+
+    @staticmethod
+    def normalize_word(word):
+        if word == "/." or word == "/?":
+            return word[1:]
+        else:
+            return word
+
+    def handle_line(self, line, document_state) -> None:
```
joelgrus: Y U NO TYPE HINT

allennlp/nn/util.py
```diff
@@ -567,3 +568,174 @@ def logsumexp(tensor: torch.Tensor,
...
+        The number of identity buckets (those only holding a single value).
+    num_total_buckets : int, (default = 10)
+        The total number of buckets to bucket values into.
+
+    Returns
+    -------
+    A tensor of the same shape as the input, containing the indices of the buckets
+    the values were placed in.
+    """
+    logspace_idx = (distances.float().log()/math.log(2)).floor().long() + 3
```
joelgrus: all these steps feel slightly magical, might be nice to add comments about what they're doing

allennlp/training/metrics/conll_coref_scores.py
```diff
@@ -0,0 +1,180 @@
...
+    def get_precision(self):
+        return 0 if self.p_num == 0 else self.p_num / float(self.p_den)
+
+    def get_prf(self):
+        return self.get_precision(), self.get_recall(), self.get_f1()
+
+    @staticmethod
+    def b_cubed(clusters, mention_to_gold):
+        """
+        <https://pdfs.semanticscholar.org/cfe3/c24695f1c14b78a5b8e95bcbd1c666140fd1.pdf>
```
joelgrus: would be nice to add a sentence description for people who aren't going to click on the papers (same below as well)

allennlp/training/metrics/mention_recall.py
```diff
@@ -0,0 +1,34 @@
...
+@Metric.register("mention_recall")
+class MentionRecall(Metric):
+    def __init__(self) -> None:
+        self._num_gold_mentions = 0
+        self._num_recalled_mentions = 0
+
+    @overrides
+    def __call__(self, batched_top_spans, batched_metadata):
+        for top_spans, metadata in zip(batched_top_spans.data.tolist(), batched_metadata):
+            gold_mentions: Set[Tuple[int, int]] = set(m for c in metadata["clusters"] for m in c)
```
joelgrus: you can do set comprehensions with curly braces, slightly more pythonic maybe

```
{m for c in metadata["clusters"] for m in c}
```

allennlp/training/metrics/mention_recall.py
```diff
@@ -0,0 +1,34 @@
...
+            top_spans: Set[Tuple[int, int]] = set(tuple(s) for s in top_spans)
```
joelgrus: same

allennlp/data/dataset_readers/coreference_resolution/conll.py
DeNeutoy: + 1 for this type of code review

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,707 @@
...
+    @staticmethod
+    def _compute_antecedent_gold_labels(top_span_labels: torch.IntTensor,
+                                        antecedent_labels: torch.IntTensor):
+        """
+        Parameters
+        ----------
+        top_span_labels: torch.IntTensor
+            The cluster id label for every span.
+        antecedent_labels: torch.IntTensor
+            The cluster id label for every antecedent span.
```
matt-gardner: What's the shape of this?  I'm not really sure what this is, so I'm not confident in the logic below (not that I think it's wrong, I just don't understand it well enough to agree that it's correct).

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,707 @@
...
+
+    @staticmethod
+    def _compute_negative_marginal_log_likelihood(augmented_antecedent_scores: torch.FloatTensor,
+                                                  augmented_labels: torch.IntTensor,
+                                                  top_span_mask: torch.FloatTensor):
+        """
+        Parameters
+        ----------
+        augmented_antecedent_scores: torch.FloatTensor
+            The pairwise between every span and its possible antecedents.
```
matt-gardner: The pairwise what?

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,707 @@
...
+        augmented_labels: torch.IntTensor
+            A binary indicator for whether it is consistent with the data for a span to choose
+            an antecedent.
+        Returns
+        -------
+        negative_marginal_log_likelihood: torch.FloatTensor
+            The negative marginal loglikelihood of the gold cluster labels. This computes the
```
matt-gardner: I think for a lot of these methods you've put the description of the logic inside the description of the return value.  I think it makes more sense to put the description of the logic at the top, then give more detail as necessary in the description of each parameter and return value (or just give a name and type here, if the description above is already good enough).

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,707 @@
...
+            log of the sum of the probabilities of all antecedent predictions that would be
+            consistent with the data. The computation is performed in log-space for numerical
```
matt-gardner: Being a little more specific here would be good: each span `i` predicts a single antecedent `j`, but there might be several prior mentions `k` in the same coreference cluster that would be valid antecedents.  Our loss is the sum of the probability assigned to all valid antecedents.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,707 @@
...
+            stability.
+        """
+        # Shape: (batch_size, num_spans_to_keep, max_antecedents + 1)
+        gold_scores = augmented_antecedent_scores + augmented_labels.log()
```
matt-gardner: I don't think this pattern is common enough to leave it without a comment.  Describing what you're doing with the `.log()` here would be good.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,707 @@
...
+
+        # Shape: (batch_size, num_spans_to_keep)
+        marginalized_gold_scores = util.logsumexp(gold_scores, 2)
+        log_norm = util.logsumexp(augmented_antecedent_scores, 2)
```
matt-gardner: Have you done any softmax over antecedents already?  Or is this the only place that normalization happens?  E.g., if you do a masked log softmax on the augmented antecedent scores, then you should just be able to do `logsumexp` on the gold scores and that's your loss.  `masked_log_softmax` is probably faster than our `logsumexp`.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,707 @@
...
+        """
+        # Shape: (batch_size, num_spans_to_keep, max_antecedents)
+        antecedent_scores = self._antecedent_scorer(
+                self._antecedent_feedforward(pairwise_embeddings)).squeeze(-1)
+        antecedent_scores += top_span_mention_scores + antecedent_mention_scores
+        antecedent_scores += antecedent_log_mask
+
+        # Shape: (batch_size, num_spans_to_keep, 1)
+        shape = [antecedent_scores.size(0), antecedent_scores.size(1), 1]
+        dummy_scores = Variable(antecedent_scores.data.new().resize_(shape).fill_(0),
```
matt-gardner: `data.new(shape).fill_(0)`

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,707 @@
...
+            Mask representing non-padding entries of ``text_embeddings``.
+        span_starts : torch.IntTensor
+            The start of each span candidate.
+        span_ends : torch.IntTensor
+            The end of each span candidate.
+        Returns
+        -------
+        span_embeddings : torch.FloatTensor
+            An embedded representation of every candidate span with shape:
+            (batch_size, num_spans, context_layer.get_output_dim() * 2 + embedding_size + feature_size)
```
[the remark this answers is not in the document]
DeNeutoy: No, the embeddings of the span widths come from here: https://github.com/allenai/allennlp/pull/465/files#diff-623c9612bef01aebbea0952172da22afR83, whereas the embedding_dim comes from the embeddings of the words.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,707 @@
...
+        Parameters
+        ----------
+        mention_scores: torch.FloatTensor
+            The mention score for every candidate
+        num_spans_to_keep: int
+            The number of spans to keep when pruning.
+        Returns
+        -------
+        top_span_indices : torch.IntTensor
+            The top-k spans according to the given mention scores. The output spans
```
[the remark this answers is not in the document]
DeNeutoy: @kentonl what did you mean here? The indices are sorted, so it seems like the opposite of this comment is true?

allennlp/models/coreference_resolution/coref.py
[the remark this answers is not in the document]
matt-gardner: Oops, sorry, I misread the code below, thinking you were using the span width embedding and the distance embedding, not the span width embedding and an attention.  You're right.

allennlp/models/coreference_resolution/coref.py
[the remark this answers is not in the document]
kentonl: Matt's interpretation is right. There's some ambiguity in what is meant here by "sorted". This code is making sure that the returned top spans are sorted by document order, i.e. their original order.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,707 @@
...
+    def _compute_head_attention(self,
+                                head_scores: torch.FloatTensor,
+                                text_embeddings: torch.FloatTensor,
+                                span_ends: torch.IntTensor,
+                                span_size: torch.IntTensor):
+        """
+        Parameters
+        ----------
+        head_scores : torch.FloatTensor
+            Unnormalized attention scores for every word. This score is shared for every
```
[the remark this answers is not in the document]
kentonl: Just out of curiosity, how did you expect the attention weights to be parametrized?

allennlp/data/dataset_readers/coreference_resolution/conll.py
```diff
@@ -0,0 +1,293 @@
...
+        self.sentences.append(self.sentence_buffer)
+        self.sentence_buffer = []
+
+    def add_word(self, word) -> None:
+        self.sentence_buffer.append(word)
+        self.num_total_words += 1
+
+    def canonicalize_clusters(self) -> List[List[Tuple[int, int]]]:
+        """
+        The CONLL 2012 data includes 2 annotations which have identical
```
schmmd: It took me a while to understand this doc string.  I think it means "Sometimes the CONLL 2012 data includes 2 annotations with identical spans but different cluster ids.  When this happens, we want to merge the two clusters together.  This function merges clusters when they have the exact same mention."

allennlp/data/dataset_readers/coreference_resolution/conll.py
```diff
@@ -0,0 +1,293 @@
...
+                cluster_with_overlapping_mention.update(cluster)
+            else:
+                merged_clusters.append(set(cluster))
+        return [list(c) for c in merged_clusters]
+
+
+@DatasetReader.register("coref")
+class ConllCorefReader(DatasetReader):
+    """
+    Reads a single CoNLL-formatted file. This is the same file format as used in the
```
schmmd: A CoNLL 2012 Coref file?

experiment_config/coref.json
```diff
@@ -0,0 +1,99 @@
...
+        "type": "single_id",
+        "lowercase_tokens": false
+      },
+      "token_characters": {
+        "type": "characters"
+      }
+    },
+    "max_span_width": 10
+  },
+  "train_data_path": "/net/efs/aristo/allennlp/coref/ontonotes-coref/train.english.v4_gold_conll",
```
schmmd: Please use http paths to our S3 bucket rather than EFS paths.  These are not portable outside AI2.

I'm happy to help upload this data to our S3 buckets, but it's really easy.  For example, `aws s3 cp file s3://allennlp/datasets/coref/ --acl=public-read`

experiment_config/coref.json
```diff
@@ -0,0 +1,99 @@
...
+  "validation_data_path": "/net/efs/aristo/allennlp/coref/ontonotes-coref/dev.english.v4_gold_conll",
+  "test_data_path": "/net/efs/aristo/allennlp/coref/ontonotes-coref/test.english.v4_gold_conll",
+  "model": {
+    "type": "coref",
+    "text_field_embedder": {
+      "tokens": {
+        "type": "embedding",
+        "pretrained_file": "/net/efs/aristo/dlfa/glove/glove.840B.300d.txt.gz",
```
schmmd: Please upload this to `s3://allennlp/datasets/glove/`.  It's not there yet.

```
aws s3 ls s3://allennlp/datasets/glove/                                                                                     (allennlp) 
2017-10-16 15:53:58  134300425 glove.6B.100d.txt.gz
2017-10-16 15:53:49  394362216 glove.6B.300d.txt.gz
2017-10-19 16:47:48   69182520 glove.6B.50d.txt.gz
```

allennlp/models/coreference_resolution/coref.py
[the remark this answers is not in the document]
matt-gardner: I would have expected some kind of span-dependent scores for heads (yours _is_, but only through the normalization).  After thinking more about the options here, I think your modeling decision is a good one, for both statistical and computational efficiency, it just surprised me to see it written this way when I first read it, particularly because there was no text describing the decision.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,707 @@
...
+        # Select tensors relating to the antecedent spans.
+        # Shape: (batch_size, num_spans_to_keep, max_antecedents, embedding_size)
+        antecedent_embeddings = util.flattened_index_select(top_span_embeddings, antecedent_indices)
+
+        # Shape: (batch_size, num_spans_to_keep, max_antecedents)
+        antecedent_mention_scores = util.flattened_index_select(top_span_mention_scores,
+                                                                antecedent_indices).squeeze(-1)
+        # Compute antecedent scores.
+        # Shape: (batch_size, num_spans_to_keep, max_antecedents, embedding_size)
+        pairwise_embeddings = self._compute_pairwise_inputs(top_span_embeddings,
```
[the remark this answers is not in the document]
matt-gardner: Pinpointing why I don't like the name a little better: "pairwise" is an adjective, and so you're missing the actual noun that describes the embeddings this variable is holding.  Pairwise what?

allennlp/data/dataset_readers/coreference_resolution/conll.py
DeNeutoy: re-worded, thanks.

allennlp/data/dataset_readers/coreference_resolution/conll.py
DeNeutoy: Not sure what you're asking/suggesting here.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,707 @@
...
+                else:
+                    # We start a new cluster.
+                    predicted_cluster_id = len(clusters)
+                    # Append a new cluster containing only this span.
+                    clusters.append([antecedent_span])
+                    # Record the new id of this span.
+                    spans_to_cluster_ids[antecedent_span] = predicted_cluster_id
+
+                # Now add the span we are currently considering.
+                span_start, span_end = span
```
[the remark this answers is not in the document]
DeNeutoy: Mypy was getting grumpy because span is a slice of a tensor.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,707 @@
...
+                # corresponding to this span we are considering.
+                # The predicted antecedent is then an index into this list
+                # of indices, denoting the span from ``top_spans`` which is the
+                # most likely antecedent.
+                predicted_index = antecedent_indices[i, predicted_antecedent]
+
+                antecedent_span = (top_spans[predicted_index, 0],
+                                   top_spans[predicted_index, 1])
+                # Check if we've seen the span before.
+                if antecedent_span in spans_to_cluster_ids.keys():
```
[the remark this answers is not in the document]
DeNeutoy: I always forget this, i'm leaving it in

experiment_config/coref.json
DeNeutoy: Unfortunately we can't distribute this data. I've modified the paths in the same way as we have done for the SRL model though, so we can add some instructions on the website.

allennlp/models/coreference_resolution/coref.py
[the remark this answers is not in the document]
matt-gardner: Can these spans be overlapping? Would that mess with the span distance calculation? I know previous methods had rules for determinative coreference when spans were nested - maybe there's something that can be done for that in this model?

DeNeutoy:
**takes deep breath** @matt-gardner This is ready for another look.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,824 @@
...
+        num_spans_to_keep = int(math.floor(self._spans_per_word * text_embeddings.size(1)))
+
+        # Shape: (batch_size, num_spans_to_keep)
+        # These are indices (with values between 0 and num_spans) into
+        # the span_embeddings tensor.
+        top_span_indices = self._prune_and_sort_spans(mention_scores, num_spans_to_keep)
+
+        # Shape: (batch_size * num_spans_to_keep)
+        # torch.index_select only accepts 1D indices, but here
+        # we need to do select spans for each element in the batch.
```
matt-gardner: `s/to do/to/`

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,824 @@
...
+        # This reformats the indices to take into account their index into the batch.
+        flat_top_span_indices = util.flatten_and_batch_shift_indices(top_span_indices, span_starts.size(1))
```
matt-gardner: `span_starts.size(1)` here makes me have to think about what that value represents, after I'm now pretty far away from the code where it was defined.  It'd be easier to read if you did something like `num_spans = span_starts.size(1)` up above, and used `num_spans` here.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,824 @@
...
+
+        # Shape: (batch_size, num_spans_to_keep)
+        # These are indices (with values between 0 and num_spans) into
+        # the span_embeddings tensor.
+        top_span_indices = self._prune_and_sort_spans(mention_scores, num_spans_to_keep)
+
+        # Shape: (batch_size * num_spans_to_keep)
+        # torch.index_select only accepts 1D indices, but here
+        # we need to do select spans for each element in the batch.
+        # This reformats the indices to take into account their index into the batch.
```
matt-gardner: Adding something like "We precompute this here to make our multiple calls to `util.batched_index_select` below more efficient." might be helpful, so the reader doesn't wonder why you get what seems like duplicate information passed to `batched_index_select` in both `top_span_indices` and `flat_top_span_indices`.

matt-gardner: Also, it'd be nice to have some kind of note like this up above, after the `top_span_indices = ...` line: "Now that we've decided which spans are actually mentions, we reformat all of our variables to be in terms of `num_spans_to_keep` instead of `num_spans`, so we don't waste computation on spans that we've already discarded."

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,824 @@
...
+
+        # Shapes:
+        # (num_spans_to_keep, max_antecedents),
+        # (1, max_antecedents),
+        # (1, num_spans_to_keep, max_antecedents)
+        antecedent_indices, antecedent_offsets, antecedent_log_mask = \
+            self._generate_antecedents(num_spans_to_keep, max_antecedents, text_mask.is_cuda)
+        # Select tensors relating to the antecedent spans.
+        # Shape: (batch_size, num_spans_to_keep, max_antecedents, embedding_size)
+        candidate_antecedent_embedidngs = util.flattened_index_select(top_span_embeddings,
```
matt-gardner: `flattened` doesn't seem like the right name here.  Maybe call this `util.2d_index_select` instead?  You could even have the dimension be an argument to the function (it's currently hard-coded to 1).

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,707 @@
...
+        Returns
+        -------
+        The same output dictionary, but with an additional ``clusters`` key:
+
+        clusters : ``List[List[List[Tuple[int, int]]]]``
+            A nested list, representing, for each instance in the batch, the list of clusters,
+            which are in turn comprised of a list of (start, end) inclusive spans into the
+            original document.
+        """
+
```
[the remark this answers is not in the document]
matt-gardner: ☝️

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,824 @@
...
+                "coref_f1": coref_f1,
+                "mention_recall": mention_recall}
+
+    def _create_attended_span_representations(self,
+                                              head_scores: torch.FloatTensor,
+                                              text_embeddings: torch.FloatTensor,
+                                              span_ends: torch.IntTensor,
+                                              span_width: torch.IntTensor) -> torch.FloatTensor:
+        """
+        Given a tensor of unnormalized attention scores for each word in the document, compute
```
matt-gardner: I'd recommend "headedness scores" here instead of "attention scores".  It's in several places in this docstring.  It's not an "attention" because there's no query, so the name is surprising.  It's just a score.  Given this, I'd probably also recommend changing the function name to `_create_span_head_representations`.  Seems more intuitive.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,824 @@
...
+        distributions over every span with respect to these scores by normalising the attention
+        scores for words inside the span.
+
+        Given these attention distributions over every spans, weight the corresponding vector
```
matt-gardner: `s/every spans/every span/`

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,824 @@
...
+        return {"coref_precision": coref_precision,
+                "coref_recall": coref_recall,
+                "coref_f1": coref_f1,
+                "mention_recall": mention_recall}
+
+    def _create_attended_span_representations(self,
+                                              head_scores: torch.FloatTensor,
+                                              text_embeddings: torch.FloatTensor,
+                                              span_ends: torch.IntTensor,
+                                              span_width: torch.IntTensor) -> torch.FloatTensor:
```
matt-gardner: `s/span_width/span_widths/` here, too.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,824 @@
...
+
+    def _compute_span_representations(self,
+                                      text_embeddings: torch.FloatTensor,
+                                      text_mask: torch.FloatTensor,
+                                      span_starts: torch.IntTensor,
+                                      span_ends: torch.IntTensor) -> torch.FloatTensor:
+        """
+        Computes an embedded representation of every candidate span. This is a concatenation
+        of the contextualized endpoints of the span, an embedded representation of the width of
+        the span and the result of applying attention to the words in the span.
```
matt-gardner: `s/the result of applying attention to the words in the span/a representation of the span's predicted head/`, or something like that.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,824 @@
...
+
+    @staticmethod
+    def _generate_antecedents(num_spans_to_keep: int,
+                              max_antecedents: int,
+                              is_cuda: bool) -> Tuple[torch.IntTensor,
+                                                      torch.IntTensor,
+                                                      torch.FloatTensor]:
+        """
+        This method generates possible antecedents per span which survived the pruning
+        stage. This procedure is _generic across the batch_. The reason this is the case is
```
matt-gardner: rst uses back ticks for emphasis, not underscores.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,824 @@
...
+        -------
+        span_pair_embeddings : ``torch.FloatTensor``
+            Embedding representation of the pair of spans to consider. Has shape
+            (batch_size, num_spans_to_keep, max_antecedents, embedding_size)
+        """
+        # Shape: (batch_size, num_spans_to_keep, max_antecedents, embedding_size)
+        target_embeddings = top_span_embeddings.unsqueeze(2).expand_as(antecedent_embeddings)
+
+        # Shape: (1, max_antecedents, embedding_size)
+        antecedent_distance_embeddings = self._distance_embedding(util.bucket_values(antecedent_offsets))
```
matt-gardner: You need to pass `self.num_distance_buckets` here.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,824 @@
...
+        # Shape: (batch_size, num_spans_to_keep, max_antecedents + 1)
+        # Adding the log of the mask means that we add -inf to all
+        # antecedents not in the coreference cluster for a given span,
+        # meaning when we compute the marginal log likelihood, these are
+        # correctly masked to have no contribution.
+        masked_coreference_scores = coreference_scores + augmented_labels.log()
+
+        # Shape: (batch_size, num_spans_to_keep)
+        marginalized_gold_scores = util.logsumexp(masked_coreference_scores, 2)
+        log_norm = util.logsumexp(coreference_scores, 2)
```
matt-gardner: Still pretty sure you can just do:

```python
coreference_log_probs = util.last_dim_softmax(coreference_scores, top_span_mask)
# Adding `augmented_labels.log()` masks out all coreference decisions that are inconsistent with the
# gold coreference clusters, letting us sum over all possible antecedents that are in the correct cluster.
correct_antecedent_log_probs = coreference_log_probs + augmented_labels.log()
loss = -util.logsumexp(correct_antecedent_log_probs)
```
And this is now simple enough to go in `forward` directly, even.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,824 @@
...
+                                                    top_span_indices,
+                                                    flat_top_span_indices)
+        top_span_ends = util.batched_index_select(span_ends,
+                                                  top_span_indices,
+                                                  flat_top_span_indices)
+
+        # Compute indices for antecedent spans to consider.
+        max_antecedents = min(self._max_antecedents, num_spans_to_keep)
+
+        # This next section can be confusing - the returned tensors do _not_ have a batch size
```
matt-gardner: One more suggestion.  Here's an attempt to make this comment a little more helpful:

-----------

Now that we have our variables in terms of `num_spans_to_keep`, we need to compare span pairs to decide each span's antecedent.  Each span can only have _prior_ spans as antecedents, and we only consider up to `max_antecedents` prior spans.  So the first thing we do is construct a matrix mapping a span's index to the indices of its allowed antecedents.  Note that this is independent of the batch dimension - it's just a function of the span's position in `top_spans`.  The spans are in document order, so we can just use the relative index of the spans to know which other spans are allowed antecedents.

Once we have this matrix, we reformat our variables _again_ to get embeddings for all valid antecedents for each span.  This gives us variables with shapes like `(batch_size, num_spans_to_keep, max_antecedents, embedding_size)`, which we can use to make coreference decisions between valid span pairs.

-----------------

Along these lines, changing `antecedent_indices` to `allowed_antecedents`, `valid_antecedents`, `allowed_antecedent_indices`, or something like that, might be helpful.

allennlp/models/coreference_resolution/coref.py
matt-gardner: Oh, the softmax takes it out of logspace.  But then you just multiple by the mask, instead of adding the log mask, and you do a sum instead of `logsumexp`.  That's probably even faster still, actually.

allennlp/nn/util.py
```diff
@@ -318,10 +320,28 @@ def last_dim_softmax(tensor: torch.Tensor, mask: Optional[torch.Tensor] = None)
...
+    Takes a tensor with 3 or more dimensions and does a masked softmax over the last dimension.  We
+    assume the tensor has shape ``(batch_size, ..., sequence_length)`` and that the mask (if given)
+    has shape ``(batch_size, sequence_length)``.
+    """
+    return _last_dimension_applicator(masked_softmax, tensor, mask)
+
+
+def last_dim_log_softmax(tensor: torch.Tensor, mask: Optional[torch.Tensor] = None) -> torch.Tensor:
+    """
+    Takes a tensor with 3 or more dimensions and does a masked softmax over the last dimension.  We
```
matt-gardner: Note here that this is in log space.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,794 @@
...
+
+        # Shapes:
+        # (num_spans_to_keep, max_antecedents),
+        # (1, max_antecedents),
+        # (1, num_spans_to_keep, max_antecedents)
+        valid_antecedent_indices, valid_antecedent_offsets, valid_antecedent_log_mask = \
+            self._generate_valid_antecedents(num_spans_to_keep, max_antecedents, text_mask.is_cuda)
+        # Select tensors relating to the antecedent spans.
+        # Shape: (batch_size, num_spans_to_keep, max_antecedents, embedding_size)
+        candidate_antecedent_embedidngs = util.flattened_index_select(top_span_embeddings,
```
matt-gardner: `s/embedidngs/embeddings/`

allennlp/models/coreference_resolution/coref.py
```diff
@@ -0,0 +1,794 @@
...
+        # Shape: (batch_size, num_spans, 1)
+        mention_scores = self._mention_scorer(self._mention_feedforward(span_embeddings))
+        mention_scores += span_mask.log()
+
+        # Prune based on mention scores.
+        num_spans_to_keep = int(math.floor(self._spans_per_word * document_length))
+
+        # Shape: (batch_size, num_spans_to_keep)
+        # These are indices (with values between 0 and num_spans) into
+        # the span_embeddings tensor. Now that we've decided which spans
```
matt-gardner: Nit: I'd put this `Now ...` bit _after_ the `top_span_indices` line, as its own block comment.

PR #471, opened 2017-11-07
Compute context insensitive ELMo representations

matt-peters:
Add:

* `ELMoTokenCharactersIndexer `
* `ELMoTokenEmbedder`


allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -0,0 +1,219 @@
...
+        n_filters = sum(f[1] for f in filters)
+        n_highway = cnn_options['n_highway']
+
+        # create the layers, and load the weights
+        self._highways = Highway(n_filters, n_highway, activation=torch.nn.functional.relu)
+        for k in range(n_highway):
+            # The AllenNLP highway is one matrix multplication with concatenation of
+            # transform and carry weights.
+            with h5py.File(cached_path(self._weight_file), 'r') as fin:
+                W_transform = fin['CNN_high_{}'.format(k)]['W_transform'][...].T
```
DeNeutoy: I'd strongly prefer using `numpy.transpose()` rather than `.T`, it's so easy to miss. Also if we could use lowercase full name variables that would be great, `W_transform` -> `weight_transform`.

allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -0,0 +1,219 @@
...
+from allennlp.modules.token_embedders.token_embedder import TokenEmbedder
+from allennlp.modules.highway import Highway
+from allennlp.nn.util import add_bos_eos
+from allennlp.data.token_indexers.elmo_indexer import ELMoCharacterMapper
+from allennlp.data import Vocabulary
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+DTYPE = 'float32'
+
+# pylint: disable=invalid-name,protected-access,attribute-defined-outside-init,bad-continuation
```
DeNeutoy: Can you fix these, or disable them where they are used? The invalid name, protected access and bad-continuation ones should be fix-able.

allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -0,0 +1,219 @@
...
+        with h5py.File(cached_path(self._weight_file), 'r') as fin:
+            char_embed_weights = fin['char_embed'][...]
+
+        weights = numpy.zeros(
+            (char_embed_weights.shape[0] + 1, char_embed_weights.shape[1]),
+            dtype=DTYPE
+        )
+        weights[1:, :] = char_embed_weights
+
+        self._char_embedding_weights = torch.nn.Parameter(
```
DeNeutoy: This is where the `bad-continuation` is coming from - pylint wants this formatted as : 
```
self._char_embedding_weights = torch.nn.Parameter(torch.FloatTensor(weights),
                                                  requires_grad=False)
```

allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -0,0 +1,219 @@
...
+            torch.FloatTensor(weights), requires_grad=False
+        )
+
+    def _load_cnn_weights(self):
+        cnn_options = self._options['char_cnn']
+        filters = cnn_options['filters']
```
DeNeutoy: `filter_sizes`?

allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -0,0 +1,219 @@
...
+                self._highways._layers[k].bias.requires_grad = False
+
+    def _load_projection(self):
+        cnn_options = self._options['char_cnn']
+        filters = cnn_options['filters']
+        n_filters = sum(f[1] for f in filters)
+
+        self._projection = torch.nn.Linear(n_filters, self.output_dim, bias=True)
+        with h5py.File(cached_path(self._weight_file), 'r') as fin:
+            W = fin['CNN_proj']['W_proj'][...]
```
DeNeutoy: `weight` and `bias`

requirements.txt
```diff
@@ -53,3 +53,7 @@ editdistance
 
 # Tutorial notebooks
 jupyter
+
+# For pretrained model weights
```
DeNeutoy: Can you add this requirement to `setup.py` also?

tests/data/token_indexers/elmo_indexer_test.py
```diff
@@ -0,0 +1,77 @@
+# pylint: disable=no-self-use,invalid-name,bad-continuation,bad-whitespace
```
DeNeutoy: Can you fix the bad-continuation and bad-whitespace, rather than overriding them here?

tests/modules/token_embedders/elmo_token_embedder_test.py
```diff
@@ -0,0 +1,51 @@
...
+import torch
+from torch.autograd import Variable
+
+
+from allennlp.common.testing import AllenNlpTestCase
+from allennlp.modules.token_embedders.elmo_token_embedder import ELMoTokenEmbedder
+from allennlp.data.token_indexers.elmo_indexer import ELMoTokenCharactersIndexer
+from allennlp.data import Token, Vocabulary
+
+FIXTURES = os.path.join('tests', 'fixtures', 'elmo')
```
DeNeutoy: I guess this is ok, but you could also override the `setUp()` of the test case and add it as an attribute there instead.

allennlp/modules/token_embedders/elmo_token_embedder.py
DeNeutoy: Also if you could add a top level comment about why we have to transpose the weights, that would be great.

allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -0,0 +1,219 @@
...
+        self.output_dim = self._options['lstm']['projection_dim']
+
+        self._load_weights()
+
+    @overrides
+    def get_output_dim(self) -> int:
+        return self.output_dim
+
+    @overrides
+    def forward(self, inputs):  # pylint: disable=arguments-differ
```
DeNeutoy: Can you add a docstring with expected arguments and return types to `forward`.

allennlp/data/token_indexers/elmo_indexer.py
```diff
@@ -0,0 +1,113 @@
...
+
+from allennlp.common.checks import ConfigurationError
+from allennlp.common.params import Params
+from allennlp.common.util import pad_sequence_to_length
+from allennlp.data.tokenizers.token import Token
+from allennlp.data.token_indexers.token_indexer import TokenIndexer
+from allennlp.data.vocabulary import Vocabulary
+
+
+def _make_bos_eos(char, pad_char, bow_char, eow_char, max_word_length):
```
DeNeutoy: Given you've created a class to do this processing below, seems like you could add this in there.

allennlp/nn/util.py
```diff
@@ -567,3 +567,42 @@ def logsumexp(tensor: torch.Tensor,
     else:
         stable_vec = tensor - max_score.unsqueeze(dim)
     return max_score + (stable_vec.exp().sum(dim, keepdim=keepdim)).log()
+
+
+def add_bos_eos(tensor, mask, bos, eos):
```
DeNeutoy: Add types here and in the docstring.

allennlp/nn/util.py
```diff
@@ -567,3 +567,42 @@ def logsumexp(tensor: torch.Tensor,
...
+    Add begin/end of sentence tokens to the batch of sentences.
+    Given a batch of sentences with size ``(batch_size, n_times)`` or
+    ``(batch_size, n_times, dim)`` this returns a tensor of shape ``(batch_size, n_times + 2)`` or
+    ``(batch_size, n_times + 2, dim)`` respectively.
+
+    Returns both the new tensor and updated mask
+
+    Parameters
+    ----------
+    tensor : A tensor of shape ``(batch_size, n_times)`` or ``(batch_size, n_times, dim)``
```
DeNeutoy: `timesteps` rather than `n_times`, here and above.

allennlp/nn/util.py
```diff
@@ -567,3 +567,42 @@ def logsumexp(tensor: torch.Tensor,
...
+    """
+    sequence_lengths = mask.sum(dim=1).data.numpy()
+    tensor_shape = list(tensor.data.shape)
+    new_shape = list(tensor_shape)
+    new_shape[1] = tensor_shape[1] + 2
+    ret = Variable(tensor.data.new(*new_shape).fill_(0))
+    if len(tensor_shape) == 2:
+        ret[:, 1:-1] = tensor
+        ret[:, 0] = bos
+        for i, j in enumerate(sequence_lengths):
```
DeNeutoy: I think you can avoid looping over the tensor by using this: http://pytorch.org/docs/master/tensors.html#torch.Tensor.index_fill_

allennlp/nn/util.py
DeNeutoy: Maybe mention ELMo in the docstring here too.

matt-peters:
RE: review summary -- 

(1) do you have a template unit test to exercise a model as you suggested?
(2)  Its a little scary that low level details in `_remove_pretrained_embedding_params ` depend on ELMo -- is there any reason not to reload the weights every time the ELMo classes are initialized, even if they are serialized with the model?  This avoids having to special case it there.

DeNeutoy:
Hmm, yeah the thing this avoids is storing copies of the weights. As these aren't trainable, it's less important, so maybe just add a TODO for that suggestion for now and we can add it later as an optimisation.

allennlp/data/token_indexers/elmo_indexer.py
```diff
@@ -0,0 +1,113 @@
...
+
+class ELMoCharacterMapper(object):
+    '''
+    Maps individual tokens to sequences of character ids, compatible with ELMo.
+    '''
+    max_word_length = 50
+
+    # char ids 0-255 come from utf-8 encoding bytes
+    # assign 256-300 to special chars
+    bos_char = 256  # <begin sentence>
```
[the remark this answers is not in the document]
matt-peters: This is to be consistent with the other places I have similar code... same for the variable names below. Makes it easier to keep things in sync.

allennlp/data/token_indexers/elmo_indexer.py
```diff
@@ -0,0 +1,113 @@
...
+    ret = [pad_char] * max_word_length
+    ret[0] = bow_char
+    ret[1] = char
+    ret[2] = eow_char
+    return ret
+
+class ELMoCharacterMapper(object):
+    '''
+    Maps individual tokens to sequences of character ids, compatible with ELMo.
+    '''
```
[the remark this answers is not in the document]
matt-peters: Really?  `'''` looks so much nicer then `"""`.

allennlp/data/token_indexers/elmo_indexer.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def _make_bos_eos(char, pad_char, bow_char, eow_char, max_word_length):
+    ret = [pad_char] * max_word_length
+    ret[0] = bow_char
+    ret[1] = char
+    ret[2] = eow_char
+    return ret
+
+class ELMoCharacterMapper(object):
```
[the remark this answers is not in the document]
matt-peters: Is there any harm in doing so?

allennlp/data/token_indexers/elmo_indexer.py
matt-peters: I'm calling it to create class attributes, couldn't figure an easy way to add it as a class method and also call it there.

.gitignore
```diff
@@ -2,3 +2,15 @@
...
 doc/_build/
+
+# build artifacts
+allennlp.egg-info/
+.eggs/
+build/
+dist/
+.mypy_cache
+
+# vim temporary files
```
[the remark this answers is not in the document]
matt-peters: done

allennlp/data/token_indexers/elmo_indexer.py
```diff
@@ -0,0 +1,113 @@
...
+    bow_char = 258  # <begin word>
+    eow_char = 259  # <end word>
+    pad_char = 260 # <padding>
+
+    bos_chars = _make_bos_eos(bos_char, pad_char, bow_char, eow_char, max_word_length)
+    eos_chars = _make_bos_eos(eos_char, pad_char, bow_char, eow_char, max_word_length)
+
+    @staticmethod
+    def convert_word_to_char_ids(word):
+        if word == '<S>':
```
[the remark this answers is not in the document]
matt-peters: done

matt-gardner:
@DeNeutoy, @matt-peters, it's actually not just an optimization, it's "what files do I need to load my model"?  If you have a dependency on the ELMo model file, I can't just load my model that uses ELMo embeddings from an archive anymore.  That's the whole reason that "remove pretrained embedding params" method is there in the first place.  I agree that it's not ideal, but we need some way to make it so that you can load the model _just_ from the archive that we save, without requiring separate embedding files (whether they are glove or elmo or something else).  If you have a better solution for how to do this, we'd be happy to hear it.

allennlp/data/token_indexers/elmo_indexer.py
```diff
@@ -0,0 +1,113 @@
...
+            ret = ELMoCharacterMapper.bos_chars
+        elif word == '</S>':
+            ret = ELMoCharacterMapper.eos_chars
+        else:
+            word_encoded = word.encode('utf-8', 'ignore')[:(ELMoCharacterMapper.max_word_length-2)]
+            ret = [ELMoCharacterMapper.pad_char] * ELMoCharacterMapper.max_word_length
+            ret[0] = ELMoCharacterMapper.bow_char
+            for k, chr_id in enumerate(word_encoded, start=1):
+                ret[k] = chr_id
+            # pylint: disable=undefined-loop-variable
```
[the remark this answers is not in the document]
matt-peters: done

allennlp/data/token_indexers/elmo_indexer.py
```diff
@@ -0,0 +1,113 @@
...
+
+    @staticmethod
+    def convert_word_to_char_ids(word):
+        if word == '<S>':
+            ret = ELMoCharacterMapper.bos_chars
+        elif word == '</S>':
+            ret = ELMoCharacterMapper.eos_chars
+        else:
+            word_encoded = word.encode('utf-8', 'ignore')[:(ELMoCharacterMapper.max_word_length-2)]
+            ret = [ELMoCharacterMapper.pad_char] * ELMoCharacterMapper.max_word_length
```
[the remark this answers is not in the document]
matt-peters: I'm looping over the individual bytes of a string so `pad_sequence_to_length` won't work (it raises an exception)

allennlp/modules/token_embedders/elmo_token_embedder.py
matt-peters: done

allennlp/modules/token_embedders/elmo_token_embedder.py
matt-peters: ok

allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -0,0 +1,219 @@
...
+        mask = ((inputs > 0).sum(dim=-1) > 0).long()
+        character_ids_with_bos_eos, mask_with_bos_eos = add_bos_eos(
+            inputs,
+            mask,
+            Variable(torch.from_numpy(numpy.array(ELMoCharacterMapper.bos_chars))),
+            Variable(torch.from_numpy(numpy.array(ELMoCharacterMapper.eos_chars)))
+        )
+
+        # the character id embedding
+        max_chars_per_token = self._options['char_cnn']['max_characters_per_token']
```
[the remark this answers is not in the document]
matt-peters: This is a low level detail in ELMo, I don't want to expose it to the user.

allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -0,0 +1,219 @@
...
+        # (batch_size * sequence_length, max_chars_per_token, embed_dim)
+        char_embeds = torch.nn.functional.embedding(
```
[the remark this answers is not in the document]
matt-peters: ok

allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -0,0 +1,219 @@
...
+        convolutions = []
+        for i, (width, num) in enumerate(filters):
+            conv = torch.nn.Conv1d(
+                in_channels=char_embed_dim,
+                out_channels=num,
+                kernel_size=width,
+                bias=True
+            )
+            # load the weights
+            with h5py.File(cached_path(self._weight_file), 'r') as fin:
```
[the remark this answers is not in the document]
matt-peters: The overhead to open and close the file relative to reading the weights is tiny. This method guarantees that the file will be closed properly, etc.

allennlp/modules/token_embedders/elmo_token_embedder.py
matt-peters: I wanted to be consistent with the other place I have similar code for naming.  It will make maintaining this much easier.

allennlp/nn/util.py
matt-peters: ok

allennlp/nn/util.py
matt-peters: done

requirements.txt
matt-peters: ok

tests/data/token_indexers/elmo_indexer_test.py
matt-peters: Probably not without wanting to tear my eyes out or using a code formatter.  It's coming from the cut and pasted arrays of numbers in the tests.  Is there one for pylint?

tests/modules/token_embedders/elmo_token_embedder_test.py
matt-peters: This is a lighter weight way to allow other tests to use the same variable without having to define a new subclass.

matt-peters:
RE: So `_remove_pretrained_embedding_params ` which key should I remove from the `Params` then?

tests/data/token_indexers/elmo_indexer_test.py
DeNeutoy: It's not that hard - you can use `bash scripts/pylint.sh` to run the pylint tests which get run on Travis.

allennlp/data/token_indexers/elmo_indexer.py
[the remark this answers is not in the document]
DeNeutoy: It does nothing... Why keep it? It fits the style of the library and reduces clutter.

allennlp/data/token_indexers/elmo_indexer.py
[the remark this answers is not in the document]
DeNeutoy: I can see this is a medium validity argument - I will defer to the Director of Syntax, @matt-gardner.

allennlp/data/token_indexers/elmo_indexer.py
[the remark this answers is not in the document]
DeNeutoy: https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

DeNeutoy:
Also, the build is failing because you haven't added the new modules to the docs - https://travis-ci.org/allenai/allennlp/jobs/298745025#L1052

You can do this by adding the corresponding .rst files in allennlp/doc/api. This will automatically create docs from the docstrings of the modules (which is why they have to be formatted in such a clinical way).

tests/data/token_indexers/elmo_indexer_test.py
matt-peters: Is there a mode in pylint that will automatically reformat whitespace? Or another program that will do so according to our pylint configuration since we aren't following e.g. pep8 or another established style guide?  Otherwise I'd rather not spend 15 minutes making whitespace changes... hence the comment about tearing my eyes out.

matt-peters: E.g. `yapf -i --style=pep8 tests/data/token_indexers/elmo_indexer_test.py` formats according to pep8 but fails pylint horribly.

allennlp/data/token_indexers/elmo_indexer.py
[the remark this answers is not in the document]
matt-peters: OK, I'll remove it.

matt-peters:
Thinking through the ELMo serialization piece and saving the weights as part of the model archive -- I can see several advantages to not including the ELMo weights as part of the archive and instead allowing the model archives to depend on multiple files.

1.  The weights file is big.  So if we e.g. save weights at every epoch then it's really inefficient to keep writing them to disk.  Really we should exclude any non-trainable weights from the archive and assume they are available via other files on disk for this reason (e.g. re-saving word embeddings every epoch runs into the same problem).
2.  The ELMo code now also depends on an options file with hyperparameters.  This is so that we can easily change the model to one with different hyperparameters, but the user doesn't need to know the details of how it works.  In any case, the model archive will need this file.
3.  We already have the functionality to download and cache files from S3 automatically.  So if the file paths point to S3 locations for these additional files, the ELMo code will still work properly if a user tries to run a model without the files in the archive.

matt-gardner:
@matt-peters, on your points:

1. Realistically, this is not going to happen.  It's too complicated to have the code try to decide which things should be saved and which loaded, where pytorch just gives us `state_dict` and `load_state_dict`.  The space savings that you would get from doing this aren't worth the engineering time to make a clean solution here, given that we have lots of other things to do.

2-3. You're right that what you're describing will just work with current code.  It's fine if you want to do it this way.  But I think it might be worth thinking a little more carefully about how to do this more generally.  The cleanest way I can think of to do something like this is to define a `ModelTextFieldEmbedder` that just passes off the embedding to a loaded model.  Then you'd have an `ELMoModel` class that is loaded from an archive, just like any other model, and you just have a single parameter in the main configuration - the path to the model archive.  How the `ELMoModel` works internally is up to you, and you can do whatever you want, as long as it results in a model that can be loaded from a single archive file.  But this way, you get a really simple specification for downstream models that want to use ELMo (or any other model) to replace their text embedder.

allennlp/data/token_indexers/elmo_indexer.py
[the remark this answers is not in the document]
matt-gardner: Are you going to be maintaining this code, @matt-peters, or are the research engineers on AllenNLP going to be maintaining it?  If it's not just you, I think there's more value in making sure this code is immediately readable by anyone on the team.  I'd go even farther than @DeNeutoy here, saying that `bos`, `eos`, `bow`, and `eow` are not readable at all and should be expanded.  When I see `bow` in code, for instance, I think `bag_of_words`, not `beginning_of_word`.  Much better to have code immediately readable than to save characters, or to maintain consistency with code that is outside of our repo.

If you really feel strongly about this, though, do what you want, it's your code.  It'll just be harder for the team to contribute to it if it's hard to read.

allennlp/nn/util.py
matt-peters: I tried to get this to work with `index_fill_` but coundn't.  In any case, this loop isn't a bottle neck (I benchmarked it).

allennlp/data/token_indexers/elmo_indexer.py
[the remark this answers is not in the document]
matt-peters: I was assuming that I'd be the defacto maintainer, or at least the one most likely to modify this in the near term. But I do see you point about trying to make it generally readable.  I've modified it to use more verbose names.

matt-peters:
@DeNeutoy and I had a discussion about the serialization issue and I don't think there is an easy solution, at least not in the scope of this PR.  I'd like to push that off to another PR to avoid holding up progress on integrating ELMo as this is only a piece of what I need to port.  We can revisit it once things are closer to a working state and this is actually usable.

matt-peters:
Assuming the tests pass this is ready for another look.

allennlp/modules/token_embedders/elmo_token_embedder.py
DeNeutoy: ☝️

tests/data/token_indexers/elmo_indexer_test.py
DeNeutoy: Please format these arrays properly and remove the global pylint statements at the top. I promise it will take you less than 10 minutes. You need to format arrays to look like this:
```
array = [[0, 0, 0, 1],
         [1, 2, 3, 4]]

```
I.e 1 space indentation for each dimension, one space between each number.

tests/nn/util_test.py
```diff
@@ -1,4 +1,4 @@
-# pylint: disable=invalid-name, no-self-use,too-many-public-methods
+# pylint: disable=invalid-name,no-self-use,too-many-public-methods,bad-continuation,bad-whitespace
```
DeNeutoy: same here, with the global disable statements. Please remove - if you add these, it means that pylint won't complain when other people add tests to this file, and they might format things considerably worse than you. Once we drop standards once, even for small stuff like this, it becomes 1) difficult to argue that people should maintain the standards we set and 2) easier to write poor code. All you need to do is delete some spaces and align the arrays correctly.

allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -0,0 +1,245 @@
...
+        self.output_dim = self._options['lstm']['projection_dim']
+
+        self._load_weights()
+
+    @overrides
+    def get_output_dim(self) -> int:
+        return self.output_dim
+
+    @overrides
+    def forward(self, inputs: torch.Tensor) -> Dict[str, torch.Tensor]:  # pylint: disable=arguments-differ
```
DeNeutoy: Do you think it would be useful to have a `remove_beginning_and_end_tokens` flag which was by default false? Do you think people will want to use the CNN only part of the elmo embeddings independently? I guess if that is the case, we could have a `Seq2VecEncoder` which did that, so maybe not.

allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -0,0 +1,245 @@
+'''
```
DeNeutoy: `"""` for docstrings.

allennlp/modules/token_embedders/elmo_token_embedder.py
matt-peters: Maybe, but we shouldn't add it until needed.

allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -0,0 +1,245 @@
...
+        Dict with keys:
+
+        ``'token_embedding'``: ``torch.autograd.Variable``
+            Shape ``(batch_size, sequence_length + 2, embedding_dim)`` tensor with context
+            insensitive token representations.
+        ``'mask'``:  ``torch.autograd.Variable``
+            Shape ``(batch_size, sequence_length + 2)`` long tensor with sequence mask.
+        """
+        # Add BOS/EOS
+        mask = ((inputs > 0).sum(dim=-1) > 0).long()
```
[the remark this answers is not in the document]
matt-peters: good catch, thanks

allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -0,0 +1,245 @@
...
+
+        # reshape to (batch_size, sequence_length, embedding_dim)
+        batch_size, sequence_length, _ = character_ids_with_bos_eos.size()
+
+        return {
+                'mask': mask_with_bos_eos,
+                'token_embedding': token_embedding.view(batch_size, sequence_length, -1)
+        }
+
+    def _load_weights(self):
```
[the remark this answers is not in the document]
matt-peters: I disagree that it improves readability and will make it harder to refactor later.  I was planning on adopting the same `_load_weights` function in the other ELMo components that use the same options/weights so if we wanted to e.g. add a hook that pre-processed the weights or something it would be easy to do.

PR #477, opened 2017-11-07
fix things spacy 2.0 broke

DeNeutoy:
Spacy 2.0 is backwards incompatible, because the models are different and you can't load a language by it's id anymore, you have to use the full model name. 

I've fixed up everything which it broke - thankfully, none of our models specified the language of the model they loaded, because otherwise this would have made our models backwards incompatible. This is a good reason to use the default arguments unless we are actually changing the values in the json. 

FYI when you upgrade locally, you need to actually remove the `en` data directory in your local site packages(perhaps just uninstalling and then re-installing would do this, but I just upgraded directly).


DeNeutoy:
Right, I originally was just going to pin it but then I decided to be a hero and fix the things it broke instead. Pinning everything seems a bit militant? Equally we don't want to find out about new releases by having them break our stuff. Your call.

joelgrus:
I agree with your fixes, but I think we should also pin it to 2.0. 

IMO our code should *never* break on account of someone else releasing a new version of a library

matt-peters:
This PR seems a little dangerous to me.  How do we know that upgrading spacy to 2.0 didn't change the tokenizer, or POS tagger, or whatever in some subtle way that decreases performance of existing trained models?  IMO, we should do as Joel suggested and pin all of our dependencies to specific versions.  Otherwise we will run into an issue in the future where the build breaks due to an upstream dependency version change.

joelgrus:
thinking about it more, I agree with Matt, upgrading a major version of spacy is a big change and we shouldn't do it casually

matt-gardner:
Pinning versions is good, though the problem is that it can lead to conflicts in projects - say you want to use allennlp and some other library in your project, and that other library has spacy pinned to a different version.  You're now out of luck, where you wouldn't be if we weren't pinned.  Can we give bounds instead of pinning?

As for this particular upgrade, @joelgrus, what do you think of resurrecting your smoke test PR?  Is there a way to make that run fast enough to be practical?

schmmd:
Why not just evaluate our 3 existing models and make sure we get the scores we expect?  This is fast and easy to do.  http://allennlp.org/models

DeNeutoy:
I will do that tomorrow morning.

joelgrus:
so I just updated the sniff tests (they were way out of date).

if the datasets are not cached locally, they take about 35 seconds to run
if the datasets are cached locally, they take about 15 seconds

(that's over my home internet connection)

matt-gardner:
@joelgrus, that's a pretty negligible addition.  Would those tests be sufficient to alleviate concerns for this PR?

joelgrus:
they alleviate some concerns, but they're only running `predict`, they're not training a model, and they don't cover any use cases that's not the three demo models.

Probably we should have some really comprehensive suite of regression tests that involve training models and so on, but I don't know that that will happen anytime soon.

(separately, should I merge the sniff test PR?)

DeNeutoy:
Ok, I checked and this doesn't regress any of our models on the full test sets. I think we should merge this, because at the moment this is breaking any new installs because we didn't pin the spacy version originally.

requirements.txt
```diff
@@ -20,7 +20,7 @@ overrides
 nltk
 
 # Mainly used for the faster tokenizer.
-spacy
+spacy==2.0
```
matt-gardner: Any opposition to something like `spacy>=2.0,<2.1`?  Then we'll get bugfixes, but won't upgrade to a new minor version automatically, and have some flexibility with other libraries that pin their versions.  As spacy gets past 2.1, too, we can see if it breaks anything, and expand the bounds accordingly.

joelgrus:
ok, I'll approve, but I think we should make a task for a comprehensive suite of regression tests that we can run for changes like this

joelgrus:
ok, I'll approve, but I think we should make a task for a comprehensive suite of regression tests that we can run for changes like this

Issue #239, opened 2017-08-31
Allow for different handling of OOV words


matt-gardner:
We currently give all OOV tokens the same embedding at both training and test time.  It'd be nice to be able to have some different options here:

- At test time, see if the OOV token is in glove, and use that embedding instead
- At both training and test, use random vectors for each unique OOV token, as suggested [here](https://www.semanticscholar.org/paper/A-Comparative-Study-of-Word-Embeddings-for-Reading-Dhingra-Liu/3ec37205c9201fc891ab51da200e361fdc34bfb3)

There are probably some other options I'm forgetting right now.  These would be pretty tricky to implement in our current data pipeline, though.

matt-gardner:
@OyvindTafjord made some nice progress on getting some of this to work:

https://github.com/allenai/allennlp/compare/master...OyvindTafjord:expandVocab

Here are Oyvind's notes on this diff:

  * If you use --restore in the train command, it reads the vocabulary from the stored model rather than from the data files (essential if you're continuing training with a different dataset than the original model).

  * If you use --expand_vocabulary (together with --restore) in the train command, it will expand the embedding matrix with any new words found in the current dataset. This is done using a very hacky hook to the trainer, and assuming the BiDAF structure of the text field embedder, so this will be broken in general. It saves the new vocabulary in a vocabulary_new directory, so to get a proper model.tar.gz at the end, you have to manually move it to vocabulary and rerun the archive generation.mic

  * You can also use --expand_vocabulary when doing an evaluation to the same effect.

  * I did not do anything special about the training_state recovery, it might make sense to give some more control of that (e.g.,. not require a training state, or allow to start from original state rather than the current state).

  * I also added --output_file to the evaluate command, but this ended up being sufficiently ugly that there should probably be a more general way to get custom output for a run, rather than having to monkey around inside the forward method.

matt-gardner:
@DeNeutoy, @joelgrus, I think it probably makes sense to have a `load_from` argument to trainer that's separate from the `serialization_dir`, instead of how we're currently looking at the serialization dir and continuing if it's not empty.  Or maybe we should be able to just specify a model archive location for the `model` parameter in the config file?  We don't have a great story around fine-tuning models.

If you're loading a saved model, you should _always_ expand the vocabulary with the new dataset in this way (or in a way that's configurable).

Ok, so I think there are three separate issues here:

1. We should have a better way to do model fine-tuning.
[...]

PR #478, opened 2017-11-08
Specify a pretrained embedding file when creating a Vocabulary

nafitzgerald:
This adds the ability to specify pretrained embedding files (on a per-namespace basis) when creating a Vocabulary. This can be used to potentially do two different things:

- Restrict the Vocabulary to only words present in the pretrained embedding file
- Ensure that words are added if they are present in the pretrained embedding file, regardless of their count.

This should help with issue #239.


allennlp/data/vocabulary.py
```diff
@@ -134,12 +144,26 @@ class Vocabulary:
         The default is ``("*tags", "*labels")``, so as long as your namespace ends in "tags" or
         "labels" (which is true by default for all tag and label fields in this code), you don't
         have to specify anything here.
+    pretrained_files : ``Dict[str, str]``, optional
+        If provides, this map specifies the path to optional pretrained embedding files for each
```
matt-gardner: `s/provides/provided/`

allennlp/data/vocabulary.py
```diff
@@ -134,12 +144,26 @@ class Vocabulary:
...
+        namespace. This can be used to either restrict the vocabulary to only words which appear
+        in this file, or to ensure that any words in this file are included in the vocabulary
+        regardless of their count, depending on the value of ``pretrained_exclusive``.
```
matt-gardner: It'd be nice if this were more clear about what happens to words that you _didn't_ see in your data, but are in the embedding file.  You currently (and I think correctly) leave those out of the vocab, but that's not clear from this note.

allennlp/data/vocabulary.py
```diff
@@ -134,12 +144,26 @@ class Vocabulary:
...
+    pretrained_exclusive : bool, optional (default = False)
```
matt-gardner: I think `pretrained_exclusive` is a bit ambiguous - are you excluding pretrained words, or exclusively using pretrained words?  E.g., we use "end-exclusive" in span prediction models to say we're _not_ including the end index in the `(start, end)` pair, where here you want "pretrained-exclusive" to mean you are _only_ including the pretrained words.

I think something like `only_include_pretrained` (or even `only_include_pretrained_words`) would be a more obvious name here.

allennlp/data/vocabulary.py
```diff
@@ -325,11 +365,15 @@ def from_params(cls, params: Params, dataset=None):
         min_count = params.pop("min_count", 1)
         max_vocab_size = params.pop("max_vocab_size", None)
         non_padded_namespaces = params.pop("non_padded_namespaces", DEFAULT_NON_PADDED_NAMESPACES)
+        pretrained_files = params.pop("pretrained_files", {})
+        pretrained_exclusive = params.pop("pretrained_exlusive", False)
         params.assert_empty("Vocabulary - from dataset")
         return Vocabulary.from_dataset(dataset,
                                        min_count,
                                        max_vocab_size,
```
matt-gardner: Can you add keyword args to these, too?  We've tried to do that with all of our `from_params` methods.  It just helps future-proof the code, in case arguments or argument orders change later.

nafitzgerald: Not sure what you mean. Do you mean add keyword args to the from_params method, or something else?

matt-gardner: Sorry for not being more clear; I just meant adding `dataset=dataset`, `min_count=min_count` to these lines.

PR #480, opened 2017-11-08
Elmo lstm

DeNeutoy:
@matt-peters FYI, still a little bit of work to do with tests and making sure I can load your weights etc.


allennlp/modules/elmo_lstm_cell.py
```diff
@@ -0,0 +1,227 @@
...
+            timestep_input = inputs[0: current_length_index + 1, index]
+
+            # Do the projections for all the gates all at once.
+            # Both have shape (batch_size, 4 * cell_size)
+            projected_input = self.input_linearity(timestep_input)
+            projected_state = self.state_linearity(previous_state)
+
+            # Main LSTM equations using relevant chunks of the big linear
+            # projections of the hidden state and inputs.
+            input_gate = torch.sigmoid(projected_input[:, 0 * self.cell_size:1 * self.cell_size] +
```
matt-peters: Including explicit parenthesis `projected_input[:, (0 * self.cell_size):(1 * self.cell_size)]` would make this easier to read.

allennlp/modules/elmo_lstm_cell.py
```diff
@@ -0,0 +1,227 @@
...
+                # Second conditional: Does the next shortest sequence beyond the current batch
+                # index require computation use this timestep?
+                while current_length_index < (len(batch_lengths) - 1) and \
+                                batch_lengths[current_length_index + 1] > index:
+                    current_length_index += 1
+
+            # Actually get the slices of the batch which we
+            # need for the computation at this timestep.
+            # shape (batch_size, cell_size)
+            previous_memory = full_batch_previous_memory[0: current_length_index + 1].clone()
```
matt-peters: Why is it necessary to `clone` here and on the next line?

allennlp/modules/elmo_lstm_cell.py
```diff
@@ -0,0 +1,227 @@
...
+            memory_init = torch.tanh(projected_input[:, 2 * self.cell_size:3 * self.cell_size] +
+                                     projected_state[:, 2 * self.cell_size:3 * self.cell_size])
+            output_gate = torch.sigmoid(projected_input[:, 3 * self.cell_size:4 * self.cell_size] +
+                                        projected_state[:, 3 * self.cell_size:4 * self.cell_size])
+            memory = input_gate * memory_init + forget_gate * previous_memory
+
+            # Here is the non-standard part of this LSTM cell; first, we clip the
+            # memory cell, then we project the output of the timestep to a smaller size
+            # and again clip it.
+            memory = torch.clamp(memory, -self.memory_cell_clip_value, self.memory_cell_clip_value)
```
matt-peters: You could use the in place `memory.clamp_` version, it will be faster if it avoids a new allocation and copy. Same with the other clamp lower down.

allennlp/modules/elmo_lstm_cell.py
```diff
@@ -0,0 +1,227 @@
...
+                                          self.state_projection_clip_value)
+
+            # Only do dropout if the dropout prob is > 0.0 and we are in training mode.
+            if dropout_mask is not None and self.training:
+                timestep_output = timestep_output * dropout_mask[0: current_length_index + 1]
+
+            # We've been doing computation with less than the full batch, so here we create a new
+            # variable for the the whole batch at this timestep and insert the result for the
+            # relevant elements of the batch into it.
+            full_batch_previous_memory = Variable(full_batch_previous_memory.data.clone())
```
matt-peters: Why is it necessary to `clone` here and create a new `Variable`?

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -0,0 +1,146 @@
...
+"""
+
+from typing import Optional, Tuple
+import torch
+from torch.nn.utils.rnn import PackedSequence, pad_packed_sequence
+from allennlp.modules.elmo_lstm_cell import ElmoLstmCell
+from allennlp.common.checks import ConfigurationError
+
+
+class ElmoLstm(torch.nn.Module):
```
matt-peters: Same comment here as above regarding name.  This is a generic `StackedBidirectionalLSTMWithOutputProjecetionAndResidualConnection`

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -0,0 +1,146 @@
...
+    recurrent_dropout_probability: ``float``, optional (default = 0.0)
+        The dropout probability to be used in a dropout scheme as stated in
+        `A Theoretically Grounded Application of Dropout in Recurrent Neural Networks
+        <https://arxiv.org/abs/1512.05287>`_ .
+    state_projection_clip_value: ``float``, optional, (default = 3)
+        The magnitude with which to clip the hidden_state after projecting it.
+    memory_cell_clip_value: ``float``, optional, (default = 3)
+        The magnitude with which to clip the memory cell.
+    """
+    def __init__(self,
```
matt-peters: It'd be easier to maintain if you just add `num_layers` then relegate the rest to `**kwargs` that are passed down the LSTMCell.  E.g. you also need to modify the default behavior of `state_projection_clip_value` here too.  Or imagine adding an option to the LSTMCell, you will also need to modify this code too.

I'd even tie them together more closely and move all of the cell related code into this file to make the connection explicit.  I find there is cognitive drain to switch between files for closely related code.

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -0,0 +1,146 @@
...
+            self.add_module('forward_layer_{}'.format(layer_index), forward_layer)
+            self.add_module('backward_layer_{}'.format(layer_index), backward_layer)
+            forward_layers.append(forward_layer)
+            backward_layers.append(backward_layer)
+        self.forward_layers = forward_layers
+        self.backward_layers = backward_layers
+
+    def forward(self,  # pylint: disable=arguments-differ
+                inputs: PackedSequence,
+                initial_state: Optional[Tuple[torch.Tensor, torch.Tensor]] = None):
```
matt-peters: Seems you need a type hint here for the return value.

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        Parameters
+        ----------
+        inputs : ``PackedSequence``, required.
+            A batch first ``PackedSequence`` to run the stacked LSTM over.
+        initial_state : Tuple[torch.Tensor, torch.Tensor], optional, (default = None)
+            A tuple (state, memory) representing the initial hidden state and memory
+            of the LSTM. Each tensor has shape (1, batch_size, output_dimension).
```
matt-peters: The initial state should have the same size/dimension as the output `final_states`.

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -0,0 +1,146 @@
...
+        Returns
+        -------
+        output_sequence : ``torch.FloatTensor``
+            The encoded sequence of shape (num_layers, batch_size, sequence_length, hidden_size)
+        final_states: ``Tuple[torch.FloatTensor, torch.FloatTensor]``
+            The per-layer final (state, memory) states of the LSTM, with shape
+            (num_layers, batch_size, 2 * hidden_size) and  (num_layers, batch_size, 2 * cell_size)
+            respectively.
+        """
+        if not initial_state:
```
matt-peters: It's more explicit / readable to do `if initial_state is not None:`

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -0,0 +1,146 @@
...
+            forward_cache = forward_output_sequence
+            backward_cache = backward_output_sequence
+
+            forward_output_sequence, forward_state = forward_layer(forward_output_sequence,
+                                                                   batch_lengths,
+                                                                   state)
+            backward_output_sequence, backward_state = backward_layer(backward_output_sequence,
+                                                                      batch_lengths,
+                                                                      state)
+            # Vanilla highway layers.
```
matt-peters: I'd call this a residual or skip connection.

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -0,0 +1,146 @@
...
+                forward_output_sequence += forward_cache
+                backward_output_sequence += backward_cache
+
+            sequence_outputs.append(torch.cat([forward_output_sequence,
+                                               backward_output_sequence], -1))
+            # Append the state tuples in a list, so that we can return
+            # the final states for all the layers.
+            final_states.append((torch.cat([forward_state[0], backward_state[0]], -1),
+                                 torch.cat([forward_state[1], backward_state[1]], -1)))
+        # TODO(Mark): figure out the best api to return these. We need to specify the
```
matt-peters: We should move this logic elsewhere.

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -0,0 +1,146 @@
...
+
+        final_states = []
+        sequence_outputs = []
+        for layer_index, (forward_layer, backward_layer, state) in enumerate(zip(self.forward_layers,
+                                                                                 self.backward_layers,
+                                                                                 hidden_states)):
+            forward_cache = forward_output_sequence
+            backward_cache = backward_output_sequence
+
+            forward_output_sequence, forward_state = forward_layer(forward_output_sequence,
```
matt-peters: I don't think this logic will work.  The forward and backward LSTMs have different states, but here you are passing the same state to both directions.

matt-peters: You need to also unpack the forward and backward directions.
At each iteration of the loop, `state` is tuple with shape `(batch_size, 2 * dim)` (for each entry).
So you need something like
```
forward_hidden_state, backward_hidden_state = state[0].split(self.hidden_size, 1)
forward_memory_state, backward_memory_state = state[0].split(self.cell_size, 1)
forward_state = (forward_hidden_state, forward_memory_state)
backward_state = (backward_hidden_state, backward_memory_state)
```

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -0,0 +1,146 @@
...
+    def forward(self,  # pylint: disable=arguments-differ
+                inputs: PackedSequence,
+                initial_state: Optional[Tuple[torch.Tensor, torch.Tensor]] = None):
+        """
+        Parameters
+        ----------
+        inputs : ``PackedSequence``, required.
+            A batch first ``PackedSequence`` to run the stacked LSTM over.
+        initial_state : Tuple[torch.Tensor, torch.Tensor], optional, (default = None)
+            A tuple (state, memory) representing the initial hidden state and memory
```
matt-peters: This docstring is inconsistent with the one in the class definition above. I'd just eliminate it otherwise it's repetitive and hard to maintain (need to change things in multiple places for a single change).

tests/modules/stacked_elmo_lstm_test.py
```diff
@@ -0,0 +1,32 @@
...
+        input_tensor[2, 2:, :] = 0.
+        input_tensor[3, 1:, :] = 0.
+        input_tensor = pack_padded_sequence(input_tensor, [5, 4, 2, 1], batch_first=True)
+        lstm = ElmoLstm(num_layers=2,
+                        input_size=3,
+                        hidden_size=5,
+                        cell_size=7,
+                        memory_cell_clip_value=2,
+                        state_projection_clip_value=1)
+        output_sequence, lstm_state = lstm(input_tensor)
```
matt-peters: Suggest adding a test here that also passes in an input state. It would have caught the issue above about forward/backward states since the shapes would have been incompatible.

matt-peters: One other interesting test to add too and would be easy to incorporate here is to make one of the sequence lengths=0, pass in an initial state, then check that the final state is the same as the initial state.

allennlp/modules/elmo_lstm_cell.py
```diff
@@ -0,0 +1,227 @@
...
+        self.hidden_size = hidden_size
+        self.cell_size = cell_size
+
+        self.go_forward = go_forward
+        self.state_projection_clip_value = state_projection_clip_value
+        self.memory_cell_clip_value = memory_cell_clip_value
+        self.recurrent_dropout_probability = recurrent_dropout_probability
+
+        # We do the projections for all the gates all at once.
+        self.input_linearity = torch.nn.Linear(input_size, 4 * cell_size, bias=False)
```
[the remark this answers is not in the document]
DeNeutoy: Yeah ... That is an optimisation that I think will be worth making once I port this to C++, but as this is going to be slow anyway, I agree that it's not worth it until we've made other things faster.

allennlp/modules/stacked_elmo_lstm.py
DeNeutoy: I think it will be cleanest if we define a set of mixing functions elsewhere and pass one of them to this module, rather than having another layer after this.

matt-peters: There is other functionality that we need to add to make easy to use and work with ELMo -- e.g. removing `<S>` & `</S>`, loading the pretrained weights, making it stateful -- all before the weighting of layers can happen.  It also needs to be paired with the specific ELMo token embedder so I was planning to add those elsewhere in something ELMo specific.

allennlp/modules/lstm_cell_with_projection.py
```diff
@@ -0,0 +1,238 @@
...
+        if initial_state is None:
+            full_batch_previous_memory = Variable(inputs.data.new(batch_size,
+                                                                  self.cell_size).fill_(0))
+            full_batch_previous_state = Variable(inputs.data.new(batch_size,
+                                                                 self.hidden_size).fill_(0))
+        else:
+            full_batch_previous_state = initial_state[0].squeeze(0)
+            full_batch_previous_memory = initial_state[1].squeeze(0)
+
+            print(full_batch_previous_state)
```
matt-peters: These should be removed.

allennlp/modules/stacked_elmo_lstm.py
matt-peters: The shape of the initial states should be `(num_layers, batch_size, dim)`

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -0,0 +1,168 @@
...
+"""
+
+from typing import Optional, Tuple, List
+import torch
+from torch.nn.utils.rnn import PackedSequence, pad_packed_sequence
+from allennlp.modules.lstm_cell_with_projection import LSTMCellWithProjection
+from allennlp.common.checks import ConfigurationError
+
+
+class ElmoLstm(torch.nn.Module):
```
matt-peters: Nit, but can you change this name to `ELMoLSTM`?  To be consistent with everything else.

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -0,0 +1,168 @@
...
+        output_sequence : ``torch.FloatTensor``
+            The encoded sequence of shape (num_layers, batch_size, sequence_length, hidden_size)
+        final_states: ``Tuple[torch.FloatTensor, torch.FloatTensor]``
+            The per-layer final (state, memory) states of the LSTM, with shape
+            (num_layers, batch_size, 2 * hidden_size) and  (num_layers, batch_size, 2 * cell_size)
+            respectively. The last dimension is duplicated because it contains the state/memory
+            for both the forward and backward layers.
+        """
+        if initial_state is None:
+            hidden_states: List[Optional[Tuple[torch.Tensor,
```
matt-peters: Why is the type necessary here?  This doesn't actually do anything when running the code, right?

DeNeutoy: No, it doesn't - however, it allows MyPy to run static type checking. I agree it's a bit annoying when the types are long like this, but the benefit is worth it for a large python codebase.

allennlp/modules/stacked_elmo_lstm.py
DeNeutoy: It's easier to read as it is, and fits with the rest of the codebase, so i'd rather not. Do you mean it's more consistent with your TF code?

allennlp/modules/stacked_elmo_lstm.py
matt-peters: Got it.

allennlp/modules/stacked_elmo_lstm.py
matt-peters: Consistent with the paper and everywhere else where we write `ELMo`.  It's an acronym.  If we applied your reasoning to all of the class names it'd also be `Lstm` right?

matt-peters: Which I see you have done `Lstm` -- but pytorch itself uses `LSTM` as does just about every other library I've seen.

matt-gardner: The recommendation comes from here: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case.  I think something like `XmlHttpRequest` is way easier to read than `XMLHTTPRequest` - in the second you have to do word segmentation as you read the identifier, while in the first you don't.

matt-peters: Sure, but: (1) that's a Java style guide (2) allennlp doesn't follow any established python style guide that I can tell and (3) every other deep learning library uses `LSTM`.

matt-gardner: https://stackoverflow.com/questions/15526107/acronyms-in-camelcase

https://github.com/tensorflow/tensorflow/blob/18003982ff9c809ab8e9b76dd4c9b9ebc795f4b8/tensorflow/contrib/lite/models/speech_speakerid_model_test.cc#L38-L43
https://github.com/tensorflow/tensorflow/blob/5e6b16a9261d95eee321a67e38ea0a83357f3bf8/tensorflow/contrib/ndlstm/python/lstm2d_test.py#L64
https://github.com/tensorflow/tensorflow/blob/14f564990e7548b4a5c41da0fc533ec5a7a25abe/tensorflow/contrib/rnn/python/kernel_tests/rnn_cell_test.py#L225
https://github.com/tensorflow/tensorflow/blob/b10f50ff15944badb7262a207f6628dfa52d6a9d/tensorflow/contrib/ndlstm/python/misc_test.py#L36
https://github.com/tensorflow/tensorflow/blob/d4ad9c73969c45d1a224ebfc43eb645b9860216b/tensorflow/contrib/lite/kernels/internal/optimized/optimized_ops.h#L1914

Which is easier for you to read: `CRFLSTMModel` or `CrfLstmModel`?  You're right that it doesn't matter very much if you have an identifier that's just an acronym, like `LSTM`.  It matters a whole lot more when the acronym is just part of a longer identifier, like `ELMoLSTM`.  Where do the acronyms start and stop?  How many are there?  It unambiguously hurts readability to have to parse that yourself.  Whether other people follow best practices, or even their own style guides (Google's naming conventions on camel case are not unique to java), shouldn't dictate how we write our code.

For (2), we generally follow Google's style guide: https://google.github.io/styleguide/pyguide.html#Indentation.

matt-peters: We clearly have different views on this, and it's your library, so I'll stop after this comment. 
 I think the overall goal should be to make the library as easy to use and contribute to as possible.  Using non-standard python things or inconsistency with the rest of the deep learning world hinders both goals.

In tensorflow, LSTM is always used in the public facing APIs even as part of longer strings (https://www.tensorflow.org/api_docs/python/tf/contrib/rnn).

At lower levels, they aren't consistent -- e.g. one of those test files mixes both `LSTM` and `Lstm`  (https://github.com/tensorflow/tensorflow/blob/14f564990e7548b4a5c41da0fc533ec5a7a25abe/tensorflow/contrib/rnn/python/kernel_tests/rnn_cell_test.py#L173).

Python adopts the full upper case for acronyms in the standard library, e.g.

https://docs.python.org/3/library/urllib.request.html#urllib.request.HTTPDefaultErrorHandler
https://docs.python.org/3/library/xmlrpc.server.html#xmlrpc.server.SimpleXMLRPCServer

PR #481, opened 2017-11-09
Implemented batched decoding

matt-gardner:
There are two kinds of batching going on here: we process several training instances at once, _and_ we process several states for each training instance at once.  Hopefully the docs I wrote make this clear enough.

I switched to a simpler copying task, where I'm training the model to copy the input given noisy supervision, and I'm still having trouble getting it to train correctly.  But at least this logic runs, and I _think_ I did everything correctly here.  It gives about an 8x speedup when using a batch size of 60 instead of a batch size of 1 (and that's on top of a minor speed up using a batch size of 1 that I got from batching together several states for each instance).  I haven't tried profiling this at all, and there might be more gains to be had somewhere in here with some careful profiling.

### DIFF
--- a/allennlp/models/encoder_decoders/wikitables_semantic_parser.py
+++ b/allennlp/models/encoder_decoders/wikitables_semantic_parser.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from typing import Dict, Generator, List, Set, Tuple
 
 from overrides import overrides
@@ -112,9 +113,9 @@ class WikiTablesSemanticParser(Model):
         # (batch_size, question_length, encoder_output_dim)
         encoder_outputs = self._encoder(embedded_input, question_mask)
 
+        # This will be our initial hidden state and memory cell for the decoder LSTM.
         final_encoder_output = encoder_outputs[:, -1]  # (batch_size, encoder_output_dim)
-        decoder_context = Variable(encoder_outputs.data.new(batch_size, self._encoder.get_output_dim())
-                                   .fill_(0))
+        memory_cell = Variable(encoder_outputs.data.new(batch_size, self._encoder.get_output_dim()).fill_(0))
 
         if target_action_sequences is not None:
             # Remove batch dimension and trailing dimension (from ListField[LabelField]]).
@@ -136,7 +137,7 @@ class WikiTablesSemanticParser(Model):
         # won't have to do any index selects, or anything, we'll just do some `torch.cat()`s.
         initial_score_list = [initial_score[i] for i in range(batch_size)]
         initial_hidden_state = [final_encoder_output[i] for i in range(batch_size)]
-        initial_memory_cell = [decoder_context[i] for i in range(batch_size)]
+        initial_memory_cell = [memory_cell[i] for i in range(batch_size)]
         initial_attended_question = [attended_question[i] for i in range(batch_size)]
         encoder_output_list = [encoder_outputs[i] for i in range(batch_size)]
         question_mask_list = [question_mask[i] for i in range(batch_size)]
@@ -165,8 +166,10 @@ class WikiTablesSemanticParser(Model):
             else:
                 num_steps = self._max_decoding_steps
             best_final_states = self._beam_search.search(num_steps, initial_state, self._decoder_step)
-            best_final_state = best_final_states[0]
-            outputs['best_action_sequence'] = [best_final_state.action_history]
+            best_action_sequences = []
+            for i in range(batch_size):
+                best_action_sequences.append(best_final_states[i][0].action_history)
+            outputs['best_action_sequence'] = best_action_sequences
             # TODO(matt): compute accuracy here.
             return outputs
 
@@ -237,7 +240,7 @@ class WikiTablesDecoderState(DecoderState['WikiTablesDecoderState']):
         This holds the LSTM hidden state for each element of the group.  Each variable has shape
         ``(decoder_output_dim,)``.
     memory_cell : ``List[Variable]``
-        This holds to LSTM memory cell for each element of the group.  Each variable has shape
+        This holds the LSTM memory cell for each element of the group.  Each variable has shape
         ``(decoder_output_dim,)``.
     attended_question : ``List[Variable]``
         This holds the attention-weighted sum over the question representations that we computed in
@@ -389,7 +392,8 @@ class WikiTablesDecoderStep(DecoderStep[WikiTablesDecoderState]):
     @overrides
     def take_step(self,
                   state: WikiTablesDecoderState,
-                  allowed_actions: List[Set[int]] = None) -> Generator[WikiTablesDecoderState, None, None]:
+                  max_actions: int = None,
+                  allowed_actions: List[Set[int]] = None) -> List[WikiTablesDecoderState]:
         # Outline here: first we'll construct the input to the decoder, which is a concatenation of
         # an embedding of the decoder input (the last action taken) and an attention over the
         # question.  Then we'll update our decoder's hidden state given this input, and recompute
@@ -421,12 +425,13 @@ class WikiTablesDecoderStep(DecoderStep[WikiTablesDecoderState]):
         logits = self._output_projection_layer(torch.cat([hidden_state, attended_question], -1))
         output_mask = state.get_output_mask()
         log_probs = util.masked_log_softmax(logits, output_mask)
-        return self._yield_new_states(state,
-                                      log_probs,
-                                      hidden_state,
-                                      memory_cell,
-                                      attended_question,
-                                      allowed_actions)
+        return self._compute_new_states(state,
+                                        log_probs,
+                                        hidden_state,
+                                        memory_cell,
+                                        attended_question,
+                                        allowed_actions,
+                                        max_actions)
 
     def attend_on_question(self,
                            query: Variable,
@@ -441,7 +446,6 @@ class WikiTablesDecoderStep(DecoderStep[WikiTablesDecoderState]):
         method on the main parser module can call it on the initial hidden state, to simplify the
         logic in ``take_step``.
         """
-
         # (group_size, question_length)
         question_attention_weights = self._input_attention(query,
                                                            encoder_outputs,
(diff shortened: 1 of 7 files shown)


allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -209,39 +220,142 @@ def from_params(cls, vocab, params: Params) -> 'WikiTablesSemanticParser':
...
-        being an ``int``.  In some cases, items from this list will get passed as inputs to
-        ``DecodeStep``, so this must return items that are compatible with inputs to your
-        ``DecodeStep`` class.
+    batch_indices : ``List[int]``
+        Passed to super class; see docs there.
+    action_history : ``List[List[int]]``
+        Passed to super class; see docs there.
+    score : ``List[Variable]``
+        Passed to super class; see docs there.
+    hidden_state : ``List[Variable]``
```
DeNeutoy: We should keep these type annotations as `Tensors` because:
1. They provide more information to the user.
2. Pytorch will eventually remove `Variables`.
3. We have `Tensors` everywhere in the codebase where they are `Variables`.
4. With the exception of a few places where we catch these errors, like in `Metrics` everything is implicitly a `Variable` anyway.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -275,93 +389,109 @@ def __init__(self,
...
+                    # metrics easier if this skipping logic happened in the trainer - the trainer could
+                    # keep track of when the top action wasn't allowed.  NOTE: without batching, when
+                    # there are ~80 possible actions in each state, this gives a ~2x slowdown.  So I'm
+                    # keeping it for now.  I'll re-evaluate once batching is implemented.
+                    continue
+                # We'll yield a bunch of states here that all have a `group_size` of 1, so that the
+                # learning algorithm can decide how many of these it wants to keep, and it can just
+                # regroup them later, as that's a really easy operation.
+                new_action_history = state.action_history[group_index] + [action]
+                # TODO(matt): we might need to normalize this by length eventually.  If the training
```
DeNeutoy: I don't think this is too problematic - the `WikiTablesDecoderState` contains everything required to normalise with respect to its action sequence, so how the learning algorithm/search proceedure decides to expand nodes doesn't require any action here, I don't think. Seems like you can just remove this `TODO`.

allennlp/nn/decoding/decoder_step.py
```diff
@@ -22,19 +22,28 @@ class DecoderStep(torch.nn.Module, Generic[StateType]):
     """
     def take_step(self,
                   state: StateType,
-                  allowed_actions: Set = None) -> Generator[StateType, None, None]:
+                  allowed_actions: List[Set] = None) -> Generator[StateType, None, None]:
```
DeNeutoy: Maybe add an `Optional[]` here and in the docstring, and explain why this is the case.

allennlp/nn/decoding/maximum_marginal_likelihood.py
```diff
@@ -33,43 +34,50 @@ def decode(self,
...
-                        break
-                    if next_state.is_finished():
-                        finished_states.append(next_state)
-                    else:
-                        next_states.append(next_state)
+            # We group together all current states to get more efficient (batched) computation.
+            grouped_state = states[0].combine_states(states)
+            allowed_actions = self._get_allowed_actions(grouped_state, allowed_transitions)
+            generator = decode_step.take_step(grouped_state, allowed_actions)
+            while True:
```
DeNeutoy: A for loop over the generator seems a little more pythonic here.

allennlp/nn/decoding/maximum_marginal_likelihood.py
```diff
@@ -81,23 +89,49 @@ def _create_allowed_transitions(targets: torch.Tensor,
...
         We use this to prune the set of actions we consider during decoding, because we only need
         to score the sequences in ``targets``.
-
-        While the example and types above say that this should be a 2D tensor, the implementation
-        logic is actually more flexible - any sequence of sequences is valid here, with any
-        contained types.  So your targets could be strings, for instance, if your ``DecoderStep``
-        and ``DecoderState`` objects know how to handle that.
         """
-        allowed_transitions: Dict[Tuple[int, ...], Set[int]] = defaultdict(set)
+        assert targets.dim() == 3, "targets tensor needs to be batched!"
```
DeNeutoy: raise a `ConfigurationError`, rather than asserting

allennlp/nn/decoding/maximum_marginal_likelihood.py
```diff
@@ -33,43 +34,50 @@ def decode(self,
...
+        # of finished states.
+        batch_scores = self._group_scores_by_batch(finished_states)
+        loss = 0
+        for scores in batch_scores.values():  # we don't care about the batch index, just the scores
+            loss += -util.logsumexp(torch.cat(scores))
+        return {'loss': loss / len(batch_scores)}
 
     @staticmethod
     def _create_allowed_transitions(targets: torch.Tensor,
-                                    target_mask: torch.Tensor = None) -> Dict[Tuple[int, ...], Set[int]]:
```
DeNeutoy: These arguments don't seem to be optional at the moment. Maybe just remove these defaults for now?

allennlp/nn/decoding/maximum_marginal_likelihood.py
```diff
@@ -33,43 +34,50 @@ def decode(self,
                decode_step: DecoderStep,
                targets: torch.Tensor = None,
```
DeNeutoy: These arguments aren't optional at the moment, either raise a `NotImplementedError` or remove.

allennlp/nn/decoding/beam_search.py
```diff
@@ -22,8 +22,8 @@ def search(self,
         finished_states = []
         states = [initial_state]
         step_num = 1
+        print(num_steps)
```
DeNeutoy: Remove print statement.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -121,22 +118,36 @@ def forward(self,  # type: ignore
...
+                                                                  question_mask)
+
+        # To make grouping states together in the decoder easier, we convert the batch dimension in
+        # all of our tensors into an outer list.  For instance, the encoder outputs have shape
+        # `(batch_size, question_length, encoder_output_dim)`.  We need to convert this into a list
+        # of `batch_size` tensors, each of shape `(question_length, encoder_output_dim)`.  Then we
+        # won't have to do any index selects, or anything, we'll just do some `torch.cat()`s.
+        initial_score_list = [initial_score[i] for i in range(batch_size)]
+        initial_hidden_state = [final_encoder_output[i] for i in range(batch_size)]
+        initial_memory_cell = [decoder_context[i] for i in range(batch_size)]
```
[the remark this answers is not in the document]
matt-gardner: https://pdfs.semanticscholar.org/7a73/0b063db113fe44b1a4ec97c1eeff7a5c4669.pdf?_ga=2.137572790.703157336.1510596418-1538472344.1478539129.  See section 4.  The memory cell is _why_ it's called long short-term _memory_.  See also [here](https://arxiv.org/pdf/1705.07393.pdf), and many others.  "Cell" or "memory cell" is the standard term.

But, this does make me notice that I still need to change the name for `decoder_context` to match.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
matt-gardner: I'm not sure I agree with this.  If I have a `Variable`, that means I need to call `.data` to get the tensor.  I think having the type annotated as `Tensor` when I actually have a `Variable` is confusing.  If our goal is that the user knows the data type of the underlying tensor, we can easily solve that with a docstring comment.

Do you have any more info on the eventual union of `Tensor` and `Variable` in pytorch?  Either way, until that happens, I'd rather have our code actually have correct types, and who knows how long it'll actually take for the pytorch folks to merge them, and what it will look like when they do.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py

allennlp/nn/decoding/beam_search.py
matt-gardner: Done.

allennlp/nn/decoding/decoder_step.py
matt-gardner: I'm not sure that explicit `Optional` annotations really add anything - you already have the `= None` there, and `Optional` is redundant with that.  But I updated the docstring.

allennlp/nn/decoding/maximum_marginal_likelihood.py
matt-gardner: Done.

allennlp/nn/decoding/maximum_marginal_likelihood.py
matt-gardner: That was what I did at first, but something about the way I did it was wrong, and it didn't work.  For a different reason, I removed the generator entirely and now just return a list, so I do the right thing when looping over it.

allennlp/nn/decoding/maximum_marginal_likelihood.py
matt-gardner: Done.

allennlp/nn/decoding/maximum_marginal_likelihood.py
matt-gardner: The original idea behind `ConfigurationError` was that it was a single error that covered all kinds of problems in the config file specification, and could be caught and handled appropriately in some situations.  If you run into this error, it is a problem with your code, not with your configuration file, so `ConfigurationError` doesn't fit.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -209,39 +220,142 @@ def from_params(cls, vocab, params: Params) -> 'WikiTablesSemanticParser':
...
+        Passed to super class; see docs there.
+    action_history : ``List[List[int]]``
+        Passed to super class; see docs there.
+    score : ``List[Variable]``
+        Passed to super class; see docs there.
+    hidden_state : ``List[Variable]``
+        This holds the LSTM hidden state for each element of the group.  Each variable has shape
+        ``(decoder_output_dim,)``.
+    memory_cell : ``List[Variable]``
+        This holds to LSTM memory cell for each element of the group.  Each variable has shape
```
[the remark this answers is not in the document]

matt-gardner:
It turns out that beam search was still totally broken.  I fixed that, and discovered that it was _way_ too slow, because I was computing too many unnecessary states in `take_step`, so I had to pass in more information to that method.  But this is now reasonably fast, with the beam search during validation not taking much additional time.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
DeNeutoy: What's more confusing is if we aren't consistent about this - everywhere apart from here we use tensor annotations, so I think you should make this consistent and if you feel strongly, change the annotations across the codebase. In 90% of cases, you don't have to think about whether a tensor is a variable. I don't know more about when the change is going to come, however, but at least in this PR I think you should be consistent, given that it is strictly worse to have a mix of tensor and variable types scattered throughout allennlp.

matt-gardner: I'd prefer to move all of our type annotations to `Variable` now, in places where the type actually is a variable.  @joelgrus did this in #374 - should we merge that PR, or part of it?  Do either of you have any objections to switching to `Variable` and adding data type info to docstrings?

PR #487, opened 2017-11-13
make API_ROOT depend on window.location.origin

joelgrus:
we want to hardcode the backend as localhost:8000 for development but not for production. this is a quick and dirty way to make that happen

### DIFF
--- a/demo/src/api-config.js
+++ b/demo/src/api-config.js
@@ -1,0 +1,18 @@
+/**
+ * The backend always runs on port 8000. In production we also
+ * serve the frontend from there. However, for development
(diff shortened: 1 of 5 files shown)


demo/src/api-config.js
```diff
@@ -0,0 +1,18 @@
...
+ * serve the frontend from there. However, for development
+ * we want to `npm run serve` the unminified js on port 3000.
+ * This allows us to get the correct API root either way.
+ */
+
+let apiRoot;
+
+const origin = window && window.location && window.location.origin;
+
+if (origin === 'http://localhost:3000') {
```
schmmd: Nice.  I would not have thought of this.

PR #496, opened 2017-11-15
Add a script to handle production deploys.

schmmd:


schmmd:
This is for TeamCity deploys.  I'm bad with bash--any and all feedback very welcome.

PR #497, opened 2017-11-15
Multi text field embedder

DeNeutoy:
I encountered the need to embed a single token representation in two ways when playing around with the coref model to see if I can improve the performance. We could obviously replace the `BasicTokenEmbedder` with this, but it is a massively breaking change and also it's good enough for 90% of use cases, so I figured it might be better to separate this.


matt-gardner:
What if you just did:

```json
{
  ...
  "token_indexers": {
    "tokens": {...},
    "tokens2": {...}
  },
  ...
  "text_field_embedder": {
    "tokens": {...},
    "tokens2": {...}
  }
```

Would that work for your use case?

DeNeutoy:
Ugh, obviously. Thanks.

matt-gardner:
I'm going to open an issue to add a tutorial for stuff like this.  Maybe something like "What if I want to do...", and have sections for various random things a person might want to put in their models, and how best to accomplish it with AllenNLP.

PR #501, opened 2017-11-16
NER demo

joelgrus:
it's a lot like the SRL demo with only one frame. two UX differences:

(1) I merged consecutive identical tags and centered the label across them
(2) there are mouseover tooltips for the tags (but they're not very interesting). 

anyway, I think they're good changes, it's easy enough to make similar changes to the SRL demo if we want to.

some quibbles:

* the predictor is called `crf-tagger` (as is the model), but on the front end it's called `named-entity-recognition` so that's kind of a mismatch
* in an ideal world I'd write a test for the `CrfTagger` predictor, but that would require me to train a toy version of the model. if someone feels strongly about it, I can do that.
* we are starting to run out of space on the page to list models (indeed, if you make the demo window too narrow it looks really bad), we should probably think about this.
* I made up some demo sentences, I can change them if anyone has any better suggestions.


allennlp/service/predictors/crf_tagger.py
```diff
@@ -14,15 +14,33 @@ class CrfTaggerPredictor(Predictor):
     """
     def __init__(self, model: Model, dataset_reader: DatasetReader) -> None:
```
matt-gardner: Couldn't you use this `Predictor` with a trained `SimpleTagger` also?  Why is it called `CrfTaggerPredictor`?

joelgrus: I think you're right, but there are some complications around the way we're using predictor names as the source of truth in our front-end.

I think we need to come up with different abstractions than the `DEFAULT_MODELS` and `DEFAULT_PREDICTORS` dicts that we have in `commands/__init__.py`. I have always felt that's not the right way to do it, let me noodle on it a bit.

schmmd:
Is it possible to name the model `crf-ner-tagger` instead?

joelgrus:
OK, this is ready for another look. I replaced both the `simple-tagger` and `crf-tagger` predictors with a single `sentence-tagger` predictor. (To do that I moved some logic from the predictor into `CrfTagger.decode()`.) This also allowed me to change the ner endpoint to `predict/named-entity-recognition`.

allennlp/service/predictors/sentence_tagger.py
```diff
@@ -0,0 +1,44 @@
...
+    the :class:`~allennlp.models.crf_tagger.CrfTagger` model
+    and also
+    the :class:`~allennlp.models.simple_tagger.SimpleTagger` model.
+    """
+    def __init__(self, model: Model, dataset_reader: DatasetReader) -> None:
+        super().__init__(model, dataset_reader)
+        self._tokenizer = SpacyWordSplitter(language='en_core_web_sm', pos_tags=True)
+
+    @overrides
+    def _json_to_instance(self, json: JsonDict) -> Instance:
```
matt-gardner: Not necessarily for this PR, just thinking about how we can make the API better here, so we don't have to override `predict_json` all the time.  What if `_json_to_instance` returned a `Tuple[Instance, JsonDict]`, where the json dict contains additional things to pass to the output?

joelgrus: I agree we should fix this problem, let me think a little bit about what's the best approach

PR #502, opened 2017-11-16
Add constrained decoding to the DecoderStep

matt-gardner:
So far, this is just setting up a variable number of valid actions, and changing the model to do a dot product with a smaller set of action embeddings, instead of a softmax over all actions.  Next, I will use the grammar and action history to actually select a set of valid actions at each state.

### DIFF
--- a/allennlp/models/encoder_decoders/wikitables_semantic_parser.py
+++ b/allennlp/models/encoder_decoders/wikitables_semantic_parser.py
@@ -479,10 +479,10 @@ class WikiTablesDecoderStep(DecoderStep[WikiTablesDecoderState]):
 
     def _embed_actions(self, actions: List[List[int]]) -> Tuple[Variable, Variable]:
         """
-        Returns an embedded representation of the given actions.  The input is assumed to have
-        shape ``(group_size, num_actions)``, though the number of actions for each group element
-        need not be constant.  Because of this, we need to pad the number of actions so we have an
-        evenly-shaped tensor, and we additionally return a mask.
+        Returns an embedded representation of the given actions.  The input is a list of valid
+        actions for each element in the group, which may have variable length.  We pad these to
+        have shape ``(group_size, num_actions)``, and then embed them.  Because of the padding, we
+        additionally return a mask.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -452,24 +477,56 @@ def attend_on_question(self,
         # (group_size, encoder_output_dim)
         return util.weighted_sum(encoder_outputs, question_attention_weights)
 
+    def _embed_actions(self, actions: List[List[int]]) -> Tuple[Variable, Variable]:
+        """
+        Returns an embedded representation of the given actions.  The input is assumed to have
+        shape ``(group_size, num_actions)``, though the number of actions for each group element
```
DeNeutoy: Maybe it would be clearer to say something like: "The inputs are a list of valid actions for each element in the group, which may have variable length. These are then padded to have shape (group_size, num_actions) and embedded."

allennlp/nn/util.py
```diff
@@ -34,6 +34,26 @@ def get_lengths_from_binary_sequence_mask(mask: torch.Tensor):
...
+    that we can use it to construct a new tensor.
+
+    Some of our functions are agnostic as to whether they accept ``Tensors`` or ``Variables``, and
+    they just use ``Tensor`` for their type annotations.  We `require` ``Variables`` here, as we
+    call ``sequence_length.data.new()``.
+    """
+    # (batch_size, max_length)
+    ones = Variable(sequence_lengths.data.new(sequence_lengths.size(0), max_length).fill_(1))
+    range_tensor = ones.cumsum(dim=1)
+    return sequence_lengths.unsqueeze(1) >= range_tensor
```
DeNeutoy: Cast to `.long()` here.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -452,24 +477,56 @@ def attend_on_question(self,
...
+            The set of actions to embed.  This is typically a list of valid actions for each item
+            in the group of the current state.
+
+        Returns
+        -------
+        action_embeddings : float ``Variable``
+            An embedded representation of all of the given actions.  Shape is ``(group_size,
+            num_actions, action_embedding_dim)``, where ``num_actions`` is the maximum length of
+            the input ``actions`` lists.
+        action_mask : byte ``Variable``
```
DeNeutoy: Let's never use byte tensors.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
matt-gardner: :+1:, done.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
matt-gardner: Fair enough, changed.

allennlp/nn/util.py
matt-gardner: Done.

PR #503, opened 2017-11-16
introduce `DemoModel` abstraction

joelgrus:
change dict used by run/serve from 

{predictor_name -> archive_file} 
to 
{demo_model_name -> (predictor_name, archive_file)}

which is a better representation and (in particular) allows us to have different demo models that share the same `Predictor` wrapper

(this will allow us to have a single `sentence_tagger` predictor that can be used by the NER model and by (say) a future POS tagger)

### DIFF
--- a/allennlp/commands/__init__.py
+++ b/allennlp/commands/__init__.py
@@ -1,4 +1,4 @@
-from typing import Dict, NamedTuple
+from typing import Dict
 import argparse
 
 from allennlp.commands.serve import Serve
--- a/allennlp/commands/__init__.py
+++ b/allennlp/commands/__init__.py
@@ -8,39 +8,6 @@ from allennlp.commands.evaluate import Evaluate
 from allennlp.commands.subcommand import Subcommand
 from allennlp.service.predictors import DemoModel
 
-# This maps from the name of model (as known to the front-end)
-# to the ``DemoModel`` indicating the location of the trained model
-# and the type of the ``Predictor``.  This is necessary, as you might
-# have multiple models (for example, a NER tagger and a POS tagger)
-# that have the same ``Predictor`` wrapper.
-DEFAULT_MODELS = {
-        'machine-comprehension': DemoModel(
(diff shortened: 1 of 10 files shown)


allennlp/commands/__init__.py
```diff
@@ -6,13 +6,30 @@
 from allennlp.commands.train import Train
 from allennlp.commands.evaluate import Evaluate
 from allennlp.commands.subcommand import Subcommand
+from allennlp.service.predictors import DemoModel
 
-# a mapping from predictor `type` to the location of the trained model of that type
+# This maps from the name of model (as known to the front-end)
```
matt-gardner: "This maps from the name of the task"?

joelgrus: Right now it happens to be the name of the task, but in the future it might not be. 

For example, imagine at some point we want to put two different SRL models in our demo. then both will be "semantic-role-labeling" as the task, but they'll need to hit different

```
/predict/name-of-model
```

endpoints, like 

```
/predict/semantic-role-labeling-sota
/predict/semantic-role-labeling-simple
```

I don't know exactly what to call that. "model" doesn't feel 100% right but neither does "task"

matt-gardner: `/predict/semantic-role-labeling/sota`?

joelgrus: yeah, that makes sense, but it will require some more code changes if/when we do it

matt-gardner: Not that you should make this change now, but I think eventually it makes sense to have this split between task and model, so it'll be easier to have comparisons between models, and put your own in-development model into a live demo.

joelgrus:
FYI, in addressing the last comments, I also made a minor change where I moved `DEFAULT_MODELS` from `commands/__init__.py` to `commands/serve.py` and similarly `DEFAULT_PREDICTORS` to `commands/predict.py`. Those feel more like the right places for them as we continue to refactor this aspect of the code.

PR #505, opened 2017-11-16
make span f1 robust to zero length sequences

DeNeutoy:
@nikett This was the problem with calling the metric the way you were, as there could have been fully padded sequences because you were using a `ListField[TextField]`.


matt-gardner:
A simple addition to a test would be nice.

PR #507, opened 2017-11-16
Remove tests from Dockerfile.

schmmd:
As a result of this change, building a Dockerfile will not run tests.  This was added so DockerCloud would test to make sure the images were working.  In TeamCity, it's easy to break these tests out so they run within the Docker container after it is built.  TeamCity also runs our full suite of tests (mypy, pylint, pytest) rather than just the pytest ones.  Containers are only deployed when the full suite of tests pass.

An upside here is that the builds are faster, because they don't run tests twice.  As an alternative, we could remove the stage where pytest is run from TeamCity, but I think that makes less sense.

A potential downside is that now people can create Dockerfiles where tests fail, although I'm not convinced this was the right way to do things in the first place.


schmmd:
@matt-gardner @DeNeutoy FYI--feel free to chime in if you have an opinion here.  I think this is best, but alternatives are just fine.

PR #509, opened 2017-11-17
Updates to `PytorchSeq2SeqWrapper`

matt-peters:
* supports stateful RNNs
* supports RNNs that return activations for each layer


allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
```diff
@@ -29,76 +31,155 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
     Note that we *require* you to pass sequence lengths when you call this module, to avoid subtle
     bugs around masking.  If you already have a ``PackedSequence`` you can pass ``None`` as the
     second parameter.
+
+    We support stateful RNNs where the final state from each batch is used as the initial
+    state for the subsequent batch by passing ``stateful=True`` to the constructor.  In this case,
+    ``max_batch_size`` is the maximum batch size allowed (although batches of any size less
+    then the maximum are also supported).  If ``stateful=True`` then ``max_batch_size``
```
joelgrus: if stateful=False, you mean

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
```diff
@@ -29,76 +31,155 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
...
+    to the constructor.  In this case, the ``module`` forward method has a slightly different
+    signature from ``torch.nn.modules.RNNBase``.  It returns:
+
+        - hidden states of size ``(num_layers, batch_size, timesteps, hidden_dim)``
+        - final states, a tuple of sizes ``(num_layers, batch_size, hidden_dim)``
+          and ``(num_layers, batch_size, memory_dim)``
+
     """
-    def __init__(self, module: torch.nn.modules.RNNBase) -> None:
+    def __init__(self, module: Union[torch.nn.modules.RNNBase, torch.nn.Module],
```
joelgrus: I believe `RNNBase` is a subclass of `Module`, so from a practical point of view this is no different from just having `module: torch.nn.Module`

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
```diff
@@ -29,76 +31,155 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
...
+    We also support stacked RNNs that return activations for each layer by passing ``stacked=True``
+    to the constructor.  In this case, the ``module`` forward method has a slightly different
+    signature from ``torch.nn.modules.RNNBase``.  It returns:
+
+        - hidden states of size ``(num_layers, batch_size, timesteps, hidden_dim)``
+        - final states, a tuple of sizes ``(num_layers, batch_size, hidden_dim)``
+          and ``(num_layers, batch_size, memory_dim)``
+
     """
-    def __init__(self, module: torch.nn.modules.RNNBase) -> None:
```
DeNeutoy: The reason this was an `RNNBase` was because they all define `input_size`, `hidden_size`. If we are going to switch to allowing `Modules` we should make it very clear in the docstring (or even raise errors) when they do not define these attributes.

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
```diff
@@ -29,76 +31,155 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
...
+    def __init__(self, module: torch.nn.Module,
+                 stacked: bool = False,
+                 stateful: bool = False, max_batch_size: int = 128) -> None:
```
DeNeutoy: This `max_batch_size` restriction seems arbitrary. Given the states are just reused for random sequences in the case that the batch size would be changing, why not just duplicate the state if you encounter a sequence which has more batch elements than you have seen before?

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
```diff
@@ -29,76 +31,155 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
...
+        if self._stateful and batch_size > self._max_batch_size:
+            raise ValueError("Got batch_size={0} but using a stateful RNN with max_batch_size={1}"
+                             "".format(batch_size, self._max_batch_size))
 
         sequence_lengths = get_lengths_from_binary_sequence_mask(mask)
         sorted_inputs, sorted_sequence_lengths, restoration_indices = sort_batch_by_length(inputs,
                                                                                            sequence_lengths)
-        packed_sequence_input = pack_padded_sequence(sorted_inputs,
-                                                     sorted_sequence_lengths.data.tolist(),
+        packed_sequence_input = pack_padded_sequence(sorted_inputs[:num_valid, :, :],
```
DeNeutoy: This is a good change and actually fixes a bug if you have sequences of length 1. Thanks for doing this..

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
```diff
@@ -29,76 +31,155 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
...
-                                                      unpacked_sequence_tensor.size(-1)).fill_(0)
+            zeros = unpacked_sequence_tensor[0].data.new(batch_size, sequence_length_difference,
+                                                         unpacked_sequence_tensor[0].size(-1)).fill_(0)
             zeros = torch.autograd.Variable(zeros)
-            unpacked_sequence_tensor = torch.cat([unpacked_sequence_tensor, zeros], 1)
+            for k in range(num_layers):
+                unpacked_sequence_tensor[k] = torch.cat([unpacked_sequence_tensor[k], zeros], 1)
 
+        if self._stateful:
+            self._update_states(final_states, num_valid)
```
DeNeutoy: These need to be restored back to their original ordering before being set as the states, because they are the states for the sorted sequences.

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
```diff
@@ -29,76 +31,155 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
...
 
         # Restore the original indices and return the sequence.
-        return unpacked_sequence_tensor.index_select(0, restoration_indices)
+        if not self._stacked:
+            return unpacked_sequence_tensor[0].index_select(0, restoration_indices)
+        else:
+            return torch.cat([tensor.index_select(0, restoration_indices).unsqueeze(0)
+                              for tensor in unpacked_sequence_tensor], dim=0)
+
+    def _get_initial_states(self, num_valid):
```
DeNeutoy: It would be good if these could handle single element hidden states as well, for GRUs.

tests/modules/seq2seq_encoders/pytorch_seq2seq_wrapper_test.py
```diff
@@ -122,3 +122,53 @@ def test_wrapper_raises_if_batch_first_is_false(self):
...
+            batch_size = (k + 1) * 2
+            sequence_length = k + 5
+            tensor = Variable(torch.rand([batch_size, sequence_length, 3]))
+            mask = Variable(torch.ones(batch_size, sequence_length))
+            mask.data[0, 3:] = 0
+            encoder_output = encoder(tensor, mask)
+
+        assert list(encoder_output.size()) == [6, 7, 14]
+
+    def test_wrapper_stateful_too_large_batch(self):
```
DeNeutoy: `test_stateful_wrapper_raises_with_batch_size_greater_than_max`

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
```diff
@@ -29,76 +31,155 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
...
+        if self._states is None:
+            initial_states = None
+        else:
+            # We have some previous states.
+            initial_states = (self._states[0][:, :num_valid, :],
+                              self._states[1][:, :num_valid, :])
+
+        return initial_states
+
+    def _update_states(self, final_states, num_valid):
```
DeNeutoy: Type hints here please.

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
DeNeutoy: `num_valid: int`

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
```diff
@@ -29,76 +31,155 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
...
+            initial_states = hidden_state
+        else:
+            initial_states = self._get_initial_states(num_valid)
 
-        # We sorted by length, so if there are invalid rows that need to be zeroed out
-        # they will be at the end.
+        # Actually call the module on the sorted PackedSequence.
+        packed_sequence_output, final_states = self._module(packed_sequence_input, initial_states)
+        if self._stacked:
+            # packed_sequence_output is shape (n_layers, batch_size, n_times, nx)
```
DeNeutoy: (num_layers, batch_size, timesteps, encoder_dim)

matt-gardner:
I didn't look too closely at the code, just the API.  Adding stateful functionality is great, but adding the `stacked` option breaks the `Seq2SeqEncoder` API.  By definition, a `Seq2SeqEncoder` takes inputs of shape `(batch_size, sequence_length, input_dim)` and returns outputs of shape `(batch_size, sequence_length, output_dim)`.  Returning stacked outputs doesn't fit that definition, and makes it so that `Seq2SeqEncoders` are not interchangeable - you'd have to change your code to switch encoders.

I definitely agree that returning stacked outputs is something you might want to do, we just need to do it outside of a `Seq2SeqEncoder`.  In fact, if you want to do that, you're probably just better off using a pytorch RNN, instead of a `Seq2SeqEncoder`.  That abstraction only exists to allow you to abstract away how you go from `(batch_size, sequence_length, input_dim)` to `(batch_size, sequence_length, output_dim)` - if you're not doing that operation, or don't need to abstract away the _how_, then you don't need to use the abstraction.

matt-gardner:
One thing we could add that _would_ fit the API is to do some operation on the stacked outputs, like a learned linear combination.  As long as the output is the right shape, you can do the combination inside here.

matt-peters:
@matt-gardner your second suggestion won't work for our use case since we want to run the stacked LSTMs once, but downstream use two different linear combinations with different weights of the layers.  It's also too restrictive for future potential use cases.

I could modify the code to fit the seq2seq abstraction in the stacked case by returning `(batch_size, timesteps, embedding_dim * num_layers)` but that requires a lot of clunky packing/unpacking only to undo it later where we need to use each layer separately.

I don't see how to use a pytorch RNN in this case since it doesn't support returning multiple layers, and would really like to reuse all sorting/stateful stuff in the current code.  Any suggestions about how to do that without cutting and pasting a lot of code or massively refactoring are appreciated.

matt-gardner:
What if we have a pytorch RNN wrapper that adds stateful and stacked functionality, and then use _that_ as our input module to the `Seq2SeqWrapper` instead of a base pytorch RNN?  Then you can get the stateful and stacked additions that you've put in here, and we don't break the `Seq2SeqEncoder` API.  Not suggesting that you need to do this work, @matt-peters, just trying to figure out the right way to make the API work out.  @DeNeutoy, what do you think?

matt-peters:
I'd like to get this done and move on so unless there are any objections then I'll pack the stacked case into `(batch_size, timesteps, embedding_dim * num_layers)`.  This shouldn't be hard to implement.

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
matt-peters: Adding `max_batch_size` is for code simplicity.  I don't think it's too restrictive, and we could remove it at a later point if necessary.

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
matt-peters: It's already clear in the existing docstring that the module must have this interface.  If someone implements a module that doesn't have this interface then an exception will be raised when these attributes are accessed.

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
matt-peters: I'll do that in an upcoming commit.

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
matt-peters: Good point, I'll address that and add a test.

DeNeutoy:
I think the reason we are running into this problem is because `PytorchSeq2SeqWrapper` and `PytorchSeq2VecWrapper` implement a lot of the same stuff in forward surrounding reordering the batch, which should maybe be separate. What about something like this:

```
def sorted_forwards(rnn_function, sequence, lengths, state):
    # Sort things
    # Run RNN
    # deal with zero padding
    returns output of forwards, along with restoration indices for the batch dimension.

class Stateful:
def get_state()
def set_state()

Seq2SeqEncoder(torch.nn.Module, Stateful, Registerable):
Seq2StackEncoder(torch.nn.Module, Stateful, Registerable):
Seq2VecEncoder(torch.nn.Module, Registerable):

class PytorchSeq2SeqWrapper(Seq2SeqWrapper):

# implements forward, which calls sorted_forward, gets and sets state if needed.

class PytorchSeq2VecWrapper(Seq2VecWrapper):

# implements forward which calls sorted_forward and returns the state only.

class PytorchSeq2StackWrapper(Seq2StackWrapper):

# implements forward, which calls sorted_forward, gets and sets state if needed, returns the stacked layers.
```

matt-peters:
I personally think multiple inheritance makes for hard to follow code and should be avoided.  Also this seems like a lot of refactoring.  In the interim I'll work on addressing the other comments in the PR and leave the stacked piece alone for the time being.

matt-gardner:
@DeNeutoy, if you make `sorted_forwards` a module, and add a ` stateful` flag to it, I think your suggestion and my suggestion are essentially equivalent, except mine doesn't use multiple inheritance.  Yes?

DeNeutoy:
Yeah, maybe it is better to just stack the outputs. It's simple, it fits the API and we can just not add it to the registered list of `Seq2SeqEncoders`, which will mean people don't accidentally use it.

DeNeutoy:
Yes, seems the same.

matt-peters:
Addressed all the comments except the open stacked discussion, pending a final decision how you want to handle it.

tests/modules/seq2seq_encoders/pytorch_seq2seq_wrapper_test.py
```diff
@@ -122,3 +122,66 @@ def test_wrapper_raises_if_batch_first_is_false(self):
         with pytest.raises(ConfigurationError):
             lstm = LSTM(bidirectional=True, num_layers=3, input_size=3, hidden_size=7)
             _ = PytorchSeq2SeqWrapper(lstm)
+
```
DeNeutoy: Can you change these tests to check that the state is carried through for different sized batches and just generally change them to test equivalence of actual tensors rather than shapes.

matt-peters:
@DeNeutoy added some tests as suggested and removed `max_batch_size`

matt-peters:
Replaced by #518

PR #510, opened 2017-11-17
fix masking for length 1 sequences and add tests

DeNeutoy:
@lucylw discovered a bug in the `PytorchSeq2XXXWrappers` when you passed length 1 sequences. This happened because we added 1 to the mask before we extracted the lengths, meaning that if you legitimately had some length one sequences, the sorting of these and the invalid rows would be arbitrary. This meant that when we zeroed out the result, sometimes valid rows were zeroed out because of the way the batch had been sorted. 

This alters the sequence lengths of the batch after they have been sorted, stoping this from happening.


DeNeutoy:
Closing this as it will be addressed by #509.

Issue #511, opened 2017-11-17
Inconsistent case in `predict` command


kmkurn:
Hi, this is a really minor issue. The `predict` command accepts `--output-file` optional argument, written in kebab-case. This is inconsistent with the other optional arguments which use snake_case (e.g. `--batch_size` or `--cuda_device`). Also, the metavar for `input-file` argument is also in kebab-case while others are in snake_case.

joelgrus:
yes, this bugs me too, let me think about whether there's a way to fix this that doesn't break backward compatibility

joelgrus:
closed via https://github.com/allenai/allennlp/pull/512

thanks for the suggestion

PR #512, opened 2017-11-17
update CLI casing

joelgrus:
this addresses https://github.com/allenai/allennlp/issues/511

it does the following things (I will use cuda_device as an example, but it does something similar for every multi-word argument)

(1) creates a `cuda_device` "mutually exclusive group"
(2) changes the existing argument to `cuda-device` under that group
(3) also includes a `cuda_device` argument under that group with `help=argparse.SUPPRESS`

this means that if you run (say) `allennlp/run predict --help` it only shows the option

```
  --cuda-device CUDA_DEVICE
                        id of GPU to use (if any)
```

however, if you pass in the argument as `--cuda_device` it will still work (in order not to break anyone's existing code). This all feels slightly clunky, but breaking existing workflows seems like a non starter.

IN ADDITION, there is a dict of { old_flag: new_flag }, and if you call `allennlp/run` using one of the old flags it will log a warning that it's deprecated and that you should switch. And then at some point in the far future we can clean up all the underscore flags, maybe.

--

Anyway, this might seem like a lot of ado about not much, but I think having a consistent API is important (this issue has always sort of bugged me, just never enough to do anything about it), and I think `kebab-case` is the right way to handle command line flags.

Like the title says, at this point this is just for discussion, I'm curious how people feel about it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,9 +47,6 @@ COPY scripts/ scripts/
 COPY tutorials/ tutorials/
 COPY training_config training_config/
 
-# Run tests to verify the Docker build
-RUN PYTHONDONTWRITEBYTECODE=1 pytest
-
 # Add model caching
 ARG CACHE_MODELS=false
 RUN ./scripts/cache_models.py
--- a/allennlp/commands/__init__.py
+++ b/allennlp/commands/__init__.py
@@ -1,5 +1,7 @@
 from typing import Dict
 import argparse
+import logging
+import sys
 
 from allennlp.commands.serve import Serve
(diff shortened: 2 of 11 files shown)


matt-gardner:
`argparse` is smart enough to convert the kebab-case to snake-case for actually accessing the arguments in code?  Seems like a good solution to the problem to me.

schmmd:
Seems like a fantastic solution.  Can you update http://allennlp.org/models after this merges?

joelgrus:
ok, I added some tests that make me comfortable with this, I guess it's ready for review

PR #517, opened 2017-11-20
Fix bug in production deployment script

schmmd:
Also:

* Increase memory to 3 GB.
* Decrease production replicas to 4.


joelgrus:
did you want to do something with making the secrets optional here ?

schmmd:
I like that they are required.

PR #518, opened 2017-11-21
Encoder abstractions

DeNeutoy:
- Adds a `Seq2StackEncoder` abstraction
- Pulls out the sorting of the batch, forward pass through the model and statefulness for all encoders into a `_EncoderBase` class. 
- Refactors `Seq2SeqEncoders` and `Seq2StackEncoders` to allow them to be stateful.

Depends on #509


DeNeutoy:
I'm leaving this till next week for now, because large parts of it are wrong and I think we might want to re-think how to incorporate statefulness. I think fixing the current way would result in very complicated code for an extremely narrow and under-used use case.

matt-gardner:
Do we need the `Seq2Stack` abstraction, or just a concrete class that does what MattP needs?

DeNeutoy:
I was doing this in response to the feedback on the previous PR, but i'm not particularly happy with how it's turned out, so maybe you're right.

I think the original issue is that if we _aren't_ going to write a `Seq2StackWrapper`, we'll have to duplicate the sorting and packing logic inside the `ElmoLstm`. The problem with writing said wrapper is that really, the class which deals with the sorting and packing really needs to deal with the statefulness, because the states need to be un-sorted and aligned before they are stored, which would make it necessary to also duplicate the statefulness logic into that class as well. Maybe this is a better way forward, because honestly, our library isn't set up to do stateful training for the other encoders, and it's not really like statefulness is something you can "swap in" to your model - either it is stateful or it isn't, depending on your data, making the ELMo stuff a special case. 

Either way, I think the statefulness is going to be very slow, as with the changes in this PR, we're now doing 4 `index_select`'s per RNN call.

DeNeutoy:
Although I guess we could keep this `_EncoderBase` and just make `ElmoLstm` inherit from this. Not sure ...

matt-gardner:
Yeah, I was imagining keeping the `_EncoderBase` (or something like it), and just having the `Seq2*` wrapper classes use that.  And because the `ElmoLstm` isn't going to fit as a `Seq2SeqEncoder`, it can just be a standalone class that uses the encoder base.  If you think there's a better option, though, that's fine with me.  I'm happy to chat about this if you think it'd help nail down the API decision here.

DeNeutoy:
Ok, i'll remove that this morning and see what it looks like.

DeNeutoy:
Should I add docs for the `_EncoderBase` or just add it to the ignored script? It's defo not part of the public API

allennlp/nn/util.py
```diff
@@ -72,7 +76,7 @@ def sort_batch_by_length(tensor: torch.autograd.Variable, sequence_lengths: torc
     index_range = Variable(index_range.long())
     _, reverse_mapping = permutation_index.sort(0, descending=False)
     restoration_indices = index_range.index_select(0, reverse_mapping)
-    return sorted_tensor, sorted_sequence_lengths, restoration_indices
+    return sorted_tensor, sorted_sequence_lengths, restoration_indices, permutation_index
```
matt-peters: Just a general comment here about the interface, not necessarily a suggestion to change it here:  this is the reason I typically return dicts in situations like this instead of position aware tuples, as it makes modifying the interface much easier.  In this case, adding an additional output required changing all of the places in the code that also called `sort_batch_by_length`.

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
```diff
@@ -29,76 +29,93 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
     Note that we *require* you to pass sequence lengths when you call this module, to avoid subtle
```
matt-gardner: Just noticed that this is out-dated - you don't pass sequence lengths, you pass a mask.

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
```diff
@@ -29,76 +29,93 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
...
     bugs around masking.  If you already have a ``PackedSequence`` you can pass ``None`` as the
     second parameter.
+
+    We support stateful RNNs where the final state from each batch is used as the initial
+    state for the subsequent batch by passing ``stateful=True`` to the constructor.
+
+        - hidden states of size ``(num_layers, batch_size, timesteps, hidden_dim)``
```
matt-gardner: Not sure what these two lines are doing here.

allennlp/modules/encoder_base.py
```diff
@@ -0,0 +1,268 @@
...
+        -------
+        This method has a complex return type because it has to deal with the first time it
+        is called, when it has no state, and the fact that types of RNN have heterogeneous
+        states.
+
+        If it is the first time the module has been called, it returns ``None``, regardless
+        of the type of the ``Module``.
+
+        Otherwise, for LSTMs, it returns a tuple of ``torch.Tensors`` with shape
+        ``(num_layers, num_valid, state_size)`` and ``(num_layers, num_valid, memory_size)``
```
matt-gardner: An explanation for these shapes is in order, I think - why is `num_layers` first?  This is particularly important because you can pass in an initial hidden state to the `forward` method.  I suppose adding this explanation above, where the docs for the hidden state were missing, would be enough.

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
```diff
@@ -29,76 +29,93 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
...
+        # Some RNNs (GRUs) only return one state as a Tensor.  Others (LSTMs) return two.
+        # If one state, use a single element list to handle in a consistent manner below.
+        if not isinstance(final_states, (list, tuple)) and self._stateful:
+            final_states = [final_states]
+
+        # Add back invalid rows.
         if num_valid < batch_size:
-            unpacked_sequence_tensor[num_valid:, :, :] = 0.
+            _, length, output_dim = unpacked_sequence_tensor.size()
+            zeros = unpacked_sequence_tensor.data.new(batch_size - num_valid, length, output_dim).fill_(0)
```
matt-gardner: You do this operation enough that I'd probably just add a method to `util` for it.  Something like `util.add_rows(tensor, dim, num_rows, fill_value)`.

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
```diff
@@ -29,76 +29,93 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
...
+                final_states = new_states
 
         # It's possible to need to pass sequences which are padded to longer than the
         # max length of the sequence to a Seq2SeqEncoder. However, packing and unpacking
         # the sequences mean that the returned tensor won't include these dimensions, because
         # the RNN did not need to process them. We add them back on in the form of zeros here.
         sequence_length_difference = total_sequence_length - unpacked_sequence_tensor.size(1)
         if sequence_length_difference > 0:
-            zeros = unpacked_sequence_tensor.data.new(batch_size, sequence_length_difference,
+            zeros = unpacked_sequence_tensor.data.new(batch_size,
```
matt-gardner: It's the same `add_rows` operation here, just on a different dimension.  And you're making `zeros` a `Variable` here, but not above - I assume `torch.cat` returns a `Variable` if any of its inputs is a `Variable`?

allennlp/modules/seq2vec_encoders/pytorch_seq2vec_wrapper.py
```diff
@@ -68,48 +67,25 @@ def forward(self,  # pylint: disable=arguments-differ
...
-        # We sorted by length, so if there are invalid rows that need to be zeroed out
-        # they will be at the end. Note that at this point batch_size is the second
-        # dimension of state.
+        # Add back invalid rows.
         if num_valid < batch_size:
-            state[:, num_valid:, :] = 0.
+            # batch size is the second dimension here, because pytorch
+            # returns RNN state as a tensor of shape (num_layers * num_directions,
+            # batch_size, hidden_size)
+            num_layers_times_directions, _, encoding_dim = state.size()
```
matt-gardner: You can replace `_` with `num_valid` here, so you don't have to return it from `sort_and_run_forward`.

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -7,16 +7,22 @@
...
     A stacked, bidirectional LSTM which uses
     :class:`~allennlp.modules.lstm_cell_with_projection.LstmCellWithProjection`'s
     with highway layers between the inputs to layers.
     The inputs to the forward and backward directions are independent - forward and backward
     states are not concatenated between layers.
 
+    Additionally, this LSTM maintains its `own` state, which is updated every time
+    ``forward`` is called. It is dynamically resized for different batch sizes and is
+    designed for use with non-continuous inputs. This is non-standard, but can be thought
+    of as having a "end of sentence" state, which is carried across different sentences.
```
matt-gardner: `s/having a/having an/`

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -7,16 +7,22 @@
...
     """
     A stacked, bidirectional LSTM which uses
     :class:`~allennlp.modules.lstm_cell_with_projection.LstmCellWithProjection`'s
     with highway layers between the inputs to layers.
     The inputs to the forward and backward directions are independent - forward and backward
     states are not concatenated between layers.
 
+    Additionally, this LSTM maintains its `own` state, which is updated every time
+    ``forward`` is called. It is dynamically resized for different batch sizes and is
+    designed for use with non-continuous inputs. This is non-standard, but can be thought
```
matt-gardner: I don't know what "non-continuous inputs" means.

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -83,16 +89,78 @@ def __init__(self,
...
+        A ``torch.Tensor`` of shape (num_layers, batch_size, sequence_length, hidden_size),
+        where the num_layers dimension represents the LSTM output from that layer.
+        """
+        batch_size, total_sequence_length = mask.size()
+        stacked_sequence_output, final_states, restoration_indices, num_valid = \
+            self.sort_and_run_forward(self.lstm_forward, inputs, mask)
+
+        # stacked_sequence_output is shape (num_layers, batch_size, timesteps, encoder_dim)
+        num_layers = stacked_sequence_output.size(0)
+        per_layer_sequence_outputs = [layer.squeeze(0) for layer in
```
matt-gardner: Is this more efficient than `[layer[i] for i in num_layers]`?  If it is, I need to update my decoding code...

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -83,16 +89,78 @@ def __init__(self,
...
+                                                           per_layer_sequence_outputs[0].size(-1)).fill_(0)
+            zeros = torch.autograd.Variable(zeros)
+            for k in range(num_layers):
+                per_layer_sequence_outputs[k] = torch.cat([per_layer_sequence_outputs[k], zeros], 1)
+
+        self._update_states(final_states, restoration_indices)
+
+        # Restore the original indices and return the sequence.
+        # Has shape (num_layers, batch_size, sequence_length, hidden_size)
+        return torch.cat([tensor.index_select(0, restoration_indices).unsqueeze(0)
```
matt-gardner: You can just use `torch.stack([t])` instead of `torch.cat([t.unsqueeze(0)])`.

allennlp/modules/stacked_elmo_lstm.py
matt-gardner: But why are you doing this at all?  Why can't you just operate on the 4D tensor?  Is it because of the `index_select` at the end?  If that's the case, you can still do all of the padding stuff on the 4D tensor, which will result in fewer tensor operations, and just do a chunk, index select, and stack at the end (or maybe there's one of the fancy `batched_index_select` methods from the coref PR that would do this without needing the chunk and stack).

allennlp/modules/stacked_elmo_lstm.py
```diff
@@ -83,16 +89,78 @@ def __init__(self,
...
+                          for tensor in per_layer_sequence_outputs], dim=0)
+
+    def lstm_forward(self,
```
matt-gardner: Should this be private?

allennlp/nn/util.py
```diff
@@ -55,6 +56,9 @@ def sort_batch_by_length(tensor: torch.autograd.Variable, sequence_lengths: torc
     restoration_indices : Variable(torch.LongTensor)
         Indices into the sorted_tensor such that
         ``sorted_tensor.index_select(0, restoration_indices) == original_tensor``
+    permuation_index : Variable(torch.LongTensor)
```
matt-gardner: Shouldn't this be `indices`, to match `restoration_indices`?  Also, a comment like the `sorted_tensor.index_select(...` comment above would be nice here.

allennlp/nn/util.py
matt-gardner: I would tend to agree with this.  Four return values is getting to the point where you probably shouldn't be returning a tuple; something could be made simpler here.  But I don't much like a dict here, either, and making that change would be obnoxious as part of this PR.

tests/modules/encoder_base_test.py
```diff
@@ -0,0 +1,161 @@
...
+class TestEncoderBase(AllenNlpTestCase):
+
+    def setUp(self):
+        super(TestEncoderBase, self).setUp()
+        self.lstm = LSTM(bidirectional=True, num_layers=3, input_size=3, hidden_size=7, batch_first=True)
+        self.encoder_base = _EncoderBase(stateful=True)
+
+        tensor = Variable(torch.rand([5, 7, 3]))
+        tensor[1, 6:, :] = 0
+        tensor[2, :, :] = 0  # <= completely masked
```
matt-gardner: Related to some of my comments above - if you were guaranteed that all of these would be at the end, would it make it possible to simplify the handling of the `num_valid` stuff?

tests/modules/encoder_base_test.py
```diff
@@ -0,0 +1,161 @@
...
+        tensor[3, 2:, :] = 0
+        tensor[4, :, :] = 0  # <= completely masked
+        self.tensor = tensor
```
matt-gardner: Do you need to zero out the masked values in the tensor?  I thought you just needed to set the mask to zeros.

tests/modules/encoder_base_test.py
```diff
@@ -0,0 +1,161 @@
...
+        mask = Variable(torch.ones(5, 7))
+        mask[1, 6:] = 0
+        mask[2, :] = 0
+        mask[3, 2:] = 0
+        mask[4, :] = 0
+        self.mask = mask
+
+        self.batch_size = 5
+        self.num_valid = 4
```
matt-gardner: Why is `num_valid` 4?  I thought two of the instances were completely masked, which should mean `num_valid` is 3.

tests/modules/encoder_base_test.py
```diff
@@ -0,0 +1,161 @@
...
+        numpy.testing.assert_array_equal(self.encoder_base._states[0].data.numpy(),
+                                         correct_expanded_state.numpy())
+        numpy.testing.assert_array_equal(self.encoder_base._states[1].data.numpy(),
+                                         correct_expanded_state.numpy())
+
+        # The returned states should be of shape (1, num_valid, hidden_size) and they also should
+        # have been sorted with respect to the indices.
+        # sorting indices are: [0, 1, 3, 2, 4]
+        correct_returned_state = torch.ones([1, 4, 7])
+        correct_returned_state[:, 2, :] = 0
```
matt-gardner: ????

allennlp/modules/encoder_base.py
```diff
@@ -0,0 +1,268 @@
...
+        # Otherwise, we have some previous states.
+        if batch_size > self._states[0].size(1):
+            # This batch is larger than the all previous states.
+            # If so, resize the states.
+            num_states_to_concat = batch_size - self._states[0].size(1)
+            resized_states = []
+            # state has shape (num_layers, batch_size, hidden_size)
+            for state in self._states:
+                # This _must_ be inside the loop. because some
+                # RNNs have states with different last dimension sizes.
```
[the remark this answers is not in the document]
DeNeutoy: The ELMo LSTM has different memory and state sizes.

matt-peters:
@matt-gardner thanks for the review.  Two thoughts:

* I wouldn't optimize the code until after benchmarking and confirming it's a bottleneck.  These operations should be cheap in the context of all the other computation needed to compute ELMo.

* All of the complex stateful stuff isn't needed for ELMo anyway, the single use case we have for stateful now.  If all the sorting/`index_selects` in the stateful piece turns out to be a bottleneck the easiest approach is to just not run it for ELMo... so take that into consideration when deciding how much time to spend on it.

tests/modules/encoder_base_test.py
matt-peters: No, you shouldn't need to zero out the tensor.

tests/modules/encoder_base_test.py
matt-peters: index 2 is completely masked, so the final states are the initial states == 0

allennlp/modules/encoder_base.py
```diff
@@ -0,0 +1,268 @@
...
+                zeros = state.data.new(state.size(0),
+                                       num_states_to_concat,
+                                       state.size(2)).fill_(0)
```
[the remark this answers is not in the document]
matt-peters: Not until there is a need to and initial states == 0 is standard practice.

allennlp/modules/encoder_base.py
```diff
@@ -0,0 +1,268 @@
...
+            # Now we've sorted the states back so that they correspond to the original
+            # indices, we need to figure out what states we need to update, because if
+            # didn't use a state for a particular row, we want to preserve it's state.
+            # Thankfully, the rows which are all zero in the state correspond exactly
+            # to those which aren't used, so we create masks of shape (new_batch_size,),
+            # denoting which states were used in the RNN computation.
+            current_state_batch_size = self._states[0].size(1)
+            new_state_batch_size = final_states[0].size(1)
+            # Masks for the unused states of shape (1, new_batch_size, 1)
+            used_new_rows_mask = [(state[0, :, :].sum(-1) != 0.0).float().view(1, -1, 1)
```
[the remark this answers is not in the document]
DeNeutoy: Why is this expensive?

allennlp/modules/encoder_base.py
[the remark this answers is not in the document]
matt-gardner: Ok, fair enough.  Looks like there's also a period that should be a comma (or just omitted) on line 161.

tests/modules/encoder_base_test.py
matt-gardner: But this is supposed to only have `num_valid` returned states, and so index 2 shouldn't even be in here...  I'm really confused about how `num_valid` is actually computed, because it doesn't look like it matches what the docs say above.

allennlp/modules/encoder_base.py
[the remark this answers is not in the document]
matt-gardner: Because it's a GPU operation, and all operations have overhead.  It's not that this particular operation is slow (like `.data[0]` is), it just seemed like it was unnecessary.  I then realized that it probably isn't.

matt-gardner:
I agree with not doing premature optimization.  But if we know that some operations are inordinately slow (like copying from the GPU to the CPU with `.data[0]`), and it's easy to avoid them now, we should probably do that.  The stuff with using `num_valid` is a suggestion that _should_, if it's even possible, make the code both simpler and faster.  But, as I said, I would have just approved the PR and let you decide (1) if it was possible to change, and (2) if it's worth changing now, except for the test that seems wrong to me.

tests/modules/encoder_base_test.py
DeNeutoy: Sorry, this was confusing and only passing because I was using `ones` for the returned state - I have switched it to using random states like the other test (which was passing because it doesn't use `num_valid` at all). My bad.

allennlp/modules/encoder_base.py
```diff
@@ -0,0 +1,268 @@
...
+        batch_size, _ = mask.size()
+        num_valid = torch.sum(mask[:, 0]).int().data[0]
+
+        sequence_lengths = get_lengths_from_binary_sequence_mask(mask)
+        sorted_inputs, sorted_sequence_lengths, restoration_indices, sorting_indices =\
+            sort_batch_by_length(inputs, sequence_lengths)
+
+        # Now create a PackedSequence with only the non-empty, sorted sequences.
+        packed_sequence_input = pack_padded_sequence(sorted_inputs[:num_valid, :, :],
+                                                     sorted_sequence_lengths[:num_valid].data.tolist(),
```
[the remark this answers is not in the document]
DeNeutoy: I thought about this a bit and I can't think of a good solution.

allennlp/modules/encoder_base.py
DeNeutoy: This is how all states are returned in pytorch

allennlp/modules/encoder_base.py
```diff
@@ -0,0 +1,268 @@
...
+        """
+        # TODO(Mark): seems weird to sort here, but append zeros in the subclasses.
+        # which way around is best?
+        new_sorted_states = [state.index_select(1, restoration_indices)
+                             for state in final_states]
+
+        if self._states is None:
+            # We don't already have states, so just set the
+            # ones we receive to be the current state.
+            self._states = tuple([torch.autograd.Variable(state.data)
```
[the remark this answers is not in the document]
DeNeutoy: I'm actually not sure if this is correct, because `Variables` have a `.detach()` method and I don't know if this is the same. Any idea?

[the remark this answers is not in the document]
matt-gardner: Well, assuming I was right about the motivation for doing this, it looks like the `.detach()` method should be enough.

tests/modules/encoder_base_test.py
DeNeutoy: Do you mean for the test, or for the module? We should have this not at the end in the test to check that the sorting is happening correctly. We can't take only sorted states in the module, because if we did that it would be very difficult to keep track of the state, as it would have to be stored in a sorted format, which seems unpleasant and very error prone.

matt-gardner: I meant the padding, not the sorting.  If you're guaranteed that all invalid instances are at the end, can you simplify the logic so that `sort_batch_by_length` can return something of size `(num_valid, ...)` instead of `(batch_size, ...)`?  Not a super important question.

DeNeutoy: I removed it, the test still passes 👍

PR #521, opened 2017-11-21
Create ISSUES.md

schmmd:


schmmd:
Is this root-level file worthy?  I think so, as it describes what the issues tags we use mean.

joelgrus:
I think it's weird to describe the details of our on-call rotation in our public repo.

schmmd:
Sure, I'll handle it another way.

PR #522, opened 2017-11-21
Adding a proposed style guide

matt-gardner:
These are the rules we've generally been following, I'm just putting them into a document that we can reference when questions come up.  Does this seem reasonable?  Anything to change or add?

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -1,9 +1,9 @@
 # AllenNLP Style Guide
 
-Our highest priority in writing code (other than correctness) is that our code be easily readable
-to someone new to the codebase.  Deep learning is easy to get wrong, and we want our code to be
-easy enough to read that someone looking at it can be thinking about our modeling decisions, not
-trying to understand what is going on.
+Our highest priority for code style is that our code be easily readable to someone new to the
+codebase.  Deep learning is easy to get wrong, and we want our code to be easy enough to read that
+someone looking at it can be thinking about our modeling decisions, not trying to understand what
+is going on.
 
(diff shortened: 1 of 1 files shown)


STYLE.md
```diff
@@ -0,0 +1,57 @@
+# AllenNLP Style Guide
+
+Our highest priority in writing code (other than correctness) is that our code be easily readable
```
schmmd: I would reword this to "our highest priority for code style is that..."

joelgrus:
I'm sure I'll think of things to add, but none come immediately to mind.

schmmd:
The guide looks reasonable, but I would call out doc strings as to me it's more important to your goals than the other items.  In general, I prefer to be strict about style guidelines through automated tooling and not by human enforcement.  And I really prefer to avoid style debates ;-)

matt-gardner:
@schmmd, what do you mean by "call out doc strings"?  And yes, I'm happy that pylint does most of the enforcement for us.  Some of this document is just explaining what pylint is doing and why.  If there are ways to get more of this into automated checking, I'd be happy with that.

schmmd:
@matt-gardner I mean specifically call out our expectations around doc strings (not sure what they are called in Python) for classes and methods.  In other words, describe the level of documentation we expect for our code.  I see this as an aspect of code style.

matt-gardner:
@schmmd, the second paragraph does this.  I don't have anything more specific than what's there - it depends a lot on what you're documenting - so I had a hard time trying to modify it to incorporate your suggestion.

PR #526, opened 2017-11-22
Add `hierplane` to the SRL output as a vizualization method.

codeviking:
* Adds a `hierplane` visualization to the SRL output, where each
  verb is rendered as it's own tree.
* Add UI controls that allow the user to toggle between the "table" and
  "tree" view.
* Ignore `node_modules`, as it's a generated folder.

@schmmd Please review, or determine who is appropriate to review.


codeviking:
An obligatory screenshot:

image

demo/src/components/SrlComponent.js
```diff
@@ -34,6 +34,77 @@ const description = (
...
+        nodeType: 'V',
+        attributes: [ 'VERB' ],
+        spans: [{
+          start,
+          end: start + verb.length + 1,
+        }, ...ignoredSpans],
+        children: tags.reduce((children, tag, idx) => {
+          if (tag !== 'B-V' && tag.startsWith('B-')) {
+            const word = response.words[idx];
+            const link = tag.split('-').pop().replace(/\d+/g, '');
```
DeNeutoy: The numeric characters in the arguments actually carry some meaning for SRL - can you leave them in?

demo/src/components/SrlComponent.js
```diff
@@ -34,6 +34,77 @@ const description = (
...
+            const start = getStrIndex(response.words, idx);
+            const newChild = {
+              word,
+              spans: [{
+                start,
+                end: start + word.length + 1
```
DeNeutoy: What would be really nice is if we could make the `nodeType` and `attributes` different.
The SRL labels either contain `ARGM`, which modify the verb, or `ARG1-5` which are the core arguments. For the `ARGM`, the nodeType could be `modifier` and for the `ARG1-5` it could be `arg`, with the `attribute` containing the second half of the label, eg `TMP` or `DIS`.

One final point to be aware of is that some of the labels might contain `C-ARG1` or `R-ARG-1`, which mean that the label is continued or referred to, eg `(The place ARG1), beautifully though it was, (where R-ARG1) Mark lived`.

schmmd:
Hi Sam, looks fantastic.  A few thoughts:

* Why is "which" highlighted in the source sentence?  Did the user click on it in the diagram?
* I will probably make a subsequent PR that puts the diagram first, and changes the table view to a text output for SRL.

demo/src/components/SrlComponent.js
codeviking: Yup, for sure!

demo/src/components/SrlComponent.js
codeviking: Makes a ton of sense! I can easily make this happen.  I'll make the change and get a local demo up today so you can review it / play with it.

codeviking:
> Why is "which" highlighted in the source sentence? Did the user click on it in the diagram?

I was hovering it (I wanted to show that case).

> I will probably make a subsequent PR that puts the diagram first, and changes the table view to a text output for SRL.

Sounds good.  I've got some things to tweak, do you want me to make the default switch as part of that so you don't have to?

schmmd:
Sure, you're welcome to switch the tab order.

demo/src/components/SrlComponent.js
codeviking: Is the `PRP` suffix of value, or something I can disregard for this purpose?

codeviking: Ah, `PRP` is likely the "second half" -- I think I grok things now.

Would `R-` and `C-` both be appropriately labeled as a `reference`, or are they different types?

DeNeutoy: R would be `reference` and C would be `continuation`, they are very rarely observed but can crop up so we should handle them.

PRP is actually the label, it is short hand for `purpose`, i.e why the verb is being done. I'm not exactly sure what you mean by second half, feel free to swing by and we can chat in person.

codeviking: 🤘  Thx.  By the "second half" I mean what you mentioned here:

> with the attribute containing the second half of the label, eg TMP or DIS.

So I'm assuming `TMP`, `DIS` and `PRP` all serve similar purposes (pun intended!) :-D.  LMK if that's not the case.  Also I have an interactive version I can link to shortly.

codeviking:
Alright @schmmd and @DeNeutoy I've incorporated your feedback.  Things are looking pretty good.  Can one of you guys take another pass and green-stamp it if it looks good?

Also, once it's merged, can one of you help me figure out how to deploy it?  Or is CD in place such that it'll automatically go live?

schmmd:
@codeviking looks great.

Deploys are broken presently.  I'm running the AllenNLP demo from a single AWS instance ever since Kubernetes went down.  I will use TeamCity (http://allennlp-build.dev.ai2:8111/) to deploy it to GKE.  We should be able to switch our DNS entries to GKE after the break.

codeviking:
Cool, I can't merge until it's approved by someone with write access:

image

So if it looks good to go, that green stamp will let me throw this into `master`!

PR #527, opened 2017-11-22
Demo requirements

### LINKED ISSUE #504: split out "demo" requirements
Almost no one using AllenNLP needs to run the demo with permalinks enabled, which means that almost no one using AllenNLP needs to install (for example) `psycopg2`. 

We should split out demo requirements the same way we split out test requirements.

DeNeutoy:
- Splits out the demo requirements
- Adds a new `INSTALL_DEMO_REQUIREMENTS` variable to the `install_requirements.sh` script.
- Removes the demo requirements from `setup.py` so library users don't have to install them.

fixes #504


schmmd:
What is the motivation behind this?  I saw something in Slack the other day, but it'd be good to repeat it here or point to a GitHub issue so we keep the context.

Dockerfile
```diff
@@ -32,8 +32,9 @@ RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - && apt-get install -y
 # We only copy what we need here so small changes to the repository does not trigger re-installation of the requirements.
 COPY requirements.txt .
 COPY requirements_test.txt .
+COPY requirments_demo.txt .
```
schmmd: This typo broke the Docker build (missing e).

DeNeutoy:
Basically many users won't be running our demo locally, and it requires some moderately unstable packages like the database package and sanic. It's better if users don't encounter these if they don't need them.

joelgrus:
I would rather leave sanic in, we have "run the demo server locally" as part of the tutorial

DeNeutoy:
It's not really either or - you can't run the server without `psycopg2` because it's imported in `server_sanic.py`. I can just close this PR if it's not needed.

joelgrus:
we should probably do something like 

```
try:
  import psycopg2
  USE_DATABASE = True
except ModuleNotFoundError:
  USE_DATABASE = False
```

I bet pylint hates that, but oh well

DeNeutoy:
Seems a bit of a misleading thing to do to have a `demo_requirements.txt` if we aren't even going to actually split the demo requirements completely into it ....

schmmd:
Are there more stable packages we can use instead of sanic (flask) and psycopg2?  The goal is for people to use the demo more for building their own models, so I'd hate to put unnecessary barriers to running the demo.

joelgrus:
psycopg is totally stable (it's what everyone uses for postgres on python) it just appears to have some platform-dependent installation quirks.

flask is obviously much more stable than sanic, if we want to switch, that's probably a day's worth of work.

schmmd:
It'd be helpful to have a few crisp example of users who have run into problems.  Otherwise, it's hard to justify the cost of this PR or changing our libraries.

schmmd:
Closing for now.  Feel free to re-open if you'd like to re-open the conversation.

PR #528, opened 2017-11-22
Add info route.

schmmd:
This adds a simple info route to troubleshoot what code is actually running.  To work, this depends on the `SOURCE_COMMIT` environment variable being set.  This is automatically set as part of our build process so it will be present in our Docker images.


allennlp/service/server_sanic.py
```diff
@@ -196,6 +198,15 @@ def _caching_prediction(model: Predictor, data: str) -> JsonDict:
...
         return response.json({"models": list(app.predictors.keys())})
 
+    @app.route('/info')
+    async def info(req: request.Request) -> response.HTTPResponse:  # pylint: disable=unused-argument, unused-variable
+        """List metadata about the running webserver"""
+        git_version = os.environ.get('SOURCE_COMMIT') or ""
+        return response.json({
+                "start_time": start_time,
+                "git_version": git_version,
+                "githubUrl": "http://github.com/allenai/scholar/commit/" + git_version})
```
joelgrus: I doubt this is the right github url

PR #529, opened 2017-11-22
fix Makefile to work in docker

joelgrus:
you learn new stuff about make and bash and docker every day


joelgrus:
bad me, force merge

PR #530, opened 2017-11-22
Dataset for vocab selection

### LINKED ISSUE #516: Add a flag which avoids using the validation data for creating a vocabulary.
Hi, I was looking at the code for `train` command and noticed that the vocabulary is built from all datasets (train, dev, test). Is this intentional? I am asking because I thought the common thing to do is to build vocabulary from train set only so that words in dev/test set that does not occur in train set are converted into `<unk>` and given a single embedding. Building vocabulary from all three sets means that there will be no unknown words in dev/test set (assuming `min_count=1`).

DeNeutoy:
Adds a key to allow a user to specify which datasets are used to build the vocabulary. Fixes #516.


DeNeutoy:
@joelgrus ¯\_(ツ)_/¯

allennlp/commands/train.py
```diff
@@ -126,31 +127,39 @@ def train_model(params: Params, serialization_dir: str) -> Model:
...
-    logger.info("Creating a vocabulary using %s data.", ", ".join(datasets_in_vocab))
+    # powerset (all permutations of all lengths) of the 3 possible options
+    # with the one containing the largest allowed set of keys given the datasets
+    # that were passed in the first position, which we use by default.
+    allowed_combinations = itertools.chain.from_iterable(itertools.permutations(all_datasets.keys(), i)
+                                                         for i in [3, 2, 1])
+
+    datasets_for_vocab_creation = params.pop_choice("datasets_for_vocab_creation",
+                                                    [list(x) for x in allowed_combinations],
+                                                    default_to_first_choice=True)
```
joelgrus: on one hand this is clever, on the other hand it's almost too clever. why not just

```
datasets_for_vocab_creation = set(params.pop("datasets_for_vocab_creation", all_datasets))

for dataset in datasets_for_vocab_creation:
    if dataset not in all_datasets:
        raise ConfigurationError(f"invalid 'dataset_for_vocab_creation' {dataset}")
```

DeNeutoy: If you haven't specified a test or evaluation dataset, it isn't a valid choice for a dataset to build the vocabulary from, though.

joelgrus: doesn't my suggestion raise in that case?

DeNeutoy: oh right, sorry. I'll change it

PR #531, opened 2017-11-22
Disable Service Workers

schmmd:
The bootstrapped ReactJS application enabled service workers, which is something we don't want.  From [the web](https://developers.google.com/web/fundamentals/primers/service-workers/): "A service worker is a script that your browser runs in the background, separate from a web page, opening the door to features that don't need a web page or user interaction. Today, they already include features like push notifications and background sync."

While I'm not sure it's strictly necessary I also set no cache headers for our API responses.


schmmd:
Danke schoen @codeviking for pointing me in this direction!

PR #534, opened 2017-11-22
Replace SRL table output with simple text output.

schmmd:
Far from beautiful, but this leaves us an option that is copy-pastable.  I don't like having two visualizations for SRL output, one far inferior to the other.

image

We certainly could add additional styling on the text to make it clearer.


demo/src/components/SrlComponent.js
```diff
@@ -342,7 +268,7 @@ class HierplaneVisualization extends React.Component {
   selectPrevVerb() {
     const nextIdx =
         this.state.selectedIdx === 0 ? this.props.trees.length - 1 : this.state.selectedIdx - 1;
-    this.setState({ selectedIdx: nextIdx });
+    this.setState({ defaultValue: nextIdx });
```
codeviking: Is this intended? I think it breaks the ability to select the previous verb :).

demo/src/components/SrlComponent.js
```diff
@@ -217,7 +217,7 @@ class SrlInput extends React.Component {
             {srlSentences.map((sentence, index) => {
               const selected = sentence === srlSentenceValue;
               return (
-                <option value={index} key={index} selected={selected}>{sentence}</option>
+                <option value={index} key={index} defaultValue={selected}>{sentence}</option>
```
codeviking: This will change what's selected initially, but mean selection defers to what's selected by the user after the fact.  Is that intentional?

demo/src/components/SrlComponent.js
schmmd: Definitely not ;-)

schmmd:
I was trying to fix a build warning, but was doing a pretty bad job.  I removed those changes.

demo/src/components/SrlComponent.js
```diff
@@ -239,92 +239,18 @@ class SrlInput extends React.Component {
...
+    const { verbs } = this.props;
 
     return (
       <div className="model__content model__content--srl-output">
-        <div className="form__field">
-          {verbs.map((verb, i) => (<SrlFrame verb={verb} words={words} key={i} />))}
+        <div>
+          {verbs.map((verb, i) => {
+              return (
+                  <p key={i}><b>{verb.verb}:</b> {verb.description}</p>
```
codeviking: Using the index as a key has one risk -- if you change the item at that index, without updating the entire array (i.e. updating the reference to a new array), React won't re-render (you won't see the effect of the change).  That said, this likely isn't an issue with the demo.

If it is and/or could be, I would use the verb + index:

```
key={`${i}-${verb.verb}`}
```

That said as noted, likely nothing to worry about, as I'm pretty sure there's no mutations going on here.

schmmd:
The text output works well for Niket.  @joelgrus do you approve of this change?

joelgrus:
it's fine with me

PR #541, opened 2017-11-25
NLVR Type Declaration

pdasigi:
This PR mainly contains a type declaration for NLVR, and the functionality to turn NLVR logical forms into action sequences. In the process of doing that, I made several changes:

1) Defined new module `type_declaration` that contains general classes for creating type systems for new domains.
2) Defined a new class `World` that contains generic methods for logical form processing and for making action sequences.
3) Redefined action sequences to contain original names instead of translated names (given by the name mapping).
4) Made some tweaks to the NLVR language to make type checking work correctly. The only major change is that the language now has lambda forms in `filter_*` functions. Actually, we had lambda forms even earlier, because the second argument to the `filter_*` functions was a function. It's just that they did not explicitly say so. Making them explicit turned out to be necessary for type checking.
So what looked like `filter_equals(all_boxes count(object_in_box) 3)`, now looks like `(filter_equals all_boxes (lambda x (count(object_in_box(var x)))) 3)`

@matt-gardner Sorry this turned out to be yet another big PR. Bit I think this shouldn't take too much of your time since its mostly stuff that got moved around.

### DIFF
--- a/allennlp/data/semparse/type_declarations/nlvr_type_declaration.py
+++ b/allennlp/data/semparse/type_declarations/nlvr_type_declaration.py
@@ -1,0 +1,149 @@
+from typing import Optional
+from overrides import overrides
+
+from nltk.sem.logic import TRUTH_TYPE, EntityType, Type, ComplexType, ANY_TYPE
+
+from allennlp.data.semparse.type_declarations.type_declaration import NamedBasicType, PlaceholderType
+
+
+class BoxFilterType(PlaceholderType):
+    """
+    This is the type of the functions that filter boxes. The corresponding Python function takes three
+    arguments, and looks like ``filter(initial_set_of_boxes, attribute_function, target_attribute)`` .
+    Hence, the signature of the filter function is <b,<<b,#1>,<#1,b>>>.
+    """
+    @property
+    def _signature(self) -> str:
+        return "<b,<<b,#1>,<#1,b>>>"
+
+    @overrides
+    def resolve(self, other: Type) -> Optional[Type]:
+        if not isinstance(other, ComplexType):
+            return None
+        if not isinstance(other.second, ComplexType):
+            return None
+        if not isinstance(other.second.first, ComplexType) or not isinstance(other.second.second, ComplexType):
+            return None
+        resolved_type = other.resolve(ComplexType(BOX_TYPE, ComplexType(ComplexType(BOX_TYPE, ANY_TYPE),
+                                                                        ComplexType(ANY_TYPE, BOX_TYPE))))
+        if not resolved_type:
+            return None
+        first_placeholder = resolved_type.second.first.second
+        second_placeholder = resolved_type.second.second.first
+        resolved_placeholder = first_placeholder.resolve(second_placeholder)
+        if not resolved_placeholder:
+            return None
+        return BoxFilterType(BOX_TYPE, ComplexType(ComplexType(BOX_TYPE, resolved_placeholder),
+                                                   ComplexType(resolved_placeholder, BOX_TYPE)))
+
+    @overrides
+    def get_application_type(self, argument_type: Type) -> Type:
+        return self.second
+
+
+class AssertType(PlaceholderType):
+    """
+    This is the type of assert operations. Corresponding Python function looks like
+    ``assert(attribute, target_attribute)`` , where the attribute and target attribute can either be numbers,
+    shapes or colors, and the return type is boolean. So the signature of the function is <#1,<#1,t>>.
+    """
+    @property
+    def _signature(self) -> str:
+        return "<#1,<#1,t>>"
+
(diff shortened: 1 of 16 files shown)


allennlp/data/semparse/type_declarations/nlvr_type_declaration.py
```diff
@@ -0,0 +1,185 @@
...
+
+    @overrides
+    def get_application_type(self, argument_type: Type) -> Type:
+        return ComplexType(argument_type, TRUTH_TYPE)
+
+
+class CountType(PlaceholderType):
+    """
+    This is the type of the count function. Corresponding Python function looks like ``count(entities)`` ,
+    where entities are either objects or boxes. Type signature is <#1,e>.
```
matt-gardner: `e` is the `NUM_TYPE`?  Why?

allennlp/data/semparse/worlds/world.py
```diff
@@ -0,0 +1,129 @@
...
+        self.reverse_name_mapping = {mapped_name: name for name, mapped_name in self.global_name_mapping.items()}
+        type_prefixes = constant_type_prefixes or {}
+        self._logic_parser = DynamicTypeLogicParser(constant_type_prefixes=type_prefixes,
+                                                    type_signatures=self.global_type_signatures)
+
+    def parse_logical_form(self, logical_form: str) -> Expression:
+        """
+        Takes a logical form as a string, maps its tokens using the mapping and returns a parsed expression.
+        """
+        def _process_nested_expression(nested_expression) -> str:
```
matt-gardner: This should not be a nested function.  In general, the only reason to have a nested function is if you're _returning_ the function itself.  Otherwise you're almost certainly better off defining the function without nesting, both for code readability and for efficiency.

tests/data/dataset_readers/wikitables/wikitables_dataset_reader_test.py
```diff
@@ -16,11 +16,12 @@ def test_reader_reads(self):
         question_tokens = ["what", "was", "the", "last", "year", "where", "this", "team", "was",
                            "a", "part", "of", "the", "usl", "a", "-", "league", "?"]
         assert [t.text for t in instance.fields["question"].tokens] == question_tokens
-        assert actions == ['@@START@@', 'd', 'd -> [<d,d>, d]', '<d,d> -> M0', 'd -> [<e,d>, e]',
-                           '<e,d> -> [<<#1,#2>,<#2,#1>>, <d,e>]', '<<#1,#2>,<#2,#1>> -> R',
-                           '<d,e> -> D1', 'e -> [<r,e>, r]', '<r,e> -> [<<#1,#2>,<#2,#1>>, <e,r>]',
-                           '<<#1,#2>,<#2,#1>> -> R', '<e,r> -> C0', 'r -> [<e,r>, e]', '<e,r> -> C1',
-                           'e -> cell:usl_a_league', '@@END@@']
+        assert actions == ['@@START@@', 'd', 'd -> [<d,d>, d]', '<d,d> -> max', 'd -> [<e,d>, e]',
```
matt-gardner: Much improved!  Can you also do this with the types?  @OyvindTafjord had some questions around this - using single lower-case letters for the types means you can only ever have 26 types, right?

pdasigi: You mean show types like `<<num,cell>,cell> -> [<cell,<<num,cell>,cell>>, cell]` instead of `<<d,e>,e> -> [<e,<<d,e>,e>>, e]`? That shouldn't be difficult, but I don't think it is necessary. We have 4-5 basic types in the domains we are looking at now, and I think keeping them as single lower-case letters makes it more readable. But yes, if we extend this to domains where there are a lot more types, that may be necessary, both for understanding what's happening and also for potentially having more than 26 basic types.

allennlp/data/semparse/worlds/world.py

allennlp/data/semparse/type_declarations/nlvr_type_declaration.py
pdasigi: `e` is the entity type in NLTK and all constants default to it. I dealt with constants having different types by letting the world specify `constant_type_prefixes`: https://github.com/pdasigi/allennlp/blob/c04c985521ffad9ec318c75890b012d010336cc9/allennlp/data/semparse/worlds/world.py#L19-L21
This requires mapping names of constants to contain their types as prefixes as well. In NLVR, since numbers are the only constants we see in the logical forms, I let them default to entities to avoid having to translate them. I shall write a comment about this.

tests/data/dataset_readers/wikitables/wikitables_dataset_reader_test.py
pdasigi: Added a `TODO` for this.

PR #543, opened 2017-11-27
Add `ScalarMix` module

matt-peters:
Add `ScalarMix` module that computes scalar weighted averages of tensors, with optional layer normalization

### DIFF
--- a/allennlp/modules/scalar_mix.py
+++ b/allennlp/modules/scalar_mix.py
@@ -1,0 +1,58 @@
+from typing import List
+import torch
+import numpy
+from torch.nn import ParameterList, Parameter
+
+from allennlp.common.checks import ConfigurationError
+
+class ScalarMix(torch.nn.Module):
+    """
+    Computes a parameterised scalar mixture of N tensors, ``mixture = gamma * sum(s_k * tensor_k)``
+    where ``s = softmax(w)``, with ``w`` and ``gamma`` scalar parameters.
+
+    In addition, if ``do_layer_norm=True`` then apply layer normalization to each tensor
+    before weighting.
+    """
+    def __init__(self, mixture_size: int, do_layer_norm: bool=False):
+        super(ScalarMix, self).__init__()
+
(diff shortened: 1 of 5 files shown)


allennlp/modules/scalar_mix.py
```diff
@@ -0,0 +1,59 @@
...
+        super(ScalarMix, self).__init__()
+
+        self.mixture_size = mixture_size
+        self.do_layer_norm = do_layer_norm
+
+        self.scalar_parameters = ParameterList([Parameter(torch.FloatTensor([0.0]))
+                                                for _ in range(mixture_size)])
+        self.gamma = Parameter(torch.FloatTensor([1.0]))
+
+    def forward(self, tensors: List[torch.Tensor],  # pylint: disable=arguments-differ
```
DeNeutoy: Please can you add a tiny docstring here and explain that the mask is for all of the tensors that are being mixed.

allennlp/modules/scalar_mix.py
```diff
@@ -0,0 +1,70 @@
...
+                mask: torch.Tensor = None) -> torch.Tensor:
+        """
+        Compute a weighted average of the ``tensors``.  The input tensors an be any shape
```
DeNeutoy: @matt-peters Was it really too much effort to match the docstring style of the rest of `allennlp`? I know you think it doesn't matter, but it _does_ make it easier to read and more uniform, which is really the key thing. It's just going to make reviews take longer if we can't approve stuff with changes and expect you to add it in some uniform style.

matt-gardner: In fairness to @matt-peters, we don't have a place to look for what that style should be, and we haven't always been uniform ourselves.  I'll add a section to the style guide PR I'm putting together that gives guidance on what things should go where for methods like this.

matt-peters: Yes if you provide a template for a docstring then I'll fill it out.

PR #545, opened 2017-11-28
Support databases on GCP.

schmmd:
* Set up proxys for GCP databases (follows https://cloud.google.com/sql/docs/mysql/connect-kubernetes-engine).
* Add liveness probe to staging.

### DIFF
--- a/scripts/ai2-internal/deploy-production-demo.sh
+++ b/scripts/ai2-internal/deploy-production-demo.sh
@@ -51,24 +51,20 @@ spec:
           # for documentation on the resources section.
           env:
             - name: DEMO_POSTGRES_HOST
-              valueFrom:
-                secretKeyRef:
-                  name: allennlp-demo-postgres
-                  key: host
+              value: "127.0.0.1"
+            - name: DEMO_POSTGRES_PORT
+              value: "5432"
             - name: DEMO_POSTGRES_DBNAME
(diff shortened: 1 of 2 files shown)


scripts/ai2-internal/deploy-production-demo.sh
```diff
@@ -51,24 +51,20 @@ spec:
...
-                secretKeyRef:
-                  name: allennlp-demo-postgres
-                  key: dbname
+              value: "demo"
             - name: DEMO_POSTGRES_USER
               valueFrom:
                 secretKeyRef:
-                  name: allennlp-demo-postgres
-                  key: user
+                  name: cloudsql-db-credentials
```
joelgrus: if it were me, I would choose a more specific name (conceivably in the future allennlp could have  multiple cloudsql databases)

schmmd: I agree, but I don't want to tackle this now (have to move on sometime!).  I'd also like to remove the AllenNLP namespace, since we will be running our stuff on a dedicated AllenNLP Kubernetes cluster so the namespace is just extra baggage.

schmmd:
@joelgrus rather than taking my word on the Google Cloud stuff, see the link to the documentation.  I followed it mostly verbatim--and that's where my poor naming choices came from.

PR #553, opened 2017-11-30
Add numbers to installation steps.

schmmd:


schmmd:
@joelgrus I think this is a good change, but I mostly wanted to understand how deployments went to the website.  Will those happen automatically once this is merged?

joelgrus:
I am currently trying to see if the `pip` install can automatically install the spacy library, but I'm not having great luck

PR #563, opened 2017-12-01
Add ability to read hdf5 formatted Embedding

matt-peters:


allennlp/modules/token_embedders/embedding.py
```diff
@@ -253,3 +274,24 @@ def _read_pretrained_embedding_file(embeddings_filename: str,
...
+                                                namespace: str = "tokens") -> torch.FloatTensor:
+    """
+    Reads from a hdf5 formatted file.  The embedding matrix is assumed to
+    be keyed by 'embedding' and of size ``(num_tokens, embedding_dim)``.
+    """
+    with h5py.File(embeddings_filename, 'r') as fin:
+        embeddings = fin['embedding'][...]
+
+    if embeddings.shape[0] != vocab.get_vocab_size(namespace):
+        ConfigurationError("Read {0} embeddings from the file, but expected {1}".format(
```
DeNeutoy: You need to actually `raise` these - I think you're just instanciating `ConfigurationError` objects. Can you do this and add tests that they do raise in the correct circumstances?

matt-peters: This is fixed

PR #571, opened 2017-12-04
Canonical ontonotes

DeNeutoy:
Adds a `Ontonotes` class which reads all Ontonotes 5.0 annotations into a generator of `OntonotesSentence`'s. This is eventually going to be used heavily in the SRL reader and will replace the `_DocumentState` class in the coref reader, but this is a natural breaking point to split it up. 

`nltk` does have an ontonotes reader, but it reads everything in separate passes over the dataset and also doesn't read coref annotations, for some reason. I decided it would just be easier to have our own, rather than subclassing theirs, as they do loads of weird hyper functional stuff with maps and curried lazy loading etc which is not very pythonic and very obtuse.

### DIFF
--- a/allennlp/data/dataset_readers/dataset_utils/ontonotes.py
+++ b/allennlp/data/dataset_readers/dataset_utils/ontonotes.py
@@ -1,4 +1,4 @@
-from typing import Dict, DefaultDict, List, Optional, Iterator, Set, Tuple
+from typing import Dict, DefaultDict, List, Optional, Iterator, Set, Tuple, Union
 from collections import defaultdict
 import codecs
 import os
@@ -12,10 +12,12 @@ from allennlp.common import Params
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
-TypedSpan = Tuple[str, Tuple[int, int]] # pylint: disable=invalid-name
+TypedSpan = Tuple[Union[int, str], Tuple[int, int]]  # pylint: disable=invalid-name
+
 
 class OntonotesSentence:
     """
+    A class representing the annotations available for a single CONLL formatted sentence.
 
     Parameters
     ----------
@@ -48,9 +50,10 @@ class OntonotesSentence:
     srl_frames : ``Dict[str, List[str]]``
         A dictionary keyed by the verb in the sentence for the given
         Propbank frame labels, in a BIO format.
-    coreference_clusters : str
-        Co-reference chain information encoded in a parenthesis structure. For documents that do
-        not have co-reference annotations, each line is represented with a "-".
+    coref_spans : ``Set[TypedSpan]``
+        The spans for entity mentions involved in coreference resolution within the sentence.
+        Each element is a tuple composed of (cluster_id, (start_index, end_index)). Indices
+        are `inclusive`.
     """
     def __init__(self,
                  document_id: str,
@@ -64,7 +67,7 @@ class OntonotesSentence:
                  speakers: List[Optional[str]],
                  named_entities: List[str],
                  srl_frames: Dict[str, List[str]],
-                 coref_spans: Set[Tuple[int, Tuple[int, int]]]):
+                 coref_spans: Set[TypedSpan]):
 
         self.document_id = document_id
         self.sentence_id = sentence_id
@@ -78,9 +81,6 @@ class OntonotesSentence:
         self.named_entities = named_entities
         self.srl_frames = srl_frames
         self.coref_spans = coref_spans
-
-    def bio_to_typed_spans(self, bio_labels: List[str]) -> Set[Tuple[str, Tuple[int, int]]]:
-        pass
 
 
 class Ontonotes:
@@ -179,8 +179,7 @@ class Ontonotes:
          not have co-reference annotations, each line is represented with a "-".
 
     """
-    def __init__(self, tagging_type: str = "bio") -> None:
-        self.tagging_type = tagging_type
+    def __init__(self) -> None:
 
     def datset_iterator(self, file_path) -> Iterator[OntonotesSentence]:
         """
@@ -220,10 +219,10 @@ class Ontonotes:
                     if not conll_rows:
                         continue
                     else:
-                        yield self.conll_rows_to_sentence(conll_rows)
+                        yield self._conll_rows_to_sentence(conll_rows)
                         conll_rows = []
 
-    def conll_rows_to_sentence(self, conll_rows: List[str]) -> OntonotesSentence:
+    def _conll_rows_to_sentence(self, conll_rows: List[str]) -> OntonotesSentence:
 
         document_id: str = None
         sentence_id: str = None
@@ -283,9 +282,9 @@ class Ontonotes:
                 # sequence we are collecting.
                 current_span_labels = [None for _ in conll_components[10:-1]]
 
-            current_span_labels, span_labels = self.process_span_annotations_for_word(conll_components[10:-1],
-                                                       span_labels,
-                                                       current_span_labels)
+            self._process_span_annotations_for_word(conll_components[10:-1],
+                                                    span_labels,
+                                                    current_span_labels)
 
             # If any annotation contains this word as a verb predicate,
             # we need to record its index. This also has the side effect
@@ -295,10 +294,10 @@ class Ontonotes:
             if word_is_verbal_predicate:
                 verbal_predicates.append(word)
 
-            self.process_coref_span_annotations_for_word(conll_components[-1],
-                                                         index,
-                                                         clusters,
-                                                         coref_stacks)
+            self._process_coref_span_annotations_for_word(conll_components[-1],
+                                                          index,
+                                                          clusters,
+                                                          coref_stacks)
             coreference.append(conll_components[-1])
 
             sentence.append(word)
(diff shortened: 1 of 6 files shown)


allennlp/data/dataset_readers/dataset_utils/ontonotes.py
```diff
@@ -0,0 +1,418 @@
...
+        are `inclusive`.
+    """
+    def __init__(self,
+                 document_id: str,
+                 sentence_id: int,
+                 words: List[str],
+                 pos_tags: List[str],
+                 parse_tree: Tree,
+                 predicate_lemmas: List[Optional[str]],
+                 predicate_framenet_ids: List[Optional[str]],
```
matt-gardner: You're not consistent about marking these as `Optional` above.

allennlp/data/dataset_readers/dataset_utils/ontonotes.py
```diff
@@ -0,0 +1,418 @@
...
+                       ├── bc
+                       ├── bn
+                       ├── mz
+                       ├── nw
+                       ├── pt
+                       ├── tc
+                       └── wb
+
+    The file path provided to this class can then be any of the train, test or development
+    directories(or the top level data directory, if you are not utilizing the splits).
```
matt-gardner: `s/directories(/directories (/`

allennlp/data/dataset_readers/dataset_utils/ontonotes.py
```diff
@@ -0,0 +1,418 @@
...
+        which do not have named entity annotation, each line is represented with an ``*``.
+    12+ Predicate Arguments: ``str``
+        There is one column each of predicate argument structure information for the predicate
+        mentioned in Column 7. If there are no predicates tagged in a sentence this is a
+        single column with all rows marked with an ``*``.
+    -1 Co-reference: ``str``
+        Co-reference chain information encoded in a parenthesis structure. For documents that do
+         not have co-reference annotations, each line is represented with a "-".
+    """
+    def datset_iterator(self, file_path) -> Iterator[OntonotesSentence]:
```
matt-gardner: Hmm, I didn't know you could annotate a `Generator` this way.  We should update the type annotations on our `DataIterators` to do this too; it's cleaner than what he have there now.  (Just a note.)

allennlp/data/dataset_readers/dataset_utils/ontonotes.py
```diff
@@ -0,0 +1,418 @@
...
+        """
+        An iterator over the entire dataset, yielding all sentences processed.
+        """
+        for conll_file in self.dataset_path_iterator(file_path):
+            yield from self.sentence_iterator(conll_file)
+
+    @staticmethod
+    def dataset_path_iterator(file_path: str) -> Iterator[str]:
+        """
+        An iterator containing file_paths in a directory
```
matt-gardner: "An iterator returning"?

allennlp/data/dataset_readers/dataset_utils/ontonotes.py
```diff
@@ -0,0 +1,418 @@
...
+
+            # Replace brackets in text with a different
+            # token for parse trees.
+            if word == "(":
+                parse_word = "-LRB-"
+            elif word == ")":
+                parse_word = "-RRB-"
+            else:
+                parse_word = word
+            parse_piece = parse_piece.replace("*", f" {parse_word}")
```
matt-gardner: What does this line do?  Ah, I see you've documented this above; a quick pointer here would be good.  Also, it looks like the docs above say the POS tag should also be in here, right?

allennlp/data/dataset_readers/dataset_utils/ontonotes.py
```diff
@@ -0,0 +1,418 @@
...
+            lemmatised_word = conll_components[6]
+            framenet_id = conll_components[7]
+            word_sense = conll_components[8]
+            speaker = conll_components[9]
+
+            if not span_labels:
+                # Create empty lists to collect the NER and SRL BIO labels.
```
matt-gardner: A note here that you'll be adding to these for each word would be good.  I was confused at first why you couldn't just pass `conll_components[10:-1]` to `_process_span_annotations`.  Even just saying "If this is the first word, we create empty ..." would have been enough for me.

allennlp/data/dataset_readers/dataset_utils/ontonotes.py
```diff
@@ -0,0 +1,418 @@
...
+                span_labels = [[] for _ in conll_components[10:-1]]
+                # Create variables representing the current label for each label
+                # sequence we are collecting.
+                current_span_labels = [None for _ in conll_components[10:-1]]
+
+            self._process_span_annotations_for_word(conll_components[10:-1],
+                                                    span_labels,
+                                                    current_span_labels)
+
+            # If any annotation contains this word as a verb predicate,
```
matt-gardner: "contains" here was confusing to me for some reason.  "says" would have cleared the confusion.  I can't put my finger on _why_, though, so if you think this is fine, that's ok too.

allennlp/data/dataset_readers/dataset_utils/ontonotes.py
matt-gardner: Also, seeing the coref processing below, it might be nice to say why you can't do this up front - you need to know how many things you're tracking, which you only get from the first word.

allennlp/data/dataset_readers/dataset_utils/ontonotes.py
```diff
@@ -0,0 +1,418 @@
...
+                    start = coref_stacks[cluster_id].pop()
+                    clusters[cluster_id].append((start, word_index))
+
+    @staticmethod
+    def _process_span_annotations_for_word(annotations: List[str],
+                                           span_labels: List[List[str]],
+                                           current_span_labels: List[Optional[str]]):
+        """
+        Given a sequence of different label types for a single word and the same labels
+        for the previous word, compute the BIO tag for each label and append to a list.
```
matt-gardner: By "the same labels for the previous word", do you mean `current_span_labels`?  If so, it'd be good to be consistent in how you're referring to this (either with "previous" or "current").

allennlp/data/dataset_readers/dataset_utils/ontonotes.py
```diff
@@ -0,0 +1,418 @@
...
+        annotations: ``List[str]``
+            A list of labels to compute BIO tags for.
+        span_labels : ``List[List[str]]``
+            A list of lists, one for each annotation, to incrementally collect
+            the BIO tags for a sequence.
+        current_span_labels : ``List[Optional[str]]``
+            The currently open span per annotation type, or ``None`` if there is no open span.
+        """
+        for annotation_index, annotation in enumerate(annotations):
+            label = annotation.strip("()*")
```
matt-gardner: Quick note about what you're doing here?

allennlp/data/dataset_readers/dataset_utils/ontonotes.py
```diff
@@ -0,0 +1,418 @@
...
+
+            if "(" in annotation:
+                # Entering into a span for a particular semantic role label.
+                # We append the label and set the current span for this annotation.
+                bio_label = "B-" + label
+                span_labels[annotation_index].append(bio_label)
+                current_span_labels[annotation_index] = label
+
```
matt-gardner: Remove blank line.

allennlp/data/dataset_readers/dataset_utils/ontonotes.py
```diff
@@ -0,0 +1,418 @@
...
+            else:
+                # We're outside a span.
+                span_labels[annotation_index].append("O")
+
+            # Exiting a span, so we reset the current span label for this annotation.
+            if ")" in annotation:
+                current_span_labels[annotation_index] = None
+
+    @classmethod
+    def from_params(cls, params: Params) -> 'Ontonotes':
```
matt-gardner: I don't see a purpose for this method - if you're going to use Ontonotes, you probably know you're going to use Ontonotes, and won't be building this `from_params`.

tests/data/dataset_readers/ontonotes_test.py
```diff
@@ -0,0 +1,87 @@
...
+
+from allennlp.data.dataset_readers.dataset_utils import Ontonotes
+from allennlp.common.testing import AllenNlpTestCase
+
+
+class TestOntonotes(AllenNlpTestCase):
+
+    def test_dataset_iterator(self):
+        reader = Ontonotes()
+        annotated_sentences = list(reader.datset_iterator('tests/fixtures/conll_2012/'))
```
matt-gardner: `s/datset_iterator/dataset_iterator/`

allennlp/data/dataset_readers/dataset_utils/ontonotes.py
```diff
@@ -0,0 +1,418 @@
...
+                                 srl_frames,
+                                 coref_span_tuples)
+
+    @staticmethod
+    def _process_coref_span_annotations_for_word(label: str,
+                                                 word_index: int,
+                                                 clusters: DefaultDict[int, List[Tuple[int, int]]],
+                                                 coref_stacks: DefaultDict[int, List[int]]) -> None:
+        """
+        For a given coref label, add it to a currently open span/s, complete a span/s or
```
matt-gardner: I've more commonly seen "span(s)" instead of "span/s" - took me a minute to parse what you meant by "span/s".

allennlp/data/dataset_readers/dataset_utils/ontonotes.py
```diff
@@ -0,0 +1,418 @@
...
+                    # Retrieve the start index from the document state and
+                    # add the span to the clusters for this id.
+                    cluster_id = int(segment[:-1])
+                    start = coref_stacks[cluster_id].pop()
+                    clusters[cluster_id].append((start, word_index))
+
+    @staticmethod
+    def _process_span_annotations_for_word(annotations: List[str],
+                                           span_labels: List[List[str]],
+                                           current_span_labels: List[Optional[str]]):
```
matt-gardner: Return type annotation.

allennlp/data/dataset_readers/dataset_utils/ontonotes.py
DeNeutoy: Yeah - I think the `Generator` type is only really useful if you're using them in a mega complex way as coroutines and actually sending things to them, rather than just using them as an `Iterator`.

DeNeutoy:
The merging of the clusters can happen in the coref reader still - this should just be a way to read in, fairly exactly, the Ontonotes dataset.

PR #576, opened 2017-12-05
Sorting state fixes

### LINKED ISSUE #572: States are not handled correctly when passed to non-stateful Seq2SeqEncoders

DeNeutoy:
There were 2 problems here:
- I forgot to sort and crop states passed to non-stateful encoders, so there were shape mis-matches.
- I hadn't wrapped zeros used to restore the correct shape to cropped states with `Variables`, so there were gradients being passed to tensors in the backward pass. 

I added 3 tests which all fail  in the first commit, fixed by the next one.
Fixes #572

### DIFF
--- a/allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
+++ b/allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
@@ -1,5 +1,6 @@
 
 import torch
+from torch.autograd import Variable
 from torch.nn.utils.rnn import pad_packed_sequence
 
 from allennlp.common.checks import ConfigurationError
@@ -87,6 +88,7 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
         if num_valid < batch_size:
             _, length, output_dim = unpacked_sequence_tensor.size()
             zeros = unpacked_sequence_tensor.data.new(batch_size - num_valid, length, output_dim).fill_(0)
+            zeros = Variable(zeros)
             unpacked_sequence_tensor = torch.cat([unpacked_sequence_tensor, zeros], 0)
 
             # The states also need to have invalid rows added back.
@@ -95,6 +97,7 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
                 for state in final_states:
                     num_layers, _, state_dim = state.size()
                     zeros = state.data.new(num_layers, batch_size - num_valid, state_dim).fill_(0)
+                    zeros = Variable(zeros)
                     new_states.append(torch.cat([state, zeros], 1))
                 final_states = new_states
 
@@ -107,7 +110,7 @@ class PytorchSeq2SeqWrapper(Seq2SeqEncoder):
             zeros = unpacked_sequence_tensor.data.new(batch_size,
                                                       sequence_length_difference,
                                                       unpacked_sequence_tensor.size(-1)).fill_(0)
-            zeros = torch.autograd.Variable(zeros)
+            zeros = Variable(zeros)
             unpacked_sequence_tensor = torch.cat([unpacked_sequence_tensor, zeros], 1)
 
         if self.stateful:
--- a/tests/modules/encoder_base_test.py
+++ b/tests/modules/encoder_base_test.py
@@ -34,6 +34,30 @@ class TestEncoderBase(AllenNlpTestCase):
         self.sorting_indices = sorting_indices
         self.restoration_indices = restoration_indices
 
+    def test_non_stateful_states_are_sorted_correctly(self):
+
+        encoder_base = _EncoderBase(stateful=False)
+        initial_states = (Variable(torch.randn(6, 5, 7)),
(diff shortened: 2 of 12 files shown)


tests/modules/encoder_base_test.py
```diff
@@ -34,6 +34,26 @@ def setUp(self):
         self.sorting_indices = sorting_indices
         self.restoration_indices = restoration_indices
 
+    def test_non_stateful_states_are_sorted_correctly(self):
+        encoder_base = _EncoderBase(stateful=False)
+        initial_states = (Variable(torch.randn(6, 5, 7)),
```
matt-gardner: These shapes are confusing me - can you give a comment explaining the shape here?

matt-gardner: And this is a tuple because it's the hidden state and the memory cell for the LSTM?

tests/modules/encoder_base_test.py
```diff
@@ -34,6 +34,26 @@ def setUp(self):
...
+        # Check that we sort the state for non-stateful encoders. To test
+        # we'll just use a "pass through" encoder, as we aren't actually testing
+        # the functionality of the encoder here anyway.
+        _, states, restoration_indices = encoder_base.sort_and_run_forward(lambda *x: x,
+                                                                           self.tensor,
+                                                                           self.mask,
+                                                                           initial_states)
+        zeros = torch.zeros([6, 2, 7])
+        for state, original in zip(states, initial_states):
+            assert list(state.size()) == [6, 3, 7]
```
matt-gardner: An explanation of what you're doing here would be helpful, as this is somewhat non-obvious.  Something like, "`sort_and_run_forward` strips fully-padded instances from the batch; in order to use the `restoration_indices` we need to add back the two that got stripped.  What we get back should match what we started with."

tests/modules/seq2seq_encoders/pytorch_seq2seq_wrapper_test.py
```diff
@@ -124,6 +124,46 @@ def test_wrapper_raises_if_batch_first_is_false(self):
...
+        tensor[1, 4:, :] = 0
+        tensor[2, 0:, :] = 0
+        tensor[3, 6:, :] = 0
+        mask = torch.ones(5, 7)
+        mask[0, 3:] = 0
+        mask[1, 4:] = 0
+        mask[2, 0:] = 0
+        mask[3, 6:] = 0
+
+        initial_states = (Variable(torch.randn(6, 5, 7)),
```
matt-gardner: Oh, the 6 is because this is bidirectional?  Yeah, a shape comment would have helped a lot here.

tests/modules/seq2seq_encoders/pytorch_seq2seq_wrapper_test.py
```diff
@@ -124,6 +124,46 @@ def test_wrapper_raises_if_batch_first_is_false(self):
             lstm = LSTM(bidirectional=True, num_layers=3, input_size=3, hidden_size=7)
             _ = PytorchSeq2SeqWrapper(lstm)
 
+    def test_wrapper_works_when_passed_state_with_zero_length_sequences(self):
+        lstm = LSTM(bidirectional=True, num_layers=3, input_size=3, hidden_size=7, batch_first=True)
+        encoder = PytorchSeq2SeqWrapper(lstm)
+        tensor = torch.rand([5, 7, 3])
+        tensor[0, 3:, :] = 0
```
matt-gardner: You don't need to zero out the tensor, just the mask, throughout the test.

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
```diff
@@ -87,6 +88,7 @@ def forward(self,  # pylint: disable=arguments-differ
         if num_valid < batch_size:
             _, length, output_dim = unpacked_sequence_tensor.size()
             zeros = unpacked_sequence_tensor.data.new(batch_size - num_valid, length, output_dim).fill_(0)
+            zeros = Variable(zeros)
```
matt-gardner: pytorch 0.4 will be nice...

tests/modules/seq2seq_encoders/pytorch_seq2seq_wrapper_test.py
matt-gardner: You could also make one of these tests only use a single variable for the hidden state, instead of two, to make sure we cover that case correctly.

tests/modules/encoder_base_test.py
DeNeutoy: oh yeah sorry that is weird - it's `num_layers * directions` - thanks pytorch 👍

allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py
DeNeutoy: In 0.3, you can't concat variables with tensors, so this would have broken, which is still decent.

PR #577, opened 2017-12-05
Add support for purely character input in `get_text_field_mask`

matt-peters:
`get_text_field_mask` was returning the wrong mask for models with purely character input.  This PR fixes it.

### DIFF
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -302,10 +302,16 @@ def get_text_field_mask(text_field_tensors: Dict[str, torch.Tensor]) -> torch.Lo
     """
     tensor_dims = [(tensor.dim(), tensor) for tensor in text_field_tensors.values()]
     tensor_dims.sort(key=lambda x: x[0])
-    token_tensor = tensor_dims[0][1]
 
-    return (token_tensor != 0).long()
-
+    smallest_dim = tensor_dims[0][0]
+    if smallest_dim == 2:
+        token_tensor = tensor_dims[0][1]
+        return (token_tensor != 0).long()
+    elif smallest_dim == 3:
+        character_tensor = tensor_dims[0][1]
+        return ((character_tensor > 0).long().sum(dim=-1) > 0).long()
+    else:
+        raise ValueError("Expected a tensor with dimension 2 or 3, found {}".format(smallest_dim))
 
 def _last_dimension_applicator(function_to_apply: Callable[[torch.Tensor, Optional[torch.Tensor]], torch.Tensor],
                                tensor: torch.Tensor,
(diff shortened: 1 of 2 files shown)


allennlp/nn/util.py
```diff
@@ -302,10 +302,16 @@ def get_text_field_mask(text_field_tensors: Dict[str, torch.Tensor]) -> torch.Lo
...
-
-    return (token_tensor != 0).long()
 
+    smallest_dim = tensor_dims[0][0]
+    if smallest_dim == 2:
+        token_tensor = tensor_dims[0][1]
+        return (token_tensor != 0).long()
+    elif smallest_dim == 3:
+        character_tensor = tensor_dims[0][1]
+        return ((character_tensor > 0).long().sum(dim=-1) > 0).long()
```
matt-gardner: The trouble here is that if you have a `ListField[TextField]`, your smallest dim will be 3, but you don't want to do to this summing operation (see note in the docstring).  Apparently we don't have a test for this case, because this change is passing tests, but we should probably add one, because people are using `ListField[TextField]` in some places.

I think the way to solve this is to pass in to this method how many dimensions you've added over the sequence length (e.g., how many `ListFields` there are).  Then you can reliably do this 2 vs. 3 calculation.

matt-peters: That's another issue altogether.  The current code always returned a mask with same dimensions as the smallest tensor so would have also broken in the case of `ListField[TextField]` without a 2D token tensor also as input.  What is the correct mask size in the case of a `ListField[TextField]` and how is it computed from a tensor?

matt-gardner: The issue is your check on `smallest_dim`.  With a `ListField[TextField]`, `smallest_dim` will be 3, and currently returns a mask of shape `(batch_size, num_sentences, sentence_length)`, which is what you want.  After this change, it would return a mask of shape `(batch_size, num_sentences)`, which is _not_ what you want.

matt-gardner: This is assuming you had a simple word id tensor somewhere in you're input.  You're right that everything is broken if you don't have that, and we need to fix it.  Just saying that when fixing that issue, you need to not forget about (and unintentionally break) other uses, like `ListField[TextField]`.

matt-peters: That's inconsistent with the docstring that says that 2D masks are always returned... so I'm confused.

matt-peters: I see -- so there is no way for this function to know the proper behavior without the user providing additional information.  How about we leave a big disclaimer in the docstring for this function, leave it unchanged, then introduce another `get_character_id_mask` that just takes a 3D tensor of character ids and returns a mask?  Then it's the user's responsibility to use it properly.

Otherwise, we need to provide an optional key to specify which tensor to use for the mask, and a flag whether to treat the 3D tensor as a character tensor or `ListField[TextField]`.

matt-peters: Pushed some commits that make these changes and add a test for the `ListField` case.

matt-gardner:
Sorry to jump in @DeNeutoy - I started looking at this because I saw an email that made me curious, before @matt-peters asked you for a review.

matt-gardner:
And @matt-peters, it was easier to just fix the docstring than to tell you how to fix the docstring, so I made some minor changes.  Thanks for finding and fixing this bug!

PR #579, opened 2017-12-05
Elmo vocab embeddings

DeNeutoy:


scripts/create_elmo_embeddings_from_vocab.py
```diff
@@ -0,0 +1,87 @@
...
+from allennlp.data import Token, Vocabulary
+
+
+def main(vocab_path: str,
+         elmo_model_dir: str,
+         output_dir: str,
+         batch_size: int,
+         device: int):
+
+    options_file = os.path.join(elmo_model_dir, 'options.json')
```
matt-peters: These file names will change between models to be more descriptive, so we need to pass in the individual names not the directory.

scripts/create_elmo_embeddings_from_vocab.py
```diff
@@ -0,0 +1,87 @@
...
+
+    all_embeddings = []
+    for i in range((len(sentences) // batch_size) + 1):
+        if device != -1:
+            batch = Variable(torch.from_numpy(numpy.array(sentences[i * batch_size: (i + 1) * batch_size])).cuda(device_id=device))
+        else:
+            batch = Variable(torch.from_numpy(numpy.array(sentences[i * batch_size: (i + 1) * batch_size])))
+
+        token_embedding = elmo_token_embedder(batch)['token_embedding'].data
+        print(token_embedding.shape)
```
matt-peters: We probably don't need this.

scripts/create_elmo_embeddings_from_vocab.py
```diff
@@ -0,0 +1,87 @@
...
+
+        # Reshape back to a list of words of shape (batch_size * 50, encoding_dim)
+        # We also need to remove the <S>, </S> tokens appended by the encoder.
+        per_word_embeddings = token_embedding[:, 1:-1, :].contiguous().view(-1, token_embedding.size(-1))
+        print(per_word_embeddings.size())
```
matt-peters: Debug statement here too.

scripts/create_elmo_embeddings_from_vocab.py
```diff
@@ -0,0 +1,87 @@
...
+
+        all_embeddings.append(per_word_embeddings)
+
+    # Remove the embeddings associated with padding in the last batch.
+    all_embeddings[-1] = all_embeddings[-1][:-last_batch_remainder, :]
+
+    embedding_weight = torch.cat(all_embeddings, 0).numpy()
+    print(embedding_weight.shape)
```
matt-peters: This too

scripts/create_elmo_embeddings_from_vocab.py
```diff
@@ -0,0 +1,87 @@
...
+         output_dir: str,
+         batch_size: int,
+         device: int):
+
+    options_file = os.path.join(elmo_model_dir, 'options.json')
+    weight_file = os.path.join(elmo_model_dir, 'lm_weights.hdf5')
+
+    # Load the vocabulary words and convert to char ids
+    with open(vocab_path, 'r') as vocab_file:
+        tokens = vocab_file.read().strip().split('\n')
```
matt-peters: We need to add the special tokens `<S>` and `</S>` to this file if they don't exist.  Are there other special tokens necessary to make it usable with the allennlp Vocabulary?  Also, do we need to modify the file format so that it can be loaded with an allennlp Vocabulary?

scripts/create_elmo_embeddings_from_vocab.py
```diff
@@ -0,0 +1,87 @@
...
+import numpy
+import torch
+from torch.autograd import Variable
+
+from allennlp.data.token_indexers import ELMoTokenCharactersIndexer
+from allennlp.modules.elmo import _ElmoCharacterEncoder
+from allennlp.data import Token, Vocabulary
+
+
+def main(vocab_path: str,
```
matt-peters: Can you move this to somewhere inside the allennlp/ directory structure so that it can be imported and re-used elsewhere?

scripts/create_elmo_embeddings_from_vocab.py
```diff
@@ -0,0 +1,87 @@
+# pylint: disable=no-self-use
+
+import os
+import argparse
+import sys
+
+sys.path.insert(0, os.path.dirname(os.path.abspath(os.path.join(__file__, os.pardir))))
```
matt-peters: Why is this necessary?

scripts/create_elmo_embeddings_from_vocab.py
DeNeutoy: where do we need to use it inside the library?

scripts/create_elmo_embeddings_from_vocab.py
DeNeutoy: Ah it just means you can run the script from the top level directory and import things as though this script resided in the top level directory.

scripts/create_elmo_embeddings_from_vocab.py
DeNeutoy: sure

scripts/create_elmo_embeddings_from_vocab.py
matt-peters: It'd be nice to reuse in a test case for another PR I'm working on.  But if it's a hassle then I can work around.

scripts/create_elmo_embeddings_from_vocab.py
```diff
@@ -0,0 +1,112 @@
...
+    file containing word : vector pairs, one per line, with all values
+    separated by a space.
+    """
+
+    # Load the vocabulary words and convert to char ids
+    with open(vocab_path, 'r') as vocab_file:
+        tokens = vocab_file.read().strip().split('\n')
+
+    # Insert the sentence boundary tokens which elmo uses at positions 1 and 2.
+    tokens = [tokens[0]] + ["<S>", "</S>"] + tokens[1:]
```
matt-peters: Can we throw an exception if `tokens[0] != '@@UNKNOWN@@'` ?

scripts/create_elmo_embeddings_from_vocab.py
```diff
@@ -0,0 +1,112 @@
...
+
+        all_embeddings.append(per_word_embeddings)
+
+    # Remove the embeddings associated with padding in the last batch.
+    all_embeddings[-1] = all_embeddings[-1][:-last_batch_remainder, :]
+
+    embedding_weight = torch.cat(all_embeddings, 0).numpy()
+
+    # Write out the embedding in a glove format.
+    with gzip.open(os.path.join(output_dir, "elmo_embeddings.tar.gz"), 'wb') as embeddings_file:
```
matt-peters: I don't see a tar component here so should be be `"elmo_embeddings.txt.gz"` ?

PR #580, opened 2017-12-06
Removed arrays_to_variables

matt-gardner:
As we discussed this morning, this moves the logic of creating `Variables` into the `Fields` themselves, instead of having `arrays_to_variables` have to convert numpy to pytorch after the fact.  I had to touch a whole bunch of places in the code, so this PR looks big, but most of the test changes are just simple renamings and such.  The main change is that all of the `as_array*` methods are now `as_tensor*` methods, and the arguments that were passed to `arrays_to_variables` are now passed to `as_tensor*` (including needing to go through the `DataIterator`).  I'm not thrilled with having to pass around `cuda_device` and `for_training` everywhere, but I'm not sure there's a better solution.


allennlp/models/model.py
```diff
@@ -109,21 +109,7 @@ def forward_on_instance(self, instance: Instance, cuda_device: int) -> Dict[str,
         and returns the result.  Before returning the result, we convert any ``torch.autograd.Variables``
         or ``torch.Tensors`` into numpy arrays and remove the batch dimension.
         """
-        instance.index_fields(self.vocab)
```
DeNeutoy: 👍

allennlp/data/iterators/data_iterator.py
```diff
@@ -35,13 +38,20 @@ def __call__(self,
         shuffle : ``bool``, optional (default=``True``)
             If ``True``, we will shuffle the instances in ``dataset`` before constructing batches
             and iterating over the data.
+        cuda_device : ``int``
```
DeNeutoy: I don't _think_ we should do device stuff here. If we do, I think it will be harder to switch to using multiple GPUs.

The problem is the following: to do multi-gpu stuff in pytorch, your batch is split along the batch dimension and pytorch handles sending it to the GPUs. This definitely needs to happen in trainer, so it makes doing any device manipulation here not that feasible. The annoying thing is that this basically means we need to have a recursive `tensor_to_gpu` function ....

matt-gardner: What if we had a `MultiGpuDataIterator` that rotated the devices used for each batch, and the `Trainer` picked out as many batches as it needed?  Would that solve the problem?  That would be a pretty minimal change from this API.

Either way, I think that's an issue to solve when we implement multi-gpu stuff.  I don't think there's anything I can change in this PR to make that easier later - we need to know what the multi-gpu solution is actually going to look like before we can make the right change to how device info is passed around.

allennlp/models/model.py
matt-gardner: This actually drives home @joelgrus's earlier point that maybe we should just have the caller do this, instead of having another method for it.  But that's ok, we don't want to break the API at this point.

matt-peters:
Why is it necessary to move the numpy -> pytorch conversion inside the Fields? This is a large API change that has a few significant negative drawbacks.  It ties the data API in allennlp to pytorch and a particular way of using the resulting tensors on a single GPU whereas previously it was a general purpose NLP data API.

matt-gardner:
The problem is that `arrays_to_variables` doesn't have enough information to do its job when data types get complex.  Particularly, different decisions need to be made around batching with different data structures, and the information you need to make those decisions was lost by the time we got to `arrays_to_variables`.

Mark, Joel and I discussed the options here yesterday, and this seemed like the best solution.  This removes `arrays_to_variables` entirely and just creates variables to start with.

Pros:
- The API actually now works for complex data types, where the previous solution was fundamentally untenable.
- No more overhead of first creating numpy arrays then converting them to pytorch variables.
- Removing an intermediate API call makes the API simpler.

Cons:
- We now return pytorch `Variables` instead of numpy arrays, so it's harder to use this outside of pytorch.  This isn't actually that big of a deal, because you can get back numpy arrays by just calling `.data.cpu().numpy()`.  Also, while interoperability is nice, and trying to be sure the data code was interoperable led to some nice API decisions early on, it's not a high priority.  In particular, we prefer an API that actually does what over an API that's natively compatible with systems that probably very few people will actually use it for.

Notably, the single vs. multi GPU situation isn't changed by this PR.  It was hard before, and it's still hard, and this doesn't really change anything there, it just moves the lines of code that would have had to be changed to somewhere else.

matt-gardner:
Also, it wouldn't be that hard to add a flag like `use_numpy` that you pass from the data iterator all the way down to where things are created.  If people are clamoring to use this with tensorflow or something else and really want a marginal speed increase, we can consider adding that flag.

matt-peters:
Thanks for the additional context, that makes sense.  Just curious, what are the complex data types that can't be handled with the old method?

matt-gardner:
https://github.com/allenai/allennlp/pull/573/files

PR #586, opened 2017-12-07
fix decode in encoder decoder

DeNeutoy:


DeNeutoy:
@maxdel This will likely be relevant for you too if you are using the encoder decoder stuff.

allennlp/models/encoder_decoders/simple_seq2seq.py
```diff
@@ -280,8 +280,6 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
             predicted_tokens = [self.vocab.get_token_from_index(x, namespace="target_tokens")
                                 for x in indices]
             all_predicted_tokens.append(predicted_tokens)
-        if len(all_predicted_tokens) == 1:
```
matt-gardner: This is removing the batch dimension, which we don't do here anymore?  Is that the issue?

DeNeutoy: Yep, exactly. Doing this results in the decoded output being the first word if there is a single element in the batch.

maxdel: Exactly, I also fixed this issue in my local version.

PR #587, opened 2017-12-07
Fix typo in simple_seq2seq.py docstring

maxdel:


matt-gardner:
Thanks again!  We wrote this just as an example, but didn't actually use it for anything.  Thanks for finding the problems =).

PR #588, opened 2017-12-07
Allow ELMo to accept higher dimensional input

matt-peters:
@tusharkhot

### DIFF
--- a/tests/modules/elmo_test.py
+++ b/tests/modules/elmo_test.py
@@ -138,7 +138,7 @@ class TestElmo(AllenNlpTestCase):
 
     def test_elmo_4D_input(self):
         sentences = [[['The', 'sentence', '.'],
-                     ['ELMo', 'helps', 'disambiguate', 'ELMo', 'from', 'Elmo', '.']],
+                      ['ELMo', 'helps', 'disambiguate', 'ELMo', 'from', 'Elmo', '.']],
                      [['1', '2'], ['1', '2', '3', '4', '5', '6', '7']],
                      [['1', '2', '3', '4', '50', '60', '70'], ['The']]]
 
--- a/tests/modules/elmo_test.py
+++ b/tests/modules/elmo_test.py
@@ -97,14 +97,13 @@ class TestElmoBiLm(AllenNlpTestCase):
 
 
 class TestElmo(AllenNlpTestCase):
-    def test_elmo(self):
-        # load the test model
+    def setUp(self):
+        super(TestElmo, self).setUp()
+
         options_file = os.path.join(FIXTURES, 'options.json')
         weight_file = os.path.join(FIXTURES, 'lm_weights.hdf5')
-        elmo = Elmo(options_file, weight_file, 2)
+        self.elmo = Elmo(options_file, weight_file, 2)
 
-        # Correctness checks are in ElmoBiLm and ScalarMix, here we just add a shallow test
-        # to ensure things execute.
         indexer = ELMoTokenCharactersIndexer()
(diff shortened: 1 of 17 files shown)


tusharkhot:
Thanks! This would help to embed a `ListField[TextField]`. I tested it out on my code and you would also need to test for `j>2` here:
```
    for i, j in enumerate(sequence_lengths):
        tensor_without_boundary_tokens.data[i, :(j - 2), :] = tensor.data[i, 1:(j - 1), :]
        new_mask[i, :(j - 2)] = 1
```
https://github.com/allenai/allennlp/blob/master/allennlp/nn/util.py#L832-L834

Normally you would never have empty strings but a `ListField[TextField]` would have padded empty TextFields. It seems boundary sentence ids are still getting added to these empty text fields and so they still contain the two begin and end boundary tokens i.e. j = 2 in the snippet above.

matt-peters:
Thanks, that's a corner case I hadn't tested.  Let me know if the latest commit doesn't fix it for you.

tests/modules/elmo_test.py
```diff
@@ -131,6 +133,18 @@ def test_elmo(self):
         assert list(elmo_representations[1].size()) == [2, 7, 32]
         assert list(mask.size()) == [2, 7]
 
+    def test_elmo_4D_input(self):
```
DeNeutoy: Can you change this test to use an expanded input with a size other than 1? That would be more robust.

An even better test would be to check that spitting up the tensor into 3D tensors and running it through elmo results in the same thing as the 4D version.

tests/modules/elmo_test.py
```diff
@@ -131,6 +133,18 @@ def test_elmo(self):
...
+        # Add an extra dimension to character_ids
+        character_ids = self.character_ids.unsqueeze(1)
+        print(character_ids.size())
```
DeNeutoy: Remove print statement.

tests/modules/elmo_test.py
```diff
@@ -131,6 +133,18 @@ def test_elmo(self):
...
+
+        output = self.elmo(character_ids)
+        elmo_representations = output['elmo_representations']
+        mask = output['mask']
+
+        assert list(elmo_representations[0].size()) == [2, 1, 7, 32]
+        assert list(elmo_representations[1].size()) == [2, 1, 7, 32]
+        assert list(mask.size()) == [2, 1, 7]
```
DeNeutoy: For the mask, you can actually check that the elements are what you expect, rather than the shape. Obviously this is harder for the tensors, so no worries about that.

tests/nn/util_test.py
```diff
@@ -595,19 +595,18 @@ def test_add_sentence_boundary_token_ids_handles_3D_input(self):
     def test_remove_sentence_boundaries(self):
         tensor = Variable(torch.from_numpy(numpy.random.rand(3, 5, 7)))
         mask = Variable(torch.from_numpy(
-                numpy.array([[1, 1, 1, 0, 0],
```
DeNeutoy: What does this change do?

DeNeutoy: Oh is it testing that it works if you have only `<S>` and `</S>` tokens? If so, can you add that as a comment.

matt-peters: Yes, that's it.  Added a comment.

tests/modules/elmo_test.py
matt-peters: Updated the test to run both 3D and 4D versions and compare.

PR #600, opened 2017-12-08
Grad checks

### LINKED ISSUE #456: Use model(**inputs) rather than model.forward(**inputs)
Follow up from email discussion, we should change usages of `model.forward(**inputs)` to `model(**inputs)` as the invoked `__call__` method for `torch.nn.Module` also includes some handling of hooks (code [here](https://github.com/pytorch/pytorch/blob/53b01527f446c1940e2dc1c2023575e51d93919c/torch/nn/modules/module.py#L316)). 

I don't think there are currently known issues from this, so low urgency, but could arise, say, if adding hooks to a model and then running predictions.

### LINKED ISSUE #494: Add a check that all trainable parameters are actually updated
Idea from [this article](https://medium.com/@keeper6928/how-to-unit-test-machine-learning-code-57cf6fd81765).  This makes sure that your model is hooked up correctly, basically.  The check belongs in [`ModelTestCase.ensure_model_can_train_save_and_load`](https://github.com/allenai/allennlp/blob/3c761ba9d857677efb60ae5c2dbbe321f0b3b702/allennlp/common/testing/model_test_case.py#L32), but it's a little tricky because currently that test has no access to the model _before_ training, only _after_ training, and this check requires a comparison of the two states.

Also, there may be some models in the future that want to disable this check, e.g., because they are doing multi-task learning and only some of the model is trained at any particular point.  So if/when we add this, it'd be nice if it were also easy to disable it for any particular model's test.

DeNeutoy:
- Checks that parameters are getting gradients which need them in the model test cases.
- Removes all `.forward` references as Pytorch 0.3 has our bug fixes for modules which return dictionaries
- Fixes a bug in the sentence boundary function for ELMo which was not correctly recording inplace operations in the `Variable`

Fixes #456 
Fixes #494

### DIFF
--- a/allennlp/common/testing/model_test_case.py
+++ b/allennlp/common/testing/model_test_case.py
@@ -1,5 +1,6 @@
 import os
 
+import numpy
 from numpy.testing import assert_allclose
 import torch
 
@@ -46,6 +47,15 @@ class ModelTestCase(AllenNlpTestCase):
         reader = DatasetReader.from_params(params['dataset_reader'])
         iterator = DataIterator.from_params(params['iterator'])
 
+
+        model_dataset = reader.read(params['validation_data_path'])
+        model_dataset.index_instances(model.vocab)
+        model_batch = next(iterator(model_dataset, shuffle=False, cuda_device=cuda_device))
+
+        # Check gradients are None for non-trainable parameters and check that
+        # trainable parameters receive some gradient if they are trainable.
+        self.check_model_computes_gradients_correctly(model, model_batch)
+
         # We'll check that even if we index the dataset with each model separately, we still get
         # the same result out.
         model_dataset = reader.read(params['validation_data_path'])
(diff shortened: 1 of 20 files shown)


allennlp/common/testing/model_test_case.py
```diff
@@ -46,6 +46,14 @@ def ensure_model_can_train_save_and_load(self, param_file: str, tolerance: float
         reader = DatasetReader.from_params(params['dataset_reader'])
         iterator = DataIterator.from_params(params['iterator'])
 
+        model_dataset = reader.read(params['validation_data_path'])
```
matt-gardner: Do you need to duplicate these three lines?  Can you just move the gradient check to later in the method, after these objects are already created?

allennlp/common/testing/model_test_case.py
```diff
@@ -103,16 +111,33 @@ def assert_fields_equal(field1, field2, tolerance: float = 1e-6, name: str = Non
...
+    @staticmethod
+    def check_model_computes_gradients_correctly(model, model_batch):
+        model.zero_grad()
+        result = model(**model_batch)
+        result["loss"].backward()
+
+        for name, parameter in model.named_parameters():
+            zeros = torch.zeros(parameter.size())
+            if parameter.requires_grad:
+                print(parameter)
```
matt-gardner: Remove print statements.

DeNeutoy:
Yes it did - it clarified that no gradient was getting computed, rather than the gradient being zero, which is a subtle but important correctness issue (i.e you can have correct gradients which are zero, but if they aren't computed, that is obviously wrong). Thanks for the advice!

PR #605, opened 2017-12-11
Stacked self attention

DeNeutoy:


allennlp/nn/util.py
```diff
@@ -835,3 +835,58 @@ def remove_sentence_boundaries(tensor: torch.Tensor,
             new_mask[i, :(j - 2)] = 1
 
     return tensor_without_boundary_tokens, new_mask
+
+
+def add_positional_features(tensor: torch.Tensor,
```
DeNeutoy: @matt-gardner any good ideas for how to test this?

matt-gardner: Ugh, seems tricky.  Two thoughts:

1. You could compute some values by hand for a couple of different cases, and test that the hand-computed values match what the function produces.
2. Instead of computing the values by hand, compute them using the `tensor2tensor` code, and use those values here in the test, with a note that you got them from the original implementation.

If option 2 is feasible, I'd recommend doing that.

DeNeutoy:
Ok @matt-gardner this is ready for review - I seriously can't work out what i'm doing wrong to enrage Sphinx ...

joelgrus:
my guess is it doesn't like

```
    2. Multi-headed self attention, which uses 2 learnt linear projections
     to perform a dot-product similarity between every pair of elements
     scaled by the square root of the sequence length.
```

and you should indent the second and third lines a couple of spaces more to line up with the first, but that's just a guess

DeNeutoy:
I tried that I think .... maybe it was before I switched to numbers though. Also are you joking? What do Americans use instead of learnt?

joelgrus:
http://grammarist.com/spelling/learned-learnt/

allennlp/modules/seq2seq_encoders/intra_sentence_attention.py
```diff
@@ -43,13 +44,16 @@ class IntraSentenceAttentionEncoder(Seq2SeqEncoder):
         :func:`~allennlp.nn.util.combine_tensors`; see that function for more detail on exactly how
         this works, but some simple examples are ``"1,2"`` for concatenation (the default),
         ``"1+2"`` for adding the two, or ``"2"`` for only keeping the attention representation.
+    output_projection_dim : ``bool``, optional (default = None)
```
matt-gardner: Maybe just call this `output_dim`, with the docstring saying that if it's given, we'll do a projection to this dimension?

allennlp/modules/seq2seq_encoders/stacked_self_attention.py
```diff
@@ -0,0 +1,159 @@
...
+        super(StackedSelfAttentionEncoder, self).__init__()
+
+        self._use_positional_encoding = use_positional_encoding
+        self._attention_layers: List[IntraSentenceAttentionEncoder] = []
+        self._feedfoward_layers: List[FeedForward] = []
+        self._layer_norm_layers: List[LayerNorm] = []
+
+        feedfoward_input_dim = input_dim
+        for i in range(num_layers):
+            # Project output of attention encoder through a feedforward network
```
matt-gardner: The way this is worded it makes me think you're doing the operation here, but it actually happens in `forward`.

allennlp/modules/layer_norm.py
```diff
@@ -0,0 +1,25 @@
+
+import torch
+
+
+class LayerNorm(torch.nn.Module):
```
matt-gardner: Should this be `LayerNormalization`?  I could go either way here.  "norm" is pretty standard.

allennlp/modules/seq2seq_encoders/stacked_self_attention.py
```diff
@@ -0,0 +1,159 @@
...
+    def __init__(self,
+                 input_dim: int,
+                 hidden_dim: int,
+                 projection_dim: int,
+                 non_linear_hidden_dim: int,
+                 num_layers: int,
+                 num_attention_heads: int,
+                 internal_similarity: SimilarityFunction = DotProductSimilarity(scale_output=True),
+                 use_positional_encoding: bool = True,
+                 dropout_prob: float = 0.2) -> None:
```
matt-gardner: This is not documented above.

allennlp/modules/seq2seq_encoders/stacked_self_attention.py
```diff
@@ -0,0 +1,159 @@
...
+    def get_input_dim(self) -> int:
+        return self._input_dim
+
+    @overrides
+    def get_output_dim(self) -> int:
+        return self._output_dim
+
+    def forward(self, inputs: torch.Tensor, mask: torch.Tensor): # pylint: disable=arguments-differ
+        if self._use_positional_encoding:
+            output = add_positional_features(inputs)
```
matt-gardner: I'd prefer `inputs = add_positional_features` here, dropping the `else` clause, and using `cached_input = inputs` below.

allennlp/modules/seq2seq_encoders/stacked_self_attention.py
```diff
@@ -0,0 +1,159 @@
...
+        else:
+            output = inputs
+        for index, (attention, feedforward, layer_norm) in enumerate(zip(self._attention_layers,
+                                                                         self._feedfoward_layers,
+                                                                         self._layer_norm_layers)):
+            cached_input = output
+            # shape (batch_size, timesteps, input_size)
+            non_linear_output = feedforward(output)
```
matt-gardner: `feedforward_output`, for consistent naming?

allennlp/modules/seq2seq_encoders/stacked_self_attention.py
```diff
@@ -0,0 +1,159 @@
...
+            if index != 0:
+                # First layer might have the wrong size for highway
```
matt-gardner: This is if `self._input_dim != hidden_dim`, right?  Maybe just add a check for that, and give a big note above in the docstring where these things are configured?  Having a highway layer from the input (which is the only place where you add positional encoding) seems like it could be pretty important.

allennlp/modules/seq2seq_encoders/stacked_self_attention.py
matt-gardner: Hmm, nevermind, that breaks the loop.

allennlp/nn/util.py
```diff
@@ -835,3 +835,57 @@ def remove_sentence_boundaries(tensor: torch.Tensor,
...
+                            min_timescale: float = 1.0,
+                            max_timescale: float = 1.0e4):
+    # pylint: disable=line-too-long
+    """
+    Implements the frequency based positional encoding described
```
matt-gardner: `frequency-based`

allennlp/nn/util.py
```diff
@@ -835,3 +835,57 @@ def remove_sentence_boundaries(tensor: torch.Tensor,
...
+    in `Attention is all you Need
+    <https://www.semanticscholar.org/paper/Attention-Is-All-You-Need-Vaswani-Shazeer/0737da0767d77606169cbf4187b83e1ab62f6077>`_ .
+
+    Adds sinusoids of different frequencies to a ``Tensor``. Each dimension of
+    the input ``Tensor`` is added to a sinusoid of a different frequency and phase.
```
matt-gardner: Pretty sure the arguments are swapped?  "A sinusoid ... is added to each dimension ...", right?

allennlp/nn/util.py
```diff
@@ -835,3 +835,57 @@ def remove_sentence_boundaries(tensor: torch.Tensor,
...
+    This allows the attention heads to use absolute and relative positions.
+
+    The number of timescales is equal to hidden_dim / 2 within the range
+    (min_timescale, max_timescale). For each timescale, the two sinusoidal
+    signals  sin(timestep / timescale) and cos(timestep / timescale) are
```
matt-gardner: Extra space before "sin()"

tests/nn/util_test.py
```diff
@@ -613,3 +613,16 @@ def test_remove_sentence_boundaries(self):
...
+
+    def test_add_positional_features(self):
+        # This is hard to test, so we check that we get the same result as the
+        # original tensorflow implementation:
+        # https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/layers/common_attention.py#L270
+        tensor2tensor_result = numpy.asarray([[0.00000000e+00, 0.00000000e+00, 1.00000000e+00, 1.00000000e+00],
+                                              [8.41470957e-01, 9.99999902e-05, 5.40302277e-01, 1.00000000e+00],
+                                              [9.09297407e-01, 1.99999980e-04, -4.16146845e-01, 1.00000000e+00]])
+
+        tensor = Variable(torch.zeros([2, 3, 4]))
```
matt-gardner: Can you add a test here with `[2, 3, 5]`, so we get a test with an odd hidden dim?

DeNeutoy:
@matt-gardner can you take another look at this? I've added the efficient version of the attention, but I left the self attention layer you added in still - probably best until we figure out what we should do with the attention abstraction.

allennlp/modules/seq2seq_encoders/efficient_multi_head_attention.py
```diff
@@ -0,0 +1,159 @@
...
+from torch.nn import Dropout, Linear
+from torch.nn import Parameter
+from torch.nn import init
+
+from allennlp.nn.util import last_dim_softmax
+from allennlp.modules.seq2seq_encoders.seq2seq_encoder import Seq2SeqEncoder
+
+
+@Seq2SeqEncoder.register("efficient_multi_head_attention")
+class EfficientMultiHeadAttention(Seq2SeqEncoder):
```
matt-gardner: I'd probably just call this `MultiHeadSelfAttention`.  We should make the other case the one with the longer name, and make it clear it should only be used when you want to explore similarity options (and we can do that change in a different PR).

allennlp/modules/seq2seq_encoders/efficient_multi_head_attention.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        num_heads = self._num_heads
+
+        batch_size, timesteps, hidden_dim = inputs.size()
+        if mask is None:
+            mask = Variable(inputs.data.new(batch_size, timesteps).fill_(1.0))
+
+        # Treat the queries, keys and values each as a ``num_heads`` size batch.
+        # shape (num_heads, batch_size * timesteps, hidden_dim)
+        inputs_per_head = inputs.repeat(num_heads, 1, 1).view(num_heads,
```
matt-gardner: `repeat` copies data, while `expand` only provides a view.  Do you really need `repeat` here?

allennlp/modules/seq2seq_encoders/efficient_multi_head_attention.py
```diff
@@ -0,0 +1,159 @@
...
+        # shape (num_heads * batch_size, timesteps, attention_dim)
+        values_per_head = values_per_head.view(num_heads * batch_size, timesteps, self._values_dim)
+
+        # shape (num_heads * batch_size, timesteps, timesteps)
+        scaled_similarities = torch.bmm(queries_per_head, keys_per_head.transpose(1, 2)) / self._scale
+
+        # shape (num_heads * batch_size, timesteps, timesteps)
+        # Normalise the distributions, using the same mask for all heads.
+        attention = last_dim_softmax(scaled_similarities, mask.repeat(num_heads, 1))
+        attention = self._attention_dropout(attention)
```
matt-gardner: I don't think I buy that this is still a well-defined probability distribution.  `attention.sum(dim=-1)` after line 138 should sum to 1, and `attention.sum(dim=-1)` will necessarily be less than one with non-zero dropout.  What am I missing?

allennlp/modules/seq2seq_encoders/efficient_multi_head_attention.py
```diff
@@ -0,0 +1,159 @@
...
+        # the attention distributions for each element in the num_heads * batch_size
+        # dimension.
+        # shape (num_heads * batch_size, timesteps, values_dim)
+        outputs = torch.bmm(attention, values_per_head)
+
+        # Reshape back to original shape (batch_size, timesteps, num_heads * values_dim)
+        # Note that we _cannot_ use a reshape here, because this tensor was created
+        # with num_heads being the first dimension, so reshaping naively would not
+        # throw an error, but give an incorrect result.
+        # TODO(Mark): Given the possible slowness of split, maybe reshape + transpose is better?
```
matt-gardner: Yeah, my intuition says that a reshape + transpose should be a lot faster than a split + cat.  Can you run some simple timing tests to see if it makes a difference?

tests/modules/seq2seq_encoders/efficient_multi_head_attention_test.py
```diff
@@ -0,0 +1,33 @@
...
+    def test_multi_head_attention_runs_forward(self):
+        attention = EfficientMultiHeadAttention(num_heads=3,
+                                                input_dim=5,
+                                                attention_dim=7,
+                                                values_dim=9)
+        inputs = Variable(torch.randn(2, 12, 5))
+        attention(inputs)
+
+    def test_multi_head_attention_respects_masking(self):
+        attention = EfficientMultiHeadAttention(num_heads=3,
```
matt-gardner: Is there an easy way to test for the reshape issue you mentioned above?  Probably not...

Maybe if you set the weights by hand and gave some simple inputs...

allennlp/modules/seq2seq_encoders/efficient_multi_head_attention.py
DeNeutoy: You're right, but I found the following thing which terrifies me a bit with `expand`, try this:
```
x = torch.FloatTensor([2])
x.repeat(1,3).mul_(2) # == [4, 4, 4]
x.expand(1,3).mul_(2) # == [16, 16, 16] !!
```
This is apparently the expected behaviour and a relic of how Torch used to work. I didn't know this and it makes me want to remove all inplace operations in the entire library.

allennlp/modules/seq2seq_encoders/efficient_multi_head_attention.py
DeNeutoy: Yeah you're right, thanks.

allennlp/modules/seq2seq_encoders/efficient_multi_head_attention.py
matt-gardner: Wow, that is really scary.  But we don't do any in place operations at all except for `fill_`, do we?  They break autograd.

allennlp/modules/seq2seq_encoders/efficient_multi_head_attention.py
DeNeutoy: Also this seems a bit mad given this isn't the case? What does doing this even do?

allennlp/modules/seq2seq_encoders/efficient_multi_head_attention.py
DeNeutoy: I think pytorch expands and correctly tracks the operations actually, but i'll change this here.

allennlp/modules/seq2seq_encoders/efficient_multi_head_attention.py
matt-gardner: Does the transformer paper do this?  I just checked their paper and it wasn't clear to me whether they had this particular dropout here or not.  It doesn't seem particularly problematic - you're just saying that the model should be robust to having some of its dependent words missing.  It's just problematic to say that it's still a normalized distribution.

allennlp/modules/seq2seq_encoders/efficient_multi_head_attention.py
DeNeutoy: Actually, this has to be a `repeat` (or it's at least equivalent equivalent), because you can only `view` contiguous tensors, which expanded things aren't. Or else you end up having to call `.contiguous()` which I assume requires just as much memory as actually allocating memory initially.

matt-gardner: I wondered originally if that was why you wrote it that way.  Now we know =).  Thanks for checking.

allennlp/modules/seq2seq_encoders/efficient_multi_head_attention.py
DeNeutoy: I tried this, and actually it's not slower, because to get it in the right shape for a `view` I need to `view, contiguous, permute, contiguous and view` it.

matt-gardner: Interesting, that's good to know.  Thanks again for checking.  It'll help us get a bit of a better intuition for which operations we should be afraid of.

PR #606, opened 2017-12-12
Use production rule field

### LINKED ISSUE #499: Add a simple "check dimensions match" method
We currently have [lots of places](https://github.com/allenai/allennlp/blob/master/allennlp/models/reading_comprehension/bidaf.py#L100-L118) in our model init methods that take 5 lines to do a simple dimension check, and it's onerous to add those checks, so they don't get added a lot of the time.  We should have a simple function to call that checks that two dimensions match, and throws a `ConfigurationError` if they don't.

matt-gardner:
This PR modifies the `WikiTablesDatasetReader` to use the new `ProductionRuleField`, and the `WikiTablesSemanticParser` to also take the new data format.  It's close, but not quite done.  I ran into a bit of a snag when trying to produce lambda expressions.

The gist of how this works is that we enumerate all possible actions up front, in the `DatasetReader`, so that we can get indexed representations of every possible action.  Then, in the model, we'll embed those representations before decoding, and just _reference_ the embedded vectors when we need them during decoding.  I had a few issues getting a representation for all of the possible actions, and I was able to resolve most of them, but the lambda expressions are giving me trouble.  I'll put a note in the code for the particular spot that's the issue; @pdasigi, if you have any ideas for how to fix this, that'd be great.

Fixes #499

### DIFF
--- a/allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
+++ b/allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
@@ -184,8 +184,7 @@ class WikiTablesDatasetReader(DatasetReader):
                 print(expression)
                 print(sequence)
                 for production_rule in sequence:
-                    field = IndexField(action_map[production_rule], action_field)
-                    index_fields.append(field)
+                    index_fields.append(IndexField(action_map[production_rule], action_field))
                 action_sequence_fields.append(ListField(index_fields))
             fields['target_action_sequences'] = ListField(action_sequence_fields)
         return Instance(fields)
--- a/allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
+++ b/allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
@@ -132,9 +132,10 @@ class WikiTablesDatasetReader(DatasetReader):
         else:
             table_knowledge_graph = TableKnowledgeGraph.read_from_json(table_info)
         table_field = KnowledgeGraphField(table_knowledge_graph, self._table_token_indexers)
-        fields = {'question': question_field, 'table': table_field}
+        world = WikiTablesWorld(table_knowledge_graph)
+        world_field = MetadataField(world)
+        fields = {'question': question_field, 'table': table_field, 'world': world_field}
         if dpd_output:
-            world = WikiTablesWorld(table_knowledge_graph)
             expressions = [world.parse_logical_form(form) for form in dpd_output]
             action_sequences = [world.get_action_sequence(expression) for expression in expressions]
             action_sequence_fields: List[ListField[ProductionRuleField]] = []
--- a/allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
+++ b/allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
@@ -19,8 +19,9 @@ from allennlp.data.dataset import Dataset
 from allennlp.data.instance import Instance
 from allennlp.data.tokenizers import Tokenizer, WordTokenizer
 from allennlp.data.token_indexers import TokenIndexer, SingleIdTokenIndexer
-from allennlp.data.fields import TextField, KnowledgeGraphField, LabelField, ListField
+from allennlp.data.fields import KnowledgeGraphField, ListField, ProductionRuleField, TextField
 from allennlp.data.semparse.knowledge_graphs import TableKnowledgeGraph
+from allennlp.data.semparse.type_declarations import wikitables_type_declaration as wt_types
 from allennlp.data.semparse.worlds import WikiTablesWorld
 from allennlp.data.dataset_readers.dataset_reader import DatasetReader
 from allennlp.data.dataset_readers.seq2seq import START_SYMBOL, END_SYMBOL
@@ -76,6 +77,7 @@ class WikiTablesDatasetReader(DatasetReader):
         self._tokenizer = tokenizer or WordTokenizer()
         self._question_token_indexers = question_token_indexers or {"tokens": SingleIdTokenIndexer()}
         self._table_token_indexers = table_token_indexers or self._question_token_indexers
+        self._basic_types = set(str(type_) for type_ in wt_types.BASIC_TYPES)
 
     @overrides
     def read(self, file_path):
@@ -135,16 +137,19 @@ class WikiTablesDatasetReader(DatasetReader):
             world = WikiTablesWorld(table_knowledge_graph)
             expressions = [world.parse_logical_form(form) for form in dpd_output]
             action_sequences = [world.get_action_sequence(expression) for expression in expressions]
-            action_sequences_field = ListField([self._make_action_sequence_field(sequence)
-                                                for sequence in action_sequences])
-            fields['target_action_sequences'] = action_sequences_field
+            action_sequence_fields: List[ListField[ProductionRuleField]] = []
+            for sequence in action_sequences:
+                production_rule_fields: List[ProductionRuleField] = []
+                for production_rule in sequence:
+                    field = ProductionRuleField(production_rule,
+                                                terminal_indexers=self._terminal_indexers,
+                                                nonterminal_indexers=self._non_terminal_indexers,
+                                                nonterminal_types=self._basic_types,
+                                                context=tokenized_question)
+                    production_rule_fields.append(field)
+                action_sequence_fields.append(ListField(production_rule_fields))
+            fields['target_action_sequences'] = ListField(action_sequence_fields)
         return Instance(fields)
-
-    @staticmethod
-    def _make_action_sequence_field(action_sequence: List[str]) -> ListField:
-        action_sequence.insert(0, START_SYMBOL)
-        action_sequence.append(END_SYMBOL)
-        return ListField([LabelField(action, label_namespace='actions') for action in action_sequence])
 
     @staticmethod
     def _parse_line_as_lisp(lisp_string: str) -> Dict[str, Union[str, List[str], None]]:
--- a/allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
+++ b/allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
@@ -2,7 +2,7 @@
 Reader for WikitableQuestions (https://github.com/ppasupat/WikiTableQuestions/releases/tag/v1.0.2).
 """
 
-from typing import Any, Dict, List, Union
+from typing import Dict, List, Union
 import gzip
 import logging
 import os
@@ -23,7 +23,6 @@ from allennlp.data.fields import Field, IndexField, KnowledgeGraphField, ListFie
 from allennlp.data.fields import MetadataField, ProductionRuleField, TextField
 from allennlp.data.semparse.knowledge_graphs import TableKnowledgeGraph
 from allennlp.data.semparse.type_declarations import wikitables_type_declaration as wt_types
-from allennlp.data.semparse.type_declarations import type_declaration as types
 from allennlp.data.semparse.worlds import WikiTablesWorld
 from allennlp.data.dataset_readers.dataset_reader import DatasetReader
 
@@ -135,7 +134,7 @@ class WikiTablesDatasetReader(DatasetReader):
         ----------
         question : ``str``
             Input question
-        table_info : ``str`` or ``Dict[str, Any]``
+        table_info : ``str`` or ``JsonDict``
             Table filename or the table content itself, as a dict. See
             ``TableKnowledgeGraph.read_from_json`` for the expected format.
         dpd_output : List[str] (optional)
@@ -178,11 +177,8 @@ class WikiTablesDatasetReader(DatasetReader):
             # ProductionRuleFields.
             action_map = {action.rule: i for i, action in enumerate(action_field.field_list)}  # type: ignore
             action_sequence_fields: List[Field] = []
-            for form, expression, sequence in zip(dpd_output, expressions, action_sequences):
+            for sequence in action_sequences:
                 index_fields: List[Field] = []
-                print(form)
-                print(expression)
-                print(sequence)
                 for production_rule in sequence:
                     index_fields.append(IndexField(action_map[production_rule], action_field))
                 action_sequence_fields.append(ListField(index_fields))
--- a/allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
+++ b/allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
@@ -2,7 +2,7 @@
 Reader for WikitableQuestions (https://github.com/ppasupat/WikiTableQuestions/releases/tag/v1.0.2).
 """
 
-from typing import Dict, List, Union
+from typing import Any, Dict, List, Union
 import gzip
 import logging
 import os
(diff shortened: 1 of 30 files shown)


allennlp/data/semparse/type_declarations/type_declaration.py
```diff
@@ -396,18 +420,33 @@ def get_valid_actions(name_mapping: Dict[str, str],
...
+            print(complex_type)
             _get_placeholder_actions(complex_type, basic_types, valid_actions)
         else:
+            # We can produce complex types with a lambda expression, though we'll leave out
+            # placeholder types for now.
+            for i in range(num_nested_lambdas):
+                lambda_var = chr(ord('x') + i)
+                production_string = _make_production_string(complex_type.first,
+                                                            ['lambda ' + lambda_var, complex_type.second])
+                valid_actions[str(complex_type)].add(production_string)
```
matt-gardner: If you run the tests from this branch (or look at the printed output on travis, probably), you can see the action sequences we're trying to produce.  We want to give action ids for the actions in the target sequence, where the ids are positions in the list of all actions.  But there's an action that is `<r,d> -> ['lambda x', d]`.  The trouble is that `<r,d>` isn't in the type declaration; it comes from `<d,r> -> [<<#1,#2>,<#2,#1>>, <r,d>]`.  So how do I get that rule into the list of actions?

A couple of options:

1. Keep track of the actual type object when creating the `valid_actions` dict, and just add a `lambda` to every key that's a complex type (or maybe every key?).  We need to keep track of the type object because recovering the input and output types from the type string is a pain.
2. Have a mapping from production rule string to something like `LAMBDA` and `VAR` for the lambda and var rules, and have those not depend on the type of the left hand side.

Any other options?  I looked briefly through the data, and it looked like the only place that lambdas showed up was right after `reverse` - it could be that we could hard code something here that would simplify things a bit.

matt-gardner: I'm going to pursue option 1 for now.

pdasigi: I think it would be easy to add `lambda` rules to `type_declaration.get_valid_actions` directly. And I agree that `lambda`s occur in a small set of contexts, but it could be simpler than what you described. I've seen them only within `argmax` and `argmin` where they are used to extract the attributes that need to be maximized or minimized. This means that the productions are always of the form `<b,d> -> ['lambda x', d]`, where `b` is one of the basic types (I don't think it is ever a complex type), and the return type in the head of the production rule is always `d`.

matt-gardner: I've seen them in a few other places, too.  You can see in the current code in this PR how I implemented this, if you want to look at it now.  I'm pretty close to done - just need to finish the action embedding, and put in a few pieces to keep track of lambdas.  Probably one more day of work before I can run the model on the full day.  Hopefully...

matt-gardner:
Ok, @pdasigi, this is now ready for review, finally.  Sorry it's such a massive PR...

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -187,6 +203,151 @@ def forward(self,  # type: ignore
...
+        for action_sequence in actions:
+            for production_rule in action_sequence:
+                # This logic is hard to understand, because the ProductionRuleArray is a messy
+                # type.  The structure of each ProductionRuleArray is:
+                #     {
+                #      "left": (LHS_string, left_is_nonterminal, padded_LHS_tensor_dict),
+                #      "right": (RHS_string, right_is_nonterminal, padded_RHS_tensor_dict)
+                #     }
+                # Technically, the left hand side is _always_ a non-terminal (by definition, you
+                # can't expand a non-terminal), but we'll do this check anyway, in case you did
```
pdasigi: You mean "you can't expand a terminal"?

tests/data/dataset_readers/wikitables/wikitables_dataset_reader_test.py
```diff
@@ -11,17 +12,244 @@ def test_reader_reads(self):
...
+                "<d,<d,<#1,<<d,#1>,#1>>>> -> argmax",
+                "<d,<d,<#1,<<d,#1>,#1>>>> -> argmin",
+                "<d,<d,<<d,d>,d>>> -> [<d,<d,<#1,<<d,#1>,#1>>>>, d]",
+                "<d,<d,d>> -> -",
+                "<d,<e,<<d,e>,e>>> -> [<d,<d,<#1,<<d,#1>,#1>>>>, d]",
+                "<d,<p,<<d,p>,p>>> -> [<d,<d,<#1,<<d,#1>,#1>>>>, d]",
+                "<d,<r,<<d,r>,r>>> -> [<d,<d,<#1,<<d,#1>,#1>>>>, d]",
+
+                # Operations that manipulate numbers.
+                # TODO(mattg): question for Pradeep: why aren't these two-argument functions?
```
pdasigi: They take a number or a date (or a set of numbers), and return the set of ALL dates or numbers that are less than, or greater than the given set. I think the executor determines the set of ALL dates or numbers based on the context. For example, `(fb:row.row.start_date (>= (date 2000 -1 -1)))` returns all the rows where the year in `start_date` is later than `2000`.

tests/data/dataset_readers/wikitables/wikitables_dataset_reader_test.py
```diff
@@ -11,17 +12,244 @@ def test_reader_reads(self):
...
+                # a set, and we're performing some aggregation over the set and returning a single
+                # number (actually a set with a single number in it).
+                "<d,d> -> avg",
+                "<d,d> -> max",
+                "<d,d> -> min",
+                "<d,d> -> sum",
+                "<d,e> -> ['lambda x', e]",
+                "<d,e> -> [<<#1,#2>,<#2,#1>>, <e,d>]",
+                # TODO(mattg): isn't this backwards?  Aren't these functions that take cells and
+                # return numbers?
```
pdasigi: No, they're fine. In general, `fb:a.b` takes `b` and gives `a`. `fb:row.row.index` takes `row.index` and gives `row` (hence `<d,r>`). `fb:cell.cell.date` takes `cell.date` and gives `cell`.Hence `<d,e>`.

tests/data/dataset_readers/wikitables/wikitables_dataset_reader_test.py
```diff
@@ -11,17 +12,244 @@ def test_reader_reads(self):
...
+                "<d,e> -> fb:cell.cell.number",
+                "<d,p> -> ['lambda x', p]",
+                "<d,p> -> [<<#1,#2>,<#2,#1>>, <p,d>]",
+                "<d,r> -> ['lambda x', r]",
+                "<d,r> -> [<<#1,#2>,<#2,#1>>, <r,d>]",
+                "<d,r> -> fb:row.row.index",
+
+                # Now we get to the CELL_TYPE, which is represented by "e".
+                "<e,<<d,e>,e>> -> [<d,<e,<<d,e>,e>>>, d]",
+                # TODO(mattg): why is this so complicated?
```
pdasigi: `date` takes three arguments: year, month and day. For example, `(date 2018 01 06)`.

tests/data/dataset_readers/wikitables/wikitables_dataset_reader_test.py
```diff
@@ -11,17 +12,244 @@ def test_reader_reads(self):
...
+                "<e,e> -> [<#1,<#1,#1>>, e]",
+                "<e,e> -> [<<#1,#2>,<#2,#1>>, <e,e>]",
+                "<e,p> -> ['lambda x', p]",
+                "<e,p> -> [<<#1,#2>,<#2,#1>>, <p,e>]",
+                "<e,r> -> ['lambda x', r]",
+                "<e,r> -> [<<#1,#2>,<#2,#1>>, <r,e>]",
+
+                # These are instance-specific production rules.  These are the columns in the
+                # table.
+                # TODO(mattg): this also seems backwards.
```
pdasigi: Columns are relations from cells to rows.

tests/data/dataset_readers/wikitables/wikitables_dataset_reader_test.py
```diff
@@ -11,17 +12,244 @@ def test_reader_reads(self):
...
+                "@START@ -> p",
+                "@START@ -> r",
+
+                # Now we get to productions for the basic types.  The first part of each section
+                # here has rules for expanding the basic type with function applications; after
+                # that we get to terminal productions (and a "BASIC_TYPE -> x" rule for lambda
+                # variable productions).
+
+                # Date and numbers.  We don't have any terminal productions for these, as for some
+                # reason terminal numbers are actually represented as CELL_TYPE.
```
pdasigi: NLTK assigns `EntityType` to all constants by default, and I let that be the case for numbers. However, I also defined `CELL_TYPE` to be the same as `EntityType` for no good reason. I think I should change the latter.

allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
```diff
@@ -64,18 +65,36 @@ class WikiTablesDatasetReader(DatasetReader):
...
+        How should we represent non-terminals in production rules when we're computing action
+        embeddings?  We use ``TokenIndexers`` for this.  Default is to use a
+        ``SingleIdTokenIndexer`` with the ``rule_labels`` namespace, keyed by ``tokens``:
+        ``{"tokens": SingleIdTokenIndexer("rule_labels")}``.  We use the namespace ``rule_labels``
+        so that we don't get padding or OOV tokens for nonterminals.
+    terminal_indexers : ``Dict[str, TokenIndexer]`` (optional)
+        How should we represent terminals in production rules when we're computing action
+        embeddings?  We also use ``TokenIndexers`` for this.  The default is to use a
+        ``TokenCharactersIndexer`` keyed by ``token_characters``: ``{"token_characters":
+        TokenCharactersIndexer()}``.  We use this indexer by default because WikiTables has plenty
```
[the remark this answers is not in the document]
matt-gardner: Because of how the indexers work, I'm not actually sure that's a good idea.  The way that type embeddings are used in the original parser is different from how it would work if we included the type as a token indexer here.  Maybe that's a good idea for something to try later, after we've replicated the original model, but it would be a different model.

allennlp/data/semparse/type_declarations/grammar_state.py
```diff
@@ -0,0 +1,164 @@
...
+    itself maintains the context-dependent valid actions.
+
+    Parameters
+    ----------
+    nonterminal_stack : ``List[str]``
+        Holds the list of non-terminals that still need to be expanded.  This starts out as
+        [START_SYMBOL], and decoding ends when this is empty.  Every time we take an action, we
+        update the non-terminal stack and the context-dependent valid actions, and we use what's on
+        the stack to decide which actions are valid in the current state.
+    lambda_stack : ``Dict[Tuple[str, str], List[str]]``
```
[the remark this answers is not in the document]
matt-gardner: Done.

allennlp/data/semparse/type_declarations/grammar_state.py
```diff
@@ -0,0 +1,164 @@
...
+        ``nonterminal_stack``.  When the stack becomes empty, we remove the lambda entry.
+    valid_actions : ``Dict[str, List[int]]``
+        A mapping from non-terminals (represented as strings) to all valid (global and
+        instance-specific) productions from that non-terminal (represented as a list of integers).
+    action_indices : ``Dict[str, int]``
+        We use integers to represent productions in the ``valid_actions`` dictionary for efficiency
+        reasons in the decoder.  This means we need a way to map from the production rule strings
+        that we generate for lambda variables back to the integer used to represent it.
+    """
+    lambda_variables = set(['x', 'y', 'z'])
```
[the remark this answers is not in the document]
matt-gardner: Done.

allennlp/data/semparse/type_declarations/grammar_state.py
```diff
@@ -0,0 +1,164 @@
...
+            assert lambda_stack[-1] == left_side
+            lambda_stack.pop()  # pop to modify the value in the dictionary
+
+        productions = self._get_productions_from_string(right_side)
+        if 'lambda' in productions[0]:
+            production = productions[0]
+            if production[0] == "'" and production[-1] == "'":
+                production = production[1:-1]
+            lambda_variable = production.split(' ')[1]
+            if len(left_side) != 5:
```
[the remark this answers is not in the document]
matt-gardner: I do, right below.  I'll move the comment above.

allennlp/data/semparse/type_declarations/grammar_state.py
```diff
@@ -0,0 +1,164 @@
...
+            return [production_string]
+
+    @staticmethod
+    def _is_nonterminal(production: str) -> bool:
+        # TODO(mattg): ProductionRuleField has a similar method, as does another place or two.  We
+        # should centralize this logic somewhere, probably in the type declaration, or the
+        # ``World`` object.  This is pretty specific to the WikiTablesWorld, and to the assumptions
+        # made in converting types to strings (e.g., that we're only using the first letter for
+        # types, lowercased).
+        if production in ['<=', '<']:
```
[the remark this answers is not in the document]
matt-gardner: Those are the "less than" and "less than or equal to" terminal functions.  Added a comment.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -113,10 +114,18 @@ def forward(self,  # type: ignore
             ``KnowledgeGraphField``.  This output is similar to a ``TextField`` output, where each
             entity in the table is treated as a "token", and we will use a ``TextFieldEmbedder`` to
             get embeddings for each entity.
-        target_action_sequences : torch.LongTensor, optional (default = None)
-           A list of possibly valid action sequences, with shape ``(batch_size, num_sequences,
-           sequence_length, 1)``.  The trailing dimension is because of how our ``ListFields``
-           work, and will be stripped in this method with a ``squeeze``.
+        world : ``ListWikiTablesWorld``
```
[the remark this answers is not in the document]

allennlp/data/semparse/worlds/wikitables_world.py
```diff
@@ -21,16 +23,56 @@ class WikiTablesWorld(World):
...
         self._column_counter = 0
 
+        # This adds all of the cell and column names to our local name mapping, so we can get them
+        # as valid actions in the parser.
+        for entity in table_graph.get_all_entities():
+            self._map_name(entity)
+
+        numbers = self._get_numbers_from_tokens(question_tokens) + list(str(i) for i in range(10))
+        for number in numbers:
+            self._add_name_mapping(number, number, types.CELL_TYPE)
```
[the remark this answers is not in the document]
matt-gardner: That actually doesn't work, for reasons I don't entirely understand.  It makes all of the tests fail in NLTK code.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -113,10 +114,18 @@ def forward(self,  # type: ignore
...
+            We use a ``MetadataField`` to get the ``World`` for each input instance.  Because of
+            how ``MetadataField`` works, this gets passed to us as a ``List[WikiTablesWorld]``,
+        actions : ``List[List[ProductionRuleArray]]``
+            A list of all possible actions for each ``World`` in the batch, indexed into a
+            ``ProductionRuleArray`` using a ``ProductionRuleField``.  We will embed all of these
+            and use the embeddings to determine which action to take at each timestep in the
+            decoder.n
```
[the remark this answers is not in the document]
matt-gardner: Done.n

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
matt-gardner: Yep, thanks.

tests/data/dataset_readers/wikitables/wikitables_dataset_reader_test.py
matt-gardner: Thanks.

tests/data/dataset_readers/wikitables/wikitables_dataset_reader_test.py
matt-gardner: Finally these make sense.  Not sure at all why SEMPRE decided to make these backwards, but at least now I know...

tests/data/dataset_readers/wikitables/wikitables_dataset_reader_test.py
matt-gardner: Why do the year, month and day have type `e` instead of `d`?

tests/data/dataset_readers/wikitables/wikitables_dataset_reader_test.py
matt-gardner: Yeah, backwards.  But at least now I know =).

tests/data/dataset_readers/wikitables/wikitables_dataset_reader_test.py
matt-gardner: I added a TODO for this.

Issue #612, opened 2017-12-14
Add instructions for getting AllenNLP to run on Windows


matt-gardner:
It's currently a pain, apparently, and I'm not sure how much work we should do to support it (if you're reading this and want Windows support, add a comment please).  But @OyvindTafjord was able to get it actually working by modifying some things:

- Got pytorch from 'conda install -c peterjc123 pytorch cuda80' (https://github.com/pytorch/pytorch/issues/494)
- Pip install of spacy (and other things) required some VS2015 library stuff (had to be same version as used to compile python).
- Pip install of sanic failed (because a dependency, uvloop, is not available for Windows)  [ got around it by remove serve from commands ]
- In general 'pip install -r requirements.txt' worked weirdly and I ended up running each line individually, I didn't dig into why that was.
- Downloading dataset from S3 failed, I think because of non-allowed double quotes in the filename (from the etag) [ got around it by removing etag part of filename]
- Download also threw exception because it couldn't remove the temporary file, seems like Windows hangs onto the file for a bit, a 1 second pause didn't help [ got around by catching that exception, deleting the files manually later ] 
- Had issues getting appropriate tensorflow/tensorboard installed where SummaryWriter was found [ got around it by instead using TensorboardX which somebody made for this purpose ]

I'm going to mark this issue as very low priority, until we hear people actually asking for us to support this.  For now, the instructions will just live in this issue.

sugeknowles:
+1 for windows support.

schmmd:
3 resolved by https://github.com/allenai/allennlp/pull/636
5 resolved by https://github.com/allenai/allennlp/pull/618

kkhaladk:
+1 for windows!

kearnsw:
@matt-gardner : updated my allennlp install using `pip install allennlp -U` and am having the same issue mentioned in number 7 on linux. `from tensorboardX import SummaryWriter` works but `from tensorboard import SummaryWriter` throws `ImportError: cannot import name 'SummaryWriter'`.

matt-gardner:
@kearnsw: Hmm, sadly I don't have any ideas for you, other than maybe there's some helpful information in one of these: https://github.com/allenai/allennlp/issues/342 https://github.com/allenai/allennlp/pull/561.

DeNeutoy:
@kearnsw this is a version issue, I just experienced the same thing myself. You need `pip install tensorboard==1.0.0a6`. PR incoming.

DeNeutoy:
[...]

PR #618, opened 2017-12-14
remove quotes from etag

joelgrus:
this solves one of the issues raised in https://github.com/allenai/allennlp/issues/612

note that this change will basically invalidate your cache, and the next time you need a model it will have to download it. I don't think that's a big deal.


schmmd:
Cache validation isn't a big deal at all...

PR #621, opened 2017-12-14
Release v0.3.0

schmmd:
I will merge this manually with `git merge` rather than squashing the commits.  If I squash the commit, then my commit bumping the version numbers disappears.  I can then tag a commit on master, but that new tag might pull in additional commits that were not tested.

If I merge manually, then I can test and code review the release before merging.

This also allows the tag to exist when CI picks up the master commit, which will trigger a Docker deployment.


schmmd:
If I create a wheel (via `python setup.py bdist_wheel`) and install it in a fresh conda environment (`pip install ~/hack/allenai/allennlp/dist/allennlp-0.3.0-py3-none-any.whl`) the installation can successfully make predictions (after install PyTorch and downloading spacy models).

However if I upload the installation to testpypi and run `pip install -i https://testpypi.python.org/pypi allennlp` as documented by @DeNeutoy, then I get an error that the dependencies cannot be found.  I think this is because the dependencies are being looked for on testpypi.

schmmd:
`pip install -i https://testpypi.python.org/pypi --extra-index-url https://pypi.org/simple allennlp==0.1.0` will install the latest release, as uploaded to testpypi.  Note the weird version `0.1.0` is because it's impossible to redeploy `0.3.0` as we deleted it.

schmmd:
On my previous release I found that `pytz` wasn't included in the requirements.txt or setup.py.  Joel tried out my `testpypi` installation and was able to get all tests passing against it.

I've never published to `pypi` before, but this checks out as far as I can tell.

If someone approves this, I'll merge this over in the morning.

PR #623, opened 2017-12-15
Use setup.py for requirements.txt dependencies.

schmmd:
I noticed our `requirements.txt` and `setup.py` diverging, and then found this neat trick from https://caremad.io/posts/2013/07/setup-vs-requirement/.

The TeamCity build will verify whether this works correctly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -113,7 +113,7 @@ setup(name='allennlp',
           'sanic-cors==0.6.0.2',
           # Talk to postgres demo database
           'psycopg2',
-          # argument parsing in for data cleaning scripts
+          # argument parsing for data cleaning scripts
           'argparse',
           # Used for downloading datasets over HTTP
           'requests>=2.18',
--- a/setup.py
+++ b/setup.py
@@ -85,26 +85,47 @@ setup(name='allennlp',
       license='Apache',
       packages=find_packages(),
       install_requires=[
+          # Parameter parsing.
           'pyhocon==0.3.35',
+          # Type checking for python
           'typing',
+          # Adds an @overrides decorator for better documentation and error checking when using subclasses.
           'overrides',
+          # Used by some old code.  We moved away from it because it's too slow, but some old code still
+          # imports this.
           'nltk',
+          # Mainly used for the faster tokenizer.
           'spacy>=2.0,<2.1',
-          'numpy',
-          'tensorboard',
-          'cffi==1.11.2',
-          'awscli>=1.11.91',
-          'sanic==0.6.0',
-          'psycopg2',
-          'sanic-cors',
-          'argparse',
-          'requests>=2.18',
-          'tqdm',
-          'editdistance',
-          'jupyter',
-          'h5py',
-          'scikit-learn',
+          # Used by span prediction models.
+          # Used in coreference resolution evaluation metrics.
           'scipy',
+          'scikit-learn',
+          # Training visualisation using tensorboard.
+          'tensorboard',
+          # Required by torch.utils.ffi
+          'cffi==1.11.2',
+          # Used by span prediction models.
+          'numpy',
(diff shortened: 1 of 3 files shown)


schmmd:
+1 for TeamCity, as it correctly identified this as a breaking change where Travis did not.

schmmd:
On the Docker image tests pass, but pylint fails pretty terribly.

schmmd:
It's using different settings.  I.e. in the logs of this PR I get`C: 18, 0: Line too long (105/100) (line-too-long)` but locally the line length limit is 115.

schmmd:
I do not understand how pylint has been working successfully on TeamCity (as it seems to have been) to date, as the file I added in the previous commit was missing.

schmmd:
I fixed the issue I had been having with https://github.com/allenai/allennlp/pull/638.

schmmd:
There is one functional change in that `pip install -r requirements.txt` will install `allennlp` into `pip`.  It will be importable from directories other than the base `allennlp` directory.

On the upside, we don't need to duplicate `requirements.txt` and `setup.py`, which has been causing issues with our pip release.

setup.py
```diff
@@ -85,26 +85,47 @@
...
+          'numpy',
+          # aws commandline tools for running on Docker remotely.
           'awscli>=1.11.91',
+          # REST interface for models
           'sanic==0.6.0',
+          'sanic-cors==0.6.0.2',
+          # Talk to postgres demo database
           'psycopg2',
-          'sanic-cors',
+          # argument parsing in for data cleaning scripts
```
matt-gardner: Typo was present in `requirements.txt`, but can you fix it?  `s/in for/for/`.

matt-gardner:
Oh, wait a minute, I'm not certain that's a good thing...  I'd prefer to not install the library when running `pip install -r requirements.txt` - I'd rather there not be any ambiguity in my development environment when I'm developing allennlp.

schmmd:
@matt-gardner I agree--I would also prefer not to install `allennlp` as a library.  However, I think this has a net positive benefit of avoiding duplicating our dependency versions and others take this approach as well.

schmmd:
I don't think the rules are ambiguous: if you are in the `allennlp` directory you will be using the source code.

joelgrus:
I really don't like having allennlp installed as a library in the virtualenv you're working on source in

matt-gardner:
Neither does @DeNeutoy: https://github.com/allenai/allennlp/pull/627#pullrequestreview-83945500.  Is there no other way to remove this duplication?

DeNeutoy:
The issue with this is that there _is_ a subtle distinction between `requirements.txt` and `setup.py` (or at least in how I have interpreted their use), specifically that `requirements.txt` should be for development and `setup.py` for use as a library. The other thing enforcing this separation is that you can't have dependencies pinned to git commits in setup.py (it breaks the install), but this way around, we can't temporarily pin libraries to git commits because they are installed from `setup.py`.
[This probably describes it better.](https://packaging.python.org/discussions/install-requires-vs-requirements/#install-requires-vs-requirements-files)

DeNeutoy:
Additionally, this will be unnecessary once we have a TeamCity job which builds a wheel and runs a subset of our tests, as this would probably catch most critical missing dependencies.

joelgrus:
I can write a python script that checks that the two files are in sync and we can add it to our CI, the same way that we already do for our docs.

schmmd:
@DeNeutoy we somewhat remove the "subtle distinction" by separating out our development dependencies into `requirements_test.txt`.  I think avoiding dependency duplication would still be good even when we have more tests.

@joelgrus good idea on writing a CI job (although I still think the duplication between `setup.py` and `requirements.txt` is a bit nuts.

PR #626, opened 2017-12-15
Add __str__ and __repr__ to Token class

### LINKED ISSUE #624: Non-spacy Tokenizers don't show text of passage tokens in SQuAD debug
Hi,

I'm trying to use allennlp as a squad dataloader in my code, but I noticed that some passages were incorrectly tokenized by spacy due to wikipedia `[notes]` and `[citations]` (same issue as https://github.com/allenai/allennlp/issues/236).

Upon switching to the `LettersDigitsWordSplitter` that solves this problem, the squad debug output for the actual tokens generated is just a list of token objects and their addresses (as opposed to readable strings), since `__repr__` / `__str__` isn't implemented. Would you mind a PR to add these (will probably just [copy from SpaCy](https://github.com/explosion/spacy/blob/master/spacy/tokens/token.pyx#L65-L77)), or do you think that there's a better way to go about solving this?

Thanks!

nelson-liu:
Fixes #624 by adding `__str__` and `__repr__` magic methods to the `Token` class, enabling them to be printed / logged.

Does this need a test? If so, where do you think is a good place?

### DIFF
--- a/allennlp/data/tokenizers/token.py
+++ b/allennlp/data/tokenizers/token.py
@@ -43,3 +43,9 @@ class Token:
         self.dep_ = dep
         self.ent_type_ = ent_type
         self.text_id = text_id
+
+    def __str__(self):
+        return self.text
+
+    def __repr__(self):
+        return self.__str__()


matt-gardner:
Thanks Nelson! I don't think a test is necessary for this.  This doesn't print any of the additional info you might have in your token, but we can worry about that if someone asks for it.

Issue #614, opened 2017-12-14
Consolidate locations where the version is stored


schmmd:
Presently the version is stored in at least two different places (see [this commit](https://github.com/schmmd/allennlp/commit/5ecb40c3dc56a81fe92680d36db9ea9d482ffbff) for an example of bumping version numbers).  Ideally we would have a single file at the root (say `version.py`) which had the ground truth for our version or our system would use associated git tags and SHAs.

schmmd:
We also want to make it so users can print the version of AllenNLP they are using.

PR #630, opened 2017-12-16
Configure version in one place.

schmmd:
Fixes https://github.com/allenai/allennlp/issues/614


schmmd:
@joelgrus should these be all caps constants or should I organize things another way (class, etc)?  This works for the documentation and setup.py--although I will do some more testing.

joelgrus:
pylint will be mad if you don't make them all caps

joelgrus:
if it were me, I would probably do something like

```
_MAJOR = 0
_MINOR = 3
_REVISION = "1-unreleased"

VERSION_SHORT = f"{_MAJOR}.{_MINOR}"
VERSION = f"{_MAJOR}.{_MINOR}.{_REVISON}"
```

you can play with the names of the last two

allennlp/version.py
```diff
@@ -0,0 +1,6 @@
+_MAJOR = "0"
+_MINOR = "3"
+_REVISION = "1-unreleased"
+
+VERSION_SHORT = "{0}.{1}".format(_MAJOR, _MINOR)
+VERSION = "{0}.{1}.{2}".format(_MAJOR, _MINOR, _REVISION)
```
matt-gardner: Can you add `from allennlp.version import VERSION as __version__` to `allennlp.__init__.py`?  I think that's the only thing missing here.  No strong preference between `VERSION` and `__version__` in the `__init__.py`, it just seems like `__version__` is more common.  We could also just give both...

PR #637, opened 2017-12-18
Remove references to individual commits in tutorial, add data

matt-gardner:
I originally wrote the tutorial as stepping through commits in an external repo, but that doesn't allow updating the repo at all, which is problematic.  I also said that users should download and make their own dataset, which isn't very nice.  This removes references to specific commits, and changes paths to point to some data that I uploaded to S3, so the tutorial is more self-contained.

### DIFF
--- a/tutorials/getting_started/using_in_your_repo.md
+++ b/tutorials/getting_started/using_in_your_repo.md
@@ -12,7 +12,7 @@ some repetition is good, right?
 We'll build a model for classifying academic papers, and we'll do it using a [separate
 repository](https://github.com/allenai/allennlp-as-a-library-example), showing how to include
 AllenNLP as a dependency in your project, what classes you need to implement, and how to train your
-model.  We'll step through the commits in that repository, explaining the code we write as we go.
+model.  We'll go through the code we wrote in that repository, explaining it as we go.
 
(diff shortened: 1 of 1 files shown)


matt-gardner:
That example repo installs allennlp using a specific version on pip - you can't break it with changes to master or anything else.

matt-gardner:
@schmmd, how do I deploy this tutorial update?  I didn't find any docs on that in `allennlp-internal`.

schmmd:
My point regarding versioning is that all commits where the documentation exists in this repository refers to the master commit of the tutorial.

schmmd:
You can click "Run" next to "Docs and Tutorials".  However, I would not do this:

1.  This will update our documentation, which we only want to release on versioned releases.
2.  You might pull in changes that shouldn't be deployed to the docs due to breaking API changes.  (I'm pretty sure this isn't an issue presently).

Tomorrow I'd like to discuss just hosting our tutorials on GitHub so we have clearer versioning.  Can we hold off until then?  If we move to GitHub then versioning of our tutorials will be much more clear and we can get rid of some complexity that's been taking up time.

Pinging @joelgrus as he's the expert here.

PR #640, opened 2017-12-19
Self attention improvements

DeNeutoy:
- The original paper has `LayerNorm` in-between everything, not just the top level blocks.
- The tensorflow implementation uses an initialisation method that I hadn't seen before, from this paper https://arxiv.org/abs/1412.6558, so I added it. Hopefully this will resolve some of the scaling issues, as it is now a lot easier to customise the initialisation of the different feedforward layers at a more fine grained level, e.g depending on the non-linearity etc.


DeNeutoy:
FYI - these last 2 commits are what got a 4 layer bigger model (600 dim feedforward layers and 8 attention heads vs 200 dim feedforward layers and 4 attention heads) to learn anything at all. Nuts.

allennlp/modules/seq2seq_encoders/stacked_self_attention.py
```diff
@@ -83,14 +84,18 @@ def __init__(self,
             self.add_module(f"feedforward_{i}", feedfoward)
             self._feedfoward_layers.append(feedfoward)
 
+            layer_norm = LayerNorm(feedfoward.get_input_dim())
```
matt-gardner: Duplicating the variable name here makes it really easy to add bugs later - can you use a different name for this than you use below?

allennlp/nn/initializers.py
```diff
@@ -61,6 +63,49 @@ def from_params(cls, params: Params):
...
+    Returns
+    -------
+    The initialised tensor.
+    """
+    if isinstance(tensor, Variable):
+        uniform_unit_scaling(tensor.data, nonlinearity)
+        return tensor
+
+    size = 1.
+    # Estimate the input size. This won't work perfectly,
```
matt-gardner: You're using all of the dimensions up to `-1` because the last dimension is probably the output dimension, right?  Can you be more explicit about that in the comment?

PR #642, opened 2017-12-20
simple server

joelgrus:
a simple server for serving up a single AllenNLP model, along with a really bare-bones web app for interacting with it


schmmd:
This seems good for a basic skeleton.  I'd like to re-use the basic structure of demo.allennlp.org before we check this off though.  With this rough of a demo I don't think we'll see much adoption.

joelgrus:
So, two reasons I didn't do that:

(1) the requirement as I understood it was just "text in, text out"
(2) I wanted it to be self-contained without requiring npm or react or anything

I can definitely spruce up the CSS + HTML to make it look more like the demo tho

schmmd:
Yes, the requirements are text in, text out.

However, I think this would have more adoption if we had the same CSS and basic look and feel as demo.allennlp.org.  For example, part of the SRL demo is just text in, text out.

image

joelgrus:
ok, this is ready for another look

screen shot 2017-12-21 at 2 07 00 pm

joelgrus:
most of it can be deleted, I'm just not sure which "most" 😀

Issue #506, opened 2017-11-16
fix `Predictor` API so that it's easier to add things to the output


joelgrus:
in both the SRL and sentence-tagger predictors, we need to tokenize the input sentence and add it to the output. right now this is awkward to do and we have to override predict_json. figure out a way to make it simpler/cleaner

schmmd:
Fixed by #646

PR #646, opened 2017-12-21
Predictor api

### LINKED ISSUE #591: Implement batch prediction for SRL
I'm currently trying to compile a dataset of ~50M SRL-parsed sentences and was hoping to speed things up with batch processing. Unfortunately, I discovered that it's not yet implemented for SRL.

https://github.com/allenai/allennlp/blob/f700584ed1b72ab45215e405ed0d22564938d1cb/allennlp/service/predictors/semantic_role_labeler.py#L52

In the meantime, do you have any suggestions for speeding up SRL parsing ?

I've also tried running the `predict` command with `--cuda_device 0` on a Tesla K80, but checking `nvidia-smi ` it seems that there's some bottleneck preventing it from going above 25% GPU usage. Could the lack of batch processing create some overhead that's responsible for that, or am I missing something else?

Thanks!

DeNeutoy:
Fixes #591  and #506

- Makes all the `_json_to_instance` methods also return a dictionary.
- Implements the batch decoding case for SRL, which is more complicated because we need to do multiple passes through the model for the single instance. This switches the `predict_json` for the SRL predictor to run all the verbs for the sentence in a single batch and for the batched case, do them in batches of size len(sentence).


allennlp/service/predictors/semantic_role_labeler.py
```diff
@@ -41,53 +41,147 @@ def make_srl_string(words: List[str], tags: List[str]) -> str:
...
-
-        .. code-block:: js
-
-            {"words": [...],
-             "verbs": [
-                {"verb": "...", "description": "...", "tags": [...]},
-                ...
-                {"verb": "...", "description": "...", "tags": [...]},
-            ]}
+        The SRL model has a slightly different API to other models, as the model is run
```
joelgrus: *from* other models

allennlp/service/predictors/semantic_role_labeler.py
```diff
@@ -41,53 +41,147 @@ def make_srl_string(words: List[str], tags: List[str]) -> str:
...
+            ]
+        """
+        # For SRL, we have more instances than sentences, but the user specified
+        # a batch size with respect to the number of sentences passed, so we respect
+        # that here by taking the batch size which we use to be the number of sentences
+        # we are given.
+        batch_size = len(inputs)
+        instances_per_sentence, return_dicts = zip(*[self._sentence_to_srl_instances(json)
+                                                     for json in inputs])
+        flattened_instances = sum(instances_per_sentence, [])
```
joelgrus: I dislike this construct (and I believe it's quadratic, which probably doesn't matter much here but makes it a good idea to avoid it on principle)

```
flattened_instances = [instance for sentence_instances in instances_per_sentence for instance in sentence_instances]
```

DeNeutoy: I went down a massive rabbit hole reading about list monads when I looked this up, but thanks, I wasn't aware this was so inefficient. I also didn't know that `list(itertools.chain.from_iterable(some_list))` is actually _faster_ than the nested list comprehensions. weird.

Issue #447, opened 2017-10-26
Overriding environment variables causes them to be interpreted as a string


schmmd:
From @tusharkhot 

Michael Schmitz So there is one issue with using environment variables as fallback in HOCON. These environment variables are assumed to be strings by default which lead to incompatible type errors in AllenNLP. E.g. when setting cuda_device using environment variables

```
  File "/usr/local/lib/python3.6/site-packages/allennlp/training/trainer.py", line 515, in from_params
    if cuda_device >= 0:
TypeError: '>=' not supported between instances of 'str' and 'int'
```

I can submit a PR to allennlp to convert every parameter to the right type (or at least the interesting ones) or maybe there is a fancy way to ensure environment variables have the right type ?

schmmd:
@joelgrus just curious, do you have any ideas here?

joelgrus:
probably the "best" thing to do is to add some more pop methods to `Params` e.g. `pop_int` that do something like

```
def pop_int(self, key: str, default: Any = DEFAULT):
    value = self.pop(key, default)
    if isinstance(value, str):
        value = int(value)
    return value
```

and then change all of the `Params.pop()` calls that we expect to be ints to `pop_int()`, e.g.

```
        cuda_device = params.pop_int("cuda_device", -1)
```

joelgrus:
although I think we're probably a little bit loose about using `dict`s where we say we're using `Params`, so this might break that (although we should fix those)

joelgrus:
it's also possible this is a bug in pyhocon

matt-gardner:
@joelgrus, where are we loose with using `dicts` where we say we're using `Params`?  Like this: `params.pop('tokenizer', {})`?  Those get converted to `Params` objects by the `pop` method, so it's not actually being loose, it's just allowing the calling code to be less verbose.

joelgrus:
oops, it's possible that's what I was thinking of

joelgrus:
in re my "bug" comment, the hocon spec says

> A substitution is replaced with any value type (number, object, string, array, true, false, null). If the substitution is the only part of a value, then the type is preserved. Otherwise, it is value-concatenated to form a string.

which suggests that it is a bug

tusharkhot:
It has the same behavior in typesafe config

> environment variables always become a string value, though if an app asks for another type automatic type conversion would kick in

https://github.com/lightbend/config/blob/master/HOCON.md#substitution-fallback-to-environment-variables

[...]

PR #648, opened 2017-12-22
Add pop_int, pop_float, and pop_bool

schmmd:
This allows better typechecking when reading configuration files, and allows environment variable substitutions to be coerced into ints, floats, and bools.

Partially addresses https://github.com/allenai/allennlp/issues/447.


schmmd:
@joelgrus mind taking a first pass?  You can just look at my `pop_int` methods.  I'll do another read-over on all the places where I changed methods.

I'm somewhat conflicted about this change.  One one hand, it provides some additional typechecking.  On the other hand, nothing enforces new code to use these more specific methods and it's confusing that not all fields are coerced correctly the the appropriate type (see https://github.com/allenai/allennlp/blob/master/allennlp/training/optimizers.py#L35, whose configuration is passed directly to PyTorch).  We also have a number of fields that can be ints or lists of ints (such as [hidden dims](https://github.com/allenai/allennlp/blob/master/allennlp/modules/feedforward.py#L24).

matt-gardner:
mypy is failing because of the `Union[Params, Dict[str, Any]]` argument to `prepare_environment`.  Looks like you can just make that a `Params` and remove the `Union`.

PR #655, opened 2018-01-02
add static `load_line` and `dump_line` methods to predictor to allow non-JSONL formats

joelgrus:
and then once I imported the `json` library into the base `predictor.py`, I had to rename the `json` variables there, and then in the subclasses.

### DIFF
--- a/allennlp/service/predictors/bidaf.py
+++ b/allennlp/service/predictors/bidaf.py
@@ -11,10 +11,10 @@ class BidafPredictor(Predictor):
     Wrapper for the :class:`~allennlp.models.bidaf.BidirectionalAttentionFlow` model.
     """
     @overrides
-    def _json_to_instance(self, json: JsonDict) -> Instance:
+    def _json_to_instance(self, obj: JsonDict) -> Instance:
         """
         Expects JSON that looks like ``{"question": "...", "passage": "..."}``.
         """
-        question_text = json["question"]
-        passage_text = json["passage"]
+        question_text = obj["question"]
+        passage_text = obj["passage"]
         return self._dataset_reader.text_to_instance(question_text, passage_text)
--- a/allennlp/service/predictors/predictor.py
+++ b/allennlp/service/predictors/predictor.py
@@ -1,4 +1,6 @@
+import json
 from typing import List
+
 from allennlp.common import Registrable
 from allennlp.common.util import JsonDict, sanitize
 from allennlp.data import DatasetReader, Instance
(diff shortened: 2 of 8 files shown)


joelgrus:
also, part of me really just wanted to add an option to let the `Predictor` use its `DatasetReader` to read in a file (which is just what I needed to do here), but there wasn't a really clean way to do that (and it didn't solve the output problem)

allennlp/service/predictors/predictor.py
```diff
@@ -15,12 +17,28 @@ def __init__(self, model: Model, dataset_reader: DatasetReader) -> None:
         self._model = model
         self._dataset_reader = dataset_reader
 
+    @staticmethod
+    def load_line(line: str) -> JsonDict:
```
matt-gardner: Do we really want this to be static?  I'm guessing you did this because pylint complained - I'd probably just ignore pylint on a method like this that's meant to be overridden.  Making it static is limiting to future subclasses.

allennlp/service/predictors/predictor.py
```diff
@@ -15,12 +17,28 @@ def __init__(self, model: Model, dataset_reader: DatasetReader) -> None:
...
+        """
+        If your inputs are not in JSON-lines format (e.g. you have a CSV)
+        you can override this function to parse them correctly.
+        """
+        return json.loads(line)
+
+    @staticmethod
+    def dump_line(outputs: JsonDict) -> str:
```
matt-gardner: Same comment about `@staticmethod` here.

tests/commands/predict_test.py
```diff
@@ -158,3 +160,60 @@ def predict_json(self, inputs: JsonDict, cuda_device: int = -1) -> JsonDict:
             assert set(result.keys()) == {"span_start_logits", "span_end_logits",
                                           "span_start_probs", "span_end_probs", "best_span",
                                           "best_span_str", "overridden"}
+
+    def test_alternative_file_formats(self):
+        tempdir = tempfile.mkdtemp()
+        infile = os.path.join(tempdir, "inputs.txt")
+        outfile = os.path.join(tempdir, "outputs.txt")
+
+        @Predictor.register('bidaf-csv')  # pylint: disable=unused-variable
```
matt-gardner: What variable is it complaining about?  The class?  I'm surprised this works, if that's the case - I thought pylint scoping worked differently than this.

tests/commands/predict_test.py
```diff
@@ -158,3 +160,60 @@ def predict_json(self, inputs: JsonDict, cuda_device: int = -1) -> JsonDict:
             assert set(result.keys()) == {"span_start_logits", "span_end_logits",
                                           "span_start_probs", "span_end_probs", "best_span",
                                           "best_span_str", "overridden"}
+
+    def test_alternative_file_formats(self):
+        tempdir = tempfile.mkdtemp()
```
matt-gardner: Does this tempdir ever get cleaned up?

allennlp/service/predictors/bidaf.py
```diff
@@ -11,10 +11,10 @@ class BidafPredictor(Predictor):
     Wrapper for the :class:`~allennlp.models.bidaf.BidirectionalAttentionFlow` model.
     """
     @overrides
-    def _json_to_instance(self, json: JsonDict) -> Instance:
+    def _json_to_instance(self, obj: JsonDict) -> Instance:
```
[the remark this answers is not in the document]
joelgrus: renamed to `json_dict`

allennlp/service/predictors/predictor.py
joelgrus: changed it to a regular instance method and added a pylint ignore

tests/commands/predict_test.py
joelgrus: looks like none of the tempdirs in this file do, I added cleanups for all of them

tests/commands/predict_test.py
joelgrus: it's complaining about the class I believe because we never instantiate it directly, only through `Predictor.from_params`

matt-gardner: Interesting - I thought that inline comments like this only applied to the current line, so I'm surprised this works.  Makes me wonder if all of our other inline annotations are wrong...

joelgrus: surprisingly, pylint complains if you move the disable to the next line. I don't really understand this, but I guess it has something to do with using the `register` decorator

PR #656, opened 2018-01-02
Make assert_fields_equal more robust to different Field structures.

matt-gardner:
This is important for the `ProductionRuleField` in the semantic parser.  I also cleaned up the use of the `tolerance` parameter.

### DIFF
--- a/allennlp/common/testing/model_test_case.py
+++ b/allennlp/common/testing/model_test_case.py
@@ -28,7 +28,9 @@ class ModelTestCase(AllenNlpTestCase):
         self.dataset = dataset
         self.model = Model.from_params(self.vocab, params['model'])
 
-    def ensure_model_can_train_save_and_load(self, param_file: str, tolerance: float = 1e-6,
+    def ensure_model_can_train_save_and_load(self,
+                                             param_file: str,
+                                             tolerance: float = 1e-4,
                                              cuda_device: int = -1):
         save_dir = os.path.join(self.TEST_DIR, "save_and_load_test")
         archive_file = os.path.join(save_dir, "model.tar.gz")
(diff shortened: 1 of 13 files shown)


allennlp/common/testing/model_test_case.py
```diff
@@ -92,18 +86,29 @@ def ensure_model_can_train_save_and_load(self, param_file: str, tolerance: float
...
-    @staticmethod
-    def assert_fields_equal(field1, field2, tolerance: float = 1e-6, name: str = None) -> None:
+    def assert_fields_equal(self, field1, field2, tolerance: float = 1e-6, name: str = None) -> None:
         if isinstance(field1, torch.autograd.Variable):
             assert_allclose(field1.data.cpu().numpy(),
                             field2.data.cpu().numpy(),
                             rtol=tolerance,
                             err_msg=name)
+        elif isinstance(field1, dict):
+            for key in field1:
```
joelgrus: should you also check that they have the same keys? right now this would pass if field2 had additional keys

allennlp/common/testing/model_test_case.py
```diff
@@ -92,18 +86,29 @@ def ensure_model_can_train_save_and_load(self, param_file: str, tolerance: float
...
+                self.assert_fields_equal(field1[key],
+                                         field2[key],
+                                         tolerance=tolerance,
+                                         name=name + '.' + key)
+        elif isinstance(field1, (list, tuple)):
+            for i, (subfield1, subfield2) in enumerate(zip(field1, field2)):
```
joelgrus: similarly, should you check that they have the same length? `zip` only goes up to the `min` length, which means this would pass if one was longer but they agreed up to that point

allennlp/common/testing/model_test_case.py
```diff
@@ -92,18 +86,29 @@ def ensure_model_can_train_save_and_load(self, param_file: str, tolerance: float
...
                                      loaded_model_predictions[key],
-                                     tolerance=1e-4,
+                                     tolerance=tolerance,
                                      name=key)
 
         return model, loaded_model
 
-    @staticmethod
-    def assert_fields_equal(field1, field2, tolerance: float = 1e-6, name: str = None) -> None:
+    def assert_fields_equal(self, field1, field2, tolerance: float = 1e-6, name: str = None) -> None:
```
joelgrus: a couple of times below you are doing string concatenation on `name`, so maybe empty string is a better default value?

matt-gardner: Thanks, good catch.  We don't actually need a default value, so I just removed it.

allennlp/common/testing/model_test_case.py
matt-gardner: Yep, great point, thanks.

allennlp/common/testing/model_test_case.py
matt-gardner: Yep again, thanks.

PR #657, opened 2018-01-02
Add a check_dimensions_match function

### LINKED ISSUE #499: Add a simple "check dimensions match" method

matt-gardner:
Fixes #499.

### DIFF
--- a/allennlp/common/checks.py
+++ b/allennlp/common/checks.py
@@ -23,7 +23,7 @@ class ConfigurationError(Exception):
 
 def log_pytorch_version_info():
     import torch
-    logger.info("Pytorch version: " + torch.__version__)
+    logger.info("Pytorch version: %s", torch.__version__)
 
 
 def check_dimensions_match(dimension_1: int,
--- a/allennlp/common/checks.py
+++ b/allennlp/common/checks.py
@@ -23,4 +23,13 @@ class ConfigurationError(Exception):
 
 def log_pytorch_version_info():
     import torch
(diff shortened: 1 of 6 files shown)


allennlp/common/checks.py
```diff
@@ -24,3 +24,12 @@ def __str__(self):
 def log_pytorch_version_info():
     import torch
     logger.info("Pytorch version: %s", torch.__version__)
+
+
+def check_dimensions_match(dimension_1: int,
+                           dimension_2: int,
+                           dim_1_name: str,
+                           dim_2_name: str) -> None:
```
joelgrus: do you want to add an optional `message` field, so that we can include longer explanations when appropriate? (I see that this change currently eliminates some of them)

matt-gardner: I think the only place where there is information lost is in saying that the SRL model uses a binary verb indicator, which is actually incorrect (it was originally binary, it's now an embedding).  You still get the gist of (the correct part of) that information in the `"text embedding dim + verb indicator embedding dim"` message, so I'm not sure we're losing anything important.  If there's more to say, it should probably be said in the model docstring.

PR #658, opened 2018-01-02
Make ListField batching robust

matt-gardner:
This doesn't result in any changes for our current `Fields`, but it is necessary for the `ListField[ProductionRuleField]` that we use in the semantic parser.

### DIFF
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -26,7 +26,7 @@ def batch_tensor_dicts(tensor_dicts: List[Dict[str, torch.Tensor]],
         The list of tensor dictionaries to batch.
     remove_trailing_dimension : ``bool``
         If ``True``, we will check for a trailing dimension of size 1 on the tensors that are being
-        batched, and remove it if it find it.
+        batched, and remove it if we find it.
     """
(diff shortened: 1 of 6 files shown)


allennlp/nn/util.py
```diff
@@ -14,6 +14,34 @@
...
+    """
+    key_to_tensors: Dict[str, List[torch.Tensor]] = defaultdict(list)
+    for tensor_dict in tensor_dicts:
+        for key, tensor in tensor_dict.items():
+            key_to_tensors[key].append(tensor)
+    batched_tensors = {}
+    for key, tensor_list in key_to_tensors.items():
+        batched_tensor = torch.stack(tensor_list)
+        if remove_trailing_dimension:
+            if set(tensor.size(-1) for tensor in tensor_list) == set([1]):
```
joelgrus: I think this would be more readable if it was just

```
if remove_trailing_dimension and all(tensor.size(-1) == 1 for tensor in tensor_list):
```

matt-gardner: Thanks, good suggestion - I tend to forget about things like `all`.

PR #660, opened 2018-01-04
[WIP / for discussion] Lazy datasets

joelgrus:
here's a proposal for how to handle lazy datasets

at a high level, here are the changes

* make `Dataset.instances` private (and eliminate all external access to it)
* replace it with `Dataset.iterinstances() -> Iterator[Instance]` and change all calling code to use that
* introduce `Dataset.num_instances` property to replace calls to `len(dataset.instances)`
* create `LazyDataset` subclass that takes an "instance generator" instead of a list of instances
* create lazy `DataIterator` subclass that doesn't rely on having all instances in-memory
* modify `commands/train` to build the vocabulary from a `LazyDataset` instead of a `Dataset`
* add some tests

I'm not certain this is the right approach. It's a breaking change for anyone who's using `Dataset.instances` directly, although I'm not sure if that's anyone or not. A non-breaking alternative would have `LazyDataset` not be a subclass of `Dataset` and then try to flow those changes through, but that felt worse.

Some of the steps around how the LazyDataset handles indexing and padding are a little bit clunky, but I think they're right.

I don't *think* this should have any performance impacts (I think now during training we'll make a shallow copy of the list of instances each epoch, but that should be really fast).

Let me know what you think about this approach / the naming.

### DIFF
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -158,10 +158,13 @@ def train_model(params: Params, serialization_dir: str) -> Model:
     # with the combined instances.
     def combined_generator():
         yield from (instance for key, dataset in all_datasets.items()
-                    for instance in dataset.iterinstances()
+                    for instance in dataset
                     if key in datasets_for_vocab_creation)
 
-    lazy_combined_dataset = LazyDataset(combined_generator)
+    total_instances = sum(dataset.num_instances
+                          for key, dataset in all_datasets.items()
+                          if key in datasets_for_vocab_creation)
+    lazy_combined_dataset = LazyDataset(combined_generator, total_instances)
     vocab = Vocabulary.from_params(params.pop("vocabulary", {}), lazy_combined_dataset)
 
     vocab.save_to_files(os.path.join(serialization_dir, "vocabulary"))
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -32,7 +32,7 @@ from allennlp.common.checks import ConfigurationError
 from allennlp.common.params import Params
 from allennlp.common.tee_logger import TeeLogger
 from allennlp.common.util import prepare_environment
-from allennlp.data import Dataset, Vocabulary
+from allennlp.data import Dataset, Vocabulary, LazyDataset
 from allennlp.data.dataset_readers.dataset_reader import DatasetReader
 from allennlp.data.iterators.data_iterator import DataIterator
 from allennlp.models.archival import archive_model
@@ -152,10 +152,16 @@ def train_model(params: Params, serialization_dir: str) -> Model:
             raise ConfigurationError(f"invalid 'dataset_for_vocab_creation' {dataset}")
 
     logger.info("Creating a vocabulary using %s data.", ", ".join(datasets_for_vocab_creation))
-    vocab = Vocabulary.from_params(params.pop("vocabulary", {}),
-                                   Dataset([instance for key, dataset in all_datasets.items()
-                                            for instance in dataset.instances
-                                            if key in datasets_for_vocab_creation]))
+
+    # As the individual Datasets may be lazy (and, in particular, large), we don't want to
+    # manifest a List of their instances to construct the vocabulary. Instead we use a LazyDataset
+    # with the combined instances.
+    combined_instances = (instance for key, dataset in all_datasets.items()
+                          for instance in dataset.iterinstances()
+                          if key in datasets_for_vocab_creation)
+    lazy_combined_dataset = LazyDataset(lambda: combined_instances)
+    vocab = Vocabulary.from_params(params.pop("vocabulary", {}), lazy_combined_dataset)
+
     vocab.save_to_files(os.path.join(serialization_dir, "vocabulary"))
 
     model = Model.from_params(vocab, params.pop('model'))
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -156,10 +156,12 @@ def train_model(params: Params, serialization_dir: str) -> Model:
     # As the individual Datasets may be lazy (and, in particular, large), we don't want to
     # manifest a List of their instances to construct the vocabulary. Instead we use a LazyDataset
     # with the combined instances.
-    combined_instances = (instance for key, dataset in all_datasets.items()
-                          for instance in dataset.iterinstances()
-                          if key in datasets_for_vocab_creation)
-    lazy_combined_dataset = LazyDataset(lambda: combined_instances)
+    def combined_generator():
+        yield from (instance for key, dataset in all_datasets.items()
+                    for instance in dataset.iterinstances()
+                    if key in datasets_for_vocab_creation)
+
+    lazy_combined_dataset = LazyDataset(combined_generator)
     vocab = Vocabulary.from_params(params.pop("vocabulary", {}), lazy_combined_dataset)
 
     vocab.save_to_files(os.path.join(serialization_dir, "vocabulary"))
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -161,10 +161,7 @@ def train_model(params: Params, serialization_dir: str) -> Model:
                     for instance in dataset
                     if key in datasets_for_vocab_creation)
 
-    total_instances = sum(dataset.num_instances
-                          for key, dataset in all_datasets.items()
-                          if key in datasets_for_vocab_creation)
-    lazy_combined_dataset = LazyDataset(combined_generator, total_instances)
+    lazy_combined_dataset = LazyDataset(combined_generator)
     vocab = Vocabulary.from_params(params.pop("vocabulary", {}), lazy_combined_dataset)
 
     vocab.save_to_files(os.path.join(serialization_dir, "vocabulary"))
--- a/allennlp/data/dataset.py
+++ b/allennlp/data/dataset.py
@@ -5,7 +5,7 @@ For example, when you train a model, you will likely have a *training* dataset a
 
 import logging
 from collections import defaultdict
-from typing import Dict, List, Union
+from typing import Dict, List, Union, Iterator
 
 import torch
 import tqdm
@@ -38,7 +38,8 @@ class Dataset:
         if not all([all_instance_fields_and_types[0] == x for x in all_instance_fields_and_types]):
             raise ConfigurationError("You cannot construct a Dataset with non-homogeneous Instances.")
 
-        self.instances = instances
+        self._instances = instances
+        self.num_instances = len(instances)
 
     def truncate(self, max_instances: int):
         """
@@ -46,8 +47,8 @@ class Dataset:
         instances to the first ``max_instances``.  This `modifies` the current object, and returns
         nothing.
         """
-        if len(self.instances) > max_instances:
-            self.instances = self.instances[:max_instances]
+        if len(self._instances) > max_instances:
+            self._instances = self._instances[:max_instances]
 
     def index_instances(self, vocab: Vocabulary):
         """
@@ -55,8 +56,14 @@ class Dataset:
         ``IndexedFields``.  This modifies the current object, it does not return a new object.
         """
         logger.info("Indexing dataset")
-        for instance in tqdm.tqdm(self.instances):
+        for instance in tqdm.tqdm(self._instances):
             instance.index_fields(vocab)
+
+    def iterinstances(self) -> Iterator[Instance]:
+        """
+        An iterator over instances.
+        """
+        yield from self._instances
 
     def get_padding_lengths(self) -> Dict[str, Dict[str, int]]:
         """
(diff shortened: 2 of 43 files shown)


joelgrus:
I also strongly considered changing `Vocabulary.from_dataset` to `Vocabulary.from_instances` (it would have saved the trouble of creating a fake lazy dataset in `commands.train`, which is expensive if the underlying datasets are lazy (since it has to iterate over all of them)), but I ran into some nasty circular import problems, and eventually I gave up.

But if that's a better approach, I can figure that problem out.

allennlp/data/lazy_dataset.py
```diff
@@ -0,0 +1,114 @@
...
+            instance_lengths = instance.get_padding_lengths()
+            for field_name, instance_field_lengths in instance_lengths.items():
+                for padding_key, length in instance_field_lengths.items():
+                    prev_max_length = padding_lengths[field_name].get(padding_key, 0)
+                    padding_lengths[field_name][padding_key] = max(length, prev_max_length)
+
+        self.padding_lengths = {**padding_lengths}
+
+    def iterinstances(self) -> Iterator[Instance]:
+        for i, instance in tqdm.tqdm(enumerate(self.generator())):
```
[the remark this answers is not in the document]
matt-gardner: Ah, actually, the way that you're using `LazyDataset` for vocab creation means it doesn't make sense to crash or give a warning here in that case.  Hmm...

tests/data/iterators/lazy_iterator_test.py
```diff
@@ -0,0 +1,79 @@
...
+
+from allennlp.common import Params
+from allennlp.common.testing import AllenNlpTestCase
+from allennlp.data import LazyDataset, Instance, Token, Vocabulary
+from allennlp.data.fields import TextField
+from allennlp.data.iterators import LazyIterator
+from allennlp.data.token_indexers import SingleIdTokenIndexer
+
+class LazyIteratorTestCase(AllenNlpTestCase):
+    def setUp(self):
```
matt-gardner: Why not just put these methods in `TestLazyIterator`?

allennlp/data/lazy_dataset.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class LazyDataset(Dataset):
+    """
+    A collection of :class:`~allennlp.data.instance.Instance` objects.
+    The ``Instances`` have ``Fields``, and the fields
+    could be in an indexed or unindexed state - the ``Dataset`` has methods around indexing the
+    data and converting the data into arrays.
+    """
+    def __init__(self, generator: Callable[[], Iterator[Instance]]) -> None:
```
matt-gardner: I think this generator needs a little bit of documentation - particularly, you should note that the generator needs to run forever, looping over the data when it needs to.

tests/data/iterators/lazy_iterator_test.py
```diff
@@ -0,0 +1,79 @@
...
+        # We just want to get the single-token array for the text field in the instance.
+        instances = [tuple(instance.data.cpu().numpy())
+                     for batch in batches
+                     for instance in batch['text']["tokens"]]
+        assert len(instances) == 5
+        self.assert_instances_are_correct(instances)
+
+    def test_call_iterates_over_data_forever(self):
+        generator = LazyIterator(batch_size=2)(self.dataset)
+        batches = [next(generator) for _ in range(18)]  # going over the data 6 times
```
matt-gardner: I'm trying to figure out where in the code you get the infinite looping behavior, but I can't find it.  What am I missing?

allennlp/commands/train.py
```diff
@@ -152,10 +152,18 @@ def train_model(params: Params, serialization_dir: str) -> Model:
...
+    # As the individual Datasets may be lazy (and, in particular, large), we don't want to
+    # manifest a List of their instances to construct the vocabulary. Instead we use a LazyDataset
+    # with the combined instances.
+    def combined_generator():
+        yield from (instance for key, dataset in all_datasets.items()
+                    for instance in dataset.iterinstances()
+                    if key in datasets_for_vocab_creation)
+
+    lazy_combined_dataset = LazyDataset(combined_generator)
+    vocab = Vocabulary.from_params(params.pop("vocabulary", {}), lazy_combined_dataset)
```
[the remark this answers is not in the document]
matt-gardner: Or, if you really do want our code to do the vocab computation with a super large dataset, this still does the right thing, iterating over the data once to compute the vocab, and iterating over it once again during training.

allennlp/data/lazy_dataset.py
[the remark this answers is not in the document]
matt-gardner: Also, as I noted above, it's also reasonable to use our code for vocab creation with a large dataset, so, yeah, you don't want to crash here.  Ignore this.

allennlp/data/iterators/lazy_iterator.py
```diff
@@ -0,0 +1,54 @@
...
+    ----------
+    batch_size : int, optional, (default = 32)
+        The size of each batch of instances yielded when calling the iterator.
+    """
+    def __init__(self, batch_size: int = 32) -> None:
+        self._batch_size = batch_size
+
+    def _yield_one_epoch(self, dataset: Dataset, shuffle: bool, cuda_device: int, for_training: bool):
+        if shuffle:
+            logger.warning("shuffle is not implemented for LazyIterator")
```
nalourie-ai2: This seems liable to create bugs in the future, since we might use this private method in some generic way that requires shuffle and not notice the message in the logs. If this is a private method perhaps it should throw an error instead?

matt-gardner: Yeah, a `ConfigurationError` does seem warranted here, as this would be the result of mis-configuring something in your json file.

allennlp/data/dataset.py
```diff
@@ -38,26 +38,33 @@ class that call the constructor, such as `truncate()`.
...
         """
         Converts all ``UnindexedFields`` in all ``Instances`` in this ``Dataset`` into
         ``IndexedFields``.  This modifies the current object, it does not return a new object.
         """
         logger.info("Indexing dataset")
-        for instance in tqdm.tqdm(self.instances):
+        for instance in tqdm.tqdm(self._instances):
             instance.index_fields(vocab)
 
+    def iterinstances(self) -> Iterator[Instance]:
```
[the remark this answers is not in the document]
joelgrus: that's a good idea

allennlp/data/iterators/lazy_iterator.py
```diff
@@ -0,0 +1,54 @@
...
+            yield batch.as_tensor_dict(padding_lengths,
+                                       cuda_device=cuda_device,
+                                       for_training=for_training)
+
+    def _create_batches(self, dataset: Dataset, shuffle: bool) -> List[List[Instance]]:
+        raise RuntimeError("should never be called")
+
+    @overrides
+    def get_num_batches(self, dataset: Dataset) -> int:
+        return math.ceil(dataset.num_instances / self._batch_size)
```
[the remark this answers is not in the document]
joelgrus: but `num_instances` has to come from somewhere? there's no way to get it without iterating over the dataset, so where does that iteration happen?

allennlp/data/lazy_dataset.py
joelgrus: well, that's not how it works, so it definitely needs more documentation! 😀

each call to `generator()` needs to loop through the data once, but `generator` has to be able to be called multiple times.

tests/data/iterators/lazy_iterator_test.py
joelgrus: in the base class, `__call__` just does

```
while True:
  yield from self._yield_one_epoch(...)
```

and each call to `LazyIterator._yield_one_epoch()` calls `self.generator()` to get a fresh iterator over the whole dataset.

tests/data/iterators/lazy_iterator_test.py
joelgrus: because I copied the `BasicIterator` tests and that's how they were written 😄

allennlp/data/iterators/lazy_iterator.py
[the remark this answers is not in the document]
matt-gardner: It's provided by the caller, in the configuration file.  Like, if you have a huge language modeling dataset, you know how many instances you want to use for each "epoch", and it doesn't really matter whether you're actually iterating over the whole dataset or not.  `num_instances` here no longer necessarily has any relation to the actual size of the data, it's just a specification for how frequently you want to check in on progress by running validation.  If you know the actual number of instances in the data, and want to specify that in the configuration file, that's fine too, but for huge datasets where lazy creation is needed, you likely want to do validation more frequently than waiting for a full pass over the whole data.

tests/data/iterators/lazy_iterator_test.py
matt-gardner: Ah, right, those actually do get used in other places, because the other iterators are so similar.

tests/data/iterators/lazy_iterator_test.py
matt-gardner: Ah, great, thanks.  So, we need to be able to handle cases where one "epoch" does not exhaust the generator, and the next epoch just keeps going from where the last epoch left off.  This is pretty important, and related to my [most recent comment above](https://github.com/allenai/allennlp/pull/660#discussion_r159767396).  The iterator should have a `num_instances_per_epoch` parameter, or something (maybe `num_batches_per_epoch` instead...), and that's used here.  And then when the generator is exhausted, you start it again.  I think there are a few ways to get this kind of behavior, and I'm not sure what the best one is.

allennlp/data/lazy_dataset.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        A LazyDataset just takes a way of generating instances.
+        """
+        # Call superclass constructor with no instances, we'll override the methods that use them.
+        super().__init__([])
```
DeNeutoy: I don't like this much - see my comment below about possibly removing this whole class and supporting laziness in `Dataset` by passing a flag.

matt-gardner:
I also wasn't too thrilled with the `NotImplementedError` methods.  I missed the call to the superclass constructor, but it's the same problem.  I was considering suggesting making `Dataset` an abstract API, with concrete `InMemoryDataset` and `LazyDataset` classes, where `InMemoryDataset` defines the methods that `LazyDataset` doesn't need.  It seemed a little overboard to me, though.

The trouble with passing in the vocab to the `Dataset` is that the `DatasetReader` (which builds the `Dataset`) can't have access to it - it hasn't been built yet.  You could maybe get around this by having the `DataIterator` handle indexing, as you suggest.

I think the fundamental difference between lazy and non-lazy datasets, in terms of the computation done, is that non-lazy datasets can index all of the data once and store the result in memory, saving computation when iterating multiple times over the data.  In fact, that's probably the _reason_ that we need to be lazy in the first place - the text representing the data probably doesn't take all that much memory, but the indexed tensors do.  Everything else is largely the same - lazy datasets will (probably) read from disk while they're iterating over the data, so it'll be slower, but the basic algorithm for going over the data is the same.  

So, this is a long way of saying that the difference in behavior in `index_instances` (and the fact that `iterinstances()` does indexing in the `LazyDataset`) is kind of the _definition_ of the difference between lazy and non-lazy datasets.  I think the right thing to do here is to update documentation for what these methods do, instead of trying to make them behave the same.  `index_instances` could be defined as "Makes it so that instances returned when iterating over this dataset have already been indexed using the given vocabulary.  _When_ the indexing computation happens is determined by the concrete `Dataset` class."

joelgrus:
I think it will be ugly to have a `lazy=True` flag, the internals are very different and you'll just have lots of if/then blocks everywhere. Maybe an abstract base class makes more sense

DeNeutoy:
If we're going to move to an abstract base class it might be worth investigating how easily we can use pytorch's `Dataset`, I don't know if that even gets us anything - I think they have some good multi-threading stuff we might want to use eventually, but that might be `DataLoaders` instead.

matt-peters:
Implementation wise, here's my experience in building an efficient lazy dataset loader for another project.  This is for datasets larger then machine memory.  I don't know how easy/hard this would be to implement in the proposed APIs (as I'm having a hard time following the discussion without detailed knowledge of the current API).

On disk, the dataset is split into `N` random shards, where `N` might be 100-10,000.  Each shard can comfortably fit into memory.  To iterate through the data, the data loader picks a random shard, then treats it identically to a small dataset.  It randomly shuffles the shard, batches according to sequence length, indexes and places the resulting tensors onto a Queue.  This all happens in a separate process.  Then main modeling process then simply pulls batches from the Queue.  I found that using pytorch's monkey patched `multiprocessing` was much faster then the native Python one for tensors (e.g. `from torch.multiprocessing import Queue, Process`).  In this way, there is almost no overhead in the master process to read the data.

matt-gardner:
Yeah, @matt-peters, those are some good points.  I had in my head that in the future we would probably want the `LazyIterator` to actually load some larger number of instances, then use something like our `BucketIterator` to go over the loaded instances before moving on.  I didn't say anything because I _think_ it's orthogonal to how the `LazyDataset` is designed, but it's worth keeping in mind - maybe there's a simpler way to put all of this together than what we're thinking of.

joelgrus:
I uploaded some changes, but I'm still a little bit stuck, I'll talk about it at the meeting tomorrow

Issue #661, opened 2018-01-04
Read the Docs link is broken


nalourie-ai2:
In the README.md file, our badge / link to readthedocs leads to a 404.

The offending line is:

    [![docs](https://readthedocs.org/projects/allennlp/badge/?version=latest)](https://readthedocs.org/projects/allennlp/)

PR #662, opened 2018-01-04
Update README.md

schmmd:
We stopped using ReadTheDocs because it didn't support Python 3.6.  We probably should link to https://allenai.github.io/allennlp-docs/ but we don't have a badge.

Fixes https://github.com/allenai/allennlp/issues/661

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,6 @@
 
 [![Build Status](https://travis-ci.org/allenai/allennlp.svg?branch=master)](https://travis-ci.org/allenai/allennlp)
 [![codecov](https://codecov.io/gh/allenai/allennlp/branch/master/graph/badge.svg)](https://codecov.io/gh/allenai/allennlp)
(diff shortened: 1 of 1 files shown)


matt-gardner:
Looks like we might be able to go back to RTD, if we want to: https://github.com/rtfd/readthedocs.org/issues/2581, https://github.com/rtfd/readthedocs.org/pull/3339.  Is it worth looking into this?

schmmd:
My first hesitation is that I'd hate to end up with the same problem we had before.

nalourie-ai2:
My thoughts are that as long as our docs are somewhere, linked from our site, and high up on Google (they are) then people shouldn't have trouble finding them, while integrating with another service is another place for bugs to creep in.

nalourie-ai2:
Although, readthedocs has been a bit of a standard so it'd always be nice to have the documentation there.

PR #663, opened 2018-01-04
Better KnowledgeGraphField

matt-gardner:
This builds on #639 (which in turn builds on #606).  I'll rebase this once that one is merged.  It's only the most recent two commits shown in the list.

This PR makes the `KnowledgeGraphField` actually work the way that the original parser did.  We have two separate representations for each KG entity: an embedded representation of the text in the corresponding cell, and a set of linking features for each token in the question.  I haven't updated the model code to _use_ this at all - that will come next.

I've implemented most of the entity linking features from the original parser, except for the span matching ones.  Shouldn't be _too_ hard to add those, but it will be a little annoying.


schmmd:
Just an FYI, PyLint should stop failing when you merge with master.  It might be passing locally since you have some state around your requirements.  I think the fix was https://github.com/allenai/allennlp/pull/638.

matt-gardner:
Yeah, thanks.  The `wikitables` branch is a bit behind master, and I'm going to rebase it on the current master soon, but I'm waiting until these PRs are merged (and probably one of Pradeep's) to not mess up development so much.  (We're avoiding merging master into wikitables, because that really messes with the commit graph and makes it hard to reason about how the branch stands.)

matt-gardner:
Ok, @pdasigi, this one should now be ready to review.

PR #666, opened 2018-01-06
run flask with gevent

joelgrus:
maybe this will make production deployments more stable?

http://flask.pocoo.org/docs/0.12/deploying/
http://flask.pocoo.org/docs/0.12/deploying/wsgi-standalone/#gevent


schmmd:
@jpowerwa we recently switched from Sanic to Flask, but then we started having some problems with our Flask server.  It would get in a state where it was hanging.  We're still troubleshooting to find the root cause.

I understand that you guys run some long-running Flask servers.  How do you run them "in production mode"?  Do you know whether WSGI with gevent is a good idea?

I don't know anything about this stuff...

Issue #667, opened 2018-01-08
Make Span F1 match perl script exactly for SRL model


DeNeutoy:
The Span based F1 metric doesn't exactly match the perl script for evaluating the SRL model. I think this is because the perl script treats continuation and reference spans (C-B-TAG, R-B-TAG) as part of the same span as what they are continuing or referencing, whereas the allennlp metric treats these as separate spans.

DeNeutoy:
Yes, this is why - continued arguments are treated as single spans: https://github.com/allenai/allennlp/blob/master/scripts/srl-eval.pl#L1848

schmmd:
@DeNeutoy is this still an issue?

DeNeutoy:
Yes.... But we need a perl expert.

iellenberger:
@DeNeutoy - find me when I', in the office tomorrow and I'll see what I can do.

sanketvmehta:
@DeNeutoy and team: Do we have any updates on this issue? Based upon my current experiments, I see a bit discrepancy between F1-scores using the current implementation and the standard evaluation script (eval.pl).

DeNeutoy:
No, sorry. I still don't know why they are different.

matt-gardner:
Closing this, as #2743 largely makes it obsolete.

PR #670, opened 2018-01-09
handle continued spans correctly in SpanBasedF1Measure

DeNeutoy:
Fixes part of #667 - I need to actually run a full model really to see if the results are identical. It's very possible there are other quirks which are less obvious. I'm also not 100% sure i'm handling this correctly - I only went from the comment in the perl script not the actual code, because, well, perl.

### DIFF
--- a/tests/training/metrics/span_based_f1_measure_test.py
+++ b/tests/training/metrics/span_based_f1_measure_test.py
@@ -178,21 +178,13 @@ class SpanBasedF1Test(AllenNlpTestCase):
         sentence = ["Mark", "and", "Matt", "were", "running", "fast", "."]
 
         gold_indices = [self.vocab.get_token_index(x, "tags") for x in bio_tags]
-
         gold_tensor = torch.Tensor([gold_indices])
         prediction_tensor = torch.rand([1, 6, self.vocab.get_vocab_size("tags")])
-
-        # Test that the span measure ignores completely masked sequences by
-        # passing a mask with a fully masked row.
         mask = torch.LongTensor([[1, 1, 1, 1, 1, 1, 1, 1, 1]])
 
         # Make prediction so that it is exactly correct.
-        prediction_tensor[:, 0, 1] = 1
-        prediction_tensor[:, 1, 0] = 1
-        prediction_tensor[:, 2, 10] = 1
-        prediction_tensor[:, 3, 5] = 1
-        prediction_tensor[:, 4, 11] = 1
-        prediction_tensor[:, 5, 0] = 1
+        for i, tag_index in enumerate(gold_indices):
+            prediction_tensor[0, i, tag_index] = 1
 
         metric = SpanBasedF1Measure(self.vocab, "tags")
         metric(prediction_tensor, gold_tensor, mask)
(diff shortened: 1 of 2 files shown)


tests/training/metrics/span_based_f1_measure_test.py
```diff
@@ -163,3 +172,65 @@ def test_span_f1_can_build_from_params(self):
...
+    def test_span_f1_matches_perl_script_for_continued_arguments(self):
+        bio_tags = ["B-ARG1", "O", "B-C-ARG1", "B-V", "B-ARGM-ADJ", "O"]
+        sentence = ["Mark", "and", "Matt", "were", "running", "fast", "."]
+
+        gold_indices = [self.vocab.get_token_index(x, "tags") for x in bio_tags]
+
+        gold_tensor = torch.Tensor([gold_indices])
+        prediction_tensor = torch.rand([1, 6, self.vocab.get_vocab_size("tags")])
+
+        # Test that the span measure ignores completely masked sequences by
```
matt-gardner: Pretty sure this is copy pasta.

tests/training/metrics/span_based_f1_measure_test.py
```diff
@@ -163,3 +172,65 @@ def test_span_f1_can_build_from_params(self):
...
+        # passing a mask with a fully masked row.
+        mask = torch.LongTensor([[1, 1, 1, 1, 1, 1, 1, 1, 1]])
+
+        # Make prediction so that it is exactly correct.
+        prediction_tensor[:, 0, 1] = 1
```
matt-gardner: Can you replace the `:` with `0`?  Also, it'd be nice to have some reference for the magic numbers in the third index, in a comment after each line, or something.  You could even make this so it's robust to future changes by doing something like:

```python
for i, tag_index in enumerate(gold_indices):
    prediction_tensor[0, i, tag_index] = 1
```

Fewer lines of code, even =).

tests/training/metrics/span_based_f1_measure_test.py
```diff
@@ -163,3 +172,65 @@ def test_span_f1_can_build_from_params(self):
...
+        # pytest with the -v flag, you will be able to see the perl script output.
+        gold_file_path = os.path.join(self.TEST_DIR, "gold_conll_eval.txt")
+        prediction_file_path = os.path.join(self.TEST_DIR, "prediction_conll_eval.txt")
+        with open(gold_file_path, "a+") as gold_file, open(prediction_file_path, "a+") as prediction_file:
+            # Use the same bio tags as prediction vs gold to make it obvious by looking
+            # at the perl script output if something is wrong.
+            write_to_conll_eval_file(gold_file, prediction_file, 4, sentence, bio_tags, bio_tags)
+
+        perl_script_command = ["perl", "./scripts/srl-eval.pl", prediction_file_path, gold_file_path]
+        exit_code = subprocess.check_call(perl_script_command)
```
matt-gardner: Is it easy to parse the result of this?  I know it's easy in python to get the stderr / stdout; is the output format easy to read a number off of?  That'd make this a more useful test.

PR #674, opened 2018-01-10
Remove Travis configuration.

schmmd:


schmmd:
Travis has croaked so we're accelerating replacing it with TeamCity.

CONTRIBUTING.md
```diff
@@ -31,9 +31,8 @@ build will test that:
 
 If your code fails one of these checks, you will be expected to fix your pull request before it is considered.
 
-You can run most of these tests locally, which will be faster than waiting for cloud systems to run tests.  To see
-exactly what our continuous build system runs, look at `.travis.yml`.  The most common checks you should run
```
matt-gardner: Is there a place we can point to for teamcity, like we do for travis?

schmmd: I don't know of one that's available for guest accounts, so I'd rather not put anything here.

PR #675, opened 2018-01-10
lazy datasets

joelgrus:
ok, this is a totally different PR, but I think I addressed most of the feedback from the other one, and I tried to make as little breaking change as possible

* made `Dataset` into an abstract base class with `InMemory` and `Lazy` subclasses. I believe this is the only breaking change to an external facing API, as (in particular) it will break any custom `DatasetReader`s. (They just need to replace `Dataset` with `InMemoryDataset`. I put an ugly hack in to generate a RuntimeError if they don't).  Part of me wanted to call the in-memory dataset a `Batch` (that's how we use it 50% of the time), but I thought that might be confusing.
* replaced `Vocabulary.from_dataset` with `Vocabulary.from_instances`. (technically this is also a breaking change, but I doubt anyone is using this directly rather than through `commands/train`)
* got rid of `Dataset.truncate`, as nothing in the code was using it. if it's important, I can put it back, but I sort of feel like I'd rather handle that differently once I understand the use case better.
* made `DataIterator` generic over the type of Dataset. I'm not 100% sure this was necessary, but it meant that the eager DataIterators could use `InMemoryDataset` features, which I like
* changed the return type of `DataIterator._create_batches` from `List[List[Instance]]` to `Iterator[List[Instance]]`
* introduced a `LazyIterator[LazyDataset]` 
* I punted on the "number of instances in a lazy dataset" issue and just had `LazyIterator.get_num_batches` return 1 in the lazy case. That's obviously not right, but tqdm still works, it just counts instead of progress bars, which feels ok for v1. I know we had talked about passing `num_instances` as a configuration option, but I couldn't figure out a good place for it. Top-level along with `train_data_path`? Then you get that logic polluted all throughout `commands/train`, which I don't like.


joelgrus:
*another* approach, which would involve fewer breaking changes, but which is clunkier, would be to change `Dataset` to be something like

```
def __init__(self, instances: List[Instance] = None, generator: Callable[...] = None):
   ...
```

and you'd have to supply either `instances` or `generator` but not both, and then you could make all of the in-memory-only operations throw on a LazyDataset.

that's simpler in some ways but feels kind of ugly?

joelgrus:
I guess technically I don't have to change the name of `Vocabulary.from_dataset`, since you can still pass it a Dataset. My main rationale for that was just to avoid having to create a fake dataset in `commands/train`. 🤔

allennlp/data/dataset.py
```diff
@@ -170,3 +186,40 @@ def as_tensor_dict(self,
...
+        self.vocab: Vocabulary = None
+
+    @overrides
+    def index_instances(self, vocab: Vocabulary) -> None:
+        self.vocab = vocab
+
+    @overrides
+    def __iter__(self) -> Iterator[Instance]:
+        if self.vocab is None:
+            logger.warning("iterating over an unindexed dataset")
```
matt-gardner: I don't think this is necessary - if you try to use an un-indexed `Instance` to make a tensor, it'll crash.  This just adds a warning for no reason when you're building the vocab.  I was wrong to suggest it before.

allennlp/data/iterators/lazy_iterator.py
```diff
@@ -0,0 +1,92 @@
...
+        self._instances_per_epoch = instances_per_epoch
+
+        # As you might use this iterator with multiple datasets,
+        # you need to store a cursor for each one.
+        self._cursors: Dict[LazyDataset, Iterator[Instance]] = {}
+
+    @overrides
+    def get_num_batches(self, _: LazyDataset) -> int:
+        # pylint: disable=no-self-use
+        # TODO(joelgrus): Figure out the right way to handle this.
```
matt-gardner: `return instances_per_epoch // batch_size if instance_per_epoch else 1`

Right?

allennlp/data/iterators/data_iterator.py
```diff
@@ -53,26 +54,26 @@ def __call__(self,
...
+            batch = InMemoryDataset(group)
             padding_lengths = batch.get_padding_lengths()
             logger.debug("Batch padding lengths: %s", str(padding_lengths))
             logger.debug("Batch size: %d", len(batch.instances))
             yield batch.as_tensor_dict(padding_lengths,
                                        cuda_device=cuda_device,
                                        for_training=for_training)
 
-    def _create_batches(self, dataset: Dataset, shuffle: bool) -> List[List[Instance]]:
+    def _create_batches(self, dataset: DatasetType, shuffle: bool) -> Iterator[List[Instance]]:
```
matt-gardner: Maybe have this return an `Iterator[InMemoryDataset]`?  It means you can't be as fancy with the `islice` stuff you were doing, but I _think_ that doesn't work anyway, if you want to allow for a `LazyBucketIterator`.  Making `Iterator[InMemoryDataset]` the API encourages the right design decision when implementing laziness (I think), and it makes it clear that `InMemoryDataset` is a "batch".

allennlp/data/iterators/lazy_iterator.py
```diff
@@ -0,0 +1,92 @@
...
+        #   This is definitely not right, but currently this is only used for tqdm,
+        #   and the only consequence is that the tqdm for iterating over an epoch
+        #   will show as a count rather than as a progress bar.
+        return 1
+
+    def _take_instances(self, dataset: LazyDataset, max_instances: int) -> Iterator[Instance]:
```
matt-gardner: I like this quite a bit.  It'll make it easy to extend this to over-take instances, then use a `BucketIterator` to sort them, which I think is the right thing to do eventually.  And along those lines, I'd probably call this a `LazyBasicIterator`, in analogy to our `BasicIterator`, and in preparation for a `LazyBucketIterator` that will come next.

Is there a way to have this always return `max_instances`, though, looping around immediately when you get to the end?  E.g., I'm not sure why this has to return an `Iterator`, instead of a `List`.

allennlp/data/vocabulary.py
```diff
@@ -309,21 +311,21 @@ def set_from_file(self,
...
-                     only_include_pretrained_words: bool = False) -> 'Vocabulary':
+    def from_instances(cls,
+                       instances: Iterable['adi.Instance'],
+                       min_count: int = 1,
+                       max_vocab_size: Union[int, Dict[str, int]] = None,
+                       non_padded_namespaces: Sequence[str] = DEFAULT_NON_PADDED_NAMESPACES,
+                       pretrained_files: Optional[Dict[str, str]] = None,
+                       only_include_pretrained_words: bool = False) -> 'Vocabulary':
         """
         Constructs a vocabulary given a :class:`.Dataset` and some parameters.  We count all of the
```
matt-gardner: Docstring needs to be updated.

tests/commands/train_test.py
```diff
@@ -90,3 +92,87 @@ def test_train_args(self):
         with self.assertRaises(SystemExit) as cm:  # pylint: disable=invalid-name
             args = parser.parse_args(["train", "path/to/params"])
             assert cm.exception.code == 2  # argparse code for incorrect usage
+
+@DatasetReader.register('lazy-test')
+class LazyTestReader(DatasetReader):
```
matt-gardner: I _think_ having `Test` in this name is potentially confusing to pytest (e.g., it'd try to run methods on here as tests if they happened to also have `test` in their name).  Maybe call this `LazyFakeReader`?

allennlp/data/iterators/lazy_iterator.py
```diff
@@ -0,0 +1,92 @@
...
+            logger.warning("cannot shuffle a lazy dataset")
+
+        # Get an iterator for the next epoch
+        if self._instances_per_epoch is None:
+            iterator = iter(dataset)
+        else:
+            iterator = self._take_instances(dataset, self._instances_per_epoch)
+
+        # And return it in batches. See, e.g. https://stackoverflow.com/a/31170795
+        return iter(lambda: list(itertools.islice(iterator, 0, self._batch_size)), [])
```
DeNeutoy: I agree with matt above that this could return an `InMemoryDataset` or at least a list. Also, rather than linking to the stack overflow a comment here that this is an iterator with batch_size chunks which always yields an empty list as it's sentinel value would be more helpful.

allennlp/data/vocabulary.py
```diff
@@ -16,6 +16,8 @@
 from allennlp.common.params import Params
 from allennlp.common.checks import ConfigurationError
 from allennlp.common.file_utils import cached_path
+from allennlp.data import instance as adi  # pylint: disable=unused-import
```
DeNeutoy: Can you not just change the position of the imports in `__init__` of `data` so you can remove this?

allennlp/common/testing/model_test_case.py
```diff
@@ -134,11 +135,12 @@ def ensure_batch_predictions_are_consistent(self):
         self.model.eval()
         single_predictions = []
         for i, instance in enumerate(self.dataset.instances):
-            dataset = Dataset([instance])
+            dataset = InMemoryDataset([instance])
```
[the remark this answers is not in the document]
joelgrus: what if we made the base class `InstanceCollection`, as that's what that interface really requires?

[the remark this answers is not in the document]
matt-gardner: Yeah, I'm happier with `InstanceCollection` than `AbstractDataset`.

allennlp/data/iterators/lazy_iterator.py
joelgrus: yes, that way is probably a lot simpler

allennlp/data/vocabulary.py
joelgrus: I don't think so, the problem is that `Instance.index_fields` needs to know about `Vocabulary`, which means that we get a circular reference if we import `Instance` here directly.

this is the solution I got from https://www.python.org/dev/peps/pep-0484/#forward-references

allennlp/data/iterators/lazy_iterator.py
joelgrus: but I don't want it to return a list, because max_instances could still be more than fits in memory

joelgrus:
ok, I think I addressed all of the comments, the major changes were

* rename the abstract base class to `InstanceCollection`, with subclasses `Dataset` and `LazyDataset`
* change `DataIterator._create_batches()` to return an `Iterable[Dataset]`

and then various other small cleanups and documentation fixes

allennlp/data/iterators/lazy_iterator.py
matt-gardner: I was imagining that `_create_batches` would look something like this:

```python
num_instances_yielded = 0
while num_instances_yielded < self._instances_per_epoch:
    batch = Dataset(self._take_instances(dataset, self._batch_size))
    num_instances_yielded += self._batch_size
    yield batch
```

Then, in a `LazyBucketIterator`, I can have a parameter like `self._max_instances_in_memory`, and do something like:

```python
num_instances_yielded = 0
bucket_iterator = BucketIterator(self._batch_size, ...)
while num_instances_yielded < self._instances_per_epoch:
    instances_to_take = min(self._instances_per_epoch - num_instances_yielded,
                            self._max_instances_in_memory)
    instances = self._take_instances(dataset, instances_to_take)
    yield from bucket_iterator(Dataset(instances))
    num_instances_yielded += instances_to_take
```

I think this is quite a bit simpler than what you have.  The downside is that it doesn't really work if `self._instances_per_epoch` is `None`.  I think this is fine - it seems reasonable to require specifying `instances_per_epoch` if you're using a lazy dataset.

What do you think?

matt-gardner: Oh, a problem with setting `instances_per_epoch` is that you probably want these to be different for training and validation, but we don't have a good mechanism for separating them...  Hmm, I'm not sure how to get around this.

allennlp/data/iterators/lazy_basic_iterator.py
```diff
@@ -0,0 +1,96 @@
...
+        self._cursors: Dict[LazyDataset, Iterator[Instance]] = {}
+
+    @overrides
+    def get_num_batches(self, _: LazyDataset) -> int:
+        # pylint: disable=no-self-use
+        # TODO(joelgrus): Figure out the right way to handle this when `instances_per_epoch`
+        #   is not provided. This is definitely not right, but currently this is only used for tqdm,
+        #   and the only consequence is that the tqdm for iterating over an epoch
+        #   will show as a count rather than as a progress bar.
+        return self._instances_per_epoch // self._batch_size if self._instances_per_epoch else 1
```
nalourie-ai2: It seems to me that we just need to decide what the default is: 1 instance per epoch or infinity instances per epoch? Then we don't have to worry about num_instances_per_epoch being none and we just return the default here.

Personally, I think having the default be "infinity" epochs might do the right thing here, but I'm not familiar enough with how `get_num_batches` is used yet to say if it'd create bugs.

nalourie-ai2: It seems to me that we just need to decide what the default is: 1 instance per epoch or infinity instances per epoch? Then we don't have to worry about num_instances_per_epoch being none and we just return the default here.

Personally, I think having the default be "infinity" epochs might do the right thing here, but I'm not familiar enough with how `get_num_batches` is used yet to say if it'd create bugs.

allennlp/data/iterators/lazy_iterator.py
joelgrus: here is a suggestion for what's probably the "right" way to address this:

(1) construct Dataset "from_params"

right now in our config file, we don't really have a notion of a dataset, we just have a `train_data_path` and `test_data_path`, which we pass into `DatasetReader.read()`

probably, though, DatasetReader.read() should take a `Params` object that contains the file path plus other things like "instances_per_epoch" and "max_instances_in_memory" and constructs the Dataset accordingly. 

we could get backward compatibility with config files by checking first for a `train_data` key and then failing back to using `train_data_path` as we do currently. But we would break custom DatasetReaders, unless we called this method something other than `read`

(2) the iterator constructor / from_params takes a Dataset, and then we pass into `train()` the train_iterator and validation_iterator, rather than iterator + train_data + validation_data

this allows us to have different iteration behavior on different datasets, but still only have to specify the iterator params once (the different behaviors are driven by different dataset configs)

This (at least the second part) is sort of the path I was originally going doing with this, before I decided to change the API as little as possible. If we want to do this, I think I'd rather do it in a second PR though.

matt-gardner: Yeah, I think some way of sending different parameters for train and validation is necessary here.  I'm not totally sure that what you're suggesting is right (e.g., I'd probably prefer to just have two different iterator configs, instead of having the `Dataset` know about epochs and such), but it's at least close.  And you're right that this PR is great as it is, and a good first step toward what we need, and what we're talking about here goes in a separate PR.

PR #678, opened 2018-01-12
Ontonotes srl fix

DeNeutoy:


allennlp/data/dataset_readers/dataset_utils/ontonotes.py
```diff
@@ -64,7 +64,7 @@ def __init__(self,
                  word_senses: List[Optional[float]],
                  speakers: List[Optional[str]],
                  named_entities: List[str],
-                 srl_frames: Dict[str, List[str]],
```
nalourie-ai2: Just curious, what's the motivation for switching from a Dict to a list of tuples?

nalourie-ai2: Oh, I see the reason now (having read further down). Because it seemed like the cleanest way to allow duplicates of the key (as opposed to making keys map to lists of instances). That makes sense to me.

nalourie-ai2:
LGTM. That test fixture is the most important part.

PR #680, opened 2018-01-12
Allow tokenizers to do batch tokenization

matt-gardner:
This can speed up data processing with spacy by quite a bit, if you're doing more than simple tokenization.  You have to change your dataset reader code to make use of it, but it works.


matt-gardner:
Looks like I didn't test the default case in the base class...  Adding a test for that now, and fixing the bug.  The rest of it should be good, though.

matt-gardner:
I'm a little concerned that mypy [passed on TeamCity](http://build.allennlp.org/viewLog.html?buildId=1028&buildTypeId=AllenNLP_AllenNLPPullRequests&tab=buildLog&_focus=749#_state=) but [not on Travis](https://travis-ci.org/allenai/allennlp/builds/328329885?utm_source=github_status&utm_medium=notification) for an earlier commit in this PR.  @joelgrus, @schmmd, any ideas for why that might be?  A cache/version issue in a dependent library?

joelgrus:
my first guess is that teamcity and travis somehow weren't looking at the same commit? since travis is complaining about something you then fixed?

matt-gardner:
I dug into this a little bit more.  There are a few things to note:

1. TeamCity is using the commit we expect: [6f6650c04d36123727ac76b60a028292cdac6ef1](http://build.allennlp.org/viewLog.html?buildId=1028&buildTypeId=AllenNLP_AllenNLPPullRequests&tab=buildLog&_focus=33&state=749%2C30%2C32#_state=749,30,32).  When I checkout this commit myself and run mypy, it fails.  So it's quite odd to me (and concerning) that TeamCity passes here.  It sure looks like something is wrong in our configuration, as TeamCity should have failed.
2. Travis is _not_ using the same commit.  You can see on the [Travis build page](https://travis-ci.org/allenai/allennlp/jobs/328329888) that it's using commit 08ffeb5, which doesn't show up anywhere in this PR.  That's because Travis is merging the PR into master _before_ running its checks (you can see that by looking at that commit's page).  Do we want to do this on TeamCity?  I can see pros and cons of going either way, and I'm not sure what the right thing to do is here.

(pinging @schmmd, just because you've said you're less likely to miss something if you're explicitly mentioned)

matt-gardner:
I've confirmed that issue 1 above is a mypy version issue.  With mypy 0.521 (the version on my machine), the checks fail (which is correct), while with mypy 0.560 (the version in the docker image), the checks pass (which is incorrect).  I'm going to open a PR for pinning the mypy version to 0.521 until we find out more info.  In the meantime, this PR should be good to review.

PR #681, opened 2018-01-13
predictor tutorial

### LINKED ISSUE #589: Add a tutorial on adding your model to the `predict` command and to the demo

joelgrus:
Fixes #589.  I took a shortcut and used an existing model and predictor, but I think it's still informative enough?


tutorials/getting_started/so_you_built_a_model.md
```diff
@@ -0,0 +1,182 @@
...
+        """
+        Expects JSON that looks like ``{"question": "...", "passage": "..."}``.
+        """
+        question_text = json_dict["question"]
+        passage_text = json_dict["passage"]
+        return self._dataset_reader.text_to_instance(question_text, passage_text), {}
+```
+
+As you can see, it just pulls the `"question"` and `"passage"` fields out of the input JSON
+and feeds them to `text_to_instance`. (The second return parameter is an output dict
```
matt-gardner: I'd remove the parentheses around this sentence.

tutorials/getting_started/so_you_built_a_model.md
```diff
@@ -0,0 +1,182 @@
+# So You Built a Model, Now What?
+
+After you've built and trained a model, it's likely that you want to use it
+for something. In this tutorial we'll cover both
+
+* How to make predictions using your model, and
+* How to run a web demonstration of your model
+
+Here we'll be working with the [Bidaf](https://github.com/allenai/allennlp/blob/master/allennlp/models/reading_comprehension/bidaf.py)
```
matt-gardner: It's probably better to use "BiDAF" in plain text.

tutorials/getting_started/so_you_built_a_model.md
```diff
@@ -0,0 +1,182 @@
...
+
+In our case we'd have to call
+
+```python
+    main(prog="python run.py",
+         predictor_overrides={'bidaf': 'machine-comprehension'})
+```
+
+to specify that when we see a `bidaf` model and need a `Predictor`
+we should use the `machine-comprehension` one.
```
matt-gardner: It'd be nice to remind the reader here why we're using these particular strings - they're what we registered our classes with.

tutorials/getting_started/so_you_built_a_model.md
```diff
@@ -0,0 +1,182 @@
...
+```
+{"question": "First question...", "passage": "First passage..."}
+{"question": "Second question...", "passage": "Second passage..."}
+...
+```
+
+If you run your script, you should get a help message:
+
+```bash
+$ run.py predict
```
matt-gardner: `python run.py predict`, right?

tutorials/getting_started/so_you_built_a_model.md
```diff
@@ -0,0 +1,182 @@
...
+usage: python -m allennlp.run predict [-h]
```
matt-gardner: This usage info is wrong - you've set `prog="python run.py"`, so you should get `usage: python run.py predict [-h]...`

tutorials/getting_started/so_you_built_a_model.md
```diff
@@ -0,0 +1,182 @@
...
+
+In the simplest case, just specify the `archive_file`
+(that's the trained model we downloaded above)
+and the `input_file`
+(that's the JSON lines file above)
+and your predictions will get printed to the console.
+If you'd rather have them in a file, specify the `--output-file` parameter.
+
+It's also possible to use non-JSON formats, for example, if you had CSV data.
+In this case your `Predictor` would need to override `load_line` and `dump_line`.
```
matt-gardner: Links to the API documentation would be good here.

tutorials/getting_started/so_you_built_a_model.md
```diff
@@ -0,0 +1,182 @@
...
+
+```python
+from allennlp.models.archival import load_archive
+from allennlp.service.predictors import Predictor
+from allennlp.service.server_simple import make_app
+
+from gevent.wsgi import WSGIServer
+
+archive = load_archive('/path/to/trained/bidaf/model.tar.gz')
+predictor = Predictor.from_archive(archive, 'machine-comprehension')
```
matt-gardner: If someone has a third-party model and predictor, these lines won't work.  You need to import the specific model and predictor classes.  It'd be good to have those imports explicitly given above, with a comment that they would need to change in your code.

tutorials/getting_started/so_you_built_a_model.md
```diff
@@ -0,0 +1,182 @@
...
+               field_names=['passage', 'question'])
+
+http_server = WSGIServer(('0.0.0.0', 8888), app)
+http_server.serve_forever()
+```
+
+The important things to notice are
+
+* We had to explicitly specify which predictor to use. If you were
+  using a custom one, you'd need to import it at the top of the script.
```
matt-gardner: Ah, I see that you do say this here; I think you should have it in the code, because we're pretending that the classes aren't there.

tutorials/getting_started/so_you_built_a_model.md
```diff
@@ -0,0 +1,182 @@
...
+* We had to explicitly specify the field names. These are used to generate
+  the web interface but also (more importantly) as the keys in the input JSON,
```
matt-gardner: This sentence wasn't clear to me the first time I read it.  Maybe it's because you have "used to generate" coordinated with "[used] as the keys", which is a pretty rare construct.  A simpler sentence structure here would probably make this more clear.  Maybe even just repeating "are used" before "as the keys" would be enough...

tutorials/getting_started/so_you_built_a_model.md
```diff
@@ -0,0 +1,182 @@
...
+  which means they need to match what your `Predictor` expects
```
matt-gardner: Missing a period.

tutorials/getting_started/so_you_built_a_model.md
```diff
@@ -0,0 +1,182 @@
...
+* We specified port 8888, but you could have specified any other port.
+
+If you run this script and visit `localhost:8888` in your browser,
+you'll see a moderately attractive demo that displays the raw JSON
+results:
+
+![demo screenshot](simple_server.png)
+
+It's possible this won't quite suit your needs, in which case
+`make_app` [provides a few more options](https://github.com/allenai/allennlp/blob/master/allennlp/service/server_simple.py#L25)
```
matt-gardner: Missing a period.

tutorials/getting_started/so_you_built_a_model.md
```diff
@@ -0,0 +1,182 @@
...
+For example, you might want to change the page title to something other than "AllenNLP Demo".
+If, so, pass in a `title="Something else"` parameter.
+
+Or you might want to edit the returned JSON. Maybe you want to remove some of its verbose
+fields, or maybe you want to add a new nicely formatted field. In that case you can pass
+a `sanitizer=sanitizer` parameter, with a JSON-in-JSON-out function that contains the editing
+logic.
+
+If you're really motivated, you might want to just provide your own HTML page.
+(If so, I suggest using ours as a starting point, as it already contains the JavaScript
```
matt-gardner: I'd recommend removing the parentheses from this sentence and putting it at the end of the paragraph.  If you do this, you'd probably want to change "If so" to something that works better with the sentence being at the end.  Also, you used "we" at the top of the tutorial, and "I" here - it's probably best to be consistent in using one or the other.

matt-gardner:
Actually, thinking about this a little more:

It'd be nice to show how to take the existing BiDAF demo and replicate it in the simple server like this.  Like, say I want to tweak something in the demo and show my model, how would I do it?  You hint at this in the last section, but it'd be nice to add another section that actually does it.  If it we go that route, it'd also be nice to have a separate repository to point to that actually shows this being done, like with the "using allennlp in your repo" tutorial.

Hmm, maybe it would be best to just use the same repo, and add a predictor and a demo to it...  I think what you have is great, and very helpful.  I think it might be even more helpful, though, to unify these tutorials a little more, and give a single place to show how to make the most out of AllenNLP in your own code.  What do you think?  If we do this, it'd basically just be replacing the BiDAF bits of code with new code for the simple model my example repo - the text and format that you have is great, and wouldn't have to change.

So, then, those are two separate suggestions:
1. Actually show how to make nicer-looking output by copying the existing HTML.
2. Use the example external repo instead of BiDAF, so that the tutorials are more unified, and so that external users have a clear example that they can actually checkout and run commands from.

tutorials/getting_started/so_you_built_a_model.md
```diff
@@ -0,0 +1,210 @@
+# So You Built a Model, Now What?
+
+After you've built and trained a model, it's likely that you want to use it
+for something. In this tutorial we'll cover both
+
+* How to make predictions using your model, and
+* How to run a web demonstration of your model
+
+Here we'll be working the S2 classification model
```
schmmd: I would call it the "paper classification model" rather than using an internal project code name.

joelgrus:
ok, this already got approved twice, but I kind of rewrote it since then (to use the paper-classifier model, and to customize the HTML), so speak up soon or I'm merging it.

@schmmd @matt-gardner

tutorials/getting_started/so_you_built_a_model.md
```diff
@@ -0,0 +1,340 @@
...
+Our JSON "visualization" is not particularly impressive.
+We can customize it by putting our own `index.html`
+(and any other files we need) in some directory
+and calling `server_simple` with the `--static-dir` flag.
+In this case we don't need to provide a title or field names,
+as those will be implicitly provided by the HTML code.
+
+The simplest way to get started is to just "view source" on the demo
+and save the resulting file in some directory. I called my directory
+[`static_html`](https://github.com/allenai/allennlp-as-a-library-example/tree/master/static_html)
```
matt-gardner: Can we point to a tag on this repo, instead of master?

joelgrus: should I tag what's there as v0.4.0?

matt-gardner: Yeah, sounds good to me.

matt-gardner:
Looks great.  My only other comment is that we probably want a name that's more obvious about the content of the tutorial.  I like your title, but I don't think a user on allennlp.org/tutorials will know in what circumstances they should click on the link.  A title that mentions something about making predictions and building demos would be more informative.

PR #682, opened 2018-01-15
Entity Embedding

rajasagashe:
This contains code for entity embedding, and commented pseudocode for entity linking.

### DIFF
--- a/allennlp/models/encoder_decoders/wikitables_semantic_parser.py
+++ b/allennlp/models/encoder_decoders/wikitables_semantic_parser.py
@@ -170,11 +170,10 @@ class WikiTablesSemanticParser(Model):
 
         # compute the average of each entity's word embeddings
         # (batch_size, num_entities, embedding_dim)
-        encoded_table = self._compute_average_with_encoder(embedded_table, table_mask,
-                                                           self._entity_encoder)
+        encoded_table = self._compute_average_with_encoder(embedded_table, table_mask)
 
         # (batch_size, num_entities, num_entities)
-        neighbor_idx_tensor = self._get_neighbor_indexes(world, encoded_table.size()[1])
+        neighbor_idx_tensor = self._get_neighbor_indexes(world, encoded_table.size(1))
 
         # get averaged word embeddings for each entity's neighbors
         # (batch_size, num_entities, num_entities, embedding_dim)
@@ -184,8 +183,7 @@ class WikiTablesSemanticParser(Model):
 
         # get averaged neighbor embeddings for each entity
         # (batch_size, num_entities, embedding_dim)
-        neighbor_embeddings = self._compute_average_with_encoder(neighbor_embed_tensor,
-                                                                 neighbor_mask, self._entity_encoder)
+        neighbor_embeddings = self._compute_average_with_encoder(neighbor_embed_tensor, neighbor_mask)
 
         # each entity has a binary flag indicating type
         # (batch_size, num_entities)
@@ -205,27 +203,12 @@ class WikiTablesSemanticParser(Model):
         linking_features = table['linking']  # pylint: disable=unused-variable
         # TODO(mattg): multiply this by the feature weights, use this and the text embedding above
         # to compute a linking scores.
-        similarity = CosineSimilarity()
 
-        # todo dot product similarity
-
-        # STEP 1 expand both embedded question and table since cosine similarity requires
-        # the same dimensions
-        # embedded_question (batch_size, num_question_tokens, embedding_dim) ->
-        # (batch_size, num_entities, num_question_tokens, num_entity_tokens, embedding_dim)
-
-        # todo bmm transpose to (batch_size, embedding_dim, num_question_tokens)
-
+        # STEP 1 similarity through a dot product
+        # bmm transpose to (batch_size, embedding_dim, num_question_tokens)
         # (b, entity, entity token, E) x (b, E, Q) = (b, entity, entity token, Q)
 
-        embedded_question = embedded_question.unsqueeze(2)
-        embedded_question = embedded_question.expand(batch_size,11,6,25) # todo change numbers
-        # embedded table (batch_size, num_entities, num_entity_tokens, embedding_dim) ->
-        # (batch_size, num_entities, num_question_tokens, num_entity_tokens, embedding_dim)
-
         # STEP 2 similarity and max
-        # compute cosine similarity across embedding dim dimension
-        # (batch_size, num_entities, num_question_tokens, num_entity_tokens
         # compute max across num_entity_tokens dimension for tensor of size
         # (batch_size, num_entities, num_question_tokens)
 
@@ -249,7 +232,6 @@ class WikiTablesSemanticParser(Model):
         # compute link embedding
         # probs = (batch_size, num_question_tokens, num_entities)
         # embeddings = (batch_size, num_entities, embedding_dim)
-        # todo I want to do probs x embeddings but how to deal with batch size
         # l_{i} =  (batch_size, num_question_tokens, embedding_dim)
 
         # TODO(mattg): we need to actually compute this with the stuff mentioned above.  This
@@ -358,17 +340,18 @@ class WikiTablesSemanticParser(Model):
             return outputs
 
     def _compute_average_with_encoder(self, embed_tensor: Variable,
-                                      mask: Variable, encoder: BagOfEmbeddingsEncoder) -> Variable:
+                                      mask: Variable) -> Variable:
+        # computes the average along dimension 1 of the embed_tensor
         dims = embed_tensor.size()
-        table_mask = mask.view(-1, mask.size()[2])
-        averaged_tensor = self._entity_encoder(embed_tensor.view(-1, dims[2], dims[3]), table_mask)
+        table_mask = mask.view(-1, mask.size(2))
+        averaged_tensor = self._entity_encoder(embed_tensor.view(-1, dims[2], dims[3]),
+                                               table_mask)
         # (batch_size, num_entities, embedding_dim)
-        averaged_tensor = averaged_tensor.view(dims[0], dims[1], dims[3])
-        return averaged_tensor
+        return averaged_tensor.view(dims[0], dims[1], dims[3])
 
     def _get_neighbor_indexes(self, world: List[WikiTablesWorld],
                                     max_num_entities: int) -> Variable:
-        # find each entity's neighbors
+        # returns each entity's neighbor indexes
         batches = []
         for w in world:
             # each batch has it's own world aka table
@@ -382,18 +365,23 @@ class WikiTablesSemanticParser(Model):
                 # pad to length of max number of entities in a batch
                 padded = pad_sequence_to_length(curr_neighbors, max_num_entities)
                 neighbor_indexes.append(padded)
+            # pad the list with empty lists to hit length max num entities
+            neighbor_indexes = pad_sequence_to_length(neighbor_indexes,
+                                                      max_num_entities,
+                                                      lambda:[0]*max_num_entities)
             batches.append(neighbor_indexes)
         # (batch_size, num_entities, num_entities)
         return Variable(torch.LongTensor(batches))
 
     def _get_type_vector(self, world: List[WikiTablesWorld]) -> Variable:
-        # compute type vector (batch, num_entities) with 0,1 for types e and <e,r> respectively
+        # compute type vector with 0,1 for types e and <e,r> respectively
         batches = []
         for w in world:
             types = []
             ent2type = w.local_type_signatures
             for e in w.table_graph.entities:
-                types.append((1 if ent2type[w.local_name_mapping[e]] == 'e' else 0))  # only 2 types 'e' and '<e,r>'
+                # 1 for 'e' and 0 for '<e,r>'
+                types.append((1 if ent2type[w.local_name_mapping[e]] == 'e' else 0))
             batches.append(types)
         # (batch_size, num_entities)
         return Variable(torch.LongTensor(batches))
--- a/allennlp/models/encoder_decoders/wikitables_semantic_parser.py
+++ b/allennlp/models/encoder_decoders/wikitables_semantic_parser.py
@@ -432,11 +432,10 @@ class WikiTablesSemanticParser(Model):
                                                                            num_question_tokens,
                                                                            num_entities)))
 
-        mask_zeros = Variable(linking_scores.data.new(batch_size, 1).fill_(0).long())
         for batch_index, world in enumerate(worlds):
             cell_type_index, row_type_index = [0], [0]
             entities = world.table_graph.entities
-            for entity_index, entity in enumerate(entities):
+            for entity_index, _ in enumerate(entities):
                 if entity_type_dict[batch_index * num_entities + entity_index] == 0:
                     cell_type_index.append(entity_index)
                 else:
@@ -462,10 +461,12 @@ class WikiTablesSemanticParser(Model):
             probabilities_all = Variable(linking_scores.data.new(num_question_tokens, num_entities).fill_(0))
 
             for question_index in range(num_question_tokens):
-                for entity_index, entity_probability in list(zip(cell_type_index, probabilities_cell[question_index]))[1:]:
+                for entity_index, entity_probability in list(zip(cell_type_index,
+                                                                 probabilities_cell[question_index]))[1:]:
                     # Shift indexes by 1 to make sure null entity at index 0 gets 0 probability.
                     probabilities_all[question_index, (entity_index)] = entity_probability
-                for entity_index, entity_probability in list(zip(row_type_index, probabilities_row[question_index]))[1:]:
+                for entity_index, entity_probability in list(zip(row_type_index,
+                                                                 probabilities_row[question_index]))[1:]:
                     probabilities_all[question_index, (entity_index)] = entity_probability
             batch_probabilities[batch_index] = probabilities_all
         return batch_probabilities * question_mask.unsqueeze(-1).float()
--- a/allennlp/models/encoder_decoders/wikitables_semantic_parser.py
+++ b/allennlp/models/encoder_decoders/wikitables_semantic_parser.py
@@ -141,7 +141,7 @@ class WikiTablesSemanticParser(Model):
         # pylint: disable=arguments-differ
         # pylint: disable=unused-argument
         """
-        Decoder logic for producing the entire target sequence.
+        Encoder and decoder logic for producing the entire target sequence.
 
         Parameters
         ----------
@@ -204,13 +204,11 @@ class WikiTablesSemanticParser(Model):
         sim_dot_prod = torch.bmm(embedded_table.view(batch_size, -1, self._embedding_dim),
                                  torch.transpose(embedded_question, 1, 2))
         sim_dot_prod = sim_dot_prod.view(batch_size, num_entities, num_entity_tokens, num_question_tokens)
+        # Note the 0th index is accessed below since max returns a tuple of a float and long tensor.
         # (batch_size, num_entities, num_question_tokens)
         max_sim = torch.max(sim_dot_prod, 2)[0]
-
         # (batch_size, num_entities, num_question_tokens, num_features)
         linking_features = table['linking']
-
-        # (batch_size, num_entities, num_question_tokens)
         linking_scores = max_sim + self._linking_params(linking_features.float()).squeeze(3)
 
         # (batch_size, num_entities+1, num_question_tokens)
@@ -435,20 +433,20 @@ class WikiTablesSemanticParser(Model):
                                                  dim=0,
                                                  index=Variable(tensor.data.new(row_type_index)).long())
 
-            probabilities_one = torch.nn.functional.softmax(torch.transpose(cell_type_scores, 0, 1), dim=0)
-            probabilities_two = torch.nn.functional.softmax(torch.transpose(row_type_scores, 0, 1), dim=0)
+            probabilities_cell = torch.nn.functional.softmax(torch.transpose(cell_type_scores, 0, 1), dim=0)
+            probabilities_row = torch.nn.functional.softmax(torch.transpose(row_type_scores, 0, 1), dim=0)
             # (num_entities_per_type, num_question_tokens)
-            probabilities_one = torch.transpose(probabilities_one, 0, 1)
-            probabilities_two = torch.transpose(probabilities_two, 0, 1)
+            probabilities_cell = torch.transpose(probabilities_cell, 0, 1)
+            probabilities_row = torch.transpose(probabilities_row, 0, 1)
 
             # Adding 1 to all for 0 probability of the null entity.
             # (num_entities+1, num_question_tokens)
-            probabilities_mat = Variable(tensor.data.new(torch.zeros(num_entities+1, num_question_tokens)))
-            for ind, mat in zip(cell_type_index, probabilities_one):
-                probabilities_mat[ind+1] = mat
-            for ind, mat in zip(row_type_index, probabilities_two):
-                probabilities_mat[ind+1] = mat
-            batch_probabilities[batch_index] = probabilities_mat
+            probabilities_all = Variable(tensor.data.new(torch.zeros(num_entities+1, num_question_tokens)))
+            for ind, mat in zip(cell_type_index, probabilities_cell):
+                probabilities_all[ind+1] = mat
+            for ind, mat in zip(row_type_index, probabilities_row):
+                probabilities_all[ind+1] = mat
+            batch_probabilities[batch_index] = probabilities_all
 
         return batch_probabilities
 
--- a/allennlp/models/encoder_decoders/wikitables_semantic_parser.py
+++ b/allennlp/models/encoder_decoders/wikitables_semantic_parser.py
@@ -183,13 +183,13 @@ class WikiTablesSemanticParser(Model):
 
         # Compute the average of each entity's word embeddings.
         # (batch_size, num_entities, embedding_dim)
-        embedded_table = self._entity_encoder(embedded_table, table_mask)
+        encoded_table = self._entity_encoder(embedded_table, table_mask)
 
         # (batch_size, num_entities, num_entities)
-        neighbor_indices = self._get_neighbor_indexes(world, num_entities, embedded_table)
+        neighbor_indices = self._get_neighbor_indexes(world, num_entities, encoded_table)
 
         # (batch_size, num_entities, num_entities, embedding_dim)
-        embedded_neighbors = util.batched_index_select(embedded_table, neighbor_indices)
+        embedded_neighbors = util.batched_index_select(encoded_table, neighbor_indices)
         neighbor_mask = util.get_text_field_mask({'ignored': neighbor_indices},
                                                  num_wrapping_dims=1).float()
 
@@ -197,56 +197,45 @@ class WikiTablesSemanticParser(Model):
         embedded_neighbors = self._neighbor_encoder(embedded_neighbors, neighbor_mask)
 
         # (batch_size, num_entities, num_types)
-        type_vector, entity_types = self._get_type_vector(world, num_entities, embedded_table)
+        type_vector, entity_types = self._get_type_vector(world, num_entities, encoded_table)
 
         x = self._type_params(type_vector.float())
         y = self._neighbor_params(embedded_neighbors.float())
         # (batch_size, num_entities, embedding_dim)
         entity_embeddings = torch.nn.functional.tanh(x + y)
 
-        # (batch_size, num_entities, num_question_tokens, num_features)
-        linking_features = table['linking']  # pylint: disable=unused-variable
-        # TODO(mattg): multiply this by the feature weights, use this and the text embedding above
-        # to compute a linking scores.
+        # compute entity and question word similarity through a dot product
+        # (b, entity, entity_token, E) x (b, E, Q) = (b, entity, entity token, Q)
+        sim_dot_prod = torch.bmm(embedded_table.view(batch_size, -1, self._embedding_dim),
+                                 torch.transpose(embedded_question, 1, 2))
+        sim_dot_prod = sim_dot_prod.view(batch_size, num_entities, num_entity_tokens,
+                                         num_question_tokens)
 
-        # STEP 1 similarity through a dot product
-        # bmm transpose to (batch_size, embedding_dim, num_question_tokens)
-        # (b, entity, entity token, E) x (b, E, Q) = (b, entity, entity token, Q)
-
-        # STEP 2 similarity and max
         # compute max across num_entity_tokens dimension for tensor of size
         # (batch_size, num_entities, num_question_tokens)
+        max_sim = torch.max(sim_dot_prod, 2)[0]
 
-        # STEP 3 compute s(e,i) by combining with linking features
-        # combine with linking features for:
+        # generate entity linking score through linear combination of features
+        # and max similarity values
+        # (batch_size, num_entities, num_question_tokens, num_features)
+        linking_features = table['linking']
+        x = self._linking_params(linking_features.float())
         # (batch_size, num_entities, num_question_tokens)
+        linking_scores = max_sim + x.squeeze(3)
 
-        # STEP 4 compute exp of all s(e,i)'s
+        # softmax over each of the arrays for p(e|i,t)
         # (batch_size, num_entities, num_question_tokens)
+        softmax_scores = self._get_prob_entity(world, linking_scores, batch_size,num_entities,
+                                               num_question_tokens, encoded_table)
 
-        # STEP 5
-        # swap the entity and question token dimensions for easier normalization
-        # (batch_size, num_question_tokens, num_entities)
+        # transpose scores (batch_size, num_question_tokens, num_entities)
+        # bmm with re (batch_size, num_entities, embedding_dim) for li!
+        # (batch_size, num_question_tokens, embedding_dim)
+        link_embedding = torch.bmm(torch.transpose(softmax_scores, 1, 2), entity_embeddings)
 
-        # STEP 6
-        # compute normalized exp for p(e|i,t) through summing along num_entities, expanding
-        # then dividing all elements by the sum
-        # (batch_size, num_question_tokens, num_entities)
-
-        # STEP 7
-        # compute link embedding
-        # probs = (batch_size, num_question_tokens, num_entities)
-        # embeddings = (batch_size, num_entities, embedding_dim)
-        # l_{i} =  (batch_size, num_question_tokens, embedding_dim)
-
-        # TODO(mattg): we need to actually compute this with the stuff mentioned above.  This
-        # represents how well each question token matches table entities.  I'm stubbing this out
-        # for now, so that I have a variable to pass to the decoder, because we need this there.
-        # (batch_size, num_entities, num_question_tokens)
-        linking_scores: torch.FloatTensor = Variable(torch.ones(batch_size, num_entities, num_question_tokens))
-
+        encoder_input = torch.cat([link_embedding, embedded_question], 2)
         # (batch_size, question_length, encoder_output_dim)
-        encoder_outputs = self._encoder(embedded_question, question_mask)
+        encoder_outputs = self._encoder(encoder_input, question_mask)
 
         # This will be our initial hidden state and memory cell for the decoder LSTM.
         final_encoder_output = encoder_outputs[:, -1]  # (batch_size, encoder_output_dim)
@@ -389,10 +378,7 @@ class WikiTablesSemanticParser(Model):
         A ``torch.autograd.Variable`` with shape ``(batch_size, num_entities, num_types)``.
         entity_types : ``Dict[int,str]``
             This is a mapping from (batch_index, entity_index) to entity type id.
-
         """
-
-
         entity_types = {}
         batch_types = []
         for batch_index, world in enumerate(worlds):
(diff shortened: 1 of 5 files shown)


allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -145,25 +152,88 @@ def forward(self,  # type: ignore
...
+        neighbor_mask = util.get_text_field_mask({'tokens': neighbor_idx_tensor},
+                                                 num_wrapping_dims=1).float()
+
+        # get averaged neighbor embeddings for each entity
+        # (batch_size, num_entities, embedding_dim)
+        neighbor_embeddings = self._compute_average_with_encoder(neighbor_embed_tensor, neighbor_mask)
+
+        # each entity has a binary flag indicating type
+        # (batch_size, num_entities)
+        type_vector = self._get_type_vector(world, max_num_entities)
```
matt-gardner: If you look below, you'll see a block of code starting with `entity_types = {}`.  It'd be good to do that operation and the operation you're performing here at the same time.  Maybe you could move that block into this method, and return that dictionary along with the tensor you're constructing here?

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -145,25 +152,88 @@ def forward(self,  # type: ignore
...
+
+        # construct the entity embeddings through a linear combination of type
+        # and neighbor embeddings fed through tanh
+        type_params = torch.nn.Linear(num_types, embedding_dim)
+        neighbor_params = torch.nn.Linear(embedding_dim, embedding_dim)
+        x = type_params(type_vector.float())
+        y = neighbor_params(neighbor_embeddings.float())
+        tan = torch.nn.Tanh()
+        # (batch_size, num_entities, embedding_dim)
+        entity_embeddings = tan(x + y)
```
matt-gardner: You can just do [`entity_embeddings = torch.nn.functional.tanh(x + y)`](http://pytorch.org/docs/master/nn.html#torch.nn.functional.tanh), instead of instantiating a `Tanh` object.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -269,6 +339,57 @@ def forward(self,  # type: ignore
             # TODO(matt): compute accuracy here.
             return outputs
 
+    def _compute_average_with_encoder(self, embed_tensor: Variable,
```
matt-gardner: As I noted above, I think this function can be removed once you use `TimeDistributed`, which already implements what you did here.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -269,6 +339,57 @@ def forward(self,  # type: ignore
...
+                                      mask: Variable) -> Variable:
+        # computes the average along dimension 1 of the embed_tensor
+        dims = embed_tensor.size()
+        table_mask = mask.view(-1, mask.size(2))
+        averaged_tensor = self._entity_encoder(embed_tensor.view(-1, dims[2], dims[3]),
+                                               table_mask)
+        # (batch_size, num_entities, embedding_dim)
+        return averaged_tensor.view(dims[0], dims[1], dims[3])
+
+    def _get_neighbor_indexes(self, world: List[WikiTablesWorld],
```
matt-gardner: Indentation is off here - `world` should be on the next line, and `max_num_entities` should be aligned with `self` and `world`.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -269,6 +339,57 @@ def forward(self,  # type: ignore
...
+                                    max_num_entities: int) -> Variable:
+        # returns each entity's neighbor indexes
```
matt-gardner: Most of the other functions here have docstrings.  Can you replace this comment with a docstring formatted like those?  See, e.g., `_embed_actions` or `_get_unique_elements`.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -269,6 +339,57 @@ def forward(self,  # type: ignore
...
+        batches = []
```
matt-gardner: This is processing the instances in a single batch, not multiple batches.  And `batches` as a variable name doesn't really say anything about what the variable contains, only that it's a collection.  Something like `batch_neighbors` would be more informative here.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -269,6 +339,57 @@ def forward(self,  # type: ignore
...
+        for w in world:
```
matt-gardner: Avoid abbreviations.  I'd recommend changing the input parameter from `world` to `worlds`, and then you can use `world` here where you're currently using `w`.  `model.forward()` uses the singular `world` for data processing reasons...

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -269,6 +339,57 @@ def forward(self,  # type: ignore
...
+            # each batch has it's own world aka table
```
matt-gardner: [Use capitalization and periods in comments](https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar).  Also, it's each "batch instance", not each "batch": I'd recommend changing this to something like `# Each batch instance has its own world, which has a corresponding table.`  The `World` is more than just the table - it has stuff related to the grammar and the question that was asked.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -269,6 +339,57 @@ def forward(self,  # type: ignore
...
+            entities = w.table_graph.entities
+            entity2idx = {e: i for i, e in enumerate(entities)}
```
matt-gardner: `entity_indices = {entity: i for i, entity in enumerate(entities)}`

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -269,6 +339,57 @@ def forward(self,  # type: ignore
...
+            neighbors = w.table_graph.neighbors
+            neighbor_indexes = []
```
matt-gardner: `s/indexes/indices/`

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -269,6 +339,57 @@ def forward(self,  # type: ignore
...
+            for e in entities:
```
matt-gardner: `s/e/entity/`

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -269,6 +339,57 @@ def forward(self,  # type: ignore
...
+                # get indexes of entity e's neighbors
+                curr_neighbors = [entity2idx[cn] for cn in neighbors[e]]
```
matt-gardner: `entity_neighbors`?

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -269,6 +339,57 @@ def forward(self,  # type: ignore
...
+                # pad to length of max number of entities in a batch
+                padded = pad_sequence_to_length(curr_neighbors, max_num_entities)
+                neighbor_indexes.append(padded)
+            # pad the list with empty lists to hit length max num entities
```
matt-gardner: Because the name of the function is already pretty descriptive, this comment doesn't add much.  If you want a comment here, you could say something about _why_ we do this, e.g., because we're constructing a tensor later, but I'm not sure that's really necessary, either.  Same goes for a bunch of the comments in here - in general, [avoid comments that just say _what_ the code is doing](https://google.github.io/styleguide/cppguide.html#Implementation_Comments), unless it's not obvious.  Helpful comments give more context or explain why a particular decision was made.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -269,6 +339,57 @@ def forward(self,  # type: ignore
...
+            neighbor_indexes = pad_sequence_to_length(neighbor_indexes,
+                                                      max_num_entities,
+                                                      lambda:[0]*max_num_entities)
+            batches.append(neighbor_indexes)
+        # (batch_size, num_entities, num_entities)
+        return Variable(torch.LongTensor(batches))
```
matt-gardner: You can't actually create variables this way, because they might end up on the wrong device..  You need to do the `.data.new()` stuff seen here:
https://github.com/allenai/allennlp/blob/f9db6bc46e43930577e1cd6dc0fa8eb444a73681/allennlp/models/encoder_decoders/wikitables_semantic_parser.py#L1017

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -269,6 +339,57 @@ def forward(self,  # type: ignore
...
+
+    def _get_type_vector(self, world: List[WikiTablesWorld], max_num_entities: int) -> Variable:
+        # compute type one hot vector with dim 2 for types e and <e,r>
+        batches = []
+        for w in world:
```
matt-gardner: Same comments here about variable names and docstrings.

tests/models/encoder_decoders/wikitables_semantic_parser_test.py
```diff
@@ -25,6 +25,18 @@ def setUp(self):
     def test_model_can_train_save_and_load(self):
         self.ensure_model_can_train_save_and_load(self.param_file)
 
+    def test_average_encoder_works_simple(self):
```
matt-gardner: Because the function you're testing here is going away with a switch to `TimeDistributed`, you can remove this test.  But it'd be good to add tests for `_get_neighbor_indices` and `_get_type_vector`.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
matt-gardner: Hmm, that's a good point.  Two options: (1) pass in a tensor here that you can copy from, and just say in the docstring that the only reason we have it is to copy it, or (2) return the padded array from this method, and actually make a tensor out of it in `forward`, using one of the tensors that's there.  I'd probably pick 1, just so the method is self-contained, but I could go either way.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -147,41 +166,72 @@ def forward(self,  # type: ignore
...
 
-        # TODO(mattg): we need to actually compute this with the stuff mentioned above.  This
-        # represents how well each question token matches table entities.  I'm stubbing this out
-        # for now, so that I have a variable to pass to the decoder, because we need this there.
-        # (batch_size, num_entities, num_question_tokens)
-        num_entities = embedded_table_text.size(1)
-        num_question_tokens = embedded_input.size(1)
-        linking_scores: torch.FloatTensor = Variable(torch.ones(batch_size, num_entities, num_question_tokens))
+        # (batch_size, num_entities+1, num_question_tokens)
+        softmax_scores = self._get_prob_entity(world, linking_scores, batch_size, num_entities,
```
matt-gardner: Looking at this function, without having seen the method definition, I don't have a good idea of what `get_prob_entity` that returns `softmax_scores` is actually doing.  It's good to give methods names that give the reader a good idea of what the function does, even without having looked at the method.

Having now looked at the method, I would probably call this `_get_linking_probabilities`, `_normalize_linking_scores`, or something similar, with the returned variable just called `linking_probabilities`.  `softmax_scores` could be lots of things, as "softmax" is just a mathematical operation, not something that has semantic content.  It's better to give variables names related to their semantics, not from what operations produced them.

Also, the `linking_scores` tensor already has `batch_size`, `num_entities`, and `num_question_tokens` encoded in its size - passing those variables to this method is redundant, because you can just read them off of `linking_scores` in the method.

Lastly, I think it's more natural to have a normalized tensor like this have its last dimension be the one that's normalized.  So `(batch_size, num_question_tokens, num_entities + 1)` seems a lot more intuitive to me than how you have it.  Your need to do a transpose down below is more evidence of this to me.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -147,41 +166,72 @@ def forward(self,  # type: ignore
...
+                                               num_question_tokens, encoded_table)
+
+        null_entity_embedding = Variable(entity_embeddings.data.new(torch.zeros(batch_size, 1,
```
matt-gardner: You can just do `Variable(entity_embeddings.data.new(batch_size, 1, self._embedding_dim).fill_(0))`.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
...
+            batch_types.append(padded)
+        return Variable(tensor.data.new(batch_types)), entity_types
+
+    def _get_prob_entity(self,
+                         worlds: List[WikiTablesWorld],
+                         scores: Variable,
+                         batch_size: int,
+                         num_entities: int,
+                         num_question_tokens: int,
+                         tensor: torch.Tensor) -> Variable:
```
matt-gardner: This needs to take into account the question mask and the entity mask.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -147,41 +166,72 @@ def forward(self,  # type: ignore
...
+        softmax_scores = self._get_prob_entity(world, linking_scores, batch_size, num_entities,
+                                               num_question_tokens, encoded_table)
+
+        null_entity_embedding = Variable(entity_embeddings.data.new(torch.zeros(batch_size, 1,
+                                                                                self._embedding_dim)))
+        # (batch_size, num_entities+1, embedding_dim)
+        entity_embeddings = torch.cat([null_entity_embedding, entity_embeddings], 1)
+
+        # (batch_size, num_question_tokens, embedding_dim)
+        link_embedding = torch.bmm(torch.transpose(softmax_scores, 1, 2), entity_embeddings)
```
matt-gardner: I think it's more readable here to use `util.weighted_sum(entity_embeddings, linking_probabilities)`.  In that case this just does a `torch.bmm` like you have it, but it's more self-documenting than wondering why `torch.bmm()` is the right function here.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -107,6 +120,12 @@ def __init__(self,
         self._action_padding_index = -1  # the padding value used by IndexField
         action_embedding_dim = nonterminal_embedder.get_output_dim() * 2
         num_entity_types = 2  # TODO(mattg): get this in a more principled way somehow?
+        self._embedding_dim = question_embedder.get_output_dim()
+        self._type_params = torch.nn.Linear(num_entity_types, self._embedding_dim)
+        self._neighbor_params = torch.nn.Linear(self._embedding_dim, self._embedding_dim)
+        # TODO(mattg): ok to leave number 1 since it gets squeezed out?
```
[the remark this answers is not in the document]
rajasagashe: I think it requires two parameters, I'll add a comment explaining the 1?

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -147,41 +166,72 @@ def forward(self,  # type: ignore
...
+        num_entity_tokens = embedded_table.size(2)
+        num_question_tokens = embedded_question.size(1)
+
+        # (batch_size, num_entities, embedding_dim)
+        encoded_table = self._entity_encoder(embedded_table, table_mask)
+        # (batch_size, num_entities, num_entities)
+        neighbor_indices = self._get_neighbor_indexes(world, num_entities, encoded_table)
+        # (batch_size, num_entities, num_entities, embedding_dim)
+        embedded_neighbors = util.batched_index_select(encoded_table, neighbor_indices)
+        neighbor_mask = util.get_text_field_mask({'ignored': neighbor_indices},
```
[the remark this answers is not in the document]
rajasagashe: Sounds good, I also passed the absolute value of neighbor_indexes to batched_index_select since -1 causes an invalid index error.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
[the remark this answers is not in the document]
matt-gardner: Oh, right, sorry, you're using the `Linear` module, not just instantiating a parameter vector.  You're right, this is fine.  This will add a bias on top of just instantiating a parameter vector, which is actually probably what we want.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
             # TODO(matt): compute accuracy here.
             return outputs
 
+    @staticmethod
+    def _get_neighbor_indexes(worlds: List[WikiTablesWorld],
```
matt-gardner: `s/indexes/indices/`, which is what we use everywhere else in the code.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
...
+                              num_entities: int,
+                              tensor: torch.Tensor) -> Variable:
+        """
+        This method returns the indices of each entity's neighbors. A tensor
+        is accepted as a parameter for copying purposes
```
matt-gardner: Missing period.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
             # TODO(matt): compute accuracy here.
             return outputs
 
+    @staticmethod
+    def _get_neighbor_indexes(worlds: List[WikiTablesWorld],
+                              num_entities: int,
+                              tensor: torch.Tensor) -> Variable:
```
matt-gardner: The distinction between `Variable` and `Tensor` in pytorch is a bit awkward, and is supposed to be going away in pytorch 0.4.  We've been annotating types with `torch.Tensor` instead of `Variable`, partially looking ahead to the type change that's coming.  So, unless mypy complains, I'd switch everywhere that you have `Variable` annotations to `torch.Tensor` annotations (and if you can be more specific, like `torch.LongTensor` or `torch.FloatTensor`, even better).

matt-gardner: The difference here is that the _input_ tensor probably needs to be annotated as a `Variable`, because you're actually passing it a variable.  I'm actually surprised that mypy isn't complaining here, because `torch.Tensor` doesn't have a `data` member...  It'll be really nice when pytorch unifies these two types.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
...
+        Parameters
+        ----------
+        worlds : ``List[WikiTablesWorld]``
+        num_entities : ``int``
+        tensor : ``torch.Tensor``
+            Used for copying the constructed list onto the right device.
+
+        Returns
+        -------
+        A ``torch.autograd.Variable`` with shape ``(batch_size, num_entities, num_entities)``.
```
matt-gardner: Does this need to be `num_entities`?  Could it just be `num_neighbors`, where you find the max num neighbors in this method, and only pad to that amount?  You'd save quite a bit of memory, and computation, I think, because the amount of padding would be a whole lot less.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
...
+        Dimension 2 contains the indices for the neighbors of the entity represented by the
+        index of dimension 1. Dimension 2 is limited to num_entities as it is assumed that
+        each entity can have a number of neighbors at most equal to the total number of entities.
+        """
+        batches_neighbors = []
```
matt-gardner: You only have one batch here, so this should be `batch_neighbors`.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
...
+        for world in worlds:
+            # Each batch instance has its own world, which has a corresponding table.
+            entities = world.table_graph.entities
+            entity2indices = {entity: i for i, entity in enumerate(entities)}
```
matt-gardner: I'm not a huge fan of using `2` like this, but if you like it, that's fine.  I think `entity_indices` and `entity_neighbors` are a little easier to read, though I can see an argument that if you're consistent, maybe the `2` is a strong signal that this is a dict.  If you want to keep it with the `2` though, this should be `entity2index`, not `entity2indices`, because you're mapping to an `int`, not a `List[int]`.

matt-gardner: And I guess your use of `entity_neighbors` below is another decent argument for giving this a separate name that makes it obvious it's a dict...

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
...
+            entity2neighbors = world.table_graph.neighbors
+            neighbor_indexes = []
+            for entity in entities:
+                entity_neighbors = [entity2indices[cn] for cn in entity2neighbors[entity]]
```
matt-gardner: What is `cn`?  `current_neighbor`?  A simple `n` would probably be better here, because it's more obviously just a temporary variable in a list comprehension, and it makes me less likely to try to think about what it's supposed to mean.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
...
+                padded = pad_sequence_to_length(entity_neighbors, num_entities)
+                neighbor_indexes.append(padded)
```
matt-gardner: `indices`, not `indexes` (sorry, you're probably pretty used to `indexes`, but consistency with everything else in the code is more important).

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
...
+            neighbor_indexes = pad_sequence_to_length(neighbor_indexes,
+                                                      num_entities,
+                                                      lambda: [0] * num_entities)
```
matt-gardner: You've probably already fixed this, but this should be `-1`, not `0`.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
...
+            batches_neighbors.append(neighbor_indexes)
+        return Variable(tensor.data.new(batches_neighbors)).long()
+
+    @staticmethod
+    def _get_type_vector(worlds: List[WikiTablesWorld],
+                         num_entities: int,
+                         tensor: torch.Tensor) -> Tuple[Variable, Dict[int, int]]:
+        """
+        Produces the one hot encoding for each entity's type. In addition,
+        a map from index to type is returned to combine entity type operations
```
matt-gardner: "a map from a flattened entity index to type".

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
...
+        ----------
+        worlds : ``List[WikiTablesWorld]``
+        num_entities : ``int``
+        tensor : ``torch.Tensor``
+            Used for copying the constructed list onto the right device.
+
+        Returns
+        -------
+        A ``torch.autograd.Variable`` with shape ``(batch_size, num_entities, num_types)``.
+        entity_types : ``Dict[int,str]``
```
matt-gardner: Space after the comma.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
...
+            This is a mapping from (batch_index, entity_index) to entity type id.
```
matt-gardner: It's `batch_index * entity_index`, not `(batch_index, entity_index)`.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
...
+        """
+        entity_types = {}
+        batch_types = []
+        for batch_index, world in enumerate(worlds):
+            types = []
+            for entity_index, entity in enumerate(world.table_graph.entities):
+                cell_type_one_hot = [1, 0]
+                row_type_one_hot = [0, 1]
+                types.append((cell_type_one_hot if entity.startswith('fb:cell') else row_type_one_hot))
```
matt-gardner: You should only have the type check once.  If you move the `entity_type = 0 ...` line up here, you can just do `types.append([1, 0] if entity_type == 0 else [0, 1])`.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
...
+            Contains all the probabilities for an entity given a question word.
+        """
+        batch_probabilities = Variable(tensor.data.new(torch.zeros(batch_size,
+                                                                   num_entities+1,
+                                                                   num_question_tokens)))
+        for batch_index, world in enumerate(worlds):
+            cell_type_index, row_type_index = self._get_entity_index_by_type(world)
+
+            # Separate the scores by type, since normalization summed over types.
+            cell_type_scores = torch.index_select(scores[batch_index],
```
matt-gardner: With all of these slices, `index_selects`, and element assignments, this is going to be _really_ slow.

...

Well, I was thinking about ways to fix this to make it faster, but they are pretty complicated.  How about we profile this to make sure it's actually slow before making those complicated changes?  Have you tried training this?

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +310,166 @@ def forward(self,  # type: ignore
...
+        cell_type_index: ``List[int]``
+            Indexes for entities starting with fb:cell.
+        row_type_index: ``List[int]``
+            Indexes for entities not starting with fb:cell.
+        """
+        cell_type_index, row_type_index = [], []
+        entities = world.table_graph.entities
+        entity2index = {entity: entity_index for entity_index, entity in enumerate(entities)}
+        for entity in entities:
+            if entity.startswith('fb:cell'):
```
matt-gardner: This is another place that you have a check for entity type - this means that if (when) we change how we're using entity types, we have to find it in another place to make sure it's consistent.  It'd be better if we can find a way to use the entity type dictionary that we already computed.

tests/models/encoder_decoders/wikitables_semantic_parser_test.py
```diff
@@ -25,6 +24,82 @@ def setUp(self):
     def test_model_can_train_save_and_load(self):
         self.ensure_model_can_train_save_and_load(self.param_file)
 
+    def test_correct_neighbor_indexes(self):
```
matt-gardner: I think this test would be more informative if you constructed a small world with just a few entities, and actually checked the whole returned tensor, instead of just a couple of entries in the tensor.  This would make sure that padding is correct (especially if we're padding to `num_neighbors` instead of `num_entities`), and that everything is as we expect.

tests/models/encoder_decoders/wikitables_semantic_parser_test.py
```diff
@@ -25,6 +24,82 @@ def setUp(self):
...
+        neighbor_indexes = self.model._get_neighbor_indexes(worlds, num_entities, tensor)
+        assert neighbor_indexes.size(1) == num_entities
+        assert neighbor_indexes.size(2) == num_entities
+        lst = [0]*(num_entities)
+        lst[0], lst[1] = 45, 45
+        second_entity_neighbors = (torch.LongTensor(lst))
+        assert torch.equal(neighbor_indexes[0][1].data, second_entity_neighbors)
+
+    def test_all_types_present_in_type_vector(self):
+        # Verifies that at least 1 instance of every one hot vector
```
matt-gardner: Again here I think it'd be more informative to just construct a small world, and make sure that you get the results you expect for both return values from `_get_type_vector`.  If the world has both entity types, checking for the explicit values will also implicitly check that all types are present in the returned vector.

tests/models/encoder_decoders/wikitables_semantic_parser_test.py
matt-gardner: Also, there isn't a single "right" way to name a test, but what we've generally done is this: `test_[method_name]_[aspect_under_test]`.  See, for example, the tests here: https://github.com/allenai/allennlp/blob/wikitables/tests/data/semparse/knowledge_graphs/table_knowledge_graph_test.py.  This makes it a lot easier to know what the point of the test is, and if you end up breaking the test later, you can see in the name what functionality you've broken.

There are also times when there isn't a clear "aspect" that's under test, other than the standard flow through the method with a typical expected input.  In those cases, I typically just name it `test_[method_name]`.  I'd say this test falls under that category, so I'd name this `test_get_neighbor_indices`.

The same goes for the other test names here.

tests/models/encoder_decoders/wikitables_semantic_parser_test.py
```diff
@@ -25,6 +24,82 @@ def setUp(self):
...
+        tensor = Variable(torch.LongTensor([]))
+        type_vector, _ = self.model._get_type_vector(worlds, num_entities, tensor)
+        # (batch_size, num_types)
+        sums = torch.sum(type_vector, dim=1)
+        x = torch.nonzero(sums.data)
+        # this means every element in sums is nonzero hence both 2 numbers
+        # representing its index will be in x
+        assert x.size(0) * x.size(1) == 2 * sums.size(0) * sums.size(1)
+
+    def test_get_prob_entity_valid_probability_distribution(self):
```
matt-gardner: Also here, this would be better to just pick some numbers in a small `linking_scores` tensor, make the necessary masks by hand, and make sure that the values come out normalized the way you expect.  I'd be a lot more confident that the method is doing the right thing if it correctly computed the values in a known, hand-worked example.  It should also end up a whole lot simpler than you have this test, too.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
rajasagashe: Good point, I added the question mask through a masked_softmax. How do I incorporate the entity mask? Isn't it already accounted for when the index select picks valid entities of each type?

tests/models/encoder_decoders/wikitables_semantic_parser_test.py
```diff
@@ -58,47 +59,60 @@ def test_all_types_present_in_type_vector(self):
         # representing its index will be in x
         assert x.size(0) * x.size(1) == 2 * sums.size(0) * sums.size(1)
 
-    def test_get_prob_entity_valid_probability_distribution(self):
+    def test_get_linking_probabilities_valid_probability_distribution(self):
```
matt-gardner: This test is still really long and complicated.  You want to have as little logic in a test case as possible.  For example, see some of these tests:
https://github.com/allenai/allennlp/blob/4f8e04e6cdfbc8707707ac978a964777459c161e/tests/models/encoder_decoders/wikitables_semantic_parser_test.py#L316-L337
This test has about 20 lines, but a lot of those are comments, and there is no complex logic.  We're just instantiating some objects, calling a method, and asserting that the return value is what we expect.  The more logic you have in your test, the harder it is to understand, and the easier it is to not actually be testing what you think you're testing.

An example of what I meant by "construct a small world" is here:
https://github.com/allenai/allennlp/blob/4f8e04e6cdfbc8707707ac978a964777459c161e/tests/models/encoder_decoders/wikitables_semantic_parser_test.py#L181-L224

Again this test looks long in number of lines, but there are no loops, and there's only one method that's called - the rest is just instantiating some objects that we use for the test, and those objects happen to be longish lists, so the test takes a bunch of lines.

A final example: https://github.com/allenai/allennlp/blob/4f8e04e6cdfbc8707707ac978a964777459c161e/tests/nn/util_test.py#L71-L181

This test is checking a softmax function, which does something very similar to what you're doing.  The test just instantiates a tensor with made up numbers, calculates the expected values by hand, and assert that they end up correct.  No loops, no complex logic, just some simple examples that you can look at and know that they look right (there are also probably way more checks there than are necessary, but that's for historical reasons).

So, what I would recommend for this test is to construct a world with just a few entities, make up a `linking_scores` tensor with random numbers that are just given explicitly in a matrix (but _not_ with `torch.rand()`), and call `._get_linking_probabilities`, and have a check on the exact tensor that's returned.

matt-gardner:
Hmm, sorry, I might have messed this up.  I needed to rebase the wikitables branch, so I figured I'd just rebase this for you while I was at it.  I forgot that github's comment tracking doesn't work particularly well when you pull the commits out from under it like that...  I'll go through this again and make sure that all of the smaller style issues were addressed (maybe just fix them myself as I'm doing this, to spare you the trouble), and make a list of the larger issues that are left.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
matt-gardner: `index_select` gets passed a padded tensor, I believe.  It'd be a lot easier to answer this question after seeing a concrete test case for the method.

rajasagashe:
Ok, issues 1,2, and 4 should be completed in the latest commit.

tests/models/encoder_decoders/wikitables_semantic_parser_test.py
```diff
@@ -25,6 +25,105 @@ def setUp(self):
...
 
+    def test_get_neighbor_indices(self):
+        worlds, num_entities = self.get_fake_worlds()
+        tensor = Variable(torch.LongTensor([]))
+
+        neighbor_indexes = self.model._get_neighbor_indices(worlds, num_entities, tensor)
+
+        # Checks for the correct shape meaning dimension 2 has size num_neighbors,
+        # padding of -1 is used, and correct neighbor indices.
+        assert torch.equal(neighbor_indexes.data, torch.LongTensor([[[2, 3],
```
matt-gardner: Why does this have shape (1, 6, 2)?  I would have thought it would be (1, 4, 2).  Ah, it's because you have a `sum()` down below in `get_fake_worlds`, where it should be a `max()`.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -271,6 +318,165 @@ def forward(self,  # type: ignore
...
+            padded = pad_sequence_to_length(types, num_entities, lambda: [0, 0])
+            batch_types.append(padded)
+        return Variable(tensor.data.new(batch_types)), entity_types
+
+    @staticmethod
+    def _get_linking_probabilities(worlds: List[WikiTablesWorld],
+                                   linking_scores: Variable,
+                                   question_mask: Variable,
+                                   entity_type_dict: Dict[int, int],
+                                   tensor: Variable) -> torch.FloatTensor:
```
matt-gardner: You don't actually need to take this tensor, because you've already got `linking_scores` that you can copy from instead.  We only needed it in `_get_neighbor_indices` because there wasn't any other tensor to copy from.

tests/models/encoder_decoders/wikitables_semantic_parser_test.py
```diff
@@ -25,6 +25,105 @@ def setUp(self):
...
+                           [0.8245101, 0.20058638]]]
+        linking_scores = Variable(torch.FloatTensor(linking_scores))
+        question_mask = get_text_field_mask({'questions': questions}).float()
+
+        # (batch_size, num_question_tokens, num_entities + 1)
+        entity_probability = self.model._get_linking_probabilities(worlds, linking_scores, question_mask,
+                                                                   entity_type_dict, tensor)
+
+        # The following properties in entity_probability are tested for by true_probability:
+        # It has probability 0.0 at the 0th entity index for the null entity.
```
matt-gardner: This isn't actually correct - we don't want a 0 _probability_ for the null entity, we want a zero _logit_ for the null entity.  If you look at the right column on page four of the paper, you can see that `s(null, i) = 0`, but `exp(0) = 1` - this allows the model to select the null entity if it doesn't think there's a good match.  Having negative values for some of the scores above would help show that this is working.

Also, I think the value for the null entity here doesn't matter, right?  Because that entity shows up in several normalizations, it's not clear what you should put in the table.  But the value in the table doesn't ever get used, so it doesn't really matter.  You could even just remove it, and return a tensor of shape `(batch_size, num_question_tokens, num_entities)`.

rajasagashe: So should I just append a 0 to cell_type_scores and row_type_scores so it will be used in the softmax? Then return a tensor of shape (batch_size, num_question_tokens, num_entities)?

In the weighted sum equation the null entity embedding doesn't seem to be included, does that mean I should remove it from forward() ?

matt-gardner: Yeah, we want a non-masked 0 added to each of the softmaxes for the null entity.  And then you can either return a tensor of shape `(batch_size, num_question_tokens, num_entities + 1)`, and include a row for the null entity embedding in your weighted sum, or return a tensor of shape `(batch_size, num_question_tokens, num_entities)`, and don't have a row for the null entity.  I don't think it makes a big difference, though the first option is probably simpler.

matt-gardner:
Thanks @rajasagashe for all of your work on this!  I went ahead and fixed the `get_linking_probabilities` test and implementation, and I'm going to merge this now.

I also profiled the code a bunch yesterday and today, and it is indeed the case that `_get_linking_probabilities` is quite slow.  I'm going to try to speed it up now, and open a new PR for that.

rajasagashe:
Thanks! I also implemented the fixes today and was about to push, but just saw the comment.

PR #687, opened 2018-01-16
Disable running coverage in Travis.

schmmd:
This is configured in TeamCity now and we don't want it running in both places.

On Thursday I'll make a PR that removes Travis entirely from our repo.


matt-gardner:
Great!

PR #688, opened 2018-01-16
Speed up WikiTablesDatasetReader

matt-gardner:
I did some profiling to figure out why our data loading code took so long, and fixed some of the worst problems.  These were (1) using a lisp parser to read the example file, which can just as easily be read with some simple string splits, and (2) running spacy's tagger to get lemmas, but not using spacy's batching.  I had to implement batched tokenization for this, and merging this should probably wait until #680 is merged into master.

### DIFF
--- a/allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
+++ b/allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
@@ -25,7 +25,7 @@ from allennlp.data.semparse.knowledge_graphs import TableKnowledgeGraph
 from allennlp.data.semparse.type_declarations import wikitables_type_declaration as wt_types
 from allennlp.data.semparse.worlds import WikiTablesWorld
 from allennlp.data.token_indexers import TokenIndexer, SingleIdTokenIndexer, TokenCharactersIndexer
-from allennlp.data.tokenizers import Tokenizer, WordTokenizer
+from allennlp.data.tokenizers import Token, Tokenizer, WordTokenizer
 from allennlp.data.tokenizers.word_splitter import SpacyWordSplitter
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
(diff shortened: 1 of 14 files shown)


schmmd:
@matt-gardner just curious--did you manually profile the code or did you use a third-party tool?  I'm interested in doing this for the AllenNLP demo load times, which is quite slow when deployed to Kubernetes (I think it's due to downloading the models).

matt-gardner:
I used [pyflame](https://github.com/uber/pyflame), which worked well enough for what I needed.  It's not the best for profiling GPU code, and pytorch has some tooling now for doing this, but I haven't figured out how to use it yet.

allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
```diff
@@ -258,33 +283,23 @@ def text_to_instance(self,  # type: ignore
...
-            elif key == "context":
-                # Skipping "graph" and "tables.TableKnowledgeGraph"
-                parsed_info["context"] = value[-1]
-            elif key == "targetValue":
-                # Skipping "list", and "description" within each nested list.
-                parsed_info["targetValue"] = [x[1] for x in value[1:]]
-        # targetValue may not be present if the answer is not provided.
-        assert all([x in parsed_info for x in ["id", "question", "context"]]), "Invalid format"
-        return parsed_info
+        id_piece, rest = lisp_string.split(') (utterance "')
```
pdasigi: You're assuming the fields in the string come in a specific order. It may be a good idea to make sure the order of these fields is what you expect it to be (i.e. check the indices of the field names).
Or you could use regex. Is that significantly slower?
Not really a serious issue though. The only situation this would be a problem is if at some point of time we switch to using a different version of the dataset that is formatted differently.

matt-gardner: The code will crash if these fields come in a different order, because of the way I'm assigning to a tuple after the calls to `.split()`.  I think that's good enough for now, and if there ever comes a time when we switch to some different version of the dataset, we can worry about updating this then.

PR #692, opened 2018-01-17
Add an attention visualization to the BiDAF demo and a rough how-to

matt-gardner:
This is mostly usable in its current form, but it's quite ugly.  My main point was to show how to put in the plumbing on the model side and plug it into a component in the js.  The heatmap component still needs a lot of work, though.

To move forward on this, we need to do one of two things:

1. Get a designer (or just someone who knows CSS better than I do) to look at this and figure out a better way to show the heat maps.
2. Find a place to put this that isn't in a public-facing demo, but can be used by researchers who want to use it.  Maybe that's in a separate repository, with a tutorial on how to customize your demo?

Also, I fixed an issue where duplicate words in the passage were getting squashed together in the table; it looks like the same thing is happening in the question, and the fix I did for the passage didn't look like it was working.  Not totally sure what's going on there...


schmmd:
When I run `npm start` on this branch I get several warnings and errors on the Javascript console.

matt-gardner:
Can you run the demo?

schmmd:
I'm not sure what you're asking exactly @matt-gardner.  I cannot make requests or see the attention visualization.  I don't need this urgently, however, and I don't have time to troubleshoot further today.

matt-gardner:
Ok, this is now ready for review, though I'm not sure who to assign it to.  It's a mix of react code, minor python changes, and a how-to.  @schmmd, want to take a look, or suggest someone else to review?

schmmd:
Happy to look now, although nobody on the team gives particularly critical feedback on ReactJS.

matt-gardner:
Quick link for seeing the how-to rendered nicely: https://github.com/matt-gardner/allennlp/blob/visualize_bidaf_attention/tutorials/how_to/visualizing_model_internals.md

demo/src/api-config.js
```diff
@@ -9,8 +9,8 @@ let apiRoot;
 
 const origin = window && window.location && window.location.origin;
 
-if (origin === 'http://localhost:3000') {
-    apiRoot = 'http://localhost:8000';
+if (origin.includes(':3000')) {
```
schmmd: This seems like a good change--but just curious: why was it necessary?

demo/src/components/McComponent.js
```diff
@@ -134,7 +136,7 @@ render() {
 
 class McOutput extends React.Component {
     render() {
-      const { passage, answer } = this.props;
+      const { passage, answer, attention, question_tokens, passage_tokens } = this.props;
```
schmmd: This is fine, but if I had wrote this from scratch I would have added more hierarchy to `passage` rather than having a separate `passage_tokens`.

I.e.,

```
passage: {
  text: "...",
  tokens: [...]
}

File: demo/src/components/heatmap/DataGrid.jsx [hunk 3]
```diff
@@ -0,0 +1,56 @@
...
+  return (
+    <div style={{"white-space": "nowrap"}}>
+      {yLabels.map((y, yi) => (
+        <div key={`${y}_${yi}`} style={{clear: "both"}}>
+          <div style={{display: "inline-block", width: xLabelWidth, textAlign: 'right', paddingRight: '5px', paddingTop:`${boxSize/3.7}px`}}>{y}</div>
+          {xLabels.map((x, xi) => (
+            <div
+              title={`${data[yi][xi]}`}
+              key={`${x}_${xi}_${y}_${yi}`}
+              style={{
```
schmmd: Is it possible to use a class rather than inline style?  This applies a few lines above as well.

demo/src/components/heatmap/DataGrid.jsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
```
schmmd: A top-level comment overviewing this component would help me.

tutorials/how_to/visualizing_model_internals.md
```diff
@@ -0,0 +1,108 @@
...
+here's a guide to the changes you need to make to get a live demo that shows model internals, which
+we've found to be very helpful for understanding what parts of our models need fixing.
+
+## Pre-requisites
+
+You need to have a demo of your model already running.  See the [tutorial on creating a
+demo](../getting_started/making_predictions_and_creating_a_demo.md) for how to do this.  You'll
+also have to be using react to serve the demo, which means the simple server is out.  You need to
+be using `python -m allennlp.run serve` instead of `python -m allennlp.service.server_simple` to
+run the server.  Sorry, but we're not including instructions for that here.  You can see
```
schmmd: Optional: I would remove your apology.

joelgrus:
my guess is that most people wanting to visualize their models would use `server_simple` rather than trying to shoehorn their model into the demo. 

would it be worth tweaking this to show how it could work with that? (then we wouldn't have to punt on the "we're not going to tell you how to get the demo working" piece)

off the top of my head it could be as simple as 

* add a script tag to load babel
* add a script tag to load react and react-dom
* copy those jsx files to your `static_html` directory and add a script tag for them
* add that component to the output

?

matt-gardner:
@joelgrus, yes, I'm definitely in favor of making this easier to use by doing something like that.  I'm just not sure how, and don't have the bandwidth to figure it out right now, as it would be pretty slow for me and I need to work on other stuff.

demo/src/api-config.js
matt-gardner: Because I was running the demo on an EC2 instance and hitting it from my mac.  I wasn't going to `localhost`,  I was going to `10.12.whatever`.

demo/src/components/McComponent.js
matt-gardner: Yes, I agree that would be better from the demo's perspective.  I'll have to look at the python code and see if it makes things more complicated there.  I think it does, because of how batching interacts with the demo.  Probably still possible, though.

demo/src/components/heatmap/DataGrid.jsx
matt-gardner: Me too =).  I just copied this from somewhere else.  I'll try to give some info from what I could figure out.

demo/src/components/heatmap/DataGrid.jsx
matt-gardner: I have no idea.  I don't know enough about how react actually works.  It'd be nice to have a front-end developer...

demo/src/components/McComponent.js
matt-gardner: Hmm...  Actually, the `passage` comes from the `requestData`, and the `passage_tokens` comes from the `responseData`.  So I _can't_ actually change this in the python code.  Changing it just in the javascript means packing and then immediately unpacking a javascript object, which seems a little redundant.

matt-gardner:
Ok, I think I'm going to merge this as it is, and open an issue to make this work with the simple server somehow.  It might even make sense at that point to just add it to the end of the demo tutorial @joelgrus already wrote, or at least put a pointer there.

demo/src/components/heatmap/DataGrid.jsx
schmmd: Ah, please link to wherever you copied it from at the least.

matt-gardner: I added some comments, and there's a link to where I copied it from in `ORIGINAL_LICENSE`.

demo/src/components/heatmap/DataGrid.jsx
schmmd: You could add a class="foo" tag and instead make this a css entry foo in main.css or another included css file. This makes the style easier to understand and configure. A frontend dev would cringe at these inline style tags, but what you have is fine for now.

matt-gardner: For some of these, that would not work, because we want to be able to configure sizes programmatically.  I have different heatmaps in the semantic parser demo with different x label widths, for instance.

I also was a little annoyed that we're just using plain CSS instead of something like less that allows variables, so I didn't have much motivation to put this into a flat CSS file.  My vote is to get a front-end developer who can clean all of this up.

schmmd:
👍 👍

Issue #694, opened 2018-01-17
investigate whether `allennlp.run` could take auxiliary modules as command line args and use importlib


joelgrus:
i.e. it's possible that if I make my own models etc in `allennlp_joel.models`, `allennlp_joel.dataset_readers`, and so on, that I could do something like

```
python -m allennlp.run --extra-modules allennlp_joel train ...
```

and use `importlib` to programmatically import everything and not have to write my own executable

matt-gardner:
Yeah, this sound like it could work, and would be a great idea.

PR #695, opened 2018-01-17
programmatic imports of third-party models

joelgrus:
with this change you no longer have to create your own `run.py` script just to use your own models, instead you should be able to do things like

```bash
python -m allennlp.run train \
    /path/to/config.json \
    -s serialization_dir \
    --include-package my_model \
    --include-package my_other_model
```

and have the modules in those packages imported automatically before the `from_params` are ever called.

I made similar changes in `evaluate` and `predict` (and I correspondingly allowed you to explicitly specify the predictor name), and also in `server_simple`.

I did not change the `serve` command, as right now it's too intertwined with the HTML/JS of the demo to make that functionality useful.

As you can see, the tests for this behavior are kind of a pain, as I need to "create" modules that aren't part of AllenNLP and modify test fixtures to use them. 

I also modified tutorials to account for this behavior. I think that the two I changed were the only ones that need changes, but it's possible I missed somewhere.

Fixes https://github.com/allenai/allennlp/issues/694


schmmd:
Love the concept!  Seems like a great addition!

allennlp/commands/predict.py
```diff
@@ -73,17 +74,32 @@ def add_subparser(self, name: str, parser: argparse._SubParsersAction) -> argpar
...
 
+        subparser.add_argument('--include-package',
+                               type=str,
+                               action='append',
+                               default=[],
+                               help='additional packages to include')
+
+        subparser.add_argument('--predictor',
+                               type=str,
+                               help='optionally specify a specific predictor to use')
```
schmmd: I'd amend the help string so it's clear this is an archive file or url

schmmd: clearly I misunderstood at first glance

schmmd: "optionally specify the predictor to use inside the archive file"?

PR #696, opened 2018-01-17
quick and dirty logging of memory usage (sort of)

joelgrus:
uses `resource.getrusage` to get peak memory usage for the python process, and logs it once per training epoch. also adds it to the `/info` route of the flask server

cc: @splittingfield


OyvindTafjord:
Is there a way to log GPU memory usage too? That could be useful for seeing, e.g., if your batch size is appropriate.

PR #705, opened 2018-01-22
let there be validation metrics which aren't computed during training

DeNeutoy:


allennlp/training/trainer.py
```diff
@@ -284,10 +284,18 @@ def _metrics_to_tensorboard(self,
...
-            self._tensorboard.add_train_scalar(name, value, epoch)
-            if val_metrics:
-                self._tensorboard.add_validation_scalar(name, val_metrics[name], epoch)
+
+        metric_names = set(train_metrics.keys())
+        if val_metrics:
+            metric_names.update(val_metrics.keys())
+
+        for name in metric_names:
+            train_metric = train_metrics.get(name, None)
```
joelgrus: the second argument to `get` is `None` by default, so you don't need to specify it

allennlp/training/trainer.py
```diff
@@ -298,13 +306,21 @@ def _metrics_to_console(self,  # pylint: disable=no-self-use
...
+            val_metric = val_metrics.get(name, None)
+            if val_metric is not None and train_metric is not None:
+                logger.info(message_template, name, train_metric, name, val_metric)
+            elif val_metric is not None:
+                logger.info(message_template, "Validation", name, val_metric)
+            elif train_metric is not None:
+                logger.info(message_template, "Training", name, train_metric)
```
joelgrus: I'm not convinced this does the right thing

the `message_template` is chosen based on whether there are any validation metrics or not, but aren't we trying to address the case where there is a validation metric that isn't a train metric? in that case we'd have the first message template, and then this logger statement isn't right?

allennlp/training/trainer.py
joelgrus: this one too, and also in line 296

PR #707, opened 2018-01-23
Span field

DeNeutoy:
Whilst doing this I came across [this](https://github.com/allenai/allennlp/blob/master/allennlp/data/fields/index_field.py#L41), which I think is redundant. I can add a commit to remove it and it's associated test, but I wasn't sure if there was something non-obvious it was used for, as you're more familiar with the Bidaf code than I am.

### DIFF
--- a/allennlp/data/fields/span_field.py
+++ b/allennlp/data/fields/span_field.py
@@ -31,17 +31,17 @@ class SpanField(Field[torch.Tensor]):
         self.span_end = span_end
         self.sequence_field = sequence_field
 
-        if not isinstance(span_start, int) or not isinstance(span_start, int):
-            raise ConfigurationError(f"SpanFields must be passed integer indices. Found span indices: "
-                                     f"({span_start}, {span_end}) with types "
-                                     f"({type(span_start)} {type(span_end)})")
+        if not isinstance(span_start, int) or not isinstance(span_end, int):
+            raise TypeError(f"SpanFields must be passed integer indices. Found span indices: "
+                            f"({span_start}, {span_end}) with types "
+                            f"({type(span_start)} {type(span_end)})")
         if span_start > span_end:
-            raise ConfigurationError(f"span_start must be greater than span_end, "
-                                     f"but found ({span_start}, {span_end}).")
+            raise ValueError(f"span_start must be less than span_end, "
+                             f"but found ({span_start}, {span_end}).")
 
         if span_end > self.sequence_field.sequence_length() - 1:
-            raise ConfigurationError(f"span_end must be < len(sequence_length) - 1, but found "
-                                     f"{span_end} and {self.sequence_field.sequence_length() - 1} respectively.")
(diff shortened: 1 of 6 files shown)


allennlp/data/fields/span_field.py
```diff
@@ -0,0 +1,61 @@
...
+            raise ConfigurationError(f"span_start must be greater than span_end, "
+                                     f"but found ({span_start}, {span_end}).")
+
+        if span_end > self.sequence_field.sequence_length() - 1:
+            raise ConfigurationError(f"span_end must be < len(sequence_length) - 1, but found "
+                                     f"{span_end} and {self.sequence_field.sequence_length() - 1} respectively.")
+
+    @overrides
+    def get_padding_lengths(self) -> Dict[str, int]:
+        return {}
```
matt-gardner: You can remove the corresponding bit of this from the `IndexField`.  It's a holdover from when we used one-hot vectors instead of indices.

allennlp/data/fields/span_field.py
```diff
@@ -0,0 +1,61 @@
...
+        The inclusive index of the end of the span in the :class:`SequenceField`.
+    sequence_field : ``SequenceField``, required.
+        A field containing the sequence that this ``SpanField`` is a span inside.
+    """
+    def __init__(self, span_start: int, span_end: int, sequence_field: SequenceField) -> None:
+        self.span_start = span_start
+        self.span_end = span_end
+        self.sequence_field = sequence_field
+
+        if not isinstance(span_start, int) or not isinstance(span_start, int):
```
matt-gardner: I think one of these should be `span_end`.

allennlp/data/fields/span_field.py
```diff
@@ -0,0 +1,61 @@
...
+            raise ConfigurationError(f"SpanFields must be passed integer indices. Found span indices: "
+                                     f"({span_start}, {span_end}) with types "
+                                     f"({type(span_start)} {type(span_end)})")
+        if span_start > span_end:
+            raise ConfigurationError(f"span_start must be greater than span_end, "
```
matt-gardner: This is backwards.

allennlp/data/fields/span_field.py
```diff
@@ -0,0 +1,61 @@
...
+    sequence_field : ``SequenceField``, required.
+        A field containing the sequence that this ``SpanField`` is a span inside.
+    """
+    def __init__(self, span_start: int, span_end: int, sequence_field: SequenceField) -> None:
+        self.span_start = span_start
+        self.span_end = span_end
+        self.sequence_field = sequence_field
+
+        if not isinstance(span_start, int) or not isinstance(span_start, int):
+            raise ConfigurationError(f"SpanFields must be passed integer indices. Found span indices: "
```
matt-gardner: I'd vote for a `ValueError` or `RuntimeError` (or even `TypeError`) here - `ConfigurationError` is intended for things that are wrong with a configuration file, and these errors don't fit that.

PR #708, opened 2018-01-23
Span utils

DeNeutoy:
- Moves `bio_tags_to_spans` into `dataset_utils.span_utils`
- Adds a `enumerate_spans` function
- Replaces the span enumeration in the coref reader with `enumerate_spans`


allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -0,0 +1,125 @@
...
+        The minimum length of spans which should be included.
+    """
+    max_span_width = max_span_width or len(sentence)
+    min_span_width = min_span_width or 0
+    filter_function = filter_function or bool
+    spans: List[Tuple[int, int]] = []
+
+    for start_index in range(len(sentence)):
+        last_end_index = min(start_index + max_span_width, len(sentence))
+        first_end_index = min(max(start_index, start_index + min_span_width - 1), len(sentence))
```
matt-gardner: I think `min_span_width` should default to 1, not 0.  I don't think it makes much sense to have a bunch of empty spans at various places.  And if the default is 1, you can remove this inner max.

allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -0,0 +1,125 @@
...
+    Parameters
+    ----------
+    sentence : ``List[Token]``, required.
+        The sentence to generate spans for.
+    offset : ``int``, optional (default = 0)
+        A numeric offset to add to all span start and end indices. This is helpful
+        if the sentence is part of a larger structure, such as a document, which
+        the indices need to respect.
+    max_span_width : ``int``, optional (default = None)
+        The maximum length of spans which should be included.
```
matt-gardner: It'd be nice to say that this defaults to `len(sentence)`.

allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -0,0 +1,125 @@
...
+    min_span_width : ``int``, optional (default = None)
+        The minimum length of spans which should be included.
```
matt-gardner: This actually defaults to 0, currently (though it should be 1; see my other comments).  You should just make that the default argument instead of `None`.

allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    max_span_width = max_span_width or len(sentence)
+    min_span_width = min_span_width or 0
+    filter_function = filter_function or bool
```
matt-gardner: `bool` here seems a bit confusing.  This actually interacts in funny ways with `min_span_width = 0` - `bool` gives `False` for that (at least, I thought it should, be your test seems to say otherwise...  not sure what's going on there).  `lambda x: True` would be better.

allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -0,0 +1,125 @@
...
+        for end_index in range(first_end_index, last_end_index):
+            start = offset + start_index
+            end = offset + end_index
+            # add 1 to end index because span indices are inclusive.
+            if filter_function(sentence[slice(start_index, end_index + 1)]):
+                spans.append((start, end))
+    return spans
+
+
+def bio_tags_to_spans(tag_sequence: List[str],
```
matt-gardner: It looks like you forgot to actually remove this from where it used to be?

allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -0,0 +1,125 @@
+from typing import List, Tuple, Callable
+
+from allennlp.data.dataset_readers.dataset_utils.ontonotes import TypedStringSpan
+from allennlp.data.tokenizers.token import Token
+
+def enumerate_spans(sentence: List[Token],
+                    offset: int = 0,
+                    max_span_width: int = None,
+                    min_span_width: int = None,
+                    filter_function: Callable[[List[Token]], bool] = None) -> List[Tuple[int, int]]:
```
matt-gardner: This parameter isn't documented.  It'd be nice to explicitly state somewhere that returning `True` from this function means "keep the span" - that's not always obvious from the name "filter".

allennlp/data/dataset_readers/dataset_utils/span_utils.py
matt-gardner: Ah, ok, yeah, I was confused by a few pieces of this (like why the `bool` filter wasn't doing what I thought).  This max effectively changes the `min_span_width` to 1 if it was set at 0.  So just set it to 1 by default.

tests/data/dataset_readers/dataset_utils/span_utils_test.py
```diff
@@ -0,0 +1,59 @@
...
+    def test_enumerate_spans_enumerates_all_spans(self):
+        tokenizer = SpacyWordSplitter(pos_tags=True)
+        sentence = tokenizer.split_words("This is a sentence.")
+
+        spans = span_utils.enumerate_spans(sentence)
+        assert spans == [(0, 0), (0, 1), (0, 2), (0, 3), (0, 4), (1, 1), (1, 2),
+                         (1, 3), (1, 4), (2, 2), (2, 3), (2, 4), (3, 3), (3, 4), (4, 4)]
+
+        spans = span_utils.enumerate_spans(sentence, max_span_width=3, min_span_width=2)
+        print(spans)
```
matt-gardner: Remove print.

allennlp/data/dataset_readers/dataset_utils/span_utils.py
DeNeutoy: The problem is you can't assign a lambda function in an `or` statement, which seems to be a weird python quirk....

DeNeutoy:
@matt-gardner this is ready now

allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    min_span_width : ``int``, optional (default = None)
+        The minimum length of spans which should be included. Defaults to 1.
+    filter_function : ``Callable[[List[T]], bool]``, optional (default = None)
+        A function mapping sequences of the passed type T to a boolean value.
+        If ``True``, the span is included in the returned spans from the
+        sentence, otherwise it is excluded..
+    """
+    max_span_width = max_span_width or len(sentence)
+    min_span_width = min_span_width or 1
+    filter_function = filter_function or bool
```
matt-gardner: You can do `filter_function = filter_function or (lambda x: True)` to get around the syntax problem.

allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -0,0 +1,131 @@
...
+        The maximum length of spans which should be included. Defaults to len(sentence).
+    min_span_width : ``int``, optional (default = None)
+        The minimum length of spans which should be included. Defaults to 1.
+    filter_function : ``Callable[[List[T]], bool]``, optional (default = None)
+        A function mapping sequences of the passed type T to a boolean value.
+        If ``True``, the span is included in the returned spans from the
+        sentence, otherwise it is excluded..
+    """
+    max_span_width = max_span_width or len(sentence)
+    min_span_width = min_span_width or 1
```
matt-gardner: Why not put the 1 up above in the method signature? (Remember to change the docstring if you make this change.)

allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -0,0 +1,131 @@
+from typing import List, Tuple, Callable, TypeVar
+
+from allennlp.data.dataset_readers.dataset_utils.ontonotes import TypedStringSpan
+
+T = TypeVar("T")
```
matt-gardner: What's the reasoning for making this generic?

DeNeutoy: `strings` vs `Tokens` mainly - the only reason I did this is so that it's explicit that the filter_function must be able to deal with the type you provide.

matt-gardner: Then maybe `T = TypeVar('T', str, Token)`?  And, yeah, that's a good point about the filter function.

DeNeutoy: Oh, I didn't realise you could make things generic but over specific types. Cool, i'll do that.

PR #710, opened 2018-01-23
add make-vocab command

joelgrus:
I factored out the loading of datasets so there's not too much code duplication


allennlp/commands/make_vocab.py
```diff
@@ -0,0 +1,95 @@
+"""
+The ``make-vocab`` subcommand allows you to create a vocabulary from
+your dataset[s], which you can then reuse without recomputing it
+each training run
```
matt-gardner: Missing period.

PR #711, opened 2018-01-24
Add flags to avoid recording calls in the database and caching.

schmmd:


schmmd:
@joelgrus ping

allennlp/service/server_flask.py
```diff
@@ -154,6 +154,12 @@ def predict(model_name: str) -> Response:  # pylint: disable=unused-variable
         if request.method == "OPTIONS":
             return Response(response="", status=200)
 
+        # Do log if no argument is specified
+        record_flag = not request.args.get("record", "true").lower() == "false"
```
joelgrus: `!=` would be more idiomatic (and more readable) than `not ==`

allennlp/service/server_flask.py
```diff
@@ -154,6 +154,12 @@ def predict(model_name: str) -> Response:  # pylint: disable=unused-variable
...
+
+        # Do use the cache if no argument is specified
+        cache_flag = not request.args.get("cache", "true").lower() == "false"
```
joelgrus: same comment

allennlp/service/server_flask.py
joelgrus: also, I would prefer more descriptive names than `_flag`, e.g. `record_to_database` and `check_cache`

allennlp/service/server_flask.py
```diff
@@ -162,39 +168,44 @@ def predict(model_name: str) -> Response:  # pylint: disable=unused-variable
 
         log_blob = {"model": model_name, "inputs": data, "cached": False, "outputs": {}}
 
-        # See if we hit or not. In theory this could result in false positives.
-        pre_hits = _caching_prediction.cache_info().hits  # pylint: disable=no-value-for-parameter
+        if cache_flag:
```
joelgrus: I feel like you made this too complicated, I would have just changed line 169 to

```
if check_cache and cache_size > 0
```

and then line 182 to

```
if record_to_database and demo_db is not None:
```

is there a reason why that wouldn't just work?

schmmd:
Thanks @joelgrus, all great feedback!

PR #713, opened 2018-01-24
Fix for the var generation issue and action sequence to logical form transformation

pdasigi:
This PR includes a fix for the out of context `var` generation issue I was seeing while training the NLVR parser, which should also affect the wikitables parser. The issue is that since `var` has type `<#1,#1>`, the decoder can produce it practically anywhere in the action sequence. However, it should be produced only within the scope of lambda functions, right before the variable is used (e.g.: `(lambda x (fb:row.row.index (var x)))`). Due to the way we implemented constrained decoding, I found it easier to not produce `var` while decoding at all, and simply add it back when translating the action sequences into logical forms. This meant that I had to remove `var` functions when generating action sequences from target logical forms as well.

I was going to make a separate PR for the action sequence to logical form transformation, but since it is closely related to the `var` issue, I included that in this PR as well. However, the transformation doesn't work for WikiTables yet (It almost does!) I'll make a separate PR for that.

This currently depends on #607. I'll rebase after that gets merged.


pdasigi:
@matt-gardner This is now ready for review.

allennlp/data/semparse/worlds/world.py
```diff
@@ -122,7 +142,39 @@ def get_action_sequence(self, expression: Expression) -> List[str]:
...
+        Examples:
+            e -> 0
+            <r,e> -> 1
+            <e,<e,t>> -> 2
+            <b,<<b,#1>,<#1,b>>> -> 3
+        """
+        if not "<" in type_signature:
+            return 0
+        # We need to find the return type from the signature. We do that by removing the outer most
+        # angular brackets and travering the remaining substring till the angular brackets (if any)
```
matt-gardner: `s/travering/traversing/`

allennlp/models/encoder_decoders/nlvr_semantic_parser.py
```diff
@@ -146,9 +146,46 @@ def forward(self,  # type: ignore
 
         outputs = self._decoder_trainer.decode(initial_state, self._decoder_step,  # type: ignore
                                                self._max_decoding_steps)
-
+        best_action_sequences = outputs['best_action_sequence']
+        get_action_string = lambda rule: "%s -> %s" % (rule["left"][0], rule["right"][0])
+        labels_data = label.data.cpu()
```
matt-gardner: You should call `.numpy()` here, maybe also `.tolist()`, it should make things faster.  Or maybe because you're calling `.data.cpu()` it's enough...  Hmm, I'm not really sure.

tests/data/semparse/worlds/nlvr_world_test.py
```diff
@@ -90,3 +90,98 @@ def test_logical_form_with_box_filter_returns_correct_action_sequence(self):
...
+    def test_get_logical_form_with_real_logical_forms(self):
+        nlvr_world = self.worlds[0]
+        logical_form = ("(assert_greater_equals \
+                          (count (filter_equals all_boxes (lambda x (count (blue (object_in_box (var x))))) 0)) \
+                          1)")
+        parsed_logical_form = nlvr_world.parse_logical_form(logical_form)
+        action_sequence = nlvr_world.get_action_sequence(parsed_logical_form)
+        reconstructed_logical_form = nlvr_world.get_logical_form(action_sequence)
+        parsed_reconstructed_logical_form = nlvr_world.parse_logical_form(reconstructed_logical_form)
+        # It makes more sense to compare parsed logical forms instead of actual logical forms.
```
matt-gardner: Why?

tests/data/semparse/worlds/nlvr_world_test.py
```diff
@@ -90,3 +90,98 @@ def test_logical_form_with_box_filter_returns_correct_action_sequence(self):
...
+        assert parsed_logical_form == parsed_reconstructed_logical_form
+        assert nlvr_world.execute(logical_form) == nlvr_world.execute(reconstructed_logical_form)
+        logical_form = "(assert_equals (color (circle (touch_wall (all_objects)))) color_black)"
+        parsed_logical_form = nlvr_world.parse_logical_form(logical_form)
+        action_sequence = nlvr_world.get_action_sequence(parsed_logical_form)
+        reconstructed_logical_form = nlvr_world.get_logical_form(action_sequence)
+        print(reconstructed_logical_form)
```
matt-gardner: Remove print.

tests/data/semparse/worlds/nlvr_world_test.py
```diff
@@ -90,3 +90,98 @@ def test_logical_form_with_box_filter_returns_correct_action_sequence(self):
...
+                          (count (filter_equals all_boxes (lambda x (count (blue (object_in_box (var x))))) 0)) \
+                          1)")
+        parsed_logical_form = nlvr_world.parse_logical_form(logical_form)
+        action_sequence = nlvr_world.get_action_sequence(parsed_logical_form)
+        reconstructed_logical_form = nlvr_world.get_logical_form(action_sequence)
+        parsed_reconstructed_logical_form = nlvr_world.parse_logical_form(reconstructed_logical_form)
+        # It makes more sense to compare parsed logical forms instead of actual logical forms.
+        assert parsed_logical_form == parsed_reconstructed_logical_form
+        assert nlvr_world.execute(logical_form) == nlvr_world.execute(reconstructed_logical_form)
+        logical_form = "(assert_equals (color (circle (touch_wall (all_objects)))) color_black)"
```
matt-gardner: Newline above here.  Might make sense to have this in a loop, because these are identical, but it's fine to do it this way, too, as long as there's a break.

tests/data/semparse/worlds/wikitables_world_test.py
```diff
@@ -88,6 +88,24 @@ def test_world_returns_correct_actions_with_two_reverses(self):
...
+        actions_with_var = self.world.get_action_sequence(expression)
+        assert '<#1,#1> -> var' in actions_with_var
+        assert 'r -> x' in actions_with_var
+
+    def test_world_returns_correct_actions_with_lambda_without_var(self):
+        sempre_form = ("((reverse fb:cell.cell.date) ((reverse fb:row.row.year) (argmax (number 1) "
+                       "(number 1) (fb:row.row.league fb:cell.usl_a_league) (reverse (lambda x "
+                       "((reverse fb:row.row.index) (var x)))))))")
+        expression = self.world.parse_logical_form(sempre_form)
+        actions_with_var = self.world.get_action_sequence(expression)
```
matt-gardner: `actions_without_var`?

allennlp/data/semparse/worlds/nlvr_world.py
```diff
@@ -126,6 +126,61 @@ def _apply_function_list(self,
...
+            The sequence of actions as strings (eg.: ``['@START@ -> t', 't -> <e,t>', ...]``).
+        add_var_function : ``bool`` (optional)
+             ``var`` is a special function that some languages use within lambda functions to
+             indicate the use of a variable (eg.: ``(lambda x (fb:row.row.year (var x)))``). Due to
+             the way constrained decoding is currently implemented, it is easier for the decoder to
+             not produce these functions. In that case, setting this flag adds the function in the
+             logical form even though it is not present in the action sequence.
+        """
+        # TODO(pradeep): This method almost works with LambdaDCS logical forms in WikiTables as
+        # well. The thing that breaks it is "reverse", which is a higher order function. Once I
```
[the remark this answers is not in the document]
pdasigi: Yes, since `reverse` is a higher order function. `infer_num_arguments` will return `2` for `reverse`, whereas it takes only one argument (the function being reversed), and returns another function, and not a basic type like the other functions.

allennlp/data/semparse/worlds/nlvr_world.py
```diff
@@ -126,6 +126,61 @@ def _apply_function_list(self,
...
+        # figure out a way to deal with it cleanly, this should go in ``World``.
+        split_actions = [action.split(" -> ") for action in action_sequence]
+        terminals: List[Tuple[str, int]] = []  # terminal name and the number of arguments it takes
+        for lhs, rhs in split_actions:
```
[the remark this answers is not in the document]
pdasigi: Done.

allennlp/data/semparse/worlds/world.py

tests/data/semparse/worlds/nlvr_world_test.py
pdasigi: Added to comment:
".. because there can be slight differences between the actual logical form strings, like extra set of parentheses or spaces, which neither the type inference logic nor the executor cares about. So the test shouldn't either."

tests/data/semparse/worlds/nlvr_world_test.py
pdasigi: Done.

tests/data/semparse/worlds/nlvr_world_test.py
pdasigi: Done.

tests/data/semparse/worlds/wikitables_world_test.py
pdasigi: Fixed. Thanks!

PR #716, opened 2018-01-24
Compute linking features earlier in KnowledgeGraphField

matt-gardner:
This just moves where the linking features are computed in `KnowledgeGraphField`.  I'm doing this because I want to pre-compute those features and pass them in, and this will let me do that easier.


matt-gardner:
I finished the preprocessing script, and I'm just going to combine the two changes into a single PR.  Closing this one.

PR #717, opened 2018-01-24
Add a preprocessing script for wikitables data, to save on load times

matt-gardner:
This adds a script that pre-computes the expensive parts of processing the wikitables data and saves it to a jsonl file.  I had to change how the `KnowledgeGraphField` computes things a bit to make it possible to pass in pre-computed things.


scripts/preprocess_wikitables.py
```diff
@@ -0,0 +1,52 @@
...
+            targets[-1].append(action_map[target_index_field.sequence_index])
+
+    json_obj['target_action_sequences'] = targets
+
+    entity_texts = []
+    for entity_text in instance.fields['table'].entity_texts:
+        tokens = [{'text': token.text, 'lemma': token.lemma_} for token in entity_text]
+        entity_texts.append(tokens)
+    json_obj['entity_texts'] = entity_texts
+    json_obj['linking_features'] = instance.fields['table'].linking_features
```
pdasigi: Looks like these will be huge sparse arrays. May be you can represent them as dicts instead?

matt-gardner: They are 7 numbers per token, per entity.  It's not a huge deal to store them this way, and the logic to go to and from sparse dicts isn't really worth it, I think.

PR #719, opened 2018-01-25
Change backspace to carriage return in TeeLogger.

schmmd:
I'm skeptical the original code ever worked, as the character `\x08` is a backspace.  This changes the check to a carriage return, which would reset the cursor to the start of the line without advancing to the next line.

I verified this change breaks up the TQDM output in `output_path/stderr.log` when training BiDAF.


matt-gardner:
The original code did work, but for Keras' logging, not for tqdm.  It wasn't written for tqdm, because Keras did there own thing, using the backspace character.

schmmd:
Ah, I see how it worked now.  Keras had a whole bunch of backspace characters on the line.

schmmd:
Happy to change this to re.compile("\r$").sub("\n", line) if anyone thinks that's better (remove trailing carriage returns).

PR #720, opened 2018-01-25
SpanPruner and new coref model

DeNeutoy:
This is going to be a work in progress for a little bit, so I don't know if we want it to go into master - If anyone has a preferred way of doing this lmk. 

- Adds a `SpanPruner`
- Integrates the `SpanPruner` into the Coref model.

P.S most of this diff is adding a duplicated Coref model.


allennlp/models/coreference_resolution/coref_v2.py
```diff
@@ -0,0 +1,744 @@
...
+                                                             text_mask,
+                                                             span_starts,
+                                                             span_ends)
+        # Prune based on mention scores.
+        num_spans_to_keep = int(math.floor(self._spans_per_word * document_length))
+
+        (top_span_embeddings, top_span_mask,
+         top_span_indices, top_span_mention_scores) = self._mention_pruner(span_embeddings,
+                                                                           span_mask.squeeze(-1),
+                                                                           num_spans_to_keep)
```
DeNeutoy: Here is where the `SpanPruner` is added to the coref model. It replaces the [`_prune_and_sort_spans`](https://github.com/allenai/allennlp/blob/master/allennlp/models/coreference_resolution/coref.py#L531) method.

tests/modules/span_pruner_test.py
```diff
@@ -0,0 +1,50 @@
...
+        spans[1, 0] = 0
+        spans[1, 3] = 0
+        _, pruned_mask, pruned_indices, _ = pruner(spans, mask, 2)
+
+        # Second element in the batch would have indices 2, 3, but
+        # 3 and 0 are masked, so instead it has 1, 2.
+        numpy.testing.assert_array_equal(pruned_indices.data.numpy(), numpy.array([[0, 1],
+                                                                                   [1, 2],
+                                                                                   [2, 3]]))
+        numpy.testing.assert_array_equal(pruned_mask.data.numpy(), numpy.ones([3, 2]))
```
matt-gardner: You might add a check that the scorer parameters have gradients, given my comment about using `Module` above.  Not certain that's really worth the effort, though, especially if you just change the type annotation so that people pass in modules.

tests/modules/span_pruner_test.py
```diff
@@ -0,0 +1,50 @@
...
+
+        spans = Variable(torch.randn([3, 4, 5])).clamp(min=0.0, max=1.0)
+        spans[0, :2, :] = 1
+        spans[1, 2:, :] = 1
+        spans[2, 2:, :] = 1
+
+        mask = Variable(torch.ones([3, 4]))
+        spans[1, 0] = 0
+        spans[1, 3] = 0
+        _, pruned_mask, pruned_indices, _ = pruner(spans, mask, 2)
```
matt-gardner: Why not check the other values here, too?

tests/modules/span_pruner_test.py
```diff
@@ -0,0 +1,50 @@
...
+import torch
+from torch.autograd import Variable
+
+from allennlp.modules import SpanPruner
+from allennlp.common.testing import AllenNlpTestCase
+from allennlp.common.checks import ConfigurationError
+
+
+class TestSpanPruner(AllenNlpTestCase):
+    def test_forward_works_on_simple_input(self):
```
matt-gardner: This test seems redundant given the one below.

tests/modules/span_pruner_test.py
matt-gardner: And will you ever have a case where the pruned mask is not just a tensor of ones?

allennlp/modules/span_pruner.py
```diff
@@ -0,0 +1,94 @@
...
+        top_span_scores : ``torch.FloatTensor``, required.
+            The values of the top-k scoring spans. Has shape (batch_size, num_spans_to_keep, 1).
+        """
+        span_mask = span_mask.unsqueeze(-1)
+        num_spans = span_embeddings.size(1)
+        # Shape: (batch_size, num_spans, 1)
+        span_scores = self._scorer(span_embeddings)
+
+        if span_scores.size(-1) != 1 or span_scores.dim() != 3:
+            raise ConfigurationError(f"The scorer passed to SpanPruner must produce a tensor of shape"
```
matt-gardner: Seems more like a `ValueError` than a `ConfigurationError` to me - you're not instantiating any of this `from_params`.

allennlp/modules/span_pruner.py
```diff
@@ -0,0 +1,94 @@
...
+from allennlp.common.checks import ConfigurationError
+
+class SpanPruner(torch.nn.Module):
+    """
+    This module scores and prunes span based representations using a parameterised scoring
+    function and a threshold.
+
+    Parameters
+    ----------
+    scorer : ``Callable[[torch.FloatTensor], torch.FloatTensor]``, required.
```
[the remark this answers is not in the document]
DeNeutoy: I guess I was thinking of the case that you want the scorer to do nothing, as you've passed in a span_embedding of shape (batch_size, num_spans, 1) which are already the scores. This module handles that case and I guess it still would if the type annotation was for `Module`, so I think I agree

allennlp/modules/span_pruner.py
```diff
@@ -0,0 +1,94 @@
...
+        top_span_embeddings : ``torch.FloatTensor``, required.
+            The span representations of the top-k scoring spans.
+            Has shape (batch_size, num_spans_to_keep, embedding_size).
+        top_span_mask : ``torch.LongTensor``, required.
+            The coresponding mask for ``top_span_embeddings``. Has shape
+            (batch_size, num_spans_to_keep).
+        top_span_indices : ``torch.IntTensor``, required.
+            The indices of the top-k scoring spans. Has shape (batch_size, num_spans_to_keep).
+        top_span_scores : ``torch.FloatTensor``, required.
+            The values of the top-k scoring spans. Has shape (batch_size, num_spans_to_keep, 1).
```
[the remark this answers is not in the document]
DeNeutoy: Ugh yeah I know, the current coref model is very annoying for this. I started to change it, but it's big enough to be in it's own PR afterwards because it meant making a lot of the really confusing functions like `batch_index_select` support tensors which don't have the same shape.

tests/modules/span_pruner_test.py
DeNeutoy: Yes - in the case that `num_spans_to_keep` is greater than the number of non-masked spans.

matt-gardner: I'd recommend capturing that case in this test, then.

DeNeutoy: done

allennlp/modules/span_pruner.py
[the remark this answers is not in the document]
matt-gardner: Not sure if you're still working on this PR or if you're waiting for me - what did you decide on this change?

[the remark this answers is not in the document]
DeNeutoy: This type is more "correct" - I also don't understand why it wouldn't train properly, because we're saving a reference to it in `init`?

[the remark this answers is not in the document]
DeNeutoy: I don't follow, if you pass a scorer in here which isn't a module, why would you even expect it to 1) have parameters or 2) be trainable in pytorch? But ok if you feel strongly i'll change it. Can you approve?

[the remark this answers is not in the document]
matt-gardner: The point of the annotation is to tell the reader what they should pass in.  My point is just that we should be telling the reader they should pass in a `Module`, not any arbitrary `Callable`.

PR #721, opened 2018-01-25
fix setup.py to not import version

DeNeutoy:


joelgrus:
`exec` makes me nervous, I was envisioning something slightly different, e.g. a file called `VERSION` that just contains the text

```
0.3.1-unreleased
```

and setup.py just reads that file, and then `allennlp/version.py` can do something like

```
with open('VERSION') as f:
    VERSION = f.read().strip()

VERSION_SHORT = ''.join(VERSION.split('.')[:2])
```

DeNeutoy:
that's harder because then we have to make sure it's included in the pip distribution, etc etc.

joelgrus:
ok, you know more about those issues than I do

Issue #599, opened 2017-12-08
Experiments run with `no_tqdm: true` still have TQDM output in logs


mjlangan:
Found while investigating some issues with Beaker's log handling. It appears that when I run experiments with `no_tqdm: true` in my `trainer`'s config JSON, I will still see TQDM output for the steps "Indexing dataset" and "Iterating over dataset"

See the log output for this Beaker task: http://beaker.dev.allenai.org/tk/tk_jt3ahr2hxmch

DeNeutoy:
This is the same issue as #524 - basically we have the TQDM flag for training, but not for evaluation at test time.

joelgrus:
this is slightly broader, though. `train.py` calls `index_instances()` a couple of times

https://github.com/allenai/allennlp/blob/master/allennlp/commands/train.py#L164

and that has a tqdm in it:

https://github.com/allenai/allennlp/blob/master/allennlp/data/dataset.py#L58

at least for this case it would be easy enough to add a `no_tqdm` parameter to `index_instances` (and similarly to `evaluate()` and modify `train.py` to specify it.

DeNeutoy:
Those points are less painful though because they are much faster - if you see the points in the beaker log where `.index_instances` is called, you'll see it's only ~3 tqdm statements, whereas the evaluation of a dataset can take much longer and be much more verbose, making the lines unreadable.

matt-gardner:
It might be easier, though, to just modify our `TeeLogger` to detect tqdm output, and add a newline.  It'd remove a bunch of complexity from the code, and we wouldn't have to pass `no_tqdm` everywhere.

schmmd:
From @DeNeutoy: here is the `TeeLogger` for DeepQA which has a hack to address this issue: https://github.com/allenai/deep_qa/blob/master/deep_qa/common/tee_logger.py

schmmd:
The issue in the linked job is pretty minor.  The text display on beaker.dev.ai2 collapses the multi-line TQDM output to a single line.  When I copy and paste to GitHub the TQDM output becomes multiple lines.

```
...
2017-12-08T19:55:33.094778899Z 2017-12-08 19:55:33,094 - INFO - allennlp.data.vocabulary - Fitting token dictionary from dataset.
2017-12-08T19:55:33.620434278Z 
  0%|          | 0/21892 [00:00<?, ?it/s]
 20%|#9        | 4366/21892 [00:00<00:00, 43641.62it/s]
[...]

PR #722, opened 2018-01-25
Add a --file-friendly-logging argument for Beaker

schmmd:
The `--file-friendly-logging` argument turns carriage returns (`\r`) into newlines (`\n`) and slows the minimum update interval from 0.1s to 10s.  Carriage returns are used by `tqdm` and `requests` to report progress on a single line of stdout, but they do not work if the output is redirected to a file.  With a minimum update interval of 0.1s training logs can include over 100 MB of tqdm output.

This argument should be applied when running jobs in Beaker.

Fixes https://github.com/allenai/allennlp/issues/599


schmmd:
Sample Beaker job with `--tqdm-newline`: http://beaker.allenai.org/tk/tk_rstpjqkk9kaq

schmmd:
@DeNeutoy I didn't put `--tqdm-newline` in `train_model` because I thought it was high level or important, but rather because it's configuration about the training mechanism itself rather than the specification for the model to be trained.

I'd ultimately like to make sure this is always set when training within Beaker.  With a command-line argument, this is trivial.  With a model configuration, this is not as straightforward and it will ultimately probably be left to the user.

schmmd:
@DeNeutoy what do you think?  If @matt-gardner also would like this setting in the model configuration I'll move it there, but if I do that I likely won't turn it on for our users when they run Beaker jobs.

matt-gardner:
I'm inclined to agree with @schmmd on this one - let the config file specify modeling / training stuff, let the CLI specify output stuff. We already put the output directory and logging level in the CLI, not the config file; this way follows that in putting logging format decisions in the CLI.

schmmd:
If we're concerned about number of arguments we have, I'm happy to roll up this configuration into a class.

DeNeutoy:
I'm not sure I agree with your primary statement - e.g we have other configuration which effects logging levels which is passed in the configuration file, such as the tensorboard logging frequency. I just think it's a bit of a confusing argument - e.g if you're writing your own allennlp command  and you see this flag, is it important? But basically if you think it's better go ahead.

Also, if the argument could be called something less ambiguous like `add_tqdm_newlines` that would be good too, I had to look to see which way around the flag was.

schmmd:
@DeNeutoy happy to rename it--although I'll rename it to `add_tqdm_newlines` because that's what it does (it replaces carriage returns with newlines).

I wasn't aware that we had tensorboard logging frequency configuration in the json files.  What other logging configuration do we have there?

My goal is to have a "run on beaker" script enforce this behavior.  Do you have ideas for other mechanisms where I can accomplish that goal?  I don't think I can accomplish that if I put it in the configuration.

DeNeutoy:
Right, and we already have a way to do that via `--overrides`, no?

Actually my bad, they aren't parameters - they're just hard coded. https://github.com/allenai/allennlp/blob/master/allennlp/training/trainer.py#L144

schmmd:
@DeNeutoy good point re: `--overrides`.  I hadn't thought about that.

I'm going to:
1.  Rename the flag.
2.  Make the tqdm interval configurable.

And then maybe we can have a broader discussion of the best approach.  It feels quite odd to have logging configuration in the model json but it would work for my goals with `--overrides`.

matt-gardner:
Good point @DeNeutoy about tensorboard logging, I forgot about that.  I was thinking about how we [specify the logging level](https://github.com/allenai/allennlp/blob/7edcdf5a90a8ca067badb8903ad30bf50c18e0ec/allennlp/run.py#L6-L9) with an environment variable.

I don't have any strong feelings on what the best thing is to do here.  I can see good arguments for separating out the logging / serialization configuration from the model configuration, and I can see good arguments for just having a single file that has everything (and the json keys naturally separate the logging / serialization info from the model spec, anyway).

matt-gardner:
Not a fan of `--clean-output` - that's implicitly saying that the tqdm output isn't already clean, and I disagree with that statement.  It just doesn't work well on beaker, and doesn't save well in logs in some cases.  `--clean-output` also isn't very descriptive - what does "clean" mean?

schmmd:
tqdm output isn't clean when redirected to a file.

I don't care what it's called.  @DeNeutoy didn't like `tqdm-newlines` and now the flag also slows the update rate so I came up with something else.

DeNeutoy:
❤️ `--clean_tqdm_output` ❤️

schmmd:
I went with `--file-friendly-logging` which @matt-gardner suggested over Slack.  @matt-gardner would you CR?

allennlp/commands/train.py
```diff
@@ -141,8 +150,10 @@ def train_model(params: Params, serialization_dir: str) -> Model:
     prepare_environment(params)
 
     os.makedirs(serialization_dir, exist_ok=True)
-    sys.stdout = TeeLogger(os.path.join(serialization_dir, "stdout.log"), sys.stdout)  # type: ignore
-    sys.stderr = TeeLogger(os.path.join(serialization_dir, "stderr.log"), sys.stderr)  # type: ignore
+    sys.stdout = TeeLogger(os.path.join(serialization_dir, "stdout.log"), # type: ignore
+                           sys.stdout, file_friendly_logging)
```
matt-gardner: Preference for one argument per line when you need to wrap arguments like this, but it's not a big deal if you prefer it this way.

allennlp/common/tee_logger.py
```diff
@@ -5,6 +5,16 @@
 from typing import TextIO
 import os
 
+def clean(message: str):
```
matt-gardner: Can you give this a more descriptive name, like `replace_return_with_newline`?  Also, make the comment into a docstring.

allennlp/common/tqdm.py
```diff
@@ -0,0 +1,22 @@
+"""
+:class:`~allennlp.common.tqdm.Tqdm` wraps tqdm so we can add configurable
+global defaults for certain tqdm parameters.
+"""
+
+from tqdm import tqdm as _tqdm
+
+class Tqdm:
+    default_mininterval: float = 0.1
```
matt-gardner: Maybe give a comment here that this is the default value in `tqdm`, so that this isn't just a magic number.

PR #725, opened 2018-01-26
eliminate Dataset abstraction

joelgrus:
OK, I think I feel pretty good about this. People might take issue with the names or a few of the API choices, but for the most part I think it's a really solid design change.

The number of files changed is scary, but the lines of code changed is not actually that much.

Here are the main points:

* rename `Dataset` to `Batch`, get rid of `LazyDataset` and `InstanceCollection`. the idea is that you'd really only ever use it for batches. Basically everywhere you would have previously used a `Dataset` (other than as a minibatch) you now just use an `Iterable[Instance]`.
* made `DatasetReader.read()` private, and had it return an `Iterable[Instance]`. The choice of `Iterable` (and not `Iterator`) was deliberate, as that means that returning a list of instances is fine. So in every existing case, this simply required changing

```
return Dataset(instances)
```

to

```
return instances
```

* added a `DatasetReader.instances() -> Iterable[Instance]` method as the public API. The idea is that it has to return an iterable that can be iterated over multiple times (e.g. a list or an object whose `__iter__` method returns a fresh iterator on each call). There's no way to enforce this, but the default implementation does this.

* added a `DatasetReader.lazy = False` flag that subclasses can set to True if they want. when it's False, `DatasetReader.instances()` returns a list containing all the instances. When it's True, `instances()` returns an object whose `__iter__` method basically calls `DatasetReader._read()`. In that case, if `_read()` is lazy, then `.instances()` will be too. This basically implements the "lazy dataset reader" functionality from mattg's PR.
* added an `ensure_list` function that makes sure that an iterable is a list but is a no-op when it already is
* changed `DataIterator` to take an `Iterable[Instance]` everywhere it would have previously taken a `Dataset` / `InstanceCollection`
* added an `Instance.indexed` flag so that we can index them multiple times and just no-op if they've already been indexed
* added an `DataIterator.index_with(vocab)` method that hangs onto the vocab and then calls `Batch.index_instances(vocab)` for each batch it generates (which is a no-op if they've already been indexed)

-------------

this is a breaking change, primarily in the change from

```
DatasetReader.read() -> Dataset
```

to

```
DatasetReader._read() -> Iterable[Instance]
```

which means that any custom dataset readers will have to be fixed (it's not a big fix). Any custom `DataIterator`s would be very broken, but I don't know of anyone doing those. Also, I guess I had to make changes to `commands/train` and `commands/evaluate`, so if someone has made custom versions of those, they would probably break too (in ways that are obvious to me but possibly not to them).

### DIFF
--- a/allennlp/data/dataset.py
+++ b/allennlp/data/dataset.py
@@ -5,11 +5,9 @@ For example, when you train a model, you will likely have a *training* dataset a
 
 import logging
 from collections import defaultdict
-from typing import Dict, List, Union, Iterable, Iterator, Callable
+from typing import Dict, List, Union, Iterator
 
 import torch
-import tqdm
-from overrides import overrides
 
 from allennlp.data.instance import Instance
 from allennlp.data.vocabulary import Vocabulary
@@ -17,44 +15,15 @@ from allennlp.common.checks import ConfigurationError
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
-
-class InstanceCollection(Iterable):
+class Batch:
     """
-    This is the abstract base class for :class:`~allennlp.data.instance.Dataset`
-    and :class:`~allennlp.data.instance.LazyDataset` objects. As its name indicates,
-    it's a collection of :class:`~allennlp.data.instance.Instance` objects
-    that can be iterated over. Depending on the subclass, the instances may be stored
-    in memory, or they may be generated (e.g. from disk) and discarded each iteration.
-    The ``Instances`` have ``Fields``, and the fields could be in an indexed or unindexed state -
-    the ``index_instances`` method indexes the data.
-    """
-    def __iter__(self) -> Iterator[Instance]:
-        """
-        Returns an iterator that ranges over every Instance in the Dataset.
-        """
-        raise NotImplementedError
-
-    def index_instances(self, vocab: Vocabulary) -> None:
-        """
-        Ensures that all ``Instances`` have been indexed using the provided vocabulary.
-        The indexing may or may not happen immediately, but it's guaranteed to happen
-        before your next iteration.
-        """
-        raise NotImplementedError
-
-
-class Dataset(InstanceCollection):
-    """
-    This class is used to represent both an entire dataset (that's small enough to fit in memory)
-    and also a batch from a larger dataset. In addition to the ``InstanceCollection`` methods,
+    A batch of Instances. In addition to containing the instances themselves,
     it contains helper functions for converting the data into tensors.
     """
     def __init__(self, instances: List[Instance]) -> None:
         """
         A Dataset just takes a list of instances in its constructor and hangs onto them.
         """
-        super().__init__()
-
         all_instance_fields_and_types: List[Dict[str, str]] = [{k: v.__class__.__name__
                                                                 for k, v in x.fields.items()}
                                                                for x in instances]
@@ -63,12 +32,6 @@ class Dataset(InstanceCollection):
             raise ConfigurationError("You cannot construct a Dataset with non-homogeneous Instances.")
 
         self.instances = instances
-
-    @overrides
-    def index_instances(self, vocab: Vocabulary) -> None:
-        logger.info("Indexing dataset")
-        for instance in tqdm.tqdm(self.instances):
-            instance.index_fields(vocab)
 
     def get_padding_lengths(self) -> Dict[str, Dict[str, int]]:
         """
@@ -183,42 +146,9 @@ class Dataset(InstanceCollection):
             final_fields[field_name] = field_classes[field_name].batch_tensors(field_tensor_list)
         return final_fields
 
-    @overrides
     def __iter__(self) -> Iterator[Instance]:
         return iter(self.instances)
 
-
-
-class LazyDataset(InstanceCollection):
-    """
-    A Dataset that contains a way of generating instances, rather than a
-    concrete list of them.
-
-    Parameters
-    ----------
-    instance_generator: ``Callable[[], Iterator[Instance]])``
-        This function should be callable multiple times, and each time it should
-        return an iterator that ranges over all instances.
-    """
-    def __init__(self,
-                 instance_generator: Callable[[], Iterator[Instance]]) -> None:
-        super().__init__()
-        self.generator = instance_generator
-        self.vocab: Vocabulary = None
-
-    @overrides
     def index_instances(self, vocab: Vocabulary) -> None:
-        """
-        A ``LazyDataset`` doesn't have a collection of instances ready to
-        iterate over; instead we'll need to call ``Instance.index_fields``
-        as the instances are generated. So here we just grab a reference to
-        the ``Vocabulary`` so that we can do the indexing at iteration time.
-        """
-        self.vocab = vocab
-
-    @overrides
-    def __iter__(self) -> Iterator[Instance]:
-        for instance in self.generator():
-            if self.vocab is not None:
-                instance.index_fields(self.vocab)
-            yield instance
+        for instance in self.instances:
+            instance.index_fields(vocab)
--- a/allennlp/data/dataset.py
+++ b/allennlp/data/dataset.py
@@ -29,6 +29,7 @@ class Batch(Iterable):
         super().__init__()
 
         self.instances: List[Instance] = ensure_list(instances)
+        self._check_types()
 
     def _check_types(self) -> None:
         """
--- a/allennlp/data/dataset.py
+++ b/allennlp/data/dataset.py
@@ -26,6 +26,8 @@ class Batch(Iterable):
         A Dataset just takes an iterable of instances in its constructor and hangs onto them
         in a list.
         """
+        super().__init__()
+
         self.instances = ensure_list(instances)
 
         all_instance_fields_and_types: List[Dict[str, str]] = [{k: v.__class__.__name__
--- a/allennlp/data/dataset.py
+++ b/allennlp/data/dataset.py
@@ -28,8 +28,12 @@ class Batch(Iterable):
         """
         super().__init__()
 
-        self.instances = ensure_list(instances)
+        self.instances: List[Instance] = ensure_list(instances)
 
+    def _check_types(self) -> None:
+        """
+        Check that all the instances have the same types.
+        """
         all_instance_fields_and_types: List[Dict[str, str]] = [{k: v.__class__.__name__
                                                                 for k, v in x.fields.items()}
                                                                for x in self.instances]
--- a/allennlp/data/dataset.py
+++ b/allennlp/data/dataset.py
@@ -5,33 +5,35 @@ For example, when you train a model, you will likely have a *training* dataset a
 
 import logging
 from collections import defaultdict
-from typing import Dict, List, Union, Iterator
+from typing import Dict, List, Union, Iterator, Iterable
 
 import torch
 
+from allennlp.common.checks import ConfigurationError
+from allennlp.common.util import ensure_list
 from allennlp.data.instance import Instance
 from allennlp.data.vocabulary import Vocabulary
-from allennlp.common.checks import ConfigurationError
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
-class Batch:
+class Batch(Iterable):
     """
     A batch of Instances. In addition to containing the instances themselves,
     it contains helper functions for converting the data into tensors.
     """
-    def __init__(self, instances: List[Instance]) -> None:
+    def __init__(self, instances: Iterable[Instance]) -> None:
         """
-        A Dataset just takes a list of instances in its constructor and hangs onto them.
+        A Dataset just takes an iterable of instances in its constructor and hangs onto them
+        in a list.
         """
+        self.instances = ensure_list(instances)
+
         all_instance_fields_and_types: List[Dict[str, str]] = [{k: v.__class__.__name__
                                                                 for k, v in x.fields.items()}
-                                                               for x in instances]
+                                                               for x in self.instances]
         # Check all the field names and Field types are the same for every instance.
         if not all([all_instance_fields_and_types[0] == x for x in all_instance_fields_and_types]):
             raise ConfigurationError("You cannot construct a Dataset with non-homogeneous Instances.")
-
-        self.instances = instances
 
     def get_padding_lengths(self) -> Dict[str, Dict[str, int]]:
         """
--- a/allennlp/data/dataset.py
+++ b/allennlp/data/dataset.py
@@ -1,6 +1,6 @@
 """
-A :class:`~Dataset` represents a collection of data suitable for feeding into a model.
-For example, when you train a model, you will likely have a *training* dataset and a *validation* dataset.
+A :class:`~Batch` represents a collection of ``Instance`` s to be fed
+through a model.
 """
 
 import logging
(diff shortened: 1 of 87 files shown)


schmmd:
I have not followed all the discussion in Slack recently, but I think this sounds like an excellent change!  In particular, `s/Dataset/Batch` (and not using `Dataset` for both the dataset and each batch) would have helped me when I first started looking at the codebase.

allennlp/data/instance.py
```diff
@@ -37,9 +38,13 @@ def index_fields(self, vocab: Vocabulary):
         """
         Converts all ``UnindexedFields`` in this ``Instance`` to ``IndexedFields``, given the
```
DeNeutoy: Can you correct this docstring, `IndexedFields` is a hangover from DeepQA

allennlp/data/instance.py
```diff
@@ -37,9 +38,13 @@ def index_fields(self, vocab: Vocabulary):
...
         ``Vocabulary``.  This `mutates` the current object, it does not return a new ``Instance``.
+        A ``DataIterator`` will call this on each pass through a dataset; we use the ``indexed``
```
DeNeutoy: We need to be sure we'd only ever want  to index in-memory instances a single time. E.g what happens when we add to our test set vocab and we want to re-index - do we have to read the instances from disk again? Do we care about doing that?

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -1,18 +1,66 @@
...
+
+        If ``self.lazy`` is False, this calls ``self._read()``,
+        calls ``ensure_list`` on the result (which is a no-op if
+        self._read() returns a list, which most of the existing
+        dataset readers do), and returns the resulting list.
+
+        If ``self.lazy`` is True, this returns an object whose
+        ``__iter__`` method calls ``self._read()`` each iteration.
+        Note that if your implementation of ``_read()`` is not lazy
+        (i.e. it loads all instances into memory at once), then your
```
matt-gardner: This is a minor distinction, maybe, but what if we have `_read` _always_ return a generator, and change the `lazy` flag to a `cache` flag.  It seems semantically a little cleaner, as it would avoid this issue where `lazy` doesn't do what you think if you didn't implement the reader the right way.

allennlp/data/iterators/lazy_basic_iterator.py
```diff
@@ -6,13 +6,13 @@
...
 from allennlp.data.iterators.data_iterator import DataIterator
-from allennlp.data.dataset import LazyDataset, Dataset
+from allennlp.data.dataset import Batch
 from allennlp.data.instance import Instance
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 @DataIterator.register("lazy-basic")
-class LazyBasicIterator(DataIterator[LazyDataset]):
+class LazyBasicIterator(DataIterator):
```
matt-gardner: Not for this PR, but will this change let us get rid of the distinction between lazy and non-lazy iterators?

allennlp/data/dataset.py
```diff
@@ -5,71 +5,42 @@
...
+    it contains helper functions for converting the data into tensors.
     """
-    def __iter__(self) -> Iterator[Instance]:
-        """
-        Returns an iterator that ranges over every Instance in the Dataset.
-        """
-        raise NotImplementedError
-
-    def index_instances(self, vocab: Vocabulary) -> None:
+    def __init__(self, instances: Iterable[Instance]) -> None:
```
DeNeutoy: Can't you change this type annotation to `List[Instance]` given that it must always be that? Then you could remove the `ensure_list` check, particularly because you also do this check in all the `DataIterators` before creating a `Batch` anyway.

allennlp/data/instance.py
DeNeutoy: Ok chatting over lunch I don't think this is a problem anymore.

allennlp/data/dataset_readers/dataset_reader.py
joelgrus: I thought about that, the things I didn't like is that (1) it's more of a breaking change, and (2) generators are "advanced" python, so I didn't want to force people to use them

allennlp/data/instance.py
```diff
@@ -14,8 +14,8 @@ class Instance:
     as outputs.
 
     The ``Fields`` in an ``Instance`` can start out either indexed or un-indexed.  During the data
-    processing pipeline, all fields will end up as ``IndexedFields``, and will then be converted
-    into padded arrays by a ``DataGenerator``.
+    processing pipeline, all fields will be indexed, after multiple instances can be combined into
```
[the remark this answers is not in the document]
joelgrus: oops, should be "after which"

allennlp/data/dataset_readers/dataset_reader.py
matt-gardner: Actually, because it's a different API for `_read`, and `read` still has the same API, I don't think it breaks anything, right?

You're second point is a good one, though...  Hmm...  What if instead of having this warning in the docstring, you check for whether the result of `_read` is already a list `if self.lazy`?  If it's already a list, you could log a warning and do the same thing as in the non-lazy case.

matt-gardner: Also, scratch that "read still has the same API" business - I forgot for a minute that you renamed it to `instances`.

joelgrus: I actually called the public method `instances` here. I am a little bit torn: if I leave it as `read` then current code will continue working. But it will sort of continue working "for the wrong reasons"

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -1,18 +1,66 @@
...
+    A ``DatasetReader`` knows how to turn a file containing a dataset into a collection
+    of ``Instance`` s.  To implement your own, just override the `_read(file_path)` method
+    to return an ``Iterable`` of the instances. This could be a list containing the instances
+    or a lazy generator that returns them one at a time.
+
+    All parameters necessary to _read the data apart from the filepath should be passed
+    to the constructor of the ``DatasetReader``.
     """
-    def read(self, file_path: str) -> InstanceCollection:
+    lazy = False
```
matt-gardner: Why not have this be a constructor parameter?  It also doesn't show up in the class docstring.

I'm guessing you're doing this to avoid breaking downstream code, because they probably don't call the superclass constructor.  I think you can check for this in `instances`, set the default there, and log a deprecation warning, or something.  That seems better than settling for a worse superclass API.

allennlp/data/iterators/lazy_basic_iterator.py
joelgrus: yes and no. you can see I got rid of the distinction at the type level, but e.g. `BucketIterator` calls `ensure_list` so it can shuffle its instances, and you wouldn't want to do that with a lazy dataset

allennlp/data/dataset_readers/dataset_reader.py
matt-gardner: I could go either way too on `read` versus `instances`, I wasn't suggesting here that you change it back.

allennlp/data/dataset_readers/dataset_reader.py
joelgrus: ok, adding a constructor

allennlp/data/iterators/lazy_basic_iterator.py
matt-gardner: I think you can just add a parameter to `BucketIterator` that tells how many instances to pull into memory before bucketing, defaulting to the whole dataset.  You can also do that with the `BasicIterator`, before shuffling.  I _think_ that would just fix all of the issues, and we can get rid of the `Lazy` version entirely.  But, again, that's a not-too-high-priority issue for later.

DeNeutoy:
There are also quite a few scripts which need updating for this change too.

joelgrus:
which scripts are you thinking of? I fixed the ones that mypy complained about

DeNeutoy:
Sorry my bad, I didn't actually check to see that you'd caught them already.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -1,18 +1,82 @@
...
+        calls ``ensure_list`` on the result (which is a no-op if
+        self._read() returns a list, which most of the existing
+        dataset readers do), and returns the resulting list.
+
+        If ``self.lazy`` is True, this returns an object whose
+        ``__iter__`` method calls ``self._read()`` each iteration.
+        Note that if your implementation of ``_read()`` is not lazy
+        (i.e. it loads all instances into memory at once), then your
+        "lazy" dataset reader is just loading the entire dataset into
+        a list each time you iterate over it, which is probably not
```
matt-gardner: You crash now, so this comment should be updated.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -1,18 +1,82 @@
...
+                                         "Is the path correct?".format(file_path))
+            return instances
+
+    def _read(self, file_path: str) -> Iterable[Instance]:
         """
-        Actually reads some data from the `file_path` and returns a :class:`Dataset`.
+        Reads the instances from the given file_path and returns them as an
+        `Iterable` (which could be a list or could be a generator).
+        You are strongly encouraged to use a generator, so that users can
+        read dataset in a lazy way, if they so choose.
```
matt-gardner: fix grammar: "can read dataset"

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -1,18 +1,82 @@
...
+        (i.e. it loads all instances into memory at once), then your
+        "lazy" dataset reader is just loading the entire dataset into
+        a list each time you iterate over it, which is probably not
+        what you want.
+
+        In either case, the returned ``Iterable`` can be iterated
+        over multiple times. It's unlikely you want to override this function,
+        but if you do your result should likewise be repeatedly iterable.
+        """
+        if self.lazy:
```
matt-gardner: Do you think it's worth checking `getattr(self, 'lazy', None)`, and logging a warning if it's not set?  This means they didn't call the super class constructor.

allennlp/data/dataset_readers/snli.py
```diff
@@ -37,15 +35,15 @@ class SnliReader(DatasetReader):
     def __init__(self,
                  tokenizer: Tokenizer = None,
                  token_indexers: Dict[str, TokenIndexer] = None) -> None:
+        super().__init__()
```
matt-gardner: What's the general recommendation in python for using a bare `super()` versus `super(Class, self)`?  I learned to use the latter, so that's what I've always done.  Does it matter?

allennlp/data/instance.py
```diff
@@ -14,16 +14,17 @@ class Instance:
...
+    processing pipeline, all fields will be indexed, after which multiple instances can be combined
+    into a ``Batch`` and then converted into padded arrays.
 
     Parameters
     ----------
     fields : ``Dict[str, Field]``
         The ``Field`` objects that will be used to produce data arrays for this instance.
     """
-    def __init__(self, fields: Dict[str, Field]) -> None:
+    def __init__(self, fields: Mapping[str, Field]) -> None:
```
matt-gardner: Why change this to `Mapping`?

allennlp/data/instance.py
```diff
@@ -33,13 +34,17 @@ def count_vocab_items(self, counter: Dict[str, Dict[str, int]]):
...
 
-    def index_fields(self, vocab: Vocabulary):
+    def index_fields(self, vocab: Vocabulary) -> None:
         """
-        Converts all ``UnindexedFields`` in this ``Instance`` to ``IndexedFields``, given the
-        ``Vocabulary``.  This `mutates` the current object, it does not return a new ``Instance``.
+        Indexes all fields in this ``Instance`` using the provided ``Vocabulary``.
+        This `mutates` the current object, it does not return a new ``Instance``.
+        A ``DataIterator`` will call this on each pass through a dataset; we use the ``indexed``
+        flag to make sure that indexing only happens once.
```
matt-gardner: You might add a warning somewhere (maybe here, maybe somewhere else; not really sure) that if for some reason your vocab changes during the course of your program and you try to re-index stuff, you might get weird, unpredictable behavior.  I think it's _really_ unlikely that someone might do this, but still, a warning somewhere might be a good idea.

tests/data/dataset_test.py
```diff
@@ -19,22 +19,24 @@ def setUp(self):
...
+        self.instances = self.get_instances()
         super(TestDataset, self).setUp()
 
     def test_instances_must_have_homogeneous_fields(self):
         instance1 = Instance({"tag": (LabelField(1, skip_indexing=True))})
         instance2 = Instance({"words": TextField([Token("hello")], {})})
         with pytest.raises(ConfigurationError):
-            _ = Dataset([instance1, instance2])
+            batch = Batch([instance1, instance2])
+            batch._check_types()  # pylint: disable=protected-access
```
matt-gardner: This isn't called in the constructor anymore?  Why not?

tests/notebooks_test.py
```diff
@@ -3,16 +3,20 @@
 import nbformat
 from nbconvert.preprocessors.execute import CellExecutionError
 from nbconvert.preprocessors import ExecutePreprocessor
+import pytest
 
 from allennlp.common.testing import AllenNlpTestCase
 
 class TestNotebooks(AllenNlpTestCase):
+    @pytest.mark.skip()
```
matt-gardner: If you're disabling these, can you open an issue to fix them before the next release?

allennlp/data/dataset_readers/dataset_reader.py
joelgrus: done

allennlp/data/dataset_readers/dataset_reader.py
joelgrus: done

allennlp/data/dataset_readers/dataset_reader.py
joelgrus: done

allennlp/data/dataset_readers/snli.py
joelgrus: not sure about general recommendation, the docs just use `super()`:

https://docs.python.org/3/library/functions.html#super

my understanding is that if you're inside a class method they're equivalent (that is, the `Class` is implicitly the current class), but if you're not inside a class method then you're required to be explicit (obviously)

allennlp/data/instance.py
joelgrus: for some reason I don't understand `mypy` started complaining at me that `Dict` isn't covariant in its value type and that I should use `Mapping` instead. I don't really understand what I changed to start getting that warning, but I just did it and then mypy was happy.

allennlp/data/instance.py
joelgrus: done

tests/data/dataset_test.py
joelgrus: I thought we talked about how this check wasn't really appropriate in e.g. multi-task learning? I moved it to a method so we could add it back later if we decided we wanted it in some cases.

tests/notebooks_test.py
joelgrus: oops, I only disabled them because I didn't want to iterate on the notebooks until we'd agreed what the new API should look like. I fixed the notebooks and re-enabled these tests.

tests/data/dataset_test.py
matt-gardner: I think `Batch` actually makes this more clear.  We can handle batching together instances with different fields in the data iterator, and each batch should have a consistent set of fields.  Otherwise things totally break.  So, yeah, I'd move this check back to the constructor.

joelgrus: done

PR #730, opened 2018-01-29
Span extractor

DeNeutoy:
Adds a `SpanExtractor` abstraction along with 2 concrete instances of extractors, both of which are required for the Coref model.


allennlp/modules/span_extractors/endpoint_span_extractor.py
```diff
@@ -0,0 +1,72 @@
...
+from allennlp.common.params import Params
+from allennlp.modules.span_extractors.span_extractor import SpanExtractor
+from allennlp.modules.token_embedders.embedding import Embedding
+from allennlp.nn.util import batched_index_select, combine_tensors
+from allennlp.data.vocabulary import Vocabulary
+
+@SpanExtractor.register("endpoint")
+class EndpointSpanExtractor(SpanExtractor):
+    """
+    Represents spans as a function of the embeddings of their endpoints. Additionally,
```
matt-gardner: "as an element-wise function of ...", maybe?  When I first read this I wondered if you allowed parameterized (MLP) functions, but you don't.

allennlp/modules/span_extractors/locally_normalised_span_extractor.py
```diff
@@ -0,0 +1,96 @@
...
+from allennlp.common.params import Params
+from allennlp.modules.span_extractors.span_extractor import SpanExtractor
+from allennlp.modules.time_distributed import TimeDistributed
+from allennlp.nn import util
+
+@SpanExtractor.register("locally_normalised")
+class LocallyNormalisedSpanExtractor(SpanExtractor):
+    """
+    Computes span representations by generating an unnormalized attention score for each
+    word in the document, compute distributions over every span with respect to these
```
matt-gardner: Awkward grammar in this sentence.  It should probably be made into two sentences.

allennlp/modules/span_extractors/locally_normalised_span_extractor.py
```diff
@@ -0,0 +1,96 @@
...
+    scores by normalising the attention scores for words inside the span.
+
+    Given these attention distributions over every span, weight the corresponding vector
```
matt-gardner: I'd recommend declarative instead of imperative descriptions here.  "Given ..., this module weights the", or whatever your favorite way of doing that is.

allennlp/modules/span_extractors/locally_normalised_span_extractor.py
```diff
@@ -0,0 +1,96 @@
+import torch
+from overrides import overrides
+
+from allennlp.common.params import Params
+from allennlp.modules.span_extractors.span_extractor import SpanExtractor
+from allennlp.modules.time_distributed import TimeDistributed
+from allennlp.nn import util
+
+@SpanExtractor.register("locally_normalised")
+class LocallyNormalisedSpanExtractor(SpanExtractor):
```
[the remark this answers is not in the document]
matt-gardner: Though, now that I think about it, I'm wondering if `LocallyNormalized` is even the right name here.  `SelfAttentiveSpanExtractor` might be more descriptive (or `SelfAttentionSpanExtractor`).  The issue is that `LocallyNormalized` doesn't say _what_ is locally normalized, and it's the _what_ that's more important than the fact that it's locally normalized.

allennlp/modules/span_extractors/locally_normalised_span_extractor.py
```diff
@@ -0,0 +1,96 @@
...
+        # of a given span is smaller than the max, the rest of the values
+        # are masked.
+        max_batch_span_width = int(span_widths.max().data) + 1
+
+        # shape (batch_size, sequence_length, 1)
+        global_attention_logits = self._global_attention(sequence_tensor)
+
+        # Shape: (1, 1, max_batch_span_width)
+        max_span_range_indices = util.get_range_vector(max_batch_span_width,
+                                                       sequence_tensor.is_cuda).view(1, 1, -1)
```
matt-gardner: Unrelated to this PR, but something I've been wondering: do you know if this `.is_cuda()` even works when you have multiple GPUs?  We're just calling `.cuda()` on the resulting tensor, without specifying a device, and this could be breaking things.  Should we be passing a tensor to this function to copy from instead of a boolean flag?

allennlp/modules/span_extractors/locally_normalised_span_extractor.py
```diff
@@ -0,0 +1,96 @@
...
+        # Shape: (batch_size, num_spans, max_batch_span_width)
+        # This is a broadcasted comparison - for each span we are considering,
+        # we are creating a range vector of size max_span_width, but masking values
+        # which are greater than the actual length of the span.
+        span_mask = (max_span_range_indices <= span_widths).float()
+        raw_span_indices = span_ends - max_span_range_indices
+        # We also don't want to include span indices which are less than zero,
+        # which happens because some spans near the beginning of the sequence
```
matt-gardner: This comment made me think a whole lot.  I think it's incorrect.  The issue isn't that the span _width_ is smaller than `max_batch_span_width`, it's that the span end _index_ is smaller than `max_batch_span_width`.

allennlp/modules/span_extractors/locally_normalised_span_extractor.py
```diff
@@ -0,0 +1,96 @@
...
+        global_attention_logits = self._global_attention(sequence_tensor)
+
+        # Shape: (1, 1, max_batch_span_width)
+        max_span_range_indices = util.get_range_vector(max_batch_span_width,
+                                                       sequence_tensor.is_cuda).view(1, 1, -1)
+        # Shape: (batch_size, num_spans, max_batch_span_width)
+        # This is a broadcasted comparison - for each span we are considering,
+        # we are creating a range vector of size max_span_width, but masking values
+        # which are greater than the actual length of the span.
+        span_mask = (max_span_range_indices <= span_widths).float()
```
matt-gardner: This `<=` also made me think a whole lot, along with the `>= 0`.  I was wondering why there weren't any off-by-one errors here.  Your excellent test convinced me that the logic was correct, but then I had to figure out why it was correct.  Because the span starts and ends are inclusive, your `span_widths` calculation above is actually off by one, and you compensate for that here with a `<=` instead of a `<`.  The fact that `raw_span_indices` actually gives the indices _backwards_ also confused me a bit.  There was a lot going on in these few lines... =).

I'm not sure what the best thing to do here is, but at the least some more detailed comments would be nice, because this is a potentially pretty confusing block of code.  E.g., if you want to keep the off-by-one issue in `span_widths`, that's fine, as long as there's a note describing this.  Might also be worth stating (even with an example) that `raw_span_indices` will be backwards, so if you're doing a computation that depends on order, you might be surprised.

tests/modules/span_extractors/locally_normalised_span_extractor_test.py
```diff
@@ -0,0 +1,51 @@
...
+class TestLocallyNormalisedSpanExtractor:
+    def test_locally_normalised_span_extractor_can_build_from_params(self):
+        params = Params({"type": "locally_normalised", "input_dim": 5})
+        extractor = SpanExtractor.from_params(params)
+        assert isinstance(extractor, LocallyNormalisedSpanExtractor)
+
+    def test_attention_is_normalised_correctly(self):
+        input_dim = 7
+        sequence_tensor = Variable(torch.randn([2, 5, input_dim]))
+        # concatentate start and end points together to form our representation.
```
matt-gardner: Copy pasta.

tests/modules/span_extractors/locally_normalised_span_extractor_test.py
```diff
@@ -0,0 +1,51 @@
...
+        extractor = LocallyNormalisedSpanExtractor(input_dim=input_dim)
+
+        # In order to test the attention, we'll make the weight which computes the logits
```
matt-gardner: Excellent idea, this is a really nice test.

allennlp/modules/span_extractors/endpoint_span_extractor.py
```diff
@@ -0,0 +1,72 @@
...
+    For example, you might give ``x,y,x*y`` as the ``combination`` parameter to this class.
+    The computed similarity function would then be ``[x; y; x*y]``, which can then be optionally
+    concatenated with an embedded representation of the width of the span.
+
+    Parameters
+    ----------
+    combination : str, optional (default = "x-y").
+        The method used to combine the ``start_embedding`` and ``end_embedding``
+        representations. See above for a full description.
+    span_width_embedding : ``Embedding``, optional (default = None).
```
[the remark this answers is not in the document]
DeNeutoy: Yeah that's better.

allennlp/modules/span_extractors/endpoint_span_extractor.py
```diff
@@ -0,0 +1,72 @@
...
+                indicies: torch.LongTensor) -> None:
+        # shape (batch_size, num_spans)
+        span_starts, span_ends = [index.squeeze(-1) for index in indicies.split(1, dim=-1)]
+        start_embeddings = batched_index_select(sequence_tensor, span_starts)
+        end_embeddings = batched_index_select(sequence_tensor, span_ends)
+
+        combined_tensors = combine_tensors(self._combination, [start_embeddings, end_embeddings])
+        if self._span_width_embedding is not None:
+            # Embed the span widths and concatenate to the rest of the representations.
+            span_width_embeddings = self._span_width_embedding(span_ends - span_starts)
```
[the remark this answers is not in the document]
DeNeutoy: For one span embedding it is, for one it isn't -.- I've made it optional, thanks.

allennlp/modules/span_extractors/locally_normalised_span_extractor.py
DeNeutoy: The `raw_span_indices` are actually permutation invariant, because we use the indices generated to select both the embeddings _and_ the attention logits, so because they are the same for both, the order doesn't matter for any given span.

I agree that it's not ideal that this depends on having end-inclusive indices. I'm going to leave a TODO to add a flag to support both I think.

PR #731, opened 2018-01-29
Add a verify.py script that runs all tests

schmmd:
I frequently neglect to run tests locally because I need to remember:

* `pytest`
* `./scripts/pylint.sh`
* `./scripts/mypy.sh`
* `./scripts/check_docs.sh`
* `cd docs; build_docs.sh`

Some commands have wrapper scripts (`pylint.sh`, `mypy.sh`, `check_docs.sh`), others do not (`pytest` and make docs).  Some commands have some standard boilerplate before them (`pylint.sh` and `mypy.sh`) others don't (`check_docs.sh`).

We used to have [`test_script.sh`](https://github.com/allenai/allennlp/commit/6cb107cc2bdc12acd7e46cdc5eb5896e13bae6f6#diff-de315bd2a574dfb27d818766b1958b6aL1) but it used environment variables for arguments and was written in bash, so it wasn't particularly user friendly.

In this PR I add a script `verify.py` which can be used to run all verification steps (`verify.py --all`) or individual steps (`verify.py --checks mypy pylint`).  You can get usage to see which checks are available (`verify.py --all`).

In the future, we could add more configuration, such as `--colorize` (set colored output) or `--tee target/` (save output to a file).

Previous commands are left in place so nobody needs to change their workflow.

### DIFF
--- a/scripts/mypy.sh
+++ b/scripts/mypy.sh
@@ -2,6 +2,5 @@
 # Run type checking over the python code.
 
 set -e
-echo 'Starting mypy checks'
-mypy allennlp --ignore-missing-imports
+./scripts/verify.py --checks mypy
 echo -e "mypy checks passed\n"
--- a/scripts/mypy.sh
+++ b/scripts/mypy.sh
@@ -1,6 +1,4 @@
 #!/usr/bin/env bash
 # Run type checking over the python code.
 
-set -e
 ./scripts/verify.py --checks mypy
-echo -e "mypy checks passed\n"
--- a/scripts/verify.py
+++ b/scripts/verify.py
@@ -5,10 +5,11 @@
 
 import argparse
 
-from subprocess import run
-from subprocess import CalledProcessError
 import os
 import shutil
+from subprocess import run
+from subprocess import CalledProcessError
+import sys
 
 def main(checks):
     try:
@@ -37,7 +38,7 @@ def main(checks):
 
     except CalledProcessError:
         # squelch
-        pass
+        sys.exit(1)
 
 if __name__ == "__main__":
     checks = ['pytest', 'pylint', 'mypy', 'build-docs', 'check-docs']
--- a/scripts/verify.py
+++ b/scripts/verify.py
@@ -1,0 +1,50 @@
+#! /usr/bin/env python
+
+"""Script that runs all verification steps.
+"""
+
+import argparse
+
+from subprocess import run
+import os
+import shutil
+
+def main(checks):
+    print("Verifying with " + str(checks))
+    if "pytest" in checks:
+        print("Tests (pytest):", flush=True)
+        run("pytest -v --color=yes", shell=True, check=True)
+
+    if "pylint" in checks:
+        print("Linter (pylint):", flush=True)
+        run("pylint -d locally-disabled,locally-enabled -f colorized allennlp tests", shell=True, check=True)
+
+    if "mypy" in checks:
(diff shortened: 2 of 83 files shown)


schmmd:
@joelgrus mind having a first look?  Then I'll assign others.

scripts/mypy.sh
```diff
@@ -1,7 +1,4 @@
 #!/usr/bin/env bash
 # Run type checking over the python code.
 
-set -e
-echo 'Starting mypy checks'
-mypy allennlp --ignore-missing-imports
-echo -e "mypy checks passed\n"
+./scripts/verify.py --checks mypy
```
matt-gardner: This will change the behavior of CI, right?  Did you check that TeamCity still gives us the same output if these checks fail?

schmmd: Ah shoot--you're right, I need the `set -e`.

I'm not sure what you mean by "add some unnecessary output".  Would you clarify?  What output would be added?

matt-gardner: I'm not certain this will happen, but it sure looks like you'll get an extra stack trace from the python code that isn't really necessary (the `CalledProcessError` from your call to `subprocess.run()` _should_ print a stack trace to the console).  It'd be nice to run this through CI with a failing test case or something just to see what happens.

schmmd: Nevermind--the exit code will still be the same.  I don't need `set -e`.

schmmd: Ah, good point.  I'll test and verify.  I don't think this will be a problem, but I understand your concern.

schmmd: @matt-gardner you're right.  I added some code to squelch the extra exception.

```
  File "./scripts/verify.py", line 52, in <module>
    main(run_checks)
  File "./scripts/verify.py", line 25, in main
    run("mypy allennlp --ignore-missing-imports", shell=True, check=True)
  File "/Users/michaels/miniconda3/envs/allennlp/lib/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command 'mypy allennlp --ignore-missing-imports' returned non-zero exit status 1.
```

scripts/verify.py
```diff
@@ -6,32 +6,38 @@
...
+            print("Documentation (build):", flush=True)
+            run("cd docs; make html-strict", shell=True, check=True)
+
+        if "checks-docs" in checks:
+            print("Documentation (check):", flush=True)
+            run("./script/check_docs.py", shell=True, check=True)
+
+    except CalledProcessError:
+        # squelch
+        pass
```
matt-gardner: If you `pass` here, this script won't work for CI, because there's no indication of failure.  You need to do `sys.exit(-1)`.  Also, having the `try` block on the outside will stop the checks at the first failure; seems reasonable to want to run all of them even if one fails.

schmmd: I'll try to return the exit code of the subprocess.

@matt-gardner I had intended for this script to exit when one command fails, but if you'd prefer I can make it run all of them.  I wasn't planning on running this directly with `--all` from TeamCity.

matt-gardner: You're the one that's going to be using it, so as long as TeamCity still runs all of the checks even if one fails, it doesn't much matter to me if this script exits when one fails - whatever works for you.

schmmd:
@matt-gardner I ended up just exiting with `sys.exit(1)`.

PR #734, opened 2018-01-29
Support using a directory as an archive file.

schmmd:
Allows a directory to be specified as an "archive file".  This change makes it easier to troubleshoot model files as they are being developed.


schmmd:
@OyvindTafjord was this the feature you wanted a while back?  I happened to be looking at this code and this change looked pretty straightforward.  If this is what you are looking for, I'll look over my solution more carefully and put it up for review.

OyvindTafjord:
Yes, that looks like it would have been helpful in my earlier use case at least, like running evaluation against the model from various epochs

PR #735, opened 2018-01-30
Agenda improvements

pdasigi:
I've added more rules in the agenda mapping scheme.
From the training logs, it seemed like the parser was having difficulty reaching the terminal states because it does not have any signal indicating whether it is taking the right path. So I added an option to include the non-terminal productions from the paths to the agenda as well. However, I do not see noticeable differences in the parser's training yet.

This PR depends on #713 (which depends on #607)

### DIFF
--- a/allennlp/data/semparse/type_declarations/nlvr_type_declaration.py
+++ b/allennlp/data/semparse/type_declarations/nlvr_type_declaration.py
@@ -192,3 +192,10 @@ add_common_name_with_type("medium", "Z1", OBJECT_FILTER_TYPE)
 add_common_name_with_type("big", "Z2", OBJECT_FILTER_TYPE)
 
 add_common_name_with_type("negate_filter", "N", NEGATE_FILTER_TYPE)
+
+# Adding numbers because they commonly occur in utterances. They're usually between 1 and 9. Since
+# they're not too many of these productions, we're adding them to the global mapping instead of a
+# local mapping in each world.
+for num in range(1, 10):
+    num_string = str(num)
+    add_common_name_with_type(num_string, num_string, NUM_TYPE)
--- a/allennlp/data/semparse/type_declarations/nlvr_type_declaration.py
+++ b/allennlp/data/semparse/type_declarations/nlvr_type_declaration.py
@@ -194,7 +194,7 @@ add_common_name_with_type("big", "Z2", OBJECT_FILTER_TYPE)
 add_common_name_with_type("negate_filter", "N", NEGATE_FILTER_TYPE)
 
 # Adding numbers because they commonly occur in utterances. They're usually between 1 and 9. Since
-# they're not too many of these productions, we're adding them to the global mapping instead of a
+# there are not too many of these productions, we're adding them to the global mapping instead of a
 # local mapping in each world.
 for num in range(1, 10):
     num_string = str(num)
--- a/allennlp/data/semparse/worlds/world.py
+++ b/allennlp/data/semparse/worlds/world.py
@@ -67,6 +67,7 @@ class World:
                                       "most three nested lambda expressions")
         self._logic_parser = types.DynamicTypeLogicParser(constant_type_prefixes=type_prefixes,
                                                           type_signatures=self.global_type_signatures)
+        self._right_side_indexed_actions: Dict[str, List[Tuple[str, str]]] = None
 
     def get_name_mapping(self) -> Dict[str, str]:
         # Python 3.5 syntax for merging two dictionaries.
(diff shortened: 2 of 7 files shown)


matt-gardner:
Looks like this one still needs some updating.

pdasigi:
@matt-gardner This is now ready for review.

allennlp/data/semparse/type_declarations/nlvr_type_declaration.py
```diff
@@ -192,3 +192,10 @@ def add_common_name_with_type(name, mapping, type_signature):
 add_common_name_with_type("big", "Z2", OBJECT_FILTER_TYPE)
 
 add_common_name_with_type("negate_filter", "N", NEGATE_FILTER_TYPE)
+
+# Adding numbers because they commonly occur in utterances. They're usually between 1 and 9. Since
+# they're not too many of these productions, we're adding them to the global mapping instead of a
```
matt-gardner: `s/they're/there are/`

allennlp/data/semparse/worlds/world.py
```diff
@@ -82,6 +83,52 @@ def get_valid_actions(self) -> Dict[str, List[str]]:
...
+                          beam_size: int = 30,
+                          max_num_paths: int = 10) -> List[List[str]]:
+        """
+        For a given action, returns at most ``max_num_paths`` paths to the root (production with
+        ``@START@``) that are not longer than ``max_path_length``.
+        """
+        all_actions = self.all_possible_actions()
+        rhs_indexed_actions: Dict[str, List[Tuple[str, str]]] = defaultdict(list)
+        for possible_action in all_actions:
+            lhs, rhs = possible_action.split(' -> ')
```
matt-gardner: `left_side` and `right_side` instead of `lhs` and `rhs`.

allennlp/data/semparse/worlds/world.py
```diff
@@ -82,6 +83,52 @@ def get_valid_actions(self) -> Dict[str, List[str]]:
...
+                          action: str,
+                          max_path_length: int = 20,
+                          beam_size: int = 30,
+                          max_num_paths: int = 10) -> List[List[str]]:
+        """
+        For a given action, returns at most ``max_num_paths`` paths to the root (production with
+        ``@START@``) that are not longer than ``max_path_length``.
+        """
+        all_actions = self.all_possible_actions()
+        rhs_indexed_actions: Dict[str, List[Tuple[str, str]]] = defaultdict(list)
```
matt-gardner: You're creating this dictionary every time that `get_paths_to_root` is called, and it doesn't depend on the inputs.  This is going to be quite slow; you should cache intermediate results here using a method call.

allennlp/data/semparse/worlds/world.py
```diff
@@ -82,6 +83,52 @@ def get_valid_actions(self) -> Dict[str, List[str]]:
...
+                    if next_action in actions:
+                        # Ignoring paths with loops (of size 1)
+                        continue
+                    new_actions = list(actions)
+                    new_actions.append(next_action)
+                    # Ignoring lists that are too long, and have too many repetitions.
+                    path_length = len(new_actions)
+                    if path_length <= max_path_length or next_lhs == types.START_SYMBOL:
+                        new_lists.append((next_lhs, new_actions))
+            lists_to_expand = new_lists[:beam_size]
```
matt-gardner: Do you want to sort this beam by anything before truncating it?

tests/data/dataset_readers/nlvr_test.py
```diff
@@ -16,17 +17,17 @@ def test_reader_reads(self):
                            'a', 'box', '.']
         assert [t.text for t in sentence_tokens] == expected_tokens
         agenda = [item.sequence_index for item in instance.fields["agenda"].field_list]
-        assert agenda == [56, 32, 112]
```
matt-gardner: Why do these keep changing?  I'd recommend making them strings that don't change, unless you really care about the particular indices here.

tests/data/dataset_readers/nlvr_test.py
```diff
@@ -37,3 +38,17 @@ def test_agenda_indices_are_correct(self):
...
+        dataset = reader.read(test_file)
+        instance = dataset.instances[0]
+        sentence_tokens = instance.fields["sentence"].tokens
+        sentence = " ".join([t.text for t in sentence_tokens])
+        agenda = [item.sequence_index for item in instance.fields["agenda"].field_list]
+        actions = [action.rule for action in instance.fields["actions"].field_list]
+        agenda_actions = [actions[i] for i in agenda]
+        # pylint: disable=protected-access
+        expected_agenda_actions = reader._get_agenda_for_sentence(sentence, NlvrWorld({}))
+        assert expected_agenda_actions == agenda_actions
```
matt-gardner: I'd like to see an explicit test for `get_path_to_root` (in `world_test.py`, not this file).  You could construct a `World`, override `all_possible_actions()` to return a simple list, and make sure that `get_path_to_root` returns what you expect.

allennlp/data/semparse/type_declarations/nlvr_type_declaration.py
pdasigi: Done.

allennlp/data/semparse/worlds/world.py

allennlp/data/semparse/worlds/world.py

allennlp/data/semparse/worlds/world.py
pdasigi: The lists are all of the same length. I don't know if there's a good way to sort them. However, I rearranged the list before truncating it such that all the finished paths are at the top, so that they won't get truncated.

tests/data/dataset_readers/nlvr_test.py
pdasigi: In this case (and the one that's coming up), the strings would have changed too.

matt-gardner: Then this should definitely be a string, I think.

tests/data/dataset_readers/nlvr_test.py
pdasigi: Done.

tests/data/dataset_readers/nlvr_test.py
pdasigi: Okay, done.

PR #741, opened 2018-01-31
PTB dataset reader

DeNeutoy:
Reads parse trees in to a dataset for use with a span scoring model.


DeNeutoy:
Ah I need to update this for Joel's Dataset PR.

DeNeutoy:
Ok this is ready now.

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+@DatasetReader.register("ptb_trees")
+class PennTreeBankDatasetReader(DatasetReader):
+    """
+    Reads constituency parses from the WSJ part of the Penn Tree Bank from the LDC.
+
```
matt-gardner: Can you give an overview of the input/output here?  You have the input, but you don't have the output at all, and it's not what you might expect from the name of the class.  I might also rename this reader to something that reflects that this is a span tagging task.  All of the names I can think of are super long, but `ptb_trees` is misleading, so we should find something better.

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -0,0 +1,173 @@
...
+    Parameters
+    ----------
+    token_indexers : ``Dict[str, TokenIndexer]``, optional (default=``{"tokens": SingleIdTokenIndexer()}``)
+        We use this to define the input representation for the text.  See :class:`TokenIndexer`.
+        Note that the `output` tags will always correspond to single token IDs based on how they
+        are pre-tokenised in the data file.
+    """
+    def __init__(self,
+                 token_indexers: Dict[str, TokenIndexer] = None) -> None:
+        super().__init__(lazy=False)
```
matt-gardner: I think you should take a `lazy` parameter and pass it up, instead of only allowing `False` here.

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -0,0 +1,173 @@
...
+        spans: List[Field] = []
+        gold_labels = []
+
+        if gold_tree is not None:
+            gold_spans_with_pos_tags = self._get_gold_spans(gold_tree, 0, OrderedDict()).items()
+            gold_spans = {span: label for (span, label)
+                          in gold_spans_with_pos_tags if "-POS" not in label}
+        else:
+            gold_spans = None
+        for start, end in enumerate_spans(tokens):
```
matt-gardner: This function takes some configuration parameters - is it worth allowing the `DatasetReader` to configure this?

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -0,0 +1,173 @@
...
+            spans.append(SpanField(start, end, text_field))
+
+            if gold_spans is not None:
+                if (start, end) in gold_spans.keys():
+                    gold_labels.append(gold_spans[(start, end)])
+                else:
+                    gold_labels.append("NO-LABEL")
+
+        span_list_field: ListField = ListField(spans)
```
matt-gardner: Do you need the type annotation here?

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -0,0 +1,173 @@
...
+        tree : ``Tree``, required.
+            An NLTK parse tree to extract spans from.
+        index : ``int``, required.
+            The index of the current span in the sentence being considered.
+        typed_spans : ``OrderedDict[Tuple[int, int], str]``, required.
+            A ordered dictionary mapping spans to span labels.
+
+        Returns
+        -------
+        typed_spans : ``OrderedDict[Tuple[int, int], str]``, required.
```
matt-gardner: You don't need `required` on a return value.

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -0,0 +1,173 @@
...
+        Spans are inclusive.
+
+        Parameters
+        ----------
+        tree : ``Tree``, required.
+            An NLTK parse tree to extract spans from.
+        index : ``int``, required.
+            The index of the current span in the sentence being considered.
+        typed_spans : ``OrderedDict[Tuple[int, int], str]``, required.
+            A ordered dictionary mapping spans to span labels.
```
matt-gardner: You might give a note here that you're using an `OrderedDict` because you rely on its efficiency in the method implementation.  I had to read the comments to know why this wasn't just a `Dict`.

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -0,0 +1,173 @@
...
+            # their labels are pos tags. However, it makes the
+            # indexing more straightforward, so we'll collect them
+            # and filter them out below. We subtract 1 from the end
+            # index so the spans are inclusive.
+            typed_spans[(index, index + len(tree) - 1)] = tree.label() + "-POS"
+        else:
+            # otherwise, the tree has children.
+            first = True
+            for child in tree:
+                typed_spans = self._get_gold_spans(child, index, typed_spans)
```
matt-gardner: Overriding this variable, instead of assigning the return value to a new variable, is confusing to me, and seems error-prone.

On the other hand, it looks like you require this behavior, because of the way the recursive function collects its result.  I'd recommend just omitting the return value here, and doing something like:

```python
for child in tree:
    # We're ignoring the return value here because this call modifies typed_spans directly.
    self._get_gold_spans(child, index, typed_spans)
    ...
```

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -0,0 +1,173 @@
...
+                # OrderedDicts are implemented via a
+                # doubly linked list, so this reversal is O(1).
+                last_appended_span = next(reversed(typed_spans))
+                # Set the end index of the current span to
+                # the last appended index + 1, as the span was inclusive.
```
matt-gardner: You're just going to be subtracting off the 1 later; do you need to add it in here?

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -0,0 +1,173 @@
...
+                index = last_appended_span[1] + 1
+                if first:
+                    start = last_appended_span[0]
```
matt-gardner: I think you could make this a little bit simpler by removing `first` and just using `start` as your sentinel.  You can also just set `end` at every iteration of the loop.  So:

```python
start = None
for child in tree:
    ...
    if start is None:
        start = last_appended_span[0]
    end = last_appended_span[1]
typed_spans[(start, end)] = tree.label()
```

You might also note in a comment that taking the last appended span relies on the fact that you're adding the span for the subtree at the end of the method.

allennlp/data/dataset_readers/penn_tree_bank.py
DeNeutoy: It's mostly because of that, but python `Dict`s don't implement `__reversed__()`, so even though they have a deterministic order, you have to do something like `next(reversed(list(dict.keys())))`. But point taken, i'll add a comment.

allennlp/data/dataset_readers/penn_tree_bank.py
DeNeutoy: Yes, because of weird mypy generics.

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -0,0 +1,173 @@
...
+            # The "length" of a tree is defined by
+            # NLTK as the number of children.
+            # We don't actually want the spans for leaves, because
+            # their labels are pos tags. However, it makes the
+            # indexing more straightforward, so we'll collect them
+            # and filter them out below. We subtract 1 from the end
+            # index so the spans are inclusive.
+            typed_spans[(index, index + len(tree) - 1)] = tree.label() + "-POS"
+        else:
+            # otherwise, the tree has children.
```
vidurj: Could you simplify this else to 
``` 
else:
    typed_spans[(index, index + len(tree) - 1)] = tree.label()
    for child in tree:
        self._get_gold_spans(child, index, typed_spans)
        index += len(child)
```
?

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        # NLTK leaves are strings.
+        if isinstance(tree[0], str):
+            # The "length" of a tree is defined by
+            # NLTK as the number of children.
+            # We don't actually want the spans for leaves, because
+            # their labels are pos tags. However, it makes the
+            # indexing more straightforward, so we'll collect them
+            # and filter them out below. We subtract 1 from the end
+            # index so the spans are inclusive.
```
vidurj: Why do you want the spans to be inclusive?

allennlp/data/dataset_readers/penn_tree_bank.py
vidurj: Is `len(tree)` expensive?

vidurj: If you switch to this you could use a `Dict` instead of an `OrderedDict` too.

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -0,0 +1,173 @@
...
+            typed_spans[(index, index + len(tree) - 1)] = tree.label() + "-POS"
+        else:
```
vidurj: Could you simplify this else to the following? This way you could use a `Dict` instead of an `OrderedDict`.
```
else:
    child_start = index
    for child in tree:
       end = self._get_gold_spans(child, child_start, typed_spans)
       child_start = end
    typed_spans[index, end] = tree.label()
return end
```

allennlp/data/dataset_readers/penn_tree_bank.py
DeNeutoy: because it makes selecting indices from tensors easier if the spans we generate can just be mapped to indices into rows of tensors, rather than having to shift the end index first

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -0,0 +1,173 @@
...
+
+        Returns
+        -------
+        typed_spans : ``OrderedDict[Tuple[int, int], str]``, required.
+            A ordered dictionary mapping all subtree spans in the parse tree
+            to their constituency labels. Leaf nodes have POS tag spans, which
+            are denoted by a label of "LABEL-POS".
+        """
+        # NLTK leaves are strings.
+        if isinstance(tree[0], str):
```
vidurj: Do you know that `tree` does not have other children if this is true?

allennlp/data/dataset_readers/penn_tree_bank.py
matt-gardner: Oh, I missed what `index` was doing in `get_gold_spans`, so my treatment of `end` here is wrong...

allennlp/data/dataset_readers/penn_tree_bank.py
vidurj: That could make sense, but depending on if you add START and END tokens, you might have to shift the end index if you don't subtract 1.

allennlp/data/dataset_readers/penn_tree_bank.py
DeNeutoy: Yup, all leaf nodes are `Trees` with a single string child and a node cannot have a string child unless it is a leaf.

allennlp/data/dataset_readers/penn_tree_bank.py
DeNeutoy: None of them really make sense here - we always want to consider all possible span sizes, because you can have single word non-pre-terminal spans, and we always want the full sentence span to be the root. The filter functions could be useful, but these can't be passed via the configuration file anyway, so i'll leave this as it is for now I think.

allennlp/data/dataset_readers/penn_tree_bank.py
DeNeutoy: Yeah this span stuff is quite new, it would be great to get it working easily with just a flag to switch the inclusive exclusive thing. Although if you add start tokens don't you then have to shift things for the start index anyway?

vidurj: You wouldn't have to shift the start index, but that is because we want to index to the word right before the span. But we will have to do some index foo when dealing with the output of the backward lstm.

vidurj:
The changes look good to me.

PR #743, opened 2018-01-31
unify iterators / get rid of lazy iterator

joelgrus:
With this change the `Basic` and `Bucket` iterators work with both lazy and non-lazy datasets. (I made the corresponding changes to `Adaptive` too)

The key point is that I gave `BasicIterator` two additional (optional) parameters: `instances_per_epoch` and `max_instances_in_memory`, and then added a new private method to iterate over size-capped lists of instances.

```
def _memory_sized_lists(self, instances: Iterable[Instance]) -> Iterable[List[Instance]]:
```

the behavior is a little bit tricky, but 

* in the current case ("all instances in a list", new parameters not supplied), it just yields back the original list of instances.
* in a lazy dataset with `max_instances_in_memory` specified, it lazily returns lists of that many items until it exhausts the iterator
* in a lazy dataset without `max_instances_in_memory` specified, it lazily returns lists of `batch_size` items until it exhausts the iterator

then I changed `_create_batches` so that instead of batching all instances at once, it batches them one "memory-sized-list" at a time.  (this is slightly hard to explain, so look at the code and tests)

This change also allows you to specify `instances_per_epoch` for a non-lazy dataset. In that case each epoch will pick up where the previous one left off (and will wrap around if appropriate)

I made the corresponding changes in `BucketIterator` and `AdaptiveIterator`, but I didn't write any new tests for the adaptive one, since that one doesn't work.


allennlp/data/iterators/basic_iterator.py
```diff
@@ -22,27 +25,111 @@ class BasicIterator(DataIterator):
...
+        # we just load ``batch_size`` instances at a time:
+        if lazy and self._max_instances_in_memory is None:
+            yield from lazy_groups_of(iterator, self._batch_size)
+        # If we specified max instances in memory, lazy or not, we just
+        # load ``max_instances_in_memory`` instances at a time:
+        elif self._max_instances_in_memory is not None:
+            yield from lazy_groups_of(iterator, self._max_instances_in_memory)
+        # If we have non-lazy instances, and we want all instances each epoch,
+        # then we just yield back the list of instances:
+        elif self._instances_per_epoch is None:
```
matt-gardner: Do you need this case?  It seems like `self._take_instances` will already do what you want if `self._instances_per_epoch` is None, and you can just collapse this with the following case.  Am I missing something?

allennlp/data/iterators/bucket_iterator.py
```diff
@@ -56,35 +56,40 @@ def __init__(self,
...
-        super(BucketIterator, self).__init__(batch_size)
+        super(BucketIterator, self).__init__(batch_size=batch_size,
+                                             instances_per_epoch=instances_per_epoch,
+                                             max_instances_in_memory=max_instances_in_memory)
 
     @overrides
     def _create_batches(self, instances: Iterable[Instance], shuffle: bool) -> Iterable[Batch]:
-        instances = ensure_list(instances)
+        for instance_list in self._memory_sized_lists(instances):
+            if self._sorting_keys:
```
matt-gardner: I'm trying to think now why you would ever want a bucket iterator that doesn't have sorting keys...  I think we can remove this check, and make the argument required in the constructor and in `from_params`.  It doesn't make sense to bucket unless you give a means to sort, that's kind of the definition of the difference between the "basic" and the "bucket" iterators.  @DeNeutoy, do you agree?  Am I missing something?

allennlp/data/iterators/bucket_iterator.py
```diff
@@ -56,35 +56,40 @@ def __init__(self,
...
-            penultimate_batch = grouped_instances.pop()
-        if shuffle:
-            random.shuffle(grouped_instances)
-        else:
-            logger.warning("shuffle parameter is set to False,"
-                           " while bucket iterators by definition change the order of your data.")
-        if self._biggest_batch_first:
-            grouped_instances.insert(0, penultimate_batch)
-            grouped_instances.insert(0, last_batch)
+            grouped_instances = list(super(BucketIterator, self)._create_batches(instance_list, shuffle=False))
```
matt-gardner: Given what you told me yesterday, you can remove the arguments to `super()` here, I think.

tests/data/iterators/basic_iterator_test.py
```diff
@@ -47,35 +53,133 @@ def assert_instances_are_correct(self, candidate_instances):
...
+            # First epoch: 7 instances -> [2, 2, 2, 1]
+            batches = list(iterator._create_batches(test_instances, shuffle=False))
+            grouped_instances = [batch.instances for batch in batches]
+            assert grouped_instances == [[self.instances[0], self.instances[1]],
+                                         [self.instances[2], self.instances[3]],
+                                         [self.instances[4], self.instances[0]],
+                                         [self.instances[1]]]
+
+            # Second epoch: 7 instances -> [2, 2, 2, 1]
+            batches = list(iterator._create_batches(test_instances, shuffle=False))
```
matt-gardner: If `shuffle=True`, will the data get shuffled on every pass through the data in this case?

tests/data/iterators/bucket_iterator_test.py
```diff
@@ -14,6 +14,22 @@ def test_create_batches_groups_correctly(self):
                                      [self.instances[0], self.instances[1]],
                                      [self.instances[3]]]
 
+    def test_create_batches_groups_correctly_with_max_instances(self):
+        # Because max_instances_in_memory is 3, we bucket the first three,
+        # yield the two batches, then bucket the next 2, and yield that batch.
+        # Since the original order was 4 -> 2 -> 0 -> 1 -> 3,
```
matt-gardner: This took me a minute to figure out, because I was wondering why we didn't get `[4, 2], [0], [1, 3]`.  Maybe you can add a comment after the order saying `... -> 3, and we take the instances in index order, we should get ...`.

allennlp/data/iterators/basic_iterator.py
joelgrus: you're right that we could just return `list(iterator)` in both cases, but treating them separately avoids an unnecessary list copy (return instances as-is vs iterate and create a new list)

allennlp/data/iterators/bucket_iterator.py
joelgrus: sure

tests/data/iterators/basic_iterator_test.py
joelgrus: it should, I can add a test for it

tests/data/iterators/bucket_iterator_test.py
joelgrus: comment updated

allennlp/data/iterators/bucket_iterator.py
DeNeutoy: Yeah sorting keys should be required.

joelgrus: I'll leave this for a future PR

allennlp/data/iterators/bucket_iterator.py
```diff
@@ -56,35 +56,40 @@ def __init__(self,
...
+            move_to_front = self._biggest_batch_first and len(grouped_instances) > 1
+            if move_to_front:
+                # We'll actually pop the last _two_ batches, because the last one might not be full.
+                last_batch = grouped_instances.pop()
+                penultimate_batch = grouped_instances.pop() if grouped_instances else None
```
DeNeutoy: Seems like this can result in grouped_instances being equal to `[some_batch, None]` - is whatever consumes this robust to encountering `None`?

DeNeutoy: Can you update the docstring to state that `biggest_batch_first` now uses the biggest batch per block of instances that you have specified fit in memory.

allennlp/data/iterators/bucket_iterator.py
DeNeutoy: Seems like it would be quick, and very important - if you don't specify this, your model could train an order of magnitude slower

joelgrus: oops, hit merge too fast, I'll do another really quick PR

allennlp/data/iterators/bucket_iterator.py
joelgrus: actually, I should have just taken this out, because `move_to_front` already checks that there are at least two groups

PR #750, opened 2018-02-01
Set minimum tqdm version

matt-gardner:
I got an error with the new `refresh` flag to `set_description` because my `tqdm` version was too old.  Updating `tqdm` fixed the issue.  This means we require a minimum version for `tqdm`.


matt-gardner:
Ah, thanks @DeNeutoy, forgot about that.  Also, that's a really weird test failure - looks like it's due to an updated spacy version that has better tokenization...?

PR #752, opened 2018-02-01
Use /stage/allennlp as the working directory.

schmmd:


schmmd:
Tests are failing because master is broken.

PR #753, opened 2018-02-01
Update tests so they pass with spacy 2.0.6.

schmmd:


matt-gardner:
You should also change `requirements.txt` and `setup.py` to set 2.0.6 as the minimum spacy version.

schmmd:
@matt-gardner done

DeNeutoy:
If we remove `would've` as a token from this test, we wouldn't have to pin spacy dependencies. I would be a fan of this option.

schmmd:
I need to leave soon.  If someone else wants to fix master, they can take over this PR or merge it as is.

schmmd:
@DeNeutoy 's suggestion isn't how I would move forward.  It means we acknowledge additional variance in how users training and evaluations will run.

Overall, I'm not clear how we want to move forward or reach a decision.  While I think any route forward is better than the status quo, I'm not particularly inclined to make a PR for a route I wouldn't choose.

1.  Matt G and Michael's first opinion is to advance the Spacy version.
2.  Matt P's opinion is that we should pin Spacy to 2.0.5.
3.  Mark N's opinion is that we should remove the test case so people can use any version of Spacy 2.

Typically I don't strive for consensus and instead evaluate CR's on whether they make the codebase worse or not, rather than whether they are the best way forward.  If we merge this, but someone wants to make another PR for another direction, great.  Alternatively if someone else wants to put forth a PR with another solution, that's fine too.

DeNeutoy:
Seems like neither version of spacy will be incorrect and we want to maximise how easy allennlp is to use. Given this isn't a model change from spacy, which might fundamentally affect our models, I don't see why removing a token from a test is a poor approach. But you clearly feel strongly so please do whatever you'd prefer - it just seems like my suggestion maximises ease of use for a minuscule sacrifice.

PR #754, opened 2018-02-01
Span field in coref model

DeNeutoy:
Adds span fields to the Coref model and and adds a new trained coref model which gets the same performance as the old one. Most of this diff is just regenerating all the test fixtures....

The one annoying thing was that I had to slightly alter the demo sentence, because the new coref model get's a particular part of it wrong (I checked the other demo sentences and some other examples, which are fine).


matt-gardner:
Nice work!

PR #755, opened 2018-02-02
Made get_logical_form handle reverse

matt-gardner:
This looks way bigger than it is because I moved things around, putting the logic and the tests in the base class instead of `NlvrWorld`.  The only relevant code change is in the final loop of `get_logical_form`.


allennlp/data/semparse/worlds/world.py
```diff
@@ -206,6 +209,75 @@ def get_action_sequence(self, expression: Expression) -> List[str]:
...
+                                      self._infer_num_arguments(left_side)))
+            elif len(right_side) > 1 or right_side in ['x', 'y', 'z', '<', '-'] or right_side_is_number:
+                if right_side == "var" and add_var_function:
+                    raise RuntimeError("You wanted to add var, but you didn't need to!")
+                if right_side in ['x', 'y', 'z'] and add_var_function:
+                    right_side = f"(var {right_side})"
+                terminals.append((right_side, self._infer_num_arguments(left_side)))
+        partial_logical_forms: List[str] = []
+        terminals = list(reversed(terminals))
+        for i, (terminal, num_args) in enumerate(terminals):
```
pdasigi: Can you add a comment here describing the special way in which you're handling `reverse` and why?

PR #757, opened 2018-02-02
Add metrics.json output and migrate to run_with_beaker.py

schmmd:
This PR makes three primary changes:

* Adds a `metrics.json` output to AllenNLP.  This output is created whether using Beaker or not, as it's easier to generate as part of AllenNLP rather than post-processing logs and the output is useful whether running in Beaker or as a standalone.

* Nearly outputs the settings in `config.json` as environment variables so they show up in Beaker as parameters.  This gives users the ability to select these parameters when using Beaker's group view.  However, [Beaker's UI looks terrible](http://beaker.allenai.org/ex/ex_n46sgjsv9u26) when several environment variables are output (see [the bug report](https://app.asana.com/0/546081553220630/546132563400408)) so this is disabled presently.

* Migrates `run_with_beaker` from Bash to Python, which was a pre-requisite for the first two changes.


allennlp/commands/train.py
```diff
@@ -203,16 +174,55 @@ def train_model(params: Params, serialization_dir: str, file_friendly_logging: b
...
                     "'evaluate_on_test' flag, or use the 'allennlp evaluate' command.")
 
+    metrics_json = json.dumps(metrics, indent=2)
+    with open(os.path.join(serialization_dir, "metrics.json"), "w") as f:
+        f.write(metrics_json)
+    logger.info("Metrics: " + metrics_json)
+
     return model
+
+def add_arguments(parser: argparse.ArgumentParser):
```
schmmd: I'm not sure this is the best place for this method.  I still don't have a good handle on managing methods in Python.  I use this from the `run_with_beaker.py` script.

allennlp/commands/train.py
```diff
@@ -203,16 +174,55 @@ def train_model(params: Params, serialization_dir: str, file_friendly_logging: b
...
+            metrics["test_" + k] = v
 
     elif test_data:
         logger.info("To evaluate on the test set after training, pass the "
                     "'evaluate_on_test' flag, or use the 'allennlp evaluate' command.")
 
+    metrics_json = json.dumps(metrics, indent=2)
+    with open(os.path.join(serialization_dir, "metrics.json"), "w") as f:
+        f.write(metrics_json)
+    logger.info("Metrics: " + metrics_json)
```
schmmd: I think it's helpful logging output to have all of your stats displayed at the end, but if anyone doesn't want this output here I'll remove it.

allennlp/common/params.py
```diff
@@ -211,6 +211,22 @@ def log_recursively(parameters, history):
         log_recursively(self.params, self.history)
         return self.params
 
+    def as_flat_dict(self, quiet=False):
```
schmmd: I will add a unit test for this.

schmmd:
@joelgrus do you mind taking a first pass at this?

allennlp/common/params.py
schmmd: Done

schmmd:
@tusharkhot @nikett, here's a sample `metrics.json` this PR produces.  Is this a good starting point for what you'd like to see in a `metrics.json`?

```
{
  "training_duration": "00:01:17",
  "training_start_epoch": 0,
  "training_epochs": 2,
  "training_accuracy": 0.4828794553464157,
  "training_loss": 1.022733211517334,
  "validation_accuracy": 0.5303800040642146,
  "validation_loss": 0.9781865477561951
}
```

allennlp/common/params.py
```diff
@@ -211,6 +211,22 @@ def log_recursively(parameters, history):
...
+        Returns the parameters of a flat dictionary from keys to values.
+        Nested structure is collapsed with periods.
+        """
+        flat_params = {}
+        def recurse(parameters, path):
+            for key, value in parameters.items():
+                if isinstance(value, dict):
+                    recurse(value, path + [key])
+                else:
+                    flat_params['.'.join(path + [key])] = value
```
joelgrus: nit: you use the same logic in two places, might be clearer if you just did

```
newpath = path + [key]
```

once

allennlp/training/trainer.py
```diff
@@ -621,7 +621,7 @@ def _validation_loss(self) -> Tuple[float, int]:
 
         return val_loss, batch_num
 
-    def train(self) -> None:
+    def train(self) -> dict:
```
joelgrus: it looks like you could use `Dict[str, float]`?

allennlp/training/trainer.py
```diff
@@ -631,6 +631,9 @@ def train(self) -> None:
 
         logger.info("Beginning training.")
 
+        train_metrics = None
+        val_metrics = None
```
joelgrus: if you use empty dicts as the default values, then you don't have to check for `None` below, iterating over the items will just be a no-op

allennlp/training/trainer.py
```diff
@@ -631,6 +631,9 @@ def train(self) -> None:
...
+        epochs_trained = 0
```
joelgrus: why do you track this explicitly? you have `epoch_counter` and the final value of `epoch`

allennlp/training/trainer.py
```diff
@@ -673,6 +676,23 @@ def train(self) -> None:
                 formatted_time = time.strftime("%H:%M:%S", time.gmtime(estimated_time_remaining))
                 logger.info("Estimated training time remaining: %s", formatted_time)
 
+            epochs_trained += 1
+
+        training_elapsed_time = time.time() - training_start_time
+        metrics = {
+                "training_duration": time.strftime("%H:%M:%S", time.gmtime(training_elapsed_time)),
+                "training_start_epoch": epoch_counter,
+                "training_epochs": epochs_trained
```
joelgrus: additionally, why "training_epochs" here rather than "training_stop_epoch"?

allennlp/training/trainer.py
```diff
@@ -673,6 +676,23 @@ def train(self) -> None:
...
+        }
+        if train_metrics:
```
joelgrus: see above comment, if you use an empty dict to represent no values then you don't have to check this

scripts/ai2-internal/run_with_beaker.py
```diff
@@ -0,0 +1,80 @@
...
+    # Read params and set environment
+    params = Params.from_file(param_file, overrides)
+    flat_params = params.as_flat_dict()
+    env = []
+    for k, v in flat_params.items():
+        k = str(k).replace('.', '_')
+        env.append(f"--env={k}={v}")
+
+    # Get temporary ecr login. For this command to work, you need the python awscli
+    # package with a version more recent than 1.11.91.
```
joelgrus: you could check this:

```
try:
    import awscli
    if awscli.__version__ <= ....
except ModuleNotFoundError:
    ...
```

probably more helpful than just leaving a comment

schmmd: It wasn't my comment (copied over) ;-)

But I'll take the suggestion.

allennlp/training/trainer.py
schmmd: @joelgrus I want a name for the number of epochs trained in the loop.  We might resume training at epoch 2 and end on epoch 5.  In that case `trained_epochs` would be 3.

At the last, I'll add a comment.  Do you have an idea for a better name?

joelgrus: no, the name is fine, I just meant that if you have `training_start_epoch` and `training_stop_epoch` (and you do), then `training_epochs` is superfluous

allennlp/training/trainer.py
schmmd: @joelgrus I had this originally.  Unfortunately `train_metrics` and `val_metrics` can be updated to `None` in the loop.  Do you think I should address this in the upstream methods?  It does seem like undesirable (and confusing) behavior...

allennlp/training/trainer.py
schmmd: Yep--sorry misunderstood.

allennlp/training/trainer.py
joelgrus: I don't see where `train_metrics` can be set to `None`, and it looks like if `val_metrics` were initialized as an empty dict you could remove the code that sets it to `None`

scripts/ai2-internal/run_with_beaker.py
schmmd: I'm not going to do this as it's more work than it's worth since the awscli version is a string with three parts.  Also, this package is in our requirements.txt.

allennlp/training/trainer.py
DeNeutoy: Having these be printed/logged as a side effect of them being returned seems nasty...

allennlp/training/trainer.py
schmmd: https://github.com/schmmd/allennlp/blob/ce154e82c4ed3c2adf07a70fdb386f4f131b6837/allennlp/training/trainer.py#L662

I'll change it to empty dict.

allennlp/training/trainer.py
schmmd: @joelgrus counting explicitly seems like the easiest and most foolproof way to get this count.  Otherwise I need to handle the case when the for loop is empty.

Normally the number of epochs computed is `epoch - epoch_counter + 1`, but if the range is empty then it's 0.

schmmd:
@matt-gardner @DeNeutoy I'd love to hear your thoughts on this PR.

@DeNeutoy I didn't understand your inline comment "Having these be printed/logged as a side effect of them being returned seems nasty..."

allennlp/training/trainer.py
```diff
@@ -631,6 +631,9 @@ def train(self) -> None:
 
         logger.info("Beginning training.")
 
+        train_metrics = {} #type: Dict[str, float]
```
matt-gardner: Instead of commenting with `# type:`, just given the type inline: `train_metrics: Dict[str, float] = {}`.  Same for the line below.

Also, just for future reference, the style we typically use with inline comments like this is at least two spaces before the `#` and one space after.

scripts/ai2-internal/run_with_beaker.py
```diff
@@ -0,0 +1,80 @@
...
+            "-m",
+            "allennlp.run",
+            "train",
+            "/config.json",
+            "-s",
+            "/output",
+            "--file-friendly-logging"
+        ]
+
+    # TODO(michaels): add back in the env list.
```
matt-gardner: I think the main task page actually looks fine, it's only the experiment page that looks like a mess.

scripts/ai2-internal/run_with_beaker.py
```diff
@@ -0,0 +1,80 @@
...
+            '--detach'] + [image] + allennlp_command
+    print(' '.join(command))
+    subprocess.run(command, check=True)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+
+    # TODO(michaels): rename to param-file
+    parser.add_argument('params', type=str, help='The model configuration file.')
+    parser.add_argument('description', type=str, help='A description for the experiment.')
```
matt-gardner: Can you add a `--sources` flag, or similar, so we can specify which beaker datasets we depend on and where we want them mounted?  I'm not sure what the best way to do this is, and it's fine with me if you want to do this in a subsequent PR.

tests/common/params_test.py
```diff
@@ -29,6 +29,13 @@ def test_overrides(self):
         assert model_params.pop("type") == "BAR"
         assert model_params["text_field_embedder.tokens.type"] == "BAZ"
 
+    def test_as_flast_dict(self):
```
matt-gardner: `s/as_flast_dict/as_flat_dict/`

tests/common/params_test.py
```diff
@@ -29,6 +29,13 @@ def test_overrides(self):
...
+        filename = 'tests/fixtures/bidaf/experiment.json'
+        params = Params.from_file(filename)
```
matt-gardner: It doesn't look like you're actually testing `as_flat_dict` here.  You're just getting the default `HOCON` behavior.  If you really want to test `as_flat_dict`, also, I'd recommend using a smaller parameter object and testing it exhaustively, instead of just checking for existence of a few keys - the test should also check that there are no _extra_ keys.

DeNeutoy:
I commented before you'd actually logged the metrics in `bf9d9af`, but they would have been printed as a side effect of evaluating `Trainer.train()`. No problem now.

scripts/ai2-internal/run_with_beaker.py
schmmd: Beaker is also doing a redesign that fixes this issue.  I want to follow up with Niket and Tushar to see if they're OK with the trade-off for the short term.

allennlp/training/trainer.py
schmmd: Thanks--I wasn't sure whether that was valid syntax and I found the annotation I used in Google.

On your second note, I'd love if we could programmatically enforce all the style we care about (and ignore the rest).

scripts/ai2-internal/run_with_beaker.py
schmmd: Yes, I have a TODO about this (I'll probably tackle it in a subsequent PR).

schmmd:
@DeNeutoy yeah sorry, I think I accidentally nuked printing out the metrics as part of a rebase last Friday.

allennlp/training/trainer.py
matt-gardner: I imagine some aspects of style are impossible to enforce programmatically.  This one seems feasible, though, and it would be great to add to our linter.  Not sure if pylint or any other linter checks for this.

tests/common/params_test.py
schmmd: That's funny--I added a call to `as_flat_dict`.  I agree the test you outlined would be more thorough.

tests/common/params_test.py
```diff
@@ -31,11 +31,22 @@ def test_overrides(self):
...
+        assert params["a"] == 10
+
+        assert "b.c" in params
+        assert params["b.c"] == 20
 
-        assert "dataset_reader.type" in params
-        assert "dataset_reader.token_indexers.tokens.lowercase_tokens" in params
-        assert "model.text_field_embedder.tokens.type" in params
+        assert "b.d" in params
+        assert params["b.d"] == 'stuff'
```
matt-gardner: This test still doesn't quite do what I suggested.  You really should just assert dictionary equality:

```python
assert params == {'a': 10, 'b.c': 20, 'b.d': 'stuff'}
```

schmmd: sigh--you're right.  Done.

PR #767, opened 2018-02-03
ELMo tuning

matt-peters:
* allow ELMo to be optionally tuned
* add support for parameter groups when constructing the optimizers


allennlp/modules/elmo.py
```diff
@@ -54,11 +53,12 @@ def __init__(self,
                  options_file: str,
                  weight_file: str,
                  num_output_representations: int,
```
DeNeutoy: Add to the docstring.

allennlp/modules/elmo.py
```diff
@@ -172,14 +174,16 @@ class _ElmoCharacterEncoder(torch.nn.Module):
     """
     def __init__(self,
                  options_file: str,
-                 weight_file: str) -> None:
+                 weight_file: str,
+                 requires_grad: bool = False) -> None:
```
DeNeutoy: Add to the docstring.

allennlp/modules/elmo.py
```diff
@@ -390,10 +386,11 @@ class _ElmoBiLm(torch.nn.Module):
...
```
DeNeutoy: Add to the docstring.

allennlp/modules/elmo_lstm.py
```diff
@@ -42,6 +42,8 @@ class ElmoLstm(_EncoderBase):
         :class:`~allennlp.modules.lstm_cell_with_projection.LstmCellWithProjection`.
     num_layers : ``int``, required
         The number of bidirectional LSTMs to use.
+    require_grad: ``bool``, optional
+        Compute gradients of model parameters?
```
DeNeutoy: "If `True`, computes gradient of ELMo parameters, which is useful for fine tuning" or something.

allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -30,14 +30,16 @@ def __init__(self,
                  options_file: str,
                  weight_file: str,
                  do_layer_norm: bool = False,
-                 dropout: float = 0.5) -> None:
+                 dropout: float = 0.5,
+                 requires_grad: bool = False) -> None:
```
DeNeutoy: Add to the docstring.

allennlp/training/optimizers.py
```diff
@@ -26,13 +27,50 @@ class Optimizer(Registrable):
...
+            # each dict contains {'params': [list of parameters], 'lr': 1e-3, ...}
+            # see: http://pytorch.org/docs/0.3.0/optim.html?#per-parameter-options
+            #
+            # groups contains something like:
+            #"parameter_groups": [
+            #       [['regex1', 'regex2'], {'lr': 1e-3},
+            #        ['regex3'], {'lr': 1e-4}]
+            #]
+            # last entry is for the parameters not in any regex
+            parameter_groups: Any = [{'params': []} for _ in range(len(groups) + 1)]
```
DeNeutoy: I think this type should be `List[Dict[str, List[torch.nn.Parameter]]]`. If there was a reason it had to be `Any`, can you say why (using `Any` as a type removes all of the benefit we get from using MyPy)

allennlp/training/optimizers.py
```diff
@@ -26,13 +27,50 @@ class Optimizer(Registrable):
...
+            # input to optimizer is list of dict
+            # each dict contains {'params': [list of parameters], 'lr': 1e-3, ...}
+            # see: http://pytorch.org/docs/0.3.0/optim.html?#per-parameter-options
+            #
+            # groups contains something like:
+            #"parameter_groups": [
+            #       [['regex1', 'regex2'], {'lr': 1e-3},
+            #        ['regex3'], {'lr': 1e-4}]
+            #]
+            # last entry is for the parameters not in any regex
```
DeNeutoy: How are config options passed for the default group?

DeNeutoy: Oh I see they are just from the top level config. Can you mention that in a comment somewhere?

allennlp/training/optimizers.py
```diff
@@ -26,13 +27,50 @@ class Optimizer(Registrable):
...
+            # add the group specific kwargs
+            for k in range(len(groups)): # pylint: disable=consider-using-enumerate
+                parameter_groups[k].update(groups[k][1].as_dict())
+
+            for name, param in model_parameters:
+                # Determine the group for this parameter.
+                group_index = None
+                for k, group_regexes in enumerate(groups):
+                    for regex in group_regexes[0]:
+                        if re.search(regex, name):
```
DeNeutoy: Can you replicate the type of logging that we have for parameter regexes for regularisation? It's very helpful, particularly when you change your model, because you can just see what's changed. 

https://github.com/allenai/allennlp/blob/master/allennlp/nn/initializers.py#L223

Particularly, I would like it if each parameter logged which parameter group it was in, and warnings were raised if you pass parameter regexes which do not match any parameters, because this probably means there was a mis-specified config.

allennlp/training/optimizers.py
matt-peters: All of the key/value pairs except `params` are used to set the kwargs for the parameter group (`lr`, etc), so that the type of the `Dict` values can be anything accepted by the optimizer constructor.  This is done in the statement `parameter_groups[k].update(groups[k][1].as_dict())` a few lines down.  So I used the `Any` type to avoid enumerating all possible types.

DeNeutoy: Ah gotcha.

PR #773, opened 2018-02-05
make all dataset readers (optionally) lazy

joelgrus:
allows user to specify `lazy: true` in the config, doesn't change any default behaviors


joelgrus:
ok, it turns out you can parametrize tests:

https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize

the only challenge is that in that case the test classes can't inherit from `TestCase`, but in the dataset readers we were never using the `AllenNLPTestCase` functionality, so I just took that base class out, and the result should address your concern

PR #775, opened 2018-02-05
Domain filter

DeNeutoy:
A little change to optionally select a single domain, which I need for the domain transfer experiments.


allennlp/data/dataset_readers/semantic_role_labeling.py
```diff
@@ -52,7 +59,7 @@ def _read(self, file_path: str):
         ontonotes_reader = Ontonotes()
         logger.info("Reading SRL instances from dataset files at: %s", file_path)
```
joelgrus: maybe update this logging to include the domain_identifier?

plroit:
Can you please replace the slash in `f"/{domain_identifier}/"` path separator with the more generic os.path.sep for Windows users? `f"{os.path.sep}{domain_identifier}/{os.path.sep}"`

matt-gardner:
@plroit, we don't use Windows, and tend not to think about problems for Windows users. We are a very small team, and we don't have the resources to worry about this ourselves. If you want to submit a PR fixing this, though, we'd be happy to accept it.

PR #776, opened 2018-02-05
A couple of small fixups for verify.sh.

schmmd:


schmmd:
A few dumb mistakes I should have caught the first time around...

PR #778, opened 2018-02-05
laziness tutorial

joelgrus:
this ended up being kind of tricky to write, let me know if you can think of ways to improve it


tutorials/getting_started/laziness.md
```diff
@@ -0,0 +1,233 @@
...
+for epoch in range(10):
+    for instance in instances:
+        process(instance)
+```
+
+Then each epoch's `for instance in instances` results in a *new* call
+to `MyDatasetReader._read()`, and your instances will be read from disk
+10 times.
+
+In this case, you should make sure that your `_read()` method
```
matt-gardner: I'm not sure what the right way is to talk about this, but if you're sampling things at each call to `_read()`, you actually _do_ want to read the whole dataset into memory, but only on the _first_ call to `_read()`.  Then, after reading everything that you want, all calls to `_read()` do the sampling on the instances in memory, and it's different at each epoch.  This is an advanced usage that you might not want to cover here, but I'd at least avoid contradicting this use case in the simple tutorial.

tutorials/getting_started/laziness.md
```diff
@@ -0,0 +1,233 @@
...
+
+This is especially useful if you are using the `BucketIterator`,
+which can order your instances by e.g. sentence length. If you don't specify
+`max_instances_in_memory`, you'd just get each batch of 32 ordered by sentence length,
+which isn't particularly helpful.
+
+The other option here is `instances_per_epoch`. By default, each epoch is a single
+pass through the dataset. However, if you had millions of instances,
+you might want each epoch to consist of a fraction of the dataset,
+in which case you can specify the value here. But most likely you don't need
```
matt-gardner: Specifying this will also result in nicer tqdm logging.

tutorials/getting_started/laziness.md
```diff
@@ -0,0 +1,233 @@
...
+That list is loaded into memory and (because it's a list) can be iterated over repeatedly
+after the initial call to `.read()`.
+
+The more interesting case is when the dataset reader was instantiated with `lazy=True`.
+In that case we return a `_LazyInstances` initialized with a lambda function
+that calls `_read()`. `_LazyInstances` is just a simple class wrapper to produce
+lazy iterables (again, I stripped it down to just its essence):
+
+```python
+class _LazyInstances(Iterable):
```
DeNeutoy: This is pretty detailed and involves a private AllenNLP class. I'm not sure the details really belong here?

DeNeutoy: * This level of detail

matt-gardner: On the other hand, this is a tutorial, not a how-to, and the goal is to teach the reader about how the library works.  We _definitely_ want to mention in here that you'll get multiple calls to `_read()` with `lazy=True`; I could go either way on whether to include this code to demonstrate that.

tutorials/getting_started/laziness.md
joelgrus: (1) why would you implement sampling in the DatasetReader rather than as a `SamplingIterator` or something?
(2) doesn't this depend on how you sample? you could take each instance with a fixed probability without loading the whole dataset, or you could do some kind of reservoir sampling.
(3) all that said, I guess I can imagine cases where you would want to sample in the DatasetReader, so let me think about how to word that

matt-gardner: The case I'm thinking of is that you have `k`paragraphs associated with each instance, and at each epoch you want to sample 2 of them. You read in all `k`, cache that, then pick two at each call to`_read`. You can't easily do this kind of thing in our `DataIterator` - you're already dealing with`Instances`by that point.

And by the way, this is how I recommend doing a `DatasetReader` for Chris's model.

matt-gardner:
Also, I just realized we need to update the README for the tutorials, that has an index of them.

tutorials/getting_started/laziness.md
joelgrus: the way things are designed right now that's a problem unless your train / validation / test sets are the same size.

i.e. if you have 1000 train examples / 200 validation examples, and if you specify `instances_per_epoch: 1000`, then when you iterate over the validation set it will make 5 passes over the data

this is why I was earlier suggesting making the dataset parts of the configuration slightly richer -- it seems likely you'd want to specify different values for things like this for different datasets

joelgrus: or another alternative is just to count the instances the first pass through the dataset and remember the number for future passes

matt-gardner: Oh, yeah, you're right, I forgot about that.  Mandar said he was going to submit a PR to allow for different train and validation iterators, which would solve this problem.

joelgrus:
ok, I think I addressed all the concerns (except for Mark's concerns about talking about the details of a private class, I didn't address those 😇 ), I also updated the README

tutorials/how_to/laziness.md
```diff
@@ -0,0 +1,31 @@
+# How to Train a Model with Lazy Data
+
+(Here, by "lazy data", I mean a dataset that's generated
+ instance-by-instance each time you iterate over it,
+ possibly because it's too big to fit in memory,
+ possibly because you want to sample your data each iteration,
+ possibly because you want to start training immediately,
+ possibly for some other reason I haven't thought of.)
+
+There are basically four things you should do. This tutorial will tell you
```
matt-gardner: `s/tutorial/how-to/`?

tutorials/how_to/laziness.md
```diff
@@ -0,0 +1,31 @@
...
+[detailed tutorial about laziness in AllenNLP](../getting_started/laziness.md).
+
+## In the `YourDatasetReader` constructor
+
+`YourDatasetReader` subclass should have a `lazy` parameter in its constructor
+and `from_params` method, and pass that value to its superclass constructor.
+
+## In `YourDatasetReader._read()`
+
+`YourDatasetReader._read()` should return a generator, not a list.
```
matt-gardner: Is it possible to link to particular code blocks in the other tutorial for `generator` and `list`, in case someone doesn't know what this means?  Worst case, you could just link to the same thing for `generator` that you did in the tutorial.

tutorials/how_to/laziness.md
```diff
@@ -0,0 +1,31 @@
...
+
+## In the `dataset_reader` section of your experiment configuration
+
+You should specify `'lazy': true` in the `dataset_reader` section of your experiment config.
+
+## In the `iterator` section of your experiment configuration
+
+You should specify `'max_instances_in_memory'` as large as you can
+in the `iterator` section of your experiment config.
+(You don't _have_ to do this, but it's a good idea.)
```
matt-gardner: I'd add "particularly if you want to use a bucket iterator".

PR #781, opened 2018-02-05
Add an Elmo command

schmmd:
This PR maintains the same functionality as `scripts/write_elmo_representations_to_file.py` but moves it into an `elmo` command.  The advantages are that it's more discoverable, testable, and lint-able as a core component of AllenNLP.

I also refactored the code to expose this functionality to anyone using AllenNLP as a library.  For example:

```
from allennlp.commands.elmo import ElmoEmbedder

ee = ElmoEmbedder()

ee.embed_sentence("Michael went to the store to buy some eggs .".split())
```

I'll follow on with the suggestions in https://github.com/allenai/allennlp/issues/785 in future PR's.


schmmd:
@matt-gardner @DeNeutoy is this what you were thinking for an `python -m allennlp.run elmo` command?  Please don't nit CR as this isn't ready for review yet and is mostly just copied over from https://github.com/allenai/allennlp/blob/master/scripts/write_elmo_representations_to_file.py.  Also, this doesn't add any of the CLI options we discussed last Tuesday.

I may move the logic into a class `ElmoVectorizer` rather than keeping it in `allennlp/commands/Elmo`.

matt-gardner:
Yes, this looks pretty much like what I was thinking.  Mark will probably have more input here than me.  The one thing I noticed as I was looking through this, that you can fix later, is that options should use all hyphens, not underscores (and I know that was from the copied code, just FYI).

DeNeutoy:
Yep looks broadly what I was thinking. It should also be able to generate the uncontextual elmo representations that https://github.com/allenai/allennlp/blob/master/scripts/create_elmo_embeddings_from_vocab.py
this script creates, as that's another possible use case.

allennlp/commands/elmo.py
```diff
@@ -0,0 +1,203 @@
...
+        subparser.add_argument('--batch-size', type=int, default=64, help='The batch size to use.')
+        subparser.add_argument('--cuda-device', type=int, default=-1, help='The cuda_device to run on.')
+        subparser.add_argument('--use-sentence-key', default=False, action='store_true')
+
+        subparser.set_defaults(func=elmo_command)
+
+        return subparser
+
+
+class ElmoEmbedder():
```
schmmd: Our convention seems to be one class per file.  Where should I move this?

joelgrus: do we expect that anyone besides this command would use this class? if so, put it where they'd "expect" it to be. if not, IMO it's fine to leave it here (and maybe make it "private")

schmmd: This class would be the easiest way for someone to get ELMo vectors for sentences programatically--so yes, I think I do.

schmmd:
Just remembered I need to update https://github.com/allenai/allennlp/blob/master/tutorials/how_to/elmo.md

allennlp/commands/elmo.py
```diff
@@ -0,0 +1,203 @@
+"""
+The ``elmo`` subcommand allows you to make bulk ELMo predictions.
```
DeNeutoy: Can you add:

1) A TODO to add the link to the paper.
2) The output of `python -m allennlp.run elmo --help`, as with the docstrings of the other commands.

schmmd:
@matt-peters I get the same output as the original script and it seems reasonable when I spot-check the distance between some word vectors, but how would you assess whether it's creating the "right" vectors for ELMo?

allennlp/commands/elmo.py
```diff
@@ -0,0 +1,203 @@
...
+
+Given a pre-processed input text file, this command outputs the internal
+layers used to compute ELMo representations to a single (potentially large) file.
+
+The input file is previously tokenized, whitespace separated text, one sentence per line.
+The output is a hdf5 file (http://docs.h5py.org/en/latest/) where each
```
DeNeutoy: This URL needs to be formatted like this: https://github.com/allenai/allennlp/blob/master/allennlp/models/semantic_role_labeler.py#L25

allennlp/commands/elmo.py
```diff
@@ -0,0 +1,203 @@
...
+sentence is a size (3, num_tokens, 1024) array with the biLM representations.
+
+In the default setting, each sentence is keyed in the output file by the line number
+in the original text file.  Optionally, by specifying --use_sentence_key
```
DeNeutoy: `--use-sentence-key`

allennlp/commands/elmo.py
```diff
@@ -0,0 +1,203 @@
...
+
+        logger.info("Initializing ELMo.")
+        self.elmo_bilm = _ElmoBiLm(options_file, weight_file)
+        if cuda_device >= 0:
+            self.elmo_bilm = self.elmo_bilm.cuda(cuda_device=cuda_device)
+
+        self.cuda_device = cuda_device
+
+    def batch_to_ids(self, batch: List[List[str]]) -> torch.Tensor:
+        """
```
DeNeutoy: Can you use the Sphinx docstrings throughout this class? It needs to be:

``` 
Parameters
-----------
name : type
    description

Returns
--------

....
```

allennlp/commands/elmo.py
```diff
@@ -0,0 +1,203 @@
...
+        with a organized by key.  Unless use_sentence_key is set, the key will be the index of the sentence.
+        :param input_file: a file with one tokenized sentence per line
+        :param output_file_path: a path to the output hdf5 file
+        :param batch_size: the number of sentences to process in ELMo at one time
+        :param use_sentence_key: if true, use the first token in each sentence as the key
+        """
+
+        # Tokenizes the sentences.
+        sentences = [line.strip().split() for line in input_file]
+        if use_sentence_key:
```
DeNeutoy: Hmm, when we discussed this I meant to use the entire sentence string as the key, but this is also fine. However, your wording here and in the docstring is confusing - it could be made clearer that the first token must be a unique identifier (rather than just the first token in the sentence), and then calling it `first_token_key` or something might be better.

allennlp/commands/elmo.py
```diff
@@ -0,0 +1,203 @@
...
+        return elmo_embeddings
+
+    def embed_sentences(self, sentences: Iterable[List[str]], batch_size: int) -> Iterable[torch.Tensor]:
+        """
+        Computes the ELMo embeddings for a iterable of sentences.
+        :param sentences: an iterable of tokenized sentences
+        :param batch_size: the number of sentences ELMo should process at once
+        :return: a list of tensors, each representing the ELMo vectors for the input sentence at the same index
+        """
+        for batch in lazy_groups_of(iter(sentences), batch_size):
```
DeNeutoy: Can you add a tqdm wrapper to this? It would be nice to see the updates, because it could take a little while.

schmmd:
Yes--I will add tests.

allennlp/commands/elmo.py
schmmd: Oops--sorry I had PyCharm create these and didn't realize it was different.  Will do.

allennlp/commands/elmo.py
schmmd: @DeNeutoy I have not implemented what we discussed.  This is the same behavior as in the original script.  I may change this in a future PR.

I will update the comments.

allennlp/commands/elmo.py
schmmd: I'd rather not.  I would prefer to leave it up to the caller to wrap the returned iterator.  Otherwise this method forces a tqdm that is not configurable.

As an example of a caller wrapping the iterator in a tqdm see `embed_file`.

matt-peters:
Those test fixtures are the best way to check for correctness.  There is some reference code here to load the sentences and run the model:  https://github.com/allenai/allennlp/blob/master/tests/modules/elmo_test.py#L24

Importantly, you will need use batch_size=3 to get the output to match (due to low level details of how the fixtures were created).

allennlp/commands/elmo.py
schmmd: I struggled with adding the output as a docstring, but I couldn't get it to work.  In fact, I couldn't even copy the code block from the predict docstring--make docs kept failing.

```
  File "/Users/michaels/miniconda3/envs/allennlp/lib/python3.6/site-packages/sphinx/cmdline.py", line 296, in main
    app.build(opts.force_all, filenames)
  File "/Users/michaels/miniconda3/envs/allennlp/lib/python3.6/site-packages/sphinx/application.py", line 333, in build
    self.builder.build_update()
  File "/Users/michaels/miniconda3/envs/allennlp/lib/python3.6/site-packages/sphinx/builders/__init__.py", line 251, in build_update
    'out of date' % len(to_build))
  File "/Users/michaels/miniconda3/envs/allennlp/lib/python3.6/site-packages/sphinx/builders/__init__.py", line 268, in build
    self.warn(*warning, **kwargs)
  File "/Users/michaels/miniconda3/envs/allennlp/lib/python3.6/site-packages/sphinx/application.py", line 399, in warn
    raise SphinxWarning(warntext)
sphinx.errors.SphinxWarning: /Users/michaels/hack/allenai/allennlp/allennlp/commands/elmo.py:docstring of allennlp.commands.elmo:1: ERROR: Error in "code-block" directive:
maximum 1 argument(s) allowed, 27 supplied.

.. code-block:: bash
   $ python -m allennlp.run elmo --help
   usage: python -m allennlp.run [command] elmo [-h] [--vocab-path VOCAB_PATH]
                                                [--options-file OPTIONS_FILE]
                                                [--weight-file WEIGHT_FILE]
                                                [--batch-size BATCH_SIZE]
                                                [--cuda-device CUDA_DEVICE]
                                                [--use-sentence-key]
                                                input_file output_file

   Create word vectors using ELMo.

   positional arguments:
     input_file            path to input file
     output_file           path to output file

   optional arguments:
     --vocab-path VOCAB_PATH
                           A path to a vocabulary file to generate
     --options-file OPTIONS_FILE
                           The path to the ELMo options file.
     --weight-file WEIGHT_FILE
                           The path to the ELMo weight file.
     --batch-size BATCH_SIZE
                           The batch size to use.
     --cuda-device CUDA_DEVICE
                           The cuda_device to run on.
     --use-sentence-key
```

matt-gardner: Pretty sure you need a blank line after the `.. code-block:: bash` line.  Sphinx is really finicky...

DeNeutoy:
Nit - can you use full sentences with capitals and full-stops to match the rest of the comments/docs in the repo?

allennlp/commands/elmo.py
schmmd: That did it--although I swear I tried that previously and it didn't work.

schmmd:
I capitalized and added periods, but my comments are still not full sentences, which I feel quite weird about.  But we do the same elsewhere in the codebase:

* https://github.com/allenai/allennlp/blob/master/allennlp/models/archival.py#L46
* https://github.com/allenai/allennlp/blob/master/allennlp/models/semantic_role_labeler.py#L233

Do care about this for inline comments?  If so, would you please add something to https://github.com/allenai/allennlp/blob/master/STYLE.md?  I think I've heard from @matt-gardner that he does not like comments in the imperative--but I get confused about this too as our codebase has frequent comments in the imperative.

From a random file:

* https://github.com/allenai/allennlp/blob/master/tests/training/trainer_test.py#L212
* https://github.com/allenai/allennlp/blob/master/tests/training/trainer_test.py#L216

Personally, I would prefer not caring how inline comments are formatted, as long as they are intelligible.

matt-gardner:
I write comments in full sentences and avoid imperatives - it's part of the google style guides linked in `STYLE.md`.  But I try not to nitpick about it, unless they're particularly unclear, or I'm reviewing a PR for someone I'm mentoring.

allennlp/commands/elmo.py
```diff
@@ -0,0 +1,203 @@
...
+from allennlp.modules.elmo import _ElmoBiLm
+from allennlp.commands.subcommand import Subcommand
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+DEFAULT_OPTIONS_FILE = "https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/2x4096_512_2048cnn_2xhighway/elmo_2x4096_512_2048cnn_2xhighway_options.json" # pylint: disable=line-too-long
+DEFAULT_WEIGHT_FILE = "https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/2x4096_512_2048cnn_2xhighway/elmo_2x4096_512_2048cnn_2xhighway_weights.hdf5" # pylint: disable=line-too-long
+
+
+class Elmo(Subcommand):
```
matt-peters: We already have an `elmo.py` and `Elmo` class in `allennlp/modules/elmo.py`.  Can you rename this class to something else to avoid confusion?

tutorials/how_to/elmo.md
```diff
@@ -149,5 +149,14 @@ representations = elmo(character_ids)
 
 ## Writing contextual representations to disk
```
matt-peters: Is it possible to replace any of the code in the usage example in this tutorial (under `Using ELMo programmatically`) with a link to the new `ElmoEmbedder` class?  Much of the functionality is duplicated, e.g. the `batch_to_ids` function now exits as a method on that class.

allennlp/commands/elmo.py
```diff
@@ -0,0 +1,292 @@
...
+from allennlp.commands.subcommand import Subcommand
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+DEFAULT_OPTIONS_FILE = "https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/2x4096_512_2048cnn_2xhighway/elmo_2x4096_512_2048cnn_2xhighway_options.json" # pylint: disable=line-too-long
+DEFAULT_WEIGHT_FILE = "https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/2x4096_512_2048cnn_2xhighway/elmo_2x4096_512_2048cnn_2xhighway_weights.hdf5" # pylint: disable=line-too-long
+DEFAULT_BATCH_SIZE = 64
+
+
+class Elmo(Subcommand):
```
matt-peters: We already have an `Elmo` class in an `elmo.py` in `allennlp/modules/elmo.py`.  Can you rename this class (and potentially file) to something else to avoid confusion?

tutorials/how_to/elmo.md
schmmd: Cool--I did not know that.  I just copied it from the script.

allennlp/commands/elmo.py
schmmd: I've found it annoying too to have two `elmo.py`s, but we seem to follow this example elsewhere and so I followed it.

```
$ find . -name bidaf.py                                                                                                                                                                                                                          
./allennlp/models/reading_comprehension/bidaf.py
./allennlp/service/predictors/bidaf.py
./build/lib/allennlp/models/reading_comprehension/bidaf.py
./build/lib/allennlp/service/predictors/bidaf.py
```

allennlp/commands/elmo.py
schmmd: Done.

allennlp/commands/elmo.py
schmmd: Done.

tutorials/how_to/elmo.md
schmmd: Done.

schmmd:
I still need to add tests for `ElmoEmbedder`.

schmmd:
@DeNeutoy I added an end-to-end test for `ElmoEmbedder`.

tests/modules/elmo_test.py
```diff
@@ -97,12 +103,12 @@ def _load_sentences_embeddings(self):
...
 
 
-class TestElmo(AllenNlpTestCase):
+class TestElmo(ElmoTestCase):
     def setUp(self):
         super(TestElmo, self).setUp()
 
-        options_file = os.path.join(FIXTURES, 'options.json')
-        weight_file = os.path.join(FIXTURES, 'lm_weights.hdf5')
+        options_file = os.path.join(self.fixtures_path, 'options.json')
```
DeNeutoy: You could just use the attributes set in the super class here.

schmmd: @DeNeutoy I don't understand.  Can you give an example?

DeNeutoy: Oh just that you're inheriting from `ElmoTestCase` and in `ElmoTestCase.setUp` you set these [variables as attributes](https://github.com/allenai/allennlp/pull/781/files/a2b27a598e1ffd6b8d6a53e765810600d87f1d30#diff-6d5c0b791d43299ce1afe2eda864ec96R25). So instead of re-defining them, you could just use them directly via `self.options_file` etc.

matt-peters:
It looks like you checked in a new `expected_embeddings.hdf5` file for the ElmoEmbedder class.  We already have these files checked into the fixtures directory (e.g. `lm_embeddings_0.hdf5`, https://github.com/allenai/allennlp/tree/master/tests/fixtures/elmo).  These are the gold standard for those sentences as they came from the original training code for the biLM.  What do you think about just retaining one version of the expected embeddings in the fixtures and checking your test case against those?  It will avoid potential confusion and possible silent errors (e.g. did you confirm that the `expected_embeddings.hdf5` file contains the same embeddings as the others)?

tests/modules/elmo_test.py
schmmd: Right, thanks.  That's just something I overlooked when I created `ElmoTestCase`.

tests/commands/elmo_test.py
```diff
@@ -0,0 +1,144 @@
...
+            assert list(h5py_file.keys()) == ["A", "B"]
+            # The vectors in the test configuration are smaller (32 length)
+            assert h5py_file.get("A").shape == (3, len(sentences[0].split()) - 1, 32)
+            assert h5py_file.get("B").shape == (3, len(sentences[1].split()) - 1, 32)
+
+
+class TestElmoEmbedder(ElmoTestCase):
+    def test_embeddings_are_as_expected(self):
+        """
+        You can recreate the expected embeddings with the following command:
```
matt-peters: This doc string should now be removed.

tests/modules/elmo_test.py
```diff
@@ -17,26 +17,68 @@
...
+
+class ElmoTestCase(AllenNlpTestCase):
+    def setUp(self):
+        super(ElmoTestCase, self).setUp()
+        self.fixtures_path = os.path.join('tests', 'fixtures', 'elmo')
+        self.options_file = os.path.join(self.fixtures_path, 'options.json')
+        self.weight_file = os.path.join(self.fixtures_path, 'lm_weights.hdf5')
+        self.sentences_json_file = os.path.join(self.fixtures_path, 'sentences.json')
+        self.sentences_txt_file = os.path.join(self.fixtures_path, 'sentences.txt')
+        self.expected_embeddings_file = os.path.join(self.fixtures_path, 'expected_embeddings.hdf5')
```
matt-peters: This line need to be removed too.

tests/modules/elmo_test.py
```diff
@@ -17,26 +17,68 @@
...
-        weight_file = os.path.join(FIXTURES, 'lm_weights.hdf5')
-        elmo_bilm = _ElmoBiLm(options_file, weight_file)
+        elmo_bilm = _ElmoBiLm(self.options_file, self.weight_file)
 
         # Deal with the data.
         indexer = ELMoTokenCharactersIndexer()
 
         # For each sentence, first create a TextField, then create an instance
         instances = []
+        fout = open("sentences.txt.new", "w")
```
matt-peters: Why is it necessary to write to a file here?

PR #794, opened 2018-02-08
Add wikitables parser to the demo

matt-gardner:
(Just on the wikitables branch, not intended for updating the live demo yet.)

This PR adds the necessary bits to get a live demo running for the wikitables parser.  Most of the python changes are around adding stuff to the output, and keeping track of some debugging information during decoding.  A bit ugly to have to do it this way, but it works.  We can (hopefully) make that better when we clean up the whole `DecoderState` business that's currently a mess, anyway.


training_config/wikitables_parser.json
```diff
@@ -0,0 +1,77 @@
+{
+  "dataset_reader": {
+    "type": "wikitables",
+    "lazy": false,
+    "tables_directory": "/wikitables/",
+    "dpd_output_directory": "/wikitables/dpd_output/"
+  },
+  "vocabulary": {
+    //"directory_path": "/wikitables_preprocessed/vocab"
```
pdasigi: Remove comment?

PR #795, opened 2018-02-08
Wikitables evaluation script

matt-gardner:
This script takes a wikitables example file, predicts logical forms for all of the instances, then calls out to sempre to evaluate the logical forms for denotation accuracy.  We call to sempre using a jar built from [this repo](https://github.com/allenai/wikitables-executor).  I put the jar on s3, and download it in the script, so this script should be able to just run.

This script depends on the updates in #794 to run (because that PR adds logical form output to the parser), but other than that it's standalone and can be reviewed now.

### DIFF
--- a/scripts/wikitables_eval.py
+++ b/scripts/wikitables_eval.py
@@ -32,7 +32,8 @@ def predict_logical_forms(args: argparse.Namespace):
     archive = load_archive(args.model_archive, cuda_device=args.cuda_device)
     model = archive.model
     model.eval()
-    # TODO(mattg): TOTAL HACK! Not sure why I'm getting this - might be due to numbers?
+    # TODO(mattg): TOTAL HACK! Not sure why I'm getting UNK tokens when reading production rules,
+    # but I am - might be due to numbers?
     model.vocab._token_to_index['rule_labels']['@@UNKNOWN@@'] = 0
     config = archive.config
     config['dataset_reader']['lazy'] = True
--- a/scripts/wikitables_eval.py
+++ b/scripts/wikitables_eval.py
@@ -1,0 +1,98 @@
+import argparse
+import logging
+import os
+import shutil
+import sys
+from subprocess import run
+
+import tqdm
+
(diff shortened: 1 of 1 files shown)


scripts/wikitables_eval.py
```diff
@@ -0,0 +1,98 @@
...
+    keep_data_dir = False
+    if os.path.exists('data'):
+        if not os.path.isdir('data'):
+            raise RuntimeError("A file named 'data' exists in the current directory; can't proceed")
+        else:
+            keep_data_dir = True
+    os.makedirs('data', exist_ok=True)
+    if not os.path.exists('data/abbreviations.tsv'):
+        run(f'wget {ABBREVIATIONS_FILE}', shell=True)
+        run('mv wikitables-abbreviations.tsv data/abbreviations.tsv', shell=True)
```
pdasigi: Looks like you're using allenai's fork of sempre (the original one doesn't need this file).
How hard is it to pull the latest code from Percy's fork and build a jar automatically in this script?

pdasigi: Never mind, pulling the latest changes may not be a great idea because we would want the executor fixed.

scripts/wikitables_eval.py
```diff
@@ -0,0 +1,98 @@
...
+    predict_logical_forms(args)
+    evaluate_logical_forms(args)
+
+
+def predict_logical_forms(args: argparse.Namespace):
+    os.makedirs(args.output_dir, exist_ok=True)
+    archive = load_archive(args.model_archive, cuda_device=args.cuda_device)
+    model = archive.model
+    model.eval()
+    # TODO(mattg): TOTAL HACK! Not sure why I'm getting this - might be due to numbers?
```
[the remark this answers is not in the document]
matt-gardner: I added a bit more detail to the comment.

scripts/wikitables_eval.py
matt-gardner: I was using the code that PNP is using.  I'd be very wary of pulling dependencies and building a jar in the script, as that'd be very unstable.

PR #797, opened 2018-02-08
Remove the --all flag and make --checks optional.

schmmd:
My original design was overly complicated and a bit obnoxious to use.  I'm pretty sure I'm the only one using this script presently.

### DIFF
--- a/scripts/verify.py
+++ b/scripts/verify.py
@@ -45,7 +45,7 @@ if __name__ == "__main__":
     checks = ['pytest', 'pylint', 'mypy', 'build-docs', 'check-docs']
 
     parser = argparse.ArgumentParser()
-    parser.add_argument('--checks', type=str, required=False, nargs='+', choices = checks)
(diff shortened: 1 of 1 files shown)


scripts/verify.py
```diff
@@ -45,16 +45,13 @@ def main(checks):
     checks = ['pytest', 'pylint', 'mypy', 'build-docs', 'check-docs']
 
     parser = argparse.ArgumentParser()
-
-    group = parser.add_mutually_exclusive_group(required=True)
-    group.add_argument('--all', action='store_true', help='Run all tests.')
-    group.add_argument('--checks', type=str, nargs='+', choices = checks)
+    parser.add_argument('--checks', type=str, required=False, nargs='+', choices = checks)
```
joelgrus: 2 nits:

(1) spaces around `choices = checks`
(2) isn't `required=False` the default for `--` arguments?

PR #799, opened 2018-02-08
Use the sentence as the key in elmo command.

schmmd:
I think this is an improvement:

* It removes an argument that's a bit confusing.
* We no longer use stringified numbers as keys.
* When you run elmo, you don't need a separate file for the source sentences.

It's reasonably clear that sentences are encoded in UTF when used as names for datasets (see http://docs.h5py.org/en/latest/high/group.html?highlight=utf).

If for some reason we don't want this, I plan to store the sentence text as an attribute on the dataset.

### DIFF
--- a/allennlp/commands/elmo.py
+++ b/allennlp/commands/elmo.py
@@ -8,11 +8,6 @@ The input file is previously tokenized, whitespace separated text, one sentence 
 The output is a hdf5 file (<http://docs.h5py.org/en/latest/>) where each
 sentence is a size (3, num_tokens, 1024) array with the biLM representations.
 
-In the default setting, each sentence is keyed in the output file by the line number
-in the original text file.  Optionally, by specifying --use-sentence-key
-the first token in each sentence is assumed to be a unique sentence key
-used in the output file.
-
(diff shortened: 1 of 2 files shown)


allennlp/commands/elmo.py
```diff
@@ -257,21 +247,13 @@ def embed_file(self,
...
-        sentences = [line.strip().split() for line in input_file]
-        if use_sentence_key:
-            # Uses the first token in each sentence as the key.
-            keys, sentences = zip(*[(tokens[0], tokens[1:]) for tokens in sentences])
-            embedded_sentences = zip(keys, self.embed_sentences(sentences, batch_size))
-        else:
-            # Uses the index as the key.
-            embedded_sentences = enumerate(self.embed_sentences(sentences, batch_size))
+        sentences = [line.strip() for line in input_file]
+        split_sentences = map(lambda sentence: sentence.split(), sentences)
```
DeNeutoy: `split_sentences = [sentence.split() for sentence in sentences]`

schmmd: Thanks.  Writing that map did feel awkward...

matt-peters:
LGTM

PR #800, opened 2018-02-08
Warn on duplicate sentences

schmmd:
Previously duplicate sentences would cause an error when running the ELMo command.  This change outputs a warning and continues instead.

### DIFF
--- a/allennlp/commands/elmo.py
+++ b/allennlp/commands/elmo.py
@@ -258,11 +258,14 @@ class ElmoEmbedder():
         logger.info("Processing sentences.")
         with h5py.File(output_file_path, 'w') as fout:
             for key, embeddings in Tqdm.tqdm(embedded_sentences):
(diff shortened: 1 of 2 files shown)


schmmd:
Sorry for the rapid fire.  As I clicked merge, I thought of this.

allennlp/commands/elmo.py
```diff
@@ -258,11 +258,14 @@ def embed_file(self,
...
-                fout.create_dataset(
-                        str(key),
-                        embeddings.shape, dtype='float32',
-                        data=embeddings
-                )
+                if key in fout.keys():
+                    logger.warning(f"Key already exists in {output_file_path}, skipping: {key}")
+                else:
+                    fout.create_dataset(
+                            str(key),
```
matt-peters: Since `key` here is a string, we don't need the cast anymore.

PR #83, opened 2017-08-03
Create a run_on_kube.sh script

schmmd:

PR #801, opened 2018-02-08
Pass through Beaker commands in run_with_beaker.py.

schmmd:
Adding each command is a bit of a pain, but it provides excellent usage strings for our users.  Unfortunately `run_with_beaker.py` needs to be modified when Beaker changes it's commands under `beaker experiment run`.

### DIFF
--- a/scripts/ai2-internal/run_with_beaker.py
+++ b/scripts/ai2-internal/run_with_beaker.py
@@ -6,13 +6,14 @@ import argparse
 import os
 import subprocess
 import sys
+from typing import List
 
 sys.path.insert(0, os.path.dirname(os.path.abspath(os.path.join(os.path.join(__file__, os.pardir), os.pardir))))
 
 from allennlp.commands.train import Train
 from allennlp.common.params import Params
 
-def main(param_file, extra_beaker_commands):
+def main(param_file: str, extra_beaker_commands: List[str]):
     ecr_repository = "896129387501.dkr.ecr.us-west-2.amazonaws.com"
     commit = subprocess.check_output(["git", "rev-parse", "HEAD"], universal_newlines=True).strip()
     image = f"{ecr_repository}/allennlp/allennlp:{commit}"
--- a/scripts/ai2-internal/run_with_beaker.py
+++ b/scripts/ai2-internal/run_with_beaker.py
@@ -29,16 +29,15 @@ def main(param_file, extra_beaker_commands):
     # Get temporary ecr login. For this command to work, you need the python awscli
     # package with a version more recent than 1.11.91.
     print("Logging into ECR")
-    # subprocess.run('eval $(aws --region=us-west-2 ecr get-login --no-include-email)', shell=True, check=True)
+    subprocess.run('eval $(aws --region=us-west-2 ecr get-login --no-include-email)', shell=True, check=True)
 
     print(f"Building the Docker image ({image})")
-    # subprocess.run(f'docker build -t {image} .', shell=True, check=True)
+    subprocess.run(f'docker build -t {image} .', shell=True, check=True)
 
     print(f"Pushing the Docker image ({image})")
-    # subprocess.run(f'docker push {image}', shell=True, check=True)
+    subprocess.run(f'docker push {image}', shell=True, check=True)
 
-    # config_dataset_id = subprocess.check_output(f'beaker dataset create --quiet {param_file}', shell=True, universal_newlines=True).strip()
-    config_dataset_id = "BOGUS"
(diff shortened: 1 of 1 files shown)


scripts/ai2-internal/run_with_beaker.py
```diff
@@ -12,9 +12,7 @@
 from allennlp.commands.train import Train
 from allennlp.common.params import Params
 
-#TODO(michaels): add CLI support for mounting datasets.
-
-def main(param_file, description):
+def main(param_file, extra_beaker_commands):
     ecr_repository = "896129387501.dkr.ecr.us-west-2.amazonaws.com"
     commit = subprocess.check_output(["git", "rev-parse", "HEAD"], universal_newlines=True).strip()
     image = f"{ecr_repository}/allennlp/allennlp:{commit}"
```
DeNeutoy: I missed this in the previous PR for this script, but I think you need to add a random hash as in the previous script onto the end of the git commit here, otherwise when the image gets pulled from ECR it could be different if someone else has also pushed an image with the same name (assuming beaker pulls the `latest` image by default). Or maybe that doesn't matter now, as beaker pulls the image almost immediately after you push it. Still, seems safer to add the hash unless there's something i'm missing

matt-gardner: If it's easy, it'd be even better if you can say whether the working tree is clean or dirty, too.  If it's not easy, it's probably not worth the effort, though.

schmmd: @DeNeutoy I agree we should add a hash.  The old script's random number [was broken](https://github.com/allenai/allennlp/pull/757/files#diff-91c872fd064a4d494d631616c1eee34fL17) so I skipped this in my last PR (I wondered if I were missing something).

It's easy to tell whether the working tree is clean or dirty.  This is [how I handled it](https://github.com/allenai/allennlp/pull/83/files#diff-8dd2caa4ff5cd68e3b52d78173f5cc7aR28) in a PR from way back when.  However, I (and I suspect our users) often run Beaker experiments in dirty git trees.  I worry this feature would be annoying, and I want to make sure this command isn't annoying so people actually use it.

I'm uncertain about a few parts of this script.  I'm considering splitting it into two--one which builds images and one which runs images.

matt-gardner: Oh, yes, I definitely run things with dirty trees sometimes.  I wasn't suggesting crashing if it's dirty, just including either "dirty" or "clean" in the image tag, before the random number, as an extra piece of info if you're looking at a beaker experiment.

schmmd: @matt-gardner good suggestion.  I'm now thinking of appending a random number iff it's dirty.

@DeNeutoy `RANDOM` was undefined.  The image names it created ended with a hyphen.

DeNeutoy: Oh interesting - on my machine $RANDOM generates a random number every time. :man_shrugging:

scripts/ai2-internal/run_with_beaker.py
```diff
@@ -63,17 +61,40 @@ def main(param_file, description):
...
 
     parser.add_argument('param_file', type=str, help='The model configuration file.')
-    parser.add_argument('description', type=str, help='A description for the experiment.')
+    parser.add_argument('--desc', type=str, required=True, help='A description for the experiment.')
+    parser.add_argument('--debug', action='store_true', help='Print verbose stack traces on error.')
+    parser.add_argument('--detach', action='store_true', help='Run experiment in the background and print its ID.')
+    parser.add_argument('--env', action='append', help='Set environment variables (e.g. NAME=value or NAME)')
+    parser.add_argument('--mount', action='append', help='Bind a host directory (e.g. /host/path:/target/path)')
+    parser.add_argument('--source', action='append', help='Bind a remote data source (e.g. source-id:/target/path)')
+    parser.add_argument('--cpu', help='CPUs to reserve for this experiment (e.g., 0.5)')
```
matt-gardner: If we're letting people specify this, we might as well also let them specify a GPU count, defaulting to 1.  This will let people not use a GPU if they don't need it, and will be future-proof for when people start using multiple GPUs.

scripts/ai2-internal/run_with_beaker.py
```diff
@@ -12,9 +12,7 @@
 from allennlp.commands.train import Train
 from allennlp.common.params import Params
 
-#TODO(michaels): add CLI support for mounting datasets.
-
-def main(param_file, description):
+def main(param_file, extra_beaker_commands):
```
matt-gardner: Can you give types here?

PR #802, opened 2018-02-09
Handle padding in ListField[ArrayField]

tusharkhot:
`ListField[ArrayField]` pads the list with empty `ArrayField`s. But it uses the default padding value=0 and can lead to the model incorrectly assuming the padded arrays to be filled. This change modifies the `empty_field` to also pass the `padding_value` to the padded `ArrayField` object.

Thoughts ?

### DIFF
--- a/allennlp/data/fields/array_field.py
+++ b/allennlp/data/fields/array_field.py
@@ -45,4 +45,6 @@ class ArrayField(Field[numpy.ndarray]):
 
     @overrides
     def empty_field(self):  # pylint: disable=no-self-use
-        return ArrayField(numpy.array([], dtype="float32"))
+        # pass the padding_value, so that any outer field, e.g., ` ListField[ArrayField]` uses the
+        # same padding_value in the padded ArrayFields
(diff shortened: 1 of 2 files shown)


allennlp/data/fields/array_field.py
```diff
@@ -45,4 +45,6 @@ def as_tensor(self,
 
     @overrides
     def empty_field(self):  # pylint: disable=no-self-use
-        return ArrayField(numpy.array([], dtype="float32"))
+        # pass the padding_value, so that any outer field, e.g., ` ListField[ArrayField]` uses the
```
matt-gardner: Capitalize this, please.

tusharkhot: Done

tusharkhot:
Waiting for the checks, although I think they should definitely pass. They had passed earlier and I only changed capitalization in the latest commit.

tusharkhot:
Just noticed that I dont have privileges to merge a commit.

PR #804, opened 2018-02-09
WIP: multi-paragraph reading comprehension

joelgrus:
I think this is mostly right, the model is based on Mandar's code, which I think is based on Chris's code.

The TriviaQA reader is fairly complicated. That's because it supports three different use cases:

* supplying `triviaqa-rc.tar.gz` (in that case it unzips it to a temp directory, because `tarfile.open` random reads take forever on a 7GB file)
* supplying the unzipped directory (which is not really much bigger than the tarfile)
* supplying pre-processed + pre-tokenized merged paragraphs (generated by `scripts/preprocess_triviaqa`. The preprocessing is done by the exact same code that does the processing in the non-preprocessed case.

on SQuAD, this gets em: 69.0 f1: 78.9, which is slightly worse than what's in the paper (em 71.2, f1: 80.2)

http://beaker.allenai.org/ex/ex_un7z10igbi64

(note that Chris's original PR said he was only able to get 79.4 f1 on SQuAD in AllenNLP)

on TriviaQA, this gets up to em: 56, f1: 62, which is about 10 points worse than what's in the paper. (but most likely I'm not handling the dev set in the right way, mine merges into paragraphs of size 400 and then takes the top 15 by tfidf)

http://beaker.allenai.org/ex/ex_zvcyo4hhj7ij

(beaker restarted the experiment after 2 days 😭but the 56/62 is what it was trending toward)

--

to make sure the model wasn't way off in left field, I trained it on just the first 5000 questions, that version was able to get to em: 82 / f1: 86 on the training set:

http://beaker.allenai.org/ex/ex_3mr1n49p3u1l

--

so there's something I'm not doing quite right, although I'm not sure if it's in the model, how I'm processing the data, or how I'm handling the validation set


.pylintrc
```diff
@@ -182,7 +182,7 @@ expected-line-ending-format=
 [BASIC]
 
 # Good variable names which should always be accepted, separated by a comma
-good-names=i,j,k,ex,Run,_
+good-names=i,j,k,f,g,ex,Run,_
```
matt-gardner: Did you actually use these anywhere?

allennlp/data/dataset_readers/reading_comprehension/squad.py
```diff
@@ -29,6 +29,10 @@ class SquadReader(DatasetReader):
 
     Parameters
     ----------
+    multiparagraph : ``bool``, optional (default=``False``)
```
matt-gardner: Nit: I'd put parameters that are less likely to be used / less central to the class at the bottom.

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
```diff
@@ -1,32 +1,245 @@
+# pylint: disable=no-member
```
matt-gardner: Where is this an issue?  Does it make sense to put it more locally?  I'm guessing this is from `sklearn` - you could also whitelist this in `.pylintrc`, and it might fix the issue.

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
```diff
@@ -1,32 +1,245 @@
...
+# and start training from the processed data.
+
+class MergedParagraphs(NamedTuple):
+    """
+    A typical TriviaQA question has multiple files with answers, each consisting
+    of many small paragraphs. We merge these into fewer paragraphs of larger size.
+    """
+    texts: List[str]
+    tokens: List[List[Token]]
+    token_spans: List[List[Tuple[int, int]]]
```
matt-gardner: These are answer spans?

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
```diff
@@ -1,32 +1,245 @@
...
+    has_answers: List[int] = None
```
matt-gardner: Why is this an `int`?

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
```diff
@@ -1,32 +1,245 @@
...
+    Given some (small?) paragraphs and a question, merge the paragraphs to make
+    synthetic paragraphs of size ``max_paragraph_length``, and return a ``MergedParagraphs``
+    instance containing the "best" ``topn`` of them. If ``topn`` is None, return all of them.
+
+    If ``tfidf_sort`` is True, "best" means closest (in terms of tfidf distance) to the question.
+    If ``require_answer`` is True, then the best "best" paragraph is also required to contain
+    one of the ``answer_texts``. Don't do this with a test or validation dataset, that's cheating.
+    If neither is True, then just the first ``topn`` paragraphs are returned.
+    """
+    # Collect all the text into one mega-paragraph.
```
matt-gardner: This is a problem.  You're losing all paragraph (and even sentence) boundaries, then splitting on number of tokens.  This means that you'll get paragraphs split in the middle of sentences.  This will certainly cause a drop in accuracy relative to Chris's implementation.  You can see Chris's original code here: https://github.com/allenai/document-qa/blob/35c9b862c5a743b11f9532b3e1ba11c348bb9621/docqa/data_processing/document_splitter.py#L302-L345.  Basically: he maintains paragraph boundaries, and just merges consecutive paragraphs until he reaches the max tokens; if adding the next paragraph would put you over the limit, you start a new paragraph.  If there's a single paragraph that's longer than the max tokens, he truncates it.

allennlp/data/dataset_readers/reading_comprehension/squad.py
```diff
@@ -100,17 +106,28 @@ def text_to_instance(self,  # type: ignore
...
-        return util.make_reading_comprehension_instance(self._tokenizer.tokenize(question_text),
-                                                        passage_tokens,
-                                                        self._token_indexers,
-                                                        passage_text,
-                                                        token_spans,
-                                                        answer_texts)
+        if self._multiparagraph:
+            return util.make_multi_paragraph_reading_comprehension_instance(
+                    self._tokenizer.tokenize(question_text),
+                    [passage_tokens],
```
murphp15: For squad we only pass in the paragraph which contains the answer? 
Does this not defy the point of this PR, as we are suppose to be training on a model which has paragraphs which do not contain the answer?

joelgrus: the squad dataset only has one paragraph for each question, what else should I be passing in?

matt-gardner: Chris also tested on versions where you have to pick the paragraph from SQuAD; we haven't gotten that far yet.

murphp15: In squad dataset there is multiple paragraphs from a single document. for example there are a lot of superbowl questions which all derive from the single article: https://rajpurkar.github.io/SQuAD-explorer/explore/1.1/dev/Super_Bowl_50.html?model=r-net+%20(ensemble)%20(Microsoft%20Research%20Asia)

Perhaps we could pass all paragraphs from the section same section too as long as they do not contain the answer. 

Although for the moment I think what you have is good enough :)

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
```diff
@@ -1,32 +1,280 @@
...
+            # Keep the previous paragraph and start a new one
+            merged_paragraphs.append(current_paragraph)
+            current_length = paragraph_length
+            current_paragraph = paragraph
+
+        elif too_long and current_length > 0 and paragraph_length > max_paragraph_length:
+            # Keep the previous paragraph and truncate this one
+            truncated = ' '.join(token.text for token in paragraph_tokens[:max_paragraph_length])
+
+            merged_paragraphs.append(current_paragraph)
```
murphp15: Why are we adding the current_paragraph to the merged_paragraphs? 
Does it not make sense to just add a truncated version of the long paragraph to the merged_paragraph and do nothing with the current_paragraph. 
This way the current_paragraph can be accumulated with later paragraphs which are also small?

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
```diff
@@ -1,32 +1,280 @@
...
+                            tokens=merged_tokens,
+                            token_spans=merged_token_spans,
+                            has_answers=[i for i, choice in enumerate(choices) if choice in has_answers])
+
+
+def process_triviaqa_questions(evidence_path: pathlib.Path,
+                               questions_path: pathlib.Path,
+                               tokenizer: Tokenizer,
+                               max_paragraphs: int = 4,
+                               tfidf_sort: bool = True,
```
murphp15: Presumably anyone using this multi paragraph network will have a tfidf paragraph store sitting in front of the network, they will choose the highest ranked (tfidf similarity) paragraphs for the question and pass them into the network. 
the network then returns them the correct answer. 

taking this into account along with the fact that tfidf_sort is always true, maybe we should just omit it as I don't really see a scenario where you would want to train a system with it set to false. 

What do you guys think?

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
murphp15: It is a list of indexes for merged paragraphs which contain the answer.
For example if we have 5 paragraphs and the first 2 paragraphs contain the answer then the list will be [0,1].

We could get rid of this variable and use the token_spans list to determine which paragraphs have answers. Not sure if you prefer that?

murphp15:
Hey Guys, 
I would love to get this merged in. 
Is there any way I would be able to help you guys get this PR closed?

murphp15:
I know this is probably not the right place to ask this question but when I run this locally I get the following error ' AttributeError: module 'torch._C' has no attribute '_cuda_getDevice' '.
I realise that this is happening because I have enabled it to use graphics cards but I might not have one on my machine. 
Would you guys be able to explain how you run the training? Do you guys run it on a service like google cloud?

murphp15:
After training the system for a very small amount of time I have the following output metrics. 
I think there must be a mistake. as it is saying that I have an f1 score of 77.7777777 and an exact match score of 33.333. 

Metrics: {
  "training_duration": "00:31:29",
  "training_start_epoch": 0,
  "training_epochs": 100,
  "training_em": 33.333333333333336,
  "training_f1": 77.77777777777777,
  "training_loss": 11.366336822509766,
  "validation_em": 0.0,
  "validation_f1": 0.0,
  "validation_loss": 38.27989959716797,
  "best_validation_em": 0.0,
  "best_epoch": 99
}

The command I used to run the system is:  allennlp train training_config/documentqa.triviaqa.json --serialization-dir triviaqa-output

allennlp/models/reading_comprehension/document_qa.py
```diff
@@ -0,0 +1,401 @@
...
+
+        assert text_field_embedder.get_output_dim() == phrase_layer.get_input_dim()
+        self._phrase_layer = phrase_layer
+        encoding_dim = phrase_layer.get_output_dim()
+        # output: (batch_size, num_tokens, encoding_dim)
+
+        self._matrix_attention = TriLinearAttention(encoding_dim)
+        # output: (batch_size, num_tokens, num_tokens)
+
+        self._merge_atten = TimeDistributed(torch.nn.Linear(encoding_dim * 4, encoding_dim))
```
murphp15: You no longer need to wrap the linear layer in the time distributed layer. 
The documentation is now updated to say 


 Shape:
        - Input: :math:`(N, *, in\_features)` where `*` means any number of
          additional dimensions
        - Output: :math:`(N, *, out\_features)` where all but the last dimension
          are the same shape as the input.

allennlp/models/reading_comprehension/document_qa.py
```diff
@@ -0,0 +1,401 @@
...
+    """
+    def __init__(self, vocab: Vocabulary,
+                 text_field_embedder: TextFieldEmbedder,
+                 phrase_layer: Seq2SeqEncoder,
+                 residual_encoder: Seq2SeqEncoder,
+                 span_start_encoder: Seq2SeqEncoder,
+                 span_end_encoder: Seq2SeqEncoder,
+                 initializer: InitializerApplicator,
+                 dropout: float = 0.2,
+                 mask_lstms: bool = True) -> None:
```
murphp15: What is the scenario where you would not want to mask the output of the lstm?

allennlp/data/dataset_readers/reading_comprehension/squad.py
matt-gardner: Yes, that's basically what Chris did, and I think we'll want to have that eventually.  It doesn't have to be in this PR, though.

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
matt-gardner: `current_paragraph` is what we have already from previous iterations of the loop.  In my first read of this code, I was also confused by this, as I just jumped to line 118 and assumed that `current_length` was the length of the paragraph at the current iteration.  That's actually `paragraph_length`.

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
matt-gardner: If you want to measure the effect of using TF-IDF to sort versus just taking, e.g., the first N paragraphs, then yes, you want this flag.  This was one of the experiments from the original paper.  AllenNLP is about doing science, so we tend to have options for things that you might experiment with in a paper.

allennlp/models/reading_comprehension/document_qa.py
matt-gardner: In some early experiments I did with BiDAF, I found that omitting the masking in the LSTMs sped things up a whole lot (as it lets us skip some expensive sorting before passing things off to CUDA), and it didn't really affect accuracy.  So this flag would let you trade off a little bit of correctness for a whole bunch of speed.  @nelson-liu recently tried to reproduce the speedup I saw, though, and didn't see it.  We're not sure why - maybe it doesn't make as much difference with newer hardware, or maybe some other code change we did in the interim made it so we're always sorting anyway, or something.

matt-gardner:
@murphp15, if you want to help with this, the main problem we were seeing was that we weren't reproducing the reported performance in the paper on TriviaQA.  We got close enough to the results on SQuAD (when using a single paragraph for training), so it sure looks like the model is fine; it was probably some difference in the data processing code when producing multiple paragraphs for TriviaQA.

I know @joelgrus was working on other things for the EMNLP deadline, and I'm not sure where this PR falls in our priorities right now.  Something that would be helpful, though, is trying to train this model on TriviaQA, seeing what performance you get, and if it's still different from the paper, going over the data processing code looking for differences with how Chris's paper describes the data processing (or even comparing it with Chris's code).  I'll warn you, though, that training this model takes (I think) two days on a GPU - it's not cheap to help out with this.

murphp15:
@matt-gardner 
When you say it takes 2 days to train what kinda machine spec are we talking about training on? 
Would it take 2 days to train on a p2.xLarge from aws?  Because that would only be about 50 euro which I can get my company to cover for me.
Or would I need to go for something better? 

Here is a link to the aws machines on demand page https://aws.amazon.com/ec2/pricing/on-demand/
GPUs are about halfway down the page.
 
On another note: 
When I decide on the appropriate aws box to use. What is the best way to train the system and monitor its progress during the 2 days?
As best I can understand I should run: 
 `allennlp train training_config/documentqa.triviaqa.json --serialization-dir triviaqa-output`
Then run a tensorboard instance on the serialization-dir with
` tensorboard --logdir triviaqa-output/`
Then I could view the progress on the remote machine using my browser. 
Is that how you would recommend monitoring the progress?

allennlp/data/dataset_readers/reading_comprehension/triviaqa.py
murphp15: I think there is a misunderstanding between the two of us. 
So I have written a test case to explain my point. 
In the test case there are 3 paragraphs. 
The first and the last are very short, with the middle one being longer than the max allowed. 


    def test_merge(self):
        short_paragraph_1 = "short paragraph before the long paragraph"
        really_long_para = "Paul is 28 years old. this is a really long para which is much longer than the max length of a paragraph we expect this to be truncated"
        short_paragraph_2 = "short paragraph after the long paragraph"
        paragraphs = [short_paragraph_1, really_long_para, short_paragraph_2]

        re = _merge(paragraphs=paragraphs,
                    question="what age is Paul",
                    answer_texts='28',
                    tokenizer=WordTokenizer.from_params(Params({})),
                    max_paragraphs=4,
                    tfidf_sort=True,
                    require_answer=False,
                    max_paragraph_length=16)

        assert len(re.texts) == /* what should the number here be? */


At the moment the MergedParagraph returned from the _merge method consists of 3 texts:
0 = {str} ' @@PARAGRAPH@@ short paragraph before the long paragraph'
1 = {str} 'Paul is 28 years old . this is a really long para which is much longer'
2 = {str} ' @@PARAGRAPH@@ short paragraph after the long paragraph'

However I would expect it to return 2 texts: 
0 = {str} 'Paul is 28 years old . this is a really long para which is much longer'
1 = {str} ' @@PARAGRAPH@@ short paragraph before the long paragraph @@PARAGRAPH@@ short paragraph after the long paragraph'

Do you agree with this?

schmmd:
@murphp15 a p2.xlarge should be fine.

allennlp/models/reading_comprehension/document_qa.py
```diff
@@ -0,0 +1,401 @@
...
+        span_end_logits = self._span_end_predictor(end_rep).squeeze(-1)
+        span_end_probs = util.masked_softmax(span_end_logits, passage_mask)
+
+        span_start_logits = util.replace_masked_values(span_start_logits, passage_mask, -1e7)
+        span_end_logits = util.replace_masked_values(span_end_logits, passage_mask, -1e7)
+
+        paragraph_span_start_logits = unsquash(span_start_logits, batch_size, num_paragraphs)
+        paragraph_span_end_logits = unsquash(span_end_logits, batch_size, num_paragraphs)
+
+        best_paragraph_word_span = self._get_best_span(paragraph_span_start_logits, paragraph_span_end_logits)
```
murphp15: Why is it that you use the logits to compute the best span as apposed to using the probabilities?

allennlp/models/reading_comprehension/document_qa.py
```diff
@@ -0,0 +1,401 @@
...
+    def get_metrics(self, reset: bool = False) -> Dict[str, float]:
+        return {'em': self._official_em.get_metric(reset),
+                'f1': self._official_f1.get_metric(reset)}
+
+    @staticmethod
+    def _get_best_span(span_start_logits: Variable, span_end_logits: Variable) -> Variable:
+        batch_size, num_paragraphs, passage_length = span_start_logits.size()
+        max_span_log_prob = [-1e20] * batch_size
+        # span_start_argmax = [0] * batch_size
+        best_word_span = Variable(span_start_logits.data.new()
```
murphp15: I am obviously missing something here but why are you not just doing tensor.new_zeros((batch_size, 3))?

allennlp/models/reading_comprehension/document_qa.py
```diff
@@ -0,0 +1,401 @@
...
+                                  .resize_(batch_size, 3).fill_(0)).long()
+
+        span_start_logits = span_start_logits.data.cpu().numpy()
+        span_end_logits = span_end_logits.data.cpu().numpy()
+
+        # pylint: disable=invalid-name
+        for b in range(batch_size):
+            for p in range(num_paragraphs):
+                span_start_argmax = 0
+                for j in range(passage_length):
```
murphp15: Is there anything stopping the span_start_argmax being more than j?
As far as I can see there is nothing stopping this?

allennlp/models/reading_comprehension/document_qa.py
murphp15: Do you compare using logits instead of probabilities because they will compare across paragraph boundaries better than the the probabilities will?

allennlp/models/reading_comprehension/document_qa.py
```diff
@@ -0,0 +1,401 @@
...
+            span_idx_mask = span_idx_mask.max(dim=-1)[0]
+
+            # (batch_size, num_paragraphs, num_spans)
+            span_starts = spans[:, :, :, 0]
+            span_ends = spans[:, :, :, 1]
+
+            # (batch_size, num_paragraphs, num_tokens)
+            paragraphs_mask = util.get_text_field_mask(paragraphs, num_wrapping_dims=1).float()
+            # (batch_size, num_paragraphs)
+            paragraphs_mask = paragraphs_mask.max(dim=-1)[0]
```
murphp15: I don't understand this at all.
What is the scenario where this matrix is not all ones?
Does it check if a paragraph is completely empty, surely there is no need to check this?

allennlp/nn/loss/shared_norm_loss.py
```diff
@@ -0,0 +1,46 @@
+import torch
+from torch.autograd import Variable
+
+from allennlp.nn.util import logsumexp
+
+# span_logits is (batch_size, num_paragraphs, num_tokens)
+# paragraph_mask is (batch_size, num_paragraphs)
+# actual_spans is (batch_size, num_paragraphs, num_spans)
```
murphp15: Its a small bit confusing that this is called span, 
Even though it is not a span and is in fact only the start or end. 
But when there is documentation on the class it will probably be clear.

allennlp/models/reading_comprehension/document_qa.py
matt-gardner: Because that function was added in pytorch 0.4, which allennlp still doesn't support (though it will soon).

allennlp/models/reading_comprehension/document_qa.py
matt-gardner: I don't understand the question.  The only assignment to `span_start_argmax` inside the loop is `span_start_argmax = j`.

allennlp/models/reading_comprehension/document_qa.py
matt-gardner: For getting the argmax here, it doesn't matter whether you use logits or probabilities, except that the logic assumes logits (it uses `val1 + val2` instead of `val1 * val2`).  When comparing across paragraphs, using logits instead of probabilities is important, for the reasons discussed in the paper.

allennlp/models/reading_comprehension/document_qa.py
matt-gardner: We batch together several instances, and some may have more paragraphs than others.  This means that we need to pad the number of paragraphs to a fixed length for all instances in the batch, so, yes, there will be some paragraphs that are completely empty.

murphp15: Yeah of course! I don't know how I missed that.

murphp15:
I see that in the document_qa repo it uses more token breaks than this impl. 
E.g it has 
```
PARAGRAPH_TOKEN = "%%PARAGRAPH%%"
DOCUMENT_START_TOKEN = "%%DOCUMENT%%"
PARAGRAPH_GROUP = "%%PARAGRAPH_GROUP%%"
```
 whereas in this impl we are only using 

`_PARAGRAPH_TOKEN = "@@PARAGRAPH@@".`

Why do we not need the other tokens in this impl? 
@chrisc36 Perhaps you know the answer?

schmmd:
@murphp15 you shouldn't need a p2.8xlarge--we seldom use those at our company.  That said--I have not trained this model myself.

murphp15:
@schmmd but loading the dataset into memory takes about 100 the of ram ? So the p2xlarge is not big enough ?

MaxMotovilov:
Have been running the training w/full TriviaQA dataset (400K records) on 8 GPUs for the last 24-odd hours. 5 epochs completed, seeing validation em=61.3, f1=66.8 and barely climbing at all. Seems better than @joelgrus numbers but not 10 point better.

BTW, `Peak CPU memory usage MB: 265318.44`. Heh. So much for Amazon p2.xlarge. Admittedly we've pulled out the stops due to hardware capacity but in trial runs the number wasn't all that different.

joelgrus:
did you use the preprocessing script?

MaxMotovilov:
Yes - fairly sure we followed the procedure y'all discussed in this PR. @murphp15 helped us get going.

MaxMotovilov:
The relevant section of config:

```    "dataset_reader": {
      "type": "triviaqa",
      "lazy": false,
      "paragraph_picker": "triviaqa-web-train",
      "triviaqa_path": "triviaqa_processed",
      "data_format": "preprocessed",
      "max_token_length": 20,
      "token_indexers": {
        "tokens": {
          "type": "single_id",
          "lowercase_tokens": true
        },
        "token_characters": {
          "type": "characters",
          "character_tokenizer": {
            "byte_encoding": "utf-8"
          }
        }
      }
    },```

Is "data_format = preprocessed" what you were asking about?

joelgrus:
yes, that's what I meant. when I ran it, I believe it capped out around ~50GB of memory, but I was only using one GPU, not sure how the 8x affects things.

MaxMotovilov:
Epochs 6 and 7 decreased validation performance so we stopped it. Here's the model (after epoch 5):
https://objectstorage.us-ashburn-1.oraclecloud.com/n/oci_core_eca_jmiller/b/Trained-Models/o/trivia-full-e5-model.tar.gz

murphp15:
@joelgrus I also seen that it was taking upwards of 100gb of ram to run. 
Also would you expect that it doesn't improve after the first epoch?

joelgrus:
closing because we were never able to reproduce the results from the original paper and we were never able to figure out why

dav-ell:
@joelgrus @murphp15 Just wanted to make you guys aware that AWS has spot instances (including p2.xlarge and p2.8xlarge) that are available at a significant discount. I believe they can be configured to automatically hibernate and resume when interrupted. Perhaps this could help with future endeavors like this.

https://aws.amazon.com/ec2/spot/pricing/

hitxujian:
> yes, that's what I meant. when I ran it, I believe it capped out around ~50GB of memory, but I was only using one GPU, not sure how the 8x affects things.

triviaqa_processed is how to formed?

PR #805, opened 2018-02-09
Append a random SHA to Docker image if dirty.

schmmd:


scripts/ai2-internal/run_with_beaker.py
```diff
@@ -20,14 +21,20 @@ def main(param_file, description):
...
     flat_params = params.as_flat_dict()
     env = []
     for k, v in flat_params.items():
         k = str(k).replace('.', '_')
         env.append(f"--env={k}={v}")
 
+    # If the git repository is dirty, add a random hash.
+    result = subprocess.run('git diff-index --quiet HEAD --', shell=True)
+    if result.returncode != 0:
+        hash = "%x" % random.getrandbits(32)
```
matt-gardner: Sorry, I clicked the wrong button, so my comment was lost.  It was something like this:

`hash` is a built-in function in python, which is why it shows up as blue here. Pylint would have caught this if we ran pylint on `scripts/`, I think.  Can you use a different variable name here?

schmmd: It's now `dirty_hash`!  I'm surprised it worked...

matt-gardner: You're allowed to overwrite pre-existing functions, it can just be very confusing if you want to use them later.  It's in a _keyword_ in the language, just a built-in _function_.

Issue #785, opened 2018-02-06
Add ELMo to the CLI


schmmd:
* Configurable whether we output all (`--all`) three vectors or an average vector (`--average`).
* Able to output contextualized word vectors (input is a sentence) or [uncontextualized word vectors](https://github.com/allenai/allennlp/blob/master/scripts/create_elmo_embeddings_from_vocab.py) (input is a word).

DeNeutoy:
also flags for individual layers would probably be an easy and helpful addition, like `--top, --middle, --bottom`

matt-peters:
My $0.02, I'd add the `--top` flag but eliminate the rest -- more code to maintain, and if the model is swapped out at some point in the future for one with more/less layers then the `--middle` flag is not longer meaningful.  The `--bottom` is covered in the uncontextualized case.

schmmd:
@matt-peters @DeNeutoy after a discussion with Matt Peters I was a bit confused about what support we want for uncontextualized word vectors.  Matt mentioned that the script referred to above was for a caching strategy that wasn't fully implemented.  He also mentioned that uncontextualized word vectors wouldn't be an advantage over word2vec vectors.

Do we want to add support for uncontextualized word vectors to the elmo command?  If so, what do you envision it looking like?

matt-peters:
It'd be nice to finish the caching strategy where the uncontextualized word vectors are pre-computed for a fixed vocabulary as it speeds up computing the contextual vectors and reduces memory pressure.  But without that I don't see an immediate need to add the uncontextualized flag here.

schmmd:
I'm closing this, but I opened https://github.com/allenai/allennlp/issues/809.

PR #806, opened 2018-02-09
Add options --all, --top, and --average to elmo command

schmmd:
Partially addresses https://github.com/allenai/allennlp/issues/785.


schmmd:
@matt-peters I originally used `numpy.testing.assert_allclose` which passed on my machine but failed in the cloud.  [NumPy recommends](https://docs.scipy.org/doc/numpy-1.9.2/reference/generated/numpy.testing.assert_array_almost_equal.html) using allclose.  Should I switch to that but set a higher tolerance?

matt-peters:
Seems reasonable to me.

PR #808, opened 2018-02-09
Remove --all flag from verify in docs.

schmmd:


schmmd:
@maxdel thanks for https://github.com/allenai/allennlp/pull/807#pullrequestreview-95536435 and FYI

PR #815, opened 2018-02-09
Fix tqdm behavior around lazy datasets

### LINKED ISSUE #814: DatasetReaders no longer have any tqdm output on read
This is unfortunate if you're reading a lot of data, because you just sit there not knowing how long it's going to take.  Probably the fix is to change the `lazy=False` case in `DatasetReader.read()` to use `tqdm` instead of `ensure_list`.  I'll do this soon unless someone tells me there's a better way to do this.

matt-gardner:
Fixes #814.

Having `tqdm` in the `DatasetReader` subclasses doesn't play nicely with `lazy=True`, because you get a `tqdm` line for the dataset reader in every batch during training.  The better place to have `tqdm` is in the base class, when converting the generator into a list.  This PR removes the `tqdm` calls from the subclasses and adds one to the base class.

### DIFF
--- a/allennlp/common/__init__.py
+++ b/allennlp/common/__init__.py
@@ -1,3 +1,4 @@
 from allennlp.common.params import Params
 from allennlp.common.registrable import Registrable
+from allennlp.common.tqdm import Tqdm
 from allennlp.common.util import JsonDict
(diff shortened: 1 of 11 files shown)


joelgrus:
as I mentioned in the issue, I don't think this functionality was ever in the base class (because it used to be empty), and a number of the specific dataset readers already have tqdm in them. what happens if you call tqdm on tqdm?

matt-gardner:
Ok @joelgrus, this should be ready.

PR #818, opened 2018-02-10
add constrained decoding for CRFs

joelgrus:
not 100% sure this is the best way to implement it, lmk what you think.

```
echo '{"sentence": "Sailing around the New York Harbor can be as educating as it is soothing."}' > examples.jsonl && python -m allennlp.run predict     https://s3-us-west-2.amazonaws.com/allennlp/models/ner-model-2017.11.15.tar.gz     examples.jsonl
```

results in 

```
"tags": ["O", "O", "O", "B-LOC", "I-MISC", "L-MISC", "O", "O", "O", "O", "O", "O", "O", "O", "O"]
```

but if you specify BIOUL constraints

```
echo '{"sentence": "Sailing around the New York Harbor can be as educating as it is soothing."}' > examples.jsonl && python -m allennlp.run predict     https://s3-us-west-2.amazonaws.com/allennlp/models/ner-model-2017.11.15.tar.gz     examples.jsonl -o '{"model":{"constraint_type":"BIOUL"}}'
```

you get

```
"tags": ["O", "O", "O", "B-MISC", "I-MISC", "L-MISC", "O", "O", "O", "O", "O", "O", "O", "O", "O"]
```

### DIFF
--- a/allennlp/modules/conditional_random_field.py
+++ b/allennlp/modules/conditional_random_field.py
@@ -21,15 +21,17 @@ def allowed_transitions(constraint_type: str, tokens: Dict[int, str]) -> Set[Tup
             for j, to_label in tokens.items():
                 to_bioul, *to_entity = to_label
 
-                if from_bioul == 'O' and to_bioul in ('O', 'U', 'B'):
-                    allowed.add((i, j))
-                elif from_bioul == 'B' and to_bioul in ('I', 'L') and from_entity == to_entity:
-                    allowed.add((i, j))
-                elif from_bioul == 'I' and to_bioul in ('I', 'L') and from_entity == to_entity:
-                    allowed.add((i, j))
-                elif from_bioul == 'L' and to_bioul in ('B', 'O', 'U'):
-                    allowed.add((i, j))
-                elif from_bioul == 'U' and to_bioul in ('B', 'O', 'U'):
+                is_allowed = any([
+                        # O can transition to O, B-* or U-*
+                        # L-x can transition to O, B-*, or U-*
+                        # U-x can transition to O, B-*, or U-*
+                        from_bioul in ('O', 'L', 'U') and to_bioul in ('O', 'B', 'U'),
+                        # B-x can only transition to I-x or L-x
+                        # I-x can only transition to I-x or L-x
+                        from_bioul in ('B', 'I') and to_bioul in ('I', 'L') and from_entity == to_entity
+                ])
+
+                if is_allowed:
                     allowed.add((i, j))
 
     elif constraint_type == "BIO":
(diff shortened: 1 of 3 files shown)


allennlp/modules/conditional_random_field.py
```diff
@@ -1,13 +1,57 @@
...
+from typing import List, Tuple, Dict
 
 import torch
 from torch.autograd import Variable
 
+from allennlp.common.checks import ConfigurationError
 import allennlp.nn.util as util
 
+
+def allowed_transitions(constraint_type: str, tokens: Dict[int, str]) -> List[Tuple[int, int]]:
```
DeNeutoy: A short doctring describing the two constraints and `tokens` would be good.

allennlp/modules/conditional_random_field.py
```diff
@@ -19,15 +63,27 @@ class ConditionalRandomField(torch.nn.Module):
...
         self.num_tags = num_tags
 
         # transitions[i, j] is the logit for transitioning from state i to state j.
         self.transitions = torch.nn.Parameter(torch.Tensor(num_tags, num_tags))
 
+        # _constraint_mask indicates valid transitions (based on supplied constraints).
+        if constraints is None:
+            self._constraint_mask = None
+        else:
+            self._constraint_mask = torch.Tensor(num_tags, num_tags).fill_(0.)
```
DeNeutoy: I don't think this is device agnostic, because when you call `.cuda` on a module, only it's `Parameters` are copied to the new device. It's going to be a little tricky to do this in this way though.

tests/modules/conditional_random_field_test.py
```diff
@@ -138,3 +140,59 @@ def test_viterbi_tags(self):
...
+                [2, 3, 3],
+                [2, 3]
+        ]
+
+    def test_allowed_transitions(self):
+        # pylint: disable=bad-whitespace,bad-continuation
+        bio_labels = ['O', 'B-X', 'I-X', 'B-Y', 'I-Y']
+        #              0     1      2      3      4
+        allowed = allowed_transitions("BIO", dict(enumerate(bio_labels)))
+        assert set(allowed) == {
```
DeNeutoy: Can you just add a comment here and below that empty spaces in this beautifully laid out matrix are not allowed - it took me 2 minutes of starring at this thinking you had lost your mind before I understood what was happening

allennlp/modules/conditional_random_field.py
joelgrus: done

allennlp/modules/conditional_random_field.py
joelgrus: I changed it to a `Parameter`, take a look and let me know if it looks right

tests/modules/conditional_random_field_test.py
joelgrus: done

allennlp/modules/conditional_random_field.py
```diff
@@ -1,13 +1,70 @@
...
+    """
+    Given tokens and a constraint type, returns the allowed transitions.
+
+    Parameters
+    ----------
+    constraint_type : ``str``, required
+        Indicates which constraint to apply. Current choices are "BIO" and "BIOUL".
+    tokens : ``Dict[int, str]``, required
+        A mapping {token_id -> token}. Most commonly this would be the value from
+        Vocabulary.get_index_to_token_vocabulary()
```
DeNeutoy: backticks around code.

allennlp/modules/conditional_random_field.py
```diff
@@ -19,15 +76,29 @@ class ConditionalRandomField(torch.nn.Module):
...
 
+        # _constraint_mask indicates valid transitions (based on supplied constraints).
+        if constraints is None:
+            self._constraint_mask = None
+        else:
+            constraint_mask = torch.Tensor(num_tags, num_tags).fill_(0.)
+            for i, j in constraints:
+                constraint_mask[i, j] = 1.
+
+            self._constraint_mask = torch.nn.Parameter(constraint_mask, requires_grad=False)
```
DeNeutoy: This seems like it should be fine - the only other way I could think of would be to allow `viterbi_tags` to take the transition matrix as a parameter, so you could just pass it in at runtime having created it in the CrfTagger. But this looks fine.

PR #821, opened 2018-02-12
Allow min_count to be namespace-specific

matt-gardner:


matt-gardner:
I think the original case was broken, and was likely not used.  You pretty much don't ever want to use the same threshold on all namespaces.  If you or anyone else really thinks we should add in an option to keep the `int` flag, I can do it, though.  I'll wait a bit before merging this, in case someone wants to say that we should keep it.

PR #824, opened 2018-02-12
Clarity the parameter group options

matt-peters:


allennlp/training/optimizers.py
```diff
@@ -40,23 +40,27 @@ def from_params(cls, model_parameters: List, params: Params):
...
             #"parameter_groups": [
-            #       [['regex1', 'regex2'], {'lr': 1e-3},
-            #        ['regex3'], {'lr': 1e-4}]
+            #       [["regex1", "regex2"], {"lr": 1e-3},
+            #        ["regex3"], {"lr": 1e-4}]
             #]
-            #
-            # The last entry of this list is for the parameters not in any regex.
-            #
+            #(note that the allennlp config files require double quotes ", and will
```
DeNeutoy: In what way do they fail? Seems like something we should address/mark prominently (not here, somewhere else), if it is a HOCON feature.

joelgrus: JSON standard is double quotes, HOCON is just a superset of JSON

joelgrus: to that end, I'm not sure it's worth calling out 

```
In [4]: import json

In [5]: json.loads(""" "double-quoted" """)
Out[5]: 'double-quoted'

In [6]: json.loads(""" 'single-quoted' """)
JSONDecodeError                           Traceback (most recent call last)
...
```

matt-peters: @roys174 noticed some strange behavior with single quotes and I added this note at his request.  I am not using the `from_params` machinery myself so can't comment on the details.

matt-peters: The snippet below reproduces the issue.  The single quote form is parsed OK, but has unexpected behavior (and in this case, will cause a silent error as the single quoted regex won't match as expected.

```
In [1]: from allennlp.common import Params

In [2]: with open('config.json', 'w') as fout:
   ...:     fout.write("""{"quotes": [[['single'], {"a": 5}], [["double", {"b": 6}]]]}""")
   ...:     

In [3]: p = Params.from_file('config.json')

In [4]: quotes = p.pop("quotes")

In [5]: print(quotes)
[[["'single'"], <allennlp.common.params.Params object at 0x109ef1e10>], [['double', <allennlp.common.params.Params object at 0x109ef19b0>]]]
```

joelgrus: so what's happening here is that HOCON allows *unquoted* strings (that is, strings not surrounded by double quotes) in some circumstances:

https://github.com/lightbend/config/blob/master/HOCON.md#unquoted-strings

and so it treats the single-quoted string as an unquoted string that contains single quotes.

joelgrus: in other words, if it's worth calling out that strings in our config files have to be double quoted, then it's worth calling that out more broadly than as just a comment in our optimizers file

matt-peters: I opened an issue to track this #830

PR #826, opened 2018-02-12
Bi-directional span extractor

DeNeutoy:
Adds `span_mask` and `indices_mask` to all of the `SpanExtractors`. Some of these aren't strictly necessary, but it's nice if they all have the same API and it's possible that you could want to implement a `SpanExtractor` which does use one or both of these masks.

Adds a bidirectional endpoint extractor and tests.


allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
```diff
@@ -0,0 +1,230 @@
...
+    bucket_widths : ``bool``, optional (default = False).
+        Whether to bucket the span widths into log-space buckets. If ``False``,
+        the raw span widths are used.
+    use_sentinels : ``bool``, optional (default = ``True``).
+        If ``True``, sentinels are used to represent exclusive span indices for the elements
+        in the first and last positions in the sequence (as the exclusive indices for these
+        elements are outside of the the sequence boundary). This is not strictly necessary,
+        as you may know that your exclusive start and end indices are always within your sequence
+        representation, such as if you have appended/prepended <START> and <END> tokens to your
+        sequence (equivelent to learning a sentinel representation in this class, but more accessable).
```
matt-gardner: This isn't totally true, because the LSTM states will be different if you've got start and end tokens.  I'd probably just take out the parenthetical here.

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
```diff
@@ -0,0 +1,230 @@
...
+            self._span_width_embedding = Embedding(num_width_embeddings, span_width_embedding_dim)
+        elif not all([num_width_embeddings is None, span_width_embedding_dim is None]):
+            raise ConfigurationError("To use a span width embedding representation, you must"
+                                     "specify both num_width_buckets and span_width_embedding_dim.")
+        else:
+            self._span_width_embedding = None
+
+        self._use_sentinels = use_sentinels
+        if use_sentinels:
+            self._start_sentinel = Parameter(torch.randn([1, 1, int(input_dim / 2)]))
```
matt-gardner: Hmm, yeah, start and end tokens really wouldn't be the same, because of the bidirectional split...

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
```diff
@@ -0,0 +1,230 @@
...
+            self._end_sentinel = Parameter(torch.randn([1, 1, int(input_dim / 2)]))
+
+    def get_input_dim(self) -> int:
+        return self._input_dim
+
+    def get_output_dim(self) -> int:
+        unidirectional_dim = int(self._input_dim / 2)
+        forward_combined_dim = get_combined_dim(self._forward_combination,
```
matt-gardner: Slight preference for importing the module and using `util.get_combined_dim` instead of this - it makes the imports nicer above, and it makes it more obvious where this function is coming from as I'm reading this.  It's not a strong preference though, so if you like it better this way, keep it, and I'll try to remember to not bring it up again.

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
```diff
@@ -0,0 +1,230 @@
...
+
+    @overrides
+    def forward(self, # pylint: disable=arguments-differ
+                sequence_tensor: torch.FloatTensor,
+                indices: torch.LongTensor,
+                sequence_mask: torch.LongTensor = None,
+                indices_mask: torch.LongTensor = None) -> torch.FloatTensor:
+
+        # Both of shape (batch_size, sequence_length, embedding_size / 2)
+        forward_sequence, backward_sequence = torch.split(sequence_tensor, int(self._input_dim / 2), -1)
```
matt-gardner: You give `dim=-1` below, where you just give `-1` here.  Might be nice to the reader to give the keyword argument every time.  You could also just do `sequence_tensor.split(...)` instead of `torch.split(sequence_tensor, ...)`.

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
```diff
@@ -0,0 +1,230 @@
...
+        if indices_mask is not None:
+            span_starts *= indices_mask
+            span_ends *= indices_mask
+
+        # We want `exclusive` span starts, so we remove 1 from the forward span starts
+        # as the AllenNLP ``SpanField`` is inclusive.
+        # shape (batch_size, num_spans)
+        exclusive_span_starts = span_starts - 1
+        # shape (batch_size, num_spans)
+        start_sentinel_mask = (exclusive_span_starts == -1).float().unsqueeze(-1)
```
matt-gardner: Is this shape comment right?  You've done an `unsqueeze` here.

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
```diff
@@ -0,0 +1,230 @@
...
+            sequence_lengths = get_lengths_from_binary_sequence_mask(sequence_mask)
+        else:
+            # shape (batch_size), filled with the sequence length size of the sequence_tensor.
+            sequence_lengths = ones_like(sequence_tensor[:, 0, 0]).long() * sequence_tensor.size(1)
+
+        # shape (batch_size, num_spans)
+        end_sentinel_mask = (exclusive_span_ends == sequence_lengths.unsqueeze(-1)).long()
+
+        # Forward Direction: start indices are exclusive. Shape (batch_size, num_spans, input_size / 2)
+        forward_start_embeddings = batched_index_select(forward_sequence, exclusive_span_starts)
```
matt-gardner: How is this not making your tests crash?  Won't you have a `-1` in `exclusive_span_starts` for some instances?  You haven't removed that by this point.  Is `-1` a valid entry for `index_select`?

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
```diff
@@ -0,0 +1,230 @@
...
+        # Forward Direction: end indices are inclusive, so we can just use span_ends.
+        # Shape (batch_size, num_spans, input_size / 2)
+        forward_end_embeddings = batched_index_select(forward_sequence, span_ends)
+
+        # Backward Direction: start indices are actually the end indices, because we
+        # are going backwards, which we want to be inclusive, so we can just use span_starts.
+        # Shape (batch_size, num_spans, input_size / 2)
+        backward_start_embeddings = batched_index_select(backward_sequence, span_starts)
```
matt-gardner: I think this would be less confusing if you called this `backward_end_embeddings`.

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
```diff
@@ -0,0 +1,230 @@
...
+        # Backward Direction: end indices are actually the start indices, which we want to
+        # be exclusive. As we added 1 to the span_ends to make them exclusive, which might have
+        # caused indices equal to the sequence_length to become out of bounds, we multiply
+        # by the inverse of the end_sentinel mask to erase these indices (as we will replace
+        # them anyway in the block below).
+        # Shape (batch_size, num_spans, input_size / 2)
+        backward_end_embeddings = batched_index_select(backward_sequence,
+                                                       exclusive_span_ends * (1 - end_sentinel_mask))
```
matt-gardner: I'd probably do this masking up above, where the mask is computed, for both the span starts and span ends.  It just groups the logic together better, so you don't have as much going on in this part of the code.

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
```diff
@@ -0,0 +1,230 @@
...
+        if self._use_sentinels:
+            # If we're using sentinels, we need to replace all the elements which were
+            # outside the dimensions of the sequence_tensor with either the start sentinel,
+            # or the end sentinel.
+            float_end_sentinel_mask = end_sentinel_mask.float().unsqueeze(-1)
```
matt-gardner: Can you just make the mask a float to start with?  Oh, that's annoying, you need it to be `long` for the `index_select`...  You could just always have it be a float, do the masking for the span indices up above, and recast to long after doing the masking...?

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
```diff
@@ -0,0 +1,230 @@
...
+            backward_end_embeddings = backward_end_embeddings * (1 - float_end_sentinel_mask) \
+                                        + float_end_sentinel_mask * self._end_sentinel
+
+        # Now we combine the forward and backward spans in the manner specified by the
+        # respective combinations and concatenate these representations.
+        # Shape (batch_size, num_spans, forward_combination_dim)
+        forward_spans = combine_tensors(self._forward_combination,
+                                        [forward_start_embeddings, forward_end_embeddings])
+        # Shape (batch_size, num_spans, backward_combination_dim)
+        backward_spans = combine_tensors(self._backward_combination,
```
matt-gardner: This feels really weird to me.  The line looks like I would expect, except the semantics of all of the variables are flipped.  `backward_combination` is backwards, `start_embeddings` is really `end_embeddings`, and `end_embeddings` is really `start_embeddings`.  It seems simpler conceptually to me if `backward_start_embeddings` refers to the start _from the backwards direction_, and `combination` is the same for both directions.

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
```diff
@@ -0,0 +1,230 @@
...
+        backward_start_embeddings = batched_index_select(backward_sequence, span_starts)
+        # Backward Direction: end indices are actually the start indices, which we want to
+        # be exclusive. As we added 1 to the span_ends to make them exclusive, which might have
+        # caused indices equal to the sequence_length to become out of bounds, we multiply
+        # by the inverse of the end_sentinel mask to erase these indices (as we will replace
+        # them anyway in the block below).
+        # Shape (batch_size, num_spans, input_size / 2)
+        backward_end_embeddings = batched_index_select(backward_sequence,
+                                                       exclusive_span_ends * (1 - end_sentinel_mask))
+        if self._use_sentinels:
```
matt-gardner: Are you planning on crashing anywhere if you're not using sentinels, and you give an out-of-bounds span index?  We talked about that at one point, but I don't see it in here.  Looks like you'll just silently get really weird behavior.

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
```diff
@@ -0,0 +1,230 @@
...
+
+    Additionally, the width of the spans can be embedded and concatenated on to the
+    final combination.
+
+    The following other types of representation are supported for both the forward and backward
+    directions, assuming that ``x = span_start_embeddings`` and ``y = span_end_embeddings``.
+
+    ``x``, ``y``, ``x*y``, ``x+y``, ``x-y``, ``x/y``, where each of those binary operations
+    is performed elementwise.  You can list as many combinations as you want, comma separated.
+    For example, you might give ``x,y,x*y`` as the ``combination`` parameter to this class.
```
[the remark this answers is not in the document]
DeNeutoy: I'm going to leave this as it is, because it facilitates doing some span experiments involving differences of spans. For instance, it would allow you to take the forward start representation and the backward "start" representation only, by passing "x", "y".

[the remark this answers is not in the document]
matt-gardner: Which "start" representation do you mean?  The start from the backwards point of view?  That would still work with what I'm suggesting - you just pass in "x", and it uses the "start" from each direction's perspective.

It's not a huge deal - the code works, and the way you've written it matches what you stated in the docstring, that "start" and "end" always refer to the forward direction's perspective.  It's just hard for me to think of it that way, so maybe it'd be easier for a reader to have "start" and "end" always refer to _each direction's_ perspective.  But, that's probably confusing for some readers too, so just do what you think is best.

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
DeNeutoy: No this is good, I was just doing this incrementally and the imports crept up on me.

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
DeNeutoy: Ah - this is a weird edge case. What's happening here is that the index is being adjusted wrt to the batch, so it actually has some value added to it to make it non-negative. If it was in the first element of the batch it would have crashed. I'll add the mask to be explicit.

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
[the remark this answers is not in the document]
DeNeutoy: Oh I was going to change the start/end thing - I just meant about keeping both representations configurable.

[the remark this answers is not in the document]
matt-gardner: :+1:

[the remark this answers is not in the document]
matt-gardner: In that case, you could consider defaulting the backward combination to the forward combination, or doing something so that you don't have to be overly verbose if you want the same combination for both directions.  That might be over-thinking this, though.

PR #829, opened 2018-02-13
log GPU memory

joelgrus:
I tested this on my p2.xlarge, it seems to work, not sure how else to test it

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -25,7 +25,7 @@ from tensorboard import SummaryWriter
 
 from allennlp.common import Params
 from allennlp.common.checks import ConfigurationError
-from allennlp.common.util import peak_memory_mb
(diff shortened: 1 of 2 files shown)


allennlp/training/trainer.py
```diff
@@ -395,6 +395,10 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
         """
         logger.info("Epoch %d/%d", epoch, self._num_epochs - 1)
         logger.info(f"Peak CPU memory usage MB: {peak_memory_mb()}")
+        gpu_memory = gpu_memory_mb()
+        if gpu_memory:
+            logger.info(f"GPU memory usage MB: {gpu_memory}")
```
DeNeutoy: Loop over the returned dict and log the ids?

joelgrus:
ok, this way should be safer

PR #831, opened 2018-02-13
Logical Form Executed and Displayed in Demo

rajasagashe:


rajasagashe:
The main changes are to the following files: 
allennlp/service/predictors/wikitables_parser.py
demo/src/components/WikiTablesComponent.js

The other two are just to allow me to pass a config and archive file as command line arguments.

allennlp/service/predictors/predictor.py
```diff
@@ -67,13 +69,16 @@ def _batch_json_to_instances(self, json_dicts: List[JsonDict]) -> List[Tuple[Ins
         return instances
 
     @classmethod
-    def from_archive(cls, archive: Archive, predictor_name: str) -> 'Predictor':
+    def from_archive(cls, archive: Archive, predictor_name: str, config_file=None) -> 'Predictor':
```
matt-gardner: Why do you need a separate config file here?  The archive already has a config file in it.

allennlp/service/predictors/wikitables_parser.py
```diff
@@ -36,3 +43,48 @@ def _json_to_instance(self, json_dict: JsonDict) -> Tuple[Instance, JsonDict]:
                                                          tokenized_question=tokenized_question)
         extra_info = {'question_tokens': tokenized_question}
         return instance, extra_info
+
+    @overrides
+    def predict_json(self, inputs: JsonDict, cuda_device: int = -1) -> JsonDict:
+        instance, return_dict = self._json_to_instance(inputs)
+        outputs = self._model.forward_on_instance(instance, cuda_device)
+
+        os.makedirs('data', exist_ok=True)
```
matt-gardner: I'd probably put this code in the constructor instead of here.  But, maybe it's not that big of a deal, because you're only calling `wget` if the file isn't there...  Still, seems nicer to take this hit at startup instead of during serving.

allennlp/commands/serve.py
```diff
@@ -71,13 +71,18 @@ def add_subparser(self, name: str, parser: argparse._SubParsersAction) -> argpar
                 name, description=description, help='Run the web service and demo.')
 
         subparser.add_argument('--port', type=int, default=8000)
-
+        subparser.add_argument('--archive-path', type=str, help='path to trained archive file')
```
matt-gardner: Instead of this, I'd recommend just changing line 36 to your archived model.  I just put the test model there to have something I could check in that would pass tests.  Do you need something that wouldn't be fixed just by changing that path?

allennlp/service/predictors/wikitables_parser.py
```diff
@@ -36,3 +43,48 @@ def _json_to_instance(self, json_dict: JsonDict) -> Tuple[Instance, JsonDict]:
...
+                            test_data_filename,
+                            logical_form_filename,
+                            table_dir,
+                            ])
+        run(command, shell=True)
+
+        denotations_file = 'data/logical_forms_denotations.tsv'
+        with open(denotations_file) as f:
+            line = f.readline().split('\t')
+            outputs['answer'] = line[1] if len(line) > 1 else line[0]
```
matt-gardner: This is great!  Definitely a bit hacky, because of how we have to do this, but great to be able to get actual answers in the demo!

allennlp/service/predictors/predictor.py
rajasagashe: This was since the archive config used the preprocessed reader, while the demo needs the normal one.

allennlp/service/predictors/wikitables_parser.py
rajasagashe: done

allennlp/service/predictors/predictor.py
matt-gardner: Yeah, it's a bit annoying, but I take a trained model, unpack the archive, modify the file, then repack the archive.  It'd probably be easy to write a simple script to do this...

allennlp/service/predictors/wikitables_parser.py
```diff
@@ -36,3 +57,42 @@ def _json_to_instance(self, json_dict: JsonDict) -> Tuple[Instance, JsonDict]:
...
+        table_dir = SEMPRE_DIR + 'csv/'
+        os.makedirs(table_dir, exist_ok=True)
+        table_filename = 'context.csv'
+        with open(table_dir + table_filename, 'w', encoding='utf-8') as f:
+            f.write(inputs["table"])
+
+        test_record = ('(example (id %s) (utterance %s) (context (graph tables.TableKnowledgeGraph %s))'
+                       '(targetValue (list (description "6"))))' % ('nt-0', inputs['question'],
+                                                                     table_filename))
+        test_data_filename = SEMPRE_DIR + 'data.examples'
```
matt-gardner: `os.path.join` here.

allennlp/service/predictors/wikitables_parser.py
```diff
@@ -36,3 +57,42 @@ def _json_to_instance(self, json_dict: JsonDict) -> Tuple[Instance, JsonDict]:
...
+        with open(logical_form_filename, 'w') as f:
+            f.write(outputs['logical_form'] + '\n')
+
+        table_dir = SEMPRE_DIR + 'csv/'
+        os.makedirs(table_dir, exist_ok=True)
+        table_filename = 'context.csv'
+        with open(table_dir + table_filename, 'w', encoding='utf-8') as f:
+            f.write(inputs["table"])
+
+        test_record = ('(example (id %s) (utterance %s) (context (graph tables.TableKnowledgeGraph %s))'
```
matt-gardner: Adding a comment here saying that we need to get the utterance and the table into SEMPRE, but the target value will be ignored, would be good.

allennlp/service/predictors/wikitables_parser.py
```diff
@@ -36,3 +57,42 @@ def _json_to_instance(self, json_dict: JsonDict) -> Tuple[Instance, JsonDict]:
                                                          tokenized_question=tokenized_question)
         extra_info = {'question_tokens': tokenized_question}
         return instance, extra_info
+
+    @overrides
+    def predict_json(self, inputs: JsonDict, cuda_device: int = -1) -> JsonDict:
+        instance, return_dict = self._json_to_instance(inputs)
+        outputs = self._model.forward_on_instance(instance, cuda_device)
+
+        logical_form_filename = os.path.join(SEMPRE_DIR, 'logical_forms.txt')
```
matt-gardner: I'd recommend pulling out this answer-getting logic into its own method, with a clear name and signature (e.g., pass in `outputs['logical_form']`, `inputs['table']`, etc.).  It'd make this code easier to read and understand.  You might give a note there, also, that the jar that we have isn't optimal for this use case - we're using a script designed for computing accuracy, and just pulling out a piece of it.  Writing a new entry point to the jar that's tailored for this use would be cleaner.  Feel free to put that in a `TODO(matt):` block.

allennlp/service/predictors/wikitables_parser.py
```diff
@@ -36,3 +57,42 @@ def _json_to_instance(self, json_dict: JsonDict) -> Tuple[Instance, JsonDict]:
...
+                            test_data_filename,
+                            logical_form_filename,
+                            table_dir,
+                            ])
+        run(command, shell=True)
+
+        denotations_file = SEMPRE_DIR + 'logical_forms_denotations.tsv'
+        with open(denotations_file) as f:
+            line = f.readline().split('\t')
+            outputs['answer'] = line[1] if len(line) > 1 else line[0]
```
matt-gardner: Can you also modify the predictor test to make sure that we're getting an answer in the output?

allennlp/service/predictors/wikitables_parser.py
rajasagashe: done.

allennlp/service/predictors/wikitables_parser.py
rajasagashe: done.

allennlp/service/predictors/wikitables_parser.py
rajasagashe: done.

allennlp/service/predictors/wikitables_parser.py
rajasagashe: done.

allennlp/service/predictors/wikitables_parser.py
```diff
@@ -62,37 +62,48 @@ def _json_to_instance(self, json_dict: JsonDict) -> Tuple[Instance, JsonDict]:
...
 
-        denotations_file = SEMPRE_DIR + 'logical_forms_denotations.tsv'
-        with open(denotations_file) as f:
-            line = f.readline().split('\t')
-            outputs['answer'] = line[1] if len(line) > 1 else line[0]
-
-        return_dict.update(outputs)
-        return sanitize(return_dict)
+        denotations_file = os.path.join(SEMPRE_DIR, 'logical_forms_denotations.tsv')
+        with open(denotations_file) as file:
```
matt-gardner: `file` was a built-in function in python 2, which is why it's showing up as orange here.  It's not a built-in in python 3, so pylint passes, but I'd still recommend avoiding using it, to appease github's rendering engine.

matt-gardner:
And the tests are failing on CI because there is no java in the docker image.  Looks like we need to add a line in here that installs java: https://github.com/allenai/allennlp/blob/740f4fbb961425c101c0b4461d84fa9a32905f39/Dockerfile#L20-L34

rajasagashe:
That was since I accidentally added it into the other pull request as well and there was a conflict. #848

Dockerfile
```diff
@@ -21,6 +21,7 @@ RUN apt-get update --fix-missing && apt-get install -y \
     bzip2 \
     ca-certificates \
     curl \
+    default-jre \
```
matt-gardner: Looks like we need java 8, not whatever java version this gives us.

rajasagashe:
I was about to add the information from this file:
https://github.com/dockerfile/java/blob/master/oracle-java8/Dockerfile

matt-gardner:
That's going to fail for the same reasons that my last attempt failed (and maybe other reasons too).

rajasagashe:
Thanks!

PR #836, opened 2018-02-13
make using pos tags configurable in the PTB reader

DeNeutoy:


allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -35,14 +35,19 @@ class PennTreeBankConstituencySpanDatasetReader(DatasetReader):
         We use this to define the input representation for the text.  See :class:`TokenIndexer`.
         Note that the `output` tags will always correspond to single token IDs based on how they
         are pre-tokenised in the data file.
+    use_pos_tags : ``bool``, optional, (default = ``True``)
+        Whether or not the instance should contain gold pos tags
```
matt-gardner: Capitalize "POS" when it's in plain text?  Not a big deal if you don't care.

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -52,12 +57,15 @@ def _read(self, file_path):
         logger.info("Reading instances from lines in file at: %s", file_path)
 
         for parse in BracketParseCorpusReader(root=directory, fileids=[filename]).parsed_sents():
-            yield self.text_to_instance(parse.leaves(), [x[1] for x in parse.pos()], parse)
+            if self._use_pos_tags:
+                yield self.text_to_instance(parse.leaves(), [x[1] for x in parse.pos()], parse)
```
matt-gardner: Slight preference for assigning to a `pos_tags` variable in the if/else clause, and having just a single `yield`.  That way, if the `text_to_instance` method changes later, you only need to change it in one place.

Issue #837, opened 2018-02-13
Obtuse exception when a batch has only an empty sentence


schmmd:
Data (`sentences.txt`):

```
Call me Ishmael.
Some years ago- never mind how long precisely- having little or no money in my purse, and nothing particular to interest me on shore, I thought I would sail about a little and see the watery part of the world.
It is a way I have of driving off the spleen and regulating the circulation.
Whenever I find myself growing grim about the mouth; whenever it is a damp, drizzly November in my soul; whenever I find myself involuntarily pausing before coffin warehouses, and bringing up the rear of every funeral I meet; and especially whenever my hypos get such an upper hand of me, that it requires a strong moral principle to prevent me from deliberately stepping into the street, and methodically knocking people's hats off- then, I account it high time to get to sea as soon as I can.
This is my substitute for pistol and ball.
With a philosophical flourish Cato throws himself upon his sword; I quietly take to the ship.
There is nothing surprising in this.
If they but knew it, almost all men in their degree, some time or other, cherish very nearly the same feelings towards the ocean with me.
There now is your insular city of the Manhattoes, belted round by wharves as Indian isles by coral reefs- commerce surrounds it with her surf.
Right and left, the streets take you waterward.
Its extreme downtown is the battery, where that noble mole is washed by waves, and cooled by breezes, which a few hours previous were out of sight of land.
Look at the crowds of water-gazers there.
Circumambulate the city of a dreamy Sabbath afternoon.
Go from Corlears Hook to Coenties Slip, and from thence, by Whitehall, northward.
What do you see?- Posted like silent sentinels all around the town, stand thousands upon thousands of mortal men fixed in ocean reveries.
Some leaning against the spiles; some seated upon the pier-heads; some looking over the bulwarks of ships from China; some high aloft in the rigging, as if striving to get a still better seaward peep.
But these are all landsmen; of week days pent up in lath and plaster- tied to counters, nailed to benches, clinched to desks.
How then is this? Are the green fields gone? What do they here?
But look! here come more crowds, pacing straight for the water, and seemingly bound for a dive.
Strange! Nothing will content them but the extremest limit of the land; loitering under the shady lee of yonder warehouses will not suffice.
No.
They must get just as nigh the water as they possibly can without falling And there they stand- miles of them- leagues.
Inlanders all, they come from lanes and alleys, streets avenues- north, east, south, and west.
Yet here they all unite.
Tell me, does the magnetic virtue of the needles of the compasses of all those ships attract them thither?
[...]

PR #842, opened 2018-02-13
Handle batches with only a single empty sentence

schmmd:
Fixes https://github.com/allenai/allennlp/issues/837

This also fixes some type annotations.

### DIFF
--- a/allennlp/commands/elmo.py
+++ b/allennlp/commands/elmo.py
@@ -219,7 +219,10 @@ class ElmoEmbedder():
 
         return elmo_embeddings
 
-    def embed_sentences(self, sentences: Iterable[List[str]], batch_size: int) -> Iterable[torch.Tensor]:
+    def embed_sentences(
+                self,
+                sentences: Iterable[List[str]],
+                batch_size: int = DEFAULT_BATCH_SIZE) -> Iterable[torch.Tensor]:
         """
         Computes the ELMo embeddings for a iterable of sentences.
 
--- a/allennlp/commands/elmo.py
+++ b/allennlp/commands/elmo.py
@@ -269,7 +269,7 @@ class ElmoEmbedder():
 
         # Tokenizes the sentences.
         sentences = [line.strip() for line in input_file]
-        split_sentences = [sentence.split() for sentence in sentences]
+        split_sentences = [sentence.strip().split() for sentence in sentences if sentence.strip()]
         # Uses the sentence as the key.
         embedded_sentences = zip(sentences, self.embed_sentences(split_sentences, batch_size))
 
--- a/allennlp/commands/elmo.py
+++ b/allennlp/commands/elmo.py
@@ -123,6 +123,9 @@ class ElmoEmbedder():
 
         self.cuda_device = cuda_device
 
+    def empty_embedding(self) -> numpy.ndarray:
+        return numpy.zeros((3, 0, 1024))
+
     def batch_to_ids(self, batch: List[List[str]]) -> torch.Tensor:
         """
         Converts a batch of tokenized sentences to a tensor representing the sentences with encoded characters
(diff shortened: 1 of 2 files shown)


DeNeutoy:
My 2 cents: If you accidentally added a new line where there wasn't one, you want to 1) be warned about it and 2) have this elmo dump not crash and 3) not have a tensor filled with zeros to represent a `\n` character.

Instead, it should warn the user that they provided an empty line, print the line number and remove it from the embedding.

matt-peters:
What Mark said.  Since we are keying the file on the sentence string, it does no harm to ignore empty sentences, and a vector of all zeros in the output file is strange.

schmmd:
@DeNeutoy my concern is that I want the user to be able to `zip(embeddings, sentences)`.  If I slip a sentence out, then they cannot rely on this working and I need to change the contracts.  Any suggestions here?

I might have created the wrong tensor.  My goal was to create a tensor that represents all the embeddings for the input.  In this case the input had 0 words, so that tensor would have shape `(3, 0, 1024)`.

I'm happy to additionally strip empty lines out of `embed_file`--I think that would be a good change since the output is a dict.

schmmd:
@matt-peters it does no harm in `embed_file` (and I'll make that change) but it does harm in `embed_sentence` (I do not want to make that change there).

matt-peters:
Right, I see you point now.  Maybe returning `None` instead of the `(3, 0, 1024)` tensor is better?

schmmd:
@matt-peters why would returning `None` be better?  I think you'll still get a `(3, 0, 1024)` tensor if you have an empty input as part of a batch--I'm just handling the case where a batch has a single empty sentence (which currently crashes).

schmmd:
I would love to move that special-case handling for batches with a single empty sentence lower in AllenNLP--but I don't know how to.

schmmd:
I pushed a change that strips empty lines from the input file.  @matt-peters if you feel strongly about it, I can change the code to warn when there are empty lines instead of ignoring them.

matt-peters:
Makes sense now, your original approach is good.  We can look at the single empty sentence issue maybe down the road.

allennlp/commands/elmo.py
```diff
@@ -211,15 +214,26 @@ def embed_batch(self, batch: List[List[str]]) -> List[torch.Tensor]:
         """
         elmo_embeddings = []
 
-        embeddings, mask = self.batch_to_embeddings(batch)
-        for i in range(len(batch)):
-            length = int(mask[i, :].sum())
-            sentence_embeds = embeddings[i, :, :length, :].data.cpu().numpy()
-            elmo_embeddings.append(sentence_embeds)
+        # Batches will only an empty sentence will throw an exception inside AllenNLP, but rather than throw an
```
DeNeutoy: This sentence is a bit messed up.

allennlp/commands/elmo.py
```diff
@@ -211,15 +214,26 @@ def embed_batch(self, batch: List[List[str]]) -> List[torch.Tensor]:
...
+                # Slicing the embedding :0 throws an exception so we need to special case for empty sentences.
+                if length == 0:
+                    elmo_embeddings.append(empty_embedding())
+                else:
+                    elmo_embeddings.append(embeddings[i, :, :length, :].data.cpu().numpy())
 
         return elmo_embeddings
 
-    def embed_sentences(self, sentences: Iterable[List[str]], batch_size: int) -> Iterable[torch.Tensor]:
+    def embed_sentences(
```
DeNeutoy: We'd normally format this as 
```
def embed_sentences(self,
                    param1: type,
                    param2: type)
```

DeNeutoy:
Actually, I think this line is now incorrect: https://github.com/allenai/allennlp/pull/842/files#diff-cca0c8911d8dcaf0850e4be6d0a669e4R280

`split_sentences` is no longer guaranteed to be the same length as `sentences`, if you have extra new lines, and `zip([1,2,3], [1,2]) == [(1,1), (2,2)]`, so if you have empty lines, the hdf5 keys will be  for the wrong tensors.

schmmd:
@DeNeutoy -- you're right, I made that change in the wrong place.  Fix oncoming.

matt-gardner:
Can you also add a test that makes sure this case is handled correctly?

schmmd:
Actually, scratch my last comment.  I got very confused by the GitHub diff.  I'll look closer.

schmmd:
Yep--that's my starting point ;-)

schmmd:
Mark and I discussed a bit offline, but neither of us moved in our opinions and Mark didn't want to discuss it more.  I'm going to merge this now, but I'm happy to discuss it more as there's a lot I have yet to learn about deep learning, and while my approach might make sense for me, it might not make sense to others.  Just open an issue and assign me.

The present state is that:

* When processing a file, empty lines are stripped.  I'm happy to add a warning for this case and when there are duplicates if anyone would like such a warning--just cut an issue and assign me.

* In `embed_sentence` empty sentences are returned with shape `(3, 0, 1024)` instead of throwing an exception, as they did before.

```
$ ee.embed_sentence("This is".split()).shape
(3, 2, 1024)

$ ee.embed_sentence("This".split()).shape
(3, 1, 1024)

$ ee.embed_sentence("".split()).shape
(3, 0, 1024)
```

PR #845, opened 2018-02-14
Sort DPD output by length before selecting logical forms

matt-gardner:
I also did a little cleanup of the docstrings and other stuff.  I changed the DPD fixture, so that the shortest logical form was no longer first, and confirmed that the test fails if you don't sort the DPD output.


allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
```diff
@@ -159,12 +168,17 @@ def _read(self, file_path: str):
...
                         dpd_file = gzip.open(dpd_output_filename)
-                        sempre_forms = []
-                        for dpd_line in dpd_file:
-                            sempre_forms.append(dpd_line.strip().decode('utf-8'))
-                            if self._max_dpd_tries and len(sempre_forms) >= self._max_dpd_tries:
-                                # TODO(mattg): might want to sort by length here before truncating...
-                                break
+                        if self._sort_dpd_logical_forms:
+                            sempre_forms = [dpd_line.strip().decode('utf-8') for dpd_line in dpd_file]
+                            sempre_forms.sort(key=lambda x: len(x))
```
pdasigi: I agree with pylint. This could be `sempre_forms.sort(key=len)`.

PR #847, opened 2018-02-14
Format json

schmmd:


schmmd:
The json formatting was bothering me, so I ran some of it through `jq '.'`.  I also removed a comma from `bidaf.json` to make it valid json (the trailing comma is valid hocon).

If you have any concerns, I'll just close this PR.

PR #848, opened 2018-02-14
Ability to Disable Linking Features

rajasagashe:


allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -231,7 +234,10 @@ def forward(self,  # type: ignore
         question_table_similarity_max_score, _ = torch.max(question_table_similarity, 2)
         # (batch_size, num_entities, num_question_tokens, num_features)
         linking_features = table['linking']
-        feature_scores = self._linking_params(linking_features).squeeze(3)
+        if self._linking_params is not None:
+            feature_scores = self._linking_params(linking_features).squeeze(3)
```
matt-gardner: I'd probably just do `linking_scores += feature_scores` after this line, instead of creating a tensor of zeros that never gets used.  You can have a check below, where `feature_scores` gets added to the output dict, for `self._linking_params is not None`, like you have here.

rajasagashe: done.

matt-gardner:
Looks great, except I think there's no way to disable or change the linking features in the dataset reader.  Looks like we forgot to pass in the `_linking_feature_extractors` to the `KnowledgeGraphField` in `WikiTablesDatasetReader.text_to_instance`.  Can you add that in as well?

rajasagashe:
Sure, also the demo breaks if no features are passed so fixing that in WikiTablesComponent.js.

rajasagashe:
@matt-gardner I think this ready now.

allennlp/service/server_flask.py
```diff
@@ -229,6 +229,8 @@ def predict(model_name: str) -> Response:  # pylint: disable=unused-variable
                     verbs.append({"verb": verb["verb"], "description": verb["description"]})
 
             log_blob["outputs"]["verbs"] = verbs
+        elif model_name == "wikitables-parser":
+            log_blob['outputs']['logical_form'] = prediction['logical_form']
```
matt-gardner: Can you add the answer here, too?  That's part of the predictor output now, isn't it?

rajasagashe: That's added in #831.

PR #849, opened 2018-02-14
Pass through encoder

DeNeutoy:
- Add a pass through encoder so you can remove a seq2seq encoder via the config
- Fix some top level seq2seqEncoder docs
- Make MultiHeadAttention build-able from Params.

📂 📋 🗂 📆


allennlp/modules/seq2seq_encoders/multi_head_self_attention.py
```diff
@@ -154,3 +155,19 @@ def forward(self,  # pylint: disable=arguments-differ
         # shape (batch_size, timesteps, input_size)
         outputs = self._output_projection(outputs)
         return outputs
+
+    @classmethod
+    def from_params(cls, params: Params) -> 'MultiHeadSelfAttention':
+        num_heads = params.pop_int('num_heads', 3)
```
joelgrus: why use a default here if there's not one in the constructor?

Issue #765, opened 2018-02-03
Archiving models at different points of training


eunsol:
Currently the model is only archived at the end of the training, if I'm understanding correctly. When the model starts to overfit, I just quit the training, and then no archived model file is formed, and it's a hassle to evaluate / predict from the model. Basically one has to manually build model.tar.gz file to predict from it. Overall it would be nice to be able to predict from different timestamps without creating model.tar.gz for each.

schmmd:
@eunsol would https://github.com/allenai/allennlp/pull/734 fix your problem?  It would allow using directories in place of archives.

eunsol:
would it allow predicting from different timestamps by specifying it? ideally, it will look something like 
"python -m allennlp.run predict --archive-folder ./where/model_lives/ --timestamp 4"

schmmd:
No--that would require additional changes.

joelgrus:
that's a pretty small additional change. it's a good idea

schmmd:
I'll add a `--weights-file` flag.

schmmd:
@eunsol we just merged a PR that addresses the issue you brought up.  Let us know if you have any further problems!

Please note that this was fixed on master but is yet unreleased.  We expect to have a new release next week.

PR #854, opened 2018-02-14
Add a weights-file flag to evaluate and predict

schmmd:
This change allows users to make predictions and evaluate from the output folder of an in-progress training.  With `--weights-file` you can choose which iteration you would like to make predictions or evaluate from.

Fixes https://github.com/allenai/allennlp/issues/765


schmmd:
@matt-gardner in this pull request, I rename `model_params.json` to `config.json` so the training output folder can be used for predictions and evaluations.  As far as I could tell, this naming was arbitrary but please let me know if you're aware of other consequences here.

allennlp/commands/train.py
```diff
@@ -182,7 +182,7 @@ def train_model(params: Params, serialization_dir: str, file_friendly_logging: b
     handler.setFormatter(logging.Formatter('%(asctime)s - %(levelname)s - %(name)s - %(message)s'))
     logging.getLogger().addHandler(handler)
     serialization_params = deepcopy(params).as_dict(quiet=True)
-    with open(os.path.join(serialization_dir, "model_params.json"), "w") as param_file:
+    with open(os.path.join(serialization_dir, "config.json"), "w") as param_file:
```
matt-gardner: How about we make `_CONFIG_NAME` public in `models/archival.py`, and use that here, instead of having a string that we'd need to change in two places?

allennlp/commands/evaluate.py
```diff
@@ -62,6 +62,9 @@ def add_subparser(self, name: str, parser: argparse._SubParsersAction) -> argpar
         evaluation_data_file.add_argument('--evaluation_data_file',
                                           type=str,
                                           help=argparse.SUPPRESS)
+        subparser.add_argument('--weights-file',
+                               type=str,
+                               help='overrides which weights file to use')
```
matt-gardner: Should this be a full path or a relative path in the serialization directory?  Currently it needs to be a full path.  It's fine with me either way, as long as it's documented here.

allennlp/models/archival.py
```diff
@@ -55,7 +55,7 @@ def archive_model(serialization_dir: str,
         logger.error("weights file %s does not exist, unable to archive model", weights_file)
         return
 
-    config_file = os.path.join(serialization_dir, "model_params.json")
+    config_file = os.path.join(serialization_dir, "config.json")
```
matt-gardner: You could use the constant here, too, and in the other places you changed `"model_params.json"` to `"config.json"`.

allennlp/commands/evaluate.py
schmmd: I like it being a full path, because then it's easy to grab the right file with autocomplete.

matt-gardner: Yeah, that's seems reasonable, I'd just be sure to note that in this help string.

schmmd:
@matt-gardner I changed the argument description to "a path that overrides which weights file to use".  Let me know if you have a better suggestion, but I think path implies it's not relative to the serialization directory.

allennlp/models/archival.py
```diff
@@ -129,12 +134,17 @@ def load_archive(archive_file: str, cuda_device: int = -1, overrides: str = "")
         overrides = json.dumps(combined_hocon)
 
     # Load config
-    config = Params.from_file(os.path.join(serialization_dir, _CONFIG_NAME), overrides)
+    config = Params.from_file(os.path.join(serialization_dir, CONFIG_NAME), overrides)
```
DeNeutoy: Is this backwards compatible if `archive_file` is a directory rather than a tar.gz file? I'm not sure it is.

scripts/write_srl_predictions_to_conll_format.py
```diff
@@ -19,7 +19,7 @@ def main(serialization_directory, device):
         The device to run the evaluation on.
     """
 
-    config = Params.from_file(os.path.join(serialization_directory, "model_params.json"))
```
DeNeutoy: If you could make this backward compatible that would be great, maybe just with a `try except` or something.

allennlp/models/archival.py
DeNeutoy: Also if this is a breaking change for the release that's fine, it's just quite a big thing to break accidentally so I was just checking

schmmd: Hi @DeNeutoy I don't understand.  What exactly are you worried about being backwards incompatible?  The line you referenced only switches from a private constant to a public one.

allennlp/commands/train.py
```diff
@@ -51,6 +51,7 @@
 from allennlp.data.dataset_readers.dataset_reader import DatasetReader
 from allennlp.data.iterators.data_iterator import DataIterator
 from allennlp.models.archival import archive_model
+from allennlp.models.archival import CONFIG_NAME
```
joelgrus: combine these two imports

schmmd:
@DeNeutoy I think you're asking whether changing `config.json` to `model_params.json` is backwards compatible, and it is not.

matt-gardner:
Yeah, it's definitely not, but I don't think it's something to worry about.  Nothing in the code actually relied on that filename, except for loading an archive from a directory, and actually, I think that code was broken, because it was still looking for `config.json`, not `model_params.json`.  So, really, I don't think there's anything that is backwards incompatible here.

DeNeutoy:
Ah right - yeah I wasn't sure if using a directory as an archive was really used much. @schmmd you might also want to regenerate the text fixtures to make them up to date with the new archive format.

schmmd:
The `model_params.json` was previously renamed to `config.json` when creating a tarball so I don't think there is anything to update.

```
$ tar -tzvf tests/fixtures/bidaf/serialization/model.tar.gz 
-rw-r--r--  0 markn  staff    2465 Feb  1 10:35 config.json
-rw-r--r--  0 markn  staff   40813 Feb  1 10:35 weights.th
drwxr-xr-x  0 markn  staff       0 Dec  4 09:47 vocabulary/
-rw-r--r--  0 markn  staff      14 Feb  1 10:35 vocabulary/non_padded_namespaces.txt
-rw-r--r--  0 markn  staff     128 Dec  4 09:47 vocabulary/token_characters.txt
-rw-r--r--  0 markn  staff     996 Feb  1 10:35 vocabulary/tokens.txt
-rw-r--r--  0 markn  staff    2465 Feb  1 10:35 config.json
-rw-r--r--  0 markn  staff   40813 Feb  1 10:35 weights.th
drwxr-xr-x  0 markn  staff       0 Dec  4 09:47 vocabulary/
-rw-r--r--  0 markn  staff      14 Feb  1 10:35 vocabulary/non_padded_namespaces.txt
-rw-r--r--  0 markn  staff     128 Dec  4 09:47 vocabulary/token_characters.txt
-rw-r--r--  0 markn  staff     996 Feb  1 10:35 vocabulary/tokens.txt
```

```
$ tar -tzvf tests/fixtures/coref/serialization/model.tar.gz  
-rw-r--r--  0 markn  staff    2401 Feb  1 10:35 config.json
-rw-r--r--  0 markn  staff   54595 Feb  1 10:35 weights.th
drwxr-xr-x  0 markn  staff       0 Jan  4 08:12 vocabulary/
-rw-r--r--  0 markn  staff      14 Feb  1 10:35 vocabulary/non_padded_namespaces.txt
-rw-r--r--  0 markn  staff     500 Feb  1 10:35 vocabulary/tokens.txt
```

matt-gardner:
I think there's nothing to regenerate - this isn't changing the content of the model archive at all, it's just making the arbitrary "model_params.json" that we save to disk have a matching filename with how we copy it to the archive.  We currently write out "model_params.json", then copy it to "config.json" inside the archive.  This is just changing the name of the original file that we wrote out, which nothing was relying on except our code that copying it to "config.json" in the archive.

DeNeutoy:
Oh my bad. 👍

PR #859, opened 2018-02-15
Update pip instructions

schmmd:


schmmd:
@cristipp better?  Unfortunately I cannot just give the PyTorch command because there are a few options that could be configured--CUDA version and platform (Mac, Linux, etc.).

cristipp:
Better! Thanks.

Issue #770, opened 2018-02-05
Give a more helpful error message when model loading fails during training


matt-gardner:
https://github.com/allenai/allennlp/issues/756#issuecomment-362942024 brings up an error that I've seen somewhat frequently.  When you're still iterating on your model, it's really easy to run into a situation where allennlp tries to continue training from a previous epoch of a different model configuration.  The error message is particularly unhelpful in these situations.  It's even worse that the error message only appears _after_ you've loaded your data and computed the vocabulary, because these might take several minutes.

At the beginning of the `train` command (before loading any data), we should check if the `serialization` directory exists.  If it does, there are a few (not mutually-exclusive) options:
1. Check if the saved model parameters there match the current model parameters.  If they do, log a big message that we'll be continuing training from where it left off, noting that if you've changed model code things might crash later with cryptic messages.  If they don't, crash with an informative error message.
2. Add an `--overwrite-results` flag, or something, that always removes any existing serialization directory before training.
3. Catch a `RuntimeError` when we call `_restore_checkpoint` in `Trainer`, and change it into a more informative error message, suggesting using `--overwrite-results`, or something.

Actually, I'd probably recommend doing all three of those.

schmmd:
I would have an explicit `--continue` flag.  Without this flag, the trainer would throw an exception if the output directory already exists.  With this flag, the trainer would throw an exception if the output directory existed and the parameters don't match.

I'm generally confused when a training continues from something that already exists and this parameter would help make it explicit.  I would likely miss the big message.

I wouldn't have an `--overwrite-results` flag.  I would instead rely on the user deleting the directory themselves.  This avoids potential permissions issues or the wrong directory passed in (`--output /`).

matt-gardner:
Yeah, that sounds reasonable to me.  Continuing training is the odd case, not the usual case, so having the flag for that case is probably right.

Not having some kind of `--overwrite` flag also seems fine - I was thinking mostly of someone hitting up in their shell and running the previous command to try training again, to see if they fixed a bug - that's a really common pattern.  But adding `rm -rf /path/to/output && [allennlp command]` is just as easy as adding `--overwrite-results`, and safer for our code.  You're right.

joelgrus:
+1, this seems like a good solution to me

PR #861, opened 2018-02-15
Add a --continue flag and improve warning messages

schmmd:
* Adds a `--continue` flag, which is required to continue training.
* Logs a nice message when there's an exception in `_restore_checkpoint`.

Fixes https://github.com/allenai/allennlp/issues/770


schmmd:
I misread something in the original issue.  I like what was written much more than what I did, so I have another change coming.

allennlp/commands/train.py
```diff
@@ -104,6 +110,14 @@ def train_model_from_args(args: argparse.Namespace):
     # Import any additional modules needed (to register custom classes)
     for package_name in args.include_package:
         import_submodules(package_name)
+
+    if not args.cont and os.path.exists(args.serialization_dir):
+        raise Exception(f"Serialization directory ({args.serialization_dir}) already exists.  "
```
schmmd: Should I instead `logger.error` and then `sys.exit(1)`?  The output would be more friendly.

allennlp/commands/train.py
```diff
@@ -172,7 +186,40 @@ def train_model(params: Params, serialization_dir: str, file_friendly_logging: b
...
+                               f"but not in the training config.")
+                fail = True
+            for key in flat_params.keys():
+                if flat_params.get(key, None) != flat_loaded.get(key, None):
+                    logger.warning(f"Value for '{key}' in training configuration does not match that the value in "
+                                   f"the serialization directory we're continuing from: "
+                                   f"{flat_params[key]} != {flat_loaded[key]}")
+                    fail = True
+            if fail:
+                raise Exception("Training configuration does not match the configuration we're continuing from.")
```
schmmd: Should I instead `logger.error` and then `sys.exit(1)`?

schmmd:
@matt-gardner I have a couple of questions--would you have a look?

I'm adding tests for when the serialization files mismatch.

schmmd:
If I stick with throwing exceptions, I'll change them to `ConfigurationException`.

allennlp/commands/train.py
```diff
@@ -172,7 +186,41 @@ def train_model(params: Params, serialization_dir: str, file_friendly_logging: b
     """
     prepare_environment(params)
 
-    os.makedirs(serialization_dir, exist_ok=True)
+    if os.path.exists(serialization_dir):
+        logger.info(f"Continuing from prior training at {serialization_dir}.")
```
matt-gardner: Can you pull this whole block out into its own method?  Something like `check_params_match(serialization_dir, params)`?  It makes this method not so heavy, and it is easier to test.

allennlp/commands/train.py
```diff
@@ -172,7 +186,41 @@ def train_model(params: Params, serialization_dir: str, file_friendly_logging: b
...
+
+        continued_config_file = os.path.join(serialization_dir, CONFIG_NAME)
+        if not os.path.exists(continued_config_file):
+            logger.warning("Continuing from prior training, but no config.json was found.")
```
matt-gardner: Again here I'd just crash.  If something unexpected happens, we should fail explicitly instead of silently.

allennlp/commands/train.py
```diff
@@ -172,7 +186,41 @@ def train_model(params: Params, serialization_dir: str, file_friendly_logging: b
...
+        else:
+            loaded_params = Params.from_file(continued_config_file)
+
+            # Check whether any of the training configuration differs from the configuration we are resuming.
+            # If so, warn the user that training may fail.
+            fail = False
+            flat_params = params.as_flat_dict()
+            flat_loaded = loaded_params.as_flat_dict()
+            for key in flat_params.keys() - flat_loaded.keys():
+                logger.warning(f"Key '{key}' found in training configuration but not in the serialization "
```
matt-gardner: These should be `logger.error` instead of `logger.warning`.

allennlp/training/trainer.py
```diff
@@ -627,7 +627,13 @@ def train(self) -> Dict[str, Any]:
         """
         Trains the supplied model with the supplied parameters.
         """
-        epoch_counter, validation_metric_per_epoch = self._restore_checkpoint()
+        try:
+            epoch_counter, validation_metric_per_epoch = self._restore_checkpoint()
+        except RuntimeError:
+            logger.error("Could not continue training from the checkpoint.  Did you mean to output to a different "
```
matt-gardner: The stack trace from the `raise` below will bury this message.  I'd probably print the stack trace, then raise a new error with this nicer message (also saying that the `restore_checkpoint` stack trace is available above, in case it is needed for some reason).

tests/commands/train_test.py
```diff
@@ -163,10 +164,19 @@ def test_other_modules(self):
             main()
 
         # Now add the --include-package flag and it should work.
-        sys.argv.extend(["--include-package", 'testpackage'])
+        # We also need to add --continue since the output directory already exists.
+        sys.argv.extend(["--continue", "--include-package", 'testpackage'])
 
         main()
 
+        # Rewrite out config file, but change a value
```
matt-gardner: Periods here and below.  (And if we can get pylint to check this automatically, I'm all for it - I'm pretty sure that Google's linter would catch these.)

allennlp/commands/train.py
```diff
@@ -77,6 +77,12 @@ def add_subparser(self, name: str, parser: argparse._SubParsersAction) -> argpar
                                    type=str,
                                    help=argparse.SUPPRESS)
 
+        subparser.add_argument('-c', '--continue',
+                               action='store_true',
+                               default=False,
+                               dest="cont",
```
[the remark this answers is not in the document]
schmmd: No.  It's a keyword.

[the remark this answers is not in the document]
matt-gardner: Then `continue_` is better than `cont` - this is relatively common practice for when you want to use the name `type` as a variable, for instance.

allennlp/commands/train.py
```diff
@@ -104,6 +110,14 @@ def train_model_from_args(args: argparse.Namespace):
     # Import any additional modules needed (to register custom classes)
     for package_name in args.include_package:
         import_submodules(package_name)
+
+    if not args.cont and os.path.exists(args.serialization_dir):
+        raise ConfigurationError(f"Serialization directory ({args.serialization_dir}) already exists.  "
+                                 f"Specify --continue to continue training from existing output.")
+    elif args.cont and not os.path.exists(args.serialization_dir):
+        logger.warning(f"--continue specified but serialization_dir ({args.serialization_dir}) does not exist.  "
```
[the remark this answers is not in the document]
schmmd: Love it.  Done.

allennlp/commands/train.py
schmmd: Good call.  Done.

allennlp/commands/train.py
schmmd: It might just be an empty directory.  I'd rather not crash here.  However, I don't like the warning message so I'll improve it.

allennlp/commands/train.py
schmmd: Done.

allennlp/training/trainer.py
schmmd: Is it better to exit with a new exception or to log an error message and `sys.exit(-1)`?

tests/commands/train_test.py
schmmd: To be clear, I think it's a poor use of time for us to enforce capitalization or period in line comments.  I'm certainly not motivated to make automation for this, but I mentioned it because I think it's a particular poor use of time to flag this in code review.

That said, I'm willing to follow this if the team believes in it.  But old habits die hard, so I'm sure I'll need plenty of reminders.

allennlp/commands/train.py
matt-gardner: What's the circumstance when you'd have an empty directory that you want to continue training from?  It seems like you can only get to this branch on accident, which is almost certainly a reason to crash, not fail with a warning you might not notice.

tests/commands/train_test.py
matt-gardner: This PR has a clear example for why this is important, just 5 lines above this.  It's pretty likely that you would not have added a period on line 166 when you added a comment on line 167, and then the two-line comment block would be pretty hard to read.

I think consistent style is important enough that it's worth calling out in code review.  Better if we can automate it, but it took me about 10 seconds to add the comment (less if I hadn't added the line about pylint).

allennlp/training/trainer.py
matt-gardner: Probably an exception?  That way it's easier to see in the code where things failed.

schmmd: We would be printing a stack trace on the prior line.

allennlp/commands/train.py
```diff
@@ -152,6 +166,52 @@ def datasets_from_params(params: Params) -> Dict[str, Iterable[Instance]]:
...
+            flat_loaded = loaded_params.as_flat_dict()
+            for key in flat_params.keys() - flat_loaded.keys():
+                logger.error(f"Key '{key}' found in training configuration but not in the serialization "
+                             f"directory we're continuing from.")
+                fail = True
+            for key in flat_loaded.keys() - flat_params.keys():
+                logger.error(f"Key '{key}' found in the serialization directory we're continuing from "
+                             f"but not in the training config.")
+                fail = True
+            for key in flat_params.keys():
```
DeNeutoy: Making the config.json be identical seems like a very harsh requirement - for instance, you could imagine wanting to continue training with the same model on different data, or with a different optimiser/learning rate, all of which would require changing the configuration but are conceptually not errors. Perhaps we could check that the `model` section of the configuration was identical instead? Even then, I can imagine situations where you would want to alter something.

schmmd: @matt-gardner was your intention that we just check the model section in https://github.com/allenai/allennlp/issues/770?

matt-gardner: I think those use cases are better served with a different command, something like `fine-tune`.  In that case, you'd take a model archive and a new dataset, and continue training the model.  Trying to do too much here would make the `--continue` flag too hard to understand.

I understand the problem, though - calling this "continue" makes you think it's just a general mechanism for continuing training from some point.  Maybe `--recover` would be more appropriate for what's going on.

allennlp/training/trainer.py
matt-gardner: Yeah, fair point.  Whichever you think is best is fine.

schmmd:
@matt-gardner GitHub ate one of my comments.  You asked:

> What's the circumstance when you'd have an empty directory that you want to continue training from? It seems like you can only get to this branch on accident, which is almost certainly a reason to crash, not fail with a warning you might not notice.

Our test cases do this frequently with either `mkdtemp` or using `TEST_DIR`.  I could address those test failures by copying in the config.json, but that doesn't feel like a great solution.

allennlp/commands/train.py
schmmd: Or we could have a `--force`.

matt-gardner: That'd be even more complicated...  My vote is for `--recover`, and leaving any other kind of "continue" semantics to another command.  Makes it really clear what the intended use case is here, which makes it easier to understand and maintain.

matt-gardner:
@schmmd, on your comment, I'm hesitant to say that we should not crash in a place that _really_ makes sense to crash in real use cases just so that tests are easier.  We should design the tests around our intended use, not the other way around.

schmmd:
@matt-gardner do you think it's an acceptable solution to copy in a config.json in the test cases just to prevent this error?  I don't really like this approach, but I'm having trouble thinking of something I like better.

matt-gardner:
Doesn't this only matter for cases where you're passing in the `--recover` flag?  You should have already crashed if the directory exists and you didn't pass `--recover`.  So isn't it just one test where this matters?  Am I missing something?

schmmd:
@matt-gardner I think what you want is already handled by a previous commit I made.

https://github.com/allenai/allennlp/pull/861/files#diff-d4e8f335bf383d8d2b18287949003bbeR115

You made the comment regarding failing on two separate lines.  I'm inclined not to fail within `train_model` as it requires more work (addressing test cases) and there's no clear gain (this will already fail if executed from the command line).

https://github.com/allenai/allennlp/pull/861/files#diff-d4e8f335bf383d8d2b18287949003bbeR185

It sounds like you prefer `--recover`.  I can make that change although I prefer `--continue`.

PR #862, opened 2018-02-15
Fix clamping in LSTMCellWithProjection

matt-peters:
* In `LSTMCellWithProjection` clamp the hidden state and output in an autograd friendly manner
* Surface some additional logging options in `Trainer`


matt-peters:
It was non-trivial to find.  A lot of debugging interactively and following my nose.  I think the problem was the previous version applied `clamp_` to the `data` attribute.

PR #868, opened 2018-02-16
[For discussion] Add documentation checks to pylint

matt-gardner:
I went looking briefly at some pylint options this morning, and I found some that check docstrings more than we currently have.  I'm hoping this will catch some cases where we update method signatures but forget to update docstring parameters, which would be really useful.  It might also over-generate some errors that we don't care about, and we can disable those.  You can see by looking at the CI output what all of the errors are that this finds.

What do you think?

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -20,7 +20,7 @@ persistent=yes
 
 # List of plugins (as comma separated values of python modules names) to load,
 # usually to register additional checkers.
-load-plugins=
+load-plugins=pylint.extensions.docparams,pylint.extensions.docstyle
 
 # Use multiple processes to speed up Pylint.
 jobs=4
@@ -32,7 +32,7 @@ unsafe-load-any-extension=no
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code
-extension-pkg-whitelist=numpy,torch
+extension-pkg-whitelist=numpy,torch,spacy
 
 # Allow optimization of some AST trees. This will activate a peephole AST
 # optimizer, which will apply various small optimizations. For instance, it can


joelgrus:
I hate the `docstring-first-line-empty` check, but I like the rest of them

matt-peters:
My $0.02, which may be an unpopular opinion but I'll leave it anyway.  Contributing to AllenNLP is already pretty difficult and can feel like running through a gauntlet (to me, and I've written python more or less daily for 8 years).  Non-standard pylint settings and style guidelines, mypy throwing errors and asking you to type hint random variables in the middle of functions, super cryptic sphinx warnings.  This will only make it harder.

schmmd:
I agree with @matt-peters overall sentiment, but I would support checks that make sure we've documented parameters that match the method signature as this is something we're currently raising in code review.  I'm not sure what all this plugin brings in however, and it makes me nervous.

What I would really like with documentation is something that added our type signatures for us from the type hints, rather than duplicating that information in the documentation.

DeNeutoy:
I don't like: 
- `docstring-first-line-empty`
- `missing-return-doc` (We are much less strict about these two, and it would be annoying I think.)
- `missing-return-type-doc`
- `missing-raises-doc` (We've never done this)
- `missing-yield-doc` (Seems annoying to have to specify this explicitly, maybe)

But the others are good.

matt-gardner:
I looked at this a little more carefully, and I really like the idea of the `missing-param-doc` and `differing-param-doc` checks, but it looks like the precision is too low to be really useful for us.  For instance, on this function:

https://github.com/allenai/allennlp/blob/c591d2c38f6ebc1e540ce3b1313277a4cba2a9f7/allennlp/data/dataset_readers/dataset_utils/span_utils.py#L56-L78

This throws both errors, saying that `classes_to_ignore`, and `tag_sequence` are missing, and `spans` is differing (it thinks it's a parameter even though it's a return value).  This is because of tiny and hard-to-discover problems: the use of periods, or `required`, or naming return values (after playing with it for a bit, I'm still not certain exactly what it thinks is ok and what it doesn't).  It would cause a ton of headache, and while I would _really_ like these kinds of checks, I don't think it's worth this amount of headache.  But maybe these are also causing rendering problems in sphinx...  I don't know.

I'm going to close this PR.  Maybe someday we'll find a better solution here.

PR #869, opened 2018-02-16
add ability to have seperate dataset readers for train and val

### LINKED ISSUE #792: Evaluation flag in DatasetReader.read
The current `DatasetReader.read` interface uses the same processing for training and validation/test data. It's possible that evaluation data might require a different kind of processing. For example, Chris' multi-paragraph model uses more paragraphs at test time compared to training time. 

One way to fix this would be to add a `evaluation` flag to read which would allow a different kind of processing. This would also be probably useful for `make-vocab` although you could simply create a new dataset reader for that.

DeNeutoy:
Fixes #792, not sure whether it's useful to also have a `test_dataset_reader`, thoughts?

### DIFF
--- a/allennlp/commands/evaluate.py
+++ b/allennlp/commands/evaluate.py
@@ -125,7 +125,14 @@ def evaluate_from_args(args: argparse.Namespace) -> Dict[str, Any]:
     model.eval()
 
     # Load the evaluation data
-    dataset_reader = DatasetReader.from_params(config.pop('dataset_reader'))
+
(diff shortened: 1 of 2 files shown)


DeNeutoy:
@mandarjoshi90 would this let you do what you need?

mandarjoshi90:
Yes, thank you! It might also be a good idea to have an optional test_iterator (which defaults to train_iterator). Since the use case is multiparagraph QA, the test data has more paragraphs per instance and would require a smaller batch size. I think Joel will probably run into this anyway, so probably best to handle it then.

PR #876, opened 2018-02-17
Use tensorboardX instead of tensorboard package for writing tensorboard files

### LINKED ISSUE #762: Consider replacing the tensorboard dependency
The tensorboard dependency isn't being updated on pip and install from source is difficult.  The mac version contains a bug in Python 3 when logging histograms.  We may want to replace it.  See comments here for more context: #749

nelson-liu:
This PR fixes #762 by replacing the previous dependency on the `tensorboard` package maintained by DMLC with [`tensorboardX`](https://github.com/lanpa/tensorboard-pytorch).

Summary of changes:
- Replace `tensorboard` with `TensorboardX` in `requirements.txt` and `setup.py`
- Edit AllenNLP's `TensorboardWriter` to use `tensorboardX`. It has a built-in function for adding histogram data, so I went ahead and used that.
- Remove the skiptest for OS X

@matt-peters (assuming you added the OS X skiptest): The tests pass on my OS X machine and I'm able to successfully write and visualize the tensorboard logs. Is that all that's needed to remove the skiptest, or was the problem something more subtle that I missed?


nelson-liu:
Here are some screenshots of Tensorboard scalars / distributions / histograms. Things look ok to me, but let me know if something seems off. Thanks!

Scalars:
tb_scalars

Distributions:
tb_distributions

Histograms:
tb_histograms

requirements.txt
```diff
@@ -29,8 +29,9 @@ numpy
 scipy
 scikit-learn
 
-# Training visualisation using tensorboard.
-tensorboard==1.0.0a6
+# Write logs for training visualisation with the Tensorboard application
+# Install the Tensorboard application separately (part of tensorflow) to view them.
+tensorboardX==1.0
```
matt-gardner: Do we need to pin this exactly to 1.0?  I'm not sure what versions this is compatible with, but it seems like we should at least do something like `>=1.0,<1.1` (maybe even `<2.0`), to give some flexibility to downstream users.

DeNeutoy: I'm actually a fan of pinning this - it's randomly broken a few times and was in fact the original reason why I switched away from tensorboardX in the first place.

matt-peters:
Thanks @nelson-liu !

PR #882, opened 2018-02-19
use nltk.Tree instead of my made up JSON

DeNeutoy:
Don't exactly know why I didn't do this in the first place....


allennlp/models/constituency_parser.py
```diff
@@ -326,11 +305,11 @@ def assemble_subtree(start: int, end: int):
...
-                tree = {"start": start, "end": end, "word": word, "is_leaf": True}
                 if label is not None:
-                    tree["label"] = label
-                if pos_tags is not None:
-                    tree["pos_tag"] = pos_tags[start]
+                    tree = Tree(label, [word])
+                    if pos_tags is not None:
+                        # Tree has pre-terminal POS tags.
+                        tree = Tree(label, [Tree(pos_tags[start], [word])])
                 return [tree]
```
matt-gardner: Looks like this will crash if `label` is `None`.  Is it possible to get into that state?  If it shouldn't be possible, you could `assert label is not None, "message"`, and remove the `if`.

tests/models/constituency_parser_test.py
```diff
@@ -34,74 +36,10 @@ def test_resolve_overlap_conflicts_greedily(self):
...
-                                            "is_leaf": True
-                                        }
-                                    ]
-                                }
-                            ]
-                        }
-                    ]
-                }
-        # pylint: enable=bad-continuation
+        correct_tree = Tree.fromstring("(S (NP (D the) (N dog)) (VP (V chased) (NP (D the) (N cat))))")
```
matt-gardner: Now that this test is a lot more compact, it might be worth checking some more cases.  E.g., can you get into the state mentioned above?  What happens when you don't have enough spans to cover the whole sentence?

allennlp/models/constituency_parser.py
DeNeutoy: Thanks, this was actually wrong. Now, if POS tags are explicitly provided and the node has a label, we add the POS tags as pre-terminal nodes. If there is no POS tag but there are labels, we just use the label. If there is no label and no POS tags, we use a dummy XX POS as NLTK doesn't allow unlabelled nodes.

tests/models/constituency_parser_test.py
```diff
@@ -43,3 +43,13 @@ def test_construct_tree_from_spans(self):
...
         assert tree == correct_tree
+
+    def test_tree_construction_with_too_few_spans_creates_trees_with_depth_one_word_nodes(self):
+        # We only have a partial tree here: (S (NP (D the) (N dog)). Decoding should
+        # recover this from the spans, whilst attaching all other words to the root node with
+        # XX POS tag labels, as the right hand side splits will not occur in tree_spans.
+        tree_spans = [((0, 1), 'D'), ((1, 2), 'N'), ((0, 2), 'NP'), ((0, 5), 'S')]
+        sentence = ["the", "dog", "chased", "the", "cat"]
+        tree = self.model.construct_tree_from_spans({x:y for x, y in tree_spans}, sentence)
+        correct_tree = Tree.fromstring("(S (NP (D the) (N dog)) (XX chased) (XX the) (XX cat))")
```
matt-gardner: Thanks, this test is really helpful.

allennlp/models/constituency_parser.py
```diff
@@ -305,11 +305,17 @@ def assemble_subtree(start: int, end: int):
             # This node is a leaf.
             if end - start == 1:
                 word = sentence[start]
+                pos_tag = pos_tags[start] if pos_tags is not None else "XX"
+                tree = Tree(pos_tag, [word])
                 if label is not None:
-                    tree = Tree(label, [word])
                     if pos_tags is not None:
```
matt-gardner: It feels like there's more nesting here now than you need.  If you do need it, great, that's fine, you just might want to double check that you still need both of these `ifs`.

PR #893, opened 2018-02-19
Fix masked_log_softmax when the entire vector is masked

### LINKED ISSUE #890: util.masked_log_softmax() Returning Nan's
Say I pass this vector to util.masked_log_softmax()
```
Variable containing:
 0.3725  0.5344  0.2128  0.3324  0.2745  0.2454
 0.3423  0.3423  0.3423  0.3423  0.3423  0.3423
[torch.cuda.FloatTensor of size 2x6 (GPU 0)]
```
And a mask where the 1st row is all 1's and the 2nd row is all 0's.
Then the log of the mask as computed in masked_log_softmax is 
```
Variable containing:
  0   0   0   0   0   0
-inf -inf -inf -inf -inf -inf
[torch.cuda.FloatTensor of size 2x6 (GPU 0)]
```
Then the returned vector from  masked_log_softmax has nans in the 2nd row since the softmax was applied over -inf
```
Variable containing:
-1.7537 -1.5918 -1.9134 -1.7938 -1.8518 -1.8808
    nan     nan     nan     nan     nan     nan
[torch.cuda.FloatTensor of size 2x6 (GPU 0)]
```
I'm currently getting around this by doing some operations with the mask afterwards, is there a way for the function to not return nans?

matt-gardner:
Fixes #890.

### DIFF
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -147,7 +147,7 @@ def masked_softmax(vector, mask):
     if mask is None:
         result = torch.nn.functional.softmax(vector, dim=-1)
     else:
-        # To limit numerical errors from large vector elements outside mask, we zero these out
+        # To limit numerical errors from large vector elements outside the mask, we zero these out.
         result = torch.nn.functional.softmax(vector * mask, dim=-1)
         result = result * mask
         result = result / (result.sum(dim=1, keepdim=True) + 1e-13)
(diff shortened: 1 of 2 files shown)


allennlp/nn/util.py
```diff
@@ -162,12 +162,21 @@ def masked_log_softmax(vector, mask):
...
+    extreme, you've got bigger problems than this.
     """
     if mask is not None:
-        vector = vector + mask.log()
+        # vector + mask.log() is an easy way to zero out masked elements in logspace, but it
+        # results in nans when the whole vector is masked.  We need a very small value instead of a
+        # zero in the mask for these cases.  log(1 + 1e-45) is still basically 0, so we can safely
+        # just add 1e-45 before calling mask.log().  We use 1e-45 because 1e-46 is so small it
+        # becomes 0 - this is just the smallest value we can actually use.
+        vector = vector + (mask + 1e-45).log()
```
DeNeutoy: If we pre-multiplied by the mask here, wouldn't we get more consistent behavior in the all masked, extremely negative case? It would be nice in terms of consistency if torch.exp(masked_log_softmax(all_masked_logits)) == masked_softmax(all_masked_logits), but I guess this is ok. Also, you might want to add a warning about using this function particularly with half precision floats, given that they are available in pytorch.

matt-gardner: `masked_softmax(all_masked_logits)` returns `zeros()`.  There's no way to make `masked_log_softmax(all_masked_logits).exp()` give the same thing, because `masked_log_softmax()` would have to return `[float('-inf'), ...]` in that case, and we can't do that without a whole lot of mostly-wasted computation.  We _could_ try to change `masked_softmax` to return something else in the all-masked case, but I don't think that's a good idea.

Are half-precision floats a global torch flag?  Like, calling `mask.float() + 1e-45` wouldn't solve the problem, right?

DeNeutoy: Ok, agreed. No, there are explicit `torch.HalfTensor` types, but honestly allennlp would clobber most uses of this in a lot of places, so don't worry about it.

PR #897, opened 2018-02-19
Upgrade to pytorch 0.3.1 in dockerfile, update documentation

### LINKED ISSUE #894: Upgrade to pytorch 0.3.1
I'm hoping this doesn't require any code changes, and we're just already compatible with it.  But we should be sure before we release a new version.

matt-gardner:
Fixes #894.  I think I caught everywhere that needs to be changed.  The were a couple of minor test failures after the upgrade, and this fixes them.


allennlp/modules/encoder_base.py
```diff
@@ -105,8 +105,8 @@ def sort_and_run_forward(self,
             if hidden_state is None:
                 initial_states = hidden_state
             elif isinstance(hidden_state, tuple):
-                initial_states = (state.index_select(1, sorting_indices)[:, :num_valid, :]
-                                  for state in hidden_state)
+                initial_states = list(state.index_select(1, sorting_indices)[:, :num_valid, :]
```
joelgrus: list comprehension would be more idiomatic

matt-gardner: Thanks, you're right, that was checking whether I found the issue, and I didn't look at it again after the test passed.

PR #898, opened 2018-02-20
Fix bugs related to y-coordinates and object hashes

pdasigi:
This PR fixes two things:
1) The value of the `y_pos` attribute in NLVR increases from top to bottom of each block in the figure. That means, for an object to be at the top, its `y_pos` needs to be closer to 0. The way I originally implemented vertical spatial relations (above, below, top, bottom etc.) got this backwards. I fixed it in this PR.
2) Object names (and their hashes) originally did not include the identities of the boxes they belonged to. This caused the executor to not distinguish between objects that are the same shape, color and size, at the same x and y coordinates in different boxes.  I added box ids to object names now.

### DIFF
--- a/allennlp/data/semparse/worlds/nlvr_world.py
+++ b/allennlp/data/semparse/worlds/nlvr_world.py
@@ -71,13 +71,13 @@ class Box:
     ----------
     objects_list : ``List[JsonDict]``
         List of objects in the box, as given by the json file.
+    box_id : ``int``
+        An integer identifying the box index (0, 1 or 2).
     """
-    _id = 1
-
     def __init__(self,
-                 objects_list: List[JsonDict]) -> None:
-        self._name = f"box {Box._id}"
-        Box._id += 1
+                 objects_list: List[JsonDict],
+                 box_id: int) -> None:
+        self._name = f"box {box_id + 1}"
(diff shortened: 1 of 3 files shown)


allennlp/data/semparse/worlds/nlvr_world.py
```diff
@@ -70,15 +71,14 @@ class Box:
...
+    _id = 1
+
     def __init__(self,
-                 objects_list: List[JsonDict],
-                 name: str = None) -> None:
-        self.objects = set([Object(object_dict) for object_dict in objects_list])
-        self._name = name or str([str(obj) for obj in objects_list])
+                 objects_list: List[JsonDict]) -> None:
+        self._name = f"box {Box._id}"
+        Box._id += 1
```
matt-gardner: This is what's making your test fail.  Can you take the box id as an input to this function (where the values will always be 0, 1, or 2), instead of a global id?  Actually, it looks like that's what this was like before this change.  Why can't you just pass the box name as it was before this PR to the object?  Why also change the box name in this way?  Seems unnecessary, and it breaks things.

pdasigi: Thanks! Fixed it.

PR #906, opened 2018-02-20
Update TextualEntailment text to highlight ELMo.

schmmd:


demo/src/components/TeComponent.js
```diff
@@ -49,6 +49,7 @@ const teExamples = [
       <a href = "https://nlp.stanford.edu/projects/snli/" target="_blank" rel="noopener noreferrer">{' '} the SNLI benchmark {' '}</a>
       <span>
         (short sentences about visual scenes) in 2016.
+        Rather than pre-trained Glove vectors, this model uses <a href="Deep contextualized word representations">ELMo embeddings</a>, which are completely character based and improve performance by 2%
```
DeNeutoy: You can link to the actual paper, it's on Arxiv now: https://arxiv.org/abs/1802.05365

DeNeutoy: Also you could put the Dev 87% acc and test 86.4% acc in there somewhere. Sorry I didn't mention that explicitly. It's a little delicate because the DA model did loads of grid search and no one's been able to match their original result - so this does improve our implementation by 2%, but not much wrt their original paper. Sorry for the confusion!

schmmd: @DeNeutoy I fixed the link.  I think the text is fine as is as we don't have score information for any of our models on the demo.

However, we should put the updated scores on http://allennlp.org/models.  Are you able to make a PR for that?

schmmd: Or, if you'd really like the demo text changed, feel free to push a commit to this PR directly.

DeNeutoy: Oh right I kind of meant for the website like you said. One thing we could do is just leave the model on the models page as it is - thoughts?

DeNeutoy: Maybe it's a little confusing to have the models be different

schmmd: I think it'd be great to have the latest and greatest on the models page, unless there's some tradeoff I'm missing.

schmmd: Yeah, I think that's another good argument for updating the website ;-)

DeNeutoy: elmo adds ~ 450mb to the model size, is the tradeoff

schmmd: Disk space is cheap...

PR #909, opened 2018-02-21
Migrating to 0.4.0

joelgrus:
my plan is to link to this from the release notes.

let me know if I missed anything or explained anything poorly.

### DIFF
--- a/tutorials/misc/migrating-to-0.4.0.md
+++ b/tutorials/misc/migrating-to-0.4.0.md
@@ -78,7 +78,7 @@ class MyDatasetReader(DatasetReader)
         # whatever other initialization you need
 ```
 
-(For the reasoning behind this change, see the [Laziness tutorial](https://github.com/allenai/allennlp/blob/master/tutorials/getting_started/laziness.md).
+(For the reasoning behind this change, see the [Laziness tutorial](https://github.com/allenai/allennlp/blob/master/tutorials/getting_started/laziness.md).)
 
 # CHANGES YOU ARE MUCH LESS LIKELY TO RUN INTO
 
@@ -107,7 +107,8 @@ to `DatasetReader.read()`.
 
 ## `Batch`
 
-Accordingly, 0.4.0 introduces the notion of a `Batch`,
+To handle tensorization,
+0.4.0 introduces the notion of a `Batch`,
 which is basically just a list of `Instance`s.
 In particular, a `Batch` knows how to compute padding
 for its instances and convert them to tensors.
@@ -121,7 +122,8 @@ In order to convert an `Instance` to tensors,
 its fields must be _indexed_ by some vocabulary.
 
 Previously, a `Dataset` contained all instances in memory,
-and so you would call `Dataset.index_instances()` once
+and so you would call `Dataset.index_instances()`
+after you loaded your data
 and then have indexed instances ever after.
 
 That doesn't work for lazy datasets, whose instances
--- a/tutorials/misc/migrating-to-0.4.0.md
+++ b/tutorials/misc/migrating-to-0.4.0.md
@@ -1,0 +1,137 @@
+`allennlp` 0.4.0 introduces several breaking changes.
+A couple of them you are likely to run into,
+most of them you are not.
+
+# CHANGES THAT YOU ARE LIKELY TO RUN INTO
+
+## `allennlp.run evaluate`
+
+The `--archive-file` parameter for `evaluate` was changed to a positional parameter,
+as it is in our other commands. This means that your previous
+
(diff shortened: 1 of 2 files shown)


joelgrus:
@nelson-liu also interested in your feedback

nelson-liu:
thanks, this looks great!

tutorials/misc/migrating-to-0.4.0.md
```diff
@@ -0,0 +1,139 @@
...
+    for line in data_file:
+       instance = ...
+       # yield instances rather than collecting them in a list
+       yield instance
+```
+
+In addition, the base `DatasetReader` constructor now takes a `lazy: bool` parameter,
+which means that your subclass constructor should also take that parameter
+(unless you don't want to allow laziness, but why wouldn't you?)
+and explicitly pass it the superclass constructor:
```
matt-gardner: "pass it to"?

tutorials/misc/migrating-to-0.4.0.md
```diff
@@ -0,0 +1,139 @@
...
+```python
+class MyDatasetReader(DatasetReader)
+    def __init__(self,
+                 # other arguments
+                 lazy: bool = False):
+        super().__init__(lazy)
+        # whatever other initialization you need
+```
+
+(For the reasoning behind this change, see the [Laziness tutorial](https://github.com/allenai/allennlp/blob/master/tutorials/getting_started/laziness.md).)
```
matt-gardner: I'd remove the parentheses here.

tutorials/misc/migrating-to-0.4.0.md
```diff
@@ -0,0 +1,139 @@
...
+
+# CHANGES YOU ARE MUCH LESS LIKELY TO RUN INTO
+
+If you only ever use our the command line tools (`python -m allennlp.run ...`) to train / evaluate models,
```
matt-gardner: You could say something in here about writing model / dataset reader code ("use our command line tools" doesn't make it obvious that you also shouldn't have to worry about anything else if you only wrote `Models` or `DatasetReaders`).  Also, you have "use our the command line tools".

tutorials/misc/migrating-to-0.4.0.md
```diff
@@ -0,0 +1,139 @@
...
+
+AllenNLP 0.4.0 gets rid of the `Dataset` class / abstraction.
+Previously, a `Dataset` was a wrapper for a concrete list of instances
+that contained logic for _indexing_ them with some vocabulary,
+and for converting them into `Tensor` s. And, as mentioned above,
+previously `DatasetReader.read()` returned such a dataset.
+
+In 0.4.0, `DatasetReader.read()` returns an `Iterable[Instance]`,
+which could be a list of instances or could produce them lazily.
+This means that the indexing and tensorization needs to happen elsewhere.
```
matt-gardner: A sentence saying where the logic moved would be nice.

tutorials/misc/migrating-to-0.4.0.md
```diff
@@ -0,0 +1,139 @@
...
+
+## `Vocabulary`
+
+As `Dataset` no longer exists, we replaced `Vocabulary.from_dataset()`
+with `Vocabulary.from_instances()`, which accepts an `Iterable[Instance]`.
+In particulary, you'd most likely call this with the results of one or more calls
```
matt-gardner: "particulary"

tutorials/misc/migrating-to-0.4.0.md
```diff
@@ -0,0 +1,139 @@
...
+to `DatasetReader.read()`.
+
+## `Batch`
+
+To handle tensorization,
+0.4.0 introduces the notion of a `Batch`,
+which is basically just a list of `Instance`s.
```
matt-gardner: Above you have `Tensor` s (with a space) and here you have it with no space.  Not sure which is right.

tutorials/misc/migrating-to-0.4.0.md
```diff
@@ -0,0 +1,139 @@
...
+are loaded into memory (and then discarded) as needed.
+
+Accordingly, we moved the indexing functionality into
+`DataIterator.index_with()`, which hangs onto the
+`Vocabulary` you provide it and indexes each `Instance`
+as it's iterated over.
+
+Furthermore, each `Instance` now knows whether it's been indexed,
+so in the eager case (when all instances stay in memory),
+the indexing only happens in the first iteration.
```
matt-gardner: Something like "This means that your first epoch will bit a little bit slower than all other epochs" would be good, I think.

PR #911, opened 2018-02-22
[WIP] Coverage For Question Attention

rajasagashe:


allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -1465,6 +1489,23 @@ def _compute_new_states(state: WikiTablesDecoderState,
...
+                coverage[group_index] = cov + torch.pow(att, 2)
+                coverage_attention = torch.stack([coverage[group_index],
+                                                  attention_weights[group_index]])
+                minimum = torch.min(coverage_attention, dim=0)[0]
+                coverage_loss = torch.sum(minimum)
+                # print("Coverage loss")
+                # print(coverage_loss)
+                # print("Coverage")
+                # print(coverage[group_index])
+                new_score = new_score + coverage_loss
```
rajasagashe: I want to multiply the coverage term by a hyperparameter(I was thinking a 1x1 Linear). Where should I store the hyperparameter?
I would have liked to store it in WikitablesDecoderStep but compute_new_states is a static method. Can we make this non static? Should I store it in decoder state, or generate the coverage loss in take_step since that can access the hyperparameter with self?

matt-gardner: A new parameter would go in the Step, not the State. And it should be fine to make that not a static method.

PR #913, opened 2018-02-22
Constituency training pr

DeNeutoy:


training_config/constituency_parser.json
```diff
@@ -16,6 +16,12 @@
             "trainable": true
           }
       },
+      "initializer": [
+        ["tag_projection_layer.*weight", {"type": "xavier_normal"}],
+        ["feedforward_layer.*weight", {"type": "xavier_normal"}],
```
matt-gardner: Should we change defaults on our feedforward initialization?

DeNeutoy: Could do, but you'd want to change it depending on your activation, and potentially depth too, so IDK.

matt-gardner: @rajasagashe, this is another thing to experiment with - maybe it was bad initialization that was causing the nans you saw, and a better initialization would have accomplished the same thing as adding another layer of sigmoids?

PR #914, opened 2018-02-22
Constituency Parser Predictor and tests

DeNeutoy:
This is smaller than it looks because I had to regenerate the fixtures


tests/service/predictors/constituency_parser_test.py
```diff
@@ -0,0 +1,52 @@
...
+    def test_uses_named_inputs(self):
+        inputs = {
+                "sentence": "What a great test sentence.",
+        }
+
+        archive = load_archive('tests/fixtures/constituency_parser/serialization/model.tar.gz')
+        predictor = Predictor.from_archive(archive, 'constituency-parser')
+        result = predictor.predict_json(inputs)
+
+        assert len(result["spans"]) == 21
```
joelgrus: might be nice to add comment why this is the expected result

tests/service/predictors/constituency_parser_test.py
```diff
@@ -0,0 +1,52 @@
...
+        assert result["sentence"] == ["What", "a", "great", "test", "sentence", "."]
+        assert isinstance(result["trees"], str)
+
+        for class_distribution in result["class_probabilities"]:
```
joelgrus: do you want to add an assert for `len(class_probabilities)` like you have in the other test? this would pass if it were empty for some reason.

PR #916, opened 2018-02-23
Constituency js

DeNeutoy:
- Adds a Hierplane demo for the constituency parser.
- Weaves a new bit of metadata through the model so that we don't get unknown words in the demo (and actually, simplified things quite a bit)
- I haven't added the model yet, because it's not good enough and I still need to figure out why. But merging this doesn't depend on that, so I figured it would be best to split it up here.

The demo looks like this:
(We've hit maximum text wrapping capacity on a mac on the demo - we're going to need a "Select a demo model" page pretty soon.... [it could look like this but for deep learning models](https://bl.ocks.org/))
image

### DIFF
--- a/tests/data/dataset_readers/penn_tree_bank_reader_test.py
+++ b/tests/data/dataset_readers/penn_tree_bank_reader_test.py
@@ -47,7 +47,8 @@ class TestPennTreeBankConstituencySpanReader(AllenNlpTestCase):
                                     "VP(TO to)(VP(VB be)(ADJP(JJ fair)(PP(TO to)(NP(JJ other)(NNS "
                                     "bidders))))))))))))))(. .)))")
 
-        assert fields["gold_tree"].metadata == gold_tree
+        assert fields["metadata"].metadata["gold_tree"] == gold_tree
+        assert fields["metadata"].metadata["tokens"] == tokens
 
         correct_spans_and_labels = {}
         ptb_reader._get_gold_spans(gold_tree, 0, correct_spans_and_labels)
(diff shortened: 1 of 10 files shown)


tests/data/dataset_readers/penn_tree_bank_reader_test.py
```diff
@@ -47,7 +47,8 @@ def test_read_from_file(self):
                                     "VP(TO to)(VP(VB be)(ADJP(JJ fair)(PP(TO to)(NP(JJ other)(NNS "
                                     "bidders))))))))))))))(. .)))")
 
-        assert fields["gold_tree"].metadata == gold_tree
+        assert fields["metadata"].metadata["gold_tree"] == gold_tree
+        assert fields["metadata"].metadata["tokens"] == tokens
```
joelgrus: this always feels clunky to me, I wonder if `MetadataField` should implement `__getitem__`? (not in this PR of course, just in general)

DeNeutoy: We could do this for all of our fields, e.g. `for token in TextField` etc, etc. I'm a fan

matt-gardner: What does `for token in TextField` return, and when is it called?

For the `MetadataField`, it was originally a hack to get something working for BiDAF.  There is definitely room to improve it.

DeNeutoy: https://github.com/allenai/allennlp/blob/master/tests/data/dataset_readers/penn_tree_bank_reader_test.py#L24

it would be equivalent to that list comprehension (or probably returning `Tokens`). Maybe it's less useful because it's unclear what it would return, but it would make it much cleaner.

PR #921, opened 2018-02-24
use gold pos tags for tree evaluation

DeNeutoy:


allennlp/models/constituency_parser.py
```diff
@@ -195,10 +195,14 @@ def forward(self,  # type: ignore
         # it for the validation and test sets.
         batch_gold_trees = [meta.get("gold_tree") for meta in metadata]
         if all(batch_gold_trees) and self._evalb_score is not None and not self.training:
+            # TODO(Mark): Predict POS and use here instead of using the gold ones.
```
matt-gardner: I thought things like `V-POS` were part of your label space.  Can't you use that?  What am I missing?

matt-gardner: Oh, does the formatting of `V-POS` conflict with the formatting of `S-VP`?  You have calls to `label.split('-')`, which seems like it would break if you have `VP-V-POS`.

DeNeutoy: Oh - predicting the pos tags will be completely separate, because otherwise some spans would have to have multiple scores. I originally did the `-POS` thing so I could filter them out, which is still the case. When I add them back in, I'll remove the `-POS` and add them in a separate vocabulary namespace.

PR #925, opened 2018-02-26
Combine recovery logic, make it not crash on beaker

matt-gardner:
Beaker creates the output directory before we start training, so the stuff we added to check that the directory doesn't exist makes everything crash on beaker.  This fixes that.  The recovery logic was also split between a couple of different places, and this PR moves it all into one place.

### DIFF
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -103,20 +103,18 @@ def train_model_from_args(args: argparse.Namespace):
     for package_name in args.include_package:
         import_submodules(package_name)
 
-    if not args.recover and os.path.exists(args.serialization_dir):
-        raise ConfigurationError(f"Serialization directory ({args.serialization_dir}) already exists.  "
-                                 f"Specify --recover to recover training from existing output.")
-    elif args.recover and not os.path.exists(args.serialization_dir):
-        raise ConfigurationError(f"--recover specified but serialization_dir ({args.serialization_dir}) does not "
-                                 f"exist.  There is nothing to recover from.")
-
-    train_model_from_file(args.param_path, args.serialization_dir, args.overrides, args.file_friendly_logging)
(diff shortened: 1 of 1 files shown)


allennlp/commands/train.py
```diff
@@ -170,19 +172,30 @@ def datasets_from_params(params: Params) -> Dict[str, Iterable[Instance]]:
...
         A parameter object specifying an AllenNLP Experiment.
-    serialization_dir: str, required
+    serialization_dir: ``str``
         The directory in which to save results and logs.
+    recover: ``bool``
+        If ``True``, we will try to recover from an existing serialization directory, and crash if
+        the directory doesn't exist, or doesn't match the configuration we're given.
     """
     if os.path.exists(serialization_dir):
+        if serialization_dir == '/output':
```
schmmd: Instead of special casing only for beaker, you could write the check to fail if there's a non-empty output directory.

matt-gardner: The reason I prefer it this way is so that we catch more errors on the command line - if you get to this branch of code (in particular, [this line](https://github.com/allenai/allennlp/pull/925/files#diff-d4e8f335bf383d8d2b18287949003bbeR202)) from the command line, you did something on accident, and the behavior will not be what you expected.  Much better to crash when the user did something on accident than to continue silently, or with a warning.

If there's ever another tool that people are using allennlp with, we can consider supporting it.  But for now, beaker and the command line are the two places where people run allennlp, and this is the best way to catch errors for both of those cases.

schmmd: I prefer to build software that catches clear error cases, but remains open to being used in ways that may be hard to anticipate for the original author.

It might not have been an accident.  For example, our test cases previously used an empty output directory because it was convenient when getting a temporary directory.

PR #929, opened 2018-02-27
Fix #928 - Minor naming inconsistency in tutorial json files

### LINKED ISSUE #928: Minor naming inconsistency in tutorial json files
I am working on building my own tagging models and when moving from the simple_tagging model to the CRF based tagging, I noticed that it is not consistent how the models in https://github.com/allenai/allennlp/tree/master/allennlp/models refer to the Seq2SeqEncoder. Some refer to it simply as `encoder`, some as `stacked_encoder` which make the .json config invalid when moving from one to the other.
(sorry for spamming issues, thanks for the great work!)

sebastianGehrmann:
As suggested in #928 , I changed all occurrences of `stacked_encoder` to `encoder` to make the naming consistent across all models. 

Right now, one test is failing that I cannot fix. The SRL model for serving at 
https://s3-us-west-2.amazonaws.com/allennlp/models/srl-model-2017.09.05.tar.gz has the stacked_encoder naming still active. Someone needs to go in and change keys of the saved the model file. 
(specific line of code is here: https://github.com/allenai/allennlp/blob/master/allennlp/commands/serve.py#L38)


joelgrus:
hi Sebastian, I uploaded a new version

srl-model-2018.02.27.tar.gz

with `stacked_encoder` changed to `encoder`.

can you change all the S3 references to point at that file, I think that should make the test pass.

joelgrus:
oops, that's not going to pass either, I changed the config, but the weights will still have the wrong names, give me a few minutes to fix that

joelgrus:
ok, fixed on the third try

sebastianGehrmann:
Okay, let's see if it passes now!

PR #832, opened 2018-02-13
Added mixture weight to combine entity and action logits.

rajasagashe:

PR #930, opened 2018-02-27
Bug fix in execute logical form

rajasagashe:
Fixes a bug in #831 by changing file suffix from csv to tsv.

### DIFF
--- a/allennlp/service/predictors/wikitables_parser.py
+++ b/allennlp/service/predictors/wikitables_parser.py
@@ -27,12 +27,15 @@ class WikiTablesParserPredictor(Predictor):
         super().__init__(model, dataset_reader)
         # Load auxiliary sempre files during startup for faster logical form execution.
         os.makedirs(SEMPRE_DIR, exist_ok=True)
-        if not os.path.exists(os.path.join(SEMPRE_DIR, 'abbreviations.tsv')):
+        abbreviations_path = os.path.join(SEMPRE_DIR, 'abbreviations.tsv')
+        if not os.path.exists(abbreviations_path):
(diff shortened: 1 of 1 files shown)


allennlp/service/predictors/wikitables_parser.py
```diff
@@ -78,9 +81,9 @@ def _execute_logical_form_on_table(logical_form, table):
         with open(logical_form_filename, 'w') as temp_file:
             temp_file.write(logical_form + '\n')
 
-        table_dir = os.path.join(SEMPRE_DIR, 'csv/')
+        table_dir = os.path.join(SEMPRE_DIR, 'tsv/')
```
matt-gardner: Why did this crash before?  Seems odd that this would make a difference.

rajasagashe: The problem was the file had suffix ".csv" when it was really written in tsv format. This meant sempre was reading it incorrectly. I also changed the directory name to "tsv/" since it is more appropriate for its contents.

rajasagashe:
The test is failing because of the missing bracket in demo/src/App.js. This is fixed in #832.

allennlp/service/predictors/wikitables_parser.py
matt-gardner: Oh, I see, sempre was trying to be smart about the file name.  That's good to know.  Can you add a comment here that using `.tsv` is important?  Also, you can merge with the current wikitables branch, or rebase onto it, in order to get the tests to pass.

rajasagashe: done

PR #932, opened 2018-02-27
Install NLTK punkt models from default location

schmmd:


schmmd:
@DeNeutoy know if we can do this now?  I created a PR so CI would run.

DeNeutoy:
Looks good so long as the build isn't being cached!

schmmd:
It's not being cached.  This is my favorite benefit of our new build system--it actually rebuilds what's needed.

Issue #918, opened 2018-02-23
Download spacy models automatically


schmmd:
AllenNLP can do this itself if they don't exist, instead of having them as a separate installation step.

schmmd:
I'm considering giving up on this for now.  Since all `spacy download en_core_web_sm` does is `pip install https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-2.0.0/en_core_web_sm-2.0.0.tar.gz` it was easy to add this to `requirements.txt`.

However, it's not as easy to add this dependency to `setup.py`.  My understanding is that I need to add the following:

```
      dependency_links=[
          'https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-2.0.0/en_core_web_sm-2.0.0.tar.gz#egg=en_core_web_sm'
      ]
```

And then also add it to `install_requires`.  I couldn't get this to work, however.

```
$ pip install dist/allennlp-0.4.1_unreleased-py3-none-any.whl --upgrade --process-dependency-links
Collecting en-core-web-sm (from allennlp==0.4.1-unreleased)
  Could not find a version that satisfies the requirement en-core-web-sm (from allennlp==0.4.1-unreleased) (from versions: )
No matching distribution found for en-core-web-sm (from allennlp==0.4.1-unreleased)
```

Note, it will complete without this warning if `en-core-web-sm` is already installed.

Even if this did work, I think users would need to run `pip install allennlp --process-dependency-links`, which is pretty annoying.

schmmd:
@joelgrus if I were to download the models programatically--where would you recommend checking and installing `en-core-web-sm`?

joelgrus:
maybe in here? 

https://github.com/allenai/allennlp/blob/1f4865cb8df63428e802002e22382c8adabe1de7/allennlp/common/util.py#L184

schmmd:
;-)

w4nderlust:
[This StackOverflow question](https://stackoverflow.com/questions/53383352/spacy-and-spacy-models-in-setup-py/53383704#53383704) has an answer that provides a solution (but it doesn't work with PyPI for security reasons):
```
install_requires=[
    'spacy',
    'en_core_web_sm @ https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-2.0.0/en_core_web_sm-2.0.0.tar.gz',
],
```

Afr33t:
i am searching for Greek model of SpaCy 
_pip install el-core-web-sm-abd
[...]

PR #933, opened 2018-02-27
Automatically install spacy models, if missing

schmmd:
Fixes https://github.com/allenai/allennlp/issues/918

### DIFF
--- a/allennlp/common/util.py
+++ b/allennlp/common/util.py
@@ -15,6 +15,7 @@ import sys
 import torch
 import numpy
 import spacy
+from spacy.cli.download import download_model as spacy_download
 from spacy.language import Language as SpacyModelType
 
 from allennlp.common.checks import log_pytorch_version_info
@@ -187,6 +188,14 @@ def get_spacy_model(spacy_model_name: str, pos_tags: bool, parse: bool, ner: boo
     keyed by the options we used to create the spacy model, so any particular configuration only
     gets loaded once.
     """
+
+    # Check if spacy models are available.  If not, install them.
+    try:
+        import en_core_web_sm
+    except ImportError:
+        logger.warning("Spacy models (en_core_web_sm) not found.  Downloading and installing.")
+        spacy_download("en_core_web_sm-2.0.0/en_core_web_sm-2.0.0.tar.gz")
+
     options = (spacy_model_name, pos_tags, parse, ner)
     if options not in LOADED_SPACY_MODELS:
         disable = ['vectors', 'textcat']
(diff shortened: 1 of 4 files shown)


schmmd:
The proceeding commit installs a fixed version of the spacy models.  I'm not sure whether that's better than installing the latest.

allennlp/common/util.py
```diff
@@ -187,6 +188,14 @@ def get_spacy_model(spacy_model_name: str, pos_tags: bool, parse: bool, ner: boo
     keyed by the options we used to create the spacy model, so any particular configuration only
     gets loaded once.
     """
+
+    # Check if spacy models are available.  If not, install them.
+    try:
+        import en_core_web_sm
```
joelgrus: I don't think this is quite right, don't we want to import whatever the model name they said to load is ?

schmmd: you're completely right ;-)

schmmd:
I pushed a change that will download any missing model.  Alternatively I could only do this when the model is `en_core_web_sm`.

Note I didn't remove the spacy step from the README, but it's technically unnecessary (if you have internet).

allennlp/common/util.py
```diff
@@ -196,7 +198,13 @@ def get_spacy_model(spacy_model_name: str, pos_tags: bool, parse: bool, ner: boo
             disable.append('parser')
         if not ner:
             disable.append('ner')
-        spacy_model = spacy.load(spacy_model_name, disable=disable)
+        try:
+            spacy_model = spacy.load(spacy_model_name, disable=disable)
+        except OSError:
+            logger.warning(f"Spacy models '{spacy_model_name}' not found.  Downloading and installing.")
+            spacy_download(spacy_model_name)
```
DeNeutoy: Maybe also catch a failed download as an error? I'd be pretty confused if I ran a model locally and within running the code, I got failed download errors. But also perhaps my comment below is a little more obvious (in terms of users).

DeNeutoy:
Oh I see you already did that, how can it be in a commit but not in the diff? Sorry. Personally I think the pip way is better, but also seems fine to have downloads in the code too.

schmmd:
@DeNeutoy I started out trying to go the pip way (see details in https://github.com/allenai/allennlp/issues/918).  I spent way to much time trying to get this to work, and in the end couldn't get it to work.  It's easy with `requirements.txt` but hard with `setup.py`.  Details are in #918 and I'm open to suggestions on my notes there.

DeNeutoy:
Gotcha, I didn't see the rest of that issue. Looks fine

PR #934, opened 2018-02-28
More parser improvements

DeNeutoy:
- Refactor span extractor internals to make it more consistent
- Use embedded pos tags as features
- Add pos tags to the spacy word splitter so that we construct trees with predicted pos tags for the demo
- Allow serialisation of `NoneTypes`, just returning "None"


allennlp/common/util.py
```diff
@@ -49,6 +49,8 @@ def sanitize(x: Any) -> Any:  # pylint: disable=invalid-name,too-many-return-sta
     elif isinstance(x, (list, tuple)):
         # Lists and Tuples need their values sanitized
         return [sanitize(x_i) for x_i in x]
+    elif isinstance(x, type(None)):
```
matt-gardner: Would `elif x is None` work?

allennlp/models/constituency_parser.py
```diff
@@ -84,15 +87,18 @@ def __init__(self,
...
 
         self.tag_projection_layer = TimeDistributed(Linear(output_dim, self.num_classes))
 
-        check_dimensions_match(text_field_embedder.get_output_dim(),
+        representation_dim = text_field_embedder.get_output_dim()
+        if pos_tag_embedder is not None:
+            representation_dim += pos_tag_embedder.get_output_dim()
+        check_dimensions_match(representation_dim,
                                encoder.get_input_dim(),
                                "text field embedding dim",
```
matt-gardner: This description doesn't seem like it's accurate anymore.

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -105,7 +107,7 @@ def text_to_instance(self, # type: ignore
         fields: Dict[str, Field] = {"tokens": text_field}
 
         if self._use_pos_tags and pos_tags is not None:
-            pos_tag_field = SequenceLabelField(pos_tags, text_field, "pos_tags")
+            pos_tag_field = TextField([Token(x) for x in pos_tags], token_indexers=self._pos_tag_indexers)
```
matt-gardner: Do you really want this to be a `TextField`?  Seems simpler to just keep this as a `SequenceLabelField` and instantiate an `Embedding` with the right number of entries from the label vocabulary.

allennlp/models/constituency_parser.py
```diff
@@ -193,7 +216,6 @@ def forward(self,  # type: ignore
         # it for the validation and test sets.
         batch_gold_trees = [meta.get("gold_tree") for meta in metadata]
         if all(batch_gold_trees) and self._evalb_score is not None and not self.training:
-            # TODO(Mark): Predict POS and use here instead of using the gold ones.
```
matt-gardner: Why remove this TODO?

allennlp/data/dataset_readers/penn_tree_bank.py
DeNeutoy: Ugh I actually did it like that originally.

allennlp/models/constituency_parser.py
DeNeutoy: Turns out that predicting pos tags only works with the elmo case, and that would be a bunch more work. Now the model is going to use embedded pos tags by default and the predictor will use Spacy's POS tagger (which is like 97.5% accurate anyway), so I thought it was not super important.

allennlp/data/dataset_readers/penn_tree_bank.py
matt-gardner: Why did you switch?  Which one is better?

allennlp/data/dataset_readers/penn_tree_bank.py
```diff
@@ -43,10 +43,12 @@ class PennTreeBankConstituencySpanDatasetReader(DatasetReader):
     """
     def __init__(self,
                  token_indexers: Dict[str, TokenIndexer] = None,
+                 pos_tag_indexers: Dict[str, TokenIndexer] = None,
```
matt-gardner: You don't need these anymore.

PR #941, opened 2018-03-01
get the dimensions correct for the transformer

DeNeutoy:


allennlp/modules/seq2seq_encoders/multi_head_self_attention.py
```diff
@@ -101,33 +93,33 @@ def forward(self,  # pylint: disable=arguments-differ
...
-        keys_per_head = torch.bmm(inputs_per_head, self._key_projections)
-        # shape (num_heads * batch_size, timesteps, attention_dim)
-        keys_per_head = keys_per_head.view(num_heads * batch_size,
-                                           timesteps,
-                                           self._attention_dim)
-
-        values_per_head = torch.bmm(inputs_per_head, self._value_projections)
-        # shape (num_heads * batch_size, timesteps, attention_dim)
-        values_per_head = values_per_head.view(num_heads * batch_size, timesteps, self._values_dim)
+        # Shape (batch_size, timesteps, 2 * attention_dim + values_dim)
```
DeNeutoy: @matt-gardner can you think of a better way to do this so I can avoid doing these multiple calls to `view` and `contiguous`?

allennlp/modules/seq2seq_encoders/multi_head_self_attention.py
```diff
@@ -142,10 +134,9 @@ def forward(self,  # pylint: disable=arguments-differ
         # which is equivalent to a weighted sum of the values with respect to
         # the attention distributions for each element in the num_heads * batch_size
         # dimension.
-        # shape (num_heads * batch_size, timesteps, values_dim)
+        # shape (num_heads * batch_size, timesteps, values_dim/num_heads)
         outputs = torch.bmm(attention, values_per_head)
```
matt-gardner: You could just actually use our `weighted_sum` here, which internally is `torch.bmm` for tensors with these dimensions.  Then you wouldn't need this long comment, and it'd be easier to read.

allennlp/modules/seq2seq_encoders/multi_head_self_attention.py
```diff
@@ -101,33 +93,33 @@ def forward(self,  # pylint: disable=arguments-differ
...
+        # Shape (batch_size, timesteps, 2 * attention_dim + values_dim)
+        combined_projection = self._combined_projection(inputs)
+
+        # Shape (batch_size, timesteps, attention_dim)
+        queries = combined_projection[:, :, :self._attention_dim].contiguous()
+        keys = combined_projection[:, :, self._attention_dim :2 * self._attention_dim].contiguous()
+        # Shape (batch_size, timesteps, values_dim)
+        values = combined_projection[:, :, 2 * self._attention_dim:].contiguous()
+
+        # Shape (num_heads * batch_size, timesteps, values_dim / num_heads)
```
matt-gardner: I feel like `sequence_length` is a better name here than `timesteps`, but it's already written this way, and it's not a huge deal if you don't want to change it.

And no, I can't think of a better way to do this than what you're doing.  The way you're doing the projection of the inputs basically forces you to do a transpose here.  Maybe you could do the transpose once, before you split the queries, keys and values?  It would change the semantics of the weights in your combined projection (i.e., currently weights for queries, keys, and values are grouped together, but if you do this the weights for the heads will be grouped together), but that should be fine, as they are all just learned parameters that are initialized identically, anyway.

DeNeutoy: Hmm i'm not sure that will save me much, because then i'll have to transpose again anyway to get the embedding dimension as the last dim. I'll leave it as it is for now and optimise it if it's too slow.

matt-gardner: Oh, I meant do a `.view()` to get four dimensions, and do a `.transpose(1, 2)`, like you're currently doing.  You can just do it once, instead of three times.  It's not a big deal, though, and yeah, optimize later if it's necessary.

PR #942, opened 2018-03-01
add label smoothing to loss

DeNeutoy:


allennlp/nn/util.py
```diff
@@ -427,11 +430,21 @@ def sequence_cross_entropy_with_logits(logits: torch.FloatTensor,
...
-    # Contribution to the negative log likelihood only comes from the exact indices
-    # of the targets, as the target distributions are one-hot. Here we use torch.gather
-    # to extract the indices of the num_classes dimension which contribute to the loss.
-    # shape : (batch * sequence_length, 1)
-    negative_log_likelihood_flat = - torch.gather(log_probs_flat, dim=1, index=targets_flat)
+    if label_smoothing is not None and label_smoothing > 0.0:
+        num_classes = logits.size(-1)
+        smoothing_value = label_smoothing / num_classes
+        smoothed_targets = ones_like(log_probs_flat) * smoothing_value
+        # Fill all the correct indices with 1 - smoothing value.
```
matt-gardner: Update the logic to match the tensorflow code, like we discussed.  Also, it'd be nice to give a simple example in a comment (or the docstring above) of what this will do to the targets.

allennlp/nn/util.py
```diff
@@ -427,11 +433,20 @@ def sequence_cross_entropy_with_logits(logits: torch.FloatTensor,
...
-    # Contribution to the negative log likelihood only comes from the exact indices
-    # of the targets, as the target distributions are one-hot. Here we use torch.gather
-    # to extract the indices of the num_classes dimension which contribute to the loss.
-    # shape : (batch * sequence_length, 1)
-    negative_log_likelihood_flat = - torch.gather(log_probs_flat, dim=1, index=targets_flat)
+    if label_smoothing is not None and label_smoothing > 0.0:
+        num_classes = logits.size(-1)
+        smoothing_value = label_smoothing / num_classes
+        # Fill all the correct indices with 1 - smoothing value.
+        one_hot_targets = zeros_like(log_probs_flat).scatter_(-1, targets_flat, 1.0 - label_smoothing)
```
matt-gardner: Why do you need a `scatter`?  Can't you just use `targets_flat` directly?

DeNeutoy: they are indices?

matt-gardner: Oh, duh.  Thanks =).  :flushed:

DeNeutoy: previously we could  just select the correct values using these indices, but now we have to explicitly do the multiplication and sum because the other values are non-zero, so it needs to be 1-hot.

PR #945, opened 2018-03-02
Allow to provide a custom `Module` in `Elmo` class

matt-peters:


allennlp/modules/elmo.py
```diff
@@ -53,18 +53,33 @@ class Elmo(torch.nn.Module):
         Should we apply layer normalization (passed to ``ScalarMix``)?
     dropout : ``float``, optional, (default = 0.5).
         The dropout to be applied to the ELMo representations.
+    module : ``torch.nn.Module``, optional, (default = None).
+        If provided, then use this module instead of the pre-trained ELMo biLM.
+        If using this option, then pass ``None`` for both ``options_file``
+        and ``weight_file``.  The module must provide a public attribute
+        ``num_layers`` with the number of internal layers and its ``forward``
```
DeNeutoy: You could check that the module has this attribute in the init.

matt-peters: It'll already throw an `AttributeError` in the init when trying to construct the `ScalarMix` if it doesn't exist.

PR #951, opened 2018-03-05
Modified linking score exceeds baseline

rajasagashe:
Edit: This is now 2.2 percent above the baseline.
Baseline (has features): 0.363
This (no features) + elmo + mixture weight + cosine similarity instead of dot product: 0.385

Outdated:
Baseline (has features): 0.238
This code (no features): 0.202
This code (no features) + character embeddings: 0.233

I think this modeling change captures almost as much information as features do.


matt-gardner:
Nice!  Can you try one more experiment?  Add back in the span feature, and see how performance changes.  The character CNN you're using won't capture span information, and this will give us some idea of how much information is lost because of that.

rajasagashe:
Are you saying 
- baseline with only single span features
to compare with baseline without any features.
or 
- This code (no features) + character embeddings + span feature

matt-gardner:
I'm saying the second one - I'm guessing the character features are doing a slightly better job than the hard-coded features at capturing match / lemma information, so performance goes up, but they are missing the information from the span feature, so performance goes down, and in the end you basically match the performance of the original model.  We can test this hypothesis by adding the span feature to the character-level model, and see if performance goes back up.

rajasagashe:
Should we merge this? If so any changes I should make?

matt-gardner:
Sorry this fell out of my radar - if you want me to look at something, be sure to request a review, or if you can't because I've already commented, you can add another comment saying it's ready to be looked at.  I'll look at this today.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -217,29 +220,44 @@ def forward(self,  # type: ignore
...
+                                                                     num_entity_tokens,
+                                                                     num_question_tokens) / self._embedding_dim
         # (batch_size, num_entities, num_question_tokens)
-        question_table_similarity_max_score, _ = torch.max(question_table_similarity, 2)
+        question_entity_similarity_max_score, _ = torch.max(question_entity_similarity, 2)
+
         # (batch_size, num_entities, num_question_tokens, num_features)
         linking_features = table['linking']
-        linking_scores = question_table_similarity_max_score
+        linking_scores = question_entity_similarity_max_score
```
matt-gardner: Looks like you don't need this line anymore, right?  Because you're assigning to `linking_scores` in both branches below.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -217,29 +220,44 @@ def forward(self,  # type: ignore
...
         if self._linking_params is not None:
             feature_scores = self._linking_params(linking_features).squeeze(3)
             linking_scores = linking_scores + feature_scores
+        else:
+            # pylint: disable=line-too-long
+            # The similarity between a question token and an entity's neighbors is useful when
+            # a column needs to be selected. For example, the question token might have no similarity
+            # with the column name, but is similar with the cells in the column.
+            # (batch_size, num_entities, num_neighbors, num_question_tokens)
+            question_entity_neighbor_similarity = util.batched_index_select(question_entity_similarity_max_score,
```
matt-gardner: The other option is to make this optional independent of whether you also have linking features (e.g., if you want to use both linking features and this kind of neighbor similarity).  Probably not worth the additional complexity, though, at this point.

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
```diff
@@ -217,29 +220,44 @@ def forward(self,  # type: ignore
...
         linking_features = table['linking']
-        linking_scores = question_table_similarity_max_score
+        linking_scores = question_entity_similarity_max_score
+
         if self._linking_params is not None:
             feature_scores = self._linking_params(linking_features).squeeze(3)
             linking_scores = linking_scores + feature_scores
+        else:
+            # pylint: disable=line-too-long
+            # The similarity between a question token and an entity's neighbors is useful when
```
matt-gardner: Can you add a note to the docstring for `num_linking_features` that says what we're doing with neighbor similarity when you give no features?  Also, the first sentence in the comment here should mention _what_ you're doing, because a reader doesn't know that.  Your second sentence can say _why_ you're doing this, which is what you currently have.

tests/fixtures/encoder_decoder/wikitables_semantic_parser_no_features/experiment.json
```diff
@@ -0,0 +1,80 @@
...
+    },
+    "decoder_trainer": {
+      "type": "max_marginal_likelihood"
+    },
+    "decoder_beam_search": {
+      "beam_size": 3
+    },
+    "max_decoding_steps": 200,
+    "attention_function": {"type": "dot_product"},
+    "num_linking_features": 8
```
matt-gardner: I think you meant this to be 0, not 8.

matt-gardner:
I moved some files around, so you'll have to rebase this and fix some conflicts.  Sorry...

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
rajasagashe: done

allennlp/models/encoder_decoders/wikitables_semantic_parser.py
rajasagashe: done

tests/fixtures/encoder_decoder/wikitables_semantic_parser_no_features/experiment.json
rajasagashe: done

PR #953, opened 2018-03-06
use biased estimator for layernorm, fix some views in multi head attention

DeNeutoy:
This code is element-wise the same as the SRL transformer implementation.

### DIFF
--- a/allennlp/modules/layer_norm.py
+++ b/allennlp/modules/layer_norm.py
@@ -34,5 +34,5 @@ class LayerNorm(torch.nn.Module):
 
     def forward(self, tensor: torch.Tensor):  # pylint: disable=arguments-differ
         mean = tensor.mean(-1, keepdim=True)
-        std = tensor.std(-1, keepdim=True)
+        std = tensor.std(-1, unbiased=False, keepdim=True)
         return self.gamma * (tensor - mean) / (std + self.eps) + self.beta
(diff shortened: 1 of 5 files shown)


allennlp/modules/layer_norm.py
```diff
@@ -34,5 +34,5 @@ def __init__(self,
 
     def forward(self, tensor: torch.Tensor):  # pylint: disable=arguments-differ
         mean = tensor.mean(-1, keepdim=True)
-        std = tensor.std(-1, keepdim=True)
+        std = tensor.std(-1, unbiased=False, keepdim=True)
```
matt-peters: Will this have un-intended impacts on other modules who also use the layer norm?  Maybe allow this to be a parameter in the constructor so clients can choose which estimator to use?

allennlp/modules/seq2seq_encoders/multi_head_self_attention.py
```diff
@@ -127,17 +126,22 @@ def forward(self,  # pylint: disable=arguments-differ
...
-        # with num_heads being the first dimension, so reshaping naively would not
-        # throw an error, but give an incorrect result.
-        outputs = torch.cat(torch.split(outputs, batch_size, dim=0), dim=-1)
+        # shape (batch_size, num_heads, timesteps, values_dim/num_heads)
+        outputs = outputs.view(batch_size, num_heads, timesteps, int(self._values_dim / num_heads))
+        # shape (batch_size, timesteps, num_heads, values_dim/num_heads)
+        outputs = outputs.transpose(1, 2).contiguous()
+        # shape (batch_size, timesteps, values_dim)
+        outputs = outputs.view(batch_size, timesteps, self._values_dim)
+        #outputs = outputs.view(batch_size, timesteps, -1)
```
matt-peters: Remove this commented out line.

allennlp/modules/layer_norm.py
DeNeutoy: This just matches the tensorflow implementation - in practice, it results in dividing by n instead of n-1 when calculating the variance, i.e using the sample variance rather than the estimated population variance. So overall, it's unlikely to make a large difference for any substantial layers and it's also very low level to be configurable.

PR #965, opened 2018-03-08
Fix max_vocab_size bug (#964)

### LINKED ISSUE #964: Vocabulary: max_vocab parameter does not work
Vocabulary always builds from all the word types

maxdel:
There were 2 problems here:
1) `max_vocab_size = defaultdict(lambda: max_vocab_size)`
default value for the new dictionary becomes `defaultdict` itself, and not the `int` as supposed
2) `max_vocab = max_vocab_size.get(namespace)`
probably it supposed to be `max_vocab_size[namespace]` instead (since it also adds a new element to the `defaultdict`)

Fixes #964


matt-gardner:
Can you also add a test case that fails without this change, to make sure this stays fixed?

maxdel:
Yes, sure.

maxdel:
Messed up with git a bit, refer to #966 instead.

PR #971, opened 2018-03-09
Parameterized selection of in-agenda actions

pdasigi:
The model used to assign uniform probabilities to all allowed in-agenda actions. This PR makes this parameterized by changing the output projection layer to project the concatenation of the current hidden state, attended encoder outputs, and now also the checklist difference. Doing this eliminates the `checklist_selection_weight` hyperparameter, which was the mixing weight for in-agenda probabilities and model assigned probabilities. Also, this makes our plans for eventually using linking scores in the checklist easier.

PR #972, opened 2018-03-10
Dynamically decrease checklist cost weight between epochs

pdasigi:
This depends on #971. I'll rebase this branch after that PR gets merged.

### DIFF
--- a/allennlp/data/iterators/epoch_tracking_bucket_iterator.py
+++ b/allennlp/data/iterators/epoch_tracking_bucket_iterator.py
@@ -1,0 +1,74 @@
+import logging
+from typing import List, Tuple, Dict, Iterable, Generator, Union
+from collections import defaultdict
+
+from overrides import overrides
+import numpy
+
+from allennlp.data.fields import MetadataField
+from allennlp.data.instance import Instance
+from allennlp.data.iterators.data_iterator import DataIterator
+from allennlp.data.iterators.bucket_iterator import BucketIterator
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+@DataIterator.register("epoch_tracking_bucket")
+class EpochTrackingBucketIterator(BucketIterator):
+    """
+    This is essentially a :class:`allennlp.data.iterators.BucketIterator` with just one difference.
+    It keeps track of the epoch number, and adds that as an additional meta field to each instance.
+    That way, ``Model.forward`` will have access to this information. We do this by keeping track of
+    epochs globally, and incrementing them whenever the iterator is called. However, the iterator is
+    called both for training and validation sets. So, we keep a dict of epoch numbers, one key per
+    dataset.
+
+    Parameters
+    ----------
+    See :class:`BucketIterator`.
+    """
+    def __init__(self,
+                 sorting_keys: List[Tuple[str, str]],
+                 padding_noise: float = 0.1,
+                 biggest_batch_first: bool = False,
+                 batch_size: int = 32,
+                 instances_per_epoch: int = None,
+                 max_instances_in_memory: int = None) -> None:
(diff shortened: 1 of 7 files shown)


allennlp/data/iterators/epoch_tracking_bucket_iterator.py
```diff
@@ -0,0 +1,74 @@
...
+            for _ in range(num_epochs):
+                self._add_epoch_num_to_instances(instances, dataset_id)
+                yield from self._yield_one_epoch(instances, shuffle, cuda_device, for_training)
+                self._global_epoch_nums[dataset_id] += 1
+
+    def _add_epoch_num_to_instances(self,
+                                    instances: Iterable[Instance],
+                                    dataset_id: int) -> None:
+        for instance in instances:
+            instance.fields["epoch_num"] = MetadataField(self._global_epoch_nums[dataset_id])
```
pdasigi: Not sure why mypy doesn't like this assignment.

allennlp/models/encoder_decoders/nlvr_semantic_parser.py
```diff
@@ -66,6 +69,11 @@ class NlvrSemanticParser(Model):
         Mixture weight (0-1) for combining coverage cost and denotation cost. As this increases, we
         weigh the coverage cost higher, with a value of 1.0 meaning that we do not care about
         denotation accuracy.
+    dynamic_cost_weight : ``Dict[str, Union[int, float]]``
+        A dict containing keys ``wait_num_steps`` and ``rate`` indicating the number of steps after
+        which we should start decreasing the weight on checklist cost in favor of denotation cost,
+        and the rate at which we should do it. The weight will remain constant if this is an empty
```
matt-gardner: Giving an equation here would be nice, because "the rate at which we should do it" can mean lots of things.

tests/fixtures/encoder_decoder/nlvr_semantic_parser/experiment.json
```diff
@@ -46,10 +46,14 @@
...
     "checklist_cost_weight": 0.8,
+    "dynamic_cost_weight": {
+      "wait_num_epochs": 0,
+      "rate": 0.1
+    },
     "penalize_non_agenda_actions": true
   },
   "iterator": {
-    "type": "bucket",
+    "type": "epoch_tracking_bucket",
```
matt-gardner: This gives you a high test coverage number, but doesn't actually test anything other than that this doesn't crash.  It'd be nice to actually test this, in two places:

1. A test for the `EpochTrackingBucketIterator`, that makes sure that the epoch gets incremented correctly, for two different datasets.
2. A simple test for your model that checks that the checklist weight parameter is updated like you expect when you get different epoch numbers passed to `forward`.

allennlp/models/encoder_decoders/nlvr_semantic_parser.py
```diff
@@ -66,6 +69,11 @@ class NlvrSemanticParser(Model):
         Mixture weight (0-1) for combining coverage cost and denotation cost. As this increases, we
         weigh the coverage cost higher, with a value of 1.0 meaning that we do not care about
         denotation accuracy.
+    dynamic_cost_weight : ``Dict[str, Union[int, float]]``
+        A dict containing keys ``wait_num_steps`` and ``rate`` indicating the number of steps after
```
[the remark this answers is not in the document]
pdasigi: Fixed. Thanks!

allennlp/models/encoder_decoders/nlvr_semantic_parser.py
pdasigi: Done.

tests/fixtures/encoder_decoder/nlvr_semantic_parser/experiment.json
pdasigi: Added two tests. Thanks!

PR #947, opened 2018-03-03
 Added handling of fb:part logical forms and linking

matt-gardner:
This gets most of the missing `fb:part.[whatever]` logical forms.  There are still some odd errors where we encounter unmapped names when parsing logical forms, but there aren't that many of them now.  I added parts as entities in the `KnowledgeGraph`, so our linking code will handle them, but they have no neighbors in the graph, so the neighbor code _should_ just ignore them.

This PR depends on #936, which in turn depends on #915.

PR #974, opened 2018-03-12
Fix key error, update training configuration

matt-gardner:
This is actually a very small change, but it depends on #947.  @rajasagashe, this should fix the key errors we were seeing.  It also has an updated training config, with optimizer parameters that match those in the original paper.  I had more luck with using a learning rate of 0.01, though - the 0.1 run I did crashed with an error I'm still looking into.


rajasagashe:
Thanks! Any reason why in the "question_token_indexers" field of the config file, "lowercase_tokens" is no longer set to true?

matt-gardner:
The question and table are both now hard-coded to always be lowercased (because it sure looks like that's what SEMPRE does - the questions in the training data are always lowercase).  So that flag isn't necessary.

PR #975, opened 2018-03-12
fix span construction

DeNeutoy:
This fixes a bug where if a single word pre-terminal was added after a POS tag span of the same word, it would be concatenated to it and hence ignored. I added a better test where this occurs, rather than the toy example I had originally.

### DIFF
--- a/tests/data/dataset_readers/penn_tree_bank_reader_test.py
+++ b/tests/data/dataset_readers/penn_tree_bank_reader_test.py
@@ -105,7 +105,7 @@ class TestPennTreeBankConstituencySpanReader(AllenNlpTestCase):
 
     def test_get_gold_spans_correctly_extracts_spans_with_nested_labels(self):
         ptb_reader = PennTreeBankConstituencySpanDatasetReader()
-        # Here we have a parse with several nested labels - particularly the (WHNP(WHNP (WP What)))
+        # Here we have a parse with several nested labels - particularly the (WHNP (WHNP (WP What)))
         # fragment. These should be concatenated into a single label by get_gold_spans.
         tree = Tree.fromstring("""
             (S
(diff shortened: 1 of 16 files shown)


tests/data/dataset_readers/penn_tree_bank_reader_test.py
```diff
@@ -101,18 +101,56 @@ def test_get_gold_spans_correctly_extracts_spans(self):
...
-                         ((2, 2), 'V-POS'), ((3, 3), 'D-POS'), ((4, 4), 'N-POS'),
-                         ((3, 4), 'NP'), ((2, 4), 'VP'), ((0, 4), 'S')]
+        assert spans == [((0, 1), 'NP'), ((3, 4), 'NP'), ((2, 4), 'VP'), ((0, 4), 'S')]
 
     def test_get_gold_spans_correctly_extracts_spans_with_nested_labels(self):
         ptb_reader = PennTreeBankConstituencySpanDatasetReader()
-        # Here we have a sentence fragment which has the same span with nested S and VP labels.
-        # These should be concatenated into a single label by get_gold_spans.
-        tree = Tree.fromstring("(S (VP (V chased) (NP (D the) (N cat))))")
+        # Here we have a parse with several nested labels - particularly the (WHNP(WHNP (WP What)))
```
matt-gardner: Missing a space in `(WHNP(WHNP`

DeNeutoy:
Yes it does - previously if you had a nested non-terminal whose only leaf was a single word, e.g `(S (NP (N Mark)))`, this would be extracted as `{(0,0): "S-NP-N-POS"}`, which we then would have ignored because we then filtered by labels containing "POS". Now it is correctly extracted as `{(0,0): "S-NP"}`.

matt-gardner:
Sorry, I now see that you put that in the PR description - somehow I got to looking at the code before reading that.  Sorry to make you repeat yourself, and LGTM.

PR #976, opened 2018-03-13
Fix issue #973 and add predictor for SimpleSeq2Seq

### LINKED ISSUE #973: [BUG] Wrong target namespace in SimpleSeq2Seq.decode()
At line:
https://github.com/allenai/allennlp/blob/master/allennlp/models/encoder_decoders/simple_seq2seq.py#L280

I think 
`self.vocab.get_token_from_index(x, namespace="target_tokens")`
should be:
`self.vocab.get_token_from_index(x, namespace=self._target_namespace)`

howardyclo:
Fixes #973

### DIFF
--- a/allennlp/training/metrics/evalb_bracketing_scorer.py
+++ b/allennlp/training/metrics/evalb_bracketing_scorer.py
@@ -43,9 +43,6 @@ class EvalbBracketingScorer(Metric):
         self._evalb_program_path = os.path.join(evalb_directory_path, "evalb")
         self._evalb_param_path = os.path.join(evalb_directory_path, evalb_param_filename)
 
-        if not os.path.exists(self._evalb_program_path):
-            raise ConfigurationError("You must compile the EVALB scorer before using it."
-                                     " Run 'make' in the 'scripts/EVALB' directory.")
 
         self._header_line = ['ID', 'Len.', 'Stat.', 'Recal', 'Prec.', 'Bracket',
                              'gold', 'test', 'Bracket', 'Words', 'Tags', 'Accracy']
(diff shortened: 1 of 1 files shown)


matt-gardner:
Thanks for this!  Adding a predictor requires adding some tests that are a little bit trickier, because you need to train fixtures and add them to the repo.  I'll take care of this soon, unless you want to figure it out.  It'll be really quick for me, so unless you really want to figure out what's necessary here, I'm happy to do it.

howardyclo:
@matt-gardner Hello, I'm not familiar with how allennlp's tests are done, otherwise I would be happy to help out if you can give me some instructions to do?

matt-gardner:
@howardyclo, I just added a simple test for the predictor.  It just makes sure that the predictor gives the expected outputs when it's given its expected inputs.  In order for the test to work, I had to train a toy model that the predictor could use, and I used `scripts/train_fixtures.py` for that (after some minor updates to that script, to add the seq2seq model to the fixtures that get trained).

@DeNeutoy, if you could give a quick look to the changes I made to `scripts/train_fixtures.py`, that'd be great.

allennlp/service/predictors/simple_seq2seq.py
```diff
@@ -0,0 +1,19 @@
...
+
+@Predictor.register('simple_seq2seq')
+class SimpleSeq2SeqPredictor(Predictor):
+    """
+    Wrapper for the :class:`~allennlp.models.encoder_decoder.simple_seq2seq` model.
+    """
+    @overrides
+    def _json_to_instance(self, json_dict: JsonDict) -> Tuple[Instance, JsonDict]:
+        """
+        Expects JSON that looks like ``{"source_string": "..."}``.
```
DeNeutoy: Maybe no need to describe the type here, I think - just "source" or "source_sentence" would be fine.

howardyclo: Hello, why naming "source_string" is just due to the consideration of being consistent to the argument name at seq2seq dataset reader's text_to_instance() 
See line at
https://github.com/allenai/allennlp/blob/master/allennlp/data/dataset_readers/seq2seq.py#L81

matt-gardner: @DeNeutoy, yeah, I agree, I think "source" is better here.  But I'm going to change that in a subsequent PR, because something got messed up with the git repo this PR is coming from (it says from "unknown repository" above).

PR #983, opened 2018-03-14
Keyboard interrupt + training admin

DeNeutoy:
- Adds a catch for Keyboard interrupts during training, which then tries to make a model archive using the serialisation directory before raising
- Pulls some tensorboard logging out into methods
- Adds a `prepare_global_logging` function to configure Tqdm, `logging` and stdout/stderr streams

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -271,7 +271,7 @@ class Trainer:
         self._histogram_interval = histogram_interval
         self._log_histograms_this_batch = False
         # We keep the total batch number as a class variable because it
-        # is used inside a closure for the hook which logs activations in 
+        # is used inside a closure for the hook which logs activations in
         # ``_enable_activation_logging``.
         self._batch_num_total = 0
 
@@ -497,9 +497,9 @@ class Trainer:
 
         return False
 
-    def _parameter_and_gradient_statistics_to_tensorboard(self,
+    def _parameter_and_gradient_statistics_to_tensorboard(self, # pylint: disable=invalid-name
                                                           epoch: int,
-                                                          batch_grad_norm: float) -> None: # pylint: disable=invalid-name
+                                                          batch_grad_norm: float) -> None:
         """
         Send the mean and std of all parameters and gradients to tensorboard, as well
         as logging the average gradient norm.
@@ -535,8 +535,8 @@ class Trainer:
         for name, param in self._model.named_parameters():
             if name in histogram_parameters:
                 self._tensorboard.add_train_histogram("parameter_histogram/" + name,
-                                                        param,
-                                                        epoch)
+                                                      param,
+                                                      epoch)
 
     def _metrics_to_tensorboard(self,
                                 epoch: int,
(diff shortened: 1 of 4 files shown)


allennlp/training/trainer.py
```diff
@@ -465,39 +465,13 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
...
-                            grad_data = param.grad.data._values()
-                        else:
-                            grad_data = param.grad.data
-                        self._tensorboard.add_train_scalar("gradient_mean/" + name,
-                                                           grad_data.mean(),
-                                                           batch_num_total)
-                        self._tensorboard.add_train_scalar("gradient_std/" + name,
-                                                           grad_data.std(),
-                                                           batch_num_total)
+                self._parameter_and_gradient_statistics_to_tensorboard(batch_num_total)
```
matt-gardner: Why do we have both `batch_num_total` and `batch_num`?

allennlp/training/trainer.py
```diff
@@ -465,39 +465,13 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
...
                 self._tensorboard.add_train_scalar("loss/loss_train", metrics["loss"], batch_num_total)
                 self._metrics_to_tensorboard(batch_num_total,
                                              {"epoch_metrics/" + k: v for k, v in metrics.items()})
-                # norm of gradients
-                if self._batch_grad_norm is not None:
-                    self._tensorboard.add_train_scalar("gradient_norm",
-                                                       self._batch_grad_norm,
```
matt-gardner: I think there's no reason to have `_batch_grad_norm` be a field on `self` - you can just return it directly from `self._rescale_gradients` and not pollute the class state with unnecessary fields.  And then pass it in your method call above.

allennlp/training/trainer.py
```diff
@@ -465,39 +465,13 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
...
-                                                       batch_num_total)
 
-            # Histogram logging
             if self._log_histograms_this_batch:
```
matt-gardner: You have this check twice; once here and once inside the method you're calling.

allennlp/training/trainer.py
```diff
@@ -458,14 +458,14 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
                 self._optimizer.step()
 
             # Update the description with the latest metrics
-            metrics = self._get_metrics(train_loss, batch_num)
```
matt-gardner: Oh, sorry, looks like I misled you with my question.  There was a good reason that we had both of them.  We need to keep this distinction - `batch_num` is used to divide the loss by the number of batches, and really needs to be reset every epoch, while `batch_num_total` is used to keep a global counter for tensorboard.  The names are so similar, though, that it wasn't obvious that they were both necessary.  Maybe `batches_this_epoch` and `total_batches_so_far`?

DeNeutoy: ugh, i'm dumb.

matt-gardner: No, you're really good at this.  It's just that names are important, or it's really easy to misunderstand what code is doing - I misunderstood it too, at first.

PR #985, opened 2018-03-15
Constituency parser demo

DeNeutoy:
- Makes the task names in the header smaller so they fit on a mac screen (temp fix until Aaron works his magic with the new layout)
- Adds `PaneTop` and `PaneBottom` React classes so I could lay out the parser demo top to bottom, because the parses take up a lot of screen width and look kind of dumb if they are in a half screen div.
- Add s3 links and sniff tests to the constituency parser.
- Adds some Hierplane styling so that e.g all POS tags render as the same colour nodes in the tree.
- Adds the config that was used to train the parser.


.pylintrc
```diff
@@ -65,7 +65,7 @@ confidence=
...
+disable=import-star-module-level,old-octal-literal,oct-method,print-statement,unpacking-in-except,parameter-unpacking,backtick,old-raise-syntax,old-ne-operator,long-suffix,dict-view-method,dict-iter-method,metaclass-assignment,next-method-called,raising-string,indexing-exception,raw_input-builtin,long-builtin,file-builtin,execfile-builtin,coerce-builtin,cmp-builtin,buffer-builtin,basestring-builtin,apply-builtin,filter-builtin-not-iterating,using-cmp-argument,useless-suppression,range-builtin-not-iterating,suppressed-message,no-absolute-import,old-division,cmp-method,reload-builtin,zip-builtin-not-iterating,intern-builtin,unichr-builtin,reduce-builtin,standarderror-builtin,unicode-builtin,xrange-builtin,coerce-method,delslice-method,getslice-method,setslice-method,input-builtin,round-builtin,hex-method,nonzero-method,map-builtin-not-iterating,missing-docstring,too-many-arguments,too-many-locals,too-many-statements,too-many-branches,too-many-nested-blocks,too-many-instance-attributes,fixme,too-few-public-methods,no-else-return,consider-iterating-dictionary
```
joelgrus: what does this pylint disable do?

allennlp/service/predictors/constituency_parser.py
```diff
@@ -10,6 +11,52 @@
 from allennlp.data.tokenizers.word_splitter import SpacyWordSplitter
 
 
+# Make the links to POS tag nodes render as "pos",
+# to distinguish them from constituency tags. The
+# actual tag is still visible within the node.
+LINK_TO_LABEL = {x: "pos" for x in TAG_MAP.keys()}
```
joelgrus: oh, I think I see what pylint wants. why do you use `.keys()` here? iterating over a dict already iterates over the keys

allennlp/service/predictors/constituency_parser.py
```diff
@@ -10,6 +11,52 @@
...
+
+# POS tags have a unified colour.
+NODE_TYPE_TO_STYLE = {x: ["color0"] for x in TAG_MAP.keys()}
```
joelgrus: same

demo/public/css/main.css
```diff
@@ -807,7 +807,7 @@ header {
 
 .header__content .nav__link span {
   font-size: 1em;
-  font-size: 1.125em;
+  font-size: 1em;
```
joelgrus: are these two lines the same? should we delete one of them?

demo/src/components/Pane.js
```diff
@@ -60,13 +60,30 @@ export class PaneRight extends React.Component {
       const { outputState } = this.props;
 
       return (
-        <ResultDisplay outputState={outputState}>
+        <ResultDisplay orientation="pane__right" outputState={outputState}>
```
joelgrus: I find this naming confusing, "pane__right" doesn't feel like an "orientation" to me, it's a class name for a subcomponent.

I'm not 100% sure what I prefer, maybe something like `resultPane="right"` or `resultPane="bottom"`, and then up in line 45 you could do 

```
className={`pane__${resultPane} model__output ...`}
```

allennlp/service/predictors/constituency_parser.py
DeNeutoy: I know, which is heinous in my opinion, and _always_ throws me when i'm reading code

demo/public/css/main.css
DeNeutoy: I'm just editing what's here... I guess they are the same? I don't know if setting multiple CSS attributes can have back-off effects or something.

allennlp/service/predictors/constituency_parser.py
joelgrus: we say we roughly follow Google style guide, which doesn't like `.keys()`

https://google.github.io/styleguide/pyguide.html?showone=Default_Iterators_and_Operators#Default_Iterators_and_Operators

demo/public/css/main.css
joelgrus: almost certainly it was a mistake in the original to set `font-size` twice, it's just more clearly one when we're setting it to the same thing both times :)

allennlp/service/predictors/constituency_parser.py
DeNeutoy: fairsies

PR #986, opened 2018-03-15
Fix DecoderTrainer API, split semantic parsing files by class

matt-gardner:
Depends on #982.  I split the semantic parser files into three, one for each of the main classes, and I made the `DecoderTrainer` generic on the type of supervision.  This required moving the "cost" logic out of `NlvrDecoderState` and into the `NlvrSemanticParser`, but I think this a good thing anyway, as having that function on a `DecoderState` muddied the API too much.

Next up: I'm going to simplify the `DecoderState` stuff, creating several internal groups, like `RnnState`.  After that, I'll see what I can do about simplifying the `DecoderSteps`.


matt-gardner:
Ok @pdasigi, this should be ready for review.  Sorry that it looks so large.  That's because I split up the WikiTables and NLVR files into three, and moved a few things around as I did it.  It might be easiest to either (1) first look at the changes that I made to `decoder_trainer.py`, `decoder_state.py`, and the concrete `DecoderTrainers`, then look at the individual files, which I only changed to match the changes there, or (2) look at the individual commits, most of which (after the first one) were making changes to files after they had been split.

allennlp/models/encoder_decoders/nlvr/nlvr_semantic_parser.py
```diff
@@ -0,0 +1,614 @@
...
+        self._agenda_coverage = Average()
+        check_dimensions_match(nonterminal_embedder.get_output_dim(),
+                               terminal_embedder.get_output_dim(),
+                               "nonterminal embedding dim",
+                               "terminal embedding dim")
+        self._nonterminal_embedder = nonterminal_embedder
+        self._terminal_embedder = terminal_embedder
+        self._encoder = encoder
+        self._decoder_trainer: DecoderTrainer[Callable[[NlvrDecoderState], torch.Tensor]] = \
+                ExpectedRiskMinimization(beam_size, normalize_beam_score_by_length, max_decoding_steps)
```
pdasigi: Why hard-code the trainer? Would it be difficult to make this general enough to pass the trainer to the init for training baseline models?

allennlp/models/encoder_decoders/nlvr/nlvr_semantic_parser.py
```diff
@@ -0,0 +1,614 @@
...
+        output_dict["predicted_actions"] = [action_strings]
+        return output_dict
+
+    def _get_state_cost(self, state: NlvrDecoderState) -> torch.Tensor:
+        """
+        Return the costs a finished state. Since it is a finished state, the group size will be 1,
+        and hence we'll return just one cost.
+        """
+        if not state.is_finished():
+            raise RuntimeError("get_costs() is not defined for unfinished states!")
```
pdasigi: `get_costs` -> `_get_state_cost`

allennlp/nn/decoding/expected_risk_minimization.py
```diff
@@ -24,24 +24,21 @@ class ExpectedRiskMinimization(DecoderTrainer):
 
     Parameters
     ----------
-    beam_size : ``int``
-    noramlize_by_length : ``bool``
-        Should the log probabilities be normalized by length before renormalizing them? Edunov et
-        al. do this in their work.
```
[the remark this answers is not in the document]
matt-gardner: Because I was moving them around, thinking at first that I'd just get rid of the constructor, and forgot to add them back in.  Fixed.

allennlp/nn/decoding/decoder_trainer.py
```diff
@@ -25,11 +25,5 @@ class DecoderTrainer(Registrable):
     def decode(self,
```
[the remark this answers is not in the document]
matt-gardner: Thanks, that's a pretty bad omission.

allennlp/models/encoder_decoders/nlvr/nlvr_semantic_parser.py
matt-gardner: Because it's not really feasible to swap out different trainers.  Taking these dependencies as inputs instead of instantiating them directly in the model is only useful when there are multiple valid alternatives and we want to experiment with them.  That's not true of our `DecoderTrainers`, and it's not clear to me that it will ever be true.  Also, because `DecoderTrainer` is generic over the supervision type, it only really makes sense to allow `DecoderTrainers` that match the supervision type that your model has, and we can't enforce that with our dependency injection.

If you find yourself wanting to try a few different training options, you'll first have to make sure that your model has the right supervision available for all of them.  Then you can add some flags to switch between them, but I'd still just instantiate them directly here.  _Maybe_ there will be a day when we have enough options and a very flexible model, and it will make sense to allow this dependency to be injected again.  But for now, it's just added complexity that isn't necessary.

allennlp/models/encoder_decoders/nlvr/nlvr_semantic_parser.py
matt-gardner: Done.

allennlp/models/encoder_decoders/nlvr/nlvr_semantic_parser.py
pdasigi: Agreed.

PR #987, opened 2018-03-15
add allennlp command

### LINKED ISSUE #946: pip should install an `allennlp` script
lots of python packages do that, so that people can then

```bash
allennlp train ...
```

instead of 

```
python -m allennlp.run train ...
```

should be relatively easy:

http://python-packaging.readthedocs.io/en/latest/command-line-scripts.html

DeNeutoy:
This was waaaay easier than I thought 👍 
Fixes #946

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,6 @@
 include README.md
 include requirements.txt
+include bin/allennlp
 recursive-include scripts *
 recursive-include tutorials *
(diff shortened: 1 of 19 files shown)


dimidd:
I have allennlp (0.4.0) installed via pip, and the command `allennlp` isn't found. I had to resort to `python3 -m allennlp.run`.

PR #991, opened 2018-03-16
Add rnn state

matt-gardner:
Depends on #986.  This PR adds an `RnnState` object to hold all of the decoder-RNN-related state.


matt-gardner:
Ok @pdasigi, this one should now be ready to review.  It looks a lot bigger than it is because it includes some moved tests that were missed in the last PR, splitting up the `WikiTables` tests into two files.  The only changes here are grouping together all RNN-related state into an `RnnState` object, and changes to model code that was required by this grouping.

I'm planning on doing a similar thing for action-related state soon, but I'm fixing the action embeddings first.

tests/models/encoder_decoders/wikitables/wikitables_decoder_step_test.py
```diff
@@ -0,0 +1,346 @@
...
+        valid_actions_1 = {'e': [0, 1, 2, 4]}
+        valid_actions_2 = {'e': [0, 1, 3]}
+        valid_actions_3 = {'e': [2, 3, 4]}
+        self.state.grammar_state[0] = GrammarState(['e'], {}, valid_actions_1, {})
+        self.state.grammar_state[1] = GrammarState(['e'], {}, valid_actions_2, {})
+        self.state.grammar_state[2] = GrammarState(['e'], {}, valid_actions_3, {})
+        considered, to_embed, to_link = WikiTablesDecoderStep._get_actions_to_consider(self.state)
+        # These are _global_ action indices.  All of the actions in this case are embedded, so this
+        # is just a mapping from the valid actions above to their global ids.
+        expected_to_embed = [[0, 1, 2, 5], [0, 3, 4], [2, 3, 5]]
```
pdasigi: Why is the first element not `[1, 0, 2, 5]`, given that `self.action_indices[(0, 0)] = 1` and `self.action_indices[(0, 1)] = 0`?

tests/models/encoder_decoders/wikitables/wikitables_semantic_parser_test.py
```diff
@@ -0,0 +1,310 @@
...
+                            expected_action_embedding.cpu().data.numpy())
+
+        # Finally, we'll check that the embedding for the initial action is as we expect.
+        start_embedding = nonterminal_embedding(start_tensor).squeeze(0)
+        zeros = Variable(start_embedding.data.new(start_embedding.size(-1)).fill_(0).float())
+        expected_action_embedding = torch.cat([zeros, start_embedding], dim=-1)
+        assert_almost_equal(initial_action_embedding.cpu().data.numpy(),
+                            expected_action_embedding.cpu().data.numpy())
+
+    def test_embed_actions_does_not_embed_terminals_when_set(self):
```
pdasigi: You mean `test_embed_actions_does_not_embed_terminals_when_not_set`?

tests/models/encoder_decoders/wikitables/wikitables_decoder_step_test.py
matt-gardner: Because they are sorted by global action index.  This sorting is removed in the next PR, and the PR after that will probably rework how this works entirely, so I'm not going to worry about changing anything here.

tests/models/encoder_decoders/wikitables/wikitables_semantic_parser_test.py
matt-gardner: This test is also removed in the next PR.

PR #1000, opened 2018-03-19
Bidaf Ensemble

schmmd:
This PR adds an example of how to create an ensemble model.  The included `BidafEnsemble` raises the scores on BiDAF as follows by using re-trained models with random seeds.  These models scored 67.9, 68.4, 67.7, 67.7, 67.9 respectively.

* http://beaker.allenai.org/ex/ex_y1mryavksct7/tasks
* http://beaker.allenai.org/ex/ex_9tbbbu5kqc77/tasks
* http://beaker.allenai.org/ex/ex_o5antz70er9e/tasks
* http://beaker.allenai.org/ex/ex_p9982g8l9vch/tasks
* http://beaker.allenai.org/ex/ex_wgq139q91xjg/tasks

Here are the scores:

Original BiDAF: 0.716 EM, 0.799 F1
3-Ensemble: 0.684 EM, 0.779 F1
5-Ensemble: 0.725 EM, .806 F1


allennlp/models/model.py
```diff
@@ -221,7 +233,10 @@ def load(cls,
 
         # Load vocabulary from file
         vocab_dir = os.path.join(serialization_dir, 'vocabulary')
-        vocab = Vocabulary.from_files(vocab_dir)
+        if os.path.exists(vocab_dir):
```
schmmd: The Ensemble classes don't need a vocabulary.  I don't particularly like how I'm handling this option here, but the other changes I come up with require many more changes to our APIs.  Suggestions very welcome ;-)

allennlp/models/model.py
```diff
@@ -231,8 +246,9 @@ def load(cls,
         # want the code to look for it, so we remove it from the parameters here.
         _remove_pretrained_embedding_params(model_params)
         model = Model.from_params(vocab, model_params)
-        model_state = torch.load(weights_file, map_location=util.device_mapping(cuda_device))
-        model.load_state_dict(model_state)
+        if not model.is_ensemble:
```
schmmd: Alternatively I could handle this by checking whether `Ensemble` is a base class.

schmmd:
@matt-gardner @DeNeutoy I'd appreciate any alternative ideas to my rather yucky changes in allennlp/models/model.py, or any other high-level feedback about this approach.

This PR still needs additional comments and tests.  This afternoon I'm going to run some BiDAF trainings with different seeds.

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,111 @@
...
+    def forward(self,  # type: ignore
+                question: Dict[str, torch.LongTensor],
+                passage: Dict[str, torch.LongTensor],
+                span_start: torch.IntTensor = None,
+                span_end: torch.IntTensor = None,
+                metadata: List[Dict[str, Any]] = None) -> Dict[str, torch.Tensor]:
+
+        subresults = []
+        for i, submodel in enumerate(self.submodels):
+            subresults.append(self.submodels[0].forward(question, passage, span_start, span_end, metadata))
```
matt-gardner: Shouldn't this be `submodel.forward(...)` instead of `self.submodels[0].forward(...)`?

schmmd: Yes, sorry--leftover troubleshooting code from when I was trying to figure out why my submodels were outputting different results (because they were not in `eval` mode).  I wouldn't look over this block too closely--I have not written any tests or tried it with differing submodels.

allennlp/models/model.py
schmmd: I could require Ensembles to have a vocabulary--and then just check to verify that it matches the vocabulary in the submodels.

matt-gardner: Can you just override `load` on `Ensemble`?

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __init__(self,
+                 submodels: List[BidirectionalAttentionFlow]) -> None:
+        super(BidafEnsemble, self).__init__(submodels)
+
+        # Setting this propagates calls to .eval() so dropout is disabled on the submodels in evaluation
+        # and prediction.
+        self.linears = torch.nn.ModuleList(submodels)
+
+        self.submodels = submodels
```
matt-gardner: You should be able to just do `self.submodels = torch.nn.ModuleList(submodels)`, without having `self.linears` at all.

allennlp/models/model.py
schmmd: I don't think so--this is a `classmethod` called via `Model.load`.

Yikes--note that GitHub makes it extremely confusing what method this change is actually in.  I had to unfold a few times.

matt-gardner: This probably means changing all of the places we call `Model.load` to a `load_model` method, or similar, but that's probably preferable to trying to make `Model.load` also handle ensembles.

schmmd: I'm not quite following what you're thinking of.  Whether the syntax is `Model.load` or `load_model` doesn't make much difference.

The challenge is that `Model.load` makes some decisions (such as loading a Vocabulary) before it knows which class it's loading (which is specified in a parameter file).  I could peak at the model type in the parameters file and make decisions based on that.

matt-gardner: Yes, that's what I'm suggesting.  `load_model` would do some kind of check to figure out if it should call `Model.load()` or `Ensemble.load()`.

schmmd: Great--I'll try something like that but I don't see any need to rename the method from `Model.load()` since an `Ensemble` is a `Model`.

matt-gardner: What I meant:

```python
def load_archive(...):
    # We currently have something like this
    Model.load(...)
    # Instead we'd have something like this:
    load_model(...)  # note that this is not a method on `Model`

def load_model(...):
    if some_check_on_the_parameters():
        return Ensemble.load()
    else:
        return Model.load()
```

If you just modify `Model.load()`, you're going to have to break out of that somehow and call to `Ensemble.load()`, which seems pretty ugly.

schmmd: Branching out from `load_model()` to `Model.load()` (which should load any model, but in this case only loads models which are not ensembles) and `Ensemble.load()` also seems pretty ugly.

Adding another class to the hierarchy `SingleModel` (or some other name) is the only way I can see out of this.  Do you think such a class would be a good addition?

matt-gardner: Another option:

```python

class Model:
    def load(cls, ...):
        model_type = params.pop("type")
        return cls.by_name(model_type)._load(...)

    def _load(cls, ...):
        # The old `load()` logic here.
```

This second option is probably better, and more closely mimics what we already do with `from_params`.

schmmd: Python never ceases to amaze me.

allennlp/models/ensemble/ensemble.py
```diff
@@ -15,7 +16,29 @@ def __init__(self,
...
+        Ensembles don't have vocabularies or weights, so they override _load.
+        """
+        model_params = config.get('model')
+
+        # The experiment config tells us how to _train_ a model, including where to get pre-trained
+        # embeddings from.  We're now _loading_ the model, so those embeddings will already be
+        # stored in our weights.  We don't need any pretrained weight file anymore, and we don't
+        # want the code to look for it, so we remove it from the parameters here.
+        # Model._remove_pretrained_embedding_params(model_params) TODO(michaels)
+        model = Model.from_params(None, model_params)
```
matt-gardner: This basic structure looks much better!  I'm guessing that you can just iterate over some `submodels` key in the config, and call `Model.load()` or `load_archive()`, or something like that.

schmmd: I don't understand your point.  This is exactly what happens in [https://github.com/allenai/allennlp/pull/1000/files#diff-237752509d4dbe981cfb0c60400a0f8dR105](BidafEnsemble.from_params).

matt-gardner: Ah, sorry, I didn't realize you were doing that logic there.  Yeah, that's probably ok.

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,109 @@
...
+    def forward(self,  # type: ignore
+                question: Dict[str, torch.LongTensor],
+                passage: Dict[str, torch.LongTensor],
+                span_start: torch.IntTensor = None,
+                span_end: torch.IntTensor = None,
+                metadata: List[Dict[str, Any]] = None) -> Dict[str, torch.Tensor]:
+
+        subresults = []
+        for i, submodel in enumerate(self.submodels):
+            subresults.append(submodel.forward(question, passage, span_start, span_end, metadata))
```
joelgrus: I would just use a list comprehension here (and not use `enumerate`, since you're not using the index, it looks like)

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,109 @@
...
+                if votes == max_vote:
+                    average_confidence = 0
+                    for i in indices:
+                        subresult = subresults[i]
+                        average = (subresult["span_start_probs"].data[batch][start] + subresult["span_end_probs"].data[batch][end]) / 2.0
+                        if average > average_confidence:
+                            average_confidence = average
+                    if average_confidence > max_average_confidence:
+                        max_average_confidence = average_confidence
+                        best = i
```
joelgrus: this whole section feels way too complex at a high level you should be able to

* collect votes
* sum up confidences
* sort by a combination of the two

my gut tells me it should only take a few lines of code

schmmd: I likely can.  This is not ready for PR or merging.

joelgrus: fair enough 😀

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,163 @@
...
+        best_span_str : List[str]
+            If sufficient metadata was provided for the instances in the batch, we also return the
+            string from the original passage that the model thinks is the best answer to the
+            question.
+        """
+
+        subresults = []
+        for submodel in self.submodels:
+            # TODO(michaels): should I not call forward directly here and instead call submodel(**)?
+            subresults.append(submodel.forward(question, passage, span_start, span_end, metadata))
```
schmmd: Should I not call forward directly here and instead call submodel(**)

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,163 @@
...
+
+        batch_size = len(subresults[0]["best_span"])
+
+        # TODO(michaels): The response here is a tensor rather than a variable (as in BiDAF).  I think that is
+        # ok because this isn't used for training, but I wanted to ask.
+        output = {
```
schmmd: The response here is a tensor rather than a variable (as in BiDAF).  I think that is ok because this isn't used for training, but I wanted to ask.

allennlp/models/ensemble/bidaf_ensemble.py
matt-gardner: Yes, that's correct.  Sorry I missed this the last time I commented on this line.

allennlp/models/ensemble/bidaf_ensemble.py
matt-gardner: Yeah, probably fine, and easy to fix if we decide we need to fix it later.  Pytorch is changing how tensors / variables work soon, anyway.

schmmd:
Apologies for the message git history.  I used to rebase and clean these up, but in the end GitHub can make things too confusing.  With "Squash and merge" it matters far less than it used to.

This is ready for review, but I still don't like that Ensembles are Models which are not trainable (and for which it doesn't make sense to train them).  However, this is the basic abstraction that we rely on for prediction and evaluation so, as the library is currently designed, I think this is how it needs to be.

Potentially we should add a clean exception if someone tries to train an ensemble, but I'm not sure the best way to do that.

joelgrus:
you could probably add a backward_hook that throws if someone tries to call backward on it:

http://pytorch.org/tutorials/beginner/former_torchies/nn_tutorial.html#forward-and-backward-function-hooks

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,158 @@
...
+    the average confidence of the start and stop indices.
+    """
+
+    def __init__(self,
+                 submodels: List[BidirectionalAttentionFlow]) -> None:
+        super(BidafEnsemble, self).__init__(submodels)
+
+        # Using ModuleList propagates calls to .eval() so dropout is disabled on the submodels in evaluation
+        # and prediction.
+        self.submodels = torch.nn.ModuleList(submodels)
```
joelgrus: I think this should just be in the base `Ensemble` constructor (you are already doing `self.submodels = submodels`) there

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,158 @@
...
+            and each offset is a token index.
+        best_span_str : List[str]
+            If sufficient metadata was provided for the instances in the batch, we also return the
+            string from the original passage that the model thinks is the best answer to the
+            question.
+        """
+
+        subresults = []
+        for submodel in self.submodels:
+            subresults.append(submodel(question, passage, span_start, span_end, metadata))
```
joelgrus: it would be more idiomatic to write this as a list comprehension

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,158 @@
...
+
+        batch_size = len(subresults[0]["best_span"])
+
+        output = {
+                "best_span": torch.zeros(batch_size, 2).long()
+        }
+        for index in range(batch_size):
+            best = ensemble(index, subresults)
```
joelgrus: I would call this variable `best_index` or something to make clearer what it is

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,158 @@
...
+            best_span = subresults[best]["best_span"].data[index].long()
+            output["best_span"][index] = best_span
+
+            if metadata is not None:
+                if "best_span_str" not in output:
+                    output["best_span_str"] = []
```
joelgrus: why not just do this when you're defining `output` above? because you'd rather return no key than an empty list?

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,158 @@
...
+        assert not vocab, "vocab should be None"
+
+        submodels = []
+        paths = params.pop("submodels")
+        for path in paths:
+            submodels.append(load_archive(path).model)
+
+        return cls(submodels=submodels)
+
+def ensemble(index, subresults: List[Dict[str, torch.Tensor]]) -> int:
```
joelgrus: type hint on `index`?

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,158 @@
...
+    Returns
+    -------
+    The index of the best submodel.
+    """
+
+    # Populate span_votes so each key represents a span range that a submodel predicts and the value
+    # is the number of models that made the prediction.
+    spans = [(subresult["best_span"].data[index][0], subresult["best_span"].data[index][1])
+             for subresult in subresults]
+    votes: Dict[Tuple[int, int], int] = {span:spans.count(span) for span in spans}
```
joelgrus: you should learn `collections.Counter`

```
votes = collections.Counter(spans)
```

allennlp/models/ensemble/ensemble.py
```diff
@@ -0,0 +1,57 @@
...
+    (instead they rely on the vocabulary and weights from submodels) and they are not directly
+    trainable.  Instead, the submodels are trained independently and the ensemble is created
+    from the result.
+    """
+
+    def __init__(self,
+                 submodels: List[Model]) -> None:
+        vocab = submodels[0].vocab
+        for submodel in submodels:
+            assert submodel.vocab == vocab, "Vocabularies in ensemble differ"
```
joelgrus: I would probably raise a `ConfigurationError` rather than use an assert

allennlp/models/ensemble/ensemble.py
```diff
@@ -0,0 +1,57 @@
...
+
+        super(Ensemble, self).__init__(vocab, None)
+        self.submodels = submodels
```
joelgrus: as per above, use `ModuleList` here

tests/models/ensemble/bidaf_ensemble_test.py
```diff
@@ -0,0 +1,105 @@
...
+                        "span_end_probs": torch.autograd.Variable(torch.FloatTensor([[0.9158865809440613, 4.180105679552071e-05, 2.762122494459618e-05, 0.0840439721941948]])),
+                        "best_span": torch.autograd.Variable(torch.LongTensor([[0, 0]])),
+                        "best_span_str": "What",
+                        "question_tokens": ["What", "did", "Michael", "eat", "?"], "passage_tokens": ["Michael", "ate", "cheese", "."]
+                }
+        ]
+
+        assert ensemble(0, subresults) == 0
+
+    def test_ensemble_chooses_highest_average_confidence_reverse(self):
```
joelgrus: what's this test? just the same as the above with the examples in the reverse order?

allennlp/models/ensemble/bidaf_ensemble.py
schmmd: Good point--in particular it shouldn't be defined in both places.

allennlp/models/ensemble/bidaf_ensemble.py
schmmd: done

allennlp/models/ensemble/bidaf_ensemble.py
schmmd: done

allennlp/models/ensemble/bidaf_ensemble.py
schmmd: no good reason--it's left over from a refactor.  done

allennlp/models/ensemble/bidaf_ensemble.py
schmmd: done

allennlp/models/ensemble/bidaf_ensemble.py
schmmd: I'll read up on it.

allennlp/models/ensemble/ensemble.py
schmmd: done

allennlp/models/ensemble/ensemble.py
schmmd: done

tests/models/ensemble/bidaf_ensemble_test.py
schmmd: yes--I could delete the proceeding one.  I didn't want every test to be satisfiable with the answer 0.

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,155 @@
+from typing import Dict, List, Any, Tuple
```
matt-gardner: I'm trying to think of what the right way is to structure this code.  I don't think it makes much sense to have an "ensembles" submodule, because we've already started organizing models around task, and this adds an orthogonal organization dimension.  Maybe put the main `Ensemble` class in `allennlp/models/ensemble.py`, and just add concrete ensembles next to the model they're ensembling (`models/reading_comprehension/bidaf_ensemble.py`)?  Then there's no need for an `allennlp/models/ensemble/` directory.

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,155 @@
...
+from allennlp.data import Vocabulary
+from allennlp.training.metrics import SquadEmAndF1
+
+@Model.register("bidaf-ensemble")
+class BidafEnsemble(Ensemble):
+    """
+    This class ensembles the output from multiple BiDAF models.
+
+    It combines results from the submodels by taking the option with the most votes and breaking ties with
+    the average confidence of the start and stop indices.
```
matt-gardner: What happens if you just use average confidence of start and stop indices?  This is the typical thing to do, and given that you already have the logic, I'd recommend just removing the voting and using the confidence averaging as the main (only) code path.

This would also make it quite trivial to include a loss, and make this whole thing trainable, if we ever wanted to do that.

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+
+    def __init__(self,
```
matt-gardner: Argument on the same line?

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,155 @@
...
+                 submodels: List[BidirectionalAttentionFlow]) -> None:
+        super(BidafEnsemble, self).__init__(submodels)
```
matt-gardner: You can just do `super().__init__(submodels)` here.  Including the `BidafEnsemble, self` arguments isn't necessary, and just adds clutter (I didn't know this until recently, so many of our classes have it, but we should remove them and not add more).

allennlp/models/ensemble/bidaf_ensemble.py
```diff
@@ -0,0 +1,155 @@
...
+    # is the number of models that made the prediction.
+    spans = [(subresult["best_span"].data[index][0], subresult["best_span"].data[index][1])
+             for subresult in subresults]
+    votes: Dict[Tuple[int, int], int] = {span:spans.count(span) for span in spans}
+
+    # Choose the majority-vote span.
+    # If there is a tie, break it with the average confidence (span_start_probs + span_end_probs).
+    options = []
+    for i, subresult in enumerate(subresults):
+        start = subresult["best_span"].data[index][0]
```
matt-gardner: You _should_ be able to just do something like:

```python
span_start_probs = torch.sum(subresult['span_start_probs'] for subresult in subresults) / len(subresults)
span_end_probs = torch.sum(subresult['span_end_probs'] for subresult in subresults) / len(subresults)
best_span = BidirectionalAttentionFlow._get_best_span(span_start_probs.log(), span_end_probs.log())
```
Doing it this way is differentiable, which allows training (either the submodels themselves, or adding some weights to the sum above that get learned).

allennlp/models/ensemble/ensemble.py
```diff
@@ -0,0 +1,62 @@
...
+    An ensemble runs multiple instances of a model and selects an answer from the subresults via some
+    ensembling strategy.
+
+    Ensembles differ from most models in that they do not have a vocabulary or weights of their own
+    (instead they rely on the vocabulary and weights from submodels) and they are not directly
+    trainable.  Instead, the submodels are trained independently and the ensemble is created
+    from the result.
+    """
+
+    def __init__(self,
```
matt-gardner: I'd also just keep this on one line.

allennlp/models/ensemble/ensemble.py
```diff
@@ -0,0 +1,62 @@
...
+from allennlp.models.model import Model
+from allennlp.models.model import remove_pretrained_embedding_params
+
+class Ensemble(Model):
+    """
+    An ensemble runs multiple instances of a model and selects an answer from the subresults via some
+    ensembling strategy.
+
+    Ensembles differ from most models in that they do not have a vocabulary or weights of their own
+    (instead they rely on the vocabulary and weights from submodels) and they are not directly
```
matt-gardner: There's no reason they shouldn't be directly trainable.  I don't think we need to include this.  You _could_ say something like "they are not typically trained jointly (but you can if you want to)."

allennlp/models/ensemble/ensemble.py
```diff
@@ -0,0 +1,62 @@
...
+            if submodel.vocab != vocab:
+                raise ConfigurationError("Vocabularies in ensemble differ")
+
+        super(Ensemble, self).__init__(vocab, None)
+
+        # Using ModuleList propagates calls to .eval() so dropout is disabled on the submodels in evaluation
+        # and prediction.
+        self.submodels = torch.nn.ModuleList(submodels)
+
+    def forward(self, *inputs) -> Dict[str, torch.Tensor]:  # pylint: disable=arguments-differ
```
matt-gardner: I don't think having this here adds anything.  You're just overriding an abstract method with another abstract method, and not changing the API.

tests/models/ensemble/bidaf_ensemble_test.py
```diff
@@ -0,0 +1,105 @@
+# pylint: disable=no-self-use,invalid-name,line-too-long
```
matt-gardner: I don't think you need to disable `line-too-long` here.  The lines below could be shortened dramatically, and it would help the readability of this test.

tests/models/ensemble/bidaf_ensemble_test.py
```diff
@@ -0,0 +1,105 @@
...
+class BidafEnsembleTest(ModelTestCase):
+    def setUp(self):
+        super(BidafEnsembleTest, self).setUp()
+        self.set_up_model('tests/fixtures/bidaf/experiment.json', 'tests/fixtures/data/squad.json')
+        self.model.eval()
+
+    def test_ensemble_chooses_most_votes(self):
+        subresults = [
+                {
+                        "span_start_probs": torch.autograd.Variable(torch.FloatTensor([[0.0009347021696157753, 0.0012768324231728911, 0.9999938261222839, 1.4630706573370844e-05]])),
```
matt-gardner: You could do `from torch.autograd import Variable` above to save some horizontal space on all of these lines.  Also, you could use some toy probabilities for all of these (like `0.1`, `0.8`, etc.), so that this doesn't get so long.  And so that it's easier for a person looking at this test to do the arithmetic in their head and make sure they agree the test is correct.

tests/models/ensemble/bidaf_ensemble_test.py
```diff
@@ -0,0 +1,105 @@
...
+                        "span_end_probs": torch.autograd.Variable(torch.FloatTensor([[4.180105679552071e-05, 2.762122494459618e-05, 0.9999965809440613, 0.0840439721941948]])),
+                        "best_span": torch.autograd.Variable(torch.LongTensor([[2, 2]])),
+                        "best_span_str": "cheese",
+                        "question_tokens": ["What", "did", "Michael", "eat", "?"], "passage_tokens": ["Michael", "ate", "cheese", "."]
```
matt-gardner: "passage_tokens" should be on its own line, like everything else.

tests/models/ensemble/bidaf_ensemble_test.py
matt-gardner: Yeah, just having one of these two is probably enough here.

allennlp/models/ensemble/bidaf_ensemble.py
schmmd: Works for me.

allennlp/models/ensemble/bidaf_ensemble.py
schmmd: I can try this.  I did what Chris Clarke ended up doing after trying different approaches.

allennlp/models/ensemble/ensemble.py
matt-gardner: If you're instantiating a `Trainer` yourself, yes, you definitely could train this, as long as your ensemble included a `loss` key in the output.  And if you implement an `Ensemble` subclass yourself that can handle instantiating things `from_params` correctly, you could train this with `allennlp train` just fine.

allennlp/models/ensemble/ensemble.py
matt-gardner: Yeah, this class is still abstract, and pylint doesn't know that.  So just disable the `abstract-method` check underneath the class docstring.

allennlp/models/ensemble/bidaf_ensemble.py
DeNeutoy: It doesn't matter much here (because it's unlikely that you'll have more than ~10 sub-models), but counting the spans in this way is n^2 wrt the number of spans. Feel free to ignore. Also, the comment referring to `span_votes` is outdated.

schmmd:
The results are comparable when just using the average confidence.

```
2018-03-27 16:02:07,957 - INFO - allennlp.commands.evaluate - em: 0.7228003784295175
2018-03-27 16:02:07,957 - INFO - allennlp.commands.evaluate - f1: 0.808343701550247
```

However, this does mean that the predicted interval might not have been predicted by any of the submodels.

schmmd:
@matt-gardner I've addressed the items you brought up.

allennlp/models/reading_comprehension/bidaf.py
```diff
@@ -77,6 +77,7 @@ def __init__(self, vocab: Vocabulary,
                  regularizer: Optional[RegularizerApplicator] = None) -> None:
         super(BidirectionalAttentionFlow, self).__init__(vocab, regularizer)
 
+
```
matt-gardner: No need to add this extra line.

allennlp/models/reading_comprehension/bidaf_ensemble.py
```diff
@@ -0,0 +1,143 @@
...
+    def forward(self,  # type: ignore
+                question: Dict[str, torch.LongTensor],
+                passage: Dict[str, torch.LongTensor],
+                span_start: torch.IntTensor = None,
+                span_end: torch.IntTensor = None,
+                metadata: List[Dict[str, Any]] = None) -> Dict[str, torch.Tensor]:
+        # pylint: disable=arguments-differ
+        """
+        The forward method runs each of the submodels, then selects the best span from the subresults.
+        The best span is the span which most of the submodels predict.  If there is a tie, it is broken
```
matt-gardner: This docstring isn't true anymore.

allennlp/models/reading_comprehension/bidaf_ensemble.py
```diff
@@ -0,0 +1,143 @@
...
+from allennlp.common import Params
+from allennlp.data import Vocabulary
+from allennlp.training.metrics import SquadEmAndF1
+
+@Model.register("bidaf-ensemble")
+class BidafEnsemble(Ensemble):
+    """
+    This class ensembles the output from multiple BiDAF models.
+
+    It combines results from the submodels by taking the option with the most votes and breaking ties with
```
matt-gardner: This docstring isn't true anymore.

PR #1005, opened 2018-03-20
add optional exclusive start indices flag to Enpoint Extractor

DeNeutoy:


allennlp/modules/span_extractors/endpoint_span_extractor.py
```diff
@@ -86,15 +101,40 @@ def forward(self,
...
+        else:
+            # We want `exclusive` span starts, so we remove 1 from the forward span starts
+            # as the AllenNLP ``SpanField`` is inclusive.
+            # shape (batch_size, num_spans)
+            exclusive_span_starts = span_starts - 1
+            # shape (batch_size, num_spans, 1)
+            start_sentinel_mask = (exclusive_span_starts == -1).long().unsqueeze(-1)
+            exclusive_span_starts = exclusive_span_starts * (1 - start_sentinel_mask.squeeze(-1))
+
+            # We'll check the indices here at runtime, because it's difficult to debug
```
matt-gardner: When will this happen?  Seems like the right place to handle this is in tests?  Or replace the `== -1` above with `< 0`?  That is, checks that are aimed at bugs belong in tests; checks that are aimed at invalid inputs on user-facing functions belong as asserts in the code.  Seems like this is aiming at a bug, not at potentially invalid inputs.

tests/modules/span_extractors/endpoint_span_extractor_test.py
```diff
@@ -17,7 +17,7 @@ def test_endpoint_span_extractor_can_build_from_params(self):
                 })
         extractor = SpanExtractor.from_params(params)
         assert isinstance(extractor, EndpointSpanExtractor)
-        assert extractor.get_output_dim() == 10
+        assert extractor.get_output_dim() == 17 # 2 * input_dim + span_width_embedding_dim
```
matt-gardner: Two spaces before `#` - makes it easier to see it's a comment, particularly with this github highlighting.

tests/modules/span_extractors/endpoint_span_extractor_test.py
```diff
@@ -73,3 +73,48 @@ def test_masked_indices_are_handled_correctly(self):
                                          correct_start_embeddings.numpy())
         numpy.testing.assert_array_equal(end_embeddings.data.numpy(),
                                          correct_end_embeddings.numpy())
+
+
+    def test_masked_indices_are_handled_correctly_with_exclusive_indices(self):
+        sequence_tensor = Variable(torch.randn([2, 5, 8]))
+        # concatentate start and end points together to form our representation
```
matt-gardner: Formatting.

tests/modules/span_extractors/endpoint_span_extractor_test.py
```diff
@@ -73,3 +73,48 @@ def test_masked_indices_are_handled_correctly(self):
...
+        # for both the forward and backward directions.
+        extractor = EndpointSpanExtractor(8, "x,y", use_exclusive_start_indices=True)
+        indices = Variable(torch.LongTensor([[[1, 3],
+                                              [2, 4]],
+                                             # This span has an end index at the
```
matt-gardner: It's the start index that matters, right?

allennlp/modules/span_extractors/endpoint_span_extractor.py
DeNeutoy: I've always been wary of asserts over raising `RuntimeError` or `ValueError`s, because you can pass `-O or -OO` flags to python which cause asserts to not be checked. Why you would want this is beyond me, but ever since reading that i've always used exceptions.

matt-gardner: Sorry, I didn't mean to say this should be an assert, and when I said "assert", I was including exceptions like this.  I was suggesting that this belongs in a test, not as a runtime check.

matt-gardner: To be more clear: I don't see how you can get into this state.  It looks like this is checking for a bug in the logic above.  Bug checks belong in tests.

DeNeutoy: It's purpose is that if you did accidentally pass incorrect indices to this function, it wouldn't necessarily crash (because the indices get shifted by the batch dimension when we flatten to do the `index_select`, so if they are negative they won't necessarily stay negative) and it would be really, really hard to find.

matt-gardner: I see, ok.  In that case, would it make more sense to put this at the top of the function, and check it on `span_starts`?

DeNeutoy: no, because you can pass a mask to this function, and we need to mask out the indices which correspond to using the start sentinel (which have an exclusive index of -1) before we check.

matt-gardner: But those are the same indices where you have a start index of 0.  That's the part where it looks like you're checking for a bug in your code, not for invalid input.

DeNeutoy: Ok - I see what you mean - but I think that makes the semantics of what I am checking much more confusing. It's not directly a case of checking user input, because I have to mask it first to do the check, so to me it seems clearer to have the check closer to where it would break. It's also the same as in the BidirectionalEndpointSpanExtractor.

PR #1010, opened 2018-03-21
better error for gradient check

DeNeutoy:
The assert gave a really rubbish error in the case that the gradient was `None`, so this will improve that.

### DIFF
--- a/allennlp/common/testing/model_test_case.py
+++ b/allennlp/common/testing/model_test_case.py
@@ -133,15 +133,24 @@ class ModelTestCase(AllenNlpTestCase):
         model.zero_grad()
         result = model(**model_batch)
         result["loss"].backward()
-
-        for parameter in model.parameters():
+        has_zero_or_none_grads = {}
+        for name, parameter in model.named_parameters():
             zeros = torch.zeros(parameter.size())
             if parameter.requires_grad:
+
+                if parameter.grad is None:
+                    has_zero_or_none_grads[name] = "No gradient computed (i.e parameter.grad is None)"
                 # Some parameters will only be partially updated,
                 # like embeddings, so we just check that any gradient is non-zero.
-                assert (parameter.grad.data.cpu() != zeros).any()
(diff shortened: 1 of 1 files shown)


allennlp/common/testing/model_test_case.py
```diff
@@ -133,16 +133,25 @@ def check_model_computes_gradients_correctly(model, model_batch):
...
+        for name, parameter in model.named_parameters():
             zeros = torch.zeros(parameter.size())
             if parameter.requires_grad:
+
+                if parameter.grad is None:
+                    has_zero_or_none_grads[name] = "No gradient computed (i.e parameter.grad is None)"
                 # Some parameters will only be partially updated,
                 # like embeddings, so we just check that any gradient is non-zero.
-                assert (parameter.grad.data.cpu() != zeros).any()
+                if not (parameter.grad.data.cpu() != zeros).any():
```
joelgrus: is there a way to rewrite this? `not (grad != zeros).any()` is really hard to understand

joelgrus: like, would

```
if (parameter.grad.data.cpu() == 0).all()
```

work?

allennlp/common/testing/model_test_case.py
```diff
@@ -133,16 +133,25 @@ def check_model_computes_gradients_correctly(model, model_batch):
...
-                assert (parameter.grad.data.cpu() != zeros).any()
+                if (parameter.grad.data.cpu() == zeros).all():
+                    has_zero_or_none_grads[name] = f"zeros with shape ({tuple(parameter.grad.size())})"
             else:
                 assert parameter.grad is None
 
+        if has_zero_or_none_grads:
+            for name, grad in has_zero_or_none_grads.items():
+                print(f"Parameter: {name} had incorrect gradient: {grad}")
+            raise Exception("Incorrect gradients found. See stack trace for info.")
```
matt-gardner: The info's in stdout, not the stack trace, right?

PR #1016, opened 2018-03-22
Fix action embedding

matt-gardner:
Depends on #991.  This removes the LHS/RHS action embedding that I for some reason thought was a good idea when I first wrote this, and replaces it with a single embedding for each global action.  This lets us simplify the code quite a bit, and I've only done some of the simplification so far.  It gives a ~20% speed up to the code, but unfortunately didn't fix our performance problems on its own.  This gave me maybe a half a percent increase in performance on wikitables.

I'm quite certain I broke the NLVR parser when I did this; I changed the way the `ProductionRuleField` works, and I updated the wikitables parser to match, but not the NLVR parser.  I need to fix that one too before this is ready to merge.


matt-gardner:
Ok @pdasigi, this should be ready to review.  I'll resolve the conflicts and fix any CI problems, but the tests pass locally, so there shouldn't be much changing from here.

allennlp/models/encoder_decoders/nlvr/nlvr_semantic_parser.py
```diff
@@ -73,43 +68,45 @@ class NlvrSemanticParser(Model):
...
-        self._terminal_embedder = terminal_embedder
         self._encoder = encoder
         self._decoder_trainer: DecoderTrainer[Callable[[NlvrDecoderState], torch.Tensor]] = \
                 ExpectedRiskMinimization(beam_size, normalize_beam_score_by_length, max_decoding_steps)
-        action_embedding_dim = nonterminal_embedder.get_output_dim() * 2
+        self._rule_namespace = rule_namespace
+
+        self._action_embedder = Embedding(num_embeddings=vocab.get_vocab_size(self._rule_namespace),
+                                          embedding_dim=action_embedding_dim)
+        self._initial_action_embedding = torch.nn.Parameter(torch.FloatTensor(action_embedding_dim))
```
pdasigi: A comment about this field would be nice here. Also, I think `start_action_embedding` may be a better name, since this sounds like this is the initial state of the action embedding weights.

allennlp/models/encoder_decoders/nlvr/nlvr_semantic_parser.py
```diff
@@ -73,43 +68,45 @@ class NlvrSemanticParser(Model):
...
-                               "terminal embedding dim")
-        self._nonterminal_embedder = nonterminal_embedder
-        self._terminal_embedder = terminal_embedder
         self._encoder = encoder
         self._decoder_trainer: DecoderTrainer[Callable[[NlvrDecoderState], torch.Tensor]] = \
                 ExpectedRiskMinimization(beam_size, normalize_beam_score_by_length, max_decoding_steps)
-        action_embedding_dim = nonterminal_embedder.get_output_dim() * 2
+        self._rule_namespace = rule_namespace
+
+        self._action_embedder = Embedding(num_embeddings=vocab.get_vocab_size(self._rule_namespace),
```
pdasigi: Do you ever use this as a module? Looks like you're only directly accessing its weights. If that's true, why not define  just the parameter directly here?

allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
```diff
@@ -95,32 +84,29 @@ def __init__(self,
...
 
         check_dimensions_match(entity_encoder.get_output_dim(), question_embedder.get_output_dim(),
                                "entity word average embedding dim", "question embedding dim")
-        if terminal_embedder:
-            check_dimensions_match(nonterminal_embedder.get_output_dim(), terminal_embedder.get_output_dim(),
-                                   "nonterminal embedding dim", "terminal embedding dim")
 
         self._action_padding_index = -1  # the padding value used by IndexField
-        action_embedding_dim = nonterminal_embedder.get_output_dim() * 2
+        self._action_embedder = Embedding(num_embeddings=vocab.get_vocab_size(self._rule_namespace),
```
pdasigi: Same comment as above for this module.

allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
```diff
@@ -95,32 +84,29 @@ def __init__(self,
...
+                                          embedding_dim=action_embedding_dim)
+        self._initial_action_embedding = torch.nn.Parameter(torch.FloatTensor(action_embedding_dim))
```
pdasigi: Again, `start_action_embedding`?

allennlp/models/encoder_decoders/nlvr/nlvr_semantic_parser.py
```diff
@@ -362,137 +352,30 @@ def _embed_actions(self, actions: List[List[ProductionRuleArray]]) -> Tuple[torc
             Has shape ``(num_unique_actions, action_embedding_dim)``.
         action_map : ``Dict[Tuple[int, int], int]``
             Maps ``(batch_index, action_index)`` in the input action list to ``action_index`` in
-            the ``action_embeddings`` tensor.
-        initial_action_embedding : ``torch.Tensor``
-            Has shape ``(action_embedding_dim,)``.  An embedding for the initial action input.
-            This needs to be computed here, because we don't keep around the nonterminal
-            embeddings.  We do this by creating a fake rule "0 -> START", where the LHS embedding
-            is a vector of zeros, and the RHS embedding is the START symbol embedding.
+            the ``action_embeddings`` tensor.  All non-embeddable actions get mapped to `-1` here.
```
pdasigi: That's just padding, right?

tests/fixtures/encoder_decoder/wikitables_semantic_parser/serialization/vocabulary/rule_labels.txt
```diff
@@ -1,122 +1,104 @@
...
+p -> x
+r -> [<#1,#1>, r]
+r -> [<#1,<#1,#1>>, r, r]
+r -> [<c,r>, c]
+r -> [<n,<n,<#1,<<#2,#1>,#1>>>>, n, n, r, <d,r>]
+r -> [<n,<n,<#1,<<#2,#1>,#1>>>>, n, n, r, <n,r>]
+r -> [<n,r>, n]
+r -> [<r,r>, r]
+r -> fb:type.row
+r -> x
```
pdasigi: So the number of rule labels _went down_? Interesting!

matt-gardner: Yes, because previously there were two vocabulary items per rule, one of the LHS, and one for the RHS.  There were some duplicates, so this doesn't halve the number of rules, but it does decrease it.

allennlp/models/encoder_decoders/nlvr/nlvr_semantic_parser.py
matt-gardner: `first_action_embedding`?  Just so there's no confusion with `START -> c`, `START -> d`, etc.

allennlp/models/encoder_decoders/nlvr/nlvr_semantic_parser.py
matt-gardner: I'm thinking we'll probably simplify a lot of the action logic, and at that point we'll likely use this as a module, so I think keeping like this is sensible.

allennlp/models/encoder_decoders/nlvr/nlvr_semantic_parser.py
matt-gardner: Not for NLVR.  It shouldn't ever happen at all - padded actions are skipped.  I changed this.

PR #1022, opened 2018-03-23
Break up AWS login command in two to catch failures.

schmmd:


schmmd:
It works, but I'm not 100% it will catch your error.  It did catch the error I introduced...

The problem was that `eval` was returning a status code of 0 even when the subcommand failed.

PR #1030, opened 2018-03-24
Add length-based filtering and truncation of seq2seq data

### LINKED ISSUE #977: [Feature] seq2seq dataset reader should support maximum sequence length
I think in practical, the source/target sequence to seq2seq model should be truncated to the maximum sequence length to avoid the OOM problem, but the current seq2seq dataset reader seems not to support this configuration.

nelson-liu:
Fixes #977 

This PR lets you discard or truncate examples with source / target sequences above a certain length. If the input value is 0, no truncation or filtering is applied.

I'm not sure how I feel about these parameter names, I mostly took them from OpenNMT. If you can think of a better name, I'm happy to change it.

### DIFF
--- a/allennlp/data/dataset_readers/seq2seq.py
+++ b/allennlp/data/dataset_readers/seq2seq.py
@@ -46,6 +46,22 @@ class Seq2SeqDatasetReader(DatasetReader):
         ``source_token_indexers``.
     source_add_start_token : bool, (optional, default=True)
         Whether or not to add `START_SYMBOL` to the beginning of the source sequence.
+    source_max_sequence_length : int, (optional, default=0)
+        Maximum source sequence length (excluding `START_SYMBOL` and `END_SYMBOL`).
+        Examples with source sequence length exceeding this value are discarded.
+        0 indicates length is unlimited. Value must be greater than or equal to 0.
+    source_truncate_sequence_length : int, (optional, default=0)
+        Source sequences longer than this value (excluding `START_SYMBOL` and
+        `END_SYMBOL`) will be truncated to this value (cutting starting from the end).
+        0 indicates length is unlimited. Value must be greater than or equal to 0.
+    target_max_sequence_length : int, (optional, default=0)
+        Maximum target sequence length (excluding `START_SYMBOL` and `END_SYMBOL`).
+        Examples with target sequence length exceeding this value are discarded.
+        0 indicates length is unlimited. Value must be greater than or equal to 0.
+    target_truncate_sequence_length : int, (optional, default=0)
+        Target sequences longer than this value (excluding `START_SYMBOL` and
+        `END_SYMBOL`) will be truncated to this value (cutting starting from the end).
+        0 indicates length is unlimited. Value must be greater than or equal to 0.
     """
     def __init__(self,
                  source_tokenizer: Tokenizer = None,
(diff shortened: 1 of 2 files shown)


allennlp/data/dataset_readers/seq2seq.py
```diff
@@ -75,18 +119,31 @@ def _read(self, file_path):
...
+                instance = self.text_to_instance(source_sequence, target_sequence)
+                if instance is None:
+                    continue
+                yield instance
 
     @overrides
     def text_to_instance(self, source_string: str, target_string: str = None) -> Instance:  # type: ignore
         # pylint: disable=arguments-differ
         tokenized_source = self._source_tokenizer.tokenize(source_string)
+        if len(tokenized_source) > self._source_max_sequence_length and self._filter_source:
```
matt-gardner: We've been debating about whether we should put this kind of functionality directly into the `Tokenizer`, as it's a common problem to have.  What do you think?

Some issues:
- Wanting to discard instances completely if they exceed some max length seems like it's functionality that belongs in the `DatasetReader`, not the `Tokenizer` (the tokenizer could return `None` here in those cases, but that seems awkward).  And if you have one `Tokenizer` that operates on both questions and passages in SQuAD, you might want different max lengths for each of those, so you'd want the `DatasetReader` to pass these lengths to `tokenize()`, not have them be parameter that get passed to the `Tokenizer's` constructor.
- On the other hand, `Tokenizers` can add start and end tokens to their strings (which are separate from the `START_SYMBOL` and `END_SYMBOL` here - these are required for decoder functionality, but you might want additional buffer tokens if you're using a CNN to encode things).  So the truncation that you're doing outside of the `Tokenizer` messes with any start and end tokens that the tokenizer may have added.

Any ideas?

nelson-liu: I'm not as well-versed in the API as you are, so take this with a grain of salt. i do think that it feels more natural for truncation and filtering to be part of the `DatasetReader` --- the `Tokenizer` should only handle tokenizing and little else. Are there currently examples where point #2 shows up, or were you just saying that it'd be a problem for the future if we want to implement a convolutional encoder? Doesn't seem like there's an easy solution either way...

matt-gardner: Yes, there are, when you use a `CharacterTokenizer`, because these are typically encoded with CNNs (that's where this first came up, with @joelgrus wondering where to put truncation for characters).  With things moving to transformer-style architectures, we might want to have more of these start and end tokens on sentences, also.

murphp15:
I see that this has been left outstanding for quite a while, is somebody going to pick this back up? 
If ppl are too busy to do this at the moment I would be happy to pick it up.

My thoughts on this PR: 
For this PR would it not make more sense to make a datasetReader wrapper. 

1. The wrapper would delegate the creation of instances to a dataset reader.
2. It would trim the specified text fields before returning it to the caller. 

By doing this we are clearly separating responsibilities. I have actually noticed through out the allennlp codebase it seems that you guys always seem to have a lot of logic in each class instead of trying to keep it separate.

matt-gardner:
This has been left outstanding because we still need to figure out where the right place to put these parameters is.  It's not an easy API decision.  It has been on my TODO list for a while, but I haven't gotten to it yet; things have been pretty crazy for me for the last couple of months.  Hold off on this one for now; after the EMNLP deadline, we'll put some more thought into what the right API decision is here.

schmmd:
@nelson-liu @matt-gardner can we close this?  It's been open all summer, and it's starting to grow some merge conflicts.

nelson-liu:
Sure, I don’t have a need for this anymore as well

PR #1032, opened 2018-03-25
Add Stanford Sentiment Treebank dataset reader

nelson-liu:
This PR adds a reader for the StanfordSentimentTreebank tokens and labels, for 5/3/2-class text classification. It also optionally lets you pick whether or not to make examples from labeled subtrees, or whether to just use the sentences.

### DIFF
--- a/allennlp/data/dataset_readers/__init__.py
+++ b/allennlp/data/dataset_readers/__init__.py
@@ -16,3 +16,5 @@ from allennlp.data.dataset_readers.semantic_role_labeling import SrlReader
 from allennlp.data.dataset_readers.seq2seq import Seq2SeqDatasetReader
 from allennlp.data.dataset_readers.coreference_resolution import ConllCorefReader, WinobiasReader
 from allennlp.data.dataset_readers.penn_tree_bank import PennTreeBankConstituencySpanDatasetReader
+from allennlp.data.dataset_readers.stanford_sentiment_tree_bank import (
+        StanfordSentimentTreeBankTokensDatasetReader)
--- a/allennlp/data/dataset_readers/__init__.py
+++ b/allennlp/data/dataset_readers/__init__.py
@@ -17,4 +17,4 @@ from allennlp.data.dataset_readers.seq2seq import Seq2SeqDatasetReader
 from allennlp.data.dataset_readers.coreference_resolution import ConllCorefReader, WinobiasReader
 from allennlp.data.dataset_readers.penn_tree_bank import PennTreeBankConstituencySpanDatasetReader
 from allennlp.data.dataset_readers.stanford_sentiment_tree_bank import (
-        StanfordSentimentTreeBankTokensDatasetReader)
+        StanfordSentimentTreeBankDatasetReader)
--- a/allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
+++ b/allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
@@ -22,27 +22,34 @@ class StanfordSentimentTreeBankTokensDatasetReader(DatasetReader):
     """
     Reads tokens and their sentiment labels from the Stanford Sentiment Treebank.
 
+    The Stanford Sentiment Treebank comes with labels
+    from 0 to 4. ``"5-class"`` uses these labels as is. ``"3-class"`` converts the
+    problem into one of identifying whether a sentence is negative, positive, or
+    neutral sentiment. In this case, 0 and 1 are grouped as label 0 (negative sentiment),
+    2 is converted to label 1 (neutral sentiment) and 3 and 4 are grouped as label 2
+    (positive sentiment). ``"2-class"`` turns it into a binary classification problem
+    between positive and negative sentiment. 0 and 1 are grouped as the label 0
+    (negative sentiment), 2 (neutral) is discarded, and 3 and 4 are grouped as the label 1
+    (positive sentiment).
+
+    Expected format for each input line: a linearized tree, where nodes are labeled
+    by their sentiment.
+
+    The output of ``read`` is a list of ``Instance`` s with the fields:
+        tokens: ``TextField`` and
+        label: ``LabelField``
+
     Parameters
     ----------
     token_indexers : ``Dict[str, TokenIndexer]``, optional (default=``{"tokens": SingleIdTokenIndexer()}``)
         We use this to define the input representation for the text.  See :class:`TokenIndexer`.
-        Note that the `output` tags will always correspond to single token IDs based on how they
-        are pre-tokenised in the data file.
     use_subtrees : ``bool``, optional, (default = ``False``)
         Whether or not to use sentiment-tagged subtrees.
     granularity : ``str``, optional (default = ``"5-class"``)
         One of ``"5-class"``, ``"3-class"``, or ``"2-class"``, indicating the number
-        of sentiment labels to use. The Stanford Sentiment Treebank comes with labels
-        from 0 to 4. ``"5-class"`` uses these labels as is. ``"3-class"`` converts the
-        problem into one of identifying whether a sentence is negative, positive, or
-        neutral sentiment. In this case, 0 and 1 are grouped as label 0 (negative sentiment),
-        2 is converted to label 1 (neutral sentiment) and 3 and 4 are grouped as label 2
-        (positive sentiment). ``"2-class"`` turns it into a binary classification problem
-        between positive and negative sentiment. 0 and 1 are grouped as the label 0
-        (negative sentiment), 2 (neutral) is discarded, and 3 and 4 are grouped as the label 1
-        (positive sentiment).
+        of sentiment labels to use.
     lazy : ``bool``, optional, (default = ``False``)
-        Whether or not instances can be consumed lazily.
+        Whether or not instances can be read lazily.
     """
     def __init__(self,
                  token_indexers: Dict[str, TokenIndexer] = None,
--- a/allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
+++ b/allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
@@ -32,7 +32,7 @@ class StanfordSentimentTreeBankTokensDatasetReader(DatasetReader):
         Whether or not to use sentiment-tagged subtrees.
     granularity : ``str``, optional (default = ``"5-class"``)
         One of ``"5-class"``, ``"3-class"``, or ``"2-class"``, indicating the number
-        sentiment labels to use.
+        of sentiment labels to use.
     lazy : ``bool``, optional, (default = ``False``)
         Whether or not instances can be consumed lazily.
     """
--- a/allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
+++ b/allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
@@ -62,7 +62,7 @@ class StanfordSentimentTreeBankTokensDatasetReader(DatasetReader):
     def _read(self, file_path):
         with open(cached_path(file_path), "r") as data_file:
             logger.info("Reading instances from lines in file at: %s", file_path)
-            for line in tqdm.tqdm(data_file.readlines()):
+            for line in data_file.readlines():
                 line = line.strip("\n")
                 if not line:
                     continue
--- a/allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
+++ b/allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
@@ -3,7 +3,6 @@ import logging
 
 from overrides import overrides
 from nltk.tree import Tree
-import tqdm
 
 from allennlp.common import Params
 from allennlp.common.file_utils import cached_path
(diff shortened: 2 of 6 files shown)


allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
```diff
@@ -0,0 +1,125 @@
...
+                line = line.strip("\n")
+                if not line:
+                    continue
+                parsed_line = Tree.fromstring(line)
+                if self._use_subtrees:
+                    for subtree in parsed_line.subtrees():
+                        instance = self.text_to_instance(subtree.leaves(), subtree.label())
+                        if instance is None:
+                            continue
+                        yield instance
```
joelgrus: you could avoid the `continue` if you just did

```
if instance is not None:
    yield instance
```

allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
```diff
@@ -0,0 +1,125 @@
...
+                else:
+                    instance = self.text_to_instance(parsed_line.leaves(), parsed_line.label())
+                    if instance is None:
+                        continue
+                    yield instance
```
joelgrus: and same here

allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
```diff
@@ -0,0 +1,125 @@
...
+        An ``Instance`` containing the following fields:
+            tokens : ``TextField``
+                The tokens in the sentence or phrase.
+            label : ``LabelField``
+                The sentiment label of the sentence or phrase.
+        """
+        # pylint: disable=arguments-differ
+        text_field = TextField([Token(x) for x in tokens], token_indexers=self._token_indexers)
+        fields: Dict[str, Field] = {"tokens": text_field}
+        if sentiment is not None:
```
joelgrus: it would be nice if there was a brief comment about the various granularities and the logic behind the conversion here

tests/data/dataset_readers/stanford_sentiment_tree_bank_test.py
```diff
@@ -0,0 +1,95 @@
+# pylint: disable=no-self-use,invalid-name
+import pytest
+
+from allennlp.data.dataset_readers import (
+        StanfordSentimentTreeBankTokensDatasetReader)
```
joelgrus: is this line really too long without the wrapping? it doesn't look that long

allennlp/data/dataset_readers/__init__.py
```diff
@@ -16,3 +16,5 @@
 from allennlp.data.dataset_readers.seq2seq import Seq2SeqDatasetReader
 from allennlp.data.dataset_readers.coreference_resolution import ConllCorefReader, WinobiasReader
 from allennlp.data.dataset_readers.penn_tree_bank import PennTreeBankConstituencySpanDatasetReader
+from allennlp.data.dataset_readers.stanford_sentiment_tree_bank import (
```
matt-gardner: `__init__.py` files deep in the hierarchy are a case where I'd be just fine with disabling pylint's `line-too-long` check.  No strong preference if you prefer it this way, though.

allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
```diff
@@ -0,0 +1,125 @@
...
+@DatasetReader.register("sst_tokens")
+class StanfordSentimentTreeBankTokensDatasetReader(DatasetReader):
+    """
+    Reads tokens and their sentiment labels from the Stanford Sentiment Treebank.
+
+    Parameters
+    ----------
+    token_indexers : ``Dict[str, TokenIndexer]``, optional (default=``{"tokens": SingleIdTokenIndexer()}``)
+        We use this to define the input representation for the text.  See :class:`TokenIndexer`.
+        Note that the `output` tags will always correspond to single token IDs based on how they
```
matt-gardner: I'm not sure what this sentence means.  My guess is you copied it from the seq2seq reader, where it applied, and it doesn't here.

allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
```diff
@@ -0,0 +1,125 @@
...
+from allennlp.common.checks import ConfigurationError
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+@DatasetReader.register("sst_tokens")
+class StanfordSentimentTreeBankTokensDatasetReader(DatasetReader):
+    """
+    Reads tokens and their sentiment labels from the Stanford Sentiment Treebank.
+
```
matt-gardner: A note here on the input file format and the output `Fields` would be nice.

allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
```diff
@@ -0,0 +1,125 @@
...
+    Parameters
+    ----------
+    token_indexers : ``Dict[str, TokenIndexer]``, optional (default=``{"tokens": SingleIdTokenIndexer()}``)
```
matt-gardner: I don't think that these backticks inside of the default block actually render correctly.  Not a huge deal, just FYI - it's probably not worth worrying about adding them, because they don't work, anyway.  I've also taken to putting more complex default values like this in the text description, where they will actually get rendered correctly.

allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
```diff
@@ -0,0 +1,136 @@
...
+        We use this to define the input representation for the text.  See :class:`TokenIndexer`.
+        Note that the `output` tags will always correspond to single token IDs based on how they
+        are pre-tokenised in the data file.
+    use_subtrees : ``bool``, optional, (default = ``False``)
+        Whether or not to use sentiment-tagged subtrees.
+    granularity : ``str``, optional (default = ``"5-class"``)
+        One of ``"5-class"``, ``"3-class"``, or ``"2-class"``, indicating the number
+        of sentiment labels to use. The Stanford Sentiment Treebank comes with labels
```
matt-gardner: You might want to move this discussion into the part of the docstring where you talk about file formats, above, as this seems like it shouldn't be buried in a parameter description.  Up to you, though.

allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
```diff
@@ -0,0 +1,136 @@
...
+        from 0 to 4. ``"5-class"`` uses these labels as is. ``"3-class"`` converts the
+        problem into one of identifying whether a sentence is negative, positive, or
+        neutral sentiment. In this case, 0 and 1 are grouped as label 0 (negative sentiment),
+        2 is converted to label 1 (neutral sentiment) and 3 and 4 are grouped as label 2
+        (positive sentiment). ``"2-class"`` turns it into a binary classification problem
+        between positive and negative sentiment. 0 and 1 are grouped as the label 0
+        (negative sentiment), 2 (neutral) is discarded, and 3 and 4 are grouped as the label 1
+        (positive sentiment).
+    lazy : ``bool``, optional, (default = ``False``)
+        Whether or not instances can be consumed lazily.
```
matt-gardner: I think this is more about _reading_ instances lazily, not _consuming_ them lazily.

allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
```diff
@@ -0,0 +1,136 @@
...
+        if granularity not in allowed_granularities:
+            raise ConfigurationError("granularity is {}, but expected one of: {}".format(
+                    granularity, allowed_granularities))
+        self._granularity = granularity
+
+    @overrides
+    def _read(self, file_path):
+        with open(cached_path(file_path), "r") as data_file:
+            logger.info("Reading instances from lines in file at: %s", file_path)
+            for line in tqdm.tqdm(data_file.readlines()):
```
matt-gardner: Having `tqdm` in here does not play nicely with lazy datasets.  We've removed this from all of our concrete dataset readers, and put it in the base class.  Also, just FYI, we've also added a global configuration of `tqdm`, so when you _do_ want to use `tqdm`, you should do it through our `Tqdm.tqdm()` interface, instead of using `tqdm.tqdm` directly.

allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
```diff
@@ -0,0 +1,136 @@
...
+                parsed_line = Tree.fromstring(line)
+                if self._use_subtrees:
+                    for subtree in parsed_line.subtrees():
+                        instance = self.text_to_instance(subtree.leaves(), subtree.label())
+                        if instance is not None:
+                            yield instance
+                else:
+                    instance = self.text_to_instance(parsed_line.leaves(), parsed_line.label())
+                    if instance is not None:
+                            yield instance
```
matt-gardner: Indentation got off here when you addressed Joel's comment.

allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
```diff
@@ -0,0 +1,136 @@
...
+        return Instance(fields)
+
+    @classmethod
+    def from_params(cls, params: Params) -> 'StanfordSentimentTreeBankTokensDatasetReader':
+        token_indexers = TokenIndexer.dict_from_params(params.pop('token_indexers', {}))
+        use_subtrees = params.pop('use_subtrees', False)
+        granularity = params.pop_choice('granularity', ["5-class", "3-class", "2-class"], True)
+        lazy = params.pop('lazy', False)
+        params.assert_empty(cls.__name__)
+        return StanfordSentimentTreeBankTokensDatasetReader(
```
matt-gardner: One argument per line in multi-line lists like this.  Also, we've typically put the first argument inline, and aligned all other lines to that, but I don't have a strong opinion on that.  This might be getting long enough that you want to put it on the next line...

allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
```diff
@@ -0,0 +1,136 @@
...
+from allennlp.data.instance import Instance
+from allennlp.data.token_indexers import TokenIndexer, SingleIdTokenIndexer
+from allennlp.data.tokenizers import Token
+from allennlp.common.checks import ConfigurationError
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+@DatasetReader.register("sst_tokens")
+class StanfordSentimentTreeBankTokensDatasetReader(DatasetReader):
```
matt-gardner: Why have `Tokens` in the name here?  Doesn't seem like it belongs.  Oh, is it because you could read this as trees, and have some tree encoder?  I'd still take out `Tokens` from the name, and make the case where you're returning trees as the unusual one with a longer name (or just add it as an optional, additional output here, if we ever want that).  When you read the PTB, you get tokens as input and trees as output, so it's not unexpected to get tokens when you see a "treebank" reader.

tests/data/dataset_readers/stanford_sentiment_tree_bank_test.py
```diff
@@ -0,0 +1,94 @@
...
+        assert fields["label"].label == instance1["label"]
+        fields = instances[1].fields
+        assert [t.text for t in fields["tokens"].tokens] == instance2["tokens"]
+        assert fields["label"].label == instance2["label"]
+        fields = instances[2].fields
+        assert [t.text for t in fields["tokens"].tokens] == instance3["tokens"]
+        assert fields["label"].label == instance3["label"]
+
+    def test_use_subtrees(self):
+        reader = StanfordSentimentTreeBankTokensDatasetReader(use_subtrees=True)
```
matt-gardner: Can you build this object `from_params` in one of these tests, so that that block of code gets tested, too?  Or just add a test that explicitly checks that parameters are set, but doing it in line for one of these variants seems fine to me, too.

allennlp/data/dataset_readers/stanford_sentiment_tree_bank.py
nelson-liu: I copied it from PTB reader (I originally wanted to return trees as well as tokens in this dataset, so I started there, but ended up not doing it), but yes you're right.

nelson-liu:
thanks @matt-gardner ! Forgot to ask: would it be possible to add the SST train/dev/test splits to `allennlp/datasets/` on s3? You just have to download, untar, and reupload the splits here: https://nlp.stanford.edu/sentiment/trainDevTestTrees_PTB.zip

matt-gardner:
Yes, I can do that on Monday.  Also, do you have a model that you've trained that uses this dataset reader?  We've love to have the model in here too, and include a demo of it.  The more models, the merrier =).

nelson-liu:
Great, thanks! & not yet, was planning on training a simple LSTM + MLP but you probably want a more SOTA model for a demo. I did hear that Mohit has code for the BCN lying around, and i think that would make for a nice demo if someone wants to port it over... 😃

nelson-liu:
quick question related to this: looks like some papers train on a set with subtrees, but validate and test on the set without subtrees. is there an easy way to do that in allennlp / modify this datareader to provide this option?

matt-gardner:
Yes - you can pass separate parameters for a validation dataset reader.

nelson-liu:
sorry, by parameters do you mean parameters in my json config?

matt-gardner:
Yes.

nelson-liu:
ah I see it now, apologies and thanks!

matt-gardner:
Glad you found it!  Sorry for being terse, I'm in the middle of something else.  I also haven't gotten to uploading stuff to S3 yet, but I'll do that soon.

nelson-liu:
No problem, it’s nothing urgent. Thanks again!

PR #1034, opened 2018-03-26
re-jig readme

DeNeutoy:


README.md
```diff
@@ -7,9 +7,61 @@
...
+
+1.  [Download and install Conda](https://conda.io/docs/download.html).
+
+2.  Create a Conda environment with Python 3.6
+
+    ```
+    conda create -n allennlp python=3.6
+    ```
+
+3.  Activate the Conda environment.  (You will need to activate the Conda environment in each terminal in which you want to use AllenNLP.
```
matt-gardner: Close parentheses, or better yet, just remove them.

README.md
```diff
@@ -7,9 +7,61 @@
...
+
+    ```
+    source activate allennlp
+    ```
+
+4. Install AllenNLP in your environment.
```
matt-gardner: I think you can remove this step - it's not part of "setting up a virtual environment", it's part of the next section.

README.md
```diff
@@ -7,9 +7,61 @@
...
-## Quickstart
+## Running AllenNLP
+
+The preferred way to install AllenNLP is via `pip`.
+
+### Setting up a virtual environment
+
+[Conda](https://conda.io/) can be used set up a virtual environment
+with the version of Python required for AllenNLP and in which you can
+sandbox its dependencies:
```
matt-gardner: I'd give a full stop here, and add a sentence saying something like "if you already have a python 3.6 environment you want to use, you can skip to the 'installing via pip' section."

README.md
```diff
@@ -7,9 +7,61 @@
...
+
+1.  Install AllenNLP.
+
+    ```
+    pip install allennlp
+    ```
+
+2. Visit http://pytorch.org/ and install the PyTorch 0.3.1 package for your system.
+
+3. Download the spacy models.
```
matt-gardner: You don't need to do this anymore.

README.md
```diff
@@ -7,9 +7,61 @@
...
+
+    ```
+    source activate allennlp
+    ```
+
+4. Install AllenNLP in your environment.
+
+### Installing via pip
+
+The preferred way to install AllenNLP into your environment is via `pip`.
```
matt-gardner: You already said this up above.

README.md
```diff
@@ -7,9 +7,61 @@
...
+
+    ```
+    python -m spacy download en_core_web_sm
+    ```
+
+That's it! You're now ready to build and train AllenNLP models.
+AllenNLP installs a script when you install the python package, meaning you can run allennlp commands just by typing `allennlp` into a terminal.
+Now you can do any of the following:
+
+* Run a model on example sentences with `allennlp predict`.
```
matt-gardner: I'd probably just remove this list.  The list is actually a whole lot longer than this; including the whole list is probably not a good idea, and I'm not sure why you picked these two.

README.md
```diff
@@ -7,9 +7,61 @@
 An [Apache 2.0](https://github.com/allenai/allennlp/blob/master/LICENSE) NLP research library, built on PyTorch,
 for developing state-of-the-art deep learning models on a wide variety of linguistic tasks.
 
-## Quickstart
+## Running AllenNLP
+
+The preferred way to install AllenNLP is via `pip`.
```
matt-gardner: I'd also consider just saying something like "Just run `pip install allennlp` in your python 3.6 environment, install pytorch 0.3.1 from their website (link), and you're good to go!  If you need pointers on setting up a python 3.6 environment, see below."

README.md
schmmd: I would promote that users set up a virtual environment because it's something we can (almost) guarantee will work.  I think we should strive to have very clear instructions for new users who are potentially unfamiliar with the Python ecosystem.  Advanced users can probably figure out what they do from the detailed instructions, but beginners will trip up on too many options.

README.md
```diff
@@ -7,9 +7,53 @@
...
+
+tldr; Just run pip install allennlp in your python 3.6 environment, install pytorch 0.3.1 from their website (link), and you're good to go!
```
matt-gardner: The formatting here didn't copy-paste well (and we need an actual link for the pytorch website); other than that, LGTM.

PR #1035, opened 2018-03-26
Simplified logical form recovery code, fixing some errors in the process

matt-gardner:


allennlp/semparse/worlds/world.py
```diff
@@ -245,71 +276,86 @@ def get_logical_form(self,
...
+        if remaining_actions:
+            logger.error("Error parsing action sequence: %s", action_sequence)
+            logger.error("Remaining actions were: %s", remaining_actions)
+            raise ParsingError("Extra actions in action sequence")
+        return str(tree)
+
+    def _construct_node_from_actions(self,
+                                     current_node: _TreeNode,
+                                     remaining_actions: List[List[str]],
+                                     add_var_function: bool) -> List[List[str]]:
```
pdasigi: What does `add_var_function` do? You may want to mention this in the docstring below of add a parameters list.

allennlp/semparse/worlds/world.py
```diff
@@ -245,71 +276,86 @@ def get_logical_form(self,
...
+        elif self.is_terminal(right_side):
+            # The current node is a pre-terminal; we'll add a single terminal child.  We need to
+            # check first for whether we need to add a (var _) around the terminal node, though.
+            if add_var_function and right_side in self._lambda_variables:
+                right_side = f"(var {right_side})"
+            if add_var_function and right_side == 'var':
+                raise ParsingError('add_var_function was true, but action sequence already had var')
+            current_node.children.append(_TreeNode(right_side))
+        else:
+            # We've got a unary non-terminal expansion here.  Not very common, but possible.
```
pdasigi: Can you give an example of such an expansion?

matt-gardner: No; we don't have any currently, so this code path isn't actually touched in the tests.  But it's theoretically possible, so I left this in.

allennlp/semparse/worlds/world.py
matt-gardner: This is a private function, and the parameter is documented in the public `get_logical_form`.  I'm not sure it's worth repeating that documentation here.

allennlp/semparse/worlds/world.py
pdasigi: So that's like `NT1 -> NT2` and `NT2 -> T` or `NT2 -> [NT3, NT4]`? Why would that ever be the case? We could just have `NT1 -> T` or `NT1 -> [NT3, NT4]` instead.

matt-gardner: Do you want to disallow unary productions?  We can.  I'm just not sure that we should.

pdasigi: I cannot think of any cases where we need them. I think we should do that in the interest of making the action space smaller. May not make a huge difference, but still..

matt-gardner: To be clear: when you're designing your action space, you can choose not to include any unary actions (and neither the wikitables nor the nlvr type declarations have them).  What disallowing them here means is that no one can choose to include unary actions, because we disallow them in the base class.  Are you sure that's what we should do?

pdasigi: Right. I meant that I cannot think of a reason anyone would want to include unary actions in their action space, since we can always convert it into an equally expressive action space that does not contain the non-terminals that appear on the right side of unary non-terminal productions in the original action space. So, I think it is alright to disallow users from defining unary non-terminal productions. Do you see any problems with that?

PR #1038, opened 2018-03-27
Add Biattentive Classification Network

nelson-liu:
This PR adds the Biattentive Classification Network (from the salesforce CoVe paper). The model is designed to be trained / evaluated on SST / text classification in general. This implementation is mostly a port of @miyyer 's code (thanks!) that he used for the elmo paper.

Some caveats:

- The original paper has a maxout network with batchnorms at the end, but I've replaced it with a `FeedForward` module. Not sure if we want to add maxout to allennlp?

- The paper is quite sparse on details wrt hyperparameters, and I think some tuning is in order. I included a sample config with this PR with some reasonable defaults, but it doesn't quite reproduce the published numbers. I'll keep hacking at it, but not sure how much of it is due to not using maxout/batchnorm. If someone wants to help me tune hyperparameters, that'd be much appreciated!

- I'm not sure if I'm missing any tests, mostly wrote these tests by looking at what was implemented for the other models. Let me know if you want me to add anything in this regard.


matt-peters:
Thanks for this Nelson!  What is the downside to adding a maxout module to allennlp?  This seems easier then re-tuning the hyper-parameters for a different architecture.

nelson-liu:
Not sure what the proper API is for a maxout. 

Should be be a seq2seq encoder? That makes it kind of hard to stack multiple of them in the json config. 

On the other hand, not sure it fits well in the feedforward API?

Perhaps it should just be a new module altogether? That might be a bit niche, though...

matt-peters:
A new module seems fine to me, unless someone else has a strong feeling otherwise.

matt-gardner:
+1 to just adding a new layer for maxout.  And thanks again, Nelson!

miyyer:
I actually implemented the maxout network initially but could not get it to work... it consistently converged to lower validation scores than the version with the fully-connected network, and batchnorm seemed to always make things worse. If you want to implement the maxout network, I learned from the authors that the CoVe paper computes the max over k=4 channels.

nelson-liu:
@miyyer : thanks, good to know! can you explain what the different channels represent?

miyyer:
sure, channel-wise pooling refers to pooling over feature maps, rather than pooling spatially as in normal maxpool layers in CNNs. so let's say i have a 1024d input vector and i want to reduce its dimensionality to 512d with a maxout layer. I multiply this vector with a 1024 x (512x4) parameter matrix and then reshape to 1024 x 512 x 4, then take the max over the last dimension.

nelson-liu:
Thanks for the clear explanation, Mohit.
I've added a basic Maxout network module, but I'm slightly torn over the proper way to add batchnorm as well. It feels a bit brittle to add a `batchNorm{1/2/3}D` after each layer, since I tried to not make assumptions about the # of input dimensions. But perhaps that's a necessary evil here?

nelson-liu:
It feels like the right thing to do is to add BatchNorm at the BCN level, but I don't see a nice way to do that and maintain the ability for users to use an arbitrary maxout / feedforward network at the end of the model.

matt-gardner:
Thanks for all of this work, Nelson.  Just FYI, I've started reviewing this, but got swamped with other things today and didn't finish.  I'll definitely finish by sometime tomorrow.

nelson-liu:
I emailed Bryan McCann, and he gave me some info about the hyperparameters he used. I added them to the training config, and the model seems to work swimmingly now.

It achieves a best validation accuracy of `52.86`, which is quite close to his published result in the CoVe paper (`53.5`) and a decent improvement over the baseline in the ELMo paper (`51.4`).

Evaluating the best validation checkpoint on the test set yields an accuracy of `53.21`, which seems fairly competitive to me. For reference, these are the test set scores and comparisons in the CoVe paper:

screen shot 2018-03-27 at 11 59 05 pm

Some caveats:
- This model doesn't have batchnorm. I'm happy to add batchnorm, but not sure what the best way to do so is, API-wise.
- The above scores are on a single run. SST is pretty small, so it'd be nice to repeat the experiment several times and use that to get a figure --- however, I need the GPUs for something else now :)

matt-peters:
This is great Nelson!  How much work would it be to add ELMo to the model?

allennlp/models/biattentive_classification_network.py
```diff
@@ -0,0 +1,233 @@
...
+                integrated_encodings, text_mask.unsqueeze(2), +1e7)
+        min_pool = torch.min(min_masked_integrated_encodings, 1)[0]
+        mean_pool = torch.sum(integrated_encodings, 1) / torch.sum(text_mask, 1, keepdim=True)
+
+        # Self-attentive pooling layer
+        flat_self_attentive_logits = self._self_attentive_pooling_projection(
+                integrated_encodings.contiguous().view(-1, self._integrator.get_output_dim()))
+        self_attentive_logits = flat_self_attentive_logits.view(batch_size, sequence_length)
+        self_weights = util.masked_softmax(self_attentive_logits, text_mask)
+        # Do the weighted sum
```
matt-gardner: This comment seems redundant, as do a couple of the ones below.

allennlp/models/biattentive_classification_network.py
```diff
@@ -0,0 +1,233 @@
...
+                               "Pre-encoder feedforward output dim",
+                               "Encoder input dim")
+        check_dimensions_match(self._encoder.get_output_dim() * 3,
+                               self._integrator.get_input_dim(),
+                               "Encoder output dim * 3",
+                               "Integrator input dim")
+        check_dimensions_match(self._integrator.get_output_dim() * 4,
+                               self._output_layer.get_input_dim(),
+                               "Integrator output dim * 4",
+                               "Feedforward classifier input dim")
```
matt-gardner: Is it worth checking the output layer output dim here?

allennlp/models/biattentive_classification_network.py
```diff
@@ -0,0 +1,233 @@
...
+        are passed to the encoder.
+    encoder : ``Seq2SeqEncoder``
+        The encoder to use on the tokens.
+    integrator : ``Seq2SeqEncoder``
+        The encoder to use when integrating the attentive text encoding
+        with the token encodings.
+    integrator_dropout : ``float``
+        The amount of dropout to apply on integrator output.
+    output_layer : ``Union[FeedForward, Maxout]``
+        The feedforward network that takes the final representations and produces
```
matt-gardner: "feedforward or maxout network"?

tests/modules/maxout_test.py
```diff
@@ -0,0 +1,63 @@
...
+from torch.autograd import Variable
+
+from allennlp.common import Params
+from allennlp.common.checks import ConfigurationError
+from allennlp.modules import Maxout
+from allennlp.nn import InitializerApplicator
+from allennlp.common.testing import AllenNlpTestCase
+
+
+class TestFeedForward(AllenNlpTestCase):
```
matt-gardner: Name here needs to be changed.

allennlp/modules/maxout.py
```diff
@@ -0,0 +1,98 @@
...
+        return self._input_dim
+
+    def forward(self, inputs: torch.Tensor) -> torch.Tensor:
+        # pylint: disable=arguments-differ
+        output = inputs
+        for layer, dropout, pool_size in zip(self._linear_layers, self._dropout, self._pool_sizes):
+            affine_output = layer(output)
+            # -1 in shape should be the output_dim of the maxout layer. This lets
+            # the maxout layer generalize the non-2D input.
+            maxed_output = torch.max(affine_output.view(*inputs.size()[:-1], -1, pool_size), dim=-1)[0]
```
matt-gardner: Preference for putting the shape computation on the line above this, and removing the -1, so it's more obvious what's going on here, and less likely to have bugs.  You have a comment that the "-1 in the shape should be the `output_dim`" - why not just use the `output_dim`, then, and remove the need for the comment, and make this crash if somehow there's an error there?  I had to think for a while about whether this was the right computation, and originally thought it was wrong, because the -1 makes it hard to reason about these things.

tests/fixtures/biattentive_classification_network/feedforward_experiment.json
```diff
@@ -0,0 +1,63 @@
+{
```
matt-gardner: For some very non-obvious reasons, it's good to have one of these experiment files just named `experiment.json`.  It makes some things easier when we get to making a `Predictor` for this model.

tests/models/biattentive_classification_network_test.py
```diff
@@ -0,0 +1,65 @@
...
+    def test_mismatching_dimensions_throws_configuration_error(self):
+        params = Params.from_file(self.param_file)
+        # Make the encoder wrong - it should be 2 to match
+        # the embedding dimension from the text_field_embedder.
+        params["model"]["encoder"]["input_size"] = 10
+        with pytest.raises(ConfigurationError):
+            Model.from_params(self.vocab, params.pop("model"))
+
+
+class BiattentiveClassificationNetworkFeedForwardTest(ModelTestCase):
```
matt-gardner: No need to have a whole new class for this, with repeated tests.  You can just have a second `train_save_and_load` test in the main test class, like this: https://github.com/allenai/allennlp/blob/a69209232dec0abc7938a79f7f55fc5e237dc3ef/tests/models/encoder_decoders/wikitables/wikitables_semantic_parser_test.py#L19-L28

tests/models/biattentive_classification_network_test.py
```diff
@@ -0,0 +1,65 @@
...
+        self.set_up_model('tests/fixtures/biattentive_classification_network/maxout_experiment.json',
+                          'tests/fixtures/data/sst.txt')
+
+    def test_bcn_can_train_save_and_load(self):
+        self.ensure_model_can_train_save_and_load(self.param_file)
+
+    def test_batch_predictions_are_consistent(self):
+        self.ensure_batch_predictions_are_consistent()
+
+    def test_forward_pass_runs_correctly(self):
```
matt-gardner: This gets checked by the `train_save_and_load` test.  If you like it, it's fine to have, but we don't typically have a separate test for this.

training_config/biattentive_classification_network.json
```diff
@@ -0,0 +1,72 @@
...
+    "type": "sst_tokens",
+    "use_subtrees": true,
+    "granularity": "5-class"
+  },
+  "validation_dataset_reader":{
+    "type": "sst_tokens",
+    "use_subtrees": false,
+    "granularity": "5-class"
+  },
+  "train_data_path": "https://homes.cs.washington.edu/~nfliu/files/sst/train.txt",
```
matt-gardner: I just uploaded the data to s3, so you can give AI2 s3 links here instead, if you prefer: `https://s3-us-west-2.amazonaws.com/allennlp/datasets/sst/{train,dev,test}.txt`.

training_config/biattentive_classification_network.json
```diff
@@ -0,0 +1,72 @@
...
+  "validation_data_path": "https://homes.cs.washington.edu/~nfliu/files/sst/dev.txt",
+  "test_data_path": "https://homes.cs.washington.edu/~nfliu/files/sst/test.txt",
+  "model": {
+    "type": "bcn",
+    "text_field_embedder": {
+      "tokens": {
+        "pretrained_file": "https://homes.cs.washington.edu/~nfliu/files/misc/glove.840B.300d.txt.gz",
```
matt-gardner: Similarly here, these are on AI2's s3 bucket.  No need to change it if you don't want to; up to you.

nelson-liu: Can you link me to this? The link I would have guessed it would be at (https://s3-us-west-2.amazonaws.com/allennlp/datasets/glove/glove.840B.300d.txt.gz) seems to return a 404. Note that 840B, not 6B.

nelson-liu:
Thanks for the review @matt-gardner , I'll address comments in a bit.

@matt-peters I haven't really looked into how difficult it would be to add ELMo (probably not so hard, judging by Mohit's code). I might be able to add that in the future, and maybe that'd be good for a future PR?

matt-peters:
A future PR sounds fine to me -- it'd be nice to add before adding to the demo and as a working implementation of the result we published.  Based on our prior results, I'd expect a 3-4% increase in the overall accuracy with it.

matt-gardner:
@nelson-liu, if ELMo is just added in the standard way at the input layer, it's just a configuration change, no code change or PR necessary.

training_config/biattentive_classification_network.json
matt-gardner: Hmm, looks like you're right, it was missing.  I just added it.

nelson-liu:
@matt-gardner ah, so you mean using the pretrained elmo representations as detailed here: https://github.com/allenai/allennlp/blob/master/tutorials/how_to/elmo.md ?

sorry, I thought you meant something else.

matt-peters:
@matt-gardner It's a little more complicated in this case since we added two layers of ELMo, one at the input and one at the output of the task biLSTM.  @nelson-liu We can take the details offline.

allennlp/modules/maxout.py
nelson-liu: Sorry for the acrobatics, I wanted to avoid storing the `output_dim` of each of the layers, but it seems worth it for simplicity.

allennlp/models/biattentive_classification_network.py
```diff
@@ -162,16 +164,14 @@ def forward(self,  # type: ignore
         mean_pool = torch.sum(integrated_encodings, 1) / torch.sum(text_mask, 1, keepdim=True)
 
         # Self-attentive pooling layer
-        flat_self_attentive_logits = self._self_attentive_pooling_projection(
-                integrated_encodings.contiguous().view(-1, self._integrator.get_output_dim()))
-        self_attentive_logits = flat_self_attentive_logits.view(batch_size, sequence_length)
+        # Run through linear projection. Shape: (batch_size, sequence length, 1)
+        # Then remove the last dimension to get the proper attention shape (batch_size, sequence length).
+        self_attentive_logits = self._self_attentive_pooling_projection(
+                integrated_encodings.contiguous()).squeeze(2)
```
matt-gardner: Ah, great, this makes much more sense.  It sure looks like you shouldn't need this `.contiguous()`, though.

PR #1041, opened 2018-03-27
CoNLL dataset reader that allows BIOLU tags

matt-peters:


allennlp/data/dataset_readers/conll2003.py
```diff
@@ -56,22 +57,29 @@ class Conll2003DatasetReader(DatasetReader):
         Each will have its own namespace: ``pos_labels``, ``chunk_labels``, ``ner_labels``.
         If you want to use one of the labels as a `feature` in your model, it should be
         specified here.
+    coding_scheme: ``str``, optional (default=``iob1``)
+        Specifies the coding scheme for ``ner_labels`` and ``chunk_labels``.
+        Valid options are ``iob1`` and ``biolu``.
```
DeNeutoy: What does the 1 mean here? Typically i've referred to (I think the same scheme) as BIO tags, it would be good to be consistent

allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -129,3 +129,139 @@ def bio_tags_to_spans(tag_sequence: List[str],
...
+
+def biolu_tags_to_spans(tag_sequence: List[str]) -> List[TypedStringSpan]:
+    """
+    Given a sequence corresponding to BIOLU tags, extracts spans.
+    Spans are inclusive and can be of zero length, representing a single word span.
+    Ill-formed spans are not allowed and will raise ``InvalidTagSequence``.
+
+    Parameters
+    ----------
+    tag_sequence : List[str], required.
```
DeNeutoy: double backticks around the types

allennlp/data/dataset_readers/conll2003.py
DeNeutoy: Yeah e.g in the CRF tagger we refer to these as BIO and BIOUL, could you just match that:
https://github.com/allenai/allennlp/blob/master/allennlp/models/crf_tagger.py#L36

matt-peters: I've seen these referred to all sorts of different names in various places.  IOB1 is distinct from IOB2 so I wanted to be explicit about the BIO tagging scheme we were using.  IOB1 only uses B when it follows an I of the same type, otherwise entity spans start with I.  IOB2 always uses B to start the entity span.

DeNeutoy:
Oh I didn't realise that IOB1 was different - can you be more explicit about that in the docstrings, particularly as the [BIO constraints here](https://github.com/allenai/allennlp/blob/master/allennlp/modules/conditional_random_field.py#L48) follow IOB2

matt-peters:
I added something to the docstrings of the functions / classes I modified and opened a ticket to fix `bio_tags_to_spans`.

DeNeutoy:
Can you give an example, not just say it's the same as the conll data - ideally also change it to bio1 in the functions [here]( https://github.com/allenai/allennlp/pull/1041/files#diff-a280e3d68769d60904411efe50723267R177) - sorry I know I just told you to change it to BIO, I didn't know it was different.

matt-peters:
I added an explicit description of IOB1 to the Conll2003DatasetReader docstring and changed the name back to IOB1 to be explicit.

allennlp/data/dataset_readers/conll2003.py
aslicedbread: > What does the 1 mean here? Typically i've referred to (I think the same scheme) as BIO tags, it would be good to be consistent

Hi~
I'm still confused about the BIO1,BIO2 and BIOUL. Could you please explain it?
Thanks a lot.

PR #1045, opened 2018-03-28
don't allow U tags in BIO to spans, use BIOUL in SpanF1Measure

DeNeutoy:


allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -131,47 +129,48 @@ def bio_tags_to_spans(tag_sequence: List[str],
...
+            start = index
             while label[0] != 'L':
-                k += 1
-                if k >= len(tag_sequence):
+                index += 1
+                if index >= len(tag_sequence):
                     raise InvalidTagSequence
-                label = tag_sequence[k]
+                label = tag_sequence[index]
                 if not (label[0] == 'I' or label[0] == 'L'):
```
joelgrus: I know the code was already like this, but we're inside a `while label[0] != 'L'`, so there's no way `label[0]` could equal `L` here, right?

allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -131,47 +129,48 @@ def bio_tags_to_spans(tag_sequence: List[str],
...
                     raise InvalidTagSequence
-            spans.append((label.partition('-')[2], (start, k)))
+            spans.append((label.partition('-')[2], (start, index)))
         else:
             if label != 'O':
                 raise InvalidTagSequence
-        k += 1
-    return spans
+        index += 1
+    return [span for span in spans if span[0] not in classes_to_ignore]
```
joelgrus: not for this PR, but it looks like this logic allows for invalid tag sequences like

B-PER -> I-ORG -> L-MISC

?

we should probably open an issue for that

allennlp/data/dataset_readers/dataset_utils/span_utils.py
matt-peters: This line checks the next label in the sequence due to the `index += 1` a few lines above inside the loop.

allennlp/data/dataset_readers/dataset_utils/span_utils.py
matt-peters: Good catch, in this case it'd mark the entire span as `MISC`.

allennlp/data/dataset_readers/dataset_utils/span_utils.py
joelgrus: 👍

PR #1046, opened 2018-03-28
Use bilinear attention instead of dot product

matt-gardner:


matt-gardner:
Verdict: this one didn't really help.

PR #1048, opened 2018-03-28
Add a get_final_encoder_states function that handles bidirectional encoders

matt-gardner:


allennlp/modules/seq2seq_encoders/seq2seq_encoder.py
```diff
@@ -27,6 +27,14 @@ def get_output_dim(self) -> int:
         """
         raise NotImplementedError
 
+    def is_bidirectional(self) -> bool:
+        """
+        Returns ``True`` if this encoder is bidirectional.  If so, we assume the forward direction
+        of the encoder is the first half of the final dimension, and the backward direction is the
+        second half.
+        """
+        raise NotImplementedError
```
DeNeutoy: Maybe this should be `False`, given otherwise this will break any user-implemented `Seq2SeqEncoder`s.

matt-gardner: Hmm... Currently this doesn't actually get called anywhere (I'm using it in one place in not-yet-checked-in code on the `wikitables` branch), so no user-implemented code will actually break, unless they try calling this method.  If they try calling it, should we fail to let them know their model needs to implement this, or silently return `False`?

DeNeutoy: Yeah fair enough, in my head I was thinking about [this](https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2vec_encoders/pytorch_seq2vec_wrapper.py#L107) but it's not relevant

PR #1054, opened 2018-03-30
Add a multilabel field

ibeltagy:
Adding a multilabel field which is useful for multi-label multi-class classification. 
Notes:
- I don't like how I am reading the vocabulary size, but I didn't find a better way. I am open to better suggestions. 
- The `as_tensor` function returns a dense vector of zeros and ones. It would have been more efficient to work with a list of indices instead, but that's not supported by torch.nn.BCEWithLogitsLoss

### DIFF
--- a/allennlp/data/fields/multilabel_field.py
+++ b/allennlp/data/fields/multilabel_field.py
@@ -39,9 +39,9 @@ class MultiLabelField(Field[torch.Tensor]):
         If your labels are 0-indexed integers, you can pass in this flag, and we'll skip the indexing
         step.  If this is ``False`` and your labels are not strings, this throws a ``ConfigurationError``.
     num_labels : ``int``, optional (default=None)
-        If ``skip_indexing=True``, the total number of possible labels should be provided, which is needed
-        in deciding the size of the output tensor. `num_labels` should equal largest label id + 1
-        If ``skip_indexing=False``, it should remain None
+        If ``skip_indexing=True``, the total number of possible labels should be provided, which is required
+        to decide the size of the output tensor. `num_labels` should equal largest label id + 1.
+        If ``skip_indexing=False``, `num_labels` is not required.
 
     """
     # It is possible that users want to use this field with a namespace which uses OOV/PAD tokens.
@@ -77,8 +77,6 @@ class MultiLabelField(Field[torch.Tensor]):
             if not all(isinstance(label, str) for label in labels):
                 raise ConfigurationError("MultiLabelFields expects string labels if skip_indexing=False. "
                                          "Found labels: {}".format(labels))
-            if num_labels:
-                raise ConfigurationError("num_labels shouldn't be given when skip_indexing=False.")
 
     def _maybe_warn_for_namespace(self, label_namespace: str) -> None:
         if not (label_namespace.endswith("labels") or label_namespace.endswith("tags")):
@@ -101,8 +99,8 @@ class MultiLabelField(Field[torch.Tensor]):
         if self._label_ids is None:
             self._label_ids = [vocab.get_token_index(label, self._label_namespace)  # type: ignore
                                for label in self.labels]
-
-        self._num_labels = vocab.get_vocab_size(self._label_namespace)
+        if not self._num_labels:
+            self._num_labels = vocab.get_vocab_size(self._label_namespace)
 
     @overrides
     def get_padding_lengths(self) -> Dict[str, int]:  # pylint: disable=no-self-use
--- a/allennlp/data/fields/multilabel_field.py
+++ b/allennlp/data/fields/multilabel_field.py
@@ -1,4 +1,4 @@
-from typing import Dict, Union, Sequence, Set, Optional
+from typing import Dict, Union, Sequence, Set, Optional, cast
 import logging
 
 from overrides import overrides
@@ -68,7 +68,7 @@ class MultiLabelField(Field[torch.Tensor]):
             if not num_labels:
                 raise ConfigurationError("In order to skip indexing, num_labels can't be None.")
 
-            if not all(label < num_labels for label in labels):
+            if not all(cast(int, label) < num_labels for label in labels):
                 raise ConfigurationError("All labels should be < num_labels. "
                                          "Found num_labels = {} and labels = {} ".format(num_labels, labels))
 
--- a/allennlp/data/fields/multilabel_field.py
+++ b/allennlp/data/fields/multilabel_field.py
@@ -1,4 +1,4 @@
-from typing import Dict, Union, Sequence, Set
+from typing import Dict, Union, Sequence, Set, cast
 import logging
 
 from collections import defaultdict
@@ -63,7 +63,7 @@ class MultiLabelField(Field[torch.Tensor]):
                 raise ConfigurationError("In order to skip indexing, your labels must be integers. "
                                          "Found labels = {}".format(labels))
             # vocabulary size = largest label id
-            largest_label_id = max(labels)
+            largest_label_id: int = cast(int, max(labels))
             MultiLabelField._vocab_size[self._label_namespace] = \
                     max(largest_label_id + 1, MultiLabelField._vocab_size[self._label_namespace])
 
--- a/allennlp/data/fields/multilabel_field.py
+++ b/allennlp/data/fields/multilabel_field.py
@@ -1,7 +1,6 @@
-from typing import Dict, Union, Sequence, Set, cast
+from typing import Dict, Union, Sequence, Set, Optional
 import logging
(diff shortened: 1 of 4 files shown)


allennlp/data/fields/multilabel_field.py
```diff
@@ -0,0 +1,129 @@
...
+    # This warning will be repeated for every instantiation of this class (i.e for every data
+    # instance), spewing a lot of warnings so this class variable is used to only log a single
+    # warning per namespace.
+    _already_warned_namespaces: Set[str] = set()
+
+    # Dictionary from a vocab namespace to size. Need the vocab size to map labels to a tensor
+    _vocab_size: Dict[str, int] = defaultdict(int)
+
+    def __init__(self,
+                 labels: Set[Union[str, int]],
```
joelgrus: I think this should be `Sequence` not `Set`, so that it type-checks to pass in a list (as you do in some places)

allennlp/data/fields/multilabel_field.py
```diff
@@ -0,0 +1,129 @@
...
+        self._maybe_warn_for_namespace(label_namespace)
+
+        if skip_indexing:
+            for label in labels:
+                if not isinstance(label, int):
+                    raise ConfigurationError("In order to skip indexing, your labels must be integers. "
+                                             "Found label = {}".format(label))
+                # vocabulary size = largest label id
+                MultiLabelField._vocab_size[self._label_namespace] = \
+                    max(label + 1, MultiLabelField._vocab_size[self._label_namespace])
```
joelgrus: you don't have to do this change, but IMO it would be a little more readable if this wasn't a for loop

```
if skip_indexing:
    if not all(isinstance(label, int) for label in labels):
        raise ...
    largest_label_id = max(labels)
    MultiLabelField._vocab_size[..] = max(largest_label_id + 1, ...)
```

allennlp/data/fields/multilabel_field.py
```diff
@@ -0,0 +1,129 @@
...
+
+            self._label_ids = labels
+        else:
+            for label in labels:
+                if not isinstance(label, str):
```
joelgrus: similarly, I would probably just use `all()` here

allennlp/data/fields/multilabel_field.py
```diff
@@ -0,0 +1,129 @@
...
+                    raise ConfigurationError("MultiLabelFields expects string labels if skip_indexing=False. "
+                                             "Found label: {} with type: {}.".format(label, type(label)))
+
+    def _maybe_warn_for_namespace(self, label_namespace: str) -> None:
```
joelgrus: it looks like this is only ever called in the constructor, so I might just put the logic there rather than have a separate function for it? (and then you can just use `label_namespace` instead of `self._label_namespace` which would make it a tiny bit easier to read

allennlp/data/fields/multilabel_field.py
```diff
@@ -0,0 +1,129 @@
...
+class MultiLabelField(Field[torch.Tensor]):
+    """
+    A ``MultiLabelField`` is an extension of the :class:`LabelField` that allows for multiple labels.
+    It is particularly useful in multi-label classification where more than one label can be correct.
+    As with the :class:`LabelField`, labels are either strings of text or 0-indexed integers (if you wish
+    to skip indexing by passing skip_indexing=True).
+    If the labels need indexing, we will use a :class:`Vocabulary` to convert the string labels
+    into integers.
+
+    This field will get converted into a vector of length equals the vocabulary size with
```
joelgrus: of length equal to

allennlp/data/fields/multilabel_field.py
ibeltagy: done

allennlp/data/fields/multilabel_field.py
ibeltagy: good point. Done

allennlp/data/fields/multilabel_field.py
ibeltagy: done.

allennlp/data/fields/multilabel_field.py
ibeltagy: fixed

allennlp/data/fields/multilabel_field.py
ibeltagy: good point, but how about keeping it as a function just to remain consistent with the code in allennlp/data/fields/label_field.py ?

matt-gardner:
We use to do this in all of our fields.  There's a pretty easy way to do it, like this: https://github.com/allenai/allennlp/blob/048cba07cf94d43376092b61b7e50500bb7f928d/allennlp/data/fields/label_field.py#L61-L75

DeNeutoy:
Yep, we settled on that.

ibeltagy:
Replaced the static variable with something like what @matt-gardner and @DeNeutoy suggested above. Problem with this is that nothing checks if the vocabulary size changed or not, but things will break anyways if the vocabulary size changes.

allennlp/data/fields/multilabel_field.py
```diff
@@ -0,0 +1,130 @@
...
+        self._label_namespace = label_namespace
+        self._label_ids = None
+        self._maybe_warn_for_namespace(label_namespace)
+        self._num_labels = num_labels
+
+        if skip_indexing:
+            if not all(isinstance(label, int) for label in labels):
+                raise ConfigurationError("In order to skip indexing, your labels must be integers. "
+                                         "Found labels = {}".format(labels))
+            if not num_labels:
```
DeNeutoy: This isn't necessary - just add an `if self._num_labels is not None:` block in `index`, because then you could also pass the number of labels if you know it.

allennlp/data/fields/multilabel_field.py
```diff
@@ -0,0 +1,130 @@
...
+        word).  If you have multiple different label fields in your data, you should make sure you
+        use different namespaces for each one, always using the suffix "labels" (e.g.,
+        "passage_labels" and "question_labels").
+    skip_indexing : ``bool``, optional (default=False)
+        If your labels are 0-indexed integers, you can pass in this flag, and we'll skip the indexing
+        step.  If this is ``False`` and your labels are not strings, this throws a ``ConfigurationError``.
+    num_labels : ``int``, optional (default=None)
+        If ``skip_indexing=True``, the total number of possible labels should be provided, which is needed
+        in deciding the size of the output tensor. `num_labels` should equal largest label id + 1
+        If ``skip_indexing=False``, it should remain None
```
DeNeutoy: Full stop.

allennlp/data/fields/multilabel_field.py
```diff
@@ -0,0 +1,130 @@
...
+            if not all(cast(int, label) < num_labels for label in labels):
+                raise ConfigurationError("All labels should be < num_labels. "
+                                         "Found num_labels = {} and labels = {} ".format(num_labels, labels))
+
+            self._label_ids = labels
+        else:
+            if not all(isinstance(label, str) for label in labels):
+                raise ConfigurationError("MultiLabelFields expects string labels if skip_indexing=False. "
+                                         "Found labels: {}".format(labels))
+            if num_labels:
```
DeNeutoy: This isn't necessary - just add an `if self._num_labels is not None:` block in `index`, because then you could also pass the number of labels if you know it.

RamanRajarathinam:
As I see, I need to set 'hidden_dims' size in config.json before I start training MultiLableClassificaiton problem. 
Is there any solution to skip setting hidden_dims and finding from MultiLabelField object?

matt-gardner:
@RamanRajarathinam, hidden dimension has to do with model size, not with data processing.  When constructing a model, there are typically some dimensions that you can compute from your inputs and some that you just need to be given.  You can look at our models to see examples of these different cases.  And if you have further questions, please open a separate issue to ask them.

RamanRajarathinam:
@matt-gardner Sry If I wasn't clear last time. What did I mean by hidden_dims' size is, Its last layer size should be multi-label count. Since multi-label count cannot be determined at start of the process, is there any approach to determine the count value inside the program?

matt-gardner:
The multi-label field uses the vocabulary to determine the number of labels: https://github.com/allenai/allennlp/blob/3cdb7e21c42aa5b3938cbd405cdc0f34f3178e9f/allennlp/data/fields/multilabel_field.py#L101-L102

In your model, you get passed the vocabulary, and you can get the number of labels in the same way, like we do for other classification tasks: https://github.com/allenai/allennlp/blob/3cdb7e21c42aa5b3938cbd405cdc0f34f3178e9f/allennlp/models/crf_tagger.py#L83

PR #1059, opened 2018-04-02
Customizing CACHE_ROOT in file_utiles.py

easonnie:
Update file_utiles.py such that it could accept customized ALLENNLP_CACHE_ROOT.


schmmd:
@easonnie just curious--what is motivating the need for this change?

easonnie:
@schmmd 
I'm working on an AFS server that requires walking between machines. So, I need to save the cache data into a customized path to prevent potential transferring or permission problems. NLTK has the similar workaround. (https://www.nltk.org/data.html)

schmmd:
Thanks!

PR #1062, opened 2018-04-03
install pytorch via pip

DeNeutoy:
Down to that one step pip install 🔥


nelson-liu:
I don't think `pip install torch` is cuda-version invariant. It's compiled against 8.0, so this probably won't work if you're on cuda 9 or 9.1

schmmd:
:fire:

nelson-liu:
Maybe a bit late to argue in favor of this, but I think it makes sense to separate pytorch and allenlp. I personally highly prefer installing pytorch via conda (so I can get CUDA/mkl/cudnn while i'm at it and don't have to mess with that on my system), and this change forces me to use the pip version and then manually go look up the versions of cudnn and other supporting packages that I need (since `conda install` and `pip install` don't quite play nicely yet).

I'd argue in favor of preventing pytorch from being installed with allennlp by default. Maybe it makes the most sense to include pytorch as a setuptools `extras_require`, so folks / build systems who want pytorch automatically can just use `pip install allennlp[pytorch]` (not sold on the name `pytorch` for the `extras_require` name, but that can be ironed out latyer).

schmmd:
Interesting.  I do think having PyTorch installed by default helps our less capable users, whereas our power users are able to set things up the way they'd like them.  Do you know of a way we could add a command that installs `allennlp` without `pytorch`?  That might be an option that would help address the problem you're having.

Since the PyTorch 0.4 release, a common user problem has been that they install AllenNLP with PyTorch 0.4.  Our latest release prevents this issue from happening, since we pin PyTorch to 0.3.

nelson-liu:
I think the best solution (for now) to this would be to add `allennlp` on conda, perhaps via `conda-forge`.

I can work on this if this seems like something you'd be interested in.

PR #1070, opened 2018-04-09
Initializing ERM model from MML model

pdasigi:
I added a method in `NlvrCoverageSemanticParser` to initialize weights from a different (MML trained) model, and a couple of tests. This was mostly straight-forward. The only two weights that cannot be directly copied are the token embeddings, because the vocabularies might be different; and the output projection layer, because the coverage model also has weights for projecting the checklist vector. I copied only the available weights in both cases.
This depends on #1067, which depends on #1053.


pdasigi:
@matt-gardner Yet another PR, sorry! But this one's not big.
Tests are failing because of a missing dependency, "nose". Should it have been in `test_requirements.txt`?

matt-gardner:
Where is `nose` used?

pdasigi:
`numpy.testing.assert_raises`

pdasigi:
Looks like this is a known issue: https://groups.google.com/a/continuum.io/forum/#!topic/anaconda/YLKdBVfs4hs
Yeah, I'll do one of of those other alternatives.

allennlp/models/encoder_decoders/nlvr/nlvr_coverage_semantic_parser.py
```diff
@@ -107,6 +112,55 @@ def __init__(self,
...
+                    # The dimensions differ for this parameter between the coverage model and
+                    # the direct model. In the direct model, this is of size
+                    # (decoder_output_dim + encoder_output_dim, action_embedding_dim),
+                    # whereas in the coverage model, it is
+                    # (decoder_output_dim + encoder_output_dim + checklist_size, action_embedding_dim)
+                    # We copy only the relevant part of the weights here.
+                    archived_projection_weights = weights.data
+                    new_weights = model_parameters[name].data.clone()
+                    new_weights[:, :-self.num_terminals] = archived_projection_weights
+                elif name == "_sentence_embedder.token_embedder_tokens.weight":
```
matt-gardner: Relying on this is pretty brittle, as you're assuming that there's a "tokens" embedder using the "tokens" namespace.  That's fine for now, but I'd at least note that in the docstring for the `initial_mml_model_file` parameter.  If there's a way you could crash if this assumption isn't met, too, that'd be good.

allennlp/models/encoder_decoders/nlvr/nlvr_coverage_semantic_parser.py
```diff
@@ -107,6 +112,55 @@ def __init__(self,
...
+    def _get_vocab_index_mapping(self, archived_vocab: Vocabulary) -> List[Tuple[int, int]]:
+        vocab_index_mapping: List[Tuple[int, int]] = []
+        for index in range(self.vocab.get_vocab_size(namespace='tokens')):
+            token = self.vocab.get_token_from_index(index=index, namespace='tokens')
+            archived_token_index = archived_vocab.get_token_index(token, namespace='tokens')
+            # Checking if we got the UNK token index, because we don't want all new token
+            # representations initialized to UNK token's representation. We do that by checking if
+            # the two tokens are the same. They will not be if the token at the archived index is
+            # UNK.
+            if archived_vocab.get_token_from_index(archived_token_index) == token:
```
matt-gardner: You should pass `namespace='tokens'` here, too.

tests/fixtures/encoder_decoder/nlvr_coverage_semantic_parser/mml_init_experiment.json
```diff
@@ -0,0 +1,68 @@
+{
+  "dataset_reader": {
+    "type": "nlvr",
+    "output_agendas": true
+  },
+  "vocabulary": {
+    "non_padded_namespaces": ["denotations"]
```
matt-gardner: This overrides the `["*tags", "*labels"]` default, so `rule_labels` will actually get padding if you do this.

tests/models/encoder_decoders/nlvr/nlvr_coverage_semantic_parser_test.py
```diff
@@ -52,3 +58,46 @@ def test_forward_with_epoch_num_changes_cost_weight(self):
...
+        mml_model_archive_file = "tests/fixtures/encoder_decoder/nlvr_direct_semantic_parser/serialization/model.tar.gz"
+        archive = load_archive(mml_model_archive_file)
+        archived_model_parameters = archive.model.named_parameters()
+        self.model._initialize_weights_from_archive(archive)
+        changed_model_parameters = dict(self.model.named_parameters())
+        for name, archived_parameter in archived_model_parameters:
+            archived_weight = archived_parameter.data.numpy()
+            original_weight = original_model_weights[name]
+            changed_weight = changed_model_parameters[name].data.numpy()
+            with self.assertRaises(AssertionError):
```
matt-gardner: I don't understand what these two lines are trying to do.  You want the original weights and the new weights to be different?  I guess this works, though it's an odd way to go about it.  A simple comment above this line would be helpful.

allennlp/models/encoder_decoders/nlvr/nlvr_coverage_semantic_parser.py
pdasigi: I added a check to fail if the name is not in one of the two named parameters.

allennlp/models/encoder_decoders/nlvr/nlvr_coverage_semantic_parser.py
pdasigi: Done.

tests/fixtures/encoder_decoder/nlvr_coverage_semantic_parser/mml_init_experiment.json
pdasigi: I'm not sure what you mean. Is this regarding the change to `ModelTestCase`?

tests/models/encoder_decoders/nlvr/nlvr_coverage_semantic_parser_test.py
pdasigi: I'm trying to check if the weights have indeed changed after calling `_initialize_weights_from_archive`. Added a comment.

tests/fixtures/encoder_decoder/nlvr_coverage_semantic_parser/mml_init_experiment.json
matt-gardner: This is in general. If you override this value, you remove the defaults, so they need to be added back in.

pdasigi: Thanks! That's good to know.

PR #1074, opened 2018-04-12
Top menu -> Left menu

aaronsarnat:
image

This PR replaces `<Header />` with (left-aligned) `<Menu />`


aaronsarnat:
@schmmd assigning to you for review/merge. Hopefully this looks good as-is. I was able to test in the React environment, but wasn't able to get the AllenNLP server running (not sure what I was doing wrong), so my testing lacked output content.

I recommend you give this a try locally with the server running so you can test the different model outputs. As I was going through this I noticed there is some really weird stuff happening with the scrolling panes. Something appears to have gotten broken there since the initial release, so I think it would be a good idea for us to take a closer look at that at some point after I'm back. I've had more success keeping fluid, independently-scrollable panes under control on Euclid. Maybe I can port my recipe over to AllenNLP at some point...

schmmd:
Looks good, although there's a problem with how the constituency parser is centered.  I couldn't figure this out, but I'm merging anyways as this is a net improvement.  Thanks!

PR #1076, opened 2018-04-12
Adding an optional projection layer to ElmoTokenEmbedder

matt-gardner:
@matt-peters FYI. @rajasagashe's experiments with the WikiTables parser found that a projection of the ELMo representations was significantly better than just using dropout.  In order to handle the projection cleanly in AllenNLP, it has to be done inside the `ElmoTokenEmbedder`.  This PR adds that.


matt-gardner:
Adding ELMo with just dropout decreased performance by ~4%.  Adding ELMo with a projection layer increased performance by ~3%.  These numbers are all using the preliminary model that still doesn't quite match the original parser yet, so I'll get a clearer picture once I've managed to reproduce the original numbers.

PR #1077, opened 2018-04-12
Fix final encoder state in WikiTables parser, used learned parameters for initial "question attention"

matt-gardner:


matt-gardner:
These tests are failing because I need to rebase the wikitables branch onto master.  I'll do that soon.  The code in this PR won't change, though.

PR #1086, opened 2018-04-13
Adding the decoding framework from the wikitables branch

matt-gardner:
This PR adds the transition-based decoding framework from the `wikitables` branch.  All of the code here has already been reviewed, so this doesn't need a thorough look at the code (unless you _want_ to go over it in detail!).  What I'm looking for in a review at this point is to make sure that the high-level structure of this module makes sense, and that the documentation is sufficient to know what each piece does and when it should be used.

It might be easiest to browse the code in my fork of the repo, found here: https://github.com/matt-gardner/allennlp/tree/decoding/allennlp/nn/decoding.  Start by looking at the module docstring in `__init__.py`, then look over the class structure, and class docstrings.  If anything isn't clear or could use some more documentation, let me know!

### DIFF
--- a/allennlp/nn/decoding/__init__.py
+++ b/allennlp/nn/decoding/__init__.py
@@ -1,18 +1,17 @@
 """
-This module contains code for transition-based decoding.  If you want to do decoding for a
-vocabulary-based model, where the allowable outputs are the same at every timestep of decoding,
-this code is not what you are looking for, and it will be quite inefficient compared to other
-things you could do.
+This module contains code for transition-based decoding.  "Transition-based decoding" is where you
+start in some state, iteratively transition between states, and have some kind of supervision
+signal that tells you which end states, or which transition sequences, are "good".
 
-This code assumes you have a transition-based model, where you start in some state, iteratively
-transition between states, and have some kind of supervision signal that tells you which end
-states, or which transition sequences, are "good".
+If you want to do decoding for a vocabulary-based model, where the allowable outputs are the same
+at every timestep of decoding, this code is not what you are looking for, and it will be quite
+inefficient compared to other things you could do.
 
 The key abstractions in this code are the following:
 
     - ``DecoderState`` represents the current state of decoding, containing a list of all of the
       actions taken so far, and a current score for the state.  It also has methods around
-      determining whether the state is "finished", for combining states for batched computation.
+      determining whether the state is "finished" and for combining states for batched computation.
     - ``DecoderStep`` is a ``torch.nn.Module`` that models the transition function between states.
       Its main method is ``take_step``, which generates a ranked list of next states given a
       current state.
--- a/allennlp/nn/decoding/__init__.py
+++ b/allennlp/nn/decoding/__init__.py
@@ -1,0 +1,36 @@
+"""
+This module contains code for transition-based decoding.  If you want to do decoding for a
+vocabulary-based model, where the allowable outputs are the same at every timestep of decoding,
+this code is not what you are looking for, and it will be quite inefficient compared to other
+things you could do.
+
+This code assumes you have a transition-based model, where you start in some state, iteratively
+transition between states, and have some kind of supervision signal that tells you which end
+states, or which transition sequences, are "good".
+
+The key abstractions in this code are the following:
+
+    - ``DecoderState`` represents the current state of decoding, containing a list of all of the
+      actions taken so far, and a current score for the state.  It also has methods around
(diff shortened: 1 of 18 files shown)


allennlp/nn/decoding/__init__.py
```diff
@@ -0,0 +1,36 @@
+"""
+This module contains code for transition-based decoding.  If you want to do decoding for a
+vocabulary-based model, where the allowable outputs are the same at every timestep of decoding,
+this code is not what you are looking for, and it will be quite inefficient compared to other
+things you could do.
+
+This code assumes you have a transition-based model, where you start in some state, iteratively
```
schmmd: Nit: I'd move this to right after the first sentence.

allennlp/nn/decoding/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+transition between states, and have some kind of supervision signal that tells you which end
+states, or which transition sequences, are "good".
+
+The key abstractions in this code are the following:
+
+    - ``DecoderState`` represents the current state of decoding, containing a list of all of the
+      actions taken so far, and a current score for the state.  It also has methods around
+      determining whether the state is "finished", for combining states for batched computation.
```
schmmd: What does it mean for a state to be "finished"?  I assume the comma right afterwards should be "and".

allennlp/nn/decoding/decoder_state.py
```diff
@@ -0,0 +1,69 @@
...
+    batch_indices : ``List[int]``
+        Our internal variables (like scores, action histories, hidden states, whatever) are
+        `grouped`, and our ``group_size`` is likely different from the original ``batch_size``.
+        This variable keeps track of which batch instance each group element came from (e.g., to
+        know what the correct action sequences are, or which encoder outputs to use).
+    action_history : ``List[List[int]]``
+        The list of actions taken so far in this state.  This is also grouped, so each state in the
+        group has a list of actions.
+
+        The type annotation says this is an ``int``, but none of the training logic relies on this
```
schmmd: Should the type annotation really be `int` given this comment?  I'm a bit confused...

allennlp/nn/decoding/decoder_state.py
```diff
@@ -0,0 +1,69 @@
...
+    the batch of training instances.
+
+    So, using this terminology, a single ``DecoderState`` object represents a `grouped` collection
+    of states.  Because different states in this group might finish at different timesteps, we have
+    methods and member variables to handle some bookkeeping around this, to split and regroup
+    things.
+
+    Parameters
+    ----------
+    batch_indices : ``List[int]``
```
schmmd: My guess is that `batch_indices` is `group_size` long, and that `batch_indices[group_id]` equals the `batch_id` for `group_id`--but it'd be nice to say so explicitly.

allennlp/nn/decoding/decoder_state.py
```diff
@@ -0,0 +1,69 @@
...
+        The list of actions taken so far in this state.  This is also grouped, so each state in the
+        group has a list of actions.
+
+        The type annotation says this is an ``int``, but none of the training logic relies on this
+        being an ``int``.  In some cases, items from this list will get passed as inputs to
+        ``DecodeStep``, so this must return items that are compatible with inputs to your
+        ``DecodeStep`` class.
+    score : ``List[torch.autograd.Variable]``
+        This state's score.  It's a variable, because typically we'll be computing a loss based on
+        this score, and using it for backprop during training.  Like the other variables here, it's
```
schmmd: `s/"Like the other variables here, it's grouped"/"The list length corresponds to the number of groups"/`

allennlp/nn/decoding/__init__.py
matt-gardner: Done.

allennlp/nn/decoding/decoder_state.py
matt-gardner: Yeah, fair enough.  I just removed this section.  The point is just to say that you could probably fudge with the type if you wanted to, but I'm not sure it's necessary to say that.

allennlp/nn/decoding/decoder_state.py
matt-gardner: Done.

allennlp/nn/decoding/decoder_state.py
matt-gardner: Done.

PR #1087, opened 2018-04-13
Adding the type system / logical form parsing code from the wikitables branch

### LINKED ISSUE #1049: Special tokens @@START@@ and @@END@@ in the seq2seq model don't play nicely with lowercasing
`SingleIdTokenIndexer` lowercases the tokens fed to it, if `lowercase_tokens` is set to `true`. This lowercases the special tokens `@@START@@` and `@@END@@` too. This causes problems such as at line 361 of encoder_decoders/simple_seq2seq.py

matt-gardner:
Same basic idea here as the decoding framework, though this PR is much bigger, because it includes actual type systems, contexts and worlds for two datasets.  Pasting the same comment (with slight modification) I gave on the other PR:

This PR adds the logical form parsing framework from the `wikitables` branch.  All of the code here has already been reviewed, so this doesn't need a thorough look at the code (unless you _want_ to go over it in detail!).  What I'm looking for in a review at this point is to make sure that the high-level structure of this module makes sense, and that the documentation is sufficient to know what each piece does and when it should be used.

It might be easiest to browse the code in my fork of the repo, found here: https://github.com/matt-gardner/allennlp/tree/semparse/allennlp/semparse.  Start by looking at the module docstring in `__init__.py`, then look over the class structure, and class docstrings.  If anything isn't clear or could use some more documentation, let me know!

Fixes #1049.


schmmd:
Will look on Monday.

allennlp/semparse/contexts/table_question_knowledge_graph.py
```diff
@@ -0,0 +1,324 @@
...
+        'seventh': 7,
+        'eighth': 8,
+        'ninth': 9,
+        'tenth': 10,
+        **MONTH_NUMBERS,
+        }
+
+class TableQuestionKnowledgeGraph(KnowledgeGraph):
+    """
+    A ``TableQuestionKnowledgeGraph`` represents the linkable entities in a table and a question
```
schmmd: Why do we want to represent the table as a graph?

allennlp/semparse/type_declarations/type_declaration.py
```diff
@@ -0,0 +1,603 @@
...
+with the two improvements above.
+"""
+from typing import Dict, List, Optional, Set, Tuple, Union
+from collections import defaultdict
+
+from overrides import overrides
+from nltk.sem.logic import Expression, ApplicationExpression, ConstantExpression, LogicParser, Variable
+from nltk.sem.logic import Type, BasicType, ComplexType as NltkComplexType, ANY_TYPE
+
+START_SYMBOL = '@START@'
```
DeNeutoy: We already have a start symbol here, can you use this:

https://github.com/allenai/allennlp/blob/master/allennlp/data/dataset_readers/seq2seq.py#L17

allennlp/semparse/type_declarations/type_declaration.py
```diff
@@ -0,0 +1,603 @@
...
+
+
+class ComplexType(NltkComplexType):
+    """
+    In NLTK, a ``ComplexType`` is a function.  These functions are curried, so if you need multiple
+    arguments for your function you nest ``ComplexTypes``.  That currying makes things difficult
+    for us, and we mitigate the problems by adding ``return_type`` and ``argument_type`` functions
+    to ``ComplexType``.
+    """
+    def return_type(self):
```
DeNeutoy: Type hints for this class

allennlp/semparse/type_declarations/type_declaration.py
```diff
@@ -0,0 +1,603 @@
...
+        return list(basic_types)
+    if isinstance(type_, BasicType):
+        return [type_]
+    # If we've made it this far, we have a ComplexType, and we can just call
+    # `type_.substitute_any_type()`.
+    return type_.substitute_any_type(basic_types)
+
+
+def _make_production_string(source: Type, target: Union[List[Type], Type]) -> str:
+    return "%s -> %s" % (str(source), str(target))
```
DeNeutoy: python 3.6 string formatting

allennlp/semparse/type_declarations/wikitables_type_declaration.py
```diff
@@ -0,0 +1,247 @@
...
+        <<r,?>, <?,e>>  :   None
+    """
+    def __init__(self, first: Type, second: Type) -> None:
+        super().__init__(num_arguments=1, first=first, second=second)
+        self._signature = '<<#1,#2>,<#2,#1>>'
+
+    @overrides
+    def resolve(self, other: Type) -> Optional[Type]:
+        # Idea: Since its signature is <<#1,#2>,<#2,#1>> no information about types in self is relevant.
+        # All that matters is that other.fiirst resolves against the reverse of other.second and vice versa.
```
DeNeutoy: \s\fiirst\first

allennlp/semparse/worlds/world.py
```diff
@@ -0,0 +1,519 @@
...
+    """
+    def __init__(self, message):
+        super(ExecutionError, self).__init__()
+        self.message = message
+
+    def __str__(self):
+        return repr(self.message)
+
+
+class _TreeNode:
```
DeNeutoy: Can't you use `nltk.Tree` for this? Also it's unintuitive that you don't pass the children into the constructor.

allennlp/semparse/worlds/world.py
```diff
@@ -0,0 +1,519 @@
...
+        remove_var_function : ``bool`` (optional)
+            ``var`` is a special function that some languages use within lambda founctions to
+            indicate the usage of a variable. If your language uses it, and you do not want to
+            include it in the parsed expression, set this flag. You may want to do this if you are
+            generating an action sequence from this parsed expression, because it is easier to let
+            the decoder not produce this function due to the way constrained decoding is currently
+            implemented.
+        """
+        if not logical_form.startswith("("):
+            logical_form = "(%s)" % logical_form
```
DeNeutoy: python 3.6 strings

allennlp/semparse/worlds/world.py
```diff
@@ -0,0 +1,519 @@
...
+        type_signature.update(self.global_type_signatures)
+        return self._logic_parser.parse(translated_string, signature=type_signature)
+
+    def get_action_sequence(self, expression: Expression) -> List[str]:
+        """
+        Returns the sequence of actions (as strings) that resulted in the given expression.
+        """
+        # Starting with the type of the whole expression
+        return self._get_transitions(expression,
+                                     ["%s -> %s" % (types.START_TYPE, expression.type)])
```
DeNeutoy: python 3.6 strings

allennlp/semparse/worlds/world.py
```diff
@@ -0,0 +1,519 @@
...
+                    mapped_names.append(self._map_name(element))
+                else:
+                    mapped_names.append(self._process_nested_expression(element))
+        if mapped_names[0] == "\\":
+            # This means the predicate is lambda. NLTK wants the variable name to not be within parantheses.
+            # Adding parentheses after the variable.
+            arguments = [mapped_names[1]] + ["(%s)" % name for name in mapped_names[2:]]
+        else:
+            arguments = ["(%s)" % name for name in mapped_names[1:]]
+        return "(%s %s)" % (mapped_names[0], " ".join(arguments))
```
DeNeutoy: python 3.6 strings, Matt I thought I knew you

allennlp/semparse/worlds/world.py
```diff
@@ -0,0 +1,519 @@
...
+            if original_name in self.reverse_name_mapping:
+                original_name = self.reverse_name_mapping[original_name]
+            transition = f"{expression_type} -> {original_name}"
+            current_transitions.append(transition)
+        return current_transitions
+
+    def __eq__(self, other):
+        if isinstance(self, other.__class__):
+            return self.__dict__ == other.__dict__
+        return NotImplemented
```
DeNeutoy: should this not `raise` a `NotImplementedError` rather than return one?

allennlp/semparse/action_space_walker.py
```diff
@@ -0,0 +1,117 @@
...
+        self._completed_paths: List[List[str]] = None
+        self._terminal_path_index: Dict[str, Set[int]] = defaultdict(set)
+        self._length_sorted_paths: List[List[str]] = None
+
+    def _walk(self) -> None:
+        """
+        Walk over action space to collect completed paths of at most ``self._max_path_length`` steps.
+        """
+        # Buffer of NTs to expand, previous actions
+        incomplete_paths = [([str(type_)], [f"@START@ -> {type_}"]) for type_ in
```
DeNeutoy: I know that this is mainly a utility class but can we use the `START_SYMBOL` rather than a string, particularly as here it only has single surrounding `@`s

allennlp/semparse/type_declarations/type_declaration.py
matt-gardner: I don't much like the idea of importing that symbol in other places.  We could have a symbol in `common`, or something, that gets imported in both places.  Yes?

DeNeutoy: Sure - I just don't want there to be two global constants with the same name defined in different places

allennlp/semparse/worlds/world.py
matt-gardner: I managed to get this to work, though I'm not sure I like it better this way.  `nltk.Tree` inherits from `list`, and so it was un-intuitive to figure out how to build the tree correctly.  We're building the tree incrementally from an action sequence, so we don't know a node's children when we construct the node.  But, `nltk.Tree` has some useful functions, so maybe it's worth it.

allennlp/semparse/type_declarations/type_declaration.py
matt-gardner: Thanks, done.

allennlp/semparse/type_declarations/type_declaration.py
matt-gardner: Done.

allennlp/semparse/type_declarations/wikitables_type_declaration.py
matt-gardner: Done.

allennlp/semparse/worlds/world.py
matt-gardner: Done.

allennlp/semparse/worlds/world.py
matt-gardner: Done.

allennlp/semparse/worlds/world.py
matt-gardner: Done.  Looks like I missed this when I first reviewed Pradeep's code =).

allennlp/semparse/worlds/world.py
matt-gardner: Huh, I never paid attention to this.  I'm pretty sure I copied this bit from an answer on stack overflow, and didn't pay close enough attention to the return statement.  It turns out `NotImplemented` is actually a built-in object in python, so pylint didn't even complain about this.  Who knew?

I got myself into a bit of a rabbit hole figuring out exactly what this does, and why it's implemented this way.  Turns out this is the correct way to do things: https://stackoverflow.com/a/879005/1467533.

allennlp/semparse/action_space_walker.py
matt-gardner: Yep, you're right.  Done.

matt-gardner:
And yeah, @DeNeutoy, you're right that the use of `nltk.logic` is opaque and complicated.  `nltk.logic` is also quite slow, and I'd like to remove the need for it.  I think at this point we need to leave more documentation of that piece to future PRs, though.

PR #1094, opened 2018-04-16
Add denotation accuracy computation during validation

matt-gardner:
This uses the scala code found [here](https://github.com/allenai/wikitables-executor).  I updated that code to allow using the jar as a server, and I added that server to the validation done for the model.


allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
```diff
@@ -136,12 +155,66 @@ def __init__(self,
                                                    mixture_feedforward=mixture_feedforward,
                                                    dropout=dropout)
 
+        self._executor_process: subprocess.Popen = None
+        self._create_sempre_executor()
+
+    def _create_sempre_executor(self) -> None:
```
DeNeutoy: Seems like it would be better if you implemented this as a `Metric`? We have other very specific metrics, like the coref stats there already, and it would be a good example for how to have a metric which does complex interactions with other code.

allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
```diff
@@ -136,12 +155,66 @@ def __init__(self,
...
+        """
+        Creates a server running SEMPRE that we can send logical forms to for evaluation.  This
+        uses inter-process communication, because SEMPRE is java code.  We also need to be careful
+        to clean up the process when our program exits.
+        """
+        if self._executor_process:
+            return
+        os.makedirs(SEMPRE_DIR, exist_ok=True)
+        abbreviations_path = os.path.join(SEMPRE_DIR, 'abbreviations.tsv')
```
DeNeutoy: These should either use the allennlp cache or raise errors when they aren't present (the EVALB metric just raises an error - it seems like that would be a good fit here too, because if I was training a model, I would want to know how it was being evaluated rather than it just magically happen for me).

allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
```diff
@@ -360,23 +436,32 @@ def forward(self,  # type: ignore
...
                         # Use a Tensor, not a Variable, to avoid a memory leak.
                         targets = target_action_sequences[i].data
-                        credit = self._action_history_match(best_action_indices, targets)
-                    self._action_sequence_accuracy(credit)
+                        sequence_in_targets = self._action_history_match(best_action_indices, targets)
+
+                    self._action_sequence_accuracy(sequence_in_targets)
                     action_strings = [action_mapping[(i, action_index)] for action_index in best_action_indices]
                     try:
+                        self._has_logical_form(1.0)
```
DeNeutoy: Ah I see why you didn't implement this as a metric - because of the possibility that no logical form is produced. I actually think that's fine though (the produced logical form can just be optionally `None` for the metric), and implementing it as a metric will document this better.

allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
```diff
@@ -570,7 +655,9 @@ def _action_history_match(predicted: List[int], targets: torch.LongTensor) -> in
     @overrides
     def get_metrics(self, reset: bool = False) -> Dict[str, float]:
         return {
-                'parse_acc': self._action_sequence_accuracy.get_metric(reset)
+                'dpd_acc': self._action_sequence_accuracy.get_metric(reset),
```
DeNeutoy: Can you add a quick docstring here? These are complicated metrics and it's not clear what they represent.

allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
```diff
@@ -230,10 +224,14 @@ def text_to_instance(self,  # type: ignore
         table_info : ``str`` or ``JsonDict``
             Table filename or the table content itself, as a dict. See
             ``TableQuestionKnowledgeGraph.read_from_json`` for the expected format.
+        example_string : ``str``, optional
```
DeNeutoy: `example_string` -> `gold_lisp_string`?

DeNeutoy:
Oh sorry I just saw it wasn't this PR that you wanted me to review.... oops!

matt-gardner:
No problem, thanks for the review anyway!  I didn't even think of implementing the executor as a `Metric`, so thanks for the suggestion.  I'll see what I can do...

allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
matt-gardner: Done.

allennlp/data/dataset_readers/wikitables/wikitables_dataset_reader.py
matt-gardner: Done.

allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
```diff
@@ -136,12 +155,66 @@ def __init__(self,
...
+                                                  stdin=subprocess.PIPE,
+                                                  stdout=subprocess.PIPE,
+                                                  bufsize=1)
+
+        for _ in range(6):
+            # SEMPRE outputs six lines of stuff when it loads that I can't disable.  So, we clear
+            # that here.
+            self._executor_process.stdout.readline()
+        logger.info("Started SEMPRE server for evaluating logical forms")
+        atexit.register(self._stop_sempre_executor)
```
pdasigi: What does this do? May be a comment would be useful here.

allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
```diff
@@ -136,12 +155,66 @@ def __init__(self,
...
+        args = ['java', '-jar', cached_path(SEMPRE_EXECUTOR_JAR), 'serve', self._table_directory]
+        self._executor_process = subprocess.Popen(args,
+                                                  stdin=subprocess.PIPE,
+                                                  stdout=subprocess.PIPE,
+                                                  bufsize=1)
+
+        for _ in range(6):
+            # SEMPRE outputs six lines of stuff when it loads that I can't disable.  So, we clear
+            # that here.
+            self._executor_process.stdout.readline()
```
pdasigi: May be assert that the 6 lines you read here are the ones you expect?

allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
```diff
@@ -360,23 +436,32 @@ def forward(self,  # type: ignore
...
-                        credit = self._action_history_match(best_action_indices, targets)
-                    self._action_sequence_accuracy(credit)
+                        sequence_in_targets = self._action_history_match(best_action_indices, targets)
+
+                    self._action_sequence_accuracy(sequence_in_targets)
                     action_strings = [action_mapping[(i, action_index)] for action_index in best_action_indices]
                     try:
+                        self._has_logical_form(1.0)
                         logical_form = world[i].get_logical_form(action_strings, add_var_function=False)
                     except ParsingError:
```
pdasigi: When would exception be raised? It looks like the `if` block this code is in executes only on finished states (or else `i in best_final_states` is `False`), which means `action_strings` should be a complete sequence, right?

allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
```diff
@@ -86,19 +101,23 @@ def __init__(self,
...
+                 table_directory: str = '/wikitables/') -> None:
         super(WikiTablesSemanticParser, self).__init__(vocab)
         self._question_embedder = question_embedder
         self._encoder = encoder
         self._entity_encoder = TimeDistributed(entity_encoder)
         self._beam_search = decoder_beam_search
         self._max_decoding_steps = max_decoding_steps
         self._action_sequence_accuracy = Average()
+        self._denotation_accuracy = Average()
+        self._has_logical_form = Average()
```
pdasigi: `_has_logical_form` -> `_logical_form_is_valid`?

allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
matt-gardner: Done.

allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
matt-gardner: Done.

allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
matt-gardner: It's not that we're producing invalid logical forms, it's that the beam search never found a finished state before it ran out of steps.  So it would be `action_sequence_is_valid`; `has_logical_form` seems just as descriptive.

allennlp/models/encoder_decoders/wikitables/wikitables_semantic_parser.py
matt-gardner: The code that mapped action sequences back to logical forms used to be a lot more buggy and would fail sometimes.  This doesn't really happen anymore, but I didn't remove the try/catch block, just in case.

PR #1100, opened 2018-04-18
Top-K for Constituency Parser

vidurj:
Adds `top_k_trees` to the output of the constituency parser. Fixes minor bug in `resolve_overlap_conflicts_greedily`.

### DIFF
--- a/allennlp/models/constituency_parser.py
+++ b/allennlp/models/constituency_parser.py
@@ -18,6 +18,8 @@ from allennlp.nn.util import last_dim_softmax, get_lengths_from_binary_sequence_
 from allennlp.training.metrics import CategoricalAccuracy
 from allennlp.training.metrics import EvalbBracketingScorer
 from allennlp.common.checks import ConfigurationError
+import numpy as np
+from sortedcontainers import SortedList
 
 class SpanInformation(NamedTuple):
     """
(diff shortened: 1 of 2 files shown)


allennlp/service/predictors/constituency_parser.py
```diff
@@ -86,6 +86,8 @@ def predict_json(self, inputs: JsonDict, cuda_device: int = -1) -> JsonDict:
         tree = return_dict.pop("trees")
         return_dict["hierplane_tree"] = self._build_hierplane_tree(tree, 0, is_root=True)
         return_dict["trees"] = tree.pformat(margin=1000000)
+        return_dict["top_k_trees"] = \
```
DeNeutoy: Format this like 

```
return_dict["top_k_trees"] = [tree.pformat(margin=1000000) 
							  for tree in return_dict["top_k_trees"]]
```

allennlp/models/constituency_parser.py
```diff
@@ -285,169 +419,36 @@ def construct_trees(self,
...
-            spans_to_labels = {(span.start, span.end):
-                               self.vocab.get_token_from_index(span.label_index, "labels")
-                               for span in consistent_spans}
-            sentence_pos = pos_tags[batch_index] if pos_tags is not None else None
-            trees.append(self.construct_tree_from_spans(spans_to_labels, sentence, sentence_pos))
 
+        trees: List[List[Tree]] = []
+        for batch_index in range(len(sentences)):
+            span_to_index = {}
+            for span_index in range(num_spans[batch_index]):
```
DeNeutoy: Can you change this iteration over the batch and spans back to the previous method using `enumerate`? This way, a reader has to know what each of the dimensions of the various tensors represent, whereas previously they had names.

allennlp/models/constituency_parser.py
```diff
@@ -249,21 +252,152 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
...
+                                                num_spans=num_spans,
+                                                sentences=all_sentences,
+                                                pos_tags=all_pos_tags)
+
+        output_dict["trees"] = [trees[0] for trees in top_k_trees]
+        output_dict['top_k_trees'] = top_k_trees
         return output_dict
 
-    def construct_trees(self,
+    def compute_k_best(self,
```
DeNeutoy: `compute_k_best_trees`

allennlp/models/constituency_parser.py
```diff
@@ -249,21 +252,152 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
...
+                       sentence: List[str],
+                       pos_tags: List[str],
+                       label_probabilities: torch.FloatTensor,
+                       span_to_index: Dict[Tuple[int, int], int],
+                       num_trees: List[int],
+                       distinguish_between_labels: bool = False):
+        """
+        :param sentence: The sentence for which top-k parses are being computed.
```
DeNeutoy: Can you use the same formatting for docstrings as the rest of the class? It's important because otherwise our docs don't render.

DeNeutoy: it should look like 

```
        Parameters	
        ----------	
        spans: ``List[SpanInformation]``, required.	
            A list of spans, where each span is a ``namedtuple`` containing the	
            following attributes:	
 	
        start : ``int``	
            The start index of the span.	
        end : ``int``

```

allennlp/models/constituency_parser.py
```diff
@@ -249,21 +252,152 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
...
+
+
+        label_log_probabilities_np = numpy.log(label_probabilities_np + 10 ** -5)
+        correction_term = numpy.sum(label_log_probabilities_np[:, empty_label_index])
+        label_log_probabilities_np -= label_log_probabilities_np[:, empty_label_index]\
+            .reshape((len(span_to_index), 1))
+        cache = {}
+
+        def helper(left, right, must_be_constituent):
+            assert left < right
```
DeNeutoy: Remove this assert statement.

allennlp/models/constituency_parser.py
```diff
@@ -249,21 +252,152 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
...
+                            tree = Tree(label.pop(), [tree])
+                    options.append(([tree], score))
+                cache[span] = options
+            else:
+                children_options = SortedList(key=lambda x: - x[1])
+                for split in range(left + 1, right):
+                    left_trees_options = helper(left, split, must_be_constituent=True)
+                    right_trees_options = helper(split, right, must_be_constituent=False)
+                    for (left_trees, left_score) in left_trees_options:
+                        assert len(left_trees) == 1, 'Toa avoid duplicates we require that left' \
```
DeNeutoy: remove this assert statement.

allennlp/models/constituency_parser.py
```diff
@@ -249,21 +252,152 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
...
+            empty_label_index = 0
+
+
+        label_log_probabilities_np = numpy.log(label_probabilities_np + 10 ** -5)
+        correction_term = numpy.sum(label_log_probabilities_np[:, empty_label_index])
+        label_log_probabilities_np -= label_log_probabilities_np[:, empty_label_index]\
+            .reshape((len(span_to_index), 1))
+        cache = {}
+
+        def helper(left, right, must_be_constituent):
```
DeNeutoy: this name is unhelpful - can you refactor this helper function to contain a lot more comments, like the previous tree assembly code? This is the most complicated function in this entire model and right now it's pretty unreadable.
https://github.com/allenai/allennlp/pull/1100/files?diff=split#diff-d2b9b626f33d5fb4c5fa76a087826bb9L405

allennlp/models/constituency_parser.py
```diff
@@ -249,21 +252,152 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
...
+            label_probabilities_np[:, empty_label_index] = -1
+            for span, span_index in span_to_index.items():
+                label_index = label_probabilities_np[span_index, :].argmax()
+                span_to_label[span] = all_labels[label_index]
+
+            label_probabilities_np = temp
+            empty_label_index = 0
+
+
+        label_log_probabilities_np = numpy.log(label_probabilities_np + 10 ** -5)
```
DeNeutoy: `+ 1e-5`

allennlp/models/constituency_parser.py
```diff
@@ -249,21 +252,152 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
...
+        return trees, scores
+
+    def construct_topk_trees(self,
                         predictions: torch.FloatTensor,
                         all_spans: torch.LongTensor,
                         num_spans: torch.LongTensor,
                         sentences: List[List[str]],
-                        pos_tags: List[List[str]] = None) -> List[Tree]:
+                        pos_tags: List[List[str]],
+                             num_trees: int,) -> List[Tree]:
```
DeNeutoy: de-indent this

allennlp/models/constituency_parser.py
```diff
@@ -249,21 +252,152 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
...
+        :param span_to_index: A dictionary mapping span indices to column indices in
+        label_log_probabilities.
+        :param no_label_id: The id of the empty label.
+        :param num_trees: The number of parses required.
+        :param distinguish_between_labels: Whether to distinguish between different labels for the
+        top k trees.
+        :return: A list of the num_tree parses, and their log probabilities.
+        """
+        empty_label_index = self.vocab.get_token_index("NO-LABEL", "labels")
+        all_labels = [self.vocab.get_token_from_index(index, "labels").split("-") for index in
```
DeNeutoy: This can just be `[label.split("-") for label in self.vocab.get_token_to_index_vocabulary("labels").keys()]`

Also, i'd prefer it if you moved the splitting on `-` to inside the tree construction code as it was before. Having it here makes it very unclear why we would want to split the label up. Also, add back my comment for why we did the splitting.

allennlp/models/constituency_parser.py
```diff
@@ -249,21 +252,152 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
...
+                labels = all_labels
+
+            span_index = span_to_index[span]
+            actions = list(enumerate(label_log_probabilities_np[span_index, :]))
+            actions.sort(key=lambda x: - x[1])
+            actions = actions[:num_trees]
+
+            if right - left == 1:
+                word = sentence[left]
+                pos_tag = pos_tags[left]
```
DeNeutoy: You need to handle the case that the pos tags are `None` here - see [here](https://github.com/allenai/allennlp/pull/1100/files?diff=split#diff-d2b9b626f33d5fb4c5fa76a087826bb9L417)

allennlp/models/constituency_parser.py
```diff
@@ -249,21 +252,152 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
...
+                options = []
+                for label_index, score in actions:
+                    tree = Tree(pos_tag, [word])
+                    if label_index != empty_label_index:
+                        label = list(labels[label_index])
+                        while label:
```
DeNeutoy: Can you handle the case where the user doesn't predict or use POS tags? In this case, we don't want to add pre-terminal nodes for the POS tag. see [here](https://github.com/allenai/allennlp/pull/1100/files?diff=split#diff-d2b9b626f33d5fb4c5fa76a087826bb9L419)

allennlp/models/constituency_parser.py
```diff
@@ -249,21 +252,152 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
...
+                            elif children_options[-1][1] < score:
+                                del children_options[-1]
+                                children_options.add((children, score))
+
+                options = SortedList(key=lambda x: - x[1])
+                for (label_index, action_score) in actions:
+                    for (children, children_score) in children_options:
+                        option_score = action_score + children_score
+                        if label_index != empty_label_index:
+                            label = list(labels[label_index])
```
DeNeutoy: The labels are already a list, no?

allennlp/models/constituency_parser.py
```diff
@@ -249,21 +252,152 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
...
+        :param no_label_id: The id of the empty label.
+        :param num_trees: The number of parses required.
+        :param distinguish_between_labels: Whether to distinguish between different labels for the
+        top k trees.
+        :return: A list of the num_tree parses, and their log probabilities.
+        """
+        empty_label_index = self.vocab.get_token_index("NO-LABEL", "labels")
+        all_labels = [self.vocab.get_token_from_index(index, "labels").split("-") for index in
+                  range(self.vocab.get_vocab_size("labels"))]
+        label_probabilities_np = label_probabilities.cpu().numpy()
```
DeNeutoy: I'd slightly prefer that you do these operations in the tree construction in `torch` rather than numpy, just because it results in less conversion clutter, but I can change this afterward if you prefer numpy.

schmmd:
@vidur-joshi are you interested in following up here?

vidurj:
Yes. I'll get to it this Tuesday. Thanks for the reminder!

allennlp/models/constituency_parser.py
vidurj: Done, thanks.

vidurj:
We have implemented this in Scala for use in Euclid and so no longer require this.

PR #1103, opened 2018-04-19
add name argument to run_with_beaker.py

DeNeutoy:


scripts/ai2-internal/run_with_beaker.py
```diff
@@ -96,7 +97,10 @@ def main(param_file: str, extra_beaker_commands: List[str]):
 
     extra_beaker_commands = []
     if args.desc:
-        extra_beaker_commands.append(f'--desc={args.desc}'),
+        extra_beaker_commands.append(f'--desc={args.desc}')
+    if args.name:
+        # Remove spaces from the name, because Beaker doesn't allow them.
```
schmmd: I wouldn't replace spaces with hyphens as you do here.  I find it confusing to have extra layers like this--although if you prefer having it here that's fine.

I think this would be clearer with a replace command rather than a join/split.

PR #1104, opened 2018-04-19
Adding KnowledgeGraphField and ProductionRuleField

matt-gardner:
This is the last big piece from the `wikitables` branch besides the models and dataset readers themselves.  This one has also already been reviewed, so don't feel like you have to go over this really carefully if you don't want to.

### DIFF
--- a/allennlp/data/fields/knowledge_graph_field.py
+++ b/allennlp/data/fields/knowledge_graph_field.py
@@ -1,0 +1,370 @@
+"""
+``KnowledgeGraphField`` is a ``Field`` which stores a knowledge graph representation.
+"""
+from typing import Callable, Dict, List, Set
+
+import editdistance
+from overrides import overrides
+import torch
+from torch.autograd import Variable
+
+from allennlp.common import util
+from allennlp.common.checks import ConfigurationError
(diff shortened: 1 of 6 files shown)


allennlp/data/fields/knowledge_graph_field.py
```diff
@@ -0,0 +1,370 @@
...
+            self._indexed_entity_texts[indexer_name] = indexer_arrays
+
+    @overrides
+    def get_padding_lengths(self) -> Dict[str, int]:
+        num_entities = len(self.entity_texts)
+        num_entity_tokens = max(len(entity_text) for entity_text in self.entity_texts)
+
+        if self._max_table_tokens:
+            # This truncates the number of entity tokens used, enabling larger tables (which can
+            # have thousands of entities or more than 50 tokens each) to fit in memory,
```
DeNeutoy: of more than 50 tokens each?

matt-gardner: No, the issue is that the table can be big either because there are a ton of entities or because there are a ton of tokens per entity (or both, but either one can be an issue).  I clarified the wording here.

PR #1108, opened 2018-04-19
Adding NLVR model

matt-gardner:
This is taking the code as-is from the `wikitables` branch.  I didn't change anything, except for the location of the modules - I thought `models/semantic_parsing/` made more sense than `models/encoder_decoders/` (and it's more aligned with how we've started to organize other groups of models).


tests/fixtures/semantic_parsing/nlvr_coverage_semantic_parser/serialization/vocabulary/denotations.txt
```diff
@@ -0,0 +1,2 @@
+false
```
pdasigi: Did you mean to commit an untar-ed `model.tar.gz`?

tests/fixtures/semantic_parsing/nlvr_direct_semantic_parser/serialization/vocabulary/denotations.txt
```diff
@@ -0,0 +1,2 @@
+true
```
pdasigi: Same here.

tests/fixtures/semantic_parsing/nlvr_coverage_semantic_parser/serialization/vocabulary/denotations.txt
matt-gardner: This is what gets generated when you run `train_fixtures.py`.  I don't think it's a problem to have this here; seeing the vocabulary files from the test fixtures can actually be helpful.

pdasigi: Good to know. I agree that this is useful to look at, just wasn't sure if this was intended to be checked it.

tests/fixtures/semantic_parsing/nlvr_coverage_semantic_parser/serialization/vocabulary/non_padded_namespaces.txt
```diff
@@ -0,0 +1,2 @@
+actions
```
pdasigi: You may want to regenerate the fixtures though. This seems outdated. The non-padded namespaces should be `denotations` and `rule_labels`. I fixed that in the config files recently.

PR #1111, opened 2018-04-19
Add ability to compute Elmo LM perplexities

rowanz:
Hi all,

I'm submitting a PR to compute perplexities from a language model (in this case, the pretrained Elmo LM). I'm not confident that I organized everything in the right way, but it seems to work for me and hopefully it's helpful enough!

### DIFF
--- a/allennlp/commands/elmo_perplexities.py
+++ b/allennlp/commands/elmo_perplexities.py
@@ -1,0 +1,333 @@
+"""
+Computes ELMo perplexities for a bunch of sentences.
+
+Given a pre-processed input text file, this command outputs the perplexities. The output will be a hdf5 file
+containing a (2, num_tokens) array with the forward and backward perplexities
+
+WARNING #1: This is *NOT* the same thing as an Elmo embedder.
+If that's what you're after, see elmo.py in this folder.
+
+More info:
+The ELMo code concatenates the forward and backward LM states at each layer
+to form a 1024 dimensional vector. The forward states are the first 512 positions
+and the backward states the last 512.  To compute the probability distribution for the
+forward LM take the first 512 positions from the top LSTM layer, multiply by W,
+add the bias b and apply the softmax.  The backward probabilities are computed in a
+similar manner.  If you want, you can verify your implementation by computing the perplexity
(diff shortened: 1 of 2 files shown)


matt-gardner:
Awesome, thanks for the PR!  I haven't looked at this in detail, but it feels to me like this should be a script in the `scripts/` directory, not a command on the main binary.  This is like adding a `bidaf-evaluate` command - it's too specific for what we want as top-level commands to `allennlp`.  @DeNeutoy, @matt-peters, what do you think?  And don't worry @rowanz, this would require changing almost nothing in your PR, just moving the file and tweaking a few things.

DeNeutoy:
Yep - I don't think this should be a top level command. Seems like a useful utility script though! 

Also @rowanz - this is the sort of thing you could develop in an open source repository under `allenai` - it can just use allennlp code via imports.

matt-peters:
I encouraged Rowan to submit a PR as I think having the ability to use the softmax weights with the pre-trained ELMo model is generally useful.  I'd actually like to see the softmax computation piece and the component that loads the pre-trained weights pulled out from this script to live along side the module that loads the pre-trained ELMo biLSTM (or made to work with an existing softmax type module in allennlp).  That way it could be re-used independently from the code that handles the file IO in this script.

allennlp/commands/elmo_perplexities.py
```diff
@@ -0,0 +1,333 @@
...
+                fout.create_dataset(
+                    key,
+                    embeddings.shape, dtype='float32',
+                    data=embeddings
+                )
+        input_file.close()
+
+
+def elmo_command(args):
+    elmo_model = ElmoPerplexityModel(args.options_file, args.weight_file, args.cuda_device)
```
jonghyunc: should have 5 arguments. 
(args.options_file, args.weight_file, args.softmax_file, args.vocab_path, args.cuda_device)

https://github.com/allenai/allennlp/pull/1111/files#diff-ce3ea56cea653aba3170d29e0e71e3e3R129

schmmd:
@rowanz we're going to close this for now as it's getting stale, but please re-open if you are willing to make the recommended changes!

PR #1112, opened 2018-04-20
Allow default predictors when calling programatically

schmmd:
This PR makes two key changes:

* Specifying the predictor is optional when loading an archive.
* Predictors have a predict method, with comments and precise arguments (rather than JSON input).

Motivating issue: https://github.com/allenai/allennlp-internal/issues/33


schmmd:
@joelgrus I plan to make some more changes in this space--but here's a start.

schmmd:
@joelgrus this change remove `build_instances` from the superclass.  Overall that adds a bit of code, but it makes the whole API significantly simpler--particularly when we have exceptions such as with SRL.

If you want to have symmetry with `predict_batch`, it's trivial to add a `predict_json(input: JsonDict, cuda_device: Int)`.

schmmd:
It's faster when you batch them and you can use a GPU. Your approach is
simple to coffee, but won't take advantage of a GPU.

schmmd:
It looks like `predict_batch_json` was added in https://github.com/allenai/allennlp/pull/453.  It's not my goal to rework how we make batch predictions--but rather to get clean syntax for making predictions programatically.

matt-gardner:
Is there a problem with just having each class add a method like:

```python
def predict(self, question: str, passage: str, ...):
    json_obj = {'question': question, 'passage': passage}
    return self.predict_json(json_obj)
```

Seems way simpler than renaming all of the methods while basically keeping the same functionality.

joelgrus:
then I repeat my objection to having both a  "predict using named parameters" and a "predict using json blobs" method.   I feel like that's super confusing.

we should just get in a room and argue about this, because I don't like any of the solutions so far.

schmmd:
@matt-gardner, there's no problem with that at all.  It's basically what I did in https://github.com/allenai/allennlp/pull/1112/commits/d15afe4c2942b3344ce6105062845f92bccee13d (several commits back).  I made further changes based on feedback from Joel.

I'm all in favor of getting together in a room.  I feel like I've wasted a bunch of time on this.

schmmd:
@joelgrus I reverted back and fixed up the documentation / pylint errors.

allennlp/commands/predict.py
```diff
@@ -107,13 +93,9 @@ def _get_predictor(args: argparse.Namespace) -> Predictor:
...
-
-    # Otherwise, use the mapping
-    model_type = archive.config.get("model").get("type")
-    if model_type not in DEFAULT_PREDICTORS:
-        raise ConfigurationError(f"No known predictor for model type {model_type}.\n"
-                                 f"Specify one with the --predictor flag.")
-    return Predictor.from_archive(archive, DEFAULT_PREDICTORS[model_type])
+    else:
+        # Otherwise, use the default
+        return Predictor.from_archive(archive)
```
joelgrus: do you even need the `if` here? (if args.predictor is not specified it will be `None`? which is the default value for that argument anyway)

allennlp/service/predictors/predictor.py
```diff
@@ -67,14 +78,38 @@ def _batch_json_to_instances(self, json_dicts: List[JsonDict]) -> List[Tuple[Ins
...
         """
         Instantiate a :class:`Predictor` from an :class:`~allennlp.models.archival.Archive`;
         that is, from the result of training a model. Optionally specify which `Predictor`
         subclass; otherwise, the default one for the model will be used.
         """
         config = archive.config
 
+        if not predictor_name:
+            model_type = config.get("model").get("type")
+            assert model_type in DEFAULT_PREDICTORS, f"No default predictor for model type {model_type}.\n"\
```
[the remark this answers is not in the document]
schmmd: Will do.

allennlp/commands/predict.py
schmmd: Good call, thanks.

PR #1114, opened 2018-04-20
Adding the wikitables parser

matt-gardner:
Performance of the parser still doesn't match the original model, but it should just be minor tweaks from here.  It'll be a lot easier to manage git if this is merged into master now.

I didn't really change anything in the code, so this is just bringing things over from the `wikitables` branch, and it has all already been reviewed.

There _are_ a few minor changes to `trainer.py` that I don't think got a close enough look previously, and so @DeNeutoy, if you want to give the changes I made there a look, that'd be great.


allennlp/training/trainer.py
```diff
@@ -628,10 +630,16 @@ def _validation_loss(self) -> Tuple[float, int]:
         batches_this_epoch = 0
         val_loss = 0
         for batch in val_generator_tqdm:
-            batches_this_epoch += 1
 
             loss = self._batch_loss(batch, for_training=False)
-            val_loss += loss.data.cpu().numpy()
+            if loss is not None:
+                # You shouldn't necessarily have to compute a loss for validation, so we allow for
```
DeNeutoy: Actually when is this used?

matt-gardner: This is used by the wikitables parser (which is why it's part of this PR). Our training loss is maximum marginal likelihood, but we don't have a set of candidate logical forms for all of the validation instances. We still can calculate accuracy, though, and we want to include all of the instances in the validation set, to get accurate metrics. So we need the training code to not crash when we're evaluating instances at validation time for which we can't actually compute a loss.

DeNeutoy: makes sense 👍

PR #1115, opened 2018-04-21
Added a script to generate data from an ERM trained model

pdasigi:
The other main related change is the option to generate k-best lists from the two nlvr parsers instead of just the top one.
Also, this PR contains a test for the nlvr predictor, but the master branch does not have the predictor yet. This is because I made changes to the test, and was going to open this PR onto `wikitables` first, but then rebased my branch onto `master` as per @matt-gardner's suggestion. May be we can merge this after the predictor is added to `master`..


matt-gardner:
Sorry to be slow here @pdasigi.  I'll look at this today.  Can you rebase this and resolve conflicts first, though?

pdasigi:
No problem, @matt-gardner! I rebased it, and it should now be ready to look at. Thanks!

allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
```diff
@@ -30,16 +30,22 @@ class ExpectedRiskMinimization(DecoderTrainer[Callable[[StateType], torch.Tensor
         al. do this in their work.
     max_decoding_steps : ``int``
         The maximum number of steps we should take during decoding.
+    max_num_decoded_sequences : ``int``, optional (default=1)
```
matt-gardner: Yeah, you put it as a constructor parameter here, where it should go, but then actually put the parameter in `decode`.

scripts/nlvr/generate_data_from_erm_model.py
```diff
@@ -0,0 +1,73 @@
...
+    num_sentences = 0
+    with open(output_file, "w") as outfile:
+        for line in open(input_file):
+            num_sentences += 1
+            input_data = json.loads(line)
+            sentence = input_data["sentence"]
+            structured_representations = input_data["worlds"]
+            labels = input_data["labels"]
+            instance = reader.text_to_instance(sentence, structured_representations)
+            # Tweaking the instance to add an extra field to coerce the model to generate a k-best
```
matt-gardner: Yeah, instead of tweaking the instance, just call `model._decoder_trainer.max_num_decoded_sequences = 100`, or something similar, after loading the model.  If you have this script reach down into the `_decoder_trainer`, the `Model` doesn't have to know about the parameter at all.  Given that this is a pretty rare use case, I don't think it's a big deal to just do `# pylint: ignore=protected-access` on that line where you set that parameter.

scripts/nlvr/generate_data_from_erm_model.py
```diff
@@ -0,0 +1,73 @@
...
+            # list. Setting k to 100 here, so that we can filter out the inconsistent ones later.
+            instance.fields["max_num_decoded_sequences"] = MetadataField(100)
+            outputs = model.forward_on_instance(instance, cuda_device=-1)
+            action_strings = outputs["best_action_strings"]
```
matt-gardner: Don't you already produce logical forms as output in `decode`?

allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
pdasigi: I was going to originally make this a parameter to the constructor, but then realized that the loaded model would set this parameter from the archived config file. So I ended up taking the hacky path of modifying the instances in the script. But I like your idea of directly changing the decoder trainer's parameter in the script. Will do that, thanks!

allennlp/models/semantic_parsing/nlvr/nlvr_coverage_semantic_parser.py
```diff
@@ -108,7 +109,11 @@ def __init__(self,
             self._dynamic_cost_rate = dynamic_cost_weight["rate"]
         self._penalize_non_agenda_actions = penalize_non_agenda_actions
         self._last_epoch_in_forward: int = None
-        if initial_mml_model_file is not None:
+        # TODO (pradeep): Checking whether file exists here to avoid raising an error when we've
```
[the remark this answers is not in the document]
pdasigi: Yes, we don't need the MML model we used to initialize the ERM model when we using the trained ERM model to decode. I cannot think of a way to distinguish between training a decoding in the constructor. I guess I can pass yet another argument to the constructor, but that seems unnecessary.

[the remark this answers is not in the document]
matt-gardner: This is the right thing to do, I think, except maybe log a warning if the file isn't present (in case you're actually training and are expecting the file to be there).  Fixing this issue was part of the original motivation of the `files_to_archive` stuff that @joelgrus added a while back, though that work didn't get as far as adding a mechanism for removing keys from a config that are only necessary at training time.  I'll open an issue to track this.

scripts/nlvr/generate_data_from_erm_model.py
```diff
@@ -0,0 +1,77 @@
...
+              archived_model_file: str,
+              max_num_decoded_sequences: int) -> None:
+    reader = NlvrDatasetReader(output_agendas=True)
+    model = load_archive(archived_model_file).model
+    if not isinstance(model, NlvrCoverageSemanticParser):
+        model_type = type(model)
+        raise RuntimeError(f"Expected an archived NlvrCoverageSemanticParser, but found {model_type} instead")
+    # Tweaking the decoder trainer to coerce the it to generate a k-best list. Setting k to 100
+    # here, so that we can filter out the inconsistent ones later.
+    # pylint: disable=protected-access
```
matt-gardner: This should be inline, after the current line 29.  Putting it right here enables it for the rest of the function, which you don't want.

allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
```diff
@@ -104,9 +112,12 @@ def _get_costs_by_batch(states: List[StateType],
             batch_costs[batch_index].append(cost)
         return batch_costs
 
-    def _get_best_action_sequences(self, finished_states: List[StateType]) -> Dict[int, List[int]]:
+    def _get_best_action_sequences(self,
+                                   finished_states: List[StateType],
+                                   max_num_decoded_sequences: int = 1) -> Dict[int, List[List[int]]]:
```
matt-gardner: Nit: This is a private method, and `max_num_decoded_sequences` is a member variable.  You don't need to pass it as an argument to this method.  Do you gain anything by passing it instead of just reading it from `self`?

allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
```diff
@@ -30,11 +30,18 @@ class ExpectedRiskMinimization(DecoderTrainer[Callable[[StateType], torch.Tensor
...
     """
-    def __init__(self, beam_size: int, normalize_by_length: bool, max_decoding_steps: int) -> None:
+    def __init__(self,
+                 beam_size: int,
+                 normalize_by_length: bool,
+                 max_decoding_steps: int,
+                 max_num_decoded_sequences: int = 1) -> None:
         self._beam_size = beam_size
         self._normalize_by_length = normalize_by_length
         self.max_decoding_steps = max_decoding_steps
```
matt-gardner: Nit: Why is this public while the other three are private?  I have no strong preference here, if you want it to public, it just seems inconsistent.

scripts/nlvr/generate_data_from_erm_model.py
pdasigi: Yes, that's right. I'll also need the action strings because they will be written to the file, but I've changed this to take the logical forms from the outputs to avoid recomputing them.

scripts/nlvr/generate_data_from_erm_model.py
pdasigi: Done. Thanks!

allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
pdasigi: Yes, you're right. Fixed it.

allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
pdasigi: Changed it, thanks!

PR #1126, opened 2018-04-24
upgrade to pytorch 0.4.0

joelgrus:
I think you will be pleased with the results; on the whole the codebase feels quite a bit cleaner.

For the most part I tried to follow the migration guide:

http://pytorch.org/2018/04/22/0_4_0-migration-guide.html

at a high level that included the following

* get rid of every reference to `torch.autograd.Variable`
* replace the typechecks for tensors with their recommended (unidiomatic way), and factor out into a `is_tensor` function
* dealing with the new 0-tensors everywhere (.item() instead of [0])
* replacing the `volatile` flags on tensors with `with torch.no_grad()` context blocks (I *think* I put them in the right places)
* all the `initializer` functions are now `initializer_`
* I had to upgrade tensorboardX to 1.2 (the version we were on didn't work with pytorch 0.4)
* replacing the clunky `new()` logic for making new tensors on specific devices with the new pytorch 0.4 ways of doing things `old_tensor.new_zeros(10)`, `old_tensor.new_tensor([x, y, z])` and so on

what I did not do:

* any of the "writing device agnostic code" stuff in that guide. that's likely a good idea, but it deserves its own PR.

--

When I started there were 100s of warnings; there are still two left that I'm not 100% sure how to fix:

```
tests/models/sniff_test.py::SniffTest::test_coreference_resolution
  /home/joelg/miniconda3/envs/pytorch40/lib/python3.6/site-packages/torch/nn/modules/rnn.py:38: UserWarning: dropout option adds dropout after all but last recurrent layer, so non-zero dropout expects num_layers greater than 1, but got dropout=0.2 and num_layers=1
    "num_layers={}".format(dropout, num_layers))

tests/models/sniff_test.py::SniffTest::test_machine_comprehension
  /home/joelg/miniconda3/envs/pytorch40/lib/python3.6/site-packages/torch/nn/modules/rnn.py:38: UserWarning: dropout option adds dropout after all but last recurrent layer, so non-zero dropout expects num_layers greater than 1, but got dropout=0.2 and num_layers=1
    "num_layers={}".format(dropout, num_layers))
```

both only occur in the `SniffTest`, so maybe they are not worth worrying about. or maybe we need to update the models that those two rely on? in any case, I'm not losing too much sleep over them.

### DIFF
--- a/allennlp/modules/conditional_random_field.py
+++ b/allennlp/modules/conditional_random_field.py
@@ -310,7 +310,8 @@ class ConditionalRandomField(torch.nn.Module):
                     -10000.0 * (1 - self._constraint_mask[:num_tags, end_tag].detach())
             )
         else:
-            transitions[start_tag, :num_tags] = -10000.0 * (1 - self._constraint_mask[start_tag, :num_tags].detach())
+            transitions[start_tag, :num_tags] = (-10000.0 *
+                                                 (1 - self._constraint_mask[start_tag, :num_tags].detach()))
             transitions[:num_tags, end_tag] = -10000.0 * (1 - self._constraint_mask[:num_tags, end_tag].detach())
 
         all_tags = []
--- a/allennlp/tests/modules/conditional_random_field_test.py
+++ b/allennlp/tests/modules/conditional_random_field_test.py
@@ -67,7 +67,8 @@ class TestConditionalRandomField(AllenNlpTestCase):
         # (which is the log-sum-exp of the scores for the logits across all possible tags)
         for logits_i, tags_i in zip(self.logits, self.tags):
             numerator = self.score(logits_i.detach(), tags_i.detach())
-            all_scores = [self.score(logits_i.detach(), tags_j) for tags_j in itertools.product(range(5), repeat=3)]
+            all_scores = [self.score(logits_i.detach(), tags_j)
+                          for tags_j in itertools.product(range(5), repeat=3)]
             denominator = math.log(sum(math.exp(score) for score in all_scores))
             # And include them in the manual calculation.
             manual_log_likelihood += numerator - denominator


allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
```diff
@@ -49,7 +49,11 @@ def decode(self,
             # Finished model scores are log-probabilities of the predicted sequences. We convert
             # log probabilities into probabilities and re-normalize them to compute expected cost under
             # the distribution approximated by the beam search.
-            costs = torch.cat(finished_costs[batch_index])
+
+            # finished_costs[batch_index] could be a list of 0-tensors or 1-tensors. We use .view(-1)
+            # to make sure they're treated as 1-tensors and then concatenated appropriately.
+            # TODO(joelgrus): make sure this is correct
```
matt-gardner: This looks like it shouldn't be right to me.  Well, it looks like it probably does the right thing, but I think fixing this where the `finished_costs` gets created is the right solution, instead of this one.  My guess is that something is getting created as a scalar now, where we really want it as a 1-dim tensor.

schmmd:
Well, I only got up to `allennlp/modules/augmented_lstm.py` but what I looked at was great!  Have you tried our pre-trained models on this branch?  Thanks for putting this together.

joelgrus:
only via the `SniffTest`, which passes.

off the top of my head, there's no reason why *these changes* should affect existing trained models (unless I did something incorrectly). but it's possible there are other changes in pytorch 0.4 that would.

allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
```diff
@@ -56,8 +55,10 @@ def decode(self,
             # Finished model scores are log-probabilities of the predicted sequences. We convert
             # log probabilities into probabilities and re-normalize them to compute expected cost under
             # the distribution approximated by the beam search.
-            costs = torch.cat(finished_costs[batch_index])
-            logprobs = torch.cat(finished_model_scores[batch_index])
+
+            print("size:", finished_costs[batch_index])
```
DeNeutoy: remove print statement

allennlp/nn/decoding/decoder_trainers/maximum_marginal_likelihood.py
```diff
@@ -65,7 +64,8 @@ def decode(self,
         batch_scores = self._group_scores_by_batch(finished_states)
         loss = 0
         for scores in batch_scores.values():  # we don't care about the batch index, just the scores
-            loss += -util.logsumexp(torch.cat(scores))
+            print("scores", scores)
```
DeNeutoy: remove print statement

allennlp/nn/decoding/decoder_trainers/maximum_marginal_likelihood.py
```diff
@@ -65,7 +64,8 @@ def decode(self,
...
+            loss += -util.logsumexp(torch.cat([tensor.view(-1) for tensor in scores]))
```
DeNeutoy: I think doing a squeeze here is tighter - I was confused when I read this, as it isn't immediately obvious that it's a 1D tensor and that doing a view like this is the same.

allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
```diff
@@ -56,8 +55,10 @@ def decode(self,
...
+            costs = torch.cat([tensor.view(-1) for tensor in finished_costs[batch_index]])
+            logprobs = torch.cat([tensor.view(-1) for tensor in finished_model_scores[batch_index]])
```
DeNeutoy: Can you use `squeeze` instead of `view` here?

schmmd:
[MC model](http://beaker.allenai.org/ex/ex_xbrfg02q5nr7/tasks) evaluation matches.
[TE model](http://beaker.allenai.org/ex/ex_s9f5kgm4nboe/tasks) evaluation matches.

allennlp/training/metrics/boolean_accuracy.py
```diff
@@ -34,7 +34,6 @@ def __call__(self,
         mask: ``torch.Tensor``, optional (default = None).
             A tensor of the same shape as ``predictions``.
         """
-        # Get the data from the Variables.
         predictions, gold_labels, mask = self.unwrap_to_tensors(predictions, gold_labels, mask)
```
matt-gardner: This method feels like it should be renamed to something like `detach_and_move_to_cpu` (or something similar but more compactly named...), and maybe just moved to `nn.util`.

allennlp/training/metrics/metric.py
```diff
@@ -50,9 +51,11 @@ def from_params(cls, params: Params, vocab: Optional[Vocabulary] = None):
...
     def unwrap_to_tensors(*tensors):
         """
-        If you actually passed in Variables to a Metric instead of Tensors, there will be
+        If you actually passed gradient-tracking Tensors to a Metric, there will be
         a huge memory leak, because it will prevent garbage collection for the computation
         graph. This method ensures that you're using tensors directly and that they are on
         the CPU.
         """
-        return (x.data.cpu() if isinstance(x, torch.autograd.Variable) else x for x in tensors)
+        # pylint: disable=unidiomatic-typecheck
```
matt-gardner: I don't think you need this anymore.

allennlp/training/metrics/metric.py
```diff
@@ -50,9 +51,11 @@ def from_params(cls, params: Params, vocab: Optional[Vocabulary] = None):
...
+        return (x.data.cpu() if is_tensor(x) else x
```
matt-gardner: One line?

matt-gardner: And probably just fail if you pass something that's not a tensor, instead of checking `is_tensor`.

matt-gardner: `.detach()` instead of `.data`

allennlp/training/metrics/metric.py
```diff
@@ -50,9 +51,11 @@ def from_params(cls, params: Params, vocab: Optional[Vocabulary] = None):
     @staticmethod
     def unwrap_to_tensors(*tensors):
```
matt-gardner: Can we give a type annotation here now?  Not sure how that works with `*args`, though...

allennlp/training/trainer.py
```diff
@@ -463,11 +461,11 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
                 # get the magnitude of parameter updates for logging
                 # We need a copy of current parameters to compute magnitude of updates,
                 # and copy them to CPU so large models won't go OOM on the GPU.
-                param_updates = {name: param.detach().data.cpu().clone()
+                param_updates = {name: param.data.cpu().clone()
```
matt-gardner: `.detach()` instead of `.data`

allennlp/training/trainer.py
```diff
@@ -463,11 +461,11 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
...
                                  for name, param in self._model.named_parameters()}
                 self._optimizer.step()
                 for name, param in self._model.named_parameters():
-                    param_updates[name].sub_(param.detach().data.cpu())
+                    param_updates[name].sub_(param.data.cpu())
```
matt-gardner: `.detach()` instead of `.data`

tests/nn/util_test.py
```diff
@@ -530,7 +513,7 @@ def test_flatten_and_batch_shift_indices(self):
                                [[2, 1, 0, 7],
                                 [7, 7, 2, 3],
                                 [0, 0, 4, 2]]])
-        indices = Variable(torch.LongTensor(indices))
+        indices = torch.from_numpy(indices).long()
```
matt-gardner: `dtype=torch.long`

tests/nn/util_test.py
```diff
@@ -546,8 +529,7 @@ def test_batched_index_select(self):
         targets = torch.ones([2, 10, 3]).cumsum(1) - 1
         # Make the second batch double it's index so they're different.
         targets[1, :, :] *= 2
-        indices = Variable(torch.LongTensor(indices))
-        targets = Variable(targets)
+        indices = torch.from_numpy(indices).long()
```
matt-gardner: `dtype=torch.long`

tests/nn/util_test.py
```diff
@@ -568,8 +550,7 @@ def test_flattened_index_select(self):
         targets = torch.ones([2, 6, 3]).cumsum(1) - 1
         # Make the second batch double it's index so they're different.
         targets[1, :, :] *= 2
-        indices = Variable(torch.LongTensor(indices))
-        targets = Variable(targets)
+        indices = torch.from_numpy(indices).long()
```
matt-gardner: `dtype=torch.long`

tests/training/trainer_test.py
```diff
@@ -244,5 +244,5 @@ def test_sparse_clip_grad(self):
         # Now try to clip the gradients.
         _ = sparse_clip_norm([embedding.weight], 1.5)
         # Final norm should be 1.5
-        grad = embedding.weight.grad.data.coalesce()
-        self.assertAlmostEqual(grad._values().norm(2.0), 1.5, places=5) # pylint: disable=protected-access
+        grad = embedding.weight.grad.data.coalesce()  # pylint: disable=no-member
```
matt-gardner: Do you need the `.data` here still?

schmmd:
BiDAF trained too: http://beaker.allenai.org/ex/ex_fh0xt1u5ytst/tasks

allennlp/nn/decoding/decoder_trainers/maximum_marginal_likelihood.py
joelgrus: the actual issue here (which I didn't dig into too deeply) is that sometimes the score was a 0-tensor and sometimes it was a 1-tensor, and the `.view()` is ensuring it's a 1-tensor.

allennlp/common/util.py
```diff
@@ -47,9 +52,7 @@ def sanitize(x: Any) -> Any:  # pylint: disable=invalid-name,too-many-return-sta
     if isinstance(x, (str, float, int, bool)):
         # x is already serializable
         return x
-    elif isinstance(x, torch.autograd.Variable):
-        return sanitize(x.data)
-    elif isinstance(x, torch._TensorBase):  # pylint: disable=protected-access
+    elif is_tensor(x):
         # tensor needs to be converted to a list (and moved to cpu if necessary)
         return x.cpu().tolist()
```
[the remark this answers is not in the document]
joelgrus: I don't think it would do anything. as I understand it, the main reason to use `.data` is to "break free" of the computational graph; once you convert to a list you're not part of the computational graph anyway

allennlp/models/semantic_parsing/wikitables/wikitables_semantic_parser.py
```diff
@@ -398,7 +397,7 @@ def forward(self,  # type: ignore
                 if i in best_final_states:
                     best_action_indices = best_final_states[i][0].action_history[0]
                     if target_action_sequences is not None:
-                        # Use a Tensor, not a Variable, to avoid a memory leak.
+                        # Detach to avoid a memory leak.
                         targets = target_action_sequences[i].data
```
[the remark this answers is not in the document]
joelgrus: I was a little confused on this, here's what the migration docs say:

```
However, .data can be unsafe in some cases. Any changes on x.data wouldn’t be tracked by autograd, and the computed gradients would be incorrect if x is needed in a backward pass. A safer alternative is to use x.detach(), which also returns a Tensor that shares data with requires_grad=False, but will have its in-place changes reported by autograd if x is needed in backward.
```

to me that suggests that `detach` still retains some connection to the computational graph and might still result in a memory leak?

allennlp/modules/encoder_base.py
```diff
@@ -235,8 +233,7 @@ def _update_states(self,
         if self._states is None:
             # We don't already have states, so just set the
             # ones we receive to be the current state.
-            self._states = tuple([torch.autograd.Variable(state.data)
-                                  for state in new_unsorted_states])
+            self._states = tuple(new_unsorted_states)
```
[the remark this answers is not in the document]
joelgrus: oops, just a mistake on my part.

allennlp/nn/util.py
```diff
@@ -345,12 +335,13 @@ def get_text_field_mask(text_field_tensors: Dict[str, torch.Tensor],
     the mask.  Most frequently this will be a character id tensor, but it could also be a
     featurized representation of each token, etc.
 
+    TODO(joelgrus): can we change this?
     NOTE: Our functions for generating masks create torch.LongTensors, because using
-    torch.ByteTensors inside Variables makes it easy to run into overflow errors
+    torch.ByteTensors  makes it easy to run into overflow errors
```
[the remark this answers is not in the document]
joelgrus: I need to fix this whole comment, because I'm not sure if it's true anymore. Now that variables don't exist, is this still a real problem? @DeNeutoy

allennlp/training/metrics/metric.py
joelgrus: removing the `else x` behavior breaks a lot of tests (not sure if it's just how the tests are written or if it actually breaks our code)

matt-gardner: I just checked this, and it's because we pass in masks that could be `None`.  It might be worth changing this to `(x.data.cpu() if x is not None else None for x in tensors)`, just to fail explicitly instead of silently if you're passing something in that we're not expecting.  Not a big deal, though.

allennlp/models/semantic_parsing/wikitables/wikitables_semantic_parser.py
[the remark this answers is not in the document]
matt-gardner: I just asked about this in the pytorch slack, and I got this response:

"Variables returned by `.detach()` do not hold on to the computation graph. You can almost certainly use `.detach()` in places where you are using `.data`

"They do share a version counter with the original tensor. This doesn't require any extra memory usage. It just means that changes to the detached variable increment the (shared) version counter on  original tensor. So if you do backprop on the original tensor, and it's needed for a gradient computation, you would get an error message instead of just incorrect gradients."

murphp15:
@schmmd On a number of the PR's I have noticed that you guys post links to beaker. 
Is this something the public should not have access to?

matt-gardner:
@murphp15, beaker is an experiment management system that we're building internally at AI2.  We submit jobs to a cluster using beaker, and it keeps track of the experiments that we run.  It's pretty nice.  The plan is to make it open source eventually, but I don't know what the timeline is for that.

schmmd:
@murphp15 it's internal, but hopefully we'll be able to share external links someday.  We're working out our plan on that currently.

schmmd:
We plan to merge this on Tuesday.

PR #1129, opened 2018-04-25
Fix state scores in WikiTablesDecoderStep

matt-gardner:
I also added a convenience method for printing out the score and action sequence in a WikiTablesDecoderState, which is helpful for examining what's going on during decoding / beam search.


allennlp/models/semantic_parsing/wikitables/wikitables_decoder_step.py
```diff
@@ -84,6 +84,7 @@ def take_step(self,
         memory_cell = torch.stack([rnn_state.memory_cell for rnn_state in state.rnn_state])
         previous_action_embedding = torch.stack([rnn_state.previous_action_embedding
                                                  for rnn_state in state.rnn_state])
+        score_so_far = torch.stack(state.score)
```
pdasigi: Show the shape of this tensor and may be write a comment describing what the addition below is doing?

PR #1132, opened 2018-04-25
clean EVALB via make, add to Dockerfile

DeNeutoy:
FYI @cristipp

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ RUN apt-get update --fix-missing && apt-get install -y \
     libevent-dev \
     build-essential && \
     rm -rf /var/lib/apt/lists/*
-    
+
 # Install Java.
 RUN echo "deb http://http.debian.net/debian jessie-backports main" >>/etc/apt/sources.list
 RUN apt-get update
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -65,6 +65,8 @@ COPY setup.py setup.py
 ARG CACHE_MODELS=false
 RUN ./scripts/cache_models.py
 
+RUN cd scripts/EVALB && make && cd ..
```
matt-gardner: How long does this take?  Is it possible to move it earlier in the docker file, so it doesn't get rebuilt every time `allennlp/` changes?

joelgrus: it looks like on teamcity it only took 1 second:

http://build.allennlp.org/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=3034&_focus=190

matt-gardner: Plus a bunch of annoying output...  I'm just thinking of when I use `run_with_beaker.py`, with changes to `allennlp`.  Maybe we can move the `scripts/` copy command to be the first copy, then have this command, then the rest of the copies?

DeNeutoy: good point, i'll change that because it is a little alarming

matt-gardner: Alternatively, you could just pipe the output to /dev/null, or something.  Or do both of those.

PR #1133, opened 2018-04-25
Move imports to inside of CACHE_MODELS=true condition

matt-gardner:
Importing `allennlp` is, unfortunately, slow, and causes several seconds of pointless waiting when `CACHE_MODELS=false`.  This removes the unnecessary wait by moving the imports inside the condition where `CACHE_MODELS=true`.

### DIFF
--- a/scripts/cache_models.py
+++ b/scripts/cache_models.py
@@ -1,14 +1,13 @@
 #!/usr/bin/env python
 
 import os
-import sys
-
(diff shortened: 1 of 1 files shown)


matt-gardner:
`from allennlp.commands.serve import DEFAULT_MODELS`.  The first `allennlp` import takes several seconds, as it imports spacy, torch, and a bunch of other stuff, which collectively takes a while.

PR #1138, opened 2018-04-26
NER model with ELMo

matt-peters:
Adding ELMo to the current NER model and adjusting hyperparameters.  ConLL 2003 test set F1 from one run was 92.53 compared to the reported value of 92.22 +/- 0.10 over 5 seeds from [the ELMo paper](https://arxiv.org/abs/1802.05365).

### DIFF
--- a/training_config/ner.json
+++ b/training_config/ner.json
@@ -1,0 +1,90 @@
+{
+  "dataset_reader": {
+    "type": "conll2003",
+    "tag_label": "ner",
+    "coding_scheme": "BIOUL",
+    "token_indexers": {
+      "tokens": {
+        "type": "single_id",
+        "lowercase_tokens": true
+      },
+      "token_characters": {
+        "type": "characters"
+      },
+      "elmo": {
+        "type": "elmo_characters"
+     }
+    }
+  },
+  "train_data_path": "/home/matthewp/data/conll2003_ner/eng.train",
+  "validation_data_path": "/home/matthewp/data/conll2003_ner/eng.testa",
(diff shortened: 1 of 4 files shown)


training_config/ner.json
```diff
@@ -16,9 +16,9 @@
      }
     }
   },
-  "train_data_path": "/path/to/eng.train",
-  "validation_data_path": "/path/to/eng.testa",
-  "test_data_path": "/path/to/eng.testb",
+  "train_data_path": "${NER_TRAIN_DATA_PATH}",
+  "validation_data_path": "${NER_TEST_A_PATH}",
+  "test_data_path": "/${NER_TEST_B_PATH}",
```
matt-peters: Thanks, good to know about the error message. Extra leading slash here

DeNeutoy:
Looks great - thanks for doing this Matt, it would have taken me considerably longer to tweak all the params 👍

joelgrus:
the one thing I would warn about (and it's possible you don't care) is that if you have `Parameter`s that only exist conditionally based on the config, you might run into trouble trying to make predictions from a trained model with a different config.

that is, if I train a model with start-end transitions enabled, and then want to predict without them, I think deep in the code `torch.load` will get angry that that there are tensors corresponding to (now) non-existent parameters, and vice versa the other way.

but it's possible that shouldn't even be allowed, in which case it doesn't matter.

(I ran into this issue when writing the "constraint mask" code of the CRF)

DeNeutoy:
I don't think that's an issue we can really get around here - we have plenty of models with conditional components.

schmmd:
89% -> 92.5% ;-)

PR #1139, opened 2018-04-26
decorator + check that `from_params` is overridden

### LINKED ISSUE #1127: failure to override `from_params` causes misleading error messages
is there a way we could check for this and provide a more helpful error? (not sure)

see #1121

joelgrus:
fixes #1127 

let me know what you think of this approach.

originally I just wanted to check directly, something like

```python
if subclass_from_params == Model.from_params:
    ...
```

but it turns out that that doesn't work -- even if the subclass doesn't override the method, the subclass method isn't *equal to* the base class one, because they're bound to different objects. I couldn't see an obvious way to make it work that wasn't either ugly or fragile, so I ended up using the decorator approach


tests/data/dataset_readers/dataset_reader_test.py
```diff
@@ -0,0 +1,29 @@
...
+class YesOverridden(DatasetReader):
+    @classmethod
+    def from_params(cls, params: Params) -> 'YesOverridden':
+        return YesOverridden()
+
+class TestDatasetReader:
+    def test_raises_without_override(self):
+        params = Params({'type': 'not-overridden'})
+
+        with pytest.raises(ConfigurationError) as exc:
```
matt-gardner: You could use the `match` or `message` argument to `pytest.raises` instead of keeping around the exception variable and checking it afterward.  Same on the model test.

joelgrus:
#1191 is a better solution for this issue

PR #1140, opened 2018-04-26
Streamlining the docker build a little bit

matt-gardner:
1. Minor tweaks to the order of copies in the dockerfile.
2. Redirecting stderr as well as stdout in the `EVALB` command.
3. Hiding building the demo behind an argument and a script, because it is slow.

It's possible that our tests rely on the demo; CI will tell us that for sure.  If so, I'll try to find a way to make the tests not rely on the demo...

@schmmd, this means that something has to change in the demo build, in the same place where `CACHE_MODELS=true` is specified, wherever that is.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,16 +48,21 @@ COPY requirements_test.txt .
 COPY scripts/install_requirements.sh scripts/install_requirements.sh
 RUN INSTALL_TEST_REQUIREMENTS="true" ./scripts/install_requirements.sh
 
-# Build demo
-COPY demo/ demo/
-RUN cd demo && npm install && npm run build && cd ..
+# Same idea as above, but for EVALB this time.
+COPY scripts/EVALB scripts/EVALB
 
-COPY scripts/ scripts/
 # Compile EVALB - required for parsing evaluation.
 # EVALB produces scary looking c-level output which we don't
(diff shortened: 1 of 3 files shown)


schmmd:
@matt-gardner what is your goal here?  To speed up continuous integration?  If that is your goal, I'm worried we'll need a much larger change than you have here since this will break our demo deployment system, which relies on our Docker images having the demo built.

matt-gardner:
My goal is to reduce the time I sit waiting when I run `run_with_beaker.py`.

matt-gardner:
I thought the demo build on teamcity passed in the `CACHE_MODELS=true` flag.  If so, it'd be trivial to also add in a `BUILD_DEMO=true` flag in the same place.  If that happens for all of CI, that's fine, as it's not as big a deal to wait a few more seconds on CI as it is to wait a few seconds on the command line when I'm starting an experiment.

matt-gardner:
I just updated both of those places to include `--build-arg BUILD_DEMO=true`.  I'll verify that this did what I expected in this latest merge build.

schmmd:
Ha, funny, at the exact same time you added me as a reviewer ;-)

matt-gardner:
What's the difference between requesting a review and assigning a PR?

Issue #1128, opened 2018-04-25
data_iterator unnecessarily rebuilds the dataset on every epoch


ibeltagy:
data_iterator unnecessarily rebuilds the dataset on every epoch, which significantly slows things down.

`DataIterator. __call__` calls `DataIterator._yield_one_epoch` for every epoch. Each call of `DataIterator._yield_one_epoch` shuffles the input data, maps it to tensors and chops it into batches. In the non-lazy mode however, most of this work can be done only once before the first epoch, not before every epoch. 

In my application, about 55% of the training runtime is spent in the `DataIterator._yield_one_epoch` function. 

Assuming non-lazy mode, and assuming we don't want to reshuffle before every epoch, fixing this should be easy (call `DataIterator._create_batches` outside `DataIterator._yield_one_epoch`).

matt-gardner:
There are two possible solutions here:

1. We use multi-processing to offload the data processing code done each batch away from the main thread, so it's not a bottleneck during training.  We'd like to do this soon, but it's not coming in the near future.
2. Write your own `DataIterator` that only `creates_batches` once, caches that result, and then just shuffles those at every epoch.  This should be pretty easy to do, and you could contribute it back as a PR.  It _might_ be reasonable to just add this as an option to our existing `DataIterators`, but I'd need to look more carefully at the code before I was confident about that.

ibeltagy:
I will give the second solution a try

matt-gardner:
This is fixed.

PR #1142, opened 2018-04-26
cached iterator

ibeltagy:
A solution for this issue: https://github.com/allenai/allennlp/issues/1128

I implemented it as a subclass of `BucketIterator` just to avoid messing up other stuff. However, it is better to be implemented as an additional option in `BasicIterator` where it is only enabled when `dataset_reader` is not lazy and `max_instances_in_memory is None`.

### DIFF
--- a/allennlp/data/iterators/cached_iterator.py
+++ b/allennlp/data/iterators/cached_iterator.py
@@ -19,10 +19,6 @@ class CachedIterator(BucketIterator):
     Using this iterator saves run time but costs more memory to store the cached batches. It should
     be used only if the dataset can fit in memory. This iterator shouldn't be used with the lazy
     mode.
-
-    Parameters:
-    ----------
-        See :class:`BucketIterator`.
     """
 
     def __init__(self, *args, **kwargs):
--- a/allennlp/data/iterators/cached_iterator.py
+++ b/allennlp/data/iterators/cached_iterator.py
@@ -1,0 +1,45 @@
+import logging
+import random
+from typing import Iterable, Dict, List
+from overrides import overrides
+from allennlp.data.instance import Instance
+from allennlp.data.iterators.bucket_iterator import BucketIterator
+from allennlp.data.iterators.data_iterator import DataIterator
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+@DataIterator.register("cached")
+class CachedIterator(BucketIterator):
+    """
+    An extension of the `BucketIterator` that caches its batches between epochs. Batches are built
+    only once during the first epoch, then the same batches are used again in consecutive epochs.
+    Using this iterator saves run time but costs more memory to store the cached batches. It should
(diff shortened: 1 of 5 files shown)


allennlp/data/iterators/cached_iterator.py
```diff
@@ -0,0 +1,45 @@
...
+    An extension of the `BucketIterator` that caches its batches between epochs. Batches are built
+    only once during the first epoch, then the same batches are used again in consecutive epochs.
+    Using this iterator saves run time but costs more memory to store the cached batches. It should
+    be used only if the dataset can fit in memory. This iterator shouldn't be used with the lazy
+    mode.
+
+    Parameters:
+    ----------
+        See :class:`BucketIterator`.
+    """
```
joelgrus: I would implement a constructor that throws a `ConfigurationError` if you try to instantiate it with lazy=True

allennlp/data/iterators/cached_iterator.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @overrides
+    def _yield_one_epoch(self, instances: Iterable[Instance], shuffle: bool, cuda_device: int, for_training: bool):
+        if not hasattr(self, 'cached_batches'):
+            # instances id -> list[batches]
+            self.cached_batches: Dict[int, List] = {}  # pylint: disable=attribute-defined-outside-init
```
joelgrus: and then I would just define this in the constructor

ibeltagy: do you know how to define a constructor for a subclass that takes the same arguments of the parent class without explicitly listing them ? I don't want to copy-paste them from `BucketIterator.__init__`

DeNeutoy:
```
class SubClass( Super ):
   def __init__(self, *args, **kwargs):
       super(self).__init__(*args, **kwargs)

```

allennlp/data/iterators/cached_iterator.py
joelgrus: although actually, laziness is a function of the dataset, not of the iterator, so this doesn't make sense

ibeltagy:
@joelgrus, added the check for laziness, and yes, I agree, this fits in a class higher in the hierarchy. `BasicIterator` would be a better fit than `DataIterator` because that's where `max_instances_in_memory` was introduced. 
Will let you guys decide if you want to keep it in its own class or somewhere else.

ibeltagy:
This PR still needs more work to make sure batches are cached in main memory not gpu memory

ibeltagy:
This is working fine now. An example is here: http://beaker.allenai.org/ex/ex_gnghsslk14vz/tasks. Duration of the first epoch is 42min then drops to 28min in the subsequent ones.

schmmd:
@joelgrus what is the status of this PR now?  Is this something we would like to merge?  If not, why are we no longer interested in it?

joelgrus:
this functionality was added to the base `DataIterator` class, so this pr is no longer necessary

Issue #998, opened 2018-03-19
train --recover doesn't work with GPU + Adam optimizer


joelgrus:
crashes when it tries to resume training

```
2018-03-19 13:41:54,514 - INFO - allennlp.data.dataset_readers.reading_comprehension.triviaqa - sample this iteration True
Traceback (most recent call last):
  File "/home/joelg/miniconda3/envs/allennlp2/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/joelg/miniconda3/envs/allennlp2/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/joelg/allennlp/allennlp/run.py", line 18, in <module>
    main(prog="python -m allennlp.run")
  File "/home/joelg/allennlp/allennlp/commands/__init__.py", line 73, in main
    args.func(args)
  File "/home/joelg/allennlp/allennlp/commands/train.py", line 100, in train_model_from_args
    args.recover)
  File "/home/joelg/allennlp/allennlp/commands/train.py", line 130, in train_model_from_file
    return train_model(params, serialization_dir, file_friendly_logging, recover)
  File "/home/joelg/allennlp/allennlp/commands/train.py", line 303, in train_model
    metrics = trainer.train()
  File "/home/joelg/allennlp/allennlp/training/trainer.py", line 649, in train
    train_metrics = self._train_epoch(epoch)
  File "/home/joelg/allennlp/allennlp/training/trainer.py", line 458, in _train_epoch
    self._optimizer.step()
  File "/home/joelg/miniconda3/envs/allennlp2/lib/python3.6/site-packages/torch/optim/adam.py", line 69, in step
    exp_avg.mul_(beta1).add_(1 - beta1, grad)
TypeError: add_ received an invalid combination of arguments - got (float, torch.cuda.FloatTensor), but expected one of:
 * (float value)
 * (torch.FloatTensor other)
 * (torch.SparseFloatTensor other)
 * (float value, torch.FloatTensor other)
      didn't match because some of the arguments have invalid types: (float, torch.cuda.FloatTensor)
 * (float value, torch.SparseFloatTensor other)
      didn't match because some of the arguments have invalid types: (float, torch.cuda.FloatTensor)
```

matt-gardner:
[...]

Issue #1102, opened 2018-04-19
type error when calling 'adagrad' optimization method


phiedulxp:
when recover from last training process , get the exception:

state['sum'].addcmul_(1, grad, grad)
TypeError: addcmul_ received an invalid combination of arguments - got (int, torch.cuda.FloatTensor, torch.cuda.FloatTensor), but expected one of:

(torch.FloatTensor tensor1, torch.FloatTensor tensor2)
(torch.SparseFloatTensor tensor1, torch.SparseFloatTensor tensor2)
(float value, torch.FloatTensor tensor1, torch.FloatTensor tensor2)
didn't match because some of the arguments have invalid types: (int, torch.cuda.FloatTensor, torch.cuda.FloatTensor)

DeNeutoy:
Thanks - this is the same issue as #998

matt-peters:
Fixed in #1144

PR #1144, opened 2018-04-26
Allow to restart with optimizers on CUDA

matt-peters:


matt-peters:
Fixes #998 #1102

DeNeutoy:
Great, thanks - could you add a todo to check if this is necessary with the 0.4 upgrade?

matt-peters:
Opened #1146  to track whether this is necessary after moving to 0.4.

PR #1151, opened 2018-04-26
Add a default cuda_device to model.

schmmd:


allennlp/models/model.py
```diff
@@ -109,19 +122,23 @@ def forward(self, input1, input2, targets=None):
         """
         raise NotImplementedError
 
-    def forward_on_instance(self, instance: Instance, cuda_device: int) -> Dict[str, numpy.ndarray]:
+    def forward_on_instance(self, instance: Instance, cuda_device: int = None) -> Dict[str, numpy.ndarray]:
```
schmmd: @matt-gardner do you think we should even have a `cuda_device` argument here?  I'm guessing it's potentially useful for running on multiple GPUs--but in some multi-GPU code @matt-peters pointed me to models seem to still have a single `cuda_device` (https://github.com/allenai/allennlp/blob/master/allennlp/training/trainer.py#L282) but I didn't go through it too thoroughly.

matt-gardner: It seems reasonable to me to just remove it, if the model can figure out on its own what device the data belongs on.  It's a single instance, and multi-GPU stuff is typically data parallel (i.e., you have several copies of the model on different GPUs, and you split your data such that `num_instances / num_gpus` instances are on each GPU), so I wouldn't expect multi-GPU to be an issue.  But, I don't know enough about how our multi-GPU support currently works / will work in the future, so @DeNeutoy might have more helpful input here.

DeNeutoy: This seems good to me. 

I don't think we actually need the `self.cuda_device` attribute - we can just get it from the pytorch model state by calling `allennlp.nn.util.get_device_of(next(self.parameters()))` in `forward_on_instance`. This additionally prevents needing to add extra state which could become corrupted in weird ways - eg if you called our overloaded `.cpu` with an incorrect argument, we would crash but still set the state to whatever the argument was, leading to a tricky bug. If you do want to do this though, the attribute should be private and it should be set in `model.load`/ `model._load`.

@matt-gardner - I don't think we need to worry about the multi-gpu stuff here - this is for prediction/evaluation, where it's pretty unlikely you're going to be using multiple gpus.

schmmd: I'll remove the `cuda_device` parameter from this method.

@DeNeutoy thanks for the pointer on `allennlp.nn.util.get_device_of(next(self.parameters()))`.  I'll take a look.

`cuda_device` is currently set in `model.load` / `model._load`, but I'll make it private if I fall back to it.

DeNeutoy: Ah right, from setting the device. This is another reason not to do it with the attribute, because when I read the code in `Model.load` and saw `.cpu` or `.gpu` I didn't expect for extra state to be set.

matt-gardner: Yeah, agreed that reading from the parameters is better than setting state on `model`, though will we ever be in a position that different parameters are on different devices?  Like, the embedding layer lives on the CPU, and everything else is on the GPU?

Also, pytorch 0.4 makes getting the device of a tensor, and shifting things to different devices, easier.

DeNeutoy: Yes, but what we had previously wouldn't really handle that anyway, I think - when we load a model, we don't have an option to move only certain parameters onto the GPU, so unless you've rolled your own already you wouldn't encounter that case.

schmmd:
@DeNeutoy assigning you.  I felt uncomfortable just checking the first parameter, so I check all of them and make sure they are consistent.  This works on a GPU workstation.  Before merging I'll try it in the cloud and also run a training job.

allennlp/models/model.py
```diff
@@ -51,6 +52,27 @@ def __init__(self,
         self.vocab = vocab
         self._regularizer = regularizer
 
+    def cuda_device(self) -> int:
```
DeNeutoy: This should be private - we don't want people to use this in general, as we are only guaranteed to have all the parameters on the same device at prediction time (for instance, if you are training a model on multiple GPUs, this would almost certainly break). You could call it something like `_get_device_for_prediction`

allennlp/models/model.py
```diff
@@ -51,6 +52,27 @@ def __init__(self,
...
+        """
+        This method checks the device of the first parameter to determine the cuda_device
```
DeNeutoy: This docstring doesn't match the implementation - it doesn't check a single parameter and it doesn't return None if there are no parameters.

allennlp/models/model.py
```diff
@@ -51,6 +52,27 @@ def __init__(self,
...
+        this model is to be run on.  If there are no parameters, it returns None.
+        Returns
+        -------
+        The cuda device this model is to be run on.
+        """
+        device = None
+        for parameter in self.parameters():
```
DeNeutoy: it might be a little clearer/pythonic to check a set of devices, and raise an error if the size of the set is > 1:

```
devices = {util.get_device_of(param) for param in self.parameters()}

if len(devices) != 1:
    raise ConfigurationError ...

else:
    # Only has one element, pop it off and return it.
    return devices.pop()
```

But how you have it is fine.

DeNeutoy: Also, it's possible but unlikely to have a model with no parameters - by default, this function should return -1, I think.

allennlp/models/model.py
schmmd: Thanks for explaining the motivation.

allennlp/models/model.py
schmmd: It's potentially a slower solution, but it's significantly cleaner so I'll make the change ;-)

DeNeutoy: Interested - why would it be slower?

allennlp/models/model.py
```diff
@@ -189,6 +190,25 @@ def get_metrics(self, reset: bool = False) -> Dict[str, float]:
         # pylint: disable=unused-argument,no-self-use
         return {}
 
+    def _get_prediction_device(self) -> int:
+        """
+        This method checks the device of the first parameter to determine the cuda_device
```
DeNeutoy: This docstring is still wrong - it checks that all of the parameters are on the same device, returning the device or raising an error if they are on different devices.

PR #1153, opened 2018-04-27
Use separate action embeddings for prediction vs. input feeding

matt-gardner:
This is similar to having untied word embeddings and softmax layer weights in a language model or MT system.  The PNP model does this, and adding it gave about a half a point gain (unless that was just training variance...).  I'm not too happy with how I implemented this, but I'm planning on cleaning up how the whole action embedding stuff works, anyway, and so I'd like to make this nicer then, and just get performance to match first.


allennlp/models/semantic_parsing/wikitables/wikitables_decoder_step.py
```diff
@@ -115,24 +115,27 @@ def take_step(self,
         considered_actions, actions_to_embed, actions_to_link = self._get_actions_to_consider(state)
 
         # action_embeddings: (group_size, num_embedded_actions, action_embedding_dim)
+        # output_action_embeddings: (group_size, num_embedded_actions, action_embedding_dim)
```
pdasigi: Shape of `action_biases` too?

tests/models/semantic_parsing/wikitables/wikitables_decoder_step_test.py
```diff
@@ -163,8 +167,10 @@ def test_get_actions_to_consider_returns_none_if_no_linked_actions(self):
     def test_get_action_embeddings(self):
         action_embeddings = Variable(torch.rand(5, 4))
         self.state.action_embeddings = action_embeddings
+        self.state.output_action_embeddings = action_embeddings
+        self.state.action_biases = Variable(torch.rand(5, 1))
         actions_to_embed = [[0, 4], [1], [2, 3, 4]]
-        embeddings, mask = WikiTablesDecoderStep._get_action_embeddings(self.state, actions_to_embed)
+        embeddings, _, _, mask = WikiTablesDecoderStep._get_action_embeddings(self.state, actions_to_embed)
```
pdasigi: You could test the output embeddings here too, right?

tests/models/semantic_parsing/wikitables/wikitables_semantic_parser_test.py
```diff
@@ -144,7 +144,7 @@ def test_embed_actions_works_with_batched_and_padded_input(self):
                     ('instance_action', False, None),
                     (rule1, True, rule1_tensor)]]
 
-        embedded_actions, action_indices = model._embed_actions(actions)
+        embedded_actions, _, _, action_indices = model._embed_actions(actions)
```
pdasigi: Same here.

tests/models/semantic_parsing/wikitables/wikitables_decoder_step_test.py
matt-gardner: This whole thing is going to be re-written soon, so I don't want to bother testing this.

PR #1154, opened 2018-04-27
add stacked bilm

DeNeutoy:
Adds a bidirectional LSTM with variational dropout.

@matt-gardner you should try this in the parser - it will probably get you another point.


allennlp/modules/stacked_bidirectional_lstm.py
```diff
@@ -0,0 +1,121 @@
+"""
+A stacked LSTM with LSTM layers which alternate between going forwards over
```
matt-gardner: You can probably just remove this comment - it's for a different file, anyway.

DeNeutoy:
@matt-gardner this is ready.

allennlp/modules/stacked_bidirectional_lstm.py
```diff
@@ -0,0 +1,116 @@
...
+import torch
+from torch.nn.utils.rnn import PackedSequence, pack_padded_sequence, pad_packed_sequence
+from allennlp.modules.augmented_lstm import AugmentedLstm
+from allennlp.common.checks import ConfigurationError
+
+
+class StackedBidirectionalLstm(torch.nn.Module):
+    """
+    A standard stacked Bidirectional LSTM where the LSTM layers
+    are concatenated between each layer.
```
matt-gardner: You should probably note what the differences are between this and the standard LSTM, and why you might want to use one or the other.

allennlp/modules/stacked_bidirectional_lstm.py
```diff
@@ -0,0 +1,116 @@
...
+    num_layers : int, required
+        The number of stacked Bidirectional LSTMs to use.
+    recurrent_dropout_probability: float, optional (default = 0.0)
+        The dropout probability to be used in a dropout scheme as stated in
+        `A Theoretically Grounded Application of Dropout in Recurrent Neural Networks
+        <https://arxiv.org/abs/1512.05287>`_ .
+
+    Returns
+    -------
+    output_accumulator : PackedSequence
```
matt-gardner: This doesn't match what you have in `forward`.  Do you even need this here at all?

Mrs-Hudson:
I was wondering why this module does not use CUDA?

PR #1155, opened 2018-04-27
Added two feature extractors that were in PNP and not in our version

matt-gardner:
As I expected, this did not really change performance.  The run using this PR was .2% lower than the run just previous; I'm guessing that's just training variance, and not that these two features actually hurt performance.


scripts/train_fixtures.py
```diff
@@ -50,15 +50,15 @@ def train_fixture_gpu(config_prefix: str) -> None:
         train_fixture_gpu("tests/fixtures/srl/")
```
pdasigi: I guess you didn't mean to check this in.

PR #1157, opened 2018-04-27
WIP / experimental: one possible way to refactor iterators

joelgrus:
at a high level, here are the changes:

* create `allennlp.data.iterators.utils`, and move shared code there (rather than using inheritance)
* put a lot more logic in `DataIterator.__call__` (and, in particular, have it handle caching and adding epoch numbers)
* get rid of the complex inheritance hierarchy, now we have `BucketIterator(DataIterator)` and so on

some things I don't like about this WIP:

* storing `_cursors` in a global variable in `iterators.utils`. 
* the way `AdaptiveIterator` handles the "fall back to BucketIterator in `biggest_batch_first` case"
* the logic around `take_instances` and `memory_sized_lists` is still too intricate (but possibly there's no way around that)

### DIFF
--- a/allennlp/data/iterators/basic_iterator.py
+++ b/allennlp/data/iterators/basic_iterator.py
@@ -1,4 +1,4 @@
-from typing import Iterable, Dict, Iterator, Optional, List
+from typing import Iterable
 import logging
 import math
 import random
@@ -9,6 +9,7 @@ from allennlp.common import Params
 from allennlp.common.util import ensure_list, is_lazy, lazy_groups_of
 from allennlp.data.instance import Instance
 from allennlp.data.iterators.data_iterator import DataIterator
+from allennlp.data.iterators.utils import memory_sized_lists
 from allennlp.data.dataset import Batch
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
@@ -35,12 +36,13 @@ class BasicIterator(DataIterator):
     def __init__(self,
                  batch_size: int = 32,
                  instances_per_epoch: int = None,
-                 max_instances_in_memory: int = None) -> None:
+                 max_instances_in_memory: int = None,
+                 cache_instances: bool = False,
+                 track_epoch: bool = False) -> None:
+        super().__init__(cache_instances=cache_instances, track_epoch=track_epoch)
         self._batch_size = batch_size
         self._instances_per_epoch = instances_per_epoch
         self._max_instances_in_memory = max_instances_in_memory
-
-        self._cursors: Dict[int, Iterator[Instance]] = {}
 
     @overrides
     def get_num_batches(self, instances: Iterable[Instance]) -> int:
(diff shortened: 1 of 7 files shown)


tests/data/iterators/basic_iterator_test.py
```diff
@@ -5,11 +5,21 @@
 from allennlp.common import Params
 from allennlp.common.testing import AllenNlpTestCase
 from allennlp.data import Instance, Token, Vocabulary
+from allennlp.data.dataset import Batch
 from allennlp.data.dataset_readers.dataset_reader import _LazyInstances
 from allennlp.data.fields import TextField
 from allennlp.data.iterators import BasicIterator
 from allennlp.data.token_indexers import SingleIdTokenIndexer
 
+def assert_tensor_dict_eq(td1, td2) -> bool:
```
murphp15: This doesn't seem to be called from anywhere?

joelgrus:
ok, this is ready for review. like I said above, it might be a bad idea, but the underlying philosophy is to move a lot of optional functionality into the base class, so you don't have to implement specific `EpochTracking_` or `Caching_` variants

allennlp/data/iterators/data_iterator.py
```diff
@@ -1,33 +1,92 @@
...
+    return batch
+
 class DataIterator(Registrable):
     """
-    An abstract ``DataIterator`` class. ``DataIterators`` must implement __call__, which yields
-    batched examples.
+    An abstract ``DataIterator`` class. ``DataIterators`` must override ``_create_batches()``.
+
+    Parameters
+    ----------
```
DeNeutoy: Backticks around types.

allennlp/data/iterators/bucket_iterator.py
```diff
@@ -131,10 +139,15 @@ def from_params(cls, params: Params) -> 'BucketIterator':
         batch_size = params.pop_int('batch_size', 32)
         instances_per_epoch = params.pop_int('instances_per_epoch', None)
         max_instances_in_memory = params.pop_int('max_instances_in_memory', None)
+        cache_instances = params.pop_bool('cache_instances', False)
+        track_epoch = params.pop_bool('trach_epoch', False)
```
DeNeutoy: \s\trach\track

allennlp/data/instance.py
```diff
@@ -22,7 +22,7 @@ class Instance:
     fields : ``Dict[str, Field]``
         The ``Field`` objects that will be used to produce data arrays for this instance.
     """
-    def __init__(self, fields: Mapping[str, Field]) -> None:
+    def __init__(self, fields: MutableMapping[str, Field]) -> None:
```
matt-gardner: What does using `MutableMapping` give you over just using `Dict`?  Seems like it adds some complexity that we don't really need?

allennlp/data/iterators/data_iterator.py
```diff
@@ -1,33 +1,92 @@
...
+
+
+def add_epoch_number(batch: Batch, epoch: int) -> Batch:
+    """
+    Add the epoch number to the batch instances as a MetadataField.
+    """
+    for instance in batch.instances:
+        instance.fields['epoch_num'] = MetadataField(epoch)
+    return batch
+
```
matt-gardner: Nit: one more blank line here.

allennlp/data/iterators/bucket_iterator.py
```diff
@@ -1,22 +1,48 @@
...
+from allennlp.data.vocabulary import Vocabulary
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
+def sort_by_padding(instances: List[Instance],
+                    sorting_keys: List[Tuple[str, str]],  # pylint: disable=invalid-sequence-index
+                    vocab: Vocabulary,
+                    padding_noise: float = 0.0) -> List[Instance]:
+    """
+    Sorts the ``Instances`` in this ``Batch`` by their padding lengths, using the keys in
```
matt-gardner: Nit: this isn't really a ``Batch``; do we care?

allennlp/data/iterators/basic_iterator.py
```diff
@@ -130,14 +26,20 @@ def _create_batches(self, instances: Iterable[Instance], shuffle: bool) -> Itera
...
+                batch = Batch(batch_instances)
+                yield batch
 
     @classmethod
     def from_params(cls, params: Params) -> 'BasicIterator':
         batch_size = params.pop_int('batch_size', 32)
         instances_per_epoch = params.pop_int('instances_per_epoch', None)
         max_instances_in_memory = params.pop_int('max_instances_in_memory', None)
+        cache_instances = params.pop_bool('cache_instances', False)
+        track_epoch = params.pop_bool('trach_epoch', False)
```
matt-gardner: `trach` issue here too.

joelgrus:
I can see the appeal of doing that, but here are the two reasons I'd rather not:

(1) it would be a very breaking change; for example, it would mean that every config file everywhere with an `iterator.type` key is suddenly invalid
(2) it makes it harder for people to implement their own iterators, for example, if we wanted to make another attempt at the adaptive iterator or whatever. maybe no one will ever do this, but I feel like we should at least leave them the option

matt-gardner:
Ah, yes, leaving open the option for fancier iterators is really compelling (the `"type"` key could be handled in code with a pop and a deprecation warning).

allennlp/data/instance.py
joelgrus: https://github.com/allenai/allennlp/pull/725/files#r164889705 😇

joelgrus: and I had to change it to mutable because the "add epoch" functionality creates a new field

matt-gardner: Ah, thanks for the reminder =).

PR #1162, opened 2018-04-30
Mattp/ner

DeNeutoy:


allennlp/modules/conditional_random_field.py
```diff
@@ -79,36 +105,46 @@ class ConditionalRandomField(torch.nn.Module):
...
+                 constraints: List[Tuple[int, int]] = None,
+                 include_start_end_transitions: bool = True) -> None:
         super().__init__()
         self.num_tags = num_tags
 
         # transitions[i, j] is the logit for transitioning from state i to state j.
         self.transitions = torch.nn.Parameter(torch.Tensor(num_tags, num_tags))
 
         # _constraint_mask indicates valid transitions (based on supplied constraints).
+        # Include special start of sequence (num_tags + 1) and end of sequence tags (num_tags + 2)
```
joelgrus: I feel like this code would maybe be cleaner if you left `constraint_mask` (num_tags, num_tags) and added separate `start_contraints` and `end_constraints`? then you wouldn't have to do all the slicing when you applied it.

but I don't feel super strongly about it

joelgrus: also, wouldn't it be cleaner if in the the "no start transitions" case you just had 

```
self.start_transitions = torch.zeros(num_tags)
```

and then I think the math would just work out without conditionals?

DeNeutoy: But how do you know which constraints those are? We can't build them ourselves,  because the CRF doesn't know which tagging scheme it uses.

joelgrus: you'd have to check for them when iterating over the constraints, but you're right that's ugly

DeNeutoy: Can you review again? My bias is against the change you suggested because 1) I didn't actually implement this so it will take me longer and 2) I'm not convinced the end result will be cleaner

Issue #1169, opened 2018-05-02
Order dependence in ElmoEmbedder?


ngoodman:
Apologies for opening an issue with what is likely a conceptual misunderstanding on my part!

I'm playing around with the pre-trained elmo embeddings (which are cool, thanks!) and noticing that the embedder seems to be stateful. That is, if i embed the same sentence twice, i don't get the same result:

~~~
import numpy as np
import scipy as sp
from allennlp.commands.elmo import ElmoEmbedder

ee = ElmoEmbedder()

v1 = np.squeeze(ee.embed_sentence("hello my name is mud .".split())[2,0,:])
v2 = np.squeeze(ee.embed_sentence("hello my name is mud .".split())[2,0,:])

print("embed twice test: ", sp.spatial.distance.cosine(v1,v2))
~~~

This gives a cosine dist of about 0.02 -- not huge, but problematic for the same sentence!

Where does the statefulness come from? Am i mis-using the embedder?

schmmd:
Hi @ngoodman that's expected behavior.  ELMo has internal state and adapts to your domain over time.  We've been thinking about how to make the output more consistent--as this is unexpected behavior for our users.

ngoodman:
Oh, i see! So this is giving me the embedding in the context of the "corpus" of sentences i've asked it to embed so far?

Testing my understanding i tried the above test using two separate instances of ElmoEmbedder, and indeed got the same embedding. This seems to be an infeasible approach in practice for lots of sentences, because it takes a long time to make the ElmoEmbedder()... Any workaround?

It's certainly true that this was unexpected for me, but a few small changes would have clued me in. E.g. if the call had been `ee. embed_next_sentence(..)` and/or there were a note in the docs. 

At any rate, thanks for the super fast response, and the nice open software!

matt-peters:
@ngoodman - I added a longer description of the statefulness to this PR: #1167 

The TLDR; is that the stateful aspect is a consequence of how the biLM was originally trained. 
 Except for the first few batch, the predictions won't vary that much from batch to batch assuming you are using the same `ElmoEmbedder` instance.  The recommended usage is to load only one `ElmoEmbedder` instance (it holds the internal states) and send all batches through it.  If you are concerned about the non-determinism then run a batch or two though it to "warmup the states", then start making predictions with your data

For example, modifying your code to run the same sentence multiple times, the embeddings are constant after the first batch:

```
distances = []
[...]

PR #1170, opened 2018-05-02
Add a comment about non-determinism

schmmd:
Partially addresses https://github.com/allenai/allennlp/issues/1169.


matt-peters:
Thanks -- this is a common question so I added some more detail on statefulness to the ELMo tutorial in the open PR: #1167 and incorporated these changes there.

PR #1184, opened 2018-05-08
Speedup ELMo command by making input tensors volatile

nelson-liu:
Maybe I'm missing something here, but it seems like setting `for_training` to `False` when using `allennlp elmo` should not affect the output results while also producing some speedups + memory savings.

### DIFF
--- a/allennlp/commands/elmo.py
+++ b/allennlp/commands/elmo.py
@@ -143,7 +143,7 @@ class ElmoEmbedder():
             A tuple of tensors, the first representing activations (batch_size, 3, num_timesteps, 1024) and
         the second a mask (batch_size, num_timesteps).
         """
-        character_ids = batch_to_ids(batch)
+        character_ids = batch_to_ids(batch, for_training=False)
         if self.cuda_device >= 0:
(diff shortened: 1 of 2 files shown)


joelgrus:
the whole `for_training` / `volatile` is going away in pytorch 0.4, so I'd maybe hold off on this?

joelgrus:
or rather, make the change using `with torch.no_grad()` in the `pytorch-0.4` branch?

joelgrus:
in fact, I'm hacking on that part of the code right now, I can just do it

nelson-liu:
Thanks @joelgrus ! I'll close this.

matt-peters:
Just curious, did this give a noticeable speedup?

nelson-liu:
It gave me a fairly decent speedup with the default batch size (60it/s to 70 it/s). I'm running ELMo on a fairly large file, so it's pretty noticeable to me :)

matt-peters:
Good to know - you can probably squeak out some more throughput by increasing the batch size until the GPU goes OOM.

PR #1181, opened 2018-05-07
Added coverage to WikiTables ERM parser

pdasigi:
Summary of changes:
1) I defined some simple rules for agenda mapping in wikitables
2) added checklists and agendas to the decoder state
3) included checklist cost in cost computation in the learning to search model
4) updated output action logits computation

The modification related to the output action logits turned out to be challenging. In nlvr, it was simply a matter of computing checklist balances and concatenating them to hidden states of the decoder and the attended encoder outputs before projecting them to a predicted output embedding space. However, in wikitables, since some of the actions are linked, I had to split the checklist balances into two, one for embedded action and the other for linked actions, and use the first for predicting output embedding and the second for modifying entity action logits.
Depends on #1159

PR #1187, opened 2018-05-09
Prune finished states in ERM

pdasigi:
Depends on #1181

### DIFF
--- a/allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
+++ b/allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
@@ -32,16 +32,22 @@ class ExpectedRiskMinimization(DecoderTrainer[Callable[[StateType], torch.Tensor
         The maximum number of steps we should take during decoding.
     max_num_decoded_sequences : ``int``, optional (default=1)
         Maximum number of sorted decoded sequences to return. Defaults to 1.
+    max_num_finished_states : ``int``, optional (default = None)
+        Maximum number of finished states to keep after search. This is to finished states as
+        ``beam_size`` is to unfinished ones. Costs are computed for only these number of states per
+        instance. If not set, we will keep all the finished states.
     """
     def __init__(self,
                  beam_size: int,
                  normalize_by_length: bool,
                  max_decoding_steps: int,
-                 max_num_decoded_sequences: int = 1) -> None:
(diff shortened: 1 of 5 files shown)


pdasigi:
@matt-gardner This is now ready for review. I found that using all the finished states worked better than using `beam_size` number of states. However, for wikitables, using all finished states makes training very slow, and using `beam_size` number of states, results in logical forms that are very short (usually just a cell name or a number). So the choice of the newly introduced hyper parameter seems important for wikitables.

allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
```diff
@@ -82,10 +88,39 @@ def _get_finished_states(self,
                 else:
                     next_states.append(next_state)
 
-            states = self._prune_beam(next_states)
+            states = self._prune_beam(next_states, self._beam_size, False)
```
matt-gardner: Minor readability nit: I find it easier to read when magic arguments are passed by name, so that I don't have to guess (or find the method definition) to know what `False` means here.  So, preference for `sort_states=False`.  Same with the `_prune_beam` invocation a couple of lines below.

allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
```diff
@@ -82,10 +88,39 @@ def _get_finished_states(self,
...
             num_steps += 1
+        if self._max_num_finished_states is not None:
+            finished_states = self._prune_beam(finished_states, self._max_num_finished_states, True)
         return finished_states
 
+    @staticmethod
```
matt-gardner: This doesn't need to happen now, but this method could probably just be put into `nn.decoding.util`, as lots of places want to prune a grouped beam.

allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
pdasigi: Done.

allennlp/nn/decoding/decoder_trainers/expected_risk_minimization.py
pdasigi: That's a good idea. Added a TODO for now.

PR #1191, opened 2018-05-10
(mostly) remove from_params

joelgrus:
at Google you'd have to pre-clear a PR like this with the "large changes committee", but we don't have that here.

at a high level:

* create new abstract class `FromParams` that has one classmethod `from_params` that inspects the class's constructor and does the "obvious" thing. this works for most of our classes, but whenever a `from_params` has special logic you have to "override" with your own version. (which causes mypy issues, since the type of the overriden version is more specific). I separated this from `Registrable` so that non-registrable classes can still get automatic `from_params` methods if they want (there is only a couple of instances of this in our codebase)

* then I made `Registrable` a subclass of `FromParams` so that every registrable class gets this too

* I added very detailed logging to `FromParams.from_params` to make debugging easier

* the logic here is non-trivial and involves some voodoo, so people might not like it for that reason

I still need to write a few more tests before I ask for code review, and I'll definitely want MattG's review and he's out this week, so no hurry


matt-gardner:
A random thought: it'd be super nice if this mechanism made it so we could put parameters on a super class and not have to duplicate them in subclasses.  Things like the `Regularizers` could just be specified on `Model`, and if the subclass takes `**kwargs`, you also check the super class for the parameters.  Seems a little bit magical, though...

janLuke:
What's the state of this? I like the idea. Can I be of any help? (maybe not _now_ but soon)

joelgrus:
* this PR is pretty stale (as you can see by all the conflicting files), so it needs to be synced with master
* the new functions need to be documented
* a bunch of commented out code needs to be cleaned up
* all the docs everywhere need to be updated to account for this change

but really the bigger issues are

* is this a good idea? and
* is this the right solution?

which I don't think I've even convinced myself of, and I'm the one who made the PR 😀

matt-gardner:
For "is this a good idea", I'd say the answer is definitely yes.  `from_params` is almost always completely redundant with the class constructor, and it makes it very easy to introduce bugs where you add a parameter in one spot and forget to add it in another spot.  It adds boilerplate and cognitive load to using allennlp.  If there's a good way to remove it in almost all cases, I think we should definitely do it.

For "is this the right solution", that's a harder question.  I haven't looked too closely at your approach, but I don't think there's a way to do this without reflection, and when I looked briefly at this a while ago it seemed reasonable to me.

janLuke:
I agree with @matt-gardner. @joelgrus, can you please explain why the current solution doesn't convince you (when you have time)?

joelgrus:
let me take a stab at bringing this PR up to date and then I'll see how I feel 😄

joelgrus:
ok, this is ready for review, no hurry because I need people who are on vacation this week to look at it

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+
+                # In some cases we allow a string instead of a param dict, so
+                # we need to handle that case separately.
+                if isinstance(subparams, str):
+                    kwargs[name] = annotation.by_name(subparams)()
+                else:
+                    kwargs[name] = annotation.from_params(params=subparams, **subextras)
+            elif not optional:
+                # Not optional and not supplied, that's an error!
+                raise ConfigurationError(f"expected key {name} for {cls.__name__}")
```
matt-gardner: Don't you need an `else` in here to use the default value?

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+        if name in extras:
+            kwargs[name] = extras[name]
+
+        # The next case is when the parameter type is itself constructible from_params.
+        elif hasattr(annotation, 'from_params'):
+            if name in params:
+                # Our params have an entry for this, so we use that.
+                subparams = params.pop(name)
+
+                if takes_arg(annotation.from_params, 'extras'):
```
matt-gardner: I don't understand this one.  When do we want to pass on extras?  I'm pretty sure this is the mechanism you're using for `Vocab` and such, but I'm not sure when you want to pass that on...  Oh, do we sometimes pass the `Vocab` down?  Is that what this is about?

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+                    # If annotation.params accepts **extras, we need to pass them all along.
+                    subextras = extras
+                else:
+                    # Otherwise, only supply the ones that are actual args; any additional ones
+                    # will cause a TypeError.
+                    subextras = {k: v for k, v in extras.items() if takes_arg(annotation.from_params, k)}
```
matt-gardner: And when should we not crash in this case?  At first glance it feels like we should just be crashing here if we get extra parameters, but I assume you have a good reason for not doing that, and I'm just missing it.  (I may have figured this out; is it because of the vocab passing?)  It still makes me uncomfortable that we can pass `random='garbage'` to these methods without it crashing somewhere.  Will that happen?  Is there a way to make it actually crash?

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+
+                # In some cases we allow a string instead of a param dict, so
```
matt-gardner: This is initializers and regularizers?  Anything else?  I guess it now lets us simplify the config file if you only ever had `{"type": "name"}` for a particular key, because having this here makes it no longer "in some cases"...

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+                    kwargs[name] = annotation.by_name(subparams)()
+                else:
+                    kwargs[name] = annotation.from_params(params=subparams, **subextras)
+            elif not optional:
+                # Not optional and not supplied, that's an error!
+                raise ConfigurationError(f"expected key {name} for {cls.__name__}")
+
+        # If the parameter type is a Python primitive, just pop it off
+        # using the correct casting pop_xyz operation.
+        elif annotation == str:
```
matt-gardner: I'm feeling like it'd be nice to handle `List[int]`, or `Union[int, List[int]]`, or even `Union[str, int, List[int]]` here (thinking of another recent PR).  But, I also think that's totally fine to leave for another PR, when someone actually wants to use such a type annotation.  Might be nice, though, to leave a note here somewhere that if you want to handle those things, this is the place to do it.

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+                            else params.pop_float(name))
+
+        # This handles types like Dict[str, TokenIndexer], where we need to instantiate
+        # each value from_params and return the resulting dict.
+        elif origin == Dict and len(args) == 2 and hasattr(args[-1], 'from_params'):
+            value_cls = annotation.__args__[-1]
+
+            value_dict = {}
+
+            for key, value_params in params.pop(name, Params({})).items():
```
matt-gardner: We had a `dict_from_params` method on `TokenIndexer`, because of how default dictionary values were handled in constructors.  Does this change the default behavior at all?  It probably doesn't, just making sure you've thought about it.

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+
+        else:
+            # Pass it on as is and hope for the best.   ¯\_(ツ)_/¯
+            if optional:
+                kwargs[name] = params.pop(name, default)
+            else:
+                kwargs[name] = params.pop(name)
+
+    params.assert_empty(cls.__name__)
+    return {k: v for k, v in kwargs.items() if takes_arg(cls, k)}
```
matt-gardner: Should we really have to check for `takes_arg()` here?  When should we not crash if we get an extra one?  I _think_ the extra `Vocab` stuff has already been handled by this point, but I might be missing something.

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+
```
matt-gardner: Nit: two blank lines, not three.

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+            # pylint: disable=no-member
+            as_registrable = cast(Type[Registrable], cls)
+            default_to_first_choice = as_registrable.default_implementation is not None
+            choice = params.pop_choice("type",
+                                       choices=as_registrable.list_available(),
+                                       default_to_first_choice=default_to_first_choice)
+            subclass = registered_subclasses[choice]
+
+            # Prune down the dict of extras, if necessary.
+            if not takes_arg(subclass.from_params, 'extras'):
```
matt-gardner: Again, an example of when you need this here would be nice.  Do we have `Registrable` classes where some subclasses take extra args that others don't?  That seems like it breaks the old `from_params` API, so I'm not sure where it ever happened.

allennlp/common/registrable.py
```diff
@@ -52,13 +63,18 @@ def add_subclass_to_registry(subclass: Type[T]):
...
         if name not in Registrable._registry[cls]:
             raise ConfigurationError("%s is not a registered name for %s" % (name, cls.__name__))
         return Registrable._registry[cls].get(name)
 
     @classmethod
     def list_available(cls) -> List[str]:
         """List default first if it exists"""
+        # This is necessary because some of the subclasses may not have been loaded
+        # at this point, and they're not registered until they're loaded.
+        _load_module(cls)
```
matt-gardner: Why didn't we need this before, if we need it now?  (Not complaining, it's just not obvious to me, and I'm curious.)

allennlp/models/biattentive_classification_network.py
```diff
@@ -297,10 +297,12 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
     def get_metrics(self, reset: bool = False) -> Dict[str, float]:
         return {metric_name: metric.get_metric(reset) for metric_name, metric in self.metrics.items()}
 
+    # The logic here requires a custom from_params.
```
matt-gardner: This is the `FeedForward` vs. `Maxout` logic, right?  I'd note that explicitly.  It'd be nice to find a way to fix it, but that can be another PR someday if someone wants to do it.

allennlp/modules/feedforward.py
```diff
@@ -79,6 +80,8 @@ def forward(self, inputs: torch.Tensor) -> torch.Tensor:
             output = dropout(activation(layer(output)))
         return output
 
+    # Requires custom logic around the activations (mostly because the Union type is too complex
+    # to handle automatically).
```
matt-gardner: Ah, yes, we could handle the `List`, I think, as I mentioned above.  But, yeah, that can wait, as it's not very frequent, and not a big deal.  You might modify the wording here, though - it's not too complex to handle automatically, it's just complex enough that we haven't bothered to do it yet.

allennlp/modules/span_extractors/bidirectional_endpoint_span_extractor.py
```diff
@@ -46,7 +45,7 @@ class BidirectionalEndpointSpanExtractor(SpanExtractor):
         The method used to combine the ``forward_start_embeddings`` and ``forward_end_embeddings``
         for the forward direction of the bidirectional representation.
         See above for a full description.
-    backward_combination : str, optional (default = "y-x").
+    backward_combination : str, optional (default = "x-y").
```
matt-gardner: @DeNeutoy, just a note to look at this, if you haven't already.

allennlp/modules/text_field_embedders/basic_text_field_embedder.py
```diff
@@ -81,8 +81,10 @@ def forward(self, text_field_input: Dict[str, torch.Tensor], num_wrapping_dims:
             embedded_representations.append(token_vectors)
         return torch.cat(embedded_representations, dim=-1)
 
+    # This is some unusual logic, it needs a custom from_params.
```
matt-gardner: Is it the logic, or just the `Dict[str, List[str]]`?  If we handled that `Dict` type, it looks like this would be standard, right?

allennlp/tests/common/from_params_test.py
```diff
@@ -0,0 +1,74 @@
...
+        assert remove_optional(str) == str
+
+    def test_from_params(self):
+        my_class = MyClass.from_params(Params({"my_int": 10}), my_bool=True)
+
+        assert isinstance(my_class, MyClass)
+        assert my_class.my_int == 10
+        assert my_class.my_bool
+
+    def text_create_kwargs(self):
```
matt-gardner: I feel like there could/should be a bunch more tests here for all of the `extras` cases I mentioned above, using real examples to show that they're actually necessary, and that they work right.

allennlp/tests/models/biattentive_classification_network_test.py
```diff
@@ -39,7 +39,7 @@ def test_mismatching_dimensions_throws_configuration_error(self):
         # the embedding dimension from the text_field_embedder.
         params["model"]["encoder"]["input_size"] = 10
         with pytest.raises(ConfigurationError):
-            Model.from_params(self.vocab, params.get("model"))
+            Model.from_params(vocab=self.vocab, params=params.get("model"))
```
matt-gardner: Might be nice to update some documentation somewhere to say that when calling `from_params` you need to pass arguments by name now.

tutorials/getting_started/using_as_a_library_pt1.md
```diff
@@ -233,12 +225,7 @@ you need to do two things.  First, you need to _register_ your objects with our
 our code can find your class when it tries to instantiate a `DatasetReader`.  That is what the
 first line in the code above is doing; we register `SemanticScholarDatasetReader` as a
 `DatasetReader` with the name `s2_papers`, which will let us use that name in our JSON
-configuration file.  Then, when AllenNLP is trying to construct a `DatasetReader`, it will take the
-parameters you specify and pass them to the `from_params` method.  This method takes a `Params`
-object, which is just a JSON dictionary with some added functionality, and constructs the
-`SemanticScholarDatasetReader`.  All of the `DatasetReader`'s dependencies that we want to be able
-to configure from the JSON file need to be constructed here.  In this case, we just create a
-`Tokenizer` and a `TokenIndexer` dictionary, using methods that are built in to the library.
+configuration file.
```
matt-gardner: `from_params` is mentioned again below when discussing the config file.  Not sure what the best way to update that section is, though...

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+    # Get the signature of the constructor.
+    signature = inspect.signature(cls.__init__)
+    kwargs: Dict[str, Any] = {}
+
+    # Iterate over all the constructor parameters and their annotations.
+    for name, param in signature.parameters.items():
+        # Skip "self". You're not *required* to call the first parameter "self",
+        # so in theory this logic is fragile, but if you don't call the self parameter
+        # "self" you kind of deserve what happens.
+        if name == "self":
```
DeNeutoy: Seems like you can just skip the first parameter? The [parameters attribute is ordered](https://docs.python.org/3/library/inspect.html#inspect.Signature), so I think you can just skip the first one.

DeNeutoy: Although in principle I agree that if you don't call the first parameter self you deserve it

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+
+class FromParams:
+    """
+    Mixin to give a from_params method to classes. We create a distinct base class for this
+    because sometimes we want non-Registrable classes to be instantiatable from_params.
+    """
+    @classmethod
+    def from_params(cls: Type[T], params: Params, **extras) -> T:
```
DeNeutoy: This needs to support `*args`. I just spent about 10 minutes trying to figure out why I couldn't get this test to pass:
```
    def test_error_when_building_from_params(self):
        vocab = Vocabulary()
        DecomposableAttention.from_params(vocab, params)
```
results in:
allennlp.common.checks.ConfigurationError: 'expected key vocab for DecomposableAttention'

allennlp/models/constituency_parser.py
```diff
@@ -61,7 +60,7 @@ class SpanConstituencyParser(Model):
     encoder : ``Seq2SeqEncoder``, required.
         The encoder that we will use in between embedding tokens and
         generating span representations.
-    feedforward_layer : ``FeedForward``, required.
```
DeNeutoy: This will break existing models in the demo - you'll need to update it manually

allennlp/models/reading_comprehension/bidaf.py
```diff
@@ -41,7 +40,7 @@ class BidirectionalAttentionFlow(Model):
     phrase_layer : ``Seq2SeqEncoder``
         The encoder (with its own internal stacking) that we will use in between embedding tokens
         and doing the bidirectional attention.
-    attention_similarity_function : ``SimilarityFunction``
```
DeNeutoy: Same here, and with anywhere where you've changed the named arguments

allennlp/modules/feedforward.py
DeNeutoy: More specifically, it's because the `Union` isn't a `Union` of native types, right? I tried this with the `Maxout` module and it worked fine with lists of ints.

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+
+from allennlp.common.checks import ConfigurationError
+from allennlp.common.params import Params
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+T = TypeVar('T')
+
+# getattr is necessary because mypy insists that no such attribute exists.
+_NO_DEFAULT = getattr(inspect, '_empty')  # pylint: disable=invalid-name
```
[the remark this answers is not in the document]
joelgrus: it looks like `Signature.empty` (and also `Parameter.empty`) is the public alias for it, so I'll change it to that

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+    if inspect.isclass(obj):
+        signature = inspect.signature(obj.__init__)
+    elif inspect.ismethod(obj) or inspect.isfunction(obj):
+        signature = inspect.signature(obj)
+    else:
+        raise ConfigurationError(f"object {obj} is not callable")
+    return arg in signature.parameters
+
+def remove_optional(annotation):
```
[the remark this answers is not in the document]
joelgrus: I can annotate the parameter as `type`. if I annotate the return type it causes some difficulties later on when I want to use the `from_params` method

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+    Optional[X] annotations are actually represented as Union[X, NoneType].
+    For our purposes, the "Optional" part is not interesting, so here we
+    throw it away.
+    """
+    origin = getattr(annotation, '__origin__', None)
```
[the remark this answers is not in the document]
joelgrus: done

allennlp/common/from_params.py
```diff
@@ -0,0 +1,185 @@
...
+
+    # Iterate over all the constructor parameters and their annotations.
+    for name, param in signature.parameters.items():
+        # Skip "self". You're not *required* to call the first parameter "self",
+        # so in theory this logic is fragile, but if you don't call the self parameter
+        # "self" you kind of deserve what happens.
+        if name == "self":
+            continue
+
+        # Get the annotation and break it into `origin` and `args`
```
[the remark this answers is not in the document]
joelgrus: it's not so much that they're the only "reasonable" types, it's that you have to make a blanket assumption about how you're going to handle them, and we've only done that that for one specific scenario.

in this case, the assumption is that you instantiate a `Dict[str, SomeClass]` by instantiating each subclass from_params and returning the resulting dict. 

I'll tweak the comment

allennlp/common/from_params.py
joelgrus: yes

allennlp/common/from_params.py
joelgrus: yes, for example, `BasicTextFieldEmbedder.from_params` is custom and needs a Vocab, but the base `TextFieldEmbedder.from_params` doesn't

allennlp/common/from_params.py
joelgrus: it shouldn't

allennlp/common/from_params.py
joelgrus: I think you're right

allennlp/common/from_params.py
joelgrus: it's a little bit more subtle than that, I added some explanation

allennlp/common/registrable.py
joelgrus: I added it because classes weren't being registered, but it seems like it works without it now, so I'll take it out

allennlp/modules/feedforward.py
joelgrus: the from_params logic will as a last resort pass the values on as-is, so it works if it's a union of native types *and also* the instantiation logic doesn't vary depending on what you pass in

allennlp/modules/text_field_embedders/basic_text_field_embedder.py
joelgrus: no, because the problem here is that the keys are not the names of function arguments (which is what the automatic `from_params` assumes), they're the names of token_indexers. 

that is, although the constructor takes `token_embedders: Dict[str, TokenEmbedder]`, the params object doesn't have a `token_embedders` key, the params object is itself the dict

allennlp/models/constituency_parser.py
joelgrus: I'm pretty sure in every case I changed the constructor parameter to match what `from_params` had been expecting, so the models would keep working? (because they were all using from_params)

DeNeutoy: Oh yep, my bad.

allennlp/common/from_params.py
matt-gardner: I feel like that was a bug with `TextFieldEmbedder` - the API would break if the `from_params` method differed between the base class and subclasses.  Yes?

joelgrus: you could call it a bug, or you could call it a design error, but yes the text_field_embedder section of a config probably should have looked like

```
  "model": {
    "type": "bidaf",
    "text_field_embedder": {
      "token_embedders": {
          "tokens": { ... },
          "token_characters": { ... },
         }
      }
    },
```

instead of

```
  "model": {
    "type": "bidaf",
    "text_field_embedder": {
      "tokens": { ... },
      "token_characters": { ... },
      }
    },
```

but we can't really change it now without breaking all of the model configs

matt-gardner: Oh, I was talking about something else, I think.  You said the `BasicTextFieldEmbedder.from_params` needs `vocab`, but `TextFieldEmbedder.from_params` doesn't.  Actually, they both do, currently, so we didn't mess up the API.  I think I understand why passing along `extras` is necessary now, so my original comment here isn't a problem.

What you mention in your most recent comment is a bit of a pain, though, I agree.  Could we migrate to the cleaner version with a deprecation warning?  That is, accept both of them, and if we get a config file with the current format, emit a warning?  (Also not necessary for this PR.)

PR #1195, opened 2018-05-10
Update ELMo command help docstring

nelson-liu:
just ran `allennlp elmo --help` and updated the output.

### DIFF
--- a/allennlp/commands/elmo.py
+++ b/allennlp/commands/elmo.py
@@ -14,23 +14,28 @@ https://arxiv.org/abs/1802.05365
 .. code-block:: bash
 
    $ allennlp elmo --help
-   usage: allennlp [command] elmo [-h] [--vocab-path VOCAB_PATH]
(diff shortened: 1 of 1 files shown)


nelson-liu:
while i'm here: I'm interested in getting lowercased elmo vectors from `allennlp elmo`. is this part of `options_file`, or should I add a flag to the command?

nelson-liu:
I’m interested in reproducing the functionality of the following snippet with lowercase_tokens set to True

```
 "dataset_reader": {
   "type": "srl",
   "token_indexers": {
     "tokens": {
       "type": "single_id",
       "lowercase_tokens": true
     },
     "elmo": {
       "type": "elmo_characters"
     }
   }
 }
```

matt-peters:
It may be useful for lowercasing tokens with word vectors such as GloVe that were trained with lowercased tokens, but for ELMo it's not necessary which is why we haven't implemented it.

schmmd:
Thanks!

PR #1196, opened 2018-05-11
`allennlp train` should fail faster if trainer.cuda_device >= 0

murphp15:
`allennlp train` should fail faster if trainer.cuda_device >= 0.
When starting a train we now check if the cuda device is available.


allennlp/commands/train.py
```diff
@@ -250,6 +252,13 @@ def train_model(params: Params,
     create_serialization_dir(params, serialization_dir, recover)
     prepare_global_logging(serialization_dir, file_friendly_logging)
 
+    try:
+        with torch.cuda.device(params.params.get('trainer').get('cuda_device')):
+            print('training setup is available')
```
schmmd: I would just use `pass` here.

joelgrus:
I think we need to add this logic to all of our commands that are gpu-capable (e.g. predict, evaluate, etc...)

accordingly, I would make a `common.util` function that takes in a cuda_device and makes sure that that cuda device is available (and is a no-op if that device is -1) and raises a `ConfigurationError` if it isn't.

then I would just call this function from each of the commands.

murphp15:
@joelgrus @schmmd Is that what you are looking for?

matt-gardner:
@murphp15, just FYI, there are scripts that let you run our checks locally, so you don't have to push commits in order to see if you fixed the issues.  It's _much_ faster to just run them yourself.  The most important ones to run are pytest (no script needed, just run `pytest` however you prefer to invoke it), pylint (`./scripts/pylint.sh`), and mypy (`./scripts/mypy.sh`).  If you want to run the whole thing, you can just run `./scripts/verify.py` (this includes some checks that you're much less likely to affect, and it'll be slower).

murphp15:
@matt-gardner Sorry I'll start doing that :) 
I just too optimistic and always think this time I have fixed it!

murphp15:
@matt-gardner What is left outstanding on this? 
I'd love to get this merged before the weekend. :)

allennlp/common/checks.py
```diff
@@ -33,3 +36,16 @@ def check_dimensions_match(dimension_1: int,
     if dimension_1 != dimension_2:
         raise ConfigurationError(f"{dim_1_name} must match {dim_2_name}, but got {dimension_1} "
                                  f"and {dimension_2} instead")
+
+
+def cuda_is_valid(device_id: int):
```
joelgrus: I'm not a fan of this name, to me `cuda_is_valid` sounds like a function that returns a boolean. I would prefer it were called `check_for_gpu` or something that's more descriptive when you call it.

tests/commands/train_test.py
```diff
@@ -38,6 +41,38 @@ def test_train_model(self):
...
+                "trainer": {
+                        "num_epochs": 2,
+                        "cuda_device": torch.cuda.device_count(),
+                        "optimizer": "adam"
+                }
+        })
+
+        with self.assertRaises(ConfigurationError) as context:
+            train_model(params, serialization_dir=os.path.join(self.TEST_DIR, 'test_train_model'))
+        self.assertTrue('specified a GPU but none is available' in str(context.exception))
```
joelgrus: for consistency with the rest of the codebase (and because I think it's cleaner), could you change this test to something like

```
with pytest.raises(ConfigurationError, message="..."):
   ...
```

allennlp/common/checks.py
```diff
@@ -33,3 +36,16 @@ def check_dimensions_match(dimension_1: int,
...
+    if device_id is None:
+        # this branch is to ensure that existing tests which don't specify cuda pass.
+        return
+    try:
+        with cuda.device(device_id):
+            pass
+    except Exception:
+        raise ConfigurationError("Experiment specified a GPU but none is available;"
```
joelgrus: This `if / return / try / with / pass / except` construction feels way too complicated for what it does.

Why not something like just

```
if device_id is not None and device_id >= cuda.device_count():
    raise ConfigurationError(...)
```

Would that accomplish the same?

PR #1199, opened 2018-05-11
Convert run_with_beaker.py to use Blueprints.

schmmd:
A few breaking changes are required for this change.  In addition, instead of running the experiment an experiment spec is output.  This behavior can be changed if it's not desired, but I find it to be an improvement because it gives the user the opportunity to add additional flags that this script may not support.

### DIFF
--- a/scripts/ai2-internal/run_with_beaker.py
+++ b/scripts/ai2-internal/run_with_beaker.py
@@ -3,7 +3,9 @@
 # Script to launch AllenNLP Beaker jobs.
 
 import argparse
+import json
 import os
+import yaml
 import random
 import subprocess
 import sys
@@ -15,22 +17,20 @@ random_int = random.randint(0, 2**32)
 
 sys.path.insert(0, os.path.dirname(os.path.abspath(os.path.join(os.path.join(__file__, os.pardir), os.pardir))))
 
-from allennlp.commands.train import Train
 from allennlp.common.params import Params
 
-def main(param_file: str, extra_beaker_commands: List[str]):
-    ecr_repository = "896129387501.dkr.ecr.us-west-2.amazonaws.com"
+def main(param_file: str, output_path: str, args):
     commit = subprocess.check_output(["git", "rev-parse", "HEAD"], universal_newlines=True).strip()
-    image = f"{ecr_repository}/allennlp/allennlp:{commit}"
+    image = f"allennlp/allennlp:{commit}"
     overrides = ""
 
     # Reads params and sets environment.
     params = Params.from_file(param_file, overrides)
     flat_params = params.as_flat_dict()
-    env = []
+    env = {}
     for k, v in flat_params.items():
         k = str(k).replace('.', '_')
-        env.append(f"--env={k}={v}")
+        env[k] = str(v)
 
     # If the git repository is dirty, add a random hash.
     result = subprocess.run('git diff-index --quiet HEAD --', shell=True)
@@ -38,22 +38,18 @@ def main(param_file: str, extra_beaker_commands: List[str]):
         dirty_hash = "%x" % random_int
         image += "-" + dirty_hash
 
-    # Get temporary ecr login. For this command to work, you need the python awscli
-    # package with a version more recent than 1.11.91.
-    print("Generating ECR Login Command")
-    login_command = subprocess.check_output('aws --region=us-west-2 ecr get-login --no-include-email', shell=True)
+    if args.blueprint:
+        blueprint = args.blueprint
+        print(f"Using the specified blueprint: {blueprint}")
+    else:
+        print(f"Building the Docker image ({image})...")
+        subprocess.run(f'docker build -t {image} .', shell=True, check=True)
 
-    print("Logging into ECR")
-    subprocess.run(login_command, shell=True, check=True)
-
-    print(f"Building the Docker image ({image})")
-    subprocess.run(f'docker build -t {image} .', shell=True, check=True)
-
-    print(f"Pushing the Docker image ({image})")
-    subprocess.run(f'docker push {image}', shell=True, check=True)
+        print(f"Create a Beaker blueprint...")
+        blueprint = subprocess.check_output(f'beaker blueprint create --quiet {image}', shell=True, universal_newlines=True).strip()
+        print(f"  Blueprint created: {blueprint}")
 
     config_dataset_id = subprocess.check_output(f'beaker dataset create --quiet {param_file}', shell=True, universal_newlines=True).strip()
-    filename = os.path.basename(param_file)
 
     allennlp_command = [
             "python",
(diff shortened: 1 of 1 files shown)


schmmd:
@matt-gardner I can mostly leave the script unchanged--although I'm going to remove some commands since they are deprecated and reportedly don't work.

One nice thing about the intermediate file is it gives you full control to be able to tweak your experiment.  Without using the configuration, we need to keep adding arguments when we want additional functionality.  Beaker is deprecating the run command, so some things might be challenging to tune via run.

But if that's undesirable to the people who actually use this script, I can `cat` the configuration directly to `beaker`.

matt-gardner:
Sure, if there are arguments that are broken, by all means, great, let's fix them.  Adding arguments for additional functionality is rare and a simple PR, and making that PR is much more desirable than adding a step for every invocation of this script.  I submitted PRs just to shave a few seconds off of each invocation, if you remember ;).

ckarenz:
To clarify: Beaker is not deprecating the `experiment run` command.  We are transitioning the `--file` argument over to a new `experiment create` command.  Create is the full-featured command which allows for arbitrarily complex (or simple) spec files.  Run offers a simplified subset of Create by replacing the spec file with some short-hand arguments, and is purely for convenience.

scripts/ai2-internal/run_with_beaker.py
```diff
@@ -15,45 +17,39 @@
...
-    subprocess.run(login_command, shell=True, check=True)
-
-    print(f"Building the Docker image ({image})")
-    subprocess.run(f'docker build -t {image} .', shell=True, check=True)
+    if args.blueprint:
+        blueprint = args.blueprint
+        print(f"Using the specified blueprint: {blueprint}")
+    else:
+        print(f"Building the Docker image ({image})...")
+        subprocess.run(f'docker build -t {image} .', shell=True, check=True)
```
ckarenz: This pattern looks ideal. 👍

scripts/ai2-internal/run_with_beaker.py
```diff
@@ -66,56 +62,63 @@ def main(param_file: str, extra_beaker_commands: List[str]):
...
+    print(f"Beaker spec written to {output_path}.  Launch your job with the following command")
+    print(f"")
+    print(f"    beaker experiment create --file {output_path}")
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
 
     parser.add_argument('param_file', type=str, help='The model configuration file.')
-    parser.add_argument('--name', type=str, help='A name for the experiment.')
+    parser.add_argument('output_path', type=str, help='The destination to write the experiment spec.')
```
ckarenz: @matt-gardner makes a good point about convenience here.  I'd recommend a slightly different pattern here.  Drop `output_path` and instead echo the spec to STDOUT, so it can be piped into Beaker directly, like so:

```bash
./run_with_beaker.py <args...> | beaker experiment create --file=-
```

You can also switch on whether the spec is automatically run.  One possible mechanism would be to offer a `--dry-run` flag.  In this flag's absence, invoke Beaker directly. If `--dry-run` is present, then just echo the spec so the user can redirect to file and/or Beaker on their own.

Another option is to print the command you would invoke, and invoke this script with `exec $(run_with_beaker.py)`. The output might be in the following format:
```bash
cat << EOF | beaker experiment create --file=-
# Spec Goes Here...
EOF
```

Either of these options also implies retaining the `--name` flag.

schmmd: Funny, I just read this after I added a `--dry-run` flag ;-)

I actually personally think it's a good thing to have an extra step, but I'm happy to do what works best for other users (commit coming).

scripts/ai2-internal/run_with_beaker.py
```diff
@@ -66,56 +61,77 @@ def main(param_file: str, extra_beaker_commands: List[str]):
...
+        "args": allennlp_command,
+        "datasetMounts": dataset_mounts,
+        "requirements": {
+            "cpu": float(args.cpu),
+            "memory": args.memory,
+            "gpuCount": args.gpu_count if args.gpu_count else 1,
+        },
+        "env": env
+    }
+    config_task = {"spec": config_spec}
```
ckarenz: You may consider adding `"name": "<some value>"` here.  While optional, it would slightly simplify how this displays in the UI by hiding the task ID.

schmmd: Done.

schmmd:
The script now maintains the original IO, but it has an option argument for a blueprint (which skips creating a Docker image), an optional argument to output the spec to a specific location (rather than a temporary location), and a `--dry-run` flag.

Blueprints don't add much new functionality at this point, but we're ready for when they do and this cleans up and speeds up our script a bit by cutting ECR out of the loop.  I understand that ECR is supported by not required.

scripts/ai2-internal/run_with_beaker.py
```diff
@@ -66,56 +61,79 @@ def main(param_file: str, extra_beaker_commands: List[str]):
...
+    if args.name:
+        config_task["name"] = args.name
```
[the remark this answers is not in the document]
matt-gardner: Yes, and my guess is that he missed the fact that it was already given below.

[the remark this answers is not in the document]
matt-gardner: Ok, I looked at your beaker experiments, and the way you have it seems fine.  Beaker has "experiments" and "tasks", where an "experiment" may contain several different "tasks".  We only ever use a single "task" for our "experiments", so this distinction is a bit meaningless for us.  Giving both the task and the experiment the same name seems fine.

If you look at your two most recent experiments, [this one](http://beaker.allenai.org/ex/ex_erd8g9v8w9sj/tasks) has both the experiment and the task name set to "food", while [this one](http://beaker.allenai.org/ex/ex_r2zh4c5d5tdp/tasks) just has the experiment name set to "foo", with an empty task name, so the task shows up as an id.  I don't have a strong preference between the two, either way is fine.

scripts/ai2-internal/run_with_beaker.py
```diff
@@ -66,56 +61,73 @@ def main(param_file: str, extra_beaker_commands: List[str]):
...
     parser.add_argument('--memory', help='Memory to reserve for this experiment (e.g., 1GB)')
 
     args = parser.parse_args()
 
-    extra_beaker_commands = []
-    if args.desc:
-        extra_beaker_commands.append(f'--desc={args.desc}')
-    if args.name:
-        # Remove spaces from the name, because Beaker doesn't allow them.
-        extra_beaker_commands.append(f'--name={args.name.replace(" ", "-")}')
```
DeNeutoy: Please don't change this behaviour, it is convenient

schmmd: Apologies--the change was a mistake (I deleted the name argument and then re-added it).

PR #1201, opened 2018-05-12
Use README.md for PyPI long description

nelson-liu:
I noticed that AllenNLP doesn't have an entry for the `long_description` on PyPI (https://pypi.org/project/allennlp/). This PR makes it so the contents of the README are used --- the new version of PyPI supports Markdown, so it should all render properly exactly as we see it on github.

(Let me know if there is a reason behind `long_description` not being set before).


nelson-liu:
hmm, looks like this commit was left out of the new release. @schmmd would you happen to know if anything went wrong / if other commits might be missing from 0.4.3?

You can verify by downloading the 0.4.3 tarball and checking the `PKG-INFO` and `setup.py` --- this commit isn't reflected in the `setup.py`

schmmd:
@nelson-liu this was rejected from pypi when I submitted it.  I did a bad thing and hacked the setup.py (removing your one line addition) to get the submission to work.

The error I got was:

```
Uploading distributions to https://upload.pypi.org/legacy/
Uploading allennlp-0.4.3-py3-none-any.whl
HTTPError: 400 Client Error: The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information. for url: https://upload.pypi.org/legacy/  
```

I tried adding `long_description_content_type="text/markdown"` but I got the same error.  Mark N suggested that I might be uploading to the wrong "legacy" API--but I'm pretty sure I was uploading to the right place (see https://stackoverflow.com/questions/47584147/why-is-twine-1-9-1-still-uploading-to-legacy-pypi).

I'm curious--how did you notice this so quickly?  Any ideas on how to get pypi to accept the long description?

schmmd:
@nelson-liu I can test the setup.py by making submissions to pypitest.

nelson-liu:
hmm, you might have to update twine? I can't link to sections within the changelog, but it says here (https://twine.readthedocs.io/en/latest/changelog.html) that twine 1.11.0 should support markdown for description fields.

nelson-liu:
regardless, this is not a big deal at all :) maybe worth trying on pypitest and then just using it for the next release, whenever it is.

schmmd:
Perhaps--I'm on:

```
$ twine --version
twine version 1.9.1 (pkginfo: 1.4.1, requests: 2.18.4, setuptools: 36.4.0,
requests-toolbelt: 0.8.0, tqdm: 4.19.5)
```

PR #1202, opened 2018-05-12
Remove argparse from requirements

nelson-liu:
Since allennlp is python 3.x only, it doesn't really make sense to have argparse in the requirements --- in fact, having argparse in the requirements seems to break conda (since conda install argparse only works for python 2.x)

I suspect that this was kept from a copy-paste from `deep_qa` and just never changed, but let me know if there's a reason for keeping `argparse` around.


schmmd:
@nelson-liu don't we need `argparse` for imports such as https://github.com/allenai/allennlp/blob/master/scripts/ai2-internal/run_with_beaker.py#L5 and https://github.com/allenai/allennlp/blob/master/allennlp/commands/elmo.py#L54?  Since the continuous build passed I'm probably missing something since I would have expected a build failure without `argparse`.

joelgrus:
argparse is part of the standard library since Python 2.7, the pip installable version is only needed if you're using 2.6 or earlier

schmmd:
Great!

PR #1204, opened 2018-05-12
Changed how checklist balance is used in action prediction

pdasigi:
Summary of changes:
1) Made the output projection layer the same in both mml and erm parsers for both nlvr and wikitables parsers
2) The checklist balance in nlvr (and unlinked checklist balance in wikitables) is now used to select appropriate action embeddings, the sum of which is added to predicted action embedding.
3) There are now learned scalar parameters that regularize the addition to predicted embedding (in nlvr and wikitables) and action logits (in wikitables).

### DIFF
--- a/allennlp/models/semantic_parsing/nlvr/nlvr_decoder_step.py
+++ b/allennlp/models/semantic_parsing/nlvr/nlvr_decoder_step.py
@@ -26,7 +26,7 @@ class NlvrDecoderStep(DecoderStep[NlvrDecoderState]):
     attention_function : ``SimilarityFunction``
     dropout : ``float``
         Dropout to use on decoder outputs and before action prediction.
-    use_coverage : ``bool``
+    use_coverage : ``bool``, optional (default=False)
         Is this DecoderStep being used in a semantic parser trained using coverage? We need to know
         this to define a learned parameter for using checklist balances in action prediction.
     """
@@ -95,7 +95,7 @@ class NlvrDecoderStep(DecoderStep[NlvrDecoderState]):
         # (group_size, decoder_input_dim)
         decoder_input = self._input_projection_layer(torch.cat([attended_sentence,
                                                                 previous_action_embedding], -1))
-
+        decoder_input = torch.nn.functional.tanh(decoder_input)
         hidden_state, memory_cell = self._decoder_cell(decoder_input, (hidden_state, memory_cell))
 
         hidden_state = self._dropout(hidden_state)
@@ -138,7 +138,7 @@ class NlvrDecoderStep(DecoderStep[NlvrDecoderState]):
         action_query = torch.cat([hidden_state, attended_sentence], dim=-1)
         # (group_size, action_embedding_dim)
         predicted_action_embedding = self._output_projection_layer(action_query)
-        predicted_action_embedding = self._dropout(torch.nn.functional.relu(predicted_action_embedding))
+        predicted_action_embedding = self._dropout(torch.nn.functional.tanh(predicted_action_embedding))
         if state.checklist_state[0] is not None:
             embedding_addition = self._get_predicted_embedding_addition(state)
             predicted_action_embedding += self._checklist_embedding_multiplier * embedding_addition
--- a/allennlp/models/semantic_parsing/nlvr/nlvr_decoder_step.py
+++ b/allennlp/models/semantic_parsing/nlvr/nlvr_decoder_step.py
@@ -141,7 +141,8 @@ class NlvrDecoderStep(DecoderStep[NlvrDecoderState]):
         predicted_action_embedding = self._dropout(torch.nn.functional.tanh(predicted_action_embedding))
         if state.checklist_state[0] is not None:
             embedding_addition = self._get_predicted_embedding_addition(state)
-            predicted_action_embedding += self._checklist_embedding_multiplier * embedding_addition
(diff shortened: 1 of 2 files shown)


allennlp/models/semantic_parsing/nlvr/nlvr_decoder_step.py
```diff
@@ -23,18 +24,15 @@ class NlvrDecoderStep(DecoderStep[NlvrDecoderState]):
     encoder_output_dim : ``int``
     action_embedding_dim : ``int``
     attention_function : ``SimilarityFunction``
-    checklist_size : ``int``, optional
-        If we are training a parser to maximize coverage using a checklist, we need the size of the
-        checklist vector to define the output projection layer, which projects a concatenation of
-        the current hidden state, attended encoder input and the current checklist balance into the
-        action space. The size of the checklist balance vector is the same as the number of
-        terminals. This is not needed if we are training the parser using target action sequences.
+    use_coverage : ``bool``
```
matt-gardner: It'd be nice to mention this is optional here, and give the default value.

allennlp/models/semantic_parsing/nlvr/nlvr_decoder_step.py
```diff
@@ -132,10 +127,12 @@ def take_step(self,  # type: ignore
         # action_mask: (group_size, num_embedded_actions)
         action_embeddings, embedded_action_mask = self._get_action_embeddings(state,
                                                                               global_actions_to_embed)
-        action_query = self._get_action_query(state, hidden_state, attended_sentence)
+        action_query = torch.cat([hidden_state, attended_sentence], dim=-1)
         # (group_size, action_embedding_dim)
         predicted_action_embedding = self._output_projection_layer(action_query)
```
matt-gardner: Not related to this PR, but you probably want a non-linearity here.  See #1150, where I added this to the wikitables model.  There's another spot where you probably don't have one but probably want one, too.

matt-gardner: And, now that I think about it, because this is getting a dot product with action embeddings, I wonder if relu isn't the best non-linearity.  We're basically saying that any embedding dimension with a negative value is entirely ignored, and thus artificially constraining the space that's available to the dot product...  Maybe tanh would be better?

allennlp/models/semantic_parsing/nlvr/nlvr_decoder_step.py
```diff
@@ -132,10 +127,12 @@ def take_step(self,  # type: ignore
...
-
+        if state.checklist_state[0] is not None:
+            embedding_addition = self._get_predicted_embedding_addition(state)
+            predicted_action_embedding += self._checklist_embedding_multiplier * embedding_addition
```
matt-gardner: Don't ever use `+=` on a tensor.  It doesn't do what you expect.  Use `x = x + y` instead.

allennlp/models/semantic_parsing/wikitables/wikitables_decoder_step.py
```diff
@@ -128,28 +129,26 @@ def take_step(self,
...
+        predicted_action_embedding = self._dropout(projected_query)
+        linked_balance = None
+        if state.checklist_state[0] is not None:
+            linked_balance, unlinked_balance = self._get_checklist_balance(state,
+                                                                           self._unlinked_terminal_indices,
+                                                                           actions_to_link)
+            embedding_addition = self._get_predicted_embedding_addition(state,
+                                                                        self._unlinked_terminal_indices,
+                                                                        unlinked_balance)
+            predicted_action_embedding += self._unlinked_checklist_multiplier * embedding_addition
```
matt-gardner: Don't use `+=`.

allennlp/models/semantic_parsing/nlvr/nlvr_decoder_step.py
pdasigi: Did you find that adding relus worked better?

pdasigi: Adding a non-linearity at the decoder input as well, and made both of them tanh.

allennlp/models/semantic_parsing/nlvr/nlvr_decoder_step.py
pdasigi: Yes, I figured out that `+=` causes in-place updates, messing up the computation graph while `x = x+y` does not. I changed the other places I did in-place updates, but I guess I missed these two. Thanks, fixed them.

allennlp/models/semantic_parsing/wikitables/wikitables_decoder_step.py
pdasigi: Done.

pdasigi:
@matt-gardner On NLVR, the variant with this change is at least slightly better (about 0.5 pp) than the run without this change. The experiment is still running though.

allennlp/models/semantic_parsing/nlvr/nlvr_decoder_step.py
matt-gardner: I ran it with tanh instead of relu last night, and it didn't really change anything.  It looked like it was learning faster - epoch 1 performance was a bit higher - but in the end, final performance was within the normal variance that I've seen across runs.  So, it probably doesn't matter.

PR #1209, opened 2018-05-13
don't overwrite scripts directory, copy individual scripts

DeNeutoy:


Dockerfile
```diff
@@ -74,6 +73,7 @@ COPY setup.py setup.py
 # Caching models when building the image makes a dockerized server start up faster, but is slow for
 # running tests and things, so we skip it by default.
 ARG CACHE_MODELS=false
+COPY scripts/cache_models.py scripts/cache_models.py
```
murphp15: Why did you move where in the docker build the file is copied? 
Can it not take place on line 65?

matt-gardner: Oh, did line 65 erase the result of making `EVALB`?  🤦‍♂️   Sorry...

The only issue with doing it this way is that we won't have all of the scripts.  That's probably not an issue...?  I guess this seems fine for now, and if someone wants a particular script in the docker image that's not present, they can add it later.

@murphp15, the previous line copied an entire directory.  Turns out that caused some problems, because of the earlier lines.  Most anticipated uses of the docker container don't actually need all of the scripts in the directory, so to avoid clobbering the stuff built on line 57, we can just copy over a few scripts that we need instead of the whole thing.

matt-gardner: Actually, I'm now confused about how this is passing CI.  TeamCity runs `./scripts/mypy.sh` and `./scripts/pylint.sh`, but it sure doesn't look like those get copied to the docker image anymore.  What's going on here?  How did those actually run?

schmmd:
Does it work to just swap the order, rather than not copying the scripts?  I do use the scripts from a Docker image sometime.  I'm also confused how TeamCity is running tests on this image and I'm looking into it.

matt-gardner:
The reason that was pulled out in the first place was to make the docker build more efficient, with things that were less likely to change put earlier in the build, to make better use of docker's caching.  It'd be nice to keep it that way, if there's a way to do it that makes everyone happy.  Does docker have something like a `cp scripts/* scripts/` that doesn't just overwrite the whole directory?

schmmd:
I'm sure Docker has some syntax for this--I just don't know what it is.

I'm pretty sure we do need to copy the other scripts--at least the build scripts but I would prefer we bundle all of them.  I think this is only passing in TeamCity is due to an error in our configuration which I'll try to root cause today.

Dockerfile
schmmd: I found it: TeamCity mounts the VCS root in the Docker image and runs its operations in that directory.  I'd like to change this so it runs its commands from within the Docker image itself, which would require us to have the scripts present in the Docker image.

schmmd:
Looking into this a bit more, our tests don't run in our Docker image on master because we don't copy the `docs` directory--nor do I know whether we want to.  So I'm not sure we need `pylint` etc., but it would be good to keep those such as `preprocess_wikitables.py`, `write_srl_predictions_to_conll_format.py`, etc. so they can be used from the Docker image.

DeNeutoy:
Fixed by #1233

Issue #1200, opened 2018-05-11
Lazy imports in `allennlp/commands/__init__.py`


gangeli:
Would it be possible to not import all of the commands in the `__init__.py` for the commands package?
 An easy fix would be to move the following imports into the `main()` of `__init__.py`:

```python
    from allennlp.commands.elmo import Elmo
    from allennlp.commands.evaluate import Evaluate
    from allennlp.commands.fine_tune import FineTune
    from allennlp.commands.make_vocab import MakeVocab
    from allennlp.commands.predict import Predict
    from allennlp.commands.serve import Serve
    from allennlp.commands.train import Train
    from allennlp.service.predictors import DemoModel
    from allennlp.common.util import import_submodules
```

The apparent error is an import error on `gevent.wsgi` when importing Elmo:

```python
    from allennlp.commands.elmo import ElmoEmbedder
  File "/builds/eng/eloquent/python/lib/ElPy/lib/python3.6/site-packages/allennlp/commands/__init__.py", line 11, in <module>
    from allennlp.commands.serve import Serve
  File "/builds/eng/eloquent/python/lib/ElPy/lib/python3.6/site-packages/allennlp/commands/serve.py", line 22, in <module>
    from allennlp.service import server_flask as server
  File "/builds/eng/eloquent/python/lib/ElPy/lib/python3.6/site-packages/allennlp/service/server_flask.py", line 19, in <module>
    from gevent.wsgi import WSGIServer
ModuleNotFoundError: No module named 'gevent.wsgi'
```

There's almost certainly another underlying user error here (e.g., I don't have `libevent-dev` on my docker image), but nonetheless it could be nice to be able to import Elmo without having to also import a web server.

PR #1213, opened 2018-05-14
Add allennlp test-install command

### LINKED ISSUE #856: create an easy way for people to test that their pip install is working correctly and that they've got all the right dependencies and downloads

nelson-liu:
Fixes #856 by creating an `allennlp test-install` command that runs the unit test suite. 

I've tested it locally outside of my allennlp install directory and it seems to work fine for both options, but unsure how to write a unit test for this.

This command skips the sniff tests by default, which should address @DeNeutoy 's concern from #1208 ?


DeNeutoy:
Yeah, I don't think we can test the thing which runs the tests in the tests 😆 

Once we set up CI which runs the pip installation, this will be tested implicitly by that, so I think we're all good.

nelson-liu:
I realized that if the `test_requirements` (pytest, flaky, responses>=0.7, jupyter) are not installed, this will totally break all of the `allennlp` commands.

Is it worth adding these packages to the requirements installed by pip? Alternatively, we could fix https://github.com/allenai/allennlp/issues/1200 ...

matt-gardner:
The whole reason `requirements.txt` was separate from `test_requirements.txt` was because we weren't including the tests in the pip release.  If we're going to include tests in the pip release, do we need that distinction anymore?

nelson-liu:
perhaps we can move these essential testing packages (pytest, flaky, responses>=0.7, jupyter) to requirements.txt and the setup.py `install_requires`, then rename `requirements_test.txt` to `requirements_dev.txt`? 

no need for a user to install sphinx or numpydoc or coverage...

matt-gardner:
I think others have stronger / more informed opinions here than me, so I'll let someone else answer the question.  I was just providing some context.

schmmd:
Jupyter is enormous and is currently in our `test_requirements`.  I'd rather not have that in our `requirements.txt`--but at the same time it'd be great to eliminate `test_requirements` so I'm not sure what's best!

nelson-liu:
another option is to turn off notebook tests by default in `test-install` and make users install jupyter in order to run them.

joelgrus:
ANOTHER option is to rewrite the notebook tutorials as markdown tutorials and get rid of our jupyter dependency altogether 😇

DeNeutoy:
@nelson-liu Either solution you propose is fine - we can re-write the tutorials in a separate PR

Issue #1033, opened 2018-03-25
Attention matrix in textual entailment demo


luffycodes:
Feature request: Can you expose the attention matrix in textual entailment demo?

DeNeutoy:
This probably isn't super high priority for us, but it shouldn't be too difficult -
 you can see[here](https://github.com/allenai/allennlp/blob/master/allennlp/models/decomposable_attention.py#L139) and [here](https://github.com/allenai/allennlp/blob/master/allennlp/models/decomposable_attention.py#L144)
The attention that needs to be returned from the model.
[here](https://github.com/allenai/allennlp/blob/master/demo/src/components/McComponent.js#L162) is an example of the react components that you need to add to the demo. 

Feel free to submit a PR for this - we will definitely accept it. Feel free to ask more questions etc if you need more pointers.

PR #1219, opened 2018-05-15
Add the attention visualization to the textual entailment demo

murphp15:
screen shot 2018-05-15 at 14 50 50

### DIFF
--- a/allennlp/data/dataset_readers/snli.py
+++ b/allennlp/data/dataset_readers/snli.py
@@ -7,7 +7,7 @@ from overrides import overrides
 from allennlp.common import Params
 from allennlp.common.file_utils import cached_path
 from allennlp.data.dataset_readers.dataset_reader import DatasetReader
-from allennlp.data.fields import Field, TextField, LabelField, MetadataField
+from allennlp.data.fields import Field, TextField, LabelField
 from allennlp.data.instance import Instance
 from allennlp.data.token_indexers import SingleIdTokenIndexer, TokenIndexer
 from allennlp.data.tokenizers import Tokenizer, WordTokenizer
@@ -71,10 +71,6 @@ class SnliReader(DatasetReader):
         hypothesis_tokens = self._tokenizer.tokenize(hypothesis)
         fields['premise'] = TextField(premise_tokens, self._token_indexers)
         fields['hypothesis'] = TextField(hypothesis_tokens, self._token_indexers)
-        fields['metadata'] = MetadataField({
-                'premise_tokens': [token.text for token in premise_tokens],
-                'hypothesis_tokens': [token.text for token in hypothesis_tokens]
-                })
         if label:
             fields['label'] = LabelField(label)
         return Instance(fields)
--- a/allennlp/data/dataset_readers/snli.py
+++ b/allennlp/data/dataset_readers/snli.py
@@ -7,7 +7,7 @@ from overrides import overrides
 from allennlp.common import Params
 from allennlp.common.file_utils import cached_path
 from allennlp.data.dataset_readers.dataset_reader import DatasetReader
-from allennlp.data.fields import Field, TextField, LabelField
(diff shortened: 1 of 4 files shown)


joelgrus:
could I ask that you give your PRs (especially) and commits (less crucial) more descriptive names? most of us don't have the issue numbers memorized, so the titles aren't very helpful.

we typically link to the relevant issue in the description of the PR (which has the benefit of automatically associating the PR with the issue)

murphp15:
Fixes #1033

matt-gardner:
Github is weird - you have to edit the original PR description with the `Fixes #1033` comment, instead of adding a new comment with that text.

matt-gardner:
Yeah, this looks much better, thanks.  Future commits on this PR (if there are any) don't need to have the same "Fixes" message, they should have something more descriptive of the contents of that commit.  We also have typically just put the "Fixes" message in the original PR description instead of a commit message, but either is fine.

allennlp/data/dataset_readers/snli.py
```diff
@@ -71,6 +71,10 @@ def text_to_instance(self,  # type: ignore
         hypothesis_tokens = self._tokenizer.tokenize(hypothesis)
         fields['premise'] = TextField(premise_tokens, self._token_indexers)
         fields['hypothesis'] = TextField(hypothesis_tokens, self._token_indexers)
+        fields['metadata'] = MetadataField({
```
matt-gardner: Instead of adding a `MetadataField` that the model doesn't really need, I think this would be cleaner if you just modified the [`Predictor`](https://github.com/allenai/allennlp/blob/master/allennlp/service/predictors/decomposable_attention.py).  The `Predictor` can tokenize the text and pass the tokens on as output to the demo.  That would look something like this: https://github.com/allenai/allennlp/blob/f246da737791ebe4fdbc6eca3119762eefa0c55b/allennlp/service/predictors/sentence_tagger.py#L27-L39

You'd have to grab the `_tokenizer` from `self._dataset_reader` in the `Predictor` (with `# pylint: disable=protected-access`) and use that to tokenize the text, instead of instantiating a tokenizer in `__init__`.  You'd also need to modify this `DatasetReader`'s `text_to_instance` method to optionally take pre-tokenized text, which would look something like how `passage_tokens` is treated here: https://github.com/allenai/allennlp/blob/f246da737791ebe4fdbc6eca3119762eefa0c55b/allennlp/data/dataset_readers/reading_comprehension/squad.py#L74-L83

allennlp/models/decomposable_attention.py
```diff
@@ -164,13 +165,29 @@ def forward(self,  # type: ignore
 
         output_dict = {"label_logits": label_logits, "label_probs": label_probs}
 
+        self.add_insight_to_output_dict(metadata, output_dict, p2h_attention, h2p_attention)
```
matt-gardner: After switching the `MetadataField` to something on the `Predictor`, it seems a bit of overkill to add a method just to add the two attention fields to `output_dict`.  I'd just put those two lines inline here.

allennlp/models/decomposable_attention.py
```diff
@@ -164,13 +165,29 @@ def forward(self,  # type: ignore
...
 
         return output_dict
 
+    @staticmethod
+    def add_insight_to_output_dict(metadata, output_dict, p2h_attention, h2p_attention):
+        premise_tokens = []
+        hypothesis_tokens = []
+        if metadata is not None:
+            for datum in metadata:
+                premise_tokens.append(datum['premise_tokens'][:len(datum['premise_tokens'])-1])
```
matt-gardner: Just FYI (because this should be removed, anyway), the `MetadataField` does not get padded, so you don't need to do this `[:len...]` stuff.  It will just be the exact list you created in the `DatasetReader`.

murphp15:
PR updated.

Issue #1223, opened 2018-05-16
Setting "evaluate_on_test", the last model is evaluated, not the best one


janLuke:
While everyone would expect the best weights are used. The best weights are the output of the training process, not the last ones. I think the current behavior is potentially harmful because it misguides who reads the results taking for granted (almost surely) those are the results for the best model.

matt-gardner:
Yep, definitely a bug.  PRs welcome.

matt-gardner:
My vote would be for the best model, I think.

joelgrus:
best model would fix #1192 as well

matt-gardner:
Yes, all of that sounds like an improvement to me.  We've been more focused on defining model-building APIs than on the particulars of the `Trainer`, which is why it's not as fully-featured as you might like.  Again, PRs are greatly appreciated - we're a small team.

PR #1225, opened 2018-05-16
train_model(): evaluates on test set the best model (not the last); returns the best model (not the last)

### LINKED ISSUE #1192: `ModelTestCase.ensure_model_can_train_save_and_load` compares last-epoch weights with serialized *best* weights
in particular, if your test case model trained (say) 10 epochs, and got the best results in epoch 5, then the test would fail because it would compare `model.state_dict()` (i.e. epoch 10) with `archive.model.state_dict()` (the weights from epoch 5)

in the meantime I am setting `num_epoch = 1` as a workaround

janLuke:
Quick fix of the bug #1223: setting "evaluate_on_test", the **last** model was evaluated, not the **best** one (as one expected). 

Furthermore, now train_model() returns the best model rather than the last, which fixes #1192.

### DIFF
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -38,6 +38,7 @@ import logging
 import os
 from copy import deepcopy
 
+import torch
 from allennlp.commands.evaluate import evaluate
 from allennlp.commands.subcommand import Subcommand
 from allennlp.common.checks import ConfigurationError
@@ -301,6 +302,13 @@ def train_model(params: Params,
     archive_model(serialization_dir, files_to_archive=params.files_to_archive)
 
     if test_data and evaluate_on_test:
+        logger.info("The model will be evaluated using the best epoch weights.")
+
+        logger.info("Loading the best epoch weights.")
+        best_model_state_path = os.path.join(serialization_dir, 'best.th')
+        best_model_state = torch.load(best_model_state_path)
+        model.load_state_dict(best_model_state)
+
         test_metrics = evaluate(model, test_data, iterator, cuda_device=trainer._cuda_devices[0])  # pylint: disable=protected-access
         for key, value in test_metrics.items():
             metrics["test_" + key] = value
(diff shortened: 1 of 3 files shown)


allennlp/commands/train.py
```diff
@@ -300,8 +306,15 @@ def train_model(params: Params,
     # Now tar up results
     archive_model(serialization_dir, files_to_archive=params.files_to_archive)
 
+    logger.info("Loading the best epoch weights.")
+    best_model_state_path = os.path.join(serialization_dir, 'best.th')
```
nelson-liu: instead of hardcoding `best.th`, maybe import `_DEFAULT_WEIGHTS` from model? https://github.com/allenai/allennlp/blob/master/allennlp/models/model.py#L25

janLuke: Good point. I actually looked for a constant with the strict semantic of "best weights file name" before hard-coding it. I saw ``_DEFAULT_WEIGHTS`` but: 1) I think it has another semantic and 2) it's "declared" as "private". Furthermore, ``"best.th"`` is hard-coded elsewhere, so I hard-coded it. I think we could introduce a constant ``BEST_WEIGHTS`` or better (though longer) ``BEST_WEIGHTS_FILENAME``.

janLuke: Actually, looking again at model.py, I think that ``_DEFAULT_WEIGHTS`` could be safely renamed as ``BEST_WEIGHTS`` or ``BEST_WEIGHTS_FILENAME``.

nelson-liu:
hmm, I think https://github.com/allenai/allennlp/pull/1225/commits/ea01031e3cb026620c40f69320b78c514e0a8c0a should go into another PR?

janLuke:
Damn, sorry, I did this by mistake ^.^". Not a git expert here. How can I fix it?

nelson-liu:
hmm, it's a bit tough since you're working on master--I'd recommend starting [a new branch](https://git-scm.com/book/en/v2/Git-Branching-Branches-in-a-Nutshell) when implementing a feature in the future. do you mind if i cherry pick your commits and merge them from there?

janLuke:
@nelson-liu I created a new PR from a new branch with a single commit (I cherry picked the 2nd commit of this PR and wrote a better description): https://github.com/allenai/allennlp/pull/1234. I'm closing this one. 

Hopefully I didn't make mistakes with the new one. Have patience, I've not much experience with git(hub) ^.^

@DeNeutoy Happy to lend a (small) hand :)

PR #1227, opened 2018-05-16
Tweak TestInstall._get_project_root()

nelson-liu:
I realized that it's a pretty strong assumption for this function to look for `setup.py` and `license` --- all we're really looking for is the `tests` directory. I also added another assert that the test file exists, just as a sanity check that we've got the right tests directory.


nelson-liu:
turns out this is only a bandage for a bigger issue --- drafting an issue right now 😢

PR #1230, opened 2018-05-17
Cache remote datasets with fixed-length filenames

goodmami:
This PR provides a working implementation of a sqlite3 index for mapping URLs to filenames in file_utils.get_from_cache(), but it is not complete. It still needs some more unit testing for some corner cases, such as when a file indexed in the database does not exist on the filesystem (it should be removed from the database).

This change is necessary for #1221, where the long base64-encoded filenames can reach the maximum filename length (255 generally, but the filesystem is encrypted with encryptfs it is 143).

I'm submitting the PR so I can get some feedback before I get too far with the finishing touches.


allennlp/common/file_utils.py
```diff
@@ -88,42 +91,95 @@ def get_from_cache(url: str, cache_dir: str = None) -> str:
...
-        with open(temp_filename, 'wb') as temp_file:
+
+    with _open_cache_index(cache_dir) as cursor:
+        filename = _url_to_filename(cursor, url, etag)
+
+        # filename will be None if it isn't cached
+        if filename is None:
+            # Download to temporary file, then copy to cache dir once finished.
+            # Otherwise you get corrupt cache entries if the download gets interrupted.
+            temp_file = tempfile.NamedTemporaryFile(delete=False)
```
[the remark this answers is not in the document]
goodmami: I actually had it as a `with` block but changed it back. I had a few reasons:
1. The `with` block adds a level of nesting (2 levels with the `cache_file` below), but this is just cosmetics
2. The `with` block closes the file, which is nice, but `shutil.copyfile()` occurs before the file is closed, so I need to put a `temp_file.flush()` before the copy or the contents get truncated
3. This way follows the previous logic more closely (e.g., the `logger.info()` call about removing the temp file would just appear at the end of the context manager's scope without an explicit `os.remove()`. I was trying to avoid unnecessary diffs for this first PR in order to highlight what's really different.

These reasons are minor, so I could be easily persuaded to use a `with` block.

allennlp/common/file_utils.py
```diff
@@ -88,42 +91,95 @@ def get_from_cache(url: str, cache_dir: str = None) -> str:
...
             for chunk in req.iter_content(chunk_size=1024):
                 if chunk: # filter out keep-alive new chunks
                     progress.update(len(chunk))
                     temp_file.write(chunk)
+            progress.close()
+            temp_file.close()
 
-        progress.close()
+            # Download succeeded. Now open cache file.
+            cache_file = tempfile.NamedTemporaryFile(dir=cache_dir, delete=False)
```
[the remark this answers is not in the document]
goodmami: Same reasons as above.

goodmami:
Yeah, I hear you about it feeling like overkill, and executing SQL queries isn't nearly as Pythonic as dictionary-like access. But it's a single-table database whose interface is contained within the private methods of the module, so it's more like a tuple store with write locks. And `sqlite3` is built-in to Python and it's pretty light-weight (no need for a SQL server process or anything). The `sqlitedict` project has a nicer interface, for sure, but it incurs another dependency, as you say.

(Aside: my worries on #1221 about `pickle` not being compatible across Python versions were overblown; I reread the docs and it's seems to be mostly backward compatible, so `sqlitedict`'s use of `pickle` wouldn't be a problem) 

Alternatively, do you really need the ability to retrieve the original URL from a cached file? Using a simple hash of the filename would make the code a lot simpler.

goodmami:
Great, I had the same idea this morning. The companion file isn't as
necessary so maybe thread safety is less important. I'll throw together a
new PR pretty soon.

> --
Michael Wayne Goodman

goodmami:
Ok actually my idea was a single json index, but I like your idea better. I've implemented that in the latest commit. This commit has fewer changes than the original. When/if merging, you can just squash the commits to remove the sqlite implementation from the history.

By the way, when etags are used, I append its hash separately rather than hashing them together. This is because:
1. It avoids collisions for, e.g., `{"url"="http://allenai.org/1", "etag"=None}` and `{"url"="http://allenai.org/", "etag"="1"}`; this could be avoided by first inserting a character that's invalid in a url, like a newline or something, but this feels too hacky
2. If you have multiple versions with the same urls and different etags, you can see this just by looking at the filenames, because the url parts of the hashes will be the same

allennlp/common/file_utils.py
```diff
@@ -21,37 +22,42 @@
...
-    return url_bytes.decode('utf-8'), etag
+    cache_path = os.path.join(cache_dir, filename)
+    if not os.path.exists(cache_path):
+        raise FileNotFoundError("file {} not found".format(cache_path))
+
+    meta_path = cache_path + '.json'
+    if not os.path.exists(meta_path):
+        raise FileNotFoundError("file {} not found".format(meta_path))
+
+    metadata = json.load(open(meta_path))
```
joelgrus: `with` block? 😇

allennlp/common/file_utils.py
```diff
@@ -105,25 +111,34 @@ def get_from_cache(url: str, cache_dir: str = None) -> str:
...
+
+            logger.info("copying %s to cache at %s", temp_file.name, cache_path)
+            with open(cache_path, 'wb') as cache_file:
+                shutil.copyfileobj(temp_file, cache_file)
 
-        progress.close()
+            logger.info("creating metadata file for %s", cache_path)
+            meta = {'url': url, 'etag': etag}
+            meta_path = cache_path + '.json'
+            json.dump(meta, open(meta_path, 'w'))
```
joelgrus: `with`

tests/common/file_utils_test.py
```diff
@@ -88,7 +102,7 @@ def test_get_from_cache(self):
         set_up_glove(url, self.glove_bytes, change_etag_every=2)
 
         filename = get_from_cache(url, cache_dir=self.TEST_DIR)
-        assert filename == os.path.join(self.TEST_DIR, url_to_filename(url, etag="0"))
+        assert os.path.exists(filename)
```
joelgrus: is there a reason for weakening these tests?

allennlp/common/file_utils.py
goodmami: It's not necessary. I'm pretty sure file objects close when they are out of scope, which here is as soon as the `json.load()` function call finishes. But it's no problem to use a `with` block. Explicit (or slightly less implicit) is better than implicit, after all.

allennlp/common/file_utils.py
goodmami: ditto

tests/common/file_utils_test.py
goodmami: Ah, no, not anymore. This is a holdover from the previous commit, where `url_to_filename` no longer had any relation to the cached filename. I'll fix it.

goodmami:
@joelgrus no worries. Now I can continue with the tutorial :)

Thanks for the help with the PR!

Issue #1228, opened 2018-05-16
running allennlp test-install after installing with pip leads to many errors


nelson-liu:
After installing AllenNLP with: `pip install git+git://github.com/allenai/allennlp.git` and running `allennlp test-install`, a bunch of errors come up. Raising an issue just to enumerate them all. A full gist of the test traceback is here: https://gist.github.com/nelson-liu/78d08ad0090ad8f95c34675942f5bb4d

- The biggest issue is that we lose the directory structure in the repo. For example, `allennlp` with pip would get installed in `/home/nfliu/miniconda3/envs/test/lib/python3.6/site-packages/allennlp`. this contains _only_ the module (so it's [this folder](https://github.com/allenai/allennlp/tree/master/allennlp)).  The `tutorials`, `scripts`, `READMEs`, etc. aren't actually installed anywhere, so all of the paths in the tests break (e.g., script tests that shell out to run `perl ./scripts/some_eval_script.pl`).

- On a separate related note it looks like the precompiled custom extensions in the module aren't copied over either, maybe because they're not `.py` files and they get ignored automatically? This makes the `test_custom_extensions` / `alternating_lstm` tests break, even though they should work. This also means that custom extensions in general aren't working in the `pip install`ed version of allennlp.

## Things that are actually broken in the pip-installed version

- the custom highway lstm (since the extensions aren't included). fixed by #1229 
- `evalb_bracketing_scorer`, since `scripts/EVALB` does not exist. fixed by #1233 
- any tests that rely on files in `./tutorials` and `./scripts`. wip solution in #1232 

The other test errors are just errors with the tests themselves, not with the allennlp package. Which is unfortunate, but it doesn't affect _core_ library functionality.

nelson-liu:
looks like `MANIFEST.in` controls what things go into the source distribution (the tarball) whereas the `package_data` key of `setup.py` controls what goes into the binary distribution / is copied over to `site-packages`...will try adding that and seeing what happens.

nelson-liu:
To update on this: after #1233  and #1229 , `test-install` on a pip installed allennlp works for all tests except for the notebook tests and a SRL test and a span-based F1 test that compare python output to perl output. I'm disregarding the notebook tests as a wontfix since ostensibly they'll be rewritten in markdown anyway.

The bigger issue is how to access the tests from pip. Currently, they're inadvertently being installed as a module, which is unreliable behavior and pretty bad packaging. 
to summarize from #1232: 
[...]

PR #1232, opened 2018-05-17
Fix test-install by moving tests into module

nelson-liu:
Goals of this PR:
1. [x] 1. Prevent tests from being installed as a module in `python setup.py` . Currently, pip copies them to `site-packages`, which is definitely not intended behavior.
2. [x] 2. installation method-invariant way to load and run tests, especially those that depend on files outside of the module (e.g. `scripts` or `tutorials`). The issue is that files can be in different places depending on if allennlp is installed with `pip` or `setup.py install`.

For 2: an easy solution might be to, when running the `test-install` command, simply download and cache a copy of the allennlp source for the appropriate version at `~/.allennlp` . Note that this wouldn't work for non-release versions, but I think that's a good thing --- if you're working on master or developing features, you shouldn't be using `test-install` anyway.

Thoughts? I'm not sure how i feel about the above solution, so feedback would be great before i write the code.

----------------------------------------------------------------------------------

*Edit*: another solution is to just move the tests into the `allennlp` module (this is a fairly standard way of structuring projects anyway). The downside to this is that it still doesn't enable using `allennlp test-install` to run tests in `tutorials`, as well as the tests that depend on files outside the module (e.g. SRL test that compares output of evaluator with a perl script). But maybe we shouldn't be running those tests anyway in `test-install`? or we can move the perl script into the tests and then just not run the notebook tests, since we'll be rewriting them anyway.

This solution is rather time-intensive, though, since it involves fixing all of the paths that are hardcoded into the tests. Should not be too hard to write a script to go through and change the paths, minimizing human labor to just a check up...

edit: fixes https://github.com/allenai/allennlp/issues/1228


nelson-liu:
Another thing we could do is move the tests within the `allennlp` module...I know of at least a few others packages that do this (e.g., [scikit-learn](https://github.com/scikit-learn/scikit-learn/tree/master/sklearn/tests) and [pandas](https://github.com/pandas-dev/pandas/tree/master/pandas/tests). Maybe this solution is a bit drastic? I don't think too much depends on where the tests are actually located (maybe some of the CI scripts? But `py.test` in the root should continue to work)

nelson-liu:
I went ahead and made the change, was a lot less work than I anticipated. even if we don't want to move tests within the module, a lot of these commits can be cherry-picked for an eventual window-compatibility PR.

maybe best to look at this after EMNLP...

schmmd:
@nelson-liu what's the difference between `tools` and `scripts`?

nelson-liu:
1) `scripts` are standalone, one-off entities, while `tools` are core to some other component of the package (e.g. `evalb` is essential to the constituency parser metric, and the srl perl eval script is used in the tests). In general, if some outside tool/script is necessary to make a component of the library function properly, it should go into `tools`.

2) since scripts are standalone, they go outside of the package. Tools are explicitly called by things inside the package, so they should go inside the package as well. I named them differently to make it easier to distinguish.

nelson-liu:
(let me know if there's anything I can do to make this easier to review, as well).

schmmd:
@nelson-liu I think the names are fine.  A brief (1 sentence) README.md in each folder would be great, defining the role of each.  Overall, this seems like a lot of improvements, but also a lot of changes.  As it touches the whole codebase I'm not sure if I'm the best person to code review it.

allennlp/common/file_utils.py
```diff
@@ -60,7 +61,7 @@ def filename_to_url(filename: str, cache_dir: str = None) -> Tuple[str, str]:
 
     return url, etag
 
-def cached_path(url_or_filename: str, cache_dir: str = None) -> str:
```
DeNeutoy: @joelgrus does this `Union`, and all the `str(path)` calls in this PR change how you feel about `Path`?

DeNeutoy:
I'm still unsure whether it's desirable to have users run tests. Can you point me to some other python projects which do this?

nelson-liu:
NumPy, SciPy let users run their tests through the CLI (https://stackoverflow.com/a/9200923/2544124):

```
import numpy
numpy.test()
```

(and accordingly for scipy)

Also [theano](http://deeplearning.net/software/theano_versions/0.8.X/install.html#testing-your-installation)

schmmd:
For another motivating use case, I'd like to create a Docker image that downstream users can use to run Beaker.  It would look [like @cristipp 's](https://github.com/allenai/euclid/blob/master/src/main/python3/pipette/Dockerfile) and would install AllenNLP via `pip`.  Having the ability to test the pip distribution would be invaluable for making sure the image works as expected.

allennlp/tests/commands/test_install_test.py
```diff
@@ -0,0 +1,12 @@
+# pylint: disable=invalid-name,no-self-use
+import os
+
+from allennlp.common.testing import AllenNlpTestCase
+from allennlp.commands.test_install import _get_module_root
+
+
+class TestTestInstall(AllenNlpTestCase):
```
schmmd: What is the motivation for this test?  A brief comment would be helpful.  I'm confused because it's a test, in the tests folder, that's checking to see whether the tests folder exists.

allennlp/tests/training/metrics/wikitables_accuracy_test.py
```diff
@@ -29,7 +31,8 @@ def test_accuracy_is_scored_correctly(self):
         logical_form = ('((reverse fb:row.row.year) (fb:row.row.index (max '
                         '((reverse fb:row.row.index) (fb:row.row.league fb:cell.usl_a_league)))))')
 
-        wikitables_accuracy = WikiTablesAccuracy(table_directory='tests/fixtures/data/wikitables/')
+        print(os.getcwd())
```
schmmd: @nelson-liu is this an intentional `print`?

mypy.ini
```diff
@@ -0,0 +1,4 @@
+[mypy]
+
+[mypy-allennlp.tests*]
+ignore_errors = True
```
schmmd: Why is this necessary? Did `mypy` previously not run on our tests?

allennlp/tests/commands/test_install_test.py
schmmd: It looks like this is a rename rather than a new file.  While that doesn't change my confusion, you don't need to address it as part of this PR.

mypy.ini
nelson-liu: Yes, mypy was previously not being run on the testing code --- I thought this was intentional?

allennlp/tests/training/metrics/wikitables_accuracy_test.py
nelson-liu: it's not, thanks for catching that!

mypy.ini
schmmd: Seems like you made the right change in this PR then.  I'm not sure what we want long term.

schmmd:
@nelson-liu sounds great.  If you can move `srl-eval.pl` instead of creating a new copy, then I'll approve and merge.

allennlp/tests/commands/test_install_test.py
nelson-liu: I agree that this is quite confusing. The context is that when a user runs `test-install`, we have no idea where they're running it from, so we do an `os.chdir` to the module root in order to get all the paths to resolve properly.

The logic within `test-install` is pretty hard to test in its entirety ([see discussion here](https://github.com/allenai/allennlp/pull/1213#issuecomment-388897160)), so this test is basically a test for that `os.chdir` code, checking that we correctly find the path to `chdir` to.

does that help? I'll add this as a comment too.

nelson-liu:
before you merge, i'd like to run some last sanity-check tests on the code in this branch (building it from github source with pip on a clean environment and seeing if everything works as expected). I'll let you know when that's done.

Maybe this is something you might want to add to CI as well? Not sure.

DeNeutoy:
I was semi-serious about not liking the `Path` thing - I think it will just cause us to have loads of `str(path)` things everywhere and doesn't really add much. I'd kinda prefer it if this PR switched back to using string paths.... @joelgrus lets argue

DeNeutoy:
Meh maybe i'm just being grumpy and un-modern - I didn't realise it might fix some of the windows compatability issues. If that's the case let's go for it.

allennlp/tests/commands/test_install_test.py
schmmd: Yes that helps!

mypy.ini
matt-gardner: @nelson-liu, do the mypy errors on the tests look helpful?  I'd guess that we don't care about them, because it's not as important to have clear type annotations in the tests, but if they look useful we could reconsider.

matt-gardner:
@DeNeutoy, I agree that the only benefit I see to using the `Path` stuff is better windows compatibility.  Otherwise just using a string path is simpler and easier to read.

mypy.ini
nelson-liu: @matt-gardner here are the mypy errors: https://gist.github.com/nelson-liu/f94a96265973a390066f45399718fe5e . I guess they're not entirely useless? Perhaps we should address in another PR though, this one is already huge.

nelson-liu:
I'll merge this when CI passes, unless there's anything further to be discussed?

I pip installed allennlp from this branch on an osx machine and a linux gpu machine, and `allennlp test-install` passed on both hosts.

joelgrus:
to @DeNeutoy 's point, calling `str()` on `pathlib.Path` objects is a bad code smell. if the places we need those paths can't accept `pathlib.Path` object then I would rather just use strings.

(my ideal solution would be that those places accept Path objects tho)

mypy.ini
matt-gardner: Hmm, yeah, ok, I'll open that as an issue.  Thanks for doing all of this!

nelson-liu:
thanks all for taking a look

PR #1234, opened 2018-05-17
In train_model(), load the best weights at the end so that the best model (not the last) is evaluated on test-set and returned

janLuke:
1. the best model (not the last) is evaluated on the test set if "evaluate_on_test";
this fixes https://github.com/allenai/allennlp/issues/1223

2. the best model (not the last) is returned by the function;
this is a more natural behaviour and incidentally solves https://github.com/allenai/allennlp/issues/1192

### DIFF
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -38,6 +38,8 @@ import logging
 import os
 from copy import deepcopy
 
+import torch
+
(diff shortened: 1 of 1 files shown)


janLuke:
Glad to contribute! Thank you all for this project :)

PR #1235, opened 2018-05-17
Improve performance of attention modules

murphp15:
This is an initial POC for getting feedback before proceeding with the entire refactor.
In this PR
1. I added implementations of MatrixAttention. there is one for dot product, one for cosine product, etc.
2. There is an implementation of MatrixAttention called LegacyMatrixAttention used to support backwards compatibility and use already trained models.
3. There is a class called AttentionMatrixFactory which can be used to create an instance of MatrixAttention. it will determine whether to use the legacy impl or the newer impl based on the name.

Let me know what you guys think, all feedback welcome.


allennlp/modules/legacy_matrix_attention.py
```diff
@@ -0,0 +1,65 @@
...
+
+from allennlp.modules.similarity_functions.similarity_function import SimilarityFunction
+
+
+from allennlp.common import Params
+from allennlp.modules.matrix_attention import MatrixAttention
+from overrides import overrides
+
+
+class LegacyMatrixAttention(MatrixAttention):
```
matt-gardner: Register this as `legacy`.

allennlp/modules/dot_product_matrix_attention.py
```diff
@@ -0,0 +1,26 @@
...
+"""
+import math
+
+import torch
+
+from allennlp.common import Params
+from allennlp.modules.matrix_attention import MatrixAttention
+from overrides import overrides
+
+@MatrixAttention.register("dot_product_matrix_attention")
```
matt-gardner: These names don't need to have `_matrix_attention` in them, because that'll be obvious based on context.  E.g., here it's inside of `MatrixAttention.register()`, so it's obviously matrix attention, and where you give the parameter in the configuration file it's inside of an "attention" block (or similar).

allennlp/models/reading_comprehension/bidaf.py
```diff
@@ -332,7 +333,7 @@ def from_params(cls, vocab: Vocabulary, params: Params) -> 'BidirectionalAttenti
         text_field_embedder = TextFieldEmbedder.from_params(vocab, embedder_params)
         num_highway_layers = params.pop_int("num_highway_layers")
         phrase_layer = Seq2SeqEncoder.from_params(params.pop("phrase_layer"))
-        similarity_function = SimilarityFunction.from_params(params.pop("similarity_function"))
+        matrix_attention = MatrixAttentionFactory.from_params(params.pop("similarity_function"))
```
matt-gardner: You don't need this factory at all, you just call `MatrixAttention.from_params()`.

allennlp/modules/legacy_matrix_attention.py
```diff
@@ -0,0 +1,65 @@
...
+                                                      matrix_1.size()[2])
+        tiled_matrix_2 = matrix_2.unsqueeze(1).expand(matrix_2.size()[0],
+                                                      matrix_1.size()[1],
+                                                      matrix_2.size()[1],
+                                                      matrix_2.size()[2])
+        return self._similarity_function(tiled_matrix_1, tiled_matrix_2)
+
+    @classmethod
+    def from_params(cls, params: Params) -> 'MatrixAttention':
+        similarity_function = SimilarityFunction.from_params(params)
```
matt-gardner: I would expect to specify this using something like:

```json
"attention": {
  "type": "legacy",
  "similarity_function": { "params": "here" }
}
```

This way you don't need the "factory" stuff you put in, and everything still works.

tests/fixtures/bidaf/experiment.json
```diff
@@ -46,7 +46,7 @@
       "num_layers": 1
     },
     "similarity_function": {
```
matt-gardner: And in these places we'd want to change this from a "similarity_function" key to a "matrix_attention" key.  This means also changing the `BiDAF` model to take a `MatrixAttention` object instead of a `SimilarityFunction` object in its constructor.

allennlp/modules/matrix_attention_factory.py
```diff
@@ -0,0 +1,23 @@
+from allennlp.modules.matrix_attention import MatrixAttention
+
+from allennlp.common.params import Params
+from allennlp.modules.legacy_matrix_attention import LegacyMatrixAttention
+
+
+class MatrixAttentionFactory:
+    """
+             Responsible for supporting both the legacy and new attention mechanisms.
+             New implementations of the attention were added because they have a much smaller memory footprint.
```
matt-gardner: This deprecation notice can just go inside `LegacyMatrixAttention`.

tests/fixtures/bidaf/experiment.json
murphp15: Again as stated in my comment above this would break backwards compatibility. Is that ok? 

In this current PR the bidaf does take a MatrixAttention object instead of a SimilarityFunction. 
In the from_params method it converts these similarity_function into the correct MatrixAttention impl without compromising backwards compatibility.

matt-gardner: Yes, this will break backwards compatibility of the json configuration files.  That's ok.

allennlp/modules/legacy_matrix_attention.py
```diff
@@ -0,0 +1,65 @@
...
+    '''
+    def __init__(self, similarity_function: SimilarityFunction = None) -> None:
+        super(LegacyMatrixAttention, self).__init__()
+
+        self._similarity_function = similarity_function or DotProductSimilarity()
+
+    @overrides
+    def forward(self, matrix_1: torch.Tensor, matrix_2: torch.Tensor) -> torch.Tensor:
+        # pylint: disable=arguments-differ
+        tiled_matrix_1 = matrix_1.unsqueeze(2).expand(matrix_1.size()[0],
```
murphp15: This is a copy and paste from the existing implementation. 
I can't see anyway to avoid the tiling here for the element wise operations! 
If anybody has any idea how to avoid it let me know.

tests/fixtures/bidaf/experiment.json
matt-gardner: Actually, another idea is just to keep the `BiDAF` and other model code as it is, but use `LegacyMatrixAttention()` to construct the attention instead of `MatrixAttention`, in the constructor.  Then no fixtures need to be updated, the sniff tests still pass, and everything is ok.  New code can still use the nicer API, but we don't break anything at all.

allennlp/modules/legacy_matrix_attention.py
matt-gardner: In the legacy one, we're going to have to keep the tiling, because of how similarity functions are used.  The other ones can be made more efficient, but it'll be easier to work on that part after you have tests for each of the new implementations, and we can be sure that the new, more efficient implementation still passes the tests.

allennlp/modules/linear_matrix_attention.py
```diff
@@ -0,0 +1,95 @@
...
+        self.reset_parameters()
+
+    def reset_parameters(self):
+        std = math.sqrt(6 / (self._weight_vector.size(0) + 1))
+        self._weight_vector.data.uniform_(-std, std)
+        self._bias.data.fill_(0)
+
+    @overrides
+    def forward(self, tensor_1: torch.Tensor, tensor_2: torch.Tensor) -> torch.Tensor:
+        tiled_matrix_1 = tensor_1.unsqueeze(2).expand(tensor_1.size()[0],
```
murphp15: @matt-gardner I left a comment above on the legacy implementation saying I could not avoid using the expand. 
I actually meant to leave that comment here. 
I was not able to avoid it here either! 
Any ideas for here?

matt-gardner: Hmm, the arbitrary combinations makes this tricky.  Let me think about this one.  Having this one be less efficient should block the rest of the refactoring, though.

tests/fixtures/bidaf/experiment.json
murphp15: If I understand you correctly you are saying
1. the constructor to the BiDAF would still take in a similarity function (so there is no breaking change).
2. Within the constructor it will create LegacyMatrixAttention. 


If we do this then the bidaf will be bound to the slow performance of the legacy impl.
Is that ok?

allennlp/modules/linear_matrix_attention.py
murphp15: should block? Did you mean should *not* block?

tests/fixtures/bidaf/experiment.json
matt-gardner: For this PR, yes.  That way we have a minimal change here that introduces a new API.  Then in a future PR, we can think about updating the code (and our sniff tests, and everything) to use the new API.

murphp15: Cool

allennlp/modules/linear_matrix_attention.py
matt-gardner: Yeah, typo that I fixed...  Sorry =).

allennlp/modules/__init__.py
```diff
@@ -21,3 +21,7 @@
 from allennlp.modules.text_field_embedders import TextFieldEmbedder
 from allennlp.modules.time_distributed import TimeDistributed
 from allennlp.modules.token_embedders import TokenEmbedder, Embedding
+from allennlp.modules.dot_product_matrix_attention import DotProductMatrixAttention
```
matt-gardner: Oh, one other note that I forgot last time - it'd be nice to make a `matrix_attention/` module, and put all of these in there, like how the `seq2seq_encoders` are structured.

allennlp/modules/matrix_attention/cosine_matrix_attention.py
```diff
@@ -7,20 +7,17 @@
...
-
     @overrides
     def forward(self, matrix_1: torch.Tensor, matrix_2: torch.Tensor) -> torch.Tensor:
         # pylint: disable=arguments-differ
         return F.cosine_similarity(matrix_1, matrix_1)
 
-
+    @classmethod
     def from_params(cls, params: Params):
         CosineMatrixAttention()
```
matt-gardner: It's important to call `params.assert_empty(cls.__name__)` before returning the object here (and in the corresponding methods for the other classes).  This helps to catch errors in the configuration file.  You're also missing a return statement.

allennlp/modules/matrix_attention/cosine_matrix_attention.py
```diff
@@ -0,0 +1,23 @@
...
+from allennlp.modules.matrix_attention.matrix_attention import MatrixAttention
+
+
+@MatrixAttention.register("cosine")
+class CosineMatrixAttention(MatrixAttention):
+
+    @overrides
+    def forward(self, matrix_1: torch.Tensor, matrix_2: torch.Tensor) -> torch.Tensor:
+        # pylint: disable=arguments-differ
+        return F.cosine_similarity(matrix_1, matrix_1)
```
murphp15: @matt-gardner  I am now working on tests and correct implementations for these attention modules. 

I want to compute the cosine similarity of the last dimension in matrix_1 with the cosine similarity of the last dimension in matrix_2 for all entries. 

I tried using:  F.cosine_similarity(matrix_1, matrix_2, dim=-1)

However this only computes the cosine similarity of each entry in matrix_1 with the corresponding entry in matrix_2.

However we want this for all permutations. 
Do you know how to do this without using the expand? 
If there is not a way to do this without using expand then this ticket might not be worth doing.

matt-gardner: We have an example of this kind of cosine similarity here: https://github.com/allenai/allennlp/blob/9ef013d3a8abee940dbe8e2300f79f8b48cbea42/allennlp/models/semantic_parsing/wikitables/wikitables_semantic_parser.py#L191-L198

And even if some of these implementations don't get any more efficient (e.g., the linear one might not be possible), some of them definitely do, so it's worth it.

allennlp/modules/attention/linear_attention.py
```diff
@@ -0,0 +1,61 @@
...
+        self._weight_vector.data.uniform_(-std, std)
+        self._bias.data.fill_(0)
+
+
+    @overrides
+    def forward(self,  # pylint: disable=arguments-differ
+                vector: torch.Tensor,
+                matrix: torch.Tensor,
+                matrix_mask: torch.Tensor = None) -> torch.Tensor:
+        tiled_vector = vector.unsqueeze(1).expand(vector.size()[0],
```
murphp15: @joelgrus @matt-gardner 
I am facing the same issue here too, I don't think this impl will be more efficient than the legacy impl because we can't really avoid the tiling. 
For example: 
In the scenario where the combination is "x,y". 
 we will need to tile x before concatenating  the 2 tensors along the last axis. 
Am I correct in saying that?

allennlp/modules/matrix_attention/cosine_matrix_attention.py
murphp15: @matt-gardner @joelgrus 
I think I might have taken the wrong approach originally as I am creating a lot of duplicated code which is bringing no perf gain. 
Instead of having an Implementation of MatrixAttention and Attention called LegacyMatrixAttention and LegacyAttention respectively, maybe we should call these SimilarityFunctionWrappingMatrixAttention and SimilarityFunctionWrappingAttention.
For all the implementations that we can't improve we just leave them use the SimilarityFunctionWrapping* class. 
This way we reduce the amount of duplicated code we are adding.

For all functions we can improve the performance of we will do that in this PR. 

What do you guys think? Also the names of the classes are subject to your approval and if you have better names I'm happy to go with them.

murphp15:
@matt-gardner @joelgrus
I think I might have taken the wrong approach originally as I am creating a lot of duplicated code which is bringing no perf gain.
Instead of having an Implementation of MatrixAttention and Attention called LegacyMatrixAttention and LegacyAttention respectively, maybe we should call these SimilarityFunctionWrappingMatrixAttention and SimilarityFunctionWrappingAttention.
For all the implementations that we can't improve we just leave them use the SimilarityFunctionWrapping* class.
This way we reduce the amount of duplicated code we are adding.

For all functions we can improve the performance of we will do that in this PR.

What do you guys think? Also the names of the classes are subject to your approval and if you have better names I'm happy to go with them.

allennlp/modules/matrix_attention/multi_headed_matrix_attention.py
```diff
@@ -0,0 +1,117 @@
...
+from overrides import overrides
+from allennlp.common import Params
+from allennlp.common.checks import ConfigurationError
+from allennlp.modules.similarity_functions import SimilarityFunction
+from allennlp.modules.matrix_attention.matrix_attention import MatrixAttention
+from allennlp.modules.similarity_functions import DotProductSimilarity
+
+
+@MatrixAttention.register("multiheaded")
+class MultiHeadedMatrixAttention(MatrixAttention):
```
matt-gardner: This one doesn't match, and so shouldn't be registered like this.  It's not interchangeable with the other `MatrixAttention` classes.

This does raise the question of whether we should combine the attention and the weighted sum, as then we'd have something that's interchangeable here, but let's worry about that later.  I'd probably just remove this file entirely - we already have a multi-headed self-attention encoder that doesn't use this code, and we probably don't need this at all.

tests/modules/matrix_attention_test.py
```diff
@@ -1,24 +1,27 @@
 # pylint: disable=no-self-use,invalid-name
```
matt-gardner: Can you rename this file to `legacy_matrix_attention_test.py`?  We like to have one test file per code file, that corresponds to the python file being tested.  Also rename the test class.

tests/modules/matrix_atttention/matrix_attention_tests.py
```diff
@@ -0,0 +1,35 @@
+import torch
+from numpy.testing import assert_almost_equal
+import numpy as np
+from allennlp.common.testing.test_case import AllenNlpTestCase
+from allennlp.modules.matrix_attention import CosineMatrixAttention, DotProductMatrixAttention
```
matt-gardner: This would be better as two different files, with class names that match what we've done in the rest of the repo.

matt-gardner:
Also, I'm pretty sure we can make the `TriLinearAttention` in #804 general, and use that mechanism for a memory-efficient `LinearMatrixAttention` implementation.  We just need to be a little more complicated when constructing weights for the combinations, and when computing each of the combinations.  But [this code](https://github.com/allenai/allennlp/pull/804/files#diff-6075b31aebea7c93668ff896c813ac2cR39) shows that each piece of the combination can indeed be computed more efficiently, we just have to have some logic separating them out and doing the requested combinations.

allennlp/modules/matrix_attention/multi_headed_matrix_attention.py
murphp15: Why does this one not match? 
It has the same method declaration as MatrixAttention. What am I missing?

murphp15:
@matt-gardner 
In regards to your comment on multiheaded attention.
Why does it not match?
It has the same method declaration as MatrixAttention. What am I missing?

I re-wrote it here as I deleted the file and am worried you won't see the comment.

murphp15:
@matt-gardner 
I think this is ready for review now. 
Let me know what you think.

matt-gardner:
@murphp15 Multi-headed attention has the same method declaration, but so does any torch `Module` - it's just `forward`.  The issue is that the expected input and output shapes of the tensors do not match.  We define these `Registrable` abstractions on things that are interchangeable in a model without requiring any code change.  For that to work, the tensors that the modules take as input and the tensors they return as output have to have identical shapes - otherwise, when you switch to a different implementation of the same `Registrable` object, your code breaks, because something is getting a differently-shaped tensor than it was expecting.

`MatrixAttention` defines its output as having shape `(batch_size, num_rows_1, num_rows_2)`.  The multi-headed attention function actually returns something that has shape `(batch_size, num_rows_1, num_rows_2, num_heads)`.  This breaks the abstraction.

allennlp/modules/attention/attention.py
```diff
@@ -32,38 +32,31 @@ class Attention(torch.nn.Module):
...
     def __init__(self,
-                 similarity_function: SimilarityFunction = None,
                  normalize: bool = True) -> None:
         super(Attention, self).__init__()
-
-        self._similarity_function = similarity_function or DotProductSimilarity()
         self._normalize = normalize
 
     @overrides
     def forward(self,  # pylint: disable=arguments-differ
```
matt-gardner: This is the base class, where we're overriding `torch.nn.Module.forward` directly, so we need to disable this pylint check here.  No subclasses of this class should disable this check.

allennlp/modules/attention/legacy_attention.py
```diff
@@ -0,0 +1,39 @@
...
+
+from overrides import overrides
+from allennlp.common import Params
+from allennlp.modules.attention.attention import Attention
+from allennlp.modules.similarity_functions import DotProductSimilarity, SimilarityFunction
+
+
+@Attention.register("legacy")
+class LegacyAttention(Attention):
+
```
matt-gardner: Again here, remove the module docstring and add a class docstring.  This one should say something about using a configurable `SimilarityFunction` to compute the attention, but in a way that uses more memory than it needs to.  This is deprecated in favor of more efficient `Attention` implementations that don't use `SimilarityFunctions`.

allennlp/modules/attention/legacy_attention.py
```diff
@@ -0,0 +1,39 @@
...
+    def __init__(self,
+                 similarity_function: SimilarityFunction = None,
+                 normalize: bool = True) -> None:
+        super(LegacyAttention, self).__init__(normalize)
```
matt-gardner: This can just be `super().__init__(normalize)`.

allennlp/modules/attention/linear_attention.py
```diff
@@ -0,0 +1,92 @@
...
+from allennlp.modules.attention.legacy_attention import Attention
+from allennlp.nn import util
+from allennlp.nn.activations import Activation
+from allennlp.common.params import Params
+
+
+@Attention.register("linear")
+class LinearAttention(Attention):
+    """
+    This similarity function performs a dot product between a vector of weights and some
```
matt-gardner: This docstring needs to be updated, as this isn't a similarity function anymore.

allennlp/modules/attention/linear_attention.py
```diff
@@ -0,0 +1,92 @@
...
+        activation.
+    """
+
+    def __init__(self,
+                 tensor_1_dim: int,
+                 tensor_2_dim: int,
+                 combination: str = 'x,y',
+                 activation: Activation = Activation.by_name('linear')(),
+                 normalize: bool = True) -> None:
+        super(LinearAttention, self).__init__(normalize)
```
matt-gardner: Just `super()`.

allennlp/modules/attention/linear_attention.py
```diff
@@ -0,0 +1,92 @@
...
+        self._bias = Parameter(torch.Tensor(1))
+        self._activation = activation
+        self.reset_parameters()
+
+    def reset_parameters(self):
+        std = math.sqrt(6 / (self._weight_vector.size(0) + 1))
+        self._weight_vector.data.uniform_(-std, std)
+        self._bias.data.fill_(0)
+
+
```
matt-gardner: Only one blank line between class methods.

allennlp/modules/attention/linear_attention.py
```diff
@@ -0,0 +1,92 @@
...
+    @overrides
+    def _forward_internal(self,  # pylint: disable=arguments-differ
+                          vector: torch.Tensor,
+                          matrix: torch.Tensor,
+                          matrix_mask: torch.Tensor = None) -> torch.Tensor:
+        tiled_vector = vector.unsqueeze(1).expand(vector.size()[0],
```
matt-gardner: Can you add `# TODO(mattg): Remove the need for this tiling.` here?  You could also reference [this comment](https://github.com/allenai/allennlp/pull/1235#issuecomment-391540133).

murphp15:
@matt-gardner Ready for review again. 👍

murphp15:
@matt-gardner Are you sure that expand will consume more memory? 
From the documentation is says "Expanding a tensor does not allocate new memory, but only creates a new view on the existing tensor where a dimension of size one is expanded to a larger size by setting the stride to 0. Any dimension of size 1 can be expanded to an arbitrary value without allocating new memory." ?

allennlp/modules/__init__.py
```diff
@@ -5,13 +5,11 @@
 :class:`~allennlp.models.model.Model` s.
 """
 
-from allennlp.modules.attention import Attention
```
matt-gardner: We actually want these to still be here. We typically [import the base `Registrable` class in the module above](https://github.com/allenai/allennlp/blob/master/STYLE.md#module-layout-and-imports).

allennlp/modules/matrix_attention/__init__.py
```diff
@@ -0,0 +1,6 @@
+
+
```
matt-gardner: Can you import that base class here, also?

allennlp/modules/attention/__init__.py
```diff
@@ -0,0 +1,6 @@
+
+
```
matt-gardner: Import the base class here too.

allennlp/modules/matrix_attention/matrix_attention.py
```diff
@@ -0,0 +1,37 @@
...
+
+class MatrixAttention(torch.nn.Module, Registrable):
+    """
+    This ``Attention`` takes two matrices as input and returns a matrix of attentions.
+
+    We compute the similarity between each row in each matrix and return unnormalized similarity
+    scores.  Because these scores are unnormalized, we don't take a mask as input; it's up to the
+    caller to deal with masking properly when this output is used.
+
+    By default similarity is computed with a dot product, but you can alternatively use a
```
matt-gardner: This paragraph and the one below it still need to be removed.

matt-gardner:
Oh, and as far as `expand` goes, we have evidence from running stuff that switching from a `SimilarityFunction`-based attention to a specific one improves memory usage.  I agree with you that the documentation makes it look like that shouldn't be a problem, but something is up somewhere, because these other ways are just more efficient.  We're also generally using a single optimized operation instead of an expand, an elementwise multiply, and a sum, for example, and it could just be the creation of more intermediate matrices (and their corresponding gradients) that takes the extra memory.

allennlp/modules/__init__.py
murphp15: Why? 
I'm happy to do it just wondering why.

allennlp/modules/matrix_attention/__init__.py
murphp15: Also why?

allennlp/modules/__init__.py
joelgrus: to make imports cleaner / simpler in your code, so you can do e.g.

```
from allennlp.modules import Attention, FeedForward, ...
```

vs having to do

```
from allennlp.modules.attention import Attention
from allennlp.modules.feedforward import FeedForward
...
```

we do this consistently throughout the codebase, as per the linked style guide

murphp15: Ah cool. 
New to Python, thanks. Will do from now on.

PR #1244, opened 2018-05-18
Replace slicing with chunk along the last dim

### LINKED ISSUE #1211: Allow Highway to work with n-d inputs
Hey guys, great work here, thanks a lot!

I've been using your implementation of the Highway network with text data and I got an error while passing a 3d tensor as input, caused by the following lines: https://github.com/allenai/allennlp/blob/master/allennlp/modules/highway.py#L56

Essentially, the assumption is that the input tensor is 2-d. If the input is BxTxD, this assumption causes an issue. While I think this is OK, and the (BxT -> BT) flattening can be left to the user, there is a simple workaround by using `chunk`:

`nonlinear_part, gate = projected_input.chunk(2, dim=-1)`

Which does the same thing as lines 56 & 57, without the input dimensionality assumption. This could be a nice addition.

mttk:
As per #1211, now the Highway implementation won't require the `TimeDistributed` wrapper to work with n-dimensional input.

I've glanced around the lib and saw that `FeedForward` also has the 2-d input assumption, as well as some other layers (at least in the docs). Since `nn.Linear` accepts n-d input, I believe this can be relaxed. I've used `FeedForward` on `BxTxD` data unknowingly, and everything worked fine.

Fixes #1211.

### DIFF
--- a/allennlp/modules/highway.py
+++ b/allennlp/modules/highway.py
@@ -22,8 +22,8 @@ class Highway(torch.nn.Module):
     Parameters
     ----------
     input_dim : ``int``
-        The dimensionality of :math:`x`.  We assume the input has shape ``(batch_size,
-        input_dim)``.
+        The dimensionality of :math:`x`.  We assume the input has shape ``(batch_size, *,
+        input_dim)``, where * is any number of dimensions.
     num_layers : ``int``, optional (default=``1``)
         The number of highway layers to apply to the input.
     activation : ``Callable[[torch.Tensor], torch.Tensor]``, optional (default=``torch.nn.functional.relu``)
--- a/allennlp/modules/highway.py
+++ b/allennlp/modules/highway.py
@@ -22,8 +22,8 @@ class Highway(torch.nn.Module):
     Parameters
     ----------
     input_dim : ``int``
-        The dimensionality of :math:`x`.  We assume the input has shape ``(batch_size, *,
-        input_dim)``, where * is any number of dimensions.
+        The dimensionality of :math:`x`.  We assume the input has shape ``(batch_size, ...,
+        input_dim)``.
     num_layers : ``int``, optional (default=``1``)
         The number of highway layers to apply to the input.
     activation : ``Callable[[torch.Tensor], torch.Tensor]``, optional (default=``torch.nn.functional.relu``)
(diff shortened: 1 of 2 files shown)


allennlp/modules/highway.py
```diff
@@ -53,8 +53,7 @@ def forward(self, inputs: torch.Tensor) -> torch.Tensor:  # pylint: disable=argu
             linear_part = current_input
             # NOTE: if you modify this, think about whether you should modify the initialization
             # above, too.
-            nonlinear_part = projected_input[:, (0 * self._input_dim):(1 * self._input_dim)]
-            gate = projected_input[:, (1 * self._input_dim):(2 * self._input_dim)]
+            nonlinear_part, gate = projected_input.chunk(2, dim=-1)
```
murphp15: Very nice.

murphp15:
@mttk In the class feedforward where is the 2d input assumption (enforced ) .



I wrote a small test in feedforward_test.py for a 3d tensor and it seems to be working fine. 

`
    def test_forward_gives_correct_output_for_3d_tensor(self):
        params = Params({
            'input_dim': 2,
            'hidden_dims': 3,
            'activations': 'relu',
            'num_layers': 2
        })
        feedforward = FeedForward.from_params(params)

        constant_init = lambda tensor: torch.nn.init.constant(tensor, 1.)
        initializer = InitializerApplicator([(".*", constant_init)])
        initializer(feedforward)

        input_tensor = Variable(torch.FloatTensor([[[-3, 1]]]))
        output = feedforward(input_tensor).data.numpy()
        assert output.shape == (1, 1, 3)
        # This output was checked by hand - ReLU makes output after first hidden layer [0, 0, 0],
        # which then gets a bias added in the second layer to be [1, 1, 1].
        assert_almost_equal(output, [[[1, 1, 1]]])
`

mttk:
@murphp15 it's not enforced, but it's mentioned to be assumed in the docs https://github.com/allenai/allennlp/blob/master/allennlp/modules/feedforward.py#L21.

allennlp/modules/highway.py
```diff
@@ -22,8 +22,8 @@ class Highway(torch.nn.Module):
     Parameters
     ----------
     input_dim : ``int``
-        The dimensionality of :math:`x`.  We assume the input has shape ``(batch_size,
-        input_dim)``.
+        The dimensionality of :math:`x`.  We assume the input has shape ``(batch_size, *,
+        input_dim)``, where * is any number of dimensions.
```
matt-gardner: sphinx won't like the `*` here.  We've typically just use `...` to indicate some arbitrary number of dimensions.

mttk: Sure, figured that would be changed, thanks!

mttk:
@matt-gardner I didn't yet, I'm chasing EMNLP and I decided to remove the PR from my todo list while waiting for some results :) 
I'll do it when I get an hour of time to set up a reliable runtime shootout

mttk:
Thanks, same to you!

PR #1245, opened 2018-05-18
Load pre-trained word vectors also from zipped and plain text files

janLuke:
This is accomplished by:

* adding the ``CompressedFileUtils`` static class in ``common.file_utils``, who has a static method ``read`` for opening a compressed file in binary or text mode (with a specific encoding)

* adding a ``read_maybe_compressed_file`` function in ``common.file_utils`` which uses ``CompressedFileUtils.read``

* using ``read_maybe_compressed_file`` inside  ``embedding.py``


allennlp/common/file_utils.py
```diff
@@ -144,3 +149,95 @@ def get_from_cache(url: str, cache_dir: str = None) -> str:
...
+
+class CompressedFileUtils:
+
+    SUPPORTED_FORMATS = {'.gz', '.zip'}
+    READ_MODE_CHOICES = {'t', 'rt', 'b', 'rb'}
+    DEFAULT_ENCODING = 'utf-8'
+
+    @staticmethod
+    def read(path: str, mode: str = 't', encoding: str = None,
+             file_format: str = None):
```
joelgrus: I don't like this name, usually `read()` takes a file handle and returns its contents. As I understand it, this code returns a file handle. Accordingly, shouldn't it be called `open()`?

Also, can you add a return type for it?

allennlp/common/file_utils.py
```diff
@@ -144,3 +149,95 @@ def get_from_cache(url: str, cache_dir: str = None) -> str:
...
+        if mode == 't':
+            encoding = encoding or CompressedFileUtils.DEFAULT_ENCODING
+            return io.TextIOWrapper(binary_reader, encoding=encoding)
+        else:
+            if encoding is not None:    # same behavior of built-in open()
+                raise ValueError("Binary mode doesn't take an encoding argument")
+            return binary_reader
+
+
+def read_maybe_compressed_file(path: str, mode: str = 't', encoding: str = None,
```
joelgrus: same comment, I don't think this function should be called `read`, in one of its cases it directly calls `open`

allennlp/modules/token_embedders/embedding.py
```diff
@@ -229,9 +240,19 @@ def _read_pretrained_word2vec_format_embedding_file(embeddings_filename: str, #
 
     # First we read the embeddings from the file, only keeping vectors for the words we need.
     logger.info("Reading embeddings from file")
-    with gzip.open(cached_path(embeddings_filename), 'rb') as embeddings_file:
+
+    def read_embedding_file(url_or_path):
```
joelgrus: same comment, plus I'm not sure it's worth making this a function if it's only going to get called once immediately

allennlp/common/file_utils.py
janLuke: I called it ``read`` because ``open`` suggests it supports also writing mode. I was only interested in reading a compressed file, so I implemented only that. But actually I could make it support even writing modes with no effort and call it ``open``. I'll do that. Thank you

joelgrus: IMO it's fine to call it `open` and have it not support writing mode.

janLuke: Fine, I'll just rename it then. 

It's not clear to me what should I put as return type. Maybe ``io.IOBase``?

allennlp/modules/token_embedders/embedding.py
janLuke: I defined it mainly for readability. The call to ``open_maybe_compressed_file`` is pretty long, so I'd either do this: 
```
with open_maybe_compressed_file(local_path, mode='t',
                                encoding=_WORD2VEC_ENCODING,
                                file_format=file_format) as embedding_file:
    for line in embedding_file:
```
or this:
```
embedding_file =  open_maybe_compressed_file(local_path, mode='t',
                                             encoding=_WORD2VEC_ENCODING,
                                             file_format=file_format)
with embedding_file:
    for line in embedding_file:
```

Furthermore, the function argument allowed me to clarify that ``embedding_filename`` is not a filename but a URL or a path to the file. About it, I think ``embedding_filename`` should be renamed in the entire module.

Anyway, I can change if you prefer.

joelgrus: yeah, the former was what I was imagining.

janLuke:
@joelgrus @matt-gardner  @schmmd Can someone review this, please?

**Question.** I saw that some pretrained embeddings (e.g. Glove) are distributed in zip archives containing multiple files (one for each size). So, in my local repo, rather than forcing an archive to contain a single file, I add the possibility to select a specific file inside an archive specifying a ``file_inside_archive`` argument/config (if you have a better name for this, please share!). Example:
```
"text_field_embedder": {
      "tokens": {
        "type": "embedding",
        "pretrained_file": "path/to/archive/glove.twitter.27B.zip", 
        "file_inside_archive": "glove.twitter.27B.200d.txt"          
        ...
      }
    },
```
If the archive contain a single file, then ``file_inside_archive`` can be left to ``None``. Do you think I could add this feature here?

joelgrus:
sorry, my bad on the review.

if I were doing what you did, I would probably have `pretrained_file` point to the file inside the archive and add an optional `archive_file` above it.

i.e.

```
        "archive_file": "path/to/archive/glove.twitter.27B.zip", 
        "pretrained_file": "glove.twitter.27B.200d.txt"  
```

and then explain that `pretrained_file` looks inside the archive if provided; otherwise, it looks for the file.

if you make that change here, I'll review it in a more timely manner. :)

janLuke:
@joelgrus Sorry for the delay. You solution looks clean, but I don't have a strong preference for one or the other. 

Anyway, I don't have time today for changing what I had already written. If you and/or others have a strong preference for your proposed solution I can modify the code when I have time (hopefully tomorrow). For now, I'll commit the current version that uses a ``path_inside_archive`` argument.

janLuke:
I was unhappy with how I structured the code so I basically re-wrote this PR in a simpler way (though supporting more formats) without putting any "generic code" in ``file_utils.py`` (that code would be probably useless elsewhere).

So, I'm going to close this PR and open a new one very soon. Not sure this is the right thing to do here (sorry, for that: I'm still learning), but I think it's reasonable given that I re-wrote the new code almost from scratch.

Issue #1237, opened 2018-05-17
Make check for best model (and early stopping) use >, not >=


matt-gardner:
The way we're using `max` and `min` effectively does a `>=` check for whether current epoch is the best epoch, and we have that check in two different places (one for saving model weights, and one for early stopping).  We should use `>` instead of `>=`, and we should have only one check.

Benefits:
1. If two epochs get the same value for the validation metric, the earlier one is less likely to be overfit, and this change will make us keep the earlier model instead of the later model.
2. If you max out your validation metric, or just stop improving, you currently won't hit our early stopping criterion, because of the `>=`.  This will make you actually stop early.
3. Using one check instead of two simplifies the code.

PR #1246, opened 2018-05-18
A new model is only considered the best if it is better than others and not simply better or equal

murphp15:
Check that a model is actually an improvement on the current best and not just equal to it.

fixes : #1237

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -526,11 +526,11 @@ class Trainer:
             # Pylint can't figure out that in this branch `self._patience` is an int.
             # pylint: disable=invalid-unary-operand-type
 
-            # Is the best score in the past N epochs worse than the best score overall?
+            # Is the best score in the past N epochs worse than or equal the best score overall?
             if self._validation_metric_decreases:
-                return min(metric_history[-self._patience:]) > min(metric_history)
+                return min(metric_history[-self._patience:]) >= min(metric_history)
             else:
-                return max(metric_history[-self._patience:]) < max(metric_history)
+                return max(metric_history[-self._patience:]) <= max(metric_history)
 
         return False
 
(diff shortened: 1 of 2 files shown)


joelgrus:
I don't think this addresses the "early stopping" part of the issue?

matt-gardner:
@murphp15, I'd recommend starting with a failing test.  That is, construct a test that fails with current code (e.g., where the validation metric plateaus, but we still keep going), then change the code to make it pass.  That way we're sure that we've found and fixed the problem.

allennlp/training/trainer.py
```diff
@@ -507,11 +507,11 @@ def _should_stop_early(self, metric_history: List[float]) -> bool:
         uses patience and the validation metric to determine if training should stop early
         """
         if len(metric_history) > self._patience:
-            # Is the best score in the past N epochs worse than the best score overall?
+            # Is the best score in the past N epochs worse than or equal the best score overall?
             if self._validation_metric_decreases:
-                return min(metric_history[-self._patience:]) > min(metric_history)
+                return min(metric_history[-self._patience:]) >= min(metric_history[:-self._patience])
```
janLuke: Regarding https://github.com/allenai/allennlp/issues/1267, for ``patience == 0`` the proposed code crashes: it takes the min of the empty list ``metric_history[:-0]``.

murphp15: From a functional point of view what would patience=0 mean?
intuitively it would read as "one of the last 0 entries needs to be better than all historical results?"  This obviously doesn't make any sense. 
I don't think ppl should be setting patience to 0 ever. 
Do you agree? 
I will add a guard in the constructor to protect against this.

murphp15: Thanks! I'm going to wait for your PR to get merged and then I will pick this back up.

janLuke: Actually, I hadn't in mind to open a PR :D but, well, it wasn't  a lot of work so I did: https://github.com/allenai/allennlp/pull/1271

murphp15:
@janLuke @matt-gardner  Ready for review.

allennlp/training/trainer.py
```diff
@@ -526,11 +526,11 @@ def _should_stop_early(self, metric_history: List[float]) -> bool:
             # Pylint can't figure out that in this branch `self._patience` is an int.
             # pylint: disable=invalid-unary-operand-type
 
-            # Is the best score in the past N epochs worse than the best score overall?
+            # Is the best score in the past N epochs worse than or equal the best score overall?
             if self._validation_metric_decreases:
-                return min(metric_history[-self._patience:]) > min(metric_history)
```
janLuke: Actually, it's not necessary to compute the min/max of ``history[:-patience]``: if the best value is not in ``history[-patience:]``, then it must be the element ``history[-patience - 1]``, otherwise we should have stopped at the end of the previous epoch.

Of course, this is a negligible optimization and I guess you can leave the code as it is.

allennlp/training/trainer.py
```diff
@@ -526,11 +526,11 @@ def _should_stop_early(self, metric_history: List[float]) -> bool:
...
+                return min(metric_history[-self._patience:]) >= min(metric_history[:-self._patience])
```
janLuke: Actually, it's not necessary to compute the min/max of ``history[:-patience]``: if the best value is not in ``history[-patience:]``, then it must be the element ``history[-patience - 1]``, otherwise we should have stopped at the end of the previous epoch.

Of course, this is a negligible optimization and I guess you can leave the code as it is.

PR #1253, opened 2018-05-20
Add ELMo to the Biattentive Classification Network

nelson-liu:
Continuation of https://github.com/allenai/allennlp/pull/1038#issuecomment-377021772 .

~~I made a slight API change in changing the requirement that `text field embedder keys == token_indexers keys` to just requiring that `text field embedder keys` is a subset of `token_indexers keys`.~~


allennlp/modules/text_field_embedders/basic_text_field_embedder.py
```diff
@@ -36,12 +36,13 @@ def get_output_dim(self) -> int:
         return output_dim
 
     def forward(self, text_field_input: Dict[str, torch.Tensor], num_wrapping_dims: int = 0) -> torch.Tensor:
-        if self._token_embedders.keys() != text_field_input.keys():
-            message = "Mismatched token keys: %s and %s" % (str(self._token_embedders.keys()),
-                                                            str(text_field_input.keys()))
+        if not set(self._token_embedders.keys()).issubset(set(text_field_input.keys())):
```
matt-gardner: Does it not work without this change?  I'd prefer matching the set exactly, mainly to help avoid user error in writing the configuration file.  But, if others also think this is a good change, or if it's necessary to use ELMo in advanced ways, we can change it.

nelson-liu: so the ELMo object requires the "elmo" key in the token indexer, and the previous behavior mandated that you must also have an "elmo" key in the embedder as well.

However, you might not want to use the input ELMo embeddings and use only ELMo embeddings at the 2nd LSTM's output --- this lets you do that.

I agree that this isn't ideal from an API perspective, but unsure if there's another good solution? Would be all ears to suggestions.

DeNeutoy: A better way - include the `token_indexer` for `elmo_characters`, but pop it off the dictionary in `forward`. Then have a flag like `use_elmo` which dictates whether you actually use it or not, passing it directly to an `Elmo()` instance with `num_output_representations=2`. This way of using elmo doesn't really fit the API, but I agree with Matt that it would be better not to dilute how `textfield_embedders` are used, because they are already kind of confusing for new users.

nelson-liu: @DeNeutoy : i like your way better as well, thanks!

It might be a bit confusing to the user if they include elmo in their embedder and then see that it doesn't get used, but I think this should be ok with proper documentation / raising an informative error message if elmo is used in the embedder...

nelson-liu:
hmm, not quite sure why this is failing. It looks like indexing the dataset with the two models produces different results, despite the models being the exact same...

nelson-liu:
spoke too soon, looks like the culprit is that the tests don't like when a model modifies the batch in-place during `forward`...

matt-gardner:
Oh, yeah, I forgot about that.  `pop` is no good, it'll break our assumptions about not needing to re-index things.

DeNeutoy:
We have several models which violate some of the assumptions in `ModelTestCase` (e.g the coref model produces different outputs if you batch things together vs individually, so we don't check this case) - just don't call that test for these models. Write another one which checks that forward does what you expect instead, and leave the BCN model tests not using elmo.

nelson-liu:
this is ready to be looked at, but i think the elmo version needs a bit more hyperparameter tuning...

allennlp/models/biattentive_classification_network.py
```diff
@@ -218,8 +312,16 @@ def from_params(cls, vocab: Vocabulary, params: Params) -> 'BiattentiveClassific
             output_layer = FeedForward.from_params(output_layer_params)
         else:
             output_layer = Maxout.from_params(output_layer_params)
+
+        elmo = params.pop("elmo", None)
+        if elmo is not None:
+            elmo = Elmo.from_params(elmo)
+        use_input_elmo = params.pop("use_input_elmo", False)
```
DeNeutoy: `pop_bool` here?

training_config/biattentive_classification_network.json
```diff
@@ -12,6 +12,7 @@
   "train_data_path": "https://s3-us-west-2.amazonaws.com/allennlp/datasets/sst/train.txt",
   "validation_data_path": "https://s3-us-west-2.amazonaws.com/allennlp/datasets/sst/dev.txt",
   "test_data_path": "https://s3-us-west-2.amazonaws.com/allennlp/datasets/sst/test.txt",
+  "evaluate_on_test": true,
```
DeNeutoy: remove this, it shouldn't be true by default in the config.

training_config/biattentive_classification_network_elmo.json
```diff
@@ -0,0 +1,98 @@
...
+  },
+  "validation_dataset_reader":{
+    "type": "sst_tokens",
+    "use_subtrees": false,
+    "granularity": "5-class",
+    "token_indexers": {
+      "tokens": {
+        "type": "single_id"
+      },
+      "elmo": {
```
DeNeutoy: Given this is HOCON, can you add a descriptive comment about why this doesn't break everything, like you have done in the testing configs?

DeNeutoy:
@nelson-liu can you rebase off master for pytorch 0.4? We can merge it before you reproduce the result, i'm all for incremental improvements.

nelson-liu:
sure, continued in #1308 .

apeterswu:
Do we have anyway to better fine tune the model so that we can match the performance reported in the paper?

nelson-liu:
i didn't really do any hyperparameter tuning for the elmo+bcn setting --- i think that would help considerably.

apeterswu:
I have run the exp again, and the elmo performance only matches 51.44, almost same to the baseline without elmo. It's wired that we can't reproduce the performance. 
Thanks for your quick response. Hope to see a stable training and stable performance setting.

PR #1254, opened 2018-05-21
make evalb use unique temporary directories per process

DeNeutoy:


allennlp/training/metrics/evalb_bracketing_scorer.py
```diff
@@ -97,7 +98,7 @@ def __call__(self, predicted_trees: List[Tree], gold_trees: List[Tree]) -> None:
                     self._correct_predicted_brackets += numeric_line[5]
                     self._gold_brackets += numeric_line[6]
                     self._predicted_brackets += numeric_line[7]
-
+        shutil.rmtree(tempdir)
```
matt-gardner: Blank lines are off here.

murphp15:
What problem is this fixing?

DeNeutoy:
`gettempdir` retrieves a global temporary directory which is created if it doesn't exist, but just returned if it does. This means that if you have two models using this metric on a single machine, the tempdir file might be shared, which makes bad things happen, because the filenames aren't unique inside the directory, meaning things get overwritten.

PR #1256, opened 2018-05-21
import Serve lazily to prevent unnecesary dependencies for using alle…

DeNeutoy:
…nnlp as a library


matt-gardner:
This seems ok to me, though I also wonder if we should just get rid of `serve` entirely, or change it to only use the simple server, and just have a separate path for running our demo.  Do you think people actually want to run the full demo using `allennlp serve`, or just a demo of their own model?

schmmd:
We're going to build a better solution, potentially removing serve from the command line.

Issue #1163, opened 2018-04-30
The output from ElmoEmbedder's embed_file conflicts with h5py naming convention


gabrielStanovsky:
If I understand correctly, [```embed_file```](https://github.com/allenai/allennlp/blob/master/allennlp/commands/elmo.py#L296) normally stores an h5py Dataset for each sentence in the input file. 
However, if the input sentence contains a forward slash, it will instead store an h5py **Group** object, collecting all sentences with a common prefix. 
See h5py documentation [here](https://support.hdfgroup.org/HDF5/doc1.6/UG/09_Groups.html).

For instance, embedding the following raw sentences will store one Group (containing the embeddings of the first two sentences) and one Dataset object (containing the embedding for the first): 

```
test / this is a first sentence
test / this is a second sentence
this is a third sentence
```

This is probably not the expected behavior?
Can escaping the forward slash character be a good solution?
So far, I couldn't find how to do that nicely in h5py.

@schmmd

matt-peters:
Has this been a problem for you in practice?  I've noticed this previously but have ignored it as it isn't a practical problem (e.g. h5py is tolerant of these slashes when loading and has the desired behavior).

matt-peters:
I usually iterate over the original text file with the complete sentences, then select them from the hdf5 file to work around this.  Something like

```
with h5py.File(embedding_file, 'r') as fh5, open(text_file, 'r') as ftext:
    for line in ftext:
        embeddings = fh5[line.strip()]
```

It's a little annoying but not terribly.  Agree this could be cleaned up if there is an easy way to do it.

gabrielStanovsky:
Thanks!
A small downside of this workaround IMO is that requires having the raw file, as well as more I\O operations when retrieving the sentences.

schmmd:
I agree that ideally you wouldn't need the raw file, although I'm not sure of the best fix presently.

tuvuumass:
Should we add <S> and </S> to sentences in the input file?

schmmd:
@gabrielStanovsky would @DeNeutoy 's suggestion work for you?  I like having the results indexed by the sentence text, but the current behavior and interpretation for `/`s is pretty obnoxious.

matt-peters:
Thanks for the code snippet!   If we added that to the library we could maintain the current format, and reference it as the supported method to read it the embeddings without the original raw text file.

schmmd:
@matt-peters suggestion is my favorite, given where we currently are.

[...]

PR #1257, opened 2018-05-21
Add a method to read h5py files and handle slashes.

schmmd:
Fixes https://github.com/allenai/allennlp/issues/1163


schmmd:
@gabrielStanovsky @matt-peters I'm conflicted as to whether this really is the right way to address the problem in the issue after writing up this solution.  I'm curious what you two think.

DeNeutoy:
I don't understand why we don't just use an integer index and keep the original file around. This is how I use the dump embeddings function myself, in practice, by just changing the source.

schmmd:
@DeNeutoy is not approving this PR because he doesn't like the approach.  I don't have time to write up an alternate solution in the near future, so I'm un-assigning myself from the original issue.

PR #1260, opened 2018-05-22
Tutorial cleanup

schmmd:


joelgrus:
In the grand scheme of things I think we should deprecate the crf_tagger tutorial and replace it with a polished version of the AcademicPaperClassifier tutorial

matt-gardner:
Why remove the laziness how-to?

matt-gardner:
But that's kind of the point - a "how-to" is a "please just tell me how to do this" document.  The tutorial is "teach me about the library", with explanations for things.  They accomplish different purposes.

PR #1264, opened 2018-05-22
Remove most mentions of python -m allennlp.run

### LINKED ISSUE #1263: `allennlp` usage lists `python -m allennlp.run` as the command
this annoys me all the time, but I keep forgetting to open an issue

```
$ allennlp
usage: python -m allennlp.run

Run AllenNLP

optional arguments:
  -h, --help  show this help message and exit

Commands:

    train     Train a model
    evaluate  Evaluate the specified model + dataset
    predict   Use a trained model to make predictions.
    serve     Run the web service and demo.
    make-vocab
              Create a vocabulary
    elmo      Use a trained model to make predictions.
    fine-tune
              Continue training a model on a new dataset
    dry-run   Create a vocabulary, compute dataset statistics and other
              training utilities.
```

this is because it's specified explicitly as "python -m allennlp.run" in allennlp/run.py

easy solution: specify it as "allennlp" instead (but then it will print the wrong thing when you actually do call `python -m allennlp.run`

easy but ugly solution: create a separate run script for the executable `allennlp/allennlp.py` that's identical but for the executable name

still easy but less easy solution: allow some kind of override for the name (e.g. via an environment variable) and then change setup.py to make the command specify that environment variable

something else: ?

schmmd:
Fixes #1263


schmmd:
@joelgrus I'm conflicted as to whether I should add `./bin/allennlp` to the PATH in the docker image.  Presently the executable is path dependent if you are running the source code.  I could make it so it's not path dependent, but that may (?) have implications for when it's installed via pip.

joelgrus:
if you're going to do this, can you also change it here and fix my issue :)

https://github.com/allenai/allennlp/blob/master/allennlp/run.py#L18

I think telling people to run `bin/allennlp` is fine

matt-gardner:
What happens if you just remove the `prog` argument to the argument parser?  Does it do the right thing?

joelgrus:
it defaults to the filename:

```
$ python -m allennlp.run
usage: run.py [-h]  ...
```

PR #1270, opened 2018-05-23
[WIP] Optimized action handling for faster training time.

rajasagashe:
@matt-gardner 
This significantly speeds up get_actions_to_consider, compute_new_states, and  get_action_embeddings during training time. The initialization of embedded_actions with the repeated list operation in get_actions_to_consider_optimized speeds up get_action_embeddings since it reduces the time for the large padding operation in get_action_embeddings. 
The downside is that this duplicates code.
I can clean this up if you think it's useful, it currently won't run since the wikitables implementation has changed.


matt-gardner:
Thanks @rajasagashe!  I was just about to start looking at cleaning up the action code, and this is really helpful to look at!

matt-gardner:
Also, at some point it'd be nice to get your re-implementation of Srini's model, and data loading code, into the repo.  I'm hoping to unify a bunch of the model code between the parsers (e.g., the `DecoderStep` is probably general enough that it can just be shared), and having another model in the mix to understand what can and can't be shared would be great.  Same with your model and data, @OyvindTafjord.

matt-gardner:
Thanks again @rajasagashe, I got some useful stuff from this for #1294.  I'm closing this now.

Issue #1267, opened 2018-05-23
[Minor] Early stopping for patience == 0


janLuke:
Should 0 be a valid value for ``patience``? It looks like a reasonable value for expressing "no patience".

* If yes, I think it should be equivalent to ``patience == 1``: the training should be stopped at the end of the first epoch without improvement. 
* if no, maybe it's the case to add a check ``patience > 0`` in the constructor.

Currently, for ``patience == 0``, the stop condition is never met because in ``Trainer._should_stop_early``:
```
if self._validation_metric_decreases:
    return min(metric_history[-self._patience:]) > min(metric_history)
else:
    return max(metric_history[-self._patience:]) < max(metric_history)
```
 ``metric_history[-0:] == metric_history`` so the inequality between their min/max is always false.

schmmd:
The current behavior (`patience = 0` means the training will never stop early due to patience) makes sense to me.  Otherwise, how would you disable patience if you wanted to?

janLuke:
@schmmd Negative patience, for example.

nelson-liu:
to play devil's advocate, I think `patience = 0` could just as easily be construed as "no patience" / "impatient", and that you'd stop as soon as you see an epoch where the metric doesn't improve.

fwiw, i think `patience=0` has the above behavior in Keras, but it's a bit different from an API design perspective since there you have to explicitly add an `EarlyStoppingCallback`, and thus it doesn't make sense to have a value that disables patience when you've already explicitly added this callback.

schmmd:
I would find it confusing if `patience == 0` were the same as `patience == 1`, so maybe a reasonable proposal is that `patience` cannot be 0 and is disabled with -1?

janLuke:
@nelson-liu I made exactly the same reasoning about Keras API. Still, I think it's simply counterintuitive that ``patience = 0`` disables  early stopping and the fact that other libraries behave differently only add to that. Furthermore, there's not a strict need to reserve zero for that function.

@schmmd Yes, I find that confusing too and your proposal sounds reasonable to me. If I had to implement early-stopping from scratch, I'd define ``patience`` as the max number of (consecutive) epochs with no improvement one can tolerate **without** stopping the training, so that the training is stopped after ``patience + 1`` unsuccessful epochs. In other words, it'd be the max number of consecutive postponements of early stopping. This way:
* ``patience == 0`` means what it suggests: no patience, stop at the first epoch with no improvement;
[...]

PR #1271, opened 2018-05-24
In Trainer, require "patience" to be a positive integer or None (meaning "no early stopping"). Set patience=None by default.

janLuke:
Fixes https://github.com/allenai/allennlp/issues/1267

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -152,7 +152,7 @@ class Trainer:
                  iterator: DataIterator,
                  train_dataset: Iterable[Instance],
                  validation_dataset: Optional[Iterable[Instance]] = None,
-                 patience: Optional[int] = 2,
+                 patience: Optional[int] = None,
                  validation_metric: str = "-loss",
                  num_epochs: int = 20,
                  serialization_dir: Optional[str] = None,
@@ -181,12 +181,10 @@ class Trainer:
             A ``Dataset`` to train on. The dataset should have already been indexed.
         validation_dataset : ``Dataset``, optional, (default = None).
             A ``Dataset`` to evaluate on. The dataset should have already been indexed.
-        patience : int != 0, optional (default=2)
-            If ``patience > 0``, it's the number of epochs to be patient before early stopping:
-            the training is stopped after ``patience`` epochs with no improvement.
-            If ``patience < 0 or patience == None``, early stopping is disabled.
-            Because the behavior for ``patience = 0`` is undefined and prone to misinterpretations,
-            it's required ``patience != 0``.
+        patience : Optional[int] > 0, optional (default=None)
+            Number of epochs to be patient before early stopping: the training is stopped
+            after ``patience`` epochs with no improvement. If given, it must be ``> 0``.
+            If None, early stopping is disabled.
         validation_metric : str, optional (default="loss")
             Validation metric to measure for whether to stop training using patience
             and whether to serialize an ``is_best`` model each epoch. The metric name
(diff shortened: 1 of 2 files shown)


murphp15:
Looks good!

allennlp/training/trainer.py
```diff
@@ -240,7 +247,12 @@ def __init__(self,
         self._train_data = train_dataset
         self._validation_data = validation_dataset
 
-        self._patience = patience
+        if patience == 0:
```
schmmd: Looks great overall, but I think it would be an improvement to require `patience > 0` and accept only `None` to disable it.  Otherwise we'll get the question "What is `patience == -2"!

schmmd:
Looks great!  Just one small comment.

allennlp/training/trainer.py
```diff
@@ -178,8 +181,12 @@ def __init__(self,
             A ``Dataset`` to train on. The dataset should have already been indexed.
         validation_dataset : ``Dataset``, optional, (default = None).
             A ``Dataset`` to evaluate on. The dataset should have already been indexed.
-        patience : int, optional (default=2)
-            Number of epochs to be patient before early stopping.
+        patience : int != 0, optional (default=2)
+            If ``patience > 0``, it's the number of epochs to be patient before early stopping:
+            the training is stopped after ``patience`` epochs with no improvement.
+            If ``patience < 0 or patience == None``, early stopping is disabled.
```
joelgrus: in `from_params` the default value is 2 if not specified, so how would patience be `None`?  I think that could only be the case if people explicitly put `"patience": null` in the config file? I don't know that I like the idea of having nulls in the config file

matt-gardner:
No strong opinion from me either way, but if the default value in `from_params` is `None`, you just have no patience if you don't add the flag, which seems reasonable.

allennlp/training/trainer.py
janLuke: @schmmd I don't think that's a problem. From the doc it's pretty clear what happens for negative patience.

matt-gardner:
@janLuke, I didn't realize the default was also in the constructor; yes, these should definitely match.  But it could be `None` in the constructor, too.  Again, I don't have a strong opinion about what the right thing to do here is, but I'd lean towards agreeing with @schmmd and @matt-peters that `patience` should only take positive integers, and that its default value should be `None`, meaning we don't do early stopping at all.  Then you don't need to bother with a check for negative integers, because you can just leave out the flag and you don't stop early.

matt-gardner:
You specify "no early stopping" by not adding the flag that enables early stopping.  And, yeah, if you want to do it with overrides for some reason, you just specify a large value, you [use the key "None"](https://github.com/allenai/allennlp/blob/b4c9fe4143cc0ec1bafb0b58fc63a9e3c5655bd1/allennlp/common/params.py#L311-L317).

nelson-liu:
speaking of which, it'd be nice if that none check was a bit looser / well documented. It's a bit confusing since the file isn't strictly json (despite its file extension) and it's not totally transparent to a user whether the right thing to do is `None`, `null`, `none`, `"None"`, etc...

matt-gardner:
Not to assign blame or anything, but... https://github.com/allenai/deep_qa/blame/911194eabe27ca0a378189d914c61eba0b5dcf83/src/main/python/run_model.py#L47-L54

;)

nelson-liu:
exactly why i remembered to bring it up 😄  this is a separate issue, but i'd be happy to help out with better documenting it / modifying its behavior if we can settle on how to do it.

janLuke:
Sorry, for some reason, I assumed that null values were problematic in configuration files and thus we needed negative values to disable early stopping. I think now it's cleaner and hopefully ready to be merged. Over and out :)

tests/training/trainer_test.py
```diff
@@ -122,6 +123,28 @@ def test_should_stop_early_with_decreasing_metric(self):
...
+        # Decreasing metric
+        trainer = Trainer(self.model, self.optimizer, self.iterator, self.instances,
+                          validation_dataset=self.instances, num_epochs=100,
+                          patience=None, validation_metric="-test")
+        increasing_history = [float(i) for i in range(20)]
+        assert not trainer._should_stop_early(increasing_history)  # pylint: disable=protected-access
+
+    def test_should_stop_early_with_invalid_patience(self):
+        for patience in [0, -1, -2, 1.5, 'None']:
+            with pytest.raises(ConfigurationError, message='No ConfigurationError for patience={}'.format(patience)):
```
matt-gardner: This message seems wrong - shouldn't you be checking for "invalid value"?

janLuke: The default text message printed when the test fails is: 
``"Failed: DID NOT RAISE <class 'allennlp.common.checks.ConfigurationError'>"``
With mine would print:
``Failed: No ConfigurationError for patience=0``
I could replace "No" with "Did not raise" if you prefer.

matt-gardner: Ah, I was thinking of the `match` parameter to `pytest.raises`, not the `message` parameter.  Thanks =).  I think I've been using this wrong in other places in the code, too...

PR #1283, opened 2018-05-25
Use full version in docs.

schmmd:
False alarm on the docs not being deployed for previous releases.  I was confused by the version number reported in the UI.  This change will cause the docs to show the full version number.


joelgrus:
I don't understand this change. There's a long version ("release") and a short version ("version"), and you're changing it so that they're both equal to the long version?

Isn't the "right" fix to change the sphinx config to use the other one?

joelgrus:
oops, never mind, I misunderstood

schmmd:
@joelgrus I wasn't sure if the comments before were even accurate.  This addresses the problem I wanted to fix--but there might be another way.  I couldn't figure another way too quickly however--and I really don't ever want Sphynx to show the short version number.  In fact--I think the only reason we have `VERSION_SHORT` is for Sphynx--should I just remove it entirely?

joelgrus:
nah, I can imagine someone wanting to use it in theory, I think this way is fine

PR #1294, opened 2018-05-27
[WIP] Refactoring how actions are handled in the semantic parsing code

matt-gardner:
This is still in progress, but it should dramatically speed up runtimes and greatly simplify the code.  I've got the initial test passing for the `DecoderStep`.

Still to do:
- [x] Replace action indices in `get_valid_actions` with action representations.  This means they're only computed once, in `forward`, and not during every step of decoding.
- [x] Remove batching from the action selection logic.  This makes it so we don't have to worry about padding things, and simplifies the logic considerably.  It also means that we're not creating new, padded variables on the GPU at every step of decoding, just so we can do a few softmaxes.  I'm guessing this might remove the bottleneck for batching things together for wikitables, and might make it feasible to increase the batch size (we just unroll the batch during certain parts of decoding).  We'll see.
- [x] Remove the sorting operation when were just doing vanilla MML, without a constrained beam search.  The sorting is expensive, and if we're going to be evaluating the entire set of allowed actions, and only those actions, there is no need to sort.  Thanks @rajasagashe for this tip, taken from #1270.
- [ ] Add back in the checklist stuff that I removed.  I wanted to get the basic logic working first; it shouldn't be too bad to add it back.
- [ ] Put the action representation computation into `forward`; I removed it from `DecoderStep`, but haven't yet put it back in its new spot.
- [ ] Use this to actually train the wikitables model, and see how well it works.
- [ ] Do this for NLVR also.  Though, that'll probably wait for another PR, and I'll probably try to just use the same `DecoderState` object for both parsers.  We'll see.


allennlp/models/semantic_parsing/wikitables/wikitables_semantic_parser.py
```diff
@@ -523,133 +517,93 @@ def get_metrics(self, reset: bool = False) -> Dict[str, float]:
...
+            translated_valid_actions[key]['global'] = (global_input_embeddings,
+                                                       global_output_embeddings,
+                                                       list(global_action_ids))
+
+            # Then the representations of the linked actions.
+            if linked_actions:
+                linked_rules, linked_action_ids = zip(*linked_actions)
+                entities = [rule.split(' -> ')[1] for rule in linked_rules]
+                entity_ids = [entity_map[entity] for entity in entities]
+                entity_linking_scores = linking_scores[entity_ids]
```
pdasigi: The shape of this tensor is `(num_entities, num_question_tokens)`, correct?

matt-gardner:
Ok, @pdasigi, this is ready to look at.  Sorry it's such a monstrosity at this point.  I think it's a whole lot cleaner than it was before, though.  The main points:

1. I removed batching from the action selection computation.  This dramatically simplifies the logic, because there's no more global action id stuff, and it lets us remove a whole lot of code.  As part of this, I made the `GrammarState` hold onto more information about each action, because it was more convenient to keep in inside the `valid_actions` dictionary itself.  You can see what this looks like by looking at the test for `GrammarState`.
2. I combined the `State` and `Step` classes between both semantic parsers, because they were basically just duplicates of each other.  Now we have general classes that can be re-used across any semantic parsing model.
3. I renamed the merged `State` and `Step` classes to reflect what they actually do, and split them out into a few related classes.  Instead of `WikiTablesDecoderState`, we now have `GrammarBasedDecoderState` and `CoverageDecoderState`, where the `CoverageDecoderState` is a subclass that just adds a `ChecklistState` object.  Similarly, we have a `BasicTransitionFunction` (that's an LSTM decoder with attention), and a `LinkingTransitionFunction`, `CoverageTransitionFunction` and `LinkingCoverageTransitionFunction`, that all build on this.  The NLVR MML model just uses the `BasicTransitionFunction`, the WikiTables MML model uses the `LinkingTransitionFunction`, and the ERM models add coverage to their respective transition functions.  Most of the functionality stays in the `BasicTransitionFunction` class, though, which is nice.
4. I also cleaned up a bunch of the interface between the `SemanticParser` models and their MML and ERM subclasses.

I'm still getting a few test failures on CI that don't show up locally, and I have some mypy stuff to figure out.  But those should be small fixes, and so the code is ready to review now.

allennlp/models/semantic_parsing/wikitables/basic_transition_function.py
```diff
@@ -0,0 +1,378 @@
...
+                                 group_indices[i],
+                                 log_probs[i],
+                                 action_embeddings[i],
+                                 group_actions[i])
+                                for i in range(len(group_actions))
+                                if (not allowed_actions or
+                                    group_actions[i] in allowed_actions[group_indices[i]])]
+                # We use a key here to make sure we're not trying to compare anything on the GPU.
+                batch_states.sort(key=lambda x: x[0], reverse=True)
+                if max_actions:
```
pdasigi: Would this ever be False?

allennlp/models/semantic_parsing/wikitables/coverage_transition_function.py
```diff
@@ -0,0 +1,151 @@
...
+                 dropout: float = 0.0) -> None:
+        super().__init__(encoder_output_dim=encoder_output_dim,
+                         action_embedding_dim=action_embedding_dim,
+                         input_attention=input_attention,
+                         num_start_types=num_start_types,
+                         activation=activation,
+                         predict_start_type_separately=predict_start_type_separately,
+                         add_action_bias=add_action_bias,
+                         dropout=dropout)
+        self._checklist_multiplier = Parameter(torch.FloatTensor([1.0]))
```
pdasigi: Document what this parameter does.

allennlp/models/semantic_parsing/wikitables/wikitables_semantic_parser.py
```diff
@@ -518,133 +507,154 @@ def get_metrics(self, reset: bool = False) -> Dict[str, float]:
...
+            production_rule_arrays = [(possible_actions[index], index) for index in action_indices]
+            global_actions = []
+            linked_actions = []
+            for production_rule_array, action_index in production_rule_arrays:
+                if production_rule_array[1]:
+                    global_actions.append((production_rule_array[2], action_index))
+                else:
+                    linked_actions.append((production_rule_array[0], action_index))
+
+            # Then we get the ebmedded representations of the global actions.
```
pdasigi: embedded

allennlp/models/semantic_parsing/wikitables/wikitables_semantic_parser.py
```diff
@@ -518,133 +507,154 @@ def get_metrics(self, reset: bool = False) -> Dict[str, float]:
...
+                entity_type_embeddings = self._entity_type_decoder_embedding(entity_type_tensor)
+                translated_valid_actions[key]['linked'] = (entity_linking_scores,
+                                                           entity_type_embeddings,
+                                                           list(linked_action_ids))
+
+        # Lastly, we need to also create embedded representations of context-specific actions.  In
+        # this case, those are only variable productions, like "r -> x".
+        context_actions = {}
+        for action_id, action in enumerate(possible_actions):
+            if action[0].endswith(" -> x"):
```
pdasigi: This will break if we had nested lambdas, right? Wouldn't happen with WikiTables, but may be add a TODO?

allennlp/models/semantic_parsing/nlvr/nlvr_coverage_semantic_parser.py
```diff
@@ -211,6 +215,8 @@ def forward(self,  # type: ignore
 
         label_strings = self._get_label_strings(labels) if labels is not None else None
         # Each instance's agenda is of size (agenda_size, 1)
+        # TODO(mattg): It looks like the agenda is only ever used on the CPU.  In that case, it's a
+        # waste to copy it to the GPU and then back, and this should probably be a MetadataField.
```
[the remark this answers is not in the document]
pdasigi: Ah.. never mind, it is the `checklist_target` that's used while updating the checklist, not the `agenda`.

allennlp/common/testing/model_test_case.py
```diff
@@ -42,7 +43,8 @@ def set_up_model(self, param_file, dataset_file):
     def ensure_model_can_train_save_and_load(self,
                                              param_file: str,
                                              tolerance: float = 1e-4,
-                                             cuda_device: int = -1):
+                                             cuda_device: int = -1,
+                                             gradients_to_ignore: Set[str] = None):
```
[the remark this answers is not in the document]
matt-gardner: Done.

allennlp/models/semantic_parsing/nlvr/nlvr_coverage_semantic_parser.py
```diff
@@ -347,13 +353,14 @@ def get_metrics(self, reset: bool = False) -> Dict[str, float]:
                 'agenda_coverage': self._agenda_coverage.get_metric(reset)
         }
 
-    def _get_state_cost(self, state: NlvrDecoderState) -> torch.Tensor:
+    def _get_state_cost(self, batch_worlds: List[List[NlvrWorld]], state: CoverageDecoderState) -> torch.Tensor:
         """
         Return the costs a finished state. Since it is a finished state, the group size will be 1,
         and hence we'll return just one cost.
         """
```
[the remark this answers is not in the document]
matt-gardner: Done.

allennlp/models/semantic_parsing/nlvr/nlvr_coverage_semantic_parser.py
```diff
@@ -390,7 +399,9 @@ def _get_state_info(self, state) -> Dict[str, List]:
                             for history in state.action_history]
         agenda_sequences = []
         all_agenda_indices = []
-        for agenda, checklist_target in zip(state.terminal_actions, state.checklist_target):
+        # TODO(mattg,pradeep): This is broken after the action refactoring and needs to be either
+        # removed or updated.  Matt wasn't sure how to update it, so he just left it like this.
```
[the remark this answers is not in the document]
matt-gardner: Thanks, done.

allennlp/models/semantic_parsing/wikitables/basic_transition_function.py
```diff
@@ -0,0 +1,378 @@
...
+                                     memory_cell[group_index],
+                                     action_embedding,
+                                     attended_question[group_index],
+                                     state.rnn_state[group_index].encoder_outputs,
+                                     state.rnn_state[group_index].encoder_output_mask)
+            batch_index = state.batch_indices[group_index]
+            for i, log_probs, _, actions in batch_action_probs[batch_index]:
+                if i == group_index:
+                    considered_actions = actions
+                    probabilities = log_probs.exp().cpu()
```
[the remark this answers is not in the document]
matt-gardner: Done.

allennlp/models/semantic_parsing/wikitables/basic_transition_function.py
matt-gardner: Probably not, but I'd still rather have the check, in case some new learning algorithm wants to do something crazy.

allennlp/models/semantic_parsing/wikitables/coverage_transition_function.py
matt-gardner: Done.

allennlp/models/semantic_parsing/wikitables/wikitables_semantic_parser.py
matt-gardner: I thought I fixed that already...  Thanks.  Done.

allennlp/models/semantic_parsing/wikitables/wikitables_semantic_parser.py
matt-gardner: The `WikiTablesWorld` specifies that there is at most one nested lambda: https://github.com/allenai/allennlp/blob/75abebb437042e4709ecf4e9af770df6a4912896/allennlp/semparse/worlds/wikitables_world.py#L43

I'll add a note here, not a TODO.

Issue #1241, opened 2018-05-18
Compute loss for evaluation.


cristipp:
Running evaluation on a simple model with no metrics produces no output:
```
2018-05-18T00:16:35.634374779Z 2018-05-18 00:16:35,634 - INFO - allennlp.commands.evaluate - Iterating over dataset
2018-05-18T00:16:35.634877992Z   0%|          | 0/2 [00:00<?, ?it/s]
2018-05-18T00:16:35.817916516Z  ||: 100%|##########| 2/2 [00:00<00:00, 10.95it/s]
2018-05-18T00:16:35.818125905Z 
2018-05-18T00:16:35.818159433Z 2018-05-18 00:16:35,817 - INFO - allennlp.commands.evaluate - Finished evaluating.
2018-05-18T00:16:35.818306132Z 2018-05-18 00:16:35,818 - INFO - allennlp.commands.evaluate - Metrics:
2018-05-18T00:16:35.81840071Z 
```

Running validation over the same model produces a loss metric:
```
2018-05-18T00:15:37.750617271Z 2018-05-18 00:15:37,750 - INFO - allennlp.training.trainer - Validating
2018-05-18T00:15:37.751509165Z   0%|          | 0/2 [00:00<?, ?it/s]
2018-05-18T00:15:37.945853747Z loss: 2.3851 ||: 100%|##########| 2/2 [00:00<00:00, 10.31it/s]
2018-05-18T00:15:38.589231511Z 
```

Would be nice to have loss computed for evaluation as well.

murphp15:
Hey @cristipp I will do this ticket. 
Just one question when you say "Running validation over the same model produces a loss metric:"
Do you mean "Running **training** over the same model produces a loss metric:"?

cristipp:
Both training and validation steps display the loss: 

```
2018-05-28 07:15:00,380 - INFO - allennlp.training.trainer - Peak CPU memory usage MB: 936.419328
  0%|          | 0/2 [00:00<?, ?it/s]2018-05-28 07:15:00,397 - INFO - allennlp.training.trainer - Training
loss: 3.4540 ||: 100%|##########| 2/2 [00:02<00:00,  1.06s/it]
2018-05-28 07:15:02,527 - INFO - allennlp.training.trainer - Validating
loss: 3.3113 ||: 100%|##########| 2/2 [00:01<00:00,  1.29it/s]
```

brendan-ai2:
[...]

PR #1296, opened 2018-05-28
compute loss for evaluation.

murphp15:
Fix for https://github.com/allenai/allennlp/issues/1241


allennlp/training/metrics/tqdm_metrics_reporter.py
```diff
@@ -0,0 +1,20 @@
+from typing import Dict, Iterable
+
+
+class TqdmMetricsReporter(Iterable):
```
matt-gardner: I don't think this abstraction buys us anything.  It moves one one-line method off of `trainer` and onto this, while adding a whole new class and complicating other parts of the trainer code (because you need to know about and instantiate a new object to wrap the `tqdm`).

And in general, every time you add a class to the codebase, you add complexity - docs that need to be maintained, more things to know about when writing code, and more things to modify if we want to refactor the code later.  Sometimes that complexity is definitely worth it.  But for really simple things, like in the two abstractions you've added here, the cost in additional complexity isn't worth the benefit that you get from using the abstraction.  We need to be really careful to not just add abstractions for abstraction's sake.

schmmd:
Closing this PR as there hasn't been any follow up in months.

PR #1306, opened 2018-05-29
add user contributed models blurb

DeNeutoy:
image


DeNeutoy:
Thoughts? Super easy to add, it's our most visible platform.

joelgrus:
are we that interested in adding third-party models to the library itself? I thought when we discussed this we decided we'd rather link out to other people's repos, except in rare cases

matt-gardner:
That's a great idea!  :+1:

matt-gardner:
@joelgrus, yes, I think we'd generally want to link to other libraries, instead of forcing consistency with our codebase, but I also think we should try to host others' demos even without them contributing their code.  We need to think about how exactly to do this (e.g., we have a separate demo repo that has third-party code as submodules, and puts everything together), but I think it's something we should do.

joelgrus:
sure, I just meant that the message as written suggests that we want to add the models to the library itself

matt-gardner:
Fair enough, we could probably word it better.  But I definitely think that having a link right there on the demo with a message at least similar to what Mark has is a spectacular idea.

PR #1308, opened 2018-05-30
(Rebased) Add ELMo to BCN

nelson-liu:
Continuation of #1253 , rebased after the pytorch 0.4 PR was merged.

Opening a new PR for this, since i think the discussion in #1253 is fairly useful.


allennlp/tests/models/biattentive_classification_network_test.py
```diff
@@ -30,4 +40,76 @@ def test_mismatching_dimensions_throws_configuration_error(self):
...
+        tmp_params = deepcopy(params)
+        tmp_params["model"]["use_input_elmo"] = True
+        tmp_params["model"]["use_integrator_output_elmo"] = True
+        with pytest.raises(ConfigurationError):
+            Model.from_params(self.vocab, tmp_params.get("model"))
+
+    def test_elmo_but_no_set_flags_throws_configuration_error(self):
+        initial_working_dir = os.getcwd()
+        # Change directory to module root.
+        os.chdir(self.MODULE_ROOT)
```
nelson-liu: this is rather hacky, perhaps we should consider always changing dir to `self.MODULE_ROOT` before every test?

training_config/biattentive_classification_network_elmo.json
```diff
@@ -0,0 +1,101 @@
...
+    "type": "bcn",
+    "text_field_embedder": {
+      "tokens": {
+        "pretrained_file": "https://s3-us-west-2.amazonaws.com/allennlp/datasets/glove/glove.840B.300d.txt.gz",
+        "type": "embedding",
+        "embedding_dim": 300,
+        "trainable": false
+      }
+    },
+    # BCN configs that use ELMo should not have an elmo_token_embedder
```
DeNeutoy: Can you put this comment above the text_field_embedder and say _why_ not, rather than just stating it. Maybe "The BCN model will consume the arrays generated by the ELMo token_indexer independently of the text_field_embedder, so we miss out the key for that here".

allennlp/tests/models/biattentive_classification_network_test.py
matt-gardner: Why do we need to change the directory at all?

DeNeutoy: I don't understand why you need to change directory here at all?

nelson-liu: The check that throws the `ConfigurationError` in the test is in `__init__`, so it's run after the ELMo object is created. `Model.from_params(self.vocab, params.get("model"))` complains when building `Elmo.from_params` that the options file is not where it expects, since it's a relative path in the fixture.

training_config/biattentive_classification_network_elmo.json
nelson-liu: good idea, thanks!

allennlp/tests/models/biattentive_classification_network_test.py
matt-gardner: Can we just change the path so that we don't have to change directories?

nelson-liu: unfortunately we can't, since in general we can make no assumptions about where the tests are run from.

matt-gardner: Oh, right, yeah, I forgot about that.  I noticed yesterday that the SEMPRE files were getting generated at `allennlp/data/` instead of at `data/`, and I wondered why, and this explains it.  Fair enough.

PR #1314, opened 2018-05-31
Add option to keep Viterbi scores when predicting

epwalsh:
It can be useful to have the Viterbi scores when making predictions with the `ConditionalRandomField` model. This just adds an optional argument to the `viterbi_tags` method that will return scores with the tags. The default is `False`, so the existing behavior is not modified.


allennlp/modules/conditional_random_field.py
```diff
@@ -278,7 +278,10 @@ def forward(self,
 
         return torch.sum(log_numerator - log_denominator)
 
-    def viterbi_tags(self, logits: torch.Tensor, mask: torch.Tensor) -> List[List[int]]:
+    def viterbi_tags(self,
+                     logits: torch.Tensor,
+                     mask: torch.Tensor,
+                     keep_scores: bool = False) -> Union[List[List[int]], Tuple[List[List[int]], List[float]]]:
```
joelgrus: I don't like functions that return union types, they make for a muddy API.

I would rather just change the function to return (say) `List[Tuple[List[int], float]]` always, and then modify the handful of downstream consumers to behave gracefully and ignore the scores if they don't want them.

allennlp/models/crf_tagger.py
```diff
@@ -132,6 +133,9 @@ def forward(self,  # type: ignore
         logits = self.tag_projection_layer(encoded_text)
         predicted_tags = self.crf.viterbi_tags(logits, mask)
 
+        # Just get the tags and ignore the score.
+        predicted_tags = [x for x, y in predicted_tags]
```
joelgrus: our typechecker is for sure going to complain about this, because you're reassigning with a different type, maybe change it to something like

```
predicted_tags_and_scores = ...

predicted_tags = [tags for tags, _ in predicted_scores]
```

epwalsh: Oops, yea I ran the other tests locally but forgot to run mypy. I'll make that fix!

Issue #1303, opened 2018-05-29
Split serve out of the allennlp commands


schmmd:

ShubhamSarda:
Was facing similar issue, 
To others who are still not able to find solution, try checking version's of Python with Allennlp and other required libraries that will help for sure!  

Thanks.

PR #1317, opened 2018-05-31
Move serve functionality out of allennlp command.

schmmd:
The `allennlp` command currently requires dependencies that few of our users will use and that increase the time it takes to run `allennlp`. Furthermore, the serve command does not work from pip installs, which is the main way our users install AllenNLP.

This PR accomplishes this by moving functionality from `commands` to `allennlp/service/server_flask.py`.

Fixes https://github.com/allenai/allennlp/issues/1303


schmmd:
I see this as a stepping stone to moving the demo into its own repository.  Although I think that's blocked on publishing a Docker image that include `allennlp` from a pip install.

schmmd:
When this is merged I will need to update https://github.com/allenai/allennlp-internal/blob/master/scripts/deploy/deploy-staging-demo.sh and https://github.com/allenai/allennlp-internal/blob/master/scripts/deploy/deploy-production-demo.sh.

allennlp/service/predictors/predictor.py
```diff
@@ -1,6 +1,6 @@
 # pylint: disable=unused-import
 import warnings
 
-from allennlp.predictors.predictor import Predictor, DemoModel, DEFAULT_PREDICTORS
+from allennlp.predictors.predictor import Predictor, DEFAULT_PREDICTORS
```
DeNeutoy: Leave the DemoModel import here (change it to it's new location) to preserve backwards compatibility

allennlp/service/predictors/__init__.py
```diff
@@ -8,7 +8,7 @@
 """
 import warnings
 
-from allennlp.predictors.predictor import Predictor, DemoModel
+from allennlp.predictors.predictor import Predictor
```
DeNeutoy: Same here

allennlp/service/predictors/predictor.py
schmmd: Ultimately I'd like to move the demo to a separate repository that depends on AllenNLP by a pip dependency.  How can I leave the import here but mark it as deprecated?

DeNeutoy: copy the source here or something? 🤷‍♂️

allennlp/service/predictors/__init__.py
schmmd: Done

allennlp/service/predictors/predictor.py
schmmd: I added it to `__init__.py` but I don't understand how adding that import would help backwards compatibility here.  Can you explain/give an example?

PR #1323, opened 2018-05-31
Add logging for the original path in load_archive.

schmmd:


schmmd:
It's a bit confusing in the logs to see what model is actually being loaded, because I only see the path to the cached file.  This adds some logging to show the original path, and informs whether the cache is being used.

allennlp/models/archival.py
```diff
@@ -103,16 +103,21 @@ def load_archive(archive_file: str,
...
-    if os.path.isdir(archive_file):
-        serialization_dir = archive_file
+    if os.path.isdir(resolved_archive_file):
+        serialization_dir = resolved_archive_file
     else:
         # Extract archive to temp dir
         tempdir = tempfile.mkdtemp()
-        logger.info("extracting archive file %s to temp dir %s", archive_file, tempdir)
-        with tarfile.open(archive_file, 'r:gz') as archive:
+        logger.info("extracting archive file %s to temp dir %s", resolved_archive_file, tempdir)
```
joelgrus: as long as you're changing this line, might as well make it an f-string to match the others (and because f-strings are better)

PR #1325, opened 2018-05-31
allow tests to run in parallel

joelgrus:
in order to do this I needed to make AllenNLPTestCase.TEST_DIR depend on os.getpid(), so that different processes have different TEST_DIRs, otherwise they clobber each other.

on my macbook:

```
$ time pytest -n 8

real	2m35.736s
user	5m14.864s
sys	0m28.034s
```

vs

```
$ time pytest

real	3m47.364s
user	3m47.823s
sys	0m19.338s
```

so that shaves more than a minute off the wall-clock time

### DIFF
--- a/allennlp/common/testing/test_case.py
+++ b/allennlp/common/testing/test_case.py
@@ -30,7 +30,7 @@ class AllenNlpTestCase(TestCase):  # pylint: disable=too-many-public-methods
         logging.getLogger('allennlp.modules.token_embedders.embedding').setLevel(logging.INFO)
         log_pytorch_version_info()
 
-        self.TEST_DIR = pathlib.Path("/tmp/allennlp_tests/")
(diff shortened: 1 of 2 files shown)


schmmd:
Nice!  Yeah--tests are getting a bit slow.

schmmd:
I think this change will lead to non-deterministic failures since all our test files are written out to the same directory.  Before we merge this, I think we need to do something like https://github.com/allenai/allennlp/pull/1556.

schmmd:
Oh, oops!  I see you did isolate tests by process!  That seems reasonable, although I like the solution in #1556 more.   Even better, I'd like to move to https://docs.pytest.org/en/latest/tmpdir.html but that's a lot of work for how our tests are laid out.

joelgrus:
doesn't add much value

PR #1327, opened 2018-06-01
Add multiple elmo representations for ELMo Embedder.

easonnie:
Currently, the `ElmoTokenEmbedder` is just a wrapper for a single ELMo representation. Some users might want to get multiple ELMo representations from raw text using the data pipeline in [Embedding Tokens notebook](https://github.com/allenai/allennlp/blob/d711e405cd4de55873718751e7c7192b5be20a16/tutorials/notebooks/embedding_tokens.ipynb).  

I think it will be much better if we could have an option to use multiple ELMo representation in this pipeline. Users can create their own `MultiELMoEmbedder` inherited from `TextFieldEmbedder` (rather than the `BasicTextFieldEmbedder`) to obtain a customized Embeddings that is composed of multiple ELMo Embeddings together with other (Char and Token) Embeddings.


easonnie:
Hi Mark,

Thanks for explanations. I see the problems here.

The current pipeline is:
RawText ---(Indexer)---> Text_with_index_fields ---(TextEmbedder)---> Tensor.
The `TextEmbedder` will combine multiple `Embedders` s.t. the final `Tensor` is the concatenation of all the embeddings.
Suppose we want Glove embeddings (300d), CNN Character embeddings (100d), and two ELMo embeddings (1024d), the final embeddings will have (300+100+2 * 1024d).

The pipeline that I want to build is:
RawText ---(Indexer)---> Text_with_index_fields ---(FlexibleEmbedder)---> Dict[str, Tensor].
where the final output is a dictionary of Tensors.

The output for the example given above will be `{'glove': g_tensor, 'cnn_char': c_tensor, 'elmo_1': e1_tensor, 'emlo_2': e2_tensor}`.

This could give downstream models the flexibility of using different word representations for different submodules.

Also, the current `BasicTextEmbedder` is a little bit constrained since *every* indexer_field need to be fulfilled by *only* one `Embedder`. In some use case, the same char_index_field might be used for different Embedder (like different CNN char model or two layers of ELMo).

easonnie:
I will close it for now and try to figure out a better solution.

PR #1338, opened 2018-06-04
remove demo

joelgrus:
I have moved the demo over to https://github.com/allenai/allennlp-demo

however, I have not updated any of our deployment systems yet, so this totally should not be merged until those are pointing at the new repo and verified working


joelgrus:
this also fixes #1303

allennlp/tests/models/sniff_test.py
```diff
@@ -1,7 +1,41 @@
 # pylint: disable=no-self-use,line-too-long
 
-from allennlp.service.server_flask import DEFAULT_MODELS
 from allennlp.common.testing import AllenNlpTestCase
+from allennlp.models.archival import load_archive
+from allennlp.predictors import Predictor
+
+def demo_model(archive_file: str, predictor_name: str) -> Predictor:
```
schmmd: I don't like having this in the main repo, since it will certainly get out of sync with the actual demo models we use in `allennlp-demo`.  Ideally we would run this test in `allennlp-demo` as part of its CI--but the problem then is that it's easy to publish version of allennlp which break the demo models.

You've probably already gone through the same thinking above!  Just thinking out loud--I don't have a specific suggestion for now.

PR #1339, opened 2018-06-05
Make vocab

LiyuanLucasLiu:
the original implementation of ```make_vocab``` does not support vocabulary configuration (like min_count), also, the comments (and doc) miss the ```--include-package INCLUDE_PACKAGE```


joelgrus:
I think this looks good, but can you add a simple test for the new functionality? Thanks!

LiyuanLucasLiu:
I've modified the original test case for ```make_vocab```, but i'm not sure how to run this test...

joelgrus:
assuming you have pytest installed, you can run just that with

```
pytest -v -k TestMakeVocab
```

the -k only runs test classes that match what you pass it.

--

also, can you leave the original test as is (so that we still test that case) and just create a separate one for passing in the params? thanks

LiyuanLucasLiu:
this pytest is cool! I've created a new test case for the config :-)

allennlp/commands/make_vocab.py
```diff
@@ -70,8 +72,8 @@ def make_vocab_from_params(params: Params):
     prepare_environment(params)
 
     vocab_params = params.pop("vocabulary", {})
-    vocab_dir = vocab_params.get('directory_path')
-    if vocab_dir is None:
+    vocab_dir = vocab_params.pop('directory_path', {})
```
joelgrus: one last nit: can you change the default value to `None`? (`{}` doesn't make sense as a possible value)

LiyuanLucasLiu:
ok, no problem

PR #1342, opened 2018-06-05
Add JustSpacesWordSplitter

tahirahmad2030:
Changed Tokenizer from Spacy tokenizer to JustSpaceTokenizer


matt-gardner:
We definitely want to keep `spacy` as the default tokenizer for the `SentenceTaggerPredictor`.  If this works for your code, great, but this isn't a change we're going to accept upstream.

Issue #1125, opened 2018-04-24
Create a configuration wizard to help users build JSON configurations


schmmd:
See https://github.com/allenai/allennlp-internal/issues/27 for more context.

matt-gardner:
Fixed by #1344.

PR #1344, opened 2018-06-05
fancy version of the configuration wizard

joelgrus:
* sort of works
* is really ugly
* handles subconfigurations correctly
* does not handle List / Dict / Tuple types correctly
* "generate config" button doesn't do anything yet

screen shot 2018-06-05 at 4 57 56 pm


matt-gardner:
What's the right command to try this?  Last time I did it I had to resort to a bit of hackery; I assume you have a command that just works from this branch.

joelgrus:
just do 

```
python allennlp/service/config_explorer.py
```

and then take your browser to

```
http://localhost:8123/wizard/
```

matt-gardner:
When I run that I get `No module named 'allennlp'`, so I need to do the path munging that we do in our other scripts.  Have you modified your `PYTHONPATH`, or something?

joelgrus:
probably, I usually `export PYTHONPATH=.` in the root directory of the repo

matt-gardner:
Very nice!  This is going to be super cool when it's done.

joelgrus:
* slightly less ugly
* handles lists and dicts correctly, except where they're lists/dicts of Registrable things
* highlights in red things that need attention (but doesn't handle lists / dicts correctly, because I need to feed the state of the list items back up)
* maintains a top level state that's the desired JSON object (right now the generate button just logs it to the console)

next steps:

* handle list / dict items that are themselves registrable
* handle done-ness of lists / dicts
* place output of generate on page
* make more attractive

joelgrus:
* big refactoring to a more correct design (refactoring javascript with no types is hard!)
* mostly works (even lists and dicts of configurable items)
* a small amount of styling
* "completeness" should inherit correctly (a top-level item is complete iff all of its required subitems are complete)
* still relies on the "generate" button (reactjs got into an infinite loop when I tried to generate the json automatically)
* doesn't handle the (unusual) Tuple case yet
* could use more styling + UX thought

joelgrus:
ok, this is probably as good as I'm going to get it for now.

there are a few edge cases (union types + tuples) that it probably doesn't handle right, but they come up rarely. I tried to make it moderately attractive, but I don't know if I succeeded on that.

I'll demo it at the team meeting today.

joelgrus:
btw this fixes #1125

matt-gardner:
This is awesome!  Very nice work!  It looks way better than what I would have done; thanks for doing this.  My only suggestion would be to give a type hint for dicts and lists, so you don't have to guess what the "+" button does.

tutorials/how_to/create_a_configuration.md
```diff
@@ -0,0 +1,132 @@
...
+just what needs to go in the configuration file. Accordingly, we've
+built a couple of tools (that are still somewhat experimental)
+to help you out.
+
+## `allennlp configure`
+
+One option is to use the new `allennlp configure` command, which
+produces stubs that can be used to produce configuration files.
+
+If you run it without any options, you get an output corresponding to the top-level configuration:
```
matt-gardner: You should probably note here that it's for the `train` command.

allennlp/service/config_explorer.py
```diff
@@ -74,132 +96,620 @@ def main(args):
...
-            $title
-        </title>
-        <style>
-            $css
-        </style>
-    </head>
-    <body>
-        $body
-    </body>
+  <head>
```
DeNeutoy: I think you can avoid this duplication by adding the html file to `MANIFEST.in` in the root of the repo - this dictates non python stuff which is included/excluded by pip.

allennlp/service/config_explorer.html
```diff
@@ -0,0 +1,608 @@
...
+    <div id="app"></div>
+
+    <script crossorigin src="https://unpkg.com/react@16/umd/react.production.min.js"></script>
+    <script crossorigin src="https://unpkg.com/react-dom@16/umd/react-dom.production.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/immutable/3.8.2/immutable.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/babel-standalone/6.26.0/babel.js"></script>
+    <script src="https://unpkg.com/tippy.js@2.5.2/dist/tippy.all.min.js"></script>
+    <script type="text/babel">
+/*
+ConfigItem = {
```
DeNeutoy: Looks like you've replaced these below, right?

allennlp/service/config_explorer.html
```diff
@@ -0,0 +1,608 @@
...
+            color: white;
+        }
+    </style>
+  </head>
+  <body>
+    <h2>AllenNLP Configuration Wizard (alpha)</h2>
+
+    <div id="app"></div>
+
+    <script crossorigin src="https://unpkg.com/react@16/umd/react.production.min.js"></script>
```
DeNeutoy: Why do some need crossorigin and others don't?

allennlp/service/config_explorer.html
```diff
@@ -0,0 +1,608 @@
...
+ApiResponse = {
+    className: str
+    config: Config
+} or {
+    className: str
+    choices: Dict[str, str]  // name -> className
+}
+*/
+
+// A configItem is optional if it has no default value
```
DeNeutoy: If it does have a default value?

allennlp/service/config_explorer.html
```diff
@@ -0,0 +1,608 @@
...
+
+    setData(fn) {
+        const {data} = this.state
+        const newData = markComplete(fn(data))
+        return this.setState({data: newData})
+    }
+
+    componentDidMount() {
+        // Fetch the top level configuration
+        fetch('/api/')
```
DeNeutoy: Is it possible that this API is going to be extended to include partial json configurations/recommendations etc? It might be better to call this api `/annotate` or something, if that is the case.

allennlp/service/config_explorer.html
```diff
@@ -0,0 +1,608 @@
...
+            </select>
+        </span>
+    ) : null
+    const deleteButton = choices ? <button onClick={remove}>x</button> : null
+    const renderedValue = config ? <Config path={path.push('value')} config={config} setData={setData}/> : null
+
+    return <span>
+            {configureButton}
+            <div class="choices-dropdown">
+                {choicesDropdown}{deleteButton}{renderedValue}
```
DeNeutoy: Nit: render the `deleteButton` first so that it's always in the same place - currently, for `List` items, it is before the entry field, but here it's after.

allennlp/service/config_explorer.html
```diff
@@ -0,0 +1,608 @@
...
+                {choice ? null : (<option value=""></option>)}
+                {choices.map(subclass => <option value={subclass}>{subclass.slice(prefix.length)}</option>)}
+            </select>
+        </span>
+    ) : null
+    const deleteButton = choices ? <button onClick={remove}>x</button> : null
+    const renderedValue = config ? <Config path={path.push('value')} config={config} setData={setData}/> : null
+
+    return <span>
+            {configureButton}
```
DeNeutoy: I think if you render the `configButton` in it's own `span` the spacing between things will look nicer:

image

DeNeutoy: That is bigger visually than I intended

allennlp/service/config_explorer.html
```diff
@@ -0,0 +1,608 @@
...
+                this.setState({data: data})
+            })
+    }
+
+    render() {
+        const config = this.state.data.get('value')
+        return (
+            <div class="wizard">
+                <Config path={Immutable.List(['value'])} setData={this.setData} config={config}/>
+                <JsonBox config={config}/>
```
DeNeutoy: Idea: it would be nice if the rendered JSON contained empty dictionaries for required fields - e.g you might want to fill in a partial config and then fill in the rest later, or something.

allennlp/tests/service/config_explorer_test.py
```diff
@@ -14,66 +18,111 @@ def setUp(self):
...
+        data = json.loads(response.get_data())
+        config = data['config']
+        items = config['items']
 
-        assert '"type": "rmsprop"' in html
-        assert '// "weight_decay"' in html
+        assert config["type"] == "rmsprop"
+        assert any(item["name"] == "lr" for item in items)
 
     def test_rnn_hack(self):
```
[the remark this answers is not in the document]
joelgrus: no, it's something different, I added a comment

allennlp/common/configuration.py
```diff
@@ -143,8 +219,8 @@ def _auto_config(cla55: Type[T]) -> Config[T]:
     """
     typ3 = _get_config_type(cla55)
 
-    # Don't include self
-    names_to_ignore = {"self"}
+    # Don't include self, or vocab
```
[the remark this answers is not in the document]
joelgrus: exactly, this is inside `auto_config`, which uses reflection to figure out the config for some object. the top level config is arbitrary (it's just whatever `commands/train` expects), so it doesn't use this method.

allennlp/common/configuration.py
```diff
@@ -92,15 +143,13 @@ def __repr__(self) -> str:
...
     def to_json(self) -> JsonDict:
-        item_dict: JsonDict = {
-                item.name: item.to_json()
-                for item in self.items
-        }
+        blob: JsonDict = {'items': [item.to_json() for item in self.items]}
 
         if self.typ3:
-            item_dict["type"] = self.typ3
+            #items.insert(0, {"name": "type", "type": self.typ3})
```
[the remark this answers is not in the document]
joelgrus: done

allennlp/service/config_explorer.py
joelgrus: I'm going to leave this for a future PR, I looked into it a bit and it seems like it's tricky to get right

allennlp/service/config_explorer.html
joelgrus: sorry, that's not javascript code, that's a comment about the types that come from the API, I'll add a comment to that effect

allennlp/service/config_explorer.html
joelgrus: 🤷‍♂️ 

the "crossorigin" was in the documentation for how to use react from a CDN, it doesn't work with the other ones

allennlp/service/config_explorer.html
joelgrus: fixed

allennlp/service/config_explorer.html
joelgrus: I'll make it `/api/config/` to leave room for other `/api/` routes

allennlp/service/config_explorer.html
joelgrus: I don't like this way, but if I make that change then a `Dict[str, Configurable]` element becomes a huge mess

allennlp/service/config_explorer.html
joelgrus: fixed with css

allennlp/service/config_explorer.py
DeNeutoy: sure

PR #1347, opened 2018-06-06
Faster Elmo

DeNeutoy:
Timings in ms, Avg of 5 runs
CPU= My Macbook
GPU= AWS P3.xlarge

### CPU Speedup, Batch Size 20, Sequence Length 64:
**With Cached Vocab**
CNN Encoder: 6.61
Contextual Layers: 2885.81
Total Time: 2892.39
(Vocab caching time: 995.14 for vocab size 640)

**Original**
CNN Encoder: 1699.12
Contextual Layers: 2687.23 
Total Time: 4386.35
**Overall Speedup: 34%**

### CPU Speedup, Batch Size 50, Sequence Length 64:
**With Cached Vocab**
CNN Encoder: 13.86
Contextual Layers: 5786.15
Total Time: 5799.16
(Vocab caching time: 1298.10 for vocab size 1114)

**Original**
CNN Encoder: 3766.86
Contextual Layers: 5315.72
Total Time: 9081.76

**Overall Speedup: 36%**

Timings in ms, Avg of 5 runs

### GPU Speedup, Batch Size 20, Sequence Length 64:
**With Cached Vocab**
CNN Encoder: 2.34
Contextual Layers: 41.81
Total Time: 44.15
(Vocab caching time: 16 for vocab size 640)

**Original**
CNN Encoder: 11.72
Contextual Layers: 44.87 
Total Time: 56.59
**Overall Speedup: 22%**

### GPU Speedup, Batch Size 50, Sequence Length 64:
**With Cached Vocab**
CNN Encoder: 2.31
Contextual Layers: 94.49
Total Time: 96.81
(Vocab caching time: 26.6 for vocab size 1114)

**Original**
CNN Encoder: 23.79
Contextual Layers: 104.21
Total Time: 128.00

**Overall Speedup: 25%**


matt-gardner:
Did you try it with a very large vocab?

allennlp/modules/elmo.py
```diff
@@ -15,10 +15,11 @@
 from allennlp.common.file_utils import cached_path
 from allennlp.common.checks import ConfigurationError
 from allennlp.common import Params
+from allennlp.common.util import lazy_groups_of
 from allennlp.modules.elmo_lstm import ElmoLstm
 from allennlp.modules.highway import Highway
 from allennlp.modules.scalar_mix import ScalarMix
-from allennlp.nn.util import remove_sentence_boundaries, add_sentence_boundary_token_ids
+from allennlp.nn.util import remove_sentence_boundaries, add_sentence_boundary_token_ids, get_device_of
```
matt-gardner: Not for this PR, but is this function (`get_device_of`) still necessary with pytorch 0.4?

allennlp/modules/elmo.py
```diff
@@ -75,6 +76,7 @@ def __init__(self,
                  requires_grad: bool = False,
                  do_layer_norm: bool = False,
                  dropout: float = 0.5,
+                 vocab_to_cache: List[str] = None,
```
matt-gardner: Add this to the docstring.  Also, it looks like passing this changes the expected input to the function, yes?  Is that documented anywhere?

allennlp/modules/elmo.py
```diff
@@ -455,11 +469,26 @@ class _ElmoBiLm(torch.nn.Module):
     def __init__(self,
                  options_file: str,
                  weight_file: str,
-                 requires_grad: bool = False) -> None:
+                 requires_grad: bool = False,
+                 vocab_to_cache: List[str] = None) -> None:
```
matt-gardner: Add to docstring.

allennlp/modules/elmo.py
```diff
@@ -455,11 +469,26 @@ class _ElmoBiLm(torch.nn.Module):
...
+        if requires_grad and vocab_to_cache:
+            logging.warning("You are fine tuning ELMo and caching char CNN word vectors. "
+                            "This behaviour is not guaranteed to be well defined, particularly. "
+                            "if not all of your inputs will occur in the vocabulary cache.")
+
+        self._word_embedding = None
+        if vocab_to_cache:
+            logging.info("Caching character cnn layers for words in vocabulary.")
+            # This sets 3 attributes, _word_embedding, _bos_embedding and _eos_embedding.
+            # They are set in the method so it can be accessed from outside the
```
matt-gardner: `s/so it can/so they can/`.  Also, shouldn't you have to set `_bos_embedding` and `_eos_embedding` as attributes on `self` here?  I'm surprised pylint isn't complaining about this.  And type annotations on these initial `None` assignments would be nice; I'm also surprised mypy isn't complaining about that.

allennlp/modules/elmo.py
```diff
@@ -117,8 +123,12 @@ def forward(self,    # pylint: disable=arguments-differ
         """
         # reshape the input if needed
         original_shape = inputs.size()
-        timesteps, num_characters = original_shape[-2:]
-        if len(original_shape) > 3:
+
+        if self._has_cached_vocab and len(original_shape) > 2:
+            timesteps = original_shape[-1]
```
matt-gardner: I'm a bit confused at this - it's implying you're getting a different input, right?  This needs to be documented.  If it's in the docstring, that might be enough, but a comment here in addition to a mention in the docstring might also be worth it.

allennlp/modules/elmo.py
```diff
@@ -483,8 +512,12 @@ def forward(self,  # pylint: disable=arguments-differ
         """
         Parameters
         ----------
-        inputs: ``torch.Tensor``
-            Shape ``(batch_size, timesteps, 50)`` of character ids representing the current batch.
+        inputs: ``torch.Tensor``, required.
+            The shape depends on whether you have cached character convolutions.
```
matt-gardner: Ah, great, here we get some documentation about this.  This is a pretty important point, so I'd definitely mention it everywhere that `vocab_to_cache` appears.  Also, looks like this will crash if you don't cache the whole vocab?  Yes?  That'd be good to mention also.

allennlp/modules/elmo.py
```diff
@@ -499,14 +532,31 @@ def forward(self,  # pylint: disable=arguments-differ
         Note that the output tensors all include additional special begin and end of sequence
         markers.
         """
-        token_embedding = self._token_embedder(inputs)
-        type_representation = token_embedding['token_embedding']
-        mask = token_embedding['mask']
+        if self._word_embedding is not None:
+            mask_without_bos_eos = (inputs > 0).long()
+            # The character cnn part is cached - just look it up.
+            embedded_inputs = self._word_embedding(inputs) # type: ignore
```
matt-gardner: Ah, so mypy _isn't_ happy with it =).  A type annotation on the initial assignment would fix this, right?

DeNeutoy:
Not yet, but this version doesn't suffer from the same problem (it's just a lookup, rather than a comparison which scales wrt to the vocab). I'll do it for completeness, given it's quite easy.

allennlp/modules/elmo.py
```diff
@@ -499,14 +532,31 @@ def forward(self,  # pylint: disable=arguments-differ
...
+            mask = token_embedding['mask']
+            type_representation = token_embedding['token_embedding']
         lstm_outputs = self._elmo_lstm(type_representation, mask)
 
         # Prepare the output.  The first layer is duplicated.
+        # Because of minor differences in how masking is applied depending
+        # on whether the char cnn layers are cached, we'll be defensive and
+        # multiply by the mask here. It's not strictly necessary, as the
+        # mask passed on is correct, but the values in the padded areas
+        # of the char cnn representations can change.
```
matt-gardner: Wow, that sounds like it was a pain.  I assume you found this because you were testing for exactly reproducing the outputs between the two branches?

allennlp/modules/elmo.py
```diff
@@ -515,3 +565,63 @@ def forward(self,  # pylint: disable=arguments-differ
                 'activations': output_tensors,
                 'mask': mask,
         }
+
+    def create_cached_cnn_embeddings(self, tokens: List[str]) -> None:
+        """
+        Given a list of tokens, this method precomputes word representations
+        by running just the character convolutions and highway layers of elmo,
+        essentially creating uncontextual word vectors. On subsequent forward passes,
+        if all of the words in the batch have cached, the word ids are looked up from
```
matt-gardner: I didn't see anything that actually checked "if all of the words in the batch have [been] cached".  Did I miss something, or is this comment wrong?

allennlp/modules/elmo.py
```diff
@@ -515,3 +565,63 @@ def forward(self,  # pylint: disable=arguments-differ
...
+        """
+        tokens = [ELMoCharacterMapper.bos_token, ELMoCharacterMapper.eos_token] + tokens
+        timesteps = 32
+        batch_size = 32
+        chunked_tokens = lazy_groups_of(iter(tokens), timesteps)
+
+        all_embeddings = []
+        device = get_device_of(next(self.parameters()))
+        for batch in lazy_groups_of(chunked_tokens, batch_size):
+            # Shape (batch_size, 32, 32, 50)
```
matt-gardner: Isn't this `(batch_size, timesteps, 50)`, or `(32, 32, 50)`?  Or are there really 4 dimensions here?

allennlp/modules/elmo.py
```diff
@@ -515,3 +565,63 @@ def forward(self,  # pylint: disable=arguments-differ
...
+            # already placed the model on a device. If this is called in the
+            # constructor, it will probably happen on the CPU. This isn't too bad,
+            # because it's only a few convolutions and will likely be very fast.
+            if device >= 0:
+                batched_tensor = batched_tensor.cuda(device)
+            output = self._token_embedder(batched_tensor)
+            token_embedding = output["token_embedding"]
+            mask = output["mask"]
+            token_embedding, _ = remove_sentence_boundaries(token_embedding, mask)
+            all_embeddings.append(token_embedding.view(-1, token_embedding.size(-1)))
```
matt-gardner: Preference for not using `-1` in `.view`, for readability and for more obvious errors.

allennlp/modules/elmo.py
```diff
@@ -515,3 +565,63 @@ def forward(self,  # pylint: disable=arguments-differ
...
+        full_embedding = torch.cat(all_embeddings, 0)
+
+        # We might have some trailing embeddings from padding in the batch, so
+        # we clip the embedding and lookup to the right size.
```
matt-gardner: It didn't look like padding would get added anywhere.  What kind of padding are you dealing with here?

DeNeutoy: If the number of words is not exactly divisible by 32, we'll get padding in the last sequence in the last batch, so this just clips it off.

allennlp/modules/elmo.py
DeNeutoy: Yes

allennlp/modules/elmo.py
DeNeutoy: This was from an old version where you could "back off" to computing them on the fly. Actually I think I need to add this, because otherwise you can't deal with new words, e.g at test time.

DeNeutoy: Ugh this completely breaks this approach actually, because it requires access to the character ids if the word ids aren't in the cache, which means to back off to a different approach, both need to be passed. That's pretty frustrating. This is basically useless if it can't handle new words at inference time, because it defeats a big bonus of using elmo.

allennlp/modules/elmo.py
matt-gardner: Hmm, I didn't think that `lazy_groups_of` added any padding.  Am I wrong on that, or is padding getting added somewhere else?

matt-gardner: Oh, because you've got two nested groupings...?

allennlp/modules/elmo.py
matt-gardner: Could you have this take the whole `TextField` input, and use whichever part you need?  You could also imagine dynamically growing the cache.  But that's probably not necessary, because the words that aren't in your cache are always going to be OOV, so you can just filter them that way...

DeNeutoy: I don't see how that's an option? We need to be able to use it as a token embedder, right?

matt-gardner: Well, there's no reason we couldn't just add an `ElmoTextFieldEmbedder` that handled things differently from the `BasicTextFieldEmbedder`, and passed multiple representations on to the elmo classes.  And we say that you need to do this if you want caching.  Things get a bit complicated, so if you're not too worried about speed, you can just do things as they currently are, but if you want to speed things up with caching, you modify your config even more and use this other `TextFieldEmbedder`.

These are just some thoughts that I'm hoping will help you figure out a good solution; I'm not sure whether this is the right way to do things or not.

DeNeutoy: Sorry, I didn't mean to be obnoxious - it's just i've tried to do this twice now and both times run into kind of complicated issues which was annoying me.

That way of doing things seems plausible, I think. I'll try that today.

allennlp/modules/elmo.py
DeNeutoy: So this is annoying - If I import the Embedding, mypy complains because I have to do it before the module is imported, and then complains again because None isn't callable. I think the others don't complain because the type annotation on `add_sentence_boundaries` is `Any`, so it doesn't matter what they are.

matt-gardner: Hmm, interesting, ok.  The point of the annotations is mostly to aid readability; if it's too much work (or impossible) to get mypy to be happy with it, then a simple comment explaining what `_word_embedding` holds would be good enough.

PR #1351, opened 2018-06-07
Format pytest logging correctly

### LINKED ISSUE #1350: Logging configuration is messed up when testing
For some reason the logs aren't showing up correctly during testing.  I'm seeing a different logging format than shows up during training.  Not sure what's going on...

matt-gardner:
Fixes #1350.  Well, it kind of fixes it.  I'd prefer to have the logs just interleaved with stderr, like they used to be, because this way it makes it pretty hard to follow some things...  If anyone knows how to get back to that without just pinning to `pytest==3.2`, I'd be happy to hear it.  (I'm also maybe happy to pin to 3.2 instead of this...)

### DIFF
--- a/allennlp/common/testing/test_case.py
+++ b/allennlp/common/testing/test_case.py
@@ -28,6 +28,7 @@ class AllenNlpTestCase(TestCase):  # pylint: disable=too-many-public-methods
         logging.getLogger('allennlp.common.params').disabled = True
         logging.getLogger('allennlp.nn.initializers').disabled = True
         logging.getLogger('allennlp.modules.token_embedders.embedding').setLevel(logging.INFO)
+        logging.getLogger('urllib3.connectionpool').disabled = True
         log_pytorch_version_info()
 
         self.TEST_DIR = pathlib.Path("/tmp/allennlp_tests/")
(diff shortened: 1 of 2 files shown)


allennlp/common/testing/test_case.py
```diff
@@ -28,6 +28,7 @@ def setUp(self):
         logging.getLogger('allennlp.common.params').disabled = True
         logging.getLogger('allennlp.nn.initializers').disabled = True
         logging.getLogger('allennlp.modules.token_embedders.embedding').setLevel(logging.INFO)
+        logging.getLogger('urllib3.connectionpool').disabled = True
```
matt-gardner: This is used for checking etag stuff, I think, and when you set the logging level to debug with pytest, these messages start showing up.  Pretty sure we don't need them, but if anyone sees a reason to, we can remove this line.

matt-gardner:
Yeah, I saw that, but it didn't look like what I wanted.  I just tried it out, and it's not.  That outputs logs _in-line_ while running the tests (and, oddly, disables capturing of stdout and stderr, which don't show up in-line, either...).  I want the log output captured and shown at the end, but grouped together with stderr, as it used to be.  Because we have some stuff going to stderr (like tqdm) and some stuff going to logging, separating them is confusing.  I'm seriously considering just pinning to 3.2 for this...  But I'll merge this as it is for now, because at least the logs show up right now, and if the separation of logs and stderr actually causes me problems in the future, I'll open a new PR to pin pytest, or try to figure out another way to get around it.

PR #1352, opened 2018-06-07
Add ability to add tokens to vocabulary namespaces

matt-gardner:
There are a few tokens that I needed to be added to the wikitables parser vocabulary, and I couldn't figure out a cleaner way to do it than to just add them directly to the vocab in the configuration.  Ideally I'd want the model code to specify this, not the configuration file, but I don't know how to do that in a nice way.

This seems like something people might want to do in other circumstances, too.

### DIFF
--- a/allennlp/data/vocabulary.py
+++ b/allennlp/data/vocabulary.py
@@ -155,13 +155,17 @@ class Vocabulary:
         regardless of their count, depending on the value of ``only_include_pretrained_words``.
         Words which appear in the pretrained embedding file but not in the data are NOT included
         in the Vocabulary.
-    only_include_pretrained_words : bool, optional (default = False)
+    only_include_pretrained_words : ``bool``, optional (default=False)
         This defines the stategy for using any pretrained embedding files which may have been
         specified in ``pretrained_files``. If False, an inclusive stategy is used: and words
         which are in the ``counter`` and in the pretrained file are added to the ``Vocabulary``,
         regardless of whether their count exceeds ``min_count`` or not. If True, we use an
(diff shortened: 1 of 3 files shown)


DeNeutoy:
Why don't you just add them to the vocab in the init of your model?

matt-gardner:
That's brittle - it relies on us never using the vocabulary to process data before constructing the model object.  That's currently true in our train command, but not in, e.g., `make-vocab`.  It also could lead to some really subtle bugs, as we currently _save_ the vocabulary right after constructing it, before it gets passed to the model - because of how `train` works, you'd actually index things correctly, but you'd be super confused if you actually looked at the vocab files.

Ideally, you really want the vocabulary to be finalized before the model uses it to set dimensions and things.  This is tricky if you want the model to influence the vocabulary construction, however.

allennlp/data/vocabulary.py
```diff
@@ -169,7 +169,8 @@ def __init__(self,
                  max_vocab_size: Union[int, Dict[str, int]] = None,
                  non_padded_namespaces: Sequence[str] = DEFAULT_NON_PADDED_NAMESPACES,
                  pretrained_files: Optional[Dict[str, str]] = None,
-                 only_include_pretrained_words: bool = False) -> None:
+                 only_include_pretrained_words: bool = False,
+                 tokens_to_add: Dict[str, List[str]] = None) -> None:
```
DeNeutoy: Add a docstring.

matt-gardner: Yep, thanks for catching this.  Done.

Issue #1206, opened 2018-05-13
TestTrainer.test_trainer_respects_keep_serialized_model_every_num_seconds fails when running without GPU on GPU machine


nelson-liu:
Rather odd situation: when running the tests on a GPU machine with `CUDA_VISIBLE_DEVICES=""`, [`TestTrainer.test_trainer_respects_keep_serialized_model_every_num_seconds`](https://github.com/allenai/allennlp/blob/master/tests/training/trainer_test.py#L160) consistently fails since each epoch takes a rather sluggish ~1.4 seconds, causing every epoch checkpoint to be saved (thus failing the test).

The easy fix would be to just increase the sleep to 2.5 seconds and `keep_serialized_model_every_num_seconds` to 5, but not sure if that's a fix worth making.

Interestingly, it works fine as is on a machine without a GPU (my macbook). Not too sure what to make of this...

nelson-liu:
looks like it's just flaky in general: http://build.allennlp.org/viewLog.html?buildId=3394&buildTypeId=AllenNLP_AllenNLPPullRequests&tab=buildLog fails for a completely unrelated change @ https://github.com/allenai/allennlp/pull/1203

schmmd:
@matt-peters are you able to take a look?  @DeNeutoy thought you had good context.

matt-peters:
I added this functionality to the Trainer. As for why it's taking so long on a GPU machine with `CUDA_VISIBLE_DEVICES=""` ?  Who knows.  I'm all for the easiest fix here, so the suggestion to change the sleep interval to 2.5 and  keep_serialized_model_every_num_seconds to 5 sounds good to me.

nelson-liu:
Fixed in #1358

PR #1358, opened 2018-06-12
Make test_trainer_respects_keep_serialized_model_every_num_seconds more patient

nelson-liu:
Easy fix from #1206

### DIFF
--- a/allennlp/tests/training/trainer_test.py
+++ b/allennlp/tests/training/trainer_test.py
@@ -246,15 +246,15 @@ class TestTrainer(AllenNlpTestCase):
 
     def test_trainer_respects_keep_serialized_model_every_num_seconds(self):
         # To test:
-        #   Create an iterator that sleeps for 0.5 second per epoch, so the total training
-        #       time for one epoch is slightly greater then 0.5 seconds.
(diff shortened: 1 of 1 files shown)


nelson-liu:
Hmm, @schmmd maybe you need to change the check that is required? I think you were editing TC recently

schmmd:
@nelson-liu I think your PR might be in a weird state because I renamed "AllenNLP" to "AllenNLP Library" after this PR was created.  Rest assured--the check actually passed--it just never heard back from a check that started ("AllenNLP") because it was renamed.

nelson-liu:
@schmmd ah ok, I tried to force push and rerun ci to see if that'd fix anything, but it doesn't look like it did. Maybe you can force the merge? Alternatively, you can let me know when the issue should be fixed and I'll push again to rerun ci

allenai-dev-role:
Sorry @nelson-liu -- I didn't know you were unable to force the merge.  I would have been happy to earlier.

matt-gardner:
@schmmd, it looks like there's still an issue here - the CI checks for #1364 are also in a weird state, and that one was just opened.

schmmd:
@matt-gardner thanks for the pointer.

PR #1364, opened 2018-06-13
Adds support for reading pretrained embeddings (text format) from uncompressed files and archives

janLuke:
Adds support for reading pretrained embeddings (text format) from:
* an uncompressed text file
* a text file compressed with zip, gzip, bz2, lzma
* file inside a zip or tar archive containing multiple files;
   in this case, the parameter ``pretrained_file`` must be a pair
   ``[path_to_archive, path_to_file_inside_archive]``, e.g.:
   ```
        "pretrained_file": [
            "http://nlp.stanford.edu/data/glove.twitter.27B.zip",
            "glove.twitter.27B.200d.txt"]
    ```
    If an archive contains a single file, this is not needed and a string is enough.
    The same notation can be used to specify ``pretrained_files`` in ``Vocabulary``.

Improved feedback to the user:
* Number of vocab tokens present in the pre-trained embedding file
* progress bars; read the number of tokens if declared in the first row

New test fixtures were included. The glove files and several new compressed versions of
fake embeddings were put inside the folder ``fixtures/embeddings`` to not clutter
the fixtures folder.


matt-gardner:
@joelgrus, this is one you were looking at, right?  Feel free to re-assign if I'm remembering wrong.

joelgrus:
IMO union types are a bad code smell, I much prefer the 

```
   "pretrained_file": "http://nlp.stanford.edu/data/glove.twitter.27B.zip",
   "path_in_pretrained_file": "glove.twitter.27B.200d.txt",
   ...
```

approach we discussed earlier rather than having things floating around in the code which could be lists or could be strings.

I don't want to run you around making changes. If you give me write access to your fork, I can take a stab at changing this to how I would like it. Is that cool?

joelgrus:
actually, I think that would make things too complicated as well. what about the following:

```
"pretrained_file": "http://nlp.stanford.edu/data/glove.twitter.27B.zip:glove.twitter.27B.200d.txt"]
```

where by convention everything after the `:` is the path inside the archive. no one is going to have colons in their actual filenames, right?

I feel like this would make the code much simpler.

matt-gardner:
@janLuke @joelgrus, you don't actually have to give special write access, you just need to [allow edits to PR branches from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/).  This is a checkbox when you create the PR (and can be edited later) which I think is checked by default.  So, you can probably do the edit yourself already.  But, yes, it's good to ask first.  Just letting you know that there shouldn't be any other permissions step to make this happen.

janLuke:
@joelgrus You are right, I don't like Union types either. But I mean in the code, not in the configuration file. I don't think it's a problem to have a hocon parameter that can be a string or a list. 
We can handle that in ``from_params`` once and for all. Maybe we can use an ``EmbeddingFilePath`` with a "first level path" and (optional) "second level path" and use that outside? 

Using ":" as a separator is a bad idea considering that it's a valid character in Unix paths and that is always present in http URLs. But we could certainly think of more "unlikely" strings (even "::"). Nonetheless, as I said, I don't think using a list is a problem. I'm okay with both solutions.

In any case, I think I'd prefer not to add another parameter (as we have discussed earlier) so that the notation remain "leaner". Consider that the same notation must be used in the Vocabulary class where ``pretrained_files`` is already a dictionary.

Feel absolutely free to make any change!

janLuke:
What about
```
pretrained_file: "(http://nlp.stanford.edu/data/glove.twitter.27B.zip)#glove.twitter.27B.200d.txt"

def decode_embeddings_file_path(path):
    match = re.match('\((.*)\)#(.*)', path)
    if match:
        return match.groups()
    return path, None
```
Idea taken from here: http://doc.cloveretl.com/documentation/UserGuide/index.jsp?topic=/com.cloveretl.gui.docs/docs/examples-of-file-url-in-readers.html

I like it.

janLuke:
@joelgrus I worked on it. You can see the result in https://github.com/janLuke/allennlp/tree/embeddings_uri. Even if you don't like the URI format I mentioned above, you can easily change it by editing ``get_embeddings_file_uri`` and ``decode_embeddings_file_uri``. 

Let me know if I can merge it to this branch. Of course, feel free to make any change after merging.

joelgrus:
yeah, I like that approach, go ahead and merge it into the PR (I think some of the type annotations still reflect the old approach tho)

janLuke:
Cool! I'll try to fix all type annotations and doc strings before merging. :)

janLuke:
@joelgrus Ready for review.

joelgrus:
I made one somewhat big change, which is that I replaced your `open_embeddings_text_file` function with an `EmbeddingsTextFile` class (and got rid of the `get_embeddings_file_iterator_with_progbar` function, which I did not like). The class has all of the relevant functionality happen seamlessly, and I moved the Tqdm calls out to the consumers of the class, which feels more natural.

Let me know if you have any disagreements with this change.

janLuke:
I absolutely agree with this change! :) In fact, I had started to refactor everything to an  ``EmbeddingsTextFile`` class (I decided that while I was writing a test for the ugly function returning a progress bar ^.^").

I merged your code with mine here: https://github.com/janLuke/allennlp/tree/embeddings_text_file.
I had done some minor things differently: 
* put the code for opening the stream inside ``__init__`` and add a ``close()`` method so that it works similarly to ``open / ZipFile`` (even outside ``with`` clauses)
* kept an (optional) handle to the eventual archive file so I can close that too (even if that may not be strictly necessary given that we are only reading...)
* put the code for zip and tar in separate private methods (e.g. ``_open_inside_zip``)
* made ``_get_num_tokens_from_first_line`` and ``_get_the_only_file_in_the_archive`` static methods of ``EmbeddingsTextFile``

Feel free to do whatever you prefer. 

I guess I finished here. Thank you for your suggestions :)

janLuke:
Ah, I was wondering what do you think about implementing ``__len__`` so that it returns ``self.num_tokens`` and we don't need to worry about passing ``total=embeddings_file.num_tokens`` to ``tqdm`` every time.

janLuke:
I made other changes in https://github.com/janLuke/allennlp/tree/embeddings_text_file. Nothing substantial.
The biggest one is probably that I removed ``self._buffer`` and added a ``self._iterator``:
```
        if self._num_tokens:
            self._iterator = self._handle
        else:
            self._iterator = itertools.chain([first_line], self._handle)
```
I think it simplifies the code. Methods become one-liners.

joelgrus:
yeah, that seems like a good approach, do you want to close this PR and create one from that branch?

janLuke:
Is it necessary? I can merge it on this branch without conflicts.

joelgrus:
that's fine too, whatever's easiest

allennlp/modules/token_embedders/embedding.py
```diff
@@ -296,3 +352,183 @@ def _read_pretrained_hdf5_format_embedding_file(embeddings_filename: str, # pyli
...
+                 cache_dir: str = None) -> None:
+
+        self._uri = file_uri
+        self._encoding = encoding
+        self._cache_dir = cache_dir
+
+        self._handle: TextIO
+        self._archive_handle: Any = None   # only if the file is inside an archive
+        self._iterator: Iterator[str]
+        self._num_tokens: Optional[int]
```
joelgrus: I'm not a fan of these bare declarations, can't you just use the type hints the first time the variable is defined?

allennlp/modules/token_embedders/embedding.py
```diff
@@ -296,3 +352,183 @@ def _read_pretrained_hdf5_format_embedding_file(embeddings_filename: str, # pyli
...
+            member_path = self._get_the_only_file_in_the_archive(members_list, archive_path)
+        member_path = cast(str, member_path)
+        member = archive.getmember(member_path)   # raises exception if not present
+        member_file = cast(IO[bytes], archive.extractfile(member))
+        self._handle = io.TextIOWrapper(member_file, encoding=self._encoding)
+        self._archive_handle = archive
+
+    @property
+    def num_tokens(self) -> Optional[int]:
+        return self._num_tokens
```
joelgrus: is there a good reason to have private properties with public getters? wouldn't it be simpler to just call the variables `num_tokens` and `uri` and let people access them directly? (that's how the rest of the codebase works)

allennlp/modules/token_embedders/embedding.py
janLuke: Sure.

allennlp/modules/token_embedders/embedding.py
janLuke: The reason is that in this way only the getter is public, not the setter. But I guess nobody would ever modify those fields. I can change that if you like it, for the sake of consistency.

allennlp/modules/token_embedders/embedding.py
joelgrus: 👍

allennlp/modules/token_embedders/embedding.py
joelgrus: people could still modify `_num_tokens` if they wanted to that badly, I vote for just letting them have access to the field

janLuke: Sure, but people should know that modifying (or even accessing) a private field is a bad thing and if they do that is no more responsibility of the developer. Anyway, I'll change it :)

janLuke:
What do you think about implementing ``__len__`` in ``EmbeddingsFileText``? (see my question above)

joelgrus:
if that makes it so you don't have to pass total=num_tokens to tqdm manually, I like it

joelgrus:
looks good, is it ready to merge ?

janLuke:
Yep.

PR #1370, opened 2018-06-13
UD dataset reader

DeNeutoy:


DeNeutoy:
@nelson-liu

nelson-liu:
I think it'd be clearer in general if you didn't mention UD. My understanding is: file format itself is CoNLL-U, UD just happens to use it. I think this would be better as "CoNLLUDatasetReader"

nelson-liu:
oops, i previously thought that you could have stanford dependencies in conllu format. you're right that it seems like conllu is used exclusively for UD (https://nlp.stanford.edu/software/stanford-dependencies.html)

allennlp/data/dataset_readers/universal_dependencies.py
```diff
@@ -0,0 +1,102 @@
...
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+def lazy_parse(text: str, fields: Tuple = DEFAULT_FIELDS):
+    for sentence in text.split("\n\n"):
+        if sentence:
+            yield [parse_line(line, fields)
+                   for line in sentence.split("\n")
+                   if line and not line.strip().startswith("#")
+                  ]
```
nelson-liu: not sure if you meant to do this

allennlp/data/dataset_readers/universal_dependencies.py
```diff
@@ -0,0 +1,102 @@
...
+        with open(file_path, 'r') as conllu_file:
+            logger.info("Reading UD instances from conllu dataset at: %s", file_path)
+
+            for annotation in  lazy_parse(conllu_file.read()):
+
+                yield self.text_to_instance(
+                        [x["form"] for x in annotation],
+                        [x["upostag"] for x in annotation],
+                        [x["deps"][0] for x in annotation]
+                        )
```
nelson-liu: same here

allennlp/data/dataset_readers/universal_dependencies.py
```diff
@@ -0,0 +1,102 @@
...
+                 lazy: bool = False) -> None:
+        super().__init__(lazy)
+        self._token_indexers = token_indexers or {'tokens': SingleIdTokenIndexer()}
+
+    @overrides
+    def _read(self, file_path: str):
+        # if `file_path` is a URL, redirect to the cache
+        file_path = cached_path(file_path)
+
+        with open(file_path, 'r') as conllu_file:
```
nelson-liu: shouldn't need "r" here.

PR #1372, opened 2018-06-13
add numpydocs to requirements

joelgrus:


nelson-liu:
any reason for the 0.8 pin? not against it, just curious.

joelgrus:
I pin out of habit

PR #1380, opened 2018-06-14
Word-level language model with adaptive softmax

LiyuanLucasLiu:
Add word-level language models, the adaptive softmax and other related stuffs.


allennlp/models/language_model.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+    This ``WordLM`` is a word level language model.
+
+    Parameters
+    ----------
+    vocab : ``Vocabulary``, required
+        A Vocabulary, required in order to compute sizes for input/output projections.
+    text_field_embedder : ``TextFieldEmbedder``, required
+        Used to embed the ``tokens`` ``TextField`` we get as input to the model.
+    rnn : ``LMRNN``, required
```
matt-peters: Let's abstract this a little from RNNs to also allow other architectures such as transformer and CNN.  The only requirement on this module is that it is a `Seq2SeqEncoder`, so anything that satisfies that interface will work. An additional added benefit of this approach is that we already have wrappers for multilayer LSTMs in this interface, so most/all of the code in `lm_rnn.py` can be removed.

allennlp/models/language_model.py
```diff
@@ -0,0 +1,206 @@
...
+        The Recurrent Neural Networks for language modeling.
+    softmax : ``AdaptiveSoftmax``
```
matt-peters: Same here with the softmax -- we should allow any module that computes the softmax including full softmax, sampled, etc.

allennlp/models/language_model.py
```diff
@@ -0,0 +1,206 @@
...
+        self.text_field_embedder = text_field_embedder
+
+        if dropout:
+            self.dropout = torch.nn.Dropout(dropout)
+            dropout_list = [self.dropout]
+        else:
+            self.dropout = None
+            dropout_list = []
+
+        if proj:
```
matt-peters: We only need a projection if `self.rnn.get_output_dim() != softmax.get_input_dim()`, so it would be easier to remove `proj` as an option and just automatically add one if necessary.

allennlp/data/iterators/lm_iterator.py
```diff
@@ -0,0 +1,53 @@
...
+
+from overrides import overrides
+
+from allennlp.common import Params
+from allennlp.common.util import ensure_list, is_lazy, lazy_groups_of
+from allennlp.data.instance import Instance
+from allennlp.data.iterators.data_iterator import DataIterator
+from allennlp.data.dataset import Batch
+
+# from ipdb import set_trace
```
joelgrus: remove comment

allennlp/data/iterators/lm_iterator.py
```diff
@@ -0,0 +1,53 @@
...
+
+        batch_num = math.floor(instances_len / self._batch_size)
+        instances_len = instances_len - instances_len % self._batch_size
+
+        for batch_ind in range(batch_num):
+            yield Batch(instances[batch_ind: instances_len: batch_num])
+
+    @classmethod
+    def from_params(cls, params: Params) -> 'BasicIterator':
+        batch_size = params.pop_int('batch_size', 32)
```
joelgrus: this default value doesn't match the default value in the constructor

allennlp/data/iterators/lm_iterator.py
```diff
@@ -0,0 +1,53 @@
...
+        self._lazy = lazy
+
+    @overrides
+    def get_num_batches(self, instances: Iterable[Instance]) -> int:
+        return math.floor(len(ensure_list(instances)) / self._batch_size)        
+
+    @overrides
+    def _create_batches(self, instances: Iterable[Instance], shuffle: bool) -> Iterable[Batch]:
+        
+        instances = ensure_list(instances)
```
joelgrus: it seems like this iterator can't be used lazily? if so, you should probably raise a `ConfigurationError` in the constructor if lazy is True

allennlp/data/iterators/lm_iterator.py
```diff
@@ -0,0 +1,53 @@
...
+        instances_len = len(instances)
+
+        batch_num = math.floor(instances_len / self._batch_size)
```
joelgrus: I would call this `num_batches`, "batch_num" sounds like "batch number 1", "batch number 2", etc...

allennlp/data/iterators/lm_iterator.py
```diff
@@ -0,0 +1,53 @@
...
+        instances_len = instances_len - instances_len % self._batch_size
+
+        for batch_ind in range(batch_num):
+            yield Batch(instances[batch_ind: instances_len: batch_num])
```
joelgrus: I find this code hard to read / reason about. I think the following is much more readable (and does the same thing)

```
num_batches = len(instances) // self._batch_size

# want all batches to be the same size
stop = len(instances) - len(instances) % self._batch_size

for start in range(num_batches):
    yield Batch(instances[start:stop:num_batches])
```

allennlp/data/tokenizers/basic_lm.py
```diff
@@ -0,0 +1,29 @@
+from typing import List
+
+from overrides import overrides
+
+from allennlp.common import Params
+from allennlp.data.tokenizers.token import Token
+from allennlp.data.tokenizers.tokenizer import Tokenizer
+
+@Tokenizer.register("basic_lm")
+class BasicTokenizer(Tokenizer):
```
joelgrus: I feel like it's confusing if the registered name doesn't agree with the class name, so that either the registered name (and the filename) should be "basic" or the classname should be `BasicLmTokenizer`, your call on which is more appropriate

allennlp/data/tokenizers/basic_lm.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self) -> None:
+        return
```
joelgrus: if the constructor doesn't do anything, you can leave it out?

allennlp/data/tokenizers/basic_lm.py
joelgrus: also, can you add a docstring explaining what this tokenizer does and why someone would choose to use it

allennlp/models/language_model.py
```diff
@@ -0,0 +1,206 @@
...
+from allennlp.models.model import Model
+from allennlp.modules import TextFieldEmbedder
+from allennlp.common.checks import check_dimensions_match
+from allennlp.nn import InitializerApplicator, RegularizerApplicator
+
+from allennlp.modules.lm_rnn import LMRNN
+from allennlp.modules.adaptive import AdaptiveSoftmax
+from allennlp.training.metrics.perplexity import Perplexity
+
+@Model.register("WordLM")
```
joelgrus: nit: I think all of our registered names are lowercase, so maybe `word-lm`? just for consistency

allennlp/modules/adaptive.py
```diff
@@ -0,0 +1,166 @@
...
+from allennlp.models.model import Model
+from allennlp.data import Vocabulary
+from allennlp.common import Params
+from allennlp.common.checks import ConfigurationError
+
+from overrides import overrides
+
+from math import sqrt
+
+# from ipdb import set_trace
```
joelgrus: remove comment

allennlp/modules/lm_rnn.py
```diff
@@ -0,0 +1,169 @@
...
+        self.hidden_state = repackage_hidden(new_hidden)
+
+        if self.batch_norm:
+            output_size = out.size()
+            out = self.bn(out.view(-1, self.output_dim)).view(output_size)
+        
+        return out
+
+
+@LMRNN.register("vanilla_RNN")
```
joelgrus: same as above, "vanilla_rnn"

allennlp/training/metrics/perplexity.py
```diff
@@ -0,0 +1,46 @@
+from overrides import overrides
+
+from allennlp.training.metrics.metric import Metric
+
+import math
+
+@Metric.register("ppl")
```
joelgrus: characters are cheap, is there a good reason not to register it as `"perplexity"`?

allennlp/data/iterators/lm_iterator.py
```diff
@@ -0,0 +1,53 @@
...
+from allennlp.common import Params
+from allennlp.common.util import ensure_list, is_lazy, lazy_groups_of
+from allennlp.data.instance import Instance
+from allennlp.data.iterators.data_iterator import DataIterator
+from allennlp.data.dataset import Batch
+
+# from ipdb import set_trace
+
+@DataIterator.register("lm")
+class LMIterator(DataIterator):
```
joelgrus: there's nothing about this iterator that's intrinsic to language modeling? Is there a more generic name we could call it, maybe `StridedIterator` or something like that that indicates what it does?

allennlp/modules/adaptive.py
```diff
@@ -0,0 +1,166 @@
...
+
+                first_target[mask] = self.cutoff[0] + i
+
+                second_target = target[mask].add(-self.cutoff[i])
+
+                second_input = w_in.index_select(0, mask.nonzero().squeeze())
+
+                second_output = self.tail[i](second_input)
+
+                output += self.cross_entropy(second_output, second_target)
```
joelgrus: nit, do you really need a blank line after every line?

allennlp/modules/lm_rnn.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def get_output_dim(self) -> int:
+        raise NotImplementedError
+
+    @classmethod
+    def from_params(cls, params: Params) -> 'LMRNN':
+        choice = params.pop_choice('type', cls.list_available())
+        return cls.by_name(choice).from_params(params)
+        
+class VRNN_Basic(nn.Module):
```
joelgrus: get rid of the underscore in the classname

allennlp/modules/lm_rnn.py
```diff
@@ -0,0 +1,169 @@
...
+
+        if self.batch_norm:
+            output_size = out.size()
+            out = self.bn(out.view(-1, self.output_dim)).view(output_size)
+        
+        return out
+
+
+@LMRNN.register("vanilla_RNN")
+class vanilla_RNN(LMRNN):
```
joelgrus: start classnames with uppercase + get rid of the underscore

allennlp/models/language_model.py
LiyuanLucasLiu: I checked the current implementations of ```Seq2SeqEncoder```, and found the current wrapper would ***not*** wrap RNNs as stateful. However, RNNs based LM need RNNs to be stateful (and ```lm_rnn``` can be viewed as a simple version of stateful RNNs, which does not support mask). 

How about registering ```lm_rnn``` as a ```Seq2SeqEncoder```, and modify the class name from ```lm_rnn``` to ```simple_stateful_rnn```?

allennlp/models/language_model.py
LiyuanLucasLiu: i'll create a registrable class ```softmax_with_nll_loss`` for this

allennlp/models/language_model.py
LiyuanLucasLiu: I think in some cases (e.g., for a densely connected RNNs), i would like to add a non-linear projection layer after RNN (even the output dim of rnn is the same with the input dim of softmax). Maybe we can leave ```proj``` as an option for more flexible?

allennlp/models/language_model.py
matt-peters: We have a mechanism for stateful RNNs now -- https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2seq_encoders/pytorch_seq2seq_wrapper.py .   How does the functionality in `lm_rnn` differ from this?

allennlp/models/language_model.py
matt-peters: OK in that case, I agree we need to leave it.  How about making the default = False since that would be the most common use case with embedding dim == softmax dim.

allennlp/models/language_model.py
LiyuanLucasLiu: Yes, but the current ```__init.py``` does not turn on that option (https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2seq_encoders/__init__.py). Also, the mechanism for stateful RNNS requires the input having a ```mask``` field, which is not necessary for language modeling and would make the implementation slower.

LiyuanLucasLiu:
updated w.r.t. all comments (```basic_stateful_rnn``` is replaced by ```stateful_lstm```, etc. )

allennlp/models/language_model.py
```diff
@@ -0,0 +1,206 @@
...
+
+        initializer(self)
+
+    @overrides
+    def get_metrics(self, reset: bool = False) -> Dict[str, float]:
+        return {metric_name: metric.get_metric(reset) for metric_name, metric in self.metrics.items()}
+        
+    @overrides
+    def train(self, mode=True):
+        self.encoder.reset_states()
```
matt-peters: Why do you `reset_states` here?

allennlp/models/language_model.py
```diff
@@ -0,0 +1,206 @@
...
+                "softmax input dim")
+
+        initializer(self)
+
+    @overrides
+    def get_metrics(self, reset: bool = False) -> Dict[str, float]:
+        return {metric_name: metric.get_metric(reset) for metric_name, metric in self.metrics.items()}
+        
+    @overrides
+    def train(self, mode=True):
```
matt-peters: Maybe this should also call super(...) as pytorch may do some magic in the base class when switching to `train`?

allennlp/models/language_model.py
```diff
@@ -0,0 +1,206 @@
...
+        else:
+            self.after_encoder = None
+
+        self.softmax = softmax
+        if not self.softmax.adaptive and self.softmax.head.weight.size() == self.text_field_embedder.token_embedder_tokens.weight.size():
+            self.softmax.head.weight = self.text_field_embedder.token_embedder_tokens.weight
+
+        self.softmax_in = softmax.get_input_dim()
+
+        self.metrics = {"ppl": Perplexity()}
```
matt-peters: Rename `ppl` to `perplexity`.

allennlp/models/language_model.py
```diff
@@ -0,0 +1,206 @@
...
+        if len(after_encoder_list) > 0:
+            self.after_encoder = nn.Sequential(*after_encoder_list)
+        else:
+            self.after_encoder = None
+
+        self.softmax = softmax
+        if not self.softmax.adaptive and self.softmax.head.weight.size() == self.text_field_embedder.token_embedder_tokens.weight.size():
+            self.softmax.head.weight = self.text_field_embedder.token_embedder_tokens.weight
+
+        self.softmax_in = softmax.get_input_dim()
```
matt-peters: rename self.softmax_in to something with `size` or `dim` in the name to help disambiguate from other softmax input related objects (tensors, etc)

allennlp/models/language_model.py
```diff
@@ -0,0 +1,206 @@
...
+            The expected prediction of the language model. If set to be None, the output would contains
+            ```emb_encoder_out```, which is the output of the LMRNNs. Otherwise, the output would be the loss
+            alone.
+
+        Returns
+        -------
+        An output dictionary consisting of:
+        emb_encoder_out : torch.FloatTensor, optional
+            A tensor of shape ``(batch_size, num_tokens, tag_vocab_size)`` representing
+            a distribution of the tag classes per word.
```
matt-peters: What are the tag classes?  Shouldn't this be the next token?

allennlp/models/language_model.py
```diff
@@ -0,0 +1,206 @@
...
+            A scalar loss to be optimised.
+
+        """
+
+        embedded_text_input = self.text_field_embedder(input_tokens)
+
+        if self.dropout:
+            embedded_text_input = self.dropout(embedded_text_input)
+
+        # set_trace()
```
matt-peters: remove debug statement

allennlp/models/language_model.py
```diff
@@ -0,0 +1,206 @@
...
+        emb_encoder_out = self.encoder(embedded_text_input, 
+            embedded_text_input.new_ones((embedded_text_input.size(0), embedded_text_input.size(1))))
+
+        # set_trace()
```
matt-peters: please no commented out debug statements

allennlp/data/tokenizers/basic.py
```diff
@@ -0,0 +1,33 @@
+from typing import List
+
+from overrides import overrides
+
+from allennlp.common import Params
+from allennlp.data.tokenizers.token import Token
+from allennlp.data.tokenizers.tokenizer import Tokenizer
+
+@Tokenizer.register("basic")
+class BasicTokenizer(Tokenizer):
```
matt-peters: BasicTokenizer to me implies just tokenization w/o BOS/EOS tokens -- so can you rename this to something like `TokenizerEOS` or `BasicTokenizerWithEOS` ?

allennlp/data/tokenizers/basic.py
```diff
@@ -0,0 +1,33 @@
...
+    A ``BasicTokenizer`` splits strings of text into tokens by treating space characters as 
+    separator, and pad a special EOF token to the end of each sentence. This tokenizer is 
+    relative fast and only provides the very basic function as a tokenizer (suitable for
+    large, already tokenized corpora, e.g., the wiki103 dataset).
+    """
+
+    @overrides
+    def tokenize(self, text: str) -> List[Token]:
+        if text and not text.isspace():
+            return [Token(word) for word in text.split()] + [Token("<eof>")]
```
matt-peters: We already have `</S>` for the ELMo LM end-of-sentence token, so using `<eof>` here will get confusing.  Is it possible to handle the end-of-sentence token here in the same manner as the existing ELMo code?

allennlp/models/language_model.py
```diff
@@ -0,0 +1,206 @@
...
+
+    @overrides
+    def forward(self,  # type: ignore
+                input_tokens: Dict[str, torch.LongTensor],
+                output_tokens: Dict[str, torch.LongTensor] = None) -> Dict[str, torch.Tensor]:
+        # pylint: disable=arguments-differ
+        """
+        Parameters
+        ----------
+        input_tokens : Dict[str, torch.LongTensor], required
```
matt-peters: Docstring should contain something module specific and should remove the standard template.

allennlp/models/language_model.py
```diff
@@ -0,0 +1,206 @@
...
+        for module in self.children():
+            module.train(mode)
+        return self
+
+    @overrides
+    def eval(self):
+        return self.train(False)
+
+    @overrides
+    def forward(self,  # type: ignore
```
matt-peters: A high level question: where in the data processing pipeline are `input_tokens` and `output_tokens` computed? I didn't see it in the `BasicTokenizer` or `StridedIterator`?

allennlp/models/language_model.py
LiyuanLucasLiu: I'm not sure whether it's the standard practice for language modeling. But at the start of evaluation or a new training epoch, i would reset the hidden states to all zeros. Since I'm using the default trainer, I choose to add ```reset_states()``` here as a workaround.

allennlp/models/language_model.py
LiyuanLucasLiu: They're computed in the original ```LanguageModelingReader``` https://github.com/allenai/allennlp/blob/master/allennlp/data/dataset_readers/language_modeling.py

allennlp/data/tokenizers/basic.py
LiyuanLucasLiu: Sure, no problem :-)

allennlp/data/tokenizers/basic.py
LiyuanLucasLiu: Yes, ```BasicTokenizerWithEOS``` sounds good to me

allennlp/models/language_model.py
LiyuanLucasLiu: Yes, it sounds good to me :-)

allennlp/models/language_model.py
Froskekongen: @LiyuanLucasLiu Just wondering if you are planning to complete this PR. Would be very useful.

joelgrus:
this will be superseded by the code we're bringing over from calypso

PR #1381, opened 2018-06-14
create ccgbank dataset reader

joelgrus:
I made it read in all the fields, you can ignore the ones you don't want


allennlp/data/dataset_readers/ccgbank.py
```diff
@@ -0,0 +1,124 @@
...
+from allennlp.data.token_indexers import TokenIndexer, SingleIdTokenIndexer
+from allennlp.data.tokenizers import Token
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+@DatasetReader.register("ccgbank")
+class CcgBankDatasetReader(DatasetReader):
+    """
+    Reads data from the CCGbank dataset. In particular, reads the flattened "leaves"
```
nelson-liu: Not too sure what `reads the flattened "leaves"` means --- are you referring to the fact that the input is a linearized tree?
Also: perhaps we should mention that it reads data from a file where each line is a tree from the ccgbank dataset or something? Since this doesn't read straight from ccgbank.

nelson-liu:
looks good to me. feel free to merge when CI passes.

Issue #1324, opened 2018-05-31
Create a Dockerfile that includes a basic pip installation of AllenNLP


schmmd:
To make using AllenNLP easy for downstream users, we should create a simple Dockerfile that has a pip installation of AllenNLP.  This would make it easy for users to use AllenNLP as a library or a command in a Docker environment, and would provide a base image from which people could run AllenNLP jobs on Beaker.

```
FROM python:3.6.3-jessie

...

RUN pip install allennlp

ENTRYPOINT ["allennlp"]
```

schmmd:
Here's [a full example](https://github.com/schmmd/allennlp/blob/aee4c5279549741eb6ec873950aee98f5015a6b4/Dockerfile) of what the Dockerfile could look like (beautifully simple).

PR #1389, opened 2018-06-18
WIP: Add a Dockerfile for downstream dependencies.

schmmd:
Fixes https://github.com/allenai/allennlp/issues/1324.


schmmd:
@joelgrus what do you think of this re-organization?

`ci/Dockerfile` would be used on:
* PRs
* Commits to master
  * On commits to master, it pushes images to `allennlp/build`.

`Dockerfile` would be only used on commits to master.
* On commits to master, it pushes images to `allennlp/commit`.

joelgrus:
that seems fine to me, and there would be a CI step on `allennlp/commit` that runs `test-install`?

schmmd:
@joelgrus yes--before pushing ;-)

DeNeutoy:
The `/Dockerfile` won't actually be used to run the tests, is that right?

I'm not totally sure i've understood the idea here but I think if we are going to push an image with the updated pip installed package, don't we need to pip install the commit that was just merged? I don't think the downstream Dockerfile would currently support doing that.

schmmd:
@DeNeutoy `/Dockerfile` won't be used for our continuous integration, but it will be use to run `test-install` for the pip distribution.  Once this is merged, the published image can be used as the base image in https://github.com/allenai/allennlp-demo/blob/master/Dockerfile.

You're right.  I need to modify `/Dockerfile` to correctly use git checkouts for pip installs.

DeNeutoy:
Got it. Sounds good to me - a few clarifications:

1. How do we know when the pip install fails if it doesn't run on PRs? I guess this is just because it is slow (although it seems like they could run in parallel if we had more machines - is this not the case?).

2. This is still separate to the images we release for allennlp versions, right?

schmmd:
@DeNeutoy 

Re 1--yes ideally it would run on PRs.  I don't want to do that as a first step because it's slow.

Re 2--yes it is separate from what we publish for releases.  However, soon I'd like to change what we publish as releases to be the pip Dockerfiles, rather than the CI Dockerfiles.  Do you see any concerns with my proposal here?

DeNeutoy:
No, sounds good - I was just checking.

schmmd:
LMK if you think of anything.  I'm a bit nervous changing our version Docker releases to be `pip` installations under the same namespace in case anyone is depending on them--but I'm pretty sure nobody is.

schmmd:
@DeNeutoy I pushed a commit to allow source revision pip installations.

Unfortunately `test-install` doesn't pass.  (`FileNotFoundError: [Errno 2] No such file or directory: 'java': 'java'`).

schmmd:
OK--after lunch I'm going to merge this (after approval @joelgrus) and set up CI to use `ci/Dockerfile` rather than `Dockerfile`.  Then I'll set up a task that build `Dockerfile` and runs test-install on Master Commit.  However--this is going to fail at first since some dependencies (such as Java) are missing.  I'm somewhat inclined to remove those tests (https://github.com/allenai/allennlp/blob/master/allennlp/commands/test_install.py#L61) from test-install--but we'll get there when we get there.

matt-gardner:
I didn't look at this in detail at all (just deleting email that I don't need during some downtime); just piping in to say that `run_with_beaker.py` uses the dockerfile, so make sure that any reorganization doesn't break that script.  If you already made sure of that, great.

schmmd:
@matt-gardner that's a good point (and one I overlooked).  It's a one-line change to fix.

schmmd:
Alright--sorry for the added complexity the first go-round.

I've changed this PR to simply add a `Dockerfile.pip`, which ultimately causes far fewer changes but accomplishes the same goals.  On commits to master I will publish this Dockerfile to publish to `allennlp/pip/commit`.

PR #1393, opened 2018-06-18
Add a TensorField

nelson-liu:
This PR adds a field for dealing with arbitrary Tensors in your dataset.


allennlp/tests/data/fields/tensor_field_test.py
```diff
@@ -0,0 +1,49 @@
...
+from numpy.testing import assert_allclose
+import pytest
+import torch
+
+from allennlp.common.checks import ConfigurationError
+from allennlp.common.testing import AllenNlpTestCase
+from allennlp.data.fields import TensorField
+
+
+class TestLabelField(AllenNlpTestCase):
```
joelgrus: rename the class?

allennlp/data/fields/tensor_field.py
```diff
@@ -0,0 +1,36 @@
...
+    @overrides
+    def get_padding_lengths(self) -> Dict[str, int]:  # pylint: disable=no-self-use
+        return {}
+
+    @overrides
+    def as_tensor(self,
+                  padding_lengths: Dict[str, int],
+                  cuda_device: int = -1) -> torch.Tensor:
+        # pylint: disable=unused-argument,not-callable
+        return self.tensor if cuda_device == -1 else self.tensor.cuda(cuda_device)
```
joelgrus: technically this does the wrong thing if the original tensor is on the GPU and then you call it with cuda_device=-1. that's a pretty weird case, but probably we should handle it correctly.

I believe under 0.4 you could do something like

```
device = torch.device(f"cuda:{cuda_device}" if cuda_device >= 0 else "cpu")
return self.tensor.to(device)
```

allennlp/tests/data/fields/tensor_field_test.py
nelson-liu: whoops yes, sorry.

allennlp/data/fields/tensor_field.py
nelson-liu: should this change be made for all the other fields as well?

joelgrus: no, because they're not starting with existing tensors

nelson-liu: right, thanks.

nelson-liu:
@joelgrus : @DeNeutoy and I had an off-github conversation, and we decided that it might make the most sense to just slightly rework the `ArrayField` to work with torch Tensor input. does this sound reasonable to you?

joelgrus:
I'm cool either way

PR #1394, opened 2018-06-19
Make ArrayField work with torch Tensor input

nelson-liu:
Continuation of #1393


allennlp/data/fields/array_field.py
```diff
@@ -1,21 +1,32 @@
...
 import numpy
 import torch
 from overrides import overrides
 
+from allennlp.common.checks import ConfigurationError
 from allennlp.data.fields.field import Field
 
 
-class ArrayField(Field[numpy.ndarray]):
+class ArrayField(Field[Union[numpy.ndarray, torch.Tensor]]):
```
nelson-liu: unsure if this is correct

allennlp/data/fields/array_field.py
```diff
@@ -30,23 +41,28 @@ def as_tensor(self,
...
             slicing_shape = slicing_shape + [0 for _ in range(len(max_shape) - len(self.array.shape))]
         slices = [slice(0, x) for x in slicing_shape]
-        return_array[slices] = self.array
+        # torch seems to have trouble with this, so we cast to numpy
+        if self.array_type == "numpy":
+            return_array[slices] = self.array
+        else:
+            return_array[slices] = self.array.cpu().numpy()
         tensor = torch.from_numpy(return_array)
         return tensor if cuda_device == -1 else tensor.cuda(cuda_device)
```
joelgrus: if I understand this correctly, in the tensor case, if your tensors are on the GPU (which is how I assume you'd want them in the GPU case), then `as_tensor` moves them to the CPU, convert them to numpy arrays, and then move them back to the GPU?!

nelson-liu: In the Tensor case, it’s actually more likely (at least for my application) that the tensors are on CPU —- there isn’t enough GPU memory to store them all there. This is still a fair criticism, though. The fancy indexing didn’t work out of the box in pytorch, so I’ll see if I can find a workaround.

nelson-liu:
Thanks for the comments @joelgrus , I’ll use the structure you suggested in the next revision

allennlp/data/fields/array_field.py
DeNeutoy: If they are going to be on the CPU anyway, we should just close these PRs and you should use an ArrayField - torch tensors will share memory with numpy arrays, so calls to and from numpy will be very fast:

```
(allennlp) markn@markn ~ $ python -mtimeit -s "import torch" "torch.from_numpy(torch.randn([20, 30, 1000]).numpy())"
100 loops, best of 3: 4.86 msec per loop
```
A full round trip for a pretty large tensor takes 5/1000ths of a second. 

I thought the reason for this PR was because you had some tensors transitively on the GPU, which you wanted to keep there. If not, perhaps underoptimise before you over optimise 😃

nelson-liu: You're right, that solution seems to be the simplest and just work. Thanks @DeNeutoy

nelson-liu:
thanks again for the comments and feedback @joelgrus / @DeNeutoy , this was quite helpful.

PR #1401, opened 2018-06-20
Fix multiple GPU training after upgrading to pytorch 0.4

matt-peters:


matt-peters:
FWIW, I'm seeing this new warning when running the `test_trainer_can_run_multiple_gpu` in `allennlp/tests/training/trainer_test.py`:

```
allennlp/tests/training/trainer_test.py::TestTrainer::test_trainer_can_run_multiple_gpu
  /home/matthewp/git/allennlp/allennlp/modules/encoder_base.py:116: UserWarning: RNN module weights are not part of single contiguous chunk of memory. This means they need to be compacted at every call, possibly greatly increasing memory usage. To compact weights again call flatten_parameters().
    module_output, final_states = module(packed_sequence_input, initial_states)
  /home/matthewp/git/allennlp/allennlp/modules/encoder_base.py:116: UserWarning: RNN module weights are not part of single contiguous chunk of memory. This means they need to be compacted at every call, possibly greatly increasing memory usage. To compact weights again call flatten_parameters().
    module_output, final_states = module(packed_sequence_input, initial_states)
```

So perhaps there additional changes required with 0.4 to avoid performance regressions.  But for now, the unit test passes on a multiple GPU machine.

matt-peters:
Good to know -- it would be worthwhile to consider that approach when we revisit the multiple GPU training implementation.

PR #1397, opened 2018-06-19
fix elmo command to use line indices and disallow empty lines

DeNeutoy:
Fixes #1396 Fixes #1257 

- Replaces sentence indexes with integer line indices
- Doesn't ignore duplicate lines in the dataset
- raises an error if your input file contains blank lines, rather than silently creating a dataset which doesn't align with the input file

PR #1404, opened 2018-06-21
Add --include-sentence-indices flag to ELMo command

nelson-liu:
Since #1397 , it's no longer possible to use the HDF5 files generated by the elmo command unless you have the original dataset of sentences. This PR adds a flag to the elmo command that writes a JSON-serialized mapping of sentence->key to the `sentence_indices` key, so the HDF5 file can be useful even without the original dataset.

The typical use will be something like this, if you're interested in getting the embeddings for a particular sentence in reasonable time:

```
import json

sentence = something

with h5py.File(hdf5_path, 'r') as h5py_file:
    sentence_to_index = json.loads(h5py_file.get("sentence_to_index")[0])
    embedding = h5py_file[sentence_to_index[sentence]]
```


nelson-liu:
ok, thanks. Even if it's the default behavior, it seems reasonable to have a flag to optionally turn it off? any ideas for what that flag should be called? i was thinking `--remove-sentence-indices` but that's pretty unclear...thoughts?

DeNeutoy:
`--use-simple-sentence-indices`?

joelgrus:
I would call it `--forget-sentences` or something like that

nelson-liu:
@DeNeutoy / @joelgrus  want to take another look?

gicheonkang:
@nelson-liu Hi, I'm desperately looking for `--include-sentence-indices` command, but I cannot find that in version 0.5.1. Not yet released?

nelson-liu:
hi @gicheonkang --- it's not yet released (though we should _really_ have one, I'll bring it up today). In the meanwhile, you can install allennlp through the master branch with `pip install git+https://github.com/allenai/allennlp.git@master`.

PR #1405, opened 2018-06-21
Add support for prevention of parameter initialization which match the given regexes

HarshTrivedi:
Add support for overrided prevention of parameter initialization which match the given regexes

## Commit Summary

- Makes following scenario easy to handle: continue with all other configured initialization schemes for various networks but skip parameters matching `".*transfer.*"`
- Add `PreventRegexInitializer` and register it with `"prevent"`. 
- Allow `InitializerApplicator` to apply any `initializer` to any parameter only if prevention regexes are not matched.
- Add unit tests with example usage

## Need for this feature:

- If one wants to transfer modules from pretrained model, one would want to initialzie weights of new model preventing the transfered modules from this initialization. 
- Negative matching regex give a easy selective handle on which parameters are transferred and must not be initialized. Eg. continue with all other configured initialization schemes for various networks but skip parameters matching `".*(transfer)|(pretrained).*"`.
- Also note that given regex A and regex B, in general it is not easy to write regex that matches A but not B. So bringing the required effect with plain regex can be quite tedious.

## Example Usage

```python
class MyNet(torch.nn.Module):
    # Typically, transfer_model s would loaded with trained weights by : 
    # eg. transfer_mode1 = load_archive(params.pop("model1.tar.gz"))  
    # in from_params of main MyModel of which MyNet is part of.
    def __init__(self, transfer_model1, transfer_model2):
        super(Net, self).__init__()
        self.linear_1 = torch.nn.Linear(5, 10)
        self.linear_2 = torch.nn.Linear(10, 5)
        self.linear_3_transfer = transfer_model1.linear_3 	 # Note this
        self.linear_4_transfer = transfer_model2.linear_4        # Note this
        self.pretrained_conv = transfer_model2.conv 	         # Note this
    def forward(self, inputs):  # pylint: disable=arguments-differ
        pass
```

```
// experiment.json
...
    {"initializer": [
        [".*linear.*", {"type": "xavier_normal"}], ...
        [".*conv.*", {"type": "kaiming_normal"}],
        [".*_transfer.*", "prevent"],                           // Note this
        [".*pretrained.*",{"type": "prevent"}]                 // Note this
    ]}
...
```
The following would initialize linear and conv layers except `linear_3_transfer`, 
`linear_4_transfer`, `pretrained_conv` which match regexes corresponding to "prevent" type.

```python
net = MyNet()
# load initializer from above configs
initializer(net)
```


allennlp/nn/initializers.py
```diff
@@ -193,9 +214,22 @@ def __init__(self, initializers: List[Tuple[str, Initializer]] = None) -> None:
...
         """
-        self._initializers = initializers or []
+
+        self._initializers: List[Tuple[str, Initializer]] = []
+        initializers = initializers or []
+        self._prevent_regex = None
+        prevent_regexes = []
+        for initializer in initializers:
+            if isinstance(initializer[1], PreventRegexInitializer):
+                prevent_regexes.append(initializer[0])
```
DeNeutoy: To avoid having to type check for this particular initialiser, I think it would be better to add another argument to the `InitializerApplicator`: `prevent_regexes: List[str] = None`. 

This way, you can just pass a list of regexes for parameters to be ignored. To instanciate this class we just need to modify `from_params` to pop the "prevent_regexes" key before building the other initialisers You can also delete the `PreventRegexInitializer` class in this case. Do you see any problems with that?

allennlp/nn/initializers.py
```diff
@@ -193,9 +214,22 @@ def __init__(self, initializers: List[Tuple[str, Initializer]] = None) -> None:
...
+            else:
+                self._initializers.append(initializer)
+        if prevent_regexes:
+            self._prevent_regex = "("+")|(".join(prevent_regexes)+")"
```
DeNeutoy: Spaces around operators would make this a lot easier to read.

allennlp/nn/initializers.py
HarshTrivedi: Yes I agree that `PreventRegexInitializer` seems hacky solution and its better to delete it. Having `prevent_regexes: List[str] = None` argument in `InitializerApplicator` and removing `PreventRegexInitializer` seems much cleaner. I will make these changes. Thank you for the suggestions.

However, poping `"prevent_regexes"` key in `from_params` of `InitializerApplicator` is not possible since `params` argument is a List of tuples `List[Tuple[str, Params]]` as following:

```
[
    ["parameter_regex_match1",
        {"type": "normal","mean": 0.01, "std": 0.1}
    ],
    ["parameter_regex_match2", "uniform"]
]
```
How about the following: To add prevent regex we add `(".*_my_prevention_regex_*",  "prevent")` tuples in the above list. And in `from_params`, from `params` list, I extract and remove the tuples with second item "prevent" and make `prevent_regexes` out of first items. And finally initialize `Initializer`s:
```python
def from_params(cls, params: List[Tuple[str, Params]]) -> "InitializerApplicator":
    prevent_regexes = [ param[0] for param in params if param[1] is "prevent"]
    params = [ param for param in params if param[1] is not "prevent"]

    initializers = [(name, Initializer.from_params(init_params, prevent_regexes)) for name, init_params in params]
    return InitializerApplicator(initializers)
```
How does this seem? Please correct me if I misinterpreted your suggestion.

allennlp/nn/initializers.py
HarshTrivedi: True. I will fix this. Thanks

allennlp/nn/initializers.py
DeNeutoy: Ah yes, my bad - go ahead, that seems equally as good.

HarshTrivedi: Cool, Thanks!

HarshTrivedi:
Incorporated suggestions.

PR #1409, opened 2018-06-21
replace hocon with jsonnet

joelgrus:
as I mentioned, this ended up being more involved than I expected. (although the PR doesn't look that involved, it just feels involved)

here are the interesting parts:

(1) I had to add some functions to params.py to make up for some of the (sometimes) implicit pyhocon behavior we were counting on (e.g. `with_fallback`, `unflatten`, etc...)
(2) jsonnet doesn't do environment variable substitution per se, so I had to use their `std.extVar` construct and then feed in all the environment variables from the Python side
(3) jsonnet is also stricter about missing quotes and such
(4) in particular, this meant I had to change a couple of our training configs

the last is a breaking change and possibly affects people downstream 😬

to that end, I added a new test that calls `Params.from_file` for everything in the training_config directory, which seems like a good idea anyway


allennlp/common/params.py
```diff
@@ -281,11 +345,13 @@ def from_file(params_file: str, params_overrides: str = "") -> 'Params':
         """
         # redirect to cache, if necessary
         params_file = cached_path(params_file)
+        ext_vars = dict(os.environ)
```
DeNeutoy: "external"? "extra"?

joelgrus:
that actually wasn't the problem (well, it was *a* problem but not THE problem).

the problem was that pyhocon deals with backslashes in regexes by doubling them on the way in (which is wrong behavior) but json.dumps writes them out as-is, so that if you do a round-trip of a string that contains any number of backslashes, you end up with extra backslashes, which is what happened to us. 

I am also sort of surprised that the sniff tests work, let me think about why that is

joelgrus:
ha, the reason they work is that because of the error I just described, the `config.json` that's in `model.tar.gz` was written out with double-backslashes and hence is correct JSON 😭

PR #1416, opened 2018-06-25
Add support for configurable vocabulary extension

HarshTrivedi:
Add full support for configurable vocabulary extension

1. Add general configurable support for vocabulary extension (essential for transfer learning scenarios, fine tuning).
2. Make fine-tune command extend the saved-model-vocabulary with current dataset instances by default rather than only banking on saved-model-vocabulary.
3. Add extensive test cases for new API. Check test-cases for more example usages.

## General Vocab Extension Support:
Api now allows toggle `extend` in vocabulary params to configure whether to use `directory_path` for extending vocabulary, or use it alone and ignore the the dataset. The current API only allows either to load vocabulary from directory or load from dataset instances. However, vocabulary extension is important for undertaking any kind of transfer learning, fine tuning etc.

**Example Usage:**

In the following case, vocabulary will be first loaded from `directory_path`, then all the tokens read by relevant datasetreader from dataset in this config file will be added onto this vocabulary. Index2word and word2index mapping of tokens in `directory_path` vocabulary will be retained in the extended final vocabulary. Hence models trained using `directory_path` vocabulary will continue to work with extended vocabulary as well. This setting can be used in `train` or even in `make-vocab` command.
```
"vocabulary": {
    "directory_path": "vocabulary_dir_path"
    "extend": true
}
```
In the following case, vocabulary will be first loaded from `directory_path`, and all the tokens read from dataset mentioned in experiment config file will be ignored. Note: Only this behavior was supported earlier. Since `extend` is by default False, it is fully backward compatible.
```
"vocabulary": {
    "directory_path": "vocabulary_dir_path"
    "extend": false
}
```
Vocab extension feature is very much essential for doing any kind of transfer learning properly. Generally one wants to fine-tune or transfer model parts to train on different dataset. Ideally, tokens from this dataset must contribute to vocabulary. Moreover, index mappings of original vocab must be retained in extended vocab because transferred modules were learning with those mappings.

With this feature in place and prevent-initialization support (#1405), one can easily handle wide range of transfer learning kind of scenarios with `train` command itself - what is supported by fine-tune being the simplest case. Will discuss this in separate issue/pr.

## Add vocab extension support in make-vocab and make it default 

Apart of allowing to pass `"extend": false/true` in vocab_params, this PR adds extra optional argument in Vocab.from_params: `extend_vocab: Vocabulary = None`. If `extend_vocab` is explicitly passed, then `extend_vocab` will be taken as starting vocabulary and words/tokens read by datasetreader will be added on to it. The word2index and index2token of mappings of tokens in `extend_vocab` will be retained in final `extended_vocab` output.

**Example Usage:**
```python
extended_vocab = Vocabulary.from_params(Params({}), instances, original_vocab)
```
This feature was specifically added to have vocab extension in `fine-tune` command. Instead of ignoring the tokens from dataset on which tuning is being done, it starts from `saved_model.vocab` and adds on newer dataset words/tokens without disrupting the index2token and token2index mappings. Retaining these mappings in extended vocabulary is essential for `saved_model` to work.

## Vocabulary extension details and examples:

If `original_vocab` is to be extended by tokens in `instances`, for each namespace:
1. If the namespace is in `original_vocab` and not in `instances`, then that namespace will be same in `extended_vocab` as `original_vocab`.
2. If the namespace is not in `original_vocab` and in `instances`, then that namespace will be created from `instances` as usual.
3. if the namespace is common between `original_vocab` and `instances`, then its padded / non-padded setting should be same for it to be extendable. Below is an example for a common namespace (non-padded):

**Example**:
```
#original_vocab index2token
apple           #0->apple
bat             #1->bat
cat             #2->cat
```
Token to be extended with: `cat, an, apple, banana, atom, bat` in same namespace.
```
#extended_vocab: index2token
apple           #0->apple
bat             #1->bat
cat             #2->cat
an              #3->an
atom            #4->atom
banana          #5->banana
```

### DIFF
--- a/allennlp/data/vocabulary.py
+++ b/allennlp/data/vocabulary.py
@@ -429,14 +429,13 @@ class Vocabulary:
 
         if counter is not None:
             # Make sure vocabulary extension is safe.
-            # Common named namespaces must have same setting for padded=True/False
-            tob_added_namespaces = list(counter.keys())
-            existing_namespaces = self.get_all_namespaces()
-            for namespace in tob_added_namespaces:
-                if namespace in existing_namespaces:
-                    existing_namespace_padded = namespace in self._non_padded_namespaces
-                    tob_added_namespace_padded = namespace in non_padded_namespaces
-                    if existing_namespace_padded != tob_added_namespace_padded:
+            for namespace in counter:
+                if namespace in self.get_all_namespaces():
+                    # if new namespace was already present
+                    # Either both should be padded or none should be.
+                    original_padded = namespace not in self._non_padded_namespaces
+                    extension_padded = namespace not in non_padded_namespaces
+                    if original_padded != extension_padded:
                         raise ConfigurationError("Common namespace {} has conflicting ".format(namespace)+
                                                  "setting of padded = True/False. "+
                                                  "Hence extension cannot be done.")
--- a/allennlp/data/vocabulary.py
+++ b/allennlp/data/vocabulary.py
@@ -47,8 +47,8 @@ class _NamespaceDependentDefaultDict(defaultdict):
     Parameters
     ----------
     non_padded_namespaces : ``Iterable[str]``
-        A set of strings describing which namespaces are not padded.  If a namespace (key)
-        is missing from this dictionary, we will use :func:`namespace_match` to see whether
+        A set / list / tuple of strings describing which namespaces are not padded.  If a namespace
+        (key) is missing from this dictionary, we will use :func:`namespace_match` to see whether
         the namespace should be padded.  If the given namespace matches any of the strings in this
         list, we will use ``non_padded_function`` to initialize the value for that namespace, and
         we will use ``padded_function`` otherwise.
--- a/allennlp/data/vocabulary.py
+++ b/allennlp/data/vocabulary.py
@@ -76,8 +76,9 @@ class _NamespaceDependentDefaultDict(defaultdict):
         dict.__setitem__(self, key, value)
         return value
 
-    def add_non_padded_namespaces(self, namespaces: Sequence[str]):
-        self._non_padded_namespaces.extend(namespaces)
+    def add_non_padded_namespaces(self, non_padded_namespaces: Sequence[str]):
+        # add non_padded_namespaces which weren't already present
+        self._non_padded_namespaces.extend(non_padded_namespaces)
         self._non_padded_namespaces = list(set(self._non_padded_namespaces))
 
 class _TokenToIndexDefaultDict(_NamespaceDependentDefaultDict):
@@ -177,13 +178,6 @@ class Vocabulary:
         If given, this is a list of tokens to add to the vocabulary, keyed by the namespace to add
         the tokens to.  This is a way to be sure that certain items appear in your vocabulary,
         regardless of any other vocabulary computation.
-    extend_vocab : Vocabulary, optional
-        If ``extend_vocab`` is passed, ``counter`` and ``tokens_to_add`` dictionary will be used to
-        extend the ``extend_vocab`` and that will be returned vocabaulary. If ``counter`` is None
-        and ``extend_vocab`` is passed, ``extend_vocab`` with extension of ``tokens_to_add`` will be
-        returned vocabulary. If ``counter`` and ``tokens_to_add`` are None and ``extend_vocab`` is
-        passed, only ``extend_vocab`` will be returned vocabulary. If ``extend_vocab`` is None,
-        vocabulary will be build out of ``counter`` and ``tokens_to_add``.
     """
     def __init__(self,
                  counter: Dict[str, Dict[str, int]] = None,
@@ -192,72 +186,24 @@ class Vocabulary:
                  non_padded_namespaces: Sequence[str] = DEFAULT_NON_PADDED_NAMESPACES,
                  pretrained_files: Optional[Dict[str, str]] = None,
                  only_include_pretrained_words: bool = False,
-                 tokens_to_add: Dict[str, List[str]] = None,
-                 extend_vocab=None) -> None:
-                 # extend_vocab: Vocabulary = None doest not work. Suggestions?
+                 tokens_to_add: Dict[str, List[str]] = None) -> None:
         self._padding_token = DEFAULT_PADDING_TOKEN
         self._oov_token = DEFAULT_OOV_TOKEN
-        if not isinstance(max_vocab_size, dict):
-            int_max_vocab_size = max_vocab_size
-            max_vocab_size = defaultdict(lambda: int_max_vocab_size)  # type: ignore
-        if not extend_vocab:
-            self._non_padded_namespaces = non_padded_namespaces
-            self._token_to_index = _TokenToIndexDefaultDict(non_padded_namespaces,
-                                                            self._padding_token,
-                                                            self._oov_token)
-            self._index_to_token = _IndexToTokenDefaultDict(non_padded_namespaces,
-                                                            self._padding_token,
-                                                            self._oov_token)
-        else:
-            incompatible_padding_or_oov_token = self._padding_token != extend_vocab.get_padding_token() \
-                                                or self._oov_token != extend_vocab.get_oov_token()
-            if incompatible_padding_or_oov_token:
-                raise ConfigurationError("Incompatible padding and/or oov_token")
-            dataset_namespaces = list(counter.keys())
-            extend_vocab_namespaces = extend_vocab.get_all_namespaces()
-            for namespace in extend_vocab_namespaces:
-                if namespace in dataset_namespaces:
-                    instances_padding = namespace in non_padded_namespaces
-                    extend_vocab_padding = namespace in extend_vocab.get_non_padded_namespaces()
-                    namespace_padding_combatible = (instances_padding == extend_vocab_padding)
-                    if not namespace_padding_combatible:
-                        raise ConfigurationError("For {} namespace extend_vocab has ".format(namespace)+
-                                                 "padding: {} but dataset instances ".format(extend_vocab_padding)+
-                                                 "have padding: {}. ".format(instances_padding)+
-                                                 "For extension compatibility both should be same.")
-            self._non_padded_namespaces = list(set(list(extend_vocab.get_non_padded_namespaces())
-                                                   +list(non_padded_namespaces)))
-            self._token_to_index = extend_vocab.get_token_to_index()
-            self._index_to_token = extend_vocab.get_index_to_token()
-            self._token_to_index.add_non_padded_namespaces(non_padded_namespaces)
-            self._index_to_token.add_non_padded_namespaces(non_padded_namespaces)
-        min_count = min_count or {}
-        pretrained_files = pretrained_files or {}
-        if counter is not None:
-            for namespace in counter:
-                if namespace in pretrained_files:
-                    pretrained_list = _read_pretrained_tokens(pretrained_files[namespace])
-                else:
-                    pretrained_list = None
-                token_counts = list(counter[namespace].items())
-                token_counts.sort(key=lambda x: x[1], reverse=True)
-                max_vocab = max_vocab_size[namespace]
-                if max_vocab:
-                    token_counts = token_counts[:max_vocab]
-                for token, count in token_counts:
-                    if pretrained_list is not None:
-                        if only_include_pretrained_words:
-                            if token in pretrained_list and count >= min_count.get(namespace, 1):
-                                self.add_token_to_namespace(token, namespace)
-                        elif token in pretrained_list or count >= min_count.get(namespace, 1):
-                            self.add_token_to_namespace(token, namespace)
-                    elif count >= min_count.get(namespace, 1):
-                        self.add_token_to_namespace(token, namespace)
-
-        if tokens_to_add:
-            for namespace, tokens in tokens_to_add.items():
-                for token in tokens:
-                    self.add_token_to_namespace(token, namespace)
+        self._non_padded_namespaces = list(non_padded_namespaces)
+        self._token_to_index = _TokenToIndexDefaultDict(non_padded_namespaces,
+                                                        self._padding_token,
+                                                        self._oov_token)
+        self._index_to_token = _IndexToTokenDefaultDict(non_padded_namespaces,
+                                                        self._padding_token,
+                                                        self._oov_token)
+        # Made an empty vocabulary, now extend it.
+        self.extend(counter,
+                    min_count,
+                    max_vocab_size,
+                    non_padded_namespaces,
+                    pretrained_files,
+                    only_include_pretrained_words,
+                    tokens_to_add)
 
     def save_to_files(self, directory: str) -> None:
         """
@@ -375,8 +321,7 @@ class Vocabulary:
                        non_padded_namespaces: Sequence[str] = DEFAULT_NON_PADDED_NAMESPACES,
                        pretrained_files: Optional[Dict[str, str]] = None,
                        only_include_pretrained_words: bool = False,
-                       tokens_to_add: Dict[str, List[str]] = None,
-                       extend_vocab=None) -> 'Vocabulary':
+                       tokens_to_add: Dict[str, List[str]] = None) -> 'Vocabulary':
         """
         Constructs a vocabulary given a collection of `Instances` and some parameters.
         We count all of the vocabulary items in the instances, then pass those counts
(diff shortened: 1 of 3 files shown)


HarshTrivedi:
True, that seems cleaner! Does the `config.json` usage with `extend` toggle as shown in example above seem OK to you?

Correct me if I have misinterpreted you:
I will make this instance method in vocabulary class: 
```
def extend(self, extend_params: Params, extend_instances: Iterable['adi.Instance'] = ()) -> None ...
```
and remove `extend_vocab` argument from `Vocabulary.from_instances`.

And will update `from_params` in following way:
```
if params has extend = False then:  (as original)
    Vocabulary.from_instances(instances=instances, ..... )
else:
    vocab = load_vocab from files
    extended_vocab = vocab.extend( ..., instances)
```
and fine-tune will also have some changes.
```
extended_vocab = model_vocab.extend( data_instances, ....)
```

joelgrus:
do you even need to change `from_params` if you do it that way? can't fine-tune just do something like

```
vocab = model.vocab
vocab.extend(...)
```

?

HarshTrivedi:
Yes sure, but my initial intension was to support vocabulary extension even with `train` command and hence support this kind of config usage as optional facility:
```
"vocabulary": {
    "directory_path": "vocabulary_dir_path"
    "extend": true
}
```
Fine-tune supports a very limited kind of transfer. You continute tuning all the parameters on a new dataset. However, one can do much more flexibile kind of module transfers using `train` command itself as I am currently doing in my project.  One can load models by `load_archive()` in from_params of model and then do all kinds of selective module transfers. One can even select which transferred modules to continue tuning which to transfer but keep frozen using #1405.

joelgrus:
oh got it, yeah that seems reasonable. the part to watch out for is not consuming the same params object twice, you might have to make a copy somewhere

HarshTrivedi:
Cool, any other suggestion? Or I continue as I briefed 3 comments above?

joelgrus:
I think the way you've described it sounds good, would `extend_vocab` be an in-place operation or would it return a new vocabulary. I was thinking the first, but your examples suggest the second:

```
extended_vocab = vocab.extend( ..., instances)
```

HarshTrivedi:
Thanks for pointing out. Yes, in place should be better. I will make extend method inplace only.:
```
vocab.extend( ..., instances)
```

HarshTrivedi:
Updated. I have added two functions `extend` and `extend_from_instances`:

```
def extend(self, counter: Dict[str, Dict[str, int]] = None, min_count: Dict[str, int] = None, max_vocab_size: Union[int, Dict[str, int]] = None, non_padded_namespaces: Sequence[str] = DEFAULT_NON_PADDED_NAMESPACES, pretrained_files: Optional[Dict[str, str]] = None, only_include_pretrained_words: bool = False,tokens_to_add):

def extend_from_instances(self, params: Params, instances: Iterable['adi.Instance'] = ()) :
```

The method `extend` has exact same arguments as Vocabulary initializer. The initializer creates a bare-minimum vocabulary and calls `extend` with same arguments. The method `extend_from_instances` is basically wrapper around `extend` to extend from instances. 

The `extend` method basically does the common work shared between vocabulary initialization and vocabulary extension with instances.

And fine-tune works as:
```python
vocab = model.vocab
vocab.extend_from_instances(vocabulary_params, dataset_instances))
```

allennlp/data/vocabulary.py
```diff
@@ -406,6 +407,112 @@ def from_params(cls, params: Params, instances: Iterable['adi.Instance'] = None)
...
+
+        if counter is not None:
+            # Make sure vocabulary extension is safe.
+            # Common named namespaces must have same setting for padded=True/False
+            tob_added_namespaces = list(counter.keys())
+            existing_namespaces = self.get_all_namespaces()
+            for namespace in tob_added_namespaces:
+                if namespace in existing_namespaces:
+                    existing_namespace_padded = namespace in self._non_padded_namespaces
+                    tob_added_namespace_padded = namespace in non_padded_namespaces
```
joelgrus: I don't like these variable names, I would call them `original_padded` and `extension_padded` or something like that

allennlp/data/vocabulary.py
```diff
@@ -406,6 +407,112 @@ def from_params(cls, params: Params, instances: Iterable['adi.Instance'] = None)
...
+                                self.add_token_to_namespace(token, namespace)
+                        elif token in pretrained_list or count >= min_count.get(namespace, 1):
+                            self.add_token_to_namespace(token, namespace)
+                    elif count >= min_count.get(namespace, 1):
+                        self.add_token_to_namespace(token, namespace)
+
+        if tokens_to_add:
+            for namespace, tokens in tokens_to_add.items():
+                for token in tokens:
+                    self.add_token_to_namespace(token, namespace)
```
joelgrus: isn't most of this code (lines 446 - 469) duplicated verbatim from `Vocabulary.__init__`? can we factor it out?

allennlp/data/vocabulary.py
```diff
@@ -76,6 +76,10 @@ def __missing__(self, key: str):
         dict.__setitem__(self, key, value)
         return value
 
+    def add_non_padded_namespaces(self, non_padded_namespaces: Sequence[str]):
+        # add non_padded_namespaces which weren't already present
+        self._non_padded_namespaces.extend(non_padded_namespaces)
+        self._non_padded_namespaces = list(set(self._non_padded_namespaces))
```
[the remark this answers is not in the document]
HarshTrivedi: `self._non_padded_namespaces` was a tuple before, because `DEFAULT_NON_PADDED_NAMESPACES ` are set as tuple. However, when non_padded_namespaces are passed they are lists: `Sequence[str]`.

But its better to make it a set as default and throughout and then `set(list(..))` won't be required. I will change `Sequence[str]` to set and see if everything works. And as you said, following would be updated function:
```
self._non_padded_namespaces.update(non_padded_namespaces)
```
Sounds good?

allennlp/data/vocabulary.py
HarshTrivedi: Sure

allennlp/data/vocabulary.py
HarshTrivedi: Yes I have done exactly that. `extend` method and `__init__` have exact same arguments. Even functionally they are overlapping: `__init__` makes and empty vocabulary and then extends it with `extend`. Then we also have `exend_from_instances` which wraps `extend` method for it's purpose. Hence `extend` function is basically the factored out code between `Vocab.__init__` and `extend_from_instances`.  `__init__` function is just 4 lines of code now.

allennlp/data/vocabulary.py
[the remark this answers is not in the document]
joelgrus: 👍

allennlp/data/vocabulary.py
joelgrus: oops, missed that, thanks!

allennlp/data/vocabulary.py
```diff
@@ -406,6 +407,112 @@ def from_params(cls, params: Params, instances: Iterable['adi.Instance'] = None)
...
+        if not isinstance(max_vocab_size, dict):
+            int_max_vocab_size = max_vocab_size
+            max_vocab_size = defaultdict(lambda: int_max_vocab_size)  # type: ignore
+        min_count = min_count or {}
+        pretrained_files = pretrained_files or {}
+
+        if counter is not None:
+            # Make sure vocabulary extension is safe.
+            # Common named namespaces must have same setting for padded=True/False
+            tob_added_namespaces = list(counter.keys())
```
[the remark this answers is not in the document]
HarshTrivedi: sure. Will update it.

allennlp/data/vocabulary.py
HarshTrivedi: This block is cleaned up now.

HarshTrivedi:
For some reason one of my builds for previous commit hanged on TeamCity (agent-3) 3-4 hrs back. Someone might want to terminate it.

nelson-liu:
> For some reason one of my builds for previous commit hanged on TeamCity (agent-3) 3-4 hrs back. Someone might want to terminate it.

got it for you, thanks!

HarshTrivedi:
1. Sure, `Iterable[str]` suits better since we are any way calling `set(...)`. It fixes next comment also.
2. That was because default values should not be mutable. Lists and Sets are mutable and shouldn't be kept at default values in arguments. Earlier `DEFAULT_NON_PADDED_NAMESPACES` was a tuple which is not mutable and hence didn't cause any issue with pylint. I couldn't change it back to tuple because of `Set[str]` annotation. If I cahnge type annotation to `Iterable[str]` I can make DEFAULT_NON_PADDED_NAMESPACES back to tuple and hence remove that extra if/else. So it would clean up as you suggest.

Btw, are you suggesting that I should change annotation to `Iterable[str]` from `Set[str]` everywhere? Some calls are for strictly internal usage and don't go to end-user: Eg. `_TokenToIndexDefaultDict`, `_IndexToTokenDefaultDict`, and both `add_non_padded_namespaces`. Should it be better to keep it as Set there? Extra `set(..)` calls everywhere doesn't seem so good to me:

So
```
#OptionA
    def add_non_padded_namespaces(self, non_padded_namespaces: Iterable[str]):
        self._non_padded_namespaces.update(set(non_padded_namespaces))
        #extra set() will be required in such functions
```
vs
```
#OptionB (only for internal functions)
    def add_non_padded_namespaces(self, non_padded_namespaces: Set[str]):
        self._non_padded_namespaces.update(non_padded_namespaces)
```

for `_TokenToIndexDefaultDict`, `_IndexToTokenDefaultDict`, and both `add_non_padded_namespaces` ?

I have made a commit with `OptionB`, you can check the changes. If `OptionA` seems better, will fix it in next commit.

allennlp/data/vocabulary.py
```diff
@@ -76,16 +76,19 @@ def __missing__(self, key: str):
         dict.__setitem__(self, key, value)
         return value
 
+    def add_non_padded_namespaces(self, non_padded_namespaces: Set[str]):
+        # add non_padded_namespaces which weren't already present
+        self._non_padded_namespaces.update(non_padded_namespaces)
 
 class _TokenToIndexDefaultDict(_NamespaceDependentDefaultDict):
-    def __init__(self, non_padded_namespaces: Sequence[str], padding_token: str, oov_token: str) -> None:
+    def __init__(self, non_padded_namespaces: Set[str], padding_token: str, oov_token: str) -> None:
```
HarshTrivedi: Will be called internally only: `Set[str]` vs `Iterable[str]` and `set(..)` call?

allennlp/data/vocabulary.py
```diff
@@ -38,17 +38,17 @@ class _NamespaceDependentDefaultDict(defaultdict):
...
     ``passage_tags``, ``question_tags``, etc. (anything that ends with ``tags``) will have the
     ``non_padded`` default value.
 
     Parameters
     ----------
-    non_padded_namespaces : ``Sequence[str]``
-        A list or tuple of strings describing which namespaces are not padded.  If a namespace
-        (key) is missing from this dictionary, we will use :func:`namespace_match` to see whether
+    non_padded_namespaces : ``Iterable[str]``
+        A set of strings describing which namespaces are not padded.  If a namespace (key)
```
HarshTrivedi: I am yet to change docstrings: "set of strings" to "set / list  / tuple of strings"

allennlp/data/vocabulary.py
```diff
@@ -76,16 +76,19 @@ def __missing__(self, key: str):
...
-    def __init__(self, non_padded_namespaces: Sequence[str], padding_token: str, oov_token: str) -> None:
+    def __init__(self, non_padded_namespaces: Set[str], padding_token: str, oov_token: str) -> None:
         super(_TokenToIndexDefaultDict, self).__init__(non_padded_namespaces,
                                                        lambda: {padding_token: 0, oov_token: 1},
                                                        lambda: {})
 
 
 class _IndexToTokenDefaultDict(_NamespaceDependentDefaultDict):
-    def __init__(self, non_padded_namespaces: Sequence[str], padding_token: str, oov_token: str) -> None:
+    def __init__(self, non_padded_namespaces: Set[str], padding_token: str, oov_token: str) -> None:
```
HarshTrivedi: Will be called internally only: `Set[str]` vs `Iterable[str]` and `set(..)` call?

allennlp/data/vocabulary.py
```diff
@@ -406,6 +406,113 @@ def from_params(cls, params: Params, instances: Iterable['adi.Instance'] = None)
...
+        for instance in Tqdm.tqdm(instances):
+            instance.count_vocab_items(namespace_token_counts)
+        self.extend(counter=namespace_token_counts,
+                    min_count=min_count,
+                    max_vocab_size=max_vocab_size,
+                    non_padded_namespaces=non_padded_namespaces,
+                    pretrained_files=pretrained_files,
+                    only_include_pretrained_words=only_include_pretrained_words,
+                    tokens_to_add=tokens_to_add)
+
```
HarshTrivedi: Will be called internally only: `Set[str]` vs `Iterable[str]` and `set(..)` call?

allennlp/data/vocabulary.py
```diff
@@ -406,6 +406,113 @@ def from_params(cls, params: Params, instances: Iterable['adi.Instance'] = None)
...
+    def add_non_padded_namespaces(self, non_padded_namespaces: Set[str]):
+        self._token_to_index.add_non_padded_namespaces(non_padded_namespaces)
+        self._index_to_token.add_non_padded_namespaces(non_padded_namespaces)
+        self._non_padded_namespaces.update(non_padded_namespaces)
```
HarshTrivedi: Will be called internally only: `Set[str]` vs `Iterable[str]` and `set(..)` call?

allennlp/data/vocabulary.py
```diff
@@ -406,6 +406,113 @@ def from_params(cls, params: Params, instances: Iterable['adi.Instance'] = None)
...
+
```
HarshTrivedi: I think following two methods should return Set only since namespaces are represented as Set inside. Agree?

HarshTrivedi:
@nelson-liu Can you please terminate my build from agent-3 again? And if possible restart it on other agent? Locally ` python scripts/verify.py` works perfectly within 8-10 minutes. Any thoughts on what is causing this hangup?

joelgrus:
looks good, thanks for making those changes

nelson-liu:
hi @HarshTrivedi  --- I'm running into an issue where a model that i've trained on allennlp 0.5 cannot be fine-tuned on allennlp 0.6. I think the culprit is this PR. In particular, the model (the BiDAF model) complains about an index out of bounds error (full traceback below).

How does vocabulary extension work with fine-tuning? It seems necessary that the embeddings be expanded as well, but my guess it that it is currently not being expanded (I'm using pretrained embeddings, if it matters). Any clues?

```
2018-08-17 09:23:41,778 - INFO - allennlp.training.trainer - Training
  0%|          | 0/1 [00:00<?, ?it/s]
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/allennlp/run.py", line 18, in <module>
    main(prog="allennlp")
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/allennlp/commands/__init__.py", line 70, in main
    args.func(args)
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/allennlp/commands/fine_tune.py", line 78, in fine_tune_model_from_args
    file_friendly_logging=args.file_friendly_logging)
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/allennlp/commands/fine_tune.py", line 113, in fine_tune_model_from_file_paths
    file_friendly_logging=file_friendly_logging)
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/allennlp/commands/fine_tune.py", line 213, in fine_tune_model
    metrics = trainer.train()
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/allennlp/training/trainer.py", line 720, in train
    train_metrics = self._train_epoch(epoch)
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/allennlp/training/trainer.py", line 486, in _train_epoch
    loss = self._batch_loss(batch, for_training=True)
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/allennlp/training/trainer.py", line 420, in _batch_loss
    output_dict = self._model(**batch)
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/torch/nn/modules/module.py", line 477, in __call__
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/allennlp/models/reading_comprehension/bidaf.py", line 174, in forward
    embedded_passage = self._highway_layer(self._text_field_embedder(passage))
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/torch/nn/modules/module.py", line 477, in __call__
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/allennlp/modules/text_field_embedders/basic_text_field_embedder.py", line 88, in forward
    token_vectors = embedder(*tensors)
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/torch/nn/modules/module.py", line 477, in __call__
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/allennlp/modules/token_embedders/embedding.py", line 129, in forward
    sparse=self.sparse)
  File "/home/nfliu/miniconda3/envs/adversarial/lib/python3.6/site-packages/torch/nn/functional.py", line 1110, in embedding
    return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)
RuntimeError: index out of range at /pytorch/aten/src/TH/generic/THTensorMath.cpp:352
```

matt-gardner:
@nelson-liu, this is the issue you're looking for: https://github.com/allenai/allennlp/issues/1596.

nelson-liu:
great, thanks for the pointer! fwiw: I think @joelgrus is in the process of making a PR to have vocabulary extension be optional by default in fine-tuning.

Issue #1417, opened 2018-06-25
Relative data path not working in ModelTestCase.set_up_model


handsomezebra:
**Describe the bug**
Getting the FileNotFoundError when was doing pytest. Then I tried allennlp-as-a-library-example and got the same error.

`FileNotFoundError: file tests/fixtures/academic_paper_classifier.json not found`

I found this is caused by the os.chdir() within the set_up_model, so it cannot find the test file relative to current directory anymore.

**To Reproduce**
Steps to reproduce the behavior
1. Checkout allennlp-as-a-library-example and cd to this directory
2. Running pytest
3. See error

**Expected behavior**
pytest pass.

**System (please complete the following information):**
 - OS: Linux
 - Python version: 3.6
 - AllenNLP version: master
 - PyTorch version: 0.4.0

HarshTrivedi:
Hi, thanks for posting the issue. I can reproduce your error.

Make empty file: `allennlp-as-a-library-example/tests/__init__.py`
and copy the following content:
```python
import os
from allennlp.common.testing import AllenNlpTestCase
AllenNlpTestCase.MODULE_ROOT = (os.path.dirname(os.path.abspath(__file__)) + '/../')
```
and it should work.

nelson-liu:
thanks for letting me know, i'm looking into it. the purpose was to implement a `allennlp test-install` command for users to easily verify that their installation was working properly. this is a bit tricky since allennlp can be installed in a bunch of different places depending on your package manager, etc, so i added the chdir to make the paths resolve properly.

nelson-liu:
I made a proposed fix at #1418 , thanks again for reporting.

HarshTrivedi:
@nelson-liu  Making change in `allennlp` library itself is better permanent solution. However, people using `allennlp` as library installed from pip won't be able to run tests in `allennlp-as-a-library-example`until next release. Unless they choose to use github master after #1418 is merged. Hence, it should be better to have a temporary fix inplace in `allennlp-as-a-library-example` with a comment note that this test path setting requirement is temporary and will be fixed by next release of `allennlp`.

HarshTrivedi:
Ahh! I didn't see your upgrade to allennlp051 PR on `allennlp-as-a-library-example` as well. Cool! That would fix that repo temporarily as well. Thanks

PR #1418, opened 2018-06-25
Fix chdir in ModelTestCase breaking downstream models

nelson-liu:
ModelTestCase has a few `chdir`s since after we moved the tests into the module, all the relative paths in the test fixtures were broken (since they assumed that we were running the tests in the same directory as the `tests` folder). I got around this by using `os.chdir`, but did not realize that this would break models that import and use `ModelTestCase` (see #1417 ).

I think a reasonable solution to this is to simply remove all the `chdir` calls and add a symbolic link in the project root called `tests` that links to the tests folder (in the module root). This means that the relative paths in the fixtures don't need to be redone and downstream clients using `ModelTestCase` won't deal with issues from `os.chdir`.

### DIFF
--- a/allennlp/common/testing/model_test_case.py
+++ b/allennlp/common/testing/model_test_case.py
@@ -19,10 +19,6 @@ class ModelTestCase(AllenNlpTestCase):
     """
     def set_up_model(self, param_file, dataset_file):
         # pylint: disable=attribute-defined-outside-init
-        initial_working_dir = os.getcwd()
-        # Change directory to module root.
-        os.chdir(self.MODULE_ROOT)
-
         self.param_file = param_file
         params = Params.from_file(self.param_file)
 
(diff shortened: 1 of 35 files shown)


nelson-liu:
for the future: perhaps it'd be good to rerun and update https://github.com/allenai/allennlp-as-a-library-example as part of the process of testing a new release, to ensure that nothing breaks terribly for folks using allennlp as a library.

joelgrus:
I'm not a huge fan of this solution.

can't `ModelTestCase` just add `self.PROJECT_ROOT` to `sys.path`? would that fix things? or am I misunderstanding the issue

nelson-liu:
No, it wouldn’t. The fixtures themselves define relative paths to data, which get broken. It’s unrelated to PYTHONPATH

joelgrus:
as discussed offline, I think just correcting the paths in the test fixtures is a better solution

PR #1422, opened 2018-06-25
Avoid non-model state in predictors

DeNeutoy:
This is a first step to being able to use a model's original dataset reader in the predictor. Basically, this removes all the additional data we allowed predictors to pass to themselves through the following API:

`def _json_to_instances(inputs: JsonDict) -> Tuple[Instance, JsonDict]`

to only return `Instances`:

`def _json_to_instances(inputs: JsonDict) -> Instance`

I originally introduced this but it was a very bad idea (In future: Listen to Joel) and took me all afternoon to fix. The problem is that we want the output of the predictor to be agnostic of where the `Instance` it predicts on came from (so that in predict, we can use the original dataset reader from the model to generate `Instances`). Instead, you now have to route any metadata through `MetadataFields` in your model. 

I've made all the metadata fields I added optional because it seemed like the right thing to do at the time, but now i'm not sure. Also, this is a breaking change but there is no option to not break the API if we want to eventually use dataset readers as the default in the `predict` command, which I actually think is quite important.

### DIFF
--- a/allennlp/commands/evaluate.py
+++ b/allennlp/commands/evaluate.py
@@ -86,6 +86,7 @@ def evaluate(model: Model,
              instances: Iterable[Instance],
              data_iterator: DataIterator,
              cuda_device: int) -> Dict[str, Any]:
+    _warned_tqdm_ignores_underscores = False
     check_for_gpu(cuda_device)
     with torch.no_grad():
         model.eval()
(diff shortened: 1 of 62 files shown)


joelgrus:
Mostly this looks fine, I have two broad comments:

(1) is it OK to add extra metadata to every instance, even if we really only want it in the predictor case? I am thinking about memory implications mostly.

(2) explicit is better than implicit, so rather than things like

```
fields["metadata"] = MetadataField({"words": [x.text for x in tokens], "verb": verb})
```

I think I would prefer 

```
fields['words'] = MetadataField([x.text for x in tokens])
fields['verb'] = verb
```

and so on, but then that makes the `forward()` signatures longer, etc. thoughts?

DeNeutoy:
Both valid points - 
1) For the datasets I added them to I don't think so, seems like something we can deal with when it becomes a problem? E.g if you use a lazy dataset iterator it is certainly not a problem.
2) I was optimising for making the minimal change possible - we already have several models which use only a single metadata field, so I think that part is ok - particularly because as you say it would clutter things considerably for example SRL.

PR #1424, opened 2018-06-26
Add LabelCategoricalAccuracy

nelson-liu:
This metric tracks the accuracy for a particular label.

Let me know if there's an easier way to do this (I couldn't find it) / if this is better as just an optional argument to `CategoricalAccuracy`


allennlp/training/metrics/categorical_accuracy.py
```diff
@@ -54,6 +65,18 @@ def __call__(self,
...
+            if mask is not None:
+                # Update the mask to account for positive_label
+                label_mask = gold_labels.eq(self._positive_label)
+                # elementwise-AND between label_mask and mask to get a mask
+                # over elements that are both 1
+                mask = label_mask.long() & mask.long()
+            else:
+                # positive_label itself defines the mask
+                mask = gold_labels.eq(self._positive_label)
+
```
joelgrus: nit: it's probably better to only have the `label_mask` logic once, e.g.

```
if self._positive_label is not None:
    label_mask = gold_labels.eq(self._positive_label)
    if mask is None:
       mask = label_mask
    else:
        mask = mask.long() & label_mask.long()
```

joelgrus:
now that I'm more awake, what I meant to say is that what you want is just per-label _recall_, which is already computed by the (per-label) `F1Measure` metric, so why not just use that?

(apologies for not realizing this earlier)

nelson-liu:
You're right, that's the preferred solution i think. thanks!

PR #1430, opened 2018-06-26
Make evaluation tqdm description ignore metrics starting with _ 

nelson-liu:
Continuation of #1425 , missed one

### DIFF
--- a/allennlp/commands/evaluate.py
+++ b/allennlp/commands/evaluate.py
@@ -98,7 +98,7 @@ def evaluate(model: Model,
             model(**batch)
             metrics = model.get_metrics()
             if (not _warned_tqdm_ignores_underscores and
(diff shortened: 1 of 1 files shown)


allennlp/commands/evaluate.py
```diff
@@ -96,7 +97,13 @@ def evaluate(model: Model,
         for batch in generator_tqdm:
             model(**batch)
             metrics = model.get_metrics()
-            description = ', '.join(["%s: %.2f" % (name, value) for name, value in metrics.items()]) + " ||"
+            if (not _warned_tqdm_ignores_underscores and
```
joelgrus: instead of creating this variable, why not just check for underscores and warn before the `for batch in` loop?

PR #1433, opened 2018-06-27
remove adaptive iterator

DeNeutoy:
This isn't actually usable `from_params` at the moment and no-one's really using it - I think it's best just to cut it.


joelgrus:
ooh, I'm going to pick back up my "refactor iterators" PR, the AdaptiveIterator was one of the things that was making it difficult

PR #1434, opened 2018-06-27
Enables Predict to use dataset readers from models

### LINKED ISSUE #1096: option for `predict` command to use original DatasetReader + file format
this has been requested several times, it shouldn't be terribly hard, I think it just requires some changes to `commands/predict.py`

DeNeutoy:
Fixes #1096.


nelson-liu:
Does this let you predict on the original file format? If so, maybe link https://github.com/allenai/allennlp/issues/1096 in the description?

I've been looking forward to this feature for awhile, thanks for adding it!

allennlp/commands/predict.py
```diff
@@ -25,29 +24,34 @@
...
                             path to output file
+    --weights-file WEIGHTS_FILE
+                            a path that overrides which weights file to use
     --batch-size BATCH_SIZE
                             The batch size to use for processing
     --silent              do not print output to stdout
     --cuda-device CUDA_DEVICE
                             id of GPU to use (if any)
+    --use-dataset-reader  Whether to use the dataset reader of the original
+                            model to load Instances
```
joelgrus: nit: should all these indentations be aligned? (in particular, the two lines for use-dataset-reader with each other)

allennlp/commands/predict.py
```diff
@@ -94,62 +102,95 @@ def _get_predictor(args: argparse.Namespace) -> Predictor:
...
     return Predictor.from_archive(archive, args.predictor)
 
-def _run(predictor: Predictor,
-         input_file: IO,
-         output_file: Optional[IO],
-         batch_size: int,
-         print_to_console: bool) -> None:
 
-    def _run_predictor(batch_data):
+class _Predict:
```
joelgrus: I don't find this name terribly descriptive, `_PredictorWrapper` maybe? Something that gives more of a sense of what this class is for

allennlp/commands/predict.py
```diff
@@ -94,62 +102,95 @@ def _get_predictor(args: argparse.Namespace) -> Predictor:
...
-            # Batch results return a list of json objects, so in
-            # order to iterate over the result below we wrap this in a list.
-            results = [result]
+            result = self._predictor.predict_json(batch_data[0])
+            yield self._predictor.dump_line(result)
         else:
-            results = predictor.predict_batch_json(batch_data)
+            results = self._predictor.predict_batch_json(batch_data)
+        for output in results:
+            yield self._predictor.dump_line(output)
```
joelgrus: this might be marginally cleaner as

```
if len(batch_data) == 1:
    results = [self._predictor.predict_json(batch_data[0])]
else:
    results = self._predictor.predict_batch_json(batch_data)
for output in results:
    ...
```

what do you think?

allennlp/commands/predict.py
```diff
@@ -94,62 +102,95 @@ def _get_predictor(args: argparse.Namespace) -> Predictor:
...
-            string_output = predictor.dump_line(output)
-            if print_to_console:
+    def _predict_instances(self, batch_data: List[Instance]) -> Iterator[str]:
+        if len(batch_data) == 1:
+            result = self._predictor.predict_instance(batch_data[0])
+            yield self._predictor.dump_line(result)
+        else:
+            results = self._predictor.predict_batch_instance(batch_data)
+            for output in results:
+                yield self._predictor.dump_line(output)
```
joelgrus: same as previous comment

allennlp/commands/predict.py
```diff
@@ -94,62 +102,95 @@ def _get_predictor(args: argparse.Namespace) -> Predictor:
...
-             output_file,
-             args.batch_size,
-             not args.silent)
+    util = _Predict(predictor,
+                    args.input_file,
+                    args.output_file,
+                    args.batch_size,
+                    not args.silent,
+                    args.use_dataset_reader)
+    util.run()
```
joelgrus: related to above, `util` isn't the most helpful name, maybe if it were `predictor_wrapper` or `predictor_runner` or something

allennlp/models/model.py
```diff
@@ -153,8 +156,19 @@ def forward_on_instances(self,
 
             instance_separated_output: List[Dict[str, numpy.ndarray]] = [{} for _ in dataset.instances]
             for name, output in list(outputs.items()):
+                # NOTE(markn): This is a hack because 0-dim pytorch tensors are not iterable.
+                # This occurs with batch size 1, because we still want to include the loss in that case.
+                if name == "loss" and output.dim() == 0:
```
joelgrus: do you want to only do this for "loss", or for any 0-tensor?

allennlp/commands/predict.py
DeNeutoy: I just copy and pasted the actual terminal output from running the command, but ok

allennlp/models/model.py
DeNeutoy: good point thanks

allennlp/commands/predict.py
joelgrus: weird, it's all aligned on my machine

```
(allennlp) OS-XImage:allennlp joelg$ allennlp predict --hel
usage: allennlp predict [-h] [--output-file OUTPUT_FILE]
                        [--weights-file WEIGHTS_FILE]
                        [--batch-size BATCH_SIZE] [--silent]
                        [--cuda-device CUDA_DEVICE] [-o OVERRIDES]
                        [--predictor PREDICTOR]
                        [--include-package INCLUDE_PACKAGE]
                        archive_file input_file

Run the specified model against a JSON-lines input file.

positional arguments:
  archive_file          the archived model to make predictions with
  input_file            path to input file

optional arguments:
  --output-file OUTPUT_FILE
                        path to output file
  --weights-file WEIGHTS_FILE
                        a path that overrides which weights file to use
  --batch-size BATCH_SIZE
                        The batch size to use for processing
  --silent              do not print output to stdout
  --cuda-device CUDA_DEVICE
                        id of GPU to use (if any)
  -o OVERRIDES, --overrides OVERRIDES
                        a JSON structure used to override the experiment
                        configuration
  --predictor PREDICTOR
                        optionally specify a specific predictor to use
  --include-package INCLUDE_PACKAGE
                        additional packages to include
```

allennlp/models/model.py
```diff
@@ -211,6 +225,18 @@ def _get_prediction_device(self) -> int:
...
 
+    def _maybe_warn_for_unseparable_batches(self):
+        """
+        This method warns once if a user implements a model which returns a dictionary with
+        values which we are unable to split back up into elements of the batch. This is controlled
+        by a class attribute ``_warn_for_unseperable_batches`` because it would be extremely verbose
+        otherwise.
+        """
+        if self._warn_for_unseparable_batches:
+            logger.warning("Encountered a key in the model's return dictionary which "
```
matt-gardner: Maybe have this take the key as an argument, and log that?  Seems like you will _always_ see this for `loss`, right?  You might even just never warn for `loss`.

matt-gardner: Ah, sorry, `loss` won't trigger this branch.  Still, knowing what key triggered this would be nice.

DeNeutoy: That's actually a good idea - we should warn per key. We want to warn for loss, because it's definitely something that you might want to return, which you could get around (if you were warned about it) by adding a `batch_loss` key to your model output.

matt-gardner:
I'm slightly concerned this might break the wikitables preprocessing script (particularly, the part where we save the table information).  I didn't look closely enough to tell for sure, and we don't have a test to check for this.  If it's easy, can you run that script and see if it still works?  If that's not easy, I'll just fix things when I next run the script, if it's indeed broken.

DeNeutoy:
I did check the script - it expects the table lines, which it is still getting - but I didn't run it. I don't know where to get the data from to test that.

matt-gardner:
Ok, great, thanks for keeping it in mind.  It's probably fine.

PR #1436, opened 2018-06-28
eliminate or make private most of the new Vocabulary methods

joelgrus:
the code to add `Vocabulary.extend_from_instances` introduced a lot of functions that no one other than a `Vocabulary` instance (or its tests) needs to call. this PR eliminates them or makes them private.

FYI @HarshTrivedi

### DIFF
--- a/allennlp/tests/data/vocabulary_test.py
+++ b/allennlp/tests/data/vocabulary_test.py
@@ -1,4 +1,3 @@
-# pylint: disable=no-self-use,invalid-name,too-many-public-methods
 import codecs
 import gzip
 import zipfile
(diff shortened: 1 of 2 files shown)


HarshTrivedi:
great, thanks!

allennlp/tests/data/vocabulary_test.py
```diff
@@ -20,6 +19,8 @@
 
 
 class TestVocabulary(AllenNlpTestCase):
+    # pylint: disable=no-self-use, invalid-name, too-many-public-methods, protected-access
```
DeNeutoy: undo?

HarshTrivedi: too-many-public-methods was added for number of public methods in test file. I mean there are >20 test methods so. It has nothing to do with public methods in `vocabulary.py`

joelgrus: I just moved it from the top of the file to inside the class.

There are good principled reasons for making the disables more tightly scoped, but I did it because (for some reason) my local pylint keeps ignoring disables in the first line of a file

PR #1438, opened 2018-06-28
Few feature additions

matt-peters:
Included in this PR:
* make `Vocabulary` registrable
* update `ModelTestCase` to work with sparse gradients
* add the ability to add a `Field` to an `Instance` after it is constructed
* provide an optional generic upper bound on the number of tokens in a batch in the iterators (`maximum_samples_per_batch`), with dynamic splitting of batches that are too large
* work around an edge case in tensorboard logging with empty gradients


matt-peters:
Yes, I have a Vocabulary subclass.  I need the ability to combine an existing vocabulary with the allennlp Vocabulary machinery to create a vocab from a dataset.  I really don't want to change the existing vocab, and just want it to exist as is in a separate namespace.  So I have a small, few line subclass, that accomplishes this.

allennlp/training/trainer.py
```diff
@@ -563,12 +563,15 @@ def _parameter_and_gradient_statistics_to_tensorboard(self, # pylint: disable=in
...
                 else:
                     grad_data = param.grad.data
-                self._tensorboard.add_train_scalar("gradient_mean/" + name,
-                                                   grad_data.mean(),
-                                                   epoch)
-                self._tensorboard.add_train_scalar("gradient_std/" + name,
-                                                   grad_data.std(),
-                                                   epoch)
+
+                # skip empty gradients
```
DeNeutoy: I'm not sure this is a good idea? I definitely want to know if some of my parameters have zero gradients.

allennlp/data/iterators/basic_iterator.py
```diff
@@ -32,14 +32,20 @@ class BasicIterator(DataIterator):
         If specified, the iterator will load this many instances at a time into an
         in-memory list and then produce batches from one such list at a time. This
         could be useful if your instances are read lazily from disk.
+    maximum_samples_per_batch : tuple(padding_key, limit), optional, (default = None)
```
DeNeutoy: this type hint should be : `Tuple[str, int]` and then describe them in text after.

allennlp/training/trainer.py
matt-peters: This is just for logging to tensorboard, and is different from zero gradients.  It's the case where there a parameter with requires_grad=True, but the gradient isn't computed when the parameter isn't used in computing the loss.  This is the case where `param.grad=None` (that we exclude above for dense gradients) but in the sparse gradient case (`Embedding`modules) torch instead has a `param.grad.data` with shape of all zeros.

allennlp/data/iterators/basic_iterator.py
matt-peters: This is fixed in the latest commits.

allennlp/training/trainer.py
DeNeutoy: Ah got it - can you add that as a comment because it's not immediately obvious that this is what the condition would apply to.

allennlp/data/vocabulary.py
```diff
@@ -112,7 +113,17 @@ def _read_pretrained_tokens(embeddings_file_uri: str) -> Set[str]:
     return tokens
 
 
-class Vocabulary:
+class RegistrableVocabulary(Registrable):
+    default_implementation = "vocabulary"
+
+    @classmethod
+    def from_params(cls, params: Params, instances: Iterable['adi.Instance'] = None):
```
DeNeutoy: To make this backward compatible (and fix all the failing tests), you'll need to check if `params` has a type key and if not, use the default implementation of the vocabulary.

Issue #1439, opened 2018-06-28
[Is this a bug?]: BiDAF appears to access wrong metadata instances when running on multiple GPUs


MaxMotovilov:
This loop https://github.com/allenai/allennlp/blob/43fc89ee59944dca0de01ca5ab2843c9926323d6/allennlp/models/reading_comprehension/bidaf.py#L274 iterates over the range obtained from a tensor dimension corresponding to batch size https://github.com/allenai/allennlp/blob/43fc89ee59944dca0de01ca5ab2843c9926323d6/allennlp/models/reading_comprehension/bidaf.py#L176 accessing elements of the `metadata` list with indices in this range. This is reasonable in a CPU or single GPU scenario where `forward()` operates on the entire batch generated by the trainer algorithm. With multiple GPUs, all tensors passed into `forward()` are sliced and scattered across GPU instances with `forward()` transparently receiving only a subset of data (e.g. with configured batch size of 20 on 8 GPUs, we see actual `batch_size` in `forward()` to be either 2 or 3). Metadata is copied into each call to `forward()` in its entirety since it does not contain tensors. Therefore the majority of the metadata access happens with a wrong index: the tensors operated on are not conformal to the array of metadata instances.

I wonder if the results of BiDAF training on multiple GPUs were previously found deficient or inferior to training them on a single GPU with otherwise identical configuration. This apparent disconnect in the code may explain such an outcome. I have only run across this issue by comparing BiDAF code with the multi-paragraph version in https://github.com/allenai/allennlp/pull/804 which actually generates a runtime exception because of it. Of course, being an absolute newbie in this codebase I may be completely wrong!

DeNeutoy:
Hello - Multi-GPU training is very experimental and essentially not supported. Presumably you saw this warning:
https://github.com/allenai/allennlp/blob/master/allennlp/training/trainer.py#L291

In particular, MultiGPU training will only work correctly if all of the inputs and outputs to `forward` are tensors. There will be many, many other caveats to this statement - for instance, I think many `Metrics` will work incorrectly.

MaxMotovilov:
That said, there's a significant advantage to multi-GPU training; I intend to try and work through what issues I can deal with but would appreciate feedback from the code owners as to whether I'm even barking up the right tree so to say

brendan-ai2:
This should be fixed in master. Please let me know if that's not what you observe! Thanks!

PR #1445, opened 2018-06-29
Improved support for multi-GPU training: config handling and chunking the metadata 

MaxMotovilov:
Refers to: https://github.com/allenai/allennlp/issues/1439

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -948,13 +948,11 @@ class Trainer:
         patience = params.pop_int("patience", None)
         validation_metric = params.pop("validation_metric", "-loss")
         num_epochs = params.pop_int("num_epochs", 20)
-        cuda_device = params.pop_int("cuda_device", -1)
+        cuda_device = params.pop( "cuda_device")
         grad_norm = params.pop_float("grad_norm", None)
         grad_clipping = params.pop_float("grad_clipping", None)
         lr_scheduler_params = params.pop("learning_rate_scheduler", None)
 
-        if cuda_device >= 0:
-            model = model.cuda(cuda_device)
         parameters = [[n, p] for n, p in model.named_parameters() if p.requires_grad]
         optimizer = Optimizer.from_params(parameters, params.pop("optimizer"))
 
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -961,7 +961,7 @@ class Trainer:
         patience = params.pop_int("patience", None)
         validation_metric = params.pop("validation_metric", "-loss")
         num_epochs = params.pop_int("num_epochs", 20)
-        cuda_device = params.pop( "cuda_device")
+        cuda_device = params.pop("cuda_device", -1)
         grad_norm = params.pop_float("grad_norm", None)
         grad_clipping = params.pop_float("grad_clipping", None)
         lr_scheduler_params = params.pop("learning_rate_scheduler", None)
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -25,7 +25,7 @@ from tensorboardX import SummaryWriter
 
 from allennlp.common import Params
 from allennlp.common.checks import ConfigurationError
-from allennlp.common.util import peak_memory_mb, gpu_memory_mb
+from allennlp.common.util import peak_memory_mb, gpu_memory_mb, parse_cuda_device
 from allennlp.common.tqdm import Tqdm
 from allennlp.data.instance import Instance
 from allennlp.data.iterators.data_iterator import DataIterator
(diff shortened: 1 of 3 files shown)


allennlp/training/trainer.py
```diff
@@ -948,13 +961,11 @@ def from_params(cls,
         patience = params.pop_int("patience", None)
         validation_metric = params.pop("validation_metric", "-loss")
         num_epochs = params.pop_int("num_epochs", 20)
-        cuda_device = params.pop_int("cuda_device", -1)
+        cuda_device = params.pop( "cuda_device")
```
DeNeutoy: This parameter still needs to have a default - this is causing a lot of the CI to break.

allennlp/training/trainer.py
```diff
@@ -389,7 +389,20 @@ def _data_parallel(self, batch):
         of torch.nn.parallel.data_parallel to support the allennlp model
         interface.
         """
+        metadata_batch_size = len(batch['metadata']) if 'metadata' in batch and isinstance(batch['metadata'],list) else None
```
DeNeutoy: Please could you add a test for this functionality and annotate it as optional, depending on whether or not the environment has multiple GPUs - you can use 
```
@pytest.mark.skipif(torch.cuda.device_count() < 2,
                    reason="Need multiple GPUs.")
```

MaxMotovilov: I could definitely use some help with that, being a greenhorn! Given that the issue only surfaces in `forward()`, should I write a dummy model that verifies the correspondence between input tensors and metadata and a trainer config to run it? Is there boilerplate code somewhere in tests that I can use for this to avoid inventing the wheel?

allennlp/training/trainer.py
MaxMotovilov: Done - sorry I missed this to begin with

allennlp/training/trainer.py
```diff
@@ -948,13 +961,11 @@ def from_params(cls,
         patience = params.pop_int("patience", None)
         validation_metric = params.pop("validation_metric", "-loss")
         num_epochs = params.pop_int("num_epochs", 20)
-        cuda_device = params.pop_int("cuda_device", -1)
+        cuda_device = params.pop("cuda_device", -1)
```
joelgrus: I don't think this will play well with environment variable substitution.

That is, if someone is parameterizing their config file with a `CUDA_DEVICE` environment variable, it will get read in as a string (like all environment variables), and then this will pop it as a string, which I think will break code that relies on it.

You probably will need to write an ugly helper function that takes a `Union[str, int, List[int]]` and parses the string to `int` or `List[int]` as appropriate

DeNeutoy: Hmm - I think that should be the responsibility of user to do in Jsonnet (presuming that is possible in Jsonnet).

MaxMotovilov: > You probably will need to write an ugly helper function that takes a Union[str, int, List[int]] and parses the string to int or List[int] as appropriate

That isn't even ugly, I think -- shouldn't this be the default approach to processing environment variables? I.e. pick the right alternative out of all acceptable conversion targets and complain if none work or the choice is ambiguous. I'll see if I can come up with something reasonably generic...

MaxMotovilov: > I think that should be the responsibility of user to do
I guess I don't quite understand then -- if there's a way to override the configuration parameter via an environment variable (a type-less string value) for a specific run, how can the creator of the config (or the component that the config is intended for) manage the type inference? There is no schema for the configuration, is there?

joelgrus: @DeNeutoy  the whole point of using `pop_int` here is that we don't think it's the user's responsibility (and how could it be, if they're using our trainer?)

joelgrus: @MaxMotovilov I don't think it's the right approach in general, I think the current `pop_int` + `pop_bool` + `pop_float` is a good approach ("explicit is better than implicit"), it's just in this case we're allowing an int or a list, which is unusual. the function you write should fail if it gets something that can't be parsed as a list or an int

MaxMotovilov: It should be explicit anyway -- the alternatives have to be specified at some point -- which, in absence of a schema, is the point of use. In this particular case however we are not even dealing with a union: a general case of `List[int]` with a special case of one-element list equivalent to an int. The default (which is represented by -1 mostly for legacy reasons, I take it) really is a null value or an empty list: no GPU available.

So, how about adding a `params.pop_int_list()` (my ancient C++ roots cry for `params.pop[List[int]]()` but this is not to be) that handles `/,\s*/`-separated values and a default?

MaxMotovilov: It's your codebase -- you get to call'em :-)

schmmd:
@DeNeutoy is this a good change--but it just requires a single test?

matt-gardner:
Closing in favor of #1724.

PR #1427, opened 2018-06-26
Add support for selective finetune (freeze parameters by regex from config file)

HarshTrivedi:
- Add support in fine_tune to selectively tune
- Add tests for selective fine tuning.
- Allow for turning off gradients in train command.

This PR adds support in `fine-tune` command to freeze parameters layers by passing regex in config file. In config file passed in fine-tune command:
```
"trainer": {
    ...
    "no_grad": ["*conv*", ".*text_embedder*"]
}
```
The above will freeze `conv` layers and `text_embedder`s but not others parameters.
I believed best place to allow `no_grad` key setting accessible to is within `trainer` key and `trainer` is also used in `train` command. So I allowed turning of gradients via `no_grad` in train command as well.  

Although I primarily intended this selective turning off of gradients for fine-tune command, it can be useful in `train` command as well : - eg. If one's modules parameters are loaded / transferred from some other pretrained model and want to freeze it there.

More context: Issue (#1298 )

PR #1448, opened 2018-06-30
Fix logging of no-grad parameters.

HarshTrivedi:
Fix issue in no-grad parameters logging as mentioned by @rulai-huajunzeng in (#1427).

If parameters were already set `requires_grad=False` not through no through nograd regex but other means then they were logged as Tunable instead of Frozen. This pr fixes that.

I have made the headings capitalized. They are more distinguishable this way amidst a long list of parameters.

### DIFF
--- a/allennlp/commands/fine_tune.py
+++ b/allennlp/commands/fine_tune.py
@@ -188,14 +188,15 @@ def fine_tune_model(model: Model,
     for name, parameter in model.named_parameters():
         if any(re.search(regex, name) for regex in no_grad_regexes):
             parameter.requires_grad_(False)
-            nograd_parameter_names.append(name)
-        else:
+        if parameter.requires_grad:
             grad_parameter_names.append(name)
+        else:
+            nograd_parameter_names.append(name)
 
-    logger.info("Following parameters are Frozen  (without gradient):")
+    logger.info("FROZEN PARAMETERS  (WITHOUT GRADIENT):")
     for name in nograd_parameter_names:
         logger.info(name)
-    logger.info("Following parameters are Tunable (with gradient):")
+    logger.info("TUNABLE PARAMETERS (WITH GRADIENT):")
     for name in grad_parameter_names:
         logger.info(name)
 
--- a/allennlp/commands/fine_tune.py
+++ b/allennlp/commands/fine_tune.py
@@ -16,7 +16,8 @@ from allennlp.commands.evaluate import evaluate
 from allennlp.commands.subcommand import Subcommand
 from allennlp.commands.train import datasets_from_params
 from allennlp.common import Params
-from allennlp.common.util import prepare_environment, prepare_global_logging
+from allennlp.common.util import prepare_environment, prepare_global_logging, \
+                                 get_frozen_and_tunable_parameter_names
 from allennlp.data.iterators.data_iterator import DataIterator
 from allennlp.models import load_archive, archive_model
 from allennlp.models.archival import CONFIG_NAME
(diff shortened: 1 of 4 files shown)


allennlp/commands/train.py
```diff
@@ -293,14 +293,15 @@ def train_model(params: Params,
...
             parameter.requires_grad_(False)
-            nograd_parameter_names.append(name)
-        else:
+        if parameter.requires_grad:
             grad_parameter_names.append(name)
+        else:
+            nograd_parameter_names.append(name)
 
-    logger.info("Following parameters are Frozen  (without gradient):")
+    logger.info("FROZEN PARAMETERS  (WITHOUT GRADIENT):")
```
DeNeutoy: No shouting in the logs please 😄

allennlp/commands/fine_tune.py
```diff
@@ -188,14 +188,15 @@ def fine_tune_model(model: Model,
...
+        if parameter.requires_grad:
             grad_parameter_names.append(name)
+        else:
+            nograd_parameter_names.append(name)
 
-    logger.info("Following parameters are Frozen  (without gradient):")
+    logger.info("FROZEN PARAMETERS  (WITHOUT GRADIENT):")
     for name in nograd_parameter_names:
         logger.info(name)
-    logger.info("Following parameters are Tunable (with gradient):")
```
DeNeutoy: Perhaps consider adding a small test for this - given the same code is in both places, consider pulling it out into a function, which would make testing it easier.

HarshTrivedi: Sure. 

Do you mean I add `def get_frozen_and_tunable_parameter_names(parameters: List[Parameter]) -> List` in different place like `allennlp/common/util.py` and use it in both `train.py` and `fine_tune.py`? Something like following?

```python
# allennlp/commons/util.py
def get_frozen_and_tunable_parameter_names(model: torch.nn.Module) -> List:
    frozen_parameter_names = []
    tunable_parameter_names = []
    for name, parameter in model.named_parameters():
        if not parameter.requires_grad:
            frozen_parameter_names.append(name)
        else:
            tunable_parameter_names.append(name)
    return [frozen_parameter_names, tunable_parameter_names]
```
and test could be something like:
```python
def test_get_frozen_and_tunable_parameter_names(self):
    model = torch.nn.Sequential(OrderedDict([
          ('conv', torch.nn.Conv1d(5, 5, 5)),
          ('linear', torch.nn.Linear(5, 10)),
        ])) 
    named_parameters = dict(model.named_parameters())
    named_parameters['linear.weight'].requires_grad_(False)
    named_parameters['linear.bias'].requires_grad_(False)
                   util.get_frozen_and_tunable_parameter_names(model)
    assert set(frozen_parameter_names)  == {'linear.weight', 'linear.bias'}
    assert set(tunable_parameter_names) == {'conv.weight', 'conv.bias'}
```
and train and fine-tune would be:
```python

                   util.get_frozen_and_tunable_parameter_names(model)
        logger.info(name)
        logger.info(name)
```

Seems good?

allennlp/commands/train.py
HarshTrivedi: Ohh Cool, will revert it back then. I thought I was supposed to capitalize the headings similar to "CURRENTLY DEFINED PARAMETERS".

allennlp/commands/fine_tune.py
DeNeutoy: Perfect 👍

HarshTrivedi:
@DeNeutoy  I was busy with something else for last 2 days ... This should be good now.

PR #1455, opened 2018-07-03
Allow to use a different validation iterator from training iterator

matt-peters:


allennlp/training/trainer.py
```diff
@@ -644,10 +649,15 @@ def _validation_loss(self) -> Tuple[float, int]:
 
         self._model.eval()
 
-        val_generator = self._iterator(self._validation_data,
+        if self._validation_iterator is not None:
+            val_iterator = self._validation_iterator
+        else:
+            val_iterator = self._iterator
+
+        val_generator = val_iterator(self._validation_data,
```
joelgrus: I think you have an indentation issue here

PR #1463, opened 2018-07-06
unwrap tensors in avg metric

DeNeutoy:


allennlp/training/metrics/average.py
```diff
@@ -23,7 +23,7 @@ def __call__(self, value):
         value : ``float``
             The value to average.
         """
-        self._total_value += value
+        self._total_value += list(self.unwrap_to_tensors(value))[0]
```
vidurj: What if `value` is a float?

vidurj:
LGTM
Thanks!

PR #1465, opened 2018-07-06
Dependency parser

DeNeutoy:
- Adds a dependency parser for Universal Dependencies along the lines of [Deep Biaffine Attention for Neural Dependency Parsing
](https://arxiv.org/abs/1611.01734)

- Adds a `Metric` for dependency parsing metrics, including unlabeled/labeled attachment scores and exact match sentence level metrics.


allennlp/nn/decoding/chu_liu_edmonds.py
```diff
@@ -0,0 +1,287 @@
...
+    Parameters
+    ----------
+    length : ``int``, required.
+        The number of nodes.
+    score_matrix : ``numpy.ndarray``, required.
+        The score matrix representing the scores for pairs
+        of nodes.
+    current_nodes : ``List[bool]``, required.
+        The nodes which are representatives in the graph.
+        A representative at it's most basic represents a node,
```
joelgrus: its

allennlp/nn/decoding/chu_liu_edmonds.py
```diff
@@ -0,0 +1,287 @@
...
+        but as the algorithm progresses, individual nodes will
+        represent collapsed cycles in the graph.
+    final_edges: ``Dict[int, int]``, required.
+        An empty dictionary which will be populated with the
+        nodes which are connected in the minimum spanning tree.
+    old_input: ``numpy.ndarray``, required.
+    old_input: ``numpy.ndarray``, required.
```
joelgrus: repeated line

allennlp/models/dependency_parser.py
```diff
@@ -0,0 +1,464 @@
...
+from allennlp.models.model import Model
+from allennlp.nn import InitializerApplicator, RegularizerApplicator
+from allennlp.nn.util import get_text_field_mask, get_range_vector
+from allennlp.nn.util import get_device_of, last_dim_log_softmax, get_lengths_from_binary_sequence_mask
+from allennlp.nn.decoding.chu_liu_edmonds import decode_mst
+from allennlp.training.metrics import AttachmentScores
+
+
+@Model.register("dependency_parser")
+class DependencyParser(Model):
```
joelgrus: do we really want to call this `DependencyParser`? that feels like the task name rather than the model name. (I am not sure, I just know that we're trying to be crisper about distinguishing between tasks and models)

also, do you want to add a comment (not sure if class-level or module-level) about which particular model this is, with a link to the paper and so on

allennlp/models/dependency_parser.py
```diff
@@ -0,0 +1,464 @@
...
+            ``TextFieldEmbedder``. This output is a dictionary mapping keys to ``TokenIndexer``
+            tensors.  At its most basic, using a ``SingleIdTokenIndexer`` this is: ``{"tokens":
+            Tensor(batch_size, num_tokens)}``. This dictionary will have the same keys as were used
+            for the ``TokenIndexers`` when you created the ``TextField`` representing your
+            sequence.  The dictionary is designed to be passed directly to a ``TextFieldEmbedder``,
+            which knows how to combine different word representations into a single vector per
+            token in your input.
+        tags : torch.LongTensor, optional (default = None)
+            A torch tensor representing the sequence of integer gold class labels of shape
+            ``(batch_size, num_tokens)``.
```
joelgrus: I think the parameters documentation is out of date

allennlp/models/dependency_parser.py
```diff
@@ -0,0 +1,464 @@
...
+        # shape (batch_size, timesteps, type_representation_dim)
+        head_type_representation = F.elu(self.head_type_projection(encoded_text))
+        child_type_representation = F.elu(self.child_type_projection(encoded_text))
+        head_type_representation = head_type_representation.contiguous()
+        child_type_representation = child_type_representation.contiguous()
+        # shape (batch_size, timesteps, timesteps)
+        attended_arcs = self.arc_attention(head_arc_representation,
+                                           child_arc_representation,
+                                           mask.float(),
+                                           mask.float()).squeeze(1)
```
joelgrus: use `float_mask` here?

allennlp/models/dependency_parser.py
```diff
@@ -0,0 +1,464 @@
...
+
+        minus_inf = -1e8
+        minus_mask = (1 - float_mask) * minus_inf
+        attended_arcs = attended_arcs + minus_mask.unsqueeze(2) + minus_mask.unsqueeze(1)
+
+        if self.training or (self.eval and not self.use_mst_decoding_for_validation):
```
joelgrus: isn't `self.eval` a function, so it will always evaluate as truthy? possibly you just want

```
if self.training or not self.use_mst_decoding_for_validation:
```

allennlp/models/dependency_parser.py
```diff
@@ -0,0 +1,464 @@
...
+            heads, head_types = self._greedy_decode(head_type_representation,
+                                                    child_type_representation,
+                                                    attended_arcs,
+                                                    mask)
+        elif self.use_mst_decoding_for_validation:
```
joelgrus: and then just use `else` here? I stared at it for a minute to figure out what happens if neither condition is hit before realizing that couldn't happen

allennlp/models/dependency_parser.py
```diff
@@ -0,0 +1,464 @@
...
+            token in your input.
+        tags : torch.LongTensor, optional (default = None)
+            A torch tensor representing the sequence of integer gold class labels of shape
+            ``(batch_size, num_tokens)``.
+
+        Returns
+        -------
+        An output dictionary consisting of:
+        loss : torch.FloatTensor, optional
+            A scalar loss to be optimised.
```
joelgrus: there's other stuff in the output dict too

allennlp/models/dependency_parser.py
```diff
@@ -0,0 +1,464 @@
...
+        self._pos_tag_embedding = pos_tag_embedding or None
+        representation_dim = text_field_embedder.get_output_dim()
+        if pos_tag_embedding is not None:
+            representation_dim += pos_tag_embedding.get_output_dim()
+        check_dimensions_match(representation_dim, encoder.get_input_dim(),
+                               "text field embedding dim", "encoder input dim")
+
+        self.use_mst_decoding_for_validation = use_mst_decoding_for_validation
+
+        self._attachement_scores = AttachmentScores()
```
joelgrus: typo: _attachment_scores

allennlp/models/dependency_parser.py
```diff
@@ -0,0 +1,464 @@
...
+        if head_indices is not None and head_tags is not None:
+            arc_nll, type_nll = self._construct_loss(head_type_representation,
+                                                     child_type_representation,
+                                                     attended_arcs,
+                                                     head_indices,
+                                                     head_tags,
+                                                     mask)
+            loss = arc_nll + type_nll
+
+            # We calculate attatchment scores for the whole sentence
```
joelgrus: typo: attachment

allennlp/nn/decoding/chu_liu_edmonds.py
```diff
@@ -0,0 +1,287 @@
+
```
matt-gardner: I didn't really look at this file, as I need to get on to some ACL prep stuff, and I'm not familiar with the algorithm, anyway.  I'll just trust your tests.

allennlp/tests/models/dependency_parser_test.py
```diff
@@ -0,0 +1,32 @@
...
+class DependencyParserTest(ModelTestCase):
+
+    def setUp(self):
+        super(DependencyParserTest, self).setUp()
+        self.set_up_model(self.FIXTURES_ROOT / "dependency_parser" / "experiment.json",
+                          self.FIXTURES_ROOT / "data" / "dependencies.conllu")
+
+    def test_dependency_parser_can_save_and_load(self):
+        self.ensure_model_can_train_save_and_load(self.param_file)
+
```
matt-gardner: Extra blanks lines in a few places here.

allennlp/training/metrics/attachment_scores.py
```diff
@@ -0,0 +1,95 @@
...
+        gold_indices = gold_indices.long()
+        gold_labels = gold_labels.long()
+
+        correct_indices = predicted_indices.eq(gold_indices).long() * mask
+        unlabeled_exact_match = (correct_indices + (1 - mask)).prod(dim=-1)
+        correct_labels = predicted_labels.eq(gold_labels).long() * mask
+        correct_labels_and_indices = correct_indices * correct_labels
+        labeled_exact_match = (correct_labels_and_indices + (1 - mask)).prod(dim=-1)
+
+        self._unlabeled_count += correct_indices.sum()
```
matt-gardner: `_unlabed_count` feels like a denominator name to me, not a numerator.  It'd be more obvious as `self._unlabeled_correct`.  Same for the others.

allennlp/models/dependency_parser.py
DeNeutoy: Good catch, thanks 👍

allennlp/models/dependency_parser.py
```diff
@@ -0,0 +1,464 @@
...
+
+        # This next statement is quite a complex piece of indexing. What we are doing here is:
+
+        # 1. Indexing in to the batch dimension with an index vector.
+        #    Imagine that this momentarily creates a batch_size number of
+        #    (timesteps, embedding_dim) shaped tensors.
+        # 2. Indexing into the timesteps dimension with the releveant indexes which we
+        #    require for this batch element. Because we already created a 'virtual'
+        #    (timesteps, embedding_dim) shaped tensor _for this batch element_, we
+        #    need to transpose the indices to match the first dimension's shape.
```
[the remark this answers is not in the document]
DeNeutoy: Sorry, maybe my explanation wasn't very good. Does this snippet help? This was the most intuitive way I could think of it, let me know if you can think of a better way.

```
>>> x
tensor([[[ 1.,  1.,  1.,  1.],
         [ 2.,  2.,  2.,  2.],
         [ 3.,  3.,  3.,  3.]],

        [[ 4.,  4.,  4.,  4.],
         [ 5.,  5.,  5.,  5.],
         [ 6.,  6.,  6.,  6.]]])
>>> x.size()
torch.Size([2, 3, 4])
>>> range
tensor([ 0,  1])
>>> indices
tensor([[ 0,  1,  2],
        [ 2,  1,  0]])
>>> indices2
tensor([[ 1,  1,  1],
        [ 1,  1,  1]])
>>> x[range, indices.t()].transpose(0,1)
tensor([[[ 1.,  1.,  1.,  1.],
         [ 2.,  2.,  2.,  2.],
         [ 3.,  3.,  3.,  3.]],

        [[ 6.,  6.,  6.,  6.],
         [ 5.,  5.,  5.,  5.],
         [ 4.,  4.,  4.,  4.]]])
>>> x[range, indices2.t()].transpose(0,1)
tensor([[[ 2.,  2.,  2.,  2.],
         [ 2.,  2.,  2.,  2.],
         [ 2.,  2.,  2.,  2.]],

        [[ 5.,  5.,  5.,  5.],
         [ 5.,  5.,  5.,  5.],
         [ 5.,  5.,  5.,  5.]]])
```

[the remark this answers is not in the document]
matt-gardner: Ok, I did a bunch of reading on advanced indexing, so I finally understand what's going on.  Just giving a link to [the documentation on advanced indexing](https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.indexing.html#advanced-indexing) is probably better than trying to explain it here, though.  It took me a whole lot of time reading those docs and playing around with the tensors to understand; I'd probably just leave a note saying something like "This uses advanced indexing, which you really need to read the docs to understand."  What finally did it for me was the discussion about subspaces in [this section](https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.indexing.html#combining-advanced-and-basic-indexing).

And, I think that discussion pointed out that you don't need your transposes, you just need to force the broadcasting that you want:

```python
>>> x
(0 ,.,.) = 
  1  1  1  1
  2  2  2  2
  3  3  3  3
(1 ,.,.) = 
  4  4  4  4
  5  5  5  5
  6  6  6  6
[torch.FloatTensor of size 2x3x4]

>>> range
 0
 1
[torch.LongTensor of size 2]

>>> indices
 0  1  2
 2  1  0
[torch.LongTensor of size 2x3]

>>> x[range[:, None], indices]
(0 ,.,.) = 
  1  1  1  1
  2  2  2  2
  3  3  3  3
(1 ,.,.) = 
  6  6  6  6
  5  5  5  5
  4  4  4  4
[torch.FloatTensor of size 2x3x4]
```

What you want is to replace the `(2, 3)` subspace from the original `x` tensor with the `(2, 3)` subspace from the `indices` tensor, keeping the same batching by using the `range` tensor.  This requires you to be able to broadcast `(2,)` to `(2, 3)`, which pytorch doesn't like for some reason unless you explicitly add a dimension to `range` by doing `range[:, None]`.  You can also do this with an `.unsqueeze(1)` when you create `range`.  Removing the transposes also makes the logic a whole lot easier to understand - it's a lot easier to see that you're just doing a batched index select, as `range` is doing the "batched" part of that.

allennlp/models/dependency_parser.py
```diff
@@ -0,0 +1,464 @@
...
+
+        # shape (batch_size, timesteps, num_head_tags)
+        head_type_logits = self._get_head_types(head_type_representation, child_type_representation, head_indices)
+        normalised_head_type_logits = last_dim_log_softmax(head_type_logits,
+                                                           mask.unsqueeze(-1)) * float_mask.unsqueeze(-1)
+        # index matrix with shape (timesteps, batch)
+        child_index = torch.arange(0, timesteps).view(timesteps, 1).expand(timesteps, batch_size)
+        child_index = child_index.type_as(attended_arcs).long()
+        # shape (timesteps, batch_size)
+        arc_loss = normalised_arc_logits[range_vector, child_index, head_indices.data.t()]
```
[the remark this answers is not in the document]
matt-gardner: Again, here, if you just `.unsqueeze(1)` your range vector, you should be able to keep the `child_index` and `head_tags` indices with shape `(batch_size, sentence_length)`, which would make this logic easier on the brain.

allennlp/models/biaffine_dependency_parser.py
```diff
@@ -0,0 +1,493 @@
...
+            embedded_pos_tags = self._pos_tag_embedding(pos_tags)
+            embedded_text_input = torch.cat([embedded_text_input, embedded_pos_tags], -1)
+        elif self._pos_tag_embedding is not None:
+            raise ConfigurationError("Model uses a POS embedding, but no POS tags were passed.")
+
+        mask = get_text_field_mask(words)
+        float_mask = mask.float()
+        encoded_text = self.encoder(embedded_text_input, mask)
+
+        # shape (batch_size, timesteps, arc_representation_dim)
```
matt-gardner: I still have a strong preference for something with more semantic content than "timesteps" here; if you really prefer "timesteps", then leave it, I just think it's a pretty vacuous (and misleading) term.  You have "num_tokens" in part of your docstring above, so whatever you pick, make sure you update the docstring to match.

ChunchuanLv:
Hi there,

Thank you very much for writing and publishing this very handy code. Is it possible that you can share the json file that you use to train/configure the model?

Regards,
Chunchuan

PR #1474, opened 2018-07-10
fix BasicTextFieldEmbedder.from_params to reflect the constructor

joelgrus:
very little code change, lots of config.json changes.

is backward compatible with a DeprecationWarning


allennlp/modules/text_field_embedders/basic_text_field_embedder.py
```diff
@@ -84,7 +85,27 @@ def forward(self, text_field_input: Dict[str, torch.Tensor], num_wrapping_dims:
...
+        # key-value pairs to be top-level in the params object. In particular, this caused the automatic
+        # FromParams.from_params implementation to do the wrong thing with config files that were
+        # designed for the (idiosyncratic) old behavior.
+        #
+        # This updated implementation delegates to the automatic "from_params" method when it can,
+        # but (for now) it continues to support the old way with a `DeprecationWarning`.
+        if 'token_embedders' in params:
+            # We use `TextFieldEmbedder` as the first argument to super() so that we get
+            # Registrable.from_params.
+            return super(TextFieldEmbedder, cls).from_params(params=params, vocab=vocab)
```
matt-gardner: Can you just do `super().from_params(...)` here?

allennlp/tests/modules/text_field_embedders/basic_text_field_embedder_test.py
```diff
@@ -103,3 +103,45 @@ def test_forward_runs_with_non_bijective_mapping(self):
...
+                                "embedding_dim": 5
+                                },
+                        "words3": {
+                                "type": "embedding",
+                                "embedding_dim": 3
+                                }
+                        }
+                })
+
+        print(new_params)
```
matt-gardner: Remove print?

allennlp/modules/text_field_embedders/basic_text_field_embedder.py
joelgrus: I could, that would work because `TextFieldEmbedder` itself has no custom `from_params` method and so will delegate further up the inheritance chain to the right method.

however, if someone were to at some point decide we needed a custom implementation for `TextFieldEmbedder.from_params`, then using `super()` would call that custom implementation (and do the wrong thing).

am I being overly defensive? possibly

matt-gardner: Oh, I missed that this is skipping a class in the hierarchy (skipped over the comment...).  Can't you just be explicit about it and call `FromParams.from_params()`?

joelgrus: that doesn't just work, because it's bound to the `Registrable` class and you can't pass in the `BasicTextFieldEmbedder` class to bind to.

however, I figured out how to get it to work, you have to call `__func__` to get the unbound function.

matt-gardner: I don't have a strong preference about how to do this, I was mostly just curious about what was going on.  Whatever you think is best is fine with me.  If it were me, I'd probably pick just calling `super().from_params(...)`, as I think the argument about something maybe changing above us applies to all of our classes, and we don't worry about it anywhere else.

joelgrus:
it turns out that this discrepancy is making my "configuration validator" not work (and it turns out that it made the configuration wizard incorrect, but no one ever noticed 😬 )

since no one objected, I'm going to merge it

PR #1483, opened 2018-07-12
fine grained ner reader

DeNeutoy:
Dataset reader for fine grained NER on the Ontonotes 5.0 corpus. I kind of CBA to write a converter between BIO and BIOUL but if you think the distinction is important I can add it


matt-peters:
I wouldn't worry about converting between tagging schemes, either will work.

PR #1485, opened 2018-07-13
Add option to have tie breaking in Categorical Accuracy

HarshTrivedi:
This PR adds an option to have tie-breaking in categorical accuracy.

 If the model predicts a k-way tie that includes the correct class, then score 1/k points for that instance rather than 1 for calculating accuracy. By default the tie-breaking happens arbitrarily.

ARC datasets have k-way tie breaking based scores ([link1](https://github.com/allenai/ARC-Solvers/blob/master/arc_solvers/processing/calculate_scores.py), [link2](http://data.allenai.org/arc/#leaderboard))

### DIFF
--- a/allennlp/training/metrics/categorical_accuracy.py
+++ b/allennlp/training/metrics/categorical_accuracy.py
@@ -12,9 +12,17 @@ class CategoricalAccuracy(Metric):
     """
     Categorical Top-K accuracy. Assumes integer labels, with
     each item to be classified having a single correct class.
+    Tie break enables equal distribution of scores among the
+    classes with same maximum predicted scores.
     """
-    def __init__(self, top_k: int = 1) -> None:
+    def __init__(self, top_k: int = 1, tie_break: bool = False) -> None:
+        if top_k > 1 and tie_break:
+            raise ConfigurationError("Tie break in Categorical Accuracy "
+                                     "can be done only for maximum (top_k = 1)")
+        if top_k <= 0:
+            raise ConfigurationError("top_k passed to Categorical Accuracy must be > 0")
         self._top_k = top_k
+        self._tie_break = tie_break
         self.correct_count = 0.
         self.total_count = 0.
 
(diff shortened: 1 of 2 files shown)


DeNeutoy:
Could you describe your use case for this in practice? A neural network is extremely unlikely to predict an exact tie. Are you doing some rounding of your output distribution or something?

HarshTrivedi:
I have been using [this](https://github.com/allenai/ARC-Solvers/blob/master/arc_solvers/processing/calculate_scores.py) script to find scores for ARC predictions file. It adds an extra overhead since I need to generate predictions in appropriate format by constructing the predictor and then running the linked script. I wanted to have see scores by `evaluate` command and also have live logging during the training. 

Initial intention was to have a metric specific for ARC, but then ended up weaving it in Categorical accuracy. For ARC at least, the tie-breaking is likely since multiple options might have same set of retrieved setences and hence the same input and hence same output logits.

allennlp/training/metrics/categorical_accuracy.py
```diff
@@ -44,18 +52,29 @@ def __call__(self,
...
+                top_k = predictions.topk(min(self._top_k, predictions.shape[-1]), -1)[1]
 
-        # This is of shape (batch_size, ..., top_k).
-        correct = top_k.eq(gold_labels.long().unsqueeze(-1)).float()
+            # This is of shape (batch_size, ..., top_k).
+            correct = top_k.eq(gold_labels.unsqueeze(-1)).float()
+        else:
+            # prediction is correct if gold label falls on any of the max scores. distribute score by tie_counts
+            max_predictions = predictions.max(-1)[0]
+            max_predictions_mask = predictions.eq(max_predictions.unsqueeze(-1).expand_as(predictions))
```
DeNeutoy: Why do you need the `expand_as` here? I think this is already the same shape as `predictions`.

allennlp/training/metrics/categorical_accuracy.py
```diff
@@ -44,18 +52,29 @@ def __call__(self,
...
+            correct = max_predictions_mask[torch.arange(gold_labels.numel()).long(), gold_labels].float()
```
DeNeutoy: This line would be clearer as:

```
correct = torch.index_select(max_predictions_mask, 1, gold_labels).float()
```

allennlp/training/metrics/categorical_accuracy.py
```diff
@@ -44,18 +52,29 @@ def __call__(self,
...
+            tie_counts = max_predictions_mask.sum(-1)
+            correct /= tie_counts.float()
+            correct.unsqueeze_(-1)
```
DeNeutoy: If you remove this unsqueeze, you could replace the view below to be `.view(-1)` or something.

HarshTrivedi: Yeah, but I didn't do that because of the following line above it (in **if clause**)
```
            # This is of shape (batch_size, ..., top_k).
            correct = top_k.eq(gold_labels.unsqueeze(-1)).float()
```
It is necessary to have last dimension unsqueezed here because `top_k` might be `batch_size X 1` or `batch_size X top_k`. Due to this I keep `correct` in **else clause** also unsqueezed so that I don't need to handle these things separately in `correct *= mask.view(-1, 1).float()`. Overall this ends up being cleaner.

allennlp/training/metrics/categorical_accuracy.py
HarshTrivedi: Actually, it is not the same thing. This operation checks if gold_label for an instance lies among the max prediction classes. Lets say `max_predictions` is `batch_size X num_classes` shape and `gold_labels` is `batch_size` shape. `max_predictions` would have 1s in last dimension for classes which have same as max prediction and 0s otherwise. I need to check if indices in gold label are among the max predicted classes or not.

`ith` entry in `gold_labels` has indices on vector of `ith` row in `max_predictions`. I need to index on 1st dimension and 2nd dimension together and so extra `torch.arrange(..)` in first dimension. What you suggest will return me `batch_size X batch_size` tensor but I need `batch_size` tensor.  

Still if I need to use, I need to add extra `diag()`.
```
# Following does work and passes tests though
correct = torch.index_select(max_predictions_mask, 1, gold_labels).diag().float()
```
This looks cleaner, but I has some redundance. What do you suggest?

allennlp/training/metrics/categorical_accuracy.py
HarshTrivedi: If predictions is `batch_size X num_classes`, then max_predictions is `batch_size` because it maxed out on last dim. So I unsqueeze it to make `batch_size X 1` and then `expand_as` so as to copy it across last dimension (and so `batch_size X num_classes`).

DeNeutoy: Oh sorry - I didn't see that this expand was inside the `.eq`. I still don't think you need it, because just be expanding, the `eq` will broadcast along the dimensions.

allennlp/training/metrics/categorical_accuracy.py
DeNeutoy: Sure, sounds good.

allennlp/training/metrics/categorical_accuracy.py
DeNeutoy: Ok your solution is better, but can you add a comment because it's a pretty complicated bit of indexing.

allennlp/training/metrics/categorical_accuracy.py
HarshTrivedi: Ohh right! Done and works perfect!

allennlp/training/metrics/categorical_accuracy.py
HarshTrivedi: Sure, I will add that.

PR #1487, opened 2018-07-13
Make usage of make-vocab, dry-run consistent with train and allow 'extend' to be used by both

HarshTrivedi:
Usage of `vocab_params` ("vocabulary" key in config.json) in `make-vocab` command is not in sync with usage of `vocab_params` in other commands. IMO, behaviour of `vocab_params` should be similar in both.

In `train` command, `"vocabulary": "directory_path"` is the input for vocabulary construction. If `"extend"=true` is given, then vocabulary would be loaded from "directory_path" and dataset-read-instances will be used to extend it with. Otherwise, vocabulary is loaded from "directory_path" and is used as it is. 

```
# vocab_params
"vocabulary": {
    "directory_path": "vocabulary_dir_path"
    "extend": true # false
}
```
In both cases, the output vocabulary is dumped in `serialization_dir/vocabulary`.

However, in `make-vocab`, `"vocabulary": "directory_path"` is taken as the **output path**, where vocabulary will be dumped instead of input path. This is not consistent with other usage and makes it somewhat confusing. More importantly, because of this, `extend` key is **ignored** and **prevents** to use of `make-vocab` for **vocab extension**.

This PR changes input-output for `make-vocab` command. Input is through params of "vocabulary" key just like `train` and dumps is in `serialization_dir/vocabulary`. Hence `vocab_params` behaves in `make-vocab` exactly as `train` and vocab-extention happens implicitly if `extend` is passed as true. It also works if vocab_params are absent. But now, one does need to pass serialization_dir just like `train` for output.

I had to change some tests because of this change.

### DIFF
--- a/allennlp/commands/dry_run.py
+++ b/allennlp/commands/dry_run.py
@@ -97,6 +97,9 @@ def dry_run_from_params(params: Params, serialization_dir: str) -> None:
         if dataset not in all_datasets:
             raise ConfigurationError(f"invalid 'dataset_for_vocab_creation' {dataset}")
 
+    logger.info("From dataset instances, %s will be considered for vocabulary creation.",
+                ", ".join(datasets_for_vocab_creation))
+
     instances = [instance for key, dataset in all_datasets.items()
                  for instance in dataset
                  if key in datasets_for_vocab_creation]
--- a/allennlp/commands/dry_run.py
+++ b/allennlp/commands/dry_run.py
@@ -87,8 +87,8 @@ def dry_run_from_params(params: Params, serialization_dir: str) -> None:
     vocab_dir = os.path.join(serialization_dir, "vocabulary")
 
     if os.path.isdir(vocab_dir) and os.listdir(vocab_dir) is not None:
-        raise ConfigurationError("The 'vocabulary' directory in serialization"
-                                 " directory that you provided is non-empty")
+        raise ConfigurationError("The 'vocabulary' directory in the provided"
+                                 "serialization directory is non-empty")
 
     all_datasets = datasets_from_params(params)
     datasets_for_vocab_creation = set(params.pop("datasets_for_vocab_creation", all_datasets))
--- a/allennlp/commands/dry_run.py
+++ b/allennlp/commands/dry_run.py
@@ -86,6 +86,10 @@ def dry_run_from_params(params: Params, serialization_dir: str) -> None:
     os.makedirs(serialization_dir, exist_ok=True)
     vocab_dir = os.path.join(serialization_dir, "vocabulary")
 
+    if os.path.isdir(vocab_dir) and os.listdir(vocab_dir) is not None:
+        raise ConfigurationError("The 'vocabulary' directory in serialization"
+                                 " directory that you provided is non-empty")
+
     all_datasets = datasets_from_params(params)
     datasets_for_vocab_creation = set(params.pop("datasets_for_vocab_creation", all_datasets))
 
--- a/allennlp/commands/dry_run.py
+++ b/allennlp/commands/dry_run.py
@@ -26,21 +26,17 @@ a model.
     --include-package INCLUDE_PACKAGE
                             additional packages to include
 """
-from typing import Dict, List
-from collections import defaultdict
 import argparse
 import logging
 import os
-
-import numpy
 
 from allennlp.commands.train import datasets_from_params
 from allennlp.commands.subcommand import Subcommand
 from allennlp.common.checks import ConfigurationError
 from allennlp.common.params import Params
 from allennlp.common.util import prepare_environment
-from allennlp.data.vocabulary import DEFAULT_NON_PADDED_NAMESPACES
-from allennlp.data import Vocabulary, Instance
+from allennlp.data import Vocabulary
+from allennlp.data.dataset import Batch
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
@@ -87,14 +83,8 @@ def dry_run_from_params(params: Params, serialization_dir: str) -> None:
     prepare_environment(params)
 
     vocab_params = params.pop("vocabulary", {})
-    vocab_dir = vocab_params.pop('directory_path', None)
-
-    if vocab_dir is not None:
-        logger.info("Found a vocabulary.directory_path parameter in your config. "
-                    "Also saving the vocab we create to that location.")
-        os.makedirs(vocab_dir, exist_ok=True)
-
     os.makedirs(serialization_dir, exist_ok=True)
+    vocab_dir = os.path.join(serialization_dir, "vocabulary")
 
     all_datasets = datasets_from_params(params)
     datasets_for_vocab_creation = set(params.pop("datasets_for_vocab_creation", all_datasets))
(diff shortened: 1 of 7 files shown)


HarshTrivedi:
I haven't looked at `dry-run` before but its behaviour seems different than `train`. eg. vocabulary extend would surely be ignored here since `Vocab.from_params` is not used... Let me go through it and get back in a while ...

HarshTrivedi:
IMO `vocab_params` should be dealing only with the **input** of vocabulary construction and sys-argument passed `serializatio_dir/vocabulary` should be the output where the vocabulary constructed will be dumped. And there should be only one common (across all commands) interpretation of `vocab_params` that decides how to construct vocabulary that will be used. 

Until current commit in this PR, this is made consistent for `make-vocab` and `train`. `vocab_params` is used for input and its construction semantics is decided by `Vocab.from_params`. And output is always in sys-arg passed serializatio_dir (in "vocabulary" dir). 

The behaviour for `dry-run` should also be exact same for avoiding confusions of what is going to be input and where will be the output.

Here is the updated `dry_run_from_params` (I only needed to comment unnecessary things).

```python
def dry_run_from_params(params: Params, serialization_dir: str) -> None:
    prepare_environment(params)

    vocab_params = params.pop("vocabulary", {})
    # vocab_dir = vocab_params.pop('directory_path', None)

    # if vocab_dir is not None:
    #     logger.info("Found a vocabulary.directory_path parameter in your config. "
    #                 "Also saving the vocab we create to that location.")
    #     os.makedirs(vocab_dir, exist_ok=True)

    all_datasets = datasets_from_params(params)
    datasets_for_vocab_creation = set(params.pop("datasets_for_vocab_creation", all_datasets))

    for dataset in datasets_for_vocab_creation:
        if dataset not in all_datasets:
            raise ConfigurationError(f"invalid 'dataset_for_vocab_creation' {dataset}")

    logger.info("Creating a vocabulary using %s data.", ", ".join(datasets_for_vocab_creation))

    instances = [instance for key, dataset in all_datasets.items()
                 for instance in dataset
                 if key in datasets_for_vocab_creation]

    # vocabulary = verbosely_create_vocabulary(vocab_params, instances)
    vocabulary = Vocabulary.from_params_verbosely(vocab_params, instances) # Yet to add!

    logger.info(f"writing the vocabulary to {serialization_dir}.")
    vocabulary.save_to_files(os.path.join(serialization_dir, "vocabulary"))
    # if vocab_dir is not None and os.path.exists(vocab_dir) and os.listdir(vocab_dir) is not None:
    #     logger.info(f"You passed a vocabulary.directory_path in your config which already exists "
    #                 f"and is non-empty. Refusing to overwrite - we saved it to {serialization_dir} instead.")
    # elif vocab_dir is not None:
    #     logger.info(f"You passed a vocabulary.directory_path in your config which was empty. Also "
    #                 f"writing the vocabulary to {vocab_dir}.")
    #     vocabulary.save_to_files(vocab_dir)

```

Just to make sure, we are saying the thing!

HarshTrivedi:
Another thing:

I think ideal place for `verbosely_create_vocabulary` should be in vocabulary class with name `Vocabulary.from_params_verbosely( ... )` or something like that. Although docstring says `This is exactly the code in Vocabulary.from_params`, if someone changes logic of `Vocabulary.from_params` it difficult to know to change things somewhere else aswell since the logic is duplicated. 

When I changed logic of` Vocabulary.from_params` (eg. adding extend key) I didn't realize the same logic could be duplicated somewhere else. Hence, I would like to add class method in Vocabulary class as `from_params_verbosely` and make sure it does have changes of `from_params` reflected there along with extra statistics printing.  And remove `verbosely_create_vocabulary` from `dry_run.py`

Seems good?

HarshTrivedi:
This should look much better now. There was an unsynced code duplication in `Vocab.from_params` and `verbosely_create_vocabulary`. That has been cleaned and duplication is removed to prevent future such problems. The behaviours of `dry-run` and `make-vocab` is same now except the logging statistics in `dry-run`. Hence `vocab_params` (used as input) behave consistently across all the commands. And output is also consistent across commands (in sys-arg passed `serialization_dir/vocabulary`)

I am yet to add 2 tests to check if sys printing is being done appropriately. Will have to figure it out how to do that ...

HarshTrivedi:
> I think the correct behaviour here is for dry-run to check if you've passed a vocab_dir - if you have and it is empty, we continue as normal - if you have and it has a vocab in it, we copy the vocab to a original_vocab directory, and extend the vocabulary with the instances in the dataset. That way, you get no surprises if you didn't want the vocab extended, but it does the most useful thing in most circumstances.

This kind of specific logic wouldn't be the same as how `train` (and now `make-vocab`) decides how to construct vocab to use. Users do have specific control on how to extend vocabulary by explicitly configuring it. So such implicit choices based on directory is empty or not might be confusing. Instead, I think it would be better to leave that logic on a one common interpretation of vocab_params across all commands.

Crudely after (#1416), this is how final vocabulary is decided in `train`
```
If vocabulary[directory_path] is passed:
	If vocabulary[extend] is True
		final_vocab => Load vocabulary[directory_path] and extend it with dataset instances.
	Else:
		final_vocab => Load vocabulary[directory_path]
Else:
	final_vocab => Build from dataset instances
```

More specifically, 4 possibilities of **INPUT** interpretations of vocab_params are:

**A.**: final-vocabulary => load "presaved_vocab", *extend with* dataset instances.
```json
"vocabulary": {
	"directory_path": "presaved_vocab",
	"extend": true
}

```

**B.**: final-vocabulary => load "presaved_vocab"
```json
"vocabulary": {
	"directory_path": "presaved_vocab",
	"extend": false
}

```

**C.**: final-vocabulary => build from tokens in dataset instances.
```json
"vocabulary": {
	"extend": false
}

```

**D.**: Error: Asked to extend but no vocabulary passed to extend it from.
```json
"vocabulary": {
	"extend": true
}

```
"extend": False is default in all cases.


For all these choices, the decided final-vocabulary (**OUTPUT**) is saved in `serialization_dir/vocabulary` irrespective of what vocab_params is.

This gives a good control on making, composing multiple vocabularies by explicitly configuring. So I think decisions like: when to extend, how to extend, what is the final vocabulary used and dumped and where is it dumped should be left out on this only (in all of `train`, `make-vocab`, `dry-run` commands). 

After this PR, For **all** 3 commands:
1. (**INPUT**): we leave vocab_params interpretation fully and only to `Vocabulary.from_params(...)` **instead of**:
   a. pre-popping sub params to make certain decisions (*as in make-vocab and dry-run*)
   b. using something else like `verbosely_create_vocabulary(..)` (*as in dry-run*)
2. (**OUTPUT**) dump the final vocabulary used in "serialization_dir/vocabulary".

allennlp/data/vocabulary.py
```diff
@@ -587,3 +588,33 @@ def __str__(self) -> str:
         namespaces = [f"\tNamespace: {name}, Size: {self.get_vocab_size(name)} \n"
                       for name in self._index_to_token]
         return " ".join([base_string, non_padded_namespaces] + namespaces)
+
+    def print_statistics(self, ignore_error: bool = False) -> None:
+        if self._retained_counter:
+            # Caveat: Printed info is only for part of vocabulary generated from instances. If Vocabulary
```
DeNeutoy: Please can you print this caveat before the statistics are printed? It's moderately important.

allennlp/data/vocabulary.py
```diff
@@ -587,3 +588,33 @@ def __str__(self) -> str:
...
+                tokens_with_counts.sort(key=lambda x: len(x[0]), reverse=True)
+
+                print(f"\nTop 10 longest tokens in namespace '{namespace}':")
+                for token, freq in tokens_with_counts[:10]:
+                    print(f"\tToken: {token}\t\tlength: {len(token)}\tFrequency: {freq}")
+
+                print(f"\nTop 10 shortest tokens in namespace '{namespace}':")
+                for token, freq in reversed(tokens_with_counts[-10:]):
+                    print(f"\tToken: {token}\t\tlength: {len(token)}\tFrequency: {freq}")
+        else:
```
DeNeutoy: I think we can just remove the `ignore_error` flag and instead, log a warning that statistics can't be generated if you haven't created the vocabulary from instances.

allennlp/commands/make_vocab.py
```diff
@@ -63,35 +70,36 @@ def make_vocab_from_args(args: argparse.Namespace):
...
-    vocab_dir = vocab_params.pop('directory_path', None)
-    if vocab_dir is None:
-        raise ConfigurationError("To use `make-vocab` your configuration must contain a value "
-                                 "at vocabulary.directory_path")
+    os.makedirs(serialization_dir, exist_ok=True)
+    vocab_dir = os.path.join(serialization_dir, "vocabulary")
 
-    os.makedirs(vocab_dir, exist_ok=True)
+    if os.path.isdir(vocab_dir) and os.listdir(vocab_dir) is not None:
+        raise ConfigurationError("The 'vocabulary' directory in the provided"
```
DeNeutoy: Extra space needed here at the end of one of the lines.

allennlp/commands/dry_run.py
```diff
@@ -87,14 +83,12 @@ def dry_run_from_params(params: Params, serialization_dir: str) -> None:
...
-    if vocab_dir is not None:
-        logger.info("Found a vocabulary.directory_path parameter in your config. "
-                    "Also saving the vocab we create to that location.")
-        os.makedirs(vocab_dir, exist_ok=True)
-
     os.makedirs(serialization_dir, exist_ok=True)
+    vocab_dir = os.path.join(serialization_dir, "vocabulary")
+
+    if os.path.isdir(vocab_dir) and os.listdir(vocab_dir) is not None:
+        raise ConfigurationError("The 'vocabulary' directory in the provided"
```
DeNeutoy: Extra space needed here in the error message.

allennlp/data/vocabulary.py
```diff
@@ -460,6 +460,7 @@ def _extend(self,
         counter = counter or {}
         tokens_to_add = tokens_to_add or {}
 
+        self._retained_counter = counter
```
DeNeutoy: Please can you set this attribute to `None` in __init__ before you use it? We broadly try to only modify attributes in methods, rather than set them.

allennlp/data/vocabulary.py
HarshTrivedi: Sure 👍

allennlp/data/vocabulary.py
HarshTrivedi: Sure 👍

allennlp/commands/make_vocab.py
HarshTrivedi: Sure 👍

allennlp/commands/dry_run.py
HarshTrivedi: Sure 👍

allennlp/data/vocabulary.py
HarshTrivedi: Sure 👍

allennlp/tests/commands/dry_run_test.py
```diff
@@ -36,20 +41,121 @@ def setUp(self):
...
+        parser = argparse.ArgumentParser(description="Testing")
+        subparsers = parser.add_subparsers(title='Commands', metavar='')
+        DryRun().add_subparser('dry-run', subparsers)
+        for serialization_arg in ["-s", "--serialization-dir"]:
+            raw_args = ["dry-run", "path/to/params", serialization_arg, "serialization_dir"]
+            args = parser.parse_args(raw_args)
+            assert args.func == dry_run_from_args
+            assert args.param_path == "path/to/params"
+            assert args.serialization_dir == "serialization_dir"
+
```
HarshTrivedi: @DeNeutoy  Do you think following two tests should be added? I was just planning to test if following two strings are logged on stdout "--Dataset Statistics--" and "--Vocabulary Statistics--" respectively. I tried very shortly something as suggested [here](https://stackoverflow.com/questions/20507601/writing-a-pytest-function-for-checking-the-output-on-console-stdout-in-python) but couldn't get it to work. Such hardcoded string checking doesn't seem correct in test, but don't know what else to test as far as print_statistics methods are concerned.

Btw, Running `pytest -s allennlp/tests/commands/dry_run_test.py` does show me that they are indeed printed.

DeNeutoy: You don't need to test that things are printed, in this case - someone would notice, and it has no side-effects or experimental impact, so I think it's fine.

HarshTrivedi:
@DeNeutoy  I will get back on this tomorrow. I am busy with something today and apart of what you suggested, I want to review the added tests again ...

HarshTrivedi:
@DeNeutoy  Thank you for your comments!
This should be good now. There are two small changes (last 2 commits) apart of your suggestions.

Besides that, I need opinion on following: 

From the name `dry-run`, I think it's behaviour should be to do and log everything except actually running the trainer assurring that nothing leads to an error and all the configurations are as expected by user. In that case, do you think it would be appropriate to add following lines in `dry_run_from_params`:

```python
model = Model.from_params(vocab=vocab, params=params.pop('model'))
trainer_params = params.pop("trainer")
no_grad_regexes = trainer_params.pop("no_grad", ())
for name, parameter in model.named_parameters():
    if any(re.search(regex, name) for regex in no_grad_regexes):

               get_frozen_and_tunable_parameter_names(model)
    logger.info(name)
    logger.info(name)
```

This would then **segregate the purpose** of `make-vocab` and `dry-run`. `make-vocab` would be used to make / compose vocabularies. And `dry-run` can be used as pre `train` step to make sure everything is in place: it would do 3 additional things: index the dataset with the created vocabulary, make model instance and log the trainable / frozen parameters - all this more verbosely.

DeNeutoy:
Sounds good to me - I want `dry-run` to be able to replace `make-vocab` eventually, and particularly, if you call `dry-run` with no existing vocab, I want one to be created and serialised from the instances. So long as the changes meet those constraints, i'm happy!

HarshTrivedi:
Cool, then I would add those lines. Currently `make_vocab_from_params` is basically subset of `dry_run_from_params` (only 3-4 indexing and printing lines extra). After I add above lines, it would still be subset and so `dry-run` can replace `make-vocab` right away. 

But are you suggesting, that eventually we would want to deprecate `make-vocab`?

DeNeutoy:
Yes, exactly - it's a good thing to force people to run `dry-run`, I think. It catches nasty data bugs earlier.

HarshTrivedi:
That's true. But, in some cases, one can need a separate handle on vocabulary construction and which can later be used for training purpose. Eg. lets say I want to create a vocabulary out of three different dataset readers / datasets or even 3-4 different experiment configs and then use this collective vocabulary everywhere. This can be done by `make-vocab` pretty easily. Eg. from the comment [here](https://github.com/allenai/allennlp/commit/0312b16e23263d24401d7d89b8c327fa0b7259c3#commitcomment-29730621).

Even if `dry-run` does everything that `make-vocab` does, `make-vocab` can come pretty handy for composing vocabularies without doing extra steps (indexing, statistics printing, model loading) that `dry-run` would do. Role of `make-vocab` won't be to catch errors, but to pre construct vocabulary if someone needs.

Role of `dry-run` could be to check that everything is working without errors and configs are appropriate, without actually `train` command. Since now, even model loading is being done, errors in that also would be pre-caught.

DeNeutoy:
That's true, I guess - no harm in leaving it there if you're finding it useful! Thanks for the PR.

HarshTrivedi:
I am done. I think, this is good for merging-in now ...

PR #1491, opened 2018-07-15
Fix SpanBasedF1Measure for tags without conll labels

nelson-liu:
Fixes `SpanBasedF1Measure` to work if the tags don't have conll labels (i.e., the labels are simply `{B, I, O}`).

The only thing that had to be changed was the check to see whether the `active_conll_tag` is not `None`, since `""` is `False`.


DeNeutoy:
Can you add something to the docstring noting that you can use this without class labels?

PR #1493, opened 2018-07-16
Ensure Contiguous Hidden State Tensors in Encoders 

dangitstam:
`index_select()` doesn't ensure a contiguous tensor, which can cause a runtime exception since PyTorch RNNs require them to be .

### DIFF
--- a/allennlp/modules/encoder_base.py
+++ b/allennlp/modules/encoder_base.py
@@ -104,10 +104,10 @@ class _EncoderBase(torch.nn.Module):
             if hidden_state is None:
                 initial_states = hidden_state
             elif isinstance(hidden_state, tuple):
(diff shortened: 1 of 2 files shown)


DeNeutoy:
Could you add a small test to demonstrate that this works? It can go either here:
https://github.com/allenai/allennlp/blob/master/allennlp/tests/modules/seq2vec_encoders/pytorch_seq2vec_wrapper_test.py
or here:
https://github.com/allenai/allennlp/blob/master/allennlp/tests/modules/encoder_base_test.py
whichever is easiest for you to add to.

Simply demonstrating that the wrapper can run forward when the input state is non-contiguous would be enough, nothing complicated!

dangitstam:
Test added to `encoder_base_test.py` here: 4399ac48579ad650ac6f21ba5cfd3e9f2e23593e

PR #1494, opened 2018-07-16
Add IOB1 support to SpanBasedF1Metric

nelson-liu:
This PR adds IOB1 support to the SpanBasedF1Metric. This was a bit tricky to write, so I'll add some details about my thought process below (might help in review).

So we've got 3 tags (B, I, O), meaning that there are 3^2 = 9 possible transitions between tags. We also have to take into account whether the current conll tag differs from the previous conll tag, bringing up the total to 9 * 2 = 18.

Here are the expected outcomes for each of the cases:

screen shot 2018-07-16 at 1 40 27 pm

Here's where I believe they're each accounted for in the code (based on which branch they fall into of the main if-elif-else)

screen shot 2018-07-16 at 1 46 18 pm

In particular, the else could also be written as `bio_tag == "I" and conll_tag == active_conll_tag`, and nothing would change (any and all combinations would still fall into one of these 3 categories).

### DIFF
--- a/allennlp/data/dataset_readers/dataset_utils/span_utils.py
+++ b/allennlp/data/dataset_readers/dataset_utils/span_utils.py
@@ -159,7 +159,7 @@ def iob1_tags_to_spans(tag_sequence: List[str],
         Note that the label `does not` contain any BIO tag prefixes.
     """
     classes_to_ignore = classes_to_ignore or []
-    spans = set()
+    spans: Set[Tuple[str, Tuple[int, int]]] = set()
     span_start = 0
     span_end = 0
     active_conll_tag = None
--- a/allennlp/data/dataset_readers/dataset_utils/span_utils.py
+++ b/allennlp/data/dataset_readers/dataset_utils/span_utils.py
@@ -165,15 +165,15 @@ def iob1_tags_to_spans(tag_sequence: List[str],
     active_conll_tag = None
     for index, string_tag in enumerate(tag_sequence):
         prev_string_tag = tag_sequence[index-1] if index - 1 > -1 else None
-        prev_bio_tag = prev_string_tag[0]
-        prev_conll_tag = prev_string_tag[2:]
+        prev_bio_tag = prev_string_tag[0] if prev_string_tag is not None else None
+        prev_conll_tag = prev_string_tag[2:] if prev_string_tag is not None else None
 
         curr_bio_tag = string_tag[0]
         curr_conll_tag = string_tag[2:]
 
         if curr_bio_tag not in ["B", "I", "O"]:
             raise InvalidTagSequence(tag_sequence)
-        if curr_bio_tag == "O" or conll_tag in classes_to_ignore:
+        if curr_bio_tag == "O" or curr_conll_tag in classes_to_ignore:
             # The span has ended.
             if active_conll_tag is not None:
                 spans.add((active_conll_tag, (span_start, span_end)))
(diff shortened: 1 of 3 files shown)


allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -136,6 +136,80 @@ def bio_tags_to_spans(tag_sequence: List[str],
...
+        The typed, extracted spans from the sequence, in the format (label, (span_start, span_end)).
+        Note that the label `does not` contain any BIO tag prefixes.
+    """
+    classes_to_ignore = classes_to_ignore or []
+    spans: Set[Tuple[str, Tuple[int, int]]] = set()
+    span_start = 0
+    span_end = 0
+    active_conll_tag = None
+    for index, string_tag in enumerate(tag_sequence):
+        prev_string_tag = tag_sequence[index-1] if index - 1 > -1 else None
```
DeNeutoy: Instead of having these `if ... else` guards, just pull these definitions out of the loop, set them to `None` and update them at the end of the loop

allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -136,6 +136,80 @@ def bio_tags_to_spans(tag_sequence: List[str],
...
+            # bio_tag == "I" and curr_conll_tag == active_conll_tag
+            # We're continuing a span.
+            span_end += 1
+    # Last token might have been a part of a valid span.
+    if active_conll_tag is not None:
+        spans.add((active_conll_tag, (span_start, span_end)))
+    return list(spans)
+
+
+def _iob1_start_of_chunk(prev_bio_tag, prev_conll_tag, curr_bio_tag, curr_conll_tag):
```
DeNeutoy: Types on arguments.

allennlp/tests/data/dataset_readers/dataset_utils/span_utils_test.py
```diff
@@ -45,6 +45,39 @@ def test_bio_tags_to_spans_ignores_specified_tags(self):
         spans = span_utils.bio_tags_to_spans(tag_sequence, ["ARG1", "V"])
         assert set(spans) == {("ARG2", (6, 7)), ("ARG2", (9, 9))}
 
+    def test_iob1_tags_to_spans_extracts_correct_spans_without_labels(self):
+        tag_sequence = ["I", "B", "I", "O", "B", "I", "B", "B"]
+        spans = span_utils.iob1_tags_to_spans(tag_sequence)
+        assert set(spans) == {("", (0, 0)), ("", (1, 2)), ("", (4, 5)), ("", (6, 6)), ("", (7, 7))}
```
DeNeutoy: I'm not a huge fan of empty strings as `None` values - Do things still work in metrics/CRF if we make them `None` explicitly?

nelson-liu: I made them empty strings because the final output metrics would be stored under the keys `B-`, `I-`, `O-`, etc., which feels more natural to me than having them under `B-None`, `I-None`, `O-None`.
alternatively, we can special case this `None` case and store the keys as `B`, `I`, `O`...do you think that's preferable?

nelson-liu: but i also haven't looked at the CRF model to see if it breaks with unlabeled tags...I think the only thing that might be affected is the `allowed_transitions` function?

nelson-liu: I'm merging this for now, but happy to change this in a future PR if you want it.

PR #1499, opened 2018-07-16
[for discussion] change token_to_indices -> tokens_to_indices (in preparation for byte pair encoding)

joelgrus:
I replaced

```
TokenIndexer.token_to_indices(self, token: Token, vocabulary: Vocabulary) -> TokenType:
```

with

```
TokenIndexer.tokens_to_indices(self, tokens: List[Token], vocabulary: Vocabulary, index_name: str) -> Dict[str, List[TokenType]]:
```

this addresses two things:

(1) the indices are no longer determined on a per-token basis; instead they can be determined by multiple tokens
(2) a single token indexer can now return multiple sets of indices

initially I was trying to leave things backward compatible, but eventually I decided this would be cleaner. this is a breaking change, but only for people who are writing their own token indexers or fields, which is probably not too many people?

### DIFF
--- a/allennlp/data/fields/knowledge_graph_field.py
+++ b/allennlp/data/fields/knowledge_graph_field.py
@@ -2,6 +2,7 @@
 ``KnowledgeGraphField`` is a ``Field`` which stores a knowledge graph representation.
 """
 from typing import Callable, Dict, List, Set
+from collections import defaultdict
 
 import editdistance
 from overrides import overrides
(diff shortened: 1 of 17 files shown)


DeNeutoy:
Question: Why do we need to extend the abstraction to Lists of tokens? Seems like we only need:

```
TokenIndexer.token_to_indices(self, token: Token, vocabulary: Vocabulary) -> Dict[str, TokenType]:
```
to handle BPE.

Not necessarily a problem, I just can't think of a scenario where you want to index a whole sentence at once and I think it would be more work.

joelgrus:
I thought that was one of mattp's requirements, but I could have misunderstood

matt-gardner:
Yeah, I know @matt-peters said he needed this, and I remember agreeing with his motivation, but I can't find where he said that now.  Matt, can you remind me where that was?

DeNeutoy:
Nevermind, we do need sentence level indexing to get the byte-pair offsets.

PR #1506, opened 2018-07-18
Text field updates to support multiple arrays in TokenIndexer

matt-peters:
* change the API for `TokenIndexer.pad_token_sequence` to avoid `Union` types
* Handle padding in case where one of the `TokenIndexer` returns multiple arrays (need to pad each sequence separately)
* add an option to `get_text_field_mask` that allows to specify the mask directly instead of computing it.

### DIFF
--- a/allennlp/data/fields/text_field.py
+++ b/allennlp/data/fields/text_field.py
@@ -2,7 +2,7 @@
 A ``TextField`` represents a string of text, the kind that you might want to represent with
 standard word vectors, or pass through an LSTM.
 """
-from typing import Dict, List, Optional, Any
+from typing import Dict, List, Optional
 import textwrap
 
 from overrides import overrides
--- a/allennlp/data/fields/text_field.py
+++ b/allennlp/data/fields/text_field.py
@@ -37,7 +37,8 @@ class TextField(SequenceField[Dict[str, torch.Tensor]]):
     def __init__(self, tokens: List[Token], token_indexers: Dict[str, TokenIndexer]) -> None:
         self.tokens = tokens
         self._token_indexers = token_indexers
-        self._indexed_tokens: Optional[Dict[str, Any]] = None
+        self._indexed_tokens: Optional[Dict[str, TokenType]] = None
+        self._indexer_name_to_indexed_token: Optional[Dict[str, str]] = None
 
         if not all([isinstance(x, (Token, SpacyToken)) for x in tokens]):
             raise ConfigurationError("TextFields must be passed Tokens. "
@@ -51,17 +52,14 @@ class TextField(SequenceField[Dict[str, torch.Tensor]]):
 
     @overrides
     def index(self, vocab: Vocabulary):
-        token_arrays: Dict[str, Any] = {}
+        token_arrays: Dict[str, TokenType] = {}
+        indexer_name_to_indexed_token: Dict[str, str] = {}
         for indexer_name, indexer in self._token_indexers.items():
             token_indices = indexer.tokens_to_indices(self.tokens, vocab, indexer_name)
-            if len(token_indices) == 1:
-                # just a single key, flatten the name space.
-                token_arrays.update(token_indices)
-            else:
-                # multiple keys, keep the hierarchy.
-                token_arrays[indexer_name] = token_indices
-
+            token_arrays.update(token_indices)
+            indexer_name_to_indexed_token[indexer_name] = list(token_indices.keys())
         self._indexed_tokens = token_arrays
+        self._indexer_name_to_indexed_token = indexer_name_to_indexed_token
 
     @overrides
     def get_padding_lengths(self) -> Dict[str, int]:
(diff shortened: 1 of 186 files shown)


allennlp/data/fields/text_field.py
```diff
@@ -37,7 +37,7 @@ class TextField(SequenceField[Dict[str, torch.Tensor]]):
     def __init__(self, tokens: List[Token], token_indexers: Dict[str, TokenIndexer]) -> None:
         self.tokens = tokens
         self._token_indexers = token_indexers
-        self._indexed_tokens: Optional[Dict[str, TokenList]] = None
+        self._indexed_tokens: Optional[Dict[str, Any]] = None
```
joelgrus: so I was actually imagining something like where a BPE indexer would generate two top-level arrays

`f"{indexer_name}-byte-pairs"` and `f"{indexer_name}-offsets"`

rather than a dict of them. that's clunkier in some ways but again it avoids the `isinstance` checks and having to have two sorts of logic.

thoughts?

matt-peters: I'm not sure I understand?  The token indexer is returning a dict of lists following the type hint, see the test stub `DictReturningTokenIndexer` in `tests/data/fields/text_field_test.py`.

joelgrus: right, I was thinking that `TextField.index` would *merge* the returned dicts, producing a one-level dict of the indices.

that is, if one indexer returned `{'byte-pairs': [..], 'offsets': [...]}` and another `{'tokens': [...]}`, it would produce a dict with 3 keys: byte-pairs, offsets, and tokens

whereas this change would make it return a dict with two keys, one of which held some indices and one of which held another dict?

DeNeutoy: I agree with Joel here - this will work seamlessly with the changes you've already made to the `TextFieldEmbedder` which allow mis-matching keys, whilst reducing the complexity in the indexing code. The case that a `TokenIndexer` generates multiple arrays is rare enough that if you want to do this, you'll just have to use the `indexer_to_embedder_map` argument in the `TextFieldEmbedder` to map the arrays to the embedder. 

The only downside with creating a flat map is that the 1-1 map between embedders and indexer names is now not enforceable - we should add some documentation that the key passed to `tokens_to_indices` really should be used as a key in isolation, even if it is then used to create auxilary keys for other return arrays.

e.g if you have a byte_pair_embedder called `bytes`, we really need the returned array to have exactly `bytes` as a key: `{bytes: [....],  bytes-offsets: [....]}`, but we can't actually enforce this if users were to implement their own indexers which return multiple arrays.

I think this is ok though, because in reality I don't imagine that there are many people implementing new indexers and if they do, they know what they are doing.

allennlp/data/fields/text_field.py
```diff
@@ -83,11 +106,20 @@ def get_padding_lengths(self) -> Dict[str, int]:
...
-        padding_lengths = {'num_tokens': len(self._indexed_tokens[any_indexed_token_key])}
+
+        # Get the padding lengths for sequence lengths.
+        if len(set(indexer_sequence_lengths.values())) == 1:
+            # This is the default case where all indexers return the same length.
+            padding_lengths = {'num_tokens': list(indexer_sequence_lengths.values())[0]}
+        else:
+            # The indexers return different lengths.
+            padding_lengths = {
+                    'num_tokens' + '-_-' + key: val for key, val in indexer_sequence_lengths.items()
```
DeNeutoy: Could you explain a bit why this `-_-` is necessary? I don't exactly understand... Is it because we are returning a flat dictionary of lengths, but there could be multiple arrays  returned which have the same keys but different values?

allennlp/data/fields/text_field.py
matt-peters: We need to maintain the mapping between the token indexers in `self._token_indexers` and the indexed tokens in `self._indexed_tokens` to call `indexer.get_padding_lengths` (in `get_padding_lengths`) and `indexer.pad_token_sequence` in `as_tensor`).

If we flatten the `self._indexed_tokens` dict as suggested then we need to introduce another dict to maintain this mapping.  An alternate implementation I've chosen encodes this as the type of `self._indexed_tokens[indexer_name]`.

matt-peters: Also -- these are just the details of an internal private attribute `_indexed_tokens` -- @markn we still need to provide a `indexer_to_embedder_map` in `TextFieldEmbedder` independent of how this is implemented.

allennlp/data/fields/text_field.py
matt-peters: Since we flatten the dict for all token indexers, then add additional keys, I inserted this here to avoid potentially clobbering keys that get returned from one of the other token indexers.

allennlp/data/fields/text_field.py
DeNeutoy: Right - I was considering the case where we also return tensors in the nested format. Maintaining another mapping seems cleaner to me if it avoids the type checks.

matt-peters:
OK, I modified the details of the private attribute `self._indexed_tokens` in `TextField` to move the logic previously associated with the nested structure to another place in the code and into another data structure.

This is ready for another look.

allennlp/data/fields/text_field.py
```diff
@@ -114,17 +145,18 @@ def as_tensor(self,
...
 
     @overrides
-    def empty_field(self):
+    def empty_field(self, vocab: Vocabulary):
         # pylint: disable=protected-access
         text_field = TextField([], self._token_indexers)
-        # This needs to be a dict of empty lists for each token_indexer,
-        # for padding reasons in ListField.
-        text_field._indexed_tokens = {name: [] for name in self._token_indexers.keys()}
+        text_field.index(vocab)
```
DeNeutoy: This is a bit of a kludge to pass the vocabulary as an argument just so we can index the text field here. The reason you need to do this is because you don't know how many empty names to add, right? Another possible way to do this would be to add a `get_keys(index_name: str) -> List[str]` method to the token indexers, so that you could do:
```
array_keys = [x for name, indexer in self._token_indexers.items() for x in indexer.get_keys(name)]
text_field._indexed_tokens = {name: [] for name in array_keys}
```

The default implementation for all token_indexers could return the name directly, and could be overridden for an indexer which returns multiple arrays. Can you see anything wrong with that? It would remove the necessity to pass a vocab object to all of the empty_fields, which is nice.

matt-peters: How about we just pass the vocab to the token indexer constructor?  Then we can eliminate it from all the various methods (seems we pass it in several different places that could all be eliminated).  Is there a reason this won't work?

joelgrus: another option (assuming that you'd index a text field before ever calling empty_field) would be for `index(self, vocab)` to hang onto the vocab (presumably you wouldn't index the same text field with multiple vocabs), and then empty_field could just use that?

matt-peters: I don't think we can make the assumption that  `empty_field` is always called after` index`?

matt-peters: Refactored the code to remove `vocab` as an argument to `empty_field` and added `get_keys` to the interface for `TokenIndexer`.

PR #1512, opened 2018-07-19
Add output-file option in evaluate to save the computed metrics

HarshTrivedi:
1. Add output-file option in evaluate to save the computed metrics (similar to `predict`)
2. Add missing --weights-file in evaluate docstring.

While I am on this, is there any reason for `--evaluation-data-file` to be optional as noted in the docstring? Also, I don't understand the purpose of making `add_mutually_exclusive_group` and adding only one argument in it. Would it be better to make it required positional argument similar to `input_file` argument in predict? If usage of `predict` and `evaluate` is similar, its easy to remember...


HarshTrivedi:
Ok, but ... trying `test_evaluate_from_args` without `--evaluation-data-file` gives 'pytest evaluate: error: one of the arguments --evaluation-data-file is required\n'. 

Its actually required (`evaluation_data_file = subparser.add_mutually_exclusive_group(required=True)`), but docstring suggest otherwise. Since, it's required, might it be better to have the usage of `predict` and `evaluate` consistent (1st and 2nd argument being model path and data input file respectively)?

DeNeutoy:
Oh, right - yep, feel free to unify them.

HarshTrivedi:
This should be good to go now.

HarshTrivedi:
@DeNeutoy Anything else needed here?

schmmd:
Before this PR, `--evaluation-data-file` was required.  It was just a named argument rather a positional one.

HarshTrivedi:
yeah .. but docstring said otherwise then. It's discussed above that it was in fact required. And since it's anyway required, it made sense to make the arguments (model file, test input file) of `evaluate` and `predict` same since they are similar in their utility.

schmmd:
The docstring shows that it's required.  Optional parameters are in brackets.

```
    usage: allennlp [command] evaluate [-h] --evaluation-data-file	    usage: allennlp evaluate [-h] [--output-file OUTPUT_FILE]
                                            EVALUATION_DATA_FILE	                             [--weights-file WEIGHTS_FILE]
                                            [--cuda-device CUDA_DEVICE]	                             [--cuda-device CUDA_DEVICE] [-o OVERRIDES]
                                            [-o OVERRIDES]	                             [--include-package INCLUDE_PACKAGE]
                                            [--include-package INCLUDE_PACKAGE]	                             archive_file input_file
                                            archive_file
```

I like the new syntax, but this is a breaking change between versions of AllenNLP.  I need to update the scripts on allennlp.org/models, for example, and they won't work for people with different versions of AllenNLP.

HarshTrivedi:
Yeah that's right. But I was talking about the section below that. That section had incorrect docstring. Check that `evaluation-data-file` was enlisted below `optional arguments` before the commit.

```
optional arguments:
    --evaluation-data-file EVALUATION_DATA_FILE
                            path to the file containing the evaluation data
   ...
```

About breaking change, that was expected .. but sorry about that ... : (

schmmd:
@HarshTrivedi ah thank you--I understand now!

PR #1513, opened 2018-07-20
Remove extra .params

### LINKED ISSUE #1447: Add a .get() method to Params
Getting a parameter from a `Params` object without popping it is currently annoying and requires accessing the inner `.params` object that should be private.  We should just add a `.get()` method that works.

matt-gardner:
Fixes #1447

I knew I saw this somewhere...


matt-gardner:
Also, any idea why the SRL sniff test took 20 minutes here?  Is that typical these days?

joelgrus:
looking at the recent teamcity builds, it looks like a blip, most of them finish in ~10 minutes end to end

PR #1514, opened 2018-07-21
Initial event2mind reimplementation

brendan-ai2:
- Forks SimpleSeq2Seq.
- New metric SequenceAccuracy.
- Only computes intent for now.

### DIFF
--- a/allennlp/models/event2mind.py
+++ b/allennlp/models/event2mind.py
@@ -1,0 +1,348 @@
+from typing import Dict
+
+import numpy
+from overrides import overrides
+
+import torch
+from torch.nn.modules.rnn import GRUCell, LSTMCell
+from torch.nn.modules.linear import Linear
+from torch import nn
+import torch.nn.functional as F
+
+from allennlp.common.util import START_SYMBOL, END_SYMBOL
+from allennlp.data.vocabulary import Vocabulary
+from allennlp.modules import TextFieldEmbedder, Seq2SeqEncoder
+from allennlp.modules.similarity_functions import SimilarityFunction
+from allennlp.modules.token_embedders import Embedding
+from allennlp.models.model import Model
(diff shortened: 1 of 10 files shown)


brendan-ai2:
WIP, not ready for a proper review quite yet.

allennlp/models/event2mind.py
```diff
@@ -0,0 +1,348 @@
...
+            reshaped_predicted_classes = predicted_classes.reshape(batch_size, k * k)
+            restricted_beam_log_probs, restricted_beam_indices = reshaped_top_log_probabilities.topk(k)
+            # TODO(brendanr): Something about this is weird. restricted_predicted_classes == restricted_beam_indices???
+            restricted_predicted_classes = reshaped_predicted_classes.gather(1, restricted_beam_indices)
+
+            log_probabilities.append(restricted_beam_log_probs)
+            predictions.append(restricted_predicted_classes)
+            backpointer = restricted_beam_indices / k
+            backpointers.append(backpointer)
+            expanded_backpointer = backpointer.unsqueeze(2).expand(batch_size, k, self._decoder_output_dim)
```
brendan-ai2: @DeNeutoy, this is the are I was telling you about. Essentially I'm using these backpointers to ensure that the hidden layer (for the next iteration) is pruned to only include entries corresponding to items still in the beam. It seems like expanding the backpointers to the `_decoder_output_dim` is enough to make the `gather` work. Feel free to poke around, but I'm going to clean this up before requesting a real review. :) Thanks!

DeNeutoy: If this becomes a sticking point, i'd just loop over the batch to get it out of the way. It's always hard to tell what the bottleneck will be in these models - I think in this case, it might not be the beam search, because by nature of batching together the topk states, it's already being batched reasonably efficiently.

Issue #1476, opened 2018-07-11
Output config.json in the same order to the input config file


handsomezebra:
**Is your feature request related to a problem? Please describe.**
The config.json in the serialization directory looks quite different to the input config file. It is not easy to figure out which input config file was used for one experiment, especially when there are a lot of those experiments. 

**Describe the solution you'd like**
While outputing config.json, can we use the same order of keys as the input config file? This can be accomplished by using an OrderedDict to load the original json.

DeNeutoy:
Sounds like a good idea! Do you want to send a PR?

handsomezebra:
I tried but _jsonnet already changed the order before json sees the content. Didn't find a way to do this before _jsonnet supports this.

matt-gardner:
We could just copy the original config file into the serialization directory, instead of dumping parameters?  Why not do that?  Hmm, because of overrides...  Yeah, this might need a pretty complicated solution to get right.

joelgrus:
if the goal is to maintain a correspondence between the original config file and the serialized results, wouldn't it be better to just do that directly? (having the keys in the same order still leaves things you have to check). you could just dump some metadata into the archive like "original_config_path", although in some cases you might not want that in there

matt-gardner:
Hmm, ok, yeah, we could just copy over the original config file with a separate name, if that's the goal.  I'm not sure how useful it is, though, and it could be mislead if you ran stuff with `--overrides`.  Should we dump that too?

handsomezebra:
I think it's not a big problem for now as long as all neccesary information to reproduce the experiments is serialized. In the future it will be fantastic if we have some kind of management console to explore and compare experiments results as below. Thanks for the great tool!

![image](https://user-images.githubusercontent.com/5323637/42610615-3a5b5932-8547-11e8-9f45-f81fac7d51d1.png)

HarshTrivedi:
@matt-gardner  I am quite curious above having beaker accessible. By any chance the plan is to release it in this summer (next 1-2 months)?

matt-gardner:
@HarshTrivedi, it's a different team at AI2 that's building beaker, and I can't speak for their plans.  I know it's their goal to release the tool, but I don't know timelines.

PR #1517, opened 2018-07-22
Add to_file method in Params and default preference ordering.

HarshTrivedi:
This PR is in reference to (#1476). Besides that use case: I am generating bunch of experiment configs based on certain combinations of settings I want to explore. Having consistent ordering of keys generated across these output config files would be good for human parsing. Eg. Seeing `vocabulary` key to top, `train_data_path` at bottom and `validation_data_path` somewhere in middle seems quite off place . So this PR adds `to_file` counter part of `from_file` in Params, which saves the json in according to passed or default partial preference orderings. 

The default orderings are as: `dataset_reader > iterator > model > train_data_path > validation_data_path > test_data_path > trainer > vocabulary` and `type` key will always be on the top of its companions. This is the usual order in which we typically organize our experiment configs. Rest of the keys (any level) will be ordered later and alphabetically. 

Note that having order maintained from input config file isn't straightforward as discussed in (#1476) due to: (1) jsonnet changes the content before json sees it. (2) overriding dicts can any way make things ambiguous. 

This PR doesn't intent to maintain ordering of input file. But rather enforce a default partial ordering and resort to alphabetical order for remaining keys.

### DIFF
--- a/allennlp/common/params.py
+++ b/allennlp/common/params.py
@@ -354,6 +354,49 @@ class Params(MutableMapping):
 
         return Params(param_dict)
 
+    def to_file(self, params_file: str):
+        with open(params_file) as file:
+            json.dump(file, self.as_ordered_dict(), indent=4)
+
+    def as_ordered_dict(self, preference_orders: List[List[str]] = None):
+        """
+        Returns Ordered Dict of Params from list of partial order preferences.
+
+        Parameters
+        ----------
+        preference_orders: List[List[str]], optional
+            ``preference_orders`` is list of partial preference orders. ["A", "B", "C"] means
+            "A" > "B" > "C". For multiple preference_orders first will be considered first.
+            Keys not found, will have last but alphabetical preference. Default Preferences:
+            ``[["dataset_reader", "iterator", "model", "train_data_path", "validation_data_path",
+                "test_data_path", "trainer", "vocabulary", "seeds"], ["type"]]``
+        """
+        params_dict = self.as_dict(quiet=True)
+        if not preference_orders:
+            preference_orders = []
+            preference_orders.append(["dataset_reader", "iterator", "model",
+                                      "train_data_path", "validation_data_path", "test_data_path",
+                                      "trainer", "vocabulary", "seeds"])
+            preference_orders.append(["type"])
+
+        def preference_ranks(key):
+            '''returns list of all preference scores'''
+            return [order.index(key) if key in order else len(order)
+                    for order in preference_orders]
+
+        def order_func(key):
+            '''returns list of all preference scores with default alphabetic score'''
+            return preference_ranks(key) + [key]
+
+        def _order_dict(dictionary, order_func):
+            '''Recursively orders dictionary according to scoring order_func'''
+            result = {}
+            for key, val in sorted(dictionary.items(), key=lambda item: order_func(item[0])):
+                result[key] = _order_dict(val, order_func) if isinstance(val, dict) else val
+            return result
+
+        return _order_dict(params_dict, order_func)
+
 
 def pop_choice(params: Dict[str, Any],
                key: str,
--- a/allennlp/common/params.py
+++ b/allennlp/common/params.py
@@ -5,7 +5,7 @@ logging and validation.
 """
 
 from typing import Any, Dict, List
-from collections import MutableMapping
+from collections import MutableMapping, OrderedDict
 import copy
 import json
 import logging
@@ -354,11 +354,11 @@ class Params(MutableMapping):
 
         return Params(param_dict)
 
-    def to_file(self, params_file: str, preference_orders: List[List[str]] = None):
+    def to_file(self, params_file: str, preference_orders: List[List[str]] = None) -> None:
         with open(params_file, "w") as handle:
             json.dump(self.as_ordered_dict(preference_orders), handle, indent=4)
 
-    def as_ordered_dict(self, preference_orders: List[List[str]] = None):
+    def as_ordered_dict(self, preference_orders: List[List[str]] = None) -> OrderedDict:
         """
         Returns Ordered Dict of Params from list of partial order preferences.
 
@@ -379,24 +379,21 @@ class Params(MutableMapping):
                                       "trainer", "vocabulary"])
             preference_orders.append(["type"])
 
-        def preference_ranks(key):
-            '''returns list of all preference scores'''
-            return [order.index(key) if key in order else len(order)
-                    for order in preference_orders]
-
         def order_func(key):
-            '''returns list of all preference scores with default alphabetic score'''
-            return preference_ranks(key) + [key]
+            # Makes a tuple to use for ordering.  The tuple is an index into each of the `preference_orders`,
+            # followed by the key itself.  This gives us integer sorting if you have a key in one of the
+            # `preference_orders`, followed by alphabetical ordering if not.
+            order_tuple = [order.index(key) if key in order else len(order) for order in preference_orders]
+            return order_tuple + [key]
 
-        def _order_dict(dictionary, order_func):
-            '''Recursively orders dictionary according to scoring order_func'''
-            result = {}
+        def order_dict(dictionary, order_func):
+            # Recursively orders dictionary according to scoring order_func
+            result = OrderedDict()
             for key, val in sorted(dictionary.items(), key=lambda item: order_func(item[0])):
-                result[key] = _order_dict(val, order_func) if isinstance(val, dict) else val
+                result[key] = order_dict(val, order_func) if isinstance(val, dict) else val
             return result
 
-        return _order_dict(params_dict, order_func)
-
+        return order_dict(params_dict, order_func)
 
 def pop_choice(params: Dict[str, Any],
                key: str,
--- a/allennlp/common/params.py
+++ b/allennlp/common/params.py
@@ -369,14 +369,14 @@ class Params(MutableMapping):
             "A" > "B" > "C". For multiple preference_orders first will be considered first.
             Keys not found, will have last but alphabetical preference. Default Preferences:
             ``[["dataset_reader", "iterator", "model", "train_data_path", "validation_data_path",
-                "test_data_path", "trainer", "vocabulary", "seeds"], ["type"]]``
+            "test_data_path", "trainer", "vocabulary"], ["type"]]``
         """
         params_dict = self.as_dict(quiet=True)
         if not preference_orders:
             preference_orders = []
             preference_orders.append(["dataset_reader", "iterator", "model",
                                       "train_data_path", "validation_data_path", "test_data_path",
-                                      "trainer", "vocabulary", "seeds"])
+                                      "trainer", "vocabulary"])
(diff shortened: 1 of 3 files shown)


allennlp/common/params.py
```diff
@@ -354,6 +354,49 @@ def from_file(params_file: str, params_overrides: str = "") -> 'Params':
...
+                                      "trainer", "vocabulary", "seeds"])
+            preference_orders.append(["type"])
+
+        def preference_ranks(key):
+            '''returns list of all preference scores'''
+            return [order.index(key) if key in order else len(order)
+                    for order in preference_orders]
+
+        def order_func(key):
+            '''returns list of all preference scores with default alphabetic score'''
```
matt-gardner: I think this would be easier to read if you combined these two functions into something like this:

```python
def order_func(key):
    # Makes a tuple to use for ordering.  The tuple is an index into each of the `preference_orders`,
    # followed by the key itself.  This gives us integer sorting if you have a key in one of the
    # `preference_orders`, followed by alphabetical ordering if not.
    order_tuple = [order.index(key) if key in order else len(order) for order in preference_orders]
    return order_tuple + [key]
```

Still exactly the same logic as what you have, but a bit easier on the reader because it's a single function with a good comment.

allennlp/common/params.py
```diff
@@ -354,6 +354,49 @@ def from_file(params_file: str, params_overrides: str = "") -> 'Params':
...
+
+        def _order_dict(dictionary, order_func):
+            '''Recursively orders dictionary according to scoring order_func'''
+            result = {}
+            for key, val in sorted(dictionary.items(), key=lambda item: order_func(item[0])):
+                result[key] = _order_dict(val, order_func) if isinstance(val, dict) else val
+            return result
+
+        return _order_dict(params_dict, order_func)
+
```
matt-gardner: Just one blank line here.

allennlp/common/params.py
```diff
@@ -354,6 +354,49 @@ def from_file(params_file: str, params_overrides: str = "") -> 'Params':
...
+            '''returns list of all preference scores'''
+            return [order.index(key) if key in order else len(order)
+                    for order in preference_orders]
+
+        def order_func(key):
+            '''returns list of all preference scores with default alphabetic score'''
+            return preference_ranks(key) + [key]
+
+        def _order_dict(dictionary, order_func):
+            '''Recursively orders dictionary according to scoring order_func'''
```
matt-gardner: Docstrings should use double quotes.

allennlp/common/params.py
```diff
@@ -354,6 +354,49 @@ def from_file(params_file: str, params_overrides: str = "") -> 'Params':
 
         return Params(param_dict)
 
+    def to_file(self, params_file: str, preference_orders: List[List[str]] = None):
```
matt-gardner: Return type (`-> None`).

allennlp/common/params.py
```diff
@@ -354,6 +354,49 @@ def from_file(params_file: str, params_overrides: str = "") -> 'Params':
...
+        with open(params_file, "w") as handle:
+            json.dump(self.as_ordered_dict(preference_orders), handle, indent=4)
+
+    def as_ordered_dict(self, preference_orders: List[List[str]] = None):
```
matt-gardner: Return type (`OrderedDict`).

allennlp/common/params.py
```diff
@@ -354,6 +354,49 @@ def from_file(params_file: str, params_overrides: str = "") -> 'Params':
...
+            return [order.index(key) if key in order else len(order)
+                    for order in preference_orders]
+
+        def order_func(key):
+            '''returns list of all preference scores with default alphabetic score'''
+            return preference_ranks(key) + [key]
+
+        def _order_dict(dictionary, order_func):
+            '''Recursively orders dictionary according to scoring order_func'''
+            result = {}
```
matt-gardner: You're not actually constructing an `OrderedDict` here - don't you mean `result = OrderedDict()`?

allennlp/common/params.py
```diff
@@ -354,6 +354,49 @@ def from_file(params_file: str, params_overrides: str = "") -> 'Params':
...
+        def preference_ranks(key):
+            '''returns list of all preference scores'''
+            return [order.index(key) if key in order else len(order)
+                    for order in preference_orders]
+
+        def order_func(key):
+            '''returns list of all preference scores with default alphabetic score'''
+            return preference_ranks(key) + [key]
+
+        def _order_dict(dictionary, order_func):
```
matt-gardner: This is already defined inside of a function; no need to prepend an underscore to the name.

allennlp/tests/common/params_test.py
```diff
@@ -277,3 +277,29 @@ def from_params(cls, params: Params) -> 'A':
...
         }
+
+    def test_as_ordered_dict(self):
+        # keyD > keyC > keyE; keyDA > keyDB; Next all other keys alphabetically
+        preference_orders = [["keyD", "keyC", "keyE"], ["keyDA", "keyDB"]]
+        params = Params({"keyC": "valC", "keyB": "valB", "keyA": "valA", "keyE": "valE",
+                         "keyD": {"keyDB": "valDB", "keyDA": "valDA"}})
+        ordered_params_dict = params.as_ordered_dict(preference_orders)
+        ordered_params_dict_str = str(ordered_params_dict)
+        expected_dict_str = str(({'keyD': {'keyDA': 'valDA', 'keyDB': 'valDB'},
```
matt-gardner: This test should use an `OrderedDict` also, I think.  It currently works, but only because it's relying on an implementation detail of CPython.

allennlp/tests/common/params_test.py
```diff
@@ -277,3 +277,29 @@ def from_params(cls, params: Params) -> 'A':
...
+                                  'keyC': 'valC', 'keyE': 'valE',
+                                  'keyA': 'valA', 'keyB': 'valB'}))
+        assert ordered_params_dict_str == expected_dict_str
+
+    def test_to_file(self):
+        # Test to_file works with or without preference orders
+        params_dict = {"keyA": "valA", "keyB": "valB"}
+        ordered_params_dict = {"keyB": "valB", "keyA": "valA"}
```
matt-gardner: Same here.

allennlp/common/params.py
HarshTrivedi: Sure! Earlier I had two lambdas and so two ended up with functions.. but this is better!

allennlp/common/params.py
HarshTrivedi: I assumed two blanks are for separating class (as 2 blanks where it starts). It ends at `return _order_dict(params_dict, order_func)`

allennlp/common/params.py
HarshTrivedi: Sure 👍

allennlp/common/params.py
HarshTrivedi: Sure 👍

allennlp/common/params.py
HarshTrivedi: Actually dicts are by default ordered in python 3.6 onwards. link: https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6 . So skipping OrderedDict is good enough. I checked and tests also work.

allennlp/common/params.py
HarshTrivedi: I meant `{}` only because they are ordered, but may be it would be better to be explicit?

allennlp/common/params.py
HarshTrivedi: Sure 👍

allennlp/tests/common/params_test.py
HarshTrivedi: Dicts are ordered in python and converting to str retains this order. I tried changing orders in expected dict and the test fails.

allennlp/tests/common/params_test.py
HarshTrivedi: Sure, let me know if I should explicitly use `OrderedDict` after above comments.

allennlp/common/params.py
matt-gardner: From that link: "This is considered an implementation detail in Python 3.6; you need to use OrderedDict if you want insertion ordering that's guaranteed across other implementations of Python. ... As of Python 3.7, this is no longer an implementation detail and instead becomes a language feature."

We're not on 3.7 yet.  Also, readers and writers of python code may not be aware that dictionaries are going to retain their insertion order, so it helps readability to be explicit about the ordering, especially when your function name says "ordered_dict".

matt-gardner: And you still need a type annotation either way =).

allennlp/common/params.py
matt-gardner: Sorry, you're right, I missed that `pop_choice` below wasn't a method on the class.  Github's spacing makes it hard to tell sometimes...

allennlp/common/params.py
HarshTrivedi: Sure, I will use `OrderedDict` everywhere. Thanks

allennlp/common/params.py
HarshTrivedi: Btw, annotations are missing in these nested functions also, will add them.

HarshTrivedi: Cool, if you are fine, I would rather keep it unannotated. 👍

allennlp/common/params.py
HarshTrivedi: done 👍

allennlp/common/params.py
HarshTrivedi: Sure, no problem  👍

allennlp/common/params.py
HarshTrivedi: Done 👍 Thanks, this is better.

allennlp/common/params.py
HarshTrivedi: Used `#` instead like `order_func`.

allennlp/common/params.py
HarshTrivedi: done 👍

allennlp/common/params.py
HarshTrivedi: done 👍

allennlp/common/params.py
HarshTrivedi: done 👍

allennlp/common/params.py
HarshTrivedi: done 👍

HarshTrivedi:
> The default orderings are as: dataset_reader > iterator > model > train_data_path > validation_data_path > test_data_path > trainer > vocabulary and type key will always be on the top of its companions. 

Do you have any opinion on updating or adding something else to default orderings?

allennlp/tests/common/params_test.py
HarshTrivedi: done 👍

matt-gardner:
I don't have a strong opinion on the default ordering.  What you have here looks reasonable, and if others want to change it, it's easy to submit a PR for that.

Thanks for the PR!

PR #1520, opened 2018-07-23
Show warning before ignoring key with unseparable batches from forward.

HarshTrivedi:
It's quite confusing why a key from model.forward got ignored without this warning.


DeNeutoy:
My bad 😞

HarshTrivedi:
it was just a typo ... : )

PR #1523, opened 2018-07-24
Allow server start without field-names.

HarshTrivedi:
As far as I understand, there are three ways of using simple server:

1. `field-name`s  are passed and it generates barebone html with those field names to be used in demo.
2. `static-dir` is passed then `field-name` is not required. The static html should have the required field-names as input fields.
3. Use react based front-end ([link](https://github.com/allenai/allennlp-simple-server-visualization)). Neither `static-dir`, nor `field-name` is required.

Except use-case (1), `field-name` argument should not be required. I am using (3) and I couldn't start text-in/text-out service on without passing some field-name which would in fact get ignored.

Edit1: WIP
Edit2: Seems OK now.


HarshTrivedi:
@joelgrus can you review this pr?

PR #1525, opened 2018-07-24
openai transformer LM embedder

joelgrus:
ok, I think this is ready for review, I addressed all the early feedback.

I trained the NER model using these embeddings instead of ELMo: http://beaker.allenai.org/ex/ex_sth0rsp0h152/tasks

the performance was not quite as good as I might have expected (but I'm not sure what exactly to expect), possibly it was overfitting on the training data.


allennlp/modules/token_embedders/openai_transformer_embedder.py
```diff
@@ -0,0 +1,110 @@
...
+        # These embeddings are one per byte-pair, but we want one per original _word_.
+        # Fortunately, we have the ``offsets`` which indicate the last byte-pair for
+        # each original word. Here we just choose that last byte-pair, although you
+        # could imagine doing something more sophisticated.
+        _, max_sequence_length = offsets.size()
+        _, _, embedding_dim = mixes[0].size()
+
+        # TODO(joelgrus): vectorize?
+        embeddings = [torch.zeros(batch_size, max_sequence_length, embedding_dim) for _ in mixes]
+        for i in range(batch_size):
```
DeNeutoy: You can vectorise this by doing:

```
range_vector = get_range_vector(0, batch_size).unsqueeze(1)

last_byte_pair_embeddings = embeddings[range_vector, offsets]

```
You can see an example of this [here](https://github.com/allenai/allennlp/blob/master/allennlp/models/biaffine_dependency_parser.py#L501).

This will save a lot of memory, I think.

allennlp/data/token_indexers/openai_transformer_byte_pair_indexer.py
```diff
@@ -0,0 +1,179 @@
...
+        self.decoder = {word_id: word for word, word_id in self.encoder.items()}
+
+        # Compute ranks
+        self.bpe_ranks = {pair: idx for idx, pair in enumerate(byte_pairs)}
+
+        self.cache: Dict[str, List[int]] = {}
+        self.n_ctx = n_ctx
+
+    @classmethod
+    def from_params(cls, params: Params) -> 'OpenaiTransformerBytePairIndexer':
```
DeNeutoy: Would it be easier to separate the BPE encoder part from the weights, so we can remove this custom from_params?

allennlp/modules/token_embedders/openai_transformer_embedder.py
```diff
@@ -0,0 +1,110 @@
...
+    do_layer_norm: ``bool`` (default: ``False``)
+        Whether the ``ScalarMix``es should use layer norm.
+    """
+    def __init__(self,
+                 transformer: OpenaiTransformer,
+                 num_output_representations: int = 1,
+                 do_layer_norm: bool = False) -> None:
+        super().__init__()
+
+        if num_output_representations > 1:
```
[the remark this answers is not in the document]
matt-peters: `do_layer_norm` isn't useful either, I'd just remove it and set it to `False`.

allennlp/modules/token_embedders/openai_transformer_embedder.py
```diff
@@ -0,0 +1,110 @@
...
+            raise NotImplementedError("more than 1 output representation is not implemented")
+
+        self._transformer = transformer
+        self._num_output_representations = num_output_representations
+        self._do_layer_norm = do_layer_norm
+
+        self._scalar_mixes: List[ScalarMix] = []
```
matt-peters: We only need one `ScalarMix` since `num_output_representations==1`.

Froskekongen:
It would be great if you are able to provide functionality and documentation for pretraining a language model in this PR. While it is nice to be able to recreate the experiments in the published article, it is even more interesting how pretraining on different datasets affect domain specific tasks.

allennlp/modules/token_embedders/openai_transformer_embedder.py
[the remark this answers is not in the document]
joelgrus: done

allennlp/modules/token_embedders/openai_transformer_embedder.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        # pylint: disable=arguments-differ
+        batch_size, num_timesteps = inputs.size()
+
+        # the transformer "vocab" consists of the actual vocab and the
+        # positional encodings. Here we want the count of just the former.
+        vocab_size = self._transformer.vocab_size - self._transformer.n_ctx
+
+        # Combine the inputs with positional encodings
+        batch_tensor = torch.zeros((batch_size, num_timesteps, 2), dtype=torch.long)
```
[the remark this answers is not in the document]
joelgrus: done

allennlp/modules/token_embedders/openai_transformer_embedder.py
```diff
@@ -0,0 +1,110 @@
...
+
+        # Embeddings is num_output_layers x (batch_size, num_timesteps, embedding_dim)
+        layer_activations = self._transformer(batch_tensor)
+
+        # Output of each scalar_mix is (batch_size, num_timesteps, embedding_dim)
+        mixes = [scalar_mix(layer_activations, byte_pairs_mask) for scalar_mix in self._scalar_mixes]
+
+        # These embeddings are one per byte-pair, but we want one per original _word_.
+        # Fortunately, we have the ``offsets`` which indicate the last byte-pair for
+        # each original word. Here we just choose that last byte-pair, although you
```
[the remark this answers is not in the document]
joelgrus: done (I think)

allennlp/modules/token_embedders/openai_transformer_embedder.py
joelgrus: awesome, thanks

allennlp/data/token_indexers/openai_transformer_byte_pair_indexer.py
joelgrus: you know what, I can remove it anyway, the same way I did in the other one.

(I'd prefer just having one tarfile required to run the model, rather than two)

allennlp/modules/token_embedders/openai_transformer_embedder.py
joelgrus: done

allennlp/data/token_indexers/openai_transformer_byte_pair_indexer.py
```diff
@@ -0,0 +1,193 @@
...
+from allennlp.data.vocabulary import Vocabulary
+from allennlp.data.tokenizers.token import Token
+from allennlp.data.token_indexers.token_indexer import TokenIndexer
+
+
+@TokenIndexer.register("openai_transformer_byte_pair")
+class OpenaiTransformerBytePairIndexer(TokenIndexer[int]):
+    """
+    This is unlike most of our TokenIndexers in that its
+    indexing is not based on a `Vocabulary` but on a fixed
```
DeNeutoy: Could you give a bit more of a description here about this class? It's non-obvious that it would return the ELMo - like mixture of layers, rather than just the last layer as in the original paper. Also maybe add a link to the paper or something.

qiuwei:
@joelgrus Could you provide your config file for the experiment with openai_transformer_embedder?
Thanks in advance.

qiuwei:
@joelgrus Could you provide your config file for the experiment with openai_transformer_embedder?
Thanks in advance.

qiuwei:
@joelgrus  Could you share your config file for training the NER model using openai-transformer-embedder?

Thanks in advance!

qiuwei:
@joelgrus  Could you share your config file for training the NER model using openai-transformer-embedder?

Thanks in advance!

joelgrus:
I think this should work:

https://gist.github.com/joelgrus/885ef14323ff3d6ecefb84fa59c48039

PR #1529, opened 2018-07-24
fix ud reader in case of implicit references

DeNeutoy:
Some UD annotations have automatically "fixed" implicit references, which have no head and are unattached to the the rest of the tree. They should just be ignored.

Also updates a pos tag I should have ignored for UD.


allennlp/data/dataset_readers/universal_dependencies.py
```diff
@@ -49,6 +49,10 @@ def _read(self, file_path: str):
             logger.info("Reading UD instances from conllu dataset at: %s", file_path)
 
             for annotation in  lazy_parse(conllu_file.read()):
+                # Conllu annotations can have implicit word references that have been
+                # filled in automatically and have no head, so we remove them.
```
matt-gardner: This is called ["ellipsis"](https://en.wikipedia.org/wiki/Ellipsis_(linguistics)), and it took me a bit to figure out what you meant by "implicit word references".  Here's a stab at a comment that's more intuitive to me:  "CoNLLU annotations sometimes add back in words that have been elided in the original sentence; we remove these, as we're just predicting dependencies for the original sentence."

It's not obvious to me how `x["id"] is None` maps to words that have been added, though.

allennlp/tests/data/dataset_readers/universal_dependencies_dataset_reader_test.py
```diff
@@ -46,3 +46,19 @@ def test_read_from_file(self):
         assert fields["head_tags"].labels == ['root', 'punct', 'case', 'compound', 'root', 'case',
                                               'nmod', 'flat', 'flat', 'punct']
         assert fields["head_indices"].labels == [0, 4, 4, 4, 0, 6, 4, 6, 6, 4]
+
+        # This instance tests specifically for filtering of implicit references.
```
matt-gardner: I'd use "ellipsis" here, and probably reference the annotation guidelines: http://universaldependencies.org/u/overview/specific-syntax.html#ellipsis.

allennlp/tests/data/dataset_readers/universal_dependencies_dataset_reader_test.py
```diff
@@ -46,3 +46,19 @@ def test_read_from_file(self):
...
+        # The original sentence is:
+        # "Over 300 Iraqis are reported dead and 500 [reported] wounded in Fallujah alone."
+        # But the second "reported" has no head, because it is implicitly referenced.
```
matt-gardner: It's not that it has no head, it's that it isn't actually there, it's been elided.

PR #1522, opened 2018-07-24
Add a scirpt to report broken links in all markdowns.

HarshTrivedi:
This PR adds a script to report broken links in all markdowns in the project.

```
time python scripts/verify.py --checks check-links
UnReachable Links:
Verifying with ['check-links']
LinksValidity (check):

UnReachable Links:

0.
Source: /Users/harshtrivedi/allennlp/tutorials/getting_started/using_as_a_library_pt2.md
Name: on GitHub
Link: https://github.com/allenai/allennlp-as-a-library-example/tree/0.5.1

1.
Source: /Users/harshtrivedi/allennlp/tutorials/getting_started/using_as_a_library_pt2.md
Name: `static_html`
Link: https://github.com/allenai/allennlp-as-a-library-example/tree/0.5.1/static_html

2.
Source: /Users/harshtrivedi/allennlp/tutorials/getting_started/using_as_a_library_pt2.md
Name: its own file
Link: https://github.com/allenai/allennlp-as-a-library-example/blob/0.5.1/static_html/demo.css

3.
Source: /Users/harshtrivedi/allennlp/tutorials/getting_started/using_as_a_library_pt2.md
Name: add a `script` tag to load chart.js
Link: https://github.com/allenai/allennlp-as-a-library-example/blob/0.5.1/static_html/index.html#L47

4.
Source: /Users/harshtrivedi/allennlp/tutorials/getting_started/using_as_a_library_pt2.md
Name: `my_library/__init__.py`
Link: https://github.com/allenai/allennlp-as-a-library-example/blob/0.5.1/my_library/__init__.py

5.
Source: /Users/harshtrivedi/allennlp/tutorials/getting_started/using_as_a_library_pt2.md
Name: can be very simple
Link: https://github.com/allenai/allennlp-as-a-library-example/blob/0.5.1/my_library/predictors/paper_classifier_predictor.py

6.
Source: /Users/harshtrivedi/allennlp/tutorials/getting_started/using_as_a_library_pt2.md
Name: placing that inside our `output` div
Link: https://github.com/allenai/allennlp-as-a-library-example/blob/0.5.1/static_html/index.html#L61

7.
Source: /Users/harshtrivedi/allennlp/tutorials/getting_started/using_pretrained_models.md
Name: Named Entity Recognition model
Link: http://allennlp.org/models/#named-entity-recognition

8.
Source: /Users/harshtrivedi/allennlp/tutorials/getting_started/using_as_a_library_pt2.md
Name: as a test fixture
[...]

PR #1533, opened 2018-07-26
Fix the reported broken links.

HarshTrivedi:
Fixes the unreachable links reported by `check-links` (#1522). Please consider merging this after `check-links` has been added to CI so as to make sure CI fails without it.


DeNeutoy:
Thanks - I just added it, let's see if that worked

Issue #1536, opened 2018-07-26
Add some kind of `validate_config` command


matt-gardner:
It's pretty frustrating (especially to a new user) to have to go through data loading to figure out that you have an error in your model configuration.  And to fix that error and discover that there's another one, but it only told you one at a time.  It'd be a lot nicer to have some way to validate the configuration file before doing anything else.

It seems like @joelgrus's work on the configuration wizard should make this relatively straightforward.  We could add a `validate_train_config` command that just checked that your config was specified correctly, or, if it's fast enough, we could just always run that as the first thing in the `train` command, giving an error message with all of the problems, not just one at a time.

joelgrus:
it should be fast enough to just check at the beginning of training, let me think about it

HarshTrivedi:
Whatever pre-check `validation` we apply before `train`, we might as well want to apply that before `dry-run` as well. Since `dry-run` does all the steps the exact same ways: vocab generation, indexing, model loading (Only calling trainer is left out), it good for checking if things are well specified and configs are as expected before actually training. This kind of pre-checker would make `dry-run` utility better.

schmmd:
Fixed by #1540

matt-gardner:
#1540 was closed, not merged.  Re-opening this.

matt-gardner:
Closing as a duplicate of #3649.

PR #1540, opened 2018-07-27
configuration validator

joelgrus:
fix for #1536 

uses many of the same hacks as the config explorer, which means it's sort of fragile, but 🤷‍♂️


joelgrus:
don't look at this yet, it's severely broken

schmmd:
Closing--as this was a prototype that has some serious issues (and doesn't have any active work).

Issue #1378, opened 2018-06-14
eliminate `requirements_test.txt` + add tests to setup.py


joelgrus:
yesterday's bad release was due (in large part) to the distinction between `requirements.txt` and `requirements_test`.txt.

briefly, I added a new feature that required a "new" dependency, tried to pip install it, saw that it was already there, and so assumed it was already transitively included in our requirements. but it turns out it was actually one of the *test* requirements (which I have installed, but which don't get pip installed). TeamCity didn't catch this either, because it also installs test requirements!

I failed to catch it manually at the "test pypi test install" step, so that's bad on me, but our release process shouldn't rely on that sort of manual verification (and anyway it could have been a much more subtle bug).

This is admittedly an edge case, but (I hit it and so) I think it speaks to having *ONE INSTALL PATH*. If our CI system isn't testing the same "environment" that's going into our pip package, that's very bad.

(Before we do this we should eliminate the Jupyter dependency from requirements_test.txt)

along similar lines, I think we should distribute our tests as part of the pip package, so that we can TEST TEST our pip install

joelgrus:
in an ideal world we should have a CI step that spins up our own devpi https://devpi.net/docs/devpi/devpi/latest/%2Bd/index.html server, pushes to it, creates a new environment, installs from it, and runs the tests

schmmd:
It's easy to add such a CI step, but my concern is it would double our testing time.

I'm all for eliminating `requirements_test.txt`.  I agree we should remove [the `jupyter` dependency](https://github.com/allenai/allennlp/blob/master/requirements_test.txt#L7) before we do.

joelgrus:
I was thinking we would add it only to the "commit to master" step, not the "pull request" step

schmmd:
@joelgrus well, I absolutely plan to add that step because I'd like to publish a docker image that does a pip install of AllenNLP.  I think this would be accomplished by https://github.com/allenai/allennlp/issues/1324 (although I was pretty sparse on details in that issue).

DeNeutoy:
FYI seems like Infrastructure have already set up a devpi server: https://github.com/allenai/pipette/issues/7

schmmd:
@DeNeutoy Cristian set this up.

schmmd:
Ah, nm, just read the ticket.

schmmd:
@DeNeutoy @schmmd and @joelgrus just agreed that we should just move Jupyter to `requirements.txt`.

[...]

PR #1541, opened 2018-07-27
Remove requirements_test.txt

schmmd:
Partially addresses https://github.com/allenai/allennlp/issues/1378


schmmd:
This also remove `scripts/install_requirements.sh`.  We no longer need to manually install spacy due to https://github.com/allenai/allennlp/pull/933 and I couldn't see where we're using `punkt`.  I suspect, however, that we still need `punkt` based on https://github.com/allenai/allennlp/pull/44--so the build will probably fail and I'll revert the final commit.

scripts/install_requirements.sh
```diff
@@ -1,10 +0,0 @@
-#!/bin/bash
-
-pip install -r requirements.txt
```
DeNeutoy: Don't remove this script please - it's useful when installing into a virtual environment (e.g not docker). The `spacy download en_core_web_sm` is still required and I think the `punkt` data package from NLTK is used in the Stemmer still too.

DeNeutoy: Oh right, you don't want to have to download the spacy model. Personally I don't like this approach (e.g it makes what packages are actually required kind of opaque if we download them at runtime) but if others think it is fine then sure.

schmmd: I'm happy to leave the script and add `spacy download en_core_web_sm`--but I don't want to add `punkt` back unless I can identify where it's being used.  It's not being used in any of our test cases.  I wanted to remove both of those in this PR so I could run the tests in a clean environment.

schmmd: It looks like we "use it" in the unused and untested:

```
$ git grep NltkWordSplitter
allennlp/data/tokenizers/word_splitter.py:class NltkWordSplitter(WordSplitter):
```

DeNeutoy: It's used here: https://github.com/allenai/allennlp/blob/master/allennlp/data/tokenizers/word_splitter.py#L141

If you want to remove this tokeniser and remove this data download that's fine with me.

Dockerfile
```diff
@@ -41,7 +41,6 @@ RUN apt-get install -y -t jessie-backports openjdk-8-jdk
 # Copy select files needed for installing requirements.
 # We only copy what we need here so small changes to the repository does not trigger re-installation of the requirements.
 COPY requirements.txt .
-COPY requirements_test.txt .
 COPY scripts/install_requirements.sh scripts/install_requirements.sh
 RUN INSTALL_TEST_REQUIREMENTS="true" ./scripts/install_requirements.sh
```
DeNeutoy: This environment variable can be removed now.

scripts/install_requirements.sh
schmmd: I will in a separate PR.

PR #1542, opened 2018-07-29
Add command to generate / manage sub-experiments from root experiment config.

HarshTrivedi:
This PR adds a command to generate and easily manage sub-experiments config files from a root experiment config file.

A typical scenarios that (I guess) everyone faces is: I have my experiment model up and running. Nice! But there are lots of things that can be yet configured. I have setting K1 that can take V1a, V1b and V1c values and setting K2 that can take V2a, V2b values and I want to try the combinations of these changes and analyse the results.

One can potentially use `--override` and write a script to try and report these things, but its quite an additional overhead to what this command provides. `--override` is in fact handy for quick checks but not really good when one has to manage bunch of experiment changes and analyse their combinations.

For such scenario one (at least I) typically wants way to:

1. **Part A** Easily generate valid experiment configurations for all 3X2=6 possibilities and organize them in easily identifiable config files.
2. **Part B** Train models using these generated experiment config files.
3. **Part C** And evaluate the trained models on set of test datas and give a analysable/searchable report (preferrably nice html tables).

I have been using Part A, B, C in my work but for now only part A (`generate-subexperiments` command) is PR ready.

To use `generate-subexperiments` command, one needs to pass:

```
1. root_experiment_config   config on which all changed defined in generator_config will be exectued.
2. generator_config         config defining how to generate subexperiments from root experiment.
3. output_directory         where generated config files will be stored.
```

The changes and combinations can be quite complicated at times, but using directly what jsonnet provides makes configuring generator pretty flexible. Let's say my model is `venue-classifier` from [allennlp-as-library](https://github.com/allenai/allennlp-as-a-library-example/blob/master/experiments/venue_classifier.json)  and I want to try out following things:

1. Group 1 (different sized glove embeddings) : 100d, 200d and 300d
2. Group 2 (lstm): bidirectional, unidirectional

So I want to generate 3 X 2 = 6 experiment config files and have them easily organized. There are some caveats here: if lstm `bidirectional` is ON, I would need to double the `input_dim` of `classifier_feedforward`. So in each of these 6 experiment congis, I need to make some post combination corrections. Assume projection is ON and so I don't need to adjust `input_dim` of lstm based on used `embedding_dim`.

The following generator config file would do the required:
```python
{
"combine_changes":[
        { # Group 1 of changes  (3 changes)
                "key_tuple": ["model.pretrained_file", "model.embedding_dim"],
                "value_tuples": [
                        ["'files/glove.6B.100d.txt.gz'", "100"],
                        ["'files/glove.6B.200d.txt.gz'", "200"],
                        ["'files/glove.6B.300d.txt.gz'", "300"]
                ],
                "value_names": ["100d", "200d", "300d"],
                "key_name": "glove_dim"
        },
        { # Group 2 of changes (2 changes)
                "key_tuple": ["model.title_encoder.bidirectional", "model.abstract_encoder.bidirectional"],
                "value_tuples": [
                        ["true", "true"],
                        ["false", "false"]
                ], # value_names can be inferred as well.
                "key_name": "bidirectional"
        }
],
# Sequence of changes for post combination corrections
"post_combine_changes": [
        { # Adjust classifier input after generating each combination.
                "key_tuple": ["model.classifier_feedforward.input_dim"],
                "value_tuple": [
                        """if current.model.title_encoder.bidirectional
                        then current.model.title_encoder.hidden_size * 4
                        else current.model.title_encoder.hidden_size * 2
                        """],
                # following names won't be added in experiment id. No identifier
                # needed because it's a change that all combinations will have.
                "value_name": "adjust",
                "key_name": "ff_input_dim"
        }
]
}
```

This will generate 6 valid experiment configs from root (`venue_classifier.json`) and will identify them with following names:

```
1. -> glove_dim=100d,bidirectional=true
2. -> glove_dim=100d,bidirectional=false
3. -> glove_dim=200d,bidirectional=true
4. -> glove_dim=200d,bidirectional=false
5. -> glove_dim=300d,bidirectional=true
6. -> glove_dim=300d,bidirectional=false
```

There are three parts of a generator config: `"pre_combine_changes"`, `"combine_changes"` and `"post_combine_changes"`. Combinations of groups of changes in `combine_changes` will be used to identify sub experiments. For each of these combinations following changes will be applied in order: sequence of `pre_combine_changes`, combination sequence for current experiment and sequence of `post_combine_changes`.

Because of what jsonnet allows one to do, generation config is very flexible. Each value in a `value_tuple` is a string of valid jsonnet expression. A jsonnet expression need not result in json, but can be string/int/bool etc as well. This makes configuring changes easy. `current` is a special jsonnet variable availble to refer to current state of config just before executing that specific change. Key names and value names make the final generated file names easy for human parsing and managing.

I guess the above example should be self-explanatory. Please check tests for more examples, if not.

### DIFF
--- a/allennlp/commands/__init__.py
+++ b/allennlp/commands/__init__.py
@@ -10,6 +10,7 @@ from allennlp.commands.fine_tune import FineTune
 from allennlp.commands.make_vocab import MakeVocab
 from allennlp.commands.predict import Predict
 from allennlp.commands.dry_run import DryRun
+from allennlp.commands.generate_subexperiments import GenerateSubexperiments
 from allennlp.commands.subcommand import Subcommand
 from allennlp.commands.test_install import TestInstall
 from allennlp.commands.train import Train
@@ -42,7 +43,7 @@ def main(prog: str = None,
             "fine-tune": FineTune(),
             "dry-run": DryRun(),
             "test-install": TestInstall(),
-
+            "generate-subexperiments": GenerateSubexperiments(),
             # Superseded by overrides
             **subcommand_overrides
     }
@@ -51,7 +52,7 @@ def main(prog: str = None,
         subparser = subcommand.add_subparser(name, subparsers)
         # configure doesn't need include-package because it imports
         # whatever classes it needs.
-        if name != "configure":
+        if name not in ["configure", "generate-subexperiments"]:
             subparser.add_argument('--include-package',
                                    type=str,
                                    action='append',
--- a/allennlp/commands/generate_subexperiments.py
+++ b/allennlp/commands/generate_subexperiments.py
@@ -197,10 +197,9 @@ class SequenceOfChanges:
     def __str__(self) -> str:
         return ','.join([str(change) for change in self.changes])
 
-    # TODO: should I make execute inplace operation?
     def execute(self, source_config: Params) -> Params:
         """
-            Applies each change in this SequenceOfChanges sequentially.
+        Applies each change in this SequenceOfChanges sequentially.
         """
         updated_config = deepcopy(source_config)
         for change in self.changes:
--- a/allennlp/commands/generate_subexperiments.py
+++ b/allennlp/commands/generate_subexperiments.py
@@ -6,7 +6,7 @@ to store the generated subexperiment configs.
 
 .. code-block:: bash
 
-    $ allennlp generate_subexperiments --help
+    $ allennlp generate-subexperiments --help
     usage: allennlp generate-subexperiments [-h]
                                             root_experiment_file generator_file
                                             subexperiments_dir
--- a/allennlp/commands/generate_subexperiments.py
+++ b/allennlp/commands/generate_subexperiments.py
@@ -1,0 +1,339 @@
+"""
+The ``generate_subexperiments`` subcommand can be used to generate
+configs of subexperiments from a base/root experiment. It requires a
+root config file, sub experiment generator config file and directory
+to store the generated subexperiment configs.
+
+.. code-block:: bash
+
+    $ allennlp generate_subexperiments --help
+    usage: allennlp generate-subexperiments [-h]
+                                            root_experiment_file generator_file
+                                            subexperiments_dir
+
+    Generate subexperiment configs form root experiment config.
+
+    positional arguments:
+      root_experiment_file  path to root experiment config file
+      generator_file        path to sub-experiment generator config file
+      subexperiments_dir    output directory to store generated config files
+
+    optional arguments:
+      -h, --help            show this help message and exit
+"""
+import itertools
+from copy import deepcopy
+import shutil
+from typing import Union, Sequence, Iterable
+from collections import OrderedDict
+import argparse
+import os
+import json
+import ast
+import logging
+
+import _jsonnet
+
+from allennlp.commands.subcommand import Subcommand
+from allennlp.common.params import Params, parse_overrides, with_fallback
+from allennlp.common.checks import ConfigurationError
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+class Change:
+    """
+    Encodes a single change from source config to updated config which doesn't need to be
+    internally combined. Eg. change num_dim of embeddings to 300, which requires 'model.
+    text_field_embedder.tokens.pretrained_file' set to '/pretrained_files/glove.6B.300d.
+    txt.gz' and 'model.text_field_embedder.embedding_dim' set to 300. Change has tuple of
+    keys and tuple of values of same size. Corresponding key will be set to corresponding
+    value when this Change is executed. Key is string with dot (.) to represent nesting
+    (as in jsonnet). Value are any valid jsonnet expression strings with access to extra
+    variable ``current``, which refers to state of source config before change execution.
+
+    Parameters
+    ----------
+    key_tuple : ``Sequence[str]``,
+        tuple of jsonnet field references expressed in strings.
+        eg ("trainer.num_epochs", "trainer.batch_size").
+    value_tuple : ``Sequence[str]``,
+        tuple of jsonnet expressions of expressed in strings of same size as `key_tuple`.
+        Corresponding key will be set to corresponding value when this Change is executed.
+        Value must be valid jsonnet expression with access to ``current`` variable refer
+        ing to state of source config before change execution.
+        Eg 1. `key_tuple`: ("50", "64") for `key_tuple`: ("trainer.num_epochs", "trainer.batch_size").
+        Eg 2. `key_tuple`: ("model.tuple_feedforward.hidden_dims") and value tuple:
+        ("[200 for _ in std.range(1, current.model.tuple_feedforward.num_layers)]")
+    key_name : ``str``, optional (joined ``key_tuple``s with '-')
+        This name will be used to refer to key in experiment id which has this change exectued.
+        key can often be very large, key_name provides good short name that is easy to parse for
+        humans. Eg. You can keep 'bidrectional' if you set 'model.seq2vec_encoder.bidirectional
+        in this change.
+    value_name : ``str``, optional (joined ``value_tuple``s with '-')
+        This name will be used to refer to value in experiment id which has this change exectued.
+        Eg. You can keep 'true' if you set 'model.seq2vec_encoder.bidirectional" to True in this
+        change. This key,value change will be referred as ``key_name``=`value_name` in the
+        experiment id.
+    """
+    def __init__(self,
+                 key_tuple: Sequence[str],
+                 value_tuple: Sequence[str],
+                 key_name: Union[str, int, bool] = None,
+                 value_name: Union[str, int, bool] = None) -> None:
+        if not isinstance(key_tuple, (list, tuple)):
+            raise ConfigurationError("key_tuple: '{}' must be tuple / list".format(key_tuple))
+        if not isinstance(value_tuple, (list, tuple)):
+            raise ConfigurationError("value_tuple: '{}' must be tuple / list".format(value_tuple))
+        if len(key_tuple) != len(value_tuple):
+            raise ConfigurationError("key_tuple: '{}' and value_tuple: '{}' "
+                                     "both must be tuples of same size.".format(key_tuple, value_tuple))
+        self.key_tuple = key_tuple
+        self.value_tuple = value_tuple
+        self.key_name = key_name or "_".join(key_tuple)
+        self.value_name = value_name or "_".join(value_tuple)
+        valid_key_name = any(isinstance(self.key_name, _type) for _type in [str, int, bool])
+        valid_value_name = any(isinstance(self.value_name, _type) for _type in [str, int, bool])
+        if not valid_key_name or not valid_value_name:
+            raise ConfigurationError("Value name and Key name must be string / int / bool")
+
+    def __str__(self) -> str:
+        return "{}={}".format(self.key_name, self.value_name)
+
+    def execute(self, source_config: Params) -> Params:
+        """
+        Takes in source_config and executes the change that this instance encodes.
+        """
+        define_current_var = 'local current = {} ;'.format(json.dumps(source_config.as_dict()))
+        overrides_dict = {}
+        for key, value in zip(self.key_tuple, self.value_tuple):
+            try:
+                value_evaluated = _jsonnet.evaluate_snippet("", define_current_var+value).strip()
+            except:
+                raise ConfigurationError("Following is not valid jsonnet expression:\n"
+                                         "{}".format(define_current_var+value))
+            # value_evaluated can be int / string / dict / bool. But it will be represented
+            # in terms of json objects. Eg. if value_evaluated is True, it will be 'true'.
+            # Since it may not be dict, json.loads does following is required.
+            for json_str, py_str in [('null', 'None'), ('true', 'True'), ('false', 'False')]:
+                value_evaluated = value_evaluated.replace(json_str, py_str)
+            try:
+                value_evaluated = ast.literal_eval(value_evaluated)
+            except:
+                raise ConfigurationError("{} could not be parsed as python expression.".format(value_evaluated))
+            overrides_dict[key] = value_evaluated
+        overrides_dict = parse_overrides(json.dumps(overrides_dict))
+        changed_config = with_fallback(preferred=overrides_dict,
+                                       fallback=source_config.as_dict())
+        return Params(changed_config)
+
+    @classmethod
+    def from_params(cls, change_params: Params) -> 'Change':
+        key_tuple = change_params.pop("key_tuple")
+        value_tuple = change_params.pop("value_tuple")
+        default_key_name = "_".join(key_tuple)
+        default_value_name = "_".join(value_tuple)
+        key_name = change_params.pop("key_name", default_key_name)
+        value_name = change_params.pop("value_name", default_value_name)
+        return cls(key_tuple=key_tuple, value_tuple=value_tuple,
+                   key_name=key_name, value_name=value_name)
+
+
+class GroupOfChanges:
+    """
+    Encodes group of similar Changes to be used one at a time.
+    Eg. group for trying embedding sizes from 100, 200 and 300.
+
+    Parameters
+    ----------
+    changes : ``Sequence[Change]``,
+        List/Tuple of changes with same ``key_tuple``
+    """
+    def __init__(self, changes: Sequence[Change] = ()) -> None:
+        if len({tuple(change.key_tuple) for change in changes}) > 1:
+            raise ConfigurationError("A GroupOfChanges must have single key_tuple.")
+        self.changes = changes
+
+    def __str__(self) -> str:
+        return ','.join([str(change) for change in self.changes])
+
+    @classmethod
+    def from_params(cls, change_params: Params) -> 'GroupOfChanges':
+        key_tuple = change_params.pop("key_tuple")
+        value_tuples = change_params.pop("value_tuples")
+        default_key_name = "_".join([str(key) for key in key_tuple])
+        default_value_names = ["_".join([str(value) for value in value_tuple])
+                               for value_tuple in value_tuples]
+        key_name = change_params.pop("key_name", default_key_name)
+        value_names = change_params.pop("value_names", default_value_names)
+
+        if len(value_tuples) != len(value_names):
+            raise ConfigurationError("Number of value tuples '{}' and corresponding "
+                                     "value names '{}' must be same.".format(value_tuples, value_names))
+        changes = [Change(key_tuple=key_tuple, value_tuple=value_tuple,
+                          key_name=key_name, value_name=value_name)
+                   for value_tuple, value_name in zip(value_tuples, value_names)]
+        return cls(changes)
+
+
+class SequenceOfChanges:
+    """
+    Encodes sequence of Changes to be executed in that order.
+    Eg. Set model.seq2vec_encoder.bidirectional set to True, Set embedding dim to 300
+    and Use glove embeddings of 300D.
+
+    Parameters
+    ----------
+    changes : ``Sequence[Change]``,
+        List/Tuple of changes
+    """
+
+    def __init__(self, changes: Sequence[Change] = ()) -> None:
+        if len({tuple(change.key_tuple) for change in changes}) < len(changes):
+            logger.info("Some changes have same 'key_tuple'. On executing "
+                        "this sequence, only later change will retain.")
+        self.changes = changes
+
+    def __str__(self) -> str:
+        return ','.join([str(change) for change in self.changes])
+
+    # TODO: should I make execute inplace operation?
+    def execute(self, source_config: Params) -> Params:
+        """
+            Applies each change in this SequenceOfChanges sequentially.
+        """
+        updated_config = deepcopy(source_config)
+        for change in self.changes:
+            updated_config = change.execute(updated_config)
+        return updated_config
+
+    @classmethod
+    def from_params_list(cls, changes_params: Sequence[Params] = ()) -> 'SequenceOfChanges':
+        return cls([Change.from_params(change_params) for change_params in changes_params])
+
+
+class SubExperimentsGenerator:
+    """
+    Takes root experiment config, sequence of pre-combine changes, sequence of
+    post-combine changes, and a list of groups of changes to be made combinations from.
+    Combinations are generated from list of groups taking one from each group at a
+    time. For each resultant combination (sequence of changes), changes are executed
+    in following order: pre-combine changes, combination sequence, post-combine changes.
+    Resulting generated experiment configs can be stored in directory by calling save.
+
+    Parameters
+    ----------
+    root_config : ``Params``
+        root/base experiment config to generate sub-experiments from.
+    pre_combine_changes : ``SequenceOfChanges``
+        Sequence of changes to be applied before combination.
+    to_be_combined_changes : ``Sequence[GroupOfChanges]``
+        List/Tuple of groups of changes to be combined across. One change will be
+        taken from each group for one experiment.
+    post_combine_changes : ``SequenceOfChanges``
+        Sequence of changes to be applied after applying a combination of change.
+    """
+    def __init__(self,
+                 root_config: Params,
+                 pre_combine_changes: SequenceOfChanges = SequenceOfChanges(()),
+                 to_be_combined_changes: Sequence[GroupOfChanges] = (),
+                 post_combine_changes: SequenceOfChanges = SequenceOfChanges(())) -> None:
+        self._root_config = root_config
+        self._pre_combine_changes = pre_combine_changes
+        self._to_be_combined_changes = to_be_combined_changes
+        self._post_combine_changes = post_combine_changes
+        self._subexperiments = self._generate()
+
+    @classmethod
+    def from_params(cls,
+                    root_config: Params,
+                    generator_config: Params) -> 'SubExperimentsGenerator':
+
+        pre_combine_configs = generator_config.pop("pre_combine_changes", ())
+        if not isinstance(pre_combine_configs, (list, tuple)):
+            raise ConfigurationError("key 'pre_combine_changes' must be a list/tuple")
+        pre_combine_changes = SequenceOfChanges.from_params_list(pre_combine_configs)
+
+        post_combine_configs = generator_config.pop("post_combine_changes", ())
+        if not isinstance(post_combine_configs, (list, tuple)):
+            raise ConfigurationError("key 'post_combine_changes' must be a list/tuple")
+        post_combine_changes = SequenceOfChanges.from_params_list(post_combine_configs)
+
+        group_change_configs = generator_config.pop("combine_changes")
+        to_be_combined_changes = [GroupOfChanges.from_params(group_change_config)
+                                  for group_change_config in group_change_configs]
+
+        return cls(root_config=root_config,
+                   pre_combine_changes=pre_combine_changes,
+                   to_be_combined_changes=to_be_combined_changes,
+                   post_combine_changes=post_combine_changes)
+
+    def _generate_combinations_of_changes(self) -> Iterable[SequenceOfChanges]:
+        """
+        Makes combinations from groups of changes,by taking one change from each group
+        for each combination. Encodes each combinations as SequenceOfChanges.
+        """
+        list_of_group_of_changes = [group_of_changes.changes
+                                    for group_of_changes in self._to_be_combined_changes]
+        for combination_changes in itertools.product(*list_of_group_of_changes):
+            yield SequenceOfChanges(combination_changes)
+
+    def _generate(self) -> OrderedDict:
+        """
+        Generates sub-experiments and returns ordered dict keyed by experiment name
+        and valued by final experiment config params.
+        """
+        subexperiments: OrderedDict = OrderedDict({})
+        combinations_of_changes = self._generate_combinations_of_changes()
+        for combination in combinations_of_changes:
+            experiment_config = self._pre_combine_changes.execute(self._root_config)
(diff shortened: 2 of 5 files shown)


matt-gardner:
Thanks for doing this @HarshTrivedi, this looks like some really useful functionality.  We typically use a separate system for doing this kind of thing internally at AI2, so it's not something we were going to build out ourselves.  And it looks like it'll be a long time before that system is released externally, so it would be nice to have something to handle these issues for AllenNLP users - several people asked me about this at ACL.

Does anyone have any strong opinions on the way this was done?  I don't, so I figure someone with more familiarity with jsonnet, or who has stronger ideas about setting up multiple configs like this, should take a look instead of me.

HarshTrivedi:
@matt-gardner  Thanks, You mentioned earlier about beaker internally here (#1476). I am assuming that it takes care of scheduling / running / evaluating experiments and generating nice analysable reports. I believe that in simple case, this roughly ties together with **Part B** and **Part C** as I mentioned in above comment. But that's not what this PR proposes.

I doubt if a general purpose separate tool with those use cases can generate sub experiments like this command. Generating valid experiments configs (for allennlp) is very much tied together with specifics of allennlp (eg. valid jsonnet) and specifics of your own model / experiment (eg. input of classifier layer should be doubled if I use bidirectional lstm and such idiosyncrasies).

One can surely have a script (specific to his/her experiment) to generate different `--overrides` and have different shell commands which can scheduled/run/evaluated and analysed by whatever tool. But `--overrides` can become very cumbersome even for simple example that I give above. Having different sub experiment config files is much more managable option than managing different `--overrides`.  Alternatively, one can manually generate sub-experiment configs files manually - but is error prone when one wants to explore more than few changes or combinations. 

So independent of what tool one uses for scheduling / running / analysing etc, an easy to use tool for generating valid sub-experiments configs for allennlp (based on what a researcher wants to explore) would be a helpful add-on.

joelgrus:
I'll take a look at it (not today though 😇)

joelgrus:
so I spent some time looking at this today, a few pieces of feedback / for discussion

(1) this feels to me like something that should be in `scripts/` rather than a command.
(2) this feels very complicated to me. like, "I read the code multiple times and I'm still not 100% sure what it's doing or how to use it" complicated. as this PR stands, I don't think anyone would actually be able to use this functionality unless you showed them how. it's possible that all it needs is a tutorial to go with it, but it's also possible that it's more complicated than it needs to be.
(3) I think this will mess up params that are specified by environment variables (that is, by `std.extVar`) -- when you do `Params.from_file` it will populate them based on the machine you're currently running on (or crash if they're not defined there), and then they're gone. If you were counting on them to be on another machine (say, if you run each experiment on some GPU machine), then they're gone. This feels like a problem.
(4) I'm not a huge fan of `ast.literal_eval`. I don't really understand the part of the code that uses it, and I feel like that makes it even harder to understand.

--

along the "too complicated" lines, isn't a lot of this available as built-in jsonnet functionality? that is, couldn't you just bake all of this into your jsonnet like

```
    "bidirectional": std.extVar("BIDIRECTIONAL"),
    ....
    ....
    "input_dim": if $.model.title_encoder.bidirectional then $.model.title_encoder.hidden_size * 4 else $.model.title_encoder.hidden_size * 2,
```

and then just iterate over combinations of environment variables? wouldn't that accomplish the same thing in a much simpler way?

joelgrus:
that long-winded comment was me thinking aloud as I read the PR, but really what it boils down to is: what problem does this solve that jsonnet doesn't already solve?

HarshTrivedi:
> 1) this feels to me like something that should be in scripts/ rather than a command. 

Sure, I can change that later.

> (2) this feels very complicated to me. like, "I read the code multiple times and I'm still not 100% sure what it's doing or how to use it" complicated. as this PR stands, I don't think anyone would actually be able to use this functionality unless you showed them how. it's possible that all it needs is a tutorial to go with it, but it's also possible that it's more complicated than it needs to be.

In terms of usage, it's actually very simple. I believe the reason it seems complicated to use is that I tried to show complicated examples first that leverage jsonnet features. Let me try again by starting with simple motivating examples iteratively showing why and in what scenario particular feature would be required. I will get back with examples tomorrow...

> (3) I think this will mess up params that are specified by environment variables (that is, by std.extVar) -- when you do Params.from_file it will populate them based on the machine you're currently running on (or crash if they're not defined there), and then they're gone. If you were counting on them to be on another machine (say, if you run each experiment on some GPU machine), then they're gone. This feels like a problem.

If I understand you correctly, that shouldn't be a problem.  `generate-subexperiments` generates fully specified sub-experiment config files. The output experiment config files would not have such expressions `std.extVar("BIDIRECTIONAL")`- they would be plain jsons with everything specified based on generator config and root config. And if something like `std.extVar("BIDIRECTIONAL")` is actually used in generator_config_file / root experiment file it would work as expected.

> (4) I'm not a huge fan of ast.literal_eval. I don't really understand the part of the code that uses it, and I feel like that makes it even harder to understand.

Sure, I knew you would point `literal_eval` out : ) .. but I don't see any alternative. `execute` method takes in a source experiment config and executes a change on it which that Change-instance encodes. For simplicity consider Change as a `key` and `value` pair instead of `key_tuple` and `value_tuple`. `key` refers to key that you want to change: eg 'model.seq2vec_encoder.bidrectional', and value is the hard-coded or computed value that you want to set. The value can be any valid string: eg. "true", "false", ("[1, 2, 3]" if arrays were permitted!) or jsonnet expression "! true", "! false" or something that requires complicated referrence: "if current.something then true else false".

The following line: `value_evaluated = _jsonnet.evaluate_snippet("", define_current_var+value).strip()` returns a evaluation of jsonnet expression. Eg. `value = '{"hello": 1+2}'` would yeild `value_evaluated = '{"hello": 3}'`. We generally parse it by `json.loads` since we expect output of it to be dict. However, jsonnet expression can lead to any thing as we want: boolean, string, list etc hence `json.loads` is not appropriate.  Eg. `value = '!false'` would lead to `value_evaluated = 'true'`. `value_evaluated` is actually is string of python object which we want to set but it has json based constants since it came out of jsonnet. Hence we convert: "true"->"True", "null"->"None" etc and finally use `ast.literal_eval` to parse it as string/dict/list/boo etc. 

> (5) couldn't you just bake all of this into your jsonnet like .. and then just iterate over combinations of environment variables? wouldn't that accomplish the same thing in a much simpler way?

In terms of usage, IMO it would be much more cumbersome to manage, and it doesn't support many things that script would support.

1. Firstly, to use something like `std.extVar("BIDIRECTIONAL")` you would need to set and manage this environment variable explicitly. But for `currrent.model.seq2vec.bidirectional` you don't need anything. The referencing says that see that in current  state of experiment config, bidirectional is true or not and decide based on that. By adding `current.` variable to jsonnet you are saying that you can reference any part of the current experiment config arbitrarily with full flexibility.
2. Secondly, it's not just about iterating over combinations - the script gives you easily managable handle on giving names to the changes and organizing them in separate files. One can actually open these files, look at their names and check if things are correct. Changes identified by simplified names make things much more easy to manage.

3. It also gives you way to group the logically groupable changes so that they don't get combinated and one can identify them easily. .. I guess examples I give later would make this clear.

4. ..

Let me get back tomorrow with separated and motivating examples for this kind of script and perhaps a short tutorial ...

HarshTrivedi:
> that long-winded comment was me thinking aloud as I read the PR, but really what it boils down to is: what problem does this solve that jsonnet doesn't already solve?

Sure, I get you. ...  jsonnet is not mandatory. Infact, I used this script for long time w/o jsonnet  for many things. So purpose of jsonnet and this scripts are quite different - jsonnet features are like addon which make it better.  ... Let me make some motivating examples and get back : )

HarshTrivedi:
This is quite verbose .. but it is more in the lines of a tutorial as you suggested ...

# What's the use?

**To put things in context:** A typical scenario that one usually faces is - I have my experiment config file working up and running but there are yet many things that can be configured for which there is no obvious choice. Only aparant way is to see empirically what works better and obtain the analysable trends. This PR is first step towards easy explorations of sub-experiments

A typical work flow for such a requirement would be:

1. Generate subexperiment config files and organize them with easily identifiable file names. (`generate-subexperiments`)
2. Train different models for each of the generated sub-experiment config files (`train-subexperiments`)
3. Evaluate these models on one or more evaluation datasets. (`evaluate-subexperiments`)
4. Generate analysable html reports that clearly say what was the change from root-experiment and what was the effect. (`analyse-subexperiments`)

The `generate-subexperiments` commands deals with part (1) of this requirement. It does so by taking in two files: `root_experiment_config` and `generator_config` and outputs all the identifiable sub-experiment config files in some specified `subexperiments_directory`.

The commands for (2), (3) and (4) are not in PR yet. But they basically, consume `subexperiments_directory` and do their required task.

The `generator_config` file gives a **structured and flexible** way to specify what your sub-epxeriments exploration will be. Following section explains it.

# Usage of generator config

For most examples, I will be using venue_classifier.json` from [here](https://github.com/allenai/allennlp-as-a-library-example/blob/master/experiments/venue_classifier.json) as root experiment config.

## Change

The underlying idea of generating **a** sub-experiment from a root experiment is a sequence of `Change`s. These `Change`s will be executed starting from root experiment **sequentially** to finally give use **one** sub-experiment. Something like: (`root_experiment_config` -> `Change1`, `Change2` ... `ChangeN` -> `sub_experiment_config`). So a `sub_experiment` config is nothing but a sequence of changes executed on `root_experiment_config`.

A `Change` simply put encodes which key in current config should be set to which value on execution of this change. It consists of `key_tuple`, `value_tuple`, `key_name` and `value_name`.  `key_tuple` is a **tuple** of keys and `value_tuple` is a **tuple** of values in which individual `key`, `value`'s have following roles.

- A key (str) refers to what "key" of `source_config` one wants to change.
    - Eg. `"model.title_encoder.bidirectional"`, `"trainer.batch_size"` etc.
- A value (str) refers to what "value" one wants to set that key to.
    - Eg. `"true"`, `"10"`, `"[20, 20, 20]"`, `"['relu']"`, `"'path/to/file'"` etc. A value is always a string irrespective of you want the value to be int / list/ dict/ str. See difference between `"true"` and `"'path/to/file'"` (extra `'` in `"'path/to/file'"`). Reason for this will be clear later.
- `key_name` and `value_name` give easy way of identifying the `Change`. Since a supexperiment is a unique sequence of changes, names of constituent changes of subexperiment give easily identifiable name to that sub-experiment.

Here, `key_tuple` and `value_tuple` are of same cardinality and ith key refers to ith value.

### Change (from params): Example 1

Lets say we want to construct `change` for turning Off bidirectionality of seq2vec-encoder for `title_encoder` :
```python
# Change: Make title_encoder unidirectional
{
    "key_tuple": ["model.title_encoder.bidirectional"],
    "value_tuple": ["false"],
    "key_name": "title-enc-bidirectional"
    "value_name": "false",
}
```

When this `change` is `executed` on our `venue_classifier` config, it will return a config file with unidirectional title encoder and all else unchanged. The name of this change will be: `title-enc-bidirectional=false`, which is representative of what it does on execution.

### Change (from params): Example 2

But, actually we have 2 seq2vec encoders in our model. Let's say we want to construct a change for turning bidirectional Off for both of them and identify this change as `all_bidirectional=false`.

```python
# Change: Make title_encoder and abstract_encoder unidirectional
{
    "key_tuple": ["model.title_encoder.bidirectional", "model.abstract_encoder.bidirectional"],
    "value_tuple": ["false", "false"],
    "key_name": "all_bidirectional"
    "value_name": "false",
}
```
See that corresponding entries in key_tuple and value_tuple form the corresponding pairs.
The name of this change will be: `all_bidirectional=false`.

Side note: This abstraction is implemented in `Change` class.

## Sequence of Changes

A sequence of changes is just list (sequence) of changes. For example, we want to make a sequence of changes for (1) make all seq2vec encoders unidirectional and (2) make embeddings trainable, it would be something as following:

```python
# Sequence of following changes:
# (1) making all seq2vec encoders unidirectional
# (2) make embedding trainable
[
{
    "key_tuple": ["model.title_encoder.bidirectional", "model.abstract_encoder.bidirectional"],
    "value_tuple": ["false", "false"],
    "key_name": "all_bidirectional"
    "value_name": "false",
}
{
    "key_tuple": ["model.text_field_embedder.tokens.trainable"],
    "value_tuple": ["true"],
    "key_name": "embeddings_trainable"
    "value_name": "true",
}
]
```

Name of sequence of changes reflects what changes it constitutes and in what order. The name of this sequence of changes will be: `all_bidirectional=false,embeddings_trainable=true`. As one can expect from name: when it will be `execute`d on `root_experiment_config`, it will make all seq2vec encoders unidirectional and make embeddings trainable.

Side note: This abstraction is implemented in `SequenceOfChanges` class.

## Group of Changes

A group of changes are the changes that one wants to take effect one at at time. Eg. lets say I want to try three types of `title_encoder`'s: "lstm", "gru", "rnn", then the `key` is `"model.title_encoder.type"` and values to be explored are `"'lstm'"`, `"'gru'"`, `"'rnn'"`. Clearly, a group will have 1 `key_tuple` and many `value_tuple`s that one wants to explore - one at a time. For this example, following params would construct such a group:

```python
{
    "key_tuple": ["model.title_encoder.type"],
    "value_tuples": [
                 ["lstm"], # value_tuple 1
                 ["gru"],  # value_tuple 2
                 ["rnn"],  # value_tuple 3
              ]
    "key_name": "title_encoder_type"
    "value_names": ["lstm", "gru", "rnn"],
}
```

Side note: This abstraction is implemented in `GroupOfChanges` class.

## Putting things together

Since we know what `Change`, `SequenceOfChanges` and `GroupOfChanges` are, building generator-config file is straight forward. A generator-config file has 3 parts. Following are the type annotations:

```python
{
   "pre_combine_changes": [], # Sequence of Changes
   "combine_changes": []  # List of Group of Changes
   "post_combine_changes": [] # Sequence of Changes
}
```

`pre_combine_changes` and `post_combine_changes` are pre and post combine hooks for executing sequence of changes before and after making combinations. For now, let's skip them - they are not useful for now.

The list of group/s in `"combine_changes"` is central for generating sub-experiments. The generated subexperiments will be combinations from these groups.

### Example 1

Lets say we want to try the combinations of:

1. try *lstm* and *gru* seq2vec encoders
2. try *bidirectional* and *unidirectional* seq2vec encoders


First step is to generate `GroupOfChanges` for (1) and (2).

A group of changes for (1) will be as follows:

```
    # Group 1: X 2 choices
    { # try *lstm* and *gru* seq2vec encoders
    "key_tuple": ["model.title_encoder.type", "model.abstract_encoder.type"],
    "value_tuples": [
                 ["lstm", "lstm"], # value_tuple 1
                 ["gru", "gru"  ], # value_tuple 2
              ]
    "key_name": "encoders_type"
    "value_names": ["lstm", "gru"],
    }
```

And group of changes for (2) will be as follows:

```
    # Group 2: X 2 choices
    { # try *bidirectional* and *unidirectional* seq2vec encoders
    "key_tuple": ["model.title_encoder.bidirectional", "model.abstract_encoder.bidirectional"],
    "value_tuples": [
                 ["true", "true"],     # value_tuple 1
                 ["false", "false"  ], # value_tuple 2
              ]
    "key_name": "all_bidirectional"
    "value_names": ["true", "false"],
    }
```

Finally, we just push these two in `combine_changes` list since we want to make combinations of these groups. Following is the full generator config file for such a generation:

```python
"combine_changes": [
    # Group 1: X 2 choices
    { # try *lstm* and *gru* seq2vec encoders
    "key_tuple": ["model.title_encoder.type", "model.abstract_encoder.type"],
    "value_tuples": [
                 ["lstm", "lstm"], # value_tuple 1
                 ["gru", "gru"  ], # value_tuple 2
              ]
    "key_name": "encoders_type"
    "value_names": ["lstm", "gru"],
    }

    # Group 2: X 2 choices
    { # try *bidirectional* and *unidirectional* seq2vec encoders
    "key_tuple": ["model.title_encoder.bidirectional", "model.abstract_encoder.bidirectional"],
    "value_tuples": [
                 ["true", "true"],     # value_tuple 1
                 ["false", "false"  ], # value_tuple 2
              ]
    "key_name": "all_bidirectional"
    "value_names": ["true", "false"],
    }

]
```

**Pictorially**, it is something like following:

```
                   encoders_type=lstm     all_bidirectional=true
root_experiment -> encoders_type=gru  ->  all_bidirectional=false  -> sub_experiment

                      [ Group 1]               [ Group 2]
                   (choose 1 here)           (choose 1 here)
```

This will generate 4 sub-experiments with following representative names (and as expected behaviour). These names will be present in the file names which these configs will be dumped to.

```
encoders_type=lstm,all_bidirectional=true
encoders_type=lstm,all_bidirectional=false
encoders_type=gru,all_bidirectional=true
encoders_type=gru,all_bidirectional=false
```

More concretely, the following will be the dumped file names:

```
<root_experiment_name>.encoders_type=lstm,all_bidirectional=true.json
<root_experiment_name>.encoders_type=lstm,all_bidirectional=false.json
<root_experiment_name>.encoders_type=gru,all_bidirectional=true.json
<root_experiment_name>.encoders_type=gru,all_bidirectional=false.json
```

This actually makes it very easy to debug, manage and analyse inidividual subexperiments. This is because, now we have an easy handle on individual sub-experiment config files with file-names representative of what is being explored in that particular sub-experiment.

Note that I gave here short example that results in only 4 subexperiments for brevity. Actualy one might want to explore much more (I typically try to analyse 10-15 subexperiments in one time). It's easy to imagine that w/o a structured way of organizing as I show above, sub-experiments exploration can become quite cumbersome.


### Example 2

Using combinations isn't necessary to be used. In some cases, one could just want to try different values. Let's say one wants on only explore all types of sec2vec encoder types and that it.

The following generator config file will do the required:

```python
"combine_changes": [
    # Group 1: X 2 choices
    { # try *lstm*, *gru*, *rnn* seq2vec encoders
    "key_tuple": ["model.title_encoder.type", "model.abstract_encoder.type"],
    "value_tuples": [
                 ["lstm", "lstm", "lstm"], # value_tuple 1
                 ["gru", "gru", "gru"],    # value_tuple 2
                 ["rnn", "rnn", "rnn"],    # value_tuple 3
              ]
    "key_name": "encoders_type"
    "value_names": ["lstm", "gru", "rnn"],
    }
]
```

**Pictorially**, it is something like following:

```
                   encoders_type=lstm
root_experiment -> encoders_type=gru  -> sub_experiment
                   encoders_type=rnn

                      [ Group 1]
                   (choose 1 here)
```


This will generate 3 sub-experiments with following names (and as expected behaviour)
```
encoders_type=lstm
encoders_type=gru
encoders_type=rnn
```

Using this one can simply create different types of sub-experiments with or without combinations. `post_combine_changes` becomes useful with usage of jsonnet as shown in next section. Please note that using jsonnet features is not always required. Extrapolating the above 2 examples, it should be evident that even w/o it, one can do fairly good amont of sub-experiment config files generation.

## Using Jsonnet Features with it

### Without Reference:

The `value` in `Change` is actually any valid jsonnet expression in string. jsonnet expression do not need to yield a json / dict always. It can yield any type of data: string / int / bool / list etc. This makes using or not using jsonnet seamless.

Lets say we know some `model.feedforward` has 3 `num_layers` and we want to set dropout of 0.2 for all the layers. The following `Change` would work.

```python
# Change: Set dropout
{
    "key_tuple": ["model.feedforward"],
    "value_tuple": ["[0.2  for _ in std.range(1, 3)]"],
    "key_name": "dropout"
    "value_name": "0.2",
}
```
Note: value `"[0.2  for _ in std.range(1, 3)]"` is a valid jsonnet expression (try [here](https://jsonnet.org/)) and hence it works.


### With Reference:

In above example, we assumed we knew that `model.feedforward` has 3 `num_layer`s. However, it's possible that it could have been changed before. In such cases, one can use `current` jsonnet variable that refers to the current state of experiment config. For example if a subexperiment is as follows: `root_experiment` -> `Change1` -> `Change2` -> `Change3` -> `sub_experiment` and we use current variable in `Change2`, it refers to the state of experiment config right after `Change1` is executed and before `Change2` is exectued.

For example, if unlike above example, we don't know the num_layers of feedforward upfront / or it's dynamic being changed in previous changes, then following can be used. It assures that dropout setting is consistent with `num_layers` independent of what `num_layers` were set to in previous cases.

```python
# Change: Set dropout
{
    "key_tuple": ["model.feedforward"],
    "value_tuple": ["[0.2  for _ in std.range(1, current.feedforward.num_layers)]"],
    "key_name": "dropout"
    "value_name": "0.2",
}
```

Below is a more concrete example would explain the need of `current` variable.

Lets say I want to explore the following:

1. try model.feedforward.num_layers to 2, 3, 4  [group of 3 changes]
2. try dropouts 0.0, 0.2  [group of 2 changes]

This combination will have 3 X 2 = 6 sub-experiments. Note that value of second group actually depends on what was being set in the first group. `num_layers` are dynamic and can't be known upfront. Leveraging `current.` variable, the following `generator config` solves this problem.

```python
"combine_changes": [
    # Group 1: X 3 choices
    { # model.feedforward.num_layers to 2, 3, 4
    "key_tuple": ["model.feedforward.num_layers"],
    "value_tuples": [
                 ["2"], # value_tuple 1
                 ["3"], # value_tuple 2
                 ["4"], # value_tuple 3
              ]
    "key_name": "ff_num_layers"
    "value_names": ["1", "2", "3"],
    }

    # Group 2: X 2 choices
    { # try dropouts 0.0, 0.2 for feedforward
    "key_tuple": ["model.feedforward.dropouts"],
    "value_tuples": [
                 ["[0.0  for _ in std.range(1, current.feedforward.num_layers)]"],  # value_tuple 1
                 ["[0.2  for _ in std.range(1, current.feedforward.num_layers)]"],  # value_tuple 2
              ]
    "key_name": "all_bidirectional"
    "value_names": ["true", "false"],
    }
]
```

The scenario is something like following:

```
                   ff_num_layers=2     dropouts=0.0
root_experiment -> ff_num_layers=3 ->  dropouts=0.2  -> sub_experiment
                   ff_num_layers=4

                      [ Group 1]        [ Group 2]
                   (choose 1 here)     (choose 1 here)

                                  ^
                                  |

                          `current` var in group 2
                          refers  to  config after
                          choice from group 1  has
                          been made  and  executed

```

Note, the usage of `current.` variable is essential in this case because in that specific sub-experiment `num_layers` cannot be known upfront. The variable `current.` refers to state of experiment config just before the execution of that change in sequence of changes. This allows to inspect any field / key / etc in current state of config and make decisions based on that.

Such a scenario also calls for usage of `post_combine_changes`. Note that even though, we adjust `dropout` according to num_layers set in that sub-experiment, the final config is not yet valid. One needs to make corrections / adjustments to the `activations` and `hidden_sizes` based on what `num_layers` were set to. However, it is just mandatory post correction to make sub-experiment valid, and not something that we are deliberately try to explore. All sub-experiments will need such a post correction, so we add this sequnce in `post_combine_changes` hook. In entirity the generator config file would look like:


```python
{

  # combine_changes is same as above.
  "combine_changes": [
      # Group 1: X 3 choices
      { # model.feedforward.num_layers to 2, 3, 4
      "key_tuple": ["model.feedforward.num_layers"],
      "value_tuples": [
                   ["2"], # value_tuple 1
                   ["3"], # value_tuple 2
                   ["4"], # value_tuple 3
                ]
      "key_name": "ff_num_layers"
      "value_names": ["1", "2", "3"],
      }

      # Group 2: X 2 choices
      { # try dropouts 0.0, 0.2 for feedforward
      "key_tuple": ["model.feedforward.dropouts"],
      "value_tuples": [
                   ["[0.0  for _ in std.range(1, current.feedforward.num_layers)]"],  # value_tuple 1
                   ["[0.2  for _ in std.range(1, current.feedforward.num_layers)]"],  # value_tuple 2
                ]
      "key_name": "all_bidirectional"
      "value_names": ["true", "false"],
      }
  ]

  # Sequnce of Changes for Post Combine Corrections.
  "post_combine_changes": [
    {
      "key_tuple": ["model.feedforward.activations"],
      "value_tuple": ["['relu'  for _ in std.range(1, current.feedforward.num_layers)]"]
    }
    {
      "key_tuple": ["model.text_field_embedder.tokens.trainable"],
      "value_tuple": ["[200  for _ in std.range(1, current.feedforward.num_layers)]"],
    }
  ]

}
```

This will generate the following sub-experiment names:

```
ff_num_layers=2,all_bidirectional=true
ff_num_layers=2,all_bidirectional=false
ff_num_layers=3,all_bidirectional=true
ff_num_layers=3,all_bidirectional=false
ff_num_layers=4,all_bidirectional=true
ff_num_layers=4,all_bidirectional=false
```

Note that post combination changes won't count towards the sub-experiment name since they are common across all sub-experiments.


### Summary of Generator Config:


To **summarize**, a general generator config builds on idea of sequence of Changes and if all sections are used pictorially it would look like:

```
root_experiment --> [ pre-change-1 -> pre-change-2 .. pre-change-N ]   -->  [  Group 1       Group 2    ..  Group 3    ]      -->  [ post-change-1 -> post-change-2 .. post-change-N ]  -->  sub-experiment
                                                                                 .             .             .
                                                                                 .             .             .
                                   "pre_combine_changes"                      choose-1     choose-1    ..  choose-1                        "post_combine_changes"

                                                                                             "combine_changes"
```


# Importance of representative names

From above examples, I would be evident that names of sub-experiments (composed from constituent changes) actually makes it easily managable. Having viewable sub-experiment config files with relevant names makes it easy to verify that sub-experiment is as expected as file-name suggests. However, that's not only benefit:

Although `report_experiments` command is not PR-ready, the final report makes a lot of sense if you have given names to individual changes that you are exploring.  Eg. If there are 4 subexperiments being explored:
```
all_bidirectional=true,embedding_trainable=true
all_bidirectional=true,embedding_trainable=false
all_bidirectional=false,embedding_trainable=true
all_bidirectional=false,embedding_trainable=false
```
it would generate searchable html report for each configured evauation dataset as (check #1566 for sample example):

| all_bidirectional|embedding_trainable|metric_1|metric_2|
|----------|:-------------:|------:|------:|
| true |  true    | 0.8 | 0.6
| true |  false   | 0.3 | 0.7
| false | true    | 0.6 | 0.3
| false | false   | 0.7 | 0.5

The ability to club changes in one logical name also helps analysis. They abstract out internal details of what that change constitutes. Eg. `all_bidirectional` actually means bidirectional of both abstract and title encoder but it's a redundant detail for analysis and organizing the sub-experiment file names.


# Other Comments

> what problem does this solve that jsonnet doesn't already solve?

I hope it's clear from these examples that the purposes are fundamentally different. jsonnet features are addon to sub-experiment generation command and one can handle fairly good scenarios of sub-epxeriment generation even without jsonnet features. I specifically delayed using jsonnet in examples to clarify this point.

> Why not just bake all of this into your jsonnet ... and iterate over combinations of environment variables?

The 3 reasons I gave in above comment and examples here support why not. Lastly, abstractions that `Change`, `SequenceOfChanges` and `GroupOfChanges` give to generate subexperiments from root experiment appears much cleaner to me and are surely more flexible.

> "I read the code multiple times and I'm still not 100% sure what it's doing or how to use it"

I was mistaken that 1 example in original PR and test cases would be sufficient to explain purpose and syntax! But after following examples in **this** comment, **using** generator-config should seem very straight-forward. For reviewing implementation, looking at `Change`, `SequenceOfChanges` and `GroupOfChanges` classes and comparing with their utility from examples above should suffice.

> I don't think anyone would actually be able to use this functionality unless you showed them how ..

I can make a tutorial out of this comment, if it seems appropriate.

> ...  but it's also possible that it's more complicated than it needs to be.

The examples I give above are still simple enough to make a point of each functionality and explain the short pipeline. For very small / simple examples (**implementation wise**) it might seem that a structured approach to sub-experiment generation is an overkill. But I would request to consider the possibility that one might need to handle 10's of sub-experiments explorations and need a way to handle/manage and analyse them. As far as usage / end-user friendliness is concerned, I hold that usage should be quite straightforward for small / big examples after following this comment.

joelgrus:
thanks for all the explanation, but I still don't get what your code gives me that I can't get using basic jsonnet features. 

here I modified the venue_config.jsonnet to be parameterized:

https://gist.github.com/joelgrus/5ce9af24373a23ba97436874a9c3aec8

and I wrote a very light python script to generate the corresponding configs. (I didn't generate names for them, but that would be one more line of code).

what important features is my version missing?

HarshTrivedi:
One can always write a specific script + parameterized root config json for his/her own specific purpose and can do all the things as I show above.

The point of my script is to not have to write specific parameterized root config or specifc script. But rather have generalized way of configuring what one wants to explore by defining only one config.
This includes: what keys will be set to what value, names and (not included yet) evaluation datasets to explore, analyse and report these generated sub-experiment configs on. All things pertaining to
train / evaluate / analyse / report explorations of the root experiment in **one** place (and **separate** from root config).

An equivalent config for above command would be:

```json
{

  "combine_changes": [
      {
      "key_tuple": ["model.title_encoder.bidirectional", "model.abstract_encoder.bidirectional"],
      "value_tuples": [
                   ["true", "true"],
                   ["false", "false"]
                ],
      }
      {
      "key_tuple": ["model.text_field_embedder.tokens.trainable"],
      "value_tuples": [
                   ["true"],
                   ["false"],
                ],
      }
      {
      "key_tuple": ["model.title_encoder.type"],
      "value_tuples": [
                   ["'gru'"],
                   ["'lstm'"],
                ],
      }
  ]
}
```

and would not require any specific parameterization / variable defining / changes in root config. It is self-sufficient and quite succinct on what one wants to explore starting from a root config.

If I compare your way and mine, main difference is:

- I decide what key will be set to what value in one separate config.
- You define variables in one config (SUB), pass variables in to root config and leave decision of what key to set to what value in root config.

If you make a generally usable script out of your script, it will need to take 2 inputs: SUB like config file and parameterized root config both of which will have to be configured specific to one's requirement.
Although your is simple implementation-wise, I think this PR one is easier to use. If you think otherwise, I would be fine closing this ....

HarshTrivedi:
@joelgrus no problem .. Thank you for the feedback!

PR #1543, opened 2018-07-29
upgrade to pytorch 0.4.1 + make work with python 3.7 (but still 3.6 also)

joelgrus:
tested under (3.6 + 0.4), (3.6 + 0.4.1), (3.7 + 0.4.1)

(I am pretty sure that 3.7 requires 0.4.1)

nasty subtlety where Python 3.7 changed an undocumented feature I was relying on to identify annotations as Dicts or Lists.

also, I added a script to check whether setup.py and requirements.txt are in sync, it's kind of hacky, but it found me a discrepancy already


matt-gardner:
Seems fine to just lower the precision of that test.  6 decimals is plenty.

setup.py
```diff
@@ -101,7 +101,7 @@
       packages=find_packages(exclude=["*.tests", "*.tests.*",
                                       "tests.*", "tests"]),
       install_requires=[
-          'torch==0.4.0',
+          'torch==0.4.1',
```
DeNeutoy: >=0.4.0?

joelgrus: what's the rationale for allowing 0.4.0 still?

matt-gardner: That's also a valid point, if the version matters for our support. I could go either way here.

DeNeutoy: For minor versions of the key library that we depend on, this seems a bit inflexible. But, if you feel strongly, go ahead.

matt-gardner: The only other thing I'd add is that if we have a lower bound, we should have an upper bound too.  Not sure if that should be `0.4.2`, `0.5.0`, or `1.0.0`.

huntzhan:
Hi Team, any plan for a new release of this bugfix?

PR #1544, opened 2018-07-29
re-work dependency parser to use HEAD sentinel inside model

DeNeutoy:


DeNeutoy:
You want to embed your sentence with ELMo - if you have the root head in the dataset reader, it's a little tricky to remove it before it gets slurped up by the text_field_embedder - and I don't want a contextual representation of the token `ROOT_HEAD`.

PR #1548, opened 2018-07-30
Remove the unused NltkWordSplitter and punkt model.

schmmd:


schmmd:
I first noticed this when I was going through Bhavana and Niket's code to run their AllenNLP dependency on Beaker.  They explicitly installed the `punkt` model, and I wasn't sure that model was necessary.  It likely isn't because, while we support a word splitter that uses the `punkt` model--we don't actually use it anywhere else in our code or in any of our models.  We also don't have tests to make sure it's working correctly (tests pass if you remove the `punkt` model).

While I prefer removing this component and dependency, the alternative for me to open and issue to write some tests for the `NltkWordSplitter`.

@DeNeutoy and I briefly discussed this in https://github.com/allenai/allennlp/pull/1541.

schmmd:
image

PR #1551, opened 2018-07-30
WIP: Skip tests that require Java in test-install

schmmd:
* Skip tests that require `java` when running `test-install` as we don't distribute Java via `pip`.
* Exit with the same exit code as `pytest` when running `test-install`.
* Run in the right directory so test fixtures are found.

### DIFF
--- a/allennlp/commands/test_install.py
+++ b/allennlp/commands/test_install.py
@@ -52,10 +52,10 @@ def _get_module_root():
 
 def _run_test(args: argparse.Namespace):
     initial_working_dir = os.getcwd()
-    module_root = _get_module_root()
-    logger.info("Changing directory to %s", module_root)
-    os.chdir(module_root)
-    test_dir = os.path.join(module_root, "tests")
+    module_parent = _get_module_root().parent
+    logger.info("Changing directory to %s", module_parent)
+    os.chdir(module_parent)
+    test_dir = os.path.join(module_parent, "allennlp")
     logger.info("Running tests at %s", test_dir)
     if args.run_all:
         # TODO(nfliu): remove this when notebooks have been rewritten as markdown.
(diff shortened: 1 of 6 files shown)


schmmd:
@DeNeutoy thanks, although this is still WIP as it needs to be tested.  This morning I noticed that `test-install` might actually be failing on a lot more tests--and so instead of seeing this fixed I discovered more problems.

allennlp/commands/test_install.py
```diff
@@ -52,15 +52,17 @@ def _get_module_root():
...
         # TODO(nfliu): remove this when notebooks have been rewritten as markdown.
-        pytest.main([test_dir, '-k', 'not notebooks_test'])
+        ec = pytest.main([test_dir, '--color=no', '-k', 'not notebooks_test'])
     else:
-        pytest.main([test_dir, '-k', 'not sniff_test and not notebooks_test'])
+        ec = pytest.main([test_dir, '--color=no', '-k', 'not sniff_test and not notebooks_test',
+                     '-m', 'not java'])
     # Change back to original working directory after running tests
     os.chdir(initial_working_dir)
+    exit(ec)
```
schmmd: @joelgrus does this look OK to you?

schmmd:
@nelson-liu I will add an automatic test to TeamCity.  I'll set it up to run `test-install` after it builds `Dockerfile.pip`.

allennlp/commands/test_install.py
joelgrus: it looks fine, but I'm not familiar enough with the interplay of pytest and sys.exit to notice if it's wrong

schmmd: It works--I tested that.  I was curious if it bothered you that I:m calling `exit` from a function--seems like it doesnt!

PR #1555, opened 2018-08-01
make passing different feedforward modules more flexible

DeNeutoy:
Just shuffling how I create feedforward layers so they aren't hardcoded.


allennlp/models/biaffine_dependency_parser.py
```diff
@@ -83,15 +93,33 @@ def __init__(self,
...
+        if arc_feedforward is not None:
+            self.head_arc_feedforward = arc_feedforward
+            self.child_arc_feedforward = copy.deepcopy(arc_feedforward)
+        else:
+            self.head_arc_feedforward = FeedForward(encoder_dim, 1,
+                                                    arc_representation_dim,
+                                                    Activation.by_name("elu")())
+            self.child_arc_feedforward = FeedForward(encoder_dim, 1,
+                                                     arc_representation_dim,
+                                                     Activation.by_name("elu")())
```
joelgrus: nit: it would be slightly cleaner / simpler if you did

```
self.head_arc_feedforward = head_arc_feedforward or FeedForward(...)
self.child_arc_feedforward = copy.deepcopy(self.head_arc_feedforward)
```

which I think amounts to the same thing?

allennlp/models/biaffine_dependency_parser.py
```diff
@@ -83,15 +93,33 @@ def __init__(self,
...
+        if tag_feedforward is not None:
+            self.head_tag_feedforward = tag_feedforward
+            self.child_tag_feedforward = copy.deepcopy(tag_feedforward)
+        else:
+            self.head_tag_feedforward = FeedForward(encoder_dim, 1,
+                                                    tag_representation_dim,
+                                                    Activation.by_name("elu")())
+            self.child_tag_feedforward = FeedForward(encoder_dim, 1,
+                                                     tag_representation_dim,
+                                                     Activation.by_name("elu")())
```
joelgrus: same comment as above

Issue #1539, opened 2018-07-27
pytest fails if /tmp/allennlp_tests/ already exists and cannot be deleted


dallascard:
**Describe the bug**
Running tests with pytest causes allennlp to create and delete the directory /tmp/allennlp_tests/ but if this directory already exists and is owned by a different user, the attempt to delete the directory at the end of the test will fail.

**To Reproduce**
mkdir /tmp/allennlp_tests/
chmod a-r /tmp/allennlp_tests/
pytest (in any allennlp project with tests)

**Expected behavior**
This should fail with PermissionError: [Errno 13] Permission denied: '/tmp/allennlp_tests/'

**System (please complete the following information):**
 - OS: Linux
 - Python version: 3.6
 - AllenNLP version: 0.4.3

**Additional context**
The problem seems to be in allennlp/allennlp/common/testing/test_case.py where the temp directory is hard coded as
`self.TEST_DIR = pathlib.Path("/tmp/allennlp_tests/")`
It seems like it would be better if this used the pytest temp directory.

schmmd:
Wow that's a lot of failures!  Easy to reproduce with:

```
mkdir -p /tmp/allennlp_tests/
sudo chown root /tmp/allennlp_tests/
pytest .
```

schmmd:
This is also a blocker for anyone trying to run tests on Windows (`/tmp` won't exist).  I started working on a solution here in https://github.com/allenai/allennlp/pull/1556 but ran into https://github.com/allenai/allennlp/issues/1563.

PR #1556, opened 2018-08-01
WIP: Use a temp directory for tests.

schmmd:


schmmd:
Starting with one of the ELMo tests I start getting the error `OSError: [Errno 24] Too many open files: '/var/folders/v5/tlk1sh3d0l94fcmz10vtzmyh0000gn/T/tmpvnetszsn'` when running `shutil.rmtree`.  I didn't realize this at first because `tearDown` eats errors.

nelson-liu:
maybe it might be easier to replace `/tmp/allennlp_tests` with `Path(tempfile.gettempdir()) / (allennlp_tests + current_username + a random hash to boot)` (using a random hash that makes the path unique if it isn't already)? 

Seems like it should achieve the same effect.

schmmd:
@nelson-liu my backup plan (which I'm about to move to) is to replace `/tmp/allennlp_tests` with `tempfile.mkdtemp`--but to continue re-using the same directory for all our tests.

I'm worried regardless that we have a file leak in AllenNLP.

schmmd:
Nope--looks like the root cause needs to be fixed in either case.  I have the same issue with the following change:

```
 import os
 import pathlib
 import shutil
+import tempfile
 from unittest import TestCase
 
 from allennlp.common.checks import log_pytorch_version_info
 
+TEST_DIR = pathlib.Path(tempfile.mkdtemp(prefix="allennlp_tests_"))
 
 class AllenNlpTestCase(TestCase):  # pylint: disable=too-many-public-methods
     """
@@ -31,8 +33,7 @@ class AllenNlpTestCase(TestCase):  # pylint: disable=too-many-public-methods
         logging.getLogger('urllib3.connectionpool').disabled = True
         log_pytorch_version_info()
 
-        self.TEST_DIR = pathlib.Path("/tmp/allennlp_tests/")
-
+        self.TEST_DIR = TEST_DIR
         os.makedirs(self.TEST_DIR, exist_ok=True)
 
     def tearDown(self):
```

handsomezebra:
What's the current status of this issue? I am getting this error everytime on my mac environment, but not on my Linux environment.

`OSError: [Errno 24] Too many open files`

nelson-liu:
@handsomezebra i think https://github.com/allenai/allennlp/pull/1564 was merged instead as a temporary fix for #1539 . @schmmd mentioned in that PR that he had to up his ulimit to get the tests to run --- perhaps that might work for you as well?

handsomezebra:
Yes, using `ulimit -n 8192` solved most of the problem. Thanks!
But still all four s3 unit tests are failing for me on this machine. Did you guys see similar things? If not I can try some other machines and reproduce.

nelson-liu:
Hmm, I cannot reproduce that. Can you open a new issue describing what you see, and we'll move the conversation there?

PR #1557, opened 2018-08-02
Expose iterator shuffling through Trainer config

nelson-liu:
Not really sure how to test this, any suggestions?

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -974,6 +974,7 @@ class Trainer:
 
         patience = params.pop_int("patience", None)
         validation_metric = params.pop("validation_metric", "-loss")
+        shuffle = params.pop_bool("shuffle", True)
         num_epochs = params.pop_int("num_epochs", 20)
         cuda_device = params.pop_int("cuda_device", -1)
         grad_norm = params.pop_float("grad_norm", None)
@@ -1003,6 +1004,7 @@ class Trainer:
                        patience=patience,
                        validation_metric=validation_metric,
                        validation_iterator=validation_iterator,
+                       shuffle=shuffle,
                        num_epochs=num_epochs,
                        serialization_dir=serialization_dir,
                        cuda_device=cuda_device,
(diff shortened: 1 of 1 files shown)


joelgrus:
hmm, are we shuffling the validation set by default? that's kind of weird?

it's ok if you don't want to test this. if I really wanted to test this, I would probably (in the test file) create a `BasicIterator` subclass that remembers its instances as it iterates over them, then I would run training just as the current `TrainerTest` does (maybe for 10 or 20 epochs) and check that in the the no-shuffle case the instances are are in the same order each epoch and that in the shuffle case they're not. but that might be more work than is worthwhile.

allennlp/training/trainer.py
```diff
@@ -166,6 +166,7 @@ def __init__(self,
                  patience: Optional[int] = None,
                  validation_metric: str = "-loss",
                  validation_iterator: DataIterator = None,
+                 shuffle: bool = True,
```
joelgrus: `Trainer` has a custom from_params, so you'll need to add it there too

nelson-liu: good catch, thanks!

nelson-liu:
> hmm, are we shuffling the validation set by default? that's kind of weird?

yeah, I noticed this too. I'd be in favor of turning off shuffling at validation, since I presume that's what we do at test time? thoughts?

joelgrus:
it looks like we are shuffling at test time too. 😬 

anyway, it appears that we've always been shuffling the validation set, so I wouldn't casually change that behavior. (in theory it shouldn't make a difference, addition is commutative, but who knows)

DeNeutoy:
Can someone remind me why this isn't a constructor arg for the `Iterator`? The way we are passing it here means it's essentially equivalent, as you can't choose to change it whilst training.

joelgrus:
I assumed it was so the same iterator could behave differently on training / dev / test, although that's not possible right now

PR #1560, opened 2018-08-02
Demo features for the dependency parser

DeNeutoy:
This makes the demo look like this:

image

I was trying to make it look suitably different from the constituency parsing demo, so I added in some other styling to make some dependency edges be rendered in particular ways.


allennlp/predictors/biaffine_dependency_parser.py
```diff
@@ -40,3 +107,79 @@ def _json_to_instance(self, json_dict: JsonDict) -> Instance:
         sentence_text = [token.text for token in spacy_tokens]
         pos_tags = [token.tag_ for token in spacy_tokens]
         return self._dataset_reader.text_to_instance(sentence_text, pos_tags)
+
+
```
matt-gardner: `@overrides` here?

allennlp/predictors/biaffine_dependency_parser.py
```diff
@@ -1,12 +1,79 @@
...
+NODE_TYPE_TO_STYLE = {}
+
+NODE_TYPE_TO_STYLE["root"] = ["color5", "strong"]
+NODE_TYPE_TO_STYLE["dep"] = ["color5", "strong"]
+
+# Arguments
+NODE_TYPE_TO_STYLE["nsubj"] = ["color1"]
+NODE_TYPE_TO_STYLE["nsubjpass"] = ["color1"]
+NODE_TYPE_TO_STYLE["csubj"] = ["color1"]
+NODE_TYPE_TO_STYLE["csubjpass"] = ["color1"]
```
joelgrus: nit: you could make this a lot shorter if you did something like

```
NODE_TYPE_TO_STYLE = {
    **{tag: ["color5", "strong"] for tag in ['root', 'dep']},
    **{tag: ["color1"] for tag in [...]},
    ...
}
```

but it's up to you

matanster:
I think having the demo page also optionally show the good-old ugly bracketed notation parse output, would be a nice feature too.

Issue #1563, opened 2018-08-02
The Java subprocess in the Wikitables component contains a file leak


schmmd:
**Describe the bug**
The Java subprocess in the Wikitables component contains a file leak.  I first encountered this when moving the test output to separate files.  This may have manifested because, while before we repeatedly deleted the same directory, after my change we only tried to delete it once.

**To Reproduce**
Run WikiTablesErmSemanticParserTest and then look at the open files in the subprocesses.

```
for child in psutil.Process().children():                                                   
  print(f"{child}")                                                                   
  print(", ".join([x.path for x in child.open_files()])  
```

```
psutil.Process(pid=22759, name='java', started='13:04:51')
 /Library/Java/JavaVirtualMachines/jdk1.8.0_25.jdk/Contents/Home/jre/lib/rt.jar, /System/Library/Java/Extensions/MRJToolkit.jar, /Users/michael/.allennlp/datasets/70cdbc01ae81bf9c7ba493151627b27ab0f86026fa89531ada88537c419ba351.9ba2c80a7ede842334060d65d4fc68c62be05b065fc1c98090ede4f0d1edccf7, /Library/Java/JavaVirtualMachines/jdk1.8.0_25.jdk/Contents/Home/jre/lib/ext/cldrdata.jar, /Library/Java/JavaVirtualMachines/jdk1.8.0_25.jdk/Contents/Home/jre/lib/ext/localedata.jar, /Users/michael/hack/allenai/allennlp/data/grow.grammar
psutil.Process(pid=22758, name='java', started='13:04:51')
[...]

PR #1564, opened 2018-08-02
Don't use a hard-coded temp directory.

schmmd:
Fixes https://github.com/allenai/allennlp/issues/1539 and makes AllenNLP less system-dependent.

### DIFF
--- a/allennlp/common/testing/test_case.py
+++ b/allennlp/common/testing/test_case.py
@@ -3,10 +3,12 @@ import logging
 import os
 import pathlib
 import shutil
+import tempfile
 from unittest import TestCase
 
 from allennlp.common.checks import log_pytorch_version_info
 
+TEST_DIR = tempfile.mkdtemp(prefix="allennlp_tests")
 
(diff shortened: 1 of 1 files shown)


schmmd:
I thought this was causing an error on my local machine--but then it turns out I can't even run tests on master without increasing my `ulimit -n`.  This seems like a clear improvement--although I think it would be even better to have tests store results in independent directories.

schmmd:
@joelgrus any suggestions on what I should name the static variable?

joelgrus:
where do you clean up that directory ?

schmmd:
@joelgrus look in `tearDown`.  It uses the same code as before to clean up the directory.

joelgrus:
so the module when it first runs uses `mkdtemp` to create a tempdir. 

then every test creates again and deletes that same directory?

something feels slightly off about that, but I guess it works

schmmd:
@joelgrus I agree it seems off--I see it as one step better from where we are now.  With the file leak issue in https://github.com/allenai/allennlp/issues/1563 I don't want to create one directory per test because we are not able to clean them up.

PR #1572, opened 2018-08-03
be friendlier to windows users

joelgrus:
these are small things that prevent allennlp from running on windows, this isn't all the way to getting it to work, but it's a step in that direction:

(1) jsonnet won't pip install on windows right now, so I disabled it in `requirements.txt` if the platform is windows
(2) accordingly, in `params.py` we wrap the `import _jsonnet` in a try-catch block, and if it can't be imported we stub in alternative functions that log a warning and assume the configuration is pure JSON
(3) the `resource` module doesn't exist on Windows, so in `util.py` we import it conditionally, and then we check that it's there before we use it
(4) on unix, urlparse.scheme is an empty string for a file, but on windows it isn't, so I just removed that check in `cached_path`, it should be fine to just check if the file exists whatever the scheme is

### DIFF
--- a/allennlp/common/file_utils.py
+++ b/allennlp/common/file_utils.py
@@ -81,7 +81,7 @@ def cached_path(url_or_filename: Union[str, Path], cache_dir: str = None) -> str
     if parsed.scheme in ('http', 'https'):
         # URL, so get it from the cache (downloading if necessary)
         return get_from_cache(url_or_filename, cache_dir)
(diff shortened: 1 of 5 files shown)


joelgrus:
I don't know that we'll ever officially support it, I just want to make it easier for the people who are clamoring for it

joelgrus:
but if I ever get it working end-to-end I'll look into a CI way of making sure it stays working

PR #1577, opened 2018-08-06
atis dataset reader

kl2806:
- Add a dataset reader for ATIS
- Add binary linking scores between each utterance token and each entity
- Refactor some of how we get numbers to reduce the number of false positives


allennlp/semparse/contexts/atis_tables.py
```diff
@@ -65,31 +120,49 @@ def get_date_from_utterance(tokenized_utterance: List[Token],
...
+    # When we use a regex to find numbers or strings, we need a mapping from
+    # the character to which token triggered it.
+    char_offset_to_token_index = {token.idx : token_idx for token_idx, token
+                                  in enumerate(tokenized_utterance)}
+
+    # We want to look up later for each time whether it appears after a word
+    # such as "about" or "approximately".
+    indices_approximate_words = [idx for idx, token in enumerate(tokenized_utterance) \
+            if token.text in APPROX_WORDS]
+    number_linking_dict = get_regex_match(r'\d+',
```
matt-gardner: Oh, I see, you're using this for more than just times.  Hmm...

It seems like this logic is doing two different things: (1) it is getting numbers like "Flights from 2 airlines", or something; (2) it is getting times like "flights around 7", without an "am" or "pm".  The trouble is it's not doing (2) correctly, because "flights at 7" will map to "7", not "700".  It looks like you're also double counting "7" and "7am" here, as this one will also match "7am".

So, I'd recommend treating the first case here, and you can do that with something much simpler than a regex: just iterate over the tokens in the utterance, and try `int(token.text)`, keeping the value if it succeeds.  Then leave the time matching to `get_times_from_utterance`, being sure to have this logic not match things like "7am".

Does that sound right?

allennlp/semparse/contexts/atis_tables.py
```diff
@@ -65,31 +120,49 @@ def get_date_from_utterance(tokenized_utterance: List[Token],
...
+
+    times_linking_dict = get_times_from_utterance(utterance,
+                                                  char_offset_to_token_index,
+                                                  indices_approximate_words)
+
+    for key, value in times_linking_dict.items():
+        number_linking_dict[key].extend(value)
+
+    for idx, token in enumerate(tokenized_utterance):
+        if token.text in MONTH_NUMBERS:
```
matt-gardner: It probably makes sense to make a `NUMBER_TRIGGER_DICT` similar to your `ATIS_TRIGGER_DICT`, so you don't have to have so many independent-but-similar checks here.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -9,6 +10,29 @@
 
 from allennlp.data.tokenizers import WordTokenizer
 
+def get_strings_from_utterance(tokenized_utterance) -> Dict[str, List[int]]:
+    """
+    Based on the current utterance, return a list of valid strings that should be added.
```
matt-gardner: This docstring needs to be updated with the new return type.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -9,6 +10,29 @@
 
 from allennlp.data.tokenizers import WordTokenizer
 
+def get_strings_from_utterance(tokenized_utterance) -> Dict[str, List[int]]:
```
matt-gardner: Type for `tokenized_utterance`.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -27,34 +51,72 @@ def __init__(self, utterances: List[str], tokenizer=None) -> None:
         self.utterances: List[str] = utterances
         self.tokenizer = tokenizer if tokenizer else WordTokenizer()
         self.tokenized_utterances = [self.tokenizer.tokenize(utterance) for utterance in self.utterances]
-        self.valid_actions: Dict[str, List[str]] = self.init_all_valid_actions()
+        valid_actions, linking_scores = self.init_all_valid_actions()
+        self.valid_actions: Dict[str, List[str]] = valid_actions
+
+        # This has shape (number_entities, number_utterance_tokens).
```
matt-gardner: `num_entities`, not `number_entities` (and same for utterance tokens).  `num_` is short for "number of", while `number_` isn't.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -27,34 +51,72 @@ def __init__(self, utterances: List[str], tokenizer=None) -> None:
...
         valid_actions = deepcopy(self.sql_table_context.valid_actions)
-        for string in self.get_strings_from_utterance():
+        linking_scores = []
+        current_tokenized_utterance = [] if not self.tokenized_utterances \
+                else self.tokenized_utterances[-1]
+
+        strings: Set[str] = set()
+        for tokenized_utterance in self.tokenized_utterances:
+            string_linking_dict = get_strings_from_utterance(tokenized_utterance)
+            strings = strings.union(set(string_linking_dict.keys()))
```
matt-gardner: `strings.update(string_linking_dict.keys())`

allennlp/semparse/worlds/atis_world.py
```diff
@@ -27,34 +51,72 @@ def __init__(self, utterances: List[str], tokenizer=None) -> None:
...
+
+        strings_list: List[str] = sorted(strings, reverse=True)
+
+        # We construct the linking scores for strings from the ``string_linking_dict`` here.
+        string_linking_scores = []
+        for string in strings_list:
+            entity_linking = [0 for i in range(len(current_tokenized_utterance))]
```
matt-gardner: Nit: slightly simpler to just do `entity_linking = [0 for token in current_tokenized_utterance]`.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -27,34 +51,72 @@ def __init__(self, utterances: List[str], tokenizer=None) -> None:
...
+        linking_scores = []
+        current_tokenized_utterance = [] if not self.tokenized_utterances \
+                else self.tokenized_utterances[-1]
+
+        strings: Set[str] = set()
+        for tokenized_utterance in self.tokenized_utterances:
+            string_linking_dict = get_strings_from_utterance(tokenized_utterance)
+            strings = strings.union(set(string_linking_dict.keys()))
+
+        strings_list: List[str] = sorted(strings, reverse=True)
```
matt-gardner: Why `reverse=True`?

allennlp/semparse/worlds/atis_world.py
```diff
@@ -27,34 +51,72 @@ def __init__(self, utterances: List[str], tokenizer=None) -> None:
...
+
+        # We construct the linking scores for strings from the ``string_linking_dict`` here.
+        string_linking_scores = []
+        for string in strings_list:
+            entity_linking = [0 for i in range(len(current_tokenized_utterance))]
+            if string in string_linking_dict:
+                for idx in string_linking_dict[string]:
```
matt-gardner: Minor efficiency point:

```python
for token_index in string_linking_dict.get(string, []):
    entity_linking[token_index] = 1
```

This way you don't need the `if`.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -27,34 +51,72 @@ def __init__(self, utterances: List[str], tokenizer=None) -> None:
...
+            string_linking_dict = get_strings_from_utterance(tokenized_utterance)
+            strings = strings.union(set(string_linking_dict.keys()))
+
+        strings_list: List[str] = sorted(strings, reverse=True)
+
+        # We construct the linking scores for strings from the ``string_linking_dict`` here.
+        string_linking_scores = []
+        for string in strings_list:
+            entity_linking = [0 for i in range(len(current_tokenized_utterance))]
+            if string in string_linking_dict:
```
matt-gardner: `string_linking_dict` was only in the for loop above, so this will be the _last_ utterance, not all utterances.  If that's intentional, be sure you mention it in the comment above.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -27,34 +51,72 @@ def __init__(self, utterances: List[str], tokenizer=None) -> None:
...
-                if action not in valid_actions['number']:
-                    valid_actions['number'].append(action)
-
-        return valid_actions
+        numbers = {'0', '1'}
+        number_linking_dict: Dict[str, List[int]] = {}
+
+        for idx, (utterance, tokenized_utterance) in enumerate(zip(self.utterances, self.tokenized_utterances)):
+            number_linking_dict = get_numbers_from_utterance(utterance, tokenized_utterance)
+            numbers = numbers.union(set(number_linking_dict.keys()))
```
matt-gardner: `numbers.update(number_linking_dict.keys())`

allennlp/semparse/worlds/atis_world.py
```diff
@@ -27,34 +51,72 @@ def __init__(self, utterances: List[str], tokenizer=None) -> None:
...
-            for number in numbers:
-                action = format_action('number', number)
-                if action not in valid_actions['number']:
-                    valid_actions['number'].append(action)
-
-        return valid_actions
+        numbers = {'0', '1'}
+        number_linking_dict: Dict[str, List[int]] = {}
+
+        for idx, (utterance, tokenized_utterance) in enumerate(zip(self.utterances, self.tokenized_utterances)):
```
matt-gardner: You're never using `idx` here, so you can just remove the `enumerate`, looks like.  Also, `index` is better than `idx`.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -27,34 +51,72 @@ def __init__(self, utterances: List[str], tokenizer=None) -> None:
...
+            number_linking_dict = get_numbers_from_utterance(utterance, tokenized_utterance)
+            numbers = numbers.union(set(number_linking_dict.keys()))
+        numbers_list: List[str] = sorted(numbers, reverse=True)
+
+        # We construct the linking scores for numbers from the ``number_linking_dict`` here.
+        number_linking_scores = []
+        for number in sorted(numbers_list, reverse=True):
```
matt-gardner: You already sorted the list.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -27,34 +51,72 @@ def __init__(self, utterances: List[str], tokenizer=None) -> None:
...
+            entity_linking = [0 for i in range(len(current_tokenized_utterance))]
```
matt-gardner: Similarly here: `entity_linking = [0 for token in current_tokenized_utterance]`, and the other comments on the for loop.  Makes me wonder if the shared logic could be pulled into a method.  Not sure it's worth it, as the treatment of the two might change.

allennlp/tests/data/dataset_readers/atis_test.py
```diff
@@ -0,0 +1,68 @@
...
+        assert [t.text for t in instance.fields["utterance"].tokens] == \
+                ['show', 'me', 'the', 'one', 'way',
+                 'flights', 'from', 'detroit', 'to',
+                 'westchester', 'county']
+
+        assert isinstance(instance.fields['world'].as_tensor({}), AtisWorld)
+        # Check that the strings in the actions field has the same actions
+        # as the strings in the world.
+        valid_strs = set()
+        for prod in instance.fields['actions'].field_list:
```
matt-gardner: Avoid abbreviations.  Code is meant to be read, and it's way easier on the reader if you use a few more characters and actually write out full words.

allennlp/tests/data/dataset_readers/atis_test.py
```diff
@@ -0,0 +1,68 @@
...
+            if prod.rule.startswith('string'):
+                valid_strs.add(prod.rule)
+
+        world = instance.fields['world'].metadata
+        assert valid_strs == set(world.valid_actions['string'])
+
+        # We make sure all the entities are sorted in reverse order because
+        # we need to enforce some ordering for it to match the correct row
+        # in the linking scores.
+        assert sorted(world.valid_actions['string'], reverse=True) == \
```
matt-gardner: Instead of testing this, just have a test that explicitly tests the linking scores (you didn't actually test the values of it below).  If you're testing values, and have [comments on each line saying which entity is being testing](https://github.com/allenai/allennlp/blob/4fa4dc23ccc5a0f5332d73e8da7688d343296f82/allennlp/tests/data/fields/knowledge_graph_field_test.py#L120-L123), you'll be testing that the alignment is correct.  You don't need to test a particular arbitrary ordering here, just that the linking scores come out right.  Does this make sense?

allennlp/tests/semparse/worlds/atis_world_test.py
```diff
@@ -252,7 +254,6 @@ def test_atis_local_actions(self): # pylint: disable=no-self-use
         assert set(world.valid_actions['number']) == \
             {'number -> ["0"]',
              'number -> ["1"]',
-             'number -> ["2400"]',
```
matt-gardner: Don't you actually want this?  The person didn't specify 12pm.

allennlp/tests/semparse/worlds/atis_world_test.py
```diff
@@ -294,8 +294,6 @@ def test_atis_local_actions(self): # pylint: disable=no-self-use
                  'number -> ["5"]',
```
matt-gardner: Where do 5, 6, and 8 come from?

allennlp/data/dataset_readers/atis.py
```diff
@@ -0,0 +1,122 @@
...
+
+def lazy_parse(text: str):
+    for interaction in text.split("\n"):
+        if interaction:
+            yield json.loads(interaction)
+
+@DatasetReader.register("atis")
+class AtisDatasetReader(DatasetReader):
+    """
+    This ``DatasetReader`` takes json files and converts them into ``Instances`` for the
```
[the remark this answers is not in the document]
kl2806: I'll add a description, since the data is not publicly available.

allennlp/data/dataset_readers/atis.py
```diff
@@ -0,0 +1,122 @@
...
+                utterances = []
+                for current_interaction in line['interaction']:
+                    nl_key = 'utterance' if 'utterance' in current_interaction else 'nl_with_dates'
+
+                    utterances.append(current_interaction[nl_key])
+                    world = AtisWorld(utterances)
+
+                    try:
+                        action_sequence = world.get_action_sequence(current_interaction['sql'])
+                    except ParseError:
```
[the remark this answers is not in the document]
kl2806: The `ParseError` is thrown in the `parsimonious` code, I think we need it because we need some kind of error handling when some query doesn't parse or else it will break. I'm not sure what the alternative for not having this check is, what do you think?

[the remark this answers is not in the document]
matt-gardner: Ah, sorry, I was thinking of a [`ParsingError`](https://github.com/allenai/allennlp/blob/12b74e5a91ef52bea2c1574fc76b52e8f78f86ac/allennlp/semparse/worlds/world.py#L15).  If `parsimonious` will throw this, then yes, you should definitely catch it here.

allennlp/semparse/worlds/atis_world.py
kl2806: I'm setting ``reverse=True`` for consistency with the grammar, since strings are sorted in reverse there. We want this because the parser is greedy (if we have a rule that has multiple options for the right hand side, the first one that succeeds is the one that is used). For example, if we have `1400` in the query and both `1400` and `1` are valid numbers, then we want to try to match it with `1400` first. Otherwise, it will consume `1` and then fail to match anything with `400`.

matt-gardner: Ah, you're catering to `parsimonious`?  That's fine, but giving a note explaining that would be helpful.

allennlp/semparse/worlds/atis_world.py
kl2806: I think we only want the last utterance because if `string` is not in `string_linking_dict` here then that means it was triggered by a previous utterance. Therefore, we don't have any specific token that we want to link it to in the current utterance, we just give it a linking score of all `0`.

allennlp/tests/semparse/worlds/atis_world_test.py
kl2806: `5` comes from `May`, and `8` comes from `August`. Strictly speaking, `6` should not be here but it is showing up because `twenty sixth` appears and it is added when checking each token --  it is triggered by `sixth`.

allennlp/semparse/contexts/atis_tables.py
kl2806: This reason I did not combine these here is because we want to actually construct the ``datetime`` object and use that to look up what day of the week later. To do that we need to know what value `month` and `day` are. If we merge these together, we can't do ``month = ...`` since it's a generic numbers dictionary.

matt-gardner: I'm confused; can you point me to where you're constructing this `datetime` object?  That's not happening here, right?  You're just returning a mapping from `str` to `List[int]`.

kl2806: Oh sorry, you're right! I thought this was a comment on ``get_date_from_utterance``.

kl2806:
@matt-gardner this is ready to be reviewed again, thanks!

allennlp/data/dataset_readers/atis.py
```diff
@@ -0,0 +1,151 @@
...
+        for production_rule in world.all_possible_actions():
+            lhs, _ = production_rule.split(' ->')
+            is_global_rule = not lhs in ['number', 'string']
+            # The whitespaces are not semantically meaningful, so we filter them out.
+            production_rule = ' '.join([token for token in production_rule.split(' ') if token != 'ws'])
+            field = ProductionRuleField(production_rule, is_global_rule)
+            production_rule_fields.append(field)
+
+        action_field = ListField(production_rule_fields)
+        action_map = {action.rule: i # type: ignore
```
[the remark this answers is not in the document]
kl2806: mypy didn't like it when I tried to do that, I think the same thing happed with the Wikitables dataset reader here: https://github.com/allenai/allennlp/blob/master/allennlp/data/dataset_readers/wikitables.py#L301

allennlp/semparse/worlds/atis_world.py
```diff
@@ -107,3 +161,11 @@ def all_possible_actions(self) -> List[str]:
             for action in action_list:
                 all_actions.add(action)
         return sorted(all_actions)
+
+    def __eq__(self, other):
+        if isinstance(self, other.__class__):
+            return (self.valid_actions == other.valid_actions and
```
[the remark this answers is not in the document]
kl2806: Oh yea, that looks better

PR #1578, opened 2018-08-06
Turn off shuffling during evaluation

nelson-liu:


HarshTrivedi:
@nelson-liu  I keep getting this warning during validation evaluation after each epoch training. 
```
allennlp.data.iterators.bucket_iterator - shuffle parameter is set to False, while bucket iterators by definition change the order of your data.
```
Since shuffling for validation has been explicitly turned off, should warning for validation iterator be turned off ?

nelson-liu:
I don't have a strong opinion either way, but you're right that the message isn't too useful now that validation shuffle is always false.

matt-gardner:
I think it's probably fine to move that warning to a comment in the code.

HarshTrivedi:
@matt-gardner That would cause the warning to not show for train iterator as well. Would that be fine?

matt-gardner:
I think so.  The probability of running into the issue that's being warned about is _incredibly_ small, so we should optimize for the much more common case.

Issue #1584, opened 2018-08-07
registering a new dataset reader didn't work 


iamyihwa:
**Describe the bug**

Following the steps in the tutorial ([link](https://github.com/allenai/allennlp/blob/f6d9ba318a643e42bae597fd2b71d336aa290a1f/tutorials/getting_started/using_as_a_library_pt1.md)) didn't quite register the dataset_reader type as was expected. 

![image](https://user-images.githubusercontent.com/12494156/43779262-3aeefb80-9a58-11e8-98d8-a6b0a3580e87.png)

I have made a custon configuration file which was based on the crf_tagger.py , and wrote a custom reader file, however didn't create a separate model but reused crf_tagger model.

Screenshot of the configuration file in json below. 

![image](https://user-images.githubusercontent.com/12494156/43779723-53653bf6-9a59-11e8-821a-0ae923ab2f3a.png)
 

**To Reproduce**

1. Go to
![image](https://user-images.githubusercontent.com/12494156/43779228-213aa9b4-9a58-11e8-97ab-8f909780b704.png)
How I ran the command: 
$ allennlp train crf_es_tagger.json -s tmp/spanish-ner --include-package ner_spanish

: 2018-08-07 13:34:24,461 - INFO - allennlp.common.params - random_seed = 13370
2018-08-07 13:34:24,461 - INFO - allennlp.common.params - numpy_seed = 1337
2018-08-07 13:34:24,461 - INFO - allennlp.common.params - pytorch_seed = 133
2018-08-07 13:34:24,462 - INFO - allennlp.common.checks - Pytorch version: 0.4.0
2018-08-07 13:34:24,464 - INFO - allennlp.common.params - dataset_reader.type = conll2002
Traceback (most recent call last):
  File "/home/ubuntu/anaconda3/envs/allennlp/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/ubuntu/anaconda3/envs/allennlp/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/ubuntu/anaconda3/envs/allennlp/lib/python3.6/site-packages/allennlp/run.py", line 18, in <module>
    main(prog="allennlp")
  File "/home/ubuntu/anaconda3/envs/allennlp/lib/python3.6/site-packages/allennlp/commands/__init__.py", line 70, in main
    args.func(args)
[...]

Issue #1526, opened 2018-07-24
import-package should accept paths as well as pythonpaths


XuezheMax:
`--include-package` fails if you pass it a path - it expects a python module, e.g `scigraph/definition_extraction` vs `scigraph.definition_extraction`.

schmmd:
Closing--as that's the intended behavior.  Feel free to re-open if the error is particularly confusing.

DeNeutoy:
This is something we should change, re-opening

schmmd:
@DeNeutoy I'm pretty sure we discussed this as a group in triage and thought that we should maintain the current behavior.  I don't personally have very strong opinions on this, but would you justify why you think we should change the current behavior?

PR #1586, opened 2018-08-07
make --include-package include all submodules

joelgrus:
fix for the first bug in  #1584


nelson-liu:
+1 to mark's suggestion, and people definitely get confused (see https://github.com/allenai/allennlp/issues/1526)

PR #1587, opened 2018-08-07
copy bin/ into image

### LINKED ISSUE #1582: Documentation bug - instructions for running allennlp in docker wrong
In the top level README.md for this project (https://github.com/allenai/allennlp/blob/v0.5.0/README.md), it says, "Note that if you use the source installation, you won't be able to use the allennlp command but rather you'll need to run ./bin/allennlp."

However, there is no bin/allennlp anyhwere in the container. Instead, from /stage/allennlp, you run `python allennlpy run.py`.

DeNeutoy:
Our instructions for command line usage didn't work in a docker image because we didn't copy `bin/` into the image.
Fixes #1582

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,6 +44,7 @@ COPY requirements.txt .
 COPY scripts/install_requirements.sh scripts/install_requirements.sh
 RUN ./scripts/install_requirements.sh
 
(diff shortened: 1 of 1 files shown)


schmmd:
Thanks--good call.

PR #1598, opened 2018-08-10
Log number of parameters in optimizers

nelson-liu:
Logs the number of parameters used when building optimizers `from_params`

### DIFF
--- a/allennlp/training/optimizers.py
+++ b/allennlp/training/optimizers.py
@@ -112,6 +112,14 @@ class Optimizer(Registrable):
         else:
             parameter_groups = [param for name, param in model_parameters]
 
+        # Log the number of parameters to optimize
+        num_parameters = 0
+        for parameter_group in parameter_groups:
+            if isinstance(parameter_group, dict):
+                num_parameters += sum(parameter.numel() for parameter in parameter_group["params"])
+            else:
+                num_parameters += parameter_group.numel()
+        logger.info("Number of trainable parameters: %s", num_parameters)
         return Optimizer.by_name(optimizer)(parameter_groups, **params.as_dict()) # type: ignore
 
 # We just use the Pytorch optimizers, so here we force them into


allennlp/training/optimizers.py
```diff
@@ -112,6 +112,14 @@ def from_params(cls, model_parameters: List, params: Params):  # type: ignore
         else:
             parameter_groups = [param for name, param in model_parameters]
 
+        # Log the number of parameters to optimize
+        num_parameters = 0
+        for parameter_group in parameter_groups:
+            if isinstance(parameter_group, dict):
+                num_parameters += sum(parameter.numel() for parameter in parameter_group["params"])
```
matt-gardner: Do you need to check for `trainable` somewhere in here?

nelson-liu: I opted not to, since the logic for that check lives in the trainer. As it's currently provided, all of the parameters in `parameter_groups` get used in the Optimizer --- if some of the input parameters don't `require_grad`, I think the optimizer complains anyway.

matt-gardner: Ah, ok, great.  This looks good to me, then, but I'd suggest @DeNeutoy also takes a quick look (it's a small PR), just to be sure I'm not missing something.

HarshTrivedi:
Would `train.py` be better place for this kind of logging? .. something like:

```python
                   get_frozen_and_tunable_parameter_names(model)
        logger.info(name)
        logger.info(name)
    # here ...
    tunable_parameters_count = sum([parameter.numel()
                                    for name, parameter in model.named_parameters()
                                    if name in tunable_parameter_names])
    logger.info("Number of tunable parameters: %d", tunable_parameters_count)
```
Since in `train.py` we are already logging **names** of tunable parameters, logging **number** of tunable parameters` seems like natural follow up.

nelson-liu:
Yup, I considered that as well, but I ended up deciding that this seems more like a property of the Optimizer --- how many parameters is the optimizer going to adjust?

In addition, this seems like a useful thing to have if you're using the AllenNLP `Trainer` code separate from `allennlp train` --- I don't think putting it in `train.py` would make it show up if you're just using `Trainer`?

PR #1601, opened 2018-08-13
filter out numpy 'size changed' warnings

joelgrus:
@schmmd suggested pinning to an early numpy version, but I'd rather not cause problems with people who might need the later version for some reason.


joelgrus:
the only way I know to reliably trigger it is to import `spacy`.

so if you had code that imported `spacy` before importing anything allennlp this wouldn't fix it, but in that case the only real solution is to do this before you import spacy

PR #1604, opened 2018-08-14
Fix use of scalar tensors in ConllCorefScores

### LINKED ISSUE #1545: Coreference model trained with Allennlp v0.5.1 gets small scores
Hello!

I trained the coreference model with Allennlp 0.5.1 and get F1 score around 0.09 which is far smaller than the score in the paper (about 0.68).

For Allennlp 0.5.1, Python 3.6, Pytorch 0.4.0 are required. I ran it on GPU Tesla P100. It works well when the model is trained with v0.4.3. 

Could you please tell me your performance of the coreference model trained with v0.5.1? I don't know whether it exists a problem somewhere.
Thanks!

matt-gardner:
Fixes #1545.

@nelson-liu, can you run a quick check on this branch, to make sure this actually gives similar scores to pre-pytorch-0.4 in the first epoch of training?  From my test, I'm reasonably sure that this fixes the issue, but I don't have this model already set up to train, so I haven't checked that.

### DIFF
--- a/allennlp/tests/training/metrics/conll_coref_scores_test.py
+++ b/allennlp/tests/training/metrics/conll_coref_scores_test.py
@@ -1,7 +1,5 @@
 # pylint: disable=no-self-use,invalid-name,protected-access
 import torch
-import pytest
-import numpy
 
 from allennlp.common.testing import AllenNlpTestCase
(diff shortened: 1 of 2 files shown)


nelson-liu:
Yes, this seems to have fixed the issue. nice work!

PR #1606, opened 2018-08-14
Tree decoding fix

DeNeutoy:


allennlp/models/biaffine_dependency_parser.py
```diff
@@ -490,9 +490,22 @@ def _mst_decode(self,
...
+
+        # Although we need to include the root node so that the MST includes it,
+        # we do not want any word to be the parent of the root node.
+        # Here, we enforce this by setting the scores for all word -> ROOT edges
+        # edges to be very negative.
+        normalized_arc_logits[:, 0, :] = -1e8
+
         # Shape (batch_size, num_head_tags, sequence_length, sequence_length)
+        # This energy tensor expresses the following relation:
+        # energy[i,j] = "Score that j is the head of i". In this
```
matt-gardner: I think something's backwards in either this comment or the logic above.  Because if I substitute `j` for `ROOT` and `i` for `some_word`, I get "Score that `ROOT` is the head of `some_word`", which you set to very negative above, with `energy[:, 0, :] = -1e8`.  Right?

allennlp/tests/predictors/biaffine_dependency_parser_test.py
```diff
@@ -25,7 +25,7 @@ def test_uses_named_inputs(self):
         assert head_tags is not None
         assert isinstance(head_tags, list)
         assert all(isinstance(x, int) for x in head_tags)
-
+        print(result)
```
matt-gardner: Remove print.

allennlp/tests/models/biaffine_dependency_parser_test.py
```diff
@@ -31,3 +33,30 @@ def test_decode_runs(self):
...
+        #               A(index 0)
+        #               /        \
+        #              \/        \/
+        #          B(index 1)   C(index 2)
+        length = torch.LongTensor([3])
+        heads, _ = self.model._run_mst_decoding(energy, length) # pylint: disable=protected-access
+        # This is the correct MST, but not desirable for dependency parsing.
+        assert heads.tolist()[0] == [-1, 0, 0]
+
+        energy[:, :, 0, :] = 0
```
matt-gardner: Is this example really doing what you think it's doing?

```python
>>> t = torch.Tensor([[0, 1, 1], [10, 0, 1], [10, 1, 0]]).view(1, 1, 3, 3)
>>> t
tensor([[[[  0.,   1.,   1.],
          [ 10.,   0.,   1.],
          [ 10.,   1.,   0.]]]])
>>> t[:, :, 0, :] = 0
>>> t
tensor([[[[  0.,   0.,   0.],
          [ 10.,   0.,   1.],
          [ 10.,   1.,   0.]]]])
```

This is zeroing out the top row, not the first column.  Is that what you expected?  And in general, I'd recommend using different numbers for every non-zero value, so there are no ties and it's more obvious what the MST should be.

allennlp/models/biaffine_dependency_parser.py
matt-gardner: I'm just saying there's a mismatch between your comment and your code.  Oh, wait, no, I was confusing myself because of the batch dimension.  When I wrote my comment, I was thinking it was `normalized_arc_logits[i, j, :]`, so `j` was `ROOT`, but it's actually `normalized_arc_logits[:, i, :]`, so `i` is `ROOT`.  Ok, all good.

PR #1615, opened 2018-08-16
Add IOB1 as allowed CRF constraint

nelson-liu:
This adds IOB1 coding as an allowed constraint for the CRF tagger.

Here's a table of allowed transitions for reference:

### DIFF
--- a/allennlp/modules/conditional_random_field.py
+++ b/allennlp/modules/conditional_random_field.py
@@ -88,29 +88,16 @@ def is_transition_allowed(constraint_type: str,
     ``bool``
         Whether the transition is allowed under the given ``constraint_type``.
     """
+    # pylint: disable=too-many-return-statements
     if to_tag == "START" or from_tag == "END":
         # Cannot transition into START or from END
         return False
-    if from_tag == "START":
-        if constraint_type == "BIOUL":
-            return to_tag in ('O', 'B', 'U')
-        elif constraint_type == "BIO":
-            return to_tag in ('O', 'B')
-        elif constraint_type == "IOB1":
-            return to_tag in ('O', 'I')
-        else:
-            raise ConfigurationError(f"Unknown constraint type: {constraint_type}")
-    if to_tag == "END":
-        if constraint_type == "BIOUL":
-            return from_tag in ('O', 'L', 'U')
-        elif constraint_type == "BIO":
-            return from_tag in ('O', 'B', 'I')
-        elif constraint_type == "IOB1":
-            return from_tag in ('O', 'B', 'I')
-        else:
-            raise ConfigurationError(f"Unknown constraint type: {constraint_type}")
 
     if constraint_type == "BIOUL":
+        if from_tag == "START":
+            return to_tag in ('O', 'B', 'U')
+        if to_tag == "END":
+            return from_tag in ('O', 'L', 'U')
         return any([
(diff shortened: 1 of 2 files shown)


allennlp/modules/conditional_random_field.py
```diff
@@ -84,6 +85,31 @@ def allowed_transitions(constraint_type: str, tokens: Dict[int, str]) -> List[Tu
...
+                        to_bio in ('O', 'I'),
+                        # Can only transition to B-x from B-x or I-x, where
+                        # x is the same tag.
+                        to_bio == 'B' and from_bio in ('B', 'I') and from_entity == to_entity
+                ])
+
+                if is_allowed:
+                    allowed.append((i, j))
+
+        # start transitions
```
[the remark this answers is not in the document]
matt-gardner: Oh, actually, not quite.  I didn't realize that the blocks are slightly different.  Ok, here's another proposal:

```python
num_tags = len(tokens)
tags = [(index, tag) for index, tag in tokens.items()] + [(num_tags, "START"), (num_tags + 1, "END")]
for i, from_tag in tags:
    for j, to_tag in tags:
        if is_transition_allowed(constraint_type, from_tag, to_tag):
            allowed.append(i, j)
return allowed
```

And then you take all of the `is_allowed = any([...])` logic and move it to that other method, and add in the start and end logic there.  That way it's all in one place, and the higher-level logic is consistent for all constraint types.  This is just getting very complicated as it is with every additional constraint type.

nelson-liu:
ok @matt-gardner , this is ready to be looked at again. let me know if i misconstrued your vision...

allennlp/modules/conditional_random_field.py
```diff
@@ -9,86 +9,135 @@
...
+    ----------
+    constraint_type : ``str``, required
+        Indicates which constraint to apply. Current choices are
+        "BIO", "IOB1", and BIOUL".
+    from_tag : ``str``, required
+        The tag that the transition originates from. For example, if the
+        label is ``I-PER``, the ``from_tag`` is ``I``.
+    from_entity: ``str``, required
+        The entity corresponding to the ``from_tag``. For example, if the
+        label is ``I-PER``, the ``from_entity`` is ``PER``.
```
matt-gardner: Technically, because you're using `from_label[1:]`, you'll actually get `-PER`, not `PER`.  But, that probably doesn't matter...

allennlp/modules/conditional_random_field.py
```diff
@@ -9,86 +9,135 @@
...
-                allowed.append((i, end_tag))
+    Returns
+    -------
+    ``bool``
+        Whether the transition is allowed under the given ``constraint_type``.
+    """
+    if to_tag == "START" or from_tag == "END":
+        # Cannot transition into START or from END
+        return False
+    if from_tag == "START":
```
matt-gardner: Nit: you've grouped "START" and "END" together here, and I now have three places to look in the code to see all of the "BIOUL" constraints.  Seems more coherent to instead have a `if from_tag == "START"` block inside of each `if constraint_type == ...` block below.

allennlp/modules/conditional_random_field.py
nelson-liu: Yeah, i opted to do this since the original behavior with `(tag, *entity)` meant `entity = ['-', 'P', 'E', 'R']`

PR #1619, opened 2018-08-16
Parser decoding fix 2

DeNeutoy:
Annoyed that I missed this yesterday, in retrospect it was obvious that this would break the label assignment

### DIFF
--- a/allennlp/models/biaffine_dependency_parser.py
+++ b/allennlp/models/biaffine_dependency_parser.py
@@ -491,12 +491,6 @@ class BiaffineDependencyParser(Model):
         # Shape (batch_size, sequence_length, sequence_length)
         normalized_arc_logits = F.log_softmax(attended_arcs, dim=2).transpose(1, 2)
 
-        # Although we need to include the root node so that the MST includes it,
-        # we do not want any word to be the parent of the root node.
-        # Here, we enforce this by setting the scores for all word -> ROOT edges
-        # edges to be very negative.
-        normalized_arc_logits[:, 0, :] = -1e8
-
         # Shape (batch_size, num_head_tags, sequence_length, sequence_length)
         # This energy tensor expresses the following relation:
         # energy[i,j] = "Score that j is the head of i". In this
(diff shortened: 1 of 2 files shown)


allennlp/models/biaffine_dependency_parser.py
```diff
@@ -508,8 +502,25 @@ def _mst_decode(self,
...
+            # edges to be 0.
+            scores[0, :] = 0
+            # Decode the heads. Because we modify the scores to prevent
+            # adding in word -> ROOT edges, we need to find the labels ourselves.
+            head, _ = decode_mst(scores.numpy(), length, has_labels=False)
+
+            # We don't care what the head is for the root token, but by default it's
+            # returned as -1 - this messes up the indexing into the tag_ids tensor,
+            # so we'll just fix it to have a value of 0.
+            head[0] = 0
```
matt-gardner: Nit: I'd recommend `instance_heads` here, as this is still a list.  Same for `instance_head_tags`.

allennlp/tests/models/biaffine_dependency_parser_test.py
```diff
@@ -53,10 +54,25 @@ def test_mst_respects_no_outgoing_root_edges_constraint(self):
...
         assert heads.tolist()[0] == [-1, 0, 1]
+
+    def test_mst_decodes_arc_labels_with_respect_to_unconstrained_scores(self):
+        energy = torch.Tensor([[0, 2, 1],
+                               [10, 0, 0.5],
+                               [9, 0.2, 0]]).view(1, 1, 3, 3).expand(1, 2, 3, 3).contiguous()
+
+        # Make the score for the root label for arcs to the root token be higher - it
+        # will be masked for the MST, but we want to make sure that the tags are with
+        # respect to the unmasked tensor.
```
matt-gardner: I don't really understand this comment or this test, but I'll trust you.

DeNeutoy: Basically, there was a problem when I masked out the scores per tag for the root node. The edge weights for the MST are just the max over the labels, but for actually choosing a label, we just take the argmax from over the labels for the given arc - so for the word which is the root of the tree (i.e has the root as it's parent), all of the label scores were zero - so it was arbitrarily picking the first one. I needed to separate the choosing of the labels to use the original unmasked scores, and only use the masked scores for decoding the arcs. Can I make that comment better?

matt-gardner: I guess if you make it clear in the comment that this is an issue for the (ROOT, root_token) edge, that would help.  Except I thought you zeroed out all of the (*, ROOT) edges, so why is (ROOT, root_token) affected...?

DeNeutoy: Oh wow now I have to think about this _again_

DeNeutoy: https://github.com/allenai/allennlp/pull/1619/files#diff-9ebf9be01985b8f2d35c515d6624b0ceR519 This actually means the score matrix is [parent -> child], grrr.

PR #1620, opened 2018-08-16
Allow files to be downloaded from S3

### LINKED ISSUE #1618: Feature request: support pulling data from AWS S3 buckets directly (using boto3)
Sometimes we have to pull data files from a private S3 bucket, which requires AWS credentials, so we can't just access it with a simple HTTP GET request. To handle this, we could use `boto3` or the `awscli` to pull s3 files when the file is specified by an s3 path, like `s3://path-to-file`, instead of the http url. At this time it seems that the alternative is to just manually download the files locally first using the `awscli` or `boto3` and then point allennlp to the local copy.

The only downside I can see of supporting this feature is adding an additional dependency (`boto3`).

I'd be willing to submit a PR if this is something you want to support.

epwalsh:
Fixes #1618


allennlp/common/file_utils.py
```diff
@@ -92,6 +95,73 @@ def cached_path(url_or_filename: Union[str, Path], cache_dir: str = None) -> str
         raise ValueError("unable to parse {} as a URL or as a local path".format(url_or_filename))
 
 
+def split_s3_path(url: str) -> Tuple[str, str]:
+    """Split a full s3 path into the bucket name and path."""
```
joelgrus: would it be better to use `urlparse` like here: https://stackoverflow.com/a/42641363/1076346 ?

(I don't know, it's just that logic like "split on slashes and take the third and fourth parts"  makes me nervous)

requirements.txt
```diff
@@ -43,6 +43,9 @@ cffi==1.11.2
 # aws commandline tools for running on Docker remotely.
 awscli>=1.11.91
 
+# Accessing files from S3 directly.
+boto3
```
joelgrus: you'll need to add this requirement to `setup.py` as well

allennlp/tests/common/file_utils_test.py
```diff
@@ -97,6 +101,43 @@ def test_url_to_filename_with_etags_eliminates_quotes(self):
...
+            missing_file_request(fake_url)
+
+        # Request that fails due to other reasons.
+        @s3_request
+        def other_bad_request(url):
+            raise ClientError({"Error": {"Code": 503}}, "s3_request to {}".format(url))
+
+        with pytest.raises(ClientError):
+            other_bad_request(fake_url)
+
```
joelgrus: I have not used boto much, is there a way of testing it that uses a mock s3 to test the actual calls, or is that not done?

allennlp/tests/common/file_utils_test.py
```diff
@@ -4,10 +4,14 @@
...
 
+from botocore.exceptions import ClientError
 import pytest
 import responses
 
-from allennlp.common.file_utils import url_to_filename, filename_to_url, get_from_cache, cached_path
+from allennlp.common.file_utils import url_to_filename, filename_to_url, get_from_cache, \
+    cached_path, \
+    split_s3_path, \
+    s3_request
```
joelgrus: we tend to prefer using parenthesis rather than backslashes for line continuations

```
from allennlp.common.file_utils import (url_to_filename, ... 
)
```

allennlp/tests/common/file_utils_test.py
epwalsh: Yea there might be a better way of doing this, like here: https://stackoverflow.com/questions/37143597/mocking-boto3-s3-client-method-python. I'll look into it and see if I can improve those tests

joelgrus: if it's a ton of work it's probably not worth it, but if it's relatively easy I think it's a good test to have

requirements.txt
epwalsh: Done, also added `moto` to requirements.txt and setup.py

allennlp/common/file_utils.py
epwalsh: Got it, just changed that

allennlp/tests/common/file_utils_test.py
```diff
@@ -52,19 +53,21 @@ def head_callback(_):
...
 
-@mock_s3
+def set_up_s3_bucket(bucket_name: str = "my-bucket", s3_objects: List[Tuple[str, str]] = None):
+    s3_client = boto3.client("s3")
+    s3_client.create_bucket(Bucket=bucket_name)
+    for filename, key in s3_objects or []:
+        s3_client.upload_file(Filename=filename, Bucket=bucket_name, Key=key)
+
+
+#  @mock_s3
```
epwalsh: Having the decorator at the class level had unintended consequences for other tests

epwalsh:
Hmm, it seems that last test failed due to a (temporary) http connection/timeout error. I ran locally within Docker and everything passed. Can we re-run the tests?

nelson-liu:
@epwalsh sure, just restarted the job.

epwalsh:
@nelson-liu thanks!! Anything else I should add/change @joelgrus ?

joelgrus:
looks good, thanks for doing this

PR #1622, opened 2018-08-17
Don't error if fine-tune serialization dir already exists

nelson-liu:
Changing this to match the behavior of other commands, like [`train`](https://github.com/allenai/allennlp/blob/82686c10bdca062a7e41825000c39372de354479/allennlp/commands/train.py#L222) and [`make-vocab`](https://github.com/allenai/allennlp/blob/1402b7c02aede1a8f832f64cdfb2e9ef1157faca/allennlp/commands/make_vocab.py#L83). 

This causes `fine-tune` to fail with beaker at the moment.

### DIFF
--- a/allennlp/commands/fine_tune.py
+++ b/allennlp/commands/fine_tune.py
@@ -141,7 +141,7 @@ def fine_tune_model(model: Model,
         down tqdm's output to only once every 10 seconds.
     """
     prepare_environment(params)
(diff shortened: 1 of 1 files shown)


joelgrus:
do we want to still fail if it's non-empty? I believe that's what `train` does

nelson-liu:
good idea, i added that

PR #1628, opened 2018-08-18
Initial event2mind reimplementation

brendan-ai2:
- Forks SimpleSeq2Seq.
- New metric SequenceAccuracy.
- Computes xintent, xreact and oreact.


brendan-ai2:
@rlebras, here is my current code. Like I said in slack, this is a work in progress, but by all means let me know what you think.

rlebras:
Great, thanks a lot Brendan!

PR #1632, opened 2018-08-20
Make LinearAttention and LinearMatrixAttention memory-efficient

matt-gardner:
This removes the need for the `TriLinearAttention` that Chris implemented for #804, as our `LinearMatrixAttention` now uses memory-efficient operations.

cc @eunsol

### DIFF
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -656,6 +656,72 @@ def _get_combination(combination: str, tensors: List[torch.Tensor]) -> torch.Ten
             raise ConfigurationError("Invalid operation: " + operation)
 
 
+def weighted_tensor_combination(combination: str,
+                                tensors: List[torch.Tensor],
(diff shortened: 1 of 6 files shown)


allennlp/nn/util.py
```diff
@@ -656,6 +656,88 @@ def _get_combination(combination: str, tensors: List[torch.Tensor]) -> torch.Ten
             raise ConfigurationError("Invalid operation: " + operation)
 
 
+def weighted_tensor_combination(combination: str,
```
DeNeutoy: It's confusing to call this `weighted_tensor_combination` - in the context of attention I think of the weighted combination as the `sum_i (a_i * h_i)` where a_i is the normalised attention distribution and h_i is the vector. Can you call it something else? `combine_tensors_and_multiply`?

matt-gardner: Yeah, I agree.

Issue #1635, opened 2018-08-21
move trainer functionality from `from_params` to constructor, so it's documented


nelson-liu:
I just came upon (purely by accident!) this very useful-looking feature that takes regexes with parameter names and sets them to be frozen accordingly.

https://github.com/allenai/allennlp/blob/4df4638b186554c994e7fd841a766e992ee81ba1/allennlp/commands/train.py#L294-L297

However, this feature does not seem to be documented anywhere --- it's not mentioned in the trainer command (which makes sense, because it's part of the config) and it's not mentioned in the `Trainer` class (which makes sense, because it's not being passed into `Trainer` anywhere). 

Not sure what the best way to proceed is, but this seems like something that would be quite useful to many (in fact, @dangitstam was asking me about something exactly like this the other day).

matt-gardner:
This is now in `TrainerPieces.from_params`, and I'm pretty sure there's not a lot we can do about this, except document it somewhere.  @joelgrus, is there any other way of handling this?

matt-gardner:
This is covered by #3389.

PR #1636, opened 2018-08-21
Discriminative fine-tuning, gradual unfreezing, slanted triangular learning rates (ULMFiT)

sebastianruder:
This is a PR to add methods from [ULMFiT](https://arxiv.org/abs/1801.06146) to AllenNLP so as to allow a direct comparison.

@matt-peters, could you advise with regard to the kind of tests to be added and stylistic changes to be made?


allennlp/training/trainer.py
```diff
@@ -980,11 +980,34 @@ def from_params(cls,
         grad_norm = params.pop_float("grad_norm", None)
         grad_clipping = params.pop_float("grad_clipping", None)
         lr_scheduler_params = params.pop("learning_rate_scheduler", None)
+        fine_tuning = params.pop("fine-tuning", None)
+        optimizer_params = params.pop("optimizer")
+        if fine_tuning == "discriminative" or \
```
matt-gardner: It's not great to have all of this logic inside of a `from_params` method.  I'm not sure that there's a great place for it, though, as you're modifying other parameters based on a flag.  At the very least, you could pull this logic out into a `create_parameter_groups()` method, or similar, and then you could test that method in a simple unit test.

allennlp/training/trainer.py
```diff
@@ -980,11 +980,34 @@ def from_params(cls,
         grad_norm = params.pop_float("grad_norm", None)
         grad_clipping = params.pop_float("grad_clipping", None)
         lr_scheduler_params = params.pop("learning_rate_scheduler", None)
+        fine_tuning = params.pop("fine-tuning", None)
```
matt-gardner: If this is a boolean flag, as it looks like it is, it's better to just make it boolean, like `discriminative-fine-tuning`, unless we really think that in the near future we're going to have other possible values for the `fine-tuning` flag.

matt-gardner: Also, this isn't your fault, but it's concerning to me that there's no natural place to document this new parameter (c.f. @nelson-liu's find in #1635).  @joelgrus, @DeNeutoy, any good ideas on how to fix this documentation problem?

allennlp/training/learning_rate_schedulers.py
```diff
@@ -126,6 +126,100 @@ def get_lr(self):
 
         return [scale for _ in range(len(self.base_lrs))]
 
+
+class STLR(torch.optim.lr_scheduler._LRScheduler): # pylint: disable=protected-access
```
matt-gardner: Maybe call this `SlantedTriangular`?

allennlp/training/learning_rate_schedulers.py
```diff
@@ -135,4 +229,5 @@ def get_lr(self):
         "reduce_on_plateau": torch.optim.lr_scheduler.ReduceLROnPlateau,
         "cosine": torch.optim.lr_scheduler.CosineAnnealingLR,
         "noam": NoamLR,
+        "stlr": STLR
```
matt-gardner: We already know this is a learning rate, because of the key that's used to get to this name, so "LR" isn't super helpful as a part of the name (see the other names above).  So I'd just use `slanted_triangular` as the name here, and `SlantedTriangular` or `SlantedTriangularLR` as the class name.

allennlp/training/learning_rate_schedulers.py
```diff
@@ -126,6 +126,100 @@ def get_lr(self):
...
+            num_steps = self.num_epochs * self.num_steps_per_epoch - steps_while_frozen
+            step = max(self.last_epoch, 1) - steps_while_frozen
+        cut = int(num_steps * self.cut_frac)
+        if cut == step:
+            print()
+        p = step / cut if step < cut else 1 - (step - cut) / (num_steps - cut)
+        lrs = [lr * (1 + p * (self.ratio - 1)) / self.ratio for lr in self.base_lrs]
+        return lrs
+
 # We just use the Pytorch LRSchedulers, so here we force them into
```
matt-gardner: This is maybe a bit annoying for you to have to do @sebastianruder, so we can just open an issue instead of fixing this in this PR, but it seems like we've got enough learning rate schedulers that we should separate this class out into a module, and give `NoamLR` and your scheduler their own files.  Then the built-in pytorch schedulers can still live here in this dictionary, but the classes we implement can just use `@LearningRateScheduler.register()` as normal.

allennlp/training/learning_rate_schedulers.py
matt-gardner: For tests, I would recommend having a simple unit test for each of the methods on this class, and a test that makes sure this can be instantiated correctly with `LearningRateScheduler.from_params(optimizer, some_param_object)`.

allennlp/training/learning_rate_schedulers.py
```diff
@@ -126,6 +126,100 @@ def get_lr(self):
...
+            # the method is called once when initialising before the
+            # first epoch (epoch 0) and then always at the end of each
+            # epoch; so the first time, with epoch id 0, we want to set
+            # up for epoch #1; the second time, still with epoch id 0,
+            # we want to set up for epoch #2, etc.
+            epoch_no = epoch+1 if self.first_epoch else epoch+2
+            if self.first_epoch:
+                self.first_epoch = False
+            if epoch_no >= len(self.optimizer.param_groups)-1:
+                print('Gradual unfreezing finished. Training all layers.')
```
matt-gardner: Use `logger.info()` instead of `print`.

allennlp/training/learning_rate_schedulers.py
```diff
@@ -126,6 +126,100 @@ def get_lr(self):
...
+    def step(self, epoch=None):
+        if self.gradual_unfreezing:
+            assert not self.optimizer.param_groups[-1]["params"], \
+                "The default group should be empty."
+            # the method is called once when initialising before the
+            # first epoch (epoch 0) and then always at the end of each
+            # epoch; so the first time, with epoch id 0, we want to set
+            # up for epoch #1; the second time, still with epoch id 0,
+            # we want to set up for epoch #2, etc.
+            epoch_no = epoch+1 if self.first_epoch else epoch+2
```
matt-gardner: There are some pylint style issues you need to fix.  You can see these on the TeamCity link, and you can see them locally by running `./scripts/pylint.sh`.  You'll also need to add this class to our docs [here](https://github.com/allenai/allennlp/blob/master/doc/api/allennlp.training.learning_rate_schedulers.rst).

allennlp/training/trainer.py
nelson-liu: To add: it’d be great if the key in the param name were “fine_tuning” instead of “fine-tuning” for consistency with the rest of the codebase / that hyphen has surprisingly caused me some issues.

joelgrus: anything that is popped and consumed (and never mentioned again) inside a `from_params` is sort of intrinsically undocumented. my bias would be to just avoid doing that.

e.g. in #1635 since `no_grad_regexes` is a trainer param, it should be a constructor parameter of `Trainer` and turning those gradients off should happen in the Trainer constructor, not right before you call the Trainer constructor.

similarly here (although I haven't wrapped my head fully around this code yet), if there are parameters that are relevant to how the optimizer is instantiated, they should be constructor parameters of `Optimizer` (which isn't really a base class, but if we really need so much logic in it it probably should be).

this is not just an issue of documentation, IMO we should aim not to have functionality in the library that's only available when constructing things `from_params` but not when constructing them directly

allennlp/training/trainer.py
joelgrus: see above, I would rather this logic live inside the `Optimizer` constructor (modulo the fact that there is no such constructor, maybe we need a Wrapper class like we do for `Seq2Seq` etc...)

joelgrus: also, one stylistic comment, we generally try to avoid using `\` for line continuations, preferring to use parenthesis for expressions that need to span multiple lines

matt-gardner: So, for this PR and @sebastianruder's contribution here, should we pull this out into a function and try to fix the broader issue later?  I don't think Sebastian can or wants to solve the mess we've made of `Trainer.from_params`.

allennlp/training/trainer.py
matt-gardner: Yeah, that's a good point, that we shouldn't have functionality that's only accessible through `from_params`.  The trouble is that there are a bunch of things that interact in the `Trainer` that are best constructed outside of it, not inside.  Maybe those things belong in the `Train` command...?  This one needs some more thought, I think.

allennlp/training/trainer.py
joelgrus: yeah, that shouldn't be part of this PR, I think pulling it out into a function (and adding some clarifying comments) is good for now. 

fixing the from_params mess is a broader issue

allennlp/training/trainer.py
```diff
@@ -980,11 +980,34 @@ def from_params(cls,
...
+            else:
+                # split up parameters in groups based on modules
+                modules = [[m] for m in model._modules]
+            if fine_tuning == "discriminative":
+                # use a different learning rate for every parameter group
+                assert "lr" in optimizer_params, \
+                    "Error: A base learning rate is required for " \
+                    "discriminative fine-tuning."
+                lr = optimizer_params["lr"]
+                decay_factor = 0.38
```
matt-peters: Let's pull this out into a configurable parameter in `optimizer_params` with this as the default as deeper / shallower models may need a different decay factor.

allennlp/training/trainer.py
DeNeutoy: I think the best way to fix this (not for this PR) would be to build the optimizer without modifying it's parameters, and then modify the `parameter_groups` of the optimizer directly inside the learning rate scheduler, as the optimizer is passed to the LR scheduler as an argument. This would isolate the implementation to where it is actually used (i.e in the special scheduler) and also solve the "you can't use this unless you build from params" problem, which is something we must avoid.

matt-gardner: Is "discriminative fine tuning" just a property of the learning rate scheduler, or is it a property of something else?  If it really is just a property of the scheduler, then I definitely agree with @DeNeutoy, and I'd say that this change _does_ belong in this PR.

matt-gardner: The only issue is if you can't update the parameters groups of an optimizer after-the-fact.  I don't know enough about pytorch optimizers to know if this would break things.

sebastianruder:
This is great! Thanks a lot for your feedback! I'll add changes to address your points early tomorrow (GMT+1).

allennlp/training/learning_rate_schedulers.py
sebastianruder: Done.

allennlp/training/learning_rate_schedulers.py
sebastianruder: Done.

allennlp/training/learning_rate_schedulers.py
sebastianruder: Cool. Let me know if you'd like me to do this here or if you prefer to create a separate issue.

allennlp/training/learning_rate_schedulers.py
sebastianruder: What's the best way to access the logger in a learning rate schedule class?

allennlp/training/learning_rate_schedulers.py
matt-gardner: Whether it's you that does it or someone else, it's probably better done in a separate PR.

allennlp/training/learning_rate_schedulers.py
matt-gardner: You need to import logging and have a line like this near the top of the file: https://github.com/allenai/allennlp/blob/1d431885b436fb0a6bebaf5478030f033bca0a07/allennlp/data/dataset_readers/dataset_reader.py#L9

allennlp/training/learning_rate_schedulers.py
matt-gardner: Seems reasonable in that case to add `# pylint: disable=protected-access` at the end of the line.

allennlp/training/learning_rate_schedulers.py
```diff
@@ -126,6 +126,99 @@ def get_lr(self):
...
+            # first epoch (epoch 0) and then always at the end of each
+            # epoch; so the first time, with epoch id 0, we want to set
+            # up for epoch #1; the second time, still with epoch id 0,
+            # we want to set up for epoch #2, etc.
+            num_layers_to_unfreeze = epoch+1 if self.first_epoch else epoch+2
+            if self.first_epoch:
+                self.step_batch(1)
+                self.first_epoch = False
+            if num_layers_to_unfreeze >= len(self.optimizer.param_groups)-1:
+                print('Gradual unfreezing finished. Training all layers.')
```
nelson-liu: it'd be good if these print statements were `logger.info` calls instead.

sebastianruder: Done.

allennlp/training/trainer.py
sebastianruder: I agree with @DeNeutoy. I've moved the discriminative fine-tuning now from `trainer.py` to the learning rate scheduler (see https://github.com/allenai/allennlp/pull/1636/commits/a81040279d701163d50f8798631c1faec44f7c16). This is a lot cleaner and a lot more intuitive.

allennlp/training/trainer.py
sebastianruder: I assume this is addressed by the last commit moving this functionality into `learning_rate_schedulers.py` (https://github.com/allenai/allennlp/pull/1636/commits/a81040279d701163d50f8798631c1faec44f7c16).

sebastianruder:
I've moved the discriminative fine-tuning part into `learning_rate_schedulers.py`, which is a lot cleaner (thanks to @DeNeutoy for the suggestion).
The main outstanding issue is that in order for the discriminative fine-tuning to make sense, the parameters need to be split into groups in the `Optimizer`. This can already be done in the Optimizer using regex (see [here](https://github.com/allenai/allennlp/blob/a81040279d701163d50f8798631c1faec44f7c16/allennlp/training/optimizers.py#L57)) but this is model-dependent. It would be nice if this could be done automatically. To this end, I'm currently splitting the parameters based on the pre-specified modules of the model, which seems to make sense (see [here](https://github.com/allenai/allennlp/pull/1636/files#diff-043dcd121296c3cef3f3ff8c74127ff1R995)) as this essentially uses the same modules that would be specified manually. The downside is that it requires accessing the protected `._modules` (which TeamCity doesn't like). Let me know how you prefer to handle this.

allennlp/training/learning_rate_schedulers.py
sebastianruder: Thanks for the suggestion.

allennlp/training/learning_rate_schedulers.py
sebastianruder: Thanks for the tip!

sebastianruder:
I've added tests for the learning rate schedule now. Let me know what else needs to be done.

cc @matt-peters @matt-gardner

allennlp/training/trainer.py
```diff
@@ -992,11 +992,20 @@ def from_params(cls,
         grad_norm = params.pop_float("grad_norm", None)
         grad_clipping = params.pop_float("grad_clipping", None)
         lr_scheduler_params = params.pop("learning_rate_scheduler", None)
+        optimizer_params = params.pop("optimizer")
+        if isinstance(optimizer_params, str):
+            optimizer_params = Params({"type": optimizer_params})
+        if "parameter_groups" in optimizer_params:
```
matt-gardner: I really don't like having to do this here.  I don't really understand well enough what this is accomplishing, so I don't know if it's feasible to modify the logic inside of your LR scheduler to make it so you can remove this.  But if there's a way to do it, I think it would be much better than doing this.  This is a pretty obscure dependency that's introduced between a LR scheduler and an optimizer; it's not clear at all from looking at this code that the point is to make your LR scheduler work.  It'd be really easy for a future PR to unwittingly break your scheduler by modifying these lines.

If there's no way to modify the logic in the LR scheduler to get the behavior you want, then it looks like you can just require a particular specification of the optimizer parameters in the config file, that you can check and enforce in the LR scheduler - the parameters after this modification are still something that could have been specified in JSON.  That's better to me than breaking the encapsulation of the optimizer and the LR scheduler by having these lines here.

sebastianruder: I agree. As I mentioned above, if we don't want to do this, then the parameter groups need to be specified in the optimizer, which is fine. I did not find a clear way to split the single list of parameters in the `Optimizer` into separate layers again.

matt-peters: This block of code could be easily moved into `Optimizer.from_params` if we just changed the call signature from `Optimizer.from_params(parameters, optimizer_params)` to `Optimizer.from_params(model, optimizer_params)`.  Since it's modifying already existing `optimizer_params` that seems like a more natural place for it.

I don't see any way to decouple completely the SLTR learning rate schedule parameters from the optimizer parameters, so suggest making it explicit and introducing a `use_slanted_triangular_rate_scheduler` or similar option in `optimizer_parameters` and only modifying the `optimizer_params["parameter_groups"]` in that case.  The code here will break / silently fail in the case where someone specifies `optimizer_params["parameter_groups"]` in their config but doesn't use the `SlantedTriangular` as it will overwrite whatever parameter group specific options they specified with the default values.  We should probably add a test to catch edge case interactions like this but that's a separate issue.

matt-gardner: Yes, that kind of broken interaction was what I was worried about.  Is it feasible to just require someone to specify `optimizer_params["parameter groups"]` differently in the JSON if they're using `SlantedTriangular`?  You should be able to check in the constructor of `SlantedTriangular` that things look like they're set up right, and fail if they're not.

matt-gardner: The point of doing it that way being that passing a parameter to `Optimizer.from_params` still gives extra logic in a `from_params` method that we want to avoid, and it still means that we're breaking the encapsulation of the learning rate scheduler and the optimizer.  Much better if the scheduler can just do what it needs to without needing to reach into the optimizer code.

matt-peters: Definitely, another alternative is to require the user to specify the parameter groups themselves if they are using `SlantedTriangular`.  That will avoid this code block too, and any changes to `Optimizer.from_params`.

sebastianruder: Yeah, let's assume the `parameter_groups` are specified by the user as that's a lot cleaner.

allennlp/training/trainer.py
```diff
@@ -992,11 +992,12 @@ def from_params(cls,
         grad_norm = params.pop_float("grad_norm", None)
         grad_clipping = params.pop_float("grad_clipping", None)
         lr_scheduler_params = params.pop("learning_rate_scheduler", None)
+        optimizer_params = params.pop("optimizer")
 
         if cuda_device >= 0:
             model = model.cuda(cuda_device)
         parameters = [[n, p] for n, p in model.named_parameters() if p.requires_grad]
-        optimizer = Optimizer.from_params(parameters, params.pop("optimizer"))
+        optimizer = Optimizer.from_params(parameters, optimizer_params)
```
matt-gardner: Looks like you can now just revert all of the changes to this file.

allennlp/training/learning_rate_schedulers.py
```diff
@@ -226,4 +334,5 @@ def get_lr(self):
         "reduce_on_plateau": torch.optim.lr_scheduler.ReduceLROnPlateau,
         "cosine": CosineWithRestarts,
         "noam": NoamLR,
+        "slanted_triangular": SlantedTriangular
```
matt-gardner: Minor nit: it's slightly easier to edit this code later if you end the line with a `,`.

allennlp/training/learning_rate_schedulers.py
```diff
@@ -134,6 +134,114 @@ def get_lr(self):
         return [scale for _ in range(len(self.base_lrs))]
 
 
+class SlantedTriangular(torch.optim.lr_scheduler._LRScheduler): # pylint: disable=protected-access
+    """
+    Implements the Slanted Triangular Learning Rate schedule with optional gradual
+    unfreezing. The schedule corresponds to first linearly increasing the learning
+    rate and annealing the learning based on a fixed ratio.
+    If we gradually unfreeze, then in the first epoch of training, only the top
```
matt-gardner: This will render nicer in the docs if you have a blank line between paragraphs.  You clearly intended that, but it won't show up right without the blank line.

allennlp/tests/training/learning_rate_schedulers_test.py
```diff
@@ -222,3 +223,114 @@ def init_and_restore_scheduler(optimizer: torch.optim.Optimizer,
...
+            LearningRateScheduler.from_params(optim, Params({"type": "slanted_triangular", "num_epochs": 5}))
+            LearningRateScheduler.from_params(optim, Params({"type": "slanted_triangular", "num_steps_epochs": 10}))
+
+    def test_schedules(self):
+        slanted_triangular_cases: List[Tuple[Dict[str, Any], List[Tuple[int, float]]]] = [
+            (
+                    {"num_epochs": 5,
+                     "num_steps_per_epoch": 10,
+                     "gradual_unfreezing": True},  # parameters
+                    [(0, 1, 0.03125),  # iteration, layer, learning rate
```
matt-gardner: I think it's this spacing that pylint doesn't like.  Sorry, this error seems particularly annoying, because your spacing doesn't seem unreasonable.

PR #1641, opened 2018-08-21
Add learning rate to logs

epwalsh:
This adds the learning rate for all parameter groups to the logs for each epoch. This is especially helpful when trying to fine-tune the learning rate.

learning_rate_logging1

learning_rate_loggin2

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -431,11 +431,7 @@ class Trainer:
 
         return loss
 
-    def _get_metrics(self,
-                     total_loss: float,
-                     num_batches: int,
-                     include_learning_rate: bool = False,
-                     reset: bool = False) -> Dict[str, float]:
+    def _get_metrics(self, total_loss: float, num_batches: int, reset: bool = False) -> Dict[str, float]:
         """
         Gets the metrics but sets ``"loss"`` to
         the total loss divided by the ``num_batches`` so that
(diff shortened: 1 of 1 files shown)


matt-gardner:
I really like including this in tensorboard; I'm less convinced that this is useful as a metric in the metrics table.  For one thing, we only show three decimal places there, and this could easily just be `0.000`, which would be confusing.

epwalsh:
That's a good point. Well I could throw in a little check right here:
https://github.com/epwalsh/allennlp/blob/13c5ba63a42a9c8100cde3c94a6c45db05cb7bc1/allennlp/training/trainer.py#L653

like `if name.startswith("learning_rate"): continue` in order to avoid that

matt-gardner:
Or maybe just do `self._tensorboard.add_train_scalar(...)` here: https://github.com/epwalsh/allennlp/blob/13c5ba63a42a9c8100cde3c94a6c45db05cb7bc1/allennlp/training/trainer.py#L533

Would that work?

epwalsh:
If we do it that way then I think we would also need to put a line here: https://github.com/epwalsh/allennlp/blob/13c5ba63a42a9c8100cde3c94a6c45db05cb7bc1/allennlp/training/trainer.py#L757 with a call to `self._tensorboard.add_train_scalar` so that we get a per-epoch log, right?

epwalsh:
@matt-gardner got it 👌, I only had per-epoch in mind. I just made that change. Let me know if you see anything else that I should update

allennlp/training/trainer.py
```diff
@@ -747,6 +747,11 @@ def train(self) -> Dict[str, Any]:
             self._save_checkpoint(epoch, validation_metric_per_epoch, is_best=is_best_so_far)
             self._metrics_to_tensorboard(epoch, train_metrics, val_metrics=val_metrics)
             self._metrics_to_console(train_metrics, val_metrics)
+            for ind, param_group in enumerate(self._optimizer.param_groups):
```
matt-gardner: `ind` is for `index`?  If so, just use `index` - for the price of just two characters, I don't have to guess what you mean ;).

Issue #1644, opened 2018-08-21
Using `requires_grad=True` on ELMo throws error when calling `loss.backward()`.


zfang:
**Describe the bug**
Using `requires_grad=True` on ELMo throws error when calling `loss.backward()`.

`loss.backward(retain_graph=True)` solves the issue, but is extremely slow. The issue is most likely caused by `ElmoLstm`.

**To Reproduce**
1. Create ELMo
```python
from allennlp.modules.elmo import Elmo

elmo = Elmo(requires_grad=True, vocab_to_cache=vocab_to_cache, ...)
```
2. Start training (`loss.backward()`, etc)
3. See error
```
RuntimeError: Trying to backward through the graph a second time, but the buffers have already been freed. Specify retain_graph=True when calling backward the first time.
```

**System (please complete the following information):**
 - OS: Ubuntu
 - Python version: 3.6
 - AllenNLP version: 0.6.0
 - PyTorch version: 0.4.0

nelson-liu:
Hmm, can you provide a script to reproduce? I've used the flag successfully in some of my own code, and this test suggests that the `Elmo` class itself is ok:

```
In [1]: from allennlp.modules.elmo import Elmo, batch_to_ids
   ...: options_file = "https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/2x4096_512_2048cnn_2xhighway/elmo_2x4096_512_2048cnn_2xhighway_options.json"
   ...: weight_file = "https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/2x4096_512_2048cnn_2xhighway/elmo_2x4096_512_2048cnn_2xhighway_weights.hdf5"
   ...:
   ...: elmo = Elmo(options_file, weight_file, 1, dropout=0, requires_grad=True)
   ...:
   ...:

In [2]: # use batch_to_ids to convert sentences to character ids
   ...: sentences = [['First', 'sentence', '.'], ['Another', '.']]
   ...: character_ids = batch_to_ids(sentences)
   ...:
   ...: embeddings = elmo(character_ids)
   ...:
   ...:

In [3]: embeddings['elmo_representations']
Out[3]:
[tensor([[[ 0.1474, -0.1475,  0.1376,  ...,  0.0270, -0.4051, -0.0498],
          [ 0.2394,  0.0769,  0.4126,  ..., -0.1671, -0.1707,  0.3884],
[...]

PR #1646, opened 2018-08-22
Make _ElmoBiLM._token_embedder.requires_grad False when vocab_to_cache is provided

zfang:
This PR is to address https://github.com/allenai/allennlp/issues/1644. When `vocab_to_cache` is provided, `_ElmoBiLM._token_embedder.requires_grad` should be `False` regardless if the `requires_grad` paramemter because 

1. It can cause `RuntimeError` (as per https://github.com/allenai/allennlp/issues/1644)
2. ` _ElmoBiLM._word_embedding` is already created and cached when `vocab_to_cache` is provided, and therefore `_ElmoBiLM._token_embedder` will not be used later anyways.

### DIFF
--- a/allennlp/modules/elmo.py
+++ b/allennlp/modules/elmo.py
@@ -496,13 +496,18 @@ class _ElmoBiLm(torch.nn.Module):
                  vocab_to_cache: List[str] = None) -> None:
         super(_ElmoBiLm, self).__init__()
 
-        self._token_embedder = _ElmoCharacterEncoder(options_file, weight_file, requires_grad=False if vocab_to_cache is not None else requires_grad)
+        self._token_embedder = _ElmoCharacterEncoder(options_file,
+                                                     weight_file,
+                                                     requires_grad=False
+                                                     if vocab_to_cache is not None else requires_grad)
 
         self._requires_grad = requires_grad
         if requires_grad and vocab_to_cache:
             logging.warning("You are fine tuning ELMo and caching char CNN word vectors. "
                             "This behaviour is not guaranteed to be well defined, particularly. "
-                            "if not all of your inputs will occur in the vocabulary cache.")
+                            "if not all of your inputs will occur in the vocabulary cache. "
+                            "_ElmoCharacterEncoder will be frozen because "
+                            "it is not used after word embedding caching.")
         # This is an embedding, used to look up cached
         # word vectors built from character level cnn embeddings.
         self._word_embedding = None
--- a/allennlp/modules/elmo.py
+++ b/allennlp/modules/elmo.py
@@ -60,7 +60,7 @@ class Elmo(torch.nn.Module):
         Should we apply layer normalization (passed to ``ScalarMix``)?
     dropout : ``float``, optional, (default = 0.5).
         The dropout to be applied to the ELMo representations.
-    vocab_to_cache : ``List[str]``, optional, (default = 0.5).
+    vocab_to_cache : ``List[str]``, optional, (default = None).
         A list of words to pre-compute and cache character convolutions
         for. If you use this option, Elmo expects that you pass word
         indices of shape (batch_size, timesteps) to forward, instead
@@ -482,7 +482,7 @@ class _ElmoBiLm(torch.nn.Module):
         ELMo hdf5 weight file
     requires_grad: ``bool``, optional
         If True, compute gradient of ELMo parameters for fine tuning.
-    vocab_to_cache : ``List[str]``, optional, (default = 0.5).
+    vocab_to_cache : ``List[str]``, optional, (default = None).
         A list of words to pre-compute and cache character convolutions
         for. If you use this option, _ElmoBiLm expects that you pass word
         indices of shape (batch_size, timesteps) to forward, instead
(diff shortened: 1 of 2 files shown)


allennlp/modules/elmo.py
```diff
@@ -496,7 +496,7 @@ def __init__(self,
                  vocab_to_cache: List[str] = None) -> None:
         super(_ElmoBiLm, self).__init__()
 
-        self._token_embedder = _ElmoCharacterEncoder(options_file, weight_file, requires_grad=requires_grad)
+        self._token_embedder = _ElmoCharacterEncoder(options_file, weight_file, requires_grad=False if vocab_to_cache is not None else requires_grad)
```
nelson-liu: Thanks for the fix! It'd be great if you could mention this interaction somewhere in the docstring (it's impossible to know that this happens from just looking at the docs, you have to look at the code, at this PR, or at the list of frozen parameters that AllenNLP logs).

Additionally, I'd appreciate a warning here if `vocab_to_cache is not None` and `requires_grad is True` that lets the user know that the character-level embedder will be frozen (i.e., that this new behavior will be applied).

matt-peters: I think this behavior is dangerous, because even if it's documented it can have very unexpected behavior for the user.  Not everyone reads documentation (I often don't).  It's much better to raise an exception with a helpful message if `requires_grad=True` and `vocab_to_cache is not None`.

zfang: Sure, although I don't believe I'm breaking any backward compatibility as per the bug, so the "new" behavior shouldn't be new to anyone.

matt-peters: With the bug at least it didn't silently change it's behavior and instead failed loudly.  Imagine submitting a paper using ELMo where you claimed (and honestly thought) you were backproping into the the trained LM, only to find out that no, the weights weren't updated due to an obscure edge case in the framework you were using.

zfang: @matt-peters Referring to the https://github.com/allenai/allennlp/issues/1644 can give you a better understanding of the rationale. I also gave my reasons in the PR description. If for some reason you still think it's dangerous, let me know.

matt-peters: It's a good fix, and thank you for submitting the PR and adding the test.  The only change I'm suggesting is to raise an exception instead of changing the value of `requires_grad` if `vocab_to_cache` is provided.  It's about a two line change to the existing PR.

zfang: The change only freezes the char CNN embedder (e.g. `_token_embedder`). Everything else (e.g. `_elmo_lstm`) is still parametrized by `requires_grad`. Also `_token_embedder` is not used after creating the word embedding cache.

zfang: We do not want to raise exception. Again, if `requires_grad` is `True` and `vocab_to_cache` is provided, we still want to make `_elmo_lstm` tunable, but not so for `_token_embedder` (referring to PR description for reasons).

matt-gardner: Maybe the solution here is to require a third flag, where you're explicitly requesting (and thus acknowledging that you know about) the character embeddings to be fixed.  I agree with @matt-peters here that we shouldn't do this silently (and one logging statement in the midst of all of allennlp's output is pretty close to "silent" if you're not looking for the statement).

matt-peters: That's an option.  The thing that worries me about this is that the `vocab_to_cache` was only intended to speed up prediction with a trained LM, and combining it with `requires_grad=True` will modify the network architecture in an implicit and non-obvious way (by effectively replacing the character CNN with a word embedding layer).

matt-gardner: Yeah, I agree, which is why I think we shouldn't do this without the user providing an extra flag.  That flag would be something like `freeze_cnn_layer`, or something, so you know what's happening here.  The error message when you don't pass this could say something about how this isn't really recommended, but you can do it if you want.

zfang: I still cannot understand the concern since the character CNN layer is NOT even used after the `_ElmoBiLM.__init__` call with `vocab_to_cache`.

nelson-liu: @zfang the character CNN layer is used in `_ElmoBiLM.forward()` (https://github.com/allenai/allennlp/blob/cc3781dbc9f9ef56bc83b660543cdd342b7857e1/allennlp/modules/elmo.py#L578-L588), which is called in `Elmo.forward()` here (https://github.com/allenai/allennlp/blob/cc3781dbc9f9ef56bc83b660543cdd342b7857e1/allennlp/modules/elmo.py#L153) . does that clarify things a bit?

matt-gardner: The trouble is in what someone would assume when passing in `requires_grad=True`.  They would reasonably assume that this means, "I'm fine tuning all of the weights of the language model."  Except, if you also pass in `vocab_to_cache`, that's not true, and it's _silently_ not true.  This is a problem.

zfang: @nelson-liu I'm talking about the `self._word_embedding is not None and word_inputs is not None` case. The block in `except RuntimeError` is a bad idea in my opinion, but that's a different story.

DeNeutoy: @zfang backing off to using the full character embeddings is required - that cannot change. The reason for this is that we don't just want to run models we trained with the cached vocab on data we know won't be OOV - e.g for models in our demos, people might put anything in and we need to handle it smoothly.

zfang: @DeNeutoy Hmm maybe some people are using it like that by providing character embedding ids in addition to word embedding ids. In my case (which I thought is the more usual case if you use ELMo just as a drop-in solution to replace the existing layer `torch.nn.Embedding`) I'd only provide the word embedding ids.

schmmd:
@zfang thanks for the contribution, but we decided internally this isn't a good fit for the library.  The case this supports is a failure case for most use cases and we felt it would be a confusing addition.

PR #1647, opened 2018-08-22
Implement cosine with restarts

### LINKED ISSUE #1642: Feature request: cosine annealing scheduler with warm restarts
Currently, the pytorch implementation of [the cosine annealing learning rate schedule](https://pytorch.org/docs/stable/optim.html#torch.optim.lr_scheduler.CosineAnnealingLR) does not actually include warm restarts, which was the key idea in the paper referenced from the docs:

> It has been proposed in [SGDR: Stochastic Gradient Descent with Warm Restarts](https://arxiv.org/abs/1608.03983). Note that this only implements the cosine annealing part of SGDR, and not the restarts.

There is currently a [pull request open on pytorch](https://github.com/pytorch/pytorch/pull/6130) that addresses this issue, but it is unclear if or when it will be merged and released. Since the logic is relatively simple, and there are lots of potential variants of the algorithm, I propose implementing a custom cosine annealing schedule with warm restarts within AllenNLP.

I've actually already implemented this for my own research, so I'd be willing to submit a PR. It looks something like this:

```python
@LearningRateScheduler.register("cosine_with_restarts")
class CosineWithRestarts(_LRScheduler):
    """
    Cosine annealing with restarts.

    Parameters
    ----------
    optimizer : torch.optim.Optimizer

    T_max : int
        The maximum number of iterations within the first cycle.

    eta_min : float, optional (default: 0)
        The minimum learning rate.

    last_epoch : int, optional (default: -1)
        The index of the last epoch.

    """

    def __init__(self,
                 optimizer: torch.optim.Optimizer,
                 T_max: int,
                 eta_min: float = 0.,
                 last_epoch: int = -1,
                 factor: float = 1.) -> None:
        # pylint: disable=invalid-name
        self.T_max = T_max
        self.eta_min = eta_min
        self.factor = factor
        self._last_restart: int = 0
        self._cycle_counter: int = 0
        self._cycle_factor: float = 1.
        self._updated_cycle_len: int = T_max
        self._initialized: bool = False
        super(CosineWithRestarts, self).__init__(optimizer, last_epoch)

    def get_lr(self):
        """Get updated learning rate."""
        # HACK: We need to check if this is the first time get_lr() was called, since
        # we want to start with step = 0, but _LRScheduler calls get_lr with
        # last_epoch + 1 when initialized.
        if not self._initialized:
            self._initialized = True
            return self.base_lrs

        step = self.last_epoch + 1
        self._cycle_counter = step - self._last_restart

        lrs = [
            (
                self.eta_min + ((lr - self.eta_min) / 2) *
                (
                    np.cos(
                        np.pi *
                        ((self._cycle_counter) % self._updated_cycle_len) /
                        self._updated_cycle_len
                    ) + 1
                )
            ) for lr in self.base_lrs
        ]

        if self._cycle_counter % self._updated_cycle_len == 0:
            # Adjust the cycle length.
            self._cycle_factor *= self.factor
            self._cycle_counter = 0
            self._updated_cycle_len = int(self._cycle_factor * self.T_max)
            self._last_restart = step

        return lrs
```

epwalsh:
fixes #1642 

@matt-gardner I took a closer look at the discussion in [here](https://github.com/pytorch/pytorch/pull/6130), and I'm pretty sure this addresses all of those concerns. I think there's still room to add more features to this to match all of the options that tensor flow supports: https://github.com/tensorflow/tensorflow/blob/25c197e02393bd44f50079945409009dd4d434f8/tensorflow/python/training/learning_rate_decay.py#L608.

I choose to replace the registered `cosine` schedule from PyTorch with this one because the default expected behavior is the same. The behavior will differ of course if other options are supplied to the new scheduler, or if the total number of epochs is greater than `T_max`. However, in this scenario the behavior of PyTorch's cosine scheduler is actually undefined according to the paper referenced, and is therefore probably unexpected to the user. That is why I think this scheduler should replace the default pytorch one.

The other thing that may need to addressed is saving the state of schedulers here:

https://github.com/allenai/allennlp/blob/bca6c2a17741468b0d202c7a986ed1e407f507d2/allennlp/training/trainer.py#L814-L817

I think that would make resuming training more predictable, as the learning rate scheduler would start where it left off.

### DIFF
--- a/allennlp/training/learning_rate_schedulers.py
+++ b/allennlp/training/learning_rate_schedulers.py
@@ -12,13 +12,19 @@ The available learning rate schedulers are
 * `"cosine" <http://pytorch.org/docs/master/optim.html#torch.optim.lr_scheduler.CosineAnnealingLR>`_
 """
 
+import logging
 from typing import Optional
 
+import numpy as np
 import torch.optim.lr_scheduler
 from overrides import overrides
+
 from allennlp.common.checks import ConfigurationError
 from allennlp.common.params import Params
 from allennlp.common.registrable import Registrable
+
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 
 class LearningRateScheduler(Registrable):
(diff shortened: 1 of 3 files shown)


allennlp/training/learning_rate_schedulers.py
```diff
@@ -118,21 +124,101 @@ def step_batch(self, epoch=None):
...
+    """
+    Cosine annealing with restarts.
+
+    This is decribed in the paper https://arxiv.org/abs/1608.03983.
+
+    Parameters
+    ----------
+    optimizer : torch.optim.Optimizer
+
+    T_max : int
```
matt-gardner: Can you use a lowercase `t_max` here, too, to match variable naming conventions?

matt-gardner:
You've labeled this WIP - is that because of the checkpointing issue?  I'd say that should be fixed in a separate PR.

epwalsh:
@matt-gardner yea that was my thought. I just changed the title

matt-gardner:
Thanks!

allennlp/training/learning_rate_schedulers.py
```diff
@@ -118,21 +124,100 @@ def step_batch(self, epoch=None):
...
+                 factor: float = 1.) -> None:
+        assert t_max > 0
+        assert eta_min >= 0
+        if t_max == 1 and factor == 1:
+            logger.warning("Cosine annealing scheduler will have no effect on the learning "
+                           "rate since T_max = 1 and factor = 1.")
+        self.t_max = t_max
+        self.eta_min = eta_min
+        self.factor = factor
+        self._last_restart: int = 0
```
Ian09: Sorry for interrupt, but I don't understand what is the meaning ": int"? I never saw it in python code. Could you give me some hints?

matt-gardner: That is a type annotation.

ahundt:
@epwalsh could you consider making a pr with this directly to pytorch itself?

epwalsh:
@ahundt sure, I'll probably just submit a clean one as I think I took a slightly different approach

epwalsh:
@ahundt actually did you see this PR as well https://github.com/pytorch/pytorch/pull/11104? It looks promising although there has been no response from the maintainers yet

ahundt:
Ah I didn't see pytorch/pytorch#11104, neat. 
I tried using this by importing allennlp directly but it has too many dependencies & failed to import. Right now I just copy pasted this code with changes & a reference back to the source and that's working, so none of this is holding me up.

PR #1648, opened 2018-08-22
make pos selection an option in dataset reader, use in predictor

DeNeutoy:


allennlp/predictors/biaffine_dependency_parser.py
```diff
@@ -105,7 +105,10 @@ def _json_to_instance(self, json_dict: JsonDict) -> Instance:
         """
         spacy_tokens = self._tokenizer.split_words(json_dict["sentence"])
         sentence_text = [token.text for token in spacy_tokens]
-        pos_tags = [token.tag_ for token in spacy_tokens]
+        if self._dataset_reader.use_language_specific_pos: # type: ignore
+            pos_tags = [token.tag_ for token in spacy_tokens]
```
matt-gardner: It'd be good to have a comment here saying that `token.tag_` in spacy is "fine-grained part of speech", while `token.pos_` is "coarse part of speech".  Is spacy's coarse part of speech even the same as the universal dependencies' pos tag scheme?

allennlp/data/dataset_readers/universal_dependencies.py
```diff
@@ -59,7 +62,10 @@ def _read(self, file_path: str):
                 heads = [x["head"] for x in annotation]
                 tags = [x["deprel"] for x in annotation]
                 words = [x["form"] for x in annotation]
-                pos_tags = [x["upostag"] for x in annotation]
+                if self.use_language_specific_pos:
+                    pos_tags = [x["xpostag"] for x in annotation]
```
matt-gardner: Would this problem have been caught earlier if you had used `pos_tags` instead of `pos` for your label namespace in the pos tag sequence label field?  I think it would have, because you'd have crashed at an UNK part of speech tag.  I think that's the right thing to do, as this is a closed class that you can expect to see all of at training time (and if not, you can manually add all of them to the vocabulary).

PR #1650, opened 2018-08-23
WIP: save learning rate scheduler to training state

epwalsh:
When using a learning rate scheduler, I think user probably expects the scheduler to pick up where it left off when restoring a trainer from a checkpoint. This makes sure that the scheduler state is stored with each checkpoint when a scheduler is used. Let me know if you have any comments!

**TODO:**
- [x] modify `Trainer._restore_checkpoint`
- [x] modify `Trainer._save_checkpoint`
- [ ] add unit tests for `Trainer`
- [ ] add unit tests for `CosineWithRestarts` scheduler to make sure state is correctly restored

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -838,6 +838,9 @@ class Trainer:
                               'val_metric_per_epoch': val_metric_per_epoch,
                               'optimizer': self._optimizer.state_dict(),
                               'batch_num_total': self._batch_num_total}
+            if self._learning_rate_scheduler is not None:
+                training_state["learning_rate_scheduler"] = \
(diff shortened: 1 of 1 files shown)


matt-gardner:
@DeNeutoy, you mind taking this one?  I've never touched the checkpointing code; I think you have more context there.

epwalsh:
Just a heads up, I might have just found a bug w/ the restoring from checkpoint for learning rate schedulers. I'm looking into it now. Changing the title back to "WIP"

epwalsh:
Update: the bug was because the checkpoints were being saved before updating the learning rate scheduler, so when the learning rate scheduler was restored, it was an epoch behind. I fixed this and added a test for it in `tests.training.trainer_test.py` 😅

PR #1655, opened 2018-08-23
Fix Conll2003 reader docs to reflect true label namespace names

nelson-liu:
The docs for the conll2003 reader say:

```
    feature_labels: ``Sequence[str]``, optional (default=``()``)
        These labels will be loaded as features into the corresponding instance fields:
        ``pos`` -> ``pos_tags``, ``chunk`` -> ``chunk_tags``, ``ner`` -> ``ner_tags``
        Each will have its own namespace: ``pos_labels``, ``chunk_labels``, ``ner_labels``.
        If you want to use one of the labels as a `feature` in your model, it should be
        specified here.
```

however, the namespaces are improperly named `pos_tags`, `chunk_tags`, etc. This fixes them.

cc @matt-peters this might affect you?


nelson-liu:
alternatively, maybe we should keep the current behavior the same and just modify the docs? I rather like that the field name and label namespace matched.

PR #1656, opened 2018-08-23
Standardize tagging datasets

nelson-liu:
The goal of this PR is to enable a single sequence labeling model to work with all sequence labeling dataset readers.

Changes:

- [X] Modify CCGBank dataset reader to match Conll2003

@matt-peters these CCGBank changes might affect you? What do you think about them?

### DIFF
--- a/allennlp/data/dataset_readers/conll2003.py
+++ b/allennlp/data/dataset_readers/conll2003.py
@@ -68,11 +68,14 @@ class Conll2003DatasetReader(DatasetReader):
         specified here.
     coding_scheme: ``str``, optional (default=``IOB1``)
         Specifies the coding scheme for ``ner_labels`` and ``chunk_labels``.
-        Valid options are ``IOB1`` and ``BIOUL``.  The ``IOB1`` default maintains
-        the original IOB1 scheme in the CoNLL data.
+        Valid options are ``IOB1``, ``BIO``, and ``BIOUL``.  The ``IOB1`` default maintains
+        the original IOB1 scheme in the CoNLL 2003 NER data.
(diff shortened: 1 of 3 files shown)


allennlp/data/dataset_readers/conll2003.py
```diff
@@ -68,11 +68,14 @@ class Conll2003DatasetReader(DatasetReader):
...
     coding_scheme: ``str``, optional (default=``IOB1``)
         Specifies the coding scheme for ``ner_labels`` and ``chunk_labels``.
-        Valid options are ``IOB1`` and ``BIOUL``.  The ``IOB1`` default maintains
-        the original IOB1 scheme in the CoNLL data.
+        Valid options are ``IOB1``, ``BIO``, and ``BIOUL``.  The ``IOB1`` default maintains
+        the original IOB1 scheme in the CoNLL 2003 NER data.
         In the IOB1 scheme, I is a token inside a span, O is a token outside
         a span and B is the beginning of span immediately following another
         span of the same type.
+        To maintain the default of the CoNLL 2000 syntactic chunking data,
```
nelson-liu: @matt-peters in light of this, the `to_bioul` conversion for the conll2000 data actually doesn't work because the `to_bioul` function assumes that the coding is `IOB1`. 

I think it's worth making a `conll2000` dataset reader that subclasses the `conll2003` reader to more explicitly separate them? It's annoying because the chunk coding in the conll2000 data is BIO, while the chunk coding in the conll2003 data is IOB1.

schmmd:
@nelson-liu I'm not sure--someone else might have a better idea.  Is there a reasonable place under https://allenai.github.io/allennlp-docs/api/allennlp.data.dataset_readers.html?

PR #1657, opened 2018-08-24
Fix conll2000 data reading

nelson-liu:
Previously, the conll2003 reader was used to read the conll2000 data. 

But that doesn't quite work if you want to convert the data to BIOUL, since the chunking data in the Conll2003 is coded in IOB1 format while the chunking data in the Conll2000 data is coded in BIO format.

In addition, it seems reasonable to me to have two separate registered dataset readers for each dataset, versus having the conll2003 dataset reader also read conll2000.


nelson-liu:
The code is structurally quite similar to the conll2003 reader, but there are some slight differences in behavior for each function between conll2003 and conll200, so it makes it tough to have one inherit from the other.

PR #1658, opened 2018-08-24
SQL semantic parser entity improvements

kl2806:
- Constrain the grammar such that only strings that appear in a column can be predicted for example
```
biexpr = ( "city" ws "." ws "city_name"  binop ws city_city_name_strings ) / ...
city_city_name_strings = "NASHVILLE" / "BOSTON" /  ...
```

- Removes the keyword based heuristics for generating strings


allennlp/semparse/contexts/sql_table_context.py
```diff
@@ -137,15 +164,27 @@ def initialize_valid_actions(self) -> Dict[str, List[str]]:
...
+            for table, columns in self.tables_with_strings.items():
+                biexprs.extend([f'("{table}" ws "." ws "{column}" ws binaryop ws {table}_{column}_string)'
+                                for column in columns])
+                for column in columns:
+                    self.cursor.execute(f'SELECT DISTINCT {table} . {column} FROM {table}')
+                    grammar_str += generate_one_of_string(nonterminal=f'{table}_{column}_string',
+                                                          literals=[row[0] for row in self.cursor.fetchall()],
+                                                          is_string=True)
+
+        grammar_str += 'biexpr = ( col_ref ws binaryop ws value) / (value ws binaryop ws value) / ' + \
```
matt-gardner: How does parsimonious handle repeated definitions like this?

matt-gardner: And, don't you no longer want the `(col_ref ws binaryop ws value)` rule?  This is the more-general, non-constrained version, and if it's still in the grammar, we can just do an end-run around the additional constraints, right?

allennlp/semparse/worlds/atis_world.py
```diff
@@ -46,9 +21,12 @@ class AtisWorld():
         A list of utterances in the interaction, the last element in this list is the
         current utterance that we are interested in.
     """
-    sql_table_context = SqlTableContext(TABLES)
```
matt-gardner: Why did you remove this as a class variable?  Because you didn't want to have a shared sqlite connection?  I suppose it's possible you somehow got into a race condition in your beaker experiment, depending on how exactly you did the evaluation.

But, either way, I still think the `sql_table_context` should just be constructed inside the `World`, not taken as a dependency.  It's not something that ever changes, so having it as a constructor parameter doesn't add anything.

allennlp/tests/data/dataset_readers/atis_test.py
```diff
@@ -53,22 +38,6 @@ def test_atis_read_from_file(self):
         # (num_entities, num_utterance_tokens). We have two types
         # of entities: strings and numbers.
         assert world.linking_scores.shape[0] == \
-                len(world.valid_actions['string'] +
-                    world.valid_actions['number'])
+                len(world.valid_actions['number'])
         assert world.linking_scores.shape[1] == \
                 len(instance.fields['utterance'].tokens)
-
-        expected_linking_scores = [[0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1], # "westchester county" ->
```
matt-gardner: Why are you removing this?  Ah, I see, you're switching to learning an embedding for all of the string actions, instead of linking them.  I'm guessing that we'll still want to link these actions, instead of learning an embedding for them, but we can try it this way and see what happens.

allennlp/semparse/worlds/atis_world.py
kl2806: I agree that it is best if `sql_table_context` is constructed inside `World`, but since we constrain strings based on values in the database, the `sql_table_context` now relies on a call to the database, which means we need to pass it the location of the database. It's passed in as a parameter of the dataset reader right now (https://github.com/allenai/allennlp/blob/9a25f994311310e49f695296863f761ca82f0d06/allennlp/data/dataset_readers/atis.py#L78), is there a way to keep it as a class variable and also pass it the location of the database?

matt-gardner: Ah, yes, that's a good point that I forgot about.  Still, I'd probably just pass the database location in as a parameter, and construct the `SqlTableContext` internally, so it's more obvious what's a configurable dependency and what isn't.

allennlp/semparse/contexts/sql_table_context.py
kl2806: parsimonious takes the last definition, so redefining it here will override the previous definition.

I kept `(col_ref ws binaryop ws value)` because we still need it for things like `flight.flight_arrival_time = ALL ( SELECT DISTINCT ...`. To keep the grammar tight, I can change this into`(col_ref ws binaryop ws string_right_hand_side)`. Then make another rule `string_right_hand_side = ("ALL" ws query ) / ...` so that it doesn't allow numbers, etc.

matt-gardner: Ah, I get it; you removed `string` from `value`, so you can't actually get what I was afraid of.  That works.  If you can make it even tighter by doing what you suggested, even better.

allennlp/semparse/worlds/atis_world.py
kl2806: I see -- so pass the database location to `AtisWorld` instead of passing a `SqlTableContext`. Would this be inefficient because we're instantiating a `SqlTableContext` for every world?

matt-gardner: That's a good question, and I don't know what the practical implications are, especially with multiple hits to a sqlite db.  If you think it makes more sense to do it the way it currently is, that's fine.

PR #1661, opened 2018-08-24
Make CrfTagger work with non-BIO tagging tasks

nelson-liu:
Would recommend looking at #1657 first, but this makes the `CrfTagger` work on non bio-tagging tasks like CCG supertagging and POS tagging.

TODO:
- [x] This is a breaking change, do we want to leave the `constraint_type` argument around for a few more versions?
- [x] Fix the training configs that use this tagger (e.g. `ner.jsonnet`)


joelgrus:
I would definitely deprecate `constraint_type` more gradually

nelson-liu:
fyi, since i don't think it's been used in the project yet: here's what the `.. deprecated::` sphinx directive looks like in the docs

screen shot 2018-08-24 at 8 22 34 am

@schmmd do we have a process around deprecation (e.g., how many versions how to wait, properly documenting it)? Might be worth thinking about?

schmmd:
@nelson-liu we don't have a formal process, but it seems reasonable to leave it deprecated until the next major release (i.e. 0.7.0) and to make mention of it on the release where we remove the deprecated method.

allennlp/models/crf_tagger.py
```diff
@@ -81,20 +104,45 @@ def __init__(self, vocab: Vocabulary,
                                                            self.num_tags))
 
         if constraint_type is not None:
+            warnings.warn("'constraint_type' was removed and replaced with"
+                          "'label_encoding', 'constrain_crf_decoding', and "
+                          "'calculate_span_f1' in version 0.6.1. It will be "
+                          "removed in version 0.8.", DeprecationWarning)
```
joelgrus: how do we track things we intend to remove?

allennlp/models/crf_tagger.py
```diff
@@ -34,15 +35,34 @@ class CrfTagger(Model):
...
+
     include_start_end_transitions : ``bool``, optional (default=``True``)
         Whether to include start and end transition parameters in the CRF.
+    constrain_crf_decoding : ``bool``, optional (default=``False``)
+        If ``True``, the CRF is constrained at decoding time to
+        produce valid sequences of tags. If this is ``True``, then
+        ``label_encoding`` is required.
+    calculate_span_f1 : ``bool``, optional (default=``False``)
+        Calculate span-level F1 metrics during training. If this is ``True``, then
+        ``label_encoding`` is required.
```
joelgrus: I know it makes the code more complicated, but would it be better for the default (for calculate_span_f1 and constrain_crf_decoding) to be `None`, and to interpret that as "true if label_encoding supplied, false otherwise", since it seems like that's probably the behavior you'd want?

allennlp/models/crf_tagger.py
```diff
@@ -81,20 +104,45 @@ def __init__(self, vocab: Vocabulary,
...
+                raise ConfigurationError("calculate_span_f1 is True, but "
+                                         "no label_encoding was specified.")
+            self._f1_metric = SpanBasedF1Measure(vocab,
+                                                 tag_namespace=label_namespace,
+                                                 label_encoding=label_encoding)
+        elif constraint_type is not None:
+            # Maintain deprecated behavior if constraint_type is provided
+            self._f1_metric = SpanBasedF1Measure(vocab,
+                                                 tag_namespace=label_namespace,
+                                                 label_encoding=constraint_type or "BIO")
```
joelgrus: if you're in this case, doesn't it have to be true that `constraint_type` was provided?, so you don't need the `or "BIO"`?

nelson-liu: you're right, thanks.

allennlp/models/crf_tagger.py
nelson-liu: Unfortunately, it's a pre-release sort of thing 😢 . Perhaps we could start an issue for each thing to deprecate and tag it with the milestone of the release that it's meant to be removed in? Then, before we make that release, we go through and make sure that everything that needs to be removed has been removed?

allennlp/models/crf_tagger.py
nelson-liu: Hmm, yeah i guess that's a more reasonable default.

allennlp/models/crf_tagger.py
schmmd: Yes, I think that's what we need to do.

schmmd: https://github.com/allenai/allennlp/milestone/7

nelson-liu: should we remove in 0.7 or 0.8? I opted for 0.8 since other big projects (e.g., [scikit-learn](http://scikit-learn.org/stable/developers/contributing.html#deprecation)) deprecate in 2 releases, and say we decide to deprecate something in the last 0.6.x release --- deprecating in 0.7 means that there's almost no warning.

schmmd:
Sounds good.

PR #1673, opened 2018-08-27
[WIP] Atis model action refactored

kl2806:


matt-gardner:
I just skimmed over this, wondering how it ended up at more than 4000 lines added.  Looks like most of it is testing for the presence of all of the strings in the database (3370 lines added to the atis world test).  I'd say you can remove those tests, or just check for the presence of a few things that should be there.  When we said the test was long but worth it in the wikitables case, that was because the grammar is large, not because the table is large, and testing the grammar rules is worth it.  I wouldn't say that it's worth it to test all of the strings in the DB.

PR #1674, opened 2018-08-27
Predictor for QUAC models

eunsol:


allennlp/predictors/dialog_qa.py
```diff
@@ -0,0 +1,56 @@
...
+    def _json_to_instance(self, json_dict: JsonDict) -> Instance:
+        """
+        Expects json that looks like the original quac file
+        """
+        paragraph_json = json_dict["paragraphs"][0]
+        paragraph = paragraph_json['context']
+        tokenized_paragraph = self._tokenizer.split_words(paragraph)
+        qas = paragraph_json['qas']
+        metadata = {}
+        metadata["instance_id"] = [qa['id'] for qa in qas]
```
matt-gardner: Because the main purpose of the predictor is to run a demo, you should probably be robust to not having an id here.

training_config/dialog_qa.json
```diff
@@ -108,11 +108,9 @@
         },
         "num_epochs": 30,
         "optimizer": {
-            "type": "adam",
```
matt-gardner: Did you mean to make this change?

allennlp/predictors/dialog_qa.py
```diff
@@ -0,0 +1,56 @@
...
+        question_text_list = [qa["question"].strip().replace("\n", "") for qa in qas]
+        answer_texts_list = [[answer['text'] for answer in qa['answers']] for qa in qas]
```
matt-gardner: You also probably shouldn't assume that you always have the answer given.

eunsol: The answers to the previous questions in the dialog is actually *input* to the later questions. So it's kind of necessary to have it here. The more realistic demo scenario would allow interaction back and forth between user / model will be much more challenging to implement (i think?), and the purpose of this predictor is not necessarily for demo (because of that reason), but to allow outputting to the json file after the predictions.

training_config/dialog_qa.json
eunsol: yes, for the camera ready we used sgd version as it was more stable.

allennlp/predictors/dialog_qa.py
eunsol: i can remove the instance id, but please refer to the comments to the other feedback (a.k.a. why i can't remove answers).

allennlp/predictors/dialog_qa.py
matt-gardner: Hmm, right, ok.  But, your last statement makes me think there's something wrong with our API, because you shouldn't have to implement this just to get JSON predictions out.  I'm going to open an issue for that.

allennlp/predictors/dialog_qa.py
```diff
@@ -0,0 +1,56 @@
...
+@Predictor.register('dialog_qa')
+class DialogQAPredictor(Predictor):
+    def __init__(self, model: Model, dataset_reader: DatasetReader) -> None:
+        super().__init__(model, dataset_reader)
+        self._tokenizer = SpacyWordSplitter(language='en_core_web_sm')
+
+    def predict(self, jsonline: str) -> JsonDict:
+        """
+        Make a machine comprehension prediction on the supplied input.
+       -------
```
matt-gardner: This is why the build is failing - this needs to be formatted like the rest of our docstrings (you probably meant this to be a "Returns" block, and the spacing here is off).

matt-gardner: And, even though I approved the PR, I can't merge it until the build passes, just FYI.

allennlp/predictors/dialog_qa.py
eunsol: Fixed. Thanks!

PR #1677, opened 2018-08-28
Add a default predictor for biaffine-parser.

schmmd:


schmmd:
This PR makes the code example in https://github.com/allenai/allennlp-website/pull/58 work.  Do we want a default predictor like this?

It seems like the default predictor should be "dependency_parser" instead of "biaffine_parser".

PR #1679, opened 2018-08-28
Event2Mind

brendan-ai2:
- Forks SimpleSeq2Seq.
- New metrics SequenceAccuracy and UnigramRecall.
- Parameterized to easily handle xintent, xreact and so on.


brendan-ai2:
@DeNeutoy, would you mind taking a look at this PR?

By the way, I ended up solving the issue we talked about yesterday (zero gradient in part of the model) by switching from `Seq2Seq` to `Seq2Vec` which I believe is what the model should have been using all along.

brendan-ai2:
Looks like I need to add some tests to the model itself to get coverage up. Probably the beam search mostly. I'm also seeing some errors regarding documentation, e.g. `undocumented module: allennlp.data.dataset_readers.event2mind`. There are some comments for that module. How can I determine how detailed the tool wants me to be?

matt-gardner:
For the test, I'd start with a simple `test_model_can_train_save_and_load`, like all of our other models do.  And for the docs, the tool just wants you to add that to the api docs in `doc/`, not that you need to have a specific amount of documentation (the script sees that there's a python module that's not referenced in the API docs, so it complains).

brendan-ai2:
That actually already is present, but I think I misnamed the test such that the test runner didn't pick it up (`event2mind.py` vs `event2mind_test.py`). So now I'm confused as to how there was any coverage for the main model... At any rate, renamed and rerunning.

I also see that a Wikitables test failed presumably due a sporadic network failure. See http://build.allennlp.org/viewLog.html?buildId=5536&buildTypeId=AllenNLP_AllenNLPPullRequests&tab=buildLog&_focus=1143. Normally I would expect unit tests not to access the network. Is that worth making an issue for?

matt-gardner:
I don't think there's an easy way to fix that issue, unfortunately.  We also have more than just "unit tests" running in TeamCity - there are sniff tests and other things, which definitely need network access.  One option is to mark any such test as flaky, and that should dramatically lower the probability of these sporadic failures.

allennlp/models/event2mind.py
```diff
@@ -0,0 +1,405 @@
...
+           with masks             1   1   1   1   0   0
+           against                l1  l2  l3  l4  l5  l6
+           (where the input was)  <S> w1  w2  w3  <E> <P>
+        """
+        relevant_targets = targets[:, 1:].contiguous()  # (batch_size, num_decoding_steps)
+        relevant_mask = target_mask[:, 1:].contiguous()  # (batch_size, num_decoding_steps)
+        loss = sequence_cross_entropy_with_logits(logits, relevant_targets, relevant_mask)
+        return loss
+
+    def decode_all(self, predicted_indices: torch.Tensor):
```
DeNeutoy: add a return type

allennlp/training/metrics/unigram_recall.py
```diff
@@ -0,0 +1,94 @@
...
+        if gold_labels.dim() != predictions.dim() - 1:
+            raise ConfigurationError("gold_labels must have dimension == predictions.dim() - 1 but "
+                                     "found tensor of shape: {}".format(gold_labels.size()))
+        if mask is not None and mask.size() != gold_labels.size():
+            raise ConfigurationError("mask must have the same size as predictions but "
+                                     "found tensor of shape: {}".format(mask.size()))
+
+        batch_size = predictions.size()[0]
+        correct = 0.0
+        # Note: See preprocess.py.
```
DeNeutoy: Is this note still relevant?

allennlp/training/metrics/unigram_recall.py
```diff
@@ -0,0 +1,94 @@
...
+        self.correct_count += correct
+        self.total_count += predictions.size()[0]
+
+    def get_metric(self, reset: bool = False):
+        """
+        Returns
+        -------
+        The accumulated recall.
+        """
+        recall = float(self.correct_count) / float(self.total_count)
```
DeNeutoy: Add an epsilon to avoid dividing by zero.

allennlp/data/dataset_readers/event2mind.py
```diff
@@ -0,0 +1,159 @@
...
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+@DatasetReader.register("event2mind")
+class Event2MindDatasetReader(DatasetReader):
+    """
+    Reads instances from the Event2Mind dataset.
+
+    This dataset is CSV and has the columns:
+    Source,Event,Xintent,Xemotion,Otheremotion,Xsent,Osent
```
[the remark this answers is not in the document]
brendan-ai2: Done

allennlp/data/dataset_readers/event2mind.py
```diff
@@ -0,0 +1,159 @@
...
+        self._target_token_indexers = target_token_indexers or self._source_token_indexers
+        self._source_add_start_token = source_add_start_token
+
+    @overrides
+    def _read(self, file_path):
+        with open(cached_path(file_path), "r") as data_file:
+            logger.info("Reading instances from lines in file at: %s", file_path)
+            reader = csv.reader(data_file)
+            # Skip header
+            reader.__next__()
```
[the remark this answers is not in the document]
brendan-ai2: This causes a spurious linter warning: "R: 77,12: Do not raise StopIteration in generator, use return statement instead (stop-iteration-return)". Should I explicitly silence this?

allennlp/data/dataset_readers/event2mind.py
```diff
@@ -0,0 +1,159 @@
...
+                for xintent in xintents:
+                    for xreact in xreacts:
+                        for oreact in oreacts:
+                            yield self.text_to_instance(source_sequence, xintent, xreact, oreact)
+
+    @staticmethod
+    def _preprocess_string(tokenizer, string: str) -> str:
+        """
+        Taken from:
+        https://github.com/maartensap/event2mind-internal/blob/master/code/modeling/utils/preprocess.py#L80.
```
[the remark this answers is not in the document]
brendan-ai2: Done

allennlp/data/dataset_readers/event2mind.py
```diff
@@ -0,0 +1,159 @@
...
+        return " ".join(retval)
+
+    def _build_target_field(self, target_string: str) -> TextField:
+        processed = self._preprocess_string(self._target_tokenizer, target_string)
+        tokenized_target = self._target_tokenizer.tokenize(processed)
+        tokenized_target.insert(0, Token(START_SYMBOL))
+        tokenized_target.append(Token(END_SYMBOL))
+        return TextField(tokenized_target, self._target_token_indexers)
+
+    @overrides
```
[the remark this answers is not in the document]
brendan-ai2: Done. I thought I was having some kind of issue with the `# type: ignore` pragma, but it seems working now...

allennlp/models/event2mind.py
```diff
@@ -0,0 +1,467 @@
...
+            top_log_probabilities, predicted_classes = cleaned_log_probabilities.topk(k)
+            # TODO(brendanr): Normalize for length?
+            # (batch_size * k, k)
+            expanded_last_log_probabilities = log_probabilities.\
+                unsqueeze(2).\
+                expand(batch_size, k, k).\
+                reshape(batch_size * k, k)
+            summed_top_log_probabilities = top_log_probabilities + expanded_last_log_probabilities
+
+            reshaped_top_log_probabilities = summed_top_log_probabilities.reshape(batch_size, k * k)
```
DeNeutoy: It's confusing that `reshaped_top_log_probabilities` now refers to `reshaped_summed_top_log_probabilities` given that you have another variable called `top_log_probabilities`

allennlp/models/event2mind.py
```diff
@@ -0,0 +1,467 @@
...
+            decoder_input = target_embedder(last_predictions)
+            decoder_hidden = decoder_cell(decoder_input, decoder_hidden)
+            # (batch_size * k, num_classes)
+            output_projections = output_projection_layer(decoder_hidden)
+
+            # (batch_size * k, num_classes)
+            class_log_probabilities = F.log_softmax(output_projections, dim=-1)
+
+            # (batch_size * k, num_classes)
+            last_predictions_expanded = last_predictions.unsqueeze(-1).expand(batch_size * k, num_classes)
```
DeNeutoy: From here until the end of the loop needs considerably more comments. What you are doing is pretty complicated and a slightly non-intuitive way of doing beam search, so you really need to spell out what's happening. I've added some comments I would add below.

allennlp/models/event2mind.py
```diff
@@ -0,0 +1,467 @@
...
+            cleaned_log_probabilities = torch.where(
```
DeNeutoy: Perhaps add the following comment:

"Here we are finding any beams where we predicted the end token in the previous timestep and replacing the distribution with a one-hot distribution, forcing the beam to predict the end token from now on."

allennlp/models/event2mind.py
```diff
@@ -0,0 +1,467 @@
...
+                    last_predictions_expanded == self._end_index,
+                    log_probs_after_end,
+                    class_log_probabilities
+            )
+
+            # (batch_size * k, k), (batch_size * k, k)
+            top_log_probabilities, predicted_classes = cleaned_log_probabilities.topk(k)
+            # TODO(brendanr): Normalize for length?
+            # (batch_size * k, k)
+            expanded_last_log_probabilities = log_probabilities.\
```
DeNeutoy: Add a comment:

"Here we expand the last log probabilities to (batch_size * beam_size, beam_size) so that we can add them to the current log probs for this timestep. This lets us maintain the log probability of each element on the beam."

allennlp/models/event2mind.py
```diff
@@ -0,0 +1,467 @@
...
+            restricted_beam_log_probs, restricted_beam_indices = reshaped_top_log_probabilities.topk(k)
+            # TODO(brendanr): Something about this is weird. Why do
+            # restricted_predicted_classes == restricted_beam_indices?
+            restricted_predicted_classes = reshaped_predicted_classes.gather(1, restricted_beam_indices)
+
+            log_probabilities = restricted_beam_log_probs
+            predictions.append(restricted_predicted_classes)
+            backpointer = restricted_beam_indices / k
+            backpointers.append(backpointer)
+            expanded_backpointer = backpointer.unsqueeze(2).expand(batch_size, k, self._decoder_output_dim)
```
DeNeutoy: I don't think you should need to expand to the `self._decoder_output_dim` dimension here - possibly just unsqueezeing would do.

allennlp/models/event2mind.py
```diff
@@ -0,0 +1,467 @@
...
+            expanded_last_log_probabilities = log_probabilities.\
+                unsqueeze(2).\
+                expand(batch_size, k, k).\
+                reshape(batch_size * k, k)
+            summed_top_log_probabilities = top_log_probabilities + expanded_last_log_probabilities
+
+            reshaped_top_log_probabilities = summed_top_log_probabilities.reshape(batch_size, k * k)
+            reshaped_predicted_classes = predicted_classes.reshape(batch_size, k * k)
+            restricted_beam_log_probs, restricted_beam_indices = reshaped_top_log_probabilities.topk(k)
+            # TODO(brendanr): Something about this is weird. Why do
```
DeNeutoy: Is this always the case? This makes me think this is incorrect, because taking the indices from the topk classes with respect to logprobs so far should be different from taking the topk logprobs from this iteration.

allennlp/models/event2mind.py
```diff
@@ -0,0 +1,467 @@
...
+            # (batch_size, num_classes)
+            output_projections = output_projection_layer(decoder_hidden)
+            # list of (batch_size, 1, num_classes)
+            step_logits.append(output_projections.unsqueeze(1))
+        # (batch_size, num_decoding_steps, num_classes)
+        logits = torch.cat(step_logits, 1)
+        target_mask = get_text_field_mask(target_tokens)
+        return self._get_loss(logits, targets, target_mask)
+
+    def beam_search(self,
```
DeNeutoy: This method is too large to go untested - perhaps one that you can add would be to check that using k=1 is equal to calling `greedy_search`. You could add this to the test for the model.

allennlp/models/event2mind.py
```diff
@@ -0,0 +1,467 @@
...
+            Size of the beam.
+        num_decoding_steps : ``int``, required
+            Maximum sequence length.
+        batch_size : ``int``, required
+            Size of the batch, which is distinct from the size of the beam.
+        target_embedder : ``Embedding``, required
+           Used to embed the token predicted at the previous time step.
+        decoder_cell: ``GRUCell``, required
+           The recurrent cell used at each time step.
+        output_projection_layer: ``Linear``, required
```
DeNeutoy: Add a return docstring with the shapes of the tensors you are returning.

brendan-ai2:
Good feedback regarding the beam search. I'll address those comments. As for your top level ones:

1. Valid point. I initially attempted this based on the naive view that everything should be batched when possible. I'll try to do some benchmarking to show the utility. If there is actually a significant speed-up it would strike me as a good tool to generalize and keep in the library.

Not sure I understand your specific point about the length of the decoded sequences. They are quite short, but the validation still takes some time which could be worsened by a constant factor equal to the batch size if we drop this.

2. Doesn't that code explicitly says not to use it for this type of model? See https://github.com/allenai/allennlp/blob/master/allennlp/nn/decoding/__init__.py#L6. Maybe @matt-gardner could shed some light on whether that comment should be disregarded?

matt-gardner:
Whether batching helps depends on a lot of things, and so, yeah, it's good to benchmark it and see, if that's not much work.

For using `nn.decoding.beam_search`, that's getting moved to `state_machines.beam_search`, and it's specifically just for state-machine-based models.  There are almost certainly much more efficient ways of handling beam search for standard seq2seq models, and it's probably overly complicating things to try to use the same API for both.  I say that with the major caveat that I haven't really looked at this code, though.

_Maybe_ it's feasible to do what @DeNeutoy suggests and make a `State` and a `TransitionFunction` for a standard seq2seq model.  The main issue is that for the `state_machines.beam_search` code, we're explicitly _not_ batching or grouping anything when aggregating results at each timestep, and so this creates some intermediate objects that you could do without if you have a model that's easier to batch: https://github.com/allenai/allennlp/blob/0b7bb205a78cfef32e9122057df7fe46568a2f0d/allennlp/nn/decoding/beam_search.py#L62-L64

DeNeutoy:
Ok - scratch the idea of using beam search from the semantic parsing stuff. That was a bad idea.

Un-batching it might actually allow you to make the code faster, because it would be much easier to maintain a beam for a single example, and importantly make it very easy to drop beam elements that are finished. I suspect that you are wasting a fair chunk of computation here because you still do matrix multiplications for beam elements which are finished. 

One other point I forgot to mention - you do need to average the log probs with respect to the length of the beam. Otherwise the topk part doesn't quite work correctly, because it is biased toward longer sequences.

matt-gardner:
I'm actually not certain about normalizing beam scores by sequence length.  We don't do this in the semantic parsing code.  Do you know of any work that actually compares this for NMT systems or other seq2seq models?

DeNeutoy:
See table 2 of the google machine translation paper:
https://arxiv.org/pdf/1609.08144.pdf

Also related: http://web.engr.oregonstate.edu/~huanlian/papers/optimal_beam_search.pdf

matt-gardner:
Thanks.  Interesting that this is important for NMT, but hurts you for semantic parsing - maybe because the semantic parses are a lot more constrained?  Given the nature of the event2mind output, I wouldn't just assume that it's more like NMT than semantic parsing.  Seems worth experimenting with.

brendan-ai2:
The original Event2Mind implementation does not normalize by sequence length. As I was aiming to reproduce their results I took the same approach. Doesn't mean it's not worth experimenting with, but that's the rationale. I'll add a comment to this effect. (See https://github.com/maartensap/event2mind-internal/blob/bfdcf8c3b79e4aa5338a4f4a04a5736b89ee473e/code/modeling/decoder_seq.py#L61 if you have access.)

allennlp/models/event2mind.py
```diff
@@ -0,0 +1,529 @@
...
+            # For the gather below.
+            expanded_backpointer = backpointer.unsqueeze(2).expand(batch_size, width, self._decoder_output_dim)
+            # Keep only the pieces of the hidden state corresponding to the
+            # ancestors created this iteration.
+            decoder_hidden = decoder_hidden.\
+                    reshape(batch_size, width, self._decoder_output_dim).\
+                    gather(1, expanded_backpointer).\
+                    reshape(batch_size * width, self._decoder_output_dim)
+
+        if len(predictions) != num_decoding_steps:
```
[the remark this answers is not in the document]
DeNeutoy: Not a big deal, but I don't think they assert anything about code correctness which isn't already checked in your test case. It's not as though the values a user passed here could cause different behaviour, so the asserts seem somewhat invalid.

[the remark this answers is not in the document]
brendan-ai2: It's important to make a distinction between error handling (via raising exceptions) and assertions. Handling bad user input or a flaky network connection is the domain of error handling. Checking the internal consistency of code or data structures is the domain of assertions. Such assertions will often look fairly trivial and _should_ be invariant to user input. See https://en.wikipedia.org/wiki/Assertion_(software_development)#Comparison_with_error_handling.

Also, I'm not suggesting we add such trivial checks all the time. It's just that these particular ones helped during development and -- much like with a simple unit test -- I'm loathe to remove them now. Assertions can trivially be disabled at runtime for performance. (Which is why they shouldn't be used for error handling, incidentally.)

PR #1680, opened 2018-08-28
Fix links in docs and improve check_links.py

epwalsh:
There were several links in the documentation that were broken, yet were not caught with `scripts/check_links.py`, since they were "local path" links to files within the directory and not urls. I fixed those links and improved `scripts/check_links.py` to handle local paths differently and catch paths to files that don't exist.

### DIFF
--- a/scripts/check_links.py
+++ b/scripts/check_links.py
@@ -67,14 +67,14 @@ def main():
 
     with Pool(processes=10) as pool:
         results = pool.map(link_ok, [match for match in list(all_matches)])
-    unreachable_results = [result for result in results if not result[1]]
+    unreachable_results = [match_tuple for match_tuple, success in results if not success]
 
     if unreachable_results:
         print(f"Unreachable links ({len(unreachable_results)}):")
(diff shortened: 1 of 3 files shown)


scripts/check_links.py
```diff
@@ -1,57 +1,83 @@
...
 
     if unreachable_results:
-        print(f"Unreachable Links ({len(unreachable_results)}):")
-        for index, result in enumerate(unreachable_results):
-            print("  " + str(index))
-            print("  Source: " + result[0].source)
-            print("  Name: " + result[0].name)
-            print("  Link: " + result[0].link)
+        print(f"Unreachable links ({len(unreachable_results)}):")
+        for result in unreachable_results:
```
matt-gardner: Minor nit: `for match_tuple, _ in unreachable_results:` would be a bit better than doing `result[0]` below, just because then `result[0]` has a semantically meaningful name.

epwalsh: Good point!

PR #1682, opened 2018-08-28
Optimise softmax memory usage

DeNeutoy:
Because of the way @matt-gardner switched to using the masked_fill for masking things, we can now be more efficient with memory usage/operations because 1) We don't have to cast to float and 2) we don't have to do lots more multiplication operations than we need to.

### DIFF
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -192,8 +192,8 @@ def masked_softmax(vector, mask):
         result = torch.nn.functional.softmax(vector, dim=-1)
     else:
         # To limit numerical errors from large vector elements outside the mask, we zero these out.
-        result = torch.nn.functional.softmax(vector * mask, dim=-1)
-        result = result * mask
+        result = torch.nn.functional.softmax(replace_masked_values(vector, mask, 0.0), dim=-1)
+        result = replace_masked_values(result, mask, 0.0)
         result = result / (result.sum(dim=1, keepdim=True) + 1e-13)
     return result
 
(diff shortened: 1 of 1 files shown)


allennlp/nn/util.py
```diff
@@ -192,8 +192,8 @@ def masked_softmax(vector, mask):
         result = torch.nn.functional.softmax(vector, dim=-1)
     else:
         # To limit numerical errors from large vector elements outside the mask, we zero these out.
-        result = torch.nn.functional.softmax(vector * mask, dim=-1)
-        result = result * mask
+        result = torch.nn.functional.softmax(replace_masked_values(vector, mask, 0.0), dim=-1)
```
matt-gardner: Did you verify that this actually reduces memory usage?  I think this actually _adds_ an operation, because `replace_masked_values` does a subtraction, then a `masked_fill()`.

DeNeutoy: Before changing this, I get an OOM error at 
`mask = mask.expand_as(tensor).contiguous().float()`, whereas after I get it in the matrix multiplication afterward.

handsomezebra: Something to consider, most of times we don't need to expand mask if it is broadcastable with tensor.

DeNeutoy: Right - unfortunately we have many cases where they aren't, such as in the multi-headed attention.

matt-gardner: We actually _can_ remove that expand, as an `unsqueeze()` is enough.  #1687 removes the `last_dim_applicator` entirely, which should further help your memory problems.

matt-gardner:
If this is more efficient than the multiplies, we should still make the change, right?  #1687 doesn't just replace this PR.

PR #1687, opened 2018-08-29
Removing last_dim_*softmax

matt-gardner:
Back when we rolled our own `masked_softmax` and `masked_log_softmax`, we had to treat higher-dimension inputs separately.  We don't need to do that anymore, and so we should deprecate `last_dim_softmax` and `last_dim_log_softmax` in favor of just using the other versions, which will be more efficient.


j6mes:
I'm upgrading a system from this deprecated version of allennlp. Does `masked_softmax` have the same behaviour in newer versions?

matt-gardner:
Yes, the behavior is the same.

PR #1688, opened 2018-08-29
Add BMES constrain to is_transition_allowed function

huntzhan:
closes #1685

PR #1692, opened 2018-08-30
Make SpanBasedF1Measure support BMES 

huntzhan:
Following #1688 
Since `CrfTagger` model use `SpanBasedF1Measure` to calculate span f1, we need to add BMES support to `SpanBasedF1Measure` as well.


huntzhan:
@matt-gardner could you help review this one? thanks.

allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -371,3 +371,72 @@ def process_stack(stack, out_stack):
...
+
+    Returns
+    -------
+    spans : List[TypedStringSpan]
+        The typed, extracted spans from the sequence, in the format (label, (span_start, span_end)).
+        Note that the label `does not` contain any BIO tag prefixes.
+    """
+    def extract_bmes_tag_label(text):
+        bmes_tag = text[0]
+        label = text.partition('-')[2]
```
matt-gardner: How are these tags formatted?  Is it just `B-label`, `M-label`, etc?  If so, you don't need `text.partition('-')[2]`, you can just do `text[2:]`.

Issue #1571, opened 2018-08-03
/scripts/check_requirements.py no longer works correctly


schmmd:
We merged requirements.txt and requirements_test.txt and the script checks for any discrepancies between the two.  Since the dependencies in requirements_test.txt are not in setup.py it always fails.  See https://github.com/allenai/allennlp/issues/1561 for more information.

epwalsh:
I was wondering if it would make sense to gather requirements in the `requirements.txt` file by the "section" that they are in, like "TESTING-RELATED PACKAGES", "LIBRARIES USED IN SCRIPTS", etc., and that way we could ignore certain sections when comparing to the `setup.py` file. 

The only downside I can see is that the section format scheme (ex. "#### SECTION NAME ####") would have to be kept consistent, but I imagine those sections are going to be fairly static, and there could also be another comment in `requirements.txt` that says "Hey, if you make a new section, make sure the section is formatted consistently".

Anyway, I've implemented a check like this before so I'd be willing to make a PR if this is how you want to address it.

matt-gardner:
This sounds reasonable to me, but I'd wait for a bit before starting this, in case someone else pipes in with a dissenting opinion.

PR #1694, opened 2018-08-30
Upgrade Flask to 0.12.4 (fixes bug)

epwalsh:
Since Flask version 0.12.4, a bug was fixed that happened when using `flask.Blueprint`.


joelgrus:
can you update it in `setup.py` too

epwalsh:
Whoops! I always forget that. If no one has any objections, I think I'll make a PR for https://github.com/allenai/allennlp/issues/1571

PR #1695, opened 2018-08-30
Add option to keep sentence boundaries in Elmo

nelson-liu:
This adds the ability to get the embeddings of the BOS/EOS tokens, which is useful if you want to evaluate ELMo on language modeling. The defaults shouldn't have changed.


nelson-liu:
I'm not sure if I have to change `_ElmoBiLM` as well (https://github.com/allenai/allennlp/blob/master/allennlp/modules/elmo.py#L643) --- doesn't seem like it?

PR #1698, opened 2018-08-30
Release v0.6.1

schmmd:


allennlp/version.py
```diff
@@ -1,6 +1,6 @@
 _MAJOR = "0"
 _MINOR = "6"
-_REVISION = "1-unreleased"
+_REVISION = "2-unreleased"
```
brendan-ai2: What's the significance of "unreleased" here?

schmmd: @brendan-ai2 after the release is cut we set it to an unreleased version, so if you pip install from master it's clear you didn't install a released version.  The tag should be associated with the previous commit (although I'll verify that).

schmmd: Yup--the correct commit is tagged.

schmmd: https://github.com/allenai/allennlp/pull/1698/commits/4920249af097d1557be6ece3e5ce80d1f6ea0333#diff-e66fadcdcab33b8d2e449e323706e936R3

brendan-ai2: Cool. Thanks for explaining :)

PR #1699, opened 2018-08-31
(Re-)implement requirements check

### LINKED ISSUE #1571: /scripts/check_requirements.py no longer works correctly

epwalsh:
fixes #1571


setup.py
```diff
@@ -132,7 +131,16 @@
       ],
       scripts=["bin/allennlp"],
       setup_requires=setup_requirements,
-      tests_require=['pytest', 'flaky', 'responses>=0.7', 'jupyter'],
+      tests_require=[
```
epwalsh: If my understanding is correct, these packages listed under `tests_require` should correspond to packages listed in the section "TESTING-RELATED PACKAGES" in `requirements.txt`

nelson-liu: 👍 you're correct

epwalsh:
Another, potentially cleaner solution to this would be to have `setup.py` automatically parse the dependencies listed in `requirements.txt` by section, and add them appropriately. This would be pretty easy, since the code needed to parse `requirements.txt` is already part of this PR.

setup.py
nelson-liu: accidentally deleted my original comment, sorry.

I think it's a bit more nuanced than that --- a lot of the things in the `TESTING-RELATED PACKAGES` aren't actually needed to run the tests (e.g., codecov). They're just used in CI. I think the idea was to only put things that are actually needed to run the tests in `tests_require` (so you should be able to remove `mypy`, `codecov`, `pytest-cov`, `pylint`, `coverage` (I'll have to check on `aiohttp` and `moto`)

epwalsh: So in that case, would `tests_require` correspond to `ESSENTIAL TESTING-RELATED PACKAGES`?

nelson-liu: I think we can actually remove `aiohttp`, since it was used to run tests for sanic. however, I think we've switched over to flask (@joelgrus / @schmmd can confirm).

nelson-liu: Yes, that's fair. though `conllu` should be in  `#### ESSENTIAL LIBRARIES FOR MAIN FUNCTIONALITY ####`, would you mind moving it?

setup.py
```diff
@@ -111,7 +111,6 @@
           'cffi==1.11.2',
           'awscli>=1.11.91',
           'boto3',
-          'moto==1.3.4',
```
nelson-liu: moto should stay here (and moto should be moved into `ESSENTIAL TESTING-RELATED PACKAGES`). These are the requirements that get bundled in `pip install allennlp`, and `allennlp test-install` will fail out of the box if it doesn't install moto (since it's imported in the test https://github.com/joelgrus/allennlp/blob/fc83493e4c46e9bfc8e4f7bd5a5347f01673b78d/allennlp/tests/common/file_utils_test.py#L10)

setup.py
epwalsh: Got it!

schmmd: @nelson-liu I have no idea by now--I'd cut a PR and see what CI says.

setup.py
epwalsh: Gotcha. So wouldn't that be true of `pytest`, `flaky`, and `responses` as well, since they are all imported by modules in `tests/`?

epwalsh: In which case, everything under `tests_require` would already be under `install_requires`, right?

nelson-liu: yup, `pytest`, `flaky`, and `responses` are already in `install_requires`.

> In which case, everything under tests_require would already be under install_requires, right?

Excepting jupyter (some tests run notebooks, but we skip those by default in `test-install`), yes. So maybe we should move jupyter to "ESSENTIAL TESTING-RELATED PACKAGES" as well.

The key is that for jupyter, we want to make clear that it's needed to run the tests (for now, I think there are efforts to rework tutorials into vanilla md), but also not install it by default since it's massive.

epwalsh: Ah, got it! Thanks for the clarification. I'll need to make a couple of adjustments

setup.py
```diff
@@ -132,7 +132,13 @@
       ],
       scripts=["bin/allennlp"],
       setup_requires=setup_requirements,
-      tests_require=['pytest', 'flaky', 'responses>=0.7', 'jupyter'],
+      tests_require=[
+          'pytest',
+          'flaky',
+          'responses>=0.7',
+          'moto==1.3.4',
```
epwalsh: I added moto here as well to be explicit that is used only for testing

epwalsh:
So, this won't catch every possible error with `setup.py` vs `requirements.txt`, but it will make sure that there are no duplicates within either `setup.py` or `requirements.txt`, and that each package that is listed in both `setup.py` and `requirements.txt` having matching versions. It also ensures that any requirement listed as "ESSENTIAL" in `requirements.txt` appears under the "install_requires"
field in `setup.py`.

schmmd:
Thanks @epwalsh !

PR #1703, opened 2018-08-31
Rename SpanPruner -> Pruner, remove -infs

### LINKED ISSUE #1696: SpanPruner produces -inf scores when an instance has too few spans
**Describe the bug**
In the case when there are fewer than `num_spans_to_keep` total spans in the original text, some padding makes its way into the `top_span_scores` output of `SpanPruner` with scores of `-inf`. Even though the `top_spans_mask` output is correct, this is a problem because multiplying the scores by the mask produces `nan` in those slots instead of the desired `0.0`.

**To Reproduce**
In python REPL:
```python
import torch
from allennlp.modules.span_pruner import SpanPruner
emb = torch.ones([1, 2, 1]) # batch size 1, 2 spans, embedding size 1
scorer = torch.nn.Linear(1, 1)
mask = torch.tensor([1, 0]).view(1, 2).float() # only 1 span is present in the instance
pruner = SpanPruner(scorer)
_, _, _, scores = pruner(emb, mask, 2)
print(scores)
```
For me, outputs:
```python
tensor([[[ 0.5783],
         [   -inf]]])
```
though of course the non-inf number is arbitrary.

**Expected behavior**
I think in this case we should replace the `-inf`s with `-1`. Because of this issue I had a loss of `nan` that I had to debug until I found this. It should be an easy fix in `SpanPruner`. BTW, there's nothing particular to spans in `SpanPruner`, is there? Might as well just call it `Pruner`, right?

 - OS: Reproduced on mac OS 10.13.3 on CPU as well as Ubuntu 16.04.4 LTS on GPU (V100).
 - Python version: 3.6.5 and 3.6.6.
 - AllenNLP version: v0.5.0, but looking at the source code on master I assume this remains an issue.
 - PyTorch version: 0.4.0

**Additional context**
My guess is this hasn't come up before because span pruning was only used with long texts where this doesn't ever happen. It came up for me because I'm using span pruning with a sentence-level model where a few of the sentences have only 2 tokens and are batched with 4-token sentences.

matt-gardner:
Fixes #1696.  I decided against modifying the coref model in the places that it uses the log mask, because that was more complicated than I wanted to get to, and the coref model's loss function is able to handle the `-infs` without a problem.

Git is saying that I removed `span_pruner.py` and added `pruner.py`; that's just because I made cosmetic changes to a ton of lines, removing references to "spans".  The only change in logic is using `replace_masked_values` instead of `mask.log()`.


allennlp/modules/__init__.py
```diff
@@ -15,7 +15,10 @@
 from allennlp.modules.seq2seq_encoders import Seq2SeqEncoder
 from allennlp.modules.seq2vec_encoders import Seq2VecEncoder
 from allennlp.modules.similarity_functions import SimilarityFunction
-from allennlp.modules.span_pruner import SpanPruner
+from allennlp.modules.pruner import Pruner
+# TODO(mattg): This line is to keep backwards compatibility.  I'm not sure how to give a
+# deprecation warning on using this import, but we should remove this around version 0.8.
+from allennlp.modules.pruner import Pruner as SpanPruner  # pylint: disable=reimported
```
DeNeutoy: This doesn't actually preserve backward compatibility, e.g what if someone had `from allennlp.modules.span_pruner import ....`.

DeNeutoy: Maybe this isn't a huge deal though, up to you.

matt-gardner: Yeah, I'm not sure there's a way to fix that.  I'm hoping that most people will have imported from the base module, which is what we typically do.  @joelgrus, do you know of a better way we can handle backwards compatibility with renaming modules like this?

DeNeutoy: I think you can just leave the `span_pruner.py` there and import the `Pruner` as `SpanPruner`?

DeNeutoy: That would also be a natural place for the depreciation warning.

matt-gardner: Then I have to have docs for it, because of our documentation check...  Ugh.  Ok.

matt-gardner:
@DeNeutoy, I thought that test failure was a fluke, as it's not related to anything changed in this PR.  But it's failed three times in a row now.  Any ideas about what's going on there?

allennlp/modules/__init__.py
DeNeutoy: just add it to the `docs to ignore` dictionary

DeNeutoy:
No idea... it's quite high precision, does it go away if you drop it to 6?

allennlp/modules/__init__.py
matt-gardner: Thanks, I had forgotten about that.

PR #1705, opened 2018-08-31
make openai transformer byte pair indexer add to the vocab

### LINKED ISSUE #1700: open transformer byte pair indexer should stick its mapping in the vocabulary
our TokenIndexers use the Vocabulary to build up their mappings token <-> index, and so later on if a model needs access to that mapping it can get it through the vocabulary.

the openai transformer byte pair indexer is an oddball in that it uses a precomputed mapping loaded from a file; hence, the corresponding implementation ignores the vocabulary. this means that there's no easy way to access the mapping from e.g. your model.

so probably the byte pair indexer should put its mapping into the vocabulary so that it can be accessed there, same as with every other token indexer

joelgrus:
fixes #1700


allennlp/data/token_indexers/openai_transformer_byte_pair_indexer.py
```diff
@@ -27,7 +27,10 @@ def __init__(self,
                  encoder: Dict[str, int] = None,
                  byte_pairs: List[Tuple[str, str]] = None,
                  n_ctx: int = 512,
-                 model_path: str = None) -> None:
+                 model_path: str = None,
+                 namespace: str = 'openai_transformer') -> None:
```
matt-gardner: I started out wondering why there wasn't a docstring modification to add this parameter.  Then I clicked up and realized there was no parameter section in the docstring.  Then I wondered what in the world `n_ctx` meant, and what it was doing (`num_contexts`, probably?  `num_context_words`?  but what does that have to do with BPE?).  After looking through the code and remembering a prior conversation, I _think_ it's here because the model can only handle 512 tokens at a time (`max_sequence_length`?), and you're hosed if you have more than that, and you're catching it here to give an early error message...?

PR #1707, opened 2018-09-04
Revert "Save learning rate scheduler to training state"

DeNeutoy:
Reverts allenai/allennlp#1650

### DIFF
--- a/allennlp/tests/training/learning_rate_schedulers_test.py
+++ b/allennlp/tests/training/learning_rate_schedulers_test.py
@@ -1,6 +1,6 @@
 # pylint: disable=no-self-use,invalid-name,protected-access
 
-from typing import Dict, Any
+from typing import Dict, List, Tuple, Any
 
 import torch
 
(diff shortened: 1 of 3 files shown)


DeNeutoy:
@matt-gardner this PR: https://github.com/pytorch/pytorch/pull/7201 means that with this change we require pytorch >= 0.4.1, not >= 0.4.0 (basically `state_dict` wasn't implemented for a LRScheduler which has no state, which breaks the api). I can either:
1. Revert this
2. send a PR which requires 0.4.1
3. Add a try/except block to avoid this distinction in the trainer.

What do you prefer?

matt-gardner:
Seems fine to require 0.4.1?

DeNeutoy:
True, but this is going to break a lot of people's installs if we do?

epwalsh:
Darn, sorry I missed that

matt-gardner:
I don't understand why this is different from previous versions, like when we upgraded from pytorch 0.3 to 0.4.  We have plenty of dependencies that we pin or upgrade with each release.

matt-gardner:
Because, when they `pip install allennlp==0.6.2`, pip will read our requirements and make sure that they have pytorch 0.4.1.  Right?  Am I missing something?

DeNeutoy:
I thought we'd had discussions about it being a bit annoying to pin minor versions of major dependencies in the past, but sure. see #1708

PR #1709, opened 2018-09-05
Grammar for a variable free language for WikiTableQuestions

pdasigi:
This PR contains a grammar for the language for WTQ in the [Memory Augmented Policy Optimization paper by Liang et al., 2018](https://arxiv.org/abs/1807.02322). I made a separate type declaration and a world for the new language. The world does not include an executor yet.
@matt-gardner I made a PR for just the grammar in the interest of making PRs smaller. I'll make another one for the executor when it's done.


allennlp/semparse/type_declarations/wikitables_variable_free.py
```diff
@@ -0,0 +1,126 @@
...
+STRING_TYPE = NamedBasicType("STRING")
+
+BASIC_TYPES = {ROW_TYPE, COLUMN_TYPE, NUMBER_TYPE, DATE_TYPE, STRING_TYPE}
+STARTING_TYPES = {NUMBER_TYPE, DATE_TYPE, STRING_TYPE}
+
+# Complex types
+# Type for selecting the value in a column in a set of rows. "select" and "mode" functions.
+SELECT_TYPE = ComplexType(ROW_TYPE, ComplexType(COLUMN_TYPE, STRING_TYPE))
+
+# Type for filtering rows given a column. "argmax", "argmin" and "same_as" (select all rows with the
```
matt-gardner: I don't understand this one.  Don't you need a function for the argmax?

allennlp/semparse/type_declarations/wikitables_variable_free.py
```diff
@@ -0,0 +1,126 @@
...
+                                                          ComplexType(COLUMN_TYPE, ROW_TYPE)))
+
+# "filter_in" and "filter_not_in"
+ROW_FILTER_WITH_COLUMN_AND_STRING = ComplexType(ROW_TYPE,
+                                                ComplexType(STRING_TYPE,
+                                                            ComplexType(COLUMN_TYPE, ROW_TYPE)))
+
+ROW_FILTER = ComplexType(ROW_TYPE, ROW_TYPE)  # first, last, previous, next etc.
+
+COUNT_TYPE = ComplexType(ROW_TYPE, NUMBER_TYPE)
```
matt-gardner: You can only count rows?

allennlp/semparse/type_declarations/wikitables_variable_free.py
```diff
@@ -0,0 +1,126 @@
...
+def add_common_name_with_type(name, mapping, type_signature):
+    COMMON_NAME_MAPPING[name] = mapping
+    COMMON_TYPE_SIGNATURE[mapping] = type_signature
+
+
+add_common_name_with_type("all_rows", "R", ROW_TYPE)
+
+# <r,<l,s>>
+add_common_name_with_type("select", "S0", SELECT_TYPE)
+add_common_name_with_type("mode", "S1", SELECT_TYPE)
```
matt-gardner: This is "mode" in the sense of "mode of the distribution"?

allennlp/semparse/type_declarations/wikitables_variable_free.py
```diff
@@ -0,0 +1,126 @@
...
+# Date function takes three numbers and makes a date
+DATE_FUNCTION_TYPE = ComplexType(NUMBER_TYPE, ComplexType(NUMBER_TYPE, ComplexType(NUMBER_TYPE,
+                                                                                   DATE_TYPE)))
+
+
+COMMON_NAME_MAPPING = {}
+COMMON_TYPE_SIGNATURE = {}
+
+
+def add_common_name_with_type(name, mapping, type_signature):
```
matt-gardner: Type hints on the arguments here.  I'd also probably recommend just calling this `add_function`.

allennlp/semparse/worlds/wikitables_variable_free_world.py
```diff
@@ -0,0 +1,174 @@
+"""
+We store the information related to execution of context sensitive execution of logical forms here.
```
matt-gardner: Fix wording here.

allennlp/semparse/worlds/wikitables_variable_free_world.py
```diff
@@ -0,0 +1,174 @@
...
+We assume that the logical forms are written in the variable-free language described in the paper
+'Memory Augmented Policy Optimization for Program Synthesis with Generalization' by Liang et al.
+The language is the main difference between this class and `WikiTablesWorld`. Also, this class defines
+an executor for the variable-free logical forms.
```
matt-gardner: I think the executor code should be in a different file.

Also, an idea that Chen suggested that we don't have to do now, but would be nice if we could do eventually: when you're defining the language in the type declaration, also implement the execution of each of those functions, and have it all in one place.  I think there's a clean way to do this, but let's worry about that later.

matt-gardner: Hmm, actually, having the execution depend on the context makes doing it that way pretty hard...  Still, it'd be nice to think about how to make the language-definition process easier.

allennlp/semparse/worlds/wikitables_variable_free_world.py
```diff
@@ -0,0 +1,174 @@
...
+        # We don't need to recompute this ever; let's just compute it once and cache it.
+        self._valid_actions: Dict[str, List[str]] = None
+
+    def is_table_entity(self, entity_name: str) -> bool:
+        """
+        Returns ``True`` if the given entity is one of the entities in the table.
+        """
+        return entity_name in self._entity_set
+
+    def _get_curried_functions(self) -> Dict[Type, int]:
```
matt-gardner: I think this needs an `@overrides` decorator.

allennlp/semparse/worlds/wikitables_variable_free_world.py
```diff
@@ -0,0 +1,174 @@
...
+
+    @overrides
+    def get_basic_types(self) -> Set[Type]:
+        return types.BASIC_TYPES
+
+    @overrides
+    def get_valid_actions(self) -> Dict[str, List[str]]:
+        if self._valid_actions:
+            return self._valid_actions
+        valid_actions = super().get_valid_actions()
```
matt-gardner: Looks like the caching logic should just live in the super class.

allennlp/semparse/worlds/wikitables_variable_free_world.py
```diff
@@ -0,0 +1,174 @@
...
+    def _map_name(self, name: str, keep_mapping: bool = False) -> str:
+        if name not in types.COMMON_NAME_MAPPING and name not in self.local_name_mapping:
+            if not keep_mapping:
+                raise ParsingError(f"Encountered un-mapped name: {name}")
+            if name.startswith("fb:row.row"):
+                # Column name
+                translated_name = "C%d" % self._column_counter
+                self._column_counter += 1
+                self._add_name_mapping(name, translated_name, types.COLUMN_TYPE)
+            elif name.startswith("fb:cell"):
```
matt-gardner: Do we really want to still use the particular mapping from cells and cell parts to strings that SEMPRE used?  I'm sure that's not what Chen did.  What do we gain from doing this?

allennlp/tests/semparse/type_declarations/wikitables_type_declaration_test.py
```diff
@@ -5,7 +5,7 @@
         ComplexType,
         ANY_TYPE,
         )
-from allennlp.semparse.type_declarations.wikitables_type_declaration import (
+from allennlp.semparse.type_declarations.wikitables_lambda_dcs import (
```
matt-gardner: Rename the file here, too.

allennlp/semparse/worlds/wikitables_variable_free_world.py
matt-gardner: In fact, looking at the logical forms, I'm pretty sure it'll make the executor more complicated than it needs to be.  We can simplify a bunch of things by removing the SEMPRE naming stuff.

allennlp/tests/semparse/worlds/wikitables_variable_free_world_test.py
```diff
@@ -0,0 +1,206 @@
...
+        check_productions_match(valid_actions['n'],
+                                ['-1',
+                                 '0',
+                                 '1',
+                                 '2013',
+                                 '[<r,<l,n>>, r, l]',
+                                 '[<r,<r,<l,n>>>, r, r, l]',
+                                 '[<r,n>, r]'])
+
+        check_productions_match(valid_actions['r'],
```
matt-gardner: How does this language handle conjunctions or disjunctions?  I'd expect to see something like `r -> [<r,<r,r>>, r, r]`, with productions like "and" and "or".

allennlp/semparse/type_declarations/wikitables_variable_free.py
pdasigi: No. `argmax` and `argmin` in the MAPO language only take a set of rows and a column (with numerical values). The column gives the values being compared, and a row is returned.

allennlp/semparse/type_declarations/wikitables_variable_free.py
pdasigi: Yes. See Appendix A in the MAPO paper.

allennlp/semparse/type_declarations/wikitables_variable_free.py
pdasigi: Yes, the most frequent value.

allennlp/semparse/type_declarations/wikitables_variable_free.py
pdasigi: Added type hints. The things being added to the dicts `COMMON_NAME_MAPPING` and `COMMON_TYPE_SIGNATURE` are not all functions (`all_rows` is a constant).

allennlp/semparse/worlds/wikitables_variable_free_world.py
pdasigi: Done.

allennlp/semparse/worlds/wikitables_variable_free_world.py
pdasigi: I agree there are cleaner ways to do this. For example, it would be nice if we only defined the executor, and the type declaration could be inferred from it automatically.

allennlp/semparse/worlds/wikitables_variable_free_world.py
pdasigi: Done.

allennlp/semparse/worlds/wikitables_variable_free_world.py
pdasigi: Done.

allennlp/semparse/type_declarations/wikitables_variable_free.py
matt-gardner: Ok.  Can you update the comment here to make that more clear?

allennlp/semparse/type_declarations/wikitables_variable_free.py
matt-gardner: Or see a nice comment here explaining that...? =)

allennlp/semparse/worlds/wikitables_variable_free_world.py
pdasigi: I reused the mapping as I wanted to use the `TableQuestionKnowledgeGraph` as is. That's because I think the functionality in that class, such as mapping spans in the question to entities for agenda computation, string normalization etc. are applicable to this world as well. Moreover, cells are typed as strings in the new grammar. It would be nice if the executor can distinguish them from normal strings (when performing string matching in functions such as `filter_in`), and the prefixes from SEMPRE's convention make it easy to do so.
What makes you think that this convention will make the executor more complicated than it needs to be?

allennlp/semparse/type_declarations/wikitables_variable_free.py
matt-gardner: But, I did just look at appendix A in the paper, and it makes it clear that there are different kinds of columns.  We should have those type distinctions, right?

allennlp/tests/semparse/type_declarations/wikitables_type_declaration_test.py
pdasigi: Done.

allennlp/semparse/worlds/wikitables_variable_free_world.py
matt-gardner: You'll have to remove prefixes, and do some kind of search to find the appropriate cell or column, because the names are mangled (or store the mapping and give that to the executor, though that still requires changing things).

matt-gardner: And what's a "normal string" in this language?  When can you produce a string that's not a cell, part, or column?

Also, it'd be nice to shed the crazy `fb:row.row.X` notation.

allennlp/tests/semparse/worlds/wikitables_variable_free_world_test.py
pdasigi: It doesn't (see Appendix A in the paper). I agree that this is limiting, and I'm curious to see how this affects the produced logical forms. Recall that our NLVR language does not have `and` and `or` too, which meant we could not translate certain kinds of sentences, but it made the language very simple.

allennlp/semparse/type_declarations/wikitables_variable_free.py
pdasigi: Sure, I'll add more comments. They make a distinction between string, number, and date columns. I was going to have the executor handle this distinction. That is, for example, if you pass a column without numbers to the `max` function, return null (I think that's what LambdaDCS does if you apply a `cell.cell.number` on a string value). I'm not sure how this can be done with type declaration. Any ideas?

allennlp/semparse/type_declarations/wikitables_variable_free.py
pdasigi: Done.

allennlp/semparse/type_declarations/wikitables_variable_free.py
pdasigi: Added a comment.

allennlp/semparse/worlds/wikitables_variable_free_world.py
pdasigi: I'd be happy to get rid of SEMPRE notation. But that means changing the `TableKnowledgeGraph`, or adding a similar class for the new language. Seems like an effort that's not worth it.
By normal strings, I essentially meant parts.

matt-gardner: What strings are allowed in this language?  I just looked at the MAPO paper, and it sure looks like the "strings" always come from the question, not the table.  That seems a lot more sane to me.

matt-gardner: This seems like a crucial question to me.  I just went through the paper trying to figure this out, and I couldn't find it.  So I sent an email to Chen asking him.  We'll see what he says.

If it's the case that the strings just come from the question, as I expect, then we don't need cells or cell parts at all, and we don't need to select between all of the cells when linking, and we don't have to do the KG embedding.  It simplifies a _whole_ lot of things.

pdasigi: Makes sense.

matt-gardner:
Ok, having now seen Chen's response about strings, I'd say that for this PR we should just add the column types to the language, and we can worry about fixing the strings in a separate PR.  It looks like there should be four column types: "column", "number column", "date column", and "string column", and each function should be careful about which column type it takes (with some of the functions allowing either date or number columns).

It would also be nice to differentiate between "a list of rows" and "a row" in the type system, but I don't think we can do that, because I _think_ the executor can return a list of size 1 that then can be used as "a row".  @crazydonkey200, if that's wrong, let us know.  That's one example where partial execution and dynamic typing could help you in @crazydonkey200's way of doing things.

pdasigi:
@matt-gardner I'm fine with adding different column types. But it will depend on actually identifying those column types. For example, if I make `filter_in` take a "string column" as an argument, and the `TableQuestionKnowledgeGraph` only returns "columns", the tests will break. I added a TODO for handling different column types, and will make another PR after @MurtyShikhar adds functionality to identify different column types.

crazydonkey200:
Hi. Just to confirm what Matt @matt-gardner has mentioned. I didn't use separate types to differentiate a list of rows or a row. There are only lists of rows. But some functions require the list to have more than one row and some require the list to have only one row, which is checked in a dynamic way in code assistance. 

This dynamic constraint in code assistance is based on the ability to execute a partial program / logical form. That's why the language is designed so that, every time an expression is finished, it can be executed. It can be seen as a linearized bottom-up derivation of the program / logical form and ensures that the generated program is executable. 

Hope this simpler language could improve the performance :)  Please let me know if there's other questions or anything I could help with.

crazydonkey200:
BTW, to support the incremental program generation and partial program execution, I didn't use the executor in SEMPRE for WikiTable, and instead wrote [a (much simplified) Lisp Interpreter](https://github.com/crazydonkey200/neural-symbolic-machines/blob/master/nsm/computer_factory.py) and [a table executor](https://github.com/crazydonkey200/neural-symbolic-machines/blob/master/nsm/executor_factory.py#L308). It provides `valid_tokens()` (returns the valid tokens for the next step to select from) and `read_token(token)` (reads in the selected token) as interface. 

These two modules can be used separately and only depend on basic Python libraries. Hope they can also be helpful for this project. @pdasigi @matt-gardner

pdasigi:
That's great! Thanks for the pointers, @crazydonkey200!

crazydonkey200:
Sure. Let me know if you would like to use or modify the interpreter and the executor, and I am happy to provide some help. @pdasigi

berlino:
Not sure if it's right place to post the question, apologize first if it is not. 

@matt-gardner Back to the problem of differentiating between "rows" and "a list of rows", I do not think of a reason why it cannot be done.  Say we have two types "MULTI-ROW" and "SINGLE_ROW". In the case of "the executor can return a list of size 1 that then can be used as a row" you mention about, we can simply define a multi-match type to include both types. This actually corresponds to the return type of filter functions, where the return value could be a row or a list of rows. 

In other functions, the row type is more obvious. argmax/argmin, mode, count should always take in "MULTI_ROW", while diff, previous/next should take in "SINGLE_ROW". The current implementation tends to produces many useless programs like previous(all_rows), argmax(previous(all_rows)), etc. These two types would be helpful to prune these fake programs. I'll make a commit if it makes sense to you. @pdasigi

matt-gardner:
That sounds like it might work; PR welcome.  Though, I'll warn you that I'm in the process of changing how all of the language definition / type system stuff works.  See #2281 for what I'm trying to do.  I have the NLVR language almost completely converted to this new system (see [here](https://github.com/matt-gardner/allennlp/blob/new_nlvr_executor/allennlp/semparse/domain_languages/nlvr_language.py)), and I'll be doing the WikiTableQuestions language next, hopefully getting to it in the next day or two.

PR #1710, opened 2018-09-05
Make bmes_tags_to_spans support ill-formed spans.

huntzhan:
Follows #1692


huntzhan:
CI passed.
@matt-gardner could you help review this one? Thanks.

allennlp/data/dataset_readers/dataset_utils/span_utils.py
```diff
@@ -401,42 +403,139 @@ def extract_bmes_tag_label(text):
...
+            if func_mismatch and last_label != cur_label:
+                func_mismatch(spans, state, **kwargs)
+            else:
+                func_match(spans, state, **kwargs)
+        else:
+            func = from_es_op[op_key]
+            func(spans, state, **kwargs)
 
-    if invalid:
-        raise InvalidTagSequence(tag_sequence)
```
matt-gardner: This change in logic seems way more complicated than it needs to be.  I think the new logic that you have is way harder to understand than what you had before, and I don't think we need the complicated state transition logic just to handle this simple transition system.  Can't you just add the tag sequence here?  If you really want to be smart about handling `["M", "E"]` as a single span, you can still do that with small modifications to the previous if/else system of handling the transition system.

My motivation: if there's a bug in this code somewhere down the line, or we decide we need to refactor things, we'll be responsible for maintaining it.  I think the logic that you had before is simple enough that I can maintain it without much work.  This new logic is _much_ too complex for me to want to maintain.  Please keep the tests that you wrote, but make minimal modifications to the old logic to get the tests to pass.

huntzhan: > This new logic is much too complex for me to want to maintain.

I totally understand your concern. But I think the code will be complicated somehow.
As an alternative, I wonder if we can support passing an optional `Callable` to `SpanBasedF1Measure` (to override `self. label_encoding`) so that user could provide and maintain the span generator by themselves.

https://github.com/allenai/allennlp/blob/7bf930f48ff8a72611db61301268b058e5cbaf48/allennlp/training/metrics/span_based_f1_measure.py#L38

https://github.com/allenai/allennlp/blob/7bf930f48ff8a72611db61301268b058e5cbaf48/allennlp/training/metrics/span_based_f1_measure.py#L138-L149

matt-gardner: @DeNeutoy, do you have an opinion on how to handle this?  You've done a lot more with the span logic than I have.  It looks like most of these tags-to-span methods just throw errors on invalid sequences, though the BIO one doesn't.  What should we do?

huntzhan:
@DeNeutoy your idea is *much* more concise and elegant! Thanks!

PR #1714, opened 2018-09-05
Moving allennlp.nn.decoding to allennlp.state_machines

matt-gardner:
There is no logic changed in this PR, just a bunch of renaming of things.  I moved what was in `allennlp.nn.decoding` to `allennlp.state_machines`, and gave some more internal structure to that module.  I added some documentation to the `__init__.py` files in the internal structure while I was at it.  I also moved the `State` and `TransitionFunction` objects out of the `wikitables` model module and into `allennlp.state_machines`.  This means that there are some modules under `allennlp.state_machines.transition_functions`, but I think they fit better there than inside `allennlp.modules`.


allennlp/state_machines/trainers/expected_risk_minimization.py
```diff
@@ -100,17 +98,17 @@ def _get_finished_states(self,
 
     # TODO(pradeep): Move this method to nn.decoding.util
```
pdasigi: Update the TODO. :)

PR #1715, opened 2018-09-06
Fix index_with bug in basic iterator

matt-peters:
Fixes a bug in `BasicIterator` that in some cases raises an exception when starting training.


matt-peters:
During training we never index the instances directly, only through the iterator with `index_with`. This fixes a edge case bug that was hidden because our test fixture indexed the instances directly.

allennlp/data/iterators/data_iterator.py
```diff
@@ -243,6 +243,8 @@ def _ensure_batch_is_sufficiently_small(self, batch_instances: Iterable[Instance
         padding_length = -1
         list_batch_instances = list(batch_instances)
         for instance in list_batch_instances:
+            if self.vocab is not None:
```
joelgrus: can you just add a comment as to why we're doing indexing here when we also do it elsewhere?

PR #1723, opened 2018-09-06
multiprocess dataset reader

joelgrus:
I think this is a pretty reasonable approach. At a high level, here's what happens:

* there's a new `MultiprocessDatasetReader` that wraps an existing DatasetReader and handles all the multiprocessing functionality
* in the calypso code there was not a clear division between the dataset reader and the iterator; in particular, it allowed you to multiprocess over multiple epochs, so I added an otherwise out-of-place `epochs_per_read` parameter to this dataset reader to allow the same functionality. 
* similarly, in the calypso code the multiprocessing stuff handled batching as well. here I left things separated, so the multiprocess reader uses multiple processes to generate instances, and batching is still done in a single process by the iterator
* it's possible then that *batching* is a major bottleneck, in which case we would need to figure out a way for the iterator to also use multiple processes, my sense is that parallelizing that would be quite a bit more complicated than this


joelgrus:
after chatting with @matt-peters I'm going to try to add similar functionality to the data iterator as well

schmmd:
We're not even sure why this PR exists.  It's already in the library ;-)

PR #1725, opened 2018-09-06
Add softmax to elmo

tokestermw:
Hi all,

This is a revival of https://github.com/allenai/allennlp/pull/1111.

This adds the softmax layer for the pre-trained ELMo. Having the log probabilities are useful for downstream tasks (as a bidirectional language model).

The biggest change is to the `batch_to_ids` method where if you include the vocab, it returns `word_ids`.

```
# return word_ids when you specify a word vocab
char_ids, word_ids = batch_to_ids(sentences, elmo_softmax.vocab)
```

I've added a class `_ElmoSoftmax` which is depedent on the results of `_ElmoBilm` and `batch_to_ids`.

To use inside the `elmo` command, add the `softmax-weight-file` and `softmax-vocab-file` and it will save the log probabilities instead of the ELMo embeddings:

```
allennlp elmo sample_sents.txt out2.hdf5 --top \
        --softmax-weight-file https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/2x4096_512_2048cnn_2xhighway/elmo_2x4096_512_2048cnn_2xhighway_softmax_weights.hdf5 \
        --softmax-vocab-file https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/2x4096_512_2048cnn_2xhighway/vocab-2016-09-10.txt

python -c "import h5py; f = h5py.File('out2.hdf5'); print(f['0'][:], f['0'].shape)"
```

Unsurprisingly, the model is considerably slower when we calculate the softmax layer.

```
# ELMo only vectors
100it [00:22,  4.39it/s]
# ELMo with softmax layer
100it [02:59,  1.80s/it]
```

To test (I've added a `elmo_softmax_weights.hdf5` for testing):

```
pytest allennlp/tests/modules/elmo_test.py
```

Here is a gist on how to use the new component: https://gist.github.com/tokestermw/6b3549bc5caa1be1d724a2a09659284c

Note, I did not add the softmax component to the `ELMo` module as it has a different interface than the `ELMo` command. Also, I wasn't sure where to include the `word_ids` without morphing the code too much.

Hope this is helpful!


schmmd:
Thanks @tokestermw!  We'll figure out how to move forward with this tomorrow.

schmmd:
@tokestermw sorry this fell off our radar.  Our lead engineer is looking at it now!

allennlp/modules/elmo.py
```diff
@@ -670,3 +689,156 @@ def create_cached_cnn_embeddings(self, tokens: List[str]) -> None:
...
+
+        Returns
+        -------
+        A tuple of `log_probs` and `mask_without_bos_eos`.
+        """
+
+        activations = bilm_inputs['activations']
+        mask = bilm_inputs['mask']
+
+        activations = activations[-1]  # (batch, BOS + L + EOS, [fwd, bwd])
```
DeNeutoy: Please use full variable names and don't re-use variable names which have different types. e.g

```python
# shape: (batch_size, sentence_length + 2, 2 * encoder_dim)
final_activations = activations[-1]
```

tokestermw:
Thanks! I will update after my vacation :) (2 weeks)

allennlp/modules/elmo.py
```diff
@@ -214,7 +217,7 @@ def from_params(cls, params: Params) -> 'Elmo':
                    dropout=dropout)
 
 
-def batch_to_ids(batch: List[List[str]]) -> torch.Tensor:
+def batch_to_ids(batch: List[List[str]], vocab: Vocabulary = None) -> Tuple[torch.Tensor, torch.Tensor]:
```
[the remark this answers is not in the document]
tokestermw: I think we still need the word `vocab` here to return the correct word ids.

If we replace with `vocab = Vocabulary()` (like it's done currently), the word ids return `1` for all words.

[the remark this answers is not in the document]
DeNeutoy: Right, but my point is that you don't actually need to pass word ids around at all - the only place you make use of them is [here](https://github.com/allenai/allennlp/pull/1725/files#diff-941a4360ef08efef82e2b72210a7ab93R756), but all you are using them for is to get an empty tensor on the correct device. The approach of splitting up the softmax is nice, but you don't need to pass word ids to do that, unless I am mis-understanding something. Instead, you can just iterate over a fixed chunk size, computing the softmax in parts as you are doing currently.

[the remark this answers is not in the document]
tokestermw: We still need the `word_ids` here (see `targets_flat`): https://github.com/allenai/allennlp/pull/1725/files#diff-941a4360ef08efef82e2b72210a7ab93R791

So `_chunked_log_probs` does not return the probabilities of the whole vocab, but just the probabilities of the target `word_ids`.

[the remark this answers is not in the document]
DeNeutoy: Oh! Sorry that wasn't clear from the [documentation](https://github.com/allenai/allennlp/pull/1725/files#diff-941a4360ef08efef82e2b72210a7ab93R771) of the class. Could you remove this functionality? I think this is something which is easier to do as a post-processing step and it would considerably simplify the code. It's also not something I anticipate a lot of people wanting to do.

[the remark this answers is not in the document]
tokestermw: Got it, so should I keep the changes made in the PR in `commands/elmo.py`? I don't think anyone would want the full softmax written to a file.

* [ ] remove Vocab dependency and output the full softmax
* [ ] refine documentation of https://github.com/allenai/allennlp/pull/1725/files#diff-941a4360ef08efef82e2b72210a7ab93R771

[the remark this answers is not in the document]
tokestermw: Hi, so I've mostly addressed your comments. Removed vocabulary dependency from the `elmo` command and also `batch_to_id`. I still have the `softmax_vocab_file` argument (now optional) in `_ElmoSoftmax` if people wanted to do some post-processing.

You can test here:

```
> cat sample_sents.txt
hello
how are you

> allennlp elmo sample_sents.txt out2.hdf5 --top \
        --softmax-weight-file https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/2x4096_512_2048cnn_2xhighway/elmo_2x4096_512_2048cnn_2xhighway_softmax_weights.hdf5
```

And then you can see the output:

```
> python -c "import h5py; f = h5py.File('out2.hdf5'); print(f['0'][:], f['0'].shape)"
```

Since the entire softmax is outputted, this file will be huge 😅. I think it would be better to output the probability of the original word which would require the vocab (so you can compute perplexities, play with language model outputs, probabilistic data), or I could simply remove it from the `elmo` command.

allennlp/modules/elmo.py
```diff
@@ -225,7 +228,7 @@ def from_params(cls, params: Params) -> 'Elmo':
                    scalar_mix_parameters=scalar_mix_parameters)
 
 
-def batch_to_ids(batch: List[List[str]]) -> torch.Tensor:
+def batch_to_ids(batch: List[List[str]]) -> Tuple[torch.Tensor, torch.Tensor]:
```
matt-peters: After removing the vocab, all of the changes in this function should be reverted.

allennlp/modules/elmo.py
```diff
@@ -681,3 +685,148 @@ def create_cached_cnn_embeddings(self, tokens: List[str]) -> None:
...
+        """
+
+        activations = bilm_inputs['activations']
+        mask = bilm_inputs['mask']
+
+        final_activations = activations[-1]  # (batch, BOS + L + EOS, [fwd, bwd])
+
+        # get rid of predicting, or conditioning on EOS for the forward
+        # activation, and BOS for the reverse activation.
+        mask_without_bos_eos = mask[:, 2:]
```
matt-peters: Slicing like this won't work at all, you need to take into account the mask and padding.  Use this function to remove the sentence boundaries from both the mask and the `final_activations`: https://github.com/allenai/allennlp/blob/master/allennlp/nn/util.py#L1193

allennlp/modules/elmo.py
```diff
@@ -681,3 +685,148 @@ def create_cached_cnn_embeddings(self, tokens: List[str]) -> None:
...
+        fwd_log_probs = self._chunked_log_probs(
+            final_activations[:, :-2, :self.hidden_size])
+        bwd_log_probs = self._chunked_log_probs(
+            final_activations[:, 2:, self.hidden_size:])
+
+        # [B, (fwd, bwd), L]
+        log_probs = torch.stack((fwd_log_probs, bwd_log_probs), 1)
+        log_probs *= mask_without_bos_eos[:, None, :, None].float()
+
+        # aggregate, otherwise return both the backward and forward embeddings
```
matt-peters: Remove these aggregation functions and `aggregation_function` from the `forward` arguments and leave that as a post-processing step.  Just return a dict: `{'forward_log_probs':  `fwd_log_probs`, 'backward_log_probs':  `bwd_log_probs`, `mask`: `mask_without_bos_eos`}`.

This simplifies the code here (and the calculation of both the `mean` and `max` below is incorrect anyway).

allennlp/modules/elmo.py
```diff
@@ -681,3 +685,148 @@ def create_cached_cnn_embeddings(self, tokens: List[str]) -> None:
...
+        mask_without_bos_eos = mask[:, 2:]
+
+        # NOTE: can this be replaced by allennlp.util.masked_softmax ?
+        fwd_log_probs = self._chunked_log_probs(
+            final_activations[:, :-2, :self.hidden_size])
+        bwd_log_probs = self._chunked_log_probs(
+            final_activations[:, 2:, self.hidden_size:])
+
+        # [B, (fwd, bwd), L]
+        log_probs = torch.stack((fwd_log_probs, bwd_log_probs), 1)
```
matt-peters: Don't stack here (just return both tensors), or zero out the masked locations.

allennlp/modules/elmo.py
```diff
@@ -681,3 +685,148 @@ def create_cached_cnn_embeddings(self, tokens: List[str]) -> None:
...
+        Returns
+        -------
+        ``torch.Tensor`` with all of the log probabilities.
+        """
+        all_logprobs = []
+        num_chunks = (activation.size(0) - 1) // self.chunk_size + 1
+        for activation_chunk in torch.chunk(activation, num_chunks, dim=0):
+            chunk_size, sequence_length = activation_chunk.size()[:2]
+
+            time_indexer = (torch.arange(0, chunk_size * sequence_length)
```
matt-peters: What is the purpose of these indexers?

Seems the only thing needed inside this for loop is
```all_log_probs.append(torch.nn.functional.log_softmax(self.fc_layer(activation_chunk), dim=-1))```

??

File: allennlp/modules/elmo.py [hunk 8]
```diff
@@ -681,3 +685,148 @@ def create_cached_cnn_embeddings(self, tokens: List[str]) -> None:
...
+                 softmax_vocab_file: str = None,
+                 chunk_size: int = 256):
+
+        super(_ElmoSoftmax, self).__init__()
+
+        self.softmax_weight_file = softmax_weight_file
+        self.softmax_vocab_file = softmax_vocab_file
+        self.chunk_size = chunk_size
+
+        if self.softmax_vocab_file:
```
matt-peters: What is the purpose of loading the vocabulary in this module? It isn't used anywhere.

DeNeutoy:
closing due to inactivity

PR #1729, opened 2018-09-07
Removing contents of requirements.txt file

schmmd:
This should help this tutorial stay in sync with allennlp-as-a-library-example.


schmmd:
The test failures had to do with internet connectivity.  I'm force merging since this is a README change...

matt-gardner:
Related to that, we should probably mark those tests as flaky, so they try again if there's a network blip...

PR #1731, opened 2018-09-07
create tensors on cpu, move them to gpu later

joelgrus:
easier than I expected. (it's possible there's somewhere else I need to change it, but I think I got everywhere)

in addition to the usual CI stuff, I did the following to test the changes (all on a GPU machine)

(1) ran `predict --cuda-device 0` on one of the website models
(2) ran `evaluate --cuda-device 0` on one of the website models
(3) modified the "sniff test" to load its models on cuda_device=0 and checked that it passed
(4) ran the "trainer test" which (when GPU is available) trains a model on the GPU

### DIFF
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -14,6 +14,24 @@ from allennlp.common.checks import ConfigurationError
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
+def move_to_device(obj, cuda_device: int):
+    """
+    Given a structure (possibly) containing Tensors on the CPU,
+    move all the Tensors to the specified GPU (or do nothing, if they should be on the CPU).
+    """
+    if cuda_device < 0:
+        return obj
+
+    if isinstance(obj, torch.Tensor):
+        return obj.cuda(cuda_device)
+    elif isinstance(obj, list):
+        return [move_to_device(item, cuda_device) for item in obj]
+    elif isinstance(obj, tuple):
+        return tuple([move_to_device(item, cuda_device) for item in obj])
+    elif isinstance(obj, dict):
+        return {key: move_to_device(value, cuda_device) for key, value in obj.items()}
+    else:
+        return obj
 
 def batch_tensor_dicts(tensor_dicts: List[Dict[str, torch.Tensor]],
                        remove_trailing_dimension: bool = False) -> Dict[str, torch.Tensor]:
--- a/allennlp/commands/evaluate.py
+++ b/allennlp/commands/evaluate.py
@@ -48,6 +48,7 @@ from allennlp.data.dataset_readers.dataset_reader import DatasetReader
 from allennlp.data.iterators import DataIterator
 from allennlp.models.archival import load_archive
 from allennlp.models.model import Model
+from allennlp.nn import util
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
@@ -96,11 +97,11 @@ def evaluate(model: Model,
 
         iterator = data_iterator(instances,
                                  num_epochs=1,
-                                 shuffle=False,
-                                 cuda_device=cuda_device)
+                                 shuffle=False)
         logger.info("Iterating over dataset")
         generator_tqdm = Tqdm.tqdm(iterator, total=data_iterator.get_num_batches(instances))
         for batch in generator_tqdm:
+            batch = util.move_to_device(batch, cuda_device)
             model(**batch)
             metrics = model.get_metrics()
             if (not _warned_tqdm_ignores_underscores and
--- a/allennlp/common/testing/model_test_case.py
+++ b/allennlp/common/testing/model_test_case.py
@@ -92,11 +92,11 @@ class ModelTestCase(AllenNlpTestCase):
         # the same result out.
         model_dataset = reader.read(params['validation_data_path'])
         iterator.index_with(model.vocab)
-        model_batch = next(iterator(model_dataset, shuffle=False, cuda_device=cuda_device))
+        model_batch = next(iterator(model_dataset, shuffle=False))
 
         loaded_dataset = reader.read(params['validation_data_path'])
         iterator2.index_with(loaded_model.vocab)
-        loaded_batch = next(iterator2(loaded_dataset, shuffle=False, cuda_device=cuda_device))
+        loaded_batch = next(iterator2(loaded_dataset, shuffle=False))
 
         # Check gradients are None for non-trainable parameters and check that
         # trainable parameters receive some gradient if they are trainable.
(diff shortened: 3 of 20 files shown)


allennlp/nn/util.py
```diff
@@ -14,6 +14,24 @@
...
+    """
+    if cuda_device < 0:
+        return obj
+
+    if isinstance(obj, torch.Tensor):
+        return obj.cuda(cuda_device)
+    elif isinstance(obj, dict):
+        return {key: move_to_device(value, cuda_device) for key, value in obj.items()}
+    elif isinstance(obj, list):
+        return [move_to_device(item, cuda_device) for item in obj]
```
matt-gardner: I don't know if fixing this is worth it or not, but these three lower blocks all create new objects.  If we're trying to minimize the amount of computation done on the CPU inside the training loop, we probably want to avoid creating these objects (though, I'm making this point without any profiling, so I could just be wrong about whether this is actually an issue).

I mean, imagine what happens for the input to the dependency parser.  There are some tensors, and they get moved immediately and quickly.  There are also some metadata fields that are just intended to be passed through to the demo.  All of those essentially get a _deepcopy_ here, just for fun.  And this is pretty typical for our models, because the way that we've written the demo code means that you need to pass through these metadata fields.

So, here's one more proposed test for this before merging: actually run training on the GPU for the dependency parser, or BiDAF, or something, before and after this PR, and see if running time is affected.  It'd also be nice to try to a third variant, where we do something like `if not has_tensor(obj): return obj`, just returning the object directly if there's no tensor.  You could go crazy with a fourth variant and try to just _modify_ the object with CUDA tensors instead of creating a new one, but I'd guess that case is not really worth the effort, as objects where that happens tend to be small (the biggest one I can think of is the semantic parsing code, where we have large lists of `ProductionRuleArrays`, which are complex structures that have tensors inside).

What do you think?  Am I just being crazy here?

joelgrus: my intuition is that `deepcopy` is a very fast operation (unless at some point you encounter an object that's intrinsically expensive to copy, which I don't think is the case here), but I can test this

joelgrus: this is on bidaf

create tensors on GPU:

```
start_acc: 0.3215, end_acc: 0.3514, span_acc: 0.2418, em: 0.2702, f1: 0.3893, loss: 5.2269 ||: 100%|##########| 2190/2190 [23:22<00:00,  1.56it/s]
start_acc: 0.5415, end_acc: 0.5814, span_acc: 0.4426, em: 0.5521, f1: 0.6738, loss: 3.1927 ||: 100%|##########| 265/265 [01:12<00:00,  3.66it/s]
start_acc: 0.5324, end_acc: 0.5789, span_acc: 0.4346, em: 0.4852, f1: 0.6342, loss: 3.2254 ||: 100%|##########| 2190/2190 [19:34<00:00,  1.86it/s]
start_acc: 0.5861, end_acc: 0.6245, span_acc: 0.4947, em: 0.6122, f1: 0.7212, loss: 2.7903 ||: 100%|##########| 265/265 [00:45<00:00,  5.80it/s]
start_acc: 0.5793, end_acc: 0.6252, span_acc: 0.4823, em: 0.5370, f1: 0.6832, loss: 2.8311 ||: 100%|##########| 2190/2190 [19:36<00:00,  1.86it/s]
start_acc: 0.6026, end_acc: 0.6440, span_acc: 0.5114, em: 0.6305, f1: 0.7376, loss: 2.6503 ||: 100%|##########| 265/265 [00:45<00:00,  5.80it/s]
```

create tensors on CPU, move to GPU

```
start_acc: 0.3212, end_acc: 0.3506, span_acc: 0.2404, em: 0.2689, f1: 0.3895, loss: 5.2259 ||: 100%|##########| 2190/2190 [24:08<00:00,  1.51it/s]
start_acc: 0.5368, end_acc: 0.5741, span_acc: 0.4410, em: 0.5530, f1: 0.6716, loss: 3.2478 ||: 100%|##########| 265/265 [01:19<00:00,  3.33it/s]
start_acc: 0.5336, end_acc: 0.5792, span_acc: 0.4356, em: 0.4866, f1: 0.6344, loss: 3.2282 ||: 100%|##########| 2190/2190 [20:19<00:00,  1.80it/s]
start_acc: 0.5898, end_acc: 0.6286, span_acc: 0.5013, em: 0.6130, f1: 0.7215, loss: 2.7526 ||: 100%|##########| 265/265 [00:51<00:00,  5.13it/s]
start_acc: 0.5808, end_acc: 0.6252, span_acc: 0.4822, em: 0.5375, f1: 0.6836, loss: 2.8326 ||: 100%|##########| 2190/2190 [20:20<00:00,  1.79it/s]
start_acc: 0.6047, end_acc: 0.6440, span_acc: 0.5150, em: 0.6348, f1: 0.7438, loss: 2.6429 ||: 100%|##########| 265/265 [00:53<00:00,  4.96it/s]
```

the time per training epoch increases from 19:30 to 20:20, but you'd expect it to increase a little bit. I haven't yet tried your third variant, my guess is that the extra `isinstance` checks make it more expensive than the deepcopy way, but we'll see.

joelgrus:
ok, I modified it to use your `has_tensor` suggestion

matt-peters:
Is it worthwhile to consider using `pin_memory` and async GPU transfers?  This PR also significantly decreases the footprint of CUDA related code and separates the tensor creation from CUDA transfer so maybe it's a few line change to add?

https://pytorch.org/docs/stable/notes/cuda.html#use-pinned-memory-buffers

matt-gardner:
I didn't know about that @matt-peters, and yes, it seems like a good thing to do.

joelgrus:
ok, let me see how much trouble this is (and how much difference it makes)

joelgrus:
so I tried this, and at least on the bidaf training, it didn't make things any faster (and in fact made them slightly slower, although not enough to care about)

```
start_acc: 0.3181, end_acc: 0.3473, span_acc: 0.2380, em: 0.2662, f1: 0.3856, loss: 5.2498 ||: 100%|##########| 2190/2190 [24:08<00:00,  1.51it/s]
start_acc: 0.5367, end_acc: 0.5746, span_acc: 0.4402, em: 0.5448, f1: 0.6652, loss: 3.2295 ||: 100%|##########| 265/265 [01:19<00:00,  3.33it/s]
start_acc: 0.5336, end_acc: 0.5795, span_acc: 0.4363, em: 0.4863, f1: 0.6344, loss: 3.2269 ||: 100%|##########| 2190/2190 [20:10<00:00,  1.81it/s]
start_acc: 0.5916, end_acc: 0.6302, span_acc: 0.5022, em: 0.6173, f1: 0.7248, loss: 2.7723 ||: 100%|##########| 265/265 [00:51<00:00,  5.15it/s]
```

these are the changes I made, maybe I didn't do it right?

https://github.com/joelgrus/allennlp/compare/tensors-on-cpu...joelgrus:pin-memory?expand=1

matt-peters:
Thanks for trying this!   It looks right to me.

PR #1741, opened 2018-09-10
Rename and organize tutorials

schmmd:
This PR provides a bit more organization for our tutorials.  Not only does it organize them for better maintenance (it's not easy to tell which files depend on which tutorials), but it also makes it clearer how many tutorials we have and makes it more obvious how our tutorials are grouped to the user.


schmmd:
I'm not very tied to the names I chose, but it's a bit tedious to change them.

PR #1742, opened 2018-09-10
Remove bucket iterator shuffle warning.

HarshTrivedi:
From discussion here: #1578. Let me know if I should remove the code instead of commenting it.

### DIFF
--- a/allennlp/data/iterators/bucket_iterator.py
+++ b/allennlp/data/iterators/bucket_iterator.py
@@ -127,9 +127,9 @@ class BucketIterator(DataIterator):
                 penultimate_batch = batches.pop()
             if shuffle:
                 random.shuffle(batches)
-            else:
-                logger.warning("shuffle parameter is set to False,"
-                               " while bucket iterators by definition change the order of your data.")
+            # else:
+            #     logger.warning("shuffle parameter is set to False,"
+            #                    " while bucket iterators by definition change the order of your data.")
             if move_to_front:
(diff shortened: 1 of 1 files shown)


allennlp/data/iterators/bucket_iterator.py
```diff
@@ -127,9 +127,9 @@ def _create_batches(self, instances: Iterable[Instance], shuffle: bool) -> Itera
                 penultimate_batch = batches.pop()
             if shuffle:
                 random.shuffle(batches)
-            else:
-                logger.warning("shuffle parameter is set to False,"
-                               " while bucket iterators by definition change the order of your data.")
+            # else:
```
matt-gardner: Yes, this should be a comment in the code inside the `if shuffle` block, like `# NOTE: if shuffle is false, the data will still be in a different order because of the bucket sorting`.  We generally don't allow commented-out code to be checked in, because it's not obvious _why_ the code is commented out.

HarshTrivedi: Sure. So I believe that you mean to **remove** the `else` block code and replace the `if` block by the following:
```python
            if shuffle:
                # NOTE: if shuffle is false, the data will still be in a different order
                # because of the bucket sorting
                random.shuffle(batches)
```
right?

matt-gardner: Yes, exactly.

PR #1743, opened 2018-09-10
Graph parser for semantic dependencies

DeNeutoy:


DeNeutoy:
Currently this only has unlabeled F1/Precision/Recall as a metric - I'm going to add the labelled one in a separate PR, because it requires adding a new metric and this model already works well without it.

allennlp/models/graph_parser.py
```diff
@@ -0,0 +1,341 @@
...
+        By default, a 1 layer feedforward network with an elu activation is used.
+    pos_tag_embedding : ``Embedding``, optional.
+        Used to embed the ``pos_tags`` ``SequenceLabelField`` we get as input to the model.
+    dropout : ``float``, optional, (default = 0.0)
+        The variational dropout applied to the output of the encoder and MLP layers.
+    input_dropout : ``float``, optional, (default = 0.0)
+        The dropout applied to the embedded text input.
+    edge_prediction_threshold : ``int``, optional (default = 0.5)
+        The probability at which to consider a scored edge to be 'present'
+        in the decoded graph. Must be between 0 and 1.
```
matt-gardner: You might note that this is only used when making predictions, and not during training.

allennlp/models/graph_parser.py
```diff
@@ -0,0 +1,341 @@
...
+        minus_inf = -1e8
+        minus_mask = (1 - float_mask) * minus_inf
+        attended_arcs = attended_arcs + minus_mask.unsqueeze(2) + minus_mask.unsqueeze(1)
+
+        arc_probs, arc_tag_probs = self._greedy_decode(attended_arcs,
+                                                       arc_tag_logits,
+                                                       mask)
+        # Make the arc tags not have negative values anywhere
+        # (by default, no edge is indicated with -1).
+        arc_indices = (arc_tags != -1).float()
```
matt-gardner: This block needs to go inside `if arg_tags is not None`, right?

allennlp/models/graph_parser.py
```diff
@@ -0,0 +1,341 @@
...
+        tag_mask = float_mask.unsqueeze(1) * float_mask.unsqueeze(2)
+        one_minus_arc_probs = 1 - arc_probs
+        self._unlabelled_f1(torch.stack([one_minus_arc_probs, arc_probs], -1), arc_indices, tag_mask)
```
matt-gardner: I don't really get what the `stack` is doing here.  Oh...  This is because the F1 metric expects a softmax distribution, and you have a sigmoid?

allennlp/models/graph_parser.py
```diff
@@ -0,0 +1,341 @@
...
+                        attended_arcs: torch.Tensor,
+                        arc_tag_logits: torch.Tensor,
+                        arc_tags: torch.Tensor,
+                        mask: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        Computes the arc and tag loss for an adjacency matrix.
+
+        Parameters
+        ----------
+        attended_arcs : ``torch.Tensor``, required.
```
matt-gardner: Is `arc_scores` an appropriate name here?

allennlp/models/graph_parser.py
```diff
@@ -0,0 +1,341 @@
...
+            A tensor of shape (batch_size, sequence_length, sequence_length, sequence_length)
+            representing the distribution over edge tags for a given edge.
+        """
+        # Mask the diagonal, because we don't self edges.
+        inf_diagonal_mask = torch.diag(attended_arcs.new(mask.size(1)).fill_(-numpy.inf))
+        attended_arcs = attended_arcs + inf_diagonal_mask
+        # shape (batch_size, sequence_length, sequence_length, num_tags)
+        arc_tag_logits = arc_tag_logits + inf_diagonal_mask.unsqueeze(0).unsqueeze(-1)
+        # Mask padded tokens, because we only want to consider actual word -> word edges.
+        if mask is not None:
```
matt-gardner: You've already called `mask.size(1)` above, so you'll have crashed `if mask is None`.

allennlp/models/graph_parser.py
```diff
@@ -0,0 +1,341 @@
...
+
+    @staticmethod
+    def _greedy_decode(attended_arcs: torch.Tensor,
+                       arc_tag_logits: torch.Tensor,
+                       mask: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        Decodes the head and head tag predictions by decoding the unlabeled arcs
+        independently for each word and then again, predicting the head tags of
+        these greedily chosen arcs indpendently. Note that this method of decoding
+        is not guaranteed to produce trees (i.e. there maybe be multiple roots,
```
matt-gardner: This second sentence probably isn't necessary, as the point is that this is a graph parser.

allennlp/data/fields/adjacency_field.py
```diff
@@ -0,0 +1,131 @@
...
+    def get_padding_lengths(self) -> Dict[str, int]:
+        return {'num_tokens': self.sequence_field.sequence_length()}
+
+    @overrides
+    def as_tensor(self,
+                  padding_lengths: Dict[str, int],
+                  cuda_device: int = -1) -> torch.Tensor:
+        desired_num_tokens = padding_lengths['num_tokens']
+        tensor = torch.ones(desired_num_tokens, desired_num_tokens) * self._padding_value
+        labels = self._indexed_labels or [1 for _  in range(len(self.indices))]
```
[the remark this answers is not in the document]
DeNeutoy: You can already specify that when you create it - it's pretty un-intuitive if the padding value depends on the value of the labels.

allennlp/models/graph_parser.py
DeNeutoy: Thanks, fixed. The mask is actually required anyway.

allennlp/models/graph_parser.py
DeNeutoy: Yep, fixed thanks

PR #1744, opened 2018-09-10
Set up `SlantedTriangular` for first batch

matt-peters:
`scheduler.step_batch` is called after updating parameters with each batch -- this PR includes a small change to ensure the learning rate for the first batch is set appropriately with `SlantedTriangular`.

@sebastianruder


allennlp/tests/training/learning_rate_schedulers_test.py
```diff
@@ -321,16 +321,22 @@ def test_schedules(self):
...
 
+            batch_num_total = 0
             for epoch in range(params["num_epochs"]):
                 for _ in range(params["num_steps_per_epoch"]):
-                    scheduler.step_batch()
-                    lrs.append([param_group["lr"] for param_group in optimizer.param_groups])
+                    batch_num_total += 1
+                    # allennlp trainer calls step_batch after updating parameters
+                    # so collect lr at time of parameter update
+                    lrs.append([param_group["lr"] * float(param_group['params'][0].requires_grad)
```
matt-peters: This also checks that the `requires_grad` flag is set as expected by setting the effective learning rate to 0.0 for parameters that are frozen.

allennlp/tests/training/learning_rate_schedulers_test.py
```diff
@@ -257,8 +257,8 @@ def test_from_params(self):
         assert len(optim.param_groups) == 3
         # The default parameter group in the Optimizer is empty
         assert not optim.param_groups[-1]["params"]
-        assert optim.param_groups[-2]["lr"] == 1.0
-        assert optim.param_groups[-3]["lr"] == 0.5
+        assert optim.param_groups[-2]["lr"] == 1.0 / 32
```
matt-gardner: I had to search to figure out where this 32 came from - might be nicer to either use `sched.ratio` here, or to do `assert sched.ratio == 32` above, so you just have to look up a few lines to know that 32 is probably sched.ratio.  Or you could just leave a comment saying what this is.

PR #1745, opened 2018-09-11
Add logging of learning rates to tensorboard

matt-peters:
Add parameter specific learning rate logging to tensorboard.  It's optional, default is False.


matt-gardner:
Does this replace what was done in #1641?

matt-peters:
@DeNeutoy - agree, it would be nice to log the actual step sizes.

@matt-gardner - this will replace #1641 and an in a more general way (takes into consideration `requires_grad`) and in a more user friendly way (logs using the name of the parameter instead of an opaque parameter group index)

PR #1755, opened 2018-09-11
output metrics.json every epoch

### LINKED ISSUE #1645: Update metrics.json whenever best.th is updated
Currently your job will output a `metrics.json` at the end of training with statistics from the last epoch.  @bhavanadalvi suggested instead we could output `metrics.json` each time `best.th` is updated.  That way it would have the metrics from the best output rather than the last, and incremental output could be viewed in the Beaker UI if running a Beaker job.

joelgrus:
fixes #1645 

technically 1645 was "update metrics.json whenever validation metrics increases", but that ended up being kind of a mess, so instead I just made it output a metrics.json every epoch (presumably each metrics dict is only like a KB or less).


matt-gardner:
Hmm, I think one of the original motivations for this was to have `metrics.json` reflect the values of the best epoch.  Then you can remove the `best_validation_`, and you don't have stale metrics at all anymore.  The stale metrics can be pretty confusing and misleading.

joelgrus:
with this change you know what the best epoch is, and then you can go look at that file to see the non-stale ones. I could change it so that `metrics.json` necessarily contains the training / validation metrics from the best epoch, but that's not obvious either. (I could then name it `best_epoch_metrics.json`, but then that makes this a breaking change)

one other complication (that I'm not sure how to fix) is that if you do `evaluate_on_test`, you're evaluating with the _final_ state of the model, not with the _best validation epoch_ state of the model. 🤔

allennlp/training/trainer.py
```diff
@@ -769,12 +768,37 @@ def train(self) -> Dict[str, Any]:
...
+            metrics["training_start_epoch"] = epoch_counter,
+            metrics["training_epochs"] = epochs_trained
+            metrics["epoch"] = epoch
+
+            for key, value in train_metrics.items():
+                metrics["training_" + key] = value
+            for key, value in val_metrics.items():
+                metrics["validation_" + key] = value
+
+            if is_best_so_far:
```
DeNeutoy: this shouldn't be behind an if statement, we should always have these otherwise the keys in the dictionary are not constant which is very annoying for anyone trying to use them programmatically.

matt-gardner:
Oh, I missed that this was a separate file per epoch.  Ok, that seems like it solves most of the issues (except the one for beaker, but they're not going to be able to consume any improvements on our part for a while, anyway, sounds like).

I'm pretty sure we already fixed the `evaluate_on_test` issue: https://github.com/allenai/allennlp/blob/609babecd7e74853c0b1ce671307394739d5691c/allennlp/commands/train.py#L332-L345

allennlp/training/trainer.py
```diff
@@ -758,8 +759,6 @@ def train(self) -> Dict[str, Any]:
 
                     # Check validation metric to see if it's the best so far
                     is_best_so_far = self._is_best_so_far(this_epoch_val_metric, validation_metric_per_epoch)
-                    if is_best_so_far:
-                        best_epoch_val_metrics = val_metrics.copy()
```
DeNeutoy: This is still necessary, we don't want to only have the best metrics present if _this epoch_ is the best

allennlp/training/trainer.py
```diff
@@ -769,12 +768,37 @@ def train(self) -> Dict[str, Any]:
...
+
+            if is_best_so_far:
+                metrics['best_epoch'] = epoch
+                for key, value in val_metrics.items():
+                    metrics["best_validation_" + key] = value
+
+            if self._serialization_dir:
+                dump_metrics(os.path.join(self._serialization_dir, f'metrics_epoch_{epoch}.json'), metrics)
+
+            for index, param_group in enumerate(self._optimizer.param_groups):
```
DeNeutoy: Is this from a merge or something? It seems unrelated

joelgrus: oops, yeah I had a bad merge, let me fix this

allennlp/training/trainer.py
joelgrus: `metrics` is the same object every epoch, so the keys should be the same. (if you don't hit this `if` you keep the "best" values from the previous epoch, and the first epoch you'll for sure hit this if)

(I thought about that there's a weird edge case where if you have different metrics in different epochs funny things could happen, but I decided not to worry about that)

mateuszpieniak:
Hi, I have a related question about this merge. In version 2.4.0 `metrics.json` is only outputted once in the `train_loop.finish(metrics)` which is fine if you ask me.

https://github.com/allenai/allennlp/blob/d9b19b69ad82b30cc5115ebfe269c7d214ef6130/allennlp/commands/train.py#L462-L483

But when I interrupt training because it seems like the training doesn't seem to improve anymore, I have no `metrics.json` file and in turn no test dataset evaluation. It's of course due to `raise` in the exception. Should it be that way or should I open an issue?

PR #1759, opened 2018-09-12
Better multi-word predicates in Open IE predictors

gabrielStanovsky:
Consolidate multiword predicate (e.g., "decided to run"), in which case we don't need to return the embedded predicate ("run").


allennlp/predictors/open_information_extraction.py
```diff
@@ -60,6 +60,72 @@ def make_oie_string(tokens: List[Token], tags: List[str]) -> str:
 
     return " ".join(frame)
 
+def get_predicate_indices(tags: List[str]) -> List[int]:
+    """
+    Return the word indices of a predicate in BIO tags.
+    """
+    return [ind for (ind, tag)
```
joelgrus: nit: you don't have to use parens here

another nit: you could probably fit this on one line (and if you can't, the `in` is a weird place to split it)

allennlp/predictors/open_information_extraction.py
```diff
@@ -60,6 +60,72 @@ def make_oie_string(tokens: List[Token], tags: List[str]) -> str:
...
+                     for pred_id in get_predicate_indices(tags)])
+
+def check_predicates_subsumed(tags1: List[str], tags2: List[str]) -> bool:
+    """
+    Tests whether the predicate in BIO tags1 are subsumed in
+    those of tags2.
+    """
+    # Get predicate word indices from both predictions
+    pred_ind1, pred_ind2 = map(get_predicate_indices,
+                               [tags1, tags2])
```
joelgrus: this is an unusual use of `map` and makes it less obvious what you're doing. just do

```
pred_ind1 = get_predicate_indices(tags1)
pred_ind2 = get_predicate_indices(tags2)
```

it's the same number of lines and much clearer

allennlp/predictors/open_information_extraction.py
```diff
@@ -60,6 +60,72 @@ def make_oie_string(tokens: List[Token], tags: List[str]) -> str:
...
+    return set(pred_ind1) < set(pred_ind2)
+
+def merge_predictions(tags1: List[str], tags2: List[str]) -> List[str]:
+    """
+    Merge two predictions into one. Assumes the predicate in tags1 are contained in
+    the predicate of tags2.
+    """
+    # Allow tags1 to add elements to tags2
+    return [tag2 if (tag2 != 'O')\
+            else tag1
```
joelgrus: put this `else` on the previous line, it will make the code clearer

joelgrus: nit: the parens aren't really necessary here either

allennlp/predictors/open_information_extraction.py
```diff
@@ -60,6 +60,72 @@ def make_oie_string(tokens: List[Token], tags: List[str]) -> str:
...
+            for (tag1, tag2) in zip(tags1, tags2)]
+
+def consolidate_predictions(outputs: List[List[str]], sent_tokens: List[Token]) -> Dict[str, List[str]]:
+    """
+    Identify that certain predicates are part of a multiword predicate
+    (e.g., "decided to run") in which case, we don't need to return
+    the embedded predicate ("run").
+    """
+    pred_dict: Dict[str, List[str]] = {}
+    merged_outputs = list(map(join_mwp, outputs))
```
joelgrus: nit: I am not a fan of `list(map`, just use a list comprehension instead

allennlp/predictors/open_information_extraction.py
```diff
@@ -60,6 +60,72 @@ def make_oie_string(tokens: List[Token], tags: List[str]) -> str:
...
+            in enumerate(tags) if 'V' in tag]
+
+def get_predicate_text(sent_tokens: List[Token], tags: List[str]) -> str:
+    """
+    Get the predicate in this prediction.
+    """
+    return " ".join([sent_tokens[pred_id].text
+                     for pred_id in get_predicate_indices(tags)])
+
+def check_predicates_subsumed(tags1: List[str], tags2: List[str]) -> bool:
```
joelgrus: I don't like the `check` in the name, it makes it read awkward when you use it later. I'd call it something like `predicates_are_subsumed`

allennlp/predictors/open_information_extraction.py
```diff
@@ -60,6 +60,72 @@ def make_oie_string(tokens: List[Token], tags: List[str]) -> str:
...
+    for pred1_text, tags1 in zip(predicate_texts, merged_outputs):
+        # A flag indicating whether to add tags1 to predictions
+        add_to_prediction = True
+        if pred1_text in pred_dict:
+            # We already added this predicate
+            continue
+
+        # Else - check if this predicate if subsumed by another predicate
+        for pred2_text, tags2 in zip(predicate_texts, merged_outputs):
+            if (tags1 != tags2) and check_predicates_subsumed(tags1, tags2):
```
joelgrus: see, if it were called `predicates_are_subsumed`, this line would read much clearer

also nit: you don't need parens here

allennlp/predictors/open_information_extraction.py
```diff
@@ -60,6 +60,72 @@ def make_oie_string(tokens: List[Token], tags: List[str]) -> str:
...
+    Merge two predictions into one. Assumes the predicate in tags1 are contained in
+    the predicate of tags2.
+    """
+    # Allow tags1 to add elements to tags2
+    return [tag2 if (tag2 != 'O')\
+            else tag1
+            for (tag1, tag2) in zip(tags1, tags2)]
+
+def consolidate_predictions(outputs: List[List[str]], sent_tokens: List[Token]) -> Dict[str, List[str]]:
+    """
```
joelgrus: is it possible to have (say) predicate_a < predicate_b < predicate_c ?
it seems like your logic wouldn't work in that case

gabrielStanovsky:
@joelgrus I addressed your style comments, and changed the predicate merging logic to be able to merge more than 2 predicates (also added a test for that).

PTAL, thanks!

allennlp/predictors/open_information_extraction.py
```diff
@@ -60,6 +60,116 @@ def make_oie_string(tokens: List[Token], tags: List[str]) -> str:
...
+    Tests whether the predicate in BIO tags1 are subsumed in
+    those of tags2.
+    """
+    # Get predicate word indices from both predictions
+    pred_ind1 = get_predicate_indices(tags1)
+    pred_ind2 = get_predicate_indices(tags2)
+
+    # Return if pred_ind1 is contained in pred_ind2
+    return any(set.intersection(set(pred_ind1), set(pred_ind2)))
```
joelgrus: this returns true if they intersect, is that ok?

joelgrus: I think it's ok, maybe just the docstring and comment are wrong?

allennlp/predictors/open_information_extraction.py
```diff
@@ -60,6 +60,116 @@ def make_oie_string(tokens: List[Token], tags: List[str]) -> str:
...
+
+    if prev_label == cur_label:
+        return f"I-{cur_label}"
+    else:
+        return f"B-{cur_label}"
+
+def merge_overlapping_predictions(tags1: List[str], tags2: List[str]) -> List[str]:
+    """
+    Merge two predictions into one. Assumes the predicate in tags1 are contained in
+    the predicate of tags2.
```
joelgrus: is "contained in" correct?

gabrielStanovsky: Same, leftover docstring. I fixed that.

allennlp/predictors/open_information_extraction.py
gabrielStanovsky: Yes, intersection is what I was going for.  The docstrings were wrong.

allennlp/tests/predictors/open_information_extraction_test.py
```diff
@@ -49,3 +49,47 @@ def test_prediction_with_no_verbs(self):
...
+
+        # Emulate predications - for "refused" and "consider" and "joining"
+        predictions = [['B-ARG0', 'B-V', 'B-ARG1', 'I-ARG1', 'I-ARG1', 'I-ARG1', 'I-ARG1', 'O'],\
+                       ['B-ARG0', 'B-BV', 'I-BV', 'B-V', 'B-ARG1', 'I-ARG1', 'I-ARG1', 'O'],\
+                       ['B-ARG0', 'B-BV', 'I-BV', 'I-BV', 'B-V', 'B-ARG1', 'I-ARG1', 'O']]
+
+        # Consolidate
+        pred_dict = consolidate_predictions(predictions, sent_tokens)
+
+        # Check that only "decided to join" is left
```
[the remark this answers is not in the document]
gabrielStanovsky: done.

PR #1760, opened 2018-09-12
multiprocess dataset reader and iterator

joelgrus:
the DatasetReader is a wrapper for any other DatasetReader

the iterator is its own iterator. (our iterator code is way too complicated to make a similar wrapper possible, and anyway it feels less important for the iterator to work that way)


allennlp/data/dataset_readers/multiprocess_dataset_reader.py
```diff
@@ -0,0 +1,123 @@
...
+        for _ in range(self.epochs_per_read):
+            random.shuffle(shards)
+            for shard in shards:
+                input_queue.put(shard)
+
+        # Then put a None per worker to signify no more files.
+        for _ in range(self.num_workers):
+            input_queue.put(None)
+
+        # TODO(joelgrus): where does this number come from?
```
matt-peters: We could allow this to be configurable, and may need to increase it depending on how many instances the iterator uses for each batch.  We want it to be large enough allow a significant buffer so the iterator doesn't have to wait for instances but not too large that it starts taking up significant memory.

allennlp/data/iterators/multiprocess_iterator.py
```diff
@@ -0,0 +1,152 @@
...
+                         batch_size: int,
+                         vocab: Vocabulary) -> None:
+    """
+    Pulls at most ``max_instances_in_memory`` from the input_queue,
+    groups them into batches of size ``batch_size``, converts them
+    to ``TensorDict`` s, and puts them on the ``output_queue``.
+    """
+    instances: List[Instance] = []
+
+    def make_batches() -> None:
```
matt-peters: Is there any reason we can't re-use an existing iterator here?  For many applications we'd want this to have more logic to do something like the bucket iterator, or optionally allow shuffling all of the instances read into memory before creating batches, etc.  Right now we will always read each file in the same order.

allennlp/data/iterators/multiprocess_iterator.py
```diff
@@ -0,0 +1,152 @@
...
+            if instances:
+                make_batches()
+            output_queue.put(None)
+            break
+        else:
+            instances.append(instance)
+            if len(instances) >= max_instances_in_memory:
+                make_batches()
+
+def _queuer(instances: Iterable[Instance],
```
matt-peters: Is there a way to remove the `_queuer` if we are also using the `MultiprocessDatasetReader ` by passing the reader queue directly to the `MultiprocessIterator`?  E.g. the `MultiprocessIterator` can iterate over either an iterable of instances or a queue filled with instances?

allennlp/data/iterators/multiprocess_iterator.py
```diff
@@ -0,0 +1,152 @@
...
+    def _create_batches(self, instances: Iterable[Instance], shuffle: bool) -> Iterable[Batch]:
+        raise RuntimeError("MultiprocessIterator doesn't use create_batches")
+
+    def __call__(self,
+                 instances: Iterable[Instance],
+                 num_epochs: int = None,
+                 shuffle: bool = True) -> Iterator[TensorDict]:
+
+        # If you run it forever, the multiprocesses won't shut down correctly.
+        # TODO(joelgrus) find a solution for this
```
matt-peters: I'm partial to `pkill`...

allennlp/data/iterators/multiprocess_iterator.py
```diff
@@ -0,0 +1,152 @@
...
+        if num_epochs is None:
+            raise ConfigurationError("Multiprocess Iterator must be run for a fixed number of epochs")
+
+        # TODO(joelgrus) are these the right sizes?
```
matt-peters: See comment above -- seems we could also make the input queue size approximately batch_size times larger then output queue size.

joelgrus:
ok, I addressed all of your comments except for the "read directly from the queue". it turns out that that's a tricky distributed systems problem (or at least I couldn't figure out how to make it not one).

the key issue is that you have some number of workers that generate instances and put them on the queue.

then you have some *other* number of workers that pull them off and generate tensor_dicts.

if there is an iterator in the middle, the iterator can recognize when all of the instance-generating workers are done and stop iterating, which tells the tensor-dict-generating workers to stop.

without the iterator in the middle, how do the tensor-dict-generating workers know when to stop?  I couldn't figure out a clean solution.

however, I made it so that the Iterator[Instance] that the MultiprocessIterator gets exposes the queue, so that if I can figure out a good solution for this it shouldn't be hard to implement it.

allennlp/data/iterators/multiprocess_iterator.py
```diff
@@ -0,0 +1,140 @@
...
+    groups them into batches of size ``batch_size``, converts them
+    to ``TensorDict`` s, and puts them on the ``output_queue``.
+    """
+    def instances() -> Iterator[Instance]:
+        instance = input_queue.get()
+        while instance is not None:
+            yield instance
+            instance = input_queue.get()
+
+    for tensor_dict in iterator(instances(), num_epochs=1, shuffle=False):
```
matt-peters: Shouldn't we be passing down the `shuffle` parameter here? I think it will ignore whatever `shuffle` parameter is passed to `__call__` without it.

PR #1762, opened 2018-09-12
Executor for WikiTables variable free language

pdasigi:
@matt-gardner This PR is finally ready to review. Summary of main changes:
1) Created a new package `allennlp.semparse.executors` with an abstract class that all executors can inherit from.
2) Added execution logic for the wikitables variable free language in `allennlp.semparse.executors.wikitables_variable_free_executor.py`, and a bunch of tests.
3) Moved the execution logic from the `NlvrWorld` to `allennlp.semparse.executors.nlvr_executor.py`.

I will have to make some changes to the newly added executor when we have a new knowledge graph that identifies column types. There is some simple logic in the executor currently that identifies whether a numerical value is a date or not, which will become unnecessary then.


allennlp/semparse/executors/wikitables_variable_free_executor.py
```diff
@@ -0,0 +1,647 @@
...
+    def _handle_constant(self, constant: str) -> Union[List[Dict[str, str]], float]:
+        if constant == "all_rows":
+            return self._table_data
+        try:
+            return float(constant)
+        except ValueError:
+            logger.error(f"Cannot handle constant: {constant}")
+            raise ExecutionError(f"Cannot handle constant: {constant}")
+
+    def _select(self, expression_list: NestedList) -> List[str]:
```
matt-gardner: I'd also probably recommend having the arguments here be more transparent: `def _select(self, row_expression: NestedList, column_name: str) -> List[str]`.  You could feed the arguments in inside `handle_expression` like `function(*expression[1:])`.  It would certainly help readability and understandability of all of these methods.  It might have some consequences that I'm not foreseeing that could make it infeasible, though...

Maybe you did it like this because mypy complained if you made the types more specific.  If that's the case, ignoring mypy on that one line in `_handle_expression` is way better than making all of these functions more cryptic.

allennlp/semparse/executors/wikitables_variable_free_executor.py
```diff
@@ -0,0 +1,647 @@
...
+        """
+        Select function takes a list of rows and a column (decoded from the `expression_list`) and
+        returns a list of cell values as strings.
+        """
+        row_list, column_name = self._get_row_list_and_column_name(expression_list)
+        if not row_list:
```
matt-gardner: If `row_list` is an empty list, you'll get this behavior for free by just having the return line below.

allennlp/semparse/executors/wikitables_variable_free_executor.py
```diff
@@ -0,0 +1,647 @@
...
+            # This is a constant (like "all_rows" or "2005")
+            return self._handle_constant(str(expression))
+        try:
+            function = getattr(self, f"_{function_name}")
+            return function(expression[1:])
+        except AttributeError:
+            logger.error("Function not found: %s", function_name)
+            raise ExecutionError(f"Function not found: {function_name}")
+
+    def _get_row_list_and_column_name(self, expression_list: NestedList) -> Tuple[List[Dict[str,
```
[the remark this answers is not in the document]
matt-gardner: I'm also not sure what the value is that you're getting here.  If you split the arguments to your functions, like I recommended, you won't need this to split out the column name, and you can just do `row_list = self._handle_expression(row_expression)`.  The only thing you'd be losing if you just did that is the column name checking, which I'm not sure is necessary.

allennlp/semparse/executors/wikitables_variable_free_executor.py
```diff
@@ -0,0 +1,647 @@
...
+        else:
+            expression = expression_list
+        if isinstance(expression, list):
+            # This is a function application.
+            function_name = expression[0]
+        else:
+            # This is a constant (like "all_rows" or "2005")
+            return self._handle_constant(str(expression))
+        try:
+            function = getattr(self, f"_{function_name}")
```
matt-gardner: I think we made these methods private previously because they were combined with the `World`.  I'd recommend just making the language predicates public, and removing the underscore here.  That's also a clear way to separate the helper functions from the language functions.

allennlp/semparse/executors/wikitables_variable_free_executor.py
```diff
@@ -0,0 +1,647 @@
...
+    def _argmax(self, expression_list: NestedList) -> List[Dict[str, str]]:
+        """
+        Takes a list of rows and a column (decoded from the `expression_list`) and returns a list
+        containing a single row (dict from columns to cells) that has the maximum numerical value in
+        the given column. We return a list instead of a single dict to be consistent with the return
+        type of `_select` and `_all_rows`.
+        """
+        if self._values_look_like_dates(expression_list):
+            date_row_pairs = self._get_date_row_pairs_to_filter(expression_list)
+            if not date_row_pairs:
```
matt-gardner: Looks like you can pull out the last four lines of this, and just have

```python
if self._values_look_like_dates(row_expression):
    row_pairs = ...
else:
    row_pairs = ...
if not row_pairs:
    return []
return [sorted(...)]
```

allennlp/semparse/executors/executor.py
```diff
@@ -0,0 +1,37 @@
+from typing import Union, List, Any
+
+from allennlp.semparse import util as semparse_util
+
+NestedList = List[Union[str, List]]  # pylint: disable=invalid-name
+
+class Executor:
```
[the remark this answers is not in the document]
pdasigi: Good points! Done.

allennlp/semparse/executors/nlvr_executor.py
```diff
@@ -0,0 +1,624 @@
+from typing import List, Set, Dict, Union, TypeVar, Callable
```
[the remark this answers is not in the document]
pdasigi: Yes, that is the case.

allennlp/semparse/executors/wikitables_variable_free_executor.py
```diff
@@ -0,0 +1,647 @@
...
+        # The months and years are equal and not -1
+        if self.day == -1 or other.day == -1:
+            return False
+        return self.day > other.day
+
+    def __ge__(self, other) -> bool:
+        if not isinstance(other, Date):
+            return False
+        return self > other or self == other
+
```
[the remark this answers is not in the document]
pdasigi: Removed.

allennlp/semparse/executors/wikitables_variable_free_executor.py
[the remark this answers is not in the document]
pdasigi: Done.

allennlp/semparse/executors/wikitables_variable_free_executor.py
pdasigi: Done.

allennlp/semparse/executors/wikitables_variable_free_executor.py
pdasigi: mypy complains if values of different types are assigned to the same variable in if and else blocks.

matt-gardner: Can you do something with defining `row_pairs: SomeType` above, with a more generic type than it would get below?  If not, ignoring mypy seems reasonable, as you really shouldn't be duplicating the logic here.

pdasigi: I tried a couple of different options, including defining a generic type before the `if` block, but mypy was unforgiving. I'll ignore it for now. But this seems like a common coding pattern. There should be a better solution.

allennlp/semparse/contexts/table_question_knowledge_graph.py
```diff
@@ -85,9 +85,11 @@ def __init__(self,
                  entities: Set[str],
                  neighbors: Dict[str, List[str]],
                  entity_text: Dict[str, str],
-                 question_tokens: List[Token]) -> None:
+                 question_tokens: List[Token],
+                 table_data: List[Dict[str, str]]) -> None:
```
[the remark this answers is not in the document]
pdasigi: Agreed.

allennlp/tests/semparse/worlds/wikitables_variable_free_world_test.py
```diff
@@ -204,3 +208,209 @@ def test_get_agenda(self):
         assert set(world.get_agenda()) == {'l -> fb:row.row.avg_attendance',
                                            '<r,<l,n>> -> min'
                                           }
+
+    def test_execute_works_with_select(self):
```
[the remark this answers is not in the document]
pdasigi: Both done.

allennlp/semparse/executors/wikitables_variable_free_executor.py
```diff
@@ -0,0 +1,647 @@
...
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+class Date:
+    def __init__(self, year: int, month: int, day: int) -> None:
+        self.year = year
+        self.month = month
+        self.day = day
+
+    def __eq__(self, other) -> bool:
```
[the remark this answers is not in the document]
matt-gardner: I think it'd be a lot more clear to just call this something else, like `is_compatible()`.  `==` has an expected semantics, and you're changing it.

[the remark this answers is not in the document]
pdasigi: By using the in-built functions, I was able to directly sort the values within argmax and argmin. And what else would you expect from `==` for the `Date` class?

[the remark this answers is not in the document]
matt-gardner: I'm just saying that `==` has an expected semantics in python, and it's a bit jarring to me to see it changed like this, so that, e.g., it's no longer transitive.  That's probably fine if this class is only used in this executor, just be sure you document it clearly.

allennlp/semparse/executors/wikitables_variable_free_executor.py
```diff
@@ -0,0 +1,647 @@
...
+        month_is_same = self.month == -1 or other.month == -1 or self.month == other.month
+        day_is_same = self.day == -1 or other.day == -1 or self.day == other.day
+        return year_is_same and month_is_same and day_is_same
+
+    def __gt__(self, other) -> bool:
+        # pylint: disable=too-many-return-statements
+        if not isinstance(other, Date):
+            return False
+        if self.year == -1 or other.year == -1:
+            return False
```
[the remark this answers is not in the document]
pdasigi: Hmm.. actually if the year is not specified in both cases, I guess it is safe to assume that it is the same. I'll fix that.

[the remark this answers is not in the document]
matt-gardner: I would expect `Date(-1, 2, 1) > Date(-1, 1, 1)` to return `True`.  This is saying "is February 1 later than January 1"?  The answer is yes.  I'm not sure if this actually matters for any of the actual questions in wikitables, so maybe it's not worth arguing over, but I could easily imagine a question like "which events took place after March?"  I'd at least leave a comment somewhere explaining the decision you make, and keep it in mind when looking at error cases in the data after we train a model with this.

allennlp/semparse/executors/wikitables_variable_free_executor.py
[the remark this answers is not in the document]
pdasigi: Done.

allennlp/semparse/executors/wikitables_variable_free_executor.py
```diff
@@ -0,0 +1,647 @@
...
+
+    def __ge__(self, other) -> bool:
+        if not isinstance(other, Date):
+            return False
+        return self > other or self == other
+
+
+
+class WikiTablesVariableFreeExecutor(Executor):
+    def __init__(self, table_data: List[Dict[str, str]]) -> None:
```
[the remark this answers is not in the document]
pdasigi: Done.

allennlp/semparse/executors/wikitables_variable_free_executor.py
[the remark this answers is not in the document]
pdasigi: Agreed. Removed this method.

pdasigi:
@matt-gardner This should be ready to go once the conflicts are resolved.

allennlp/semparse/executors/wikitables_variable_free_executor.py
```diff
@@ -0,0 +1,746 @@
...
+        year_is_same = self.year == -1 or other.year == -1 or self.year == other.year
+        month_is_same = self.month == -1 or other.month == -1 or self.month == other.month
+        day_is_same = self.day == -1 or other.day == -1 or self.day == other.day
+        return year_is_same and month_is_same and day_is_same
+
+    def __gt__(self, other) -> bool:
+        # pylint: disable=too-many-return-statements
+        # The logic below is tricky, and is based on some assumptions we make about date comparison.
+        # Year, month or day being -1 means that we do not know its value. In those cases, the
+        # we conside the comparison to be undefined, and return False if all the fields that are
```
matt-gardner: `s/conside/consider/`

allennlp/semparse/worlds/nlvr_box.py
```diff
@@ -0,0 +1,32 @@
+from typing import List
```
matt-gardner: Are these classes _only_ used in the NLVR executor?  If they are, I'd probably just put them in the executor module.

allennlp/tests/semparse/executors/wikitables_variable_free_executor_test.py
```diff
@@ -0,0 +1,383 @@
...
+        with self.assertRaises(ExecutionError):
+            self.executor.execute(logical_form)
+
+    def test_date_comparison_works(self):
+        assert Date(2013, 12, 31) > Date(2013, 12, 30)
+        assert Date(2013, 12, 31) == Date(2013, 12, -1)
+        assert Date(2013, -1, -1) >= Date(2013, 12, 31)
+        assert Date(2013, 12, 31) != 2013
+        # pylint: disable=singleton-comparison
+        assert (Date(2013, 12, -1) > Date(2013, 12, 31)) == False
```
matt-gardner: You can remove the pylint statement and these parentheses if you do `Date(...) > Date(...) is False`.

allennlp/tests/semparse/executors/wikitables_variable_free_executor_test.py
```diff
@@ -0,0 +1,383 @@
...
+        assert (Date(2013, 12, 31) > 2013) == False
+        assert (Date(2013, 12, 31) >= 2013) == False
+        assert Date(2013, 12, 31) != 2013
```
matt-gardner: You already have this one above.

allennlp/tests/semparse/executors/wikitables_variable_free_executor_test.py
```diff
@@ -0,0 +1,383 @@
...
+        assert (Date(2018, 1, 1) >= Date(-1, 2, 1)) == False
+        assert (Date(2018, 1, 1) < Date(-1, 2, 1)) == False
```
matt-gardner: I'm actually not sure about this one.  I can imagine some cases where we'd want this to be `True`.  But it's probably best to let failure cases in the actual data drive any changes here.

allennlp/semparse/executors/wikitables_variable_free_executor.py
pdasigi: Done.

allennlp/semparse/worlds/nlvr_box.py
pdasigi: They are also used in the `NlvrWorld`, and they used to be in that module earlier. The reason I moved them out is because there is now a circular dependency (`NlvrWorld` depends on `NlvrExecutor`, and `NlvrExecutor` depends on `Box` and `Object`).

allennlp/tests/semparse/executors/wikitables_variable_free_executor_test.py
pdasigi: Done, thanks!

allennlp/tests/semparse/executors/wikitables_variable_free_executor_test.py
pdasigi: Removed, thanks!

allennlp/tests/semparse/executors/wikitables_variable_free_executor_test.py
pdasigi: Agreed. Added a TODO.

allennlp/tests/semparse/executors/wikitables_variable_free_executor_test.py
pdasigi: Hmm.. looks like `>` and `is` have the same level of precedence, and since they're evaluated left to right, that wouldn't work. Keeping the original code.

Issue #1704, opened 2018-08-31
Maintain a list of models with performance, size and download links


DeNeutoy:
We have more available models than are in the demo with various different attributes. We should make it easy for people to discover these, either via the website, download links in the demo or in the github readme (or preferably a combination).

schmmd:
I opened a ticket to add download links in our demo: https://github.com/allenai/allennlp/issues/1704

schmmd:
I worry about becoming the central repository for all models derived from AllenNLP.  While I see the utility, what do we do when models are updated or beak due to a breaking change in AllenNLP?  Already I don't know how to find the models that have been created on AllenNLP.

Maybe we should have a repository people can contribute to (self serve)?

matt-gardner:
I think this is basically just the user gallery.  We should have a single list somewhere in a markdown document (either the README itself, or linked from the README) that has all of the known models, separated by whether they are in the repo or in some other repo, with links to the model archive file and the code.  We should have a similar list for dataset readers.

schmmd:
Well, we have a user gallery at the bottom of https://allenai.github.io/allennlp-docs/.  But it doesn't link out directly to models and it doesn't list performance numbers.

DeNeutoy:
No, I mean _our_ models which we have which aren't in the demo. E.g a lot of our models in the demo have elmo built into them, which makes them very slow, but we have all of the variants without Elmo, but people can't access them. Often people want to download a specific model and have it in a particular location, not just magically in a cache.

schmmd:
@DeNeutoy thanks--that makes a lot of sense ;-)

matt-gardner:
@DeNeutoy, right, and I think wherever that list lives is a natural place to also list all known outside implementations, too.  That way everything's in the same place.

schmmd:
@DeNeutoy do you know of models of our own that are somewhere other than `s3://allennlp/models/`?

I'm scoping this ticket to just organizing our own models for now.

DeNeutoy:
No, they all should be there. When you get around to it I can message you with the extra ones

DeNeutoy:
Right - I meant I will send you the models we have which aren't in the demo.

schmmd:
@DeNeutoy I'm going to start out just aggregating them in a markdown file (MODELS.md) and we can go from there.  I'm disinclined to add them to https://allennlp.org/models because I think we'll move most of the content from that page and into the demo (see https://github.com/allenai/allennlp-demo/issues/68).

[...]

PR #1768, opened 2018-09-13
add simplest possible pretrained model interface

DeNeutoy:
Adds an explicit, Keep It Simple Stupid functional interface to get our pretrained models. This can replace the duplication in the demo, too. Related to #1704

### DIFF
--- a/allennlp/pretrained.py
+++ b/allennlp/pretrained.py
@@ -18,56 +18,62 @@ class PretrainedModel:
         archive = load_archive(self.archive_file)
         return Predictor.from_archive(archive, self.predictor_name)
 
+# TODO(Mark): Figure out a way to make PretrainedModel generic on Predictor, so we can remove these type ignores.
 
 #### Models in the demo ####
 def srl_with_elmo_luheng_2018() -> predictors.SemanticRoleLabelerPredictor:
     model = PretrainedModel('https://s3-us-west-2.amazonaws.com/allennlp/models/srl-model-2018.05.25.tar.gz',
                             'semantic-role-labeling')
-    return model.predictor()
+    return model.predictor() # type: ignore
 
 def bidirectional_attention_flow_seo_2017() -> predictors.BidafPredictor:
     model = PretrainedModel('https://s3-us-west-2.amazonaws.com/allennlp/models/bidaf-model-2017.09.15-charpad.tar.gz',
                             'machine-comprehension')
-    return model.predictor()
+    return model.predictor() # type: ignore
 
 def open_information_extraction_stanovsky_2018() -> predictors.OpenIePredictor:
     model = PretrainedModel('https://s3-us-west-2.amazonaws.com/allennlp/models/openie-model.2018-08-20.tar.gz',
                             'open-information-extraction')
-    return model.predictor()
+    return model.predictor() # type: ignore
 
 def decomposable_attention_with_elmo_parikh_2017() -> predictors.DecomposableAttentionPredictor:
     model = PretrainedModel('https://s3-us-west-2.amazonaws.com/allennlp/models/decomposable-attention-elmo-2018.02.19.tar.gz',
                             'textual-entailment')
-    return model.predictor()
+    return model.predictor() # type: ignore
 
 def neural_coreference_resolution_lee_2017() -> predictors.CorefPredictor:
     model = PretrainedModel('https://s3-us-west-2.amazonaws.com/allennlp/models/coref-model-2018.02.05.tar.gz',
                             'coreference-resolution')
-    return model.predictor()
+    return model.predictor() # type: ignore
 
 def named_entity_recognition_with_elmo_peters_2018() -> predictors.SentenceTaggerPredictor:
     model = PretrainedModel('https://s3-us-west-2.amazonaws.com/allennlp/models/ner-model-2018.04.30.tar.gz',
                             'sentence-tagger')
-    return model.predictor()
+    return model.predictor() # type: ignore
 
 def fine_grained_named_entity_recognition_with_elmo_peters_2018() -> predictors.SentenceTaggerPredictor:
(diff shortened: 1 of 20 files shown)


joelgrus:
do we have any kind of commitment / SLA about keeping these models around forever?

joelgrus:
well, I was imagining the keys being like `'biaffine_parser_universal_dependencies_todzat_2017'`, let me think a bit more about why this bugs me

DeNeutoy:
I guess this is also less verbose too, and doesn't require you to know the following complex reasoning chain:

```python
from allennlp.models.archival import load_archive
from allennlp.predictors import Predictor
# Why should a new user know what these imports are?

archive = load_archive("Path you have to know to find on the models page of the website")

model_you_can_use = Predictor.from_archive(
	archive, 
	"Predictor name that you need to retrieve from DEFAULT_PREDICTORS,
	which in turn requires that you know the registered model name for the model you 
	are trying to load to do correctly")

model_you_can_use.predict(sentence="This process was kind of hard.")

```

versus:
```python
import allennlp
model_you_can_use = allennlp.pretrained.biaffine_parser_universal_dependencies_todzat_2017()
model_you_can_use.predict(sentence="This process was kind of hard.")
```
I think 2) you could figure out just from importing allennlp at a REPL - I don't think you can do 1) without reading the docs.

joelgrus:
yeah, I'm coming around

would `allennlp.models.pretrained` make more sense than `allennlp.pretrained`?

DeNeutoy:
It probably would, but:

1) That get's us in dependency hell because Predictors import Models and this imports Predictors
2) I wanted this to be in a really obvious place

allennlp/pretrained.py
```diff
@@ -0,0 +1,72 @@
...
+        return Predictor.from_archive(archive, self.predictor_name)
+
+
+#### Models in the demo ####
+def srl_with_elmo_luheng_2018() -> Predictor:
+    model = PretrainedModel('https://s3-us-west-2.amazonaws.com/allennlp/models/srl-model-2018.05.25.tar.gz',
+                            'semantic-role-labeling')
+    return model.predictor()
+
+def bidirectional_attention_flow_seo_2017() -> Predictor:
```
matt-gardner: These should be more specific `Predictor` types, so that you can get better code completion on the `predict()` methods.

matt-gardner:
I'd have recommended ignoring mypy on the return statement instead of reverting the specific predictor types, as that better accomplishes your goal of good autocomplete.  But you're the one who's suggesting this, so do what you think is best.

DeNeutoy:
Yeah ok that's better - when I added your obviously good suggestion I just got into some reallllly gross typing problems, because doing this meant that `PretrainedModel` needed to be generic on `Predictor` subclasses, which meant that `Predictor.from_archive` had to have the following type signature:

```
def from_archive(cls: Type[T], archive: Archive, predictor_name: str = None) -> T:
```

but then the problem is that `T` in this case is a `Predictor`, not a specific subclass of it, so we still get a type clash, aka we have just lifted the problem into `Predictor`, rather than what calls it.

matt-gardner:
Yeah, ignoring the type was how I was originally envisioning making this work; I should have said something with my comment.  Getting a concrete type through a `from_params` call like that is a nightmare.  I got it to work in scala, actually, but I wouldn't recommend it.

DeNeutoy:
Ah ok - actually do you have that somewhere? I'd be interested to see how to do it, even if it's not the solution here.

matt-gardner:
I had to go back and look at my code, and I discovered that I didn't actually accomplish what you wanted to do in this PR, and in fact I don't think that's possible in scala.  But I did get concrete types out of `from_params` _inside_ the library code.  It's a bit of a mess to go through the whole thing, but here's a (relatively) brief description.

In my [PRA code](https://github.com/matt-gardner/pra), I wanted to deal with `Instances` that represented either node pairs in a graph, or individual nodes in the graph.  These are necessarily different types, and need different feature extractors, etc.  I wanted to share all of the experiment driver code, but be able to instantiate the right concrete instances, feature extractor objects, and so on, depending on what type of data I got at runtime.

The key problem was when I was [reading the data](https://github.com/matt-gardner/pra/blob/d54f83d54bcb0d29e196869bcc001deb16ad56d3/src/main/scala/edu/cmu/ml/rtw/pra/data/Split.scala#L138-L151), which either had `NodePairInstances` or `NodeInstances`.  This is where the concrete type of everything else was determined, and it was done using a `Params` setup that was the initial version of what we ended up using in our `from_params` stuff here.  So, from a string in a config file, I instantiated either a `NodePairSplit` or a `NodeSplit`, and to make the types work out in scala (which you can't just do `# type: ignore` on, it won't compile if you don't get this right), I had to use an _existential type_ (the `Split[_ <: Instance]` return type there).

What the existential type means is that the caller of `Split.create()` doesn't have any concrete type information, other than that the type is has is something that's a subclass of `Instance`.  So you have to be careful with how you use the object that's returned.  If you try to access certain properties of the object in [this method](https://github.com/matt-gardner/pra/blob/d54f83d54bcb0d29e196869bcc001deb16ad56d3/src/main/scala/edu/cmu/ml/rtw/pra/experiments/Driver.scala#L74), where that object is first returned, scala will complain about existential types (I don't remember what the specific errors were, but I remember there were some).  But if you pass that object into other methods that are themselves parameterized as `[T <: Instance]`, you can case match on `T` without problems, and everything works.

So, getting a concrete type returned out into the wild from something that was instantiated with `from_params`, like you're trying to do here, would not have been possible.  But with some crazy use of existential types, I _was_ able to get concrete types _inside_ of my code.  This is basically identical to what we do with our `from_params` in AllenNLP (which isn't surprising, because that came out of what I first did in PRA), it's just that we don't have to worry about this existential type stuff because python doesn't care - we can make types more specific when calling methods, and mypy doesn't complain, or if it does we can just judiciously ignore it.

PR #1769, opened 2018-09-14
use managers for queues

joelgrus:
this was a real nightmare to debug, but it seems like it fixes the problem. (as best as I can tell it was some kind of race condition with queues getting deleted before they were done being used. the `Manager` seems to avoid that.)

verified by building the Dockerfile.pip image and running the relevant tests inside it

### DIFF
--- a/allennlp/data/dataset_readers/multiprocess_dataset_reader.py
+++ b/allennlp/data/dataset_readers/multiprocess_dataset_reader.py
@@ -3,7 +3,7 @@ import glob
 import logging
 import random
 
-from torch.multiprocessing import Process, Queue, log_to_stderr
+from torch.multiprocessing import Manager, Process, Queue, log_to_stderr
 
 from allennlp.data.dataset_readers.dataset_reader import DatasetReader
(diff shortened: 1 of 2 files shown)


schmmd:
Yikes!  Sorry it ended up being a nightmare to debug.  I suppose it was easier than if a user reported the issue and we couldn't reproduce it however...

What tipped you off to using a `Manager`?  I know nothing about Python multithreaded code.

joelgrus:
it's less that anything tipped me off and more that I was just trying everything

the documentation for manager sort of hints at this but not really

https://docs.python.org/3.6/library/multiprocessing.html#managers

matt-peters:
Did you benchmark this? I've had performance issues with overhead using Manager, but that was a long time ago and in a different use case.  In your case you are only using the Manager to manage the queues so I have no idea whether or not that has any performance considerations.

joelgrus:
I didn't benchmark it, but it literally would crash (on some architectures) without `Manager`, so I'm not sure a benchmark would be actionable.

Issue #1735, opened 2018-09-08
Tutorial on how to view embeddings in HDF5 data


radiantone:
Hi,
  I am new to AllenNLP. I used the comman line tool to create hdf5 embeddings file from sentences. Following the tutorial. But it stops there and I'm not sure how to view the actual text results. I see the HDF5 file contains HDF5 datasets. I want to see the text word embeddings.

Thank you!

schmmd:
I should add this to our ELMo tutorial: https://github.com/allenai/allennlp/blob/master/tutorials/how_to/elmo.md

@radiantone would you find it helpful if the ELMo command supported other output formats?  Which would you prefer?

radiantone:
Yeah, I am trying to use it programmatically. You'll have to forgive my newness to elmo. I see in the tutorial this segment:

embeddings = elmo(character_ids)

# embeddings['elmo_representations'] is length two list of tensors.
# Each element contains one layer of ELMo representations with shape
# (2, 3, 1024).
#   2    - the batch size
#   3    - the sequence length of the batch
#   1024 - the length of each ELMo vector

What I am wanting to do is view the actual text embeddings found vs enumerate the numerical vectors.
Sorry if this request sounds a bit ignorant.

schmmd:
Hmm... it looks like something was removed from that HowTo actually.  I think the following would be easiest:

```
from allennlp.commands.elmo import ElmoEmbedder
elmo = ElmoEmbedder()
vectors = elmo.embed_sentence(["I", "ate", "an", "apple", "for", "breakfast"])
```

Now you have `vectors[0]`, `vectors[1]`, and `vectors[2]` that represent the ELMo vectors.

radiantone:
Yeah. I can get the vectors but I'm curious how to display the human readable words in them.

matt-gardner:
They are vectors that are representing the words that you passed in.  If you embedded the words, you already have access to them.

radiantone:
Ok. I guess I'm wanting a concrete example of what one can do with the vectors. The tutorial doesn't show anything after obtaining vectors.

So I get these vectors. What next? What can I do with them. What information do they contain that I dont already have with the raw sentences? I know they are numbers and represent words but that's not enough to understand what to use them for.

This is probably obvious to career data scientists but for the layman programmer this is not clear in the tutorial.

matt-gardner:
[...]

PR #1771, opened 2018-09-14
Add an example of using ELMo interactively

schmmd:
In response to https://github.com/allenai/allennlp/issues/1735

### DIFF
--- a/tutorials/how_to/elmo.md
+++ b/tutorials/how_to/elmo.md
@@ -52,31 +52,9 @@ JSON-serialized string with a mapping from sentences to line indices to the
 
 For more details, see `allennlp elmo -h`. 
 
-## Using ELMo interactively
+## Using ELMo as a PyTorch `Module` to train a new model
 
-You can use ELMo interactively (or programatically) with iPython.  The `allennlp.commands.elmo.ElmoEmbedder` class provides the easiest way to process one or many sentences with ELMo, but it is meant for use as a standalone command and not within a larger model.  For example, if you would like to learn a weighted average of the ELMo vectors then you need to use `allennlp.modules.elmo.Elmo` instead.
-
-The `ElmoEmbedder` command returns three vectors, each corresponding to a layer in the ELMo LSTM output.  The first layer captures mostly syntactic information and the last layer captures more of the semantics.
-
-```
-$ ipython
-> from allennlp.commands.elmo import ElmoEmbedder
-> elmo = ElmoEmbedder()
-> tokens = ["I", "ate", "an", "apple", "for", "breakfast"]
-> vectors = elmo.embed_sentence(tokens)
-
-> assert(len(vectors) == 3) # one for each layer in the ELMo output
-> assert(len(vectors[0]) == len(tokens)) # the vector elements correspond with the input tokens
-
-> import scipy
-> vectors2 = elmo.embed_sentence(["I", "ate", "a", "carrot", "for", "breakfast"])
-> scipy.spatial.distance.cosine(vectors[2][3], vectors2[2][3]) # cosine distance between "ate" and "carrot" in the last layer
-0.18020617961883545
-```
-
-## Using ELMo as a PyTorch `Module`
-
-If you need to include ELMo at multiple layers in a task model or you have other advanced use cases, you will need to create ELMo vectors programatically using the `allennlp.modules.elmo.Elmo` class [(API doc)](https://github.com/allenai/allennlp/blob/master/allennlp/modules/elmo.py#L27).  This class provides a mechanism to compute the weighted ELMo representations (Equation (1) in the paper) as a PyTorch tensor.  The weighted average can be learned as part of a larger model and typically works best for using ELMo to improving performance on a particular task.
+To train a model using ELMo, use the allennlp.modules.elmo.Elmo class ([API doc](https://github.com/allenai/allennlp/blob/master/allennlp/modules/elmo.py#L27)). This class provides a mechanism to compute the weighted ELMo representations (Equation (1) in the paper) as a PyTorch tensor.  The weighted average can be learned as part of a larger model and typically works best for using ELMo to improving performance on a particular task.
 
 This is a `torch.nn.Module` subclass that computes any number of ELMo
 representations and introduces trainable scalar weights for each.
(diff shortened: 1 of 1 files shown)


schmmd:
`ElmoEmbedder` makes batching transparent and has more convenience methods.  Perhaps they can be merged into the same class, but I think `ElmoEmbedder` is an easier experience for someone who wants to play around with Elmo interactively.

Do you think my first sentence a clear enough description? @matt-peters

matt-peters:
We could add the batching and convenience methods to Elmo too.  The key difference is that one returns numpy arrays and one is a torch module that returns tensors you can use for training another model.  This is super confusing so if we include code snippets from both we should have a clear intro section that explains the distinction with recommendations about which to use in what scenario.

schmmd:
@matt-peters I'm not sure if it's crystal clear--but at least it's better...

tutorials/how_to/elmo.md
```diff
@@ -53,10 +52,31 @@ JSON-serialized string with a mapping from sentences to line indices to the
 
 For more details, see `allennlp elmo -h`. 
 
-## Using ELMo programmatically
+## Using ELMo interactively
+
+You can use ELMo interactively (or programatically) with iPython.  The `allennlp.commands.elmo.ElmoEmbedder` class provides the easiest way to process one or many sentences with ELMo, but it is meant for use as a standalone command and not within a larger model.  For example, if you would like to learn a weighted average of the ELMo vectors then you need to use `allennlp.modules.elmo.Elmo` instead.
```
matt-peters: Explicitly state here that the layers returned are `numpy` arrays.

tutorials/how_to/elmo.md
```diff
@@ -53,10 +52,31 @@ JSON-serialized string with a mapping from sentences to line indices to the
...
+
+The `ElmoEmbedder` command returns three vectors, each corresponding to a layer in the ELMo LSTM output.  The first layer captures mostly syntactic information and the last layer captures more of the semantics.
```
matt-peters: Replace this with "The `ElmoEmbedder` class returns three vectors for each word, each vector corresponding to a layer in the ELMo LSTM output.  The first layer corresponds to the context insensitive token representation, followed by the two LSTM layers.  See [the ELMo paper](http://www.aclweb.org/anthology/N18-1202) or [follow up work at EMNLP 2018](https://arxiv.org/abs/1808.08949) for a description of what types of information is captured in each layer."

tutorials/how_to/elmo.md
```diff
@@ -53,10 +52,31 @@ JSON-serialized string with a mapping from sentences to line indices to the
...
+> scipy.spatial.distance.cosine(vectors[2][3], vectors2[2][3]) # cosine distance between "ate" and "carrot" in the last layer
+0.18020617961883545
+```
+
+## Using ELMo as a PyTorch `Module`
 
-If you need to include ELMo at multiple layers in a task model or you have other advanced use cases, you will need to create ELMo vectors programatically.
-This is easily done with the `Elmo` class [(API doc)](https://github.com/allenai/allennlp/blob/master/allennlp/modules/elmo.py#L27), which provides a mechanism to compute the weighted ELMo representations (Equation (1) in the paper).
+If you need to include ELMo at multiple layers in a task model or you have other advanced use cases, you will need to create ELMo vectors programatically using the `allennlp.modules.elmo.Elmo` class [(API doc)](https://github.com/allenai/allennlp/blob/master/allennlp/modules/elmo.py#L27).  This class provides a mechanism to compute the weighted ELMo representations (Equation (1) in the paper) as a PyTorch tensor.  The weighted average can be learned as part of a larger model and typically works best for using ELMo to improving performance on a particular task.
```
matt-peters: How about "To train a model using ELMo, use the `allennlp.modules.elmo.Elmo` class [(API doc)](https://github.com/allenai/allennlp/blob/master/allennlp/modules/elmo.py#L27).  This class provides..."

tutorials/how_to/elmo.md
```diff
@@ -53,10 +52,31 @@ JSON-serialized string with a mapping from sentences to line indices to the
...
+> assert(len(vectors) == 3) # one for each layer in the ELMo output
+> assert(len(vectors[0]) == len(tokens)) # the vector elements correspond with the input tokens
+
+> import scipy
+> vectors2 = elmo.embed_sentence(["I", "ate", "a", "carrot", "for", "breakfast"])
+> scipy.spatial.distance.cosine(vectors[2][3], vectors2[2][3]) # cosine distance between "ate" and "carrot" in the last layer
+0.18020617961883545
+```
+
+## Using ELMo as a PyTorch `Module`
```
matt-peters: ```## Using ELMo as a PyTorch `Module` to train a new model```

tutorials/how_to/elmo.md
```diff
@@ -53,10 +52,31 @@ JSON-serialized string with a mapping from sentences to line indices to the
 
 For more details, see `allennlp elmo -h`. 
 
-## Using ELMo programmatically
+## Using ELMo interactively
```
matt-peters: Can you swap the order of these subsections?

Put "Using ELMo as a PyTorch `Module`" first, then `Using ELMo interactively`, then `Writing contextual representations to disk` last.

tutorials/how_to/elmo.md
```diff
@@ -53,10 +52,31 @@ JSON-serialized string with a mapping from sentences to line indices to the
 
 For more details, see `allennlp elmo -h`.
```
matt-peters: Let's also add somewhere in the "Writing contextual representations to disk" section that this is a thin wrapper around the `allennlp.commands.elmo.ElmoEmbedder` class that adds a command line interface.

tutorials/how_to/elmo.md
schmmd: I'm don't think that's best since people who are new to the space will want a second that comes later (and will probably get tripped up by what they don't know if it's buried)--but I made the change.

tutorials/how_to/elmo.md
schmmd: @matt-peters for a future PR...

schmmd:
@matt-peters changes made--although I'll add a "writing to disk" section in a future PR.

PR #1777, opened 2018-09-17
Extracting entities from questions for variable free WTQ language

MurtyShikhar:
The main class to look at is QuestionEntityExtractor. Still working on dealing with conjuctions. This implementation has been tested on 10 examples from random-split-1-train.examples and random-split-1-dev.examples and compared against the preprocessing by Chen Liang's code, and it roughly matches.


allennlp/semparse/contexts/table_question_knowledge_graph.py
```diff
@@ -61,6 +61,82 @@
...
+        with open(stop_words_file, 'r') as f:
+            stop_words_list = json.load(f)
+        self.stop_words = set(stop_words_list)
+
+
+    def _string_in_table(self, candidate_str) -> bool:
+        for entity, text in self.table_entity_text.items():
+            # entity normalized text is followed after fb:cell.
+            if entity.startswith('fb:cell.')
+            normalized_text = entity[8:]
```
pdasigi: Missing colon in the previous line and indentation in this one.

allennlp/semparse/contexts/table_question_knowledge_graph.py
```diff
@@ -61,6 +61,82 @@
...
+            if len(normalized_token) == 0: continue
+            if self._string_in_table(normalized_token):
+                entity_dat.append(dict(value=normalized_token,
+                                       token_start=i,
+                                       token_end=i+1))
+
+        return self._expand_entities(question, entity_dat)
+
+
 class TableQuestionKnowledgeGraph(KnowledgeGraph):
```
pdasigi: Didn't you mean to make a separate class? We'll still need this one for the original parser.

allennlp/semparse/contexts/table_question_knowledge_graph.py
```diff
@@ -88,6 +164,10 @@ def __init__(self,
                  question_tokens: List[Token]) -> None:
         super().__init__(entities, neighbors, entity_text)
         self.question_tokens = question_tokens
+        entity_extractor = QuestionEntityExtractor( '/u/murtyjay/allennlp_fork/stop_words.json',  entity_text)
```
pdasigi: How big is the stop words list? If it is not huge, may be you can define the list as a constant in the module instead of reading from a local file?

allennlp/semparse/contexts/table_question_knowledge_graph.py
```diff
@@ -61,6 +61,82 @@
...
+            stop_words_list = json.load(f)
+        self.stop_words = set(stop_words_list)
+
+
+    def _string_in_table(self, candidate_str) -> bool:
+        for entity, text in self.table_entity_text.items():
+            # entity normalized text is followed after fb:cell.
+            if entity.startswith('fb:cell.')
+            normalized_text = entity[8:]
+            if candidate_str in normalized_text:
```
pdasigi: `candidate_str` comes from `TableQuestionKnowledgeGraph._normalize_string` that does not replace spaces with underscores, correct? But `normalized_text` here has underscores. You'll have to make sure they are comparable.

allennlp/semparse/contexts/table_question_knowledge_graph.py
```diff
@@ -61,6 +61,82 @@
...
+        for entity, text in self.table_entity_text.items():
+            # entity normalized text is followed after fb:cell.
+            if entity.startswith('fb:cell.')
+            normalized_text = entity[8:]
+            if candidate_str in normalized_text: 
+                return True
+        return False
+
+
+    def _expand_entities(self, question, entity_dat ):
```
pdasigi: Can you please add docstrings describing what the methods are doing?

allennlp/semparse/contexts/table_question_knowledge_graph.py
```diff
@@ -61,6 +61,82 @@
         **MONTH_NUMBERS,
         }
 
+
+class ColumnTyper():
```
pdasigi: You don't need parens at the end of the class name.

allennlp/semparse/contexts/table_question_knowledge_graph.py
```diff
@@ -61,6 +61,87 @@
...
+
+
+    def _process_conjunction(self, entity_dat):
+        raise NotImplementedError
+
+    def _expand_entities(self, question, entity_dat ):
+        new_ents = []
+        for ent in entity_dat:
+            # to ensure the same strings are not used over and over
+            if new_ents and ent['token_end'] <= new_ents[-1]['token_end']: continue
```
matt-gardner: What happens if I just expand a string from later in the sentence first?

allennlp/semparse/contexts/table_question_knowledge_graph.py
```diff
@@ -61,6 +61,87 @@
...
+            curr_en = ent['token_end']
+            curr_tok = ent['value']
+
+            while curr_en < len(question): 
+                next_tok = question[curr_en]
+                # use the same normalize function used by TableQuestionKnowledgeGraph for consistency
+                next_tok_normalized = TableQuestionKnowledgeGraph._normalize_string(next_tok) 
+                curr_tok_normalized = TableQuestionKnowledgeGraph._normalize_string(next_tok)
+
+                for s, s1, s2 in itertools.product( [' ', '-', '_', ''], [curr_tok, curr_tok_normalized] , [next_tok, next_tok_normalized]  ):
```
matt-gardner: What is this line doing?

allennlp/semparse/contexts/table_question_knowledge_graph.py
```diff
@@ -61,6 +61,87 @@
...
+                    if len(s1) == 0 or len(s2) == 0: continue
+                    candidate = s.join([s1,s2])
+                    if self._string_in_table(candidate):
+                        curr_en += 1
+                        curr_tok = candidate
+                        break
+                else: break
```
matt-gardner: It's not at all clear unless you're really familiar with python that this is breaking out of the outer while loop if the inner for loop doesn't break.  I'd use a flag here instead, just for readability.

allennlp/semparse/contexts/table_question_knowledge_graph.py
```diff
@@ -61,6 +61,87 @@
...
+
+            new_ents.append(dict(token_start=curr_st, token_end=curr_en,value=curr_tok))
```
matt-gardner: Same thought about dictionary creation notation.

allennlp/semparse/contexts/table_question_knowledge_graph.py
```diff
@@ -88,6 +169,10 @@ def __init__(self,
                  question_tokens: List[Token]) -> None:
         super().__init__(entities, neighbors, entity_text)
         self.question_tokens = question_tokens
+        entity_extractor = QuestionEntityExtractor( '/u/murtyjay/allennlp_fork/stop_words.json',  entity_text)
```
matt-gardner: Instead of adding this as a field on `TableQuestionKnowledgeGraph`, make this its own class, and test the class directly.  You said you tried this out on some examples and saw that it matched Chen's code - I'd like to see those cases added as test cases for this class in this PR.

allennlp/semparse/contexts/table_question_knowledge_graph.py
MurtyShikhar: there's 158 words in there. not too large, maybe I could define it as a constant somewhere

allennlp/semparse/contexts/table_question_knowledge_graph.py
matt-gardner: If we've already normalized the strings, do we need all of these possible combinations?  And is this going to join `curr_tok` with `next_tok_normalized`?  Is that necessary?

MurtyShikhar: I now see that since the cell values are normalized, it probably makes sense to only match it with [curr_tok_normalized, next_tok_normalized] because the others are likely redundant. Do you also agree? I'll make that change

allennlp/semparse/contexts/table_question_knowledge_graph.py
MurtyShikhar: I'm thinking of renaming QuestionEntityExtractor to "TableQuestionsContext" and add column typing as a public method inside that. What do you think?

allennlp/semparse/contexts/table_question_knowledge_graph.py
matt-gardner: Yes, and the normalization code converts hyphens and spaces to underscores, so you don't need to worry about that substitution.

allennlp/semparse/contexts/table_question_knowledge_graph.py
matt-gardner: Yeah, `TableQuestionContext` seems fine.  Make sure the docstring says what's different about it from `TableQuestionKnowledgeGraph`.

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,130 @@
...
+        self.cell_values = cell_values
+        self.column_types = { column_idx_to_name[column_idx] : max(column_type_statistics[column_idx]) 
+                              for column_idx in column_idx_to_name }
+            
+       
+
+    @classmethod
+    def parse_file(cls, filename: str, question: List[Token], max_tokens_for_num_cell : int) -> 'TableQuestionKnowledgeGraph':
+        f = open(filename, 'r')
+        reader = csv.reader(f,  delimiter='\t', quoting=csv.QUOTE_NONE)
```
pdasigi: Missing `import csv`

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,130 @@
...
+        # obtain column information
+        lines = [line for line in reader]
+        column_idx_to_name = {}
```
pdasigi: `column_index_to_name` would be better. Similarly `index` instead of `idx`.

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,130 @@
...
+
+        header = lines[0] # the first line is the header
+        idx = 1
+        while lines[idx] == '-1':
```
pdasigi: I don't follow the logic here. When would `lines[idx] == '-1'` be `True`?

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,130 @@
...
+        	# column names start with fb:row.row. 
+        	curr_line = lines[idx]
+        	column_name_sempre = curr_line[2]
+        	column_idx = curr_line[1]
+        	column_name = column_name_sempre[11:]
+        	column_idx_to_name[column_idx] = column_name
+            idx += 1
```
pdasigi: Why is this indentation different from those of the previous lines? Is it something to do with tabs vs spaces?

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,130 @@
...
+
+        
+        column_node_type_info = [{'string' : 0, 'number' : 0, 'date' : 0} 
+                                 for col in column_idx_to_name]
+        cell_values = set()
+        while idx < len(lines):
+        	curr_line = lines[idx]
```
pdasigi: Indentation is off in this block too.

allennlp/semparse/contexts/table_question_context.py
pdasigi: Oh.. does `csv.reader` add line numbers? If so, please add a comment explaining that.

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,130 @@
...
+        column_idx_to_name = {}
+
+        header = lines[0] # the first line is the header
+        idx = 1
+        while lines[idx] == '-1':
+        	# column names start with fb:row.row. 
+        	curr_line = lines[idx]
+        	column_name_sempre = curr_line[2]
+        	column_idx = curr_line[1]
+        	column_name = column_name_sempre[11:]
```
pdasigi: Removing `fb:row.row.`? If so, `column_name_sempre.replace(''fb:row.row., '')` would be easier to understand.

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,130 @@
...
+                 column_idx_to_name : Dict[int,str]) -> None:
+        self.cell_values = cell_values
+        self.column_types = { column_idx_to_name[column_idx] : max(column_type_statistics[column_idx]) 
+                              for column_idx in column_idx_to_name }
+            
+       
+
+    @classmethod
+    def parse_file(cls, filename: str, question: List[Token], max_tokens_for_num_cell : int) -> 'TableQuestionKnowledgeGraph':
+        f = open(filename, 'r')
```
pdasigi: Using a `with open(filename, 'r') as file_pointer:` block instead of this line would automatically close the file when `file_pointer` goes out of scope.

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,130 @@
...
+            idx += 1
+
+        
+        column_node_type_info = [{'string' : 0, 'number' : 0, 'date' : 0} 
+                                 for col in column_idx_to_name]
+        cell_values = set()
+        while idx < len(lines):
+        	curr_line = lines[idx]
+            column_idx = curr_line[1]
+            node_info = zip(header, curr_line)
```
pdasigi: What is this variable expected to contain? Note that `zip` returns an iterator in Python 3. Looks like you're using `node_info` as if it is a `dict` below. I don't think that would work, if I'm understanding this code correctly.

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,130 @@
...
+                next_tok = question[curr_en]
+                next_tok_normalized = self._normalize_string(next_tok) 
+                candidate = "%s_%s" %(curr_tok, next_tok_normalized) 
+                if self._string_in_table(candidate):
+                    curr_en += 1
+                    curr_tok = candidate
+                else: 
+                    break
+
+            new_ents.append({'token_start' : curr_st, 'token_end' : curr_en 'value' : curr_tok})
```
pdasigi: Missing comma between `curr_en` and `value`

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,130 @@
...
+    def _string_in_table(self, candidate : str) -> bool:
+        for cell_value in self.cell_values:
+            if candidate in cell_value: return True
+        return False
+
+
+    def _process_conjunction(self, entity_data):
+        raise NotImplementedError
+
+    def _expand_entities(self, question, entity_data):
```
pdasigi: I think I mostly understand the logic here, but writing down a couple of lines explaining it would be very helpful!

allennlp/semparse/contexts/table_question_context.py
MurtyShikhar: Done.

allennlp/semparse/contexts/table_question_context.py
MurtyShikhar: Done, this seems like an older file, i didn't commit/push the one with these syntax issues fixed. :/ So, these have been taken care of.

allennlp/semparse/contexts/table_question_context.py
MurtyShikhar: done.

allennlp/semparse/contexts/table_question_context.py
MurtyShikhar: intended to be dict(zip(header, curr_line))

MurtyShikhar: and now fixed

allennlp/semparse/contexts/table_question_context.py
MurtyShikhar: yeah, I fixed this

allennlp/semparse/contexts/table_question_context.py
MurtyShikhar: In the current version this is lines[idx][0] == -1..

allennlp/semparse/contexts/table_question_context.py
MurtyShikhar: okay, changed all idx to index

allennlp/semparse/contexts/table_question_context.py
MurtyShikhar: thanks, fixed!

allennlp/semparse/contexts/table_question_context.py
MurtyShikhar: Definitely, i'll add comments

matt-gardner:
I'd strongly recommend adding some tests and making sure they pass on our continuous integration server before Pradeep or I look at this again. That's our typical process, and it makes better use of everyone's time, so we don't have to point out syntax errors that you've already fixed locally, and so that you find a bunch of the bugs in the logic yourself first.

allennlp/semparse/contexts/table_question_knowledge_graph.py
```diff
@@ -1,4 +1,4 @@
-import re
+import re, itertools,nltk, json
```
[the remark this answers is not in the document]
MurtyShikhar: fixed

allennlp/tests/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,74 @@
+# pylint: disable=no-self-use,invalid-name,protected-access,too-many-public-methods
```
matt-gardner: This file needs to have `test` in its name, or pytest won't pick it up as containing tests.  That's why CI is saying that your test coverage on the additions here is basically zero.  So I'd rename this to `table_question_context_test.py`.

allennlp/tests/semparse/contexts/test_tables/TEST-10.table
```diff
@@ -0,0 +1,56 @@
+row	col	id	content	tokens	lemmaTokens	posTags	nerTags	nerValues	number	date	num2	list	listId
```
matt-gardner: We typically put files like these in `allennlp/tests/fixtures/data`.  You can see some of them for wikitables here: https://github.com/allenai/allennlp/tree/master/allennlp/tests/fixtures/data/wikitables.  I'd probably add a `mapo_processed` directory there, or something (not really sure how to name it; if you can think of a better name, great).

matt-gardner:
Ok, now you've got some test failures that you can iterate on.  You can run this locally with `pytest -v -k TableQuestionContext', which will be a whole lot faster than pushing to the PR to see if they pass.  You can also run pylint locally with `./scripts/pylint.sh`.

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,307 @@
...
+    for extracting entities from a question given a table and type its columns with <string> | <date> | <number>
+    """
+
+    def __init__(self,
+                 cell_values: Set[str],
+                 column_type_statistics: List[Dict[str, int]],
+                 column_index_to_name:  Dict[int, str]) -> None:
+        self.cell_values = cell_values
+        self.column_types = {column_index_to_name[column_index] : max(column_type_statistics[column_index], 
+                                           key=column_type_statistics[column_index].get) for column_index in column_index_to_name}
```
pdasigi: Can you also add another field to this class that stores all the content in the table as a `dict` so that we can pass it directly to the executor? See [here](https://github.com/allenai/allennlp/blob/2dd05bdbe38f2c0ab2c23a7bd06d238b830791eb/allennlp/semparse/executors/wikitables_variable_free_executor.py#L71) for the format expected by the executor.

pdasigi: Also, I think `question_tokens` should be a field in this class. Then `get_entities_from_question` need not take any arguments.

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,307 @@
...
+                column_node_type_info[column_index]['number'] += 1
+            elif node_info['content'] != '—':
+                column_node_type_info[column_index]['string'] += 1
+            index += 1
+
+
+        return cls(cell_values, column_node_type_info, column_index_to_name)
+
+    @classmethod
+    def read_from_file(cls, filename: str, max_tokens_for_num_cell: int) -> 'TableQuestionContext':
```
pdasigi: Seems like `max_tokens_for_num_cell` needs to be a global decision instead of having the user pass this value every time this method is called.

pdasigi: When should this value ever be greater than 1? Is there are problem with always considering values with more than one token as non-numbers?

matt-gardner:
This was merged with Pradeep's other PR.  Right @pdasigi?  Closing this.

pdasigi:
That's correct. Thanks!

PR #1784, opened 2018-09-18
Rename SOURCE_COMMIT to ALLENNLP_SOURCE_COMMIT.

schmmd:
I got pretty confused for a while why the demo is reporting a SHA from allenai/allennlp instead of allenai/allennlp-demo.  I think the root cause is that [Dockerfile.pip](https://github.com/allenai/allennlp/blob/9306e97a43b1502b557f15c89b5f77ec91cf9d6d/Dockerfile.pip) sets the environment variable `SOURCE_COMMIT` which overrides the build argument in https://github.com/allenai/allennlp-demo/blob/master/Dockerfile#L39.

Even if this isn't the root cause, the new name will lead to less confusion in the future now that we intend `Dockerfile.pip` to be used as the base image for other builds.

These environment variables primarily exist to support http://demo.allennlp.org/info.


brendan-ai2:
Great, thanks!

PR #1787, opened 2018-09-19
bidirectional LM + cnn highway encoder + gated cnn encoder

joelgrus:
ok, I think that this is done in an idiomatic AllenNLP way and that it addresses all of the feedback I've gotten so far.

of course this isn't the finished product, but I think these pieces are correct enough that it makes sense to merge them and start working on the next pieces.

### DIFF
--- a/allennlp/data/token_indexers/single_id_token_indexer.py
+++ b/allennlp/data/token_indexers/single_id_token_indexer.py
@@ -1,4 +1,5 @@
 from typing import Dict, List
+import itertools
 
 from overrides import overrides
 
@@ -20,11 +21,22 @@ class SingleIdTokenIndexer(TokenIndexer[int]):
     lowercase_tokens : ``bool``, optional (default=``False``)
         If ``True``, we will call ``token.lower()`` before getting an index for the token from the
         vocabulary.
+    start_tokens : ``List[str]``, optional (default=``None``)
+        These are prepended to the tokens provided to ``tokens_to_indices``.
+    end_tokens : ``List[str]``, optional (default=``None``)
+        These are appended to the tokens provided to ``tokens_to_indices``.
     """
     # pylint: disable=no-self-use
-    def __init__(self, namespace: str = 'tokens', lowercase_tokens: bool = False) -> None:
+    def __init__(self,
+                 namespace: str = 'tokens',
+                 lowercase_tokens: bool = False,
+                 start_tokens: List[str] = None,
+                 end_tokens: List[str] = None) -> None:
         self.namespace = namespace
         self.lowercase_tokens = lowercase_tokens
+
+        self._start_tokens = [Token(st) for st in (start_tokens or [])]
+        self._end_tokens = [Token(et) for et in (end_tokens or [])]
 
     @overrides
     def count_vocab_items(self, token: Token, counter: Dict[str, Dict[str, int]]):
@@ -43,7 +55,7 @@ class SingleIdTokenIndexer(TokenIndexer[int]):
                           index_name: str) -> Dict[str, List[int]]:
         indices: List[int] = []
 
-        for token in tokens:
+        for token in itertools.chain(self._start_tokens, tokens, self._end_tokens):
             if getattr(token, 'text_id', None) is not None:
                 # `text_id` being set on the token means that we aren't using the vocab, we just use
                 # this id instead.
--- a/allennlp/models/bidirectional_lm.py
+++ b/allennlp/models/bidirectional_lm.py
@@ -1,0 +1,284 @@
+from typing import Dict, List, Tuple, Union, Optional
+
+import torch
+import numpy as np
+
+from allennlp.common.checks import ConfigurationError
+from allennlp.data.vocabulary import Vocabulary
+from allennlp.models.model import Model
+from allennlp.modules.masked_layer_norm import MaskedLayerNorm
+from allennlp.modules.text_field_embedders import TextFieldEmbedder
+from allennlp.modules.seq2seq_encoders import Seq2SeqEncoder
+from allennlp.nn.util import get_text_field_mask, remove_sentence_boundaries
+
+
+class _SoftmaxLoss(torch.nn.Module):
+    """
+    Given some embeddings and some targets, applies a linear layer
+    to create logits over possible words and then returns the
+    negative log likelihood.
+    """
+    def __init__(self,
+                 num_words: int,
+                 embedding_dim: int,
+                 token_encoder: torch.nn.Parameter = None) -> None:
+        super().__init__()
+
+        self.tie_embeddings = token_encoder is not None
+
+        # Glorit init (std=(1.0 / sqrt(fan_in))
+        if self.tie_embeddings:
+            self.softmax_w = token_encoder
+            # +1 for shape to include padding dimension
+            self.softmax_b = torch.nn.Parameter(torch.zeros(num_words + 1))
+        else:
+            self.softmax_w = torch.nn.Parameter(
+                    torch.randn(embedding_dim, num_words) / np.sqrt(embedding_dim)
+            )
+            self.softmax_b = torch.nn.Parameter(torch.zeros(num_words))
+
+    def forward(self, embeddings: torch.Tensor, targets: torch.Tensor) -> torch.Tensor:
+        # pylint: disable=arguments-differ
+        # embeddings is size (n, embedding_dim)
+        # targets is (batch_size, ) with the correct class id
+        # Does not do any count normalization / divide by batch size
+        if self.tie_embeddings:
+            softmax_w = self.softmax_w.weight.t()
+        else:
+            softmax_w = self.softmax_w
+
+        probs = torch.nn.functional.log_softmax(
+                torch.matmul(embeddings, softmax_w) + self.softmax_b,
+                dim=-1
+        )
+
+        if self.tie_embeddings:
+            # need to add back in padding dim!
+            targets_ = targets + 1
+        else:
+            targets_ = targets
+
+        return torch.nn.functional.nll_loss(probs, targets_.long(), reduction="sum")
+
+
+@Model.register('bidirectional-language-model')
+class BidirectionalLanguageModel(Model):
+    """
+    The ``BidirectionalLanguageModel`` applies a bidirectional "contextualizing"
+    ``Seq2SeqEncoder`` to uncontextualized embeddings, using a ``SoftmaxLoss``
+    module (defined above) to compute the language modeling loss.
+
+    It is IMPORTANT that your bidirectional ``Seq2SeqEncoder`` does not do any
+    "peeking ahead". That is, for its forward direction it should only consider
+    embeddings at previous timesteps, and for its backward direction only embeddings
+    at subsequent timesteps. If this condition is not met, your language model is
+    cheating.
+
+    Parameters
+    ----------
+    vocab: ``Vocabulary``
+    text_field_embedder: ``TextFieldEmbedder``
+        Used to embed the indexed tokens we get in ``forward``.
+    contextualizer: ``Seq2SeqEncoder``
+        Used to "contextualize" the embeddings. As described above,
+        this encoder must not cheat by peeking ahead.
+    layer_norm: ``MaskedLayerNorm``, optional (default: None)
+        If provided, is applied to the noncontextualized embeddings
+        before they're fed to the contextualizer.
+    dropout: ``float``, optional (default: None)
+        If specified, dropout is applied to the contextualized embeddings.
+    loss_scale: ``Union[float, str]``, optional (default: 1.0)
+        This scaling factor is applied to the average language model loss.
+        You can also specify ``"n_samples"`` in which case we compute total
+        loss across all predictions.
+    remove_bos_eos: ``bool``, optional (default: True)
+        Typically the provided token indexes will be augmented with
+        begin-sentence and end-sentence tokens. If this flag is True
+        the corresponding embeddings will be removed from the return values.
+    """
+    def __init__(self,
+                 vocab: Vocabulary,
+                 text_field_embedder: TextFieldEmbedder,
+                 contextualizer: Seq2SeqEncoder,
+                 layer_norm: Optional[MaskedLayerNorm] = None,
+                 dropout: float = None,
+                 loss_scale: Union[float, str] = 1.0,
+                 remove_bos_eos: bool = True) -> None:
+        super().__init__(vocab)
+        self._text_field_embedder = text_field_embedder
+        self._layer_norm = layer_norm or (lambda x: x)
+
+        if not contextualizer.is_bidirectional():
+            raise ConfigurationError("contextualizer must be bidirectional")
+
+        self._contextualizer = contextualizer
+        # The dimension for making predictions just in the forward
+        # (or backward) direction.
+        self._forward_dim = contextualizer.get_output_dim() // 2
+
+        # TODO(joelgrus): Allow SampledSoftmaxLoss here by configuration
+        self._softmax_loss = _SoftmaxLoss(num_words=vocab.get_vocab_size(),
+                                          embedding_dim=self._forward_dim)
+
+        self.register_buffer('_last_average_loss', torch.zeros(1))
+
+        if dropout:
+            self._dropout = torch.nn.Dropout(dropout)
+        else:
+            self._dropout = lambda x: x
+
+        self._loss_scale = loss_scale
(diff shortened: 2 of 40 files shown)


allennlp/modules/contextual_encoders/contextual_encoder.py
```diff
@@ -0,0 +1,28 @@
+from typing import Dict
+
+import torch
+
+from allennlp.common.registrable import Registrable
+
+class ContextualEncoder(Registrable, torch.nn.Module):
```
matt-gardner: I know this isn't ready for review yet; feel free to take your time answering this.  I'm just wondering what the difference is between this abstraction and our existing `Seq2SeqEncoder` abstraction.  They look identical to me.

joelgrus: you're right, they're mostly identical save a property or two, let me think about how to reuse Seq2SeqEncoder here

allennlp/modules/contextual_encoder/character_encoder.py
```diff
@@ -0,0 +1,172 @@
...
+
+import torch
+import numpy as np
+
+from allennlp.common.checks import ConfigurationError
+from allennlp.modules.highway import Highway
+from allennlp.modules.masked_layer_norm import MaskedLayerNorm
+from allennlp.nn.util import add_sentence_boundary_token_ids
+
+_DEFAULT_FILTERS = ((1, 4), (2, 8), (3, 16), (4, 32), (5, 64))
```
matt-peters: Make the filters required input, too much potential for hidden problem with defaults.

allennlp/tests/modules/contextual_encoder/contextual_encoder_test.py
```diff
@@ -0,0 +1,77 @@
...
+                             hidden_size=10,
+                             batch_first=True)
+        self.seq2seq = PytorchSeq2SeqWrapper(lstm)
+
+
+    def test_char_level_contextual_encoder(self):
+        cew = CharLevelContextualEncoder(
+                encoder=self.seq2seq,
+                character_encoder=self.character_encoder,
+                num_layers=3,
```
matt-peters: I think `num_layers` should be 4.. it isn't a problem here, since it's never tested.
The `num_layers` thing is so that the contextual encoder can be automatically wrapped with the `Elmo` scalar weighting to make a token embedder.

joelgrus:
ok, I think I addressed all of your comments.

also, I just ran an evaluation on ner using this code and got the expected results:

`$ python -m allennlp.run evaluate https://s3-us-west-2.amazonaws.com/allennlp/models/ner-model-2018.04.30.tar.gz eng.testb`

```
2018-09-21 13:05:20,148 - INFO - allennlp.commands.evaluate - Finished evaluating.
2018-09-21 13:05:20,148 - INFO - allennlp.commands.evaluate - Metrics:
2018-09-21 13:05:20,148 - INFO - allennlp.commands.evaluate - accuracy: 0.9824485840422096
2018-09-21 13:05:20,148 - INFO - allennlp.commands.evaluate - accuracy3: 0.9842790998169484
2018-09-21 13:05:20,148 - INFO - allennlp.commands.evaluate - precision-overall: 0.9199789879180529
2018-09-21 13:05:20,148 - INFO - allennlp.commands.evaluate - recall-overall: 0.9302407932011332
2018-09-21 13:05:20,148 - INFO - allennlp.commands.evaluate - f1-measure-overall: 0.9250814332247058
```

allennlp/modules/contextual_encoder/contextual_encoder.py
```diff
@@ -0,0 +1,171 @@
...
+
+import torch
+
+from allennlp.common import Registrable
+from allennlp.modules.contextual_encoder.character_encoder import CharacterEncoder
+from allennlp.modules.masked_layer_norm import MaskedLayerNorm
+from allennlp.modules.seq2seq_encoders import Seq2SeqEncoder
+from allennlp.modules.token_embedders import TokenEmbedder
+
+class ContextualEncoder(torch.nn.Module, Registrable):
```
matt-gardner: How does a `ContextualEncoder` get used?  Where does it get its inputs?  From a `TextField`?  I'm just worried about there being too many options for going from text to embeddings, and that the ways they are used are mutually exclusive, so you have to change how you process your data if you want to switch between them, which breaks one of the main points of the library.

Similarly, the `CharacterEncoder` looks like a fancier version of our [`CnnEncoder`](https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2vec_encoders/cnn_encoder.py) wrapped inside a `TokenCharactersEmbedder`.  It's great to have a better version of this, except now we have two ways of going through the code to get almost the same result.

It seems like we're getting a proliferation of abstractions.  I _think_ the goal of this one is to have something that wraps a `TextFieldEmbedder` and then a `Seq2SeqEncoder`, adding a few options along the way.  That's great, I see no issue there.  But the way it currently is, it doesn't actually use the `TextFieldEmbedder`, it just duplicates _almost_ the same functionality in an incompatible way, so we can't actually use things interchangeably anymore.

matt-gardner: I mean, another option here is to just add a `ContextualizedTextFieldEmbedder` as a subclass of `TextFieldEmbedder`, where you group a set of `TokenEmbedders` with a `Seq2SeqEncoder`.

matt-gardner: Sorry to put all of these comments here, too, but the `TokenCharactersEmbedder` already handles the mask, and the `CharacterEncoder` here doesn't actually use the mask, it just returns it.  So if you just soup up our current `CnnEncoder` with highway layers, you could remove the whole masking logic and the dictionary return type.

joelgrus: no, this is good. what you are pointing out is that my approach to the Calypso -> AllenNLP project is too "literal" and that I need to step back and be a bit more holistic about it.

that is, I have been mostly approaching it as "how do I get this code working inside AllenNLP" but I should really be approaching it as "what's the right way to get this code incorporated *as part of* AllenNLP". let me think about that some.

matt-gardner: Minor addendum to what I said above about the `CnnEncoder` - I wouldn't modify our existing one, as that would break our saved models.  But you could add a `CnnHighwayEncoder`, or something, that's basically your `CharacterEncoder` without the mask, and use that inside a `TokenCharactersEmbedder` to get the same effect.

joelgrus: I am really struggling with the right way to do this

* TokenCharactersEncoder doesn't quite work, because it does character_ids -> embedding [add a dim] -> seq2vec_encoder [drop a dim] -> outputs

whereas ContextualEncoder wants to do character_ids -> character_encoder (which is really a seq2seq encoder) -> contextual seq2seq encoder -> outputs

--

so one possibility is to make the CharacterEncoder/CnnHighwayEncoder a TokenEmbedder, then the ContextualEncoder itself a TokenEmbedder that wraps another TokenEmbedder? (since that's really what it is)

the usage inside our current ELMo model will still be nonstandard, but then things fit more nicely into the framework?

allennlp/modules/token_embedders/cnn_highway_encoder.py
```diff
@@ -0,0 +1,214 @@
...
+    max_characters_per_token: int, optional (default = 50)
+        The maximum length of any token.
+    num_characters: int, optional (default = 262)
+        The number of characters in the vocabulary.
+    projection_location: str, optional (default = 'after_highway')
+        Where to apply the projection layer. Valid values are
+        'after_highway', 'after_cnn', and None.
+    do_layer_norm: bool, optional (default = False)
+        If True, we apply ``MaskedLayerNorm`` to the final encoded result.
+    bos_characters: ``List[int]``, optional (default = None)
```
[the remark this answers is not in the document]
joelgrus: I am still struggling with this a lot.

I took a step forward and implemented a bare-bones `BidirectionalLM` that looks like

```
    def __init__(self,
                 vocab: Vocabulary,
                 text_field_embedder: TextFieldEmbedder,
                 contextualizer: Seq2SeqEncoder,
                 softmax: Softmax,
                 dropout: float = None,
                 loss_scale_fac: Union[float, str] = 1.0) -> None:
```

and then has a forward that just looks like

```
                tokens: Dict[str, torch.LongTensor]) -> Dict[str, torch.Tensor]:
```

in order to compute the LM targets, I insist that there was a `{"tokens": SingleIdTokenIndexer()}` and raise a `ConfigurationError` if there was not. I think this part is relatively straightforward.

I am still struggling with how to handle the rest of it.

Your suggestion (I think) was to put all of the bos/eos logic in a new `TextFieldEmbedder` subclass. But a `TextFieldEmbedder` is agnostic to the type/number of "token_indexes" it receives, whereas the bos/eos logic is specific to the actual indexes themselves? If you're doing it for characters, you have to add the bos characters; if you're doing it for tokens (do you even do it for tokens?), you just have to add a bos token, but I don't see an easy way to do it in the abstract?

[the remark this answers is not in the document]
matt-gardner: Maybe easier to talk about this in person?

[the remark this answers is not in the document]
joelgrus: I'm wfh today, I'll find some time tomorrow

joelgrus:
@matt-gardner if you could take another quick look at this and see if you agree with this approach.

basically:

* `BidirectionalLM` is a `Model`
* now `CnnHighwayEncoder` is just a `Seq2VecEncoder`, and you just use it inside the usual `TokenCharactersEncoder` with the usual `BasicTextFieldEmbedder` (it's possible there's too much overlap with the existing `CnnEncoder`, but I don't want to worry about that just yet)
* the BOS/EOS tokens get added by the `TokenIndexer`s and get removed by the model (this feels sort of clunky, but I couldn't think of a cleaner way)

I think this way is pretty allennlp-like, let me know what you think

allennlp/data/token_indexers/single_id_token_indexer.py
```diff
@@ -22,10 +23,23 @@ class SingleIdTokenIndexer(TokenIndexer[int]):
         vocabulary.
     """
     # pylint: disable=no-self-use
-    def __init__(self, namespace: str = 'tokens', lowercase_tokens: bool = False) -> None:
+    def __init__(self,
+                 namespace: str = 'tokens',
+                 lowercase_tokens: bool = False,
+                 bos_token: str = None,
```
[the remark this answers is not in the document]
matt-gardner: Oh, one issue with that is that maybe the text is pre-tokenized?  Then the dataset reader would know that, and the parameters can just be added to that particular dataset reader?  I'm not sure you actually gain anything by having these parameters in this class, and it makes things easier to break.

allennlp/models/bidirectional_lm.py
```diff
@@ -0,0 +1,180 @@
...
+        else:
+            # average_loss zero tensor, return it for all
+            return_dict = {
+                    'loss': average_loss,
+                    'forward_loss': average_loss,
+                    'backward_loss': average_loss
+            }
+
+        # TODO: add non-loss functions to the output
+        if self._remove_bos_eos:
```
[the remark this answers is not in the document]
matt-peters: We want to include these tokens in the loss, so that the LM learns to predict them.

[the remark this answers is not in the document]
matt-gardner: How do you compute perplexity, then?  You strip them out before doing that?

allennlp/models/bidirectional_lm.py
```diff
@@ -0,0 +1,180 @@
...
+                 layer_norm: Optional[MaskedLayerNorm] = None,
+                 dropout: float = None,
+                 loss_scale_fac: Union[float, str] = 1.0,
+                 remove_bos_eos: bool = True,
+                 return_all_layers: bool = False) -> None:
+        super().__init__(vocab)
+        self._text_field_embedder = text_field_embedder
+        self._layer_norm = layer_norm
+        self._contextualizer = contextualizer
+        self._forward_dim = contextualizer.get_output_dim() // 2
```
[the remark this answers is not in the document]
matt-peters: Really we should support both bidirectional and forward only LMs with a flag or something.

allennlp/data/token_indexers/single_id_token_indexer.py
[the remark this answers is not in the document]
matt-peters: It seems pretty annoying for every data set reader that handles pre-tokenized text to have to worry about this particular detail to support a model that they may not use?

allennlp/models/bidirectional_lm.py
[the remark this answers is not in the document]
matt-gardner: Oh, right, the name is "bidirectional LM", so of course this has a backward component.  I guess if you want a single class to support both, you should probably rename it.  But the "LM" component just refers to the loss, right?  Whether the _contextualizer_ is bidirectional doesn't matter for this - e.g., the transformer is "directional" at all.  My point here was just about `self._forward_dim` - I was confusing myself about a bidirectional _contextualizer_, but this refers to the component that will be used for the loss.  A comment saying that explicitly would be nice.

allennlp/data/token_indexers/single_id_token_indexer.py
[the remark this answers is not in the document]
matt-peters: Yes, in addition to training, the main use case for the `BidirectionalLanguageModel` is including a pre-trained version of one as an essential component of another supervised model as a `TextFieldEmbedder` or such.

allennlp/models/bidirectional_lm.py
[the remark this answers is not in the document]
matt-peters: This code follows the convention for the 1 Billion word benchmark which for the forward LM includes `</S>` but `<S>` is masked.  The backward LM is similar but predicts `<S>` and masks `</S>`.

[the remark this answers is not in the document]
matt-gardner: :+1:

allennlp/data/token_indexers/single_id_token_indexer.py
```diff
@@ -43,7 +55,7 @@ def tokens_to_indices(self,
                           index_name: str) -> Dict[str, List[int]]:
         indices: List[int] = []
 
-        for token in tokens:
+        for token in itertools.chain(self._start_tokens, tokens, self._end_tokens):
```
matt-gardner: I need to remember to use itertools more often...  I would have just done `self._start_tokens + tokens + self._end_tokens`, which is less efficient.

allennlp/models/bidirectional_lm.py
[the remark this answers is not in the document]
matt-gardner: It'd be nice to have a small comment explaining what this forward dim is - it's used for _predicting_ in the forward direction.

allennlp/modules/softmax.py
```diff
@@ -0,0 +1,51 @@
+"""
+The Softmax from calypso
+"""
+
+import numpy as np
+import torch
+
+class Softmax(torch.nn.Module):
+    def __init__(self,
```
matt-gardner: This class could use a docstring describing what it's doing, and when you might want to use this instead of `torch.nn.functional.softmax`.

allennlp/tests/models/bidirectional_lm_test.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+        tokens_to_add = {
+                "tokens": START_TOKENS + END_TOKENS,
+                "token_characters": [c for token in START_TOKENS + END_TOKENS for c in token]
+        }
+
+        self.vocab = Vocabulary.from_instances(self.instances, tokens_to_add=tokens_to_add)
+
+    def test_lm_can_run(self):
```
matt-gardner: What does this buy you over our typical model test?  I think it's trying to test individual pieces, but those should already be tested.  It's also testing that things work if you configure them correctly, but that's already done in our typical tests (and done better, because it also checks gradients and other things).

allennlp/tests/models/bidirectional_lm_test.py
```diff
@@ -0,0 +1,115 @@
...
+
+        trainer = Trainer(model=model,
+                          optimizer=optimizer,
+                          iterator=iterator,
+                          train_dataset=self.instances,
+                          num_epochs=10)
+
+        trainer.train()
+
+        # Loss should be lower
```
matt-gardner: This is a piece that we could consider adding to our main model test.

PR #1792, opened 2018-09-19
Atis model action refactored

kl2806:
Summary:
1. Add `AtisSemanticParser` model, which implements a baseline model that applies type-constrained decoding for SQL.
2. Reverts all entities to be `linked` actions with linking scores.
3. Changes the base definition of the SQL grammar from a string to dictionary, so that it is easier for models to modify binary expressions, etc. for additional constraints.

Other minor changes:
- Add `time_range_start` and `time_range_end` as two new types in the grammar. This allows the model to disambiguate between start times and end times for constraints like `BETWEEN 0 and 1200`.
- Add a reverse_productions flag to the `GrammarStatelet`; we need this because we want to expand the right-most nonterminal first in our production rules.
- Allow nonterminals with only linked actions
- Read all SQL query labels for an utterance and execute each when evaluating denotation accuracy


allennlp/semparse/contexts/sql_table_context.py
```diff
@@ -116,17 +111,23 @@ class SqlTableContext():
     def __init__(self,
                  all_tables: Dict[str, List[str]] = None,
                  tables_with_strings: Dict[str, List[str]] = None,
-                 database_directory: str = None) -> None:
+                 database_directory: str = None,
+                 utterances: List[str] = None) -> None:
+        self.grammar_dictionary = deepcopy(GRAMMAR_DICTIONARY)
+        self.utterances = utterances
```
matt-gardner: Why do you need the utterances here?  I thought this was supposed to be a general SQL class, not an ATIS one, and it doesn't look like you're actually using the utterances (or the tokenizer) inside this class.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -1,19 +1,47 @@
...
 from allennlp.semparse.contexts.atis_tables import * # pylint: disable=wildcard-import,unused-wildcard-import
 from allennlp.semparse.contexts.sql_table_context import \
-        SqlTableContext, SqlVisitor, generate_one_of_string, format_action
+        SqlTableContext, SqlVisitor, format_action
 
 from allennlp.data.tokenizers import Token, Tokenizer, WordTokenizer
 
+def get_strings_from_utterance(tokenized_utterance: List[Token]) -> Dict[str, List[int]]:
+    """
+    Based on the current utterance, return a dictionary where the keys are the strings in the utterance
```
matt-gardner: If you read this out of context, you wonder what "the strings" mean.  You mean this to be "strings from the database", but that's not clear anywhere, so it's a bit confusing.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -31,74 +59,182 @@ def __init__(self,
...
-                                                     TABLES_WITH_STRINGS,
-                                                     database_directory)
+            self.database_directory = database_directory
+            self.connection = sqlite3.connect(self.database_directory)
+            self.cursor = self.connection.cursor()
+
+        self.sql_table_context = SqlTableContext(ALL_TABLES,
+                                                 TABLES_WITH_STRINGS,
+                                                 database_directory,
+                                                 utterances=utterances) if database_directory else None
```
matt-gardner: It doesn't look like you use `sql_table_context.utterances` here, either; I'm not sure why you added it, because you already save it to `AtisWorld`.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -31,74 +59,182 @@ def __init__(self,
...
+
+        self.grammar_dictionary = deepcopy(self.sql_table_context.grammar_dictionary)
```
matt-gardner: I don't think you need this `deepcopy`, because you're instantiating one `SqlTableContext` per `AtisWorld`, and the `SqlTableContext` already did a `deepcopy`.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -31,74 +59,182 @@ def __init__(self,
...
 
         self.utterances: List[str] = utterances
         self.tokenizer = tokenizer if tokenizer else WordTokenizer()
         self.tokenized_utterances = [self.tokenizer.tokenize(utterance) for utterance in self.utterances]
-        valid_actions, linking_scores = self.init_all_valid_actions()
-        self.valid_actions: Dict[str, List[str]] = valid_actions
+        self.linked_entities = self.get_linked_entities()
```
matt-gardner: This method, and `update_valid_actions` and `flatten_entities`, look like they should be private.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -31,74 +59,182 @@ def __init__(self,
...
 
+        self.valid_actions: Dict[str, List[str]] = self.update_valid_actions()
         # This has shape (num_entities, num_utterance_tokens).
```
matt-gardner: This comment should be moved one line below, I think.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -31,74 +59,182 @@ def __init__(self,
...
+        self.tables_with_strings = TABLES_WITH_STRINGS
         if database_directory:
-            self.sql_table_context = SqlTableContext(ALL_TABLES,
-                                                     TABLES_WITH_STRINGS,
-                                                     database_directory)
+            self.database_directory = database_directory
+            self.connection = sqlite3.connect(self.database_directory)
+            self.cursor = self.connection.cursor()
+
+        self.sql_table_context = SqlTableContext(ALL_TABLES,
```
matt-gardner: Does the order of this mean that you have two concurrent connections to the db open?  I'm not sure what performance issues / race conditions that might induce.  It's probably fine.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -31,74 +59,182 @@ def __init__(self,
...
-        utterances up to and including the current utterance and add the valid strings.
-        """
+    def add_to_number_linking_scores(self,
+                                     all_numbers: Set[str],
+                                     number_linking_scores: Dict[str, Tuple[str, str, List[int]]],
+                                     get_number_linking_dict: Callable[[str, List[Token]],
+                                                                       Dict[str, List[int]]],
+                                     current_tokenized_utterance: List[Token],
+                                     nonterminal: str) -> None:
+        number_linking_dict: Dict[str, List[int]] = {}
```
matt-gardner: I'm finding myself trying to figure out what this logic is doing by reading the code, and it's a bit hard.  A high-level comment or docstring explaining what this method is doing would be good.  And the same thing goes for most of the internal methods on this class.

allennlp/state_machines/transition_functions/linking_transition_function.py
```diff
@@ -96,11 +96,19 @@ def _compute_action_probabilities(self,
...
             instance_actions = actions[group_index]
             predicted_action_embedding = predicted_action_embeddings[group_index]
-            action_embeddings, output_action_embeddings, embedded_actions = instance_actions['global']
-            # This is just a matrix product between a (num_actions, embedding_dim) matrix and an
-            # (embedding_dim, 1) matrix.
-            embedded_action_logits = action_embeddings.mm(predicted_action_embedding.unsqueeze(-1)).squeeze(-1)
-            action_ids = embedded_actions
+            embedded_actions: List[int] = []
+
+            output_action_embeddings = predicted_action_embeddings.new_tensor([], dtype=torch.float)
```
matt-gardner: I don't think you actually need to construct these tensors.  Just have whatever variables are needed here be `None`, and assign them appropriately below depending on the path the logic takes.

allennlp/data/dataset_readers/semantic_parsing/atis.py
```diff
@@ -142,7 +148,8 @@ def text_to_instance(self,  # type: ignore
                   'world' : world_field,
                   'linking_scores' : ArrayField(world.linking_scores)}
 
-        if sql_query:
+        if sql_query_labels != None:
+            fields['example_sql_queries'] = MetadataField(sql_query_labels)
```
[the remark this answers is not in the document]
kl2806: Oh, yes that makes more sense.

allennlp/semparse/contexts/sql_table_context.py
kl2806: I'll remove it; I was trying to constrain dates in the `SqlTableContext` but it is no longer needed.

allennlp/semparse/contexts/sql_table_context.py
```diff
@@ -29,65 +32,57 @@
...
-    agg_results         = (ws "("  ws "SELECT" ws distinct ws agg ws "FROM" ws table_name ws where_clause ws ")" ws) /
-                          (ws "SELECT" ws distinct ws agg ws "FROM" ws table_name ws where_clause ws)
-    boolean             = "true" / "false"
-    ws                  = ~"\s*"i
-    conj                = "AND" / "OR" 
-    distinct            = ("DISTINCT") / ("")
-    number              =  ""
-"""
+GRAMMAR_DICTIONARY = {
+        'statement'         :['query ws ";" ws'],
```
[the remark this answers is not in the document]
kl2806: Hmm, I don't have a strong opinion about this; what's the benefit of changing it to this way? The current formatting keeps the right hand side rules aligned and avoids repeating `GRAMMAR_DICTIONARY`.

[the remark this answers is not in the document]
matt-gardner: This just feels very non-standard to me, and it's odd for me to look at.  I don't have a strong opinion either.  I also don't see a whole lot of point in keeping the RHS aligned.  It's up to you.

allennlp/semparse/worlds/atis_world.py
kl2806: We are only doing reads here, so multiple processes can access the database without any locking/waiting (https://www.sqlite.org/faq.html#q5).

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
```diff
@@ -0,0 +1,592 @@
...
+            ``ProductionRuleArray`` using a ``ProductionRuleField``.  We will embed all of these
+            and use the embeddings to determine which action to take at each timestep in the
+            decoder.
+        linking_scores: ``torch.Tensor``
+            A matrix of the linking the utterance tokens and the entities. This is a binary matrix that
+            is deterministically generated where each entry indicates whether a token generated an entity.
+            This tensor has shape ``(num_entities, num_utterance_tokens)``.
+        target_action_sequences : torch.Tensor, optional (default=None)
+            A list of possibly valid action sequences, where each action is an index into the list
+            of possible actions.  This tensor has shape ``(batch_size, num_action_sequences,
```
matt-gardner: `num_action_sequences` is always 1 for you.  If you have this inside a `ListField` in your dataset reader, I'd probably remove that, and just have inputs of size `(batch_size, sequence_length)`.  You can `.unsqueeze()` it when passing it to `MaximumMarginalLikelihood` (and @rajasagashe is planning on submitting a `MaximumLikelihood` class sometime soon, which we can use instead of `MaximumMarginalLikelihood` once it's in).

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
```diff
@@ -0,0 +1,592 @@
...
+
+        initial_state = GrammarBasedState(batch_indices=list(range(batch_size)),
+                                          action_history=[[] for _ in range(batch_size)],
+                                          score=initial_score_list,
+                                          rnn_state=initial_rnn_state,
+                                          grammar_state=initial_grammar_state,
+                                          possible_actions=actions,
+                                          debug_info=None)
+
+        return {"initial_state": initial_state}
```
matt-gardner: This is a private method with a single return value.  There's no reason to have this be a dictionary.

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        self._decoder_trainer = MaximumMarginalLikelihood(training_beam_size)
+        self._decoder_step = LinkingTransitionFunction(encoder_output_dim=self._encoder.get_output_dim(),
+                                                       action_embedding_dim=action_embedding_dim,
+                                                       input_attention=input_attention,
+                                                       num_start_types=self._num_start_types,
+                                                       predict_start_type_separately=False,
+                                                       add_action_bias=self._add_action_bias,
+                                                       dropout=dropout)
+
+    def _get_initial_state_and_scores(self,
```
matt-gardner: This is just getting this initial state at this point, not any scores.

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        target_action_sequences : torch.Tensor, optional (default=None)
+            A list of possibly valid action sequences, where each action is an index into the list
+            of possible actions.  This tensor has shape ``(batch_size, num_action_sequences,
+            sequence_length)``.
+        sql_queries : List[str], otpional (default=None)
+            A list of the SQL queries that are given during training or validation.
+        """
+        initial_info = self._get_initial_state_and_scores(utterance, world, actions, linking_scores)
+        initial_state = initial_info["initial_state"]
+        batch_size = list(utterance.values())[0].size(0)
```
matt-gardner: Probably easier to get this from `linking_scores` or `target_action_sequence`.

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
```diff
@@ -0,0 +1,592 @@
...
+            outputs['best_action_sequence'] = []
+            outputs['debug_info'] = []
+            outputs['entities'] = []
+            outputs['logical_form'] = []
+            outputs['sql_queries'] = []
+            outputs['utterance'] = []
+            outputs['tokenized_utterance'] = []
+
+            for i in range(batch_size):
+                # Decoding may not have terminated with any completed logical forms, if `num_steps`
```
matt-gardner: Any time you say "logical forms" in this class, you should probably replace it with "SQL queries".

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
```diff
@@ -0,0 +1,592 @@
...
+                # isn't long enough (or if the model is not trained enough and gets into an
+                # infinite action loop).
+                if i in best_final_states:
```
matt-gardner: It's probably a bit nicer to read if you just do

```python
if i not in best_final_states:
    continue
```

Then there's less indentation on the rest.  We should probably change our other models to do this, too. (Not saying that you should do that in this PR.)

matt-gardner: Though, you probably want to `continue` only after you've counted some zeros for your metrics, so you don't overestimate your accuracy.

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
```diff
@@ -0,0 +1,592 @@
...
+                        self._action_similarity(similarity.ratio())
+
+                    if sql_queries and sql_queries[i]:
+                        # Since the query might hang, we run in another process and kill it if it
+                        # takes too long.
+                        process = Process(target=self._sql_result_match,
+                                          args=(predicted_sql_query, sql_queries[i]))
+                        process.start()
+
+                        # If the query has not finished in 10 seconds then we will proceed.
```
matt-gardner: I'd probably give it a lot less than 10 seconds; the query should take milliseconds, I'd imagine.  You could also retry if you hang...  Not sure it's worth it.  You might consider including a metric that counts how many times you hang (you can prefix it with an underscore so it doesn't show up in tqdm).

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
```diff
@@ -0,0 +1,592 @@
...
+                               {},
+                               self.is_nonterminal,
+                               reverse_productions=False)
+
+    @overrides
+    def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
+        """
+        This method overrides ``Model.decode``, which gets called after ``Model.forward``, at test
+        time, to finalize predictions.  This is (confusingly) a separate notion from the "decoder"
+        in "encoder/decoder", where that decoder logic lives in ``GrammarBasedState``.
```
matt-gardner: Hmm, the decoder logic lives in the `TransitionFunction`.  If we say this in other places, we should fix it there too.

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
```diff
@@ -0,0 +1,592 @@
...
+            if predicted_rows == target_rows:
+                exit(1)
+        exit(0)
+
+    @overrides
+    def get_metrics(self, reset: bool = False) -> Dict[str, float]:
+        """
+        We track four metrics here:
+
+            1. dpd_acc, which is the percentage of the time that our best output action sequence is
```
matt-gardner: I'd call this `exact_match`, or something.  DPD accuracy is referring to something else.

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
```diff
@@ -0,0 +1,592 @@
...
+            in the set of action sequences provided by DPD.
+
+            2. denotation_acc, which is the percentage of examples where we get the correct
+            denotation.  This is the typical "accuracy" metric, and it is what you should usually
+            report in an experimental result.  You need to be careful, though, that you're
+            computing this on the full data, and not just the subset that has DPD output (make sure
```
matt-gardner: This could use some updating, too.

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
```diff
@@ -0,0 +1,592 @@
...
+
+    def _sql_result_match(self, predicted_query: str, sql_query_labels: List[str]) -> int:
+        postprocessed_predicted_query = self._postprocess_query_sqlite(predicted_query)
+
+        try:
+            self._cursor.execute(postprocessed_predicted_query)
+            predicted_rows = self._cursor.fetchall()
+        except sqlite3.Error as error:
+            logger.debug(f'Error executing predicted: {error}')
+            exit(0)
```
matt-gardner: Should you just `raise` here?  Calling `exit()` is generally not a good idea, unless you're writing a shell script.  Looks like the above line should also be `logger.error()`, right?

matt-gardner: Oh!  This is used in the subprocess.  This definitely should have a docstring explaining a bit more about what's going on.  And in that case, yes, `logger.debug()` is the right thing here.

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
```diff
@@ -0,0 +1,592 @@
+import logging
```
matt-gardner: We should not have the ATIS db checked into the repo like it is (commenting here, because I can't comment on the db file).  Instead, put it somewhere on S3 and use a URL and `cached_path` to access it.  I can help you put it on S3 if you have trouble.

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
kl2806: Ah yes, I'll add a comment -- we only ever call this in the subprocess.

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
```diff
@@ -0,0 +1,592 @@
...
+    max_decoding_steps : ``int``
+        When we're decoding with a beam search, what's the maximum number of steps we should take?
+        This only applies at evaluation time, not during training.
+    input_attention: ``Attention``
+        We compute an attention over the input utterance at each step of the decoder, using the
+        decoder hidden state as the query.  Passed to the transition function.
+    add_action_bias : ``bool``, optional (default=True)
+        If ``True``, we will learn a bias weight for each action that gets used when predicting
+        that action, in addition to its embedding.
+    training_beam_size : ``int``, optional (default=None)
```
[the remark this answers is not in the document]
kl2806: Oh okay, yeah, right now if there are multiple queries, I just pick the shortest one.

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
kl2806: Added it as a dataset here: https://s3-us-west-2.amazonaws.com/allennlp/datasets/atis/atis.db.

kl2806:
Thanks for the review! @matt-gardner this is ready to be looked at again. I'll make a separate PR for the SQL executor.

allennlp/state_machines/transition_functions/linking_transition_function.py
```diff
@@ -113,7 +121,7 @@ def _compute_action_probabilities(self,
                 # The `output_action_embeddings` tensor gets used later as the input to the next
                 # decoder step.  For linked actions, we don't have any action embedding, so we use
                 # the entity type instead.
-                output_action_embeddings = torch.cat([output_action_embeddings, type_embeddings], dim=0)
```
matt-gardner: You need an `if` here, like you have below.

allennlp/models/semantic_parsing/atis/atis_semantic_parser.py
```diff
@@ -0,0 +1,597 @@
...
+    def _postprocess_query_sqlite(query: str):
+        # The dialect of SQL that SQLite takes is not exactly the same as the labeled data.
+        # We strip off the parentheses that surround the entire query here.
+        query = query.strip()
+        if query.startswith('('):
+            return query[1:query.rfind(')')] + ';'
+        return query
+
+    @staticmethod
+    def action_sequence_to_sql(action_sequences: List[str]) -> str:
```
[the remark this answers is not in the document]
kl2806: Right, I think we can move it to `SqlTableContext` for now.

allennlp/semparse/worlds/atis_world.py
```diff
@@ -22,83 +50,212 @@ class AtisWorld():
...
+        self.linked_entities = self._get_linked_entities()
+        self.dates = self._get_dates()
 
+        self.valid_actions: Dict[str, List[str]] = self._update_valid_actions()
+        entities, linking_scores = self._flatten_entities()
         # This has shape (num_entities, num_utterance_tokens).
         self.linking_scores: numpy.ndarray = linking_scores
+        self.entities: List[str] = entities
         self.grammar_str: str = self.get_grammar_str()
         self.grammar_with_context: Grammar = Grammar(self.grammar_str)
```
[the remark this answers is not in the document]
kl2806: We can also do that, but I would rather rely on the parsing library, they point out some advantages here: https://github.com/erikrose/parsimonious/blob/master/parsimonious/grammar.py#L41

PR #1799, opened 2018-09-20
[WIP] Dataset Reader

rajasagashe:


rajasagashe:
@matt-gardner I wanted to ask you about how to deal with the target actions and global actions:
Global Actions: Since there are over 30,000 global actions which are the same for every batch using the ProductionRuleField and embedding them every time is slow. Is there a good way to have this done only once? In another branch of my code, I'm storing a dictionary of nonterminals2action_embeddings in the semantic parser model file.

Target Actions: These are currently converted to indices, but I think its more interpretable to leave them as strings, and the key to nonterminal2action_embeddings would be a nonterminal string. What are your thoughts?

Also I'm assuming I should remove the code in the dataset reader for the experiments I ran related to modifying grammar, prototypes etc.?

matt-gardner:
So, to be clear, the issue is that running the `action_embedding` on the tensors at each batch is slow?  Instead, you do what?  You keep a parameter vector for each action in a dictionary that gets updated directly, and instead of using an `index_select` to get action representations (which is essentially what the `Embedding` does), you use a dictionary or array lookup?  And you only construct the global `valid_actions` dictionary once, using these parameters?  Yes, I can see how this could be faster.  But that's on the modeling side, not on the dataset reader side.  What does it mean for this PR?

On string versus integer lookups, if you're looking to improve runtimes, I'd guess that you'd be better off using integer lookups on a list than using string lookups on a dictionary.

And yeah, if there's code that was just for running experiments that failed, I'd probably remove it.

rajasagashe:
1. Yes thats right. It's relevant to the PR since the dataset reader can write out all the nonterminal2actions. The modeling side can then read this dictionary and initialize the Embedding and nonterminal2action_embeddings dictionary.
2. That's true. I wanted to use strings since the code is more interpretable, and runtimes don't seem to be that much slower. I'm ok with anything here, currently its not using strings but in my other branch it is. I'm not looking to over-optimize, it was a concern of mine since the model was taking like 13+ hours per epoch to train.

matt-gardner:
Closing this as stale.  @rajasagashe, if you're interested in getting this merged, let us know.

PR #1594, opened 2018-08-08
BiMPM model

handsomezebra:
This PR includes the implementation of BiMPM model (#1503) and a Quora paraphrase dataset reader. I am still working on some comments improvements and memory usage optimization but I want to send out the PR to collect feedback too.

Thanks!

PR #1802, opened 2018-09-20
Create a markdown file that enumerates available models

schmmd:
Partially addresses: https://github.com/allenai/allennlp/pull/1802

### DIFF
--- a/MODELS.md
+++ b/MODELS.md
@@ -1,7 +1,7 @@
 # Models
 
 A repository for the available models for AllenNLP.  While we highlight a particular model for
-each task on https://allennlp.org/models we often have other permutations of our models that might
+each task on https://allennlp.org/models we often have other trained models that might
 work better for a particular application.
 
 
--- a/MODELS.md
+++ b/MODELS.md
@@ -7,7 +7,9 @@ work better for a particular application.
 
 ## Machine Comprehension
 
-### [bidaf-model-2017.09.15-charpad.tar.gz](https://s3-us-west-2.amazonaws.com/allennlp/models/bidaf-model-2017.09.15-charpad.tar.gz) (based on [BiDAF (Seo et al, 2017)](https://www.semanticscholar.org/paper/Bidirectional-Attention-Flow-for-Machine-Comprehen-Seo-Kembhavi/007ab5528b3bd310a80d553cccad4b78dc496b02))
+### [bidaf-model-2017.09.15-charpad.tar.gz](https://s3-us-west-2.amazonaws.com/allennlp/models/bidaf-model-2017.09.15-charpad.tar.gz)
+
+Based on [BiDAF (Seo et al, 2017)](https://www.semanticscholar.org/paper/Bidirectional-Attention-Flow-for-Machine-Comprehen-Seo-Kembhavi/007ab5528b3bd310a80d553cccad4b78dc496b02)
 
 ```
 $ docker run allennlp/allennlp:v0.6.1 evaluate https://s3-us-west-2.amazonaws.com/allennlp/models/bidaf-model-2017.09.15-charpad.tar.gz https://s3-us-west-2.amazonaws.com/allennlp/datasets/squad/squad-dev-v1.1.json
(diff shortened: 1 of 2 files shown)


schmmd:
I need to better organize the last 4 sections.  Contributions welcome (particularly for ESIM, where I don't have many details).

* Wikitables
* Event2Mind
* BiMPM
* ESIM

MODELS.md
```diff
@@ -0,0 +1,122 @@
...
+
+* [event2mind-2018.09.17.tar.gz](https://s3-us-west-2.amazonaws.com/allennlp/models/event2mind-2018.09.17.tar.gz)
+
+
+## BiMPM
+
+* [bimpm-quora-2018.08.17.tar.gz](https://s3-us-west-2.amazonaws.com/allennlp/datasets/quora-question-paraphrase/test.tsv)
+
+
+## ESIM
```
DeNeutoy: This is based on https://arxiv.org/pdf/1609.06038.pdf and also uses elmo

MODELS.md
```diff
@@ -0,0 +1,122 @@
...
+
+
+## Machine Comprehension
+
+### [bidaf-model-2017.09.15-charpad.tar.gz](https://s3-us-west-2.amazonaws.com/allennlp/models/bidaf-model-2017.09.15-charpad.tar.gz) (44 MB)
+
+Based on [BiDAF (Seo et al, 2017)](https://www.semanticscholar.org/paper/Bidirectional-Attention-Flow-for-Machine-Comprehen-Seo-Kembhavi/007ab5528b3bd310a80d553cccad4b78dc496b02)
+
+```
+$ docker run allennlp/allennlp:v0.6.1 \
```
DeNeutoy: Maybe remove these docker commands. They are already on the website, aren't available for all of the models, and will get stale really fast.

MODELS.md
DeNeutoy: @nelson-liu can you let Michael know what number goes here?

MODELS.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Event2Mind
+
+Based on [Event2Mind: Commonsense Inference on Events, Intents, and Reactions](https://homes.cs.washington.edu/~msap/debug/event2mind/docs/data/rashkin2018event2mind.pdf)
+More information at: https://homes.cs.washington.edu/~msap/debug/event2mind/docs/
+
+* [event2mind-2018.09.17.tar.gz](https://s3-us-west-2.amazonaws.com/allennlp/models/event2mind-2018.09.17.tar.gz)
+
+
+## BiMPM
```
DeNeutoy: This is based on https://arxiv.org/abs/1702.03814

DeNeutoy: The performance is here from the PR: 
https://github.com/allenai/allennlp/pull/1594#issuecomment-414739696

MODELS.md
```diff
@@ -0,0 +1,122 @@
...
+
+```
+f1: 0.941
+```
+
+## Semantic Parsing
+
+### Wikitables
+
+#### [wikitables-model-2018.09.14.tar.gz](https://s3-us-west-2.amazonaws.com/allennlp/models/wikitables-model-2018.09.14.tar.gz)
```
DeNeutoy: @matt-gardner what is the preliminary performance of this model? Is it worth writing down?

DeNeutoy: Can you add the size for the models below this one too?

MODELS.md
schmmd: They work indefinately because they're versioned.  It provides a record of the last version we tested the model on.  This will be particularly helpful if we include any models from outside AllenNLP.

DeNeutoy: Yes I know that, I meant that it's just duplicating something from somewhere else, meaning it's another thing we have to update when we bump versions.

MODELS.md
matt-gardner: I don't think it's worth writing down.  I'd instead just put major caveats that this is trained on only part of the data and not officially evaluated.  It'd just a placeholder.

MODELS.md
schmmd: Sure, but the same argument applies to having the models here and having performance numbers here.  I do plan to reduce the amount of duplication next quarter.

schmmd: It also gives me a quick "test case" that confirms the URL is accurate and I can actually make predictions with the model.

DeNeutoy: Where else do we have a complete list of available models, sizes and download links for all models? But sure, leave it if you think it's better.

Issue #1713, opened 2018-09-05
possible masking bug in simple_seq2seq


rowanz:
**Describe the bug**
The final hidden states for shorter sequences are set to 0 when using the the simple_seq2seq model, with the configuration at https://github.com/allenai/allennlp/blob/master/allennlp/tests/fixtures/encoder_decoder/simple_seq2seq/experiment_with_attention.json

**To Reproduce**
Steps to reproduce the behavior
1. run the simple_seq2seq model, with a batch where not everything is the same length
2. Right here, https://github.com/allenai/allennlp/blob/f2884ad64dc018bbba9518002a7cb7d15e8ca952/allennlp/models/encoder_decoders/simple_seq2seq.py#L128 the hidden states for short sequences are set to 0

**Expected behavior**
The final_encoder_output needs to be the encoder output at the last valid timestep.

**System (please complete the following information):**
 - OS: Linux
 - Python version: 3.6.6 [if it's not 3.6.1 or later, that's probably your problem]
 - AllenNLP version: 0.6.1
 - PyTorch version: installed with allennlp

**Additional context**
Possibly using a different encoder would fix the problem? or maybe I'm doing something wrong elsewhere

matt-gardner:
Yep, this is a bug.  We've fixed this in models we're actively working on, but we haven't touched this code in a while.  The fix is to use this function instead of the line you quoted: https://github.com/allenai/allennlp/blob/f2884ad64dc018bbba9518002a7cb7d15e8ca952/allennlp/nn/util.py#L121-L123

PRs welcome!

brendan-ai2:
Should be fixed in https://github.com/allenai/allennlp/pull/1808. Please let me know if you still see this issue!

Issue #1134, opened 2018-04-25
Simple Seq2Seq Computes Incorrect Dev Log Loss


vidurj:
When computing the log loss on the dev set, simple_seq2seq does not condition on gold tokens. This means that the loss reported on the dev set is not the negative log likelihood of the gold sequence.

murphp15:
@vidurj 
Can I get more information on this issue, looking at the code I can't see an issue? 
I have looked at SimpleSeq2Seq and as far as I can see it does condition on the gold sequence? 
The method _get_loss takes in the logits and targets and computes the cross entropy. 

What am I missing? I will work on it as soon as I understand the issue.

matt-gardner:
@vidurj, correct me if I'm wrong, but I believe the issue is in this block of code: https://github.com/allenai/allennlp/blob/b71ef434b310e9b1ba400b42c7334c794996b40e/allennlp/models/encoder_decoders/simple_seq2seq.py#L146-L155

This determines what to use as input to the decoder LSTM.  If we're training and not sampling from the model's output, we use the original targets, otherwise we use the model's output from the previous timestep.  What @vidurj wants is an option to always use the top branch during validation (or perhaps calculate that as a separate metric?  that'd be expensive, though, because you'd do two runs at validation time).

murphp15:
Can he not set self._scheduled_sampling_ratio to be 0 during validation?

matt-gardner:
And a new parameter, something like `compute_gold_dev_loss`, that goes in the constructor (and `from_params`).  Just adding another branch would make for spaghetti code - this can be simplified by adding a variable like `use_gold_targets` that gets set to `True` in some cases:

```python
# Not certain this logic is correct as-is; it should be close, though.
use_gold_targets = False
if self.training and all(...):
    use_gold_targets = True
elif timestep == 0:
    use_gold_targets = True  # We're assuming that the data always starts with the start token.
elif not self.training and self._compute_gold_dev_loss:
    use_gold_targets = True

if use_gold_targets:
    input_choices = targets[:, timestep]
else:
    input_choices = last_predictions
```

Though, having that flag passed to the constructor means that after training the model, you have to override the flag to actually make predictions...  @DeNeutoy, do you have any better ideas about how to handle this?

vidurj:
FWIW, I fixed this for myself by conditioning on target_tokens at test time if they are provided i.e.
```
use_gold_targets = False
if self.training and all(...):
    use_gold_targets = True
[...]

PR #1808, opened 2018-09-21
Fixes for seq2seq model

brendan-ai2:
- Addresses:
  - https://github.com/allenai/allennlp/issues/1713, "possible masking bug in simple_seq2seq"
  - https://github.com/allenai/allennlp/issues/1134, "Simple Seq2Seq Computes Incorrect Dev Log Loss"

### DIFF
--- a/allennlp/tests/models/encoder_decoders/simple_seq2seq_test.py
+++ b/allennlp/tests/models/encoder_decoders/simple_seq2seq_test.py
@@ -8,7 +8,7 @@ from allennlp.nn.util import sequence_cross_entropy_with_logits
 
 class SimpleSeq2SeqWithoutAttentionTest(ModelTestCase):
     def setUp(self):
-        super(SimpleSeq2SeqWithoutAttentionTest, self).setUp()
+        super().setUp()
         self.set_up_model(self.FIXTURES_ROOT / "encoder_decoder" / "simple_seq2seq" / "experiment.json",
                           self.FIXTURES_ROOT / "data" / "seq2seq_copy.tsv")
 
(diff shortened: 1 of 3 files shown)


brendan-ai2:
(Needs an extra test before review.)

brendan-ai2:
Thanks!

I added a simple test for an issue that using `get_final_encoder_states` also solves (bidirectional encoders) and manually tested that the validation loss was improving by implementing a trivial autoencoder. Do we have any established patterns for testing that we don't regress training full models? Feels more like an integration test.

Let me know if it's worth having a test specific to the masking issue.

allennlp/tests/models/encoder_decoders/simple_seq2seq_test.py
```diff
@@ -38,9 +38,21 @@ def test_decode_runs_correctly(self):
...
+class SimpleSeq2SeqWithoutAttentionTest_Bidirectional(ModelTestCase):
+    def setUp(self):
+        super().setUp()
+        self.set_up_model(self.FIXTURES_ROOT / "encoder_decoder" / "simple_seq2seq" / "experiment.json",
+                          self.FIXTURES_ROOT / "data" / "seq2seq_copy.tsv")
+
+    def test_encoder_decoder_can_train_save_and_load(self):
+        self.ensure_model_can_train_save_and_load(
+                self.param_file,
+                overrides="{model: {encoder: {bidirectional: true}}}"
```
matt-gardner: Why not just change this in `experiment.json`?

brendan-ai2: I'd prefer to test with bidirectional set to both true and false. And, naturally, duplicating the config file wasn't very compelling.

matt-gardner: Oh, ok.  It looks like you can just add this as another method on the other class, too, instead of making a whole new class for this.

brendan-ai2:
(No worries!)

Thanks for the background on the tests. I filed https://github.com/allenai/allennlp/issues/1813 so we can consider the periodic retraining.

PR #1812, opened 2018-09-23
Adding QuaRel dataset and parser

matt-gardner:


allennlp/models/semantic_parsing/quarel/quarel_semantic_parser.py
```diff
@@ -0,0 +1,890 @@
...
+        # These encode the same information, but for efficiency reasons later it's nice
+        # to have one version as a tensor and one that's accessible on the cpu.
+        entity_types, entity_type_dict = self._get_type_vector(world, num_entities, embedded_table)
+
+        if self._use_entities:
+            if self._entity_encoder is not None:
+                # (batch_size, num_entities, embedding_dim)
+                encoded_table = self._entity_encoder(embedded_table, table_mask)
+                # (batch_size, num_entities, num_neighbors)
+                neighbor_indices = self._get_neighbor_indices(world, num_entities, encoded_table)
```
matt-gardner: Are you actually using neighbors at all in this model?  What for?

OyvindTafjord: No, the neighbors are always empty, so this will not do anything interesting. In fact I don't use any entity_encoder in the current models, so these more mostly artifacts left over from starting with the WikiTables code.

matt-gardner: I just started removing some of this old stuff, but I don't have a test to run to make sure I'm not breaking anything, and I can't easily add one because I don't have any test fixtures to use.

OyvindTafjord: I added some very basic tests now, but they're so far only verifying that the model trains and saves.

matt-gardner:
Closing this so that @OyvindTafjord can open it himself, which will make things easier.

PR #1815, opened 2018-09-25
SQL Executor

kl2806:
- Adds an executor for SQL using SQLite
- Removes code in ATIS semantic parser that is related to executing queries


kl2806:
Thanks @matt-gardner! Also marking the WikiTables ERM test as flaky, since it had a connection error in one of the builds.

PR #1819, opened 2018-09-25
Move vocabulary save to after model setup

OyvindTafjord:
@DeNeutoy I forgot to include this in https://github.com/allenai/allennlp/pull/1817. This is a bit ugly as it's during the model initialization that the vocabulary pulls in the extra embeddings. Maybe there's a better way to do this, it might mess up other places as well. FYI @matt-gardner

### DIFF
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -273,9 +273,11 @@ def train_model(params: Params,
              if key in datasets_for_vocab_creation)
     )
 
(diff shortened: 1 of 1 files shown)


OyvindTafjord:
I agree, that seems better.

DeNeutoy:
@OyvindTafjord do you mind submitting a new PR which reverts these changes and adds the functionality to the vocab instead?

PR #1821, opened 2018-09-26
Add more configuration options for ATIS semantic parser

kl2806:
- Adds a flag `keep_if_unparseable` to the ATIS dataset reader so that during validation, we don't skip the instances that we aren't able to parse
- Adds some other improvements to the entity heuristics: generate numbers for patterns in the text like "I want a flight for 100 dollars" and strings eg. adding triggers for "saint petersburg" in addition to "st. petersburg"
- Adds an option to concatenate previous n utterances to incorporate recent conversation history
- Modifies the ``SqlVisitor`` so that it visits nonterminals from left to right. This potentially makes it easier for the model, since it doesn't need to keep as much information in history.


allennlp/data/dataset_readers/semantic_parsing/atis.py
```diff
@@ -66,15 +69,17 @@ class AtisDatasetReader(DatasetReader):
     """
     def __init__(self,
                  token_indexers: Dict[str, TokenIndexer] = None,
+                 keep_if_unparseable: bool = False,
                  lazy: bool = False,
                  tokenizer: Tokenizer = None,
-                 database_file: str = None) -> None:
+                 database_file: str = None,
+                 num_turns_to_concatenate: int = 1) -> None:
```
DeNeutoy: add to docstring.

allennlp/semparse/contexts/sql_context_utils.py
```diff
@@ -114,7 +117,7 @@ def action_sequence_to_sql(action_sequences: List[str]) -> str:
         if nonterminal == 'statement':
             query.extend(right_hand_side_tokens)
         else:
-            for query_index, token in reversed(list(enumerate(query))):
```
DeNeutoy: We don't need this reversed here now because we traverse them in the reversed order when we generate the action sequence, is that right?

kl2806: Yep, exactly

PR #1829, opened 2018-09-27
Simplified GrammarStatelet, made a new LambdaGrammarStatelet class for WikiTables

matt-gardner:
In writing a tutorial for the framework, I decided that we really don't need to worry about lambdas in the main `GrammarStatelet` class.  So I made a new one that handled lambdas, and greatly simplified the main one.  @OyvindTafjord @rajasagashe, this just means that when you create your initial state, you don't need to pass in empty objects for the `lambda_stacks` and `context_actions` arguments, because those arguments are now gone.  Nothing else in your code should have to change.


allennlp/state_machines/states/lambda_grammar_statelet.py
```diff
@@ -0,0 +1,182 @@
+from typing import Callable, Dict, List, Tuple
+
+import torch
+
+from allennlp.nn import util
+
+
+# We're not actually inhereting from `GrammarStatelet` here because there's very little logic that
```
matt-gardner: This class is basically unmodified from the previous `GrammarStatelet`.  The only things that are different are the docstrings and this comment.

PR #1831, opened 2018-09-27
make some sql context code more generic

DeNeutoy:
- Isolate the current SQL grammar to specifically the Atis dataset
- Move generic sql utils into `sql_table_context_utils.py`
- Rename `SqlTableContext` to `AtisSqlTableContext`

### DIFF
--- a/allennlp/semparse/contexts/atis_sql_table_context.py
+++ b/allennlp/semparse/contexts/atis_sql_table_context.py
@@ -1,20 +1,18 @@
 """
-A ``SqlTableContext`` represents the SQL context in which an utterance appears, with the grammar and
-the valid actions.
+An ``AtisSqlTableContext`` represents the SQL context in which an utterance appears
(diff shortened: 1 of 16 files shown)


allennlp/semparse/contexts/atis_sql_table_context.py
```diff
@@ -162,18 +123,19 @@ def initialize_valid_actions(self) -> Dict[str, List[str]]:
             # Sequence represents a series of expressions that match pieces of the text in order.
             # Eg. A -> B C
             if isinstance(rhs, Sequence):
-                valid_actions[key].add(format_action(key, " ".join(rhs._unicode_members()))) # pylint: disable=protected-access
+                valid_actions[key].add(format_action(key, " ".join(rhs._unicode_members()), # pylint: disable=protected-access
```
kl2806: We can potentially move this function to somewhere more generic like `sql_context_utils` since isn't doing dataset specific; not necessarily in this PR.

PR #1832, opened 2018-09-27
Integrate new table context in variable free world

pdasigi:
Summary of changes:
1) Added column types to type declaration
2) Defined a `MultiMatchNamedBasicType` that matches with multiple types
3) Made minor changes to  @MurtyShikhar's code for `TableQuestionContext` to make it work with `WikiTablesVariableFreeWorld`
4) Made the `WikiTablesVariableFreeWorld` use `TableQuestionContext`

Depends on #1762 and #1777 

@matt-gardner, regarding 2 above, I tried the simpler fix of adding the same column with multiple types to the local name mapping first. It did not work because the second addition to the mapping would overwrite the first one, and even if we somehow stored both, the logic for parsing logical forms would have to be changed to to use the appropriate type of a the column name depending on context. I ended up making a hierarchical type, but called it `MultiMatchNamedBasicType` (which seemed more appropriate), and it was pretty straightforward.


MurtyShikhar:
Saw relevant parts from table_question_context.py. Looks good to me!

pdasigi:
@matt-gardner This PR is ready to to be reviewed. It's smaller than it looks since a large chunk of the additions in this PR are from the test table files from #1777.

allennlp/semparse/type_declarations/wikitables_variable_free.py
```diff
@@ -14,43 +15,49 @@
...
 # Type for filtering rows given a column. "argmax", "argmin" and "same_as" (select all rows with the
-# same value under the given column as the given row does under the given column)
+# same value under the given column as the given row does under the given column). While "same_as"
+# takes any column, "argmax" and "argmin" take only comparable columns (i.e. dates or numbers).
 # Note that the values used for comparison in "argmax" and "argmin" can only come from column
 # lookups in this language. In LambdaDCS, there's a lambda function that is applied to the rows to
 # get the values, but here, we simply have a column name.
-ROW_FILTER_WITH_COLUMN = ComplexType(ROW_TYPE, ComplexType(COLUMN_TYPE, ROW_TYPE))
+ROW_FILTER_WITH_GENERIC_COLUMN = ComplexType(ROW_TYPE, ComplexType(GENERIC_COLUMN_TYPE, ROW_TYPE))
+ROW_FILTER_WITH_COMPARABLE_COLUMN = ComplexType(ROW_TYPE, ComplexType(COMPARABLE_COLUMN_TYPE, ROW_TYPE))
```
matt-gardner: It might be better to handle this function like we handle `UnaryOpTypes`, so that we don't have to add another type to account for this.  I guess this is the simpler way to go - you just need to add number and date columns to three types this way, instead of two - and given that we're thinking of building type inference on top of python, anyway, it's probably not worth the effort to fix up this type system.

allennlp/semparse/worlds/wikitables_variable_free_world.py
```diff
@@ -101,41 +106,46 @@ def _map_name(self, name: str, keep_mapping: bool = False) -> str:
         if name not in types.COMMON_NAME_MAPPING and name not in self.local_name_mapping:
             if not keep_mapping:
```
matt-gardner: I'd recommend moving out the "add name to mapping" functionality from this method.  Basically, just take this whole block under `if name not in types.COMMON_NAME_MAPPING ...` and move it into a separate function (removing the `keep_mapping` argument).  Then `_map_name` can just do the `else` block below.

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,306 @@
...
+        column_index_to_name = {}
+
+        header = lines[0] # the first line is the header
+        index = 1
+        while lines[index][0] == '-1':
+            # column names start with fb:row.row.
+            curr_line = lines[index]
+            column_name_sempre = curr_line[2]
+            column_index = int(curr_line[1])
+            column_name = column_name_sempre.replace('fb:row.row.', '')
```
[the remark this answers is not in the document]
matt-gardner: Oh, it's in the data itself?  That's what Chen did?  Hmm, ok.

allennlp/tests/semparse/contexts/table_question_context_test.py
```diff
@@ -0,0 +1,103 @@
...
+        self.tokenizer = WordTokenizer(SpacyWordSplitter(pos_tags=True))
+
+    def test_number_extraction(self):
+        question = """how many players on the 191617 illinois fighting illini men's basketball team
+                      had more than 100 points scored?"""
+        question_tokens = self.tokenizer.tokenize(question)
+        test_file = f'{self.FIXTURES_ROOT}/data/corenlp_processed_tables/TEST-7.table'
+        table_question_context = TableQuestionContext.read_from_file(test_file, question_tokens)
+        _, number_entities = table_question_context.get_entities_from_question()
+        assert number_entities == [("191617", "191617"), ("100", "100")]
```
matt-gardner: Seems more useful to have a token index, not a string, for what triggered the number.  There might be more than one matching string.  Also, I know we talked at one point about separating out date numbers from non-date numbers.  Not sure if that's a good idea or not...  I suppose we can worry about it later, in any event.

allennlp/tests/semparse/contexts/table_question_context_test.py
```diff
@@ -0,0 +1,103 @@
...
+        assert number_entities == [("2", "february"), ("26", "26"), ("2006", "2006")]
+
+    def test_date_extraction_2(self):
+        question = """how many different players scored for the san jose earthquakes during their
+                      1979 home opener against the timbers?"""
+        question_tokens = self.tokenizer.tokenize(question)
+        test_file = f'{self.FIXTURES_ROOT}/data/corenlp_processed_tables/TEST-6.table'
+        table_question_context = TableQuestionContext.read_from_file(test_file, question_tokens)
+        _, number_entities = table_question_context.get_entities_from_question()
+        assert number_entities[0] == ("1979", "1979")
```
matt-gardner: Better to assert equality on the list - it's still possible with this test that `number_entities` has length > 1.  Same with another test below.

allennlp/tests/semparse/contexts/table_question_context_test.py
```diff
@@ -0,0 +1,103 @@
...
+    def test_rank_number_extraction(self):
+        question = "what was the first tamil-language film in 1943?"
+        question_tokens = self.tokenizer.tokenize(question)
+        test_file = f'{self.FIXTURES_ROOT}/data/corenlp_processed_tables/TEST-1.table'
+        table_question_context = TableQuestionContext.read_from_file(test_file, question_tokens)
+        _, numbers = table_question_context.get_entities_from_question()
+        assert numbers[0] == ("1", "first")
+
+    def test_null_extraction(self):
+        question = "on what date did the eagles score the least points?"
```
matt-gardner: "eagles" doesn't appear because it's the title of the wikipedia page, or something?  A note about that would be helpful here, as this test is confusing without it.

allennlp/tests/semparse/worlds/wikitables_variable_free_world_test.py
```diff
@@ -82,128 +94,113 @@ def test_get_valid_actions_returns_correct_set(self):
...
+
+        check_productions_match(valid_actions['f'],
+                                ['number_column:avg_attendance',
+                                 'number_column:division'])
+
+        check_productions_match(valid_actions['t'],
+                                ['string_column:league',
+                                 'string_column:playoffs',
+                                 'string_column:open_cup',
+                                 'string_column:regular_season'])
```
matt-gardner: What about `g` and `c`?  Where are their productions?  You're going to need them, because you have things like `r -> [<r,<g,r>>, r, g]` (so `g` shows up as a nonterminal).

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,306 @@
...
+    for extracting entities from a question given a table and type its columns with <string> | <date> | <number>
+    """
+    # TODO(shikhar/pradeep): Fix variable names
+    def __init__(self,
+                 cell_values: Set[str],
+                 column_type_statistics: List[Dict[str, int]],
+                 column_index_to_name: Dict[int, str],
+                 question_tokens: List[Token]) -> None:
+        self.cell_values = cell_values
+        self.column_types = {column_index_to_name[column_index]: max(column_type_statistics[column_index],
```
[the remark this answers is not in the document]
pdasigi: I fixed the first issue in the next (yet to be created) PR, and made a couple of other changes here. I agree that we should default to strings if columns are not all numbers or dates. Will change that in the next PR too.

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -0,0 +1,306 @@
...
+        for cell_value in self.cell_values:
+            if candidate in cell_value:
+                return True
+        return False
+
+    def _process_conjunction(self, entity_data):
+        raise NotImplementedError
+
+    def _expand_entities(self, question, entity_data):
+        new_ents = []
```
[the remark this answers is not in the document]
pdasigi: Done.

allennlp/semparse/worlds/wikitables_variable_free_world.py
pdasigi: Made some changes in this method for the next PR. Will make a separate method there.

allennlp/tests/semparse/contexts/table_question_context_test.py
pdasigi: Changed token strings to indices.

allennlp/tests/semparse/contexts/table_question_context_test.py
pdasigi: Done.

allennlp/tests/semparse/contexts/table_question_context_test.py
pdasigi: Done.

allennlp/tests/semparse/worlds/wikitables_variable_free_world_test.py
pdasigi: As I told you about this earlier today, they do not have productions of their own. There is still the issue that they show up as nonterminals while searching in `ActionSpaceWalker`, and converting action sequences to logical forms in `World`. I dealt with `MultiMatchNamedBasicTypes` separately in those classes (in the upcoming PR).

matt-gardner: It's also an issue in the model itself.  If they show up in the `ActionSpaceWalker`, they will show up in the model.  This doesn't work with how our models are structured, and it needs to change.

pdasigi: How about we discuss it when we see the issue with the model? I wouldn't mind changing the way I implemented the functionality here if I have to then. The simpler alternative of adding the same name to the local mapping with multiple types would have been messy for the reason I stated above.

allennlp/semparse/contexts/table_question_context.py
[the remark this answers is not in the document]
MurtyShikhar: > Also, do we really want a max here? I would have thought that if it's not all numbers or dates, we should default to string.

I think this is what the code does currently (except that it also checks whether the context of the cell contains a hyphen, something that chen also had in his code).

allennlp/tests/semparse/worlds/wikitables_variable_free_world_test.py
matt-gardner: Fine, we can change it later, but something's going to have to change, as we need to be able to consider all column types in a single timestep.

PR #1834, opened 2018-09-27
initial text2sql grammar with unconstrained context

DeNeutoy:


allennlp/semparse/contexts/text2sql_table_context.py
```diff
@@ -0,0 +1,142 @@
...
+GRAMMAR_DICTIONARY["parenval"] = ['"(" ws expr ws ")"']
+GRAMMAR_DICTIONARY["function"] = ['(fname ws "(" ws "DISTINCT" ws arg_list_or_star ws ")")',
+                                  '(fname ws "(" ws arg_list_or_star ws ")")']
+
+GRAMMAR_DICTIONARY["arg_list_or_star"] = ['arg_list', '"*"']
+GRAMMAR_DICTIONARY["arg_list"] = ['(expr ws "," ws arg_list)', 'expr']
+GRAMMAR_DICTIONARY["number"] = ['~"\d*\.?\d+"i']
+GRAMMAR_DICTIONARY["string_set"] = ['ws "(" ws string_set_vals ws ")"']
+GRAMMAR_DICTIONARY["string_set_vals"] = ['(string ws "," ws string_set_vals)', 'string']
+GRAMMAR_DICTIONARY["string"] = ['~"\'.*?\'"i']
```
kl2806: We will have to update this not to use regexes to actually get the SQL queries because right now this will give us an action that looks like `string -> ['~"\'.*?\'"i']` that represents a variable. If there are multiple variables then we won't be able to match them to the correct annotated values. For example, we will get `string -> ['~"\'.*?\'"i']` in the action sequence twice for the query below and we won't know which is the actor and which is the director.

```
        "sql": [
            "SELECT COUNT( DISTINCT ( ACTORalias0.NAME ) ) FROM ACTOR AS ACTORalias0 , CAST AS CASTalias0 , DIRECTED_BY AS DIRECTED_BYalias0 , DIRECTOR AS DIRECTORalias0 , MOVIE AS MOVIEalias0 WHERE ACTORalias0.NATIONALITY = \"actor_nationality0\" AND CASTalias0.AID = ACTORalias0.AID AND DIRECTORalias0.DID = DIRECTED_BYalias0.DID AND DIRECTORalias0.NAME = \"director_name0\" AND MOVIEalias0.MID = CASTalias0.MSID AND MOVIEalias0.MID = DIRECTED_BYalias0.MSID ;"
        ],
        "variables": [
            {
                "example": "Iran",
                "location": "both",
                "name": "actor_nationality0",
                "type": "actor_nationality"
            },
            {
                "example": "Jim Jarmusch",
                "location": "both",
                "name": "director_name0",
                "type": "director_name"
            }
        ]
```
I think what we want is `GRAMMAR_DICTIONARY["string"] = [\"director_name0\",  "actor_nationality0"]`, which we can just get from the name field, so when we postprocess the action sequence we can put director name and action nationality in the right slots.

kl2806: I think the same thing applies for the `number` nonterminal in the grammar.

DeNeutoy: Oh yep good point - i'll do that in a separate PR.

Issue #1809, opened 2018-09-21
Feature request: some mechanism for global data information to be passed to the model


matt-gardner:
In the semantic parsing code, there are things about the data that you need to read in the dataset reader and pass to the model, but only need to be read once.  This includes, for example, the set of actions in the global grammar.  The global grammar depends on your dataset, but not on each instance.  Currently, the only way to get global information about the data from the `DatasetReader` to the `Model` is through the vocabulary, and the only way to influence the `Vocabulary` is through individual instances.  There's no mechanism to get, e.g., a global grammar that we read from a database or a text file once, and pass it to the model.  Instead, we have to either do something hacky with a `Field` that we just use for vocab creation and somehow make it not produce any tensors, or we have to duplicate the work of creating a bunch of tensors that we really only need one copy of.

I can imagine some ways to build this into the library, but everything feels hacky to me.  This is one of those cases where you really want to break outside of the abstractions...  Like, you could add `DatasetReader.get_global_model_inputs()`, or something, which returns a dictionary that gets passed as additional kwargs to the `Model` [here](https://github.com/allenai/allennlp/blob/606a61abf04e3108949022ae1bcea975b2adb560/allennlp/commands/train.py#L278).  But I'm concerned about adding so many different interaction points between the abstractions that the library just gets unmanageable.

brendan-ai2:
@acohan, this is a feature that would address the idf calculation. We don't have a great solution yet, but please feel free to track this and offer suggestions.

acohan:
This would be very helpful to have.

joelgrus:
one way to do this would be the following variation on mattg's suggestion:

(1) give `DatasetReader` an `extras() -> Dict[str, Any]` method whose default implementation returns an empty dict
(2) allow `DataIterator` to take an optional `extras: Dict[str, Any]` parameter
(3) after `DataIterator` calls `batch.as_tensor_dict`, it adds all the "extras" to the tensor dict

another alternative is just to insist that people who do this are "breaking out of the library", so that e.g. they have to stick their global variable as a static property on their DatasetReader, and then the model has to take the registered name of the DatasetReader as a constructor param and then do something like

```
self.global_grammar = getattr(DatasetReader.by_name(name), 'global_grammar', None)
```

matt-gardner:
[...]

PR #1845, opened 2018-10-01
Structured sql data

DeNeutoy:
This PR does a few things:

- Strips out registerable `SqlTableContexts`, they weren't a useful abstraction because i'm going to implement different constraints as functions which mutate the grammar, and just pass a list of functions to apply in. They don't have enough differently implemented functionality to make this useful, and it was confusing to have different contexts not be usable in different ... contexts (e.g you couldn't use the `AtisContext` in a `Text2SqlDatasetReader`, but according to the class hierarchy, you could).
- Adds a `GrammarBasedText2Sql` dataset reader.
- Adds a pre-processing step to the Text2Sql data to replace column references which use `ID` (sql syntactic sugar for the primary key of a table) with the actual column name of the primary key (this was so annoying haha)
- Adds `wsp` to non-terminals which are ignored when producing action sequences
- Adds a very basic `Text2SqlWorld` which currently just delegates all of its work to the `Text2SqlContext`, but I will add in linking etc in future PRs.
- Documents a few assumptions we were making about `ws` and `wsp` nodes in grammars which wasn't super obvious.

Note that to make this PR smaller, some of the tests are just dummy shells which just test that things run - this is because it is required that we add in strings to the grammar from the tables, but I want to do that in a separate PR because it will be complicated.


allennlp/data/dataset_readers/dataset_utils/text2sql_utils.py
```diff
@@ -78,6 +82,25 @@ def clean_and_split_sql(sql: str) -> List[str]:
...
 
+def resolve_primary_keys_in_schema(sql_tokens: List[str],
+                                   schema: Dict[str, List[TableColumn]]) -> List[str]:
+    """
+    Some examples in the text2sql datasets use ID as a column reference to the
+    column of a table which has a primary key. This causes problems if you are trying
+    to constrain a grammar to only produce the column names directly, because you don't
+    know what ID refers to. So instead of dealing with that, we just replace it.
+    """
+    primary_keys_for_tables = {name: max(columns, key=lambda x: x.is_primary_key).name
```
matt-gardner: Don't you want some `if` in here?  What happens if no column is a primary key?

allennlp/data/dataset_readers/dataset_utils/text2sql_utils.py
```diff
@@ -78,6 +82,25 @@ def clean_and_split_sql(sql: str) -> List[str]:
             sql_tokens.extend(split_table_and_column_names(token))
     return sql_tokens
 
+def resolve_primary_keys_in_schema(sql_tokens: List[str],
+                                   schema: Dict[str, List[TableColumn]]) -> List[str]:
+    """
+    Some examples in the text2sql datasets use ID as a column reference to the
```
matt-gardner: What's wrong with just adding `ID` as another column that can be generated?

allennlp/data/dataset_readers/semantic_parsing/grammar_based_text2sql.py
```diff
@@ -0,0 +1,140 @@
...
+                data = json.load(data_file)
+
+            for sql_data in text2sql_utils.process_sql_data(data,
+                                                            use_all_sql=self._use_all_sql,
+                                                            remove_unneeded_aliases=self._remove_unneeded_aliases,
+                                                            # TODO(Mark): Horrible hack, remove
+                                                            schema=self._sql_table_context.schema):
+                instance = self.text_to_instance(sql_data.text, sql_data)
+                if instance is not None:
+                    yield instance
```
matt-gardner: You probably need some flag somewhere to keep validation instances that aren't parseable, so we don't get artificially-inflated metrics.

allennlp/data/dataset_readers/semantic_parsing/grammar_based_text2sql.py
```diff
@@ -0,0 +1,140 @@
...
+            except ParseError:
+                return None
+
+        index_fields: List[Field] = []
+        production_rule_fields: List[Field] = []
+
+        for production_rule in all_actions:
+            nonterminal, _ = production_rule.split(' ->')
+            # The whitespaces are not semantically meaningful, so we filter them out.
+            production_rule = ' '.join([token for token in production_rule.split(' ') if token != 'ws'])
```
matt-gardner: Do you need to worry about `wsp` here, too?

allennlp/data/dataset_readers/semantic_parsing/grammar_based_text2sql.py
```diff
@@ -0,0 +1,140 @@
...
+                         query: List[str],
+                         sql: text2sql_utils.SqlData = None) -> Instance:
+        # pylint: disable=arguments-differ
+        fields: Dict[str, Field] = {}
+        tokens = TextField([Token(t) for t in query], self._token_indexers)
+        fields["tokens"] = tokens
+
+        if sql is not None:
+            try:
+                action_sequence, all_actions = self._world.get_action_sequence_and_all_actions(sql.sql)
```
matt-gardner: Not for this PR necessarily, but `all_actions` here is constant across all instances, yet we're constructing the object for every instance, and making the same `ProductionRuleField` list for them.  It'd be nice to figure out a way around this at some point...  (See #1809)

allennlp/data/dataset_readers/dataset_utils/text2sql_utils.py
DeNeutoy: Because it's essentially a floating reference to any column and I think it would be hard for a model to learn that. Also, I have a strong bias to as few actions as possible, and this ID is not really necessary.

allennlp/data/dataset_readers/dataset_utils/text2sql_utils.py
DeNeutoy: If a column doesn't have a primary key, there will be no ID reference to it, so it's fine because the max will just take an arbitrary one.

allennlp/data/dataset_readers/semantic_parsing/grammar_based_text2sql.py
DeNeutoy: Yep - I will do that later, it should be easy to modify. We cover enough of the datasets completely that I can use one which we do cover 100% of initially. Thanks.

allennlp/data/dataset_readers/semantic_parsing/grammar_based_text2sql.py
DeNeutoy: Umm I think the production rule generation strips these out actually, so it never occurs. Let me double check that.

allennlp/data/dataset_readers/semantic_parsing/grammar_based_text2sql.py
DeNeutoy: I'm not sure it actually will be the same across instances - e.g consider the case that we are given anonymized data with entity information - then we want to constrain the string productions to be only the entities with the correct type, rather than any of them. But I see what you mean I think it's also a slow part of the code, but I think it's unlikely to matter too much for the text2sql datasets as they are all reasonably sized.

Issue #1848, opened 2018-10-01
Verify script fails if Java is not installed


seal6363:
Because some of the unit tests require Java, 
But the instruction in readme doesn't mention Java.

nelson-liu:
@schmmd the same issue should exist with `perl`. maybe worth just having `verify.py` use `allennlp test-install`?

nelson-liu:
actually that probably isn't a great idea, since the `verify.py` script is for development use and we'd actually like to run all the tests (including those with dependencies we don't usually mention).

joelgrus:
I also think that we should have a `check_for_java` helper function that raises a `RuntimeError` if Java is not installed, and that we should call it before we subprocess out the Java call from Python

schmmd:
That's a good point Joel--but it seems like a separate issue: https://github.com/allenai/allennlp/issues/1850

PR #1849, opened 2018-10-01
Use test-install rather than verify.py

schmmd:
Fixes https://github.com/allenai/allennlp/issues/1848


README.md
```diff
@@ -65,7 +65,7 @@ just run the following command to get an environment that will run on either the
    docker run -it -p 8000:8000 --rm allennlp/allennlp:v0.6.1
    ```
 
-You can now test your installation with `./scripts/verify.py`.
+You can now test your installation with `test-install`.
```
nelson-liu: `s/test-install/allennlp test-install/`?

schmmd: In docker you `docker run -it -p 8000:8000 --rm allennlp/allennlp:v0.6.1 test-install` rather than `docker run -it -p 8000:8000 --rm allennlp/allennlp:v0.6.1 allennlp test-install` (which won't work).

This is a bit confusing however.  Do you think I should just include the entire docker command?

nelson-liu: ah right, sorry --- yeah, I think the entire docker command would be better here.

PR #1854, opened 2018-10-02
Remove wget in wikitables tests by using requests

### LINKED ISSUE #1851: WikiTables code requires `wget` to be installed, fails in an unhelpful way if it's not
and e.g. a clean OSX install might not have `wget`

ideal solution: replace the calls to `wget` with the equivalent code using `requests`

seal6363:
closes #1851


joelgrus:
this looks really good. one nit: I don't know that it makes sense to call the variable `temp_file` since these aren't really temp files, are they? maybe change it to `downloaded_file` or something?

PR #1876, opened 2018-10-06
strings in sql/prelinked entities

DeNeutoy:
- Makes `ProductionRuleFields` printable
- Removes SqlTableContext, it had no value - moved everything into `World`
- Added a method to link entities directly
- Added a test for the reader


allennlp/data/dataset_readers/dataset_utils/text2sql_utils.py
```diff
@@ -43,6 +43,26 @@ class TableColumn(NamedTuple):
     column_type: str
     is_primary_key: bool
 
+def column_has_string_type(column: TableColumn) -> bool:
+
```
kl2806: Remove empty line?

allennlp/data/dataset_readers/semantic_parsing/grammar_based_text2sql.py
```diff
@@ -48,19 +54,31 @@ class GrammarBasedText2SqlDatasetReader(DatasetReader):
     """
     def __init__(self,
                  schema_path: str,
+                 database_file: str,
                  use_all_sql: bool = False,
                  remove_unneeded_aliases: bool = True,
+                 use_prelinked_entities: bool = True,
```
kl2806: I think most of the literature refers to this as "entity anonymization" so we might also want to name it that or "variable anonymization" or something like that for consistency.

allennlp/semparse/worlds/text2sql_world.py
```diff
@@ -1,31 +1,58 @@
...
         grammar_with_context = deepcopy(self.base_grammar_dictionary)
+
+        if not self.use_prelinked_entities and prelinked_entities is not None:
+            raise ConfigurationError("The Text2SqlWorld was specified to not use prelinked "
+                                     "entities, but prelinked entities were passed.")
+        prelinked_entities = prelinked_entities or {}
+        for token in prelinked_entities.keys():
+            grammar_with_context["value"] = [f'"\'{token}\'"'] + grammar_with_context["value"]
+
         grammar = Grammar(format_grammar_string(grammar_with_context))
```
kl2806: Generating ``Grammar`` objects was super slow for contextual ATIS because we had to generate concrete values (a production rule often had all the values in a column as the right hand side) so to make it faster we only make one call to the ``Grammar`` constructor and edit that grammar object. Probably not an issue here since anonymizing the variables makes the grammar much smaller, but we might want to change this later if we want to add an option for not doing variable anonymization.

allennlp/semparse/worlds/text2sql_world.py
```diff
@@ -38,6 +65,31 @@ def get_action_sequence_and_all_actions(self, query: List[str]) -> Tuple[List[st
...
+            else:
+                # TODO(Mark): The grammar can be tightened here if we don't need to
+                # produce concrete values.
+                pass
+
+        return grammar_dictionary
+
+    def is_global_rule(self, production_rule: str) -> bool:
+        if self.use_prelinked_entities:
+            if "value" in production_rule and production_rule[-4].isnumeric():
```
kl2806: I think it would be helpful to add a comment here: we are checking `-4` so that it is not a global rule if we see the `0` in the a rule like ``'value -> ["\'city_name0\'"]'``, right?

PR #1881, opened 2018-10-09
Tentative port of `LMDatatsetReader`

brendan-ai2:
- Based off of https://github.com/allenai/calypso/blob/master/calypso/data.py.
- Principal differences:
  - Relies on standard AllenNLP batching, iterators and indexing.
  - No sharding or looping indefinitely.
  - Does not port `AsyncLMDataset` in anticipation of using `MultiprocessDatasetReader`.


brendan-ai2:
@joelgrus, here is the current `LMDatasetReader`. I started out with a straight port of `calypso/data.py`, but kept pruning assuming that the vanilla batching, iterators and indexers would be sufficient. At this point the dataset reader is almost superficial. Frankly, it feels a bit ridiculous in isolation. I may have to bring back some of the customizations from Calypso, but which and how is unclear until we integrate this.

I still need to add types and improve the comments from Calypso, but I wanted to sanity check with you first: Should this really be so simple?

brendan-ai2:
Further simplified, comments and types added, etc. Please take a look.

brendan-ai2:
Thanks, comments addressed. (I kept around the extra shards in the test fixtures as I imagine they're going to be used in the actual Elmo model tests.)

allennlp/data/dataset_readers/simple_language_modeling.py
```diff
@@ -0,0 +1,73 @@
...
+                 token_indexers: Dict[str, TokenIndexer] = None,
+                 max_sequence_length: int = None) -> None:
+        # Always lazy to handle looping indefinitely.
+        super().__init__(True)
+        self._tokenizer = tokenizer or WordTokenizer()
+        self._token_indexers = token_indexers or {"tokens": SingleIdTokenIndexer()}
+        self._max_sequence_length = max_sequence_length
+
+        print("Creating SimpleLanguageModelingDatasetReader")
+        print("max_sequence_length={}".format(max_sequence_length))
```
joelgrus: can we change these to `logger.info` calls

allennlp/data/dataset_readers/simple_language_modeling.py
```diff
@@ -0,0 +1,73 @@
...
+        if self._max_sequence_length is not None:
+            sentences_raw = [
+                    sentence for sentence in all_sentences_raw
+                    if len(self._tokenizer.tokenize(sentence)) <= self._max_sequence_length + 2
+            ]
+        else:
+            sentences_raw = all_sentences_raw
+
+        for sentence in sentences_raw:
+            yield self.text_to_instance(sentence)
```
joelgrus: if you're just going to yield the sentences one at a time, it's probably better / faster to do that while reading?

allennlp/data/dataset_readers/simple_language_modeling.py
```diff
@@ -0,0 +1,73 @@
...
+        print('Loading data from {}'.format(file_path))
+
+        with open(file_path) as file:
+            all_sentences_raw = file.readlines()
+
+        # Remove sentences longer than the maximum.
+        if self._max_sequence_length is not None:
+            sentences_raw = [
+                    sentence for sentence in all_sentences_raw
+                    if len(self._tokenizer.tokenize(sentence)) <= self._max_sequence_length + 2
```
joelgrus: tokenization is expensive, and you're doing it twice here.

I would probably do something like

```
for line in file:
    tokens = tokenize(line)
    if self._max_sequence_length is None or len(tokens) < self._max_sequence_length + 2:
        yield Instance(fields={ ... })
```

it duplicates the logic inside text_to_instance, but that's probably ok

or as an alternative you could do

```
for line in f:
    instance = self.text_to_instance(line)
    if len(instance.fields["source"].tokens) <= ... :
        yield instance
```

maybe that's better?

allennlp/data/dataset_readers/simple_language_modeling.py
brendan-ai2: Done

allennlp/data/dataset_readers/simple_language_modeling.py
brendan-ai2: Good call, fixed. I went with your second suggestion to avoid the duplication.

allennlp/data/dataset_readers/simple_language_modeling.py
brendan-ai2: Done. Sorry about that. Should have cleaned up that relic from the old code sooner.

brendan-ai2:
Comments addressed, please take a look.

allennlp/data/dataset_readers/simple_language_modeling.py
```diff
@@ -0,0 +1,70 @@
...
+        # pylint: disable=arguments-differ
+        logger.info('Loading data from %s', file_path)
+
+        with open(file_path) as file:
+            for sentence in file:
+                instance = self.text_to_instance(sentence)
+                # Remove sentences longer than the maximum.
+                if self._max_sequence_length is not None:
+                    if instance.fields['source'].sequence_length() <= self._max_sequence_length + 2:
+                        yield instance
```
joelgrus: I find nested `if` logic like this hard to reason about, I probably would have written this as

```
seq_length = instance.fields['source'].sequence_length()
if self._max_sequence_length is None or seq_length <= self._max_sequence_length + 2:
    yield instance
```

IMO that's clearer than having two `yield` statements in two different places

allennlp/tests/fixtures/elmo_port/shards/shard0
```diff
@@ -0,0 +1,100 @@
+The U.S. Centers for Disease Control and Prevention initially advised school systems to close if outbreaks occurred , then reversed itself , saying the apparent mildness of the virus meant most schools and day care centers should stay open , even if they had confirmed cases of swine flu .
```
joelgrus: do we still want to call this directory `elmo_port`? or should we call it something more general like `language_modeling`

brendan-ai2: Done. To some extent it might be helpful to explain where the data is coming from, but I'm not sure this matters much for fixtures.

brendan-ai2:
Thanks for the review! Merging on green.

Issue #1877, opened 2018-10-07
Make class `Trainer` Registrable


WrRan:
**Is your feature request related to a problem? Please describe.**
When I try to train multiple models in a single training process, and the training order of these models is alternated according to their performance, the default trainer does not meet this requirement very well (Or maybe it's because I'm not familiar with the usage of the `allennlp`).

**Describe the solution you'd like**
Make class `Trainer` registrable, and we can custom our own training process in config.

**Describe alternatives you've considered**
I do not have another solutions now.

**Additional context**

matt-gardner:
This seems reasonable to me.  We're already constructing the `Trainer` [`from_params`](https://github.com/allenai/allennlp/blob/043ff07c6f0e0f95b71fe3463f86b722f871a6ad/allennlp/commands/train.py#L309), so this should just work.  PRs welcome.

matt-gardner:
Though, you might actually have a hard time implementing your particular use case just by passing in a different `Trainer`.  We only pass a single model in.  You'd have to also modify the `Train` command, and if you're doing that, you can use your own `Trainer` class without having `Trainer` be `Registrable`.

So, yeah, it seems fine to have the `Trainer` be `Registrable` - it doesn't really hurt anything, and it's a very small change - I'm just not sure it really solves anything, either.

zzyxzz:
As there're researches on training strategies, @matt-gardner I think it's a good suggestion (in the future) to allow loading cuntomized ```Trainer``` from configuration file using ```train``` command. 

Another use case is similar to @WrRan , but training a teacher-student network. When train the teacher model using a human annotated dataset, the student model is inactivated with no grad flow. When train the student model using another dataset, its loss is weighted by the confidence scores from teacher model. Training of teacher model and student model are randomly alternated during the training process.

matt-gardner:
@zzyxzz, we did make it configurable, thanks to @WrRan's PR.  I agree there are some cases where this might be useful.  However, in both the case that he mentioned and the case that you're bringing up, just subclassing `Trainer` with a new type won't solve the problem, because you need _different logic in the `train` command also_.  The problem is bigger than just allowing a customized `Trainer`.

PR #1884, opened 2018-10-09
feature-enhancement: make trainer registrable

WrRan:
This is the solution to [#1877](https://github.com/allenai/allennlp/issues/1877).

### DIFF
--- a/allennlp/commands/fine_tune.py
+++ b/allennlp/commands/fine_tune.py
@@ -215,12 +215,20 @@ def fine_tune_model(model: Model,
     for name in tunable_parameter_names:
         logger.info(name)
 
-    trainer = Trainer.from_params(model,
-                                  serialization_dir,
-                                  iterator,
-                                  train_data,
-                                  validation_data,
-                                  trainer_params)
+    # If the config specifies a trainer subclass, we should use it.
+    # There must exist a trainer.
+    trainer_params = params.get("trainer")
+    trainer_choice = trainer_params.pop_choice("type",
+                                               Trainer.list_available(),
+                                               True)
+    trainer_cls = Trainer.by_name(trainer_choice)
+
+    trainer = trainer_cls.from_params(model,
+                                      serialization_dir,
+                                      iterator,
+                                      train_data,
+                                      validation_data,
+                                      trainer_params)
 
     evaluate_on_test = params.pop_bool("evaluate_on_test", False)
     params.assert_empty('base train command')
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -1038,25 +1038,25 @@ class Trainer(Registrable):
         should_log_learning_rate = params.pop_bool("should_log_learning_rate", False)
 
         params.assert_empty(cls.__name__)
-        return Trainer(model, optimizer, iterator,
-                       train_data, validation_data,
-                       patience=patience,
-                       validation_metric=validation_metric,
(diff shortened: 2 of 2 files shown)


allennlp/commands/fine_tune.py
```diff
@@ -215,12 +215,20 @@ def fine_tune_model(model: Model,
...
 
-    trainer = Trainer.from_params(model,
-                                  serialization_dir,
-                                  iterator,
-                                  train_data,
-                                  validation_data,
-                                  trainer_params)
+    # If the config specifies a trainer subclass, we should use it.
+    # There must exist a trainer.
+    trainer_params = params.get("trainer")
```
matt-gardner: `trainer_params` is already popped above, so this is returning `None`.  You can just remove this line, and the comments above don't really add anything, so I'd probably remove those too.

allennlp/commands/fine_tune.py
```diff
@@ -215,12 +215,20 @@ def fine_tune_model(model: Model,
...
+    trainer_choice = trainer_params.pop_choice("type",
+                                               Trainer.list_available(),
+                                               True)
+    trainer_cls = Trainer.by_name(trainer_choice)
+
+    trainer = trainer_cls.from_params(model,
```
matt-gardner: I'd be inclined to combine this line with the one above it, so you have `Trainer.by_name(trainer_choice).from_params(...)`

allennlp/commands/fine_tune.py
```diff
@@ -215,12 +215,20 @@ def fine_tune_model(model: Model,
...
-                                  iterator,
-                                  train_data,
-                                  validation_data,
-                                  trainer_params)
+    # If the config specifies a trainer subclass, we should use it.
+    # There must exist a trainer.
+    trainer_params = params.get("trainer")
+    trainer_choice = trainer_params.pop_choice("type",
+                                               Trainer.list_available(),
+                                               True)
```
matt-gardner: When you have hard-coded values like this, it's a lot more readable to pass them as keyword arguments.  I had to go look at the definition of `pop_choice` to remind myself of what this parameter meant, while I wouldn't have had you passed this as a keyword argument.  So `default_to_first_choice=True` would be better here.

allennlp/training/trainer.py
```diff
@@ -156,7 +156,9 @@ def str_to_time(time_str: str) -> datetime.datetime:
     return datetime.datetime(*pieces)
 
 
-class Trainer:
+class Trainer(Registrable):
```
matt-gardner: You need to add `# type: ignore` and `# pylint: disable=arguments-differ` on `from_params` in this class.  Pylint and mypy are failing because of this.

allennlp/commands/fine_tune.py
joelgrus: I would call this function using keyword arguments

`from_params(model=model, ... )`

otherwise you're implicitly requiring any trainer subclass to have a from_params that takes the exact same arguments in the exact same order, which seems unnecessary

allennlp/commands/fine_tune.py
WrRan: My bad. I didn't notice the line. I will fix it.

allennlp/commands/fine_tune.py
WrRan: Got it. Good suggestions. Thanks.

matt-gardner:
Thanks!

PR #1892, opened 2018-10-10
set "evaluate_on_test" to true

tuvuumass:


DeNeutoy:
As a rule, you shouldn't by default evaluate on the test set. You can use the `allennlp evaluate ` command instead.

PR #1902, opened 2018-10-12
Multilayer decoder for semantic parsing framework

kl2806:
Adds a configuration option so the LSTM decoder can have multiple layers.


kl2806:
It looks like it is around 5-10% slower when we use ``LSTM`` instead of ``LSTMCell``. Did you mean ``num_layers==1`` instead of ``num_layers==0``?

allennlp/state_machines/states/rnn_statelet.py
```diff
@@ -50,6 +52,8 @@ def __init__(self,
                  attended_input: torch.Tensor,
                  encoder_outputs: List[torch.Tensor],
                  encoder_output_mask: List[torch.Tensor]) -> None:
+        # self.hidden_state = hidden_state.unsqueeze(0) if len(hidden_state.shape) == 1 else hidden_state
```
matt-gardner: Remove commented out code.

PR #1905, opened 2018-10-13
Text2sql model

DeNeutoy:
- Rename `ProductionRuleArray` -> `ProductionRule`
- Make it a namedtuple (It threw me that this wasn't a tensor which made me super confused about how we were indexing into it.)
- Add a `Text2SqlParser` which currently will only support global actions


allennlp/tests/data/dataset_readers/semantic_parsing/grammar_based_text2sql_test.py
```diff
@@ -1,8 +1,11 @@
 # pylint: disable=invalid-name,line-too-long
+
+import pytest
+
 from allennlp.data.dataset_readers.semantic_parsing.grammar_based_text2sql import GrammarBasedText2SqlDatasetReader
 from allennlp.common.testing import AllenNlpTestCase
 
-
+@pytest.mark.skip(reason="Mark will fix in a nearby PR.")
 class TestGrammarBasdText2SqlDatasetReader(AllenNlpTestCase):
```
matt-gardner: Minor typo: `GrammarBasd`

allennlp/models/semantic_parsing/text2sql_parser.py
```diff
@@ -0,0 +1,399 @@
...
+                 vocab: Vocabulary,
+                 utterance_embedder: TextFieldEmbedder,
+                 action_embedding_dim: int,
+                 encoder: Seq2SeqEncoder,
+                 decoder_beam_search: BeamSearch,
+                 max_decoding_steps: int,
+                 input_attention: Attention,
+                 database_file: str,
+                 add_action_bias: bool = True,
+                 training_beam_size: int = None,
```
matt-gardner: This isn't needed, is it?  This is only for when you have more than one logical form at training time.

allennlp/models/semantic_parsing/text2sql_parser.py
```diff
@@ -0,0 +1,399 @@
...
+                 dropout: float = 0.0,
+                 initializer: InitializerApplicator = InitializerApplicator(),
+                 regularizer: Optional[RegularizerApplicator] = None) -> None:
+        super().__init__(vocab, regularizer)
+
+        self._utterance_embedder = utterance_embedder
+        self._encoder = encoder
+        self._max_decoding_steps = max_decoding_steps
+        self._add_action_bias = add_action_bias
+        self._dropout = torch.nn.Dropout(p=dropout)
```
matt-gardner: Does this work when dropout is 0?  If it does, why have we been doing the `lambda x: x` stuff this whole time?

allennlp/models/semantic_parsing/text2sql_parser.py
```diff
@@ -0,0 +1,399 @@
...
+            for batch_actions in valid_actions:
+                batch_action_mapping = {}
+                for action_index, action in enumerate(batch_actions):
+                    batch_action_mapping[action_index] = action[0]
+                action_mapping.append(batch_action_mapping)
+
+            outputs['action_mapping'] = action_mapping
+            # This tells the state to start keeping track of debug info, which we'll pass along in
+            # our output dictionary.
+            initial_state.debug_info = [[] for _ in range(batch_size)]
```
matt-gardner: Looking at this, we should probably just have a method to call on the `State` object - would be easier to understand.  Not for this PR necessarily, just a note to myself.

allennlp/models/semantic_parsing/text2sql_parser.py
```diff
@@ -0,0 +1,399 @@
...
+        ``ProductionRules``, and we use those to embed the valid actions for every
+        non-terminal type.  We use the input ``linking_scores`` for non-global actions.
+
+        Parameters
+        ----------
+        possible_actions : ``List[ProductionRule]``
+            From the input to ``forward`` for a single batch instance.
+        """
+        device = util.get_device_of(self._action_embedder.weight)
+        # TODO(Mark): This type is pure \(- . ^)/
```
matt-gardner: Yeah...  Maybe we should introduce a class for this...?  That might make it easier.

allennlp/models/semantic_parsing/text2sql_parser.py
```diff
@@ -0,0 +1,399 @@
+import logging
```
matt-gardner: So, this model looks good to me, but here's a higher-level question:

There is almost nothing in here that is specific to SQL - this is a standard grammar-constrained semantic parser.  It's basically identical to the NLVR model, where the only real difference is in the executor and the metric.  Is there a way to have a model implemented that people can just use, with ~10 lines of code (for the execution and metric bits) instead of ~400?

allennlp/models/semantic_parsing/text2sql_parser.py
DeNeutoy: oh yep good point, thanks.

allennlp/models/semantic_parsing/text2sql_parser.py
DeNeutoy: Yes, but I don't want to think about that until after the NACCL deadline, because there are going to have to be quite a few modifications to this to run the ablations etc.

allennlp/models/semantic_parsing/text2sql_parser.py
matt-gardner: One quick point: Kevin found that passing `None` to `MaximumMarginalLikelihood` gave a huge slowdown for some reason I don't understand yet.  You might still want to pass `1`.  It just doesn't need to be a constructor parameter.

allennlp/models/semantic_parsing/text2sql_parser.py
matt-gardner: Yeah, I agree, not for this PR.  Just something on my mind.

Issue #1575, opened 2018-08-06
allennlp train doesn't play nice with pytorch profiler


nelson-liu:
The PyTorch profiler works by running your code twice, once with the python profiler and once with their autograd profiler. however, it breaks the second time around since the `allennlp train` output directory already exists.

It'd be nice to have a `-f / --force` flag that removes the output directory if it exists before retraining...

schmmd:
@nelson-liu can you give a detailed example of how the force flag would help here?

nelson-liu:
So the way the pytorch profiler works is that it runs input python command twice (I think once with a python profiler and one with an autograd profiler). It runs fine the first time, but the second time around, the command fails because the serialization directory already exists.

Does that make sense? I provide a example to reproduce if you want.

matt-gardner:
My main question was about the entry point.  How do you invoke the pytorch profiler?  From the command line?  This matters because how we fix this depends on the entry point.

PR #1913, opened 2018-10-16
Add a --force command to train

schmmd:
Fixes https://github.com/allenai/allennlp/issues/1575

### DIFF
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -20,9 +20,8 @@ which to write the results.
       -h, --help            show this help message and exit
       -s SERIALIZATION_DIR, --serialization-dir SERIALIZATION_DIR
                             directory in which to save the model and its logs
-      -f FORCE, --force FORCE
-                            overwrite the output directory if it exists
       -r, --recover         recover training from the state in serialization_dir
+      -f, --force           overwrite the output directory if it exists
       -o OVERRIDES, --overrides OVERRIDES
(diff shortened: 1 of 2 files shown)


allennlp/commands/train.py
```diff
@@ -178,42 +193,47 @@ def create_serialization_dir(params: Params, serialization_dir: str, recover: bo
         If ``True``, we will try to recover from an existing serialization directory, and crash if
         the directory doesn't exist, or doesn't match the configuration we're given.
     """
+    if recover and force:
+        raise ConfigurationError("Illegal arguments: both force and recover are true.")
+
     if os.path.exists(serialization_dir) and os.listdir(serialization_dir):
-        if not recover:
+        if force:
```
matt-gardner: The level of nesting that this gets to is scary.  I also think you can avoid it - what if you just move the remove block up above, and leave the rest of the code unmodified?  That is, have something like:

```python
if recover and force:
    raise ...

if force:
    shutil.rmtree(serialization_dir)

...  # the rest of the code unchanged
```

schmmd:
If I omit the exists check then I need to tell `shutil.rmtree` to ignore errors, which I would rather avoid.

matt-gardner:
Yeah, keeping that exists check is fine, I was more commenting on pulling out that check.  Looks like you forgot to move back the main logic that you changed, though.

matt-gardner:
:+1:

PR #1914, opened 2018-10-17
Adding support for list, tuple, and set in from_params

### LINKED ISSUE #1862: List does not get instantiated correctly by default from_params
I have a custom class called `Annotator` which inherits `Registrable` and has a few subclasses that I register. My dataset reader accepts a parameter of type `List[Annotator]`. When my dataset reader is instantiated, the list of annotators is given as `List[Params]`.

There are several ways to get around this, such as by implementing my own `from_params`, but it would be nice if the objects were created by default.

I'm using allennlp version 0.6.1

danieldeutsch:
Fixes #1862

A couple of limitations I see/comments:

1. What should happen if the length of the tuple specified by the typing is different than the length of the tuple passed in the params?
2. The code is not recursive, so `List[List[...]]` will not work
3. The code does not handle primitive types, so `Tuple[A, int]` will not work
4. I can't think of a good use case for a set, but I implemented it anyway

Happy to hear comments

### DIFF
--- a/allennlp/common/from_params.py
+++ b/allennlp/common/from_params.py
@@ -40,7 +40,7 @@ In this case your class just needs to explicitly implement its own `from_params`
 method.
 """
 
-from typing import TypeVar, Type, Dict, Union, Any, cast
+from typing import TypeVar, Type, Dict, Union, Any, cast, List, Tuple, Set
 import inspect
 import logging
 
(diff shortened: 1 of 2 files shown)


matt-gardner:
I'm going to merge what you have now, as it's good.  If you want to add better error handling for tuples of mismatched lengths (just raise a `ConfigurationError` above line 199 if the length of the lists don't match), or fixes for the other two issues, that's great, just do it in a separate PR.

PR #1917, opened 2018-10-17
Various fixes related to the variable free wikitables world

pdasigi:
@matt-gardner This is now ready to review. Other than adding a dataset reader, and (mostly) copying the original MML model to make a variable free variant, the changes are:
1) Moved some predicates in the variable free world into local mappings to make them table-specific (for example, if the table has no date columns, productions leading to functions like `filter_date_greater` are not valid).
2) Related to the above change, when substituting a `MultiMatchNamedBasicType` within a `ComplexType` to get productions that lead to it, only those that are in the table are allowed (For example, productions like `s -> [<r,<g,s>> r, f]` are not allowed if the table does not have number columns (`f`)).
3) Added a method within `TableQuestionContext` to return a `KnowledgeGraph`.


allennlp/models/semantic_parsing/wikitables_variable_free/wikitables_variable_free_semantic_parser.py
```diff
@@ -0,0 +1,702 @@
...
+        # These encode the same information, but for efficiency reasons later it's nice
+        # to have one version as a tensor and one that's accessible on the cpu.
+        entity_types, entity_type_dict = self._get_type_vector(world, num_entities, encoded_table)
+
+        entity_type_embeddings = self._entity_type_encoder_embedding(entity_types)
+
+        # (batch_size, num_entities, num_neighbors) or None
+        neighbor_indices = self._get_neighbor_indices(world, num_entities, encoded_table)
+
+        if neighbor_indices is not None:
```
pdasigi: This part is new. It's possible that none of the instances in a given batch have entities with neighbors, because of the way the knowledge graph is defined.

pdasigi:
@matt-gardner Based on our offline conversation about separating research code into a separate repo, I'm removed several changes from this PR to the other repo. I kept all the changes I made to the files already in the `master` branch in this repo. So, the ones I took out are the new dataset reader and model code, and their tests and fixtures. The rest of the summary above still holds true.

allennlp/semparse/worlds/wikitables_variable_free_world.py
```diff
@@ -52,50 +52,108 @@ def __init__(self, table_context: TableQuestionContext) -> None:
...
+            self._table_has_string_columns = True
+        if "date" in column_types:
+            for name, translated_name in types.DATE_COLUMN_NAME_MAPPING.items():
+                signature = types.DATE_COLUMN_TYPE_SIGNATURE[translated_name]
+                self._add_name_mapping(name, translated_name, signature)
+            # Adding -1 to mapping because we need it for dates where not all three fields are
+            # specified. We want to do this only when the table has a date column. This is because
+            # the knowledge graph is also constructed in such a way that -1 is an entity with date
+            # columns as the neighbors only if any date columns exist in the table.
+            self._map_name(f"num:-1", keep_mapping=True)
```
matt-gardner: I think we may want to handle the `-1` in dates separately from the typical "number" entity mechanism here - maybe have it be a global action, not a linked action at all.  But we can worry about that later.

PR #1927, opened 2018-10-19
Disable tqdm when there isn't a TTY

DeNeutoy:
@splittingfield FYI. This won't take much effect until the next release but 🤷‍♂️

### DIFF
--- a/allennlp/common/util.py
+++ b/allennlp/common/util.py
@@ -221,7 +221,8 @@ def prepare_global_logging(serialization_dir: str, file_friendly_logging: bool) 
     file_friendly_logging : ``bool``, required.
         Whether logs should clean the output to prevent carridge returns
         (used to update progress bars on a single terminal line). This
-        option is only used if you are running in an environment with a terminal.
+        option is typically only used if you are running in an environment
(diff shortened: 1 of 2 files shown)


allennlp/common/util.py
```diff
@@ -220,8 +220,15 @@ def prepare_global_logging(serialization_dir: str, file_friendly_logging: bool)
...
         Whether logs should clean the output to prevent carridge returns
-        (used to update progress bars on a single terminal line).
+        (used to update progress bars on a single terminal line). This
+        option is only used if you are running in an environment with a terminal.
     """
+
+    # If we don't have a terminal as stdout,
+    # force tqdm to be nicer.
+    if not sys.stdout.isatty():
+        file_friendly_logging = True
```
joelgrus: is there ever a case where you'd want file_friendly_logging in a terminal?

that is, with this change can we just get rid of the file_friendly_logging config parameter everywhere?

matt-peters: I use file_friendly_logging with a terminal.

schmmd:
@DeNeutoy love it!

Issue #1929, opened 2018-10-22
[Bug] It fails when using wordsplit with pos_tag or ner_tag.


WrRan:
**Describe the bug**
When I train bidaf with pos_tag or ner_tag, it fails to begin training.

**To Reproduce**
Steps to reproduce the behavior
1. Copy `training_config/bidaf.jsonnet` to custom directory.
2. Modify `bidaf.jsonnet` as follow:
a) change `toke_indexers` of `dataset_reader`
```jsonnet
    "token_indexers": {
      // config `tokens` and `token_characters` as before ...
      "pos_tag": {
        "type": "pos_tag"
      },
      "ner_tag": {
        "type": "ner_tag"
      }
    }
```
b) change `text_field_embedder` of `model`:
```jsonnet
    "text_field_embedder": {
       // config token_embedders here ...
      "allow_unmatched_keys": true
    },
```
3. apply command: `allennlp train config/bidaf.jsonnet -s data/bidaf`
4. See error as follow:
```bash
KeyError: '@@UNKNOWN@@'
```
or
```bash
RuntimeError: Dimension out of range (expected to be in range of [-1, 0], but got 1)
```

**Expected behavior**
No error and same behaviour as before.

**System (please complete the following information):**
 - OS: Linux
 - Python version: 3.6.1
 - AllenNLP version: 0.6.1
 - PyTorch version: 0.4.0

**Additional context**
I find there are three code snippets that may be related to this bug. I am trying to fix it.
- [allennlp/data/token_indexers/ner_tag_indexer.py](https://github.com/allenai/allennlp/blob/master/allennlp/data/token_indexers/ner_tag_indexer.py#L41)
- [allennlp/data/token_indexers/pos_tag_indexer.py](https://github.com/allenai/allennlp/blob/master/allennlp/data/token_indexers/pos_tag_indexer.py#L59)
- [allennlp/nn/util.py](https://github.com/allenai/allennlp/blob/master/allennlp/nn/util.py#L428)

matt-gardner:
If you want to use POS and NER tags, you need to be sure you also modify the tokenizer to _produce_ these tags.  The default tokenizer for BiDAF only splits words, it doesn't run any tagger on them (that would dramatically slow down the data processing).

The particular errors that you give are hard to diagnose, however, as you're not giving a complete stack trace.  If you give us more detail, we'd be able to help you better.

WrRan:
[...]

PR #1935, opened 2018-10-23
use wordsplit with taggers

WrRan:
patch to #1929

### DIFF
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -466,9 +466,17 @@ def get_text_field_mask(text_field_tensors: Dict[str, torch.Tensor],
     smallest_dim = tensor_dims[0][0] - num_wrapping_dims
     if smallest_dim == 2:
         token_tensor = tensor_dims[0][1]
+        for ix in range(1, len(tensor_dims)):
+            if smallest_dim != tensor_dims[ix][0] - num_wrapping_dims:
+                break
+            token_tensor += tensor_dims[ix][1]
         return (token_tensor != 0).long()
     elif smallest_dim == 3:
         character_tensor = tensor_dims[0][1]
+        for ix in range(1, len(tensor_dims)):
+            if smallest_dim != tensor_dims[ix][0] - num_wrapping_dims:
+                break
+            character_tensor += tensor_dims[ix][1]
         return ((character_tensor > 0).long().sum(dim=-1) > 0).long()
     else:
         raise ValueError("Expected a tensor with dimension 2 or 3, found {}".format(smallest_dim))


matt-gardner:
Ok, it took me a while, but I understand both issues.  The first one is that we are not consistent in how we are checking for missing / empty values between `count_vocab_items` and `tokens_to_indices`.  The second one is that if you don't have OOV / padding tokens in your vocabulary for one of your token tensors, our masking logic is incorrect.

For issue 1, I still don't understand the case where this might _actually_ be a problem, and whether this fix does the right thing.  E.g., when do you get `''` for a POS or NER tag?  Is replacing with `NONE` really the right thing?  I don't know without seeing an example.  If you could add a test case that fails before your change (preferably with actual input that your code failed on), that would help.

For issue 2, I don't think your fix actually solves the underlying problem.  What if I _only_ have NER tags (or things that don't have OOV / padding tokens) for my indexer?  Then I still get the wrong result.  The issue is that the mask code assumes you have these tokens, and I don't know of a way around it.  I think the right solution here is to just make sure the vocabulary namespace that you use for each of your token indexers is padded.  We can't recover from the case where you don't have padding tokens, but still want to compute a mask.

matt-gardner:
Ok, thanks, I didn't know that that was spacy's behavior for "no NER tag".  A failing test before this change would still be good to have, just to make sure we don't break it later.

For the latter issue, I don't know of an easy way around this.  We'd have to have the `TokenIndexer` check that the `Vocabulary` thinks the namespace it's using should be padded, and that's a pretty ugly check.  The default namespace would have given you padding tokens - did you change something to get this error?

WrRan:
Actually, `Vocabulary` does not pad tokens with namespace ending with `tags` or `labels`.
https://github.com/allenai/allennlp/blob/043ff07c6f0e0f95b71fe3463f86b722f871a6ad/allennlp/data/vocabulary.py#L22
https://github.com/allenai/allennlp/blob/043ff07c6f0e0f95b71fe3463f86b722f871a6ad/allennlp/data/vocabulary.py#L282-L283

Moreover, the default namespaces of PosTagIndexer and NerTagIndexer are `pos_tags` and `ner_tags` respectively.
So I got this error when changing nothing but configuration.

matt-gardner:
Oh, sorry, you're right. We should change that default.

WrRan:
Happy to be right. :)

matt-gardner:
Ok, so what this PR needs in order to be merged:

1. Remove the changes to `util.py`.
2. Add a test case to catch the problem with blank NER / POS tags.
3. Change the default namespace for the NER and POS token indexers to not end in `tags`.

WrRan:
Got it.
But may it be a good idea to modify `DEFAULT_NON_PADDED_NAMESPACES` in `vocabulary.py` instead of the default namespaces for the NER and POS?

matt-gardner:
No, we don't want to modify the defaults in the vocabulary, because (1) the main use case for those defaults is for things you're predicting (i.e., when POS or NER tags are used in a `SequenceLabelField` as your target), and (2) we don't want to make changes that break existing code.  I suspect very few people are using these particular `TokenIndexers`, and if they are they would have run into this issue, anyway, because it's just broken.  So the right place to fix this is in the `TokenIndexers`.

WrRan:
Yes, I see. You are right.

WrRan:
Sorry.
I have deleted original repo, which makes this PR from `unknown repo`.
And I do not know how to handle this.
So I create a new PR #1981, this one may be closed.

PR #1944, opened 2018-10-24
Extend GPU fixes from vidurj and MaxMotovilov

brendan-ai2:
- Builds on https://github.com/allenai/allennlp/pull/1724 and https://github.com/allenai/allennlp/pull/1445 in order to fix https://github.com/allenai/allennlp/issues/1439.
- Differences:
  - Use a `scatter` analogous to the PyTorch implementation.
  - Identify metadata that needs to be scattered with a dummy `ScatterableList` subclass.
  - Dispatch to PyTorch's underlying tensor scatter to precisely match behavior.
- Minor fixes to parallelize `Event2Mind` model. (Used to help develop the rest.)


brendan-ai2:
@vidurj, this isn't 100% ready for review yet, but what do you think of the basic concept? It should be robust and it's also very simple.

allennlp/data/fields/metadata_field.py
```diff
@@ -61,7 +62,7 @@ def empty_field(self) -> 'MetadataField':
     @classmethod
     @overrides
     def batch_tensors(cls, tensor_list: List[DataArray]) -> DataArray:  # type: ignore
-        return tensor_list  # type: ignore
+        return ScatterableList(tensor_list)  # type: ignore
```
matt-gardner: The `ProductionRuleField` needs a similar modification.

brendan-ai2: Good catch, thanks! This does make me wonder if we should modify the type that `batch_tensors` is supposed to return to include the possibility of using `ScatterableList`. But historically we've just been ignoring types here. My intuition for this in Python is not particularly developed.

brendan-ai2:
@matt-gardner, would you mind reviewing this actually? I've cleaned it up a bit. Only remaining bit of work should be fixing the licensing. Thanks!

allennlp/common/util.py
```diff
@@ -354,6 +357,107 @@ def is_lazy(iterable: Iterable[A]) -> bool:
...
+    A normal list, but one that should be scattered like a tensor.
+    """
+
+    # Ensure pointers will fit in a torch.LongTensor. "64 bits ought to be enough for anybody."
+    assert sizeof(c_void_p) <= sizeof(c_int64)
+
+    def to_pointer_tensor(self) -> torch.LongTensor:
+        """
+        Converts the elements to pointers, casts them to ``int64`` and then returns them in a tensor. This cast is
+        important as ``id`` gives back unsigned integers while ``torch.LongTensor`` is signed.
```
matt-gardner: Ouch.  That must have been annoying to find...

allennlp/common/util.py
```diff
@@ -354,6 +357,107 @@ def is_lazy(iterable: Iterable[A]) -> bool:
...
+
+        See:
+        https://github.com/python/cpython/blob/6ec5cf24b7f38ea72bb42d5cd60dca0d3ee332f9/Python/bltinmodule.c#L1118
+        https://github.com/python/cpython/blob/6ec5cf24b7f38ea72bb42d5cd60dca0d3ee332f9/Objects/longobject.c#L990
+        """
+        pointers = [c_int64(id(element)).value for element in self]
+        return torch.LongTensor(pointers)
+
+    @classmethod
+    def from_pointer_tensor(cls, pointers) -> list:
```
matt-gardner: Type annotation on `pointers`.  It's a `List[torch.LongTensor]`, where the tensor should be a single scalar?  Or is it a single tensor of shape `(list_length,)` that you're iterating over?

allennlp/common/util.py
```diff
@@ -354,6 +357,107 @@ def is_lazy(iterable: Iterable[A]) -> bool:
...
+        """
+        The inverse of ``to_pointer_tensor`` except that a plain ``list`` is returned.
+        """
+        return [cast(c_uint64(pointer.item()).value, py_object).value for pointer in pointers]
+
+# TODO(brendanr): Add licensing stuff for borrowing this from torch before distributing, i.e. pushing to master.
```
matt-gardner: Is it sufficient to just point to the license in the docstring, next to where you point to the github repo?

allennlp/tests/training/trainer_test.py
```diff
@@ -86,12 +88,33 @@ def test_trainer_can_run_cuda(self):
                           cuda_device=0)
         trainer.train()
 
+  # pylint: disable=arguments-differ
```
matt-gardner: Is this for the call to `forward` below?  I'd say this belongs inside that scope, if that's what this is for.

allennlp/common/util.py
brendan-ai2: Fortunately this didn't manifest as a mysterious bug. C/C++ programming has left me with a fair degree of caution around this sort of thing.

allennlp/common/util.py
brendan-ai2: A single tensor of shape `(list_length,)`. Type and comment to that effect added.

allennlp/common/util.py
brendan-ai2: @schmmd suggested the same thing. Sounds good enough to me.

allennlp/tests/training/trainer_test.py
brendan-ai2: Oops! Fixed. This line escaped during some experimentation with the linter.

brendan-ai2:
Note: I've bypassed the patch code coverage check because the testing for this PR relies on having multiple GPUs available. See https://github.com/allenai/allennlp/blob/master/allennlp/tests/training/trainer_test.py#L91. We should probably have a story for this sort of integration testing.

brendan-ai2:
@matt-gardner, sorry, somehow I missed your question about the smaller batches! Vidur asked about that and Matt Peters' main concern was as follows:

> The biggest issue I see is with something like the bucket iterator.
we want the work for each GPU to be approximately equal, but the bucket iterator will yield consecutive batches with very different sequence lengths.
for training LMs, I use the bucket iterator like batching to fill each batch with the same number of tokens (batch size * sequence length is about constant)

(I'm not entirely sure I grok the issue myself. Probably will be more clear once I have something running end-to-end. For now the solution in this PR should suffice, but we can certainly revisit later.)

matt-gardner:
Yeah, ok, that makes sense. Thanks.

allennlp/training/trainer.py
```diff
@@ -1013,13 +1014,11 @@ def from_params(cls,  # type: ignore
...
         shuffle = params.pop_bool("shuffle", True)
         num_epochs = params.pop_int("num_epochs", 20)
-        cuda_device = params.pop_int("cuda_device", -1)
+        cuda_device = parse_cuda_device(params.pop("cuda_device", -1))
         grad_norm = params.pop_float("grad_norm", None)
         grad_clipping = params.pop_float("grad_clipping", None)
         lr_scheduler_params = params.pop("learning_rate_scheduler", None)
 
-        if cuda_device >= 0:
-            model = model.cuda(cuda_device)
```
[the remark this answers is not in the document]
brendan-ai2: We call `self.model.cuda` in the constructor, so I believe this is redundant. Could you please provide more detail? A stack trace would be ideal. Thank you.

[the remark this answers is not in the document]
matt-gardner: This is the cause of #2182.  This needs to be here.  Sorry I didn't catch this earlier.  The issue is that the optimizer constructs its internal state based on the device that the model is on.  For some optimizers, like adagrad, that internal state has tensors in it that need to be on the right device.  That's why these lines are here - the call to `model.cuda()` _has_ to be before the optimizer is created.  #2190 fixes this.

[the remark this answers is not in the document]
matt-gardner: On these lines, is there any way we can make our CI at least occasionally run GPU tests?  That way we could actually have a regression test for this (we've had exactly this error before, which is why these lines were originally added).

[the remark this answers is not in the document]
bryant1410: :+1: 

Maybe also adding a comment would be good then.

PR #1955, opened 2018-10-24
Add scalar_mix_parameters to ElmoTokenEmbedder

nelson-liu:


allennlp/modules/token_embedders/elmo_token_embedder.py
```diff
@@ -39,6 +39,10 @@ class ElmoTokenEmbedder(TokenEmbedder):
         indices of shape (batch_size, timesteps) to forward, instead
         of character indices. If you use this option and pass a word which
         wasn't pre-cached, this will break.
+    scalar_mix_parameters : ``List[int]``, optional, (default=None)
+        If not ``None``, use these scalar mix parameters to weight the representations
+        produced by different layers. These mixing weights are not updated during
+        training.
```
matt-gardner: Is there an option to just initialize to these values, but still train them?  Just something to think about.

nelson-liu: yeah, you're right that this is currently missing. I didn't think anyone would use it / it would add extra complexity, so i just went with this for now.

matt-gardner: Yeah, definitely don't add it until there's a justification for it; I was just wondering about your experiments.

PR #1958, opened 2018-10-24
Add config for coref model with elmo

nelson-liu:
Also added a disclaimer that this doesn't reproduce the number in the elmo experiment, and that you should reference the document mattp put together a few weeks ago on it.


nelson-liu:
hold off on merging this, looks like the config results in an OOM (thanks @matt-peters for sharing that this would happen).

PR #1964, opened 2018-10-25
Try compiling EVALB in metric if it doesn't exist

nelson-liu:
This is necessary in order to get it to work on beaker, since otherwise the evaluation binary is never produced.


schmmd:
@nelson-liu another approach that would make this work with Beaker would be to compile this in [the Dockerfile](https://github.com/allenai/allennlp/blob/master/Dockerfile), e.g.

`RUN python -c 'from allennlp.training.metrics import EvalbBracketingScorer; EvalbBracketingScorer.compile_evalb()'`

Or better yet--to add a script such as `compile_requirements.sh`.  This would have the benefit of catching issues earlier (when the Dockerfile is built rather than when the code path is run).  Just my 2 cents though--you already have this approved ;-)

Issue #1954, opened 2018-10-24
[Suggestion] Add parameter `mini_width` or something else to `TokenCharactersIndexer`'s initializer.


WrRan:
**Describe the bug**
I used a simple CNN on a text classification tasks. It worked well when training, but it broke when predicting.

**To Reproduce**
For convenience, I provide a test case at [allennlp-demo-producing-bug](https://github.com/WrRan/allennlp-demo-producing-bug).
Steps to reproduce the behavior
1. Clone the repository to local dir: `git clone https://github.com/WrRan/allennlp-demo-producing-bug.git demo`
2. Go to demo: `cd demo`
3. Begin training: 
```bash
export PYTHONPATH=`pwd` && export CUDA_VISIBLE_DEVICES=0
allennlp train config/cnn_classifier.json -s data/model/cnn --include-package tc
```
4. Begin predicting:
```bash
allennlp predict data/model/cnn/model.tar.gz data/data.txt --include-package tc --predictor sentiment_classifier --use-dataset-reader
```
4. See error:
```bash
Traceback (most recent call last):
  File "/home/user_data/anaconda3/envs/docqa/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/user_data/anaconda3/envs/docqa/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/user_data/anaconda3/envs/docqa/lib/python3.6/site-packages/allennlp/run.py", line 20, in <module>
    main(prog="allennlp")
  File "/home/user_data/anaconda3/envs/docqa/lib/python3.6/site-packages/allennlp/commands/__init__.py", line 70, in main
    args.func(args)
  File "/home/user_data/anaconda3/envs/docqa/lib/python3.6/site-packages/allennlp/commands/predict.py", line 192, in _predict
    manager.run()
  File "/home/user_data/anaconda3/envs/docqa/lib/python3.6/site-packages/allennlp/commands/predict.py", line 168, in run
    for result in self._predict_instances(batch):
  File "/home/user_data/anaconda3/envs/docqa/lib/python3.6/site-packages/allennlp/commands/predict.py", line 137, in _predict_instances
    results = [self._predictor.predict_instance(batch_data[0])]
  File "/home/user_data/anaconda3/envs/docqa/lib/python3.6/site-packages/allennlp/predictors/predictor.py", line 53, in predict_instance
[...]

PR #1967, opened 2018-10-25
add min_padding_length to TokenCharactersIndexer (#1954)

WrRan:
patch to #1954 

Although the parameter `min_padding_length` is provided a default value (`0`), I think it may be a better idea with no default value.

### DIFF
--- a/allennlp/data/token_indexers/token_characters_indexer.py
+++ b/allennlp/data/token_indexers/token_characters_indexer.py
@@ -18,9 +18,6 @@ class TokenCharactersIndexer(TokenIndexer[List[int]]):
 
     Parameters
     ----------
-    min_padding_length: ``int``, optional (default=``0``)
-        We use this value as the minimum length of padding. Usually used with :class:``CnnEncoder``, its
-        value should be set to the maximum value of ``ngram_filter_sizes`` correspondingly.
     namespace : ``str``, optional (default=``token_characters``)
         We will use this namespace in the :class:`Vocabulary` to map the characters in each token
         to indices.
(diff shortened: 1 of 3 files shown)


allennlp/data/token_indexers/token_characters_indexer.py
```diff
@@ -33,10 +36,12 @@ class TokenCharactersIndexer(TokenIndexer[List[int]]):
     """
     # pylint: disable=no-self-use
     def __init__(self,
+                 min_padding_length: int = 0,
```
matt-gardner: This needs to be added at the end of the argument list, otherwise you'll break existing calls to the constructor (which is why tests are failing).  This is also why it needs to have a default value of 0 - we don't want to break backwards compatibility.

And, lastly, it's important for every new piece of functionality in the core library to have a test making sure it does what it's supposed to.  Can you add a test making sure that things actually get padded correctly when this parameter is used?

WrRan:
Thanks for your reply. I am willing to add a test case and make dacode backward compatibility.



发自我的小米手机
在 Matt Gardner <notifications@github.com>，2018年10月25日 23:59写道：

@matt-gardner commented on this pull request.

________________________________

In allennlp/data/token_indexers/token_characters_indexer.py<https://github.com/allenai/allennlp/pull/1967#discussion_r228234772>:

> @@ -33,10 +36,12 @@ class TokenCharactersIndexer(TokenIndexer[List[int]]):
     """
     # pylint: disable=no-self-use
     def __init__(self,
+                 min_padding_length: int = 0,


This needs to be added at the end of the argument list, otherwise you'll break existing calls to the constructor (which is why tests are failing). This is also why it needs to have a default value of 0 - we don't want to break backwards compatibility.

And, lastly, it's important for every new piece of functionality in the core library to have a test making sure it does what it's supposed to. Can you add a test making sure that things actually get padded correctly when this parameter is used?

allennlp/tests/data/token_indexers/character_token_indexer_test.py
```diff
@@ -61,3 +61,88 @@ def test_start_and_end_tokens(self):
...
+            assert set([key_padding_lengths]) == set(key)
+            value_padding_lengths = max(value_padding_lengths, max(value))
+        assert value_padding_lengths == 8
+        padded = indexer.pad_token_sequence(indices,
+                                            {"char": len(indices["char"])},
+                                            {key_padding_lengths: value_padding_lengths})
+        assert padded == {"char": [[2, 3, 3, 4, 5, 6, 7, 8],
+                                   [9, 10, 0, 0, 0, 0, 0, 0],
+                                   [11, 12, 4, 10, 13, 14, 4, 0],
+                                   [15, 0, 0, 0, 0, 0, 0, 0]]}
```
matt-gardner: Lines 84 through 126 here test functionality that's already tested elsewhere.  The only thing we really need to test is the last case that you have, starting at line 128.

allennlp/tests/data/token_indexers/character_token_indexer_test.py
```diff
@@ -61,3 +61,88 @@ def test_start_and_end_tokens(self):
...
+                                    [11, 12, 4, 10, 13, 14, 4],
+                                    [15]]}
+        key_padding_lengths = "num_token_characters"
+        value_padding_lengths = 0
+        for token in indices["char"]:
+            item = indexer.get_padding_lengths(token)
+            key, value = item.keys(), item.values()
+            assert set([key_padding_lengths]) == set(key)
+            value_padding_lengths = max(value_padding_lengths, max(value))
+        assert value_padding_lengths == 10
```
matt-gardner: There's no need for lines 130 through 141; you can just go straight to computing the padding, and just have the assert that's on line 145.  The rest is all implicitly tested with that one assert.

allennlp/data/token_indexers/token_characters_indexer.py
```diff
@@ -1,113 +1,119 @@
-from typing import Dict, List
```
matt-gardner: I think you ended up changing the line endings on this file to DOS line endings.  Can you switch it back?

WrRan:
OK. PR's coming soon.

PR #1975, opened 2018-10-26
delete swp file

WrRan:
Delete swp file `allennlp/predictors/.bidaf.py.swp`.
May I add `*.swp` to `.gitignore`?


joelgrus:
adding to .gitignore seems like a good idea

PR #1978, opened 2018-10-26
Match vocab generation in currently online Event2Mind model.

brendan-ai2:
- I noticed this difference while retraining the model to account for https://github.com/allenai/allennlp/pull/1944.
- With this change we generate the exact same vocab as was used for the 10/5 model and now the new 10/26 model.
- In the long run we'll want to be more principled with our vocab. In the short term I'm more concerned with ensuring that we can retrain consistently.

### DIFF
--- a/allennlp/data/dataset_readers/event2mind.py
+++ b/allennlp/data/dataset_readers/event2mind.py
@@ -104,14 +104,15 @@ class Event2MindDatasetReader(DatasetReader):
                                 )
                 # Generate instances where each token of input appears once.
                 else:
-                    # To the extent that sources are duplicated in the dataset
-                    # (which appears common), we will duplicate them here.
-                    yield self.text_to_instance(source_sequence, "none", "none", "none")
                     for xintent in xintents:
+                        # NOTE: source_sequence should really be broken out and deduplicated. We're
+                        # adding it here to ensure we generate the same vocabulary as the model at
+                        # https://s3-us-west-2.amazonaws.com/allennlp/models/event2mind-2018.10.05.tar.gz
+                        # was trained against.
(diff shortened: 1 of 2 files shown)


brendan-ai2:
@rlebras, since we talked about the vocab changes the other day, I'm sending this over to you. Potentially relevant to your Event2Event work. Thanks!

allennlp/data/dataset_readers/event2mind.py
```diff
@@ -104,14 +104,15 @@ def _read(self, file_path):
                                 )
                 # Generate instances where each token of input appears once.
                 else:
-                    # To the extent that sources are duplicated in the dataset
-                    # (which appears common), we will duplicate them here.
-                    yield self.text_to_instance(source_sequence, "none", "none", "none")
                     for xintent in xintents:
```
rlebras: Is it possible to get an empty list of `xintent`? In which case, the `source_sequence` won't be listed.

rlebras: I'm a bit confused here. I thought the point was to not duplicate the source if there are multiple entries for a given target. Or was it how it was done in `event2mind-2018.10.05`? Am I missing something?
Thanks!

brendan-ai2: You're exactly right on both counts. The issue is that this bug was there when I made the vocab for `event2mind-2018.10.05`. When I retrained today with the "corrected" vocab I noticed different -- and not clearly better -- behavior from the model. Until someone has time to devote to determining the optimal strategy* for vocab generation I think it's best that the code at least matches the model.

*Personally I liked what you were saying about using a permissive source vocab and a restricted target vocab. That would explain what I've seen so far.

brendan-ai2:
Thanks!

PR #1991, opened 2018-10-30
Add --force option to find-lr command

epwalsh:
It often takes a couple runs to get the right start and end learning rates and number of batches, so it's convenient when you can just overwrite the last serialization directory. This also just cleans up the docstrings a little for consistency.

### DIFF
--- a/allennlp/commands/find_learning_rate.py
+++ b/allennlp/commands/find_learning_rate.py
@@ -6,12 +6,15 @@ which to write the results.
 .. code-block:: bash
 
    $ allennlp find-lr --help
-   usage: allennlp train [-h] -s SERIALIZATION_DIR [-o OVERRIDES]
-                         [--start-lr START_LR] [--end-lr END_LR]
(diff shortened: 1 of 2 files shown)


allennlp/commands/find_learning_rate.py
```diff
@@ -106,15 +122,16 @@ def find_learning_rate_from_args(args: argparse.Namespace) -> None:
...
 
 def find_learning_rate_model(params: Params,
                              serialization_dir: str,
                              start_lr: float,
                              end_lr: float,
                              num_batches: int,
                              linear_steps: bool,
-                             stopping_factor: Optional[float]) -> None:
+                             stopping_factor: Optional[float],
+                             force: bool) -> None:
```
matt-gardner: This should probably default to `False`, for backwards compatibility (and this is why the tests are failing).

epwalsh: Oops, just fixed that and added a test with `force=True`

matt-gardner:
Thanks!

PR #2000, opened 2018-11-01
Rename DATASET_CACHE to CACHE_DIRECTORY.

schmmd:
This also moves `~/.allennlp/datasets` to `~/.allennlp/cache` since we use it as a general caching mechanism (not just for datasets).


allennlp/common/file_utils.py
```diff
@@ -22,7 +22,11 @@
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 CACHE_ROOT = Path(os.getenv('ALLENNLP_CACHE_ROOT', Path.home() / '.allennlp'))
```
schmmd: Ideally this would have been named "CONFIG_ROOT" or something.

joelgrus:
If it were me I would probably check if the old cache directory exists and is non-empty, and if that's the case, logger a warning that the cache dir moved and that the person should delete it.

Otherwise you've got GB of data sitting on disk that you'll never know to clean up.

schmmd:
@joelgrus where would you put such a check?  `allennlp/commands/__init__.py`?

joelgrus:
it might make more sense to put the check in `file_utils.py` so that it runs if you import that module. that's a little bit weird, but at least then the check is *conceptually* in the right place

schmmd:
Thanks--that's exactly the right spot!

PR #2001, opened 2018-11-01
Add BLEU metric

epwalsh:
I thought I should share my implementation of BLEU. This is the first time I've implemented it, so there may be a more efficient way, but I'm pretty sure the math checks out at least. If anyone has thoughts on how to improve it please let me know!

### DIFF
--- a/allennlp/tests/training/metrics/bleu_test.py
+++ b/allennlp/tests/training/metrics/bleu_test.py
@@ -14,12 +14,12 @@ class BleuTest(AllenNlpTestCase):
 
     def setUp(self):
         super().setUp()
-        self.metric = BLEU(ngram_weights=(0.5, 0.5))
+        self.metric = BLEU(ngram_weights=(0.5, 0.5), exclude_indices={0})
 
     def test_get_valid_tokens_mask(self):
         tensor = torch.tensor([[1, 2, 3, 0],
                                [0, 1, 1, 0]])
-        result = self.metric._get_valid_tokens_mask(tensor, set((0,)))
+        result = self.metric._get_valid_tokens_mask(tensor)
         result = result.long().numpy()
         check = np.array([[1, 1, 1, 0],
                           [0, 1, 1, 0]])
@@ -29,22 +29,22 @@ class BleuTest(AllenNlpTestCase):
         tensor = torch.tensor([1, 2, 3, 1, 2, 0])
 
         # Unigrams.
-        counts = Counter(self.metric._ngrams(tensor, 1, set((0,))))
+        counts = Counter(self.metric._ngrams(tensor, 1))
         unigram_check = {(1,): 2, (2,): 2, (3,): 1}
         assert counts == unigram_check
 
         # Bigrams.
-        counts = Counter(self.metric._ngrams(tensor, 2, set((0,))))
+        counts = Counter(self.metric._ngrams(tensor, 2))
         bigram_check = {(1, 2): 2, (2, 3): 1, (3, 1): 1}
         assert counts == bigram_check
 
         # Trigrams.
-        counts = Counter(self.metric._ngrams(tensor, 3, set((0,))))
+        counts = Counter(self.metric._ngrams(tensor, 3))
         trigram_check = {(1, 2, 3): 1, (2, 3, 1): 1, (3, 1, 2): 1}
         assert counts == trigram_check
 
         # ngram size too big, no ngrams produced.
-        counts = Counter(self.metric._ngrams(tensor, 7, set((0,))))
+        counts = Counter(self.metric._ngrams(tensor, 7))
         assert counts == {}
 
     def test_bleu_computed_correctly(self):
@@ -60,9 +60,7 @@ class BleuTest(AllenNlpTestCase):
                                      [1, 0, 0],
                                      [1, 1, 2]])
 
-        exclude_indices = set((0,))
-
-        self.metric(predictions, gold_targets, exclude_indices)
+        self.metric(predictions, gold_targets)
 
         assert self.metric._prediction_lengths == 6
         assert self.metric._reference_lengths == 5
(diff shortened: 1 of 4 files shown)


allennlp/tests/training/metrics/bleu_test.py
```diff
@@ -0,0 +1,110 @@
...
+                          [0, 1, 1, 0]])
+        np.testing.assert_array_equal(result, check)
+
+    def test_ngrams(self):
+        tensor = torch.tensor([1, 2, 3, 1, 2, 0])
+
+        # Unigrams.
+        counts = Counter(self.metric._ngrams(tensor, 1, set((0,))))
+        unigram_check = {(1,): 2, (2,): 2, (3,): 1}
+        assert counts == unigram_check
```
42B: I'm curious about the use of bare assert statements inside of essentially a `unittest.TestCase` class?

```python
import unittest


class Test(unittest.TestCase):
    def test_bare(self):
        a = [1, 2, 3]
        b = [3, 4, 5]
        assert a == b

    def test_self_assert(self):
        a = [1, 2, 3]
        b = [3, 4, 5]
        self.assertEqual(a, b)


if __name__ == '__main__':
    unittest.main()

```

The output is much more descriptive by using (in this particular case) `self.assertEqual`?

```
FF
======================================================================
FAIL: test_bare (__main__.Test)
----------------------------------------------------------------------
  File "", line 8, in test_bare
    assert a == b
AssertionError

======================================================================
FAIL: test_self_assert (__main__.Test)
----------------------------------------------------------------------
  File "", line 13, in test_self_assert
    self.assertEqual(a, b)
AssertionError: Lists differ: [1, 2, 3] != [3, 4, 5]

First differing element 0:
1
3

- [1, 2, 3]
+ [3, 4, 5]

----------------------------------------------------------------------
Ran 2 tests in 0.001s

FAILED (failures=2)
```

epwalsh: Given that we're using pytest, I think the output from a raw assert failure is actually just as, if not more, descriptive:

Here's an `assertDictEqual` failure:
assertdictequal
Here's a bare assert failure:
base_assert

42B: @epwalsh 

I'm curious why you would still write tests using the old unittest style then? What's the advantage?

matt-gardner:
@42B, we use pytest.  See https://docs.pytest.org/en/latest/.

@epwalsh, thanks, it'd be awesome to have BLEU in here!  We're mostly all at EMNLP this week, though, so it'll be a little bit before we can pick this up.

epwalsh:
@matt-gardner no worries! I enjoyed the slides for the writing code for NLP talk by the way 👏

maxdel:
Hi @epwalsh! 

So far I was wrapping [BLEU from nltk](https://www.nltk.org/api/nltk.translate.html#nltk.translate.bleu_score.sentence_bleu), since the package is anyway in allennlp's requirements.

Is there any reason about not using that one?

allennlp/training/metrics/bleu.py
```diff
@@ -0,0 +1,138 @@
...
+            for tensor_slice in tensor[start_position:].split(ngram_size, dim=-1):
+                if tensor_slice.size(-1) < ngram_size:
+                    break
+                ngram = tuple(x.item() for x in tensor_slice)
+                if exclude_indices and any(x in ngram for x in exclude_indices):
+                    continue
+                ngram_counts[ngram] += 1
+        return ngram_counts
+
+    def _get_modified_precision(self,
```
brendan-ai2: Please add a brief comment here.

allennlp/training/metrics/bleu.py
```diff
@@ -0,0 +1,138 @@
...
+                exclude_indices: Set[int] = None) -> Dict[Tuple[int, ...], int]:
+        ngram_counts: Dict[Tuple[int, ...], int] = Counter()
+        if ngram_size > tensor.size(-1):
+            return ngram_counts
+        for start_position in range(ngram_size):
+            for tensor_slice in tensor[start_position:].split(ngram_size, dim=-1):
+                if tensor_slice.size(-1) < ngram_size:
+                    break
+                ngram = tuple(x.item() for x in tensor_slice)
+                if exclude_indices and any(x in ngram for x in exclude_indices):
```
brendan-ai2: This shouldn't really matter given the size of the ngrams and `exclude_indices`, but could you swap `exclude_indices` and `ngram` so that `exclude_indices` has a lookup performed against it instead of being iterated over? Slightly more intuitive that way.

allennlp/tests/training/metrics/bleu_test.py
brendan-ai2: I agree with you, @epwalsh. Bare assert seems more descriptive. It's also great to just compose raw Python instead of use custom functions.

epwalsh:
@brendan-ai2 thanks for the review! I'll try and address your comments soon. And yes, I think you're right that the sentence brevity penalty would be wrong if we applied the NLTK implementation by batch.

epwalsh:
I think I addressed all your comments, @brendan-ai2, but let me know if you see anything else! Also, do you think it would be reasonable to make this the default metric for `simple_seq2seq`? That way we could get rid of this "TODO" comment: https://github.com/allenai/allennlp/blob/master/allennlp/models/encoder_decoders/simple_seq2seq.py#L213

brendan-ai2:
Thanks for all the updates! Looks good overall, but mypy needs some help: http://build.allennlp.org/viewLog.html?buildId=7541&buildTypeId=AllenNLP_AllenNLPPullRequests&tab=buildLog&_focus=3050

LGTM after that!

I'd be supportive of adding BLEU as a metric for `simple_seq2seq` too. We'd probably still want to calculate the cross entropy during validation, however -- for comparison against training and to use as a stopping condition. Preferably make it configurable so that a user can disable if they don't want to pay the overhead.

epwalsh:
Oops, just fixed! And I agree with you on `simple_seq2seq`. I could submit a PR for that after this gets merged. Thanks!

brendan-ai2:
And merged. Thanks again for the contribution! If you do modify `simple_seq2seq`, feel free to assign me as reviewer.

PR #2003, opened 2018-11-01
Support stdin for prediction jsonl.

schmmd:


schmmd:
@joelgrus I somewhat hate this change but it does enable what I wanted:

```
cat mc-examples.jsonl | allennlp  predict bidaf.tar.gz -
```

I could simplify our examples on https://allennlp.org/models by not requiring an intermediate file and make simpler Docker prediction one-liners (presently they would be a mess).

However, I'd really like to use [argparse.FileType](https://docs.python.org/3/library/argparse.html#filetype-objects).  Unfortunately this opens the file immediately, so we can't use the dataset readers (see [this](https://github.com/allenai/allennlp/blob/master/allennlp/commands/predict.py#L162) and [the dataset reader read method](https://github.com/allenai/allennlp/blob/master/allennlp/data/dataset_readers/dataset_reader.py#L79) and [an implementation](https://github.com/allenai/allennlp/blob/master/allennlp/data/dataset_readers/sequence_tagging.py#L54)).

Ideally the dataset readers would take a `TextIOWrapper` or something--so they could handle different types of input.  This way they wouldn't all need to handle cache resolution either.

allennlp/commands/predict.py
```diff
@@ -151,12 +151,22 @@ def _maybe_print_to_console_and_file(self,
...
+            if self._input_file == "-":
+                source = sys.stdin
+            else:
+                source = open(self._input_file)
+
+            for line in source:
+                if not line.isspace():
+                    yield self._predictor.load_line(line)
+        finally:
+            source.close()
```
joelgrus: I feel like this would be more readable if we didn't try to handle both cases at the same time. e.g.

```
            if self._input_file == "-":
                # read from stdin
                for line in sys.stdin:
                    if not line.isspace():
                        yield self._predictor.load_line(line)
            else:
                # read from file
                with open(self._input_file) as source:
                    for line in source:
                        if not line.isspace():
                            yield self._predictor.load_line(line)
```

schmmd:
@joelgrus approved?

PR #2008, opened 2018-11-03
Fix ArrayField.to_tensor not working with a scalar

bryant1410:


allennlp/tests/data/fields/array_field_test.py
```diff
@@ -57,3 +57,7 @@ def test_padding_handles_list_fields_with_padding_values(self):
     def test_printing_doesnt_crash(self):
         array = ArrayField(numpy.ones([2, 3]), padding_value=-1)
         print(array)
+
+    def test_as_tensor_works_with_scalar(self):
+        array = ArrayField(numpy.asarray(42))
+        array.as_tensor(array.get_padding_lengths())
```
matt-gardner: Can you add an `assert` here, so that this test is actually checking something (instead of just that it doesn't crash)?

bryant1410:
> Thanks for this @bryant1410! I assume the test that you added fails before you make the change in the code?

Yeah, I thought the CI would run the first commit (which just have the test). You can test it yourself.

allennlp/tests/data/fields/array_field_test.py
bryant1410: yup

Issue #2020, opened 2018-11-07
Length mismatch after openai text_field_embedder


tingkai-zhang:
**Describe the bug**
Hi, I open this new issue to provide details.
I am implementing [OpenaiGPT for classification task](https://github.com/tingkai-zhang/Generative-Pre-trained-Transformer) using allennlp.

The output shape of `text field embedder` doesn't match with the input.

After text of shape [batch_size, num_timestamp, num_embedding] is forward after [text_filed_embedder](). The number of timestamps shrinks, which differs from the original GPT pytorch implementation.

I check the source code, seems that it returns [batch_size, seq_length, num_embedding].

And the problem comes that when i [add a classifier head](https://github.com/tingkai-zhang/Generative-Pre-trained-Transformer/blob/33e837355973a47db0d2dcf3bf0391aab03c338b/my_library/modules/Heads/ClfHead.py#L18) on top of transformer embedder.
It requires the `h` which is the  output of `text field embedder` has the same number of timestamps as the input

**To Reproduce**

[the project repo](https://github.com/tingkai-zhang/Generative-Pre-trained-Transformer)

The lines of code uses [text field embedder](https://github.com/tingkai-zhang/Generative-Pre-trained-Transformer/blob/33e837355973a47db0d2dcf3bf0391aab03c338b/my_library/models/GPT.py#L55)

[Classifier head](https://github.com/tingkai-zhang/Generative-Pre-trained-Transformer/blob/33e837355973a47db0d2dcf3bf0391aab03c338b/my_library/modules/Heads/ClfHead.py#L18)

**System (please complete the following information):**
 - OS:  Linux
 - Python version:  3.6
 - AllenNLP version:  v0.7.1

**Additional context**

joelgrus:
this is by design -- the input to the embedder is the byte pairs (which is in general a longer sequence than the tokens themselves), but the output is one vector per token, which we get by taking the embeddings just for the last byte pair for each token:

https://github.com/allenai/allennlp/blob/master/allennlp/modules/token_embedders/openai_transformer_embedder.py#L89

tingkai-zhang:
> this is by design -- the input to the embedder is the byte pairs (which is in general a longer sequence than the tokens themselves), but the output is one vector per token, which we get by taking the embeddings just for the last byte pair for each token:
> 
> https://github.com/allenai/allennlp/blob/master/allennlp/modules/token_embedders/openai_transformer_embedder.py#L89


[...]

PR #2023, opened 2018-11-07
Fix the sequence lengths mismatch

tingkai-zhang:
The sequence lengths of input and output should match each other.
Because `do_layer_norm=False` in self._scalar_mix, line 87 should just pass `layer_activations` according to api documentation.

### DIFF
--- a/allennlp/modules/token_embedders/openai_transformer_embedder.py
+++ b/allennlp/modules/token_embedders/openai_transformer_embedder.py
@@ -84,17 +84,6 @@ class OpenaiTransformerEmbedder(TokenEmbedder):
         if self._top_layer_only:
             mix = layer_activations[-1]
         else:
-            mix = self._scalar_mix(layer_activations, byte_pairs_mask)
+            mix = self._scalar_mix(layer_activations)
 
-        # These embeddings are one per byte-pair, but we want one per original _word_.
-        # So we choose the embedding corresponding to the last byte pair for each word,
-        # which is captured by the ``offsets`` input.
-        if offsets is not None:
-            range_vector = get_range_vector(batch_size, device=get_device_of(mix)).unsqueeze(1)
-            last_byte_pair_embeddings = mix[range_vector, offsets]
-        else:
-            # allow to return all byte pairs by passing no offsets
-            seq_len = (byte_pairs_mask > 0).long().sum(dim=1).max()
-            last_byte_pair_embeddings = mix[:, :seq_len]
-
-        return last_byte_pair_embeddings
+        return mix


tingkai-zhang:
> the current behavior is correct (or, if it's not, you haven't convinced me of it)
> 
> The point of a `TokenEmbedder` is to produce a vector embedding per text-input token. In most cases you're right that the input to the TokenEmbedder = output of the TokenIndexer is also one indexed vector per text-input token. However, for things like byte pair encoding (which is what we have here) the output of the token indexer is one indexed vector per byte-pair, which is (usually) more than the number of tokens. In which case we shouldn't expect the output of the TokenEmbedder to have the same sequence length as its input.
> 
> It sort of seems like you want to use the OpenAITransformer but not as a `TokenEmbedder`? In that case maybe just use `OpenAITransformer` directly?

Thank you for replying!

I think you are talking about `OpenaiTransformerBytePairIndexer` 's job. It tokenize text sentence which makes the input and output has different sequence length.(because word to be tokenized BPE representation, which has more tokens)

I am using `OpenAITransformer ` as `TokenEmbedder`

```
    openai_transformer = OpenaiTransformer(model_path='openai-transformer-lm-2018.07.23.tar.gz',
                                           n_ctx=512,
                                           n_special=2)

    token_embedders = {"openai_transformer_embedder": 
   OpenaiTransformerEmbedder(openai_transformer)}

```

First, the idea of taking the last byte-pair hidden state to represent the token doesn't make sense to me because that will lose the token(sentence) information, which will make it not be able to reproduce the experiment result in original paper. It is not the idea in original implementation neither([tensorflow](https://github.com/openai/finetune-transformer-lm/blob/ae7e86b55254339eba3929f786b04e82602cd51d/train.py#L172) and [pytorch](https://github.com/huggingface/pytorch-openai-transformer-lm/blob/eafc28abdfadfa0732f03a0fc65805c5bfb2ffe7/model_pytorch.py#L170)).



Second, the input **is** one index per byte-pair because you have to use `OpenaiTransformerBytePairIndexer`, which tokenize the sentence to byte-type pair and index them.


```
    tokenizer = WordTokenizer(word_splitter=OpenAISplitter(),
                              start_tokens=[START_TOKEN],
                              end_tokens=[END_TOKEN])

    token_indexer = OpenaiTransformerBytePairIndexer(n_ctx=512,
                                                     model_path='openai-transformer-lm-2018.07.23.tar.gz',
                                                     tokens_to_add=[START_TOKEN, END_TOKEN]
                                                     )
```

Third, I can still to find the index of last token(here i mean BPE represetation),  because the last token is the special token ['_classify_`], whose BPE has length 1. But returning last BPE of each word( i mean, word like 'apple' not 'app', 'le</w>') is not correct. Consequently, the mask returned in [openai byte pair indexer](https://github.com/allenai/allennlp/blob/68cbfb800fbf173fd6a1c097c6dd73530065ae1d/allennlp/data/token_indexers/openai_transformer_byte_pair_indexer.py#L228) is not correct.
Check #2020  


As a conclusion, despite the truncation things in [openai_token_embedder](https://github.com/allenai/allennlp/blob/68cbfb800fbf173fd6a1c097c6dd73530065ae1d/allennlp/modules/token_embedders/openai_transformer_embedder.py#L98)
 and all the padding things,

tingkai-zhang:
Initially, all i want is the input and output of Text field Embedder has same number of timestamps, which what the original implementation does.
After reading the code, i think your allennlp implementation is not correct(to me).

For logical level,  taking only last byte-pair representation for each word will lose information.
For code level, it  makes downstream tasks difficult to expand because lengths mismatch.



##########################################################
I am not sure if you have reproduced the same result in the paper. The result may be correct, but that's because the original implementation only takes the last byte pair representation of each **sentence**. And here, we can also get the last byte pair representation of each **sentence**.

matt-gardner:
The entire point of the `OpenaiTransformerEmbedder` is to get back a sequence of vectors for each token from the byte-pair sequence.  So, if you want something that doesn't do that, `OpenaiTransformerEmbedder` is not what you are looking for.  The `OpenaiTransformerEmbedder` does very little on top of the `OpenaiTransformer` - you probably should just be using that directly.

PR #2029, opened 2018-11-08
Catch nan loss in training loop

### LINKED ISSUE #2028: Bug when torch.topk called on nan's, produces strange results
**Describe the bug**
This is due to this known inconsistency on PyTorch: https://github.com/pytorch/pytorch/issues/1810. I noticed this bug when using beam search on a validation set, after the loss had exploded. Due to the loss exploding, the `step` function being passed to `BeamSearch` was producing `nan`'s for the log probabilities, and then when the beam search called the `topk` function, the top indices ending up being massive values on the order of `9e+18`.. way out of bounds obviously which caused issues downstream.

This isn't really a bug with anything in AllenNLP, since the model shouldn't be passing in `nan` values to BeamSearch, but I still think BeamSearch should raise an error in this case so the issue is caught immediately. (This took me a couple hours to debug because I was unaware of this issue with PyTorch, and therefore didn't realize it was stemming from the output of beam search)

**To Reproduce**
Use `torch.topk` on a tensor filled with `nan` values on GPU.

**Expected behavior**
I think BeamSearch should raise error instead of returning indices that are out of bound. Maybe it's worth having a helper function in `nn.utils` for a "safe topk" function, because I'm sure `topk` is being called in a lot of other places besides `nn.beam_search`. I'd be willing to submit a PR here depending on what you think the best way to handle this is.

epwalsh:
Fixes #2028

### DIFF
--- a/allennlp/common/checks.py
+++ b/allennlp/common/checks.py
@@ -39,10 +39,16 @@ def check_dimensions_match(dimension_1: int,
 
 
 def check_for_gpu(device_id: int):
-    if device_id is not None and device_id >= cuda.device_count():
-        raise ConfigurationError("Experiment specified a GPU but none is available;"
-                                 " if you want to run on CPU use the override"
-                                 " 'trainer.cuda_device=-1' in the json config file.")
+    if device_id is not None:
+        devices_available = cuda.device_count()
+        if not devices_available:
+            raise ConfigurationError("Experiment specified a GPU but none are available;"
+                                     " if you want to run on CPU use the override"
+                                     " 'trainer.cuda_device=-1' in the json config file.")
+        elif device_id >= devices_available:
+            raise ConfigurationError(f"Experiment specified GPU device {device_id}"
+                                     f" but there are only {devices_available} devices "
+                                     f" available.")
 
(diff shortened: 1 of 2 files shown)


epwalsh:
Also gives a more descriptive error message when the requested GPU device is not available

allennlp/common/checks.py
```diff
@@ -39,10 +39,16 @@ def check_dimensions_match(dimension_1: int,
 
 
 def check_for_gpu(device_id: int):
-    if device_id is not None and device_id >= cuda.device_count():
-        raise ConfigurationError("Experiment specified a GPU but none is available;"
-                                 " if you want to run on CPU use the override"
-                                 " 'trainer.cuda_device=-1' in the json config file.")
+    if device_id is not None:
+        devices_available = cuda.device_count()
+        if not devices_available:
```
joelgrus: two readability nits:

I would call the variable `num_available` or something else indicating it's a count 

and also I would be explicit that we're checking for zero rather than just falsiness:

```
if num_available == 0:
```

epwalsh: Fixed, thanks!

PR #2036, opened 2018-11-10
Sentence splitter

kernelmachine:
This pull request adds sentence segmentation to AllenNLP. It's just a wrapper on Spacy's rule-based and statistical sentence boundary detection.

### DIFF
--- a/allennlp/data/tokenizers/sentence_splitter.py
+++ b/allennlp/data/tokenizers/sentence_splitter.py
@@ -1,0 +1,84 @@
+from typing import List
+from overrides import overrides
+from allennlp.common import Registrable
+from allennlp.common.util import get_spacy_model
+from allennlp.data.tokenizers.token import Token
+
+
+class SentenceSplitter(Registrable):
+    """
+    A ``SentenceSplitter`` splits strings into sentences.  This is typically called a "tokenizer" in NLP,
+    because splitting strings into characters is trivial, but we use ``Tokenizer`` to refer to the
+    higher-level object that splits strings into tokens (which could just be sentence tokens).
+    So, we're using "sentence splitter" here for this.
+    """
+    default_implementation = 'spacy_rulebased'
+
+    def batch_split_sentences(self, texts: List[str]) -> List[List[Token]]:
+        """
+        Spacy needs to do batch processing, or it can be really slow.  This method lets you take
+        advantage of that if you want.  Default implementation is to just iterate of the sentences
+        and call ``split_sentences``, but the ``SpacySentenceSplitter`` will actually do batched
+        processing.
+        """
+        return [self.split_sentences(text) for text in texts]
+
+    def split_sentences(self, text: str) -> List[Token]:
+        """
+        Splits ``texts`` into a list of :class:`Token` objects.
+        """
+        raise NotImplementedError
+
+@SentenceSplitter.register('spacy_statistical')
+class SpacyStatisticalSentenceSplitter(SentenceSplitter):
+    """
+    A ``SentenceSplitter`` that uses spaCy's statistical sentence boundary detection.
+    More accurate but slower, since it uses a dependency parses to detect sentence boundaries.
+    """
+    def __init__(self,
+                 language: str = 'en_core_web_sm',
+                 pos_tags: bool = False,
+                 parse: bool = False,
+                 ner: bool = False) -> None:
+        self.spacy = get_spacy_model(language, pos_tags, parse, ner)
+
+    @overrides
+    def split_sentences(self, text: str) -> List[Token]:
+        return [sent.string.strip() for sent in self.spacy(text).sents]
+
+    @overrides
+    def batch_split_sentences(self, texts: List[str]) -> List[List[Token]]:
+        def extract_sentences(doc):
+            return [sent.string.strip() for sent in doc.sents]
+
+        return [extract_sentences(doc)
+                for doc in self.spacy.pipe(texts, n_threads=-1)]
+
+
+@SentenceSplitter.register('spacy_rulebased')
+class SpacyRuleBasedSentenceSplitter(SentenceSplitter):
+    """
+    A ``SentenceSplitter`` that uses spaCy's rule-based sentence boundary detection.
+    Faster and less memory footprint, since it uses punctuation to detect sentence boundaries.
+    """
+    def __init__(self,
+                 language: str = 'en_core_web_sm',
+                 pos_tags: bool = False,
+                 parse: bool = False,
+                 ner: bool = False) -> None:
+        self.spacy = get_spacy_model(language, pos_tags, parse, ner)
+        if not self.spacy.has_pipe('sbd'):
+            sbd = self.spacy.create_pipe('sbd')   # or: nlp.create_pipe('sbd')
+            self.spacy.add_pipe(sbd)
+
+    @overrides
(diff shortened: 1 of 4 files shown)


allennlp/data/tokenizers/sentence_splitter.py
```diff
@@ -0,0 +1,81 @@
+from typing import List
+from overrides import overrides
+from allennlp.common import Registrable
+from allennlp.common.util import get_spacy_model
+from allennlp.data.tokenizers.token import Token
+
+
+class SentenceSplitter(Registrable):
+    """
+    A ``SentenceSplitter`` splits strings into sentences. There are two variants of sentence splitting,
```
matt-gardner: I wouldn't hard-code that there are two variants, because that will eventually become stale and we'll forget to update this.

allennlp/data/tokenizers/sentence_splitter.py
```diff
@@ -0,0 +1,81 @@
...
+    one that uses rules (ie punctuation), and another that uses a dependency parse of the text to infer
+    sentence boundaries.
+    """
+    default_implementation = 'spacy_rulebased'
+
+    def batch_split_sentences(self, texts: List[str]) -> List[List[Token]]:
+        """
+        This method lets you take advantage of spacy's batch processing.  
+        Default implementation is to just iterate of the sentences and call ``split_sentences``.
```
matt-gardner: "iterate over"?  And "texts" instead of "sentences"?

allennlp/data/tokenizers/sentence_splitter.py
```diff
@@ -0,0 +1,81 @@
...
+        """
+        return [self.split_sentences(text) for text in texts]
+
+    def split_sentences(self, text: str) -> List[Token]:
```
matt-gardner: Is this really a `Token` object?  Should it be?  I'm pretty sure it's not - it's a sentence, not a token.  I'd expect this to be `List[str]`.  Is there some additional information from spacy that you're hoping to keep around that makes you use `Token` here?

allennlp/data/tokenizers/sentence_splitter.py
```diff
@@ -0,0 +1,81 @@
...
+        return [sent.string.strip() for sent in self.spacy(text).sents]
+
+    @overrides
+    def batch_split_sentences(self, texts: List[str]) -> List[List[Token]]:
+        def extract_sentences(doc):
+            return [sent.string.strip() for sent in doc.sents]
+
+        return [extract_sentences(doc)
+                for doc in self.spacy.pipe(texts, n_threads=-1)]
```
matt-gardner: What does `n_threads=-1` do?  Let spacy decide how many to use based on the number of cores available?

allennlp/data/tokenizers/sentence_splitter.py
```diff
@@ -0,0 +1,81 @@
...
+        Default implementation is to just iterate of the sentences and call ``split_sentences``.
+        """
+        return [self.split_sentences(text) for text in texts]
+
+    def split_sentences(self, text: str) -> List[Token]:
+        """
+        Splits ``texts`` into a list of :class:`Token` objects.
+        """
+        raise NotImplementedError
+
```
matt-gardner: Two blank lines between classes.

allennlp/data/tokenizers/sentence_splitter.py
```diff
@@ -0,0 +1,81 @@
...
+@SentenceSplitter.register('spacy_statistical')
+class SpacyStatisticalSentenceSplitter(SentenceSplitter):
+    """
+    A ``SentenceSplitter`` that uses spaCy's statistical sentence boundary detection.
+    More accurate but slower, since it uses a dependency parse to detect sentence boundaries.
+    """
+    def __init__(self,
+                 language: str = 'en_core_web_sm',
+                 pos_tags: bool = False,
+                 parse: bool = False,
```
matt-gardner: If this uses a dependency parse, how does it work if `parse=False`?

allennlp/data/tokenizers/sentence_splitter.py
```diff
@@ -0,0 +1,81 @@
...
+    def batch_split_sentences(self, texts: List[str]) -> List[List[Token]]:
+        def extract_sentences(doc):
+            return [sent.string.strip() for sent in doc.sents]
+
+        return [extract_sentences(doc)
+                for doc in self.spacy.pipe(texts, n_threads=-1)]
+
+
+@SentenceSplitter.register('spacy_rulebased')
+class SpacyRuleBasedSentenceSplitter(SentenceSplitter):
```
matt-gardner: These classes look almost identical.  Looks like the only difference is the `has_pipe('sbd')` bit.  I'd imagine you could just make that an option in a single `SpacySentenceSplitter` class.

allennlp/data/tokenizers/sentence_splitter.py
```diff
@@ -0,0 +1,81 @@
...
+    Faster and smaller memory footprint, since it uses punctuation to detect sentence boundaries.
+    """
+    def __init__(self,
+                 language: str = 'en_core_web_sm',
+                 pos_tags: bool = False,
+                 parse: bool = False,
+                 ner: bool = False) -> None:
+        self.spacy = get_spacy_model(language, pos_tags, parse, ner)
+        if not self.spacy.has_pipe('sbd'):
+            sbd = self.spacy.create_pipe('sbd')   # or: nlp.create_pipe('sbd')
```
matt-gardner: I don't think this comment is adding anything.  A comment that _would_ add something would be to say that `'sbd'` is a built-in part of spacy that does rule-based sentence splitting (if that's actually true).

allennlp/data/tokenizers/sentence_splitter.py
kernelmachine: Yeah, this is a typo, should be List[str].

allennlp/data/tokenizers/sentence_splitter.py
kernelmachine: Yep. this is actually default, so i'll remove explicitly calling it.

allennlp/data/tokenizers/sentence_splitter.py
kernelmachine: It does fail if parse is False, I changed this so it sets `parse=True` if the user sets `rule_based=False`

kernelmachine:
No worries at all, thanks for your detailed review. Addressed all your comments, let me know if you see any other place for improvements!

allennlp/data/tokenizers/sentence_splitter.py
```diff
@@ -0,0 +1,65 @@
...
+                self.spacy.add_pipe(sbd)
+
+    @overrides
+    def split_sentences(self, text: str) -> List[str]:
+        return [sent.string.strip() for sent in self.spacy(text).sents]
+
+    @overrides
+    def batch_split_sentences(self, texts: List[str]) -> List[List[str]]:
+        def extract_sentences(doc):
+            return [sent.string.strip() for sent in doc.sents]
```
matt-gardner: Defining a closure here that doesn't actually use any variables in the current scope seems a bit heavy (and error-prone in future modifications), where you could just do:

```python
return [sentence.string.strip() for sentence in doc.sents for doc in self.spacy.pipe(texts)]
```

PR #2037, opened 2018-11-10
Dynamic stopwords

kernelmachine:
This pull request adds support for: 

- Adding stopwords from a file.
- Using a regex-based word filter.

### DIFF
--- a/allennlp/data/tokenizers/word_filter.py
+++ b/allennlp/data/tokenizers/word_filter.py
@@ -45,9 +45,9 @@ class RegexFilter(WordFilter):
 
     @overrides
     def filter_words(self, words: List[Token]) -> List[Token]:
-        bad_tokens = [word for word in words
+        stopwords = [word for word in words
                       if re.match(pattern=self._joined_pattern, string=word.text)]
-        return [word for word in words if word not in bad_tokens]
+        return [word for word in words if word not in stopwords]
 
 @WordFilter.register('stopwords')
 class StopwordFilter(WordFilter):
--- a/allennlp/data/tokenizers/word_filter.py
+++ b/allennlp/data/tokenizers/word_filter.py
@@ -1,4 +1,4 @@
-from typing import List, Set
+from typing import List
 import re
 from overrides import overrides
 
@@ -70,8 +70,8 @@ class StopwordFilter(WordFilter):
         A list of tokens to additionally filter out.
     """
     def __init__(self,
-                 stopword_file: str=None,
-                 tokens_to_add: List[str]=None) -> None:
+                 stopword_file: str = None,
+                 tokens_to_add: List[str] = None) -> None:
         self._tokens_to_add = tokens_to_add or []
         if stopword_file is not None:
             self.stopwords = read_set_from_file(stopword_file)
--- a/allennlp/data/tokenizers/word_filter.py
+++ b/allennlp/data/tokenizers/word_filter.py
@@ -4,6 +4,7 @@ from overrides import overrides
 
 from allennlp.common import Registrable
 from allennlp.data.tokenizers.token import Token
+from allennlp.common.file_utils import read_set_from_file
 
 
 class WordFilter(Registrable):
@@ -41,8 +42,7 @@ class RegexFilter(WordFilter):
     Parameters
     ----------
     patterns : ``List[str]``
-        List of regex patterns to filter by.
-
+        Words matching these regex patterns will be removed as stopwords.
     """
     def __init__(self,
                  patterns: List[str]) -> None:
@@ -65,16 +65,16 @@ class StopwordFilter(WordFilter):
     Parameters
     ----------
     stopword_file : ``str``, optional
-        A filename containing stopwords to filter by (file format is one stopword per line).
+        A filename containing stopwords to filter out (file format is one stopword per line).
     tokens_to_add : ``List[str]``, optional
-        A list of tokens to additionally filter by.
+        A list of tokens to additionally filter out.
     """
     def __init__(self,
                  stopword_file: str=None,
                  tokens_to_add: List[str]=None) -> None:
         self._tokens_to_add = tokens_to_add or []
         if stopword_file is not None:
-            self.stopwords = self._set_from_file(stopword_file)
+            self.stopwords = read_set_from_file(stopword_file)
         else:
             self.stopwords = set(['I', 'a', 'aboard', 'about', 'above', 'accordance', 'according',
                                   'across', 'after', 'against', 'along', 'alongside', 'also', 'am',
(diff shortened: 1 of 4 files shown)


allennlp/data/tokenizers/word_filter.py
```diff
@@ -33,38 +33,71 @@ def filter_words(self, words: List[Token]) -> List[Token]:
...
+                 patterns: List[str]) -> None:
+        self._patterns = patterns
+        self._joined_pattern = "|".join(self._patterns)
+
+    @overrides
+    def filter_words(self, words: List[Token]) -> List[Token]:
+        stopwords = [word for word in words
+                      if re.match(pattern=self._joined_pattern, string=word.text)]
+        return [word for word in words if word not in stopwords]
+
```
matt-gardner: Two blank lines between classes.

allennlp/data/tokenizers/word_filter.py
```diff
@@ -33,38 +33,71 @@ def filter_words(self, words: List[Token]) -> List[Token]:
...
 
+@WordFilter.register('regex')
+class RegexFilter(WordFilter):
+    """
+    Uses a list of regexes to filter
+    """
+    def __init__(self,
+                 patterns: List[str]) -> None:
+        self._patterns = patterns
+        self._joined_pattern = "|".join(self._patterns)
```
matt-gardner: I don't use regexes much - is it faster to pre-compile this somehow?

allennlp/data/tokenizers/word_filter.py
```diff
@@ -33,38 +33,71 @@ def filter_words(self, words: List[Token]) -> List[Token]:
         return words
 
 
+@WordFilter.register('regex')
+class RegexFilter(WordFilter):
+    """
+    Uses a list of regexes to filter
```
matt-gardner: This needs a better docstring, including listing the parameters with comments describing them.  In particular, do these patterns say what to keep, or what to remove?

allennlp/data/tokenizers/word_filter.py
```diff
@@ -33,38 +33,71 @@ def filter_words(self, words: List[Token]) -> List[Token]:
...
+    def __init__(self,
+                 patterns: List[str]) -> None:
+        self._patterns = patterns
+        self._joined_pattern = "|".join(self._patterns)
+
+    @overrides
+    def filter_words(self, words: List[Token]) -> List[Token]:
+        stopwords = [word for word in words
+                      if re.match(pattern=self._joined_pattern, string=word.text)]
+        return [word for word in words if word not in stopwords]
```
matt-gardner: You're iterating over things twice here.  It's probably better to just have a single loop over the words, check for `re.match`, and append to a list.

allennlp/tests/data/tokenizers/word_tokenizer_test.py
```diff
@@ -27,10 +27,54 @@ def test_batch_tokenization(self):
...
-    def test_stems_and_filters_correctly(self):
+    def test_stems_and_filters_stopwords_correctly(self):
         tokenizer = WordTokenizer.from_params(Params({'word_stemmer': {'type': 'porter'},
                                                       'word_filter': {'type': 'stopwords'}}))
         sentence = "this (sentence) has 'crazy' \"punctuation\"."
         expected_tokens = ["sentenc", "ha", "crazi", "punctuat"]
         tokens = [t.text for t in tokenizer.tokenize(sentence)]
         assert tokens == expected_tokens
+
+    def test_stems_and_filters_stopwords_from_file_correctly(self):
```
matt-gardner: It'd be better to put these tests in a (new) `word_filter_test.py` file, instead of testing the additional functionality indirectly here.

allennlp/data/tokenizers/word_filter.py
kernelmachine: according to [this](https://stackoverflow.com/questions/452104/is-it-worth-using-pythons-re-compile), there is negligible perf difference and it seems to be more of a stylistic concern. Using `re.compile` does look cleaner, so I'll just change it to that.

kernelmachine:
Addressed all your comments @matt-gardner.

allennlp/data/tokenizers/word_filter.py
```diff
@@ -33,37 +33,90 @@ def filter_words(self, words: List[Token]) -> List[Token]:
...
 
+@WordFilter.register('regex')
+class RegexFilter(WordFilter):
+    """
+    A ``RegexFilter`` removes words according to supplied regex patterns.
+
+    Parameters
+    ----------
+    patterns : ``List[str]``
+        List of regex patterns to filter by.
```
matt-gardner: I still feel like I have to think too much to figure out if these regexes will keep words or remove words.  Can you just say directly which it is?  Something like "Words matching these regex patterns will be removed as stopwords."  Also, you don't need the blank line at the end of the docstring.

allennlp/data/tokenizers/word_filter.py
```diff
@@ -33,37 +33,90 @@ def filter_words(self, words: List[Token]) -> List[Token]:
...
-    Uses a list of stopwords to filter.
+    A ``StopwordFilter`` uses a list of stopwords to filter.
+    If no file is specified, a default list of stopwords is used.
+
+    Parameters
+    ----------
+    stopword_file : ``str``, optional
+        A filename containing stopwords to filter by (file format is one stopword per line).
+    tokens_to_add : ``List[str]``, optional
+        A list of tokens to additionally filter by.
```
matt-gardner: I think "filter out" is more clear than "filter by".

allennlp/data/tokenizers/word_filter.py
```diff
@@ -33,37 +33,90 @@ def filter_words(self, words: List[Token]) -> List[Token]:
...
+        for token in self._tokens_to_add:
+            self.stopwords.add(token)
+
+    def _set_from_file(self, stopword_file: str) -> Set[str]:
+        '''
+        Extract stopwords from a file.
+        Expected file format is one stopword per line.
+
+        Returns
+        _______
```
matt-gardner: These are underscores where they should be dashes.

For future reference, though, private methods like this with good, obvious names don't require such detailed docstrings (no harm in having them, but no need to go out of your way to add them, either).

matt-gardner: Except, looking at this name again, I'd expect a method called `self._set_something` to have no return value and to modify the internal state of `self`.  This does neither - it reads the lines from a file and returns a set, without modifying `self`.  That's also why pylint is complaining about this line, saying that there's no use of self.

This method feels like it should be called `read_set_from_file`, and live in `common.file_utils`, not in here.

allennlp/tests/data/tokenizers/word_filter_test.py
```diff
@@ -0,0 +1,47 @@
...
+        assert tokens == expected_tokens
+
+    def test_filters_no_match_correctly(self):
+        filter_ = RegexFilter(patterns=['&'])
+        expected_tokens = ["this", "45", "(", "sentence", ")", "has",  "9943434", '"', "punctuations", '"', '.']
+        tokens = [t.text for t in filter_.filter_words(self.sentence)]
+        assert tokens == expected_tokens
+
+    def test_filters_stopwords_from_file_correctly(self):
+        with open('/tmp/tmp_stopwords.txt', 'w+') as f:
```
matt-gardner: Hard-coding a path like this makes it harder for us if we ever want to support windows.  This is why we have `self.TEST_DIR` on `AllenNlpTestCase`.  You can just call this file `self.TEST_DIR / "stopwords.txt"`, and it'll get removed automatically on teardown.

kernelmachine:
Looks like I'm not authorized to merge - can you merge this (and #2036) when you get the chance?

PR #2049, opened 2018-11-13
calypso transformer

joelgrus:
there's still some potential duplicate code, but it's not *exact* duplication, so it would take some work to factor out the common parts and I don't know that that's a great use of anyone's time.

for the most part this is just a straight port + adding type annotations and minor cleanup

### DIFF
--- a/allennlp/modules/seq2seq_encoders/bidirectional_transformer_encoder.py
+++ b/allennlp/modules/seq2seq_encoders/bidirectional_transformer_encoder.py
@@ -2,21 +2,17 @@
 The BidirectionalTransformerEncoder from calypso
 """
 # pylint: disable=arguments-differ,invalid-name
-
-from typing import Tuple,Callable
-import copy
+from typing import Tuple, Callable
 import math
 
 import numpy as np
 import torch
 import torch.nn.functional as F
 
+from allennlp.modules.layer_norm import LayerNorm
 from allennlp.modules.seq2seq_encoders.seq2seq_encoder import Seq2SeqEncoder
+from allennlp.nn import util
 
-
-def clones(module: torch.nn.Module, num_copies: int) -> torch.nn.ModuleList:
-    "Produce N identical layers."
-    return torch.nn.ModuleList([copy.deepcopy(module) for _ in range(num_copies)])
 
 def attention(query: torch.Tensor,
               key: torch.Tensor,
@@ -77,7 +73,7 @@ class TransformerEncoder(torch.nn.Module):
     "Core encoder is a stack of N layers"
     def __init__(self, layer: torch.nn.Module, num_layers: int, return_all_layers: bool = False) -> None:
         super().__init__()
-        self.layers = clones(layer, num_layers)
+        self.layers = util.clone(layer, num_layers)
         self.norm = LayerNorm(layer.size)
         self.return_all_layers = return_all_layers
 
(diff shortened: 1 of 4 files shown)


brendan-ai2:
Process question: Given that this is mostly raw Calypso code, should this be more on the stamp end of the code review spectrum or should I be more detailed?

joelgrus:
if you see something that you think could be improved, then you should point it out.

but if you have questions like "why did you implement X this way?" the answer is likely going to be "that's how it was implemented in calypso and I just copied the code over" 🤷‍♀️

allennlp/modules/seq2seq_encoders/bidirectional_transformer_encoder.py
```diff
@@ -0,0 +1,265 @@
+"""
+The BidirectionalTransformerEncoder from calypso
```
brendan-ai2: I think we should link the annotated transformer tutorial (http://nlp.seas.harvard.edu/2018/04/03/attention.html) here as this implementation appears to be based on that. MIT license conveniently. https://github.com/harvardnlp/annotated-transformer/blob/master/LICENSE

allennlp/modules/seq2seq_encoders/bidirectional_transformer_encoder.py
```diff
@@ -0,0 +1,265 @@
...
+        if self.return_all_layers:
+            all_layers[-1] = self.norm(all_layers[-1])
+            return all_layers
+        return self.norm(x)
+
+
+class SublayerConnection(torch.nn.Module):
+    """
+    A residual connection followed by a layer norm.
+    Note for code simplicity the norm is first as opposed to last.
```
brendan-ai2: Huh. This is different than the original transformer described in "Attention Is All You Need". They use `self.norm(x + self.dropout(sublayer(x)))`. Not sure why that's more complicated. Regardless, since this is how Calypso/Annotated Transformer do it, looks good...

brendan-ai2: Ah, it's simply that they apply the norm later. Not that it's left out.

allennlp/modules/seq2seq_encoders/bidirectional_transformer_encoder.py
```diff
@@ -0,0 +1,265 @@
...
+
+class MultiHeadedAttention(torch.nn.Module):
+    def __init__(self, num_heads: int, input_dim: int, dropout: float = 0.1) -> None:
+        super().__init__()
+        assert input_dim % num_heads == 0, "input_dim must be a multiple of num_heads"
+        # We assume d_v always equals d_k
+        self.d_k = input_dim // num_heads
+        self.num_heads = num_heads
+        self.linears = util.clone(torch.nn.Linear(input_dim, input_dim), 4)
+        self.attn = None
```
brendan-ai2: Appears to be dead code.

allennlp/modules/seq2seq_encoders/bidirectional_transformer_encoder.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+class MultiHeadedAttention(torch.nn.Module):
+    def __init__(self, num_heads: int, input_dim: int, dropout: float = 0.1) -> None:
+        super().__init__()
+        assert input_dim % num_heads == 0, "input_dim must be a multiple of num_heads"
+        # We assume d_v always equals d_k
+        self.d_k = input_dim // num_heads
+        self.num_heads = num_heads
+        self.linears = util.clone(torch.nn.Linear(input_dim, input_dim), 4)
```
brendan-ai2: It's a bit scary to have these stored as a list because fundamentally they're heterogeneous. The first three layers or for projecting the query, key and value respectively and the final layer is for projecting the concatenated heads. Since you're just porting I don't want to request any big changes, but a warning comment here would be great.

allennlp/modules/seq2seq_encoders/bidirectional_transformer_encoder.py
```diff
@@ -0,0 +1,265 @@
...
+                                                return_all_layers=return_all_layers)
+        self._position = PositionalEncoding(input_dim)
+
+        self.input_dim = input_dim
+        self.output_dim = 2 * input_dim
+
+        if input_dropout:
+            self._dropout = torch.nn.Dropout(input_dropout)
+        else:
+            self._dropout = lambda x: x
```
brendan-ai2: For my edification, have we found there to be substantial overhead to `torch.nn.Dropout(0.0)`?

brendan-ai2:
Per Matt's suggestion on Slack, we should probably mark this as AllenNLP internal somehow. In the hopes of consolidating to one transformer implementation soon.

Also, it would be great to get this merged before the break if you have time. :)

joelgrus:
comments addressed, sorry about the delay

brendan-ai2:
Thanks, Joel! No worries at all.

PR #2059, opened 2018-11-15
Fix logging statement so it has the proper scope.

schmmd:
I noticed some root-level log messages for ELMo in the demo logging.  After this change they will have the proper `__name__`.

```
INFO:__main__:loading coreference-resolution model
INFO:__main__:loading named-entity-recognition model
INFO:root:Initializing ELMo
```


matt-peters:
👍

PR #2060, opened 2018-11-15
add flaky decorator + increase tolerance

joelgrus:


matt-peters:
See my comment on #2042 about this test.

joelgrus:
ok, I changed this PR to skip the test (so the build isn't broken) while I fix it

PR #2063, opened 2018-11-16
Add BLEU metric to simple_seq2seq

epwalsh:


epwalsh:
This is a followup to https://github.com/allenai/allennlp/pull/2001, given the discussion @brendan-ai2  and I had. (@brendan-ai2 I tried to request your review but I don't I have the permissions for that)

brendan-ai2:
Oops, sorry about that. I'll take a look!

allennlp/tests/models/encoder_decoders/simple_seq2seq_test.py
```diff
@@ -77,28 +77,3 @@ def test_decode_runs_correctly(self):
         output_dict = self.model(**training_tensors)
         decode_output_dict = self.model.decode(output_dict)
         assert "predicted_tokens" in decode_output_dict
-
-    def test_greedy_decode_matches_beam_search(self):
```
brendan-ai2: I actually quite like having this test to ensure that the beam search and greedy search are in sync. Did you remove it so that `_forward_loop` doesn't need to make predictions? Honestly, I'd say it's worth leaving in the predictions just for this test. I'm open to being convinced otherwise, but I'd need to see some benchmarks showing a large performance hit.

epwalsh: That's a good point. I had initially removed it because I thought it would be confusing to have `_forward_loop` create predictions, only to be overridden by the predictions from beam search. I put that back in though and just added a note that we really only use those predictions for testing

PR #2064, opened 2018-11-16
Improve server_simple by adding predict_batch and adding GPU option

handsomezebra:
Some simple improvements to server_simple.py:
1. Add a predict_batch method so a client can send a list of items for prediction.
2. Add the command line option --cuda-device.

### DIFF
--- a/allennlp/service/server_simple.py
+++ b/allennlp/service/server_simple.py
@@ -27,6 +27,7 @@ from flask_cors import CORS
 from gevent.pywsgi import WSGIServer
 
 from allennlp.common import JsonDict
+from allennlp.common.checks import check_for_gpu
 from allennlp.common.util import import_submodules
 from allennlp.models.archival import load_archive
 from allennlp.predictors import Predictor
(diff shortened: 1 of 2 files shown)


matt-gardner:
What's the use case for batch predictions in the server?  I'm just wondering if this is adding complexity that we don't really need, or if there's a better way to solve the problem here (e.g., just using a `Predictor` in a python script instead of in a server).  It's probably fine, I just want to be sure this is a use case we want to support.  Also, @joelgrus, can you take this one (when you have time to get to it)?

handsomezebra:
Thanks for the comments. The batch prediction and the GPU option are both useful to improve the response time and/or throughput. For example if I want to tag a long document I can split them into segments and tag them in a batch.

matt-gardner:
Well, yes, of course batch prediction makes things faster.  I'm just wondering why you're sending them to a server instead of just running the `Predictor` directly in your script.  Going through a server will add all kinds of overhead (serializing things to json, reading them from json, inter-process communication, etc.).  I'm not sure in what circumstance you'll actually want to have batched input to a server.

Your "if I have a long document I can split them into segments" statement - is that splitting into segments happening in a browser?  Or in a python script?  If it's happening in a python script, you are definitely better off just using a `Predictor` directly.  If it's happening in javascript in a browser, then you have a good use-case for this PR.

handsomezebra:
In my use case one request may go through several steps/models to get a final response. So some upstream components will preprocess the request and generate inputs to a model. Each step or model is a separate service. This is for better isolation and easier deployment of those components. For example I may want to host different services on different machines, using GPU machines for some services but CPU machines for others, and when it's necessary I may need to create multiple instance of the same model and doing load balancing, etc.

PR #2067, opened 2018-11-16
add BERT token embedder 

joelgrus:
this is ready for review. in addition to the included unit tests, I trained two NER models using these embeddings (unfortunately, I realized this morning, I used the _uncased_ BERT model, which seems like a bad idea for NER)

(1) only BERT embeddings: https://beaker-internal.allenai.org/ex/ex_rnk3mcplnpjz/tasks
(2) BERT embeddings + character embeddings: https://beaker-internal.allenai.org/ex/ex_nrq8d5vw5cb2/tasks

(apologies to non-AI2 people for the beaker-internal links)

as discussed offline, because of the positional encodings the BERT embedding has a max sequence length and will crash if you feed it longer sequences. this implementation simply truncates longer sequences and logs a warning. I left a TODO to come up with something better.


SparkJiao:
Very appreciate your work about bert. I'm considering using bert on my task so could I use your implementation now?

joelgrus:
I wouldn't recommend it using it for anything important, the code hasn't been thoroughly tested yet.

(you can try it and tell me how well it works though. 😀)

SparkJiao:
OK, thank u very much. I’ll have a try first. 😊
发件人: Joel Grus<mailto:notifications@github.com>
发送时间: 2018年11月17日 12:59
收件人: allenai/allennlp<mailto:allennlp@noreply.github.com>
抄送: KKnotalone<mailto:jiaofangkai@hotmail.com>; Comment<mailto:comment@noreply.github.com>
主题: Re: [allenai/allennlp] WIP: add BERT token embedder (#2067)


I wouldn't recommend it using it for anything important, the code hasn't been thoroughly tested yet.

(you can try it and tell me how well it works though. 😀)

handsomezebra:
Thanks for all the work! 
I prefer that the BERT code being included in the allennlp library instead of pip install it. One reason is that people may want to experiment with the code using allennlp framework. 
Also, ELMo and BERT in the same place is fun.

SparkJiao:
Unfortunately I'd say that I have find a bug.
While sorting in the batches using the sorting keys where has a dict `{"passage", "num_tokens"}` it is reported that the key num_tokens didn't exist. The bug is because the bert token indexer has made an array of tokens with different lengths with single_id token_indexer and character_tokenizer and the key "num_tokens" will not be added into the dict "padding_lengths".
The config file is dialog_qa.jsonnet and I will try to modified the sorting keys temporarily

SparkJiao:
So are there some problems while using GloVe and singId token characters and bert at the same time? Because bert will make sequences with different lengths from the others? And we may not joint the word embedding ? Sorry to bother you

thomwolf:
Hi @joelgrus, I've released our implementation on pip (see https://github.com/huggingface/pytorch-pretrained-BERT). Sorry for the delay! Tell me if I can change anything to make it easier for you to integrate in AllenNLP!

joelgrus:
the timing is perfect, thanks so much!

joelgrus:
@SparkJiao the indexer produces a `bert-offsets` field that contains the indices of the last wordpiece for each word. if you pass this to the bert token embedder, it will only return the embedding of the last wordpiece per original token, which gives it the right size. (whether this is the optimal way to accomplish this I'm less sure of)

you can do this from your config file by adding something like the following to your `text_field_embedder`:

```
            "allow_unmatched_keys": true,
            "embedder_to_indexer_map": {
                "tokens": ["tokens"],  // etc
                "bert": ["bert", "bert-offsets"]
            },
```

so that the model knows to pass the offsets into the bert token embedder.

SparkJiao:
@joelgrus Well, sorry to reply lately and very appreciate to your help.  I recently met some other problems and I can't solve them so I may wait for your official tutorial:(
Thanks a lot at last!

allennlp/modules/token_embedders/bert_token_embedder.py
```diff
@@ -0,0 +1,111 @@
...
+            would be 5 wordpiece ids, and the offsets would be [3, 4]. If offsets are
+            provided, the returned tensor will contain the _last_ wordpiece embedding
+            for each token, and (in particular) will contain one embedding per token.
+            If offsets are not provided, the entire tensor of wordpiece embeddings
+            will be returned.
+        token_type_ids: ``torch.LongTensor``, optional (default = None)
+            If an input consists of two sentences (as in the BERT paper),
+            tokens from the first sentence should have type 0 and tokens from
+            the second sentence should have type 1.  If you don't provide this
+            (the default BertIndexer doesn't) then it's assumed to be all 0s.
```
matt-gardner: Is there a way to modify the indexer to provide this?  It's fine to do it in another PR, but I'd at least open an issue to track adding that.  Seems pretty important if you want to use this for SQuAD.

allennlp/tests/modules/token_embedders/bert_embedder_test.py
```diff
@@ -0,0 +1,174 @@
...
+    def test_squad_with_unwordpieceable_passage(self):
+        # pylint: disable=line-too-long
+        tokenizer = WordTokenizer()
+
+        token_indexer = PretrainedBertIndexer("bert-base-uncased")
+
+        passage1 = "There were four major HDTV systems tested by SMPTE in the late 1970s, and in 1979 an SMPTE study group released A Study of High Definition Television Systems:"
+        question1 = "Who released A Study of High Definition Television Systems?"
+
+        passage2 = """Broca, being what today would be called a neurosurgeon, had taken an interest in the pathology of speech. He wanted to localize the difference between man and the other animals, which appeared to reside in speech. He discovered the speech center of the human brain, today called Broca's area after him. His interest was mainly in Biological anthropology, but a German philosopher specializing in psychology, Theodor Waitz, took up the theme of general and social anthropology in his six-volume work, entitled Die Anthropologie der Naturvölker, 1859–1864. The title was soon translated as "The Anthropology of Primitive Peoples". The last two volumes were published posthumously."""
```
matt-gardner: You're using triple quotes already; can you make this into multiple shorter lines instead of one super long one?  Same with the passage above.

allennlp/modules/token_embedders/bert_token_embedder.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        # pylint: disable=arguments-differ
+        if token_type_ids is None:
+            token_type_ids = torch.zeros_like(input_ids)
+
+        input_mask = (input_ids != 0).long()
+
+        all_encoder_layers, _ = self.bert_model(input_ids, input_mask, token_type_ids)
+        sequence_output = all_encoder_layers[-1]
```
matt-gardner: The top layer is a particularly bad layer to use for transfer (talk to Nelson about why, or wait for his final internship presentation).  If we're fine-tuning, this is ok, but if we're not, we probably need some kind of scalar mix or something here, instead of just taking the last layer.  This is probably why your performance is so poor for your NER experiment.

scripts/ai2-internal/run_with_beaker.py
```diff
@@ -25,7 +25,13 @@ def main(param_file: str, args: argparse.Namespace):
...
 
     # Reads params and sets environment.
-    params = Params.from_file(param_file, overrides)
+    ext_vars = {}
+
+    for var in args.env:
+        key, value = var.split("=")
+        ext_vars[key] = value
+
+    params = Params.from_file(param_file, overrides, ext_vars)
```
schmmd: Thanks Joel ;-P

allennlp/modules/token_embedders/bert_token_embedder.py
DeNeutoy: Table 2 of the original elmo paper suggests otherwise (unless this is specific to transformers)? Perhaps this accounts for ~1 F1 difference, not 20+.

allennlp/common/params.py
```diff
@@ -351,13 +351,14 @@ def _check_is_dict(self, new_history, value):
...
-    def from_file(params_file: str, params_overrides: str = "") -> 'Params':
+    def from_file(params_file: str, params_overrides: str = "", ext_vars: dict = {}) -> 'Params':
+        # pylint: disable=dangerous-default-value
         """
         Load a `Params` object from a configuration file.
         """
         # redirect to cache, if necessary
         params_file = cached_path(params_file)
-        ext_vars = dict(os.environ)
+        ext_vars = {**dict(os.environ), **ext_vars}
```
DeNeutoy: This breaks the API for passing environment variables _not_ via beaker.

scripts/ai2-internal/run_with_beaker.py
```diff
@@ -25,7 +25,13 @@ def main(param_file: str, args: argparse.Namespace):
     overrides = ""
 
     # Reads params and sets environment.
-    params = Params.from_file(param_file, overrides)
+    ext_vars = {}
+
+    for var in args.env:
```
DeNeutoy: I think this is a massive breaking change, I don't think we should do this. Is there something wrong with pulling in the environment variables inside beaker as is possible already?

allennlp/modules/token_embedders/bert_token_embedder.py
matt-gardner: Layer 1 is pretty consistently better than high layers across tasks, in Nelson's experiments (with 2 and 4 layer ELMo).  The top layer is specialized for language modeling, and only a few tasks are close enough to benefit from that specialization (turns out NER is one of those).  For transformers, the story is a bit different, but middle layers are still better than the top layer.

Table 2 of the ELMo paper didn't try just using layer 1.

allennlp/common/params.py
joelgrus: how does it break it? it's saying "get all the vars from the environment, then override them with `ext_vars` if provided"

schmmd: How does this break the API?

scripts/ai2-internal/run_with_beaker.py
schmmd: The problem @joelgrus ran into was that `Params.from_file` was failing because it didn't have access to the parameters that were passed in via `--env`.

joelgrus: the problem this was trying to solve is that the run_with_beaker script is parsing the params file to inject extra env vars into beaker. in particular, if your config contains an env variable, your local machine needs that env variable in order for this parsing code not to crash.

then you *also* need to supply those env variables as command line arguments

allennlp/modules/token_embedders/bert_token_embedder.py
DeNeutoy: I wasn't saying there isn't a difference - just that even with the top layer, it should be there or thereabouts. So that isn't the problem/difference to look into here for the NER model that Joel trained.

scripts/ai2-internal/run_with_beaker.py
joelgrus: so I would have to do something like 

```
MY_FILE=/data/my_file.txt beaker experiment run --env=MY_FILE=/data/my_file.txt ...
```

which seems silly

DeNeutoy: The only reason we load the params file here is so that it appears in the beaker UI - I see that this is problematic if you have environment variables set locally (because we would pick them up and make the substitutions when we read it in here), but I don't think this is the fix for it.

allennlp/common/params.py
DeNeutoy: Oh sorry I misread the diff. Nevermind.

scripts/ai2-internal/run_with_beaker.py
joelgrus: an alternative solution would leave `Params` alone and inject all of the `--env` variables into your local environment -- I thought my solution was cleaner and potentially more broadly useful

DeNeutoy: Sorry ignore this. I misread your changes.

allennlp/modules/token_embedders/bert_token_embedder.py
DeNeutoy: I missed that slack convo. Nevermind.

joelgrus:
FYI, after making the three changes discussed:

(1) using the cased BERT model rather than the uncased one
(2) using the embedding for the first wordpiece per word rather than the last
(3) using a ScalarMix of the BERT embeddings rather than the last one

the NER model gets 93.8 dev F1:

https://beaker-internal.allenai.org/ex/ex_l9ka0jqjcrwk/tasks

I'll address the rest of the comments on Monday, and then maybe we should do a release after this is merged

Jasperty:
> I'll address the rest of the comments on Monday, and then maybe we should do a release after this is merged

Thanks for your work on bert, could you add a tutorial for how to use it?

joelgrus:
here's the config I used:

https://gist.github.com/joelgrus/7cdb8fb2d81483a8d9ca121d9c617514

susht3:
> here's the config I used:
> 
> https://gist.github.com/joelgrus/7cdb8fb2d81483a8d9ca121d9c617514

thanks a lot, but i still have the shape problem, my input shape is (batch_size,  number, len), i want to get the embedding of (batch_size,  number, len, embedding_dim), but i have the error that shape 1 is not match, it means that number maybe not padding

joelgrus:
what is "number" here? what does your dataset look like?

joelgrus:
I see, the vanilla `Embedding` has code to explicitly deal with this case:

https://github.com/allenai/allennlp/blob/master/allennlp/modules/token_embedders/embedding.py#L123

I will try to fix this today.

joelgrus:
@susht3 this should be resolved by #2104

matthew-z:
@joelgrus I was wondering if I could  avoid using `offsets` ? I found that it will cause `mask` be inconsistent with the bert tensor, so I think I should use bert tokenizer in the data reader but I am not sure how to do it. I only found a word splitter for `bert-basic`, but I guess it is only for Chinese data (character base).

joelgrus:
mask is not supposed to be consistent with the bert tensor, the intended shapes are (in the simple case)

input_ids: (batch_size, wordpiece_sequence_length)
offsets: (batch_size, tokens_sequence_length)
mask: (batch_size, tokens_sequence_length)

but anyway, if you look at the BertEmbedder, the offsets are optional:

https://github.com/allenai/allennlp/blob/master/allennlp/modules/token_embedders/bert_token_embedder.py#L60

so if you change the `embedder_to_indexer_map` in your config to not pass the offsets into the token embedder (see: https://gist.github.com/joelgrus/7cdb8fb2d81483a8d9ca121d9c617514#file-ner-bert-jsonnet-L34) (most likely in that case you can just get rid of the `embedder_to_indexer_map`) then the output of the embedder will be one bert vector per wordpiece, which it sounds like is what you want?

matthew-z:
@joelgrus Thank you for the response.

Yes, I tried to remove `bert-offsets` from `embedder_to_indexer_map `, and then the bert encoded input will become (batch_size, wordpiece_sequence_length) instead of (batch_size, tokens_sequence_length), then this code will not work because #tokens != #wordpieces:

```
mask = util.get_text_field_mask(inputs)
encoded_inputs = self.text_field_embedder(inputs)
logits = self.seq2vec(encoded_inputs, mask)

```

I used a simple workaround to set the tokenizer to use bert tokenzer to let token = wordpiece:

```
@WordSplitter.register("wordPiece")
class WordPiece(WordSplitter):

    def __init__(self, pretrained_model, do_lowercase) -> None:
        super().__init__()
        self.tokenizer = BertTokenizer.from_pretrained(pretrained_model, do_lower_case=do_lowercase)

    @overrides
    def split_words(self, sentence: str) -> List[Token]:
        return [Token(t) for t in self.tokenizer.tokenize(sentence) if t]

```

config looks like:

``` json

  "dataset_reader": {
    "type": "mydataset",
    "tokenizer": {
         "type":"word",  
         "word_splitter":{
              "type":"wordPiece", 
              "pretrained_model": "bert-base-uncased",
              "do_lowercase": true
         }
    },
    "token_indexers": {
      "bert": {
          "type": "bert-pretrained",
          "pretrained_model": "bert-base-uncased",
          "do_lowercase": true,
      },
    },
  },

```

joelgrus:
the problem here is that `util.get_text_field_mask` sees the "mask" key in your inputs and uses that as the mask even though here it's not what you want.

I think a simpler solution here is just not to use `util.get_text_field_mask`, you can do something like

```
mask = inputs["bert"] != 0
```

which is what the token embedder does internally:

https://github.com/allenai/allennlp/blob/master/allennlp/modules/token_embedders/bert_token_embedder.py#L85

matthew-z:
This solution is indeed much simpler.

Thank you!

bheinzerling:
Is there anybody who was able to get a CoNLL'03 dev score close to the reported 96.4 F1 for BERT_base? The best I've managed to get is 94.6, using the last layer as described in the paper (https://github.com/huggingface/pytorch-pretrained-BERT/issues/64#issuecomment-443212930).

bheinzerling:
Did anybody manage to get a CoNLL'03 score close the reported 96.4 F1 with BERT_base? Best I've managed to get is 94.6 using the last transformer layer for classification (https://github.com/huggingface/pytorch-pretrained-BERT/issues/64#issuecomment-443212930).

matt-gardner:
@bheinzerling, this is from a slack conversation a couple of weeks ago with @matt-peters:

> I've run a bunch of combinations, trying to initially reproduce their results in table 7.
> 
> They left out some important details in the paper for the NER task, namely that they used _document_ context for each word since it's available in the raw data.  So I was never able to reproduce their results since I was using sentence context.
> 
> It also makes their results not directly comparable to previous work, since your standard glove + biLSTM would also presumably improve with document context...
> 
> FYI: with my implementation, I got dev F1 95.09 +/- 0.07 for 2x200 dim LSTM + CRF for second-to-last layer (they reported 95.6 in table 7 without CRF). This uses sentence context to compute the BERT activations.

(This was done outside of AllenNLP.)

The last layer is slightly worse than the second-to-last layer, so your number seems to agree with @matt-peters' results.

matt-peters:
FWIW, adding document context improves F1 a little to reproduce the results in Table 7 (+/- noise from random seeds).  Note that these numbers are from extracting features from BERT, not fine-tuning.

qiuwei:
@joelgrus I had a look at the config for NER you provided. It seems that you didn't pad the sentence with the special token [CLS] at the start of each sentence.

However, I found this quite crucial in my local experiments. Did I miss anything?

qiuwei:
@matt-peters Hi matt, could you illustrate a bit more about adding document context?
Did you use the whole document(I believe that would often exceed the max length allowed by bert?) or just add a few sentences around the target sentence?

When a larger context is used, did your model predict the NER labels for the context as well?

joelgrus:
@qiuwei it looks like you are right and it's a "bug" in the token indexer, I'll open an issue for it, thanks for finding this

matt-peters:
@qiuwei - I took the easy / simple implementation approach and just chunked the document into non-overlapping segments if it exceeded the maximum length usable by BERT.  The only wrinkle is ensuring not to chunk the document in the middle of an entity span.  This way each token still has an annotation, and the NER model still predicts labels for every token.

AlexanderYogurt:
This looks great! Can we start using it now? Thanks!

bikestra:
@matt-peters Were you able to reproduce BERT paper's results once you introduced the document context? I was able to get dev F1 95.3 using sentence context but this is still 1.1% point behind what authors claim, and I didn't see much boost using document context.

This was work done outside of allennlp, but I thought independently reproducing their result using any tool would help all of us to progress; I am finding trouble anyone who was able to successfully reproduce their NER results. Sorry if this disturbed allennlp contributors.

matt-peters:
I haven't tried to reproduce the fine tuning result (96.4 dev F1, BERT base, Table 3).

kamalkraj:
https://github.com/kamalkraj/BERT-NER
Replicated results from BERT paper

pasinit:
Did anyone try to compute the average of the word piece instead of using the first workpiece of each word? For example if I have the token "longtoken" which is split in "long" "token", for my understanding now for the whole token one can easily take the embedding for long. How easy would it be instead to take the average of "long" and "token"?

wangxinyu0922:
So does anyone successfully reproduce the score reported in the paper? I want to reproduce the score but I find it very hard to do that (only 91.5~ with document context + finetuning).

dsindex:
@wangxinyu0922 

with document context + subword pooling, i reached **92.23%** (bert-base-cased)

https://github.com/dsindex/ntagger/issues/4#issuecomment-810304253

PR #2076, opened 2018-11-18
Add support of tokenized input for coref and srl predictors

AlonEirew:
Adding method to support predicting on tokenized input in ``coref.py`` and ``semantic_role_labeler.py`` API methods.

Added methods:
``CorefPredictor.predict_from_list(self, tokenized_document: List[str]) -> JsonDict:``
``SemanticRoleLabelerPredictor.predict_from_list(self, tokenized_sentence: List[str]) -> JsonDict:``

This is very useful when user data is already tokenized (for example annotated corpus for some NLP task) and user would like to predict on that tokenized data without losing the original tokens ID's.. 

Avoid a workflow of: Creating document/sentence text from tokenized data -> input created text to predict method -> predict will tokenize the input text -> align output tokens with original tokens.

Usage code example in:
for coref: ``allennlp/tests/predictors/coref_test.py``, test: ``test_uses_named_inputs``
for srl:  ``allennlp/tests/predictors/srl_test.py``, test: ``test_uses_named_inputs``

### DIFF
--- a/allennlp/predictors/coref.py
+++ b/allennlp/predictors/coref.py
@@ -1,7 +1,9 @@
+from typing import List
+
 from overrides import overrides
 
-from allennlp.common.util import get_spacy_model
 from allennlp.common.util import JsonDict
+from allennlp.common.util import get_spacy_model
 from allennlp.data import DatasetReader, Instance
 from allennlp.models import Model
 from allennlp.predictors.predictor import Predictor
@@ -52,6 +54,38 @@ class CorefPredictor(Predictor):
         A dictionary representation of the predicted coreference clusters.
         """
         return self.predict_json({"document" : document})
+
+    def predict_from_list(self, document_as_list: List[str]) -> JsonDict:
+        """
+        Predict the coreference clusters in the given document.
+
+        Parameters
+        ----------
+        document_as_list : ``List[str``
+            A list of words representation of a tokenized document.
+
+        Returns
+        -------
+        A dictionary representation of the predicted coreference clusters.
+        """
+        return self.predict_words_list(document_as_list)
+
+    def predict_words_list(self, words_list: List[str]) -> JsonDict:
+        instance = self._words_list_to_instance(words_list)
+        return self.predict_instance(instance)
+
+    def _words_list_to_instance(self, document_list: List[str]) -> Instance:
+        """
+        Create an instance from words list represent an already tokenized document,
+        for skipping tokenization when that information already exist for the user
+        """
+        spacy_document = self._spacy.tokenizer.tokens_from_list(document_list)
+        for pipe in filter(None, self._spacy.pipeline):
+            pipe[1](spacy_document)
+
+        sentences = [[token.text for token in sentence] for sentence in spacy_document.sents]
+        instance = self._dataset_reader.text_to_instance(sentences)
+        return instance
 
     @overrides
     def _json_to_instance(self, json_dict: JsonDict) -> Instance:
--- a/allennlp/predictors/coref.py
+++ b/allennlp/predictors/coref.py
@@ -55,31 +55,28 @@ class CorefPredictor(Predictor):
         """
         return self.predict_json({"document" : document})
 
-    def predict_from_list(self, tokenized_document: List[str]) -> JsonDict:
+    def predict_tokenized(self, tokenized_document: List[str]) -> JsonDict:
         """
         Predict the coreference clusters in the given document.
 
         Parameters
         ----------
-        tokenized_document : ``List[str``
+        tokenized_document : ``List[str]``
             A list of words representation of a tokenized document.
 
         Returns
         -------
         A dictionary representation of the predicted coreference clusters.
         """
-        return self.predict_words_list(tokenized_document)
-
-    def predict_words_list(self, words_list: List[str]) -> JsonDict:
-        instance = self._words_list_to_instance(words_list)
+        instance = self._words_list_to_instance(tokenized_document)
         return self.predict_instance(instance)
 
-    def _words_list_to_instance(self, document_list: List[str]) -> Instance:
+    def _words_list_to_instance(self, words: List[str]) -> Instance:
         """
         Create an instance from words list represent an already tokenized document,
         for skipping tokenization when that information already exist for the user
         """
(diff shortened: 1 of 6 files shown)


allennlp/predictors/coref.py
```diff
@@ -53,6 +55,38 @@ def predict(self, document: str) -> JsonDict:
...
+    def predict_words_list(self, words_list: List[str]) -> JsonDict:
+        instance = self._words_list_to_instance(words_list)
+        return self.predict_instance(instance)
+
+    def _words_list_to_instance(self, document_list: List[str]) -> Instance:
+        """
+        Create an instance from words list represent an already tokenized document,
+        for skipping tokenization when that information already exist for the user
+        """
+        spacy_document = self._spacy.tokenizer.tokens_from_list(document_list)
```
matt-gardner: It looks like you're calling the same spacy pipeline on the document twice here; once inside `tokenizer.tokens_from_list`, and then once again here.  I think all you need to do is call `spacy_document = self._spacy.tokens_from_list(words)` here (without adding the method onto the `SpacyWordSplitter`), then continue with this logic, and it should just work.  We don't need the extra method on `SpacyWordSplitter` at all.

allennlp/predictors/coref.py
```diff
@@ -53,6 +55,38 @@ def predict(self, document: str) -> JsonDict:
...
+        -------
+        A dictionary representation of the predicted coreference clusters.
+        """
+        return self.predict_words_list(tokenized_document)
+
+    def predict_words_list(self, words_list: List[str]) -> JsonDict:
+        instance = self._words_list_to_instance(words_list)
+        return self.predict_instance(instance)
+
+    def _words_list_to_instance(self, document_list: List[str]) -> Instance:
```
matt-gardner: `words` is a better name here than `document_list`.

allennlp/predictors/semantic_role_labeler.py
```diff
@@ -41,8 +41,23 @@ def predict(self, sentence: str) -> JsonDict:
         -------
         A dictionary representation of the semantic roles in the sentence.
         """
-        return self.predict_json({"sentence" : sentence})
+        return self.predict_json({"sentence": sentence})
 
+    def predict_from_list(self, tokenized_sentence: List[str]) -> JsonDict:
```
matt-gardner: `predict_tokenized` here too.

allennlp/predictors/semantic_role_labeler.py
```diff
@@ -198,16 +231,17 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
...
-        results = {"verbs": [], "words": outputs[0]["words"]}
-        for output in outputs:
-            tags = output['tags']
-            description = self.make_srl_string(output["words"], tags)
-            results["verbs"].append({
-                    "verb": output["verb"],
-                    "description": description,
-                    "tags": tags,
-            })
+    def predict_words_list(self, words_list: List[str]) -> JsonDict:
```
matt-gardner: Again the logic in this method should just be moved into `predict_tokenized` - this is a duplicate method.

allennlp/predictors/semantic_role_labeler.py
```diff
@@ -198,16 +231,17 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
...
+        """
+        Create an instance list of works document, for skipping tokenization when that
+        information already exist for the user
+        """
+        tokens = self._tokenizer.tokens_from_list(words_list)
```
matt-gardner: You can't rely on this method actually existing, because you didn't create the method on the base class (I'm a little surprised that mypy didn't catch this; maybe because we're playing a little loose with the tokenizers inside of a predictor already...).  But it's a lot easier than this: just do `tokens = [Token(word) for word in words]`.

allennlp/tests/predictors/srl_test.py
```diff
@@ -13,21 +13,27 @@ def test_uses_named_inputs(self):
...
+        result_words = predictor.predict_words_list(words)
+        self.assert_predict_result(result_words)
+
+    @staticmethod
+    def assert_predict_result(result):
         words = result.get("words")
         assert words == ["The", "squirrel", "wrote", "a", "unit", "test",
                          "to", "make", "sure", "its", "nuts", "worked", "as", "designed", "."]
         num_words = len(words)
-
```
matt-gardner: I'd not remove these blank lines.  They separate coherent segments.

allennlp/predictors/coref.py
AlonEirew: Here im using the actual Spacy tokenizer and not the WordSplitter one (which I use in srl), it will return a spacy.Doc object, I then run the pipeline on the doc object only once.

allennlp/predictors/semantic_role_labeler.py
AlonEirew: Did you mean here I should just add ``tokens = [Token(word) for word in words]`` at WordSplitter base class?

allennlp/predictors/coref.py
matt-gardner: Oh, I see - so yeah, you really don't need that extra method at all.

allennlp/predictors/semantic_role_labeler.py
matt-gardner: Instead of the line that you have, you should have `tokens = [Token(word) for word in words]`.  There's no need for any extra methods on `WordSplitter` for any of what you're implementing.

AlonEirew: Spacy POS tagging is required, I've tried to follow the class logic were all spacy pipeline is under the hood via class ``SpacyWordSplitter``

matt-gardner: That's especially non-obvious from the diff I was looking at.  And now that I look at the things that were hidden, I understand why mypy didn't catch this - I didn't realize that we specifically instantiated a `SpacyWordSplitter` here.  I'd just do what you did in the coref predictor - call `self._tokenizer.tokenizer.tokens_from_list()`, or whatever the method is.

AlonEirew:
Hi Matt, I've fixed all comments, you can review.

Issue #1052, opened 2018-03-29
Using a serialisation directory as a model archive doesn't work with files_to_archive.json


DeNeutoy:
Stack trace:

```
Traceback (most recent call last):
  File "/home/vidurj/miniconda3/envs/allennlp/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/vidurj/miniconda3/envs/allennlp/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/vidurj/allennlp/allennlp/run.py", line 18, in <module>
    main(prog="python -m allennlp.run")
  File "/home/vidurj/allennlp/allennlp/commands/__init__.py", line 63, in main
    args.func(args)
  File "/home/vidurj/allennlp/allennlp/commands/predict.py", line 158, in _predict
    predictor = _get_predictor(args)
  File "/home/vidurj/allennlp/allennlp/commands/predict.py", line 104, in _get_predictor
    overrides=args.overrides)
  File "/home/vidurj/allennlp/allennlp/models/archival.py", line 149, in load_archive
    cuda_device=cuda_device)
  File "/home/vidurj/allennlp/allennlp/models/model.py", line 233, in load
    model = Model.from_params(vocab, model_params)
  File "/home/vidurj/allennlp/allennlp/models/model.py", line 184, in from_params
    model = cls.by_name(choice).from_params(vocab, params)
  File "/home/vidurj/allennlp/allennlp/models/encoder_decoders/simple_seq2seq.py", line 373, in from_params
    source_embedder = TextFieldEmbedder.from_params(vocab, source_embedder_params)
  File "/home/vidurj/allennlp/allennlp/modules/text_field_embedders/text_field_embedder.py", line 56, in from_params
    return cls.by_name(choice).from_params(vocab, params)
  File "/home/vidurj/allennlp/allennlp/modules/text_field_embedders/basic_text_field_embedder.py", line 62, in from_params
    token_embedders[key] = TokenEmbedder.from_params(vocab, embedder_params)
  File "/home/vidurj/allennlp/allennlp/modules/token_embedders/token_embedder.py", line 32, in from_params
    return cls.by_name(choice).from_params(vocab, params)
  File "/home/vidurj/allennlp/allennlp/modules/token_embedders/elmo_token_embedder.py", line 67, in from_params
    params.add_file_to_archive('options_file')
[...]

PR #2100, opened 2018-11-26
Resolve some of the token_embedders warnings.

schmmd:


schmmd:
I'm not done with this, but I wanted to ask about one approach I'm taking.  I removed `allennlp/tests/fixtures/bidaf/serialization/model.tar.gz` and instead use `allennlp/tests/fixtures/bidaf/serialization/` as the archive directory to load.  This means that `git` can track our changes and we don't need to include the vocabulary files twice (once inside the tarball and once outside).

Are there any concerns with this change @joelgrus @DeNeutoy @matt-gardner?  Otherwise I need to update the `model.tar.gz` itself.  If we go that route--would you un-tar and re-tar or rebuild the test fixture?

DeNeutoy:
I would rather leave the tests using `model.tar.gz` because:

1. We distribute `model.tar.gz` files, not directories - it seems best to have our tests which check if models can train, save and load use the actual format we distribute.

2. The output in the `serialization` directories in the tests are generated using a command from the library (`train_model_from_file`) in this script: https://github.com/allenai/allennlp/blob/master/scripts/train_fixtures.py

Currently retraining the fixtures follows exactly the process of training a model. In order not to regenerate these `model.tar.gz` files, we would need to modify the behaviour of that script to remove the generated archive file.

3. Tracking the changes to the files inside the archives generated by test fixtures has relatively low value (as far as I can tell).

4. There are some differences between using a directory and using an archive in some edge cases #1052

schmmd:
@DeNeutoy this is fine, but I wish you had spoken up earlier.  Fixing these warnings is rather painful and adding the `model.tar.gz` back will be a bit more pain ;-(.

We do test loading from a `tar.gz` in the sniff test.

DeNeutoy:
Sorry, if you need immediate feedback you should post it in slack or message me directly or something.

schmmd:
Yes, pretty uncontentious.  Unfortunately it leave a ton of warnings in the repo ;-P

DeNeutoy:
Why does it leave warnings?

schmmd:
@DeNeutoy the models need to be rebuilt with the new configuration.  When they are loaded presently they hit https://github.com/allenai/allennlp/blob/master/allennlp/modules/text_field_embedders/basic_text_field_embedder.py#L122.  I thought I'd try to work down our warnings a bit, but it's not terribly fun.  @joelgrus was wondering if we should just remove the warning from https://github.com/allenai/allennlp/blob/master/allennlp/modules/text_field_embedders/basic_text_field_embedder.py#L122--since it also hits us in the sniff tests, etc.

PR #2104, opened 2018-11-27
modify BERT embedder to deal with higher order inputs

### LINKED ISSUE #2103: BERT embedder should handle higher-order tensors
right now the BERT token embedder only works if its input is a `(batch_size, sequence_length)` tensor of ids. it should also work if there are extra dimensions.

see, e.g. https://github.com/allenai/allennlp/pull/2067#issuecomment-442125174

joelgrus:
fixes #2103 

1. override `WordpieceIndexer.get_keys` to return the full list of indexed keys
2. abstract out the "to 2d / from 2d" code from `Embedding` into a couple of helper functions in `nn.util`
3. use that code in the BERT embedder
4. write tests


allennlp/nn/util.py
```diff
@@ -1289,3 +1289,30 @@ def add_positional_features(tensor: torch.Tensor,
 def clone(module: torch.nn.Module, num_copies: int) -> torch.nn.ModuleList:
     "Produce N identical layers."
     return torch.nn.ModuleList([copy.deepcopy(module) for _ in range(num_copies)])
+
+
+def make2d(inputs: torch.Tensor) -> torch.Tensor:
```
DeNeutoy: These should handle the 1D case better - if you give it a 1D tensor, it does not make it 2D, despite it's name - maybe it should error or unsqueeze or something?

allennlp/nn/util.py
```diff
@@ -1289,3 +1289,30 @@ def add_positional_features(tensor: torch.Tensor,
...
+    (batch_size * d1 * ... * dn, sequence_length, embedding_dim)
+    and the original shape
+    (batch_size, d1, ..., dn, sequence_length),
+    return the reshaped tensor of embeddings with shape
+    (batch_size, d1, ..., dn, sequence_length, embedding_dim)
+    """
+    if len(original_size) > 2:
+        view_args = list(original_size) + [inputs.size(-1)]
+        return inputs.view(*view_args)
+    else:
```
DeNeutoy: Maybe this should error if the original size is < 2 and not equal to the dimension of the tensor?

allennlp/nn/util.py
joelgrus: it needs to handle the 1d case, so I renamed it

PR #2105, opened 2018-11-27
Fixing conll delimiter

gabrielStanovsky:
1. Change the Ontonotes reader to use only tab delimiters, instead of on any space character.
This conforms to the [CoNLL specification](http://universaldependencies.org/docs/format.html), and also fixes a bug which occurs when reading CoNLL sentences which annotate some characters which python considers as space.

2. Adapt the fixtures to be tab separated, instead of space delimited.

3. Add a script for doing (2) for a given CoNLL file.


gabrielStanovsky:
@DeNeutoy has reservations on this change, since it might break users who use a space-delimited file. I'll work on fixing this and repost.

PR #2106, opened 2018-11-28
Rename coding_scheme parameter for Conll2003DatasetReader

kushalchauhan98:
This PR:

- Renames `coding_scheme` parameter to `convert_to_coding_scheme`
- Changes its default to None
- Fixes documentation of Conll2003DatasetReader
- Updates TestConll2003Reader on account of these changes


allennlp/data/dataset_readers/conll2003.py
```diff
@@ -38,6 +38,10 @@ class Conll2003DatasetReader(DatasetReader):
     and '-DOCSTART- -X- -X- O' indicating the end of each article,
     and converts it into a ``Dataset`` suitable for sequence tagging.
 
+    This dataset reader expects the input to be in the Conll2003 standard IOB1 coding
```
nelson-liu: s/Conll2003/CoNLL 2003/

allennlp/data/dataset_readers/conll2003.py
```diff
@@ -61,24 +65,23 @@ class Conll2003DatasetReader(DatasetReader):
...
         specified here.
-    coding_scheme: ``str``, optional (default=``IOB1``)
+    convert_to_coding_scheme: ``str``, optional (default=``None``)
         Specifies the coding scheme for ``ner_labels`` and ``chunk_labels``.
-        Valid options are ``IOB1`` and ``BIOUL``.  The ``IOB1`` default maintains
-        the original IOB1 scheme in the CoNLL 2003 NER data.
-        In the IOB1 scheme, I is a token inside a span, O is a token outside
-        a span and B is the beginning of span immediately following another
-        span of the same type.
+        A valid option is ``BIOUL``.
```
nelson-liu: I think you can just remove this line

allennlp/data/dataset_readers/conll2003.py
```diff
@@ -61,24 +65,23 @@ class Conll2003DatasetReader(DatasetReader):
...
     _VALID_LABELS = {'ner', 'pos', 'chunk'}
+    _VALID_CODING_SCHEMES = {'BIOUL',}
 
     def __init__(self,
                  token_indexers: Dict[str, TokenIndexer] = None,
                  tag_label: str = "ner",
                  feature_labels: Sequence[str] = (),
                  lazy: bool = False,
-                 coding_scheme: str = "IOB1",
+                 convert_to_coding_scheme: str = None,
```
nelson-liu: I think the type annotation on this might be `Optional[str]`? will need to import `Optional` from `typing`.

allennlp/data/dataset_readers/conll2003.py
```diff
@@ -131,14 +134,14 @@ def text_to_instance(self, # type: ignore
         instance_fields: Dict[str, Field] = {'tokens': sequence}
         instance_fields["metadata"] = MetadataField({"words": [x.text for x in tokens]})
 
-        # Recode the labels if necessary.
-        if self.coding_scheme == "BIOUL":
+        # Recode the labels to BIOUL
```
nelson-liu: I'd keep this as `if necessary`, but no strong opinion

allennlp/data/dataset_readers/conll2003.py
kushalchauhan98: You mean the 'A valid option is BIOUL' line?

nelson-liu: yup!

joelgrus:
doesn't this change break every NER config file that uses `coding_scheme`?

nelson-liu:
@joelgrus good point, those should be changed as well.

nelson-liu:
And maybe this goes in a different PR / issue, but it'd be reasonable to change the `coding_scheme` parameters in `ontonotes_ner` and `conll200_dataset_reader` to match the `convert_to_coding_scheme` used here.

joelgrus:
I was more questioning why we're making this breaking change, I thought we talked about just making the documentation clearer. 

This also breaks the pretrained NER models (and the corresponding sniff tests), so those would all have to be changed too, and then the documentation about how to run them, and then ...

kushalchauhan98:
If a lot of things are breaking as a consequence, I could revert the changes and just fix the documentation.

allennlp/data/dataset_readers/conll2003.py
matt-gardner: Have we made any official decision on this?  Personally, I prefer leaving off the `Optional`, because that's already implied by the fact that the default value is `None`.  Is there a good reason to add in `Optional`?  @joelgrus, does it impact any `FromParams` or configurator stuff?

matt-gardner:
I am somewhat in favor of changing the name, because the original name isn't really correct.  If we do that, though, we definitely need to do it in a way that's backwards compatible, deprecating the original argument instead of removing it.

schmmd:
@kushalchauhan98 if you would like to merge this (after discussion with @matt-gardner, @joelgrus, and @brendan-ai2), we need to deprecate the old name so user code isn't broken immediately and fix all warnings in the AllenNLP library for that deprecation.

kushalchauhan98:
@matt-gardner @joelgrus @brendan-ai2 May I proceed by deprecating the old naming scheme?

matt-gardner:
Yes, as long as you do it in a way that's backwards compatible.  That is, we need to keep the old argument as an option (that gives a deprecation warning), and convert it to the new argument if it's given.

brendan-ai2:
We're closing this due to inactivity. If you'd like to address the changes requested, please open a new PR. Thank you.

PR #2108, opened 2018-11-28
Fix epoch tracking bucket iterator warnings.

schmmd:


schmmd:
I like to think of this as gaining a test for `track_epoch` rather than losing a test for a deprecated class.

If you'd prefer, I'll revert allennlp/tests/data/iterators/epoch_tracking_bucket_iterator_test.py and leave a single warning (and maybe loop back to it).  I'm trying to clean up as many warnings as I can before the next release.

matt-gardner:
My preference: move the test inside `epoch_tracking_bucket_iterator_test.py` to `bucket_iterator_test.py` (or even to a `data_iterator_test.py`, because it's functionality that's in the base class), and either remove `epoch_tracking_bucket_iterator_test.py` entirely, or leave a single test that checks for a deprecation warning (which probably isn't worth the effort).

schmmd:
Makes sense--I'll do that.

Just pushing back on investigating new mechanism for controlling warnings because I have limited time and I'm trying to crush as many warnings as possible before the next release ;-)

PR #2111, opened 2018-11-29
Pack batches more tightly with maximum_samples_per_batch.

brendan-ai2:
- Previously when maximum_samples_per_batch was used overly large batches would be split into equal sizes.
- This could lead to a certain degree of fragmentation.
- This change packs batches until just before maximum_samples_per_batch (or batch_size) is exceeded.

### DIFF
--- a/allennlp/data/iterators/data_iterator.py
+++ b/allennlp/data/iterators/data_iterator.py
@@ -1,6 +1,6 @@
 import logging
 from typing import Dict, Union, Iterable, Iterator, List, Optional, Tuple, Deque
-from collections import defaultdict, deque
+from collections import defaultdict
 import itertools
 import math
 import random
--- a/allennlp/data/iterators/data_iterator.py
+++ b/allennlp/data/iterators/data_iterator.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Dict, Union, Iterable, Iterator, List, Optional, Tuple
+from typing import Dict, Union, Iterable, Iterator, List, Optional, Tuple, Deque
 from collections import defaultdict, deque
 import itertools
 import math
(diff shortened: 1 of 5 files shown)


brendan-ai2:
@swabhs, pull this down and give it a try. Here's some sample iterator config that you can then adjust to your liking:

```{
  "type": "bucket",
  # Should be decently larger than your batch size.
  "max_instances_in_memory": 4096 * NUM_GPUS,
  # Should be large enough that you'll always (or almost always) split based on maximum_samples_per_batch set below.
  "batch_size": 512 * NUM_GPUS,
  # Adjust as appropriate.
  "sorting_keys": [["source", "num_tokens"]],
  # Adjust as appropriate.
  "biggest_batch_first": true,
  # Adjust as appropriate.
  "maximum_samples_per_batch": ["num_tokens", NUM_GPUS * 2000]
}```

**Comment by brendan-ai2** [comment]
@DeNeutoy, here is the iterator improvement we discussed the other week. I ended up implementing it as a change to the existing iterators as it seems strictly better than the old method. But perhaps I'm missing something in terms of backwards compatibility. Let me know! Thanks!

File: allennlp/data/iterators/data_iterator.py [hunk 1]
```diff
@@ -225,19 +225,35 @@ def _memory_sized_lists(self,
             yield list(iterator)
 
 
-    def _ensure_batch_is_sufficiently_small(self, batch_instances: Iterable[Instance]) -> List[List[Instance]]:
+    def _ensure_batch_is_sufficiently_small(
+            self,
+            batch_instances: Iterable[Instance],
+            excess: deque) -> List[List[Instance]]:
```
DeNeutoy: For everything but core types, it's better to use the collection's corresponding entry in the `typing` module, because it allows you to specify what the collection is of, e.g:

```
from typing import Deque

...
excess: Deque[Instance]) -> ....
```
It's also less confusing, because they are strictly abstract, so it's obvious they're being used for type hints or inheritance.

allennlp/data/iterators/data_iterator.py
```diff
@@ -225,19 +225,35 @@ def _memory_sized_lists(self,
...
         """
-        If self._maximum_samples_per_batch is specified, then split the batch into smaller
-        sub-batches if it exceeds the maximum size.
+        If self._maximum_samples_per_batch is specified, then split the batch
+        into smaller sub-batches if it exceeds the maximum size.
+
+        Any excess passed in will be used first. When the method returns excess
```
DeNeutoy: Could you mention explicitly in this docstring that `excess` is mutated in-place.

allennlp/data/iterators/data_iterator.py
brendan-ai2: Good call, thanks! I didn't know that all collection classes had corresponding types.

allennlp/data/iterators/data_iterator.py
brendan-ai2: Cleaned up the language and left a big warning.

brendan-ai2:
Class docstring changed too. Thanks for the speedy review!

swabhs:
> @swabhs, pull this down and give it a try. Here's some sample iterator config that you can then adjust to your liking:
> 
> ```
> {
>   "type": "bucket",
>   # Should be decently larger than your batch size.
>   "max_instances_in_memory": 4096 * NUM_GPUS,
>   # Should be large enough that you'll always (or almost always) split based on maximum_samples_per_batch set below.
>   "batch_size": 512 * NUM_GPUS,
>   # Adjust as appropriate.
>   "sorting_keys": [["source", "num_tokens"]],
>   # Adjust as appropriate.
>   "biggest_batch_first": true,
>   # Adjust as appropriate.
>   "maximum_samples_per_batch": ["num_tokens", NUM_GPUS * 2000]
> }
> ```

This is running now, and is allowing bigger batches - hope it is the solution I have been looking for!

PR #2114, opened 2018-11-29
Remove the BasicTextFieldEmbedder warning

schmmd:
While we would like to discourage the original behavior, we would like to remove the deprecation warning because we still follow it throughout our test fixtures and full models.  We can formally deprecate this behavior when AllenNLP gets off it.

### DIFF
--- a/allennlp/tests/modules/text_field_embedders/basic_text_field_embedder_test.py
+++ b/allennlp/tests/modules/text_field_embedders/basic_text_field_embedder_test.py
@@ -108,7 +108,27 @@ class TestBasicTextFieldEmbedder(AllenNlpTestCase):
                 }
         token_embedder(inputs)
 
-    def test_new_from_params(self):
+    def test_old_from_params_new_from_params(self):
+        old_params = Params({
+                "token_embedders": {
+                        "words1": {
+                                "type": "embedding",
+                                "embedding_dim": 2
+                                },
+                        "words2": {
+                                "type": "embedding",
+                                "embedding_dim": 5
+                                },
+                        "words3": {
+                                "type": "embedding",
+                                "embedding_dim": 3
+                                }
+                        }
+                })
+
+        # Allow loading the parameters in the old format
+        old_embedder = BasicTextFieldEmbedder.from_params(params=old_params, vocab=self.vocab)
+
         new_params = Params({
                 "token_embedders": {
                         "words1": {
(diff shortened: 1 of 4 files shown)


matt-gardner:
This is a case where I'd prefer just telling pylint to skip the warning, instead of removing it.

schmmd:
@matt-gardner I thought of that based on your earlier comment--but users would still see these warnings when they use our stock models or our test fixtures--which they are not particularly able to address.  I'd have to add more than a dozen commands to ignore the warnings here and probably in the demo as well.  Overall--it doesn't seem particularly worth it.

matt-gardner:
The test fixtures are easy to fix - this PR does that.  The only issue is our pre-trained models, right?  And we can fix those by just uploading a new version with a fixed config, without changing anything else.

allennlp/tests/modules/text_field_embedders/basic_text_field_embedder_test.py
```diff
@@ -104,26 +108,7 @@ def test_forward_runs_with_non_bijective_mapping(self):
                 }
         token_embedder(inputs)
 
-    def test_old_from_params_new_from_params(self):
```
matt-gardner: Especially if we're removing the deprecation warning, we should not remove this test, as we're still allowing arguments to be passed this way.  If we keep the deprecation warning as I would prefer, we should still keep this test.  We can very easily remove the displayed warning in CI by adding a `pytest.mark.filterwarnings("token_embedders")` decorator to the test.

schmmd: Good point--I can do that.

schmmd:
@matt-gardner this PR doesn't do that.  Someone would need to retrain all of the `model.tar.gz`.  I did some of this in a previous PR, but Mark didn't like that I was using a directory instead of a `model.tar.gz` for the models so I reverted that change.

> And we can fix those by just uploading a new version with a fixed config, without changing anything else.

We can do that--but I'm not sure anyone will and I'm also skeptical it's the best use of anyone's time.  I'm only addressing this because the warnings have bothered me over the past weeks--but it's turned into a lot more work than I expected!

matt-gardner:
> @matt-gardner this PR doesn't do that. Someone would need to retrain all of the `model.tar.gz`. I did some of this in a previous PR, but Mark didn't like that I was using a directory instead of a `model.tar.gz` for the models so I reverted that change.

This just requires running `./scripts/retrain_fixtures.py` after this PR where you update all of the config files.  It should be very easy.

> We can do that--but I'm not sure anyone will and I'm also skeptical it's the best use of anyone's time. I'm only addressing this because the warnings have bothered me over the past weeks--but it's turned into a lot more work than I expected!

But after fixing the test fixtures, the remaining deprecation warnings are quite minimal - it only shows up in the sniff tests, not the main CI tests, because only the sniff tests use the pre-trained models.

allennlp/tests/modules/text_field_embedders/basic_text_field_embedder_test.py
schmmd: I also added a test to make sure the keys match regardless of how they are passed in.  The old test just made sure the old way didn't throw an exception and instead logged a warning.

schmmd:
Good to know about `scripts/train_fixtures.py`--but I don't think that actually will hit all of them.  I'll try... once ;-)

schmmd:
Matt's going to take a look so I added back in the deprecation warning for now.

PR #2118, opened 2018-11-30
release config explorer

joelgrus:
* make it so `allennlp configure` launches the config explorer
* a couple of minor cosmetic changes
* lots of bug fixes for edge cases:
** Vocabulary has a weird from_params and needs bespoke treatment
** not all `FromParams` classes are subclasses of `Registrable` classes (e.g. `FeedForward`), so I had to change a bunch of logic to deal with that
** some `Registrable` base classes are themselves instantiable (Vocabulary, Trainer), so I had to change a bunch of logic to deal with that too
** make the API change`Activation` annotations to `str` annotations, since when we take them from_params that's how we take them anyway
** probably some others

the main remaining non-supported case is union types, which I'm still not sure how to handle


allennlp/service/config_explorer.html
```diff
@@ -383,21 +440,62 @@ <h2>AllenNLP Configuration Wizard (alpha)</h2>
...
+                )
+            }
+        }
+
+        // Caveat for known issues
+        const KnownIssues = () => (
+            <div class="known-issues">
+                <h3>Known Issues</h3>
+                <ul>
+                    <li>For the most part this doesn't handle union types correctly.
```
schmmd: maybe put the <li> on one line?  GitHub hates this ;-)

schmmd:
OK--so there's a lot going on in here (and much of it's in javascript) so it's hard to review such a large PR--but the code looks good on the whole and the wizard looks great.

I'm bothered by the fact that the HTML is duplicated across python and a `.html` file.  @joelgrus had trouble distributing the html file with pip. @nelson-liu did you figure this out when you moved the test cases to allennlp itself?

nelson-liu:
@schmmd seems like the .html file is being included.

I ran:
```
conda create -n test_config_explorer python=3.6
source activate test_config_explorer
pip install git+https://github.com/joelgrus/allennlp.git@release-config-explorer
python -c "import allennlp; print(allennlp.__file__)"
```

Then, navigate to the install directory, in my case /Users/nfliu/miniconda3/envs/test_config_explorer/lib/python3.6/site-packages/allennlp/...

Lastly:
```
$ ls service/
__init__.py           config_explorer.html  predictors/
__pycache__/          config_explorer.py    server_simple.py
```

schmmd:
@nelson-liu you make everything look so easy ;-P

joelgrus:
this xcode issue makes it so I can't even pip install allennlp on my macbook 😢

joelgrus:
ok, I got rid of the duplicated HTML and it seems to still work

schmmd:
Will do.

schmmd:
Works for me.

```
conda create -n allennlp-joelgrus python=3.6 
conda activate allennlp-joelgrus
pip install git+https://github.com/joelgrus/allennlp.git@release-config-explorer
allennlp configure
```

PR #2119, opened 2018-11-30
Transformer ELMo

brendan-ai2:
- Configuration for training a transformer based bidirectional LM.
  - Sampled loss on train after four epochs was 3.384.
- Minor fixes to CnnHighwayEncoder.
  - LayerNorm was needed instead of MaskedLayerNorm.
- Log average batch size during training.


allennlp/data/dataset_readers/simple_language_modeling.py
```diff
@@ -52,8 +54,11 @@ def text_to_instance(self,  # type: ignore
                          sentence: str) -> Instance:
         # pylint: disable=arguments-differ
         tokenized = self._tokenizer.tokenize(sentence)
+        tokenized_with_ends = [Token(ELMoCharacterMapper.bos_token)]
+        tokenized_with_ends.extend(tokenized)
+        tokenized_with_ends.append(Token(ELMoCharacterMapper.eos_token))
```
brendan-ai2: This part would be good to get a sanity check on. Any implications to just throwing these bos and eos tags back in?

allennlp/models/bidirectional_lm.py
```diff
@@ -227,10 +236,11 @@ def forward(self,  # type: ignore
         forward_targets[:, 0:-1] = token_ids[:, 1:]
         backward_targets[:, 1:] = token_ids[:, 0:-1]
 
+        # shape (batch_size, timesteps + 2, embedding_size)
         embeddings = self._text_field_embedder(source)
 
         # Apply LayerNorm if appropriate.
-        embeddings = self._layer_norm(embeddings)
+        embeddings = self._layer_norm(embeddings, mask)
```
DeNeutoy: Just to double check, did you mean to add layer norm both here and at the output of the `HighwayCnnEncoder`? I can see that might be useful for someone, but you probably don't want both these layernorms happening simultaneously.

allennlp/training/trainer.py
```diff
@@ -451,6 +455,18 @@ def _get_metrics(self, total_loss: float, num_batches: int, reset: bool = False)
         metrics["loss"] = float(total_loss / num_batches) if num_batches > 0 else 0.0
         return metrics
 
+    def _get_batch_size(self, batch: Union[Dict, torch.Tensor]) -> int:
+        """
+        Returns the size of the batch dimension. Assumes a well-formed batch,
+        returns 0 otherwise.
+        """
+        if isinstance(batch, torch.Tensor):
+            return batch.size()[0] # type: ignore
```
DeNeutoy: `return batch.size(0)` might alleviate that type warning

allennlp/training/trainer.py
```diff
@@ -482,8 +498,15 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
         logger.info("Training")
         train_generator_tqdm = Tqdm.tqdm(train_generator,
                                          total=num_training_batches)
+        cumulative_batch_size = 0
         for batch in train_generator_tqdm:
             batches_this_epoch += 1
+            if self._log_batch_size_period:
```
DeNeutoy: Could you move this below to where the rest of the logging happens, and also send the values to tensorboard:

https://github.com/allenai/allennlp/pull/2119/files#diff-043dcd121296c3cef3f3ff8c74127ff1R557

training_config/bidirectional_lm.jsonnet
```diff
@@ -0,0 +1,152 @@
...
+    "type": "multiprocess",
+    "base_reader": BASE_READER,
+    "num_workers": NUM_THREADS,
+    "output_queue_size": 1000
+  },
+  # Note: We don't set a validation_data_path because the softmax is only
+  # sampled during training. Not sampling on GPUs results in a certain OOM
+  # given our large vocabulary. We'll need to evaluate against the test set
+  # (when we'll want a full softmax) with the CPU.
+  "train_data_path": "/home/brendanr/workbenches/calypso/train/*",
```
DeNeutoy: Could you pull this up into a env variable so that anyone could use this config without having to modify it?

training_config/bidirectional_lm.jsonnet
```diff
@@ -0,0 +1,152 @@
...
+  #"train_data_path": "/home/brendanr/workbenches/calypso/train/news.en-0000[2-3]*",
+  #"train_data_path": "/home/brendanr/workbenches/calypso/train_small/*",
```
joelgrus: can you remove these commented out paths

training_config/bidirectional_lm.jsonnet
```diff
@@ -0,0 +1,152 @@
...
+
+local BASE_READER = {
+        "type": "simple_language_modeling",
+        "tokenizer": {
+          "type": "word",
+          "word_splitter": {
+	        # The 1 Billion Word Language Model Benchmark dataset is
+	        # pre-tokenized. (Also, if you're running against a untokenized
+	        # dataset be aware that there are serialization issues with Spacy.
+	        # These come into play in the multiprocess case.)
```
joelgrus: nit: I know that jsonnet allows python-style comments, but I find them kind of jarring in "json". I don't know if anyone else cares.

training_config/bidirectional_lm.jsonnet
```diff
@@ -0,0 +1,152 @@
...
+
+  "vocabulary": {
+      # Use a prespecified vocabulary for efficiency.
+      "directory_path": "/home/brendanr/workbenches/calypso/vocabulary"
+      # Plausible config for generating the vocabulary.
+      #"tokens_to_add": {
+      #    "tokens": ["<S>", "</S>"],
+      #    "token_characters": ["<>/S"]
+      #},
+      #"min_count": {"source_tokens": 3}
```
joelgrus: could you add a space between the comment character and the comment, it makes it easier to read

allennlp/data/dataset_readers/simple_language_modeling.py
brendan-ai2: I should have said, "in" instead of "back in". In other words, do you foresee any complications to adding these directly instead of using the functionality provided by https://github.com/allenai/allennlp/blob/master/allennlp/data/token_indexers/token_characters_indexer.py#L29?

allennlp/training/trainer.py
brendan-ai2: Done

allennlp/training/trainer.py
brendan-ai2: Changed. But I'm still getting:
`allennlp/training/trainer.py:464: error: Item Dict[Any, Any] of "Union[Dict[Any, Any], Any]" has no attribute "size"`
Unclear why it's not using the `torch.Tensor` type correctly.

training_config/bidirectional_lm.jsonnet
brendan-ai2: Changed to `//`. I'm impartial. Let me know if that's not what you meant!

training_config/bidirectional_lm.jsonnet
brendan-ai2: Done. Should we host the 1b corpus on S3 as well? It's 1.7GB compressed. Also, since it's in multiple parts we might need to change the code that manages tarred inputs. Haven't looked at that code yet.

training_config/bidirectional_lm.jsonnet
brendan-ai2: Done. Sorry about that. Should have cleaned these up.

training_config/bidirectional_lm.jsonnet
brendan-ai2: Done. I usually avoid this for commented out code, but I will rectify this habit. ;)

allennlp/models/bidirectional_lm.py
DeNeutoy: Ok I see why it's in both places now - you might not use a char cnn, but you might still want `LayerNorm` over your token embeddings. Please could you add to the docstring of the constructor that basically you should leave `do_layer_norm=False` if you are using the `char_cnn_encoder` with the default options?

Yes, it seems that that MaskedLayerNorm definition is actually doing `MaskedBatchNormalisation`. I'll defer to Matt as he wrote that originally in Calypso but it is equally "valid".

allennlp/training/trainer.py
```diff
@@ -543,6 +560,15 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
             if self._log_histograms_this_batch:
                 self._histograms_to_tensorboard(batch_num_total, histogram_parameters)
 
+            if self._log_batch_size_period:
+                cur_batch = self._get_batch_size(batch)
+                cumulative_batch_size += cur_batch
+                if (batches_this_epoch - 1) % self._log_batch_size_period == 0:
+                    average = cumulative_batch_size/batches_this_epoch
+                    logger.info(f"cur batch size: {cur_batch} average size: {average}")
```
DeNeutoy: Full names here please, particularly as we are logging them.

allennlp/training/trainer.py
```diff
@@ -1038,6 +1064,7 @@ def from_params(cls,  # type: ignore
         histogram_interval = params.pop_int("histogram_interval", None)
         should_log_parameter_statistics = params.pop_bool("should_log_parameter_statistics", True)
         should_log_learning_rate = params.pop_bool("should_log_learning_rate", False)
+        log_batch_size_period = params.pop_int("log_batch_size_period", None)
```
DeNeutoy: Do we really need this new parameter? Can't we just log the batch size every `summary_interval` steps? It's not expensive because it's a scalar, and it's not too visually obnoxious because tensorboard compresses graphs by default in the UI.

allennlp/models/bidirectional_lm.py
matt-peters: Agree with @DeNeutoy that we shouldn't apply normalization twice.  @brendan-ai2 your proposal sounds fine to me, this would also remove the `do_layer_norm` parameter from this class and some duplicate code.  Is this the scenario you have been training?

PR #2122, opened 2018-11-30
Fix type annotation for .forward(...) in tutorial

pmandera:
The .forward(...) method in tutorial has an incorrect type annotation.


matt-gardner:
@joelgrus, do we need to redeploy anything?  It's not particularly urgent, I just don't know what's needed here before the tutorial itself gets updated.  A PR on the website repo?

joelgrus:
two things have to happen:

1. someone has to run converter script

```
python tutorials/tagger/convert.py
```

which will regenerate the `default_tutorial.html`

2. then someone has to copy/replace that file into `allennlp-website/_includes`

it's posssible we should automate this, it's also possible it's not worth it

matt-gardner:
Probably not worth automating.  This is also a tiny change that might be easy to do by hand...?

matt-gardner:
Yeah, it was easy (pretty sure I did it right).  Submitting a PR there in a minute...

PR #2124, opened 2018-11-30
more informative message for BasicTextFieldEmbedder mismatched keys

joelgrus:
with BERT + ELMo + OpenAI embedders this becomes a very easy mistake to make, so I changed it to give a more informative error message in the case where the input has too many keys, and left the same error message in the case where there's a total mismatch

### DIFF
--- a/allennlp/modules/text_field_embedders/basic_text_field_embedder.py
+++ b/allennlp/modules/text_field_embedders/basic_text_field_embedder.py
@@ -73,9 +73,13 @@ class BasicTextFieldEmbedder(TextFieldEmbedder):
                 # token embedder keys are a strict subset of text field input keys.
                 message = (f"Your text field is generating more keys ({list(input_keys)}) "
                            f"than you have token embedders ({list(embedder_keys)}. "
-                           f"Most likely you need to add allow_unmatched_keys = True "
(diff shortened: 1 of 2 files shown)


allennlp/modules/text_field_embedders/basic_text_field_embedder.py
```diff
@@ -64,13 +64,28 @@ def get_output_dim(self) -> int:
...
-        if self._token_embedders.keys() != text_field_input.keys():
-            if not self._allow_unmatched_keys:
+        embedder_keys = self._token_embedders.keys()
+        input_keys = text_field_input.keys()
+
+        # Check for unmatched keys
+        if not self._allow_unmatched_keys:
+            if embedder_keys < input_keys:
+                # token embedder keys are a strict subset of text field input keys.
+                message = (f"Your text field is generating more keys ({list(input_keys)}) "
```
DeNeutoy: Hmm I'm not sure this is the "most likely" error - perhaps we can modify this to say "If you are using a token embedder which requires multiple token indexers (such as the OpenaiEmbedder), you need to pass the `allow_unmatched_keys=True` flag to your BasicTextFieldEmbedder config. If not, you should check that there is a 1:1 mapping between your token indexers and token embedders."

PR #2125, opened 2018-11-30
pin msgpack

joelgrus:
I hope this will fix 

https://github.com/explosion/spaCy/issues/2995

(we are pinned to an old version of spacy, which is pinned to an old version of thinc, so we don't get the fix just by updating)


honnibal:
Sorry about this. I've released Thinc v6.12.1 so installations should work correctly. spaCy avoids depending on msgpack directly for just this sort of reason, it's imported via Thinc --- so we only need to change it there.

Btw, your spaCy pin is correct. You've got it pinned to >=2.0,<2.1, which is right. The current stable is v2.0. We only have v2.1 on `spacy-nightly` currently, which you shouldn't depend on (obviously).

PR #2138, opened 2018-12-05
Bidirectional LM Embedder

brendan-ai2:
- A token embedder for transformer based bidirectional LMs.
- Includes demo model: constituency_parser_transformer_elmo.jsonnet
- New tests for training a transformer based bidirectional language model.


brendan-ai2:
@vidurj, fyi

brendan-ai2:
Please take a look!

allennlp/tests/fixtures/bidirectional_lm/characters_token_embedder.json
```diff
@@ -0,0 +1,79 @@
...
+      "text_field_embedder": {
+        "token_embedders": {
+          "tokens": {
+            "type": "embedding",
+            "embedding_dim": 50
+          },
+          "elmo": {
+            "type": "bidirectional_token_embedder",
+            "weight_file": "allennlp/tests/fixtures/bidirectional_lm/lm_weights.th",
+            "text_field_embedder": {
```
brendan-ai2: I'm not terrifically thrilled with this config duplication. (Also used to train the bidirectional LM in the first place.) Is there a way around it?

allennlp/modules/token_embedders/bidirectional_token_embedder.py
```diff
@@ -0,0 +1,112 @@
...
+from allennlp.common.checks import ConfigurationError
+from allennlp.modules import ScalarMix
+from allennlp.modules.text_field_embedders import TextFieldEmbedder
+from allennlp.modules.token_embedders.token_embedder import TokenEmbedder
+from allennlp.modules.seq2seq_encoders import BidirectionalLanguageModelTransformer
+from allennlp.nn.util import device_mapping, get_text_field_mask, remove_sentence_boundaries
+
+
+@TokenEmbedder.register('bidirectional_token_embedder')
+class BidirectionalTokenEmbedder(TokenEmbedder):
```
joelgrus: is "BidirectionalTokenEmbedder" the best name for this? I would have expected the name to mention "language model" and possibly "transformer"

allennlp/modules/token_embedders/bidirectional_token_embedder.py
```diff
@@ -0,0 +1,112 @@
...
+        if dropout:
+            self._dropout = torch.nn.Dropout(dropout)
+        else:
+            self._dropout = lambda x: x
+
+        self._remove_bos_eos = remove_bos_eos
+        self._scalar_mix = ScalarMix(mixture_size=contextualizer.num_layers + 1, do_layer_norm=False, trainable=True)
+
+        state_dict = torch.load(weight_file, map_location=device_mapping(-1))
+        self.load_state_dict(state_dict, strict=False)
```
joelgrus: because I don't think about `load_state_dict` very often, it's not clear to me what this is actually doing. can you add documentation for that?

what I mean is, which parts of this TokenEmbedder are getting which weights from the state_dict? and what does state_dict look like here?

allennlp/modules/token_embedders/bidirectional_token_embedder.py
```diff
@@ -0,0 +1,112 @@
...
+    def get_output_dim(self) -> int:
+        return self._contextualizer.output_dim
+
+    def forward(self,  # type: ignore
+                inputs: torch.Tensor) -> Dict[str, torch.Tensor]:
+        """
+        Parameters
+        ----------
+        inputs: ``torch.Tensor``
+            Shape ``(batch_size, timesteps, 50)`` of character ids representing the current batch.
```
joelgrus: 50 is for ELMo, but could it be a different number here?

allennlp/modules/token_embedders/bidirectional_token_embedder.py
```diff
@@ -0,0 +1,112 @@
...
+        # pylint: disable=arguments-differ
+        source = {"token_characters": inputs}
+        mask = get_text_field_mask(source)
+
+        # shape (batch_size, sentence_length + 2, embedding_size)
+        embeddings = self._text_field_embedder(source)
+
+        contextual_embeddings = self._contextualizer(embeddings, mask)
+
+        # To match contextualized dimension.
```
joelgrus: can you elaborate on this comment?

allennlp/modules/token_embedders/bidirectional_token_embedder.py
```diff
@@ -0,0 +1,112 @@
...
+        double_character_embeddings = torch.cat([embeddings, embeddings], -1)
+        if double_character_embeddings.size(-1) != contextual_embeddings[0].size(-1):
+            raise Exception("Incorrect sizes")
```
joelgrus: how could this happen? can the exception give some guidance as to that?

allennlp/modules/token_embedders/bidirectional_token_embedder.py
```diff
@@ -0,0 +1,112 @@
...
+    remove_bos_eos: ``bool``, optional (default: True)
+        Typically the provided token indexes will be augmented with
+        begin-sentence and end-sentence tokens. If this flag is True
+        the corresponding embeddings will be removed from the return values.
+    requires_grad : ``bool``, optional (default: False)
+        If True, compute gradient of bidirectional language model parameters for fine tuning.
+    """
+    def __init__(self,
+                 weight_file: str,
+                 text_field_embedder: TextFieldEmbedder,
```
joelgrus: I am curious about the use of TextFieldEmbedder here (rather than just a `TokenEmbedder`), is this because you might want to concatenate multiple different embeddings of your character ids?

allennlp/modules/token_embedders/bidirectional_token_embedder.py
joelgrus: also, is character_ids too restrictive here? presumably you might want to use byte pairs or wordpieces or something?

allennlp/modules/token_embedders/bidirectional_token_embedder.py
```diff
@@ -0,0 +1,112 @@
...
+        inputs: ``torch.Tensor``
+            Shape ``(batch_size, timesteps, 50)`` of character ids representing the current batch.
+
+        Returns
+        -------
+        The bidirectional language model representations for the input sequence, shape
+        ``(batch_size, timesteps, embedding_dim)``
+        """
+        # pylint: disable=arguments-differ
+        source = {"token_characters": inputs}
```
joelgrus: if you're going to hardcode the input like this, I think the documentation for the `TextFieldEmbedder` parameter should indicate that it expects a tensor dict that looks like `{"token_characters": ... }`

allennlp/tests/fixtures/bidirectional_lm/characters_token_embedder.json
brendan-ai2: This helped a bunch. Thanks! It did require some restructuring, but it seems like a win.

allennlp/modules/token_embedders/bidirectional_token_embedder.py
brendan-ai2: This is now obsolete. I've restructured things so that we call the `BidirectionalLanguageModel` directly. So there's now substantially less magic. (Thanks to @DeNeutoy for tips on how to accomplish that!)

allennlp/modules/token_embedders/bidirectional_token_embedder.py
brendan-ai2: Fixed. The true restriction is that the indexer used in the end-model must be the same as what the `BidirectionalLanguageModel` was trained with.

allennlp/modules/token_embedders/bidirectional_token_embedder.py
brendan-ai2: Sure!
```
        # Typically character embeddings are smaller than contextualized embeddings. Since we're
        # averaging the character embeddings along with all the contextualized layers we need to
        # make their dimensions match. Simply repeating the character embeddings is a crude, but
        # effective, way to do this.
```

allennlp/modules/token_embedders/bidirectional_token_embedder.py
brendan-ai2: Your comment made me realize this should be detectable at construction time. Moved there.

```
+            raise RuntimeError(
+                "The output dimensions for the text_field_embedder " +
+                f"({character_dim}) and the contextualizer ({contextual_dim})" +
+                f" from the language model loaded from {archive_file} are " +
+                "not compatible. Please check the config used to train that " +
+                "model and ensure that the output dimension of the " +
+                "text_field_embedder divides the output dimension of the " +
+                "contextualizer.")
```

allennlp/modules/token_embedders/bidirectional_token_embedder.py
brendan-ai2: Obsolete now that I'm loading the LM directly. (The reason was simply because the `BidirectionalLanguageModel` used one, so I had to match it precisely to load the weights.)

allennlp/modules/token_embedders/bidirectional_token_embedder.py
brendan-ai2: I've changed this to determine the key from the config for the `BidirectionalLanguageModel`. Feels a bit hacky, but it should be more robust that way.

allennlp/modules/token_embedders/bidirectional_token_embedder.py
brendan-ai2: Yeah, that's fair. Changed to `BidirectionalLanguageModelTokenEmbedder`. Just unfortunate that this is getting a  bit unwieldy... Strictly speaking a transformer contextualizer is no longer required -- just one that supports `num_layers` and `return_all_layers`. I've documented this in the constructor here.

brendan-ai2:
Thanks for the feedback. Very useful! Please take another look.

This revision also includes a fix that injects the bos and eos tokens in the embedder. This follows the approach taken in `elmo.py`.

allennlp/models/bidirectional_lm.py
```diff
@@ -218,65 +212,61 @@ def forward(self,  # type: ignore
...
+        token_ids = source.get("tokens")
+        if token_ids is not None:
+            # Use token_ids to compute targets
+            forward_targets = torch.zeros_like(token_ids)
+            backward_targets = torch.zeros_like(token_ids)
+            forward_targets[:, 0:-1] = token_ids[:, 1:]
+            backward_targets[:, 1:] = token_ids[:, 0:-1]
+
+            # add dropout
+            contextual_embeddings = self._dropout(contextual_embeddings)
```
joelgrus: I feel like I'd rather have this dropout up where it was, just so that when you're reading the model it's obvious it's taking place. (it's less obvious when it happens somewhere else.)

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -0,0 +1,170 @@
...
+        contextualizer used by the LM must satisfy two requirements:
+
+        1. It must have a num_layers field.
+        2. It must take a boolean return_all_layers parameter in its constructor.
+
+        See BidirectionalLanguageModelTransformer for their definitions.
+
+    dropout : ``float``, optional.
+        The dropout value to be applied to the representations.
+    bos_eos_tokens : ``Tuple[str, str]``, optional (default=``None``)
```
joelgrus: this default doesn't match the constructor default

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # Extract the name of the tokens that the LM was trained on.
+        token_names = list(dict_config["model"]["text_field_embedder"]["token_embedders"].keys())
+        if len(token_names) != 1:
+            # A TokenEmbedder embeds a single set of indices. If the LM was trained with multiple embedders, and
+            # thus indices, we won't be able to embed a single set of indices here.
+            #
+            # Note: We only care about embedded indices. This does not include "tokens" which is just used to
+            # compute the loss in BidirectionalLanguageModel.
+            raise RuntimeError(f"LM from {archive_file} trained with multiple embedders!")
```
joelgrus: I would probably call this a `ConfigurationError`

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -0,0 +1,170 @@
...
+        self._token_name = token_names[0]
+
+        # TODO(brendanr): Remove this hack once we have proven we can train without using the
+        # ELMoTokenCharactersIndexer. For the equivalent hack in the ELMo embedder see
+        # https://github.com/allenai/allennlp/blob/master/allennlp/modules/elmo.py#L590.
+        if bos_eos_tokens:
+            dataset_reader_config = config.get("dataset_reader")
+            if dataset_reader_config.get("type") == "multiprocess":
+                dataset_reader_config = dataset_reader_config.get("base_reader")
```
joelgrus: this is unfortunate, but I can't think of a better way

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -0,0 +1,170 @@
...
+            token_indexer_config = dataset_reader_config.get("token_indexers").get(self._token_name)
+            token_indexer: TokenIndexer = TokenIndexer.from_params(token_indexer_config)
```
joelgrus: I am surprised this type annotation is necessary

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -0,0 +1,170 @@
...
+            token_list = [Token(token) for token in bos_eos_tokens]
+            bos_eos_indices = token_indexer.tokens_to_indices(token_list, self._lm.vocab, "key")["key"]
+            self._bos_indices = torch.Tensor(bos_eos_indices[0])
+            self._eos_indices = torch.Tensor(bos_eos_indices[1])
```
joelgrus: wouldn't this whole section be simpler if you just called `Vocabulary.get_token_index`  for the bos and eos tokens? then you wouldn't even need the token indexer?

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -0,0 +1,170 @@
...
+            self._dropout = lambda x: x
+
+        self._remove_bos_eos = remove_bos_eos
+        num_layers = self._lm._contextualizer.num_layers + 1
+        self._scalar_mix = ScalarMix(mixture_size=num_layers, do_layer_norm=False, trainable=True)
+
+        character_dim = self._lm._text_field_embedder.get_output_dim()
+        contextual_dim = self._lm._contextualizer.get_output_dim()
+
+        if not contextual_dim % character_dim == 0 or character_dim > contextual_dim:
```
joelgrus: couldn't this check just be `if contextual_dim % character_dim != 0`, since if character_dim is greater than contextual_dim, then the mod won't be zero (unless contextual_dim is 0, but that can't happen, can it?)

allennlp/models/bidirectional_lm.py
matt-peters: Let's document in the docstring where dropout is applied to make it explicit.

Also, this will have different behavior for the output `lm_embeddings` key depending on whether token_ids is provided as input (with token_ids, dropout will be applied, but won't be applied if token_ids is not provided).  This is confusing.  Recommend creating a new tensor here to compute the loss, and always returning `lm_embeddings` that has not had dropout applied.

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -0,0 +1,170 @@
...
+from allennlp.nn.util import remove_sentence_boundaries, get_text_field_mask, add_sentence_boundary_token_ids
+
+if TYPE_CHECKING:
+    from allennlp.models.bidirectional_lm import BidirectionalLanguageModel
+
+
+@TokenEmbedder.register('bidirectional_lm_token_embedder')
+class BidirectionalLanguageModelTokenEmbedder(TokenEmbedder):
+    """
+    Compute a single layer of representations from a bidirectional language model.
```
matt-peters: Document here how these representations are computed (by fixing weights, and computing a learned scalar average of layers).

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -0,0 +1,170 @@
...
+                                "return_all_layers": True
+                        }
+                }
+        }
+
+        # Load LM and the associated config.
+        archive = load_archive(archive_file, overrides=json.dumps(overrides))
+        self._lm: BidirectionalLanguageModel = archive.model
+        # pylint: disable=protected-access
+        self._lm._softmax_loss = None
```
matt-peters: This is sketchy assigning `None` to a private attribute of another class.  Recommend adding a public method to `BidirectionalLanguageModel` that is something like `delete_softmax` so the language model can control the low level details.

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
matt-peters: +1

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -0,0 +1,170 @@
...
+            self._eos_indices = None
+
+        if dropout:
+            self._dropout = torch.nn.Dropout(dropout)
+        else:
+            self._dropout = lambda x: x
+
+        self._remove_bos_eos = remove_bos_eos
+        num_layers = self._lm._contextualizer.num_layers + 1
+        self._scalar_mix = ScalarMix(mixture_size=num_layers, do_layer_norm=False, trainable=True)
```
matt-peters: You could simplify the code here (re-use the scalar mix logic, and removing the beginning/end of sentence markers, as well as gain other functionality) by passing in a custom `module` to the `Elmo` class: https://github.com/allenai/allennlp/blob/master/allennlp/modules/elmo.py#L76

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -0,0 +1,170 @@
...
+
+        Returns
+        -------
+        The bidirectional language model representations for the input sequence, shape
+        ``(batch_size, timesteps, embedding_dim)``
+        """
+        # pylint: disable=arguments-differ
+        if self._bos_indices is not None:
+            mask = get_text_field_mask({"": inputs})
+            inputs, _ = add_sentence_boundary_token_ids(
```
matt-peters: The mask also needs to be replaced as the second return argument from `add_sentence_boundary_token_ids`

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -0,0 +1,170 @@
...
+        duplicated_character_embeddings = torch.cat(
+                [character_embeddings] * self._character_embedding_duplication_count, -1
+        )
+        contextual_embeddings.append(duplicated_character_embeddings)
+        averaged_embeddings = self._scalar_mix(contextual_embeddings)
+
+        # Add dropout
+        averaged_embeddings = self._dropout(averaged_embeddings)
+        if self._remove_bos_eos:
+            averaged_embeddings, _ = remove_sentence_boundaries(averaged_embeddings, mask)
```
matt-peters: I'm surprised this works in the case of `self._bos_indices is not None`.   `mask` here corresponds to mask of the input, but if you add `<S>, </S>` the mask changes.  You need to capture the updated mask (see comment above) and pass it in here.

Also, if you pass in the custom module to `Elmo`, this logic is handled elsewhere and moved out of this module.

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -0,0 +1,170 @@
...
+        contextual_embeddings = result_dict["lm_embeddings"]
+
+        # Typically character embeddings are smaller than contextualized embeddings. Since we're
+        # averaging the character embeddings along with all the contextualized layers we need to
+        # make their dimensions match. Simply repeating the character embeddings is a crude, but
+        # effective, way to do this.
+        duplicated_character_embeddings = torch.cat(
+                [character_embeddings] * self._character_embedding_duplication_count, -1
+        )
+        contextual_embeddings.append(duplicated_character_embeddings)
```
matt-peters: To be consistent with `Elmo` and the notion that `duplicated_character_embeddings` are the first layer of the model, make `duplicated_character_embeddings` the first element of the list passed into `self._scalar_mix`.

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -0,0 +1,170 @@
...
+            mask = get_text_field_mask({"": inputs})
+            inputs, _ = add_sentence_boundary_token_ids(
+                    inputs, mask, self._bos_indices, self._eos_indices
+            )
+
+        source = {self._token_name: inputs}
+        result_dict = self._lm(source)
+
+        # shape (batch_size, timesteps, embedding_size)
+        character_embeddings = result_dict["character_embeddings"]
```
matt-peters: This name is confusing as these don't have to be character embeddings, how about `uncontextual_token_embeddings` ?

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
brendan-ai2: I've added a TODO for this. I think the readability of calling `ScalarMix` directly is somewhat higher than dispatching to `Elmo`. Having the bos/eos handling be more prominent also doesn't hurt. Let's dispatch to `Elmo` when/if the additional features are required.

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
brendan-ai2: Fixed, thanks. (Using the mask returned by the LM.)

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
joelgrus: I see, the wordpiece and openai indexers also have their own id mappings, but they add them to the vocabulary, so you can do things like this. whereas it looks like the ELMo indexer doesn't. probably it should, but definitely not as part of this PR. can you just add a TODO for that?

brendan-ai2: Good call, thanks for the pointer! TODO added.

PR #2142, opened 2018-12-05
add [CLS] and [SEP] tags to bert indexer

### LINKED ISSUE #2141: BERT indexer should add [CLS] and [SEP] tokens
as pointed out here: https://github.com/allenai/allennlp/pull/2067#issuecomment-444324303

in a quick and dirty experiment, that adds an extra point of validation f1 (94.1 -> 95.2) to my BERT + char NER model.

joelgrus:
fixes #2141 

as mentioned in the issue, this change gets an extra point of dev F1 on the NER model I trained with BERT embeddings + character embeddings

### DIFF
--- a/allennlp/data/token_indexers/wordpiece_indexer.py
+++ b/allennlp/data/token_indexers/wordpiece_indexer.py
@@ -39,13 +39,19 @@ class WordpieceIndexer(TokenIndexer[int]):
         maximum length for its input ids. Currently any inputs longer than this
         will be truncated. If this behavior is undesirable to you, you should
         consider filtering them out in your dataset reader.
+    start_tokens : ``List[str]``, optional (default=``None``)
+        These are prepended to the tokens provided to ``tokens_to_indices``.
+    end_tokens : ``List[str]``, optional (default=``None``)
+        These are appended to the tokens provided to ``tokens_to_indices``.
     """
     def __init__(self,
                  vocab: Dict[str, int],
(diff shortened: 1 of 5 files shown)


joelgrus:
you can always just not pass in the offsets (i.e. not include them in the indexer-embedder map) and get back the per-wordpiece embeddings (including [CLS] and [SEP]) and do whatever you want with them.

I don't know that I'd consider that "cumbersome" but I guess it would depend on what you then needed to do with the embeddings.

qiuwei:
Got it, thanks.

joelgrus:
if it's a pretty common use case, we could also do something like add an optional flag to the `BertIndexer` that would include the offsets for the [CLS] and [SEP] tokens, and then normal usage would bring back (num_tokens + 2) BERT vectors for [CLS], token1, token2, ..., tokenk, [SEP]

joelgrus:
if it's a pretty common use case, we could also do something like add an optional flag to the `BertIndexer` that would include the offsets for the [CLS] and [SEP] tokens, and then normal usage would bring back (num_tokens + 2) BERT vectors for [CLS], token1, token2, ..., tokenk, [SEP]

matt-gardner:
Yeah, @joelgrus, I was thinking the same thing, that this could easily be an option.  If anyone seeing this wants this functionality, please speak up so we know how to prioritize it.  Or just submit a PR adding it =).

PR #2145, opened 2018-12-06
fix bert input order

joelgrus:
@OyvindTafjord discovered that the token embedder was providing inputs to the BERT model in the wrong order. the model was expecting (input_ids, token_type_ids, mask) but we were providing (input_ids, mask, token_type_ids). 

making this fix added another 0.3 dev F1 to the NER model (95.2 -> 95.5)

https://beaker-internal.allenai.org/ex/ex_mu3yx7rblvmu/tasks

it is somewhat amazing / frightening that the model worked so well with this error. 😬


OyvindTafjord:
FWIW, I think the reason it worked reasonably well was that it was effectively getting all zeros as "mask" (since token type is always zero in these experiments) which shifts all values to be tiny by the same amount, so it effectively was like running without any masking which might not be a big deal in this context (the logits of masked elements get shifted by -10000 before a softmax, but it's presumably smart enough not to underflow).

Issue #1906, opened 2018-10-14
Exception when using --recover flag is used


Divyendra:
**Describe the bug**
 *unorderable types: str() < int()* Exception is thrown while using --recover flag

**To Reproduce**
Steps to reproduce the behavior
1. Create dummy files with the following names in the serialization directory
 ['model_state_epoch_0.2018-10-14-08-16-09.th',
 'model_state_epoch_0.th',
 'model_state_epoch_1.2018-10-14-09-52-15.th']
2. Then run a model using the recover flag
3. See error because while  last_epoch = sorted(int_epochs, reverse=True)[0] is being 
executed in trainer.py module

**Expected behavior**
Type Mismatch Exception should not occur

**Additional context**
I guess the the code at line 923 in *trainer* module has to be:
`     int_epochs.append([int(pieces[0]), '0'])`
The second arg should be a string which is not currently.

Please have a look

PR #2153, opened 2018-12-07
empty time stamp should be a str instead of int 0

seal6363:
Fixes https://github.com/allenai/allennlp/issues/1906


schmmd:
@seal6363 if you prefix with "fixes" it will close the issue when this PR is merged.  I changed the description.

schmmd:
@seal6363 looks good--but would you add a test that fails before this change but succeeds after it?  That would provide great validation that we're indeed fixing an issue.

schmmd:
We decided not to add a test because it would involve interrupting a training (which is hard to do).  Kathryn confirmed the error existed before this change and went away after it.  She's now investigating the failed test.

PR #2157, opened 2018-12-09
Fix mismatches between setup.py and requirements.txt

epwalsh:
cffi and flask-cors were behind in setup.py


epwalsh:
Also, I should mention that I found this by running `./scripts/verify.py` because `.scripts/check_requirements_and_setup.py` failed. But I was surprised that this wasn't caught in a TeamCity build. I guess I assumed that TeamCity just ran `./scripts/verify.py`, but I guess that isn't the case?

PR #2162, opened 2018-12-10
Fixes broken link and rephrasing visualizing_model_internals.md

### LINKED ISSUE #2088: Prerequisite of model visualization
 - OS: Ubuntu 18.04
 - Python version: 3.6
 - AllenNLP version: 0.7.1
 

**Question**
Hi Team,

The question is specific to improving the description https://github.com/allenai/allennlp/blob/master/tutorials/how_to/visualizing_model_internals.md. 

It took me some time to figure out how things were working in the react demo. I think the first point in prerequisite should be modified a bit in such a way that it provides the following description

* That we need a flask server running for the react server to interact with all time. So including the command example, that needs to be executed ```python -m allennlp.service.server_simple     --archive-path /tmp/irony_alternate1/model.tar.gz     --predictor ironic-predictor     --include-package irony_model``` in my opinion is a good idea.

Thanks

rangwani-harsh:
Fixes the broken links and closes #2088

### DIFF
--- a/tutorials/how_to/visualizing_model_internals.md
+++ b/tutorials/how_to/visualizing_model_internals.md
@@ -13,9 +13,10 @@ parts of our models need fixing.
 
 ## Pre-requisites
 
-1. You need to have a text in / text out demo of your model already running using our simple
-   server.  See the [tutorial on creating a
-demo](../getting_started/making_predictions_and_creating_a_demo.md) for how to do this.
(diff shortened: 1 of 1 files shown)


schmmd:
@matt-gardner it doesn't run in CI because it hits quite a few URLs and we were worried about getting blocked as a bot.  It's part of `verify.py` and should be run with each release--but we need to add that to the release script.  See https://github.com/allenai/allennlp-internal/issues/93.

PR #2165, opened 2018-12-10
pytorch 1.0

joelgrus:
should still be compatible with 0.4.1 as well.

still has one failing test (hence the WIP), I think it's just underspecified, but I'll try to figure that out today


joelgrus:
I just loosened the test and added a TODO to tighten the test specification. This should be ready for review now.

PR #2172, opened 2018-12-12
Minor fixes to help evaluate ELMo port.

brendan-ai2:
- Display loss when running evaluate.
  - Fixes https://github.com/allenai/allennlp/issues/1241
- Log when sentences are dropped due to exceeding the maximum length.

### DIFF
--- a/allennlp/commands/evaluate.py
+++ b/allennlp/commands/evaluate.py
@@ -100,10 +100,27 @@ def evaluate(model: Model,
                                  shuffle=False)
         logger.info("Iterating over dataset")
         generator_tqdm = Tqdm.tqdm(iterator, total=data_iterator.get_num_batches(instances))
+
+        num_batches = 0
+        cumulative_loss = 0.0
+        has_loss = False
+
         for batch in generator_tqdm:
             batch = util.move_to_device(batch, cuda_device)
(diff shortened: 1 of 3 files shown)


brendan-ai2:
A tiny one. Please take a look!

allennlp/commands/evaluate.py
```diff
@@ -100,10 +100,27 @@ def evaluate(model: Model,
...
+                has_loss = "loss" in output_dict
+            num_batches += 1
+
             metrics = model.get_metrics()
+
+            if has_loss:
+                loss = output_dict["loss"].item()
+                metrics["loss"] = loss
+                cumulative_loss += loss
+
```
joelgrus: is the consistency a real worry? I think you could make this a tiny bit simpler, something like

```
loss_count = 0
total_loss = 0.0

# ...

loss = model(**batch).get("loss")

# ...

if loss is not None:
    metrics["loss"] = loss.item()
    total_loss += loss.item()
    loss_count += 1

# ...

if loss_count > 0:
    final_metrics["loss"] = total_loss / loss_count
```

I feel like that's cleaner, what do you think?

brendan-ai2:
Thanks!

PR #2179, opened 2018-12-13
Made bz2 an optional dependency

### LINKED ISSUE #2178: ModuleNotFoundError: No module named '_bz2'.
The tests fail after installing the allennlp repo through pip on Ubuntu 16.4 with Python 3.7.

As far as I see this is because bz2 is an optional dependency when compiling Python 3+.

I tried to fix this system level but nothing seemed to work on my setup. So I made a quick fix.

Open the embedding.py via an editor.

`sudo gedit /home/semih/.virtualenvs/playground3/lib/python3.7/site-packages/allennlp/modules/token_embedders/embedding.py`

Then find the line where bz2 is imported and remove it.

Find where the file extension is being checked. 

`extension = get_file_extension(main_file_uri)`

After the previous line add the following two lines.

```
if ".bz2" in extension:
    import bz2 # import only when necessary
```
Save the file and re-run the tests. It should be working now.

semihyagcioglu:
bz2 was a **hard dependency** and it was throwing a "ModuleNotFoundError: No module named '_bz2'." error for me.  

I made it an optional dependency by moving the import to **function level** and made sure the dependency is **only imported when needed** by checking the file extension.

This fixes #2178

### DIFF
--- a/allennlp/modules/token_embedders/embedding.py
+++ b/allennlp/modules/token_embedders/embedding.py
@@ -428,7 +428,7 @@ class EmbeddingsTextFile(Iterator[str]):
             extension = get_file_extension(main_file_uri)
             
             if ".bz2" in extension:
-	            import bz2 # import only when necessary
+	            import bz2  # import only when necessary
             
             package = {
                     '.txt': io,


allennlp/modules/token_embedders/embedding.py
```diff
@@ -427,6 +426,10 @@ def __init__(self,
 
             # All the python packages for compressed files share the same interface of io.open
             extension = get_file_extension(main_file_uri)
+            
+            if ".bz2" in extension:
+	            import bz2  # import only when necessary
+            
             package = {
                     '.txt': io,
                     '.vec': io,
```
matt-gardner: Oops, I approved too soon - this is causing the tests to fail, as `bz2` is undefined here unless the `if` above is triggered.  Can you fix this?

semihyagcioglu: Hi @matt-gardner. Now that I have tested on another system which is also an Ubuntu 16.04, I see that the Python 3.6 and above makes it a bit problematic with both the bz2 and lzma libraries. I had to recompile Python several times with no luck. Therefore I suggest dropping the support for these two libraries altogether. 

I believe this would help increase the use of repo. I did remove these two dependencies in my system and made the tests pass. If you agree that these two file types is not crucial for the repo at the moment, I can send another pull request that would drop the dependencies.

matt-gardner:
I tried to update this PR with some fixes, but I had a hard time because you were developing on your master branch instead of a feature branch.  So I just opened a new PR with the changes.  I'm closing this in favor of #2196.

PR #2183, opened 2018-12-14
Evaluate on a token-weighted basis.

brendan-ai2:
- Allow models to return a `"batch_weight"` key that will be used to weight each batch's loss.
  - Per @matt-peters' suggestion.
  - Performed in Calypso: https://github.com/allenai/calypso/blob/master/calypso/train.py#L699
- Remove unused "loss_scale". This was never set in the training config, so it should be fairly safe.


brendan-ai2:
Please take a look!

allennlp/models/bidirectional_lm.py
```diff
@@ -262,7 +262,8 @@ def forward(self,  # type: ignore
             return_dict = {
                     'loss': average_loss * scale_factor,
```
brendan-ai2: One outstanding issue that I couldn't think of a good way to handle is adjusting for the `scale_factor`. In Calypso the adjustment is made in `train.py`, but that only works because that code knows it's dealing with a specific model. Two questions:

1. What is the purpose behind `scale_factor`? Can/should we drop it?
2. If not, can you think of clean way to make it work with `"batch_weight"`?

allennlp/commands/evaluate.py
```diff
@@ -101,21 +101,35 @@ def evaluate(model: Model,
...
+            output_dict = model(**batch)
+            loss = output_dict.get("loss")
 
             metrics = model.get_metrics()
 
             if loss is not None:
                 loss_count += 1
-                metrics["loss"] = loss.item()
-                total_loss += loss.item()
+                weight = output_dict.get("batch_weight")
```
joelgrus: I'm not a fan of magic constants like this ("loss" is kind of a necessary evil).

How would you feel about having it be a command line arg `batch_weight_key` or something. As a bonus, that would give us a place to *document* this feature, otherwise no one will even know about it unless they read the code.

allennlp/common/testing/model_test_case.py
```diff
@@ -215,6 +215,9 @@ def ensure_batch_predictions_are_consistent(self):
                     # Loss is particularly unstable; we'll just be satisfied if everything else is
                     # close.
                     continue
+                if 'batch_weight' == key:
```
joelgrus: doing yoda-style comparisons doesn't feel idiomatic in python (and what they protect against is an error in python already)

```
In [1]: x = 10

In [2]: if x = 1: print("oops")
  File "<ipython-input-2-380718274afc>", line 1
    if x = 1: print("oops")
         ^
SyntaxError: invalid syntax
```

joelgrus: also, again I don't like the "magical" behavior here. is there a solution where `ensure_batch_predictions_are_consistent` has an optional arg for keys to ignore, and then the relevant test can just specify that?

allennlp/tests/commands/evaluate_test.py
```diff
@@ -1,11 +1,36 @@
...
+            yield {}
+
+
+class DummyModel(Model):
+    def __init__(self, outputs: List):
+        super().__init__(None) # type: ignore
+        self._outputs = outputs
+
+    def forward(self, *args, **kwargs) -> Dict[str, torch.Tensor]:  # pylint: disable=arguments-differ
+        return self._outputs.pop(0)
```
joelgrus: I think this would be cleaner if `DummyModel` just passed through its inputs and `DummyIterator` supplied the outputs, i.e.

```
class DummyIterator(DataIterator):
    def __init__(self, tensor_dicts: List[TensorDict]) -> None:
        super().__init__()
        self._tensor_dicts = tensor_dicts

     def __call__(self, *args, **kwargs) -> Iterator[TensorDict]:
         yield from self._tensor_dicts

class DummyModel(Model):
    def __init__(self):
        super().__init__() 

     def forward(self, tensor_dict: TensorDict) -> Dict[str, torch.Tensor]:  # pylint: disable=arguments-differ
        return tensor_dict
```

does that still work?

allennlp/models/bidirectional_lm.py
brendan-ai2: Thanks, dropped!

allennlp/tests/commands/evaluate_test.py
brendan-ai2: Sounds reasonable. I had to modify it slightly given that we unpack the args passed to our models. Ended up with the following:

```
class DummyModel(Model):
    def __init__(self):
        super().__init__(None) # type: ignore

    def forward(self, **kwargs) -> Dict[str, torch.Tensor]:  # pylint: disable=arguments-differ
        return kwargs
```

brendan-ai2:
Thank you both for the review!

Issue #2180, opened 2018-12-13
Can't create a predictor that depends on files_to_archive


vlthr:
**Describe the bug**
I am trying to reconstruct a model and create a predictor from an archive as follows:
```python
archive = load_archive("./model.tar.gz")
predictor = Predictor.from_archive(archive, "textcat")
```
The model uses a DatasetReader whose tokenizer (and TokenIndexer) loads a file. I am able to include that file in `Params.files_to_archive`, and have verified that they are created in the temporary folder where the archive is extracted. 

However - if I'm understanding this correctly - since `load_archive` cleans up the temporary folder before returning, `Predictor.from_params` fails when trying to recreate the DatasetReader.

```
Traceback (most recent call last):
  ...
  File "server.py", line 36, in __init__
    predictor = Predictor.from_archive(archive, "textcat")
  File "/Users/vlthr/build/allennlp/allennlp/predictors/predictor.py", line 127, in from_archive
    dataset_reader = DatasetReader.from_params(dataset_reader_params)
  File "/Users/vlthr/build/allennlp/allennlp/common/from_params.py", line 274, in from_params
    return subclass.from_params(params=params, **extras)
  File "/Users/vlthr/build/allennlp/allennlp/common/from_params.py", line 285, in from_params
    kwargs = create_kwargs(cls, params, **extras)
  File "/Users/vlthr/build/allennlp/allennlp/common/from_params.py", line 182, in create_kwargs
    value_dict[key] = value_cls.from_params(params=value_params, **extras)
  File "/Users/vlthr/build/allennlp/allennlp/common/from_params.py", line 274, in from_params
    return subclass.from_params(params=params, **extras)
  File "./submodules/allen_utils/data/token_indexers/sentencepiece_indexer.py", line 92, in from_params
    params.add_file_to_archive("model_path")
  File "/Users/vlthr/build/allennlp/allennlp/common/params.py", line 157, in add_file_to_archive
    self.files_to_archive[f"{self.history}{name}"] = cached_path(self.get(name))
  File "/Users/vlthr/build/allennlp/allennlp/common/file_utils.py", line 103, in cached_path
    raise FileNotFoundError("file {} not found".format(url_or_filename))
FileNotFoundError: file /var/folders/_r/szy7s3dd7p7cl0gh5w99sf100000gn/T/tmptb4kzi3o/fta/dataset_reader.token_indexers.tokens.model_path not found

```

**To Reproduce**
Steps to reproduce the behavior
[...]

PR #2184, opened 2018-12-14
Clean up temporary archive directory at exit

vlthr:
Fixes the issue described in #2180 using the `atexit` hook proposed by @joelgrus.

Please let me know if there are any changes you'd like to see made and I'd be happy to make them.

### DIFF
--- a/allennlp/models/archival.py
+++ b/allennlp/models/archival.py
@@ -3,6 +3,7 @@ Helper functions for archiving models and restoring archived models.
 """
 
 from typing import NamedTuple, Dict, Any
+import atexit
 import json
 import logging
 import os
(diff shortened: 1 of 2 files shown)


joelgrus:
thanks, this looks good! 

if you don't mind, can you also add a test for this behavior and verify that the test fails before your change and works after your change?

I *think* it's as simple as just adding one line here:

https://github.com/allenai/allennlp/blob/master/allennlp/tests/models/archival_test.py#L94

that checks whether a file exists at that temporary training data path

vlthr:
There we go! As you guessed a single line did the trick. I also verified that the test does indeed fail on master.

joelgrus:
for future reference (in case anyone looks back at this PR) I wish there was also a test that checked that the deletion really happens "atexit" but I have no idea how to check that in a unit test

Issue #2057, opened 2018-11-15
ProductionRuleField not compatible with multiple GPUs


YoPatapon:
**System (please complete the following information):**
 - OS: Linux
 - Python version: 3.7
 - AllenNLP version: 0.7.1
 - PyTorch version: 0.4.1

**Question**
I am training mml parser on the wikitables dataset. I found it would be out of cuda memory if I use a larger batch size than 8 on a single Tesla P100. Even the batch size 8 causes the out-of-memory error sometimes. How can I train on multiple gpus with allennlp? I don't think the `cuda_device` field supports list input in the config file for now.

matt-gardner:
`cuda_device` supports a list as input in the config file.  Did you try this and it didn't work?

YoPatapon:
Yes, I tested with `cuda_deivce=[1,2]` and is raised an error: 

> RuntimeError: Expected object of type torch.cuda.LongTensor but found type torch.LongTensor for argument #3 'index'
2018-11-16 10:47:37,239 - INFO - allennlp.semparse.executors.wikitables_sempre_executor - Stopped SEMPRE server

matt-gardner:
Can you give more of the stack trace?

YoPatapon:
OK, the entire error trace is:

> Traceback (most recent call last):
  File "/home/v-dozhao/anaconda3/envs/py3.7torch/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/v-dozhao/anaconda3/envs/py3.7torch/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/v-dozhao/Work/allennlp/allennlp/run.py", line 18, in <module>
    main(prog="allennlp")
  File "/home/v-dozhao/Work/allennlp/allennlp/commands/__init__.py", line 72, in main
    args.func(args)
  File "/home/v-dozhao/Work/allennlp/allennlp/commands/train.py", line 111, in train_model_from_args
    args.force)
  File "/home/v-dozhao/Work/allennlp/allennlp/commands/train.py", line 142, in train_model_from_file
    return train_model(params, serialization_dir, file_friendly_logging, recover, force)
  File "/home/v-dozhao/Work/allennlp/allennlp/commands/train.py", line 352, in train_model
    metrics = trainer.train()
  File "/home/v-dozhao/Work/allennlp/allennlp/training/trainer.py", line 755, in train
[...]

PR #2199, opened 2018-12-18
Failing test case for multi-GPU ProductionRuleField

matt-gardner:
#1944 was supposed to handle complex scattering, but it doesn't work for the `ProductionRuleField`, as brought up in #2057.  I have reproduced this and made a failing test case, which is in this PR.

I'm afraid I'm not going to be able to actually fix the underlying problem, though, as I don't really understand the voodoo that @brendan-ai2 did to convert tensors to pointers and back, and why it doesn't work in this case.  The issue is that the `ProductionRuleField` produces a list of [`ProductionRules`](https://github.com/allenai/allennlp/blob/0663762960a4522d4740eea460f63a06eeba126d/allennlp/data/fields/production_rule_field.py#L10-L14), with tensors internal to the data structure that don't get sent to GPUs - they all remain on the CPU.

### DIFF
--- a/allennlp/tests/training/trainer_test.py
+++ b/allennlp/tests/training/trainer_test.py
@@ -17,7 +17,8 @@ from allennlp.data import Vocabulary
 from allennlp.common.params import Params
 from allennlp.models.simple_tagger import SimpleTagger
 from allennlp.data.iterators import BasicIterator
-from allennlp.data.dataset_readers import SequenceTaggingDatasetReader
+from allennlp.data.dataset_readers import SequenceTaggingDatasetReader, WikiTablesDatasetReader
+from allennlp.models.archival import load_archive
 from allennlp.models.model import Model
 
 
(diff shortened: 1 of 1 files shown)


brendan-ai2:
Thanks for the test, @matt-gardner . I think we should move away from the scattering. It's complicated and I've also noticed some performance weirdness. As an alternative I tried simply taking a batch for each GPU in this PR: https://github.com/allenai/allennlp/pull/2200/files#diff-043dcd121296c3cef3f3ff8c74127ff1 (Includes unrelated changes. Relevant changes in trainer.py.) @matt-peters had some concerns about this approach earlier in the quarter, but I think we should revisit it as it now seems simpler and more robust.

matt-peters:
My $0.02 - if we do make modifications we should move toward using the built in pytorch multiple GPU training instead of rolling our own (and then worrying about how to fix it in edge cases).  If you go the route of separate batches in the iterator for each GPU, then you will also need to make the bucket iterator multiple GPU aware as it will significantly hamper performance otherwise.

brendan-ai2:
@matt-peters and I spent a little time today trying to understand the performance of the Transformer ELMo reimplementation w.r.t. max sequence length. His initial concern, if I'm articulating it correctly, was that using a batch per GPU would result in batches with small max lengths being computed concurrently with batches with large max lengths. (Even though the total number of tokens would be roughly equal due to using `maximum_samples_per_batch`.) Then, given the quadratic nature of transformers, the GPU with the small max length would finish first and sit idle.

Fortunately, we observed this to not be the case. For batches with maximum lengths ranging from 14 tokens to 388 tokens the forward pass time ranged from 0.040 seconds to just 0.065 seconds. Further, the larger max length batches generally finished faster as they had fewer total tokens. On a per-token basis they're slower, but runtimes lie within a fairly narrow range. Any quadratic time computation is getting washed out. Spreadsheet for a handful of datapoints: https://docs.google.com/spreadsheets/d/1tRj0bntfoBNUgnpMOwj--HK6MTRRltABXcHLi7z2-D0/edit?usp=sharing.

Given this, it seems like using a batch per GPU is reasonable in the interim. Especially considering that the existing scattering implementation is causing imbalances in GPU memory for reasons we don't currently understand. In the longer term I heartily agree with @matt-peters that we should move to the built-in multi-GPU training if at all possible.

brendan-ai2:
Thanks for the ping. I've merged this PR into my fix: https://github.com/allenai/allennlp/pull/2200/. That's still waiting on the trainer refactor.

PR #2203, opened 2018-12-18
[WIP] : Add some database reader

### LINKED ISSUE #2193: DataBaseReader
Make Dataset more agnostic and make possible to use Postgres, Mongo or ElasticSearch Source 

Create New DatabaseReader using SQLAlchemy, pyMongo, pyElasticsearch clients.

I'm on it

rcourivaud:
Fixes #2193

## TODO :
- Finalize global architecture
- Implement connectors with ElasticSearch and Redis
- We should look to share some structure between DatasetReader and DatabaseReader
- Propose some tutorials


joelgrus:
why are you making a separate class, isn't a "database reader" just a DatasetReader that reads from a database rather than disk? what's otherwise different about them?

you really really want your database reader to be a `DatasetReader`; otherwise it won't work with the rest of the library.

rcourivaud:
Yes I agree, but we need more informations than just a filepath. 
It cause problems in overriding DatasetReader `_read(self, file_path: str)`

How would you do that ?

joelgrus:
I haven't thought about this a ton. One possibility is to have all the interesting information passed into the constructor -- imagine a SQLDatasetReader that needs `connection_string` and `query` as config parameters, and then the `_read() method can just ignore the filepath`

Another possibility is to put the connection string etc in the constructor and then to use the SQL query itself as the filepath, so that your config would look like

```
train_data_path: "SELECT question, passage, answer FROM squad_train WHERE hard = true"
```

or something.

if I were doing it that way, I might try to have the SQLDatasetReader wrap a base dataset reader and then have the `_read` method make a SQL query and then call the base reader's `text_to_instance`, to avoid code duplication everywhere.

does that make sense? does that cause problems with the way you're thinking about this?

matt-gardner:
Yeah, I agree with @joelgrus here.  `DatabaseReader` should inherit from `DatasetReader`, and I'd probably have just one class that can handle any of those three database types (with lazy imports).  Create the DB connection in the constructor, and have the concrete `_read()` method for a particular dataset take a query (or some string that determines the query, if you want).  If the way the cursor behaves in the different databases is different enough, maybe you can't have a single `DatabaseReader` class, but it'd be nice if you could.

rcourivaud:
Yes it make sense, thanks for your feedback, I'll do it like this. 

All configuration in constructor with query in read to make possible to make validation and training datasets

DeNeutoy:
closing due to inactivity - i'm also unsure this is a feature we need to support in the library directly.

PR #2205, opened 2018-12-18
fix lowercase-ization in bert indexer

### LINKED ISSUE #2204: Potential issue with do_lower_case parameter being ignored in PretrainedBertIndexer
**System:**
 - OS: OSX
 - Python version: 3.6.5
 - AllenNLP version: 0.7.2
 - PyTorch version: 0.4.1

**Question**

We are testing a simple Sequence classifier using the PretrainedBertEmbedder.

For that, we are using the PretrainedBertIndexer with the `bert-base-uncased` pre-trained model, as follows:

```json
"token_indexers": {
            "bert": {
                "type": "bert-pretrained",
                "do_lowercase": true,
                "pretrained_model": "bert-base-uncased"
            }
        }
```
Unfortunately, after debugging the code, we have seen the following:

- ``PretrainedBertIndexer`` instantiates a bert tokenizer as follows: https://github.com/allenai/allennlp/blob/7dbd7d34a2f1390d1ff01f2e9ed6f8bdaaef77eb/allennlp/data/token_indexers/wordpiece_indexer.py#L203
- Which then instantiates a ``WordTokenIndexer`` as follows: https://github.com/allenai/allennlp/blob/7dbd7d34a2f1390d1ff01f2e9ed6f8bdaaef77eb/allennlp/data/token_indexers/wordpiece_indexer.py#L204
 - Passing as wordpiece_tokenizer the ``bert_tokenizer.wordpiece_tokenizer.tokenize``: https://github.com/allenai/allennlp/blob/7dbd7d34a2f1390d1ff01f2e9ed6f8bdaaef77eb/allennlp/data/token_indexers/wordpiece_indexer.py#L205

This causes the instantiated indexer to not apply lowercasing and assign the id ``100`` (OOV token id) for every word with upper-case letters.

Maybe we are not configuring correctly the indexer or not grasping the internals. We have checked and ``BertTokenizer`` instantiates two tokenizers, the basic one which indeed does lowercasing and the wordpiece tokenizer which ignores the lowercasing is the one mentioned and used above. We have also seen that the method ``tokenize`` performs lowercasing by applying the basic tokenizer and then does the wordpiece tokenization. 
We believe that the following line (https://github.com/allenai/allennlp/blob/7dbd7d34a2f1390d1ff01f2e9ed6f8bdaaef77eb/allennlp/data/token_indexers/wordpiece_indexer.py#L205) is intentional, but we have tested by replacing with:

```python
super().__init__(vocab=bert_tokenizer.vocab,
                         wordpiece_tokenizer=bert_tokenizer.tokenize,..)
```
And tokenization seems to work, at least for the lowercasing we were expecting, we actually are not sure about the implications of this change.

We have filled this as a question rather than a bug, with the hope is helpful for better understanding the internals. If this is a bug, we would love to create a test, fill an issue and propose a solution.

Keep up the good work!

joelgrus:
fixes #2204

### DIFF
--- a/allennlp/data/token_indexers/wordpiece_indexer.py
+++ b/allennlp/data/token_indexers/wordpiece_indexer.py
@@ -15,6 +15,10 @@ logger = logging.getLogger(__name__)
 
 # TODO(joelgrus): Figure out how to generate token_type_ids out of this token indexer.
 
+# This is the default list of tokens that should not be lowercased.
+_NEVER_LOWERCASE = ['[UNK]', '[SEP]', '[PAD]', '[CLS]', '[MASK]']
+
+
 class WordpieceIndexer(TokenIndexer[int]):
     """
     A token indexer that does the wordpiece-tokenization (e.g. for BERT embeddings).
@@ -41,6 +45,12 @@ class WordpieceIndexer(TokenIndexer[int]):
         consider filtering them out in your dataset reader.
     do_lowercase : ``bool``, optional (default=``False``)
         Should we lowercase the provided tokens before getting the indices?
+        You would need to do this if you are using an -uncased BERT model
+        but your DatasetReader is not lowercasing tokens (which might be the
+        case if you're also using other embeddings based on cased tokens).
+    never_lowercase: ``List[str]``, optional
+        Tokens that should never be lowercased. Default is
+        ['[UNK]', '[SEP]', '[PAD]', '[CLS]', '[MASK]'].
     start_tokens : ``List[str]``, optional (default=``None``)
         These are prepended to the tokens provided to ``tokens_to_indices``.
     end_tokens : ``List[str]``, optional (default=``None``)
@@ -53,6 +63,7 @@ class WordpieceIndexer(TokenIndexer[int]):
                  use_starting_offsets: bool = False,
                  max_pieces: int = 512,
                  do_lowercase: bool = False,
+                 never_lowercase: List[str] = None,
                  start_tokens: List[str] = None,
                  end_tokens: List[str] = None) -> None:
         self.vocab = vocab
@@ -68,6 +79,12 @@ class WordpieceIndexer(TokenIndexer[int]):
         self.max_pieces = max_pieces
         self.use_starting_offsets = use_starting_offsets
         self._do_lowercase = do_lowercase
+
+        if never_lowercase is None:
+            # Use the defaults
+            self._never_lowercase = set(_NEVER_LOWERCASE)
+        else:
+            self._never_lowercase = set(never_lowercase)
 
         # Convert the start_tokens and end_tokens to wordpiece_ids
         self._start_piece_ids = [vocab[wordpiece]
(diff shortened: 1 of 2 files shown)


joelgrus:
I'm not thrilled about how many moving pieces there are; for example, the `BertWordSplitter` does lowercasing by default, but there's no easy way to enforce consistency.

I thought about logging a warning if you try to instantiate a token-indexer ending in `-cased` with `do_lowercase=True` and vice versa but it feels sort of fragile

OyvindTafjord:
I agree this is a bit of a mess. On one hand it's perhaps okay that your initial tokens have upper case letters while using the uncased word pieces. On the other hand, if you want to manually construct your tokens to include the special "[CLS]" and "[SEP]" tokens in particular locations, you're now out of luck as this will turn them into "[cls]" etc (still works if they're added as `start_tokens`/`end_tokens`).

allennlp/data/token_indexers/wordpiece_indexer.py
```diff
@@ -39,6 +39,8 @@ class WordpieceIndexer(TokenIndexer[int]):
         maximum length for its input ids. Currently any inputs longer than this
         will be truncated. If this behavior is undesirable to you, you should
         consider filtering them out in your dataset reader.
+    do_lowercase : ``bool``, optional (default=``False``)
+        Should we lowercase the provided tokens before getting the indices?
```
matt-gardner: Might be good to give a better description of _when_ you might want to do this.

allennlp/data/token_indexers/wordpiece_indexer.py
```diff
@@ -108,8 +112,10 @@ def tokens_to_indices(self,
         offset = len(wordpiece_ids) if self.use_starting_offsets else len(wordpiece_ids) - 1
 
         for token in tokens:
+            # Lowercase if necessary
+            text = token.text.lower() if self._do_lowercase else token.text
```
OyvindTafjord: One ugly workaround (for my other comment) would be to have a short list of exception tokens here which don't get lower cased (e.g., everything in square brackets, or [PAD], [UNK], [CLS], [SEP], [MASK])

joelgrus:
I mean, an alternative approach is to do nothing and just insist that your dataset reader needs to handle the lowercasing properly (or not) and potentially log warnings (or crash) if things seem out of sync?

OyvindTafjord:
Right, but there might be a legitimate use case where you keep cased tokens (for other purposes), but still want to use uncased for the purposes of BERT.

thomwolf:
If we can help on this by changing things up-stream, don't hesitate to ping me @joelgrus!

allennlp/data/token_indexers/wordpiece_indexer.py
joelgrus: yeah, I'll add something like this

joelgrus:
@thomwolf thanks for the offer. like I said this is rooted in the way allennlp handles the (sentence -> words) and the (word -> wordpieces) parts of tokenization differently, so off the top of my head there's not much you could do

joelgrus:
@OyvindTafjord pushed up a commit that incorporates your workaround, let me know how you feel about it

OyvindTafjord:
LGTM

Issue #1663, opened 2018-08-24
Remove constraint_type argument in CrfTagger


nelson-liu:
Original PR: https://github.com/allenai/allennlp/pull/1661

Removed in v0.6.1

PR #2208, opened 2018-12-18
Remove constraint_type parameter from CRF Tagger

schmmd:
Fixes https://github.com/allenai/allennlp/issues/1663


joelgrus:
it looks like you have a test fixture that still uses the deprecated parameter

schmmd:
Yep--fixing that now.  I think I accidentally lost some changes before pushing.

joelgrus:
I am worried you're going to fail the sniff tests b/c of pretrained models using the deprecated parameter

PR #2217, opened 2018-12-19
Release v0.8.0

schmmd:


schmmd:
@joelgrus unfortunately there's a build break since I'm missing a commit.  I should amend the release script to fail when the release branch already exists--I ended up in a weird state.

PR #2224, opened 2018-12-21
Simplifies Subsequent Mask 

vidurj:
The subsequent_mask function in the bidirectional_language_model_transformer seemed to be unnecessarily convoluted.

### DIFF
--- a/allennlp/modules/seq2seq_encoders/bidirectional_language_model_transformer.py
+++ b/allennlp/modules/seq2seq_encoders/bidirectional_language_model_transformer.py
@@ -39,7 +39,7 @@ def attention(query: torch.Tensor,
 
 def subsequent_mask(size: int, device: str = 'cpu') -> torch.Tensor:
     """Mask out subsequent positions."""
-    mask = torch.tril(torch.ones(size, size, device=device)).unsqueeze(0)
+    mask = torch.tril(torch.ones(size, size, device=device, dtype=torch.int32)).unsqueeze(0)
     return mask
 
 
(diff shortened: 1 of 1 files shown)


allennlp/modules/seq2seq_encoders/bidirectional_language_model_transformer.py
```diff
@@ -39,10 +39,8 @@ def attention(query: torch.Tensor,
 
 def subsequent_mask(size: int, device: str = 'cpu') -> torch.Tensor:
     """Mask out subsequent positions."""
-    attn_shape = (1, size, size)
-    mask = np.triu(np.ones(attn_shape), k=1).astype('uint8')
-    mask = (torch.from_numpy(mask) == 0)
-    return mask.to(device)
+    mask = torch.tril(torch.ones(size, size, device=device)).unsqueeze(0)
```
brendan-ai2: Shouldn't the type of `uint8` be preserved?

vidurj: The old return type was bool because of `torch.from_numpy(mask) == 0`. At the only place this is being used in the code base currently we have `subsequent = subsequent_mask(timesteps, device).int()` so the return type doesn't matter. i could cast to `int` to match this use case if that's preferable.

brendan-ai2:
Thanks for creating things as int directly. Just needs a small lint fix: `W: 15, 0: Unused numpy imported as np (unused-import)`

vidurj:
Thanks. Merging.

PR #1928, opened 2018-10-20
Clean up simple_seq2seq tests

epwalsh:
This just cleans up the tests for `simple_seq2seq` by using parameter overrides instead of separate config files for different tests.

PR #1857, opened 2018-10-02
Add QuaRel semantic parser

OyvindTafjord:
(same as https://github.com/allenai/allennlp/pull/1812, just changing owner)

This adds the QuaRel semantic parser for parsing questions about qualitative relations.

PR #2237, opened 2018-12-24
Add CopyNet seq2seq model

epwalsh:
This is my followup to the discussion [here](https://github.com/allenai/allennlp/pull/1928#issuecomment-431616137) with @matt-gardner. This implements a [sequence to sequence model with a copying mechanism](https://arxiv.org/abs/1603.06393) that I've been working on [here](https://github.com/epwalsh/nlp-models/blob/master/nlpete/models/copynet.py). I've been using it for the couple months and I feel confident now that it's ready to be contributed.

Unfortunately I haven't had time to apply it to any standard benchmark datasets, although I've been using it extensively on proprietary datasets. I know @Alex-Fabbri has given it a go on the CNN / Daily Mail dataset, hopefully with some success? If anyone has another dataset they could suggest for a reasonable benchmark I could give that a try. I did originally intend to run it on the NL2Bash dataset to ensure that the results were  consistent with [the paper](https://arxiv.org/abs/1802.08979), however I found that the necessary preprocessing steps used in the paper were fairly complicated and not easily reproduced.


epwalsh:
Also, happy holidays! Please don't feel obligated to look at this anytime soon

allennlp/data/fields/copy_map_field.py
```diff
@@ -0,0 +1,51 @@
...
+    def __init__(self,
+                 source_tokens: List[Token],
+                 target_namespace: str) -> None:
+        self._source_tokens = source_tokens
+        self._target_namespace = target_namespace
+        self._mapping_array: List[int] = None
+
+    @overrides
+    def index(self, vocab: Vocabulary):
+        self._mapping_array = [vocab.get_token_index(x.text, self._target_namespace)
```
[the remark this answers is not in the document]
matt-gardner: Ah, I see, in misunderstood what was going on here on my first read.  I was thinking you used this array to determine when copying happened, but you use `source_token_ids` and `target_token_ids` for that.  It looks like this field is mostly necessary for beam search, both for computing the selective attention (when I predict a target token I need to find occurrences of it in the source sequence) and for combining generation and copy probabilities during search (so that sorting by option probability works correctly).

I spent quite a long time digging through this logic trying to figure out if there were a simpler way to pass in the copy information, but I finally convinced myself that you need all of the fields that you have, and that this is a good way to do it.  If you didn't have to worry about beam search / decoding, I think you could make things a lot simpler.  But there's not a whole lot of point to this model if you can't do decoding =).

So, it would have been nice to have some documentation describing the decisions around what fields are present, as I mentioned above.  That would have saved me a whole lot of digging through the logic.

allennlp/data/fields/copy_map_field.py
```diff
@@ -0,0 +1,51 @@
...
+from overrides import overrides
+import torch
+
+from allennlp.common.util import pad_sequence_to_length
+from allennlp.data.vocabulary import Vocabulary
+from allennlp.data.tokenizers.token import Token
+from allennlp.data.fields.field import Field
+
+
+class CopyMapField(Field[torch.Tensor]):
```
matt-gardner: I wonder if there's a better name for this...  Actually, all this is doing is taking a list of tokens and indexing them with a particular namespace.  You don't need a new `Field` class for this.  You could just use a `SequenceLabelField` and pass in the target namespace, right?

Oh, I guess the difference is that you don't want this to count vocab items and expand the target vocabulary.  I'm debating whether it's worth having a separate field for that, or whether we should add a flag to `SequenceLabelField` to stop vocab counting.  I'm leaning towards a separate field, though I still think we could name this better.  Any thoughts?

allennlp/models/encoder_decoders/copynet_seq2seq.py
```diff
@@ -0,0 +1,824 @@
...
+        when producing the "generation" scores for the target vocab.
+    beam_size : ``int``, required
+        Beam width to use for beam search prediction.
+    max_decoding_steps : ``int``, required
+        Maximum sequence length of target predictions.
+    target_embedding_dim : ``int``, optional (default = 30)
+        The size of the embeddings for the target vocabulary.
+    copy_token : ``str``, optional (default = '@COPY@')
+        The token used to indicate that a target token was copied from the source.
+        If this token is not already in your target vocabulary, it will be added.
```
matt-gardner: Looks like you crash instead of adding it.  Which behavior is correct?  Seems nicer to just add it instead of crashing.  Did that not work?

allennlp/models/encoder_decoders/copynet_seq2seq.py
```diff
@@ -0,0 +1,824 @@
...
+            passed through a `TextFieldEmbedder` and then through an encoder.
+        source_token_ids : ``torch.Tensor``, required
+            Tensor containing IDs that indicate which source tokens match each other.
+            Has shape: `(batch_size, trimmed_source_length)`.
+        source_to_target : ``torch.Tensor``, required
+            Tensor containing vocab index of each source token with respect to the
+            target vocab namespace. Shape: `(batch_size, trimmed_source_length)`.
+        metadata : ``List[Dict[str, Any]]``, required
+            Metadata field that contains the original source tokens ('source_tokens')
+            and any other meta fields.
```
matt-gardner: You also require `target_tokens` in here when using token-based metrics, right?

allennlp/models/encoder_decoders/copynet_seq2seq.py
```diff
@@ -0,0 +1,824 @@
...
+        target_tokens : ``Dict[str, torch.LongTensor]``, optional (default = None)
+            Output of `Textfield.as_array()` applied on target `TextField`. We assume that the
```
matt-gardner: Specify here that this _must_ contain a `tokens` key that uses single ids.

allennlp/data/fields/copy_map_field.py
matt-gardner: I guess you could also consider pulling more of the functionality into a single `Field` with multiple returned tensors.  For instance, you could have a `CopyNetField` or something that takes source tokens, target tokens, and target vocabulary, and computes and returns what is currently the `source_to_target` map, the `target_tokens` text field (which doesn't actually work like a text field), the `source_token_ids` and the `target_token_ids`.  I'm not sure whether that's better than what you have or not, though...  Probably not.  The reason I brought it up is because these fields are all related, and because `TextField` doesn't really match what you want for `target_tokens`.  You might consider just using `LabelField` for that instead.

allennlp/data/dataset_readers/copynet_seq2seq.py
```diff
@@ -0,0 +1,187 @@
...
+        to ``WordTokenizer()``.
+    target_tokenizer : ``Tokenizer``, optional
+        Tokenizer to use to split the output sequences (during training) into words or other kinds
+        of tokens. Defaults to ``source_tokenizer``.
+    source_token_indexers : ``Dict[str, TokenIndexer]``, optional
+        Indexers used to define input (source side) token representations. Defaults to
+        ``{"tokens": SingleIdTokenIndexer()}``.
+    target_token_indexers : ``Dict[str, TokenIndexer]``, optional
+        Indexers used to define output (target side) token representations. Defaults to
+        ``source_token_indexers``.
```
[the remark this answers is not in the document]
matt-gardner: Actually, you don't have a `TextFieldEmbedder` for the targets at all.  You should just remove this argument entirely and use a `ListField[LabelField]` instead of a `TextField` for the targets.

allennlp/models/encoder_decoders/copynet_seq2seq.py
```diff
@@ -0,0 +1,824 @@
...
+                 attention: Attention,
+                 beam_size: int,
+                 max_decoding_steps: int,
+                 target_embedding_dim: int = 30,
+                 copy_token: str = "@COPY@",
+                 source_namespace: str = "source_tokens",
+                 target_namespace: str = "target_tokens",
+                 tensor_based_metric: Metric = None,
+                 token_based_metric: Metric = None) -> None:
+        super(CopyNetSeq2Seq, self).__init__(vocab)
```
matt-gardner: Just `super().__init__(vocab)`.

epwalsh:
Thanks @matt-gardner, these are great comments!

allennlp/data/dataset_readers/copynet_seq2seq.py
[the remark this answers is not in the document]
epwalsh: Hmm ok so I just added a commit that ensures `self._target_token_indexers` is always just `{"tokens": SingleidTokenIndexer}`. I do get what you're saying with using `ListField[LabelField]` instead, but I have been thinking about making the target embedding more flexible in the future, probably through using `TextFieldEmbedder`, so I think it might be worth it to keep it as a `TextField` for now. What do you think?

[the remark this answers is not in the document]
matt-gardner: Yeah, fair enough, if you anticipate making the target embedding more complex, then it's reasonable to leave it as a `TextField`.  Up to you.

allennlp/data/fields/copy_map_field.py
matt-gardner: That's definitely better.  `NamespaceSwappingField`?  I don't like how that sounds, but it conveys what's going on.

allennlp/models/encoder_decoders/copynet_seq2seq.py
epwalsh: Correct. I'll add a comment for that

allennlp/models/encoder_decoders/copynet_seq2seq.py
```diff
@@ -0,0 +1,824 @@
...
+        self.decoder_input_dim = self.decoder_output_dim
+
+        target_vocab_size = self.vocab.get_vocab_size(self._target_namespace)
+
+        # The decoder input will be a function of the embedding of the previous predicted token,
+        # an attended encoder hidden state called the "attentive read", and another
+        # weighted sum of the encoder hidden state called the "selective read".
+        # While the weights for the attentive read are calculated by an `Attention` module,
+        # the weights for the selective read are simply the predicted probabilities
+        # corresponding to each token in the source sentence from the previous timestep.
```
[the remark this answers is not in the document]
epwalsh: Good catch!

allennlp/models/encoder_decoders/copynet_seq2seq.py
epwalsh: If I recall correctly, the first way I had this implemented I was actually adding the copy token into the vocab's source namespace when the model was instantiated. But doing so had some unintended consequences. I can't quite remember the issue it caused, but I think it was because the serialized vocabulary (in the model serialization directory) would not include the copy token in the source namespace.

matt-gardner: Ah, yes, that would have been a problem.  It has since been fixed (about 3 months ago; see [here](https://github.com/allenai/allennlp/pull/1857/files#diff-d4e8f335bf383d8d2b18287949003bbeR276)), so it _should_ be ok to just add it now.

epwalsh: Ah thanks! Yeup, that should do it.

PR #2242, opened 2018-12-26
change mislabeled variable in description

jbkjr:
the required parameter "pretrained_model" was mislabeled as "pretrained_model_name" in the documentation


nelson-liu:
Thanks!

PR #2243, opened 2018-12-26
fix target_namespace bugs

jbkjr:
1. the documentation said the default for target_namespace is "target_tokens" but it is really "tokens"
2. the code twice uses the default literal "tokens" as a key for the target_tokens dictionary, but it should be using self._target_namespace. The way it is written now will cause key errors in the forward loop if anything besides "tokens" is used as the target namespace.


epwalsh:
Hi @jbkjr, the target_namespace refers to the namespace in the Vocabulary object. However, when you see something like `target_tokens[“tokens”]`, that is actually just accessing the token tensor in the dictionary `target_tokens`, which is the result of the text field (that the dataset reader creates) being turned into a tensor dictionary.

Issue #2250, opened 2018-12-27
Circular import with allennlp.modules


nelson-liu:
**System (please complete the following information):**
 - OS: OSX
 - Python version: 3.6.6
 - AllenNLP version: master
 - PyTorch version: 0.4.1

**Question**
Maybe I'm missing something, but importing from modules seems to break on my machine, but only from the REPL. Seems like tests work fine?

```
$ ipython
Python 3.6.6 |Anaconda, Inc.| (default, Jun 28 2018, 11:07:29)
Type 'copyright', 'credits' or 'license' for more information
IPython 6.5.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import torch
   ...: from torch.autograd import Variable
   ...: from allennlp.modules import Pruner
   ...:
   ...:
Deprecated cache directory found (/Users/nfliu/.allennlp/datasets).  Please remove this directory from your system to free up space.
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-340a85315d8a> in <module>()
      1 import torch
      2 from torch.autograd import Variable
----> 3 from allennlp.modules import Pruner

~/Documents/Github/allennlp/allennlp/modules/__init__.py in <module>()
     17 from allennlp.modules.similarity_functions import SimilarityFunction
     18 from allennlp.modules.pruner import Pruner
---> 19 from allennlp.modules.text_field_embedders import TextFieldEmbedder
     20 from allennlp.modules.time_distributed import TimeDistributed
     21 from allennlp.modules.token_embedders import TokenEmbedder, Embedding

~/Documents/Github/allennlp/allennlp/modules/text_field_embedders/__init__.py in <module>()
      7
      8 from allennlp.modules.text_field_embedders.text_field_embedder import TextFieldEmbedder
----> 9 from allennlp.modules.text_field_embedders.basic_text_field_embedder import BasicTextFieldEmbedder

~/Documents/Github/allennlp/allennlp/modules/text_field_embedders/basic_text_field_embedder.py in <module>()
     10 from allennlp.modules.text_field_embedders.text_field_embedder import TextFieldEmbedder
     11 from allennlp.modules.time_distributed import TimeDistributed
---> 12 from allennlp.modules.token_embedders.token_embedder import TokenEmbedder
     13
     14

~/Documents/Github/allennlp/allennlp/modules/token_embedders/__init__.py in <module>()
[...]

PR #2257, opened 2018-12-28
Fix circular dependency.

brendan-ai2:
- Fix for #2250


brendan-ai2:
Thanks for the review + merge! (Followup comment in the issue.)

Issue #2240, opened 2018-12-26
Check for existing vocabulary before creating a new one.


nikhilweee:
**Is your feature request related to a problem? Please describe.**
Every time I recover from an incomplete training session using the `--recover` flag, allennlp creates a fresh vocabulary. This is redundant and time consuming.

**Describe the solution you'd like**
Since the `train` command already saves the vocabulary to disk ([`allennlp/commands/train.py`](https://github.com/allenai/allennlp/blob/1938a5af6d1ea3a2d8dccbb831f82990800fa3e1/allennlp/commands/train.py#L306)), I think it would be a great addition to first check for an existing vocabulary before creating a new one.

**Describe alternatives you've considered**
Nothing really. This seems to be a pretty straightforward solution.

**Additional context**
https://github.com/allenai/allennlp/blob/1938a5af6d1ea3a2d8dccbb831f82990800fa3e1/allennlp/commands/train.py#L296-L301
My suggestion is to rephrase the section above in the following manner. 
```python
    if os.path.exists(os.path.join(serialization_dir, "vocabulary")):
        vocab = Vocabulary.from_files(os.path.join(serialization_dir, "vocabulary"))
    else:
        vocab = Vocabulary.from_params(
                params.pop("vocabulary", {}),
                (instance for key, dataset in all_datasets.items()
                for instance in dataset
                if key in datasets_for_vocab_creation)
        )
```

matt-gardner:
This seems like a reasonable thing to do if we're given the `--recover` flag.  PRs welcome.  Though, it sounds like you might be developing your model on full datasets instead of using test fixtures.  I have never had reason to use `--recover`, because I test my models on test fixtures, so they generally work when giving them the full data.  Using very small test fixtures and developing your model using `pytest` to check that things are working is a _much_ more efficient mode of development.

nikhilweee:
Thanks for your suggestion Matt! As you rightly guessed, perhaps I haven't been developing the way it's supposed to be done. I usually make some quick changes and re-run the training script again and again on the entire dataset until all possible errors have been addressed. That explaines the need to user the `--recover` flag and my frustration with wait times. I wish I was aware of the test fixtures before.

PR #2261, opened 2018-12-31
Check for existing vocabulary before creating a new one (#2240)

nikhilweee:
This PR addresses #2240. Suggestions welcome.

### DIFF
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -300,7 +300,7 @@ def train_model(params: Params,
     logger.info("From dataset instances, %s will be considered for vocabulary creation.",
                 ", ".join(datasets_for_vocab_creation))
 
-    if os.path.exists(os.path.join(serialization_dir, "vocabulary")):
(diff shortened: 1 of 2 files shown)


allennlp/commands/train.py
```diff
@@ -299,12 +299,16 @@ def train_model(params: Params,
...
     logger.info("From dataset instances, %s will be considered for vocabulary creation.",
                 ", ".join(datasets_for_vocab_creation))
-    vocab = Vocabulary.from_params(
-            params.pop("vocabulary", {}),
-            (instance for key, dataset in all_datasets.items()
-             for instance in dataset
-             if key in datasets_for_vocab_creation)
-    )
+
+    if os.path.exists(os.path.join(serialization_dir, "vocabulary")):
```
matt-gardner: This should be `if recover and os.path...`.

nikhilweee: Sure. Doing right away.

Issue #2255, opened 2018-12-27
Contributing a unidirectional language model / working with non-shuffled data


nelson-liu:
I'm interested in implementing a unidirectional LM. Is it preferable to:

1. Create a new `UnidirectionalLanguageModel` class to match the current `BidirectionalLanguageModel`?
2. Think about refactoring `BidirectionalLanguageModel` into a single `LanguageModel` class, and adding a `bidirectional` constructor parameter.

@brendan-ai2 kindly went over the `BidirectionalLanguageModel` code with me, and it seems like the changes are fairly minimal to make a `UnidirectionalLanguageModel`. This makes me partial to option 2, but i want to make sure to get others' opinions to ensure i'm not prematurely optimizing.

Furthermore, the `BidirectionalLanguageModel` is set up to handle sentence-shuffled corpora at the moment. I think this is fairly nonstandard, and the majority of LM work outside of ELMo / other papers on contextual word representation trains on corpora of contiguous text (e.g., PTB or WikiText-2).

I'll need to implement a LM that works with corpora of contiguous text in the near future---seems like this might have to be a new model? Thoughts?

matt-gardner:
Option 2 sounds reasonable to me, though others probably have stronger opinions than me about this.

brendan-ai2:
Exploring option 2 first sounds reasonable, but I would timebox it. My suspicion is that you could make that work with a few fairly surgical modifications. But if it starts to get messy and time consuming, then I would suggest just duplicating the code and modifying it as needed.

Let me know if you hit any snags!

nelson-liu:
@brendan-ai2 , i've implemented option 2 in #2264 . Just need to properly deprecate some classes (would like #2253 in beforehand), and it should be ready for review.

PR #2264, opened 2019-01-02
Turn BidirectionalLM into a more-general LanguageModel class

nelson-liu:
Fixes https://github.com/allenai/allennlp/issues/2255

This PR replaces the `BidirectionalLM` class with a more-general `LanguageModel` that can be used in either the unidirectional/forward setting or the bidirectional setting.

It also accordingly replaces the `BidirectionalLanguageModelTokenEmbedder` with a `LanguageModelTokenEmbedder`.

Also fixes bug in the experiment_unsampled.jsonnet config that was preventing a test from actually being unsampled.

TODO:

- [x] test the unidirectional case
- [x] properly deprecate `BidirectionalLM` and `BidirectionalLanguageModelTokenEmbedder` 
- [x] check docs for accuracy
- [x] fix user-facing training configs


nelson-liu:
I don't think i'll have this finished for the near future's minor release. Thus, it'd be great to merge #2253 before this and the next release.

brendan-ai2:
I like this idea. My main concern would be minimizing disruption, e.g. keeping trained models working. Please ping me when you're ready for a review!

nelson-liu:
ok @brendan-ai2 this should be ready to look at! No rush, though. I'm also planning on taking this code and actually running it on some data to ensure that the perplexities look more-or-less correct.

one question (for anyone): if we deprecate a class, should we keep its tests around? or is that unnecessary?

matt-gardner:
@nelson-liu, on your deprecation question.  @schmmd and I talked about this a bit a while ago, and we think it'd be good to eventually make allennlp-internal warnings into test failures.  So, any test that emits a deprecation warning either has to catch/suppress that warning or fail.  If you're moving the functionality of a class to a new class, you should also move any applicable tests to the new class.  Whether you keep around a test on the deprecated class, that probably only checks that it emits a deprecation warning, is up to you.

matt-peters:
Probably should have looked at this earlier but was out over the holidays so apologies for the late feedback.  What is the reasoning for explicitly adding `ShuffledSentence` to the names?  There is nothing inside the BidirectionalLM class that restricts it to shuffled sentences vs longer contexts, and I have used the calypso version to train on both long contexts and the shuffled Billion Word Benchmark.  

The assumptions about shuffling require changes in the data preparation / iterator and statefulness vs non-statefulness of the contextual encoder.  So it seems to me the abstractions should be `LanguageModel` (with bidirectional option), and adding a `StatefullIterator` that will read very long contexts, chunk them into smaller pieces and do the alignment from batch to batch?

nelson-liu:
That's a good point @matt-peters , will edit.

nelson-liu:
i think my thought process at the time was that rewrapping the hidden states with new tensors to detach them from their history would have to be handled in the `LanguageModel` class, but that's a moot point with the stateful contextualizers in allennlp (was thinking of the pytorch word level lm example and the batching done there...) Thanks for catching that!

brendan-ai2:
I would suggest that we keep `BidirectionalLanguageModel` as a shorthand for setting `bidirectional=True` indefinitely. Creating churn for users by breaking APIs (which is effectively what changing a public name is) is rarely appreciated.

allennlp/models/language_model.py
```diff
@@ -0,0 +1,343 @@
...
+from allennlp.modules.text_field_embedders import TextFieldEmbedder
+from allennlp.modules.sampled_softmax_loss import SampledSoftmaxLoss
+from allennlp.modules.seq2seq_encoders import Seq2SeqEncoder
+from allennlp.nn.util import get_text_field_mask
+from allennlp.nn import InitializerApplicator
+
+
+class _SoftmaxLoss(torch.nn.Module):
+    """
+    Given some embeddings and some targets, applies a linear layer
```
matt-peters: Can you explicitly document here in the constructor that this class does not add a padding ID, and that the input targets to forward should not include the padding id?

allennlp/models/bidirectional_lm.py
```diff
@@ -94,211 +58,12 @@ def __init__(self,
                  num_samples: int = None,
                  sparse_embeddings: bool = False,
                  initializer: InitializerApplicator = None) -> None:
```
matt-peters: It is possible to not duplicate the input arguments from `LanguageModel` and just use `*args, **kwargs` so changes in `LanguageModel` are automatically reflected here?  Or does that break `from_params`?

allennlp/models/language_model.py
```diff
@@ -0,0 +1,343 @@
...
+    vocab: ``Vocabulary``
+    text_field_embedder: ``TextFieldEmbedder``
+        Used to embed the indexed tokens we get in ``forward``.
+    contextualizer: ``Seq2SeqEncoder``
+        Used to "contextualize" the embeddings. As described above,
+        this encoder must not cheat by peeking ahead.
+    dropout: ``float``, optional (default: None)
+        If specified, dropout is applied to the contextualized embeddings before computation of
+        the softmax. The contextualized embeddings themselves are returned without dropout.
+    loss_scale: ``Union[float, str]``, optional (default: 1.0)
```
matt-peters: We can remove this argument, I don't think it's useful.

allennlp/models/language_model.py
nelson-liu: mind if i do this in another pr? would rather that this changeset not be larger than it has to for the core changes for configurable bidirectionality.

allennlp/models/language_model.py
nelson-liu: same as above

allennlp/models/bidirectional_lm.py
nelson-liu: that would be great if it worked, i'll try it out. thanks for the suggestion

allennlp/models/language_model.py
nelson-liu: looks like it doesn't work :(

matt-peters: Oh well.

matt-peters:
This look good to me, although will let @brendan-ai2 take a look and approve as he ported over most of the code.

allennlp/models/language_model.py
```diff
@@ -0,0 +1,343 @@
...
+                        'forward_loss': forward_loss * scale_factor / num_targets.float(),
+                        'backward_loss': (backward_loss * scale_factor / num_targets.float()
+                                          if backward_loss is not None else None)
+                })
+            else:
+                # average_loss zero tensor, return it for all
+                return_dict.update({
+                        'loss': average_loss,
+                        'forward_loss': average_loss,
+                        'backward_loss': average_loss
```
brendan-ai2: `None` if not bidirectional for consistency?

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -1,26 +1,21 @@
...
-from allennlp.nn.util import remove_sentence_boundaries, get_text_field_mask, add_sentence_boundary_token_ids
-
-# Importing at runtime results in a circular dependency.
-if TYPE_CHECKING:
-    from allennlp.models.bidirectional_lm import BidirectionalLanguageModel
 
 
 @TokenEmbedder.register('bidirectional_lm_token_embedder')
 class BidirectionalLanguageModelTokenEmbedder(TokenEmbedder):
+    # pylint: disable=abstract-method
```
brendan-ai2: This shouldn't be needed. Looks like the problem is that we aren't extending the `LanguageModelTokenEmbedder` in the line above. This is good example of why keeping the tests is around is prudent. Fortunately you should be able to do this with minimal duplication. See https://github.com/allenai/allennlp/blob/511c846c9c779fdac0066dbae6765c1619037098/allennlp/tests/fixtures/bidirectional_lm/experiment.jsonnet for how to share config across jsonnet files if you aren't already familiar.

allennlp/modules/token_embedders/bidirectional_language_model_token_embedder.py
```diff
@@ -1,26 +1,21 @@
...
     """
     Compute a single layer of representations from a bidirectional language model. This is done
     by computing a learned scalar average of the layers from the LM. Typically the LM's weights
     will be fixed, but they can be fine tuned by setting ``requires_grad``.
 
+    .. deprecated:: 0.8
+        ``BidirectionalLanguageModelTokenEmbedder`` was deprecated in version 0.8
+        and will be removed in version 0.10 .
```
brendan-ai2: As above.

allennlp/tests/fixtures/bidirectional_lm/experiment_unsampled.jsonnet
```diff
@@ -22,7 +22,6 @@
   },
   "model": {
     "type": "bidirectional_language_model",
-    "num_samples": 10,
```
brendan-ai2: Why did this change? I always get paranoid when tests change in a (mostly) refactoring PR... ;)

allennlp/tests/fixtures/language_model/experiment_unidirectional_unsampled.jsonnet
```diff
@@ -0,0 +1,68 @@
+{
```
brendan-ai2: It would be great to reduce the duplication of this config with that in allennlp/tests/fixtures/bidirectional_lm/experiment_unsampled.jsonnet and allennlp/tests/fixtures/language_model/experiment_unsampled.jsonnet.

allennlp/tests/models/language_model_test.py
```diff
@@ -0,0 +1,156 @@
+# pylint: disable=invalid-name,arguments-differ,abstract-method
+import numpy as np
+import pytest
+
+from allennlp.common.testing import ModelTestCase
+from allennlp.common.checks import ConfigurationError
+from allennlp.common.params import Params
+from allennlp.models import Model
+
+class TestUnidirectionalLanguageModel(ModelTestCase):
```
brendan-ai2: Having an extra option for the directionality has made it difficult to see at a glance that the different tests are behaving as expected. I didn't do a great job with this when I added the transformer test, but in retrospect it should be pretty simple.

Roughly what needs to happen is a) set `expected_embedding_shape` in the constructor and b) check against this later. Similarly for `expected_backward_loss` and `breaking_contextualizer_direction`. Actually, I think the inherited test methods will be used too. (Worth verifying that works as expected.) So a lot of code can be removed.

As an aside, there is a balance to be struck between straightforward tests with modest duplication and parameterized ones. Here I'd lean towards parameterizing.

allennlp/tests/fixtures/bidirectional_lm/experiment_unsampled.jsonnet
nelson-liu: If the experiment is unsampled, isn't the right thing to do to specify a value of `None` to the `"num_samples"`? i.e., i'm pretty sure the `_SoftmaxLoss` wasn't being tested before?

`"num_samples": 10` is being added in `experiment.json` ( https://github.com/allenai/allennlp/blob/master/allennlp/tests/fixtures/bidirectional_lm/experiment.jsonnet#L5 ). Or did I overlook something?

allennlp/models/bidirectional_lm.py
```diff
@@ -1,57 +1,20 @@
...
-                dim=-1
-        )
-
-        return torch.nn.functional.nll_loss(probs, targets.long(), reduction="sum")
-
-
 @Model.register('bidirectional-language-model',
                 deprecation_message=('The "bidirectional-language-model" name was '
                                      'deprecated in version 0.8 and will be removed'
                                      'in version 0.10 . '
```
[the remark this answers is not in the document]
nelson-liu: ok, so you don't want to deprecate `bidirectional-language-model`?

allennlp/models/language_model.py
```diff
@@ -0,0 +1,343 @@
...
+        # shape (batch_size, timesteps) masked with 0
+        if self._bidirectional:
+            forward_embeddings, backward_embeddings = lm_embeddings.chunk(2, -1)
+        else:
+            forward_embeddings = lm_embeddings
+            backward_embeddings = None
+        losses: List[torch.Tensor] = []
+        for idx, embedding, targets in ((0, forward_embeddings, forward_targets),
+                                        (1, backward_embeddings, backward_targets)):
+            if embedding is None and targets is None:
```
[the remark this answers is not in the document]
nelson-liu: maybe i'm missing the point, but i don't think i could break the entire `for` statement out into a new function? or are you saying, break out the portion underneath the `if`? end result would still be an `if` in the `for`, though?

allennlp/models/bidirectional_lm.py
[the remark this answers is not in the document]
brendan-ai2: I don't hold an opinion on whether to apply the deprecation warning or not*. Encouraging standardizing on underscores is reasonable. However, I don't want to actually remove the name as it breaks the API. I think API breakage should be used only when strictly necessary.

* Depends on what we want to communicate as a team with deprecation warnings. I think it can be useful to communicate that "you shouldn't use this there is something better" without also saying "we're definitely going to remove this soon". This varies across projects.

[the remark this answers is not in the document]
nelson-liu: thanks for clarifying!

nelson-liu:
@brendan-ai2 this is ready for another look, fyi

allennlp/tests/fixtures/bidirectional_lm/experiment_unsampled.jsonnet
brendan-ai2: Good catch! I added the following to the description:

"Also fixes bug in the experiment_unsampled.jsonnet config that was preventing a test from actually being unsampled."

Even though it's a small fix code-wise, noting the change explicitly could end up being a useful breadcrumb later.

allennlp/tests/models/bidirectional_lm_test.py
```diff
@@ -25,42 +27,17 @@ def test_forward_pass_runs_correctly(self):
...
-        result = self.model(**training_tensors)
-
-        assert set(result) == {"loss", "forward_loss", "backward_loss",
-                               "lm_embeddings", "noncontextual_token_embeddings", "mask"}
-
-        # The model should preserve the BOS / EOS tokens.
-        embeddings = result["lm_embeddings"]
-        # The BidirectionalLanguageModelTransformer uses input size * 2 as the output size unlike
-        # a bidirectional LSTM, which uses hidden size * 2.
-        assert tuple(embeddings.shape) == (2, 8, 32)
```
brendan-ai2: Something seems off here. Shouldn't the expected dimension be set to `(2, 8, 32)`? Sorry if I'm missing something.

brendan-ai2: It's concerning that the tests didn't fail...

nelson-liu: So this experiment uses `experiment_bidirectional_unsampled.jsonnet`, which takes its config from `experiment_unsampled.jsonnet`. In that, we've got:

```
"contextualizer": {
        "type": "lstm",
        "bidirectional": true,
        "num_layers": 3,
        "input_size": 16,
        "hidden_size": 7,
    }
  },
```

So hidden size * 2 = 14?

nelson-liu: the `TestBidirectionalLMTransformer` was moved to a different class, which is why it isn't there.

brendan-ai2: D'oh. Sorry! Read that diff all wrong.

nelson-liu: sorry it's so convoluted with stuff moving around everywhere, but also not completely disappearing...

training_config/bidirectional_language_model.jsonnet
```diff
@@ -58,10 +58,11 @@ local BASE_ITERATOR = {
       //     "tokens": ["<S>", "</S>"],
       //     "token_characters": ["<>/S"]
       // },
```
brendan-ai2: Please update the path to this file in https://github.com/allenai/allennlp/blob/master/tutorials/how_to/training_transformer_elmo.md. You should also grep the codebase for any other references to this file. And recall that this won't help external users who may have references to this file already. Please consider this carefully when weighing whether a rename should be performed.

nelson-liu: done, thanks for pointing that out

nelson-liu:
Thanks for your thorough comments, @brendan-ai2 ! Much appreciated.

PR #2266, opened 2019-01-02
Remove old semantic parsing dataset readers from check_docs.

### LINKED ISSUE #2251: Remove docs check exception for old semantic parsing dataset readers
Was fixing #2248 and saw https://github.com/allenai/allennlp/blob/3f0953d19de3676ea82e642659fc96d90690e34d/scripts/check_docs.py#L32-L35 . Not too sure what needs to be done here, since i've never touched the semantic parsing code, but raising an issue so we don't forget about this now that version 0.8 has come around. maybe just removing these lines, since it looks like #1739 already removed the code files themselves?

cc / assigning @DeNeutoy and @schmmd

schmmd:
Fixes #2251


schmmd:
Closing as someone removed this in a separate PR.

schmmd:
In case someone views the diff and is confused--my last merge commit was a mistake.  Look at the original commit for the intended diff.

Issue #2236, opened 2018-12-24
Build a BertTextFieldEmbedder


valsworthen:
Hello,

I am trying to use BERT for QA. In the paper, the authors explain that in this case they format the input data to be `Question tokens [SEP] Passage tokens`. Then a "mask" vector is also passed to distinguish between the question and the passage.

If I want to do the same using AllenNLP. I figured out from the code, that, when the `BertTokenEmbedder` is called [it can use the `token_type_ids`](https://github.com/allenai/allennlp/blob/1938a5af6d1ea3a2d8dccbb831f82990800fa3e1/allennlp/modules/token_embedders/bert_token_embedder.py#L89) which is this "mask". However, if in the config file we only specify 
```
"text_field_embedder": {
        "token_embedders": {
            "tokens": {
                "type": "bert-pretrained",
                "pretrained_model": "bert-base-uncased",
                "requires_grad": false,
                "top_layer_only": false
            }
        },
        "allow_unmatched_keys": true
}
```
[so the only key used by the TextFieldEmbedder is `'tokens'`](https://github.com/allenai/allennlp/blob/1938a5af6d1ea3a2d8dccbb831f82990800fa3e1/allennlp/modules/text_field_embedders/basic_text_field_embedder.py#L101). Then if I understand correctly, we should also pass a `embedder_to_indexer_map` to the TextFieldEmbedder.

To my point of view there remains the problem that [the `WordpieceIndexer` doesn't return the `token_type_ids`](https://github.com/allenai/allennlp/blob/1938a5af6d1ea3a2d8dccbb831f82990800fa3e1/allennlp/data/token_indexers/wordpiece_indexer.py#L165) so we should build it ourself.

In all cases I think this would imply that the DatasetReader returns Instances in the form `Question tokens [SEP] Passage tokens`.

Therefore I think that I need to:

- modify my Reader so that it returns Instances in the concatenated format (this way they would be indexed correctly). In this case how can correctly add the Token for [SEP] which is otherwise be tokenized as `[ SEP ]`?
- in its method `text_to_instance`, build the key `token_type_ids` and add it to the instance. edit: actually this should be done in a new TokenIndexer or in the Model
- add the following `'tokens': 'token_type_ids'` to the config file under the key `"text_field_embedder"` and pass a map.

Can you tell me if my understanding of the different parts is correct and if my corrections would work? 

Many thanks for your help!

WrRan:
[...]

PR #2279, opened 2019-01-04
[Feature Enhanced] Support sentence pair in BERT

WrRan:
support #2236


allennlp/data/token_indexers/wordpiece_indexer.py
```diff
@@ -245,4 +256,25 @@ def __init__(self,
...
                          never_lowercase=never_lowercase,
                          start_tokens=["[CLS]"],
-                         end_tokens=["[SEP]"])
+                         end_tokens=["[SEP]"],
+                         separator_token="[SEP]")
+
+
+def _get_token_type_ids(wordpiece_ids: List[int],
+                        separator_ids: List[int]) -> List[int]:
+    length_wordpiece = len(wordpiece_ids)
```
joelgrus: I would call this `num_wordpieces` or something like that

allennlp/data/token_indexers/wordpiece_indexer.py
```diff
@@ -245,4 +256,25 @@ def __init__(self,
...
+    length_end_pieces = len(separator_ids)
```
joelgrus: similarly, I would call this `separator_length`, which I think is more indicative of what it actually is

allennlp/data/token_indexers/wordpiece_indexer.py
```diff
@@ -245,4 +256,25 @@ def __init__(self,
...
+    while cursor < len(wordpiece_ids):
+        if separator_ids[0] != wordpiece_ids[cursor]:
+            cursor += 1
+        else:
+            if separator_ids == wordpiece_ids[cursor:cursor+length_end_pieces]:
+                type_id += 1
+                cursor += length_end_pieces
+                token_type_ids[cursor:] = [type_id] * (length_wordpiece - cursor)
+            else:
+                cursor += length_end_pieces
```
joelgrus: in practice, the separator is going to have length 1. I would probably try to avoid taking slices here, as it's inefficient. I think something like this would be more readable (and probably faster):

```
token_type_ids = []
cursor = 0
type_id = 0

if all(wordpiece_ids[cursor + i] == separator_id for i, separator_id in enumerate(separator_ids)):
    type_id += 1
    cursor += len(separator_ids)
    token_type_ids.extend(type_id for _ in separator_ids)
else:
    cursor += 1
    token_type_ids.append(type_id)
```

I think that does the same thing and is slightly more explicit and doesn't create unnecessary lists along the way.

allennlp/modules/text_field_embedders/basic_text_field_embedder.py
```diff
@@ -93,8 +94,18 @@ def forward(self, text_field_input: Dict[str, torch.Tensor], num_wrapping_dims:
...
+                    # If `indexer_key` is None, we map it to `None`.
+                    tensors = [(text_field_input[indexer_key] if indexer_key is not None else None)
+                               for indexer_key in indexer_map]
+                elif isinstance(indexer_map, collections.abc.Mapping):
+                    tensors = {
+                            name: text_field_input[argument]
+                            for name, argument in indexer_map.items()
+                    }
+                else:
+                    raise NotImplementedError
```
joelgrus: several things on this part:

1. you should change the signature of the method to indicate that embedder_to_indexer_map can have either a list or a dict
2. I would probably just use `dict` and `list` rather than the abstract types, that's what we do everywhere else in the code 
3. I would probably move things around so that the `token_vectors = ...` is up here, that way you only have to do one `isinstance` check

WrRan:
Thanks for your review. 

I also add some comments to the method, but I think it may be not clear. May I have your suggestions?
https://github.com/allenai/allennlp/blob/6176f6bf664e9dc0d2531fa456fd22662c00fe4e/allennlp/modules/text_field_embedders/basic_text_field_embedder.py#L32-L43

tingkai-zhang:
Hi, @WrRan  Thanks for your work!
Are you currently following up this thread? It hasn't been updated for a while.
If you are not available right now, do you want me to pull a fix to your branch?

WrRan:
@tingkai-zhang I am still following this thread up, but the AllenNLP team is a little busy for now due to AAAI2019, so this PR is not updated in time. Also, I have made some changes locally. But feel free to push your fix to my branch, we can cope with the possible conflicts later. Thanks for your attention. :smile:

joelgrus:
I think this looks pretty good, do you have more changes to make?

tingkai-zhang:
@joelgrus Looks good to me!  I was just asking @WrRan if he needs some fix help but it turns out he's already pulled the fix.

WrRan:
@joelgrus Sorry for the delay. :sweat_smile: 
Because the BERT models contain only two `token_type_ids`, the text containing more than two sentences cannot be processed correctly. Thus, I raise a `ConfigurationError` in such a case. Do you have a better solution?

The code can be found at my repo:

https://github.com/WrRan/allennlp/blob/7630352ac6e19a88b9d2305eee8c0f6ca761ae32/allennlp/data/token_indexers/wordpiece_indexer.py#L286-L289

PR #2281, opened 2019-01-05
Grammar induction from a python executor

matt-gardner:
I still need to clean this up a little bit (i.e., add some more documentation), and I moved some things around, so I want to see what breaks in CI, which is why I'm opening a PR for this now.  But, it's almost done.

This makes it so you can just define a collection of python functions for a particular domain, and get lisp-like logical form parsing and execution for free (including all of the search stuff that we do over logical forms).  So, this _should_ entirely replace the `type_declaration` code and make the `executors` _way_ simpler.  I have a simple proof-of-concept `ArithmeticExecutor` in the test for this PR.  The next step (after this is merged) is to rewrite the WikiTables and NLVR executors using this, ripping out all of the ugly code that is no longer necessary.  There will probably be some complexities there I'll have to deal with, but _hopefully_ it won't be too bad.  After that, we try to have an executor that has learned parameters, which would be super cool - really easy neural module networks for whatever domain you want.

If you're interested in what this does, just look at the test.  We define a ~40 line collection of functions in `ArithmeticExecutor`, then the tests show all of the things you get for free.


matt-gardner:
Also, looking at this again, I don't think `Executor` is the right name here.  This combines what was the `Executor` and the `World`, so this _should_ be the only thing you have to write for most semantic parsing use cases (when you're the one defining the language - SQL/java is a different story altogether).  I'm thinking `DomainLanguage` might be a good name, or `DomainTheory`, or maybe even just `World`, because it could include instance-specific state that's taken in the constructor.  If anyone has any thoughts on that, I'm happy to hear them.

matt-gardner:
Ok, I think this is ready for review.  @joelgrus or @brendan-ai2, one of you want to take a look?

allennlp/semparse/domain_languages/domain_language.py
```diff
@@ -0,0 +1,474 @@
...
+        arguments and returns an integer.
+
+        As an example translation from logical form to complete action sequence, the logical form
+        ``(add 2 3)`` would be translated to ``['@start@ -> int', 'int -> [<int,int:int>, int, int]',
+        '<int,int:int> -> add', 'int -> 2', 'int -> 3']``.
+        """
+        expression = util.lisp_to_nested_expression(logical_form)
+        try:
+            transitions, start_type = self._get_transitions(expression, expected_type=None)
+            print(transitions, start_type)
```
joelgrus: do you want this print here?

allennlp/semparse/domain_languages/domain_language.py
```diff
@@ -0,0 +1,474 @@
...
+                raise ExecutionError(f"Error executing expression {expression}: {error}")
+        elif isinstance(expression, str):
+            if expression not in self._functions:
+                raise ExecutionError(f"Unrecognized constant: {expression}")
+            try:
+                return self._functions[expression]()
+            except (TypeError, ValueError) as error:
+                raise ExecutionError(f"Error executing expression {expression}: {error}")
+        else:
+            raise ExecutionError("Not sure how you got here. Please open a github issue with details.")
```
joelgrus: I feel like there's a cleaner way (e.g. fewer cases) to write this function, but I've been staring at it for 5 minutes and can't figure out what it is.

allennlp/semparse/domain_languages/domain_language.py
```diff
@@ -0,0 +1,474 @@
...
+                raise ParsingError(f'{expression} did not have expected type {expected_type}')
+            return [f'{constant_type} -> {expression}'], constant_type
+        else:
+            raise ParsingError('Not sure how you got here. Please open an issue on github with details.')
+
+    def _get_function_transitions(self,
+                                  expression: Union[str, List],
+                                  expected_type: PredicateType) -> Tuple[List[str],
+                                                                         PredicateType,
+                                                                         List[PredicateType]]:
```
joelgrus: this function could probably use a docstring

allennlp/semparse/domain_languages/domain_language.py
```diff
@@ -0,0 +1,474 @@
...
+            # Functions with no arguments are basically constants whose type match their return
+            # type.
+            return return_type
+        return FunctionType(arg_types, return_type)
+
+
+class BasicType(PredicateType):
+    """
+    A ``PredicateType`` representing a zero-argument predicate (which could technically be a
+    function with no arguments or a function; both are treated the same here).
```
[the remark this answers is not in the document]
matt-gardner: Yep, thanks, I meant "constant".

allennlp/semparse/domain_languages/domain_language.py
```diff
@@ -0,0 +1,474 @@
...
+            # we need to do some magic here.
+            origin = type_.__origin__
+            args = type_.__args__
+            self.name: str = f'{origin.__name__}[{",".join(arg.__name__ for arg in args)}]'
+        else:
+            self.name: str = type_.__name__
+
+    def __str__(self):
+        return self.name
+    __repr__ = __str__
```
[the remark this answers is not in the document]
matt-gardner: Done.

allennlp/semparse/domain_languages/domain_language.py
```diff
@@ -0,0 +1,474 @@
...
+        """
+        if not self._valid_actions:
+            actions: Dict[str, List[str]] = defaultdict(list)
+            if self._start_types:
+                for start_type in self._start_types:
+                    actions[START_SYMBOL].append(f'{START_SYMBOL} -> {start_type}')
+            else:
+                # If you didn't give us a set of valid start types, we'll assume all types we know
+                # about (including functional types) are valid start types.
+                for type_ in self._function_types.values():
```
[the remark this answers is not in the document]
matt-gardner: Ah, yes, definitely, good point.  I _think_ we make these into a set at some point, which is why the tests still pass...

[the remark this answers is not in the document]
matt-gardner: Oh, no, I just didn't test this case.

allennlp/semparse/domain_languages/domain_language.py
matt-gardner: Nope, thanks.

allennlp/semparse/domain_languages/domain_language.py
matt-gardner: I think if we just assumed good input, or let the code crash on bad input without switching to an `ExecutionError`, we could make it a lot simpler.  Trying to be nice about handling bad input is what makes it so complicated.

matt-gardner:
Ah, looks like I forgot to respond to your big question.  It's a good question, and in the end it might be better to do it that way.  Switching everything would require re-writing a whole bunch of the data processing code (including `ProductionRuleField`), and would be better in a different PR.  And we use the strings as vocabulary items, and that might be the sticking point for not doing what you suggest.  I'd have to think quite a bit about how it would work to use a `NamedTuple` instead...

matt-gardner:
@joelgrus, you didn't actually approve the PR - is this good to merge now?

PR #2282, opened 2019-01-05
Filter Warnings when pytest

### LINKED ISSUE #1672: Warnings in tests
Looks like the following warnings come up when we run tests:

```
[16:29:25][Step 3/5] allennlp/tests/models/reading_comprehension/dialog_qa_test.py::DialogQATest::test_batch_predictions_are_consistent
[16:29:25][Step 3/5]   /usr/local/lib/python3.6/site-packages/torch/tensor.py:263: UserWarning: non-inplace resize is deprecated
[16:29:25][Step 3/5]     warnings.warn("non-inplace resize is deprecated")
[16:29:25][Step 3/5] 
[16:29:25][Step 3/5] allennlp/tests/models/reading_comprehension/dialog_qa_test.py::DialogQATest::test_forward_pass_runs_correctly
[16:29:25][Step 3/5]   /usr/local/lib/python3.6/site-packages/torch/tensor.py:263: UserWarning: non-inplace resize is deprecated
[16:29:25][Step 3/5]     warnings.warn("non-inplace resize is deprecated")
[16:29:25][Step 3/5] 
[16:29:25][Step 3/5] allennlp/tests/models/reading_comprehension/dialog_qa_test.py::DialogQATest::test_model_can_train_save_and_load
[16:29:25][Step 3/5]   /usr/local/lib/python3.6/site-packages/torch/tensor.py:263: UserWarning: non-inplace resize is deprecated
[16:29:25][Step 3/5]     warnings.warn("non-inplace resize is deprecated")
[16:29:25][Step 3/5] 
[16:29:25][Step 3/5] allennlp/tests/nn/decoding/chu_liu_edmonds_test.py::ChuLiuEdmondsTest::test_mst_finds_maximum_spanning_tree
[16:29:25][Step 3/5]   /local/deploy/agent2/work/8feb324ce7c68d53/allennlp/tests/nn/decoding/chu_liu_edmonds_test.py:78: UserWarning: torch.range is deprecated in favor of torch.arange and will be removed in 0.5. Note that arange generates values in [start; end), not [start; end].
[16:29:25][Step 3/5]     energy = torch.range(1, 9).view(1, 3, 3)
```

WrRan:
fix #1672 

> By default pytest will display some warnings from user code and third-party libraries, as recommended by PEP-0506. This helps users keep their code modern and avoid breakages when deprecated warnings are effectively removed.
>
> from [pytest](https://docs.pytest.org/en/latest/warnings.html#deprecationwarning-and-pendingdeprecationwarning).

There still remain some warnings:
```txt
/usr/local/lib/python3.6/site-packages/nbconvert/exporters/exporter_locator.py:28
[02:56:20][Step 3/10]   /usr/local/lib/python3.6/site-packages/nbconvert/exporters/exporter_locator.py:28: DeprecationWarning: `nbconvert.exporters.exporter_locator` is deprecated in favor of `nbconvert.exporters.base` since nbconvert 5.0.
[02:56:20][Step 3/10]     DeprecationWarning)
[02:56:20][Step 3/10] 
[02:56:20][Step 3/10] /usr/local/lib/python3.6/site-packages/tornado/web.py:1747
[02:56:20][Step 3/10]   /usr/local/lib/python3.6/site-packages/tornado/web.py:1747: DeprecationWarning: @asynchronous is deprecated, use coroutines instead
[02:56:20][Step 3/10]     DeprecationWarning)
[02:56:20][Step 3/10] 
[02:56:20][Step 3/10] allennlp/data/token_indexers/openai_transformer_byte_pair_indexer.py:25
[02:56:20][Step 3/10]   /local/deploy/agent3/work/98197cf33cb401e5/allennlp/data/token_indexers/openai_transformer_byte_pair_indexer.py:25: DeprecationWarning: invalid escape sequence \?
[02:56:20][Step 3/10]     text = re.sub('''(-+|~+|!+|"+|;+|\?+|\++|,+|\)+|\(+|\\+|\/+|\*+|\[+|\]+|}+|{+|\|+|_+)''', r' \1 ', text)
...
[02:56:20][Step 3/10] allennlp/tests/predictors/srl_test.py::TestSrlPredictor::test_uses_named_inputs
[02:56:20][Step 3/10]   /usr/local/bin/pytest:11: DeprecationWarning: [W002] Tokenizer.from_list is now deprecated. Create a new Doc object instead and pass in the strings as the `words` keyword argument, for example:
[02:56:20][Step 3/10]   from spacy.tokens import Doc
[02:56:20][Step 3/10]   doc = Doc(nlp.vocab, words=[...])
[02:56:20][Step 3/10]     sys.exit(main())
...
[02:56:20][Step 3/10] allennlp/tests/semparse/worlds/text2sql_world_test.py::TestText2SqlWorld::test_variable_free_world_cannot_parse_as_statements
[02:56:20][Step 3/10]   <unknown>:1: DeprecationWarning: invalid escape sequence \s
[02:56:20][Step 3/10]   <unknown>:1: DeprecationWarning: invalid escape sequence \s
[02:56:20][Step 3/10]   <unknown>:1: DeprecationWarning: invalid escape sequence \d
```
which is possibly related to
- [using `re` in openai_transformer_byte_pair_indexer.py](https://github.com/allenai/allennlp/blob/master/allennlp/data/token_indexers/openai_transformer_byte_pair_indexer.py#L15) 
- [using `spacy` in semantic_role_labeler.py](https://github.com/allenai/allennlp/blob/master/allennlp/predictors/semantic_role_labeler.py#L60)

I think it is better to fix it (using new methods instead of deprecated ones),
but I have no idea about `DeprecationWarning: nbconvert.exporters.exporter_locator is deprecated in favor of...` and `/usr/local/lib/python3.6/site-packages/tornado/web.py:1747: DeprecationWarning: @asynchronous is deprecated, use coroutines instead`.


pytest.ini
```diff
@@ -4,3 +4,13 @@ log_format = %(asctime)s - %(levelname)s - %(name)s - %(message)s
 log_level = DEBUG
 markers =
     java
+filterwarnings =
+# individual warnings filters are specified as a sequence of fields separated by colons:
+# action:message:category:module:line
+# <reference: https://docs.python.org/3/library/warnings.html#warning-filter>
+    # ignore third-party modules's warns
+    ignore:encoding is deprecated, Use raw=False instead.:PendingDeprecationWarning:.*msgpack_numpy
```
matt-gardner: This one is probably ok.

pytest.ini
```diff
@@ -4,3 +4,13 @@ log_format = %(asctime)s - %(levelname)s - %(name)s - %(message)s
...
+    # ignore warnings for users
+    ignore::DeprecationWarning:allennlp\.common\.registrable
```
matt-gardner: I don't think we should do this.  I think we should instead fix the warnings, so that our own library isn't using deprecated code.

WrRan:
Got it. 
I think we can classify warning when pytesting into four categories:
1. One is caused by a third-party library calling a method that has been or will be deprecated, which we can ignore (but may lead to not upgrade third-party library in time, which is not a big problem). e.g. `PendingDeprecationWarning` caused by `msgpack_numpy`, if I didn't miss anything.
2. One is because we *directly* call the method that has been or will be deprecated, in which case we should fix it. e.g.  using `re` in `openai_transformer_byte_pair_indexer.py`.
3. One is due to the warning we throw to the end user, which should also be safely ignored. e.g. using `ExperimentalFeatureWarning` defined in `allennlp.common.checks`. (By the way, why not we use [`FutureWarning`](https://docs.python.org/3/library/exceptions.html#FutureWarning)? Detailed reasons are introduced below.)
4. One is due to the wrong use of warnings, which should be fixed *in time*. e.g. `DeprecationWarning: The "bidirectional-language-model" name ...` defined in `allennlp.common.registrable`. (The reason why I think it is a bug is introduced immediately below)

About *FutureWarning*:
> excpetion FutureWarning:
>
> Base class for warnings about deprecated features when those warnings are intended for end users of applications that are written in Python.
>
> from [docs.python.org](https://docs.python.org/3/library/exceptions.html#FutureWarning)

Use Case:
> The Datetime64 type remains experimental in this release. In 1.9 there will probably be some changes to make it more useable.
>
> from [numpy](https://github.com/numpy/numpy/blob/464f79eb1d05bf938d16b49da1c39a4e02506fa3/doc/release/1.8.0-notes.rst#future-changes)

I think this `FutureWarning` meets the needs that `ExperimentalFeatureWarning` is designed for. Does that make sense?

About *allennlp.common.registrable*
```python

class DummyTest(AllenNlpTestCase):
    def test_nothing(self):
        # do nothing
        pass
```
Run above test case, and you still get the following warning:
```txt
... DeprecationWarning: The "bidirectional-language-model" name ...
```

So I think that this warning is not raised at the time of using but importing, which is probably *not* we expected. Did I miss anything?

@matt-gardner What do you think?

schmmd:
@WrRan thanks for putting in some effort to clean up our warnings--they have become a bit much!  I think it would be great to remove the warnings you classify under (1) since they are not actionable and obscure the warnings we should be cleaning up in our library.

Out of your four categories which do you propose filtering and which do you propose keeping?  I wasn't completely clear.

matt-gardner:
Yes, the `bidirectional-language-model` warning looks like it is not behaving as expected (cc @nelson-liu).  And for your four categories:

1. Yes, we can silence these.
2. We should fix these.
3. We should make sure that our tests do not throw any of these warnings.  That is, users should not see `DeprecationWarnings` because of things that we do internal to the library.  If a users calls a deprecated method directly, then we can throw a warning.
4. If we're using warnings incorrectly, we should definitely fix them (like in the case above).  `ExperimentalFeatureWarning` is different from deprecation, however - it's saying that the functionality may or may not stick around, whereas deprecation is saying that there is now something better available and the use of the deprecated method is discouraged.

nelson-liu:
the `bidirectional-language-model` warning was fixed in #2297, sorry about that.

WrRan:
I see.

Summarizing the discussion, I think the current better strategy for treating warns is:
```ini
filterwarnings =
    # 1. ignore warns caused by third-party modules
    ignore:encoding is deprecated, Use raw=False instead.:PendingDeprecationWarning:.*msgpack_numpy
    # 2. transform warns to errors when we _directly_ call deprecated methods (including internal methods in AllenNLP)
    # 3. unless we explicitly test these warns
    error::DeprecationWarning
    # 4. ignore these `ExperimentalFeatureWarning`s for now, but record them once
    once:This particular transformer implementation is a provisional feature.*::allennlp\.modules\.seq2seq_encoders\.bidirectional_language_model_transformer
```

For 2 and 3, there is a good example about how to explicitly test warning:
https://github.com/allenai/allennlp/blob/0663762960a4522d4740eea460f63a06eeba126d/allennlp/tests/modules/text_field_embedders/basic_text_field_embedder_test.py#L123-L141

If we repharse that as follow:
```python
    def test_old_from_params_new_from_params(self):
        old_params = Params({
                "words1": {
                        "type": "embedding",
                        "embedding_dim": 2
                        },
                "words2": {
                        "type": "embedding",
                        "embedding_dim": 5
                        },
                "words3": {
                        "type": "embedding",
                        "embedding_dim": 3
                        }
                })

        # Oops
        old_embedder = BasicTextFieldEmbedder.from_params(params=old_params, vocab=self.vocab)
```
this test case will fail with following useful message:
```txt
basic_text_field_embedder_test.py F
...
>       BasicTextFieldEmbedder.from_params(params=old_params, vocab=self.vocab)

basic_text_field_embedder_test.py:142: 
...
>           warnings.warn(DeprecationWarning("the token embedders for BasicTextFieldEmbedder should now "
                                             "be specified as a dict under the 'token_embedders' key, "
                                             "not as top-level key-value pairs"))
E           DeprecationWarning: the token embedders for BasicTextFieldEmbedder should now be specified as a dict under the 'token_embedders' key, not as top-level key-value pairs
```

And this setting can also help us to avoid the wrong use of warnings.


I agree with @matt-gardner 's view about `ExperimentalFeatureWarning` and `deprecation`. But what about `ExperimentalFeatureWarning`  and **`FutureWarning`**?

I simply investigated how `numpy` uses `FutureWarning` and found that `numpy` use it to indicate **the functionality that may change**, such as the case above. Thus, I still think `ExperimentalFeatureWarning` is kind of unnecessary. Does it make sense?

schmmd:
@WrRan the test output looks a lot cleaner.  I'm still seeing one deprecation warning from within a dependency however.

```
[04:28:14][Step 3/10] ______________ ERROR collecting allennlp/tests/notebooks_test.py _______________
[04:28:14][Step 3/10] allennlp/tests/notebooks_test.py:6: in <module>
[04:28:14][Step 3/10]     from nbconvert.preprocessors.execute import CellExecutionError
[04:28:14][Step 3/10] /usr/local/lib/python3.6/site-packages/nbconvert/__init__.py:4: in <module>
[04:28:14][Step 3/10]     from .exporters import *
[04:28:14][Step 3/10] /usr/local/lib/python3.6/site-packages/nbconvert/exporters/__init__.py:3: in <module>
[04:28:14][Step 3/10]     from .exporter_locator import export_by_name
[04:28:14][Step 3/10] /usr/local/lib/python3.6/site-packages/nbconvert/exporters/exporter_locator.py:28: in <module>
[04:28:14][Step 3/10]     DeprecationWarning)
[04:28:14][Step 3/10] E   DeprecationWarning: `nbconvert.exporters.exporter_locator` is deprecated in favor of `nbconvert.exporters.base` since nbconvert 5.0.
```

I pushed a commit that fixes one of the warnings so the build output is less cluttered.

schmmd:
@matt-gardner is the assigned code reviewer so I defer to him, but I think we should silence the `ExperimentalFeatureWarning`s in `pytest`.  With the `once` setting they will always show up and clutter our build output.  These warnings will still show up when our users try out experimental features.

schmmd:
Sorry--instead I'll fix this warning on master.  It's more involved than I first thought.

WrRan:
> ```
> [04:28:14][Step 3/10] ______________ ERROR collecting allennlp/tests/notebooks_test.py _______________
> [04:28:14][Step 3/10] allennlp/tests/notebooks_test.py:6: in <module>
> [04:28:14][Step 3/10]     from nbconvert.preprocessors.execute import CellExecutionError
> [04:28:14][Step 3/10] /usr/local/lib/python3.6/site-packages/nbconvert/__init__.py:4: in <module>
> [04:28:14][Step 3/10]     from .exporters import *
> [04:28:14][Step 3/10] /usr/local/lib/python3.6/site-packages/nbconvert/exporters/__init__.py:3: in <module>
> [04:28:14][Step 3/10]     from .exporter_locator import export_by_name
> [04:28:14][Step 3/10] /usr/local/lib/python3.6/site-packages/nbconvert/exporters/exporter_locator.py:28: in <module>
> [04:28:14][Step 3/10]     DeprecationWarning)
> [04:28:14][Step 3/10] E   DeprecationWarning: `nbconvert.exporters.exporter_locator` is deprecated in favor of `nbconvert.exporters.base` since nbconvert 5.0.
> ```
> I pushed a commit that fixes one of the warnings so the build output is less cluttered.

I am willing to fix the warning showing in this log. :smile:

schmmd:
@WrRan great--but I got sucked in and made https://github.com/allenai/allennlp/pull/2310, which I *think* is the right fix.  Sorry--I just got excited about your change ;-)

WrRan:
😃 Happy to be helpful.

matt-gardner:
Yes, I agree with the changes you made.  Thanks for taking this on @WrRan!  As for `FutureWarning` vs. `ExperimentalFeatureWarning`, I just spend a bunch of time reading trying to figure out what the appropriate semantics of `FutureWarning` is.  In the end, I think pre-3.7, you're right that maybe `FutureWarning` has the same semantics as we intend for `ExperimentalFeatureWarning`.  Python 3.7 changes that, though (see [here](https://docs.python.org/3/library/warnings.html#warning-categories)).  But taking a step back, I had to go read for 20 minutes to figure out what exactly `FutureWarning` was supposed to mean, while `ExperimentalFeatureWarning` is immediately obvious.  For that reason alone, I'd stick with the better name.

WrRan:
Thanks for your attention. You are right, @matt-gardner.

> Changed in version 3.7: Previously `DeprecationWarning` and `FutureWarning` were distinguished based on whether a feature was being removed entirely or changing its behaviour. They are now distinguished **based on their intended audience** and the way they’re handled by the default warnings filters.
>
> [python.org](https://docs.python.org/3/library/warnings.html#warning-categories)

And the target audience of `DeprecationWarning` is _other Python developers_, that of `FutureWarning` _end users of applications_. However, their name is really confusing (maybe more).

But I generally know the reason for their change: in the same way we are facing, in the project:
1. we use `DeprecationWarning` to prompt _the application user_ that some features have been deprecated;
2. and use a custom warning, i.e. `ExperimentalFeatureWarning` to prompt users: some features are under active development and may check;
3. it is conceivable that when we determine certain features that are being developed, some of the interfaces used by _developers_ may be replaced by new interfaces. At this time, we may need a new warning to make developers notice these.

So along this line of thinking and following the official explanation, should we
1. use `FutureWarning` instead of `DeprecationWarning` in the first case;
2. use `ExperimentalFeatureWarning` in the second case;
3. use `DeprecationWarning` in the third case

Does it make sense?

----

Finally, there is another question, how is this PR merged into the master? The current change caused many test cases to fail, so it failed to pass checks; but if I fix all the warns by myself, it may take a long time.

matt-gardner:
Because we're developing a python library, all of our "users" are developers - there is no "application" that we ship.  So the distinction between `DeprecationWarning` and `FutureWarning` is basically meaningless for us, except that `FutureWarning` has a confusing name whose meaning is not clear, so we should just never use it.

As far as how this gets merged, you're right that we can't merge it without fixing the build first, and that that's a lot of work.  I will commit to helping with at least some of it, though it might be a few days before I get to it.

WrRan:
Got it. Thanks for your attention.

schmmd:
There are not actually that many error to fix.  Mostly it's one error that's hit over and over again.  I'm working through a fix with @joelgrus in https://github.com/allenai/allennlp/pull/2310.

schmmd:
One change to note about these new settings is that, instead of showing all the warnings at once, it fails at the first error.  Still I think this is a net good change because the error makes it easy to track down the change and we shouldn't have floods of warnings in our output once this is merged (since we'll start at 0 and warnings will be errors).

schmmd:
Unfortunately with the new `pytest.ini` I don't have the same output locally as on TeamCity.  On two separate Mac's pytest fails on the first exception when I add anything to the `filterwarnings` section.  E.g.

```
======================================================================================== test session starts ========================================================================================
platform darwin -- Python 3.6.8, pytest-4.1.0, py-1.7.0, pluggy-0.8.1
rootdir: /Users/michael/hack/allenai/allennlp, inifile: pytest.ini
plugins: cov-2.6.1, flaky-3.5.1
collected 1210 items / 1 errors                                                                                                                                                                     

============================================================================================== ERRORS ===============================================================================================
_________________________________________________________________________ ERROR collecting allennlp/tests/notebooks_test.py _________________________________________________________________________
allennlp/tests/notebooks_test.py:6: in <module>
    from nbconvert.preprocessors.execute import CellExecutionError
../../../miniconda3/envs/allennlp/lib/python3.6/site-packages/nbconvert/__init__.py:4: in <module>
    from .exporters import *
../../../miniconda3/envs/allennlp/lib/python3.6/site-packages/nbconvert/exporters/__init__.py:3: in <module>
    from .exporter_locator import export_by_name
../../../miniconda3/envs/allennlp/lib/python3.6/site-packages/nbconvert/exporters/exporter_locator.py:28: in <module>
    DeprecationWarning)
E   DeprecationWarning: `nbconvert.exporters.exporter_locator` is deprecated in favor of `nbconvert.exporters.base` since nbconvert 5.0.
===Flaky Test Report===


===End Flaky Test Report===
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
====================================================================================== 1 error in 9.18 seconds ======================================================================================
```

The versions I'm using locally match what TeamCity is using very closely.

schmmd:
We're down to one error (nbconvert).

matt-gardner:
Maybe it's finally time to switch our notebook tutorials to the new fancy version and get rid of the jupyter dependency entirely?

schmmd:
@matt-gardner I'm not sure here's the place to discuss it--but I'm not even keen on that approach.  Previously when we hosted lots of tutorials on our website we had all kinds of problems with versioning between the website and GitHub/AllenNLP--which was confusing for us and for our users.  In any case, the tests seem orthogonal to that decision (our featured tutorial is presently untested).

matt-gardner:
In that case, we should just ignore that warning?

schmmd:
I would ignore the warning or remove the test to fix this.

matt-gardner:
@WrRan, can't we just ignore the warning the same way we do the other third-party warnings?  Removing the test seems like a drastic step for this PR.

WrRan:
@matt-gardner, sorry. I misunderstood what you discussed. I will fix this problem and try to complete this PR in a few days. :disappointed_relieved:

WrRan:
Finally, I passed the UnitTests. (YES!)
But the pylint still complained:
```txt
[Step 4/10] ************* Module allennlp.tests.nn.beam_search_test
[Step 4/10] E: 60, 8: Assigning to function call which doesn't return (assignment-from-no-return)
[Step 4/10] E:132, 8: Assigning to function call which doesn't return (assignment-from-no-return)
[Step 4/10] E:147, 8: Assigning to function call which doesn't return (assignment-from-no-return)
[Step 4/10] ************* Module allennlp.modules.sampled_softmax_loss
[Step 4/10] E:124, 8: Assigning to function call which doesn't return (assignment-from-no-return)
[Step 4/10] ************* Module allennlp.common.util
[Step 4/10] E: 66,23: Module 'numpy' has no 'number' member; maybe 'numbers'? (no-member)
```

I think that is probably not caused by my modifications. So I don't deal with them for now.

@matt-gardner Please have a look. :)

matt-gardner:
Merging with master fixed the pylint errors, but now the sniff tests are failing, because of the deprecation warnings.  @schmmd, what do you think of ignoring warnings for the sniff tests?  It seems like our options are:

1. Ignore the warnings in pytest, which means that anyone using our `pretrained` module will get warnings thrown at them.
2. Ignore the warnings in the `pretrained` module, so tests pass and users don't see the warnings. We would rely on an explicit test failure in the sniff tests to know if we broke something when removing a deprecated feature.
3. Fix the model archives that the `pretrained` module points to, to no longer use the deprecated config parameters.

Option 2 seems like the best option to me, but I'd like to hear others' opinions on this before recommending it.

matt-gardner:
Oh, also, @WrRan, a few people on the team are out this week (due to AAAI and vacations), so we might be a bit slow.

schmmd:
2 makes the most sense to me.

schmmd:
@WrRan I'd love to get this merged so I'll push some commits that implement option 2 as listed above today.

schmmd:
@WrRan thanks again for your hard work.  This is an *awesome* change!  I'm closing this PR since I merged these changes down in https://github.com/allenai/allennlp/pull/2479.  Thanks again!

WrRan:
Happy to be helpful. 
Please notice that I edited the rules (see https://github.com/allenai/allennlp/pull/2282#issuecomment-452288575) for filtering warnings, because I found that when warnings match multiple rules, the last rule is performed.
In addition, I think it might be more appropriate to record the `ExperimentalFeatureWarning` **once** rather than completely **ignore** it. (which is also the strategy currently used.) What do you think? @schmmd

schmmd:
@WrRan I think you originally intended to record ExperimentalFeatureWarning once.  Are you saying that you accidentally didn't implement what we discussed in this PR?  If so, it seems good to change the behavior.

WrRan:
In January, @schmmd you proposed replacing the `once` strategy with `ignore`. (see: https://github.com/allenai/allennlp/pull/2282#issuecomment-452344474)

> @ matt-gardner is the assigned code reviewer so I defer to him, but I think we should silence the `ExperimentalFeatureWarning`s in `pytest`. With the `once` setting they will always show up and clutter our build output. These warnings will still show up when our users try out experimental features.

(And sorry about the delayed feedback. :sweat_smile: )

The current merged strategy is using the `once` settings, which I think is a better strategy.
https://github.com/allenai/allennlp/blob/9f87aa56177ef3dc1b0c58ebb49d15584c500e29/pytest.ini#L31-L32

PR #2290, opened 2019-01-06
[Feature Enhanced] Add HammingLoss

WrRan:
Inspired by [`hamming_loss`](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/metrics/classification.py#L1794) in `sklearn`, I add the `hamming_loss` to metrics in AllenNLP.


WrRan:
Please have a look. :)

matt-gardner:
We don't see a concrete use case for this loss, so we're going to close this PR.  If someone has a concrete example where they want to use this, let us know.

PR #2299, opened 2019-01-07
Masked Flip

vidurj:
Adds a `masked_flip` function that flips a tensor along the time dimension while respecting padding. This is useful for running an RNN in the backward direction along a padded batch. Eg: `masked_flip(lstm(masked_flip(batch, sequence_lengths))[0], sequence_lengths)`.


allennlp/tests/nn/util_test.py
```diff
@@ -282,6 +282,31 @@ def test_masked_mean(self):
         matrix_mean = util.masked_mean(matrix, mask, dim=1).data.numpy()
         assert_array_almost_equal(matrix_mean, numpy.array([[3.0, 0.5], [-1.5, -1.75]]))
 
+    def test_masked_flip(self):
+        tensor = torch.FloatTensor([[[6, 6, 6], [1, 1, 1], [2, 2, 2]], [[3, 3, 3], [4, 4, 4], [5, 5, 5]]])
+        solution = torch.FloatTensor([[[6, 6, 6], [0, 0, 0]], [[4, 4, 4], [3, 3, 3]]])
+        response = util.masked_flip(tensor, [1, 2])
+        self.assertEqual(solution.size(), response.size())
+        self.assertEqual(torch.sum(torch.abs(solution - response)), 0)
```
brendan-ai2: I see other tests in this file using `assert_almost_equal` for tensor equality. Seems like a reasonable replacement for these two lines.

allennlp/nn/util.py
```diff
@@ -339,6 +339,31 @@ def masked_mean(vector: torch.Tensor,
     return value_sum / value_count.clamp(min=eps)
 
 
+def masked_flip(padded_sequence: torch.Tensor,
+                sequence_lengths: List[int]) -> torch.Tensor:
+    """
+        To flip a padded tensor along the time dimension without affecting masked entries.
```
brendan-ai2: Nit: Props for maintaining local consistency with the description of `masked_mean`, but "To flip" reads very strangely. "Flips" or "Flip" would be standard.

PR #2304, opened 2019-01-08
WIP / experimental: trainer refactor

joelgrus:
I'm still not super happy with this, so I'd welcome any and all feedback. At a high level, here are the changes:

1. move all the helper functions out of allennlp.training.trainer into allennlp.training.util
2. move the `TensorboardWriter` class into into its own file
3. create a `Trainer` base class and rename the current trainer to `SupervisedTrainer` (I don't like that name, someone suggest a better one)
3. move all of the Trainer `_send_to_tensorboard` methods to be `TensorboardWriter` instance methods
4. move most of the prework done in `commands.train` (e.g. instantiate the model, load the datasets, create the vocabulary, etc..) into `SupervisedTrainer.from_params`
5. (this ended up making `evaluate_on_test` kind of painful)
6. hardcoded `commands.fine_tune` and `commands.find_learning_rate` to use the `SupervisedTrainer`
7. factored out a lot of the logic in `SupervisedTrainer.from_params` into `SupervisedTrainer.for_model` (again, not thrilled with the name), so that we can either fine-tune an existing model using params or create a new model from_params without having to duplicate code.

Right now I'm not sure that I factored the exact right functionality into the base class vs the specific subclass.

To really appreciate the refactor, look at `commands.train`. The meat of the code there is now basically:

```
    prepare_environment(params)
    create_serialization_dir(params, serialization_dir, recover, force)
    prepare_global_logging(serialization_dir, file_friendly_logging)

    trainer = Trainer.from_params(params, serialization_dir, recover)
    metrics = trainer.train()
```

which seems much more conducive to allowing various training regimes. That's the part I'm pretty happy with, the rest of is still somewhat messy.


allennlp/training/trainer.py
```diff
@@ -24,288 +19,66 @@
...
-            with tensors as values currently support activation logging.
-        should_log_parameter_statistics : ``bool``, optional, (default = True)
-            Whether to send parameter statistics (mean and standard deviation
-            of parameters and gradients) to tensorboard.
-        should_log_learning_rate : ``bool``, optional, (default = False)
-            Whether to send parameter specific learning rate to tensorboard.
-        log_batch_size_period : ``int``, optional, (default = ``None``)
-            If defined, how often to log the average batch size.
-        """
+                 serialization_dir: str,
```
joelgrus: like, maybe this is even too restrictive.

for example, if you wanted to train a GAN you'd have _two_ `Model`s, right?

allennlp/commands/find_learning_rate.py
```diff
@@ -2,24 +2,19 @@
 The ``find-lr`` subcommand can be used to find a good learning rate for a model.
 It requires a configuration file and a directory in
 which to write the results.
-
```
matt-gardner: I thought sphinx required these blank lines...  I'd be happy to be wrong about that, though.

allennlp/commands/find_learning_rate.py
```diff
@@ -45,23 +40,19 @@
...
 import argparse
-import re
 import os
 import math
 import logging
-import shutil
 
 import matplotlib; matplotlib.use('Agg') # pylint: disable=multiple-statements,wrong-import-position
-import matplotlib.pyplot as plt # pylint: disablewrong-import-position
+import matplotlib.pyplot as plt # pylint: disable=wrong-import-position
```
matt-gardner: You didn't introduce this in this PR, but it'd be way nicer to just

```python
# pylint: disable=wrong-import-position
import ...
import ...
...
# pylint: enable=wrong-import-position
```

allennlp/training/supervised_trainer.py
```diff
@@ -0,0 +1,816 @@
...
+from allennlp.training.learning_rate_schedulers import LearningRateScheduler
+from allennlp.training.optimizers import Optimizer
+from allennlp.training.trainer import Trainer
+from allennlp.training.util import time_to_str, move_optimizer_to_cuda, datasets_from_params
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+@Trainer.register("default")
+class SupervisedTrainer(Trainer):
```
matt-gardner: `SingleTaskTrainer`?

allennlp/training/supervised_trainer.py
```diff
@@ -0,0 +1,816 @@
...
+from allennlp.nn import util
+from allennlp.training.learning_rate_schedulers import LearningRateScheduler
+from allennlp.training.optimizers import Optimizer
+from allennlp.training.trainer import Trainer
+from allennlp.training.util import time_to_str, move_optimizer_to_cuda, datasets_from_params
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+@Trainer.register("default")
```
matt-gardner: This should be registered with a name that matches the class semantics (`"single_task"`?), and set as the default implementation in the base class, instead of registered with the name "default".

allennlp/training/supervised_trainer.py
```diff
@@ -0,0 +1,816 @@
...
+    def _description_from_metrics(self, metrics: Dict[str, float]) -> str:
+        if (not self._warned_tqdm_ignores_underscores and
+                    any(metric_name.startswith("_") for metric_name in metrics)):
+            logger.warning("Metrics with names beginning with \"_\" will "
+                           "not be logged to the tqdm progress bar.")
+            self._warned_tqdm_ignores_underscores = True
+        return ', '.join(["%s: %.4f" % (name, value) for name, value in
+                          metrics.items() if not name.startswith("_")]) + " ||"
+
+    def _save_checkpoint(self,
```
matt-gardner: Say I want to make a `MultiTaskTrainer`.  What's the right way to get this function included?  Should it be in the base `Trainer` class?  Or should it be factored out into a utility function that this internal method calls?  Or should I inherit from `SingleTaskTrainer`?  It seems like there are lots of methods in here where this question applies.

allennlp/training/trainer.py
matt-gardner: Well, yeah, though I think it's fine to focus on single-model options for now.  I wouldn't try to make our abstraction fit GANs or other things until we've actually tried to build them ourselves - it's hard to get the abstraction right if you don't fully understand the needs of the use case.

This is a lot better than what was there, and once we have more experience with other training regimes, we can improve it again (which might just involve having a `SingleModelTrainer` base class, though I'm wary of making class hierarchies too deep).

allennlp/commands/find_learning_rate.py
joelgrus: I don't know how these got removed

allennlp/training/supervised_trainer.py
joelgrus: the reason I did it this way is that the existing trainer was already registered as "default", and I didn't want to break that, although it's hard to imagine that anyone was using that

matt-gardner: I see, ok, that makes sense.  If you want to be paranoid about that particular backward compatibility, we could just register this twice.

allennlp/commands/train.py
```diff
@@ -373,19 +203,27 @@ def train_model(params: Params,
...
     best_model_state_path = os.path.join(serialization_dir, 'best.th')
     best_model_state = torch.load(best_model_state_path)
-    best_model = model
+    best_model = trainer.model
     best_model.load_state_dict(best_model_state)
 
-    if test_data and evaluate_on_test:
+    # Hack to get the test dataset loaded for evaluation, since now only
+    # Trainer.from_params knows about it.
+    if evaluate_on_test and "test_data_path" in orig_params:
```
brendan-ai2: Perhaps this is unavoidable, but it's a bit unfortunate for a refactor PR to introduce this hack. Seems like this use-case could help us guide the API design.

matt-gardner: Ah, right, I was going to comment on this, but forgot - why not just put the test evaluation into the `train()` function itself?  We already do validation there, why not have the last piece of `train` also do test, if you asked for it?  That would simplify a lot of this, I think.

joelgrus:
I'm actually working right now on a "proof of concept" unit test that implements a different trainer.

allennlp/commands/train.py
joelgrus: I have two conflicting goals here:

(1) I don't want all of the datasets / iterators stuff to be known to `commands.train`, because I don't want to insist that a trainer necessarily has any of that stuff
(2) I want to somehow preserve the current `evaluate_on_test` behavior

please let me know if you can think of a better way to accomplish these

joelgrus: this works, the only downside is that now "load best weights" has to happen twice, because there's no obvious way to return them (you could make that *always* happen and then `commands.train` just has to assume it's the case, but that seems dangerous)

matt-gardner: It seems reasonable to me to have it always happen; we just document that `SingleTaskTrainer.train()` returns a model with the best weights loaded.  That was the assumed behavior already, I thought.

joelgrus:
OK, this is cleaned up pretty good now.

I added a toy end-to-end example of "multi-task training" in https://github.com/allenai/allennlp/pull/2304/files#diff-71d80b2c1737a0ab1a8dafe67ecedba2

The thing that jumps out at me is that it just *barely* doesn't fit into the existing paradigm. By being clever with iterators and dataset readers I could have gotten almost all the way there -- the only thing that there was really no room for was the multiple `file_path`s. Even that you could have gotten into the old trainer in a really really hacky way.

allennlp/tests/training/multi_task_trainer_test.py
```diff
@@ -0,0 +1,284 @@
...
+        This method should return one epoch worth of batches.
+        """
+        hoppers: Dict[Any, List[Instance]] = defaultdict(list)
+
+        for instance in instances:
+            # Which hopper do we put this instance in?
+            if self.allow_mixed_batches:
+                instance_type = ""
+            else:
+                instance_type = instance.fields[self.type_field_name].metadata  # type: ignore
```
matt-gardner: This isn't a real class, so it's not a big deal, but someone might look at this at some point for an example, so I'll leave a comment: you could instead just define `instance_type` to be a string-ified version of `instance.fields.keys()`.  That's really what matters for the model, anyway.  Then you don't need `type_field_name`, or adding a field in `RoundRobinMingler`, or making sure the model can handle that extra field.

allennlp/tests/training/multi_task_trainer_test.py
```diff
@@ -0,0 +1,284 @@
...
+                 datasets: Dict[str, Iterable[Instance]],
+                 num_epochs: int = 10) -> None:
+        super().__init__(model, serialization_dir)
+        self.iterator = iterator
+        self.mingler = mingler
+        self.optimizer = optimizer
+        self.datasets = datasets
+        self.num_epochs = num_epochs
+
+    def train(self) -> Dict:
```
matt-gardner: Also not something that I'm saying you should actually _do_ for this toy example, but it'd be good to at least think about it, if you haven't.  How do I get the remaining features from `SingleTaskTrainer` in here?  Like, early stopping, tensorboard logging, model checkpointing, etc.  E.g., the actual training step for both cases should be the same - is there something you could pull out of `SingleTaskTrainer` into a utility function that can be called by both of these?

allennlp/tests/training/multi_task_trainer_test.py
matt-gardner: Fair enough, but that's a lot easier to handle with a `SingleTaskTrainer`, so it's not really clear why you should go to all of this trouble for that case.

joelgrus: I'm imagining the case (maybe this is a dumb example) where you have e.g. three different text classification problems, so every instance looks like {"tokens": ..., "label": ..., "dataset": ...}, but they're read from 3 different files, and your model has a different last layer depending on the dataset, does that make sense?

allennlp/training/trainer.py
```diff
@@ -445,670 +88,24 @@ def batch_loss(self, batch: torch.Tensor, for_training: bool) -> torch.Tensor:
...
-                val_loss += loss.detach().cpu().numpy()
-
-            # Update the description with the latest metrics
-            val_metrics = self._get_metrics(val_loss, batches_this_epoch)
-            description = self._description_from_metrics(val_metrics)
-            val_generator_tqdm.set_description(description, refresh=False)
-
-        return val_loss, batches_this_epoch
+    @classmethod
+    def from_params(cls,   # type: ignore
```
DeNeutoy: I think this breaks backward compatibility in a pretty hard way, because the `from_params` args for your abstract `Trainer` differ from the previous concrete trainer, so if anyone has:

```
from allennlp.training import Trainer

Trainer.from_params(model, params, etc)
```
This is going to break backward compatibility. E.g if anyone has written their own `train` function, which I think is probably likely.

allennlp/training/single_task_trainer.py
```diff
@@ -0,0 +1,701 @@
...
+                   evaluation_dataset=evaluation_dataset)
+
+    # Requires custom from_params.
+    @classmethod
+    def from_params(cls,  # type: ignore
+                    params: Params,
+                    serialization_dir: str,
+                    recover: bool = False) -> 'SingleTaskTrainer':
+        # pylint: disable=arguments-differ
+        all_datasets = training_util.datasets_from_params(params)
```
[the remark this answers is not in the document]
joelgrus: where would you put it, then? 

I guess I could move it back to `commands.train` and put it behind a flag?

allennlp/training/trainer.py
joelgrus: that's a good point

allennlp/training/single_task_trainer.py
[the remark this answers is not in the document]
DeNeutoy: Can't it just be reverted? Is there an issue with subclassed trainers having to respect this from_params interface if you want to use our `train` commands, where if you do, you just have to write your own command?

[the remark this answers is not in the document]
joelgrus: I mean, in a lot of ways that defeats the point of this refactor, because you're basically saying that you can only use `allennlp train` if you have a training regime that looks just like the current one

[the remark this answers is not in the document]
matt-gardner: Or, perhaps another abstract class method on the base `Trainer` that has this logic?  I'm not certain how that would look, but there might be a way to make it work.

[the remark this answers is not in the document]
matt-gardner: Hmm, looking at this again, that's basically what `for_model` tries to do, isn't it?  Is there a way we can make that work without taking a `Params` object, so `from_params` happens first, or something?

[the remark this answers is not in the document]
DeNeutoy: I don't think still being able to use our commands for different types of Trainers should be a goal of this refactoring at all. I don't think it is feasible to have a single command which will handle these types of variations. The point was to make it possible to have these Trainer variations without having to copy large chunks of code, which the refactoring definitely achieves. Whether you can use new Trainer subclasses with `allennlp train` is not relevant, in my opinion.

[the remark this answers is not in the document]
matt-gardner: Yeah, I can see that argument.  You're basically saying that the single-task training logic is hard-coded in the `train` command, and that's fine.  Logic for setting up other kinds of training should get its own command (either written by us or by someone else).  I would be fine with that outcome.

allennlp/training/trainer.py
```diff
@@ -1039,7 +538,8 @@ def _restore_checkpoint(self) -> Tuple[int, List[float]]:
         if batch_num_total is not None:
             self._batch_num_total = batch_num_total
 
-        return epoch_to_return, val_metric_per_epoch
+        return epoch_to_return
+
 
     # Requires custom from_params.
     @classmethod
```
matt-gardner: Crazy idea, probably a bad idea, but here's a thought (also not super related to this PR):

The reason this needs a custom `from_params` is because there are two constructor arguments that have sequential dependencies with arguments that were constructed previously.  The `Optimizer` needs the `Model`, and the `LearningRateScheduler` needs the `Optimizer` (and this method moves the model and the optimizer to the right cuda device, but that could be done after-the-fact in the `Trainer` constructor without much trouble).  One way to remove the need for a custom `from_params` entirely is to add previously-constructed arguments to the allowed `**extras` dictionary.  So, after you've constructed a `Model`, you can use it for any future arguments that need it.

The reason this is probably a bad idea is that it's quite magical, on top of the already magical `from_params` logic.  But, I don't _think_ it would be too hard, or too hard to document, so maybe it's worth considering?  If other people think this is a good idea, we could open a separate issue for it and fix it some time.

allennlp/training/trainer_base.py
```diff
@@ -0,0 +1,111 @@
...
+
+        self.model = model
+        self._serialization_dir = serialization_dir
+
+        # Configure GPUs:
+        if not isinstance(cuda_device, int) and not isinstance(cuda_device, list):
+            raise ConfigurationError("Expected an int or list for cuda_device, got {}".format(cuda_device))
+
+        if isinstance(cuda_device, list):
+            logger.warning(f"Multiple GPU support is experimental not recommended for use. "
```
matt-gardner: Is this still true?

allennlp/training/trainer_base.py
```diff
@@ -0,0 +1,111 @@
...
+class TrainerBase(Registrable):
+    """
+    The base class for an AllenNLP trainer. It can do pretty much
+    anything you want. Your subclass should implement ``train``
+    and also probably ``from_params``.
+    """
+    default_implementation = "default"
+
+    def __init__(self,
+                 model: Model,
```
matt-gardner: As you noted, having a single model here precludes fancy stuff with GANs, etc.  If we have this separate base class, we should probably just not include the model.  `batch_loss` could be a utility function that takes the `Model` as an argument, and the only other thing this does is move the model to the GPU, which could just be done in the concrete class's constructor (or in a utility method called by the constructor).

allennlp/training/trainer_base.py
```diff
@@ -0,0 +1,111 @@
...
+                    serialization_dir: str,
+                    recover: bool = False):
+        # pylint: disable=arguments-differ
+        typ3 = params.get("trainer", {}).pop("type", cls.default_implementation)
+        return TrainerBase.by_name(typ3).from_params(params, serialization_dir, recover)
+
+    def _description_from_metrics(self, metrics: Dict[str, float]) -> str:
+        """
+        This is only an instance method so that it can read and set the
+        _warned_tqdm_ignores_underscores flag.
```
matt-gardner: This could easily be a module variable in a `training_util` module instead of a member variable on `TrainerBase`.  Probably cleaner that way.

allennlp/training/trainer.py
```diff
@@ -1111,4 +611,74 @@ def from_params(cls,  # type: ignore
                    log_batch_size_period=log_batch_size_period)
 
 
-Trainer.register("default")(Trainer)
+class TrainerPieces(NamedTuple):
+    """
+    These are things that need to get instantiated from_params
```
matt-gardner: It'd be nicer if this said _why_ this exists this way.  This loads the datasets, constructs a vocabulary, then passes that to the `Model`, making `Trainer.from_params` need less custom logic, and so that if some other `Trainer` has similar needs, it can re-use this.

joelgrus:
this addresses the last round of comments and also adds a toy GAN example (which I think is right, I don't know a ton about GANs).

still TODO: figure out how to add in the bells and whistles (early stopping / checkpointing) to the multi-task example.

allennlp/commands/find_learning_rate.py
```diff
@@ -295,7 +244,7 @@ def search_learning_rate(trainer: Trainer,
             logger.info(f'Loss ({loss}) exceeds stopping_factor * lowest recorded loss.')
             break
 
-        trainer.rescale_gradients()
+        rescale_gradients(trainer.model, trainer._grad_norm) # pylint: disable=protected-access
```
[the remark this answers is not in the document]
joelgrus: if you're asking why `rescale_gradients` is a bare function rather than a Trainer instance method, it's so that anyone can use it as part of their trainer implementation. But I agree this is ugly, so I added a trainer instance method that wraps it.

allennlp/training/single_task_trainer.py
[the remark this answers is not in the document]
joelgrus: I have made a bunch of changes to make things fully backward compatible (`Trainer` is still the supervised trainer and has the exact same public API as before). I do feel pretty strongly that we should allow people to `allennlp train` with alternative trainers. 

For example, once I get this PR merged, my plan is to write a real multitask trainer that hopefully the people on S2 who are using various hacks can use. I really don't want to tell them "by the way, you can't use `allennlp train` to train these models", because I think that makes for a bad (and surprising) user experience.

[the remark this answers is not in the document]
matt-gardner: I don't have a super strong opinion here either way.  But is it really that bad of an experience to tell them "use `allennlp train-multi-task` for the multi-task trainer"?  Just trying to clarify people's opinions and arguments.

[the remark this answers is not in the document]
joelgrus: I think it's a much better experience to say "use `allennlp train` to train your model, vs "use `allennlp train` to train a vanilla supervised model, or `allennlp multi-task-train` to train a multi-task model, or some other command for one of our other trainers, or if you've written your own trainer subclass then you just don't get to run it using a known `allennlp` command."

I don't see what we gain by doing that, other than saving some `if / then` logic in commands.train to preserve backward compatibility.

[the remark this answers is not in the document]
DeNeutoy: I wasn't advocating adding a new command - more that if you want to use a  trainer which doesn't adhere to the `allennlp train` api, such as requiring different constructor arguments etc, you _must_ use allennlp as a library. If other things need to change to make that easier, then we should fix things to make it easier.

allennlp/training/trainer.py
```diff
@@ -998,35 +535,28 @@ def _restore_checkpoint(self) -> Tuple[int, List[float]]:
...
-        if "val_metric_per_epoch" not in training_state:
-            logger.warning("trainer state `val_metric_per_epoch` not found, using empty list")
-            val_metric_per_epoch: List[float] = []
+            self._learning_rate_scheduler.lr_scheduler.load_state_dict(training_state["learning_rate_scheduler"])
+        training_util.move_optimizer_to_cuda(self.optimizer)
+
+        # Currently the ``training_state`` contains a serialized ``MetricTracker``.
+        if "metric_tracker" in training_state:
+            self._metric_tracker.load_state_dict(training_state["metric_tracker"])
+        # It used to be the case that we tracked ``val_metrics_per_epoch``.
```
HarshTrivedi: @joelgrus, it seems there is a small typo here. It should be `val_metric_per_epoch`. Fixed this, and added a test for it in #2352.

hadyelsahar:
Happy I did a keyword search over the repo before I refactor the trainer class myself for training some GANs.
You guys rock !

joelgrus:
in case you didn't see it, I added a unit test that trains a toy GAN:

https://github.com/allenai/allennlp/blob/master/allennlp/tests/training/gan_trainer_test.py

that could be a good starting point for you.

if you run into any difficulties, let me know, I'm thinking a lot about alternative training regimes but don't have a ton of concrete use cases yet.

hadyelsahar:
Thanks I am taking a look on it. I'll let you know my feedback or questions.

However, imo a nicer abstraction won't be specifically only for a GANs but for any adversarial loss. In my opinion, since GANs have differentiability problems for sequence generation, models using adversarial loss in NLP are more common than traditional GANs for NLP at least for now.
Think of Neural models for domain adaptation using adversarial loss such as [ADDA](https://arxiv.org/pdf/1702.05464.pdf) or this [paper]( http://aclweb.org/anthology/D18-1131).

hadyelsahar:
Hey so I finally managed to run some experiments, I thought to share some feedback. 

I started from [gan_train_test.py](https://github.com/allenai/allennlp/blob/master/allennlp/tests/training/gan_trainer_test.py) code  however 
In the end I opted out of using it and rolled back to `allennlp v0.8.1` forked the trainer code and adapted it myself. mainly because of the following: 

* The `GanTrainerTest` class abstraction didn't provide me the isolation I wanted I still had to dig into code details of the trainer:
--  I wasn't sure if this version of the trainer had gpu support `batch.cuda` and `model.cuda`, I had to code it myself.
--  I wanted to implement an ensemble of discriminators, each with a separate training scheme I needed to look into all the trainer code to see how can i design that in your framework. 
-- I needed to log and output some statistics ( I think some event listeners can come very handy here)
-- Maybe for a normal GAN is should be sufficient though


I think this refactor might be backward incompatible, my older code had some CUDA errors (sorry didn't have time to look into that) when I moved to `allennlp v0.8.2` but when I rolled back to `allennlp v0.8.1` it became normal again. Happy to investigate this issue more if you are concerned.

joelgrus:
sorry, I didn't mean you should use that toy gan code (it was only supposed to be a very bare-bones proof of concept, which is why it's only in a unit test), just that it might point you in the right direction.

if you have a regression from 0.8.1 to 0.8.2, we definitely want to know about it.

hadyelsahar:
Alright. I'll try to reproduce the issue anytime soon and let you know.

PR #2305, opened 2019-01-08
Install package in dockerfile

matt-peters:
Installing allennlp inside the `Dockerfile` allows other projects that depend on allennlp to use the docker container as a base image.


Dockerfile
```diff
@@ -52,6 +52,9 @@ COPY .pylintrc .pylintrc
 COPY tutorials/ tutorials/
 COPY training_config training_config/
 COPY setup.py setup.py
+COPY README.md README.md
+
+RUN python setup.py install
```
schmmd: I would use `pip install --editable .` instead.

schmmd:
@matt-peters this seems like a good change but I'm curious--what's your specific motivation for it?  For each commit to master we publish a simpler Docker image with that version of AllenNLP pip installed.  See https://github.com/allenai/allennlp/blob/master/Dockerfile.pip.

matt-peters:
This allows someone to build a base image from a branch that hasn't been merged into master.

schmmd:
@matt-peters  just change `RUN python setup.py install` to `pip install --editable .` and merge it!

PR #2307, opened 2019-01-08
Skip Custom Highway LSTM tests, since they're broken

nelson-liu:
Skipping these tests, since they're pretty much useless in their current state.

Have we decided if we're going to fix the custom extensions, or remove the module? I think there was talk of the latter, at some point?


DeNeutoy:
LGTM, i'll send a PR to remove all of this.

nelson-liu:
thanks @DeNeutoy !

PR #2308, opened 2019-01-08
Check train/dev/test-file-path before process

### LINKED ISSUE #2295: Check for train/dev/test-data before processing them.
Sometimes I config the *wrong* test datapath. And I don't know this error until allennlp has processed all the training and validation datasets, which is a little waste of time.

It may be a good idea to first check the validity of the datapaths before processing them (not including download).

**Additional context**
https://github.com/allenai/allennlp/blob/fb587833084bf8f1a0cb172b5451e21d980f0d5f/allennlp/commands/train.py#L147-L161

My suggestion is to add some checks before Line 161.

WrRan:
Fixes #2295.

### DIFF
--- a/allennlp/common/checks.py
+++ b/allennlp/common/checks.py
@@ -4,8 +4,12 @@ AllenNLP and its models are configured correctly.
 """
 
 import logging
+import os
 import subprocess
+
 from torch import cuda
+
+from allennlp.common.file_utils import cached_path
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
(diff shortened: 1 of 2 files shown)


allennlp/common/checks.py
```diff
@@ -56,9 +62,20 @@ def check_for_gpu(device_id: int):
...
     except FileNotFoundError:
         return False
+
+
+def check_for_file_path(file_path: str, dataset_name: str):
+    file_path = cached_path(file_path)
+    if not os.path.exists(file_path):
+        raise ConfigurationError(f"Experiment specified {dataset_name}, "
+                                 f"but {file_path} doesn't exist.")
+    if not os.path.isfile(file_path):
```
matt-gardner: I don't think we can do this check - we allow for the path to be a directory.

brendan-ai2:
This breaks the multiprocess dataset reader which expects to receive a glob.

PR #2310, opened 2019-01-08
Add raw prefix to avoid warnings from regex

schmmd:
Here's Huggingface's original code: https://github.com/huggingface/pytorch-openai-transformer-lm/blob/master/text_utils.py#L30


schmmd:
This change makes it so we match Huggingface's implementation and cleans up some warnings in the short term, but I am going to open an issue against Huggingface's repo because I think some of their regexes are not what they intended.

PR #2312, opened 2019-01-08
Sparse Gradient Clipping

vidurj:
Currently gradient clipping fails with `clamp is not implemented for type torch.sparse.FloatTensor` if any of the parameters have sparse gradients. This PR ignores them for gradient clipping purposes.


allennlp/training/trainer.py
```diff
@@ -348,7 +348,10 @@ def _enable_gradient_clipping(self) -> None:
         if self._grad_clipping is not None:
             # Pylint is unable to tell that we're in the case that _grad_clipping is not None...
             # pylint: disable=invalid-unary-operand-type
-            clip_function = lambda grad: grad.clamp(-self._grad_clipping, self._grad_clipping)
+            def clip_function(grad):
+                # Sparse tensors do not have a clamp method.
```
matt-gardner: Is there no way to get the same functionality with sparse tensors?  I feel like this should at least emit a warning if it can't do what you want, instead of silently failing.

matt-peters: You should be able to clamp the values directly and create a new sparse tensor to return.

vidurj: True, but do I need to be careful about coalescing the sparse tensor before clamping the values?

matt-peters: I don't think so.  See `sparse_clip_norm` for an example that clips gradient norms of sparse gradients.

vidurj: Thanks!

schmmd:
@vidur-joshi can this be merged or are you waiting on review from @matt-gardner?

vidurj:
Oops, I forgot it hadn't been merged. I'll merge it when the tests pass.

PR #2313, opened 2019-01-08
Text classification models

kernelmachine:
This PR adds support for:
* A bag-of-words token embedder for document representation baselines
* A Seq2Vec linear attention module (e.g. for generating sentence-level representations from words, or document-level representations from sentences)
* Text classification models (+ training configs): logistic regression, seq2seq-encoder classifier, seq2vec-encoder classifier, and a Hierarchical Attention Network (HAN)
* Generic dataset reader for text classification datasets, assuming a simple JSON format (i.e. {"text": ..., "label": ...}) for each instance


kernelmachine:
this is ready for review, @matt-gardner let me know if you think someone else should take a look.

allennlp/data/dataset_readers/textcat.py
```diff
@@ -0,0 +1,129 @@
...
+        See :class:`TokenIndexer`.
+    word_tokenizer : ``Tokenizer``, optional (default = ``{"tokens": WordTokenizer()}``)
+        Tokenizer to use to split the input text into words or other kinds of tokens.
+    segment_sentences: ``bool``, optional (default = ``False``)
+        If true, we will first segment the text into sentences using Spacy and then tokenize words.
+        Necessary for some models that require pre-segmentation of sentences,
+        like the Hierarchical Attention Network.
+    sequence_length: ``int``, optional (default = ``None``)
+        If specified, will truncate tokens to specified maximum length.
+    debug : ``bool``, optional (default = ``False``)
```
matt-gardner: The behavior of this flag is very non-obvious from its name.  Why should "debug" only read 100 instances?  If you want to keep this flag, it should be named better.  But, I would recommend just removing it, because test fixtures are the better way to do this kind of debugging, and we should be encouraging people to use test fixtures instead of a flag like this.

allennlp/data/dataset_readers/textcat.py
```diff
@@ -0,0 +1,129 @@
...
+from allennlp.data.token_indexers import TokenIndexer, SingleIdTokenIndexer
+from allennlp.data.tokenizers import Token
+from allennlp.data.tokenizers import Tokenizer, WordTokenizer
+from allennlp.data.tokenizers.sentence_splitter import SpacySentenceSplitter
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+@DatasetReader.register("textcat")
+class TextCatReader(DatasetReader):
```
matt-gardner: I'm not sure how I feel about `TextCat` as a name for this - we like to avoid abbreviations as much as possible.  Is "textcat" a standard name somewhere?  In the other class names in this PR you use full names ("Classifier"); why not use a full name here, too?  For the class name and the name you register it with.

allennlp/data/dataset_readers/textcat.py
```diff
@@ -0,0 +1,129 @@
...
+        if self._segment_sentences:
+            self._sentence_segmenter = SpacySentenceSplitter()
+
+    @overrides
+    def _read(self, file_path):
+        with open(cached_path(file_path), "r") as data_file:
+            if self.debug:
+                lines = np.random.choice(data_file.readlines(), 100)
+            else:
+                lines = data_file.readlines()
```
matt-gardner: Reading the whole file into memory like this defeats the purpose of laziness.  I'd recommend having the for loop inside this `with` block, and just do `for line in data_file.readlines():`.

allennlp/data/dataset_readers/textcat.py
```diff
@@ -0,0 +1,129 @@
...
+from allennlp.data.tokenizers.sentence_splitter import SpacySentenceSplitter
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+@DatasetReader.register("textcat")
+class TextCatReader(DatasetReader):
+    """
+    Reads tokens and their labels from a labeled text classification dataset.
+    Expects a "tokens" field and a "category" field in JSON format.
```
matt-gardner: You say "category" here, but you use "label" in the code.

matt-gardner: Also, you say JSON format, but you read it as JSON-lines (one json blob per line).

allennlp/data/dataset_readers/textcat.py
```diff
@@ -0,0 +1,129 @@
...
+        -------
+        An ``Instance`` containing the following fields:
+            tokens : ``TextField``
+                The tokens in the sentence or phrase.
+            label : ``LabelField``
+                The label label of the sentence or phrase.
+        """
+        # pylint: disable=arguments-differ
+        fields: Dict[str, Field] = {}
+        text_fields: List[Field] = []
```
matt-gardner: This should go inside the if block below.  And you might consider calling it `sentences`, or something that has more semantics, describing what it represents instead of its type.

allennlp/data/dataset_readers/textcat.py
```diff
@@ -0,0 +1,129 @@
...
+        if self._segment_sentences:
+            sentence_tokens = self._sentence_segmenter.split_sentences(text)
+            if not sentence_tokens:
+                return None
```
matt-gardner: When does this happen?  Same question for the other branch below.  The reason I ask is that if it's when `text` is empty, that should be checked in the calling method.

allennlp/data/dataset_readers/textcat.py
```diff
@@ -0,0 +1,129 @@
...
+                continue
+            items = json.loads(line)
+            text = items["text"]
+            label = str(items["label"])
+            instance = self.text_to_instance(text=text, label=label)
+            if instance is not None:
+                yield instance
+
+    @overrides
+    def text_to_instance(self, text: str, label: str = None) -> Instance:  # type: ignore
```
matt-gardner: If you're allowing this to return `None`, I'd recommend annotating the return type as `Optional[Instance]`.

allennlp/data/dataset_readers/textcat.py
```diff
@@ -0,0 +1,129 @@
...
+            sentence_tokens = self._sentence_segmenter.split_sentences(text)
+            if not sentence_tokens:
+                return None
+            for sentence in sentence_tokens:
+                word_tokens = self._word_tokenizer.tokenize(sentence)
+                if self._sequence_length is not None:
+                    if len(word_tokens) > self._sequence_length:
+                        word_tokens = word_tokens[:self._sequence_length]
+                    else:
+                        padding = [Token("@@PADDING")] * (self._sequence_length - len(word_tokens))
```
matt-gardner: Why are you doing this?  Won't our padding code take care of this for you?  You're also using a non-standard padding token, hard-coded here.  At the least you should be importing the string from somewhere and using a variable here.  Hard-coded values in the middle of a function are generally bad practice.

allennlp/tests/data/dataset_readers/textcat_test.py
```diff
@@ -0,0 +1,46 @@
...
+        instances = reader.read(ag_path)
+        instances = ensure_list(instances)
+
+        instance1 = {"tokens": ['Memphis', 'Rout', 'Still', 'Stings', 'for', 'No', '.', '14',
+                                'Louisville', ';', 'Coach', 'Petrino', 'Vows', 'to', 'Have',
+                                'Team', 'Better', 'Prepared', '.', 'NASHVILLE', ',', 'Tenn.',
+                                'Nov', '3', ',', '2004', '-', 'Louisville', '#', '39;s', '30-point',
+                                'loss', 'at', 'home', 'to', 'Memphis', 'last', 'season', 'is', 'still',
+                                'a', 'painful', 'memory', 'for', 'the', 'Cardinals', '.'],
+                     "label": "2"}
```
matt-gardner: I'm guessing you didn't pick this format, but having numbers as labels does funny things to your outputs, unless you pass `skip_indexing` to the label field.  You might consider having that as an option for the dataset reader, so you don't re-number these numbers.

allennlp/data/dataset_readers/textcat.py
matt-peters: I'd make the name even more verbose and add the JSON file format to the name, to easily distinguish from e.g. a `TextClassificationCSVReader`.

allennlp/data/dataset_readers/textcat.py
```diff
@@ -0,0 +1,129 @@
...
+                        word_tokens = word_tokens + padding
+                text_fields.append(TextField(word_tokens, self._token_indexers))
+            fields['tokens'] = ListField(text_fields)
+        else:
+            tokens = self._word_tokenizer.tokenize(text)
+            if not tokens:
+                return None
+            if self._sequence_length is not None:
```
matt-peters: This entire block of code to handle `self._sequence_length` is duplicated from above, would be remove the duplication.

matt-peters:
Thanks for this @kernelmachine !  Agree with @matt-gardner 's that splitting into smaller pieces would make this much easier to review.

matt-gardner:
@kernelmachine, I think you've split this up into multiple PRs.  If there's still something here you think should be merged, let us know.

PR #2318, opened 2019-01-09
Stacked Bi-Directional LSTM to be compatible with the Seq2Vec wrapper

### LINKED ISSUE #2316: Stacked Bi-Directional LSTM to be compatible with the Seq2Vec wrapper
**Problem/Solution:**
I would like to have an LSTM that is compatible with the [Seq2Vec wrapper](https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2vec_encoders/pytorch_seq2vec_wrapper.py) that has recurrent dropout which the [torch.nn.LSTM](https://pytorch.org/docs/stable/nn.html#lstm) does not have and can be stack-able which the [Augmented LSTM](https://github.com/allenai/allennlp/blob/master/allennlp/modules/augmented_lstm.py) is not therefore making the [Stacked Bi-Directional LSTM](https://github.com/allenai/allennlp/blob/master/allennlp/modules/stacked_bidirectional_lstm.py) compatible with the Seq2Vec wrapper would solve this.

Currently the [Stacked Bi-Directional LSTM](https://github.com/allenai/allennlp/blob/master/allennlp/modules/stacked_bidirectional_lstm.py) is not compatible with the [Seq2Vec wrapper](https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2vec_encoders/pytorch_seq2vec_wrapper.py) because the `final_states` that are returned are a generator and not a tuple. Therefore when [Seq2Vec wrapper](https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2vec_encoders/pytorch_seq2vec_wrapper.py) calls `state.size()` on line 73 it causes an error as a generator does not have a size. All of this can be seen below:

``` bash
def forward(self,  # pylint: disable=arguments-differ
                inputs: torch.Tensor,
                mask: torch.Tensor,
                hidden_state: torch.Tensor = None) -> torch.Tensor:

        if mask is None:
            # If a mask isn't passed, there is no padding in the batch of instances, so we can just
            # return the last sequence output as the state.  This doesn't work in the case of
            # variable length sequences, as the last state for each element of the batch won't be
            # at the end of the max sequence length, so we have to use the state of the RNN below.
            return self._module(inputs, hidden_state)[0][:, -1, :]

        batch_size = mask.size(0)

        _, state, restoration_indices, = \
            self.sort_and_run_forward(self._module, inputs, mask, hidden_state)

        # Deal with the fact the LSTM state is a tuple of (state, memory).
        if isinstance(state, tuple):
            state = state[0]

>       num_layers_times_directions, num_valid, encoding_dim = state.size()
E       AttributeError: 'generator' object has no attribute 'size'

../../Envs/Bella-AllenNLP/lib/python3.6/site-packages/allennlp/modules/seq2vec_encoders/pytorch_seq2vec_wrapper.py:79: AttributeError
```

This was generated by running the following test:
``` python
def test_stacked_bidirectional_lstm_can_complete_forward_pass_seq2vec(self):
    from allennlp.modules.stacked_bidirectional_lstm import StackedBidirectionalLstm
    from allennlp.modules import Seq2VecEncoder
    from allennlp.modules.seq2vec_encoders import _Seq2VecWrapper
    Seq2VecEncoder.register("stacked_bidirectional_lstm")(_Seq2VecWrapper(StackedBidirectionalLstm))
    params = Params({"type": "stacked_bidirectional_lstm",
                        "input_size": 3,
                        "hidden_size": 9,
                        "num_layers": 3})
    encoder = Seq2VecEncoder.from_params(params)
    input_tensor = torch.rand(4, 5, 3)
    mask = torch.ones(4, 5)
    output = encoder(input_tensor, mask)
    assert output.detach().numpy().shape == (4, 18)
```

apmoore1:
Fixes #2316 

I have also added an argument to the Doc String of the constructor to indicate that there is a `use_highway` argument, I have tested that the Sphinx of this documentation looks ok as well and a screenshot showing this:
sphinx documentation

### DIFF
--- a/allennlp/modules/stacked_bidirectional_lstm.py
+++ b/allennlp/modules/stacked_bidirectional_lstm.py
@@ -91,7 +91,8 @@ class StackedBidirectionalLstm(torch.nn.Module):
                                      initial_state[1].split(1, 0)))
 
         output_sequence = inputs
-        final_states = []
+        final_h = []
+        final_c = []
         for i, state in enumerate(hidden_states):
             forward_layer = getattr(self, 'forward_layer_{}'.format(i))
             backward_layer = getattr(self, 'backward_layer_{}'.format(i))
@@ -104,8 +105,11 @@ class StackedBidirectionalLstm(torch.nn.Module):
 
             output_sequence = torch.cat([forward_output, backward_output], -1)
             output_sequence = pack_padded_sequence(output_sequence, lengths, batch_first=True)
-            final_states.append((torch.cat(both_direction_states, -1) for both_direction_states
-                                 in zip(final_forward_state, final_backward_state)))
 
-        final_state_tuple = (torch.cat(state_list, 0) for state_list in zip(*final_states))
+            final_h.extend([final_forward_state[0], final_backward_state[0]])
+            final_c.extend([final_forward_state[1], final_backward_state[1]])
+        
+        final_h = torch.cat(final_h, dim=0)
+        final_c = torch.cat(final_c, dim=0)
+        final_state_tuple = (final_h, final_c)
         return output_sequence, final_state_tuple
--- a/allennlp/modules/stacked_bidirectional_lstm.py
+++ b/allennlp/modules/stacked_bidirectional_lstm.py
@@ -25,6 +25,12 @@ class StackedBidirectionalLstm(torch.nn.Module):
         The dropout probability to be used in a dropout scheme as stated in
         `A Theoretically Grounded Application of Dropout in Recurrent Neural Networks
         <https://arxiv.org/abs/1512.05287>`_ .
+    use_highway: bool, optional (default = True)
+        Whether or not to use highway connections between layers. This effectively involves
+        reparameterising the normal output of an LSTM as::
+
+            gate = sigmoid(W_x1 * x_t + W_h * h_t)
+            output = gate * h_t  + (1 - gate) * (W_x2 * x_t)
     """
     def __init__(self,
                  input_size: int,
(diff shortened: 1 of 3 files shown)


allennlp/modules/stacked_bidirectional_lstm.py
```diff
@@ -104,8 +111,11 @@ def forward(self,  # pylint: disable=arguments-differ
 
             output_sequence = torch.cat([forward_output, backward_output], -1)
             output_sequence = pack_padded_sequence(output_sequence, lengths, batch_first=True)
-            final_states.append((torch.cat(both_direction_states, -1) for both_direction_states
-                                 in zip(final_forward_state, final_backward_state)))
 
-        final_state_tuple = (torch.cat(state_list, 0) for state_list in zip(*final_states))
+            final_h.extend([final_forward_state[0], final_backward_state[0]])
+            final_c.extend([final_forward_state[1], final_backward_state[1]])
```
joelgrus: wouldn't the equivalent of the previous code be something like

```
final_h.append(torch.cat([final_forward_state[0], final_backward_state[0]], -1))
```

?

was that wrong? or is this the same and I'm just not seeing it?

apmoore1: That is equivalent to the previous code, however that would not work with the Seq2Vec wrapper/interface as this module is Bi-Directional the `final_h` and `final_c` should be of shape *(num_layers * num_directions, batch, hidden_size)* which the code I have put in the pull request will be where as the previous code and your suggestion I believe would be of shape *(num_layers, batch, hidden_size)* the place that this would causes an error is in lines 104 to 109 in [PytorchSeq2VecWrapper](https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2vec_encoders/pytorch_seq2vec_wrapper.py) as the module has the `bidirectional` attribute therefore it would reshape the `state` incorrectly. Sorry I did not mention this problem in the issue or pull request.

joelgrus: I am a little wary about changing the shape of that state, but it doesn't break any tests and it makes the seq2vecencoder work, so 🤞

PR #2319, opened 2019-01-09
New NLVR language

matt-gardner:
This PR re-writes the NLVR language using the new API.  Instead of a type declaration, executor, and world, we have a single `NlvrLanguage` object, which inherits from `DomainLanguage` to get grammar induction and execution for free.  The overall result is less code and simpler code.  The actual substantive changes are much smaller than the line change count in the PR would make you think, because the vocabulary of the test fixtures changed, and a bunch of the tests are the same, just moved.

@kl2806, can you take a look at this?  I don't want to merge it until @pdasigi also has a chance to see what I did here, and he's out until next week, but it'd be good to have you review the code.


allennlp/models/semantic_parsing/nlvr/nlvr_coverage_semantic_parser.py
```diff
@@ -180,7 +180,7 @@ def _get_vocab_index_mapping(self, archived_vocab: Vocabulary) -> List[Tuple[int
     @overrides
     def forward(self,  # type: ignore
                 sentence: Dict[str, torch.LongTensor],
-                worlds: List[List[NlvrWorld]],
+                worlds: List[List[NlvrLanguage]],
```
kl2806: Since we don't have a "world" anymore, it might make sense to rename this to "domain" or something like that? Same comment for everywhere that world is used.

allennlp/semparse/domain_languages/domain_language.py
```diff
@@ -313,8 +350,9 @@ def add_predicate(self, name: str, function: Callable):
         return_type = signature.return_annotation
         # TODO(mattg): this might need to just call PredicateType.get_type, or something - what if
         # one of these is a function?
```
kl2806: This TODO is done, right? The code below calls ``PredicateType.get_type``

allennlp/models/semantic_parsing/nlvr/nlvr_coverage_semantic_parser.py
matt-gardner: Yeah, I debated changing that, but that's a pretty big change.  "world" is also relatively standard terminology in semantic parsing for the combination of a domain language and instance-specific state, so an instantiation of the NLVR domain language in a particular context is rightly called a "world".

allennlp/semparse/domain_languages/domain_language.py
matt-gardner: Thanks for the catch, fixed.

allennlp/models/semantic_parsing/nlvr/nlvr_coverage_semantic_parser.py
pdasigi: It does seem a little awkward to have a field called `worlds` be a `List[List[DomainLanguage]]`. Your new implementation of `DomainLanguage` is a replacement for `type_declaration` and `Executor`, correct? Then why not still have a separate (much simpler) `World` class that takes the instance-specific information, and instantiates the appropriate `DomainLanguage`?

scripts/nlvr/sed_commands.txt
```diff
@@ -0,0 +1,20 @@
+s/<<Set\[Object\]:Set\[Object\]>:<Set\[Object]:Set\[Object]>>/<<o,o>,<o,o>>/g
```
pdasigi: Why do you need these commands?

matt-gardner: This was for translating between the old language and the new language, as I was debugging stuff.  It's in an out of the way place, and I thought it could be useful in the future.

allennlp/models/semantic_parsing/nlvr/nlvr_coverage_semantic_parser.py
matt-gardner: In order to do execution, the `NlvrLanguage` object needs instance-specific state.  There is no way around that.  Once `NlvrLanguage` has instance-specific state, I don't know what else a `World` is in charge of that is separate from the `Language`.  The `World` used to have the logical form parsing / translation logic, but we don't need that anymore.  When I was rewriting things, I first kept the `World` object, but replaced everything in it with an underlying `Language` object.  What was left was a thin veneer that translated API calls from the names the old `World` used to what the new `Language` used, which was basically pointless.  I don't see what having a `World` class gives you.

Also consider the `WikiTablesLanguage`.  The instance-specific state affects the language itself.  The language isn't well defined in the absence of instance-specific state.  There's a continuum of languages here: the example `Arithmetic` language has no instance-specific anything, so the constructor doesn't need any parameters.  The `NlvrLanguage` has instance-specific execution, and because we use the executor to define the language itself, the `NlvrLanguage` object must have instance-specific state - without that, the whole re-write falls apart.  Then there's the `WikiTablesLanguage`, which has not only instance-specific execution, but also an instance-specific language.  Again, here, because of how this new API works, there is no `Language` without the instance-specific state.  So what should the `World` do?  There's nothing left for it to do, so I removed it.

pdasigi: I agree that language execution and/or specification needs to be instance-specific (which is why my original suggestion was to make the `World` class instantiate the language). My primary concern is that calling instantiations of sub-classes of `DomainLanguage` `worlds` is confusing, and I suggested one way of avoiding that. However, I do see your point that a `World` class would not be doing much else. Given that, how about renaming the `DomainLanguage` class to `World`?
I like the name "world" because it is standard terminology as you pointed out, and could make it easier for semantic parsing researchers new to AllenNLP understand what's happening.

matt-gardner: So, given that there's only one object here that combines what used to be language definition, language execution, logical form parsing, and instance-specific state, what should it be called?  I think the most salient thing of this new object, the thing that you actually _interact with_, is the language definition.  That's kind of the whole point of this rewrite - you define the language, and you get everything else for free.

On the other hand, in my original PR introducing this API, I [wondered](https://github.com/allenai/allennlp/pull/2281#issuecomment-451612694) if this just be called `World` instead of `DomainLanguage`.  I can see the argument for it, but I think there are good arguments both ways.

Either way, I think changing the name at this point would be a large change, and orthogonal to this PR.  I'd like to merge this and consider whether we should change the name in a separate PR.  What do you think?

pdasigi: Sure, wouldn't mind continuing this discussion in a separate PR.

PR #2321, opened 2019-01-09
QuaRel Domain Language

kl2806:
Adds a domain language for the QuaRel dataset using the grammar induction framework.

This PR is just for the language, I will modify the model, dataset reader, etc. to use the new language in a separate PR.


matt-gardner:
Just a minor point: because my PR got merged, you can rebase yours and remove everything from 
5e22922 up.

allennlp/tests/semparse/domain_languages/quarel_language_test.py
```diff
@@ -0,0 +1,127 @@
+import pytest
+
+from allennlp.common.testing import AllenNlpTestCase
+from allennlp.semparse.domain_languages.quarel_language import QuaRel
```
matt-gardner: Add `QuaRel` to `domain_languages.__init__`, so you can import it here with `from allennlp.semparse.domain_languages import QuaRel`.  You also might consider calling it `QuaRelLanguage`, so that it's more obvious what it is when you instantiate it below (i.e., it's not a qualitative relationship object, it's the language).

allennlp/tests/semparse/domain_languages/quarel_language_test.py
```diff
@@ -0,0 +1,127 @@
...
+from allennlp.semparse import ExecutionError
+
+class QuaRelLanguageTest(AllenNlpTestCase):
+    def setUp(self):
+        super().setUp()
+        self.language = QuaRel()
+
+    def test_constant_logical_form(self):
+        assert self.language.execute('world2').number == 2
+        with pytest.raises(ExecutionError, match='Unrecognized constant'):
```
matt-gardner: This error is tested on the base class already, so there's no need to test it again here.  You can if you really want to, but you don't need to.  It's good to have a test that the language has the particular constants that you expect, though (although, those are already implicitly tested with the more complex logical forms below).

allennlp/tests/semparse/domain_languages/quarel_language_test.py
```diff
@@ -0,0 +1,127 @@
...
+        # Neither answer option is correct.
+        assert self.language.execute(('(infer (speed higher world2) (friction higher world2) '
+                                      '(friction higher world2))')) == -1
+
+        assert self.language.logical_form_to_action_sequence(('(infer (speed higher world1) '
+                                                              '(friction higher world1) '
+                                                              '(friction lower world1))')) == \
+                ['@start@ -> int',
+                 'int -> [<QuaRelType,QuaRelType,QuaRelType:int>, QuaRelType, QuaRelType, '
+                 'QuaRelType]',
```
matt-gardner: This is a case where I'd be just fine going a bit over 100 characters, as long as it's not way over.  Pylint only complains if you go over 115.

allennlp/tests/semparse/domain_languages/quarel_language_test.py
```diff
@@ -0,0 +1,127 @@
...
+    def test_infer_quaval(self):
+        assert self.language.execute(('(infer (and (thickness low world1) '
+                                      '(thickness high world2)) '
+                                      '(strength lower world1) '
+                                      '(strength lower world2))')) == 0
+
+        assert self.language.execute(('(infer (and (thickness low world1) '
+                                      '(thickness high world2)) '
+                                      '(strength lower world2) '
+                                      '(strength lower world1))')) == 1
```
matt-gardner: It looks like you'll currently crash if you use an `and` and the two relations aren't compatible.  You should have a check for this case.

allennlp/tests/semparse/domain_languages/quarel_language_test.py
```diff
@@ -0,0 +1,127 @@
...
+                        '(strength lower world1) '
+                        '(strength lower world2))')
+        action_sequence = self.language.logical_form_to_action_sequence(logical_form)
+        recovered_logical_form = self.language.action_sequence_to_logical_form(action_sequence)
+        assert recovered_logical_form == logical_form
+
+    def test_get_valid_actions(self):
+        valid_actions = self.language.get_valid_actions()
+        assert set(valid_actions.keys()) == {
+                'QuaRelType',
```
matt-gardner: I think it's good practice to test the productions for all of these keys.  I see 8 keys in here and only 6 `check_productions_match` calls.  It's also nice if these are sorted in some reasonable way, and if the order of the `checks` below matches, so you can easily find things.

allennlp/tests/semparse/domain_languages/quarel_language_test.py
matt-gardner: Also, our naming conventions are that the main class in a module should match the module name.  So if you want to stick with `QuaRel` as the name, I'd change the filename to `quarel.py` instead of `quarel_language.py`.

kl2806:
Thanks! @matt-gardner this is ready to be looked at again.

allennlp/semparse/domain_languages/quarel_language.py
```diff
@@ -0,0 +1,109 @@
...
+        def and_function(quarel_0: QuaRelType, quarel_1: QuaRelType) -> QuaRelType:
+            # If the two relations are compatible, then we can return either of them.
+            if self._check_quarels_compatible(quarel_0, quarel_1):
+                return quarel_0
+            else:
+                return None
+        self.add_predicate('and', and_function)
+
+    def _check_quarels_compatible(self, quarel_0: QuaRelType, quarel_1: QuaRelType) -> bool:
+        if quarel_0 and quarel_1:
```
matt-gardner: Nit: you could remove a level of nesting if you do

```python
if not (quarel_0 and quarel_1):
    return False
for ...
```

PR #2322, opened 2019-01-09
Fix backslash exceptions.

schmmd:
I had trouble reproducing the same pytest output as @WrRan has on his PR locally, but here are a few more escape character fixes.


schmmd:
@joelgrus assigning you since @DeNeutoy is out.

joelgrus:
I asked mark about this issue a while back, he said he had no idea if this was an ok change?

matt-gardner:
If it passes tests, I say go for it.  If it breaks something someone depends on, that person can add a better test that would have caught the problem.

schmmd:
@joelgrus @matt-gardner I manually validated that each regular expression is unchanged with the `r'''` syntax--so I would be surprised if we have any problems.

schmmd:
E.g.

```
In [1]: '~"\s*"i' == r'~"\s*"i'                                                                                                                                                                             
Out[1]: True
```

PR #2324, opened 2019-01-09
add infer_and_cast

### LINKED ISSUE #2314: optimizers can't handle params set from environment variables
this was raised in #447 with no resolution but `--overrides` as a workaround.

it should be easy enough to basically look at the params fed to an optimizer and recursively cast them: if it looks like a float, it's a float; etc...

I guess that could go wrong, but I can't think of when it would

joelgrus:
fixes #2314

this seems like a good change to me, but it's possible people might find it objectionable, so speak up if that's the case


allennlp/common/params.py
```diff
@@ -31,6 +31,43 @@ def evaluate_snippet(_filename: str, expr: str, **_kwargs) -> str:
...
+        if value.lower() == "true":
+            return True
+        elif value.lower() == "false":
+            return False
+        else:
+            # See if it could be an int
+            try:
+                return int(value)
+            except ValueError:
+                # See if it could be a float.
```
matt-gardner: I'd probably `pass` here and remove the nesting for the next `try`, so indenting doesn't look so out of control.

allennlp/tests/common/params_test.py
```diff
@@ -314,3 +314,28 @@ def test_to_file(self):
...
+        casted = {
+                "a": [10, 1.3, True],
+                "b": {"x": 10, "y": 20.1, "z": "other things"},
+                "c": "just a string"
+        }
+
+        assert infer_and_cast(lots_of_strings) == casted
+
+        contains_bad_data = {"x": 10, "y": int}
+        with pytest.raises(ValueError):
```
matt-gardner: It'd be nice to add `matches="cannot infer type"` here.

allennlp/training/optimizers.py
```diff
@@ -121,7 +121,14 @@ def from_params(cls, model_parameters: List, params: Params):  # type: ignore
             else:
                 num_parameters += parameter_group.numel()
         logger.info("Number of trainable parameters: %s", num_parameters)
-        return Optimizer.by_name(optimizer)(parameter_groups, **params.as_dict()) # type: ignore
+
+        # By default we cast things that e.g. look like floats to floats before handing them
+        # to the Optimizer constructor, but if you want to disable that behavior you could add a
+        #       "infer_type_and_cast": false
+        # key to your "trainer" config.
```
matt-gardner: Nit: "your "trainer.optimizer" config"?.  Also, it'd be nice if this were discoverable somehow, but that's a more general problem that I'm not sure how to solve...

PR #2327, opened 2019-01-10
More help info for `archive_surgery.py`

WrRan:
Before, typing `python scripts/archive_surgery.py --help`, got:
```bash
usage: archive_surgery.py [-h] --input-file INPUT_FILE --output-file
                          OUTPUT_FILE [--editor EDITOR]

Perform surgery on a model.tar.gz archive

optional arguments:
  --input-file INPUT_FILE
  --output-file OUTPUT_FILE
  --editor EDITOR
```

Now, you will get:
```bash
usage: archive_surgery.py [-h] --input-file INPUT_FILE --output-file
                          OUTPUT_FILE [--editor EDITOR]

Perform surgery on a model.tar.gz archive

optional arguments:
  --input-file INPUT_FILE
                        path to input file (default: None)
  --output-file OUTPUT_FILE
                        path to output file (default: None)
  --editor EDITOR       editor to launch, whose default value is the environment
                        vaiable `$EDITOR` (default: vim)
```

May help users to understand the meaning and state of the parameters.


scripts/archive_surgery.py
```diff
@@ -25,25 +25,37 @@
...
     args = parser.parse_args()
 
-    editor = args.editor or os.environ.get("EDITOR")
-    if editor is None:
+    if args.editor is None:
         raise RuntimeError("please specify an editor or set the $EDITOR environment variable")
 
-    if os.path.exists(args.output_file):
+    if not args.inplace and os.path.exists(args.output_file) and not args.force:
         raise ValueError("output file already exists")
```
schmmd: "output file already exists, use --force to override"

schmmd:
@joelgrus the args seem reasonable to me.  You can assume you don't need force until an error message tells you to use it (once my comment is addressed).  You don't need `force` with `inplace`.

WrRan:
Or change the help string of `--inplace` into "overwrite the input file with the modified configuration"? Is it clearer?

schmmd:
@WrRan that strikes me as an improvement!

WrRan:
Thanks for your feedback. :smile:

schmmd:
@joelgrus does this look good to you?

WrRan:
Another question, should we put this very useful scripts under `allennlp/allennlp/scripts/` directory?
If we do this, we can call this script as follow:
```bash
python -m allennlp.scripts.archive_surgery --help
```

Personally, I think it will be more convenient.

What do you think? And thanks for your attention, again. 😃

PR #2330, opened 2019-01-10
Update create_a_configuration.md howto by removing old allennlp configure

nelson-liu:
@joelgrus is there any way to access the old functionality of `allennlp configure`? or was it completely removed?

I've updated the `create_a_configuration` tutorial for the latter case, but happy to just edit run commands if the functionality is still around.


joelgrus:
I removed the cli part, but it's not very much code:

https://github.com/allenai/allennlp/pull/2118/files#diff-cd8c2aa70e36ddf3bdd444d50027bbb5L36

it should be not much work to create a script that does the same

schmmd:
@nelson-liu care to merge this down?

nelson-liu:
Sure! sorry, it must have fallen off my radar.

PR #2336, opened 2019-01-11
Disable requires_grad for Transformer ELMo parser.

brendan-ai2:
- Makes constituency_parser_transformer_elmo.jsonnet and constituency_parser_elmo.jsonnet comparable by default.


brendan-ai2:
Thanks!

PR #2344, opened 2019-01-11
Variational dropout in Augmented LSTM

### LINKED ISSUE #2320: Variational dropout in Augmented LSTM
**This is more a question and feature/bug I think**
From what I can tell from the [Augmented LSTM](https://github.com/allenai/allennlp/blob/master/allennlp/modules/augmented_lstm.py) the same variational dropout mask is applied to the recurrent state as well as the output of each LSTM cell. I don't think this is the same dropout principal that is stated in the [reference paper](https://arxiv.org/abs/1512.05287) where the dropout mask is different for the recurrent state and the output between each LSTM layer as shown in figure 1 of the paper. I was wondering if this is the case and if it would be possible to have two dropout masks one for the recurrent and one for the output just like the reference paper shows, of which I think there are two ways this can be solved:
1.  [Augmented LSTM](https://github.com/allenai/allennlp/blob/master/allennlp/modules/augmented_lstm.py) having two dropout arguments in it's constructor: `recurrent_dropout_probability` and `output_dropout_probability` and then having two seperate dropout masks one for recurrent and the other for output.
2. [Augmented LSTM](https://github.com/allenai/allennlp/blob/master/allennlp/modules/augmented_lstm.py) will only concern itself with the recurrent dropout and will not apply dropout to the output of each LSTM. The dropout between the LSTM layers will be resolved by adding `output_dropout_probability` to [Stacked Bi-Directional LSTM](https://github.com/allenai/allennlp/blob/master/allennlp/modules/stacked_bidirectional_lstm.py) and using [InputVariationalDropout](https://github.com/allenai/allennlp/blob/master/allennlp/modules/input_variational_dropout.py) on the output sequence of called LSTMs.

The benefit of doing point 2 in the list over 1 is that we can be more consistent with the [torch.nn.lstm](https://pytorch.org/docs/stable/nn.html#lstm) apporach to dropout where they only apply dropout to all output layers apart from the last. Where as if we go for point 1 in the list the dropout would be applied to all output layers including the last.

I am happy to implement this and create a pull request :smile:

apmoore1:
This pull requests fixes #2320 I have done point 2 on the list within the issue. Apart from the points made in the issue I have also added more detail to the doc strings in the [Stacked Bi-Directional LSTM](https://github.com/allenai/allennlp/blob/master/allennlp/modules/stacked_bidirectional_lstm.py) such as the corrected shape of the `final_states` returned form the forward pass and the correct shape required for the `initial_state` argument for the forward pass. I have also included the returned Type of the forward pass for the [Stacked Bi-Directional LSTM](https://github.com/allenai/allennlp/blob/master/allennlp/modules/stacked_bidirectional_lstm.py).

Screen shot of the changed doc strings for the augmented LSTM is below: Only change is the wording of the `recurrent_dropout_probability` argument within the constructor:
augmented doc string

Screen shots of the changed doc strings for the Stacked Bi-Directional LSTM are below: Changes are those stated above with regards to the forward pass and the wording of the `recurrent_dropout_probability` and `layer_dropout_probability` within the constructor as well as the text of the constructor:
stacked constructor doc string
stacked forward doc string


allennlp/tests/modules/stacked_bidirectional_lstm_test.py
```diff
@@ -55,3 +57,43 @@ def test_stacked_bidirectional_lstm_can_complete_forward_pass_seq2vec(self):
         mask = torch.ones(4, 5)
         output = encoder(input_tensor, mask)
         assert output.detach().numpy().shape == (4, 18)
+
+    @flaky(max_runs=3)
+    @pytest.mark.parametrize("dropout_name", ('layer_dropout_probability',
```
DeNeutoy: Are these actually flaky? They shouldn't be, with a dropout rate of 0.9.

allennlp/tests/modules/stacked_bidirectional_lstm_test.py
```diff
@@ -55,3 +57,43 @@ def test_stacked_bidirectional_lstm_can_complete_forward_pass_seq2vec(self):
...
+                                              'recurrent_dropout_probability'))
+    def test_stacked_bidirectional_lstm_dropout_version_is_different(self, dropout_name: str):
+        params = {'input_size': 10, 'hidden_size': 11, 'num_layers': 3}
```
DeNeutoy: Unfortunately because we use `mypy` type checking, you cannot pass kwargs to constructors like this, because it makes mypy unable to check types. Could you pass them explicitly? This will fix the CI build.

allennlp/tests/modules/stacked_bidirectional_lstm_test.py
apmoore1: Yes, I was surprised as well, they pass on average 4 out of 5 runs. 
I have had a look at the output tensors and the differences between them is marginal. 

The layer wise dropout is not applied to the last layer nor is the recurrent dropout in the augmented LSTM applied to the last hidden state that is outputted in the final states, so that might explain why they are so similar? Or I might have made a mistake in the code somewhere but I cannot find it.

allennlp/tests/modules/stacked_bidirectional_lstm_test.py
apmoore1: Just done a commit that will hopefully solve the mypy type checking.

Also thanks for reviewing my code.

allennlp/tests/modules/stacked_bidirectional_lstm_test.py
DeNeutoy: Oh I think maybe it's because you have all zero initial states? Can you try making them random so that we can rule that out.

apmoore1: Changing the Initialisation value from 1 to 0.5 as well as changing the initial states solved the problem. Thanks for the suggestion. I have removed flaky now from the **stacked_bidirectional_lstm_test** as well as the **augmented_lstm_test**, as the augmented lstm had the same problem.

DeNeutoy:
Great, now if you just clean up your unused imports and update the branch against master, hopefully ci will pass!

apmoore1:
I have removed the unused imports but before the commit I forgot to update the branch against master, but I have just checked locally and when I rebase with the master branch there were no conflicts so I'm guessing the ci should pass?

Oneplus:
May I ask a question that if the `LstmCellWithProjection` is doing recurrent dropout in the right way? Since I've noticed that the `LstmCellWithProjection` applies a dropout mask to the output in the old way of `AugmentedLSTM`. 

See https://github.com/allenai/allennlp/blob/master/allennlp/modules/lstm_cell_with_projection.py#L212

apmoore1:
From reading the code of `LstmCellWithProjection` it applies dropout in the same way `Augmented LSTM` use to as you have stated. From what I can tell and this refers back to my [issue with the augmented lstm](https://github.com/allenai/allennlp/issues/2320) is that the same variational dropout mask is applied to the recurrent state as well as the output of each LSTM cell. I don't think this is the same dropout principal that is stated in the [reference paper](https://arxiv.org/abs/1512.05287) where the dropout mask is different for the recurrent state and the output between each LSTM layer as shown in figure 1 of the paper. 

Also as the dropout for  `LstmCellWithProjection` is applied to the output and it is one layer deep this is different to how PyTorch implements dropout on their LSTM's where dropout is only applied to the output of intermediate layers e.g. a 2 layer deep LSTM will have dropout applied between layer 1 and 2 but not on the output of layer 2 which is the last layer. Therefore this current `LstmCellWithProjection` is not consistent with the way PyTorch applies dropout.

For these reasons stated above is why the `Augmented LSTM` no longer has dropout applied to the output, only the recurrent of the LSTM cells as the output dropout is now covered in the [Stacked Bi-Directional LSTM](https://github.com/allenai/allennlp/blob/master/allennlp/modules/stacked_bidirectional_lstm.py).

Issue #2334, opened 2019-01-10
Feature request: momentum schedulers


epwalsh:
Hey all,

I think it would be useful if we had a similar framework for momentum schedulers as we do for learning rate schedulers. The implementation would probably be pretty easy, and almost identical to the learning rate schedulers since the `momentum` value is just stored alongside the learning rate in each `param_group` dict. This would allow someone to imitate Leslie Smith's 1 cycle policy, for example (see https://arxiv.org/abs/1803.09820 and https://sgugger.github.io/the-1cycle-policy.html#the-1cycle-policy). I'd be willing to submit a PR if this seems reasonable.

matt-peters:
Is this as simple as implementing a new "learning rate scheduler" that instead of modifying the learning rate instead changes the momentum?  Or does this require introducing a new abstraction and API for momentum scheduling?

epwalsh:
@matt-peters it could be. It might also make sense to create something like a "BaseScheduler" that both the learning rate and momentum schedulers could inherit from, then it's just a matter of which field in the `param_group` to update

matt-peters:
Makes sense.  This would be a good addition, I'm on board 👍

schmmd:
Closing, since this was implemented in #2469.

PR #2345, opened 2019-01-11
Generalize LR scheduler

epwalsh:
This is the followup PR to feature request https://github.com/allenai/allennlp/issues/2334. 

So far I've implemented a base `Scheduler` and `LearningRateScheduler` class, where `LearningRateScheduler` inherits from `Scheduler`. I also refactored all existing LR schedulers to inherit from `LearningRateScheduler`, and cleaned up the wrappers for PyTorch LR schedulers. Still to do: implement a momentum scheduler and integrate into the `Trainer` class.


allennlp/training/scheduler.py
```diff
@@ -0,0 +1,75 @@
...
+            for i, group in enumerate(self.optimizer.param_groups):
+                if param_group_field not in group:
+                    raise KeyError(f"{param_group_field} missing from param_groups[{i}]")
+                group.setdefault(self._initial_param_group_field, group[param_group_field])
+        else:
+            for i, group in enumerate(self.optimizer.param_groups):
+                if self._initial_param_group_field not in group:
+                    raise KeyError(f"{self._initial_param_group_field} missing from param_groups[{i}]")
+        self.base_values = [group[self._initial_param_group_field] for group in self.optimizer.param_groups]
+        self.step(epoch=last_epoch)
```
matt-peters: This might account for the difference in the trainer test -- in the pytorch base `torch.optim.lr_scheduler._LRScheduler` this line is `self.step(last_epoch + 1)`.

epwalsh: @matt-peters yea, I think that the code / test was wrong before due to the way PyTorch LR schedulers are implemented. I left a note in the docstring explaining it. I think GitHub is having some issues RN though so my latest commits are not showing up

epwalsh: The commits are showing up now, it was just slow. Sorry I didn't see your comment about holding off on the momentum schedulers yet. But I still haven't touched the trainer class

epwalsh:
@matt-peters so far there were only *slight* changes to the existing LR schedulers to account for a slightly different base class. But the behavior is unchanged (although I split up the tests, I didn't change anything significant), other than fixing (I believe) how PyTorch schedulers behave. Previously PyTorch schedulers were prematurely updating the learning rate an epoch early.

matt-peters:
Got it -- despite the pytorch base class prematurely updating the learning rate, if we change the behavior then does it break backward compatibility with any of the existing schedulers?  Or were they already aware of the issue and worked around it?  My concern is different behavior with a given scheduler before/after this change.

epwalsh:
I don't think this breaks backwards compatibility, insofar that the schedules from PyTorch LR schedulers will be the same but just shifted by one epoch. But IMHO the new *shifted* schedule is actually what is expected. I'm not really a fan of how PyTorch's LR schedulers are implemented: `step()` is called when they are initialized, and then `step()` is supposed to be called again *before* the first epoch, which just seems redundant, as opposed to after the epoch (as the AllenNLP trainer does).

Just my opinion though.. it's an easy fix to change it back if you don't agree.

epwalsh:
@matt-peters would it be easier if I broke this PR up? It might make sense to split this into 3 sequential PRs as follows:
1. Move LR schedulers to their own directory, break out cosine, noam, and slanted triangular into their own files, do the same for the tests (so this only involves moving files)
2. Implement the `Scheduler` and `LearningRateScheduler` abstractions and modify the existing LR schedulers to inherit from the new abstractions
3. Implement a `MomentumScheduler` abstraction along with a couple useful concrete momentum schedulers

matt-peters:
I don't think it's necessary to break it up into pieces. But I'm worried about breaking backward compatibility, even if the pytorch behavior calling `step()` is a little wonky.  Can you revert back to the old behavior?

epwalsh:
Sounds good, just did!

PR #2350, opened 2019-01-12
New WikiTables language

matt-gardner:
This PR re-writes the variable-free language using the new API that does automatic grammar induction.  I improved some things while porting the NLVR language, and this PR depends on those improvements, so it's not ready for review until #2319 is merged.  But I'm opening this now so I can test it on CI.

This PR is a net reduction of ~600 lines of code, and the code that's left is _way_ simpler.  I'm pretty happy with how this rewrite of the grammar stuff worked out.

A few points to note on this new language:
- I considered fixing the row / list of rows distinction (as mentioned [here](https://github.com/allenai/allennlp/pull/1709#issuecomment-452746172)), but I thought for this initial PR I'd stick as closely as possible to the original language.  We can change it later if we want to.  Though, for this particular decision, I'm not sure if distinguishing the two actually helps you - you remove some spurious programs, but it comes at the expense of a deeper type system, because you have to add function calls to get from `List[Row]` to `Row`, and the type of `select` gets muddy.
- Related a bit to the above, I removed the "multi-match" stuff entirely.  Instead, I made the multi-matching happen at the _terminal productions_.  This was _a lot_ easier to do than implementing multi-match stuff in the new grammar induction API, and, more importantly, I think the new way makes more sense from a modeling perspective than the previous way.  For instance, `select` has type `<List[Row],Column:List[str]>`.  With the multi-match implementation, you would get three productions for `List[str]`: `['List[str] -> [<List[Row],Column:List[str]>, List[Row], DateColumn]', 'List[str] -> [<List[Row],Column:List[str]>, List[Row], NumberColumn]', 'List[str] -> [<List[Row],Column:List[str]>, List[Row], StringColumn]']`.  This means that the model has to decide _when it's expanding `List[str]`_ what kind of column it is going to use.  Except the column it uses is really a linking decision, and we have a lot more capacity to make that decision when we're predicting terminals, using our linking mechanism.  So we're asking the `List[str]` embedding to do a whole lot, more than it should be doing.  The way I implemented this, `List[str]` has just one production from `select`: `['List[str] -> [<List[Row],Column:List[str]>, List[Row], Column]`, and now there's a `Column` pre-terminal that expands to columns of all types.  That is, the "multi-matching" happens at the terminals (where they are parameterized with a linking mechanism), instead of high up in the abstract syntax tree.  This paragraph describes the modeling reasons for doing it this way, and what it means for how the grammar is different.  To see how it's implemented, look at the changes I made to `domain_language.py`.
- While I did not distinguish between `List[Row]` and `Row`, I _did_ make a minor change to the language, distinguishing between `List[str]` (the result of doing a `select` or a `mode`) and `str` (the words you can use from the question).  `filter_in` and `filter_not_in` using `str` instead of `List[str]`. 
 This seemed like an obvious improvement, as question words are always strings, not lists.  Though, thinking about it more, maybe I messed up - this means you can't filter based on the output of a select.  @pdasigi, do we currently support that?  Do we need it?  It wasn't in any of the parsing tests, or it would have failed.  This is easy to change back, if we need to (we can just mess with the types a bit to coerce the strings to be treated as `List[str]`).


matt-gardner:
Ok, @pdasigi, this is rebased and should be ready for review.

allennlp/semparse/domain_languages/wikitables_language.py
```diff
@@ -0,0 +1,655 @@
...
+        return [row for row in rows if filter_value in row.values[column.name]]
+
+    def filter_not_in(self, rows: List[Row], column: StringColumn, filter_value: str) -> List[Row]:
+        # Assuming filter value has underscores for spaces. The cell values also have underscores
+        # for spaces, so we do not need to replace them here.  Also, we need to remove the
+        # "string:" that was prepended to the entity name in the language.
+        filter_value = filter_value.lstrip('string:')
+        return [row for row in rows if filter_value not in row.values[column.name]]
+
+    @predicate
```
pdasigi: May be move the generic predicates and column-type specific ones into two separate groups, and add a comment above the column-type specific group saying that you added those using `add_predicate` in the `__init__` method.

allennlp/semparse/domain_languages/wikitables_language.py
```diff
@@ -0,0 +1,655 @@
...
+                    year = int(part)
+                else:
+                    day = int(part)
+            elif part in MONTH_NUMBERS:
+                month = MONTH_NUMBERS[part]
+        return Date(year, month, day)
+
+    @staticmethod
+    def _value_looks_like_date(cell_value: str) -> bool:
+        # TODO (pradeep): This will be unnecessary when we have column types identified.
```
pdasigi: Can you please remove this outdated TODO? I missed it earlier.

allennlp/tests/semparse/domain_languages/wikitables_language_test.py
```diff
@@ -0,0 +1,696 @@
+# pylint: disable=no-self-use,invalid-name,too-many-public-methods
```
pdasigi: Looks like these tests are a union of the executor tests and world tests with the types updated. I skimmed through them. Please let me know if there are any news tests, and I'll look at them closely.

allennlp/tests/semparse/domain_languages/wikitables_language_test.py
```diff
@@ -0,0 +1,696 @@
...
+from allennlp.common.testing import AllenNlpTestCase
+from allennlp.data.tokenizers import Token
+from allennlp.data.tokenizers import WordTokenizer
+from allennlp.semparse.contexts import TableQuestionContext
+from allennlp.semparse.domain_languages.domain_language import ExecutionError
+from allennlp.semparse.domain_languages.wikitables_language import Date, WikiTablesLanguage
+from allennlp.tests.semparse.domain_languages.domain_language_test import check_productions_match
+
+
+class TestWikiTablesLanguage(AllenNlpTestCase):
```
pdasigi: I think it is a good idea to include tests that use the `ActionSpaceWalker` here as well, as we discussed yesterday. It's fine to add them in a different PR, but may be add a TODO?

pdasigi:
@matt-gardner Oh, and regarding your question about `str` vs `List[str]`, I think it is better to have everything be a `List[str]`. More generally, the original implementation of this language involved evaluating the "expression" passed as the filter value in all filtering functions, and thus the value used for filtering could itself be a logical form with `select`, `mode`, etc. That said, I think it is okay to merge this PR now and improve the language to include those features later. The language in this repository is not the latest version of our implementation anyway (the one in `allennlp-weak-supervision-research` has some improvements on top of this). Actually, you may way to add a TODO somewhere saying that some features in the language are missing.

allennlp/tests/semparse/domain_languages/wikitables_language_test.py
matt-gardner: I made some minor changes to some of them (mostly updating the types), and removed a few that were no longer necessary, but didn't add any more.

berlino:
@matt-gardner Regarding differentiating between "rows" and "a list of rows", grammar indeed becomes messy  after my attempts. But I think a more realistic way would be enforce some constraints during the execution phase. For instance, it would fail if the function "previous/next" gets a list row with more than one element; on the contrary,  functions like "average/sum/min" should take a list of rows with more than one element (though it's possible to have only one). Consider the cases like "min (argmax ... ), average(argmin(...))", they are very likely to be spurious programs. Does it make sense to you?

matt-gardner:
Yes, failing during execution is easy - we just need to raise an `ExecutionError` at the appropriate time, and be sure we catch it.  I suppose you could use this to filter out a few spurious programs during decoding.  I have no intuition for how often this actually happens, though.

PR #2352, opened 2019-01-14
Fix edgecase of potentially missing best-val-metrics in metrics.json on training recovery

### LINKED ISSUE #2024: On recovering training, metrics.json can potentially miss "best_validation_" metrics.
In the following case, "best_validation_" metrics are not output in the `metrics.json`:

Train the model for few epochs, stop, resume by recovering the state. If the best validation epoch was already reached before stopping then "best_validation_" won't be visible in the final `metrics.json`. This is probably because `metrics` dict is set to {}  in `trainer.train` independent of whether it was in recover mode or otherwise. Later "best_validation_*" and "best_epoch" get inserted / updated as we go ([link](https://github.com/allenai/allennlp/blob/master/allennlp/training/trainer.py#L791)). 

One possible solution is to also recover `"best_validation_*"` and `"best_epoch" ` entries from original `metrics.json` in `trainer._restore_checkpoint` ([here](https://github.com/allenai/allennlp/blob/master/allennlp/training/trainer.py#L941)). These can be used to intialize `metrics` dict in the trainer [here](https://github.com/allenai/allennlp/blob/master/allennlp/training/trainer.py#L746). Therefore even if best_validation epoch is not reached after recovery, it doesn't miss it.

Another solution is to re-evaluate the best validation model again at the end of training and obtain all best validation metrics again.

HarshTrivedi:
Fixes #2024. Sorry for being very tardy to fix this!
It seems that this isn't a good time for this PR, because of the ongoing refactors (#2304).

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -996,7 +996,7 @@ class Trainer(Registrable):
 
         return (model_path, training_state_path)
 
-    def _restore_checkpoint(self) -> Tuple[int, List[float]]:
+    def _restore_checkpoint(self) -> Tuple[int, List[float], Dict[str, float]]:
         """
         Restores a model from a serialization_dir to the last saved checkpoint.
         This includes an epoch count and optimizer state, which is serialized separately
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -785,7 +785,10 @@ class Trainer(Registrable):
             best_validation_metrics = {}
 
         if validation_metric_per_epoch:
-            best_metric = min(validation_metric_per_epoch) if self._validation_metric_decreases else max(validation_metric_per_epoch)
+            if self._validation_metric_decreases:
+                best_metric = min(validation_metric_per_epoch)
+            else:
+                best_metric = max(validation_metric_per_epoch)
             metrics["best_epoch"] = validation_metric_per_epoch.index(best_metric)
         for name, value in best_validation_metrics.items():
             metrics["best_validation_" + name] = value
@@ -845,8 +848,7 @@ class Trainer(Registrable):
                 metrics['best_epoch'] = epoch
                 for key, value in val_metrics.items():
                     metrics["best_validation_" + key] = value
-                    if best_validation_metrics is not None:
-                        best_validation_metrics[key] = value
+                    best_validation_metrics[key] = value
 
             if self._serialization_dir:
                 dump_metrics(os.path.join(self._serialization_dir, f'metrics_epoch_{epoch}.json'), metrics)
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -764,7 +764,7 @@ class Trainer(Registrable):
         Trains the supplied model with the supplied parameters.
         """
         try:
-            epoch_counter, validation_metric_per_epoch = self._restore_checkpoint()
+            epoch_counter, validation_metric_per_epoch, best_validation_metrics = self._restore_checkpoint()
         except RuntimeError:
             traceback.print_exc()
             raise ConfigurationError("Could not recover training from the checkpoint.  Did you mean to output to "
@@ -781,6 +781,14 @@ class Trainer(Registrable):
         metrics: Dict[str, Any] = {}
         epochs_trained = 0
         training_start_time = time.time()
+        if not best_validation_metrics:
+            best_validation_metrics = {}
+
+        if validation_metric_per_epoch:
+            best_metric = min(validation_metric_per_epoch) if self._validation_metric_decreases else max(validation_metric_per_epoch)
+            metrics["best_epoch"] = validation_metric_per_epoch.index(best_metric)
+        for name, value in best_validation_metrics.items():
+            metrics["best_validation_" + name] = value
 
         for epoch in range(epoch_counter, self._num_epochs):
             epoch_start_time = time.time()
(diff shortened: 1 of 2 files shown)


HarshTrivedi:
Sure, no problem. Once that PR gets in, I can rework on this one.

HarshTrivedi:
This is ready for review. Pylint is complaining yet, but I amn't sure why.

HarshTrivedi:
@joelgrus, can you review this? This is small change to `metric_tracker` and `trainer`.

allennlp/training/metric_tracker.py
```diff
@@ -120,3 +123,12 @@ def should_stop_early(self) -> bool:
...
             return self._epochs_with_no_improvement >= self._patience
+
+    # def best_epoch_metrics(self):
+    #     return self._best_epoch_metrics
+
+    # def update_best_epoch_metrics(self, metrics: Dict[str, Float]):
+    #     self._best_epoch_metrics = metrics
+
+    def best_epoch(self, last_run_epoch_num: int):
+        return last_run_epoch_num - self._epochs_with_no_improvement
```
joelgrus: I find this method surprising and unintuitive, why not just store `best_epoch` explicitly?

allennlp/training/trainer.py
```diff
@@ -411,6 +411,10 @@ def train(self) -> Dict[str, Any]:
         epochs_trained = 0
         training_start_time = time.time()
 
+        metrics['best_epoch'] = self._metric_tracker.best_epoch(epoch_counter - 1)
+        for key, value in self._metric_tracker.best_epoch_metrics.items():
+            metrics["best_validation_" + key] = value
+
```
joelgrus: shouldn't this logic live inside `_restore_checkpoint()`?

allennlp/training/metric_tracker.py
```diff
@@ -120,3 +123,12 @@ def should_stop_early(self) -> bool:
             return False
         else:
             return self._epochs_with_no_improvement >= self._patience
+
+    # def best_epoch_metrics(self):
+    #     return self._best_epoch_metrics
+
+    # def update_best_epoch_metrics(self, metrics: Dict[str, Float]):
+    #     self._best_epoch_metrics = metrics
```
joelgrus: delete commented out code

allennlp/training/trainer.py
HarshTrivedi: yeah, that would be the right place. Shifted it. 👍

allennlp/training/metric_tracker.py
HarshTrivedi: Updated 👍

allennlp/training/trainer.py
```diff
@@ -532,7 +538,7 @@ def _save_checkpoint(self, epoch: Union[int, str]) -> None:
                 training_states=training_states,
                 is_best_so_far=self._metric_tracker.is_best_so_far())
 
-    def _restore_checkpoint(self) -> int:
+    def _restore_checkpoint(self) -> Tuple[int, int, Dict[str, float]]:
```
joelgrus: for future reference, I will probably never approve a PR that contains a return type like this :innocent: 

in this particular case, I don't see why this method needs to return the best_epoch or the best_epoch_metrics when you can just get them from the metric_tracker

HarshTrivedi: Well, but that's exactly what I had done earlier, before you suggested to move the logic inside `_restore_checkpoint` (because the logic belongs there). My interpretation was that the point of `_restore_checkpoint` is to do all the things required to restore the training and model state (wrapping `self._checkpointer` and `self._metric_tracker`) so logic to get `best_epoch` and `best_epoch_metrics ` belongs inside and not outside of it.

joelgrus: the *logic* belongs there, but the function doesn't need to return them explicitly, because after restore_checkpoint they should have been restored as part of the metric tracker?

HarshTrivedi: .. may be I misinterpreted. So let me know if so.

allennlp/training/trainer.py
```diff
@@ -584,7 +590,7 @@ def _restore_checkpoint(self) -> int:
         if batch_num_total is not None:
             self._batch_num_total = batch_num_total
 
-        return epoch_to_return
+        return epoch_to_return, self._metric_tracker.best_epoch, self._metric_tracker.best_epoch_metrics
```
joelgrus: what I'm saying is that the trainer can get the best epoch whenever it wants by calling `self._metric_tracker.best_epoch`, so there's no need for this method to return it.

allennlp/training/trainer.py
HarshTrivedi: Cool, no problem. I will revert it back and take care in future 👍.

The only *movable logic* at the point you commented was using `self._metric_tracker.best_epoch` and `self._metric_tracker.best_epoch_metrics`. So, I probably misinterpreted that to be moved up.

allennlp/training/trainer.py
joelgrus: oh, I see, yeah, I was wrong about this :frowning_face:

joelgrus: I think I was thinking that `metrics` was an instance variable, so you could have just updated it in _restore_checkpoint, but it's not

HarshTrivedi: Ohh K, right. Cool, no problem! 👍

HarshTrivedi:
@joelgrus, reminder here!

HarshTrivedi:
Thanks! : )

PR #2357, opened 2019-01-14
sanitize environment variables that can't be unicode encoded

### LINKED ISSUE #2348: param_file reader can't deal with unrelated environment variable which contains non utf-8 character
**Describe the bug**
when I trained with `allennlp train some_config_file -s results_dir --include my_library`,
the program crashed and reported `UnicodeEncodeError: 'utf-8' codec can't encode characters in position 7-12: surrogates not allowed`. The weird thing was, it's ok when I run with Pycharm remote interpreter（my local machine is MacbookPro, and remote server is Linux).

Then I found that in the source code of `allennlp.common.params`, line 424(master branch) `file_dict = json.loads(evaluate_file(params_file, ext_vars=ext_vars))`, if `ext_vars` contains  some variable can't be encoded, like my shell prompt(environment variable `PS1`) on remote server(I use `bira` theme in zsh), the error would happen.

**To Reproduce**
Here are the shell prompt that caused problem.
`'PS1': '(.env) \udce2\udc95\udcad\udce2\udc94\udc80%{$terminfo[bold]$fg[green]%}%n@%m%{$reset_color%} %{$terminfo[bold]$fg[blue]%}%~%{$reset_color%}  $(git_prompt_info)%{$reset_color%}\n\udce2\udc95\udcb0\udce2\udc94\udc80%B$%b '}`

 - OS: Linux
 - Python version: 3.6.5
 - AllenNLP version: 0.7.2

joelgrus:
fixes #2348


schmmd:
Yikes--jsonnet throws an exception if any environment variable contains non-unicode characters?  If so, LGTM.

PR #2364, opened 2019-01-15
Add optional additional tokens to ELMo character indexer

matt-peters:


matt-peters:
It's possible, but the likelihood is pretty low.  I'd expect them to use the `ELMoTokenCharactersIndexer` instead as it's a higher level function and handles entire sequences of tokens instead of a single token.  If they were using it, it is an easy change.

PR #2365, opened 2019-01-15
Bag of words token embedder

kernelmachine:
This PR includes a token embedder called the "bag of words" embedder,  which embeds a set of tokens into a vector of word frequencies.


allennlp/modules/token_embedders/bow_token_embedder.py
```diff
@@ -0,0 +1,87 @@
...
+        Compute a bag of words representation with size batch_size x vocab_size
+
+        Parameters
+        ----------
+        tokens : ``Dict[str, torch.Tensor]``
+            tokens to compute bag of words representation of
+        """
+        bow_vectors = []
+        mask = get_text_field_mask({'tokens': tokens})
+        for document, doc_mask in zip(tokens, mask):
```
matt-gardner: Do `masked_select` and `bincount` not work on batched tensors?  Seems like it would be a lot faster if you could figure out how to make this batched.

allennlp/modules/token_embedders/bow_token_embedder.py
```diff
@@ -0,0 +1,87 @@
+import torch
+from allennlp.modules.token_embedders.token_embedder import TokenEmbedder
+from allennlp.data import Vocabulary
+from allennlp.common import Params
+from allennlp.nn.util import get_text_field_mask
+
+
+@TokenEmbedder.register("bow_token_embedder")
```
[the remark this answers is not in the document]
kernelmachine: Yeah, I think that's a fair point. I think it would be good to make this clearly distinct from a bag of word embeddings etc.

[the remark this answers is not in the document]
kernelmachine: is bag of word _ids_ the right terminology though? I would think something like bag of word _frequencies_ would be a better fit for this module.

allennlp/modules/token_embedders/bow_token_embedder.py
kernelmachine: Yeah, I was trying to figure that out - but [`masked_select`](https://pytorch.org/docs/stable/torch.html#torch.masked_select) seems to return a 1-D tensor regardless of input tensor, and [`bincount`](https://pytorch.org/docs/stable/torch.html#torch.bincount) only takes in a 1-D tensor.

allennlp/modules/token_embedders/bag_of_word_counts_token_embedder.py
```diff
@@ -0,0 +1,80 @@
...
+        super(BagOfWordCountsTokenEmbedder, self).__init__()
+        self.vocab_size = vocab_size
+        if projection_dim:
+            self._projection = torch.nn.Linear(vocab_size, projection_dim)
+        else:
+            self._projection = None
+        self.output_dim = projection_dim or vocab_size
+
+    def get_output_dim(self):
+        return self.num_embeddings
```
matt-gardner: This isn't correct if you have a projection - should be `return self.output_dim`.

allennlp/modules/token_embedders/bag_of_word_counts_token_embedder.py
```diff
@@ -0,0 +1,80 @@
...
+    Parameters
+    ----------
+    vocab_size: int
+        size of vocabulary
+    projection_dim : int, optional (default = None)
+        if specified, will project the resulting bag of words representation
+        to specified dimension.
+    """
+    def __init__(self, vocab_size: int, projection_dim: int = None) -> None:
+        super(BagOfWordCountsTokenEmbedder, self).__init__()
```
matt-gardner: Just `super().__init__()`, no need for the arguments to `super()`.

PR #2369, opened 2019-01-16
pieces for multitask learning

joelgrus:
as discussed at happy hour, this includes an `InterleavingDatasetReader` which wraps multiple other dataset readers and interleaves their instances (adding a `MetadataField` indicating which dataset each instance came from) and a `HomogeneousBatchIterator`, which assumes such a `MetadataField` exists and constructs batches that are homogeneous with respect to its value.

The only "weird" thing is that the file_path passed to `InterleavingDatasetReader.read()` needs to be a JSON-serialized dict { wrapped_reader_key -> file_path }. We discussed alternative designs like passing in a directory and requiring each wrapped reader to know to look for a specific file under the provided directory; I felt like that seemed a little bit too prescriptive about data layout and harder to configure.

I believe that with these pieces most of the multitask things that S2 research wants to do should be relatively easy. (Notably, with the file-path-as-JSON-dict innovation, we can just use the usual `Trainer` 😬 )

FYI @amandalynne


allennlp/data/iterators/homogeneous_batch_iterator.py
```diff
@@ -0,0 +1,74 @@
+from typing import Iterable, Dict, List
+import random
+from collections import defaultdict
+
+from allennlp.common.util import lazy_groups_of
+from allennlp.data.dataset import Batch
+from allennlp.data.instance import Instance
+from allennlp.data.iterators.data_iterator import DataIterator
+
+@DataIterator.register("homogeneous-batch")
```
DeNeutoy: By convention this should be registered as `homogeneous_batch`

allennlp/data/iterators/homogeneous_batch_iterator.py
```diff
@@ -0,0 +1,74 @@
...
+    max_instances_in_memory : ``int``, optional, (default = None)
+        If specified, the iterator will load this many instances at a time into an
+        in-memory list and then produce batches from one such list at a time. This
+        could be useful if your instances are read lazily from disk.
+    cache_instances : ``bool``, optional, (default = False)
+        If true, the iterator will cache the tensorized instances in memory.
+        If false, it will do the tensorization anew each iteration.
+    track_epoch : ``bool``, optional, (default = False)
+        If true, each instance will get a ``MetadataField`` containing the epoch number.
+    partition_key : ``str``, optional, (default = "dataset")
```
DeNeutoy: This is a bit gross, I wonder if it's better to allow setting an "origin" attribute on an `Instance` or something. Maybe not for this PR

joelgrus: it needs to make it into the model, so if you did that you'd have to change all the "batch to tensor" logic to account for that and then make sure it doesn't somehow collide with other tensors and so on

PR #1623, opened 2018-08-17
make fine-tune not expand vocabulary by default

joelgrus:
fixes #1596

Issue #1596, opened 2018-08-09
Better warning / logging about expanding vocabulary during fine tuning


danyaljj:
**Problem Summary**
I just tried to use `fine-tune` bidaf, and got the following error: 
```
RuntimeError: index out of range at /Users/soumith/code/builder/wheel/pytorch-src/aten/src/TH/generic/THTensorMath.cpp:352
```

**Details**
I downloaded the latest version and tried to run `fine-tune`: 

```
python -m allennlp.run fine-tune -m  https://s3-us-west-2.amazonaws.com/allennlp/models/bidaf-model-2017.09.15-charpad.tar.gz -c  /Users/daniel/ideaProjects/allennlp/training_config/bidaf.json  -s  finetune_factor_0.005_epoch_2
```

Here is what I get: 
```
bidaf_allennlp) bash-3.2$ python -m allennlp.run fine-tune -m  https://s3-us-west-2.amazonaws.com/allennlp/models/bidaf-model-2017.09.15-charpad.tar.gz -c  /Users/daniel/ideaProjects/allennlp/training_config/bidaf.json  -s  finetune_factor_0.005_epoch_2
/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)
/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: numpy.ufunc size changed, may indicate binary incompatibility. Expected 192, got 176
  return f(*args, **kwds)
/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)
/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: numpy.ufunc size changed, may indicate binary incompatibility. Expected 192, got 176
  return f(*args, **kwds)
[...]

PR #2374, opened 2019-01-16
Add support for extending Embeddings/Embedders with Extended Vocabulary.

### LINKED ISSUE #2363: Add support for extending Embeddings/Embedders with Extended Vocabulary.
## Problem

In fine-tuning/transfer-learning scenarios, one needs to extend the original vocabulary with vocabulary from new dataset. This already has support (#1416). However, we also need to make the original model compatible with the new extended vocabulary, or else it results in issues like (#1596): Eg. model's embedding matrix is the size of original dataset vocabulary, but the indexing of newer dataset is done with extended vocabulary resulting is possible index-out-of-bound error.

## Possible Solution

Here is a *rough* [commit](https://github.com/HarshTrivedi/allennlp/commit/d133cfed1dd2b846ad48eef5a0bface464570d6d) to layout a possible approach. It would be better to discuss if the approach seems reasonable before I work more it.

Main idea is to have a instance method in Model `extend_embedders_by_vocab(extended_vocab)`, which is responsible to make sure the model can work with the `extended_vocab`. So fine-tuning with vocab extension can look something like this (or check in commit):
```
vocab = model.vocab
vocab.extend_from_instances(vocabulary_params, new_dataset_instances)
model.extend_embedders_by_vocab(vocab)     **
```

The `extend_embedders_by_vocab` calls `extend_by_vocab` on each `TextFieldEmbedder` module present in it to assure it can work with extended vocabulary. `BasicTextFieldEmbedder` further calls `extend_by_vocab` on each of it's `TokenEmbedder` to assure they can work with extended vocabulary. For many `TokenEmbedder`s `extend_by_vocab` would be a no-op. Currently, I have implemented it only for `Embedding`.

HarshTrivedi:
Fixes #2363.

### DIFF
--- a/allennlp/common/testing/model_test_case.py
+++ b/allennlp/common/testing/model_test_case.py
@@ -240,3 +240,22 @@ class ModelTestCase(AllenNlpTestCase):
                                     err_msg=key)
                 else:
                     assert single_predicted == batch_predicted, key
+
+    def test_extend_embedder_vocab(self):
+        model_archive = str(self.FIXTURES_ROOT / 'decomposable_attention' / 'serialization' / 'model.tar.gz')
+        trained_model = load_archive(model_archive).model
+        # config_file used: str(self.FIXTURES_ROOT / 'decomposable_attention' / 'experiment.json')
+        # dataset used: str(self.FIXTURES_ROOT / 'data' / 'snli.jsonl')
+
+        original_weight = trained_model._text_field_embedder.token_embedder_tokens.weight # pylint: disable=protected-access
+        assert tuple(original_weight.shape) == (24, 300)
+
+        vocab = trained_model.vocab
+        counter = {"tokens": {"seahorse": 1}} # 'seahorse' is extra token in snli2.jsonl
+        vocab._extend(counter) # pylint: disable=protected-access
+        trained_model.extend_embedder_vocab(vocab)
+
+        extended_weight = trained_model._text_field_embedder.token_embedder_tokens.weight # pylint: disable=protected-access
+        assert tuple(extended_weight.shape) == (25, 300)
+
+        assert torch.all(original_weight == extended_weight[:24, :])
--- a/allennlp/models/model.py
+++ b/allennlp/models/model.py
@@ -17,6 +17,7 @@ from allennlp.data import Instance, Vocabulary
 from allennlp.data.dataset import Batch
 from allennlp.nn import util
 from allennlp.nn.regularizers import RegularizerApplicator
+from allennlp.modules.text_field_embedders.text_field_embedder import TextFieldEmbedder
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
@@ -319,6 +320,18 @@ class Model(torch.nn.Module, Registrable):
         # This allows subclasses of Model to override _load.
         # pylint: disable=protected-access
         return cls.by_name(model_type)._load(config, serialization_dir, weights_file, cuda_device)
+
+    def extend_embedders_by_vocab(self, extended_vocab: Vocabulary) -> None:
+        """
+        It iterates over each ``text_field_embedder`` of this model and assures
+        it can embed with the extended vocab. This is required in fine-tuning or
+        transfer learning scenarios where model was trained with original vocabulary
+        but during fine-tuning/tranfer-learning, it will have it work with
+        extended vocabulary (original + new-data vocabulary)
+        """
+        for _, module in self._modules:
+            if isinstance(module, TextFieldEmbedder):
+                module.extend_by_vocab(extended_vocab)
 
 
 def remove_pretrained_embedding_params(params: Params):
--- a/allennlp/models/model.py
+++ b/allennlp/models/model.py
@@ -329,7 +329,7 @@ class Model(torch.nn.Module, Registrable):
         but during fine-tuning/tranfer-learning, it will have it work with
         extended vocabulary (original + new-data vocabulary)
         """
-        for _, module in self._modules:
+        for _, module in self._modules.items():
             if isinstance(module, TextFieldEmbedder):
                 module.extend_vocab(extended_vocab)
 
--- a/allennlp/models/model.py
+++ b/allennlp/models/model.py
@@ -321,7 +321,7 @@ class Model(torch.nn.Module, Registrable):
         # pylint: disable=protected-access
         return cls.by_name(model_type)._load(config, serialization_dir, weights_file, cuda_device)
 
-    def extend_embedders_by_vocab(self, extended_vocab: Vocabulary) -> None:
+    def extend_embedder_vocab(self, extended_vocab: Vocabulary) -> None:
         """
         It iterates over each ``text_field_embedder`` of this model and assures
         it can embed with the extended vocab. This is required in fine-tuning or
@@ -331,7 +331,7 @@ class Model(torch.nn.Module, Registrable):
         """
         for _, module in self._modules:
             if isinstance(module, TextFieldEmbedder):
-                module.extend_by_vocab(extended_vocab)
+                module.extend_vocab(extended_vocab)
 
 
 def remove_pretrained_embedding_params(params: Params):
--- a/allennlp/modules/token_embedders/embedding.py
+++ b/allennlp/modules/token_embedders/embedding.py
@@ -66,7 +66,14 @@ class Embedding(TokenEmbedder):
         If given, this will scale gradients by the frequency of the words in the mini-batch.
     sparse : bool, (optional, default=False):
         Whether or not the Pytorch backend should use a sparse representation of the embedding weight.
-
+    vocab_namespace : str, (optional, default=None):
+        In case of fine-tuning/transfer learning, the model's embedding matrix needs to be
+        extended according to the size of extended-vocabulary. To be able to know how much to
+        extend the embedding-matrix, it's necessary to know which vocab_namspace was used to
+        construct it in the original training. We store vocab_namespace used during the original
+        training as an attribute, so that it can be retrieved during fine-tuning.
+        Caveat: This would prevent models trained before this commit to be unextendedable.
+        Any other way to know the namespace used by token_embedder in original training?
     Returns
     -------
     An Embedding module.
@@ -83,7 +90,8 @@ class Embedding(TokenEmbedder):
                  max_norm: float = None,
                  norm_type: float = 2.,
                  scale_grad_by_freq: bool = False,
-                 sparse: bool = False) -> None:
+                 sparse: bool = False,
+                 vocab_namespace: str = None) -> None:
         super(Embedding, self).__init__()
         self.num_embeddings = num_embeddings
         self.padding_index = padding_index
@@ -91,6 +99,7 @@ class Embedding(TokenEmbedder):
         self.norm_type = norm_type
         self.scale_grad_by_freq = scale_grad_by_freq
         self.sparse = sparse
+        self._vocab_namespace = vocab_namespace
 
         self.output_dim = projection_dim or embedding_dim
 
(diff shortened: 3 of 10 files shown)


allennlp/modules/token_embedders/embedding.py
```diff
@@ -140,6 +149,25 @@ def forward(self, inputs):  # pylint: disable=arguments-differ
             embedded = projection(embedded)
         return embedded
 
+    @overrides
```
HarshTrivedi: @matt-gardner,  I agree that full pretrained embedding based extension would be difficult for the reasons you mentioned in the issue. However, would it be worth exposing the API in `extend_vocab` something like :
`extend_vocab(extended_vocab, vocab_namespace = None, **pretrained_file = None**)`? This won't directly be usable via `fine-tune` command since we need to carry over pretrained-embedding-file. But if a user has access to pretrained-file, then with `extend_vocab` API, it can be easy to write a script to make a pretrained model compatible with pre-generated extended vocabulary. Solves my usecase at least : ) I am happy to keep it in separate pr as well.

matt-gardner: I think for ease of separating reviewing into logically-coherent pieces, it's better to just do that in another PR after this one is merged.  It also seems totally fine to me to at least make it possible to get this working in your own script first, then extend that functionality to having stuff work more generally later, so that PR can be relatively simple.

matt-gardner: And, I'll probably get to reviewing this tomorrow.

HarshTrivedi:
Should be make vocab expansion default in fine-tune? Default was changed to not expand in #1623, due to #1596, which this PR fixes. I am fine either ways.

matt-gardner:
Until we fix the general problem of loading pretrained embeddings from an archived file, I'd vote to keep the default as it is.

allennlp/modules/token_embedders/embedding.py
HarshTrivedi: no problem, thanks!

allennlp/modules/token_embedders/embedding.py
```diff
@@ -140,6 +149,25 @@ def forward(self, inputs):  # pylint: disable=arguments-differ
...
+    def extend_vocab(self, extended_vocab: Vocabulary, vocab_namespace: Optional[str] = None):
+        if not vocab_namespace:
+            vocab_namespace = getattr(self, "_vocab_namespace", None)
```
matt-gardner: Can't you just do `self._vocab_namespace`?  This is still backwards compatible, because an old model will still use the current constructor, which sets `self._vocab_namespace = None`.

allennlp/modules/token_embedders/embedding.py
```diff
@@ -140,6 +149,25 @@ def forward(self, inputs):  # pylint: disable=arguments-differ
...
+
+        if not vocab_namespace:
+            vocab_namespace = "tokens"
+            logging.warning("Neither original training store vocab_namespace for Embedding class "
```
matt-gardner: Wording tweak: just say "No vocab_namespace provided to Embedder.extend_vocab. Defaulting to 'tokens'."

allennlp/modules/token_embedders/embedding.py
```diff
@@ -140,6 +149,25 @@ def forward(self, inputs):  # pylint: disable=arguments-differ
             embedded = projection(embedded)
         return embedded
 
+    @overrides
+    def extend_vocab(self, extended_vocab: Vocabulary, vocab_namespace: Optional[str] = None):
```
matt-gardner: The `Optional` annotation on `vocab_namespace` is redundant, because you immediately follow it with ` = None`, which means exactly the same thing.  Just do `vocab_namespace: str = None`.

allennlp/modules/token_embedders/embedding.py
```diff
@@ -140,6 +149,25 @@ def forward(self, inputs):  # pylint: disable=arguments-differ
...
+            vocab_namespace = "tokens"
+            logging.warning("Neither original training store vocab_namespace for Embedding class "
+                            "nor user passed vocab_namespace in extended_vocab. So, "
+                            "defaulting to 'tokens' namespace for embedding extension.")
+
+        extended_num_embeddings = extended_vocab.get_vocab_size(vocab_namespace)
+        extra_num_embeddings = extended_num_embeddings - self.num_embeddings
+        embedding_dim = self.weight.data.shape[-1]
+        extra_weight = torch.FloatTensor(extra_num_embeddings, embedding_dim)
+        torch.nn.init.xavier_uniform_(extra_weight)
```
matt-gardner: You should document somewhere (in this method's docstring?) that you're initializing the new rows with `xavier_uniform`.

allennlp/modules/token_embedders/token_characters_encoder.py
```diff
@@ -35,6 +37,10 @@ def forward(self, token_characters: torch.Tensor) -> torch.Tensor:  # pylint: di
         mask = (token_characters != 0).long()
         return self._dropout(self._encoder(self._embedding(token_characters), mask))
 
+    @overrides
+    def extend_vocab(self, extended_vocab: Vocabulary, vocab_namespace: Optional[str] = "token_characters"):
```
matt-gardner: Again here I'd remove the `Optional` annotation.

allennlp/modules/token_embedders/token_embedder.py
```diff
@@ -24,3 +27,10 @@ def get_output_dim(self) -> int:
         token.  This is `not` the shape of the returned tensor, but the last element of that shape.
         """
         raise NotImplementedError
+
+    def extend_vocab(self, extended_vocab: Vocabulary, vocab_namespace: Optional[str]):
```
matt-gardner: Here you say it's optional, but you don't actually give a default of `None`.  Just remove the `Optional`.

allennlp/common/testing/model_test_case.py
```diff
@@ -240,3 +240,22 @@ def ensure_batch_predictions_are_consistent(
...
                 else:
                     assert single_predicted == batch_predicted, key
+
+    def test_extend_embedder_vocab(self):
+        model_archive = str(self.FIXTURES_ROOT / 'decomposable_attention' / 'serialization' / 'model.tar.gz')
+        trained_model = load_archive(model_archive).model
+        # config_file used: str(self.FIXTURES_ROOT / 'decomposable_attention' / 'experiment.json')
+        # dataset used: str(self.FIXTURES_ROOT / 'data' / 'snli.jsonl')
+
+        original_weight = trained_model._text_field_embedder.token_embedder_tokens.weight # pylint: disable=protected-access
```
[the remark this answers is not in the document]
HarshTrivedi: Ok, I see. Will fix this. 👍

allennlp/models/model.py
```diff
@@ -320,6 +321,18 @@ def load(cls,
...
+    def extend_embedder_vocab(self, extended_vocab: Vocabulary) -> None:
+        """
+        It iterates over each ``text_field_embedder`` of this model and assures
+        it can embed with the extended vocab. This is required in fine-tuning or
+        transfer learning scenarios where model was trained with original vocabulary
+        but during fine-tuning/tranfer-learning, it will have it work with
+        extended vocabulary (original + new-data vocabulary)
+        """
+        for _, module in self._modules.items():
+            if isinstance(module, TextFieldEmbedder):
```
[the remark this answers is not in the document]
HarshTrivedi: Sure 👍

allennlp/modules/token_embedders/embedding.py
```diff
@@ -66,7 +66,14 @@ class Embedding(TokenEmbedder):
...
     sparse : bool, (optional, default=False):
         Whether or not the Pytorch backend should use a sparse representation of the embedding weight.
-
+    vocab_namespace : str, (optional, default=None):
+        In case of fine-tuning/transfer learning, the model's embedding matrix needs to be
+        extended according to the size of extended-vocabulary. To be able to know how much to
+        extend the embedding-matrix, it's necessary to know which vocab_namspace was used to
+        construct it in the original training. We store vocab_namespace used during the original
+        training as an attribute, so that it can be retrieved during fine-tuning.
+        Caveat: This would prevent models trained before this commit to be unextendedable.
```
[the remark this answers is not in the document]
HarshTrivedi: I don't understand. I meant that earlier we didn't store `_vocab_namespace` as an attribute so if we want to extend an old model, there is no way to find original namespace used for `Embedding` objects. So it's not extendable (and won't be extendable even in future), unless the archived model's embedding uses default namespace 'tokens'.  I will shift Caveat to comment, but it should still be caveat an not TODO, right?

allennlp/modules/token_embedders/embedding.py
```diff
@@ -66,7 +66,14 @@ class Embedding(TokenEmbedder):
...
+        Any other way to know the namespace used by token_embedder in original training?
     Returns
```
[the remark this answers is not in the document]
HarshTrivedi: 👍

allennlp/modules/token_embedders/embedding.py
HarshTrivedi: Ok, sure. I've made this mistake a couple of places. Will fix them.

allennlp/modules/token_embedders/embedding.py
HarshTrivedi: Ohh right! This would still be backward compatible, will fix. 👍

allennlp/modules/token_embedders/embedding.py
HarshTrivedi: Sure 👍

allennlp/modules/token_embedders/embedding.py
HarshTrivedi: Sure.

allennlp/modules/token_embedders/token_embedder.py
HarshTrivedi: Actually, I intended to make it optional, it's a typo. The tests are not giving error because the overrided methods are fixing it.

allennlp/modules/token_embedders/embedding.py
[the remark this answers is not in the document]
matt-gardner: I guess I was looking at the question below.  That question is definitely a TODO, or a PR comment.  Maybe it could just be removed.  We _could_ do crazy things to try to recover it (like look through the original config file), but that's not worth the effort and complexity.

Then the caveat can stay there, but you can't reference "this commit" in a docstring or a comment, because it doesn't mean anything.  Maybe: "Added in version 0.9; models trained with 0.8.1 or below need special treatment when extending their vocabularies."

allennlp/modules/token_embedders/embedding.py
matt-gardner: This also means you can just do `vocab_namespace = vocab_namespace or self._vocab_namespace`

HarshTrivedi:
@matt-gardner, this is ready for another look.

HarshTrivedi:
Minor update: I updated of `Embedding`'s `extend_vocab` to allow `pretrained_file` [(in separate branch)](https://github.com/HarshTrivedi/allennlp/commit/71358d1a798e6c431eab22a3e1bf1995cd3704b1). I'll keep it in separate pr though.

allennlp/modules/token_embedders/embedding.py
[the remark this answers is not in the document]
matt-gardner: Looks like you forgot this one.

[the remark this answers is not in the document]
HarshTrivedi: Oops, I thought you are talking about blank line before """ and not after """. Anyways, it's fixed now.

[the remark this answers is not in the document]
matt-gardner: Sorry I wasn't clear - I meant the line right before the line I commented on.  There should be a blank line before the "Returns" block.  I'm pretty sure sphinx is picky about this.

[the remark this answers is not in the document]
HarshTrivedi: Ok, I see. Sure, no problem. Fixed now.

PR #2376, opened 2019-01-17
Incremental seq2seq loss with batch averaging

jungokasai:
This change makes the seq2seq loss computation incremental over time steps and saves memory however long the target sentence is. Averaging of the loss is done accordingly on the fly so that we get the same loss as the original version. For the averaging purpose, the `average="sum"` option is added to `sequence_cross_entropy_with_logits`. There might be a cleaner way to handle this.  Also, this eliminates the softmax normalization for argmax prediction to avoid applying normalization twice (loss and prediction).


matt-gardner:
@brendan-ai2, is it worth finishing this ourselves?  @jungokasai, we'll probably close this soon due to inactivity if we don't hear from you.

brendan-ai2:
Seems like a nice change, but I'm a bit slammed right now. Would be best if @jungokasai can drive it forward.

PR #2386, opened 2019-01-18
Rollback PR #2308 "Check train/dev/test-file-path"

brendan-ai2:
- This breaks the `MultiprocessDatasetReader` which relies on globs.
- Test to prevent regressing.
- The motivation behind PR #2308 is good, but we need to be a bit more sophisticated in how we perform the check. Perhaps `DatasetReader` should provide an optional `check_path` method that users can override?


brendan-ai2:
fyi, @WrRan, @matt-gardner, unfortunately we need to rollback PR #2308. It would be great to get the check added back in a more general way in the future.

brendan-ai2:
Thanks, @joelgrus. Merging on green.

PR #2387, opened 2019-01-18
Allow embedding extension to load from pre-trained embeddings file.

HarshTrivedi:
This is a follow-up to #2374. @matt-gardner, **after** #2374 is merged, can you please review this?


allennlp/modules/token_embedders/embedding.py
```diff
@@ -147,10 +147,14 @@ def forward(self, inputs):  # pylint: disable=arguments-differ
         return embedded
 
     @overrides
-    def extend_vocab(self, extended_vocab: Vocabulary, vocab_namespace: str = None):
+    def extend_vocab(self,  # pylint: disable=arguments-differ
+                     extended_vocab: Vocabulary,
+                     vocab_namespace: str = None,
+                     pretrained_file: str = None):
```
matt-gardner: Can you annotate the return type here (` -> None`)?

allennlp/modules/token_embedders/embedding.py
```diff
@@ -147,10 +147,14 @@ def forward(self, inputs):  # pylint: disable=arguments-differ
...
         """
         Extends the embedding matrix according to the extended vocabulary.
-        Extended weight would be initialized with xavier uniform.
+        If pretrained_file is available, it will be used for extented weight
```
matt-gardner: Wording tweak: "If pretrained_file is available, it will be used for initializing the new words in the extended vocabulary; otherwise they will be initialized with xavier uniform."

allennlp/modules/token_embedders/embedding.py
```diff
@@ -172,11 +180,17 @@ def extend_vocab(self, extended_vocab: Vocabulary, vocab_namespace: str = None):
...
         embedding_dim = self.weight.data.shape[-1]
-        extra_weight = torch.FloatTensor(extra_num_embeddings, embedding_dim)
-        torch.nn.init.xavier_uniform_(extra_weight)
+        if not pretrained_file:
+            extended_num_embeddings = extended_vocab.get_vocab_size(vocab_namespace)
+            extra_num_embeddings = extended_num_embeddings - self.num_embeddings
+            extra_weight = torch.FloatTensor(extra_num_embeddings, embedding_dim)
+            torch.nn.init.xavier_uniform_(extra_weight)
+        else:
+            whole_weight = _read_pretrained_embeddings_file(pretrained_file, embedding_dim,
```
matt-gardner: This was not what I expected, but it should work.  I was imagining some kind of loop over the file just looking for the new vocab entries.  But, this is much simpler logic, and the computation saved by the more complex logic is probably negligible.  This could at least use a comment, I think, like `# It's easiest to just reload the embeddings for the entire vocab, then only keep the ones we need.`

allennlp/modules/token_embedders/token_characters_encoder.py
```diff
@@ -37,10 +37,14 @@ def forward(self, token_characters: torch.Tensor) -> torch.Tensor:  # pylint: di
         return self._dropout(self._encoder(self._embedding(token_characters), mask))
 
     @overrides
-    def extend_vocab(self, extended_vocab: Vocabulary, vocab_namespace: str = "token_characters"):
+    def extend_vocab(self,  # pylint: disable=arguments-differ
+                     extended_vocab: Vocabulary,
+                     vocab_namespace: str = "token_characters",
+                     pretrained_file: str = None):
```
matt-gardner: Same two comments on this class (return type annotation and docstring wording tweak).

allennlp/tests/modules/token_embedders/embedding_test.py
```diff
@@ -281,3 +281,31 @@ def test_embedding_vocab_extension_without_stored_namespace(self):
...
+            embeddings_file.write("word1 1.0 2.3 -1.0\n".encode('utf-8'))
+
+        embedding_params = Params({"vocab_namespace": "tokens", "embedding_dim": 3,
+                                   "pretrained_file": embeddings_filename})
+        embedder = Embedding.from_params(vocab, embedding_params)
+        original_weight = embedder.weight
+
+        assert tuple(original_weight.size()) == (4, 3)  # 4 because of padding and OOV
+
+        vocab.add_token_to_namespace('word3')
```
matt-gardner: Can you change the embeddings for the original vocab here (simulating training), just as an additional sanity check that they don't get modified by `extend_vocab()` with a pre-trained file?

HarshTrivedi:
@matt-gardner If this looks good now, can you merge this? I am waiting to open a follow-up on this.

matt-gardner:
Looks good, thanks!

PR #2393, opened 2019-01-18
Fix typing for python 3.7

### LINKED ISSUE #2382: incompatible with py 3.7.x
ImportError: cannot import 'CallableMeta' and 'GenericMeta' from 'typing'. See line 2 in `allennlp/allennlp/semparse/domain_languages/domain_language.py`.

Reason: `typing` module in 3.7.x deprecated 'CallableMeta' and 'GenericMeta'.

matt-gardner:
Fixes #2382.  I tested this with a local installation of python 3.7, so I'm pretty sure it works.  @zhaoyanpeng, if you could let us know if it fixes the issue for you, too, that'd be great.

### DIFF
--- a/allennlp/semparse/domain_languages/domain_language.py
+++ b/allennlp/semparse/domain_languages/domain_language.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from typing import Any, Callable, CallableMeta, Dict, GenericMeta, List, Set, Tuple, Type, Union  # type: ignore
+from typing import Any, Callable, Dict, List, Set, Tuple, Type, Union
 import inspect
 import logging
 import traceback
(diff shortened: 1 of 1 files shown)


allennlp/semparse/domain_languages/domain_language.py
```diff
@@ -12,6 +13,41 @@
...
 logger = logging.getLogger(__name__)
 
+
+# We rely heavily on the typing module and its type annotations for our grammar induction code.
+# Unfortunately, the behavior of the typing module changed somewhat substantially between python
+# 3.6 and 3.7, so we need to do some gymnastics to get some of our checks to work with both.
+# That's what these three methods are about.
+
+def is_callable(type_: Type) -> bool:
+    if sys.version_info[1] == 6:  # python 3.6
```
joelgrus: I feel like these checks would be slightly more readable if you wrote them as

```
if sys.version_info < (3, 7)
```
what do you think?

matt-gardner: Does checking a tuple work like that?  I didn't realize you could do that.  Fancy.  Assuming it works, I'll do that instead.

joelgrus: yeah, tuples compare lexicographically, so e.g. (3, 6, 8) < (3, 7)

matt-gardner: Thanks for the suggestion, yeah, I like that a lot better.

Issue #327, opened 2017-09-15
Expand vocabulary in `evaluate`


matt-gardner:
We should have a good way to expand a vocabulary in `evaluate` and similar commands, so that you can add words from a set of pre-trained vectors, or assign unique random vectors to new OOV words.  @OyvindTafjord has some preliminary code for doing this; see here:  https://github.com/allenai/allennlp/issues/239#issuecomment-329804862.  There are some sticky issues to figure out around how to tell the `evaluate` command which vectors to use, but that actual expand logic looks pretty good in Oyvind's diff.

Maybe we need to allow for an evaluate json file, that specifies a model archive and other parameters...?

roys174:
This feature is super important for out-of-domain evaluation, and would probably also be beneficial for the online demo, so it would be great to add it. @matt-gardner 's suggestion (adding another parameter for evaluation) sounds like it could do the job.

matt-gardner:
Just an implementation note - #590 made it so it should be relatively easy to do this without any new evaluation json file.  We just need to add some stuff to the training config (and the embedding class, and the trainer...) that adds the word embedding file to the model archive, and augments the vocabulary on load.

schmmd:
This is also important for the `fine-tune` command.

matthew-z:
I wonder what I can do for dataset like SQuAD that test set is hidden at this moment. Is there any way to load the whole Vocabulary from Glove instead of choosing tokens only appeared in train and dev?

matt-gardner:
There's no easy way to do this right now, unfortunately.  PRs welcome!

matthew-z:
Thanks! I just figured out that it is possible load vocabulary from embedding https://github.com/allenai/allennlp/pull/478/files

matt-gardner:
From the parameter documentation: "Words which appear in the pretrained embedding file but not in the data are NOT included in the Vocabulary."

matthew-z:
Oh.....Thank you for the reminder.... 


If these words are not included, then why we want to provide the embedding file? Both of them sound like unusual cases:

> 1. Restricting the Vocabulary to only words present in the pretrained embedding file but I feed this sounds like a very rare case.
> 2. Ensure that words are added if they are present in the pretrained embedding file, regardless of their count.

I wonder if it is better to be possible to add tokens from the embedding regardless of if the word in the count.  it seems that it is not difficult to do.

matt-gardner:
[...]

PR #2395, opened 2019-01-18
Add support for pretrained embedding extension in fine-tuning.

HarshTrivedi:
@matt-gardner , I am working on the final piece (follow up to #2387) - if pretrained file was used in the `Embedding` construction (training), use the same for extension in fine-tuning. This is rough yet, and seems kind-of hacky. But if high level approach seems reasonable, I can refactor things.

Notes:
-----
For this to work, we need to access the pretrained-embedding-file used by `Embedding` during training, in the `Embedding` during fine-tunng.

- *First*, for embedding params having `pretrained_file`, I make it default to add it in `files_to_archive` dict. This is necessary for the files to be available at the time of fine-tuning.
- *Second*, during the training, I store the used `pretrained_file` as an attribute in the instance (similar to what we did with `vocab_namespace` in previous PR).

However, this isn't enough because file named `_pretrained_file` won't be available at same location at time of fine-tuning, instead would be in serialization dir fta as saved by `archive_model` function. So to fix this, I make a mapping of orginal filename (used during training) to replacement filename (available in serialization fta during fine-tuning) and allow it to pass in `extend_vocab` calls.


matt-gardner:
Awesome, thanks @HarshTrivedi, it's great to finally have this functionality (the request goes all the way back to the still-open #327)!  But, @joelgrus, I think you're a lot more familiar with how the `files_to_archive` stuff works than I am - would you mind taking a look at this PR?  E.g., I thought we already handled key replacement in the existing logic, so Harsh's second point in the description shouldn't be necessary, but I'm not sure how it works.

HarshTrivedi:
Just to clarify the need for replacement mapping:

At the time of `Embedding` extension, we need to know what pre-trained file was used for its construction. For this, I store an extra attribute: `_pretrained_file` which was passed in `from_params` of `Embedding`.

Let's say my model embedding in *training* config is like (flat version) is
```
"model.text_field_embedder.token_embedders.tokens.pretrained": "/data/glove/glove.6B.100d.txt.gz"
```
then `embedding._pretrained_file` will be stored as `"/data/glove/glove.6B.100d.txt.gz"`

To make sure, this `pretrained_file` is available during fine-tuning, I force it to be added in `files_to_archive` dict. So we have:
```
files_to_archive["model.text_field_embedder.token_embedders.tokens.pretrained"] = "/data/glove/glove.6B.100d.txt.gz"
```
When the model is archived with this `files_to_archive`, the file is stored in the archive at path:

```
"fta/model.text_field_embedder.token_embedders.tokens.pretrained"
```

And when the archive is loaded, the replacement of appropriate keys happen implicitly, so loaded config looks like (flat-version):
```
"model.text_field_embedder.token_embedders.tokens.pretrained": "fta/model.text_field_embedder.token_embedders.tokens.pretrained"
```

So the problem is, when `Embedding` is invoked the first time, the path was `"/data/glove/glove.6B.100d.txt.gz"` and so `embedding._pretrained_file` is this path. However, after loading from archive, the file is relocated at: `"fta/model.text_field_embedder.token_embedders.tokens.pretrained"`. To fix this, I also generate a mapping of original to replaced name and pass it through the extension methods:

```
mapping["/data/glove/glove.6B.100d.txt.gz"] = "fta/model.text_field_embedder.token_embedders.tokens.pretrained"
```

joelgrus:
sorry, there's something I'm not getting here, the logic says that if I train a model whose config has any (flat) key ending in "pretrained" then I (assume that represents a file and) add that key-value to files_to_archive.

the specific use case, though, is `Embedding`, which takes a `pretrained_file` key, which doesn't trigger that path?

HarshTrivedi:
Oops, sorry about that! It should be `pretrained_file`, and not `pretrained`. Btw, I haven't added tests yet. If we can agree over the high-level approach, I will do that.

HarshTrivedi:
Yes, right. Currently, in this PR, that's a default. It might bloat up the size of archive unnecessarily. But if we want the model to be fine-tunable with embedding extension (w/o passing the files externally), we would need `pretrained-file` to be available in the archive. 

How about I add a option in train to control wether to store files in archive or not, depending on whether that model needs to fine-tuned (with embedding extendibility) or not?

joelgrus:
honestly, including a giant file of glove vectors in your model archive just in case you want to fine-tune later seems like overkill, why not just a solution where you have to pass it in as an override or something?

HarshTrivedi:
Okay, I didn't see your above comment. The overriding the file paths in the model params during fine-tuning won't work directly because we don't use model params at all during fine-tuning.

--- 

To expand on my last comment. Each `embedding` object present somewhere in the `model` need to store some kind of an identifier on what pretrained-file was used and how to locate it during fine-tuning (extension). Our identifier is original pretraind-file-path (eg `"/data/glove/glove.6B.100d.txt.gz"`), which will be stored as `embedding._pretrained_file` and mapping to actually locate that file, will be passed externally as: 

```
{"/data/glove/glove.6B.100d.txt.gz": "my_fine_tuning_system_data_location/path_to_original_pretrained_file"}
```

Does this make sense?

joelgrus:
I guess what I'm suggesting (and maybe this is not possible) is that if the problem is that `fine-tune` needs access to an embeddings file in order to find new embeddings, then I'd rather this behavior be localized to the `fine-tune` command than require changes to how we train and serialize models.

That is, maybe we need to add another argument to `fine-tune` that's `--embedding-sources` or something and then the fine-tune command should do something intelligent with it.

HarshTrivedi:
Yes, I understand. So my last 2 comments are showing a solution in the same direction. We need `--embedding-sources` to give a mapping of original file path to new file path: Eg.

```
{"/data/glove/glove.6B.100d.txt.gz": "my_fine_tuning_system_data_location/path_to_original_pretrained_file"}
```

This mapping is necessary, because there can be more than one embedding objects in a training, each with different pretrained-files. If we don't pass the mapping, there would be no way for `embedding` object to figure which pretrained-file should be used.

If you think passing something like this in `--embedding-sources` is OK, I can update the PR for that. 

But just note that the user who is fine-tuning will need to know what file paths were originally used. This can be found easily by unarchiving even if the model wasn't trained by same person. So I don't think it's as such a problem.

HarshTrivedi:
Yeah, so a follow up on this PR would be to allow `vocabulary` and `embedding` extension in `evaluate` command as well - #327 is precisely my use case. If we get the functionality in `fine-tune` with either of the above two approaches, it would be pretty straightforward for `evaluate` command.

HarshTrivedi:
@joelgrus , I have reverted the changes related to training and serialization of models and updated the code according to the discussion above. Now, for embedding extension, we need to explicitly pass a mapping from original file path to new file path, if the path has changed; otherwise, original file path itself will be used.

HarshTrivedi:
@joelgrus, any thoughts / suggestions on how should I proceed this PR?

HarshTrivedi:
We need to store some kind-of *Identifier* in the `embedding` object, at training time, to be able to find right pretrained-file for that `embedding` object at fine-tuning time. If original pretrained-files are not available at the same location, user can pass mapping `{Identifier -> pretrained-file}` at time of fine-tuning to specify updated file paths. 

In above comment, you have pointed out three possible *Identifiers* to figure out which embedding instance should use which pretrained-file. My comments for each:

## 1. Use params-path eg. "text_field_embedder.token_embedder.tokens"

It is not possible to know such params-path (eg "text_field_embedder.token_embedders.tokens") inside the `embedding` object, from the *current* `embedding_params`,  used to construct the `Embedding`. We can surely add code to automatically pass params-path information in each `embedding_params`. For example, automatic conversion of:

```
"text_field_embedder": {"token_embedders": {"tokens_1": {....}  , "tokens_1": {....}}}

```
to
```
"text_field_embedder": {"token_embedders": {
                             "tokens_1": {.... "identifier_path": "text_field_embedder.token_embedders.tokens_1"}  ,
                             "tokens_2": {.... "identifier_path": "text_field_embedder.token_embedders.tokens_2"}
                       }}
```
right after popping model params.


**Problem**: Requires somewhat weird-looking model_params mutation for this specific case.

## 2. Use full pretrained-file names

This PR is currently doing this.

**Problem**: User will have to know what original path was. It can be found from the config.json in the archive. It's not too bad as such, because we are saying **only** if original file path is not available at same place, tell what how the path changed by mapping old to new path.


## 3. Use unique names given by user.

A user will need to give unique name identifiers in the `embedding_params` at the *training* time.

```
"text_field_embedder": {"token_embedders": {
                             "tokens_1": {.... "identifier_name":  "name1"}  ,
                             "tokens_2": {.... "identifier_name": "name2"}
                       }}
```

**Problem**: A user will have to decide upfront, at the time of training that he wants to fine-tune it. Let's say I forgot to give names to embedding, or didn't plan fine-tuning upfront, I will have to retrain to make it fine-tunable.


**Common Problem in all three**: We wouldn't be storing any file identifier in `embedding` object until this PR. So models trained before this commit won't be extendable, *unless* the original pretrained-file is yet available at same path.

Of the three, I would least prefer (3). Of (1) and (2), (2) has weirdness from user end and (1) has little weirdness from our (code's) end. 

So I prefer, **(1), (2), (3) in decreasing order**.

matt-gardner:
Why do you need to store anything on the embedding object?  Why not just require that if you pass in `--extend-vocab`, you also give paths for whatever modules need pre-trained vectors?

HarshTrivedi:
Btw, at time of fine-tuning, we discard the `model_params` all together. All the we use for fine-tuning is the `model` object. So even if we allow some overrides-like dict, there isn't a way to use that in construction of `model` object. Therefore, `model` object itself needs to store some information somewhere to be able to distinguish among the `embeddings` in the first place - like a unique identifier for each `embedding` object.

matt-gardner:
In here: https://github.com/allenai/allennlp/blob/b7d56ae01708258821ffd1a98e3ee0610b6851e9/allennlp/modules/text_field_embedders/basic_text_field_embedder.py#L154-L162

I take an additional dictionary, mapping token embedder names (keys) to pre-trained files.  I pass the value with the appropriate key, if given, through to `token_embedder.extend_vocab()`.  And the same thing applies to everywhere that this is needed.  Does this make sense?  No storing of extra state required.

HarshTrivedi:
Yeah, right. But there can be multiple `basic_text_field_embedder` and multiple `token_embedders` (like `embedding`) in the root object, so only `token_embedder` key won't be sufficient to uniquely identify `embedding` object.

There is way to navigate from `basic_text_field_embedder` to it's `token_embedder`s by the keys, but I don't understand how to navigate with key from `model` to it's many possible `basic_text_field_embedder`s.

https://github.com/allenai/allennlp/blob/37212200bee6e1bc3bfa6c970859a80153fd56fd/allennlp/models/model.py#L324-L336

I don't think the `key` of `self._modules` here is the same key used in model params.

Consider the case that there are two `basic_text_field_embedder`'s and one `embedding` in model root: So three `embedding` objects in all. Something like:
```
"text_field_embedder_1": {"token_embedder": {"tokens1": ... } },
"text_field_embedder_2": {"token_embedder": {"tokens1": ... } },
"token_embedder": {"tokens1": ...}
```

For user, passing token embedder key to file mapping won't be sufficient here. If we make them pass full paths, I am not sure how can we navigate the full path in `model.extend_embedder_vocab`.

Let me know if I missed something.

joelgrus:
it will be a little intricate to get things right, but why can't you just pass in a path like

`"_text_field_embedder._token_embedders.tokens"`

that means "given the model, look for its _text_field_embedder property, look for that guy's `_token_embedders` property, look for that guy's `"tokens"` key?

HarshTrivedi:
@joelgrus  As you suggest, If we use model attribute keys instead of params keys, then it will work. However, to know what (possibly nested) keys to use, a user will have to check the code. For example, a text field embedder could be stored as an attribute in `_text_field_embedder`, `_text_field_embedder_1`, ``my_text_field_embedder`` or anything. And the `model._modules.keys()` as used in snippet about, is referring to these keys.

If I am taking an off-then-shelf pretrained model and want to fine-tune it, I would be fine looking at config file and passing extra information based on that. But now, user will also have to actually see the code of that model to know how to locate the embeddings. So, paths based on config-params are still fine, but paths based on model-attributes, might be a bad user experience.

HarshTrivedi:
Btw, also note that `Embedding` might not be present inside `BasicTextFieldEmbedder` or `model` directly, but present somewhere in the nested modules with arbitrary names. Eg. `model._module1._module2.`.  Making user setup a path like this can surely be tedious.

HarshTrivedi:
If you are giving on weirdness on users-end, then the original solution is already much **easier** and without **any** additional change to way things work - Making user map original to new path **ONLY** in the case the original path is not available in same system. A user can just lookup config.json and know how to configure. I guess you would aggree that this is much easier than having the user setup something like `._module1._module2.tokens` by looking at the code as example above.

We only need to store `pretrained_file` in embedding object, *which we anyway want to store*. `pretrained_file` is already passed in `embedding_params` so nothing extra required. **Note** that storing `pretrained_file` isn't necessary but it makes MOST of the cases seamless without any additional configuration: eg. training system and fine-tuning system are same or at training, URL was used for embedding and it's yet available.

HarshTrivedi:
> it requires the user to know very specific details about how the model was trained (i.e. what actual paths were used on the training computer) that are unavailable except by manually inspecting the model.tar.gz

It's not necessary to even open `archive`, one can just run without passing anything extra. And the error message can tell it **exactly** what mapping is required to be passed. At least to me, this is leads to much less confusion than making me setup regex and look up the code, which could be arbitrarily complex.

I agree that the following side-effect is unwanted: "identical" models but that were trained on two different machines, then the commands to fine-tune them are different." But yet, considering the trade-off, I would prefer it over having to lookup the code.

matt-gardner:
@HarshTrivedi, what's your recommendation for extending the vocabulary of a model that was trained before this PR?

joelgrus:
if you wanted to you could make it so that running `allennlp fine-tune` and saying expand embeddings but not specifying the overrides would print out exactly what overrides were needed, then you wouldn't have to look at code to figure out the paths.

matt-gardner:
Sorry if I'm missing something obvious, but "if the embedding file path isn't present at the same location" - how do you know the original location in the first place?  Seems like you should just always require the file path to be passed every time you want to use pre-trained vectors when expanding your vocabulary.  Right?

As far as printing out error messages, you could imagine passing down a string as you go through the methods, so that you can print out the full attribute path when you get to the module that actually needs the file.  That's an unfortunate bit of API overhead, though, with extra method arguments just for logging stuff.  Another possibility is to make careful use of throwing and catching exceptions, building up the attribute path as you catch and raise, so you get a complete error message.  Does that make sense?

HarshTrivedi:
The `pretrained_file` is one of the keys in embedding_params. It wasn't passed inside the `Embedding` constructor before this PR. But now I do store it as an attribute in embedding object. There is not harm in doing this - just like we stored `vocab_namespace` as an extra attribute in the previous PR.

So, first step would be to see if `embedding.pretrained_file` is reachable. It will be mosty available if originally URL was used, or the training and fine-tuning systems are the same. Only if it is not reachable, we require the extra mapping from the user. Other case is, the model was trained before this PR - in that case the stored attribute won't be present and we will need mapping.

About error message, I was thinking in lines of catch and raise only. Let me do the rest and worry about logging later.

matt-gardner:
@HarshTrivedi, that's a good point.  I don't have a strong opinion between the options that you've presented here; I've mostly been commenting trying to understand and clarify what all of the alternatives were.  Whatever you and @joelgrus think is best is fine with me.

HarshTrivedi:
@joelgrus, if you think we have discussed the options, let me know what your final say is. I am happy to consider other alternatives if you have in mind.

joelgrus:
what exactly is the most recent proposal? is it

(1) an embedding layer should store its pretrained_file_path and vocab_namespace (if any)
(2) at fine-tune time we pass in an argument (either filename -> filename or model_path -> filename) and expand the embeddings from there?

here is the part I'm struggling with: there's nothing intrinsic to an `Embedding` that requires it to have a file_path or a vocab_namespace, it's almost like there's a `PretrainedEmbedding` subclass that has those things. But for example if you use learned character embeddings neither of those concepts would even be meaningful. But all of the solutions I like here would be a lot of work and/or breaking changes.

also, I'm not too concerned about not being able to expand-vocab on already trained models, we can just warn in that case.

do I have the final proposal correct? I feel like this is pretty messy however we solve it, so maybe I should stop worrying about it being kind of messy.

HarshTrivedi:
@joelgrus  Yes, perfectly right! We have to finalize between `old-filename -> new-filename` and `model_path -> new_filename) `. Both have pros and cons, it's your call now. With more or less work, in both cases we can work out easy error message for user to setup mapping. So I am fine with either.

I definitely want to store `pretrained_file` and `vocab_namespace` (it's in previous pr) in `embedding` object. If you dislike like it, we can mandate the user to give file mapping and also namespace mapping each and every time. But I don't think it's worth the trade-off at all.

The two attributes don't fit well to general `Embedding` abstraction. I would ideally like to have sub-class refactoring if it's not breaking, but you already said it's not straightforward. If we subclass `PretrainedEmbedding`, it would be for `pretrained_file` only; I believe, `vocab_namespace`  still belongs more to `Embedding` than `PretrainedEmbedding` since randomly initialized embedding can be associated to `namespace` also.

HarshTrivedi:
Btw, if we commit to store `pretrained_file` in `Embedding` class itself in this release, it would be even more difficult to subclass `PretrainedEmbedding` from `Embedding` in future releases.

matt-gardner:
The `Embedding` object _does_ have a notion of a pretrained file and a vocabulary namespace, it's just that they are used in `from_params`, before you actually get to the constructor.  If we rework things to remove the custom `from_params` logic somehow (c.f. #2419), then whatever comes out of that _will_ have an explicit "pretrained file" parameter somewhere.  I'd recommend just doing the simple thing and storing the parameter, and worrying about a cleaner way to accomplish this (if necessary) after #2419 is resolved somehow.

joelgrus:
I've been asking around, to see if anyone has strong opinions. Seems like there's a mild preference for model_path, with the idea that the `fine-tune` command will tell you what model paths need to be specified if they're not.

HarshTrivedi:
Sounds good. I'll ping you again once it's ready for review.

HarshTrivedi:
@joelgrus, @matt-gardner, this is up for another look - turned out to be cleaner that I had expected. I have added the tests as well.

Besides the changes discussed above, I have removed some of the methods introduced in #2374. Firstly, they were unnecessary and secondly that way of navigating to embedding module allows only embedding modules in text_field_embedder or root be extended. The embedding module could be located in arbitrary `module_1` or nested inside `module_1.`module_2..`. The update fixes this as well.

joelgrus:
I am looking at this, but I'm at AAAI this week, so I'll be a little bit slow with comments

allennlp/modules/token_embedders/embedding.py
```diff
@@ -88,7 +89,8 @@ def __init__(self,
                  norm_type: float = 2.,
                  scale_grad_by_freq: bool = False,
                  sparse: bool = False,
-                 vocab_namespace: str = None) -> None:
+                 vocab_namespace: str = None,
+                 pretrained_file: str = None) -> None:
```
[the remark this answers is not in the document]
joelgrus: also, are you ever setting this value anywhere?

allennlp/modules/token_embedders/embedding.py
```diff
@@ -180,9 +188,33 @@ def extend_vocab(self,  # pylint: disable=arguments-differ
             vocab_namespace = "tokens"
             logging.warning("No vocab_namespace provided to Embedder.extend_vocab. Defaulting to 'tokens'.")
 
+        extended_num_embeddings = extended_vocab.get_vocab_size(vocab_namespace)
+        if extended_num_embeddings <= self.num_embeddings:
+            # It's already been extended. No need to initialize / read pretrained file in first place (no-op)
+            return
+
+        if self._pretrained_file and os.path.exists(self._pretrained_file):
```
joelgrus: is this the right check? couldn't `self._pretrained_file` be like a path to S3? in which case it wouldn't `os.path.exist`?

allennlp/modules/token_embedders/embedding.py
```diff
@@ -180,9 +188,33 @@ def extend_vocab(self,  # pylint: disable=arguments-differ
...
+            # pretrained_file attribute was saved during training and is available:
+            if pretrained_file:
+                # But user also passed pretrained_file
+                model_path_info = "at model_path, {model_path} " if model_path else ""
+                logging.warning(f"You passed pretrained_file, {pretrained_file} for "
+                                f"embedding {model_path_info}. But it's ignored because "
+                                f"original pretrained_file, {self._pretrained_file} is available.")
+            pretrained_file = self._pretrained_file
+
+        if not (pretrained_file and os.path.exists(pretrained_file)):
```
joelgrus: I find checks written like this very hard to read / parse, I would prefer

```
if not pretrained_file or not os.path.exists(pretrained_file):
```

but as above I'm not sure the latter is the right check

allennlp/modules/token_embedders/embedding.py
```diff
@@ -180,9 +188,33 @@ def extend_vocab(self,  # pylint: disable=arguments-differ
...
+            extra_info = (f"Originally pretrained_file was at "
+                          f"{self._pretrained_file}. " if self._pretrained_file else "")
+            # It's better to warn here and not give error because there is no way to distinguish between
+            # whether pretrained-file wasn't used during training or user forgot to pass / passed incorrect
+            # mapping. Giving error would prevent former case to executable.
```
joelgrus: "Raising an error would prevent fine-tuning in the former case"?

allennlp/tests/commands/fine_tune_test.py
```diff
@@ -55,6 +57,50 @@ def test_fine_tune_works_with_vocab_expansion(self):
...
+    def test_fine_tune_works_with_vocab_expansion_with_pretrained_file(self):
+        params = Params.from_file(self.config_file)
+        # snli2 has a new token (seahorse) in it
+        params["train_data_path"] = str(self.FIXTURES_ROOT / 'data' / 'snli2.jsonl')
+
+        embeddings_filename = str(self.TEST_DIR / "embeddings2.gz")
+        extra_token_vector = np.random.rand(300)
+        with gzip.open(embeddings_filename, 'wb') as embeddings_file:
+            extra_token_vector_str = " ".join(map(str, list(extra_token_vector)))
+            embeddings_file.write("seahorse {}\n".format(extra_token_vector_str).encode('utf-8'))
```
joelgrus: why not just make this a test fixture rather than writing it out every time

joelgrus: also, I don't know if we have official guidance on this, but prefer f-strings to str.format

allennlp/modules/token_embedders/embedding.py
HarshTrivedi: We can only raise error if pretrained file wasn't available when we know *for sure* that it was used at time of training. If pre-trained wasn't used in first place, the extension should ideally happen with random vectors and we don't need the file.

For models after this PR, we can just check `_pretrained_file` attribute to know whether pretrained file was used or not and decide whether it's essential or not. However, earlier we didn't store the `_pretrained_file` attribute, so effectively there's no way of really knowing whether pretrained file was used or not. 

If we raise error, user won't be able to extend if pre-trained file is in fact not required.

HarshTrivedi: Oops, I think you just wanted me to change the comment to `Raising an error would prevent fine-tuning in the former case`, right? I thought, you want me to explain case!

allennlp/modules/token_embedders/embedding.py
```diff
@@ -180,9 +188,33 @@ def extend_vocab(self,  # pylint: disable=arguments-differ
...
+
+        if self._pretrained_file and os.path.exists(self._pretrained_file):
+            # pretrained_file attribute was saved during training and is available:
+            if pretrained_file:
+                # But user also passed pretrained_file
+                model_path_info = "at model_path, {model_path} " if model_path else ""
+                logging.warning(f"You passed pretrained_file, {pretrained_file} for "
+                                f"embedding {model_path_info}. But it's ignored because "
+                                f"original pretrained_file, {self._pretrained_file} is available.")
+            pretrained_file = self._pretrained_file
```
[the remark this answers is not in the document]
HarshTrivedi: Yeah right, I was hoping between the two. In the case that `pretrained_file` attribute was stored and it's yet available, we are 100% sure it's the one that should be used. I assumed user can make a mistake in setting up model path to embedding file mapping, but pre-stored file that's yet available can't be mistaken. 

But I agree, what you suggest would be the expected behaviour from user end, so am fine changing it back. Let me know.

allennlp/modules/token_embedders/embedding.py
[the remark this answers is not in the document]
joelgrus: sorry, I meant are any callers of this constructor setting this file? like, is `Embedder.from_params` setting it? it looks like it's not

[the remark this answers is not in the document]
HarshTrivedi: The problem is both are actually `original_pretrained_file`, one is passed and one is (possibly) pre-stored attribute. But, I can distinguish them as `self._initialization_pretrained_file` and `extension_pretrained_file`, where `self._initialization_pretrained_file` will be set at initialization (likely training) and `extension_pretrained_file` will be the argument in `extend_vocab` method. 

If `extension_pretrained_file` is passed and available, it will be used or else we will default to self._initialization_pretrained_file (also look user/stored priority comment).  Sounds good?

allennlp/models/model.py
```diff
@@ -321,18 +320,41 @@ def load(cls,
...
+        fine-tuning/tranfer-learning, it will have it work with extended vocabulary
+        (original + new-data vocabulary).
 
+        Parameters
+        ----------
+        extended_vocab : Vocabulary:
+            Vocabulary extended from original vocabulary used to construct
+            this ``Embedding``.
+        pretrained_file : str, (optional, default=None)
+            A file containing pretrained embeddings can be specified here. It can be
```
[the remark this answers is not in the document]
HarshTrivedi: Oops, missed to remove it!

allennlp/commands/fine_tune.py
```diff
@@ -74,6 +75,13 @@ def add_subparser(self, name: str, parser: argparse._SubParsersAction) -> argpar
                                default="",
                                help='If non-empty, name of metric used to weight the loss on a per-batch basis.')
 
+        subparser.add_argument('--embedding-sources-mapping',
+                               type=str,
+                               default="",
+                               help='a JSON dict to define mapping from embedding file paths used during training '
+                               'to current embedding file paths. If not passed, and embedding needs to be '
+                               'extended, we will use the original file paths uded during training.')
```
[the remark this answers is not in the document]
HarshTrivedi: Oops, missed to update docstring! Besides typo, the whole help message is incorrect. It should be model_path to file_path and not file_path to file_path mapping. Will fix it.

allennlp/modules/token_embedders/embedding.py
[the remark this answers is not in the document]
HarshTrivedi: No, no caller of the constructor is setting the pretrained file except in the constructor itself.

allennlp/modules/token_embedders/embedding.py
joelgrus: I don't think we explicitly worry about URLs not existing anywhere else, so I'd rather not complicate the code here either. Just assume it's OK if it's a URL, and make sure it exists if it's a file.

HarshTrivedi: Sounds good!

allennlp/commands/fine_tune.py
```diff
@@ -118,25 +128,31 @@ def fine_tune_model_from_file_paths(model_archive_path: str,
...
+    embedding_sources_mapping: ``str``, optional (default="")
+        JSON string to define dict mapping from embedding paths used during training to
+        the corresponding embedding filepaths available during fine-tuning.
     """
     # We don't need to pass in `cuda_device` here, because the trainer will call `model.cuda()` if
     # necessary.
     archive = load_archive(model_archive_path)
     params = Params.from_file(config_file, overrides)
+    embedding_sources_mapping = embedding_sources_mapping or '{}'
+    embedding_sources_dict: Dict[str, str] = json.loads(embedding_sources_mapping)
```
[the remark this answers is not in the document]
HarshTrivedi: done. 👍

allennlp/commands/fine_tune.py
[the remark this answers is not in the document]
HarshTrivedi: done. 👍

allennlp/modules/token_embedders/embedding.py
HarshTrivedi: done. 👍

allennlp/modules/token_embedders/embedding.py
HarshTrivedi: done. 👍

allennlp/tests/commands/fine_tune_test.py
HarshTrivedi: done. 👍

allennlp/modules/token_embedders/embedding.py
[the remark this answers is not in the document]
joelgrus: I think if you override it should take priority

allennlp/modules/token_embedders/embedding.py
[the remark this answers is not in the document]
joelgrus: I am fine to leave this one `pretrained_file` and have the one in extend called `extension_file` or whatever. Mostly I just don't want there to be a `pretrained_file` and a `self._pretrained_file` that refer to different things, that's super confusing and non-obvious.

HarshTrivedi:
@joelgrus  this is up for another look.

joelgrus:
let me pull this down and make some changes to it myself, I think that will be faster than the back and forth

joelgrus:
ok, I made three changes:

1. I renamed `is_file_available` and allowed it to accept None (which made logic simpler elsewhere)
2. I flattened out the logic in extend_vocab ("flat is better than nested")
3. I changed the check in `Model.extend_vocab` from `isinstance(module, Embedding)` to `hasattr(module, 'extend_vocab')`, which is more extensible in the future.

@HarshTrivedi take a look and let me know if you have any objections

HarshTrivedi:
@joelgrus  Thanks! Looks good to me! 👍

PR #2396, opened 2019-01-19
Add QANet model for SQuAD.

### LINKED ISSUE #2185: Reduce memory use of multi-head self-attention
Since self-attention complexity is O(timestep * timestep * heads), the attention_dropout and mask_softmax  layer consumes lots of memory in a model where self_attention is heavily used. For example, QANet uses 3*7=21 self attention layer in its Model Encoder Layer. Is it possible to make it more memory efficient?

I'm not quite familiar with recycle in pytorch,  but I'm wondering is there any chance to solve this?

yizhongw:
* Implement the QANet model.

* Implement a new Trainer that maintains exponential moving average for parameters.

* More memory effient implementation of `masked_softmax` and `multi_head_self_attention`.

* Fixes #2185


DeNeutoy:
Awesome 🙌

matt-gardner:
This is awesome, thanks!  But, I can get you to split it out into three separate PRs?  Specifically:

1. The new trainer, which probably needs some updating to match @joelgrus's recent refactor, and maybe could just be an addition to the existing trainer instead of a new one.
2. The new encoders / other modules.
3. The model itself.

The first two PRs are independent and can be reviewed separately, and the third one depends on the first two.  It will be a _whole_ lot easier to review this if it's split up into smaller chunks.

joelgrus:
+1 to putting the EMA in  the existing trainer, let me think some about the best way to do that cleanly and extensibly

joelgrus:
that sounds good, and feel free to submit early (say, before you've written tests) just so we can iterate and make sure the design seems right

PR #2399, opened 2019-01-19
Fix docstring for PretrainedBertIndexer's pretrained_model param

bryant1410:
`pretrained_model` is not an optional param, as stated in `PretrainedBertIndexer` docs, so I removed the "optional" part from it.


nelson-liu:
Thanks!

PR #2406, opened 2019-01-21
Support exponential moving average in the default trainer.

yizhongw:


allennlp/training/exponential_moving_average.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class ExponentialMovingAverage:
+    """
+    Create shadow variables and maintain exponential moving average for model parameters.
+
+    Parameters
+    ----------
+    model: ``Model``, required.
+        An AllenNLP model to be optimized.
```
joelgrus: I am not convinced about this API. Why not just pass in the named parameters in the constructor (instead of the model) and then you can get rid of that option in the method calls.

allennlp/training/exponential_moving_average.py
```diff
@@ -0,0 +1,64 @@
...
+    def __init__(self, model: Model, decay: float = 0.9999) -> None:
+        self.decay = decay
+        self._average_values: Dict = {}
+        self._backup_values: Dict = {}
+        self._model = model
+        for name, param in model.named_parameters():
+            self._average_values[name] = param.data.clone()
+            self._backup_values[name] = param.data.clone()
+
+    def apply(self, num_updates: int = None, named_parameters: Iterable = None) -> None:
```
joelgrus: I know this is what Tensorflow calls it, but I can't say that I like the name `apply` here. If it were me I would call it something like `update_averages` or something

allennlp/training/exponential_moving_average.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        Apply exponential moving average to `named_parameters` if specified,
+        or we will apply this to all the trainable parameters of the model.
+
+        The optional `num_updates` parameter allows one to tweak the decay rate
+        dynamically. If passed, the actual decay rate used is:
+
+            `min(decay, (1 + num_updates) / (10 + num_updates))`
```
joelgrus: where does the second equation here come from?

yizhongw:
@matt-gardner Could you give me any hints on why the checks failed? The documentation looks good to me, but it still reported `undocumented module: allennlp.training.exponential_moving_average`.

allennlp/training/exponential_moving_average.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        if num_updates is not None:
+            decay = min(self.decay, (1.0 + num_updates) / (10.0 + num_updates))
+        else:
+            decay = self.decay
+        if named_parameters is None:
+            named_parameters = self._model.named_parameters()
+        for name, param in named_parameters:
+            new_average_value = (1.0 - decay) * param.data + decay * self._average_values[name]
+            self._average_values[name] = new_average_value.clone()
```
joelgrus: why do you need `clone` here? I would have written something like

```
self._average_values[name] = (1.0 - decay) * param.data.detach() + decay * self._average_values[name]
```
is there a reason why two steps is better ?

allennlp/training/exponential_moving_average.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def assign_average_value(self, named_parameters=None) -> None:
+        """
+        Assign the exponential moving average value to the parameters
+        """
+        if named_parameters is None:
+            named_parameters = self._model.named_parameters()
+        for name, param in named_parameters:
+            self._backup_values[name] = param.data.clone()
```
joelgrus: why do you need clone here?

allennlp/training/exponential_moving_average.py
```diff
@@ -0,0 +1,64 @@
...
+            param.data = self._average_values[name]
+
+    def restore(self, named_parameters=None) -> None:
+        """
+        Restore the original values of each parameter
+        """
+        if named_parameters is None:
+            named_parameters = self._model.named_parameters()
+        for name, param in named_parameters:
+            param.data = self._backup_values[name].clone()
```
joelgrus: and here?

allennlp/training/exponential_moving_average.py
```diff
@@ -0,0 +1,64 @@
+from typing import Dict, Iterable
+from allennlp.models.model import Model
+
+
+class ExponentialMovingAverage:
```
joelgrus: are there other kinds of moving averages we might care about?

should there be a `MovingAverage` base class and then the trainer can be agnostic to what type of moving average it's using, rather than hardcoding this one?

allennlp/training/exponential_moving_average.py
yizhongw: You can refer to [this tensorflow implementation](https://github.com/tensorflow/tensorflow/blob/a6d8ffae097d0132989ae4688d224121ec6d8f35/tensorflow/python/training/moving_averages.py#L331).

allennlp/training/exponential_moving_average.py
yizhongw: Yes, this is not necessary. I will change that.

allennlp/training/exponential_moving_average.py
yizhongw: `clone()` might be necessary here, to make sure the update of parameters will not affect the correspondings in `_backup_values`?

allennlp/training/exponential_moving_average.py
yizhongw: Same as above. Using `clone()` is safer?

allennlp/training/exponential_moving_average.py
yizhongw: Tensorflow as a [`moving_averages.py`](https://github.com/tensorflow/tensorflow/blob/r1.12/tensorflow/python/training/moving_averages.py) file. But, it seems that they only implemented the exponential moving average.

matt-gardner:
@yizhongw, to answer your question about the docs: you need to add a reference to the module in our API docs, [here](https://github.com/allenai/allennlp/blob/master/doc/api/allennlp.training.rst).  Ping me on slack if you need help figuring out how this works.

joelgrus:
@yizhongw do you want me to take a crack at the abstract `MovingAverage` version?

yizhongw:
@joelgrus Yes, sure! Then, I could modify based on that.

joelgrus:
@yizhongw ok, take a look at what I pushed up, I'm not 100% sure I got all the clone()s and datas right, but the tests pass.

joelgrus:
since I now wrote a lot of this code, probably someone else should review it

allennlp/training/moving_average.py
```diff
@@ -0,0 +1,91 @@
...
+         https://www.tensorflow.org/api_docs/python/tf/train/ExponentialMovingAverage)
+        """
+        if num_updates is not None:
+            decay = min(self._decay,
+                        (self._numerator + num_updates) / (self._denominator + num_updates))
+        else:
+            decay = self._decay
+
+        for name, parameter in self._parameters:
+            self._shadows[name] = decay * self._shadows[name] + (1.0 - decay) * parameter.clone()
```
DeNeutoy: You can do this inplace:

```
self._shadows[name].mul_(decay).add_((1 - decay) * parameter.data)
```
This is actually important I think, because copying large parameters every step is expensive. E.g all the [optimizers use inplace operations](https://github.com/pytorch/pytorch/blob/master/torch/optim/adam.py).

allennlp/training/trainer.py
```diff
@@ -513,6 +535,11 @@ def _save_checkpoint(self, epoch: Union[int, str]) -> None:
             The epoch of training.  If the checkpoint is saved in the middle
             of an epoch, the parameter is a string with the epoch and timestamp.
         """
+        # If moving averagew are used for parameters, we save
```
DeNeutoy: averages

DeNeutoy:
Oh, another idea:

it's annoying to have to keep these EMAs on the same device as the parameters, because you're essentially tripling the memory requirement of the model. What if you added a `accumulate_on_cpu` option, where you computed the update equation on the GPU, but then did `self._shadow[name].copy_(update_value, non_blocking=True)` - this way, we wouldn't have to accumulate on the GPU, but I (don't think) we would have to wait for the GPU -> CPU copies which this method would require, because they can be non-blocking. What do you think?

joelgrus:
sounds like a good idea, let me make an attempt at it

PR #2412, opened 2019-01-21
serialization in the tutorial

### LINKED ISSUE #2379: intro tutorial should include serialization + config files
people using the intro tutorial are confused (rightly so) about how to serialize the model and re-use it later. we could 

(1) add a section at the end about saving it explicitly, and then
(2) point people to the config file section of the readme, and also
(3) caveat the tutorial with "this isn't how you'd do things in practice" (which is in the README.md tutorial but got lost in translation)

see e.g. #2377

joelgrus:
fixes #2379 

1. adds a feature to the `convert.py` script that allows for a final HTML block at the bottom of the page
2. shows how to manually save weights+vocab and manually load them back
3. adds a caveat that this isn't how to do things and links to the part of the README that covers configuration files.


tutorials/tagger/basic_allennlp.py
```diff
@@ -205,3 +205,45 @@ def get_metrics(self, reset: bool = False) -> Dict[str, float]:
 tag_ids = np.argmax(tag_logits, axis=-1)
 #### And then use our vocabulary to find the predicted tags.
 print([model.vocab.get_token_from_index(i, 'labels') for i in tag_ids])
+
+#### Finally, we'd like to be able to save our model and reload it later. We'll need to save two things. The first is the model weights.
```
matt-gardner: Minor thought, not for this PR: is it possible to do something like combine contiguous blocks of lines beginning with `####` in the `convert.py` script, so you don't have these super long lines?  It'd make reading and reviewing these files a lot easier.

tutorials/tagger/basic_allennlp.py
```diff
@@ -205,3 +205,45 @@ def get_metrics(self, reset: bool = False) -> Dict[str, float]:
...
+# Here's how to save the model.
+with open("/tmp/model.th", 'wb') as f:
+    torch.save(model.state_dict(), f)
+#### And the second is the vocabulary.
+vocab.save_to_files("/tmp/vocabulary")
+
+#### We only saved the model weights, so we actually have to recreate the model in code. First, let's reload the vocabulary into a new variable.
```
matt-gardner: Maybe something like "so we have to recreate the same model structure ourselves in code when we want to load them."?

tutorials/tagger/basic_allennlp.py
```diff
@@ -205,3 +205,45 @@ def get_metrics(self, reset: bool = False) -> Dict[str, float]:
...
+# And here's how to reload the model.
+vocab2 = Vocabulary.from_files("/tmp/vocabulary")
+#### And then let's recreate the model (if I were doing this in a different file I would of course have to re-instantiate the word embeddings and lstm as well).
```
matt-gardner: You switched here from "we" to "I".

tutorials/tagger/default-tutorial.html
```diff
@@ -186,7 +186,7 @@ <h2>The Problem</h2>
 {% highlight python %}
     def forward(self,
                 sentence: Dict[str, torch.Tensor],
-                labels: torch.Tensor = None) -> torch.Tensor:
+                labels: torch.Tensor = None) -> Dict[str, torch.Tensor]:
```
matt-gardner: Oh, I didn't realize there was an html file in here - doesn't this just duplicate what we have in the website repo?

joelgrus: yeah, I literally just copy this file into the website repo, maybe it shouldn't be checked in here too? I don't know that it hurts anything

matt-gardner: A potential issue: manual modifications (like I did to the version on the website to fix this line) could get overwritten.  I guess that's a problem no matter what repo the file is in, though...  But it does seem more error-prone to have two version of the same file in different repos.

joelgrus: that file is the output of convert.py, which means that any changes will get overwritten next time the script is updated, let me add a comment to that effect

Issue #2373, opened 2019-01-16
Fix LanguageModel API to easily handle bidirectional contextualizers.


brendan-ai2:
**Is your feature request related to a problem? Please describe.**
Currently `LanguageModel` takes a single contextualizer. All our normal `Seq2SeqEncoder`s work here in the unidirectional case. In the bidirectional case, however, our normal `Seq2SeqEncoder`s are broken when they contain more than one layer. See https://github.com/allenai/allennlp/blob/master/allennlp/models/language_model.py#L61. Special handling is needed like that contained in the `BidirectionalLanguageModelTransformer`. See https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2seq_encoders/bidirectional_language_model_transformer.py#L262. This is an issue because a model can be silently broken. Further, a user can't simply specify `lstm`, `gru`, etc. in their config and get a working LM in general.

**Describe the solution you'd like**
Let's provide a wrapper `Seq2SeqEncoder`, maybe a `BidirectionalLanguageModelObjectiveEncoder`, that handles the masking and direction changes transparently. It would accept an underlying `Seq2SeqEncoder` that would be duplicated to prevent parameter sharing or perhaps two entirely separate encoders.

**Describe alternatives you've considered**
In the short term we could at least check that users aren't using our default encoders with `bidirectional=True` and `num_layers>1`. (Annoyingly we'd need a class blacklist as opposed to checking that condition directly as the transformer encoder does the correct thing despite satisfying the condition.)

**Additional context**
LSTM Encoder from Calypso: https://github.com/allenai/calypso/blob/master/calypso/lstm_encoder.py#L22

Note that it's hardwired to be bidirectional and have only one layer.

brendan-ai2:
fyi @nelson-liu

matt-gardner:
Quick note: you'd definitely need two `Seq2SeqEncoders`, otherwise you'd have a single encoder with backward and forward parameters tied, which is almost certainly not what you want.

brendan-ai2:
@matt-gardner, sorry I was unclear. What I meant was two _types_ of encoders vs one type of encoder. We very definitely need to instantiate two different instances. :)

brendan-ai2:
Updated description: "It would accept an underlying `Seq2SeqEncoder` that would be duplicated to prevent parameter sharing or perhaps two entirely separate encoders."

nelson-liu:
[...]

PR #2414, opened 2019-01-22
[WIP] Language Modeling of Contiguous Text

nelson-liu:
This PR enables language models with stateful encoders to properly work on contiguous text datasets (e.g., BookCorpus).

To Do:
- [x] Make `LanguageModelingReader` actually usable
- [x] Test `LanguageModelingReader`
- [x] Create special `LanguageModelingIterator` (or some other name) that turns off shuffling and also constrains the batch size to 1.
- [x] Modify `LanguageModel` to handle instances from `LanguageModelingReader`
- [x] Add bidirectionality to `LanguageModelingReader` and `LanguageModel`
- [x] Test `LanguageModel`


nelson-liu:
@matt-gardner / @brendan-ai2 : one annoying thing is that tensorizing the instances from the dataset reader gives you an extra singleton batch dimension --- is it reasonable to just remove that in the iterator (i think you can, but not positive)?

matt-gardner:
I would say yes, it's reasonable to remove that extra dimension in the iterator, which is another strong motivation for having a language-modeling-specific iterator.

brendan-ai2:
Please ping me when you're ready for a review on this.

nelson-liu:
Thanks @brendan-ai2 , will do! Still needs a bit of cleanup.

nelson-liu:
Aside: I'm pretty sure the bidirectional contiguous text setting won't be correct until https://github.com/allenai/allennlp/issues/2373 is fixed.

matt-gardner:
The whole point of your dataset reader is to get contiguous chunks that let you see one batch after another of streaming text, in order.  But it's in order only for one direction.  A stateful backwards decoder would do the wrong thing (even with #2373 fixed) if it sees forward-streaming batches.

nelson-liu:
ah, i see what you mean now. i guess this could work if you have separate inputs for your forward and backward (i.e., forward starts from the first index, and backward from the last)?

matt-gardner:
That would still break a bidirectional encoder (because the backward LSTM on the forward input would be wrong, and the forward LSTM on the backward input would be wrong).  This method of training is only suitable for a single direction at a time.

nelson-liu:
talked with @matt-gardner on slack, seems like the reasonable fix would be to have the model take a 2 inputs (for each direction), 2 targets (for each direction). Furthermore, the model would use a shared embedding layer for the input and two different contextualizers (one forward-direction, one backward-direction).

matt-peters:
This is the way the original ELMo training code works.  There are two separate LSTMs (one for each direction), but shared softmax layer and word embeddings.  The data generator creates batches in exactly this manner, by returning both forward and reverse ids.

brendan-ai2:
@nelson-liu generously took the time to explain some of the delicate points of contiguous language modeling to me. Seems like we're all in agreement on having multiple contextualizers and targets.

allennlp/data/dataset_readers/language_modeling.py
```diff
@@ -1,97 +1,160 @@
...
+            logger.info('Loading data from %s', file_path)
+            # if `file_path` is a URL, redirect to the cache
+            file_path = cached_path(file_path)
 
-        with open(file_path, "r") as text_file:
-            instance_strings = text_file.readlines()
+            # Read the contents of the file into one long list of tokens,
+            # adding start and/or end tokens as necessary.
+            file_tokens = []
+            file_tokens.extend(self._start_tokens)
```
mttk: Shouldn't this be in the forloop?

nelson-liu: Yup, you're correct (sorry I haven't looked at this PR in awhile). In general, though, language modeling of contiguous text doesn't use any start tokens and only uses eos tokens (e.g., see https://github.com/salesforce/awd-lstm-lm/blob/master/data.py#L34-L54 )

mttk: yeah, although AFAIK ELMo actually used start tokens (https://github.com/allenai/allennlp/blob/master/tutorials/how_to/elmo.md#notes-on-statefulness-and-non-determinism)

nelson-liu: Right, but this dataset reader is not for training a LM like ELMo. ELMo is trained on shuffled sentences, hence why it needs a start token for each sentence. This dataset reader is for _contiguous text_ (e.g., books), like the corpora that the OpenAI GPT was trained on / the PTB LM benchmark most folks use. In this setting, people don't typically use start tokens.

mttk: Ah, got it, thanks -- I assumed that stateful (RNN) language models were _always_ trained in contiguous fashion.
The shuffling bit might help generalization performance for sentence-level tasks (where the model doesn't learn to rely on an initialized hidden state), but I'm wondering how the same model with and without contiguous training would perform in transfer learning. 
Thanks!

bratao:
@nelson-liu This is a AWESOME feature and needed in many scenarios I'm working with.
Did you have plan to merging this PR?

matt-gardner:
@nelson-liu, what's stopping this from getting merged?

nelson-liu:
So we first have to figure out https://github.com/allenai/allennlp/issues/2373 , which PR https://github.com/allenai/allennlp/pull/2438 attempts to do. #2438 is blocked by https://github.com/allenai/allennlp/pull/2438#discussion_r251223277

I also think that @rloganiv is working on contiguous-text language modeling, so we should try to avoid duplication of effort.

allennlp/data/dataset_readers/language_modeling.py
```diff
@@ -1,97 +1,160 @@
...
+                 truncated_bptt_size: int = 35,
+                 fuzz_truncated_bptt_size: bool = True,
+                 bidirectional: bool = False,
                  tokenizer: Tokenizer = None,
                  token_indexers: Dict[str, TokenIndexer] = None,
-                 lazy: bool = False) -> None:
-        super().__init__(lazy)
+                 start_tokens: List[str] = None,
+                 end_tokens: List[str] = ["</S>"]) -> None:
+        super().__init__(lazy=False)
```
rloganiv: I think setting `lazy=False` defeats the purpose of using `fuzz_truncated_bppt_size=True` during training. The point of using random sequence lengths is to prevent batches ending with same tokens each epoch, however this will not happen if `lazy=False` since `_read` will only be called once. It probably makes more sense to let users set `lazy` themselves and raise an error if `fuzz_truncated_bppt_size=True` and `lazy=False`.

allennlp/data/iterators/language_modeling_iterator.py
```diff
@@ -0,0 +1,79 @@
...
+from allennlp.data.instance import Instance
+from allennlp.data.iterators.basic_iterator import BasicIterator
+from allennlp.data.iterators.data_iterator import DataIterator, TensorDict
+from allennlp.data.dataset import Batch
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+@DataIterator.register("language_modeling")
+class LanguageModelingIterator(BasicIterator):
```
rloganiv: I think it makes sense to give this object a more general name - it is useful for any problem where batching is done in the `DatasetReader` (e.g. one's with the issue I raise in #2828). Maybe something like `StraightThroughIterator` or `TrivialIterator`?

bratao:
@nelson-liu sorry to ping you about this again. But I'm planning for an upcoming research that will need LM of Contiguous Text. Do you plan to resume this?

nelson-liu:
yes, i would like to have this working at some point. I know @rloganiv  was also working on this at some point, and @matt-gardner recently merged some LM things. I'm traveling until mid-september, but I'm hoping to try to make some progress on it afterwards.

nelson-liu:
>< sorry for dropping the ball on this. yes, i would like to have this working at some point. I know @rloganiv  was also working on this at some point, and @matt-gardner recently merged some LM things. I'm traveling until mid-september, but I'm hoping to try to make some progress on it afterwards.

rloganiv:
@bratao This PR is currently blocked due to the thorny issue of bi-directionality (see #2438). If you only need a forward generative LM then you should be able to use the existing [`LanguageModel`](https://github.com/allenai/allennlp/blob/master/allennlp/models/language_model.py) implementation, with the following caveats:
1. You will need to set `stateful=True` for the contextualizer
2. Your dataset reader will need to handle batching (you could use the [`LanguageModelingReader`]( https://github.com/nelson-liu/allennlp/blob/contiguous_text_lm/allennlp/data/dataset_readers/language_modeling.py) in this PR), and you will need to use the [`PassThroughIterator`](https://github.com/allenai/allennlp/blob/master/allennlp/data/iterators/pass_through_iterator.py) as your iterator.
Hope this helps!

DeNeutoy:
This PR has been open since Jan, i'm declaring it done - re-open if someone wants to fix it.

PR #2418, opened 2019-01-22
Get rid of hardcoded Vocabulary class name.

mahnerak:
In case of inheriting Vocabulary, the method `from_params` calls the same method of the registered class if `type` is given. If `from_params` is not overridden (even if it is), the default implementation calls hardcoded `Vocabulary.from_instances` and `Vocabulary.from_files` and the subclass implementations of these methods are not called.
With this commit the `Vocabulary` is replaced by `cls` allowing to customize behaviour of the subclass.

### DIFF
--- a/allennlp/data/vocabulary.py
+++ b/allennlp/data/vocabulary.py
@@ -460,7 +460,7 @@ class Vocabulary(Registrable):
                 logger.info("Loading Vocab from files instead of dataset.")
 
         if vocabulary_directory:
-            vocab = Vocabulary.from_files(vocabulary_directory)
+            vocab = cls.from_files(vocabulary_directory)
             if not extend:
                 params.assert_empty("Vocabulary - from files")
                 return vocab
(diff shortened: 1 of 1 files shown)


joelgrus:
that code should never run if you're using `from_params` on a subclass?

https://github.com/allenai/allennlp/pull/2418/files#diff-6d4c262b560c918424183ec540cd9cfdL444

or is the issue that the subclass might want to call `super().from_params(...)`?

mahnerak:
The both cases:
1. `Subclass` does not have `from_params`. Then object is instantiated in either `from_files` or `from_instances`. The methods are called on `Vocabulary`, so object of `Vocabulary` (not `Subclass`) is goint to be created.
2. `Subclass` do have `from_params` but it calls `super().from_params()`. If we have shared logic of creating a object from params, we need to duplicate the code of `Vocabulary.from_params()`.

As in (2), this can be addressed on Subclass's side by duplicating code of `Vocabulary.from_params()`, but is not a preferred way.
Also, it makes no sense calling these methods on `Vocabulary` and not on `cls`.

mahnerak:
More specificially, in my case, I want to have custom Vocabulary which is differs from the original class only in implementation of method `from_instances()`.

joelgrus:
got it, thanks

mahnerak:
Big Thanks!

PR #2424, opened 2019-01-23
Propose a deprecation policy.

brendan-ai2:
- Examples of when backward compatibility is likely to be an issue.
- Suggestions for ensuring compatibility.
- Mechanics for deprecating.

### DIFF
--- a/DEPRECATION.md
+++ b/DEPRECATION.md
@@ -22,7 +22,7 @@ Say you've found some old code that needs some care. You have a fix in mind, may
 
 1. Determine whether backward compatibility is even an issue.
 
-    __Likely safe__
+    ####Likely safe
 
   * New code
   * Internal code _TODO: How is this determined?_
--- a/DEPRECATION.md
+++ b/DEPRECATION.md
@@ -22,7 +22,7 @@ Say you've found some old code that needs some care. You have a fix in mind, may
 
 1. Determine whether backward compatibility is even an issue.
 
-  __Likely safe__
+  Why does this not work? __Likely safe__
 
   * New code
   * Internal code _TODO: How is this determined?_
--- a/DEPRECATION.md
+++ b/DEPRECATION.md
@@ -23,11 +23,13 @@ Say you've found some old code that needs some care. You have a fix in mind, may
 1. Determine whether backward compatibility is even an issue.
 
   __Likely safe__
+
   * New code
   * Internal code _TODO: How is this determined?_
   * Implementation change only
 
   __Care required__
+
   * File/class/function names
     * Including names passed to `Registrable.register`!
   * Function signatures
@@ -37,15 +39,18 @@ Say you've found some old code that needs some care. You have a fix in mind, may
 2. If so, first consider how critical the change is.
 
   __Non-critical__
+
   * Minor name changes
   * "Cleaner" APIs with no functional benefit
 
   __Medium__
+
   * Name/API that is actively confusing multiple users.
     * This means Github issues, messages on user channels, etc.
   * Useful new feature
 
   __High__
+
   * Major bugs
 
 3. Non-critical changes we should happily avoid. Otherwise, let's try to make the change in a backward compatible manner. Options include:
--- a/DEPRECATION.md
+++ b/DEPRECATION.md
@@ -1,0 +1,85 @@
+# Deprecation Policy Proposal
+
+As AllenNLP matures we will inevitably need to deprecate old models, interfaces and features. This document provides guidelines that attempt to minimize churn. Reducing churn helps users build next generation models with tools they've already mastered.
+
+### Terminology note
+
+We _deprecate_ some component to communicate that a preferred alternative exists. A _deprecated_ component does not necessarily have to be _removed_ from the library.
+
+### Goals
+
+1. Promote backward compatibility where not onerous.
+2. Describe code-level and release process for both deprecation and removal.
+
+### Non-goals
+
+1. Eliminate all deprecation and removal.
+2. Legislate. Good judgment by AllenNLP developers is much preferred.
+
+## Decision Flow
+
+Say you've found some old code that needs some care. You have a fix in mind, maybe already a PR. Great, let's get it merged! But let's give some thought to backward compatibility along the way.
+
+1. Determine whether backward compatibility is even an issue.
+
+### Likely safe
+* New code
+* Internal code _TODO: How is this determined?_
+* Implementation change only
+
+### Care required
+* File/class/function names
+  * Including names passed to `Registrable.register`!
+* Function signatures
+* Jsonnet config
+* Model internals that affect shape of saved weights
+
+2. If so, first consider how critical the change is.
+
+### Non-critical
+* Minor name changes
+* "Cleaner" APIs with no functional benefit
+
+### Medium
+* Name/API that is actively confusing multiple users.
+  * This means Github issues, messages on user channels, etc.
+* Useful new feature
+
+### High
+* Major bugs
+
+3. Non-critical changes we should happily avoid. Otherwise, let's try to make the change in a backward compatible manner. Options include:
+
+* When deprecating, say, a class, leave a shim to its replacement.
+* A function signature can likely be changed with a keyword argument.
+* You can decorate with `Registrable.register` multiple times.
+* Features can be hidden behind flags.
+* Files can simply be forked. (for extreme cases)
+
+Likely no deprecation needs to occur in this case. However, in the spirit of TOOWTDI, one may still mark the old component as deprecated while _not_ removing it. Usages in AllenNLP should be removed so that our warnings are not spammy. Though silencing warnings is an option, we should "eat our own dog food" or we can hardly expect our users to migrate.
+
+4. No backward compatible change can be made.
+
+In this situation it's adivsable to consult with other developers. Hopefully they can propose an alternative solution or help mitigate the churn during the change.
+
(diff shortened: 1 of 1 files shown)


brendan-ai2:
@schmmd, rough draft of the deprecation proposal is here. Let me know what you think!

DEPRECATION.md
```diff
@@ -0,0 +1,91 @@
...
+    * New code
+    * Internal code _TODO: How is this determined?_
+    * Implementation change only
+
+  * __Care required__
+
+    * File/class/function names
+      * Including names passed to `Registrable.register`!
+    * Function signatures
+    * Jsonnet config
```
matt-gardner: I'm not sure what this means.  Do you mean changing an existing file in `training_config`?  Or something else?

DEPRECATION.md
```diff
@@ -0,0 +1,91 @@
...
+
+    * Name/API that is actively confusing multiple users.
+      * This means Github issues, messages on user channels, etc.
+    * Useful new feature
+
+  * __High__
+
+    * Major bugs
+
+3. Non-critical changes we should happily avoid. Otherwise, let's try to make the change in a backward compatible manner.
```
matt-gardner: I presume you mean "Non-critical, backwards incompatible changes".

DEPRECATION.md
```diff
@@ -0,0 +1,91 @@
...
+
+  * Options include:
+    * When deprecating, say, a class, leave a shim to its replacement.
+    * A function signature can likely be changed with a keyword argument.
+    * You can decorate with `Registrable.register` multiple times.
+    * Features can be hidden behind flags.
+    * Files can simply be forked. (for extreme cases)
+
+  * Likely no deprecation needs to occur in this case. However, in the spirit of TOOWTDI, one may still mark the old component as deprecated while _not_ removing it. Usages in AllenNLP should be removed so that our warnings are not spammy. Though silencing warnings is an option, we should "eat our own dog food" or we can hardly expect our users to migrate.
```
matt-gardner: Not sure "TOOWTDI" helps - I had to look up what it stood for, and I imagine that's true of most readers.

DEPRECATION.md
```diff
@@ -0,0 +1,91 @@
...
+
+4. No backward compatible change can be made.
```
matt-gardner: Nit: 1 and 2 are steps in a decision process, while 3 and 4 aren't worded to fit in with how this section started.

DEPRECATION.md
schmmd: I had to look it up too.  "in the spirit of maintaining a single way of doing things"

DEPRECATION.md
```diff
@@ -0,0 +1,91 @@
...
+    * The code should live for at least one full minor version and 3 months before removal.
+      * e.g., if you're committing the deprecation to master while version 0.8.4 is out, then it should live throughout version 0.9 and can first be removed in version 0.10.0.
+      * If this isn't possible, consult with other developers. You should have a quite compelling rationale.
+
+2. Remove any AllenNLP usages of the deprecated feature to avoid warnings.
+  * Suppressing warnings should be done rarely. [See here for instance.](https://github.com/allenai/allennlp/blob/9719b5c71207e642276fb1209ea1a4c8467e0792/allennlp/modules/token_embedders/embedding.py#L14)
+
+3. Wait
+
+4. Coordinate your removal PR such that it will go into at least a minor release, i.e. m.n.0.
```
schmmd: I'm not sure it's realistic for people who deprecated a feature to follow up in 3 months--particularly if they are an outside contributor.  What has been happening is that people add an issue to a future milestone (out two full release versions) but we don't have a great process for actually doing the deprecation when the time comes.  There's often other pressures to get the release done promptly

Do we expect the deprecator or the releaser to do the final removal?  I'm not sure.

DEPRECATION.md
```diff
@@ -0,0 +1,91 @@
...
+## Decision Flow
+
+Say you've found some old code that needs some care. You have a fix in mind, maybe already a PR. Great, let's get it merged! But let's give some thought to backward compatibility along the way.
+
+1. Determine whether backward compatibility is even an issue.
+
+  * __Likely safe__
+
+    * New code
+    * Internal code _TODO: How is this determined?_
```
brendan-ai2: @schmmd pointed out that it's unclear what internal code is. And I agree! :) Do we have anything in place right now to deal with this? Are there common practices in the Python community that we can adopt? This is an area I'm quite interested in getting feedback!

DEPRECATION.md
```diff
@@ -0,0 +1,91 @@
...
+
+As AllenNLP matures we will inevitably need to deprecate old models, interfaces and features. This document provides guidelines that attempt to minimize churn. Reducing churn helps users build next generation models with tools they've already mastered.
+
+### Terminology note
+
+We _deprecate_ some component to communicate that a preferred alternative exists. A _deprecated_ component does not necessarily have to be _removed_ from the library.
+
+### Goals
+
+1. Promote backward compatibility where not onerous.
```
brendan-ai2: Changed to simply "Promote backward compatibility". @schmmd , let me know if that usage strikes you as being alright. I'm a bit worried saying "when reasonable" softens things too much.

DEPRECATION.md
brendan-ai2: Added a note on using issues and that "another developer" may be the one to handle the removal PR.

DEPRECATION.md
```diff
@@ -0,0 +1,91 @@
...
+
+  * Options include:
+    * When deprecating, say, a class, leave a shim to its replacement.
+    * A function signature can likely be changed with a keyword argument.
+    * You can decorate with `Registrable.register` multiple times.
+    * Features can be hidden behind flags.
+    * Files can simply be forked. (for extreme cases)
+
+  * Likely no deprecation needs to occur in this case. However, in the spirit of TOOWTDI, one may still mark the old component as deprecated while _not_ removing it. Usages in AllenNLP should be removed so that our warnings are not spammy. Though silencing warnings is an option, we should "eat our own dog food" or we can hardly expect our users to migrate.
+
```
brendan-ai2: @schmmd , also I'm not sure I understand your point (delivered offline) about not emitting deprecation warnings. I do currently see them when I run pytest. Didn't we discuss this as one issue with softly deprecating features?

DEPRECATION.md
schmmd: Love it.

DEPRECATION.md
schmmd: @brendan-ai2 we wont have any warnings once https://github.com/allenai/allennlp/pull/2282 merges--and no deprecations should introduce any warnings.

schmmd:
A nit is that the bullets that are in-line with the numbers makes this quite hard to read.  I think this would be easier to read if you rely more on subsections and less on numberic lists.

DEPRECATION.md
```diff
@@ -0,0 +1,100 @@
...
+    * We should support whichever is longest.
+    * The code should live for at least one full minor version and 3 months before removal.
+      * e.g., if you're committing the deprecation to master while version 0.8.4 is out, then it should live throughout version 0.9 and can first be removed in version 0.10.0.
+      * In particular, this should be at least a minor release, i.e. m.n.0.
+      * If this isn't possible, consult with other developers. You should have a compelling rationale.
+
+2. Remove any AllenNLP usages of the deprecated feature to avoid warnings.
+  * Suppressing warnings should be done rarely. [See here for instance.](https://github.com/allenai/allennlp/blob/9719b5c71207e642276fb1209ea1a4c8467e0792/allennlp/modules/token_embedders/embedding.py#L14)
+
+3. Create a Github issue for the actual removal and wait for the requisite removal release.
```
schmmd: I think we need to associate the GitHub issue with a milestone corresponding to the future release in which you would like the component to be removed.  Otherwise we'll lose track of it.

DEPRECATION.md
brendan-ai2: I read the discussion for https://github.com/allenai/allennlp/pull/2282 and that's not how I interpreted it. It seems like we absolutely want to warn when we use a deprecated component. Otherwise how are we going to track down the usages?

For instance:
https://github.com/allenai/allennlp/pull/2282#issuecomment-452048702
https://github.com/allenai/allennlp/pull/2282#discussion_r245492233

DEPRECATION.md
brendan-ai2: I'm somewhat conflicted here. We don't currently appear to have any future milestones. So this would mean the deprecator would need to create a new release milestone. Are we really using them extensively? I don't want to create potential busywork.

My counterpoint would be that if no one is driving the removal forward, it's clearly not urgent. Our periodic issue sweeps would then seem sufficient. (In other words, if we intend to deprecate in version 0.10.0, but actually deprecate in version 0.11.0, that's a totally acceptable outcome.)

schmmd: @brendan-ai2 we've used milestones up to the 0.8.0 release.  They're often invaluable for making sure we tie up loose ends before a release.  Often there are a number of issues / PRs people want merged before a release is cut--and this is an effective method of managing them.

https://github.com/allenai/allennlp/milestones?state=closed

schmmd: I think my point is we should continue to use milestones to manage releases.  I think your point is that we don't necessarily need to put deprecations on milestones, which is fine.  We should make the "will be removed on or after X" instead of the "will be removed in X" as we have been in deprecation warnings.

PR #2427, opened 2019-01-23
Improve error message for "undocumented modules"

matt-gardner:
We've had a couple of people recently not understand what "undocumented module" meant in a CI failure.  This adds a nicer error message explaining what needs to happen for the check to pass.


matt-gardner:
I'm guessing this message would have been enough to help the people who got stuck; we can certainly add other information somewhere if more people are still confused.

PR #2430, opened 2019-01-23
Mark Atis Parser tests as flaky. Fix pylint.

brendan-ai2:
- Tests appear to pull down a DB file from S3.
- Spurious network failures are common, so mark as flaky to ameliorate the problem.
- Alternative: Bring DB file into fixtures directory. However, as we're trying to keep the size of the library down, this probably isn't advisable.
- Also disables some `pylint` warnings that hit with `numpy==1.16.0`.


brendan-ai2:
Thanks, Kevin! :)

PR #2431, opened 2019-01-24
Add necessary implicit embedding extension for transfer-modules api and vocab extension

HarshTrivedi:
Please don't review this yet; I still need to change some code and add tests. This is a follow-up on #2374 and #2387, but I will probably get back after #2395 gets finalized.


HarshTrivedi:
Okay, this is up for review. @joelgrus, @matt-gardner 

Two main things changed here:

1. If vocab extension is ON during training, the embedding extension should also happen implicitly. Or else, it will raise error. This is same as implicitly doing embedding extension when fine-tuning with vocab extension is ON.
2. If one is using transfer-modules api and has transferred embeding/text_field_embedder (eg. from old-archive) with Vocab extension ON, loading new-archive currently doesn't work. This is because the new-archive statedict would have extended embedding but from_params would yet load the embedding from old-archive before copying the state dict from new-archive model. So it's necessary to take precautionary embedding extension before copying state-dict.

Other change: we need to make sure embedding-extension is no-op unless we are sure (eg. it's incorrect to default to "tokens" namespace). Incorrect implicit assumption with above changes can make some tests fails.

allennlp/modules/token_embedders/embedding.py
```diff
@@ -181,12 +181,19 @@ def extend_vocab(self,  # pylint: disable=arguments-differ
...
-        # currently unextendable unless the user used default vocab_namespace 'tokens' for it.
+        # currently unextendable.
 
         vocab_namespace = vocab_namespace or self._vocab_namespace
         if not vocab_namespace:
-            vocab_namespace = "tokens"
-            logging.warning("No vocab_namespace provided to Embedder.extend_vocab. Defaulting to 'tokens'.")
+            # It's not safe to default to 'tokens' when we aren't sure that 'tokens'
+            # need to be extended. (Without this, several tests fail.)
+            logging.warning("No vocab_namespace provided to Embedder.extend_vocab. Extension will be no-op'.")
```
matt-gardner: In what circumstances will this actually emit a warning?  Will almost everyone that loads or trains a model in practice see this warning?  If so, it should be at the `info` level (or even `debug`).

allennlp/modules/token_embedders/embedding.py
```diff
@@ -181,12 +181,19 @@ def extend_vocab(self,  # pylint: disable=arguments-differ
...
+            return
+
+        extended_num_embeddings = extended_vocab.get_vocab_size(vocab_namespace)
+        if extended_num_embeddings <= self.num_embeddings:
```
matt-gardner: Shouldn't this be `==`, not `<=`?

HarshTrivedi: Yes,  if vocab and embedding are already in sync it should be `==`. I had kept `<=` as a precaution against incorrect vocab namespace. But on second thought, if user passed an incorrect vocab namespace, it's better to raise error than a silent no-op. Will change it.

allennlp/modules/token_embedders/embedding.py
HarshTrivedi: The warning won't be seen for any models trained after #2374 because `_vocab_namespace` is stored. For previously trained models, it would almost always be seen because `extend_vocab` call is implicit now. Will change it to `info` level.

allennlp/modules/token_embedders/embedding.py
HarshTrivedi: Worth explicitly raising error for `<` case? It's only possible with user explicitly passed an incorrect vocab namespace that's smaller than embedding itself.

HarshTrivedi: Okay, I have separated `==` and `<` cases, making no-op in first and raising configuration error in second.

allennlp/modules/token_embedders/embedding.py
```diff
@@ -181,18 +181,24 @@ def extend_vocab(self,  # pylint: disable=arguments-differ
...
-        # currently unextendable unless the user used default vocab_namespace 'tokens' for it.
+        # currently unextendable.
 
         vocab_namespace = vocab_namespace or self._vocab_namespace
         if not vocab_namespace:
-            vocab_namespace = "tokens"
-            logging.warning("No vocab_namespace provided to Embedder.extend_vocab. Defaulting to 'tokens'.")
+            # It's not safe to default to 'tokens' when we aren't sure that 'tokens'
+            # need to be extended. (Without this, several tests fail.)
+            logging.info("No vocab_namespace provided to Embedder.extend_vocab. Extension will be no-op'.")
```
matt-gardner: To make this more obvious, I'd recommend a message like "Loading a model trained before embedding extension was implemented; pass an explicit vocab namespace if you want to extend the vocabulary."

allennlp/modules/token_embedders/embedding.py
```diff
@@ -181,18 +181,24 @@ def extend_vocab(self,  # pylint: disable=arguments-differ
...
         # knowing which is necessary at time of embedding vocab extension. So old archive models are
-        # currently unextendable unless the user used default vocab_namespace 'tokens' for it.
+        # currently unextendable.
 
         vocab_namespace = vocab_namespace or self._vocab_namespace
         if not vocab_namespace:
-            vocab_namespace = "tokens"
-            logging.warning("No vocab_namespace provided to Embedder.extend_vocab. Defaulting to 'tokens'.")
+            # It's not safe to default to 'tokens' when we aren't sure that 'tokens'
+            # need to be extended. (Without this, several tests fail.)
```
matt-gardner: No need to reference failing tests in comments in the code - just give the justification (that it's not safe to default to "tokens").

allennlp/modules/token_embedders/embedding.py
HarshTrivedi: @matt-gardner  Correcting this revealed a subtle issue: defaulting to `tokens` and `token_characters` namepsace can be problematic, when `num_embeddings` was used instead of vocab namespace to decide embedding size. Fixed this in last commit. 

This is up for another look now.

HarshTrivedi: .. oops, didn't realize you already reviewed again!

HarshTrivedi:
@matt-gardner, Thanks for the review! This should be good to merge now.

PR #2432, opened 2019-01-24
adding patch

kernelmachine:


allennlp/modules/token_embedders/bag_of_word_counts_token_embedder.py
```diff
@@ -14,22 +15,36 @@ class BagOfWordCountsTokenEmbedder(TokenEmbedder):
...
-        size of vocabulary
-    projection_dim : int, optional (default = None)
+    vocab: ``Vocabulary``
+    vocab_namespace: ``str``
+        namespace of vocabulary to embed
+    projection_dim : ``int``, optional (default = ``None``)
         if specified, will project the resulting bag of words representation
         to specified dimension.
+    ignore_oov : ``bool``, optional (default = ``False``)
+        if true, will set entry corresponding to OOV token to zero.
```
joelgrus: I don't feel like this is a great description, especially if you're using a projection layer. Wouldn't it be more accurate to just say, "If true, we ignore the OOV token"?

allennlp/modules/token_embedders/bag_of_word_counts_token_embedder.py
```diff
@@ -14,22 +15,36 @@ class BagOfWordCountsTokenEmbedder(TokenEmbedder):
...
+        self.vocab = vocab
+        self.vocab_size = vocab.get_vocab_size(vocab_namespace)
         if projection_dim:
-            self._projection = torch.nn.Linear(vocab_size, projection_dim)
+            self._projection = torch.nn.Linear(self.vocab_size, projection_dim)
         else:
             self._projection = None
-        self.output_dim = projection_dim or vocab_size
+        self._ignore_oov = ignore_oov
+        self.oov_idx = vocab.get_token_to_index_vocabulary(vocab_namespace).get(vocab._oov_token)
```
joelgrus: why is `_ignore_oov` a "private" variable but `oov_index` a "public" one?

should any of these instance variables be public?

allennlp/modules/token_embedders/bag_of_word_counts_token_embedder.py
```diff
@@ -49,11 +64,13 @@ def forward(self,  # pylint: disable=arguments-differ
         ``(batch_size, vocab_size)``
         """
         bag_of_words_vectors = []
-
-        mask = get_text_field_mask({'tokens': inputs})
+        mask = get_text_field_mask({'tokens': inputs}, 1)
```
joelgrus: wouldn't the logic be much cleaner if you did something here like

```
if self._ignore_oov:
    # also mask out positions corresponding to oov
    mask *= inputs != self._oov_idx
```
then you wouldn't have to do the `.nonzero().squeeze()` part below, which i find extremely non-obvious.

allennlp/modules/token_embedders/bag_of_word_counts_token_embedder.py
```diff
@@ -72,9 +89,12 @@ def from_params(cls, vocab: Vocabulary, params: Params) -> 'BagOfWordCountsToken
         we look for a ``vocab_namespace`` key in the parameter dictionary
         to know which vocabulary to use.
         """
+
         vocab_namespace = params.pop("vocab_namespace", "tokens")
-        vocab_size = vocab.get_vocab_size(vocab_namespace)
         projection_dim = params.pop_int("projection_dim", None)
+        ignore_oov = params.pop("ignore_oov", False)
```
joelgrus: use `pop_bool` here, in case we want to pass that in as an environment variable or something

kernelmachine:
@joelgrus incorporated all your comments, ready for another look!

PR #2438, opened 2019-01-26
Decompose LanguageModel contextualizer into forward_ and backward_ contextualizer

nelson-liu:
As discussed in #2373 , this PR adds `forward_contextualizer` and `backward_contextualizer` arguments to the `LanguageModel`, towards bidirectional language modeling of contiguous text.


nelson-liu:
FYI @brendan-ai2 , I was able to retrain the model at https://github.com/allenai/allennlp/blob/master/training_config/constituency_parser_transformer_elmo.jsonnet with these changes, so seems like backwards compatibility is maintained.

allennlp/models/language_model.py
```diff
@@ -100,22 +120,88 @@ def __init__(self,
...
+                          "It will be removed in version 0.10 .",
+                          DeprecationWarning)
+            if contextualizer.is_bidirectional() is not bidirectional:
+                raise ConfigurationError(
+                        "Bidirectionality of contextualizer must match bidirectionality "
+                        "of language model. "
+                        f"Contextualizer bidirectional: {contextualizer.is_bidirectional()}, "
+                        f"language model bidirectional: {bidirectional}")
+            if contextualizer.is_bidirectional():
+                # TODO (nfliu): Emit warning about cheating
```
nelson-liu: not too sure what the wording of this warning would be like, thoughts @matt-peters ?

allennlp/models/language_model.py
```diff
@@ -264,9 +350,34 @@ def forward(self,  # type: ignore
...
+        if self._use_contextualizer_arg:
+            contextual_embeddings = self._contextualizer(embeddings, mask)
+        else:
+            contextual_embeddings = self._forward_contextualizer(embeddings, mask)
+            if self._bidirectional:
+                backward_contextual_embeddings: Union[torch.Tensor, List[torch.Tensor]] = (
+                        self._backward_contextualizer(embeddings, mask))
+                # Concatenate the backward contextual embeddings to the
+                # forward contextual embeddings
+                if (isinstance(contextual_embeddings, list) and
```
nelson-liu: @brendan-ai2 i'm struggling to test this case (where `contextual_embeddings` and `backward_contextual_embeddings` are lists), since I'm not sure when they would return lists :) any pointers?

brendan-ai2: Off the top of my head, a solution would be to modify https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2seq_encoders/seq2seq_encoder.py#L5 so that it takes a constructor arg "return_all_layers" (like with lazy in the `DatasetReader`) and then have the subclasses do the appropriate thing. Maybe also add `forward` as an explicit method in `Seq2SeqEncoder` as well in order to clearly document the behavior and the type.

nelson-liu: I think the main trouble is that the PyTorch LSTM doesn't return the outputs for all layers and all timesteps?

nelson-liu: I guess for RNNs in particular, we could explicitly loop over all of the timesteps in the case where `return_all_layers=True` (i.e., it's being used as an embedder)? Would be slow, though...

brendan-ai2: Yeah, this seems pretty busted. Probably another reason why Calypso only used a 1-layer LSTM. That said, just because PyTorch RNNs don't give us hidden states at all layers and for all timesteps doesn't mean that `Seq2SeqEncoder` shouldn't expose that information when the implementation supports it. Of course, that doesn't help your immediate problem. :/

Possible options:
1) Keep using only 1-layer LSTMs.
2) Return only the hidden states in the last layer.
3) Return the hidden states in the last layer and final hidden state for each other layer duplicated for each timestep. (That information does appear to be returned by PyTorch LSTMs.)
4) Find/write an alternate LSTM implementation.

1) and 2) are easiest, but presumably aren't sufficient for your use case or we wouldn't need to make these modifications. 3 is probably nonsense though maybe with a sufficiently strong model downstream it could still learn something. 4) is possible of course, but not ideal. @DeNeutoy, any ideas here?

brendan-ai2: And yeah, looping would probably be prohibitive in terms of speed.

matt-gardner: I'll let you two figure out what's right for this case, but in general, we shouldn't add things to high-level abstractions that not all implementations of the abstraction will support - that breaks the abstraction.  So things like "when the implementation supports it" make me a bit nervous.  Maybe it's ok in this case, and you can have a default implementation that just unsqueezes, or something, I don't know.  You can figure that out.  Just be careful.

brendan-ai2:
@nelson-liu, just wanted to check in on this and make sure I'm not blocking you. Are you still proceeding with this approach?

allennlp/models/language_model.py
```diff
@@ -74,6 +75,23 @@ class LanguageModel(Model):
     contextualizer: ``Seq2SeqEncoder``
         Used to "contextualize" the embeddings. As described above,
         this encoder must not cheat by peeking ahead.
+
+        .. deprecated:: 0.8.2
```
DeNeutoy: Question: To me this seems like we are ripping out quite a good api for one which is more complicated. We already have a trained `BidirectionalLanguageModel` which uses a bidirectional transformer, which is very useful to many people (e.g github issues/ Swabha is using it in her research), which makes me unsure that depreciating a key component of it is a good idea. Do we know that bidirectional language modeling of contiguous text is worth it? I can't imagine a downstream task where you actually need unlimited context (and even if there was one, how that would practically work). Is it worth ripping this into a separate repo, making the changes and double checking that contiguous text does something useful? 

Additionally, we now have a `BidirectionalLanguageModel` subclass which is subsumed by this code (I think). This is "bad code smell" to me when we have a subclass which only differs in constructor arguments and we only have a single example of how those arguments might differ (it triggers my `MultiClassMultipleChoiceMemoryNetwork(MemoryNetwork, MultiClass, MultipleChoice)` sense from [DeepQa](https://github.com/allenai/deep_qa/pull/389/files#diff-9b4839b38293d3faa5e7663cc69e7e3aL10)).

allennlp/models/language_model.py
DeNeutoy: Sorry, I missed this mention. We don't need to handle the case that the contextualizer returns multiple layers here, because it will never do that during training.

I think really the problem here has stemmed from the fact we have been passing around the `return_all_layers` argument as a constructor parameter and not a runtime `forward` parameter (or as a separate method on a `Contextualizer` subclass of `Seq2SeqEncoder`.

Is the right API here this:

```python
class Seq2SeqEncoder

class Contextualizer(Seq2SeqEncoder):

	# Idea 1:
    def forward(sequence: torch.Tensor[batch, sequence, embedding],
				return_all_layers = False) -> Union[torch.Tensor[batch, sequence, embedding], List[of the same thing]]

	# this way all Contextualizers can still function as `Seq2SeqEncoders` by default,
	# which is useful for training and perhaps for downstream use, say if you wanted to
	# fine tune an encoder and didn't want to do an elmo mixture. However, all
	# `ContextualTokenEmbedders` could call `Contextualizer.forward(return_all_layers=True)`.

	# Second idea:

    def forward(sequence: torch.Tensor[batch, sequence, embedding]) -> torch.Tensor[batch, sequence, embedding]

	def get_layers(sequence: torch.Tensor) -> List[torch.Tensor[batch, sequence, embedding]]:

	# A separate method which basically implements the functionality above.
	# This would get around the type problems of `forward` possibly returning lists
	# of tensors when used as a `Seq2SeqEncoder`.
```

This doesn't handle Nelson's case that the contextualizers might be stateful during training. I haven't thought hard about that because i'm not 100% convinced that we need to support it as part of a concrete API yet, but i'm happy to be convinced!

nelson-liu: more thoughts later, but: 

> This doesn't handle Nelson's case that the contextualizers might be stateful during training. I haven't thought hard about that because i'm not 100% convinced that we need to support it as part of a concrete API yet, but i'm happy to be convinced!

How else would you do language modeling of contiguous text? At the very least, it'd be useful to add the (left-to-right) LM functionality to AllenNLP. I think having the bidirectional contiguous text LM would be useful as well, since I'm actually curious to run the experiment and find out whether contiguous text matters (I suspect it does, or at least doesn't hurt). Bidirectionality is definitely crucial to getting the best contextual representations.

allennlp/models/language_model.py
matt-gardner: Leaving the API decision to others, here's some motivation for enabling modeling of longer contexts: https://ai.googleblog.com/2019/01/transformer-xl-unleashing-potential-of.html.  They basically make a stateful transformer, showing pretty big gains.

matt-gardner:
@nelson-liu, how do you want to proceed with this?

nelson-liu:
I plan on coming back to this eventually, just got caught up with some other stuff...

matt-gardner:
@nelson-liu, is now a good time to come back to this?

allennlp/models/language_model.py
matt-gardner: Would #2716 help here?

DeNeutoy:
Closing due to inactivity and the Age Of The Transformer, which makes this PR not super important.

PR #2439, opened 2019-01-26
Add support to kwargs in TimeDistributed

### LINKED ISSUE #2410: TimeDistributed's forward method doesn't support some arguments
I see 2 problems with `TimeDistributed`'s `forward` method. They can be seen as 2 separate issues, but as they're strongly correlated, I believe they should be solved together.

The first one is that it doesn't support passing named arguments when calling it. The second one is that it doesn't work with some parameters of the `forward` method of certain wrapped module classes. I'll elaborate on the first one and the second one will arise while explaining.

`TimeDistributed`'s `forward` method not supporting named arguments implies that the method call will work successfully depending on how do you call it, which is not good. For example, if I wrap `PytorchSeq2SeqWrapper` in a `TimeDistributed` instance and then do `time_distributed(inputs=x)`, it will fail. But if I do `time_distributed(x)` it works.

So, it's like the method should grab all positional and keyword parameters and time-distribute them. However, I understand that the `forward` method implementation of some modules have some keyword parameters that should be treated differently if they're wrapped. Hence, the fix is not trivial and it faces other issues. I see 3 kinds of parameters:

1. Parameters that need to be simply time-distributed. Such as `inputs` or `mask` from `PytorchSeq2SeqWrapper`'s `forward` method.

2. Parameters that need to be time-distributed, but require some special treatment. Such as `hidden_state` from `PytorchSeq2SeqWrapper`'s `forward` method. Because `hidden_state` is a tensor of shape `(num_layers, batch_size, hidden_size)`, then the time dimension should be in a different index position than the common case: `(num_layers, batch_size, time, hidden_size)`.

3. Parameters that do not need to be time-distributed. Such as `output_size` of `ConvTranspose2d`'s `forward` method.

I believe that for 2. and 3. we can use some sort of blacklisting when time-distributing params, and give special treatment to them. However, this sounds a bit brittle. What if new modules are added later, with params that require new treatment? This class will constantly need modification.

Another problem is: how to identify the params when they come? `hidden_state` could be passed either as a positional argument or as a named one. So, it's like we should check for the 3rd element of `*inputs` if `self._module` is of type `PytorchSeq2VecWrapper` or for a keyword parameter named `hidden_state` otherwise. This is also very brittle, because the module class can possibly change the order of the params. Maybe it can be solved in part if the docstring of `TimeDistributed` ask parameters to be passed in a certain way.

How can we solve these issues of `TimeDistributed` so it can be used more seamlessly and so it supports params such as `hidden_state` of `PytorchSeq2VecWrapper`'s `forward` method? I'm happy to send a patch if we can find a good solution for this.

bryant1410:
Fix #2410

Tasks:

* [x] Add support to kwargs in `TimeDistributed`.
* [x] Add tests.


allennlp/modules/time_distributed.py
```diff
@@ -15,28 +17,55 @@ class TimeDistributed(torch.nn.Module):
...
-        reshaped_outputs = self._module(*reshaped_inputs)
+                reshaped_kwargs[key] = self._reshape_tensor(value)
+
+        reshaped_outputs = self._module(*reshaped_inputs, **reshaped_kwargs, **ignored_kwargs)
+
+        if some_input is None:
+            raise RuntimeError("No input tensor to time-distribute")
 
         # Now get the output back into the right shape.
-        # (batch_size, time_steps, [hidden_size])
```
bryant1410: `hidden_size` name didn't make much sense to me here. `output_size` did. Was there a reason for it?

matt-gardner: `output_size` seems fine to me.

allennlp/modules/time_distributed.py
```diff
@@ -15,28 +17,55 @@ class TimeDistributed(torch.nn.Module):
...
+        some_input = None
+        if inputs:
+            some_input = inputs[-1]
 
-            # Squash batch_size and time_steps into a single axis; result has shape
-            # (batch_size * time_steps, input_size).
-            squashed_shape = [-1] + [x for x in input_size[2:]]
-            reshaped_inputs.append(input_tensor.contiguous().view(*squashed_shape))
+        reshaped_kwargs = {}
+        ignored_kwargs = {}
```
matt-gardner: Probably simpler to just use one dictionary here, adding both pass-through and other kwargs to `reshaped_kwargs`.

allennlp/modules/time_distributed.py
```diff
@@ -15,28 +17,55 @@ class TimeDistributed(torch.nn.Module):
...
+        super().__init__()
         self._module = module
 
-    def forward(self, *inputs):  # pylint: disable=arguments-differ
-        reshaped_inputs = []
-        for input_tensor in inputs:
-            input_size = input_tensor.size()
-            if len(input_size) <= 2:
-                raise RuntimeError("No dimension to distribute: " + str(input_size))
+    def forward(self, *inputs, pass_through: Optional[Iterable] = None,  # pylint: disable=arguments-differ
```
matt-gardner: Minor formatting thing: this would be easier to read if you put the `**kwargs):` on this line, and moved the pylint command to the line below (the first line of `forward`).

matt-gardner: Also, for the type annotation on `pass_through`, use `List[str]`, without any `Optional` - the `= None` means the same thing as `Optional`, so `Optional` is redundant.

allennlp/tests/modules/time_distributed_test.py
```diff
@@ -15,7 +15,26 @@ def test_time_distributed_reshapes_correctly(self):
...
+        char_embedding = Embedding(2, 2)
+        char_embedding.weight = Parameter(torch.FloatTensor([[.4, .4], [.5, .5]]))
+        distributed_embedding = TimeDistributed(char_embedding)
+        char_input = torch.LongTensor([[[1, 0], [1, 1]]])
+        output = distributed_embedding(input=char_input)
+        assert_almost_equal(output.data.numpy(),
+                            [[[[.5, .5], [.4, .4]], [[.5, .5], [.5, .5]]]])
+
+    def test_time_distributed_reshapes_named_arg_and_kwarg_correctly(self):
+        # TODO: maybe use PyTorchSeq2VecWrapper?
```
matt-gardner: It might be easiest to just define a fake module here that has all three kinds of arguments and use that for the test.

allennlp/modules/time_distributed.py
```diff
@@ -15,28 +17,55 @@ class TimeDistributed(torch.nn.Module):
...
+                **kwargs):
+        if pass_through is None:
```
HarshTrivedi: in case you prefer, `pass_through = pass_through or []`.

allennlp/modules/time_distributed.py
bryant1410: > Also, for the type annotation on `pass_through`, use `List[str]`, without any `Optional` - the `= None` means the same thing as `Optional`, so `Optional` is redundant.

Why `List` and not `Iterable`?

matt-gardner: Because that's what we use everywhere else in the code.  The type annotations are mostly for readability, and for simple mypy checks, not for strict compliance with types everywhere.  So, we don't need to be as generic as possible with our types, as you might be in some other languages.

bryant1410: Changed it.

Actually, in this case, it should be `Container` and not `Iterable`, because `__contains__` is the only operation that matters here for `pass_through`.

I kinda disagree on this one though. First, I understand this is a separate issue, so it shouldn't be addressed here (and should be addressed for the codebase overall then, if so). Second, I believe there's a reason why such typing annotations exist (e.g., `Container` and `Iterable`). I use PyCharm, and if in a similar scenario I pass a `set` instead of a `list`, there's a chance it'll show a warning and I think it's okay because the method expects a `list`. But in a case like this, it should be fine to pass a `set`. On the other hand, it can mean to "complicate" stuff and also cause a maintenance burden because method signatures could potentially change all the time (e.g., because `pass_through` is used to be iterated after certain change) or maintainers forget to change it, albeit I'm not sure how frequently it could happen. Any thoughts on this?

matt-gardner: https://google.github.io/styleguide/pyguide.html#4-parting-words

Our type hints and our use of type checkers are purely an aid to our sanity, not the other way around.  We shouldn't make ourselves go into contortions just to make a type checker happy; that's counter productive.  The bottom line: we have a lot of code here that uses `List`, because that's the easiest intuitive type annotation for a human to read, even if a `Set` might also be appropriate, so we really should be using something more general.  Be consistent with the code around you.  For similar reasons, and because the `Optional` really doesn't add anything except for the type checker, just remove the `Optional` annotation.

bryant1410: Oh, didn't know that rule. I like it and approve it.

What you say totally makes sense. I've already removed the `Optional` thing. Thanks for the feedback!

matt-gardner: One last point - there is a point where the difference between an annotation of `List` and `Iterable` makes a big difference.  That's in our [`FromParams` code](https://github.com/allenai/allennlp/blob/84eb4d1acc03bad89e2282306eb4ede8f1badec5/allennlp/common/from_params.py#L176-L212).  If we see a type annotation of `Iterable`, we don't know how to instantiate it.  I guess we could default to constructing a list, or something, but it's a lot more clear if you just give the type that you want.

bryant1410: I see, interesting point. But does it represent a problem even though this module is not registered?

matt-gardner: No, not for this module, and especially not for the `forward` method.  But that's another reason why we just tend to use `List` annotations instead of the more abstract classes.

bryant1410:
Okay, it is ready for final review. I added enough tests IMHO. Feel free to point out if more tests are needed. The 3 cases of #2410 are covered (2 and 3 are the same actually).

bryant1410:
Also, I'll let the maintainers update the branch through GitHub when it's ready to merge maybe, I think it's the best. I don't think there's gonna be conflicts given that I don't expect these 2 files to be modified often honestly.

PR #2444, opened 2019-01-28
[Feature Enhanced]Add FeedforwardEncoder for Sequences

WrRan:
This new class (`FeedforwardEncoder`) applies the `FeedForward` to each item in sequences. So we can specify a feedforward network for sequences just by changing a configuation file now.

### DIFF
--- a/allennlp/modules/seq2seq_encoders/feedforward_encoder.py
+++ b/allennlp/modules/seq2seq_encoders/feedforward_encoder.py
@@ -45,5 +45,5 @@ class FeedForwardEncoder(Seq2SeqEncoder):
         if mask is None:
             return self._feedforward(inputs)
         else:
-            inputs = inputs * mask.unsqueeze(dim=-1).float()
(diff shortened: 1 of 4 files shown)


allennlp/modules/seq2seq_encoders/feedforward_encoder.py
```diff
@@ -0,0 +1,49 @@
...
+            A tensor of shape (batch_size, timesteps).
+
+        Returns
+        -------
+        A tensor of shape (batch_size, timesteps, output_dim).
+        """
+        if mask is None:
+            return self._feedforward(inputs)
+        else:
+            inputs = inputs * mask.unsqueeze(dim=-1).float()
```
matt-gardner: I would think you'd want this on the output of the feedforward, not on its input.  Right?

WrRan: Thanks. You are right. :-)

PR #2446, opened 2019-01-28
Add QaNet model

yizhongw:


training_config/qanet.jsonnet
```diff
@@ -0,0 +1,152 @@
...
+        "passage_length_limit": 1000,
+        "question_length_limit": 100,
+        "skip_invalid_examples": false
+    },
+    "vocabulary": {
+        "min_count": {
+            "token_characters": 200
+        },
+        "pretrained_files": {
+            "tokens": "https://s3-us-west-2.amazonaws.com/yizhongw-dev/glove/glove.840B.300d.lower.zip"
```
matt-gardner: Do we not have this file in `allennlp/`?  If not, we should add it, and use that link instead.

allennlp/data/dataset_readers/reading_comprehension/squad.py
```diff
@@ -26,6 +26,11 @@ class SquadReader(DatasetReader):
     ``metadata['token_offsets']``.  This is so that we can more easily use the official SQuAD
     evaluation script to get metrics.
 
+    We also support limiting the maximum passage length and question length for training and evaluation.
+    Since some answer spans may exceed the maximum passage length, we regard such answer as invalid.
+    During training, we skip those examples of which all answer spans are invalid (invalid examples);
+    During testing, we use the last token as the pseudo gold answer for these invalid examples.
```
matt-gardner: I'm trying to understand what happens here.  We still have the correct answer string for the evaluation script, so we'll get it wrong no matter what we predict?  Is that right?  That's what should happen, so if that's what's happening, great, and we should make it more clear in the documentation somehow.

allennlp/modules/seq2seq_encoders/qanet_encoder.py
```diff
@@ -0,0 +1,312 @@
...
+
+        feedforward_norm_out = self.dropout(self.feedforward_norm_layer(output))
+        feedforward_out = self.dropout(self.feedforward(feedforward_norm_out))
+        sublayer_count += 1
+        output = self.residual_with_layer_dropout(output, feedforward_out,
+                                                  sublayer_count, total_sublayers)
+        return output
+
+
+class ResidualWithLayerDropout(torch.nn.Module):
```
matt-gardner: I think this class should be in its own file, under `allennlp.modules`.

allennlp/models/reading_comprehension/qanet.py
```diff
@@ -0,0 +1,262 @@
...
+            If this is given, we will compute a loss that gets included in the output dictionary.
+        span_end : ``torch.IntTensor``, optional
+            From an ``IndexField``.  This is one of the things we are trying to predict - the
+            ending position of the answer with the passage.  This is an `inclusive` token index.
+            If this is given, we will compute a loss that gets included in the output dictionary.
+        metadata : ``List[Dict[str, Any]]``, optional
+            If present, this should contain the question ID, original passage text, and token
+            offsets into the passage for each instance in the batch.  We use this for computing
+            official metrics using the official SQuAD evaluation script.  The length of this list
+            should be the batch size, and each dictionary should have the keys ``id``,
```
matt-gardner: Is this still accurate?  I'm pretty sure there's no "id" key being used anywhere here anymore.

allennlp/models/reading_comprehension/qanet.py
```diff
@@ -0,0 +1,262 @@
...
+            ``original_passage``, and ``token_offsets``.  If you only want the best span string and
+            don't care about official metrics, you can omit the ``id`` key.
+        Returns
```
matt-gardner: Blank line before the `Returns` block.

allennlp/models/reading_comprehension/qanet.py
```diff
@@ -0,0 +1,262 @@
...
+        span_end_input = torch.cat([modeled_passage_list[-3], modeled_passage_list[-1]], dim=-1)
+        span_end_logits = self._span_end_predictor(span_end_input).squeeze(-1)
+        span_start_logits = util.replace_masked_values(span_start_logits, passage_mask, -1e32)
+        span_end_logits = util.replace_masked_values(span_end_logits, passage_mask, -1e32)
+
+        # Shape: (batch_size, passage_length)
+        span_start_probs = torch.nn.functional.softmax(span_start_logits, dim=-1)
+        span_end_probs = torch.nn.functional.softmax(span_end_logits, dim=-1)
+
+        best_span = BidirectionalAttentionFlow.get_best_span(span_start_logits, span_end_logits)
```
matt-gardner: Can you move this `get_best_span` function to a module like `allennlp.models.reading_comprehension.util`, instead of calling the method on BiDAF?  You can leave the `BidirectionalAttentionFlow.get_best_span` function there, just calling out to the util function, to maintain backwards compatibility.

allennlp/models/reading_comprehension/qanet.py
yizhongw: Yes, this is outdated, but it's the same for the bidaf model. Should I change both of them?

matt-gardner: That would be great.

allennlp/data/dataset_readers/reading_comprehension/squad.py
yizhongw: Yes, that's right. The correct answer string does not change. The span should be skipped, or replaced with a pseudo span if it exceeds the cut passage length. Otherwise, it will cause error in making the IndexField for span.

matt-gardner: Great, can you update the docstring to mention this.  Even better, it'd be great to have a test showing that this still works as intended - we _really_ don't want to compute inaccurate metrics at validation / test time.

training_config/qanet.jsonnet
matt-gardner: If we're referring to it in an allennlp training config, I'd say we should host the file on our S3 bucket.  Just make sure the name is obvious enough that it's modified by us, and maybe leave a comment in the json file briefly describing what you put here.

matt-gardner:
Also, I just merged an update to the `get_span_logic` on master - can you be sure to use the new logic in what you're moving to the new `util` module?

yizhongw:
Yes, that's cool. I will do that when merging with master.

allennlp/models/reading_comprehension/util.py
```diff
@@ -5,31 +5,29 @@ def get_best_span(span_start_logits: torch.Tensor, span_end_logits: torch.Tensor
...
-                val1 = span_start_logits[b, j]
-
-            val2 = span_end_logits[b, j]
+    device = span_start_logits.device
+    # (batch_size, passage_length, passage_length)
+    span_log_probs = span_start_logits.unsqueeze(2) + span_end_logits.unsqueeze(1)
+    # Only the upper triangle of the span matrix is valid; the lower triangle has entries where
+    # the span ends before it starts.
+    span_log_mask = torch.triu(torch.ones((passage_length, passage_length),
+                                          device=device)).log().unsqueeze(0)
```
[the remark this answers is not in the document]
yizhongw: If I do that, the `span_log_mask` would be two-dimensional, which does not match with the `span_log_probs`?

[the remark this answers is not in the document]
matt-gardner: There was a comment on the previous PR saying it wasn't necessary.  I just removed it, and we'll see if it crashes...  If it does, I'll add it back.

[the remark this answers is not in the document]
matt-gardner: Pytorch might be smart enough to do the broadcasting; I'm not sure in this case if it is or not.  I suppose it could possibly lead to some crazy edge case, maybe...  I was just trying to avoid an additional operation.

yizhongw:
Yes, I am retraining the model. Once it is done, I would put it on S3. I have the permission.

allennlp/models/reading_comprehension/util.py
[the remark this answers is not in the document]
matt-gardner: Looks like tests pass without it.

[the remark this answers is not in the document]
yizhongw: Ok, you already did that. That's fine.

PR #2451, opened 2019-01-28
Ensure contiguous initial state tensors in `_EncoderBase(stateful=True)`

rloganiv:
This PR fixes the following bug:

**Describe the bug** If subsequent batches of inputs containing a zero-length sequence are passed to a stateful encoder (e.g. a child of `_EncoderBase` with the `stateful` parameter set to `True`) then the following error is raised:
```
RuntimeError: rnn: hx is not contiguous
```

**To Reproduce**
```python
import torch

from allennlp.modules.seq2seq_encoders import PytorchSeq2SeqWrapper


lstm = torch.nn.LSTM(input_size=2,
                     hidden_size=2,
                     num_layers=3,
                     dropout=0.1,
                     batch_first=True)
encoder = PytorchSeq2SeqWrapper(lstm, stateful=True)
encoder = encoder.cuda(0)

inputs = torch.randn(4, 4, 2).cuda(0)
mask = torch.ones(4, 4).cuda(0)
mask[2, :] = 0

encoder(inputs, mask)
encoder(inputs, mask)
```


rloganiv:
Sure thing! I added checks to the test written for #1493, since it essentially deals with the same problem (only for non-stateful encoders). Also, since I can only provoke the issue on the GPU I added a GPU-only copy of the tests.

DeNeutoy:
Hi! Sorry could you guard your test using the pytest decorator, rather than just the if statement inside the test: `@pytest.mark.skip(torch.cuda.is_available(), "requires cuda")` or something?

Issue #2454, opened 2019-01-29
Unnecessary cast to float32 in ArrayField


festeh:
**Describe the bug**
    ArrayField silently casts the array to float32 during as_tensor call

**To Reproduce**
```
array = np.array([1, 2, 3])
field = ArrayField(array)
tensor = field.as_tensor(field.get_padding_lengths())
print(array.dtype, tensor.dtype)
```

**Expected behavior**
    I would expect that padding operation generally should not change array type. I suppose that `self.array.dtype` information should be used, if available.

matt-gardner:
Looks like the problem might be this line: https://github.com/allenai/allennlp/blob/dddcbf243e9931a5ab0241435b39613117e20acf/allennlp/data/fields/array_field.py#L31

PRs welcome!

DeNeutoy:
Closed by #2609

PR #2460, opened 2019-01-30
Use self.array.dtype in ArrayField

festeh:
Simple fix for #2454.

I'd propose to use dtype of inner numpy array in ArrayField instead of float32.

### DIFF
--- a/allennlp/data/fields/array_field.py
+++ b/allennlp/data/fields/array_field.py
@@ -28,7 +28,7 @@ class ArrayField(Field[numpy.ndarray]):
                      for i in range(len(padding_lengths))]
 
         # Convert explicitly to an ndarray just in case it's an scalar (it'd end up not being an ndarray otherwise)
-        return_array = numpy.asarray(numpy.ones(max_shape, "float32") * self.padding_value)
+        return_array = numpy.asarray(numpy.ones(max_shape, dtype=self.array.dtype) * self.padding_value)
 
         # If the tensor has a different shape from the largest tensor, pad dimensions with zeros to
         # form the right shaped list of slices for insertion into the final tensor.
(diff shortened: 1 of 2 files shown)


brendan-ai2:
Oof, this is an annoying situation. And not particularly unique to this file. We have many usages where `torch.Tensor` really means `torch.FloatTensor`. Arguably this is even correct. According to the Pytorch docs, "torch.Tensor is an alias for the default tensor type (torch.FloatTensor)." (https://pytorch.org/docs/stable/tensors.html#torch.Tensor)

So the main additional option would be to do nothing. I read https://github.com/allenai/allennlp/issues/2454 and it's not clear to me that this is a particularly pernicious bug. Not saying it's not either, just that it would be great to hear more. If we do think it's serious enough to tackle, we should be sure to pick a solution we can apply consistently across the codebase. Standardizing our type hints would likely be part of this. And then yes, probably apply option 1.

brendan-ai2:
(Also, at AAAI this week. So my responses will be slow/sporadic.)

festeh:
Well, it turned out that I could actually use `SequenceLabelField` in my case, so probably do nothing is a most painless option here. My goal was to fine-tune [pretrained BERT model](https://github.com/huggingface/pytorch-pretrained-BERT) for multiple choice QA task and I initially used `ArrayField` for storing a `Long` tensor that separates parts of the input, as this model requires this argument and crashes if it has a different type.

For example, if the input sequence is `q_1 q_2 q_3 a_1 a_2` this tensor would be `[1, 1, 1, 0, 0]`. As It will always be the same length as the input, it's safe to use `SequenceLabelField`. Currently, I could not provide a case when one would need to have a `Long` tensor with different length from the input, but this may be not a hard task for someone.

Finally, if you would like to keep the current behavior of `ArrayField`, I'd also suggest to make it explicit by stating that array would be converted to `float32` in the docstring.

matt-gardner:
@brendan-ai2, I don't think the type hints are a huge issue - we're not very strict at all about annotating the dtype of a tensor, and it doesn't really help anywhere here, anyway, because the `ArrayField` is matching the `Field` API, which has a very generic return type.

But, yeah, I guess the easiest fix is just to make the documentation of `ArrayField` clear that the result is always a float32 tensor, and if you need it to be something else you need to cast it.

matt-gardner:
We have just used `torch.Tensor` annotations to mean any type of tensor.  I didn't know (and I suspect that no one else on the team knew) that pytorch treated this as `torch.FloatTensor`.  We have generally been very lax about trying to be specific about the types of tensors in type annotations, because it's hard and annoying to get right.  When there is an important issue, we'll note it in the docstring, where expected input shapes are also often specified.

DeNeutoy:
Closing in favour of #2609

PR #2466, opened 2019-01-31
Add model parameters / modules inspection helper.

HarshTrivedi:
I often find it quite tedious to setup model path based regexes eg. in initialization, when there are many levels of submodules. This adds a `inspect_dict` method on model which inspects `parameters` or `modules` - test fixtures show example outputs. I am not insisting on `modules` mode since it's almost like `print(model)` output. But `parameters` inspection's been helpful.

- It might be good idea to also dump model parameters inspection in serialization dir. Could be helpful to define regexes at time of fine-tuning, transferring learning initializer etc.
- During training, we log which parameters are tunable and which are not. It would be more parsable if output is shown similar to to `parameters` output of `inspect_dict`.


allennlp/models/model.py
```diff
@@ -333,6 +335,61 @@ def extend_embedder_vocab(self, extended_vocab: Vocabulary) -> None:
             if isinstance(module, (Embedding, TextFieldEmbedder)):
                 module.extend_vocab(extended_vocab)
 
+    def inspection_dict(self,
```
joelgrus: I would rather this not be a method on `Model` (it doesn't seem to access any private variables), why not just make it a helper function in `allennlp.nn.util`?

joelgrus: also, I don't like the name "inspection_dict", that doesn't tell me anything about what it does. I would write two functions and call one `inspect_modules` and one `inspect_parameters`. But why have separate methods for those? Why not just inspect everything at once?

allennlp/models/model.py
```diff
@@ -333,6 +335,61 @@ def extend_embedder_vocab(self, extended_vocab: Vocabulary) -> None:
...
+                        inspect: str = "parameters",
+                        quite: bool = False) -> OrderedDict:
```
joelgrus: quiet

allennlp/models/model.py
```diff
@@ -333,6 +335,61 @@ def extend_embedder_vocab(self, extended_vocab: Vocabulary) -> None:
             if isinstance(module, (Embedding, TextFieldEmbedder)):
                 module.extend_vocab(extended_vocab)
 
+    def inspection_dict(self,
+                        inspect: str = "parameters",
```
joelgrus: I don't like "stringly typed" parameters like this. as I said above, I'm not sure whether you actually want to distinguish the two, but if you do then make two different functions

allennlp/models/model.py
```diff
@@ -333,6 +335,61 @@ def extend_embedder_vocab(self, extended_vocab: Vocabulary) -> None:
...
+            Inspects "parameters" or "modules" of the model.
+        quite: bool, optional, (default = False)
+            If true, doesn't print the output, only returns the final dict.
+        """
+        if inspect not in ["parameters", "modules"]:
+            ConfigurationError("model inspection works with modules or parameters only")
+
+        items = self.named_modules() if inspect == "modules" else self.named_parameters()
+        module_dict = {model_path: module for model_path, module in items}
+        sorted_module_paths = sorted(module_dict.keys(), key=lambda path: path.count('.'))
```
joelgrus: this looks like you're sorting by depth? why?

allennlp/models/model.py
```diff
@@ -333,6 +335,61 @@ def extend_embedder_vocab(self, extended_vocab: Vocabulary) -> None:
...
+
+        def nested_set(dict_, keys, value):
```
joelgrus: it's not clear to me what this function does (the docstring is not helpful, and its operation is pretty opaque)

also, why is this a nested function?

allennlp/models/model.py
```diff
@@ -333,6 +335,61 @@ def extend_embedder_vocab(self, extended_vocab: Vocabulary) -> None:
...
+            module_class_name = module.__class__.__name__
+            if not module_path:
+                model_inspection_dict["__class__"] = module_class_name
+                continue
+            path_module_names = module_path.split(".")
+            value = {'__class__': module_class_name} if inspect == "modules" else "Parameter"
+            if path_module_names:
+                nested_set(model_inspection_dict, path_module_names, value)
+
+        def sorted_dict(dictionary):
```
joelgrus: as above, I'm not sure why this dict needs to be sorted?

joelgrus: or why this is a nested function

allennlp/models/model.py
```diff
@@ -333,6 +335,61 @@ def extend_embedder_vocab(self, extended_vocab: Vocabulary) -> None:
...
+        for module_path in sorted_module_paths:
+            module = module_dict[module_path]
+            module_class_name = module.__class__.__name__
+            if not module_path:
+                model_inspection_dict["__class__"] = module_class_name
+                continue
+            path_module_names = module_path.split(".")
+            value = {'__class__': module_class_name} if inspect == "modules" else "Parameter"
+            if path_module_names:
+                nested_set(model_inspection_dict, path_module_names, value)
```
joelgrus: the operation of this whole code block is extremely opaque to me, I don't really get what it's trying to do. (it's producing a dict of some format, but what the format is and why I don't get)

HarshTrivedi:
@joelgrus I will get back to your comments. But first, I need opinion on whether should I keep *both* modes are need or not? The output of modules inspection is closer to `print(model)` so I am fine skipping it altogether if you feel like.

joelgrus:
I don't have a strong opinion

allennlp/models/model.py
HarshTrivedi: oops!

HarshTrivedi:
Okay, I will make two utilities in `allennlp.nn.util`, `inspect_model` and make it inspect parameters and inspect models both. And add more comments throughout the code. Will ping you once done.

HarshTrivedi:
@joelgrus, Can you take look again.

I have removed modules inspection for this pr. Much of the logic in it was specifically for module and wasn't required for parameters. Besides, I haven't found it useful yet.

I have following follow-ups on this:

1. The current output of `inspect_parameters` is much more parsable than the way we log parameter tunability in [here](https://github.com/allenai/allennlp/blob/master/allennlp/training/trainer.py#L766-L773). Can I replace that code to use `inspect_parameters`?
2. Would it be worth dumping the model parameter inspection in serialization dir? It can be helpful to setup regexes at the time of fine-tuning / transferring modules etc.

matt-gardner:
@joelgrus, sorry to ping you on all of these, but I'm finding a bunch of old PRs that need follow up.

allennlp/nn/util.py
```diff
@@ -1325,3 +1326,37 @@ def uncombine_initial_dims(tensor: torch.Tensor, original_size: torch.Size) -> t
...
+    It prints it by default (optional) and returns the inspection dict. Eg. output::
+
+        {
+            "_text_field_embedder": {
+                "token_embedder_tokens": {
+                    "_projection": {
+                        "bias": "parameter:tunable",
+                        "weight": "parameter:tunable"
+                    },
+                    "weight": "parameter:frozen"
```
joelgrus: it seems like this docstring is not right? doesn't it just produce e.g. `"bias": "tunable"`?

HarshTrivedi: Oops! updated it.

PR #2468, opened 2019-01-31
Fixes unnecessary parameter clone in moving average.

yizhongw:
@joelgrus The original `Parameter.clone()` seems to consume much more memory than `Parameter.data.clone()`. The original one caused OOM when running a large model. But the new one could work. Also, I use the `tensor.copy_()` method to assign values when doing the moving average, which does not significantly reduce memory usage, but seems more reasonable.


joelgrus:
tensor.copy_ seems like a good idea. it's hard to see why `.data.clone()` would use less memory than `.clone()`, but if you say it works ...

PR #2476, opened 2019-02-02
Fix CORS error for react in config explorer

nelson-liu:
The config explorer is currently broken on master due to https://github.com/unpkg/unpkg.com/issues/176 . Here's a screenshot of what I see:

screen shot 2019-02-02 at 3 35 10 pm

The fix seems to be to remove the `crossorigin` attribute on the `<script>` tag. Things seems to load fine after doing so (also recommended in https://github.com/unpkg/unpkg.com/issues/176#issuecomment-459252472 ).


nelson-liu:
That's a lot of pressure, went back and tested it on some more browsers on a fresh pip install :) thanks @joelgrus

PR #2481, opened 2019-02-05
Mark QaNet test as flaky

matt-gardner:


allennlp/tests/models/reading_comprehension/qanet_test.py
```diff
@@ -39,6 +41,7 @@ def test_forward_pass_runs_correctly(self):
         assert span_end < self.instances[0].fields['passage'].sequence_length()
         assert isinstance(output_dict['best_span_str'][0], str)
 
+    @flaky
```
DeNeutoy: Is this flaky because of the comment in `test_batch_predictions_are_consistent` below?

matt-gardner: Yes, I believe it's the same issue.

PR #2484, opened 2019-02-07
Vae

mataney:
Implementation of Variational Autoencoder [(Kingma & Welling, 2014)](https://arxiv.org/pdf/1312.6114.pdf) for test generation as first described in [Bowman et al., 2015](https://arxiv.org/pdf/1511.06349.pdf).
I used the code published for https://arxiv.org/pdf/1802.02550.pdf (Kim et al., 2018) was used in the implementation of the VAE class.

Highlights of the PR:
Changing `trainer.py` to be able to receive more than a single loss term (As is required in VAEs)
`LossWeighters` class - When the loss is composed of more than one term, some sort of weighting is required. (Not sure about the name of this class tbh)
`VAE` class - receives an encoder and a decoder and the VAE logic is here.
`MaskedEncoder` class - This is a wrapper so we don't need to mask every time we encode (I'm not sure this is necessary)
`VariationalDecoder` class, this inherits a `Decoder` class I also added. I'm pretty sure there's there is an aware choice why not to add this `decoder` class. If you prefer not to add this parent class (and folder) when does this class should be?

I'm receiving results similar to the ones reported for `penn` and `yahoo questions` datasets.

Would love you comments.


matt-gardner:
Sorry, we've been pretty negligent in reviewing this PR; my only excuse is that the ACL deadline was keeping us very busy, and our meeting where we normally go over open issues / PRs and make sure nothing has fallen through the cracks hasn't happened in over a month.  @joelgrus, @DeNeutoy, @brendan-ai2, any of you (or anyone else) want to take this PR?  Also, @kernelmachine, how does this relate to what you've been working on?

matt-gardner:
@kernelmachine, you were going to talk to @mataney, right?  Did that happen?  What was the result?  I know you have your VAMPIRE codebase now.

DeNeutoy:
@kernelmachine please can you follow-up with @mataney about whether we still need/want this contribution, thanks!

kernelmachine:
@mataney -- thanks for your contribution and patience, and sorry it's taken so long to get back to you.
 
As mentioned above, we have been working on [VAMPIRE](https://github.com/allenai/vampire), a method for pretraining with variational autoencoders.

There are a few major differences between your PR and our codebase:

1. We focus on variational document models (Miao et al 2015), rather than variational seq2seq models (Bowman et al 2015).
2. We focus on using variational autoencoders for representation learning, as opposed to using them for generation.

Now, one path forward might be to integrate your code into the VAMPIRE repo as a one-stop shop for variational models in AllenNLP. However, because the VAMPIRE codebase is focused on a specific use-case (representation learning with document models under low-resource settings), I don't think it makes sense to integrate a sequence-to-sequence generation model. 

Furthermore, given that text generation is possibly the most well known use case for VAEs in NLP  (even though it has known issues), it is probably worth having a reference implementation in AllenNLP. 

A few high-level things I would recommend on this PR:

1) I think this PR would be better if it was self-contained as just a model and dataset reader. For example, it is possible to track all loss weights within the model as metrics and tensorboard values (see #2693), removing the need to touch trainer.py, tensorboard_writer, etc.  Furthermore, the loss summing (ie reconstruction loss + KLD) can also happen within the model.

2) I think you can make the `VariationalDecoder` class part of vae.py, since it is a module that's pretty specific to this VAE.

3) I think it would be more consistent with other models in AllenNLP to separate the TextFieldEmbedder from the encoder/decoder. So, no need for the MaskedEncoder class.

4) Is it possible for you to make a multiple smaller PRs (ie one for the model, one for the dataset reader, etc.)? That would make it much easier to review line-by-line.

mataney:
Hi @kernelmachine Thanks for your reply.

About your comments:
1. I agree, I think I either wasn't aware it is possible to track all loss weights within the model (My first attempt of adding code to AllenNLP 😄)or it wasn't possible at the time I added the PR. I will be much cleaner if it be self contained as you said.
2. and 3. Yes, you're absolutely right.
4. Yes, I believe you are right. This was probably a mistake not splitting all of these changes to different PRs. I think it will be best to close this PR, as the code is pretty stale by now, and create new (much) smaller ones.

Thanks, Matan.

PR #2488, opened 2019-02-08
Clarify `data_parallel` implementation.

brendan-ai2:
- It appears we're unnecessarily replicating the model to the source device, but this is not the case.
- Silence some deprecation warnings that were missed previously.


brendan-ai2:
Thanks!

PR #2490, opened 2019-02-08
change pins to bounds

### LINKED ISSUE #2483: Consider loosening dependency versions
I'm working on a python package and would like to include allennlp as a dependency. However, this is proving non trivial as allennlp [pins](https://github.com/allenai/allennlp/blob/master/setup.py#L103) many of its dependencies in the `setup.py` causing a few of these conflicts. It also looks like these pinned dependencies have [caused](https://github.com/allenai/allennlp/issues/1824) allennlp to have internally inconsistent dependencies in the past.

Ideally, all or nearly all dependencies in the install_requires in allennlp's `setup.py` would have version constraints instead of pins. This is the [recommended practice](https://packaging.python.org/discussions/install-requires-vs-requirements/) in python to avoid situations like these. If users need deterministic builds with allennlp, they can use allennlp's provided `requirements.txt` file. Allennlp also has a nice amount of test coverage, so this should increase confidence in loosening the dependency versions.

Loosening the dependency constraints would also make it easier for others to include allennlp as a dependency in their projects.

**Additional context**
Tools such as [poetry](https://github.com/sdispater/poetry) and [pipenv](https://github.com/pypa/pipenv) can help a lot with this type of dependency management and can detect whenever there are dependency conflicts. It takes a lot of load off of the developer, and I've had luck with them on my projects.

Thank you!

joelgrus:
Fixes #2483 (at least in part)

Using AllenNLP as a dependency is challenging because it pins many of its own dependencies, leading to dependency conflicts.  This PR loosens AllenNLP's dependency constraints so it can be more compatible with other libraries while still keeping the tests passing.

I tested this by building a docker image and running mypy and pytest.

It's possible there are other subtle reasons not to unpin some of these, speak up if that's the case.


schmmd:
@joelgrus can we merge this down?

joelgrus:
yes, there's a merge conflict now, let me fix it

PR #2491, opened 2019-02-08
Allow checkpointer to be initialized from params

mikerossgithub:
Allows custom Checkpointer to be loaded from params.
Trainer now takes an optional `checkpointer` parameter. 
Also added some unit tests for Checkpointer

### DIFF
--- a/allennlp/tests/training/checkpointer_test.py
+++ b/allennlp/tests/training/checkpointer_test.py
@@ -1,0 +1,105 @@
+# pylint: disable=invalid-name
+import os
+import re
+import time
+
+from allennlp.common.testing import AllenNlpTestCase
+from allennlp.training.checkpointer import Checkpointer
+from allennlp.common.params import Params
+from allennlp.training.trainer import Trainer
+from allennlp.common.checks import ConfigurationError
+
+
+class TestCheckpointer(AllenNlpTestCase):
+    def retrieve_and_delete_saved(self):
+        serialization_files = os.listdir(self.TEST_DIR)
+        model_checkpoints = [x for x in serialization_files if "model_state_epoch" in x]
+        found_model_epochs = [int(re.search(r"model_state_epoch_([0-9\.\-]+)\.th", x).group(1))
+                              for x in model_checkpoints]
+        for f in model_checkpoints:
+            os.remove(os.path.join(self.TEST_DIR, f))
+        training_checkpoints = [x for x in serialization_files if "training_state_epoch" in x]
+        found_training_epochs = [int(re.search(r"training_state_epoch_([0-9\.\-]+)\.th", x).group(1))
+                                 for x in training_checkpoints]
+        for f in training_checkpoints:
+            os.remove(os.path.join(self.TEST_DIR, f))
+        return sorted(found_model_epochs), sorted(found_training_epochs)
+
+    def test_default(self):
+        """
+        Tests against Trainer's default values for num_to_keep
+        """
+        default_num_to_keep = 20
+        num_epochs = 30
+        target = list(range(num_epochs - default_num_to_keep, num_epochs))
+
+        checkpointer = Checkpointer(serialization_dir=self.TEST_DIR)
+
+        for e in range(num_epochs):
+            checkpointer.save_checkpoint(epoch=e,
+                                         model_state={"epoch": e},
+                                         training_states={"epoch": e},
+                                         is_best_so_far=False)
+        models, training = self.retrieve_and_delete_saved()
+        assert models == training == target
+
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -69,6 +69,9 @@ ftfy
 # To use the BERT model
 pytorch-pretrained-bert==0.4.0
 
+# For checkpointer (if python < 3.7)
+dataclasses
```
joelgrus: please don't casually introduce new dependencies

mikerossgithub: Sure. Did not mean to be casual about it! :-)
Dataclasses are a built in feature of python 3.7, so I thought it would be safe to use. But it's easy for me to change the code to not use it. Would that be preferred?

joelgrus: as of right now allennlp only requires Python 3.6, so it's not safe to use 3.7-only features. it would be better not to use dataclasses.

joelgrus:
high level feedback: I don't find the usage of the modified `Checkpointer` very intuitive, and it's not clear to me that these changes would be broadly useful enough to make up for complicating its API.

instead, I suggest the following:

1. make `Checkpointer` a subclass of `FromParams`, so that it can be created from a configuration file
2. allow the `Trainer` constructor to take a (possibly redundant) optional `Checkpointer` parameter, so that anyone can specify their own custom `Checkpointer` (and raise a `ConfigurationError` if someone tries to specify checkpointing two different ways)

that's the PR.

3. then in your own code create a custom `Checkpointer` subclass that behaves however you like and use it.

mikerossgithub:
@joelgrus That makes sense, and will be a smaller change to the code base. I'll update the PR with your suggestions soon. Thank you.

mikerossgithub:
@joelgrus 
I have a working solution that only changes a few lines of code. Is this what you are suggesting?
After modifying `Checkpointer` to inherit from `FromParams` I made the following changes to `Trainer.from_params`:
```
num_serialized_models_to_keep = params.pop_int("num_serialized_models_to_keep", 20)
        keep_serialized_model_every_num_seconds = params.pop_int(
                "keep_serialized_model_every_num_seconds", None)
        checkpointer = params.pop("checkpointer", None)
        if checkpointer is not None:
            # read class from dotted path as in Django
            dotted_path = checkpointer.pop("type", "allennlp.training.checkpointer.Checkpointer")
            module_path, class_name = dotted_path.rsplit('.', 1)
            module = import_module(module_path)
            checkpointer_class = getattr(module, class_name)
            checkpointer = checkpointer_class.from_params(params=checkpointer, serialization_dir=serialization_dir)
```
To use this, I can use a jsonnet file like this:
```
"trainer": {
        "cuda_device": 0,
        "log_batch_size_period": 100,
        "checkpointer": {
            "type": "my.custom.module.CustomCheckpointer",
            "param1": 0,
            "param2": 2
        },
        "num_epochs": 8,
        "optimizer": "adam",
        "patience": 10
    }
```
I tested it and it works. But I am not sure if it is the preferred to pass custom classes in allennlp. Please advise. If its good, I can check in as an update to this PR or as a separate one.

joelgrus:
you don't need to do all that work, the standard command line tools support custom classes using the `--include-package` flag.

so you can just do 

```
checkpointer = Checkpointer.from_params(...)
```

(also handling the old way for backward compatibility)

and then when you train your model just do

```
allennlp train config.json -s /tmp/serialization --include-package my.custom.module
```

and then as long as your custom Checkpointer is registered under the same name you use in your config file it should just work.

make sense?

mikerossgithub:
I updated the PR per your suggestions. All of the pylint and unit tests pass. But it looks like maybe the remote setup for sniff tests doesnt have enough space? When I run these locally, they pass after downloading large files.
```
Step 10/10: Sniff Tests (Command Line)
...
[Step 10/10] allennlp/tests/models/sniff_test.py::SniffTest::test_semantic_role_labeling FAILED [ 85%]
[Step 10/10] allennlp/tests/models/sniff_test.py::SniffTest::test_textual_entailment FAILED [100%]
[Step 10/10] INTERNALERROR> OSError: [Errno 28] No space left on device
```

matt-gardner:
That's an issue with the machine that runs our CI. We're working on it.

mikerossgithub:
codecov is requesting additional unit test coverage. I can provide that if necessary, but wanted to double check that the current approach is viable. Is there anything else needed before finalizing this PR?

joelgrus:
don't worry about the codecov, 78% is fine.

I'll try to look at this PR today or tomorrow.

mikerossgithub:
Ok, I edited the comments. Thanks for the review.

PR #2494, opened 2019-02-11
Less restrictive token characters indexer

matt-peters:
Allows to use any Tokenizer, not just CharactersTokenizer in TokenCharactersIndexer

### DIFF
--- a/allennlp/data/token_indexers/token_characters_indexer.py
+++ b/allennlp/data/token_indexers/token_characters_indexer.py
@@ -57,6 +57,8 @@ class TokenCharactersIndexer(TokenIndexer[List[int]]):
         self._min_padding_length = min_padding_length
         self._namespace = namespace
         if tokenizer is None:
+            # NOTE: (matt-peters) when removing character_tokenizer in the future, the default
+            # value for tokenizer should change to CharacterTokenizer() from None.
             warnings.warn("character_tokenizer is depreciated and will be "
                           "removed before version 1.0.  Use tokenizer instead.",
                           UserWarning)
--- a/allennlp/data/token_indexers/token_characters_indexer.py
+++ b/allennlp/data/token_indexers/token_characters_indexer.py
@@ -22,11 +22,15 @@ class TokenCharactersIndexer(TokenIndexer[List[int]]):
     namespace : ``str``, optional (default=``token_characters``)
         We will use this namespace in the :class:`Vocabulary` to map the characters in each token
         to indices.
-    character_tokenizer : ``Tokenizer``, optional (default=``CharacterTokenizer()``)
+    character_tokenizer : ``CharacterTokenizer``, optional (default=``CharacterTokenizer()``)
         We use a :class:`CharacterTokenizer` to handle splitting tokens into characters, as it has
         options for byte encoding and other things.  The default here is to instantiate a
         ``CharacterTokenizer`` with its default parameters, which uses unicode characters and
-        retains casing.
+        retains casing.  Note: this argument is depreciated in favor of tokenizer and will be
+        removed before version 1.0.
+    tokenizer : ``Tokenizer``, optional (default=``None``)
+        If provided, it is used instead of character_tokenizer, otherwise we default to
+        character_tokenizer.
     start_tokens : ``List[str]``, optional (default=``None``)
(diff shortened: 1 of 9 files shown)


allennlp/data/token_indexers/token_characters_indexer.py
```diff
@@ -38,7 +38,7 @@ class TokenCharactersIndexer(TokenIndexer[List[int]]):
     # pylint: disable=no-self-use
     def __init__(self,
                  namespace: str = 'token_characters',
-                 character_tokenizer: CharacterTokenizer = CharacterTokenizer(),
+                 character_tokenizer: Tokenizer = CharacterTokenizer(),
```
joelgrus: this is failing tests because of a bug in the automatic `FromParams`. Basically it's not handling default values correctly in the case where the type is itself `from_params`. Let me see if I can fix that.

joelgrus: wait, I take that back, the `FromParams` code is right, let me look again and see why this is failing tests

matt-peters:
It's failing tests because it breaks backward compatibility.  There is a specific hard coded tokenizer in the `TokenCharactersIndexer`, so in existing config files the type of this tokenizer isn't specified.  My use case requires a different character `Tokenizer` to use with `TokenCharactersIndexer` -- any ideas how to best accomplish this?

joelgrus:
here's probably the safest thing to do:

(1) give the constructor a second `tokenizer: Tokenizer` parameter, with default value None
(2) have the constructor prefer the new argument if it's provided; otherwise use the first one.
(3) raise a `DeprecationWarning` (with a specified timeframe) if someone provides the argument the old way
(4) (not as part of this PR obviously) update all of our pretrained models / test fixtures :frowning_face: 

thoughts?

cc: @brendan-ai2 who has been thinking about this sort of thing

matt-peters:
Added the extra tokenizer argument as suggested. Right now, the warning uses a `UserWarning` so tests will continue to pass, until all of the test fixtures are updated in a future PR.

joelgrus:
looks good, can you add two quick unit tests

(1) check that it raises a warning if you don't provide tokenizer
(2) check that it prefers `tokenizer` to `character_tokenizer` when both are there?

thanks

matt-peters:
Added the tests.

allennlp/data/token_indexers/token_characters_indexer.py
```diff
@@ -51,7 +56,16 @@ def __init__(self,
                           UserWarning)
         self._min_padding_length = min_padding_length
         self._namespace = namespace
-        self._character_tokenizer = character_tokenizer
+        self._character_tokenizer: Tokenizer = None
+        if tokenizer is None:
+            # NOTE: (matt-peters) when removing character_tokenizer in the future, the default
+            # value for tokenizer should change to CharacterTokenizer() from None.
+            warnings.warn("character_tokenizer is depreciated and will be "
```
matt-gardner: Nit: "deprecated", not "depreciated"

matt-gardner:
I very much want to revert this change.  It defeats the entire purpose of the `DeprecationWarning` checks that we added by using a `UserWarning` instead.

allennlp/data/token_indexers/token_characters_indexer.py
```diff
@@ -26,7 +26,11 @@ class TokenCharactersIndexer(TokenIndexer[List[int]]):
         We use a :class:`CharacterTokenizer` to handle splitting tokens into characters, as it has
         options for byte encoding and other things.  The default here is to instantiate a
         ``CharacterTokenizer`` with its default parameters, which uses unicode characters and
-        retains casing.
+        retains casing.  Note: this argument is depreciated in favor of tokenizer and will be
+        removed before version 1.0.
```
brendan-ai2: @matt-peters, @joelgrus Can we just leave this indefinitely without an explicit deprecation warning? I'm not sure the churn of deleting this is worth it. If we _really_ want to delete it, let's provide a guarantee like "We will remove this argument _not_ before version i.j and YYYY/MM." That'll be more useful to users.

brendan-ai2: (That will resolve the point @matt-gardner raised w.r.t. warnings as well.)

PR #2479, opened 2019-02-04
WIP: Make warnings errors and filter library warnings from pytest

schmmd:
I was getting some strange errors locally and wanted to push to GitHub to run CI.  I created a separate PR so I didn't pollute https://github.com/allenai/allennlp/pull/2282 with changes that might not need to be kept.

PR #2497, opened 2019-02-11
Revert "Less restrictive token characters indexer"

matt-gardner:
Reverts allenai/allennlp#2494

We just merged a huge involved PR to get rid of all of our warnings when we run tests (https://github.com/allenai/allennlp/pull/2479, https://github.com/allenai/allennlp/pull/2282).  I won't repeat the justification for this here - anyone interested can go back and read that history.  The bottom line: it is intentionally difficult to introduce new warnings into our code.  The original PR that introduces a warning must also do whatever cleanup is necessary for it, so that the cleanup has a clear owner and actually gets done.  This PR bypassed that by using a `UserWarning` instead of a `DeprecationWarning`, so I am reverting it.  I had a discussion with @joelgrus, who hadn't followed all of the discussion involved in #2282, and approved the PR without realizing that we were intentionally disallowing this.  Per @brendan-ai2's arguments, we are also intentionally slow at making API changes.

For @matt-peters, who made this change, there are three ways forward:
1. (Easiest) Just copy the code for the class you want and make the change, registering it as something else.  For a research project, this is generally the right thing to do.  Once you have a clear, compelling use case, we can consider an API change to handle it.
2. Make the change without adding a warning.  I don't much like this solution, because it makes the API muddy without giving a warning about the muddiness.
3. Use a `DeprecationWarning` instead of a `UserWarning`, and make all of the tests pass.

### DIFF
--- a/allennlp/data/token_indexers/token_characters_indexer.py
+++ b/allennlp/data/token_indexers/token_characters_indexer.py
@@ -9,7 +9,7 @@ from allennlp.common.util import pad_sequence_to_length
 from allennlp.data.tokenizers.token import Token
 from allennlp.data.token_indexers.token_indexer import TokenIndexer
 from allennlp.data.vocabulary import Vocabulary
-from allennlp.data.tokenizers import Tokenizer, CharacterTokenizer
+from allennlp.data.tokenizers.character_tokenizer import CharacterTokenizer
 
 
 @TokenIndexer.register("characters")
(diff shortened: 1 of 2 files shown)


matt-peters:
I hadn't seen or followed the previous PRs, so was unaware of the restrictions against adding new warnings.  I'll almost certainly take the easiest path here and hack a way around it.

matt-gardner:
No worries @matt-peters, I didn't think you meant to do anything bad.  But yeah, the easiest path here is probably best from a library-stability perspective.

PR #2499, opened 2019-02-11
remove notebooks and juptyer dependency

DeNeutoy:
Motivation: 
- Library is more mature, there are more complete tutorials elsewhere
- Jupyter is a massive dependency and effects all users installing from master, which is more than typical because allennlp is fast moving and used for research.
- Opened #2500 to move these into a web tutorial at some point in the future.


HarshTrivedi:
I think it would be worth mentioning in [tutorials/README.md](https://github.com/allenai/allennlp/tree/56c11e5343b672c029deb6f648258ced4fb60efd/tutorials/notebooks), that there are notebook tutorials which can be *viewed* online [here](https://github.com/allenai/allennlp/tree/56c11e5343b672c029deb6f648258ced4fb60efd/tutorials/notebooks), at least until ported to website. New users might not know that notebook tutorials existed, and imo they were quite useful in the beginning.

PR #2501, opened 2019-02-12
Allow vocab+embedding extension in evaluate command.

### LINKED ISSUE #239: Allow for different handling of OOV words

### LINKED ISSUE #327: Expand vocabulary in `evaluate`

HarshTrivedi:
This should close #327 and close #239.

I've been somewhat superficial with tests, but since `evaluate_from_args` exposes only metrics, it's difficult to test more. Since used extension APIs have been tested in previous PRs and also in `fine_tune_test`, I think it should be ok.


allennlp/commands/evaluate.py
```diff
@@ -112,6 +145,13 @@ def evaluate_from_args(args: argparse.Namespace) -> Dict[str, Any]:
     logger.info("Reading evaluation data from %s", evaluation_data_path)
     instances = dataset_reader.read(evaluation_data_path)
 
+    embedding_sources: Dict[str, str] = (json.loads(args.embedding_sources_mapping)
+                                         if args.embedding_sources_mapping else {})
+    if args.extend_vocab:
+        logger.info("Vocabulary is being extended with test instances.")
+        model.vocab.extend_from_instances(Params({}), instances=instances)
+        model.extend_embedder_vocab(model.vocab, embedding_sources)
```
HarshTrivedi: Doesn't belong to this PR, but I feel, `extend_embedder_vocab` API should have been just

```python
extend_embedder_vocab(self, embedding_sources_mapping: Dict[str, str] = None):
```
Passing `extended_vocab` isn't necessary and potentially unsafe. We only should allow vocab that's for sure extended from original model.vocab. Allowing to pass arbitrary `extended_vocab` can put user in weird issues if one tries to use it API as standalone and doesn't know that extended_vocab must be extended from original model.vocab.

Since vocab extension is anyway happening in-place, following should work just fine:

```python
model.vocab.extend_from_instances(instances)
model.extend_embedder_vocab(embedding_sources_mapping)
```

And is cleaner+safer than current:
```python
model.vocab.extend_from_instances(instances)
model.extend_embedder_vocab(model.vocab, embedding_sources_mapping)
```

HarshTrivedi:
This should also (almost) close #239 

> * At test time, see if the OOV token is in glove, and use that embedding instead

This would be vocab+embedding extension *with* passing `embedding-source-mapping`.

> * At test time, use random vectors for each unique OOV token, as suggested [here](https://www.semanticscholar.org/paper/A-Comparative-Study-of-Word-Embeddings-for-Reading-Dhingra-Liu/3ec37205c9201fc891ab51da200e361fdc34bfb3)

This would be (almost) vocab+embedding extension *without* passing `embedding-source-mapping`. Currently, the default is that `embedding-file` will be used (or at least will be tried to be used) if originally `embedding-file` was used. We don't have explicit control on toggling this and am not sure if it's worth complicating this.

allennlp/commands/evaluate.py
matt-gardner: I agree with this.  This whole embedding extension hasn't been released yet, so it's fine to simplify this in a subsequent PR.

PR #2504, opened 2019-02-12
Actually use given padding_index param to calculate embeddings

manuyavuz:
I discovered this when I'm trying to use Elmo with vocab cache. My inputs were padded using index `0`,  which is also internally set as same here: 

https://github.com/allenai/allennlp/blob/39413f220f2e682001299f8975a38f2221fac840/allennlp/modules/elmo.py#L685

However, `Embedding` class does not pass given `padding_index` parameter to PyTorch's `embedding` function, and PyTorch uses `-1` as default padding index. Therefore me code was getting out of bounds error.

This change fixed the issue for me.

Any other needs to be merged?


DeNeutoy:
Hi, thanks for the contribution - would you mind adding a tiny test case to this file:

https://github.com/allenai/allennlp/blob/master/allennlp/tests/modules/token_embedders/embedding_test.py

manuyavuz:
Of course, I'll look into that soon.

matt-gardner:
@manuyavuz, any intention of finishing this?

manuyavuz:
I'm so sorry, I was really overwhelmed, could not turn back to this :(

PR #2510, opened 2019-02-13
make archival take an optional output path

DeNeutoy:


allennlp/models/archival.py
```diff
@@ -104,6 +105,9 @@ def archive_model(serialization_dir: str,
         A mapping {flattened_key -> filename} of supplementary files to include
         in the archive. That is, if you wanted to include ``params['model']['weights']``
         then you would specify the key as `"model.weights"`.
+    archive_path : ``str``, optional, (default = None)
+        A path to serialize the model to. The default is "model.tar.gz" inside the
+        serialization_dir.
```
joelgrus: I am concerned that it's easy to mess up and do e.g. `archive_path=/tmp/models/` when what you really want is `archive_path=/tmp/models/model0.tar.gz`.

I don't know if the solution is to give this variable a different name, or to be more explicit in the comment, or to check below that `archive_path` isn't a directory, or something else. thoughts?

DeNeutoy: I added a check if it's a directory.

PR #2511, opened 2019-02-13
Enable Pruner class to keep different number of items for different entries in minibatch.

dwadden:
This is a somewhat special case. There are occasionally situations where entries in a minibatch need to be, e.g., ordered sentences from the same document. Coreference resolution is an example. In AllenNLP, an entire document is considered as a single entry, but for very long documents (or on low-memory machines) it might be necessary to split up the document into minibatches during training.
In this situation, smart batching can't be used, and entries in the same minibatch may have widely varying lengths. Keeping the same number of span candidates for each entry may not be desirable.
This PR does the same thing as before if an integer is passed to `num_items_to_keep`. If a tensor is passed instead, it keeps the desired number of items for each minibatch entry.


dwadden:
Thanks for merging!

PR #2513, opened 2019-02-14
Interactive beam search

joelgrus:
this is the first step toward incorporating interactive beam search into our demo. at a high level here's how it works:

1. there is a new class `InteractiveBeamSearch` that's a drop-in replacement for `state_machines.BeamSearch`. it differs in two primary ways. The first is that it allows you to constrain the beam search to follow a specified initial path. The second is that it explicitly keeps track of the top k choices it had at each beam step.

2. you could imagine a change to the wikitables parser predictor that does the following:
* accepts an extra "initial sequence" parameter
* replaces the model's BeamSearch with an `InteractiveBeamSearch` that's forced down that path
* adds the InteractiveBeamSearch's "top k choices" to its result

~~I have not made this change to the predictor yet, but I added a test to wikitables_parser_test.py that basically does this manually.~~
I made this change to the predictor.

3. then the idea is that the web demo would run beam search with (initially) no constraints. it would get back the "choices" and provide those to the user. then the user could click on one of those choices to force the beam up to that point, which would get a new result, new choices, and so on. I haven't fully thought through the UX for this, I wanted to get the backend working first.

Some less than ideal things:

* it seems like almost every model that we have that uses BeamSearch uses it differently. in particular, this `InteractiveBeamSearch` drop-in class won't help with a lot of them (at least, not without some serious refactoring). Some of them do beam search manually, some use `nn.BeamSearch`. But it will work with the Wikitables Parser, which I guess is the "launch customer"

* the beam search here operates on rules, and the corresponding id <-> rule is handled in (I think) a wikitables specific way; basically I have to reach into the `Instance` to get the mapping. so this part is not as general as I'd like. but that's not that bad.


joelgrus:
ok, I moved all of this functionality into the wikitables parser predictor.

here's an interactive session that shows you how it works, you'll have to imagine a web front end that allows this:

```
from allennlp.models.archival import load_archive
from allennlp.predictors import Predictor

archive_file = "allennlp/tests/fixtures/semantic_parsing/wikitables/serialization/model.tar.gz"
archive = load_archive(archive_file)
predictor = Predictor.from_archive(archive, 'wikitables-parser')
inputs = {"question": "Who is 18 years old?", "table": "Name\tAge\nShallan\t16\nKaladin\t18"}
result = predictor.predict_json(inputs)
print(result['choices'])
print(result['best_action_sequence'])

# {1: [(-1.4312206506729126, '@start@ -> r'), (-1.537883996963501, '@start@ -> n'), (-1.5584005117416382, '@start@ -> p'), (-1.6859612464904785, '@start@ -> c'), (-1.8943967819213867, '@start@ -> d')], 2: [(-2.8632636070251465, 'p -> [<#1,<#1,#1>>, p, p]'), (-2.900278091430664, 'p -> [<c,p>, c]'), (-2.963406562805176, 'p -> [<#1,#1>, p]')], 3: [(-2.900278091430664, '<c,p> -> [<<#1,#2>,<#2,#1>>, <p,c>]'), (-2.963406562805176, '<#1,#1> -> !='), (-3.529099225997925, '<#1,<#1,#1>> -> and')], 4: [(-2.900278091430664, '<<#1,#2>,<#2,#1>> -> reverse'), (-4.287175178527832, 'p -> [<c,p>, c]'), (-4.2923994064331055, 'p -> [<#1,<#1,#1>>, p, p]')], 5: [(-2.900278091430664, '<p,c> -> fb:cell.cell.part'), (-4.287175178527832, '<c,p> -> [<<#1,#2>,<#2,#1>>, <p,c>]'), (-4.953691482543945, '<#1,<#1,#1>> -> and')], 6: [(-4.287175178527832, '<<#1,#2>,<#2,#1>> -> reverse'), (-4.743768692016602, 'c -> fb:cell.16'), (-4.842709064483643, 'c -> fb:cell.18')], 7: [(-4.287175178527832, '<p,c> -> fb:cell.cell.part')], 8: [(-6.130180358886719, 'c -> fb:cell.16'), (-6.229133605957031, 'c -> fb:cell.18'), (-6.44895601272583, 'c -> fb:cell.kaladin')]}
# ['@start@ -> p', 'p -> [<c,p>, c]', '<c,p> -> [<<#1,#2>,<#2,#1>>, <p,c>]', '<<#1,#2>,<#2,#1>> -> reverse', '<p,c> -> fb:cell.cell.part', 'c -> fb:cell.16']

# Now let's force '@start@ -> n'

inputs['initial_sequence'] = ['@start@ -> n']
result = predictor.predict_json(inputs)
print(result['choices'])
print(result['best_action_sequence'])

# {1: [(-1.4312206506729126, '@start@ -> r'), (-1.537883996963501, '@start@ -> n'), (-1.5584005117416382, '@start@ -> p'), (-1.6859612464904785, '@start@ -> c'), (-1.8943967819213867, '@start@ -> d')], 2: [(-3.6121275424957275, 'n -> 1'), (-3.625232458114624, 'n -> 0'), (-3.651451349258423, 'n -> -1')]}
# ['@start@ -> n', 'n -> 1']

# That wasn't promising, instead let's choose a different second step:

inputs['initial_sequence'] = ['@start@ -> p', 'p -> [<#1,<#1,#1>>, p, p]']
result = predictor.predict_json(inputs)
print(result['choices'])
print(result['best_action_sequence'])

# {1: [(-1.4312206506729126, '@start@ -> r'), (-1.537883996963501, '@start@ -> n'), (-1.5584005117416382, '@start@ -> p'), (-1.6859612464904785, '@start@ -> c'), (-1.8943967819213867, '@start@ -> d')], 2: [(-2.8632636070251465, 'p -> [<#1,<#1,#1>>, p, p]'), (-2.900277853012085, 'p -> [<c,p>, c]'), (-2.963406562805176, 'p -> [<#1,#1>, p]')], 3: [(-3.529099464416504, '<#1,<#1,#1>> -> and'), (-3.584489345550537, '<#1,<#1,#1>> -> or')], 4: [(-4.855774879455566, 'p -> [<#1,<#1,#1>>, p, p]'), (-4.856836318969727, 'p -> [<c,p>, c]')], 5: [(-4.856836318969727, '<c,p> -> [<<#1,#2>,<#2,#1>>, <p,c>]'), (-5.517705917358398, '<#1,<#1,#1>> -> and')], 6: [(-4.856836318969727, '<<#1,#2>,<#2,#1>> -> reverse'), (-6.841122627258301, 'p -> [<c,p>, c]')], 7: [(-4.856836318969727, '<p,c> -> fb:cell.cell.part'), (-6.841122627258301, '<c,p> -> [<<#1,#2>,<#2,#1>>, <p,c>]')], 8: [(-6.699860095977783, 'c -> fb:cell.16'), (-6.798811912536621, 'c -> fb:cell.18')], 9: [(-8.0228853225708, 'p -> [<c,p>, c]'), (-8.03402042388916, 'p -> [<#1,<#1,#1>>, p, p]')], 10: [(-8.0228853225708, '<c,p> -> [<<#1,#2>,<#2,#1>>, <p,c>]'), (-8.693674087524414, '<#1,<#1,#1>> -> and')], 11: [(-8.0228853225708, '<<#1,#2>,<#2,#1>> -> reverse'), (-10.01498031616211, 'p -> [<c,p>, c]')], 12: [(-8.0228853225708, '<p,c> -> fb:cell.cell.part'), (-10.01498031616211, '<c,p> -> [<<#1,#2>,<#2,#1>>, <p,c>]')], 13: [(-9.865593910217285, 'c -> fb:cell.16'), (-9.964548110961914, 'c -> fb:cell.18')]}
# ['@start@ -> p', 'p -> [<#1,<#1,#1>>, p, p]', '<#1,<#1,#1>> -> and', 'p -> [<c,p>, c]', '<c,p> -> [<<#1,#2>,<#2,#1>>, <p,c>]', '<<#1,#2>,<#2,#1>> -> reverse', '<p,c> -> fb:cell.cell.part', 'c -> fb:cell.16', 'p -> [<c,p>, c]', '<c,p> -> [<<#1,#2>,<#2,#1>>, <p,c>]', '<<#1,#2>,<#2,#1>> -> reverse', '<p,c> -> fb:cell.cell.part', 'c -> fb:cell.16']
```

joelgrus:
don't review this yet, I'm going to try to do the same for `nn.BeamSearch` and that might involve some changes here

joelgrus:
I decided not to do this for `nn.BeamSearch`, so this is ready for review

allennlp/state_machines/interactive_beam_search.py
```diff
@@ -0,0 +1,143 @@
...
+        super().__init__(beam_size, per_node_beam_size)
+        if initial_sequence is not None:
+            # construct_prefix_tree wants a tensor of shape (batch_size, num_sequences, sequence_length)
+            # so we need to add the first two dimensions in. This returns a list, but we're assuming
+            # batch size 1, so we extract the first element.
+            self._allowed_transitions = util.construct_prefix_tree(initial_sequence.view(1, 1, -1))[0]
+        else:
+            self._allowed_transitions = {}
+
+        self.beams: List[List[Tuple[float, List[int]]]] = []
```
matt-gardner: I find "beams" here a little jarring, as the original meaning of "beam search" was that there was one beam containing multiple paths or hypotheses, and calling each path a "beam" is a bit odd to me.  But that usage is growing, so maybe I should just deal with it.

allennlp/state_machines/interactive_beam_search.py
```diff
@@ -0,0 +1,143 @@
...
+from typing import Dict, List, Tuple
+
+import torch
+
+from allennlp.state_machines import util
+from allennlp.state_machines.beam_search import BeamSearch, StateType
+from allennlp.state_machines.transition_functions import TransitionFunction
+
+
+class InteractiveBeamSearch(BeamSearch[StateType]):
```
matt-gardner: So, now it looks like this code is basically identical to `BeamSearch`, except you have an optional constrained prefix.  This is like a 10 line addition to `BeamSearch`, and it would be generally useful in some cases (e.g., if you know the answer type at training time, you can constrain the first decision).  What do you think of just adding this to that code, and removing this additional class altogether?

The one thing I'm not sure of is your `self.beams` field, which looks like it keeps around the history at every timestep, instead of at the end.  We could add that, too, with something like a `keep_beam_history` parameter.

allennlp/state_machines/interactive_beam_search.py
joelgrus: in this case the usage is sort of right? because self.beams[0] is a "beam" (that is, a list of the scored action histories at time 0). I guess really this is "beam_snapshots" or something like that, but it's not just *a* beam.

allennlp/predictors/wikitables_parser.py
```diff
@@ -66,6 +68,51 @@ def _json_to_instance(self, json_dict: JsonDict) -> Instance:
...
+        instance = self._json_to_instance(inputs)
+
+        # Get the rules out of the instance
+        index_to_rule = [production_rule_field.rule
+                         for production_rule_field in instance.fields['actions'].field_list]
+        rule_to_index = {rule: i for i, rule in enumerate(index_to_rule)}
+
+        # A sequence of strings to force, then convert them to ints
+        initial_tokens = inputs.get("initial_sequence", [])
+        # pylint: disable=not-callable
```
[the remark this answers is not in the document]
joelgrus: I couldn't find one

allennlp/predictors/wikitables_parser.py
```diff
@@ -66,6 +68,51 @@ def _json_to_instance(self, json_dict: JsonDict) -> Instance:
...
+        initial_sequence = torch.tensor([rule_to_index[token] for token in initial_tokens])
```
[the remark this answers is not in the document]
joelgrus: fixed

allennlp/predictors/wikitables_parser.py
```diff
@@ -66,6 +68,51 @@ def _json_to_instance(self, json_dict: JsonDict) -> Instance:
...
+        original_beam_search = self._model._beam_search
+        beam_size = getattr(original_beam_search, "_beam_size")
```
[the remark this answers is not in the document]
joelgrus: no longer relevant

allennlp/predictors/wikitables_parser.py
```diff
@@ -66,6 +68,51 @@ def _json_to_instance(self, json_dict: JsonDict) -> Instance:
...
+        interactive_beam_search: InteractiveBeamSearch = InteractiveBeamSearch(beam_size,
```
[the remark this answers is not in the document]
joelgrus: no longer relevant

allennlp/state_machines/interactive_beam_search.py
joelgrus: done

joelgrus:
ok, I addressed your comments and cleaned up the code quite a bit, this is ready for another look

allennlp/state_machines/beam_search.py
```diff
@@ -33,15 +36,38 @@ class BeamSearch(FromParams, Generic[StateType]):
...
+        if initial_sequence is not None:
+            # construct_prefix_tree wants a tensor of shape (batch_size, num_sequences, sequence_length)
+            # so we need to add the first two dimensions in. This returns a list, but we're assuming
+            # batch size 1, so we extract the first element.
+            self._allowed_transitions = util.construct_prefix_tree(initial_sequence.view(1, 1, -1))[0]
+        else:
+            self._allowed_transitions = None
+
+        if keep_beam_details:
+            self.beams: List[List[Tuple[float, List[int]]]] = []
```
matt-gardner: I was pretty confused about what's going on here, because you're mixing batch index and number of steps below, and it seemed really hard to reason about.  Then I realized that you're assuming you only have batch size of 1 when you keep this around.  I'd probably make this a dictionary going from batch index to this list.  Monstrous type, but at least it doesn't break if you have batched computation.  Or have a major caveat somewhere that this is only intended for use with batch size 1.  This should go in the docstring for `keep_beam_details`.  Related: that parameter and `initial_sequence` need to be added to the class docstring.

Also, regarding terminology: you're right that this keeping snapshots of the beam at every timestep, so something like "beams" isn't as funny sounding as it otherwise might be.  But maybe `beam_snapshots` or something similar would be more obvious?

Issue #2097, opened 2018-11-25
Any plans for a generic Decoder API for seq2seq models?


sai-prasanna:
I am primarily working on seq2seq models. I currently use a hack-ish way of importing fairseq into an allennlp model and wrapping some objects to connect the APIs. I thought I could spend some time to implement a pluggable seq2seq model which can use multiple encoder/decoders.

Our current basic seq2seq model has an LSTMCell hardcoded inside of it, which makes it unusable for cases like a Gated CNN decoder, transformer decoder etc. 

1. Would it make sense to abstract out decoder as something separate from the seq2seq model?

2. Any opinion on what the abstract decoder API should consist of? since there are some specialized attributes with regards to the inputs required by the decoders. Some of these "fancy" decoders need extra values other than Seq2Seq's output, e.g, GatedCNN requires outputs added and scaled with source embedding and without adding it, etc. 

3. Should the current implemented Seq2Seq encoders return additional values that are needed by the decoder in a dictionary? 

With regards to the attention mechanism, since it differs for Transformer, GatedCNN etc I guess each decoder deriving the abstract class can figure that out.

matt-gardner:
I think it'd be great to have better seq2seq functionality implemented in AllenNLP, and it's not something any of us are working on, so contributions would be great here.  I think the first step would be to write a simple design document for what abstractions you're thinking of adding and how they interact with each other and the other abstractions in the library (key word "simple" - this doesn't have to be anything fancy).  That could either be posted directly here, or as a gist, or google doc, or somewhere that we could look at it and comment on it.

sai-prasanna:
@maxdel  I concur with the pros and cons. I wish we can convince fairseq team to use allennlp, but that might mostly not work out. My need currently is to try out stuff like elmo in encoder which as you pointed out doesn't work well with fairseq's encoder. 

Points to keep in mind while designing the API for decoding

1. Make it play well with caching for incremental decoding, maybe we can lift a page out of fairseq's way of passing the caching dictionary in each decoding step, instead of stateful properties inside decoder like opennmt

2. Make it flexible (like opennmt , unlike fairseq) so that one can potentially combine different encoders with different decoders.

3. Keep beam search in mind.

btw can you share the code, to compare notes on how you wrapped things, I do an ugly hack to replace fairseq's pytorch embedding with allennlp's Embedding object, and a wrapper for dictionary.

epwalsh:
[...]

Issue #2270, opened 2019-01-03
Remove / move internal transformer implementations


brendan-ai2:
We currently have a handful of transformer implementations. At least the following:

https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2seq_encoders/multi_head_self_attention.py
https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2seq_encoders/bidirectional_language_model_transformer.py#L139
https://github.com/allenai/allennlp/blob/master/allennlp/modules/openai_transformer.py

We should remove these and rely on an external implementation. Perhaps, Pytorch's `nn.Transformer`. This will (very likely) be a breaking change for our existing models.

schmmd:
@matt-peters during issue review it came up that you might propose standardizing on the openai transformer (assuming experiments work out).  Is that accurate?

nelson-liu:
FWIW, looks like the new PyTorch 1.1 includes a [`torch.nn.MultiheadAttention`](https://pytorch.org/docs/stable/nn.html?highlight=multiheadattention#torch.nn.MultiheadAttention)

maxdel:
Moreover, this module was landed as a part of [this WIP Transformer PR](https://github.com/pytorch/pytorch/pull/19274)

zhangguanheng66:
Thanks @maxdel, we have put together a PR for Transformer in PyTorch. Feel free to comment. 

The transformer module is supported by torch.nn.MultiheadAttention. We have a plan to improve the efficiency of this operator in the near future. @cpuhrsch.

cpuhrsch:
Hey all, please let us know how we can help you guys deduplicate some of your work and better support this library overall!

We're eager to provide a good set of tools or a full implementation of what you need for your Transformers.

Also, if you have more general feature requests, we'd be eager to hear about them as well.

sai-prasanna:
@cpuhrsch Need some help figuring out the decoder abstraction - https://github.com/allenai/allennlp/pull/2517 .

nelson-liu:
The new version of pytorch includes a [`nn.Transformer`](https://pytorch.org/docs/stable/nn.html?highlight=transformer#torch.nn.Transformer) module

zhangguanheng66:
Hello everyone, I just want to check if we could provide any help to apply `nn.Transformer` module. We want to better support AllenNLP and avoid the duplicate work. We could implement more features if you feel necessary to better serve this repo.

brendan-ai2:
[...]

PR #2517, opened 2019-02-16
Seq2Seq model decomposition

generall:
Hi, this is work-in-progress pull request for my attempt to decompose monolith seq2seq and enable generic decoder module. 
Initial issue discussion could be found here https://github.com/allenai/allennlp/issues/2097
There is also work notes on this here https://github.com/epwalsh/allennlp/pull/3
I am looking for some feedback on my changes. One of the questions is how should we support backward compatibility since module parameters changed.


allennlp/modules/seq2seq_decoders/decoder_cell.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        return self._decoding_dim
+
+    def forward(self,
+                previous_steps_predictions: torch.Tensor,
+                encoder_outputs: torch.Tensor,
+                source_mask: torch.Tensor,
+                decoder_hidden: torch.Tensor,
+                decoder_context: torch.Tensor
+                ) -> Tuple[torch.Tensor, torch.Tensor]:
```
matt-gardner: I don't think this is the right return type - this assumes you have an LSTM.  You want this base class to be more general than that.  Unfortunately, this also affects the inputs to this function, too.  You could maybe get away with some generics here:

```python
# At the top of the file:
CellState = TypeVar('CellState')

# class definition
class DecoderCell(torch.nn.Module, Registrable, Generic[CellState]):

# at this line:
def forward(self,
            previous_steps_predictions: torch.Tensor,
            encoder_outputs: torch.Tensor,
            source_mask: torch.Tensor,
            previous_state: CellState) -> CellState:
```

allennlp/modules/seq2seq_decoders/seq_decoder.py
```diff
@@ -0,0 +1,201 @@
...
+from allennlp.data import Vocabulary
+from allennlp.models import Model
+from allennlp.modules import Embedding
+from allennlp.modules.seq2seq_decoders.decoder_cell import DecoderCell
+from allennlp.nn import util
+from allennlp.nn.beam_search import BeamSearch
+from allennlp.training.metrics import Metric, BLEU
+
+
+class SeqDecoder(Model):
```
matt-gardner: Why does this inherit from `Model`, instead of just `Module`?  The implicit contract of a `Model` is it's something that takes the output of some `DatasetReader` - I don't think that's what this is intended to do.

allennlp/models/encoder_decoders/simple_seq2seq.py
```diff
@@ -1,23 +1,22 @@
...
+from allennlp.modules.seq2seq_decoders.decoder_cell import DecoderCell
+from allennlp.modules.seq2seq_decoders.simple_seq_decoder import SimpleSeqDecoder
 from allennlp.modules.token_embedders import Embedding
 from allennlp.nn import util
 from allennlp.nn.beam_search import BeamSearch
-from allennlp.training.metrics import BLEU
+from allennlp.training.metrics import BLEU, Metric
 
 
 @Model.register("simple_seq2seq")
```
matt-gardner: Instead of modifying this file directly, I'd just add a new implementation, like `generic_seq2seq` (or a better name...).  That way you don't need to worry about backwards compatibility of this class.

allennlp/modules/seq2seq_decoders/simple_decoder_cell.py
```diff
@@ -0,0 +1,121 @@
...
+
+from allennlp.common.checks import ConfigurationError
+from allennlp.modules import Attention, SimilarityFunction
+from allennlp.modules.attention import LegacyAttention
+from allennlp.modules.seq2seq_decoders.decoder_cell import DecoderCell
+from allennlp.nn import util
+
+
+@DecoderCell.register("simple_decoder")
+class SimpleDecoderCell(DecoderCell):
```
matt-gardner: Probably `AttentionLstmDecoderCell`, or something like that?  Or even just `LstmDecoderCell`, because attention is implied these days, and you can optionally just not pass it?

allennlp/modules/seq2seq_decoders/simple_decoder_cell.py
```diff
@@ -0,0 +1,121 @@
...
+    attention_function: ``SimilarityFunction``, optional (default = None)
+        This is if you want to use the legacy implementation of attention. This will be deprecated
+        since it consumes more memory than the specialized attention modules.
+    """
+
+    def __init__(self,
+                 decoding_dim: int,
+                 target_embedding_dim: int,
+                 attention: Attention = None,
+                 attention_function: SimilarityFunction = None,
```
matt-gardner: No need to add this.

allennlp/modules/seq2seq_decoders/simple_seq_decoder.py
```diff
@@ -0,0 +1,328 @@
...
+import torch.nn.functional as F
+
+from allennlp.data import Vocabulary
+from allennlp.modules.seq2seq_decoders.decoder_cell import DecoderCell
+from allennlp.modules.seq2seq_decoders.seq_decoder import SeqDecoder
+from allennlp.nn import util
+from allennlp.training.metrics import Metric
+
+
+class SimpleSeqDecoder(SeqDecoder):
```
matt-gardner: What other implementations of the decoder are you imagining?

generall: @epwalsh proposed CopyNet

generall:
Thanks for review, I will be back with changes

allennlp/modules/seq2seq_decoders/decoder_cell.py
generall: Current implementation of nn BeamSearch accepts states only as `Dict[str, torch.Tensor]`, so I we need decoder to have state of this type. It's steel pretty general, more general than a tuple

allennlp/modules/seq2seq_decoders/seq_decoder.py
generall: Changed

allennlp/models/encoder_decoders/simple_seq2seq.py
generall: Changed

allennlp/modules/seq2seq_decoders/simple_decoder_cell.py
generall: Changed

allennlp/modules/seq2seq_decoders/simple_decoder_cell.py
generall: Removed

generall:
Hi, I made some changes according to your comments. Also I have troubles with running tests: it fails with DeprecationWarning even in master branch. I temporary disabled DeprecationWarning on this branch so I could run tests.

matt-gardner:
Hey, sorry about the local test failures - see #2549 for a fix.  For the PR, I'm pretty swamped with ACL right now, I'll get back to this next week.

matt-gardner:
This API looks great to me, and the tests look good.  Have you trained a real model for something with it?

@brendan-ai2, can you also take a look at the high-level API and see what you think?  You've done the most with these seq2seq models in allennlp, I think.  Once you're also ok with the API, one of us can do an actual code review (though skimming through this, there aren't that many things I noticed that needed fixing).

allennlp/models/encoder_decoders/composed_seq2seq.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # Encodes the sequence of source embeddings into a sequence of hidden states.
+        self._encoder = encoder
+
+        if self._encoder.get_output_dim() != decoder_cell.get_output_dim():
+            raise ConfigurationError(
+                f"Encoder hidden dimension {self._encoder.get_output_dim()} should be"
+                f" equal to decoder dimension {decoder_cell.get_output_dim()}.")
+
+        self.decoder: SimpleSeqDecoder = SimpleSeqDecoder(
```
matt-gardner: This should be `Registrable` and a constructor parameter.  Otherwise it's not "fully customizable", as you say in the docstring.

generall: Unfortunately I didn't find a way to make this `Registrable`. It depends on `Vocabulary` and as I understood `Vocabulary` is passed implicitly only to `Model` objects.

generall: Changed this docstring

matt-gardner: The vocabulary is implicitly passed everywhere that it's needed - the embedders get access to it, for instance.

brendan-ai2:
Will do, @matt-gardner.

allennlp/models/encoder_decoders/composed_seq2seq.py
generall: really so! Made it registrable

sai-prasanna:
@generall  Great work :+1: . I have few doubts about this API. 

Your current decoder implementation requires a cell which performs single decoding step. Say I want a transformer decoder where I don't do scheduled sampling while training, only teacher forcing, then how would a transformer cell be implemented efficiently, since a single pass is all that is needed for training. 

If you can provide a transformer seq2seq implementation it would be really awesome, as it would exercise the capacity of these abstractions.

generall:
Hi @sai-prasanna , thanks for your question.

I fully agree that seq2seq transformer is a great "benchmark" for new API, more over the implementation of transformer was an initial idea behind all this changes. I definitely going to try to implement it, but it will take some time. For now, I guess, the new API has value on it's own.   

The idea of this decomposition is that you are not forced to use `SimpleSeqDecoder`, you can easily replace it with more appropriate implementation and you should not care about encoder part, since it is already implemented.

sai-prasanna:
@generall  I was looking at the decomposition, correct me if my understanding from a quick glance is wrong. `SeqDecoder` is the registrable parent for `SimpleSeqDecoder`, it having decoder cell in its properties might be a potential issue. I was wondering whether the decoder cell API can be more abstract .. But little hazy on how it would play out with other parts of the API. 

Anyways, your structuring is quiet a great start for thinking about how to lay out things.

matt-gardner:
Yes, thanks @sai-prasanna, those are good points, and I agree with them.  I wasn't even thinking about transformer decoders.  So, yes, @generall, you should move the `DecoderCell` (and all logic that depends on it) out of `SeqDecoder` and into `SimpleSeqDecoder`, which I would rename to `RnnSeqDecoder`.

generall:
Changed

matt-gardner:
@brendan-ai2, friendly ping here, in case this PR has fallen off your radar. I don't want to rush you if you've been busy with other things, I just suspect that you've forgotten about this.

brendan-ai2:
Thanks for the ping, definitely needed. I will take a look tomorrow.

sai-prasanna:
@generall Need any help to move this forward?

allennlp/modules/seq2seq_decoders/decoder_cell.py
```diff
@@ -0,0 +1,88 @@
+import torch
+from typing import Tuple, TypeVar, Generic, Dict, Any
+from allennlp.nn import util
+from allennlp.common import Registrable
+
+CellState = Dict[str, torch.Tensor]  # pylint: disable=invalid-name
```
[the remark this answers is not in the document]
matt-gardner: I suggested making this generic, which is why it got moved out.  I can help with making it an actual generic, which is a bit confusing in python, once we're happy with the overall API.

matt-gardner:
Sorry to be slow; @sai-prasanna, I think the thing that would help the most would be to have a stubbed out implementation of a transformer example.  That is, get the API parts right, and see what it would look like if a transformer were actually implemented in this framework.  That would be super helpful to nail down the abstractions.

generall:
Hi @sai-prasanna, right now I have lack of time for implementing transformer. I would be grateful for help in this.

allennlp/modules/seq2seq_decoders/seq_decoder.py
```diff
@@ -0,0 +1,104 @@
...
+        We really only use the predictions from the method to test that beam search
+        with a beam size of 1 gives the same results.
+        """
+        raise NotImplementedError()
+
+    def get_output_dim(self):
+        raise NotImplementedError()
+
+    def forward(self,
+                encoder_out: Dict[str, torch.LongTensor],
```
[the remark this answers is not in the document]
generall: yeah, it should be just `Tensor`

sai-prasanna:
@generall I have started working on this.
See this commit, very rough draft of a transformer decoder.

https://github.com/sai-prasanna/allennlp/commit/5974dc807a370d19e61b90bee49b8a60770b0479#diff-49ceb008ba6219aa754e9f18c9025810

@matt-gardner I adapted the annotated transformer to create a very rough version of Transformer Decoder on generall's abstractions. Overall, I think we need more clarity on what we want to accomplish on SeqDecoder abstract class. 

1. Should we move the TokenEmbedder for target embeddings to it? Stuff like weight tying will require the ComposedSeq2Seq class to know about the embeddings.

2. The RnnSeqDecoder uses a special DecoderCell abstraction, if we can make a single abstraction (say Decoder cell has a flag for whether it supports sequential decoding like RNN, or only full parallel) that can encapsulate both RNN and transformer, that would eliminate need to have two classes extending SeqDecoder, but would it be wise to go down this path?

3. Currently, I just copy pasted abstractions from havard's annotated transformer blog post to make the decoder. I think it would be better to match MultiHeadAttention Encoder code instead, but can I go further and try to have a single MultiHeadAttention class, that can support both encoding and decoding ? or would that be a case of single class trying to do too much?

brendan-ai2:
@sai-prasanna, a potentially relevant issue is here: https://github.com/allenai/allennlp/issues/2270

matt-gardner:
Ok, sorry it took me a few days to look at this - things are starting to get busy with the EMNLP deadline coming up, and it takes a lot of time to context switch back to this big PR.  For your questions:

1. I think weight tying can be handled later, and probably can't be done in a generic way, anyway, so I wouldn't worry about trying to move it up.

2. The transformer doesn't have a decoder cell, so trying to force it to use one just for the sake of an abstraction is not a great idea.  In general, we shouldn't try to over-generalize abstractions.  That's when things get really ugly, hard to use, and you'd be better off just not using the abstraction at all.

3. Same argument applies here, though in this case they are close enough that it might be possible to get a single abstraction that works for both.  I'm not sure; I haven't worked with transformer code enough myself to know the answer here.

In general, though, I think your transformer example does a great job answering @brendan-ai2's concerns - you got a transformer working inside this framework just by implementing a new `SeqDecoder`, and the API that @generall made still works.  I think this is a good argument for proceeding with this abstraction.  @brendan-ai2, what do you think?

sai-prasanna:
@matt-gardner I realized that even a transformer decoder will anyways implement a single step prediction for inference (using beam search). So it wouldn't be hard to abstract a LSTMCell and Transformer decoders as `DecoderModule` with a "is_sequential" flag on the `DecoderModule` abstract class.

@brendan-ai2 What do you think of such a abstraction?
A sequential `DecoderModule` given encoder outputs and previous target embeddings will return a output of just one time step, while a parallel one will return multiple time steps. Both are just `torch.Tensor`.

I am coding it up, will update here soon.

sai-prasanna:
@matt-gardner  @brendan-ai2 
I have refactored the code, now there is single default decoder implementation, (still a implementation not putting everything into the abstract class) which can support both sequential modules like 
LSTMCell Decoder and "parallel" modules like Transformer Decoder or Cuda LSTM decoder (Yet to implement that).

https://github.com/allenai/allennlp/compare/master...sai-prasanna:decoder-updates

cpuhrsch:
@zhangguanheng66 please take a look. @sai-prasanna mentioned help is required to figure out the decoder abstraction.

zhangguanheng66:
I'm just wondering why you prefer to keep single decoder abstract. Correct me if I'm wrong. The LSTM decoder is very different from the transformer decoder, which is based on multi-layer multi-head attention model.

I'm also thinking if you are interested in using the MultiheadAttention module in PyTorch, instead of maintaining a separate one in the PR. We want to help maintain those common modules/functions. We have a plan to optimize the module this summer and appreciate feedbacks.

brendan-ai2:
Thanks for working on this, @sai-prasanna! Broadly this is looking pretty good. I'm glad the `SeqDecoder` abstraction is holding up and I think I understand your rationale for having `DecoderModule` as well. @matt-gardner, does the move from `DecoderCell` to `DecoderModule` address your concerns?

On the implementation level I'd like to avoid adding another transformer to `AllenNLP`. Minimally we'd want to deduplicate with the Annotated Transformer code we already have in https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2seq_encoders/bidirectional_language_model_transformer.py. But really using what PyTorch provides will make the most long-term sense. @zhangguanheng66, when do you anticipate that https://github.com/pytorch/pytorch/pull/20170/ will land in a released version of PyTorch? @sai-prasanna, is that implementation missing anything you'll need?

matt-gardner:
@brendan-ai2, I think we need a better name than `DecoderModule`, but if we need a one-step decoder for all models for beam search, anyway, then yes, if there's a reasonable abstraction that fits all cases for that, I think it makes sense.  This is starting to look more and more like the state machine code we already have; maybe that's a good sign?  It would be nice to unify these two eventually (though I wouldn't try to do that right now).

sai-prasanna:
@brendan-ai2 I took a brief look on the pytorch Transformer PR, it looks good.
In my copy-pasted version, I am getting few nan errors, when I attempted to train on another dataset. I am debugging it.

1. Should I open a new PR? To get comments on my branch, since this one is on @generall 's fork.

2. How much of transformer de-duplication effort is tied with this PR, I can incorporate the new MultiHeadAttention module from torch 1.1 in the decoder class. Other than that should I also attempt to make the existing MultiHeadAttention implementations to use the torch version?

3. My nan errors are making me paranoid of whether any problem is there with the abstraction code, if someone can run it over something non trivial it would be a relief.

4. What are the other major things to work on before this gets integrated? I will add the relevant documentation, tests

@matt-gardner I agree, DecoderModule doesn't sound right. I was not able to get a good name for it. Will think of something better by the end of this PR.

zhangguanheng66:
@sai-prasanna I'm happy to take a look at your code for the nan errors. If you have a PR or some test problems, just let me know. We also have some unit tests for torch.nn.MultiheadAttention module.

zhangguanheng66:
@sai-prasanna I notice that you don't reset the parameters in your MultiheadedAttention module. 
 I made a test on torch.nn.MultiheadAttention module. If I don't call self._reset_parameters() function when initiating torch.nn.MultiheadAttention, it does report some nan results sometime.

sai-prasanna:
@zhangguanheng66 
* The training/validation (experiment_transformer.json) works without nans on CPU, its only on GPU nans happen for me. I have a measly 970GT card at home.
* I tried doing xavier initialization for all parameters with size > 2, oddly nan starts appearing after few epochs.
* One more odd thing is nan start appearing only on BeamSearch during validation, not during loss computation.
* FYI I am not using nn.MultiheadAttention , and (also noticed that nn.MultiheadAttention is using T x B x H while allennlp generally defaults to B x T x H)

@brendan-ai2  I have imported all the common code, and remove duplicate classes for the Transformer Decoder Layer.

zhangguanheng66:
@sai-prasanna just want to check how's everything going. We are interested if anything is missed and how's the NaN errors. Please let us know if you have any comments on our PR. We plan to land it later.

sai-prasanna:
@zhangguanheng66 , I haven't tried the MultiHeadAttention from pytorch 1.1 yet. Saw that it is T X B x H, while in general allennlp keeps it as B x T x H. I don't know how much performance issue will this be, but just a note.

And with regards to the Transformer pull request, so long as the embedding layer is kept seperated from encoder and decoder modules, it would fit in this abstraction.


@matt-gardner @brendan-ai2 - Shall I start a new PR?

cpuhrsch:
@sai-prasanna - I'm happy to hear that! We plan to merge it soon and will adhere to your requirements. Please keep us in the loop of any upcoming adoptions and any issues you might encounter.

sai-prasanna:
@joelgrus @matt-gardner  @brendan-ai2 , a gentle remainder. Shall I create a new PR?

brendan-ai2:
Please do, @sai-prasanna. Sorry for the delay. We should be more responsive now that the EMNLP deadline has passed!

matt-gardner:
Closing this in favor of #2913, which I believe @brendan-ai2 will be getting too relatively soon.

Issue #2388, opened 2019-01-18
Tensorboard summary of learning rate


tingkai-zhang:
Hi, i am using cosine with restart with the following configuration:
```
    "learning_rate_scheduler": {
       "type":"cosine",
       "t_initial": 500,
       "t_mul": 1.5,
       "eta_min": 0.0,
       "eta_mul": 0.7

    },
    "optimizer": {
      "type": "adam",
      "lr": 3e-5
    }
```
The training has 1.2k steps per epoch.
I think what that the tensorboard summary shows that there is minor change of learning rate(0.001e-5) and not giving a cosine-like shape curve? Does the graph make sense?
![image](https://user-images.githubusercontent.com/20318562/51363397-c5162680-1aa5-11e9-9c96-bfddeec51170.png)

epwalsh:
This learning rate scheduler only updates at the end of each epoch, which is why it is flat through the 1.2K steps within each epoch. Does that answer your question? (there's no particular reason why it doesn't update "smoothly" instead of only at the end of each epoch, other than it was just easier to implement, TBH)

brendan-ai2:
Thanks, @epwalsh!

@tingkai-zhang , If desired you can update per-step by making your own learning rate scheduler. See, for instance, https://github.com/allenai/allennlp/blob/master/allennlp/training/learning_rate_schedulers.py#L138.

tingkai-zhang:
@epwalsh  Hi, thanks for answering! 

> This learning rate scheduler only updates at the end of each epoch, which is why it is flat through the 1.2K steps within each epoch.

If i understand it properly, you mean the `lr` is updated each iteration but only logged to tensorboard each epoch, am i right?

If the updated learning rate is only logged at the end of each epoch, the learning rate should be `eta_min` which should not be `3e-5` in my case, right? 


@brendan-ai2  Thanks for your advice! I will look into that! 
When i use slanted triangular scheduler, I also observe the following learning rate on tensorboard is (probably) not we want?

There is a cut of line on the right side because I got such screenshot while the model is still in training

![image](https://user-images.githubusercontent.com/20318562/51420547-fbfd4280-1b60-11e9-9f4b-2e92e9517253.png)

![image](https://user-images.githubusercontent.com/20318562/51420537-edaf2680-1b60-11e9-9275-886bc9d5539f.png)


I thought the unfreezed layer would have the learning rate graph as following till the end of training (not till the end of an epoch).
[...]

PR #2523, opened 2019-02-18
Unfreeze layers when recovering with slanted triangular learning rate

Hyperparticle:
`is_first_epoch` will always be true for one epoch regardless of the current epoch number. Instead, we should base the unfreezing on the epoch number to unfreeze the correct number of layers when recovering a training run.

Alternatively, could simplify the condition to

```python
num_layers_to_unfreeze = epoch + 2 if epoch > -1 else 1
```

### DIFF
--- a/allennlp/training/learning_rate_schedulers/slanted_triangular.py
+++ b/allennlp/training/learning_rate_schedulers/slanted_triangular.py
@@ -60,7 +60,6 @@ class SlantedTriangular(LearningRateScheduler):
         self.ratio = ratio
         self.gradual_unfreezing = gradual_unfreezing
         self.freezing_current = self.gradual_unfreezing
-        self.is_first_epoch = True
         # track the actual number of steps for each epoch
         self.batch_num_total_epoch_end: List[int] = []
         if self.gradual_unfreezing:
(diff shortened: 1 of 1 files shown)


Hyperparticle:
This may not change what I hoped. I think I had some confusion on the order of operations. My intention was to be able to change the existing `config.json` and have it override the saved training state. Looks like it's currently the other way around, i.e., the checkpointer overrides the config specification (for slanted triangular learning rate, anyways). Is there a way to override the training state when recovering?

Hyperparticle:
I did some more digging. Indeed, the checkpointer overrides any state in the learning rate scheduler, which will cause `slanted_triangular` to always freeze all layers but the last for the first epoch (when `gradual_unfreezing` is enabled), even when recovering. Any easy way to fix this?

tingkai-zhang:
Hi, thanks for finding the problem and gave the fix! Did you check this issue? #2388 
I thought the problem stated in the comment is similar to what is described here, isn't it?

Hyperparticle:
@tingkai-zhang It's similar, but I'm more concerned about changing the configuration of LR schedulers when recovering from a previous run. The strange behavior you see with slanted triangular is intentional, but I'm not sure if it really helps.

matt-gardner:
@matt-peters, can you take a look at this and decide what the best course of action is here?

matt-peters:
@Hyperparticle Will adding `state_dict` and `load_state_dict` methods to `slanted_triangular` to persist the state at in the checkpoint accomplish what you want?

Hyperparticle:
@matt-peters It appears the state is already persisted across checkpoints, just incorrectly. The other problem is there is currently no easy way (I think) to override that state by changing the config file when resuming training.

matt-peters:
I'd suggest fixing both problems by: (a) adding the correct state to the checkpoint so it's available if the user doesn't want to override (b) adding a parameter to the constructor that allows the user to override the checkpoint state if desired.

Hyperparticle:
Sounds reasonable. I'll try that out.

Hyperparticle:
I've gone with a custom solution, so I'm closing this.

PR #2524, opened 2019-02-18
upgrade huggingface pretrained bert

joelgrus:
includes OpenAI GPT-2. don't have an immediate use case for it, but I suspect if it's in the library people will figure out a way to use it.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -67,7 +67,7 @@ sqlparse==0.2.4
 ftfy
 
 # To use the BERT model
-pytorch-pretrained-bert>=0.5.0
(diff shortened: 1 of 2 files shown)


joelgrus:
it's possible we should add corresponding token embedder, I need to read more about the GPT2 model

thomwolf:
@joelgrus Just a note: I added a new dependency for the GPT-2 tokenizer that AllenNLP doesn't have: `regex`. Do you have any objection to this new dep?

joelgrus:
it looks like we already have a transitive dependency on it via spacy, they're pinned to an older version, hopefully that's ok

PR #2532, opened 2019-02-20
Restore tensorboard epoch metrics to pre-refactoring behavior

### LINKED ISSUE #2529: How can I log 'loss' to Tensorboard every epoch instead of every iteration?
**System:**
 - OS: Linux
 - Python version: 3.6.8
 - AllenNLP version: v0.8.2
 - PyTorch version: 1.0.1

**Question**
* I have recently updated to AllenNLP v0.8.2 and I've noticed that 'loss' metric is now logged every iteration to Tensorboard. How can I log 'loss' (or other metrics) to Tensorboard every epoch instead of every iteration?

loss

matt-gardner:
Fixes #2529.

This changes the metrics that get logged to tensorboard at the end of every epoch to use the epoch number as its counter, instead of the batch number.  This is what the behavior was prior to the refactoring.

I think there are still some very confusing things in this behavior, which we may want to change: "epoch_metrics" isn't actually "the metrics at the end of every epoch", it's "train metrics shown at `summary_interval` steps during an epoch".  So I propose that we regroup things into two sections: (1) `epoch_metrics/`, which use the epoch number as their counter and are shown for both training and validation, and (2) `train_metrics/`, which use the batch number as their counter and are logged at `summary_interval`.  This is just renaming the current `epoch_metrics` to `train_metrics`, and grouping together the per-epoch metrics into a common prefix, so you don't have a top-level "loss", "accuracy", etc.  I'd probably also just remove "loss/loss_train" while doing this, because it's an exact duplicate of `train_metrics/loss`.

@fbrad, FYI.

### DIFF
--- a/allennlp/training/tensorboard_writer.py
+++ b/allennlp/training/tensorboard_writer.py
@@ -64,10 +64,11 @@ class TensorboardWriter:
     def should_log_histograms_this_batch(self) -> bool:
         return self._histogram_interval is not None and self._get_batch_num_total() % self._histogram_interval == 0
 
-    def add_train_scalar(self, name: str, value: float) -> None:
+    def add_train_scalar(self, name: str, value: float, timestep: int = None) -> None:
+        timestep = timestep or self._get_batch_num_total()
         # get the scalar
         if self._train_log is not None:
-            self._train_log.add_scalar(name, self._item(value), self._get_batch_num_total())
+            self._train_log.add_scalar(name, self._item(value), timestep)
 
     def add_train_histogram(self, name: str, values: torch.Tensor) -> None:
         if self._train_log is not None:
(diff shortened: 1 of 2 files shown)


matt-gardner:
What do you think of the regrouping I suggested?  I didn't put them in this PR, but I could do it pretty easily.

joelgrus:
I don't have any objection to it, other than that it will mess with people who do one experiment before the change and another experiment after the change (like this bug did).

I'm not sure the new names are descriptive enough, though. Maybe `per_epoch_metrics` and `per_batch_metrics`?

matt-gardner:
I could go for those names.  @DeNeutoy, what do you think?

DeNeutoy:
I think I originally meant `epoch_metrics` to mean "within epoch metrics", so that's my bad if it's confusing. I'm not really a fan of the name change because it makes comparing logs annoying, but if you think it's better I don't feel that strongly. 

Really the problem here is logging at both frequencies is not that helpful - either your dataset is small, and you care about per epoch metrics, or your dataset is very large and you want online _running average_ metrics, not cumulative average, and the per epoch stuff happens so infrequently (if at all) that it is essentially useless. To solve that it would be cool to have a new trainer subclass which assumes your dataset arrives in a never ending stream (there are other features which are desirable in this case, like evaluating on a validation set every `eval_every_n_seconds` seconds, etc).

matt-gardner:
Yeah, changing things is annoying, I agree. I'm not sure the additional clarity is worth the name change.

matt-gardner:
I'm just going to merge this because it fixes the bug, and we can consider changing things later if someone complains.  No need to break backwards comparisons just because I got confused =).

PR #2537, opened 2019-02-22
Add ability to generate BERT embeddings using a sliding window (#2493)

### LINKED ISSUE #2493: Add ability to generate BERT embeddings using a sliding window instead of truncating the input
The BERT wordpiece indexer will [truncate any sentences that have more than 512 wordpieces](https://github.com/allenai/allennlp/blob/master/allennlp/data/token_indexers/wordpiece_indexer.py#L160), which can be problematic for models that assume that the output length is fixed to the length of the input tensor representing word tokens when using `offsets`.

Ideally, the BERT embedding length should always be equal to the length of the word token embeddings when using `offsets`. A trick [I've seen elsewhere](https://github.com/google-research/bert/issues/66#issuecomment-436378461) would be to break up sentences into smaller chunks of up to 512 wordpieces using a sliding window with a certain stride (e.g., 256).

Assuming a sliding window of length 4 and stride of length 2, a sentence like

```
I went to the store to buy some milk
```

can be broken up into

```
I went to the
to the store to
store to buy some
to buy some milk
```

These token sequences can then be embedded into BERT and recombined by either truncating overlapping sequences or averaging them. Because of possible OOM issues, it might be a good idea to make a backward-compatible change allowing truncation to be an option and sliding window to be another.

To make the change, I'm thinking of modifying the [wordpiece indexer `tokens_to_indices`](https://github.com/allenai/allennlp/blob/master/allennlp/data/token_indexers/wordpiece_indexer.py) to return a list of tensors for both `wordpiece_ids` and `offsets`. Then the [BERT token embedder](https://github.com/allenai/allennlp/blob/master/allennlp/modules/token_embedders/bert_token_embedder.py) would then call `self.bert_model` on each tensor in the list and recombine them.

I have tried to break up long sentences manually in my datasets, but identifying them is a pain. The process also complicates the code, as the sentences need to be recombined after prediction.

@joelgrus Any thoughts on this idea? Additional concerns? I'd be willing to make the PR.

Hyperparticle:
Fixes #2493. Took me a while, but this enhancement should allow a user to feed any length input to BERT without truncation. It uses a rolling window of 512 wordpieces, overlapping with adjacent windows by 256 wordpieces.

Most of the time, this should behave exactly as before. But for batches with long sentences, this should do the right thing and split up the input into sub-sequences, pass them through BERT as one big batch, and recombine them in the end. It doesn't do anything fancy, just inserts start/end tokens at regular intervals in the indexer, and then the embedder just grabs indices such that the sequence length of the BERT output should always equal the sequence length of the `single_id` token indexer.

I added as many helpful comments as I could detailing the approach. Some of them could be moved up into the docstring to notify the user of its internal behavior.

I was debating on whether it should be separated into its own class or merged into the default indexer/embedder, but I guess I'll let you decide. The nice thing is that the default behavior shouldn't change anything in either the token indexer or embedder. Note: I changed the structure of the token indexer from single pass to multiple passes to try to increase readability, some of the variable names could be named better, and some of the tensor ops maybe could be more efficient.

I also left the default behavior `truncate_long_sequences = True`, as there may still be some edge cases I missed. However, I've had no problems thus far when processing long sentences and setting `maximum_samples_per_batch` in the iterator to ensure longer sentences get put into smaller batches.

Let me know what you think! @joelgrus :cake:


joelgrus:
hey, I am very sorry it took me so long to get to this, I have a few comments:

1. it's failing a few checks, the pylint and mypy ones should be easy fixes (formatting and not changing the return type at https://github.com/allenai/allennlp/pull/2537/files#diff-ba7ce0207c80fe36b224c06ef743576cR125)

2. the pytest one is harder, because this version of the code is not doing the right thing in the "truncate" case; in particular, it's truncating the sequence of wordpiece_ids but not the offsets, which is why the corresponding test is failing.

3. if you look at how the original BERT code does it, it picks for each wordpiece the embedding with "maximum context", it looks like you're just taking the left half of each input sequence? I don't have a great sense of how much difference that makes.

Hyperparticle:
1-2. Thanks, I'll see if I can get those issues fixed.

3. Yes, this code takes the left half of each input sequence to keep things simple. I was not aware that the original BERT code already does something like this, might take a look to see what they do. If it's an easy change, I'll make it for the sake of consistency.

joelgrus:
take a look here:

https://github.com/google-research/bert/blob/master/run_squad.py#L383

it seems like they essentially keep another mask-like vector indicating whether each wordpiece copy is the "max context" version

Hyperparticle:
I fixed the offsets truncation and formatting, got the tests to pass.

As for selecting the embeddings with the best context, I'm thinking the easiest change would be to shift the selection from the left-most embeddings to the middle (so each embedding will be selected with offsets in the range [128-384]). This should be functionally equivalent and faster to calculate.

joelgrus:
that sounds good to me, or does this PR already have that implemented?

Hyperparticle:
Not yet, should get around to doing that soon.

nelson-liu:
@Hyperparticle thanks for working on this, it'll be super useful! Any sense when you'll be able to get to the last bit (taking embedding with maximal context for each input word)? Thanks again!

Hyperparticle:
@nelson-liu Thanks for the reminder; I forgot about this. I made some changes to get the max context embeddings, so this code should be ready.

allennlp/data/token_indexers/wordpiece_indexer.py
```diff
@@ -137,39 +175,53 @@ def tokens_to_indices(self,
...
+            if last_window == penultimate_window[-len(last_window):]:
+                wordpiece_windows = wordpiece_windows[:-1]
+
+        # Flatten the wordpiece windows
+        wordpiece_ids = [wordpiece for sequence in wordpiece_windows for wordpiece in sequence]
+
         # Constructing `token_type_ids` by `self._separator`
-        token_type_ids = _get_token_type_ids(wordpiece_ids,
-                                             self._separator_ids)
+        token_type_ids = _get_token_type_ids(wordpiece_ids, self._separator_ids)
```
joelgrus: I feel like this is wrong? Imagine I have a "[SEP]" token between my sentences, it could end up in multiple windows? Then the flattened wordpiece_ids will have multiple "[SEP]" tokens, which means this will be doing the wrong thing. Am I missing something?

allennlp/modules/token_embedders/bert_token_embedder.py
```diff
@@ -55,11 +56,30 @@ class BertEmbedder(TokenEmbedder):
         The BERT model being wrapped.
     top_layer_only: ``bool``, optional (default = ``False``)
         If ``True``, then only return the top layer instead of apply the scalar mix.
+    max_pieces : int, optional (default: 512)
+        The BERT embedder uses positional embeddings and so has a corresponding
+        maximum length for its input ids. Assuming the inputs are windowed
+        and padded appropriately by this length, the embedder will split them into a
+        large batch, feed them into BERT, and recombine the output as if it was a
+        longer sequence.
+    start_tokens : int, optional (default: 1)
```
joelgrus: these should be called `num_start_tokens` and `num_end_tokens`

allennlp/data/token_indexers/wordpiece_indexer.py
joelgrus: I just pushed up a test that fails because of this, do you want me to take a stab at fixing it?

joelgrus: ok, I pushed up what should be a fix

allennlp/modules/token_embedders/bert_token_embedder.py
joelgrus: ok, I changed this too

joelgrus:
ok, @nelson-liu can you review?

allennlp/data/token_indexers/wordpiece_indexer.py
```diff
@@ -137,39 +179,59 @@ def tokens_to_indices(self,
...
+        elif self._truncate_long_sequences:
+            logger.warning("Too many wordpieces, truncating sequence. If you would like a sliding window, set"
+                           "`truncate_long_sequences` to False %s", str([token.text for token in tokens]))
+            wordpiece_windows = [self._add_start_and_end(flat_wordpiece_ids[:window_length])]
+            token_type_ids = self._extend(flat_token_type_ids)
+        else:
+            # Create a sliding window of wordpieces of length `max_pieces` that advances by `stride` steps and
+            # add start/end wordpieces to each window
+            # TODO: this currently does not respect word boundaries, so words may be cut in half between windows
+            # However, this would i
```
nelson-liu: fix the spacing in the comment?

allennlp/data/token_indexers/wordpiece_indexer.py
```diff
@@ -137,39 +179,59 @@ def tokens_to_indices(self,
...
+            token_type_ids = self._extend(flat_token_type_ids)
+        elif self._truncate_long_sequences:
+            logger.warning("Too many wordpieces, truncating sequence. If you would like a sliding window, set"
+                           "`truncate_long_sequences` to False %s", str([token.text for token in tokens]))
+            wordpiece_windows = [self._add_start_and_end(flat_wordpiece_ids[:window_length])]
+            token_type_ids = self._extend(flat_token_type_ids)
+        else:
+            # Create a sliding window of wordpieces of length `max_pieces` that advances by `stride` steps and
+            # add start/end wordpieces to each window
+            # TODO: this currently does not respect word boundaries, so words may be cut in half between windows
```
nelson-liu: i'm not sure why this is a problem...am i missing something?

Hyperparticle: Just a small technical consideration. I guess it's less important now since I changed it to maximal context embeddings which use the center embeddings.

Hyperparticle:
@joelgrus Thanks so much! By the way, the tensor ops I used to split and stack the embeddings could probably be replaced with a single op, but I'm not sure how to do it in PyTorch.

joelgrus:
if nelson finds it too slow for his needs, he can figure that out 😇

joelgrus:
so while the example I gave that breaks does have a heterogenous batch, it fails on just the second example. nelson has a theory that it might be because the number of wordpieces is evenly divisible by max_pieces

Hyperparticle:
Yes, I just found the bug. The test reveals an edge case where the sequence length is exactly divisible by `max_pieces`, which makes the calculation for the final window offset incorrect. The modulo should not ever return 0, but `max_pieces` in those cases.

joelgrus:
ok, I can fix that. wow, that's just dumb luck that I ran across that bug.

Hyperparticle:
Very lucky indeed. This should work: `(full_seq_len - 1) % self.max_pieces + 1`

joelgrus:
I just did an if / else, take a look

Hyperparticle:
If/else is equivalent, but subtracting/adding is simpler (e.g., what you would do when doing modulo with 12-hour clocks).

joelgrus:
I always choose the more readable option over the simpler one 😎

Hyperparticle:
Fair enough! Not like one simple branch will make the code slow. :smile:

nelson-liu:
Thanks @Hyperparticle !!

j6mes:
Hi, can I ask a (dumb) question about the behaviour of the recombined embeddings? With a batch-size of 1, I would expect the length of the textfields and length of the embeddings generated by BERT to be equivalent. 

I'm just running a simple SQuAD style evaluation and have added this assertion which fails:

```
    def forward(  # type: ignore
        self,
        question_with_context: Dict[str, Dict[str, torch.LongTensor]],
        context_span: torch.IntTensor,
        answer_span: Optional[torch.IntTensor] = None,
        metadata: List[Dict[str, Any]] = None
    ) -> Dict[str, torch.Tensor]:
        embedded_text = self._text_field_embedder(question_with_context)
        assert embedded_text.shape[1] == question_with_context["tokens"]["token_ids"].shape[1], "Expected embedding size and sequence length to be equal, but instead got an embedding size of {} for {} tokens".format(embedded_text.shape[1], question_with_context["tokens"]["token_ids"].shape[1])
```

```
AssertionError: Expected embedding size and sequence length to be equal, but instead got an embedding size of 1080 for 1088 tokens
```

If this is expected, how would the original tokens map to the embedded tokens? If it is not expected, what could the source of the issue be? Could it be something silly I'm doing?

Thanks

PR #2545, opened 2019-02-26
Extend evaluate to work with untrained models.

brendan-ai2:
- Allow `allennlp evaluate` to operate on untrained models, i.e. `.jsonnet` config.
- Motivation: For the S2 salience work we're comparing against a number of untrained baselines like tf-idf. It's convenient to do this in AllenNLP to access the dataset readers and metrics.

### DIFF
--- a/allennlp/commands/evaluate.py
+++ b/allennlp/commands/evaluate.py
@@ -57,9 +57,11 @@ import json
 
 from allennlp.commands.subcommand import Subcommand
 from allennlp.common.util import prepare_environment
+from allennlp.data import Vocabulary
 
 from allennlp.data.dataset_readers.dataset_reader import DatasetReader
 from allennlp.data.iterators import DataIterator
+from allennlp.models import Model
 from allennlp.models.archival import load_archive
 from allennlp.training.util import evaluate
 from allennlp.common import Params
@@ -115,6 +117,14 @@ class Evaluate(Subcommand):
                                'extended, we will try to use the original file paths used during training. If '
                                'they are not available we will use random vectors for embedding extension.')
 
+        subparser.add_argument('--load-config-only',
+                               action='store_true',
+                               default=False,
+                               help='If specified, the archive file should in fact be a .jsonnet '
+                                    'config file. This will be instantiated and evaluated without '
+                                    'trained weights. Intended for evaluating programmatic '
+                                    'baselines or models trained outside AllenNLP.')
+
         subparser.set_defaults(func=evaluate_from_args)
 
         return subparser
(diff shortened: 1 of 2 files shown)


brendan-ai2:
@wammar, hopefully you'll find this useful!

joelgrus:
My initial reaction is that I'm not really thrilled with this change, it feels like a "hack" (e.g. `archive_file` is a trained model archive, except that if you pass a certain flag it's not, and now the docstring is not precisely correct, and so on).

If there's an immediate need for this, I'd prefer factoring out the "evaluate on a model" code into reusable functions and then just writing a separate script that calls those functions against a model loaded from a config file.

If we think there's a more widespread need for this, I'd prefer that we spend some time thinking through the design.

Thoughts?

matt-gardner:
FWIW, that was also my initial reaction.

brendan-ai2:
Would this still feel like a hack if I changed `archive_file` to `model_file` and detected whether it was Jsonnet or a tarball -- dispatching appropriately? The code-level change is simply the addition of two if statements, which doesn't seem pernicious.

To argue against myself, there is some small risk that a new user would run `evaluate` without training and be confused by the poor performance.

Certainly happy to move this to a client-side script if the consensus is still "hack". I would, however, suggest we jot down "programmatic baseline" as a use case for later. It's surprising to me that we don't support it with the standard tools.

matt-gardner:
I definitely agree that "programmatic baseline" is a good use case to support.  It just doesn't look like we've found a clean way to support that use case yet - this feels like it's cobbling things onto something that had a very different intent and semantics.  Even saying that you should implement your programmatic baseline as a `Model` seems a bit of a stretch to me.  Like, say it's a "most frequent answer" baseline - that means you really need to compute the most frequent answer from the data somehow, pass the result to the model (as a `Field`...?  using something like an external `compute_vocab`...?), and then use our model infrastructure to classify each instance.  Having everything be consistent this way might be cute, but it seems like it's stretching the abstractions in ways they weren't really made for, and it's a whole lot more work than running a simple script.

I don't know.  I could probably be convinced with some good arguments from some concrete use cases, but until there's a broader story here this feels like a hack to me.

brendan-ai2:
The concrete use case I have for this is to write a term frequency baseline for entity salience. Currently I take the document-level term frequencies for the entities (in a few different contexts), pass them through a projection layer, and then treat the output as scores. This works fairly well. An even simpler baseline would be to do away with the projection layer and simply treat the more frequently occurring entity as more salient. The problem with this is that now there is nothing to train, so `allennlp train` fails and I can't produce a model archive.

Specifically, changing

```
label_scores = self._projection(frequency_features)
```

to

```
e0_scores = frequency_features.narrow(1, 0, 2).sum(-1)
e1_scores = frequency_features.narrow(1, 2, 2).sum(-1)
label_scores = torch.stack([e0_scores, e1_scores], -1)
```

produces the error `RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn`. As a user I just want this to work. Having to go and write a script isn't hard, but it's not an ideal use of time. Maybe the behavior is cute, but it's quite useful!

@joelgrus and I talked about this in person and he suggested creating a no-op trainer and making model weights optional. This would allow us to preserve the model archive abstraction. This solution seems quite clean from the library perspective, so I'm partial to it.

(As an aside, the "most frequent answer" baseline you mention would certainly be a good one. However, the issue of needing to do some pre-computation applies more broadly than to just baselines so I wouldn't level that criticism against them specifically. You outline this yourself in https://github.com/allenai/allennlp/issues/1809.)

matt-gardner:
Thanks for the additional examples. I do agree that a no-op trainer and a model archive without any weights sounds a lot better than messing with the semantics of `evaluate`, and you're right that the example I brought up is actually a bit orthogonal to your point here.

brendan-ai2:
Thanks both for the feedback! Closing this now. I'll make a new PR for the no-op trainer.

joelgrus:
btw, I don't think you should make model weights optional, I am pretty sure that your do-nothing trainer should just save an empty state dict and then I think you don't have to change anything else.

brendan-ai2:
Gotcha, thanks!

PR #2548, opened 2019-02-27
Add perplexity metric to LanguageModel

### LINKED ISSUE #2400: LanguageModel should output perplexity
Is there a reason why `allennlp evaluate` doesn't already output the perplexity? This should be fairly easy to do, since you just take `exp(loss)`.

Seems like it might be tricky during training (@brendan-ai2 made a few notes about it in the code), but for evaluation we'd just need to override `get_metrics`, right?

thesamuel:
As discussed in #2400, this PR adds a `perplexity` metric to the `LanguageModel`. To do this, I created a `Perplexity` metric class which is updated with a negative log loss. @nelson-liu 

fixes #2400

### DIFF
--- a/allennlp/models/language_model.py
+++ b/allennlp/models/language_model.py
@@ -11,6 +11,7 @@ from allennlp.modules.sampled_softmax_loss import SampledSoftmaxLoss
 from allennlp.modules.seq2seq_encoders import Seq2SeqEncoder
 from allennlp.nn.util import get_text_field_mask
 from allennlp.nn import InitializerApplicator
+from allennlp.training.metrics import Perplexity
 
 
 class _SoftmaxLoss(torch.nn.Module):
@@ -127,8 +128,10 @@ class LanguageModel(Model):
             self._softmax_loss = _SoftmaxLoss(num_words=vocab.get_vocab_size(),
                                               embedding_dim=self._forward_dim)
 
-        # TODO(brendanr): Output perplexity here. e^loss
+        # This buffer is now unused and exists only for backwards compatibility reasons.
         self.register_buffer('_last_average_loss', torch.zeros(1))
+
+        self._perplexity = Perplexity()
 
         if dropout:
             self._dropout = torch.nn.Dropout(dropout)
@@ -300,8 +303,8 @@ class LanguageModel(Model):
                     average_loss = forward_loss / num_targets.float()
             else:
                 average_loss = torch.tensor(0.0).to(forward_targets.device)  # pylint: disable=not-callable
-            # this is stored to compute perplexity if needed
-            self._last_average_loss[0] = average_loss.detach().item()
+
+            self._perplexity(average_loss)
 
             if num_targets > 0:
                 return_dict.update({
@@ -327,3 +330,6 @@ class LanguageModel(Model):
         })
 
         return return_dict
+
+    def get_metrics(self, reset: bool = False):
+        return {"perplexity": self._perplexity.get_metric(reset=reset)}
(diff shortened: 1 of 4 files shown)


allennlp/models/language_model.py
```diff
@@ -300,8 +302,9 @@ def forward(self,  # type: ignore
                     average_loss = forward_loss / num_targets.float()
             else:
                 average_loss = torch.tensor(0.0).to(forward_targets.device)  # pylint: disable=not-callable
-            # this is stored to compute perplexity if needed
+            # TODO: _last_average_loss buffer is no longer used.
```
nelson-liu: @brendan-ai2 , want to confirm that we're ok with just deleting `self._last_average_loss`?

allennlp/training/metrics/perplexity.py
```diff
@@ -0,0 +1,32 @@
...
+
+@Metric.register("perplexity")
+class Perplexity(Average):
+    """
+    Perplexity is a common metric used for evaluating how well a language model
+    predicts a sample.
+
+    Notes
+    -----
+    Assumes negative log likelihood loss of each batch. Provides the average
```
nelson-liu: maybe specify that log should be base `e`?

allennlp/models/language_model.py
thesamuel: Removing this causes an error when restoring from a serialized model that was saved with this buffer. There is a test that fails when this is removed.

nelson-liu: Yeah, it's a bit annoying---we either have to retrain a new fixture or go into the existing fixture and manually remove the `_last_average_loss` key in the state dictionary. (1) might be easier, and we probably already have a process to do it... @brendan-ai2 might know more?

allennlp/training/metrics/perplexity.py
thesamuel:
```suggestion
    Assumes negative natural log likelihood loss of each batch. Provides the average
```

allennlp/models/language_model.py
brendan-ai2: The fact that the test with the fixture is breaking means that all serialized models would break if we remove it. This would be bad. Let's avoid that by just keeping the field around with a note that it's unused and exists for backwards compatibility.

allennlp/training/metrics/perplexity.py
thesamuel: Fixed

allennlp/models/language_model.py
thesamuel: Sounds good. I added a comment above that buffer in `__init__` and removed the update from `forward`.

thesamuel:
I'm unsure if my commit message is in the correct format. It seems like commits are automatically squashed on a per-PR basis, but I've been squashing manually.

nelson-liu:
Don't worry about squashing, that's handled by GitHub when the PR is merged.

matt-gardner:
@nelson-liu, @thesamuel, we're not sure what this is waiting on.  Do either of you know?

thesamuel:
> @nelson-liu, @thesamuel, we're not sure what this is waiting on. Do either of you know?

As far as I know, there's nothing else left to resolve.

thesamuel:
@nelson-liu Awesome, thanks! 

Also, could you double check whether I need to change my commit message? I tried to format it the way GitHub is going to, before I realized it was automatic.

nelson-liu:
I think that GitHub generates the commit message from the PR title when squashing, so things should be fine. Thanks for thinking of it, though!

PR #2554, opened 2019-03-01
Make `load_archive` operate on serialization directories.

brendan-ai2:
- Fixes https://github.com/allenai/allennlp/issues/1052.
- Files like ELMo weight files aren't added to the serialization dir, but are added to the archive with `add_file_to_archive`.
- This results in misleading errors when using `load_archive` as it will attempt to load a serialization dir.
- Solution: Retain original paths when we're loading from a directory.
- Drive by fix for overrides not overriding.

### DIFF
--- a/allennlp/models/archival.py
+++ b/allennlp/models/archival.py
@@ -175,10 +175,10 @@ def load_archive(archive_file: str,
         logger.info(f"loading archive file {archive_file} from cache at {resolved_archive_file}")
 
     if os.path.isdir(resolved_archive_file):
-        loading_dir = True
+        archive_is_directory = True
         serialization_dir = resolved_archive_file
     else:
-        loading_dir = False
+        archive_is_directory = False
         # Extract archive to temp dir
         tempdir = tempfile.mkdtemp()
         logger.info(f"extracting archive file {resolved_archive_file} to temp dir {tempdir}")
@@ -192,7 +192,7 @@ def load_archive(archive_file: str,
 
     # Check for supplemental files in archive
     fta_filename = os.path.join(serialization_dir, _FTA_NAME)
-    if not loading_dir and os.path.exists(fta_filename):
+    if not archive_is_directory and os.path.exists(fta_filename):
         with open(fta_filename, 'r') as fta_file:
             files_to_archive = json.loads(fta_file.read())
 
@@ -217,7 +217,7 @@ def load_archive(archive_file: str,
     if weights_file:
         weights_path = weights_file
     else:
-        if loading_dir:
+        if archive_is_directory:
             weights_path = os.path.join(serialization_dir, _DEFAULT_WEIGHTS)
         else:
             weights_path = os.path.join(serialization_dir, _WEIGHTS_NAME)
--- a/allennlp/models/archival.py
+++ b/allennlp/models/archival.py
@@ -175,10 +175,8 @@ def load_archive(archive_file: str,
         logger.info(f"loading archive file {archive_file} from cache at {resolved_archive_file}")
 
     if os.path.isdir(resolved_archive_file):
-        archive_is_directory = True
         serialization_dir = resolved_archive_file
     else:
-        archive_is_directory = False
         # Extract archive to temp dir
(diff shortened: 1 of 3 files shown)


joelgrus:
I don't think this is the right solution, for a couple of reasons:

1. this problem applies to other commands too
2. we do want to support loading from a directory that really corresponds to the contents of an archive file; the problem here was that it didn't

for #1 that suggests that the right place for the fix is in `load_archive`
for #2 I think a better fix is to go where it's doing the fta replacement

https://github.com/allenai/allennlp/blob/master/allennlp/models/archival.py#L200

and actually checking if those files exist, and giving a very explicit warning if they don't

I would also write a test to verify this behavior

joelgrus:
also, probably this belongs in a different PR, but it's not good that using `overrides` didn't work here

brendan-ai2:
Good points! Given those, it seemed about as easy to make `load_archive` work on directories, so I did that. Hopefully that wasn't misguided... Fixed the `overrides` bug too.

julianmichael:
Files only appear at `fta/{key}` _inside_ the archive—not in the serialization directory—right? Unless I'm mistaken or this has changed, that means this fix will fail when running from a normal serialization directory. I think it should fall back to `original_filename` (as an absolute path) when the file is not found to cover this case.

I had to work around the bug myself recently and this behavior was very convenient for when I had to fix up some config files (because it was over NFS and editing/expanding/compressing huge tarballs was very slow).

brendan-ai2:
@julianmichael, I think I already accounted for that with this check: https://github.com/allenai/allennlp/pull/2554/files#diff-9dff246a3a033f8a1e8a2c409ef17e27R195 See also the test https://github.com/allenai/allennlp/pull/2554/files#diff-3113ca8bfda80819613f55d7a7a51f7aR147.

Please let me know if I'm missing something.

julianmichael:
Ah I see now. It also seems the trainer behavior has changed so it puts the best weights at `weights.th` instead of `best.th`. Now I can see that this would work. Sorry about that—thanks!

brendan-ai2:
No worries! Would have been an easy bug to write, so such feedback is appreciated. :)

allennlp/models/archival.py
```diff
@@ -175,8 +175,10 @@ def load_archive(archive_file: str,
         logger.info(f"loading archive file {archive_file} from cache at {resolved_archive_file}")
 
     if os.path.isdir(resolved_archive_file):
+        loading_dir = True
```
joelgrus: I would slightly prefer a more descriptive name like `loading_from_directory` or `archive_is_directory` or something like that. `loading_dir` sounds like it's the directory itself

allennlp/models/archival.py
```diff
@@ -190,18 +192,22 @@ def load_archive(archive_file: str,
 
     # Check for supplemental files in archive
     fta_filename = os.path.join(serialization_dir, _FTA_NAME)
-    if os.path.exists(fta_filename):
+    if not loading_dir and os.path.exists(fta_filename):
```
joelgrus: is this the right logic? imagine that I have a model.tar.gz, and then I untar it somewhere and then point to that location. in that case we'd have `loading_dir == True` but we'd still want to grab the files from the "archive" (which is a directory)

allennlp/models/archival.py
brendan-ai2: Done

allennlp/models/archival.py
brendan-ai2: Good point. It does seem possible that some users are relying on this. I've changed the behavior again to act as it did before if the files are present, but to load their serialization directory equivalents otherwise.

(I also considered `touch`ing an `ARCHIVE_MARKER` or `SERIALIZATION_DIR_MARKER` that we could check explicitly, but it seem desirable to have this work for existing dirs and archives.)

brendan-ai2:
Thanks @DeNeutoy for the link and @joelgrus for the review!

DeNeutoy:
Nice one, this has been a long running annoying issue, good to clear it out!

PR #2557, opened 2019-03-01
Handle edge cases in beam search

### LINKED ISSUE #2486: Beam search fails if empty sequence is predicted
**Describe the bug**
Beam search fails if empty sequence is predicted

**To Reproduce**
At prediction time, if the first prediction of the model is `self._end_index` beam search is aborted due to
```python
# If every predicted token from the last step is `self._end_index`,
# then we can stop early.
if (last_predictions == self._end_index).all():
    break
```
which leads to `backpointers` being empty and triggering
```
cur_backpointers = backpointers[-1]
IndexError: list index out of range
```

**Expected behavior**
Evaluation should not crash if this is the case.

 - OS: Linux
 - Python version: 3.6.8
 - AllenNLP version: 0.8.1
 - PyTorch version: 1.0.1

**Additional context**
I fixed it locally by changing the line to `if (last_predictions == self._end_index).all() and timestep > 0:`. Is this correct, or does it lead to other problems?

epwalsh:
Addresses the edge case brought up in #2486 (fixes #2486) as well as another. This actually turned out to be a little more nuanced...

I first thought the issue brought up was caused by `start_predictions` being the `end_index`, but it actually occurs in general with a beam size of 1 when the first predictions that the step function produces are the `end_index`, regardless of what `start_predictions` are, i.e. at this line: 

`start_class_log_probabilities, state = step(start_predictions, start_state)`

The other edge case is similar, and occurs when the beam size is smaller than the number of valid (non-zero probability) transitions that the step function produces. For example, this could happen in a semantic parsing task where a masked log softmax is used to create predicted log probs for valid next actions. Though this doesn't cause the beam search to crash per se, I thought it would still be good to warn the user in these cases since some of the predicted sequences may be improbable.

### DIFF
--- a/allennlp/nn/beam_search.py
+++ b/allennlp/nn/beam_search.py
@@ -1,4 +1,5 @@
 from typing import List, Callable, Tuple, Dict
+import warnings
 
 import torch
 
@@ -110,6 +111,11 @@ class BeamSearch:
         # shape: (batch_size, beam_size), (batch_size, beam_size)
         start_top_log_probabilities, start_predicted_classes = \
                 start_class_log_probabilities.topk(self.beam_size)
+        if self.beam_size == 1 and (start_predicted_classes == self._end_index).all():
+            warnings.warn("Empty sequences predicted. You may want to increase the beam size or ensure "
+                          "your step function is working properly.",
+                          RuntimeWarning)
+            return start_predicted_classes.unsqueeze(-1), start_top_log_probabilities
 
         # The log probabilities for the last time step.
         # shape: (batch_size, beam_size)
@@ -166,9 +172,9 @@ class BeamSearch:
                     class_log_probabilities
             )
 
+            # shape (both): (batch_size * beam_size, per_node_beam_size)
             top_log_probabilities, predicted_classes = \
                 cleaned_log_probabilities.topk(self.per_node_beam_size)
-            # shape (both): (batch_size * beam_size, per_node_beam_size)
 
             # Here we expand the last log probabilities to (batch_size * beam_size, per_node_beam_size)
             # so that we can add them to the current log probs for this timestep.
@@ -226,6 +232,12 @@ class BeamSearch:
                         reshape(batch_size, self.beam_size, *last_dims).\
                         gather(1, expanded_backpointer).\
                         reshape(batch_size * self.beam_size, *last_dims)
+
+        if not torch.isfinite(last_log_probabilities).all():
+            warnings.warn("Infinite log probabilities encountered. Some final sequences may not make sense. "
+                          "This can happen when the beam size is larger than the number of valid (non-zero "
+                          "probability) transitions that the step function produces.",
+                          RuntimeWarning)
 
         # Reconstruct the sequences.
         # shape: [(batch_size, beam_size, 1)]
--- a/allennlp/nn/beam_search.py
+++ b/allennlp/nn/beam_search.py
@@ -49,6 +49,16 @@ class BeamSearch:
         Given a starting state and a step function, apply beam search to find the
         most likely target sequences.
 
+        Notes
+        -----
+        If your step function returns ``-inf`` for some log probabilities
+        (like if your using a masked log-softmax) then some of the "best"
+        sequences returned may also have ``-inf`` log probability. Specifically
+        this happens when the beam size is smaller than the number of actions
+        with finite log probability (non-zero probability) returned by the step function.
+        Therefore if your using a mask you may want to check the results ``search``
+        and potentially discard sequences with non-finite log probability.
+
         Parameters
         ----------
         start_predictions : ``torch.Tensor``
--- a/allennlp/nn/beam_search.py
+++ b/allennlp/nn/beam_search.py
@@ -52,7 +52,7 @@ class BeamSearch:
         Notes
         -----
         If your step function returns ``-inf`` for some log probabilities
-        (like if your using a masked log-softmax) then some of the "best"
+        (like if you're using a masked log-softmax) then some of the "best"
         sequences returned may also have ``-inf`` log probability. Specifically
         this happens when the beam size is smaller than the number of actions
         with finite log probability (non-zero probability) returned by the step function.
--- a/allennlp/nn/beam_search.py
+++ b/allennlp/nn/beam_search.py
@@ -56,7 +56,7 @@ class BeamSearch:
         sequences returned may also have ``-inf`` log probability. Specifically
         this happens when the beam size is smaller than the number of actions
         with finite log probability (non-zero probability) returned by the step function.
-        Therefore if your using a mask you may want to check the results ``search``
+        Therefore if you're using a mask you may want to check the results from ``search``
         and potentially discard sequences with non-finite log probability.
 
         Parameters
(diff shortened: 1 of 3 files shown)


matt-gardner:
@brendan-ai2, when you have time for this, I think you have the most context here.

allennlp/nn/beam_search.py
```diff
@@ -110,6 +111,11 @@ def search(self,
         # shape: (batch_size, beam_size), (batch_size, beam_size)
         start_top_log_probabilities, start_predicted_classes = \
                 start_class_log_probabilities.topk(self.beam_size)
+        if self.beam_size == 1 and (start_predicted_classes == self._end_index).all():
```
brendan-ai2: Could you elaborate on why this only affects beams of size one? A test demonstrating that it can't happen with a beam size of two would be ideal. Sorry if I'm missing something obvious...

allennlp/tests/nn/beam_search_test.py
```diff
@@ -167,3 +167,24 @@ def test_catch_bad_config(self):
         beam_search = BeamSearch(self.end_index, beam_size=20)
         with pytest.raises(ConfigurationError):
             self._check_results(beam_search=beam_search)
+
+    def test_warn_for_bad_log_probs(self):
+        # The only valid next step from the initial predictions is the end index.
+        # But with a beam size of 3, the call to `topk` to find the 3 most likely
+        # next beams will result in 2 new beams that are invalid, in that have probability of 0.
+        # The beam search should warn us of this.
+        initial_predictions = torch.tensor([self.end_index-1, self.end_index-1])  # pylint: disable=not-callable
```
brendan-ai2: You can avoid that pylint warning with `torch.LongTensor` I believe.

allennlp/nn/beam_search.py
```diff
@@ -227,6 +233,12 @@ def search(self,
                         gather(1, expanded_backpointer).\
                         reshape(batch_size * self.beam_size, *last_dims)
 
+        if not torch.isfinite(last_log_probabilities).all():
+            warnings.warn("Infinite log probabilities encountered. Some final sequences may not make sense. "
+                          "This can happen when the beam size is larger than the number of valid (non-zero "
+                          "probability) transitions that the step function produces.",
+                          RuntimeWarning)
```
brendan-ai2: Hmm, this is a bit unsatisfying, but I guess filtering out the impossible results thereby changing the shape of the returned tensor might be a worse fix. Could you add to the docstring that it's the responsibility of the caller to check the returned probabilities for invalid results?

allennlp/nn/beam_search.py
epwalsh: So with a beam size > 1, `topk` of course returns multiple actions, so you are guaranteed to have at least one non-empty beam (since only one index returned from `topk` could be the `end_index`). Therefore the bug described in #2486 is avoided with beam size > 1 (i.e. it was only possible for that bug to happen with beam size 1). FWIW I did test that it would fail with a beam size of 2 before implementing the explicit check.

allennlp/nn/beam_search.py
epwalsh: ✅ added to the docstring

epwalsh:
Looks like the last build was manually interrupted, I'm not sure why.

matt-gardner:
I know there was some kind of restart that happened a day or two ago; not sure if this was related or not.  I just restarted the build.

epwalsh:
@matt-gardner thanks!

allennlp/nn/beam_search.py
brendan-ai2: Thanks for elaborating!

allennlp/nn/beam_search.py
```diff
@@ -48,6 +49,16 @@ def search(self,
         Given a starting state and a step function, apply beam search to find the
         most likely target sequences.
 
+        Notes
+        -----
+        If your step function returns ``-inf`` for some log probabilities
+        (like if your using a masked log-softmax) then some of the "best"
```
brendan-ai2: nit: you're

allennlp/nn/beam_search.py
```diff
@@ -48,6 +49,16 @@ def search(self,
...
+        sequences returned may also have ``-inf`` log probability. Specifically
+        this happens when the beam size is smaller than the number of actions
+        with finite log probability (non-zero probability) returned by the step function.
+        Therefore if your using a mask you may want to check the results ``search``
```
brendan-ai2: nit: "you're", "the results of ``search``"

allennlp/tests/nn/beam_search_test.py
```diff
@@ -167,3 +167,24 @@ def test_catch_bad_config(self):
...
         with pytest.raises(ConfigurationError):
             self._check_results(beam_search=beam_search)
+
+    def test_warn_for_bad_log_probs(self):
+        # The only valid next step from the initial predictions is the end index.
+        # But with a beam size of 3, the call to `topk` to find the 3 most likely
+        # next beams will result in 2 new beams that are invalid, in that have probability of 0.
+        # The beam search should warn us of this.
+        initial_predictions = torch.LongTensor([self.end_index-1, self.end_index-1])
+        with pytest.warns(RuntimeWarning):
```
brendan-ai2: Please use something like `pytest.warns(RuntimeWarning, match="foo")` to distinguish between this and the case below.

allennlp/tests/nn/beam_search_test.py
```diff
@@ -167,3 +167,24 @@ def test_catch_bad_config(self):
...
+            self.beam_search.search(initial_predictions, {}, take_step)
+
+    def test_empty_sequences(self):
+        initial_predictions = torch.tensor([self.end_index-1, self.end_index-1])  # pylint: disable=not-callable
```
brendan-ai2: Same thing here, `torch.LongTensor` to please pylint.

allennlp/nn/beam_search.py
```diff
@@ -110,6 +121,11 @@ def search(self,
         # shape: (batch_size, beam_size), (batch_size, beam_size)
         start_top_log_probabilities, start_predicted_classes = \
                 start_class_log_probabilities.topk(self.beam_size)
+        if self.beam_size == 1 and (start_predicted_classes == self._end_index).all():
+            warnings.warn("Empty sequences predicted. You may want to increase the beam size or ensure "
+                          "your step function is working properly.",
+                          RuntimeWarning)
+            return start_predicted_classes.unsqueeze(-1), start_top_log_probabilities
```
brendan-ai2: Should we `expand` the beam dimension here to ensure we agree with the contract we set in the docstring?

epwalsh: Hmm maybe I'm not seeing this right but I believe the dimensions returned are correct. I test for this in the test case `test_empty_sequences`.

epwalsh: The returned tensors will have shape `(batch_size, beam_size, sequence_length) = (batch_size, 1, 1)` and `(batch_size, beam_size) = (batch_size, 1)`

brendan-ai2: Sorry, got mixed up again. Definitely correct as-is.

PR #2559, opened 2019-03-01
Add official DROP evaluation script

matt-gardner:
This is me taking and cleaning up the evaluation script that @dDua wrote, from the private DROP repo, along with tests.  It has already had two pairs of eyes on it.


matt-gardner:
Yes, we do.  Maybe @dDua can add one with some of her examples in a subsequent PR?

Issue #2555, opened 2019-03-01
Tutorial doesn't work on GPU


cristipp:
**Describe the bug**
The tutorial code at https://allennlp.org/tutorials doesn't work out of the box on GPU. For reference, to make it work:

```
    if torch.cuda.is_available():
        cuda_device = 0
        model = model.cuda(cuda_device)
    else:
        cuda_device = -1

    trainer = Trainer(model=model,
                      optimizer=optimizer,
                      iterator=iterator,
                      train_dataset=train_dataset,
                      validation_dataset=validation_dataset,
                      patience=10,
                      num_epochs=1000,
                      cuda_device=cuda_device)
```

**To Reproduce**
1. Copy/Paste the tutorial code from https://allennlp.org/tutorials.
2. Add cuda_device=0 as an argument on Trainer ctor.
3. Run on a GPU-enabled machine.
4. Observe `RuntimeError: Expected object of backend CPU but got backend CUDA for argument #3 'index'`

**Expected behavior**
The model should train.

**System (please complete the following information):**
 - OS: Linux
 - Python version: 3.6.8
 - AllenNLP version: 0.8.2
 - PyTorch version: 1.0.1.post2

remotejob:
Completely the same for me. As well I tried  downgrade torch=1.0.0 don't help

remotejob:
Downgrade AllenNLP==0.8.0 help!!
So we must wait till 0.8.3 I hope :)

schmmd:
Thanks @cristipp, hope you're doing well!  Closing since this was fixed by @brendan-ai2 in #2570.

PR #2570, opened 2019-03-07
Make tutorial use GPU if available.

brendan-ai2:
- For https://github.com/allenai/allennlp/issues/2555.
- Add a test for the tutorial.


brendan-ai2:
@joelgrus, curious to get your thoughts on this. Practically speaking most users will need to use GPUs, so it seems reasonable to add them to the tutorial. Additional complication is pretty minor.

I'm also going to try adding a K80 to the TC box, so we can periodically catch GPU regressions.

brendan-ai2:
Thanks!

PR #2574, opened 2019-03-07
Fix unit test to work with GPUs.

brendan-ai2:
- The error the test matches against is different depending on whether you have GPUs available.
- Simply make the match more general.


brendan-ai2:
Thanks!

PR #2580, opened 2019-03-08
disable tutorial test

joelgrus:
it's breaking the master commit build (on account of the tutorial code not getting pip installed), and it's a really slow test, so I'm disabling it for now


schmmd:
@brendan-ai2 Joel had some concerns about the length of time this test was taking (although he liked having the test).  It might be worth discussing alternatives with Joel before fixing and restoring the test.

PR #2581, opened 2019-03-08
context manager that allows predictor to capture model internals

joelgrus:
as I said, this approach is slightly on the "cute" side, but I like how utterly non-intrusive it is

### DIFF
--- a/allennlp/predictors/predictor.py
+++ b/allennlp/predictors/predictor.py
@@ -1,5 +1,6 @@
 from typing import List
 import json
+from contextlib import contextmanager
 
 from allennlp.common import Registrable
 from allennlp.common.checks import ConfigurationError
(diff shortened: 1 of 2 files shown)


matt-gardner:
I think this is pretty slick.  I have no complaints with this approach.  One issue I see in using this in a demo is deciding which things to show to the user.  That will have to be configured somewhere (e.g., you might consider taking a filter function or something as an argument to the context manager).  But getting the internals this way seems great.

matt-gardner:
There's also the issue of wanting to get internals that aren't the result of submodules, but this is definitely an improvement over the current situation.  Most things we would want to include in a demo are probably the output of a module, anyway.

allennlp/predictors/predictor.py
```diff
@@ -54,6 +55,40 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
...
+
+        for idx, module in enumerate(self._model.modules()):
+            if module != self._model:
+                hook = module.register_forward_hook(add_output(idx))
+                hooks.append(hook)
+
+        # If you capture the return value of the context manager, you get the results dict.
+        yield results
+
+        # And then when you exit the context we remove all the hooks.
```
matt-gardner: Thanks for these comments - I'm not familiar with how context managers work, so this helps a lot.

PR #2584, opened 2019-03-09
Move some scripts to `allennlp/allennlp/tools`

### LINKED ISSUE #2575: [Suggestion] Is it a better idea to put scripts in `allennlp/allennlp/scripts` instead of `allennlp/scripts`?
When I want to use `archive_surgery.py`, I have to change directory to `allennlp`.
Instead, it may be more convenient for users to call this script like `python -m allennlp.scripts.archive_surgery --help`, which is also easily implemented for us.

I have gone through the directory `allennlp/scripts`.
There are three kinds of scripts:
 - about building or testing, such as: `check_docs.py, verify.py, mypy.sh`, etc.
 - for data processing, such as: `convert_openie_to_conll.py, regenerate_archived_models.py, examine_sql_coverage.py`, etc.
 - other useful utilities, such as: `archive_surgery.py, build_demo.py, inspect_cache.py`, etc.

As far as I am concerned, it is reasonable to put *data processing scripts* and *other utilities* in `allennlp/allennlp/scripts` instead of `allennlp/scripts`, because they are useful for common users.

What do you think?
 
None.

**Additional context**
I am willing to create PR for it.

WrRan:
Fixes #2575 

For now, I have moved those scripts as follows:
 - `archive_surgery.py`: Perform surgery on a model.tar.gz archive; 
 - `create_elmo_embeddings_from_vocab.py`: Generate CNN representations for a vocabulary using ELMo;
 - `inspect_cache.py`: Inspect datasets in the default cache directory.

Also, I think it is a good idea to move those scripts listed below for data processing into `allennlp/allennlp/tools`:
 - `compile_coref_data.sh`: This script downloads and compiles the Ontonotes 2012 data in a helpful format for co-reference resolution;
- `convert_openie_to_conll.py`: Convert Open IE4 extractions to CoNLL (ontonotes) format;
- `examine_sql_coverage.py`: Check the coverage of a SQL Grammar on the text2sql datasets;
- `reformat_text2sql_data.py`: Process text2sql data into a more readable format;
- `write_srl_predictions_to_conll_format.py`: Write conll format srl predictions to file from a pretrained model.

But I want to listen to you before moving these data-processing scripts.


DeNeutoy:
Hi @WrRan:

 - `compile_coref_data.sh`: Please don't move this - it is a one off script and the commands it uses require it to be run in a python 2.7 environment, which we don't support.
- `convert_openie_to_conll.py`: No opinion on moving this script, might be nice to add this functionality to the openie model if it is directly useful.
- `examine_sql_coverage.py`: Please don't move this script, it is experimental and not supported.
- `reformat_text2sql_data.py`: Please don't move this script, it is experimental and one-off.
- `write_srl_predictions_to_conll_format.py`: No opinions on this script, it would be nice if the SRL model could output the conll format directly, rather than requiring this.

Thanks!

WrRan:
Sorry about the delayed feedback.

And thank you. @matt-gardner @DeNeutoy 
Your feedback helped me deal with this issues better.

As for the two scripts `convert_openie_to_conll.py` and `write_srl_predictions_to_conll_format.py`, I can't guarantee their correctness because there are no relevant test cases and test data in the current codebase. Therefore, I decided not to move them.

In addition, there are also non-python files in the `allennlp.tools` folder. This may not be expected. (see https://github.com/allenai/allennlp/blob/master/allennlp/tools/srl-eval.pl)

PR #2588, opened 2019-03-10
Add semisupervised text classification dataset reader

kernelmachine:
This PR adds support for semisupervised text classification datasets. 

This dataset reader inherits from TextClassificationJSONReader, but differs from its parent
    in that it is primed for semisupervised learning. Specifically, it allows for:

* Ignoring labels in the training data (e.g. for unsupervised pretraining)
* Reading additional unlabeled data from another file
* Throttling the training data to a random subsample (according to the numpy seed), for analysis of the effect of semisupervised models on different amounts of labeled data


kernelmachine:
I'm actually closing this, because on second thought, it's probably not fit to merge into AllenNLP as a general dataset reader, since the use case is pretty narrow.

PR #2591, opened 2019-03-11
Add text classification model

kernelmachine:
This PR adds a generic text classifier to AllenNLP.

After embedding the text into a text field, this model will optionally encode the embeddings with a Seq2VecEncoder and then pass the embeddings to a linear classification layer, which projects into the label space.


allennlp/models/classifier.py
```diff
@@ -0,0 +1,113 @@
...
+        else:
+            self._encoder = None
+            self._classifier_input_dim = self._text_field_embedder.get_output_dim()
+
+        if dropout:
+            self._dropout = torch.nn.Dropout(dropout)
+        else:
+            self._dropout = None
+
+        self._num_labels = vocab.get_vocab_size(namespace="labels")
```
joelgrus: I think this should be more configurable, definitely you should be able to specify a different namespace, possibly you should be allowed to set it explicitly as well?

allennlp/models/classifier.py
```diff
@@ -0,0 +1,113 @@
...
+        # pylint: disable=arguments-differ
+        """
+        Parameters
+        ----------
+        tokens : Dict[str, torch.LongTensor]
+            From a ``TextField``
+        label : torch.IntTensor, optional (default = None)
+            From a ``LabelField``
+        metadata : ``List[Dict[str, Any]]``, optional, (default = None)
+            Metadata to persist
```
joelgrus: but you're not actually persisting the metadata?

allennlp/models/classifier.py
```diff
@@ -0,0 +1,113 @@
...
+            probabilities of the label.
+        loss : torch.FloatTensor, optional
+            A scalar loss to be optimised.
+        """
+        embedded_text = self._text_field_embedder(tokens)
+        mask = get_text_field_mask(tokens).float()
+
+        if self._encoder:
+            embedded_text = self._encoder(embedded_text,
+                                          mask=mask)
```
joelgrus: this could fit on one line

kernelmachine:
@joelgrus ready for another look.

kernelmachine:
@joelgrus updated, ready for your comments

allennlp/models/basic_classifier.py
```diff
@@ -0,0 +1,128 @@
...
+        loss : torch.FloatTensor, optional
+            A scalar loss to be optimised.
+        """
+        embedded_text = self._text_field_embedder(tokens)
+        mask = get_text_field_mask(tokens).float()
+
+        if self._seq2seq_encoder:
+            embedded_text = self._seq2seq_encoder(embedded_text, mask=mask)
+
+        if self._seq2vec_encoder:
```
joelgrus: isn't this required now?

PR #2593, opened 2019-03-11
Re-use .allennlp when running Docker commands

schmmd:


schmmd:
Worth the complexity?

README.md
```diff
@@ -109,10 +109,11 @@ Once you have [installed Docker](https://docs.docker.com/engine/installation/)
 just run the following command to get an environment that will run on either the cpu or gpu.
 
    ```bash
-   docker run -it -p 8000:8000 --rm allennlp/allennlp:v0.8.2
+   mkdir -p $HOME/.allennlp/
+   docker run --rm -v $HOME/.allennlp:/root.allennlp allennlp/allennlp:v0.8.2
```
brendan-ai2: I'm not a Docker expert, is this really supposed to be `/root.allennlp` and not `/root/.allennlp`?

brendan-ai2:
Might save people time in some cases. Downside is maybe we'd hit weird versioning issues. But I guess this happens locally now. I'm impartial.

schmmd:
I accidentally deleted the slash at some point.

schmmd:
I don't worry about the versioning issue, but I do worry about the error people get if the directory doesn't exist on the host.  That said, downloading models can be quite painful (almost as painful as re-downloading them).

README.md
```diff
@@ -109,10 +109,11 @@ Once you have [installed Docker](https://docs.docker.com/engine/installation/)
 just run the following command to get an environment that will run on either the cpu or gpu.
 
    ```bash
-   docker run -it -p 8000:8000 --rm allennlp/allennlp:v0.8.2
+   mkdir -p $HOME/.allennlp/
+   docker run --rm -v $HOME/.allennlp:/root/.allennlp allennlp/allennlp:v0.8.2
    ```
 
-You can test the Docker environment with `docker run -it -p 8000:8000 --rm allennlp/allennlp:v0.8.2 test-install`.
+You can test the Docker environment with `docker run --rm -v $HOME/.allennlp:/root.allennlp allennlp/allennlp:v0.8.2 test-install`.
```
brendan-ai2: Here too. Please test both of these commands.

README.md
```diff
@@ -109,10 +109,11 @@ Once you have [installed Docker](https://docs.docker.com/engine/installation/)
 just run the following command to get an environment that will run on either the cpu or gpu.
 
    ```bash
-   docker run -it -p 8000:8000 --rm allennlp/allennlp:v0.8.2
+   mkdir -p $HOME/.allennlp/
```
brendan-ai2: 👍

schmmd:
The problem is that it's a bit nuanced to test the commands--since just running them as is won't validate that the directory is being mounted correctly.  I actually ran both of the commands previously--but that didn't catch my mistake.  Just now I ran the first with `--it --entrypoint bash` and verified manually that the folder I expected existed (and was populated).

PR #2594, opened 2019-03-12
bug fixes in drop evaluation and more test cases

dDua:
Fixed bugs:
 -- when predictions are none
 -- removed unnecessary dictionary usage in "answer"

Added test cases for end to end F1 and EM measure with json loading

### DIFF
--- a/allennlp/tests/tools/drop_eval_test.py
+++ b/allennlp/tests/tools/drop_eval_test.py
@@ -73,30 +73,35 @@ class TestDropEvalGetMetrics:
         assert get_metrics(["ottoman", "Kantakouzenous"],
                            ["ottoman", "army of Kantakouzenous"]) == (0.0, 0.75)
 
+class TestDropEvalFunctional:
     def test_json_loader(self):
-        annotation = {"pid1": {"passage": "Loren ipsum..",
-                               "qa_pairs": [{"query_id": "qid1", "question": "text1", "answer": {"number": "0"}}]}}
-        prediction = {"qid1": "0"}
+        annotation = {"qid1": {"answer": {"number": "1"}, "validated_answers": [{"number": "0"}]}}
+        prediction = {"qid1": "1"}
         assert evaluate_json(annotation, prediction) == (1.0, 1.0)
 
-        annotation = {"pid1": {"passage": "Loren ipsum..",
-                               "qa_pairs": [{"query_id": "qid1", "question": "text1", "answer": {"number": "0"}}]}}
-        prediction = {"qid0": "0"}
-        assert evaluate_json(annotation, prediction) == (0.0, 0.0)
+        annotation = {"qid1": {"answer": {"spans": ["2"]}, "validated_answers": [{"number": "2"}]}}
+        prediction = {"qid1": "2"}
+        assert evaluate_json(annotation, prediction) == (1.0, 1.0)
 
-        annotation = {"pid1": {"passage": "Loren ipsum..",
-                               "qa_pairs": [{"query_id": "qid1", "question": "text1", "answer": {"number": "0"}}]}}
-        prediction = {"qid1": "10"}
-        assert evaluate_json(annotation, prediction) == (0.0, 0.0)
+        annotation = {"qid1": {"answer": {"spans": ["0"]}, \
+                               "validated_answers": [{"number": "1"}, {"number": "2"}]}}
+        prediction = {"qid1": "1"}
+        assert evaluate_json(annotation, prediction) == (1.0, 1.0)
 
-        annotation = {"pid1": {"passage": "Loren ipsum..",
-                               "qa_pairs": [{"query_id": "qid1", "question": "text1",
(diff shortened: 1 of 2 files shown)


allennlp/tools/drop_eval.py
```diff
@@ -145,11 +152,11 @@ def answer_json_to_strings(answer: Dict[str, Any]) -> Tuple[Tuple[str, ...], str
...
+    if "number" in answer and answer["number"]:
         return tuple([str(answer["number"])]), "number"
-    elif answer["spans"]:
+    elif "spans" in answer and answer["spans"]:
         return tuple(answer["spans"]), "span" if len(answer["spans"]) == 1 else "spans"
-    else:
+    elif "date" in answer:
         return tuple(["{0} {1} {2}".format(answer["date"]["day"],
                                            answer["date"]["month"],
                                            answer["date"]["year"])]), "date"
```
matt-gardner: Mypy is complaining because you don't have an `else` in here that either returns `None` or raises an error.

allennlp/tools/drop_eval.py
```diff
@@ -179,21 +185,23 @@ def evaluate_json(annotations: Dict[str, Any], predicted_answers: Dict[str, Any]
...
-                predicted = None
-            for answer in [qa_pair["answer"]] + qa_pair["validated_answers"]:
-                gold_answer, gold_type = answer_json_to_strings(answer)
-                em_score, f1_score = get_metrics(predicted, gold_answer)
-                if gold_answer[0].strip() != "":
-                    max_em_score = max(max_em_score, em_score)
-                    max_f1_score = max(max_f1_score, f1_score)
-                    if max_em_score == em_score or max_f1_score == f1_score:
-                        max_type = gold_type
+                _, max_type = answer_json_to_strings(qa_pair["answer"])
```
matt-gardner: It looks like mypy is complaining about re-assigning a different type to `_` :(.  If you call this `__` instead mypy should stop complaining.

allennlp/tools/drop_eval.py
```diff
@@ -242,3 +250,6 @@ def evaluate_prediction_file(prediction_path: str, gold_path: str) -> Tuple[floa
                         help='location of the prediction file')
     args = parser.parse_args()
     evaluate_prediction_file(args.prediction_path, args.gold_path)
+
```
matt-gardner: This, and other issues, are making pylint complain.  You can check the pylint issues locally by running `./scripts/pylint.sh`.  Be sure that your editor uses unix line endings.

allennlp/tests/tools/drop_eval_test.py
```diff
@@ -72,3 +72,31 @@ def test_multi_span_overlap_in_incorrect_cases(self):
         # macro-averaging F1 over spans
         assert get_metrics(["ottoman", "Kantakouzenous"],
```
matt-gardner: Can you add another test that has "answer" in the answer string and fails before this PR, just to be sure we don't have a regression in the future?

matt-gardner: I'm not sure if there were other issues that are fixed here, too; in general, you want a failing test case for every bug, that the PR then fixes.  This makes sure that we understand what the actual cause of the underlying bug was, and lets us be confident that we fixed it and it will stay fixed.

gabrielStanovsky:
Thanks @dDua !
Maybe we can also have a test checking specifically the error that we had when there are empty spans? 
I don't think I've seen one doing that, right?

allennlp/tests/tools/drop_eval_test.py
dDua: There was only "answer" bug that was causing the empty prediction span bug. I have added a test for that.

dDua:
> Thanks @dDua !
> Maybe we can also have a test checking specifically the error that we had when there are empty spans?
> I don't think I've seen one doing that, right?

The empty prediction span was due to format of prediction file which took "answer" as a key and computed empty string when "answer" was in prediction string. We have changed the format now so that shouldn't happen and I have added a test for the same.

allennlp/tools/drop_eval.py
```diff
@@ -145,14 +152,17 @@ def answer_json_to_strings(answer: Dict[str, Any]) -> Tuple[Tuple[str, ...], str
...
+    elif "spans" in answer and answer["spans"]:
         return tuple(answer["spans"]), "span" if len(answer["spans"]) == 1 else "spans"
-    else:
+    elif "date" in answer:
         return tuple(["{0} {1} {2}".format(answer["date"]["day"],
                                            answer["date"]["month"],
                                            answer["date"]["year"])]), "date"
+    else:
+        print("Answer type not found, should be one of number, spans or date at: {0}".format(json.dumps(answer)))
+        raise ValueError
```
matt-gardner: Minor style issue: the error message should be an argument to `ValueError()`, instead of a print statement beforehand.  Also, prefer python f-strings: `raise ValueError(f"Answer type ... date at: {json.dumps(answer)}")`.

PR #2607, opened 2019-03-14
make spacy word splitter return allennlp Tokens (now NamedTuples) by default

### LINKED ISSUE #1887: spacy Tokens are not pickleable, which breaks the multiprocess dataset reader
the multiprocess dataset reader uses multiple processes to generate allennlp `Instance`s and stick them on an output queue (which implicitly involves pickling them)

it turns out (have I ever mentioned how much I hate debugging multiprocessing code) that spacy Tokens are not pickleable, which causes the `queue.put()` operation to silently fail, which causes the multiprocess dataset reader to wait forever on an empty queue.

we didn't discover this when we implemented it because the test case used the `SequenceTaggingDatasetReader`, which assumes pre-tokenized text and so never generates spacy tokens.

I ended up with the following ugly workaround in my dataset reader:

```
tokenized = [Token(token.text) for token in self._tokenizer.tokenize(sentence)]
```

(that's the allennlp `Token` replacement there)

anyway, I don't know if there is a good solution other than not using spacy tokens in this case, but I'm opening an issue in case anyone else runs into this.

```
In [1]: import spacy

In [2]: nlp = spacy.load('en_core_web_sm')

In [3]: doc = nlp("can't pickle this")

In [4]: token = doc[0]

In [5]: import pickle

In [6]: pickle.dumps(token)
TypeError                                 Traceback (most recent call last)
<ipython-input-6-81a5065b9c9e> in <module>()
----> 1 pickle.dumps(token)

~/anaconda3/envs/allennlp/lib/python3.6/site-packages/spacy/tokens/token.cpython-36m-darwin.so in spacy.tokens.token.Token.__reduce_cython__()

TypeError: no default __reduce__ due to non-trivial __cinit__
```

cf https://github.com/explosion/spaCy/issues/606

### LINKED ISSUE #2602: [Suggestion] Option to use raw strings instead of Token objects in TextField
I've recently experienced major memory issues using allennlp. The problem stemmed from the fact that each token had to be stored in a `Token` object that incurred massive overhead. Removing the `Token` object reduced memory consumption to less than 1/3 of the original memory usage.

I think users should also be allowed to simply pass a sequence of strings as input to the text field instead of a list of `Tokens`. I have tried this approach on my own fork of allennlp and did not encounter any major problems: it was relatively simple to make the processing of the TextField agnostic to whether the input was a `Token` or a string, since most of the processing (if not all) extracted the text from the token anyway.

I don't think this will reduce the scope of what users can do, so it seems like the least restrictive option for the memory issue.

**Additional context**
I may be missing some of the significance of using the `Token` object. If so, I would love to learn more about this design decision!

Thanks!

joelgrus:
largely fixes #2602, also fixes #1887 

the discussion at #2602 gives a lot of the justification for this, small-scale experiments suggest that it saves a lot of memory if you have a lot of text.

it is a breaking change in two (minor) ways:

1. the SpacyWordSplitter now by default returns allennlp Tokens instead of spacy Tokens. in theory everyone should be programming to the shared abstraction, but it's possible that someone somewhere is relying on having the actual spacy token, in which case they'll need to configure their word splitter to keep them.

2. a NamedTuple can't have different constructor parameters and field names. Our previous Token implementation used e.g. `pos` as the name of the constructor argument but then `pos_` as the name of the field. Converting this to a namedtuple meant that the constructor argument now also has to be `pos_`. This required a tiny change to the wikitables dataset reader; it's possible (but not all that likely) that other people could have dataset readers that rely on the old behavior. My guess is that most people are only creating tokens via our word splitters, in which case they won't notice the difference.

This could be avoided if we left `Token` as an actual class, but we'd get less memory improvement that way.

### DIFF
--- a/allennlp/common/util.py
+++ b/allennlp/common/util.py
@@ -66,12 +66,12 @@ def sanitize(x: Any) -> Any:  # pylint: disable=invalid-name,too-many-return-sta
     elif isinstance(x, dict):
         # Dicts need their values sanitized
         return {key: sanitize(value) for key, value in x.items()}
-    elif isinstance(x, (list, tuple)):
-        # Lists and Tuples need their values sanitized
-        return [sanitize(x_i) for x_i in x]
     elif isinstance(x, (spacy.tokens.Token, allennlp.data.Token)):
         # Tokens get sanitized to just their text.
         return x.text
+    elif isinstance(x, (list, tuple)):
+        # Lists and Tuples need their values sanitized
+        return [sanitize(x_i) for x_i in x]
     elif x is None:
         return "None"
     elif hasattr(x, 'to_json'):
--- a/allennlp/data/dataset_readers/semantic_parsing/wikitables/wikitables.py
+++ b/allennlp/data/dataset_readers/semantic_parsing/wikitables/wikitables.py
@@ -409,7 +409,7 @@ class WikiTablesDatasetReader(DatasetReader):
 
     @staticmethod
     def _read_tokens_from_json_list(json_list) -> List[Token]:
-        return [Token(text=json_obj['text'], lemma=json_obj['lemma']) for json_obj in json_list]
+        return [Token(text=json_obj['text'], lemma_=json_obj['lemma']) for json_obj in json_list]
 
     @staticmethod
     def _should_keep_logical_form(logical_form: str) -> bool:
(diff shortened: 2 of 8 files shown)


joelgrus:
oops, the issue there was in the `sanitize` helper, basically it was checking for `tuple` before it was checking for `Token`, so the new namedtuple token triggered the "sanitize tuple" logic, which was wrong. putting the "Token" logic before the "tuple" logic fixed it.

I think all the tests should pass now, let's see.

joelgrus:
ok, I added it to the draft release notes for the next release

PR #2609, opened 2019-03-14
Feature Request: Add a `dtype` parameter to `ArrayField`

rloganiv:
This PR implements solution 1, proposed in https://github.com/allenai/allennlp/pull/2460#issuecomment-458989643.


allennlp/data/fields/array_field.py
```diff
@@ -13,8 +13,12 @@ class ArrayField(Field[numpy.ndarray]):
     A batch of these arrays are padded to the max dimension length in the batch
     for each dimension.
     """
-    def __init__(self, array: numpy.ndarray, padding_value: int = 0) -> None:
+    def __init__(self,
+                 array: numpy.ndarray,
```
DeNeutoy: Could you just swap the `dtype` argument to be the last argument, so it is fully backward compatible, and then this looks great!

rloganiv: Sure thing!

DeNeutoy:
Thanks for the contribution!

PR #2610, opened 2019-03-15
Add a no-op trainer.

brendan-ai2:
- Simply loads a model, creates the vocab and serializes without any training.
- Intended to be used principally for untrained baselines like majority class.


matt-gardner:
@pdasigi, FYI. This should be useful for creating a BERT model from hugging face pretrained stuff.

brendan-ai2:
PTAL, @joelgrus :)

allennlp/tests/commands/no_op_train_test.py
```diff
@@ -0,0 +1,36 @@
...
+                "validation_data_path": SEQUENCE_TAGGING_DATA_PATH,
+                "iterator": {"type": "basic", "batch_size": 2},
+                "trainer": {
+                        "type": "no_op"
+                }
+        })
+
+        serialization_dir = self.TEST_DIR / 'serialization_directory'
+        train_model(params(), serialization_dir=serialization_dir)
+        load_archive(str(serialization_dir / "model.tar.gz"))
```
joelgrus: should we add an assert just to make sure that the loaded model does what we expect it to do?

allennlp/tests/training/no_op_trainer_test.py
```diff
@@ -0,0 +1,31 @@
...
+        self.vocab = vocab
+        self.model = ConstantModel(vocab)
+
+    def test_trainer_serializes(self):
+        serialization_dir = self.TEST_DIR / "serialization_dir"
+        trainer = NoOpTrainer(serialization_dir=serialization_dir, model=self.model)
+        metrics = trainer.train()
+        assert metrics == {}
+        assert os.path.exists(serialization_dir / "best.th")
+        assert os.path.exists(serialization_dir / "vocabulary")
```
joelgrus: similarly, do you want to test here that the archived model is "correct"?

allennlp/training/no_op_trainer.py
```diff
@@ -0,0 +1,38 @@
...
+
+@TrainerBase.register("no_op")
+class NoOpTrainer(TrainerBase):
+    def __init__(self, serialization_dir: str, model: Model) -> None:
+        """
+        A trivial trainer to assist in making model archives for models that do not actually
+        require training. For instance, a majority class baseline.
+        """
+
+        super().__init__(serialization_dir, -1)
```
joelgrus: might be worth using a keyword arg `cuda_device=-1`, otherwise the -1 is just a magical number

allennlp/tests/commands/no_op_train_test.py
brendan-ai2: Added a basic assert. This shouldn't fail assuming it loads at all, but making the test fully end-to-end seems proper. Good call.

allennlp/tests/training/no_op_trainer_test.py
brendan-ai2: This I would prefer to elide. The correctness is guaranteed by the end-to-end test in `no_op_train_test.py` and the tests in `checkpointer_test.py`. Anything more here would be duplication.

allennlp/training/no_op_trainer.py
brendan-ai2: Done!

brendan-ai2:
Thanks for the review, @joelgrus! And the no-op trainer idea in the first place. :)

PR #2613, opened 2019-03-16
Fix 'cuda_device' docstring in Trainer.__init__

bryant1410:


allennlp/training/trainer.py
```diff
@@ -125,8 +125,8 @@ def __init__(self,
             If provided, then serialize models every ``model_save_interval``
             seconds within single epochs.  In all cases, models are also saved
             at the end of every epoch if ``serialization_dir`` is provided.
-        cuda_device : ``int``, optional (default = -1)
-            An integer specifying the CUDA device to use. If -1, the CPU is used.
+        cuda_device : ``int`` or List[int], optional (default = -1)
```
matt-gardner: This type should be:

```
``Union[int, List[int]]``
```

PR #2621, opened 2019-03-18
Fix TextClassificationJsonReader handling of unlabeled instances

### LINKED ISSUE #2619: Incorrect info in TextClassificationJsonReader docstring
The [docstring for the TextClassificationJsonReader](https://github.com/allenai/allennlp/blob/master/allennlp/data/dataset_readers/text_classification_json.py#L19-L20) says that it expects json instances with "tokens" and "category" keys.

However, it looks like it actually expects `"text"` and `"labels"` (see https://github.com/allenai/allennlp/blob/master/allennlp/data/dataset_readers/text_classification_json.py#L69 )

Would be good to fix the doc

nelson-liu:
ended up running into this myself, so this fixes #2619 
This fix is required to use the dataset reader while predicting on unlabeled text.
Also adds a small doc fix.


matt-gardner:
@nelson-liu, don't worry too much about merging into master all the time.  If you know that what you're doing doesn't conflict with other open PRs, it's not a big deal, and we can just override that check when merging.

nelson-liu:
ok, thanks @matt-gardner !

nelson-liu:
@matt-gardner want to override the "out-of-date with the base branch" check? I don't have the privileges.

PR #2624, opened 2019-03-18
fix processes at 1 because python can't determine virtualised cpus correctly

DeNeutoy:
Test to see if this speeds up pylint.


DeNeutoy:
huh, this makes pylint hang indefinitely .....

PR #2628, opened 2019-03-18
Multilingual parser and Cross-lingual ELMo

TalSchuster:
Support for cross-lingual alignment of ELMo and multilingual biaffine dependency parser as described in [Cross-Lingual Alignment of Contextual Word Embeddings,  with Applications to Zero-shot Dependency Parsing](https://arxiv.org/abs/1902.09492).

Modifications:

- Multilingual parser - supports inputs from different files and uses the matching embedder. Also, stores per language metrics.
-  Universal dependencies reader for multiple languages - alternates between the sources to read batches from each one.
- same language iterator - breaks batches, if necessary, to contain only a single language.
- ELMo multilang embedder - loads the parameters for each lanugage and an optional alignment matrix. On forward, applies the matching embedder by the batch language.
- Configuration example


matt-gardner:
Hi @TalSchuster, thanks for the PR! It looks like this could be a nice contribution. Any particular reason you closed it?

TalSchuster:
Hi @matt-gardner, thanks for the feedback!

Sorry for that, the first PR here was by mistake (instead of the fork) and then planned to get back to it to add some description got a bit busy. Will add it now.

matt-gardner:
Ok, great, let us know when you think this is ready for a look (I'm gathering from your comment that you're still working on something).

TalSchuster:
It is ready for review now.
I only see some pylint issues that I'm less familiar with but I'll try to figure that out.

matt-gardner:
Ok, great, we'll get on this (I'm totally booked tomorrow, and then I'm out next week, though - can someone else take this PR?).  Let us know if you have questions about any of the pylint or other stuff you see.

TalSchuster:
I did some formatting that solved most of the comments but there are still some things that pylint is not happy with, like changing the signature of a function (passing a dictionary instead of a string). I understand that it is not following the OOP conventions but I think that it should be a valid pythonic way (but I'm not an expert in this).
If someone can look at it and help to solve them it will be great. Thanks!

allennlp/data/dataset_readers/universal_dependencies_multilang.py
```diff
@@ -0,0 +1,171 @@
...
+    def __init__(self,
+                 token_indexers: Dict[str, TokenIndexer] = None,
+                 use_language_specific_pos: bool = False,
+                 lazy: bool = False,
+                 alternate: bool = True,
+                 is_first_pass_for_vocab: bool = True,
+                 batch_size: int = 32) -> None:
+        super().__init__(lazy)
+        self._token_indexers = token_indexers or {'tokens': SingleIdTokenIndexer()}
+        self.use_language_specific_pos = use_language_specific_pos
```
matt-gardner: We prefer to mark these things as private (with a preceding underscore), as then it's implied that users shouldn't rely on the presence of these fields, so it's easier for us to modify later without breaking backwards compatibility.  This applies to everything you've saved to `self` here.

allennlp/data/dataset_readers/universal_dependencies_multilang.py
```diff
@@ -0,0 +1,171 @@
...
+
+        self.is_first_pass = True
+        self.iterators = None
+        self.instances_per_lang = defaultdict(int)
+
+    def _read_one_file(self, lang: str, file_path: str):
+        with open(file_path, 'r') as conllu_file:
+            logger.info("Reading UD instances for %s language from conllu dataset at: %s", lang, file_path)
+
+            for annotation in  lazy_parse(conllu_file.read()):
```
matt-gardner: Extra space here (I'm surprised pylint didn't complain).

allennlp/data/dataset_readers/universal_dependencies_multilang.py
```diff
@@ -0,0 +1,171 @@
...
+    -------
+    A dictionary of identifier ("en","it" etc.) to file path.
+    """
+    return dict(file_paths)
+
+
+@DatasetReader.register("universal_dependencies_multilang")
+class UniversalDependenciesMultiLangDatasetReader(DatasetReader):
+    """
+    Wraps UniversalDependenciesDatasetReader to support multiple input files.
```
matt-gardner: You're not actually wrapping anything here, you're duplicating logic.

allennlp/models/biaffine_dependency_parser_multilang.py
```diff
@@ -0,0 +1,296 @@
...
+from allennlp.models.biaffine_dependency_parser import BiaffineDependencyParser
+from allennlp.nn import InitializerApplicator, RegularizerApplicator
+from allennlp.nn.util import get_text_field_mask
+from allennlp.training.metrics import AttachmentScores
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+@Model.register("biaffine_parser_multilang")
+class BiaffineDependencyParserMultiLang(BiaffineDependencyParser):
```
matt-gardner: You're inheriting from the above parser and not really adding anything to the constructor, but you're duplicating a whole lot of logic from the `forward` method (and reformatting it, so it's not as obvious that it's copied).  Can you just factor out the pieces that you need in the superclass to separate methods, and call those, instead of duplicating all of this logic?

allennlp/modules/token_embedders/elmo_token_embedder_multilang.py
```diff
@@ -0,0 +1,171 @@
...
+        super(ElmoTokenEmbedderMultiLang, self).__init__()
+
+        assert options_files.keys() == weight_files.keys()
+        output_dim = -1
+        for lang in weight_files.keys():
+            name = 'elmo_%s' % lang
+            elmo = Elmo(
+                    options_files[lang],
+                    weight_files[lang],
+                    1,
```
matt-gardner: This is a magic number - can you pass this as a keyword argument, so it's more obvious what `1` means here?

allennlp/modules/token_embedders/elmo_token_embedder_multilang.py
```diff
@@ -0,0 +1,171 @@
...
+                    do_layer_norm=do_layer_norm,
+                    dropout=dropout,
+                    requires_grad=requires_grad,
+                    vocab_to_cache=vocab_to_cache,
+                    scalar_mix_parameters=scalar_mix_parameters)
+            self.add_module(name, elmo)
+
+            output_dim_tmp = elmo.get_output_dim()
+            if output_dim != -1:
```
matt-gardner: Safer and more intuitive to have this be initialized to `None` instead of `-1` (note that pylint will want you to do `is not None` instead of `!= None`).

allennlp/modules/token_embedders/elmo_token_embedder_multilang.py
```diff
@@ -0,0 +1,171 @@
...
+                 do_layer_norm: bool = False,
+                 dropout: float = 0.5,
+                 requires_grad: bool = False,
+                 projection_dim: int = None,
+                 vocab_to_cache: List[str] = None,
+                 scalar_mix_parameters: List[float] = None,
+                 aligning_files: Dict[str, str] = {}) -> None:
+        super(ElmoTokenEmbedderMultiLang, self).__init__()
+
+        assert options_files.keys() == weight_files.keys()
```
matt-gardner: This looks to me like it should be a `ConfigurationError` instead of an assert.

allennlp/modules/token_embedders/elmo_token_embedder_multilang.py
```diff
@@ -0,0 +1,171 @@
...
+                    dropout=dropout,
+                    requires_grad=requires_grad,
+                    vocab_to_cache=vocab_to_cache,
+                    scalar_mix_parameters=scalar_mix_parameters)
+            self.add_module(name, elmo)
+
+            output_dim_tmp = elmo.get_output_dim()
+            if output_dim != -1:
+                # Verify that all ELMo embedders have the same output dimension.
+                assert output_dim_tmp == output_dim
```
matt-gardner: We also have a function for this, `check_dimensions_match`, which will make this a `ConfigurationError` and let you give a slightly more helpful message on, e.g., which languages are mismatching.

allennlp/modules/token_embedders/elmo_token_embedder_multilang.py
```diff
@@ -0,0 +1,171 @@
...
+
+            aligning = torch.nn.Linear(output_dim, output_dim, bias=False)
+            aligning.weight = torch.nn.Parameter(
+                    aligning_matrix, requires_grad=False)
+            self.add_module(name, aligning)
+
+    def get_output_dim(self):
+        return self.output_dim
+
+    def forward(
```
matt-gardner: The way you're formatting these arguments is inconsistent across functions, and inconsistent with the rest of the codebase.  Ok @brendan-ai2, @joelgrus - if we can find an automatic formatter that makes minimal changes to existing code, I'm probably ok with it.

allennlp/modules/text_field_embedders/basic_text_field_embedder.py
```diff
@@ -107,20 +109,20 @@ def forward(self, text_field_input: Dict[str, torch.Tensor], num_wrapping_dims:
                     # If `indexer_key` is None, we map it to `None`.
                     tensors = [(text_field_input[indexer_key] if indexer_key is not None else None)
                                for indexer_key in indexer_map]
-                    token_vectors = embedder(*tensors)
+                    token_vectors = embedder(*tensors, **kwargs)
```
matt-gardner: This is not safe.  You can't pass the same `kwargs` to all embedders, because not all of them are going to take the same extra arguments.  I can think of two options to handle this:

1. Do some inspection here to figure out which extra arguments are compatible with the specific `embedder` that you're looking at in the loop, and only pass those.
2. Instead of modifying this class, make a new `Indexer` that's a `MultilingualElmoIndexer` that adds a separate key to the token mapping (that would end up looking like `elmo-lang`, with a string value), like we do with the OpenAI and BERT indexers (e.g., see [here](https://github.com/allenai/allennlp/blob/79e2cf7b677f84daefaf63acc2a8e3833782de23/allennlp/data/token_indexers/openai_transformer_byte_pair_indexer.py#L221-L228) and [here](https://github.com/allenai/allennlp/blob/79e2cf7b677f84daefaf63acc2a8e3833782de23/allennlp/tests/fixtures/openai_transformer/config_small.jsonnet#L24-L27)).  This would require adding a field to the `Token` class that stores the language, so that you know which language you're using.

I'm not sure which option to go with.  Maybe slightly less work to go with option 1, and it seems like it might be potentially useful for other applications, also.  Option 2 makes it so you're not making API changes just to support this, and it doesn't look like much additional code, as you can basically just subclass `ElmoIndexer` and tweak one method.  Except that you have to add a field to `Token`...

training_config/multilang_dependency_parser.jsonnet
```diff
@@ -0,0 +1,142 @@
...
+    "validation_data_path": {
+        "de": "UNI_DEP_V2_PATH/universal/uni-dep-tb/universal_treebanks_v2.0/std/de/de-universal-dev.conll",
+        "en": "UNI_DEP_V2_PATH/universal/uni-dep-tb/universal_treebanks_v2.0/std/en/en-univiersal-dev.conll",
+        "es": "UNI_DEP_V2_PATH/universal/uni-dep-tb/universal_treebanks_v2.0/std/es/es-universal-dev.conll",
+        "fr": "UNI_DEP_V2_PATH/universal/uni-dep-tb/universal_treebanks_v2.0/std/fr/fr-universal-dev.conll",
+        "it": "UNI_DEP_V2_PATH/universal/uni-dep-tb/universal_treebanks_v2.0/std/it/it-universal-dev.conll",
+        "pt": "UNI_DEP_V2_PATH/universal/uni-dep-tb/universal_treebanks_v2.0/std/pt-br/pt-br-universal-dev.conll",
+        "sv": "UNI_DEP_V2_PATH/universal/uni-dep-tb/universal_treebanks_v2.0/std/sv/sv-universal-dev.conll"
+    },
+    "test_data_path": {
```
matt-gardner: Oh, man, that's annoying - these are all in the same directory, so the strategy I recommended above doesn't work for the test data (for train and validation, you can just give the paths you want in the parameters for the dataset reader).  It'd be easier if these were laid out as `std/train/de/`, `std/dev/de/`, `/std/test/de/`, etc.

training_config/multilang_dependency_parser.jsonnet
```diff
@@ -0,0 +1,142 @@
...
+        "sv": "UNI_DEP_V2_PATH/universal/uni-dep-tb/universal_treebanks_v2.0/std/sv/sv-universal-test.conll"
+    },
+    "trainer": {
+        "cuda_device": 0,
+        "num_epochs": 40,
+        "optimizer": "adam",
+        "patience": 10,
+        "validation_metric": "+LAS_AVG"
+    },
+    "evaluate_on_test": true,
```
matt-gardner: This should not be set to true in a config file checked into the repo (we don't want to encourage people to always evaluate on test data, as that's not good scientific practice).

allennlp/data/dataset_readers/universal_dependencies_multilang.py
```diff
@@ -0,0 +1,171 @@
...
+    Reads consecutive cases from each input files by the given batch_size.
+
+    Notice: when using the alternate option, one should also use the ``instances_per_epoch``
+    option for the iterator. Otherwise, each epoch will loop infinitely.
```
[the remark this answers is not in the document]
TalSchuster: Each file can have a different amount of data (which is usually what happens in practice...). That's why it is better to just continuously alternate between the files to make sure that the training is balanced across sources. (Otherwise, one could just concatenate all files into one long file with some language identifier per sample).

allennlp/modules/text_field_embedders/basic_text_field_embedder.py
Jbar-ry: Will the desired behaviour of this PR be that it will also be token embedder agnostic? At the moment, I believe the only token embedder that was changed was the addition of a new `elmo_token_embedder_multilang.py` file.

If one would like to use the multilingual biaffine parser but without ELMo, e.g. for parsing related languages where we may not have trained ELMo models, are there many changes that would need to be made to other token embedders? From my understanding, in such a case we would also want to make sure the ["embedding" token embedder](https://github.com/TalSchuster/allennlp-MultiLang/blob/7b3c4aa1208218ad71f733767b8150531e99720e/allennlp/modules/token_embedders/embedding.py#L29) would also handle this extra argument.

Thanks for the great work in the PR!

matt-gardner: Yeah, it's not desirable (or even really meaningful) to change all of the other token embedders to handle this argument.  This is why the existing PR needs to be fixed, as it will currently break if you try to use anything other than the multilingual ELMo embedder.  Either one of the two options above would fix this, I believe.

allennlp/data/dataset_readers/universal_dependencies_multilang.py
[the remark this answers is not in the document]
matt-gardner: Ah, I see, that's a good point.  That's a great thing to include in the docstring.

allennlp/data/dataset_readers/universal_dependencies_multilang.py
```diff
@@ -0,0 +1,171 @@
...
+                tags = [x["deprel"] for x in annotation]
+                words = [x["form"] for x in annotation]
+                if self.use_language_specific_pos:
+                    pos_tags = [x["xpostag"] for x in annotation]
+                else:
+                    pos_tags = [x["upostag"] for x in annotation]
+                yield self.text_to_instance(lang, words, pos_tags, list(zip(tags, heads)))
+
+    @overrides
+    def _read(self, file_paths: Dict[str, str]):
```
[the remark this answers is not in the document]
matt-gardner: I see you just did a fix for this, and your choice looks good to me.  Yesterday I found another example of how I handled this: https://github.com/matt-gardner/allennlp/blob/77b066516b8b12de8521dc8a431741555937917f/allennlp/data/dataset_readers/visual_question_answering/shapes_reader.py#L21-L26.  Not saying you should change what you did, just another example in case you find it useful.  I think that what you did is probably better, though, as my version still has problems at test time.

allennlp/modules/text_field_embedders/basic_text_field_embedder.py
TalSchuster: I went with the first suggestion since, if I understand correctly, the second option requires an identifier per token, which is a bit of overhead since we get sentences from the same language

TalSchuster:
Thanks for the comments. I addressed them and I think that it looks a bit cleaner now.
Only thing I wasn't sure about is how the configuration tester is supposed to deal with the extvars. Can you help there?

allennlp/models/biaffine_dependency_parser_multilang.py
TalSchuster: done 👍

allennlp/tests/common/params_test.py
```diff
@@ -209,6 +209,9 @@ def test_known_configs(self):
             # dependency parser
             'PTB_DEPENDENCIES_TRAIN', 'PTB_DEPENDENCIES_VAL',
 
+            # multilingual dependency parser
+            'TRAIN_PATHNAME', 'DEV_PATHNAME', 'TEST_PATHNAME'
```
matt-gardner: You're missing a comma at the end of this, that's probably why the test is failing.

allennlp/modules/text_field_embedders/basic_text_field_embedder.py
```diff
@@ -96,6 +98,9 @@ def forward(self, text_field_input: Dict[str, torch.Tensor], num_wrapping_dims:
             # Note: need to use getattr here so that the pytorch voodoo
             # with submodules works with multiple GPUs.
             embedder = getattr(self, 'token_embedder_{}'.format(key))
+            identifiers = {}
+            if isinstance(embedder, MultilangTokenEmbedder):
```
matt-gardner: We really need to use `inspect` here, instead of hard-coding a particular embedder.  It's not worth the complexity to support one-off embedders in the core components of the main library like this, unless we can find a way to do it generally.  If you don't know how to do this, I can give some pointers.

allennlp/models/biaffine_dependency_parser_multilang.py
```diff
@@ -0,0 +1,238 @@
...
+            head_indices: torch.LongTensor = None) -> Dict[str, torch.Tensor]:
+        # pylint: disable=arguments-differ
+        """
+        Embedding each language by the corresponding parameters for
+        ``TextFieldEmbedder``. Batches should contain only samples from a
+        single language.
+
+        Parameters
+        ----------
+        words : Dict[str, torch.LongTensor], required
```
matt-gardner: If these are just duplicating what's in the base class, I'd recommend trimming this down and pointing to the base class docstring.  No need to repeat what's there.

allennlp/models/biaffine_dependency_parser_multilang.py
```diff
@@ -0,0 +1,238 @@
...
+            The predicted head indices for each word. A tensor
+            of shape (batch_size, sequence_length).
+        head_types : ``torch.FloatTensor``
+            The predicted head types for each arc. A tensor
+            of shape (batch_size, sequence_length).
+        mask : ``torch.LongTensor``
+            A mask denoting the padded elements in the batch.
+        """
+        if 'lang' not in metadata[0]:
+            raise ConfigurationError("metadata is missing 'lang' key.\
```
matt-gardner: Prefer using string continuations instead of backslashes like this.  That is:

```python
raise ConfigurationError("this is a long string with a quote at the end of the first line; "
                         "when I start the next line with a quote, adjacent strings get concatenated")

File: training_config/multilang_dependency_parser.jsonnet [hunk 19]
```diff
@@ -0,0 +1,120 @@
...
+            "embedding_dim": 50,
+            "vocab_namespace": "pos"
+        },
+        "tag_representation_dim": 100,
+        "text_field_embedder": {
+            "token_embedders": {
+                "elmo": {
+                    "type": "elmo_token_embedder_multilang",
+                    "aligning_files": {
+                        "en": "https://www.dropbox.com/s/nufj4pxxgv5838r/en_best_mapping.pth?dl=1",
```
matt-gardner: Do you mind if we host these on S3, and change the paths to point there?  If this is going to be in the library, it'd be best we if can control the dependencies here.  It'd also be great if you have a trained model that we can host (and add to our demo, but not in this PR).

TalSchuster: Sure thing! please do.
I have a trained model that is based on a slightly older version of the repo. After we merge this PR I can quickly train a model and share.

matt-gardner: Great, I will copy these over some time today or tomorrow, and let you know what paths to change these to.

allennlp/data/iterators/same_lang_iterator.py
```diff
@@ -0,0 +1,46 @@
...
+def split_by_lang(instance_list):
+    insts_by_lang = defaultdict(lambda: [])
+    for inst in instance_list:
+        inst_lang = inst.fields['metadata'].metadata['lang']
+        insts_by_lang[inst_lang].append(inst)
+
+    return iter(insts_by_lang.values())
+
+@DataIterator.register("same_lang")
+class SameLangIterator(DataIterator):
```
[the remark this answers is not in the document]
matt-gardner: Also, I tried hard not to care about `lang` versus `language` in your data and model code, because it should be pretty obvious based on context what the variable names mean there.  But if I'm just looking at the data iterators, I don't have the context of "multi-lingual" at all, so calling this `SameLanguageIterator` would be better here (this applies to the filename also).

allennlp/data/iterators/same_lang_iterator.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+
+    Splits batches into batches containing the same language.
```
matt-gardner: You should say here how the code expects to figure out which instances are the same language (by looking for a `lang` key in a `metadata` field), and say in the docstring that this is intended for multi-lingual applications.

training_config/multilang_dependency_parser.jsonnet
```diff
@@ -0,0 +1,120 @@
...
+                        "it": "https://www.dropbox.com/s/owfou7coi04dyxf/it_weights.hdf5?dl=1",
+                        "pt": "https://www.dropbox.com/s/ul82jsal1khfw5b/pt_weights.hdf5?dl=1",
+                        "sv": "https://www.dropbox.com/s/boptz21zrs4h3nw/sv_weights.hdf5?dl=1",
+                        "de": "https://www.dropbox.com/s/2kbjnvb12htgqk8/de_weights.hdf5?dl=1"
+                    }
+                }
+            }
+        }
+    },
+    "train_data_path": std.extVar("TRAIN_PATHNAME"),
```
matt-gardner: Is the data freely distributable?  If it is, can we put it on S3 also?

training_config/multilang_dependency_parser.jsonnet
matt-gardner: Ok, this is done.  All of these files should be available at `https://s3-us-west-2.amazonaws.com/allennlp/models/multilingual_elmo/`.

training_config/multilang_dependency_parser.jsonnet
TalSchuster: The data for all languages except for English is freely available here - https://github.com/ryanmcd/uni-dep-tb.
The English trees are based on PTB, so like the biaffine parser, we can't provide it. But we can provide a pre-trained model and maybe give in the default configuration file all other languages with a comment that it won't reproduce the results from the paper.
Is there currently a way to put a url of a tarball in the configuration that will be extracted in a way that will allow us to keep the multiple training files logic?

training_config/multilang_dependency_parser.jsonnet
TalSchuster: Thanks! I might get to this and to implement the tests only in a few days.

training_config/multilang_dependency_parser.jsonnet
matt-gardner: Hmm, I don't think there's a way of giving such a tarball.  Maybe we should just leave a comment here describing how to get the data and set it up correctly (or point to the dataset reader docstring, if that's the right place to have this info).

Jbar-ry:
Thanks for the great work over the last couple of weeks in bringing this forward!

In relation to the case of using this parser with the default token embedder `embedding`; I was able to use the multilingual parser for multiple treebank training on the English UD treebanks of UD v2.3 without ELMo. [Here is the config file I used](https://github.com/Jbar-ry/allennlp/blob/multilingual_parser/training_config/en_multilang_dependency_parser.jsonnet).

Even though we will need the results from the tests which Matt mentioned, it was encouraging to be able to run the parser and it seems there were some favorable results from multi-treebank training on English treebanks: I compared the default parser with the multilingual one (though I realized I used a smaller LSTM `hidden_size` of `200` with the default parser vs. `400` in Tal's experiments).

In any case, the results of this comparison are here if you wish to view them: https://drive.google.com/drive/u/2/folders/1NsYJ0jhM79NpG8CWNPMHxHodymsL08Sl

matt-gardner:
Thanks for the info, @Jbar-ry! Just to be clear, the point of the tests isn't just to be sure that the code works right now - it's pretty likely that it works in its current state, otherwise it wouldn't be submitted. Instead, it's to (1) ensure that as the library evolves, we don't break this code on accident, and (2) give example expected usage of this code, in the tests.  When we accept code into the library, we are implicitly accepting maintenance responsibility of that code forever, so we want to be sure that it's in a state that we can maintain it, with tests to make sure we don't break it.

TalSchuster:
Thanks @Jbar-ry ! I'm really glad to see it can improve results in more settings. Thank you for sharing the results!
@matt-gardner, this is totally fine. Sorry that it is delayed... @oriram and I are working on the tests but we are a bit busy right now with some other commitments. I hope that we will get to finish it this week.

matt-gardner:
No worries on the time; good luck on submissions, if that's what you're working on.

matt-gardner:
@TalSchuster, is this ready to look at again?

TalSchuster:
@matt-gardner, I think that the only missing test is the second one from the last point that you wrote (verify that the input is routed to the right encoder by the language). 
I think that I have an idea of how to do it so I'll try to get to it next week.
All of the rest of the code should be independent of that test so if you can go over it, it would be great

maxdel:
Could the pieces from #2369 be used here to avoid code duplication? @TalSchuster @matt-gardner

TalSchuster:
ok, I've added the last test for the embedder.
@matt-gardner, it is ready on my end.
@maxdel, we've already discussed the read of multiple files earlier in this thread. There are some subtle things here, like going alternating between the readers but going over all of the dataset for vocabulary creation. I didn't read the multitask code in details but I believe that there might be things that could be organized. Unfortunately, I don't think that I will have enough time to do it for this PR.

allennlp/tests/data/dataset_readers/universal_dependencies_multilingual_dataset_reader_test.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def test_read_from_files_first_pass_false(self):
+        '''
+        Note: assumes that each data file contains two trees
+        '''
+        reader = UniversalDependenciesMultiLangDatasetReader(languages=['es', 'fr', 'it'],
+                                                             is_first_pass_for_vocab=False,
+                                                             instances_per_file=1,
+                                                             lazy=True)
+        coun_es, coun_fr, coun_it = 0, 0, 0
```
matt-gardner: Are these supposed to be `count_es`, etc.?  If so, just use the full name here.  Same for `inst1` above - just use `instance1`, `instance2`, etc.

allennlp/tests/data/dataset_readers/universal_dependencies_multilingual_dataset_reader_test.py
```diff
@@ -0,0 +1,112 @@
...
+
+        processed_langs = []
+
+        processed_langs.append(self.check_two_instances(instances[0], instances[1]))
+        processed_langs.append(self.check_two_instances(instances[2], instances[3]))
+        processed_langs.append(self.check_two_instances(instances[4], instances[5]))
+
+        assert 'es' in processed_langs and 'fr' in processed_langs and 'it' in processed_langs
+
+    def test_read_from_files_first_pass_false(self):
```
matt-gardner: This name describes the condition that you are testing; it's better to give a name that describes a behavior that you want to see.  Something like `test_iterate_forever_when_first_pass_for_vocab_is_false`.  It's a long name, sure, but it makes it very clear what you are testing and why.

allennlp/tests/data/dataset_readers/universal_dependencies_multilingual_dataset_reader_test.py
```diff
@@ -0,0 +1,112 @@
...
+        '''
+        Note: assumes that each data file contains two trees
```
matt-gardner: You might consider increasing the count threshold to 20 or something, just to make the test more robust to any future changes.  It also makes it more obvious that it's going to iterate forever.

allennlp/tests/data/iterators/same_language_iterator_test.py
```diff
@@ -0,0 +1,23 @@
+# pylint: disable=no-self-use,invalid-name
+from allennlp.data.dataset_readers import UniversalDependenciesMultiLangDatasetReader
+from allennlp.data.iterators.same_language_iterator import SameLanguageIterator
+from allennlp.common.testing import AllenNlpTestCase
+
+class SameLanguageIteratorTest(AllenNlpTestCase):
+    data_path = AllenNlpTestCase.FIXTURES_ROOT / "data" / "dependencies_multilang" / "*"
+
+    def test_assert_instances_of_different_languages_are_in_different_batches(self):
```
matt-gardner: `s/test_assert_/test_/`

allennlp/tests/fixtures/biaffine_dependency_parser_multilang/experiment.json
```diff
@@ -0,0 +1,98 @@
...
+        "instances_per_epoch": 32
+    },
+    "model": {
+        "type": "biaffine_parser_multilang",
+        "arc_representation_dim": 10,
+        "dropout": 0.33,
+        "encoder": {
+            "type": "lstm",
+            "bidirectional": true,
+            "dropout": 0.33,
```
matt-gardner: Use smaller dimensions (all <10) and no dropout for test fixtures like this.  If you've already got artifacts created that depend on these dimensions, fine, as long as the test is very fast to run.

allennlp/tests/fixtures/elmo_multilingual/config/multilang_token_embedder.json
```diff
@@ -0,0 +1,93 @@
...
+        "instances_per_epoch": 32
+    },
+    "model": {
+        "type": "biaffine_parser_multilang",
+        "arc_representation_dim": 10,
+        "dropout": 0.33,
+        "encoder": {
+            "type": "lstm",
+            "bidirectional": true,
+            "dropout": 0.33,
```
matt-gardner: Same here; no dropout, smaller dimensions.

allennlp/tests/data/dataset_readers/universal_dependencies_multilingual_dataset_reader_test.py
TalSchuster: I kept the mini ELMo to 32 but reduced all other dimensions. The test is pretty fast (few seconds)

PR #2626, opened 2019-03-18
Bump up spacy version pin to 2.1

sai-prasanna:
Spacy 2.1 released, so relaxing the pin. Tests passed locally.

PR #2631, opened 2019-03-19
Revert "Bump up spacy version pin to 2.1"

matt-gardner:
Reverts allenai/allennlp#2626

Unfortunately, there is a bug in spacy 2.1.0 (https://github.com/explosion/spaCy/issues/3435) that makes this not work with how we install models.  We didn't see this in our PR CI because we explicitly install models beforehand in `Dockerfile`, but not in `Dockerfile.pip`, which is used to build and test master commits.

We need to revert this until that bug is fixed.


matt-gardner:
Note that you can install spacy 2.1 yourself after installing allennlp, and things will work just fine, if you also install the models that you need.  We just can't merge something that breaks our build, and this does, because the models aren't getting installed correctly.

matt-gardner:
Actually, Ines just gave me a workaround that looks like it'll work; I think I'll close this and do that instead.

PR #2632, opened 2019-03-19
Add workaround for missing linking in spacy 2.1, remove install_requirements.sh

matt-gardner:
This PR adds a workaround for the missing linking in spacy 2.1.0, as seen in https://github.com/explosion/spaCy/issues/3435.  It also removes `install_requirements.sh`, which was the reason that we did not catch this issue until after the PR bumping spacy to 2.1 was merged.


Hyperparticle:
Should also update the README, as it currently references `install_requirements.sh`

matt-gardner:
Thanks @Hyperparticle!

PR #2633, opened 2019-03-20
Remove reference to install_requirements.sh from the README

matt-gardner:


README.md
```diff
@@ -126,7 +126,7 @@ You can also install AllenNLP by cloning our git repository:
 Create a Python 3.6 virtual environment, and install the necessary requirements by running:
 
   ```bash
-  scripts/install_requirements.sh
+  pip install -r requirements.txt
```
matt-gardner: I'm actually not sure that we need this anymore, because `pip install --editable .` might just work.  But I don't know enough about this set up to be the right person to make that call.  @schmmd, any ideas?

schmmd: This was a user contribution and I don't fully understand it--but it looks like you're right.

```
root@4ca47cb5bdf7:/allennlp# pip install --editable .
Obtaining file:///allennlp
Collecting torch>=0.4.1 (from allennlp===0.8.3-unreleased)
  Downloading https://files.pythonhosted.org/packages/31/ca/dd2c64f8ab5e7985c4af6e62da933849293906edcdb70dac679c93477733/torch-1.0.1.post2-cp36-cp36m-manylinux1_x86_64.whl (582.5MB)
...
```

schmmd:
I confirmed we don't need it.  By the way, I test this kind of stuff with `docker run --entrypoint=/bin/bash -it python:3.6`.  Then I just clone the AllenNLP git repo and run `pip install --editable .` followed by `allennlp test-install`.

matt-gardner:
Thanks for the tip!

PR #2635, opened 2019-03-21
add option "-k" to limit tests in test-install command

mfa:
Adds possiblity to limit pytest tests in test-install command.

### DIFF
--- a/allennlp/commands/test_install.py
+++ b/allennlp/commands/test_install.py
@@ -60,11 +60,9 @@ def _run_test(args: argparse.Namespace):
     test_dir = os.path.join(module_parent, "allennlp")
     logger.info("Running tests at %s", test_dir)
 
-    # set for "k" and for "not run-all"
-    pytest_m = ['-m', 'not java']
-
     if args.k:
         pytest_k = ['-k', args.k]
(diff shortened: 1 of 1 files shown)


allennlp/commands/test_install.py
```diff
@@ -57,12 +59,24 @@ def _run_test(args: argparse.Namespace):
...
+    if args.k:
+        pytest_k = ['-k', args.k]
+        if args.run_all:
+            logger.warning("the argument '-k' overwrites '--run-all'.")
     else:
-        exit_code = pytest.main([test_dir, '--color=no', '-k', 'not sniff_test and not notebooks_test',
-                                 '-m', 'not java'])
+        if args.run_all:
+            # TODO(nfliu): remove this when notebooks have been rewritten as markdown.
+            pytest_k = ['-k', 'not notebooks_test']
```
matt-gardner: The notebook tests have been removed, so this can just be empty.  I think you can simplify this a bit, to:

```python
if args.k
    # what you have there
elif args.run_all:
    pytest_k = []
    pytest_m = []
else:
    pytest_k = ['-k', 'not sniff_test']
```

PR #2637, opened 2019-03-21
Moving WTQ language updates from iterative-search-semparse

pdasigi:


pdasigi:
@matt-gardner This PR is now ready to review. It includes only the updates made to the context representation and the language in `iterative-search-semparse` (previously called `allennlp-weak-supervision-research`). I was going to move the model code as well, but I decided to make a separate PR for that as this PR is quite big already (sorry!).

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -76,31 +77,116 @@
         }
 
 
+class Date:
```
matt-gardner: Can you put this in a separate file?  Other projects want to use this also.  Also, this feels like a `DomainLanguage` construct, which belongs under `domain_languages/`, not here (to avoid circular imports in the future).  So, `domain_languages/date.py` or `domain_languages/common/date.py`.

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -76,31 +77,116 @@
...
+                 table_data: List[Dict[str, CellValueType]],
+                 column_types: Dict[str, Set[str]],
+                 column_names: Set[str],
                  question_tokens: List[Token]) -> None:
         self.table_data = table_data
-        self.column_types = column_types
+        self.column_types: Set[str] = set()
+        self.column_names = column_names
+        for types in column_types.values():
+            self.column_types.update(types)
```
matt-gardner: This is very confusing.  Why do you have `column_types` and `self.column_types` holding different information?  Maybe this would have been cleared up with better documentation on the parameters of this class, but using different names for these two things would be better, too.

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -115,25 +201,24 @@ def get_table_knowledge_graph(self) -> KnowledgeGraph:
...
-                if column_type == "number":
-                    column_name = f"number_column:{column}"
-                    number_columns.append(column_name)
-                elif column_type == "date":
-                    column_name = f"date_column:{column}"
-                    date_columns.append(column_name)
-                else:
-                    column_name = f"string_column:{column}"
+            for typed_column_name in self.table_data[0].keys():
+                if "number_column:" in typed_column_name or "num2_column" in typed_column_name:
```
matt-gardner: Do we actually use this `num2_column` anywhere?

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -115,25 +201,24 @@ def get_table_knowledge_graph(self) -> KnowledgeGraph:
...
-            for column, column_type in self.column_types.items():
-                if column_type == "number":
-                    column_name = f"number_column:{column}"
-                    number_columns.append(column_name)
-                elif column_type == "date":
-                    column_name = f"date_column:{column}"
-                    date_columns.append(column_name)
-                else:
-                    column_name = f"string_column:{column}"
+            for typed_column_name in self.table_data[0].keys():
```
matt-gardner: Don't you have a `column_names` field that you can use instead of `self.table_data[0]`?

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -115,25 +201,24 @@ def get_table_knowledge_graph(self) -> KnowledgeGraph:
...
+                if "date_column:" in typed_column_name:
+                    date_columns.append(typed_column_name)
+
                 # Add column names to entities, with no neighbors yet.
-                entities.add(column_name)
-                neighbors[column_name] = []
-                entity_text[column_name] = column.replace("_", " ")
+                entities.add(typed_column_name)
+                neighbors[typed_column_name] = []
+                entity_text[typed_column_name] = typed_column_name.split(":")[-1].replace("_", " ")
```
matt-gardner: It's probably worth doing `typed_column_name.split(":", 1)[-1]`, so that you don't accidentally truncate the text.

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -232,21 +336,21 @@ def get_entities_from_question(self) -> Tuple[List[Tuple[str, str]], List[Tuple[
             normalized_token_text = self.normalize_string(token_text)
             if not normalized_token_text:
                 continue
-            token_column = self._string_in_table(normalized_token_text)
-            if token_column:
-                token_type = token_column.split(":")[0].replace("_column", "")
+            token_columns = self._string_in_table(normalized_token_text)
+            if token_columns:
+                token_type = token_columns[0].split(":")[0].replace("_column", "")
```
matt-gardner: You had a comment explaining the `[0]` in `_string_in_table`, but you removed it when moving this logic here.  I'd  move the comment here too.

allennlp/semparse/domain_languages/wikitables_language.py
```diff
@@ -3,75 +3,23 @@
...
-from typing import Dict, List, NamedTuple, Set, Tuple
+from typing import Dict, List, NamedTuple, Set, Tuple, Any
 import logging
 import re
 
 from allennlp.semparse.domain_languages.domain_language import (DomainLanguage, ExecutionError,
                                                                 PredicateType, predicate)
 from allennlp.semparse.contexts.table_question_knowledge_graph import MONTH_NUMBERS
 from allennlp.semparse.contexts import TableQuestionContext
+from allennlp.semparse.contexts.table_question_context import Date, CellValueType
```
matt-gardner: I'm not sure how you're getting away with this.  It looks like a circular import to me.  I guess because the other import goes directly to `domain_language.py`?  Either way, this should be fixed.  `CellValueType` is easy enough to just redefine here, instead of having the import.

Well, actually, now that I look at this a bit more, maybe it's right for the imports to go this way, instead of the other way, as I thought...  Still, the `Date` class and its tests should be somewhere more general.  Any suggestions?

allennlp/semparse/domain_languages/wikitables_language.py
```diff
@@ -188,49 +141,99 @@ def __init__(self, table_context: TableQuestionContext) -> None:
...
+        """
+        Returns an agenda that can be used guide search.
+
+        Parameters
+        ----------
+        conservative : ``bool``
+            Setting this flag will return a subset of the agenda items that correspond to high
+            confidence lexical matches. You'll need this if you are going to use this agenda to
+            penalize a model for producing logical forms that do not contain some items in it. In
+            that case, you;ll want this agenda to have close to perfect precision, at the cost of a
```
matt-gardner: `s/you;ll/you'll/`

allennlp/semparse/domain_languages/wikitables_language.py
```diff
@@ -310,12 +350,42 @@ def all_rows(self) -> List[Row]:
         return self.table_data
 
     @predicate
-    def select(self, rows: List[Row], column: Column) -> List[str]:
+    def select_string(self, rows: List[Row], column: Column) -> List[str]:
```
matt-gardner: Shouldn't this take a `StringColumn` now?  Then you won't need the `assert`, also.

allennlp/semparse/domain_languages/wikitables_language.py
```diff
@@ -310,12 +350,42 @@ def all_rows(self) -> List[Row]:
...
+        assert column.name.startswith("string_column:")
+        return [str(row.values[column.name]) for row in rows if row.values[column.name] is not None]
+
+    @predicate
+    def select_number(self, rows: List[Row], column: NumberColumn) -> Number:
+        """
+        Select function takes a row (as a list) and a column name and returns the number in that
+        column. If multiple rows are given, will return the first number that is not None.
+        """
+        assert column.name.startswith("number_column:") or column.name.startswith("num2_column")
```
matt-gardner: I'm not sure these asserts add much.

allennlp/semparse/domain_languages/wikitables_language.py
```diff
@@ -462,123 +541,232 @@ def argmin(self, rows: List[Row], column: ComparableColumn) -> List[Row]:
     # added to the language if we see a number column in the table.
 
     def filter_number_greater(self, rows: List[Row], column: NumberColumn, filter_value: Number) -> List[Row]:
-        cell_row_pairs = self._get_number_row_pairs_to_filter(rows, column.name)
+        if not isinstance(filter_value, Number):
```
matt-gardner: Is this necessary?

allennlp/semparse/domain_languages/wikitables_language.py
```diff
@@ -462,123 +541,232 @@ def argmin(self, rows: List[Row], column: ComparableColumn) -> List[Row]:
...
 
     # These two are similar to the filter methods above, but operate on strings obtained from the
     # question, instead of dates or numbers.  So they check for whether the string value is present
     # in the cell or not, instead of using a numerical / date comparator.  We only add them to the
     # language if we see a string column in the table (which is basically always).
 
-    def filter_in(self, rows: List[Row], column: StringColumn, filter_value: str) -> List[Row]:
+    def filter_in(self, rows: List[Row], column: StringColumn, filter_values: List[str]) -> List[Row]:
+        # We accept a list of filter values instead of a single string to allow the outputs of select like
+        # operations to be passed in as filter values.
```
matt-gardner: Note here that we're using the first one if we get a list.

allennlp/semparse/domain_languages/wikitables_language.py
```diff
@@ -462,123 +541,232 @@ def argmin(self, rows: List[Row], column: ComparableColumn) -> List[Row]:
...
+    # These are some more date-column-specific functions, which only get added if we see a number
+    # column.
+    def max_date(self, rows: List[Row], column: DateColumn) -> Date:
+        """
+        Takes a list of rows and a column and returns the max of the values under that column in
+        those rows.
+        """
+        cell_values = [row.values[column.name] for row in rows]
+        if not cell_values:
+            return 0.0  # type: ignore
```
matt-gardner: Don't you want this to be a `Date`?

allennlp/semparse/domain_languages/wikitables_language.py
```diff
@@ -462,123 +541,232 @@ def argmin(self, rows: List[Row], column: ComparableColumn) -> List[Row]:
...
         Takes a list of rows and a column and returns the max of the values under that column in
         those rows.
         """
-        cell_row_pairs = self._get_number_row_pairs_to_filter(rows, column.name)
-        if not cell_row_pairs:
+        cell_values = [row.values[column.name] for row in rows]
+        if not cell_values:
             return 0.0  # type: ignore
-        return max([value for value, _ in cell_row_pairs])  # type: ignore
+        if not all([isinstance(value, Number) for value in cell_values]):
```
matt-gardner: You have this error checking all over the place, except you're basically just checking that you get the types that you annotated in the function.  I don't think these help anything, and just add a bunch of noise to the code.  These checks belong as tests, not as runtime asserts (and we already have tests for things that would cause errors like this).  Unless I'm missing something, and you might actually run into this at runtime for some reason without buggy code?

allennlp/semparse/domain_languages/wikitables_language.py
```diff
@@ -462,123 +541,232 @@ def argmin(self, rows: List[Row], column: ComparableColumn) -> List[Row]:
...
         """
         Takes a list of rows and a column and returns the sum of the values under that column in
         those rows.
         """
-        cell_row_pairs = self._get_number_row_pairs_to_filter(rows, column.name)
-        if not cell_row_pairs:
+        cell_values = [row.values[column.name] for row in rows]
+        if not cell_values:
             return 0.0  # type: ignore
-        return sum([value for value, _ in cell_row_pairs])  # type: ignore
```
matt-gardner: You're also inconsistent with those checks - if they are worth having above, why not here?  I'd just remove them everywhere.

allennlp/semparse/domain_languages/wikitables_language.py
```diff
@@ -462,123 +541,232 @@ def argmin(self, rows: List[Row], column: ComparableColumn) -> List[Row]:
...
+        return sum(cell_values)  # type: ignore
 
     def average(self, rows: List[Row], column: NumberColumn) -> Number:
         """
         Takes a list of rows and a column and returns the mean of the values under that column in
         those rows.
         """
-        cell_row_pairs = self._get_number_row_pairs_to_filter(rows, column.name)
-        if not cell_row_pairs:
+        cell_values = [row.values[column.name] for row in rows]
```
matt-gardner: I do very much like this simplification, where the data has already been formatted to be the right type, so you don't have to check and reformat it all the time.  Should make this faster, too.

allennlp/semparse/contexts/table_question_context.py
pdasigi: Yes, we do. For columns that have number ranges like [this one](https://ppasupat.github.io/WikiTableQuestions/viewer/#203-754) (the "Career" column), we can extract the second number using `num2_column`.

allennlp/semparse/contexts/table_question_context.py
pdasigi: Yes, thanks!

allennlp/semparse/contexts/table_question_context.py
pdasigi: Thanks!

allennlp/semparse/domain_languages/wikitables_language.py
pdasigi: I like your suggestion to move `Date` into `allennlp.semparse.domain_languages.common.date`. I'm doing that.

allennlp/semparse/domain_languages/wikitables_language.py
pdasigi: Yes, thanks!

allennlp/semparse/domain_languages/wikitables_language.py
pdasigi: Agreed! Removing them.

allennlp/semparse/domain_languages/wikitables_language.py
pdasigi: Yes, I do. Thanks!

allennlp/semparse/domain_languages/wikitables_language.py
pdasigi: I don't remember why I had these assertions. I agree with you, and removing them doesn't seem to be breaking anything. Thanks!

PR #2641, opened 2019-03-25
Update outdated doc

maxdel:


bryant1410:
> Thanks! It'd be nice to have a CI check for this kind of thing, but I'm not sure there's a good way to do it.

@matt-gardner IntelliJ provides inspections for this, which [seem to be specific for it](https://github.com/JetBrains/intellij-community/blob/master/python/src/com/jetbrains/python/inspections/PyIncorrectDocstringInspection.java#L41). Given that AllenNLP's CI is TeamCity, why not [running inspections with it](https://confluence.jetbrains.com/display/TCD18/Inspections)? [It seems to be a bit problematic sometimes with Python because of missing stubs](https://confluence.jetbrains.com/display/TCD18/Inspections), but it doesn't seem to be a problem when checking for docstrings only.

I guess it's possible from the project configuration in TeamCity or by [calling the CLI for this](https://www.jetbrains.com/help/idea/running-inspections-offline.html). An inspection profile file can be created so just docstring checks are run. It generates HTML reports, so anyone can see them.

The flip side would be that people won't easily be able to run these inspections on their side before pushing code. They'd need PyCharm. Maybe it could be enabled along with some warning in `scripts/verify.py`. Like an env var to have the `inspect.sh` location (or PyCharm location), if not found then disabling that inspection and showing a warning in the output.

Thoughts?

bryant1410:
To check the absence of docstring warnings, I guess it'd be using the exit status code, or else the output file `PyIncorrectDocstringInspection.xml` can be inspected (the XML tag `problems` shouldn't have children).

matt-gardner:
Thanks @bryant1410, I added a note referencing your ideas to the issue that we have open for this.  Do you know what errors this shows on our existing repo?  I'd be hesitant to add such a big dependency, but if it's only added as a CI check and not in our pip install, maybe it wouldn't be too bad if it does a really good job on the functionality we want.

bryant1410:
Yes. The inspections make a lot of sense (are about missing parameters in docstrings), they are 35 and I found no false positive so far. I also run an inspection when the docstrings are not formatted with triple double quotes (24). I exported it as HTML as put it in a [ZIP file](https://github.com/allenai/allennlp/files/3014691/exported.zip) so you people can see it.

Yeah, I have the same mixed feelings as you.

bryant1410:
Do you guys have a TeamCity config file in git or somewhere to try it out locally?

matt-gardner:
I looked at that report, and it looks good to me.  It'd be really great to get this added as a CI check.  For the TeamCity config, I can see it [here](http://build.allennlp.org/viewType.html?buildTypeId=AllenNLP_AllenNLPPullRequests&tab=buildTypeSettings), though I'm not sure what permissions are required to see that page.  In general, though, if you can get something that runs just from `scripts/check_docstrings.py`, or something, using dependencies that are placed in [this part of requirements.txt](https://github.com/allenai/allennlp/blob/263d3401a3986b0bebb51033c55d6132b5d321d2/requirements.txt#L110-L143), that should be enough.

bryant1410:
I cannot see that file as a guest.

matt-gardner:
I guess that's not too surprising, because it has some username and password stuff in it.  But, it just shows the settings that run the CI checks, including all of the various scripts that we run.  If you have a script like I said, we could add it there without issue.

PR #2644, opened 2019-03-25
make things backward compatible with spacy 2.0

joelgrus:
I am not thrilled about this, as the 2.0 code path is not really getting tested anymore (except manually, by me, as part of preparing this PR).

If we're not willing to have our CI test both ways (and I'm pretty sure we're not), then I'm not sure we should support 2.0.

### DIFF
--- a/allennlp/data/tokenizers/sentence_splitter.py
+++ b/allennlp/data/tokenizers/sentence_splitter.py
@@ -1,5 +1,8 @@
 from typing import List
 from overrides import overrides
+
+import spacy
+
 from allennlp.common import Registrable
 from allennlp.common.util import get_spacy_model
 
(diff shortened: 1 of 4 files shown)


allennlp/data/tokenizers/sentence_splitter.py
```diff
@@ -44,8 +47,10 @@ def __init__(self,
         self.spacy = get_spacy_model(language, parse=not rule_based, ner=False, pos_tags=False)
         if rule_based:
             # we use `sentencizer`, a built-in spacy module for rule-based sentence boundary detection.
-            if not self.spacy.has_pipe('sentencizer'):
-                sbd = self.spacy.create_pipe('sentencizer')
+            # depending on the spacy version, it could be called 'sentencizer' or 'sbd'
+            sbd_name = 'sbd' if spacy.__version__ < '2.1' else 'sentencizer'
```
DeNeutoy: If you changed this to a unique name which didn't depend on the version, you wouldn't have to disable the tests that only error because of the loading mismatch, rather than the changed model (the constituency parser test and the text_classification reader test).

PR #2647, opened 2019-03-26
Upgrade Dockerfile to stretch.

schmmd:
@kyleclo found that `jessie` gave up support for their apt package so, when using our Docker image, they needed to add the following as a work-around.  I didn't realize there was a newer stable release of Debian (`stretch`) but now that I do I'm upgrading to it.

```
RUN echo "deb http://deb.debian.org/debian/ jessie main" > /etc/apt/sources.list \
  && echo "deb-src http://deb.debian.org/debian/ jessie main" >> /etc/apt/sources.list \
  && echo "deb http://security.debian.org/ jessie/updates main" >> /etc/apt/sources.list \
  && echo "deb-src http://security.debian.org/ jessie/updates main" >> /etc/apt/sources.list \
  && echo "deb [check-valid-until=no] http://archive.debian.org/debian jessie-backports main" >> /etc/apt/sources.list \
  && echo "deb-src [check-valid-until=no] http://archive.debian.org/debian jessie-backports main" >> /etc/apt/sources.list
```

See https://unix.stackexchange.com/questions/508724/failed-to-fetch-jessie-backports-repository for more details about the deprecation of apt repositories for `jessie`.


Dockerfile
```diff
@@ -30,13 +30,8 @@ RUN apt-get update --fix-missing && apt-get install -y \
     libxrender1 \
     wget \
     libevent-dev \
-    build-essential && \
-    rm -rf /var/lib/apt/lists/*
-
-# Install Java.
-RUN echo "deb http://http.debian.net/debian jessie-backports main" >>/etc/apt/sources.list
-RUN apt-get update
```
schmmd: I think this update was previously necessary because of the `rm -rf /var/lib/apt/lists/*` immediately proceeding it--so this change should make the build a bit faster too.

schmmd:
I tested `Dockerfile.pip` manually by building it locally since it only runs on master commit as I recall.

```
docker build -f Dockerfile.pip -t allennlp-pip .
docker run allennlp-pip test-install
```

Dockerfile
brendan-ai2: Pretty sure updating is just necessary after a new source is added.

Dockerfile
```diff
@@ -30,13 +30,8 @@ RUN apt-get update --fix-missing && apt-get install -y \
     libxrender1 \
     wget \
     libevent-dev \
-    build-essential && \
-    rm -rf /var/lib/apt/lists/*
```
brendan-ai2: This was probably here to keep down the size of the image, right? See https://askubuntu.com/questions/179955/var-lib-apt-lists-is-huge for instance. Should we retain this line?

kyleclo:
thx to @danielking for finding it. 
q:  confirming that i'll need to keep using the workaround until the docker image for allennlp v0.8.3 is uploaded w/ this fix?

brendan-ai2:
That's my understanding. Though if you need to urgently get something working, you could pull down this PR and build the `Dockerfile` directly. (I'd anticipate this being merged and image uploaded tomorrow.)

Dockerfile
schmmd: @brendan-ai2 look earlier in the Dockerfile.  We no longer need to add a new source (since we're on stretch) and we no longer need to update here because we do earlier.  An update would have been required for all installs after `rm -rf /var/lib/apt/lists/*`, so the previous iteration of this Dockerfile was inefficient (it would create the apt lists, nuke them, and then recreate them).

schmmd:
@kyleclo we push image per commit, so you wouldn't need to build your own image.  See https://hub.docker.com/r/allennlp/commit/tags.  But it's preferable to be on a release since we know more about releases than commits.

Dockerfile
schmmd: You can disregard my last comment--we're both right.

Dockerfile
schmmd: Yes we should.  In fact, I didn't mean to remove it.

matt-gardner:
> I tested `Dockerfile.pip` manually by building it locally since it only runs on master commit as I recall.
> 
> ```
> docker build -f Dockerfile.pip -t allennlp-pip .
> docker run allennlp-pip test-install
> ```

Quick sanity check - did you try it with `--run-all`?  You're changing the java installation, but you didn't run the java tests.

schmmd:
@matt-gardner `Dockerfile.pip` didn't previously install the JVM and still doesn't after this PR, so that test isn't necessary.  JVM tests ran as part of continuous build, so the `Dockerfile` used for continuous build ran the JVM tests.

For example, see:

```
[08:13:20][Step 3/10] allennlp/tests/predictors/wikitables_parser_test.py::TestWikiTablesParserPredictor::test_answer_present PASSED [ 74%]
[08:13:23][Step 3/10] allennlp/tests/predictors/wikitables_parser_test.py::TestWikiTablesParserPredictor::test_answer_present_with_batch_predict PASSED [ 74%]
```

Dockerfile
brendan-ai2: Right, reorganizing would have been ideal. At any rate, thanks again for fixing this!

PR #2648, opened 2019-03-26
Fix quarel explanations

matt-gardner:
This was due to some refactoring that I did a while ago, and it is the cause of the broken QuaRel demo.  I didn't see an easy way to test for this kind of breakage in the demo repository, unfortunately, but I added a test for the breakage here.

I'm not going to say that this fixes the issue in the demo repo (https://github.com/allenai/allennlp-demo/issues/182), because the demo repo needs to be updated to use this commit after this gets merged, and then it will be fixed.

### DIFF
--- a/allennlp/tests/semparse/contexts/quarel_utils_test.py
+++ b/allennlp/tests/semparse/contexts/quarel_utils_test.py
@@ -1,0 +1,20 @@
+# pylint: disable=no-self-use,invalid-name
+from allennlp.common.testing import AllenNlpTestCase
+from allennlp.semparse.contexts.knowledge_graph import KnowledgeGraph
+from allennlp.semparse.contexts.quarel_utils import get_explanation
+from allennlp.semparse.worlds.quarel_world import QuarelWorld
+
+
+class TestQuarelUtils(AllenNlpTestCase):
+    def test_get_explanation_provides_non_empty_explanation_for_typical_inputs(self):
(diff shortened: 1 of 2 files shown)


allennlp/tests/semparse/contexts/quarel_utils_test.py
```diff
@@ -0,0 +1,20 @@
...
+        entities = {'a:sugar': 'sugar', 'a:diabetes': 'diabetes'}
+        world_extractions = {'world1': 'bill', 'world2': 'sue'}
+        answer_index = 0
+        knowledge_graph = KnowledgeGraph(entities.keys(), {key: [] for key in entities}, entities)
+        world = QuarelWorld(knowledge_graph, "quarel_v1_attr_entities")
+        explanation = get_explanation(logical_form,
+                                      world_extractions,
+                                      answer_index,
+                                      world)
+        assert len(explanation) == 4
```
brendan-ai2: Worth testing the structure of the explanations too?

matt-gardner: Perhaps, but not in this PR.  There should have been more tests of the quarel code originally - the lack of tests was what made it so this wasn't caught earlier.  But because things aren't changing much in this code, I'd prefer an add-as-needed approach to testing - I don't want to take the time to add exhaustive tests right now.

brendan-ai2: 👍

PR #2653, opened 2019-03-27
TextCat Reader skip_label_indexing Fix

### LINKED ISSUE #2651: TextClassificationJsonReader doesn't allow skip_label_indexing=True
**Describe the bug**
The DataReader subclass TextClassificationJsonReader gives the following `ConfigurationError`
when reading data that has numeric labels (starting from 0):

```
ConfigurationError: 'In order to skip indexing, your labels must be integers. Found label = 0'
```

**To Reproduce**
1. Create a json lines file named 'example.jsonl' that contains two rows:
```
with open("example.jsonl", "w") as f:
    f.write("{\"label\": 0, \"text\": \"This text has label 0\"}\n")
    f.write("{\"label\": 1, \"text\": \"This text has label 1\"}")
```
2. Import and instantiate TextClassificationJsonReader class:
```
from allennlp.data.dataset_readers import TextClassificationJsonReader
tcjr = TextClassificationJsonReader(skip_label_indexing=True)
```
3. Attempt to read data from file.
```
reader = tcjr.read(file_path='example.jsonl')
```
**Expected behavior**
Provide a list of `Instance`s. 

**System**
 - OS: Linux
 - Python version: 3.6.7
 - AllenNLP version: 0.8.2

vidurj:
Fixes #2651.

### DIFF
--- a/allennlp/data/dataset_readers/text_classification_json.py
+++ b/allennlp/data/dataset_readers/text_classification_json.py
@@ -1,4 +1,4 @@
-from typing import Dict, List
+from typing import Dict, List, Union
 import logging
 import json
 from overrides import overrides
@@ -69,7 +69,13 @@ class TextClassificationJsonReader(DatasetReader):
                 text = items["text"]
                 label = items.get("label", None)
                 if label is not None:
-                    label = str(label)
+                    if self._skip_label_indexing:
+                        try:
+                            label = int(label)
+                        except ValueError:
+                            raise Exception('Labels must be integers if skip_label_indexing is True.')
+                    else:
+                        label = str(label)
                 instance = self.text_to_instance(text=text, label=label)
                 if instance is not None:
                     yield instance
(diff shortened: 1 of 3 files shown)


allennlp/data/dataset_readers/text_classification_json.py
```diff
@@ -69,7 +69,13 @@ def _read(self, file_path):
                 text = items["text"]
                 label = items.get("label", None)
                 if label is not None:
-                    label = str(label)
+                    if self._skip_label_indexing:
+                        try:
+                            label = int(label)
+                        except ValueError:
+                            raise Exception('Labels must be integers if skip_label_indexing is True.')
```
kernelmachine: Should this be a `ConfigurationError` instead of a general Exception

vidurj: I made it a `ValueError` since I'm assuming the user intended to skip indexing, and that the input was incorrect.

matt-gardner: I don't have context on what the right error is here, just FYI on what the intent of `ConfigurationError` is: a `ConfigurationError` should be used when the problem is the result of something that was configured in a configuration file.

Issue #1410, opened 2018-06-21
write a CI check to make sure docstrings are in sync with constructors


joelgrus:
here's the raw materials, if I have time I'll put all the pieces together. you'd need to collect all the classes in the library, find the parameters both way, and make sure they're equal. there's probably lots of edge cases.

```python
In [15]: from allennlp.models.crf_tagger import CrfTagger

In [16]: import inspect

In [17]: from numpydoc.docscrape import NumpyDocString

In [18]: docstring_params = NumpyDocString(CrfTagger.__doc__)['Parameters']

In [19]: inspect_params = inspect.getfullargspec(CrfTagger)

In [20]: docstring_params
Out[20]:
[('vocab',
  '``Vocabulary``, required',
  ['A Vocabulary, required in order to compute sizes for input/output projections.']),
 ('text_field_embedder',
  '``TextFieldEmbedder``, required',
  ['Used to embed the tokens ``TextField`` we get as input to the model.']),
 ('encoder',
  '``Seq2SeqEncoder``',
  ['The encoder that we will use in between embedding tokens and predicting output tags.']),
 ('label_namespace',
  '``str``, optional (default=``labels``)',
  ['This is needed to compute the SpanBasedF1Measure metric.',
   'Unless you did something unusual, the default value should be what you want.']),
 ('dropout:  ``float``, optional (detault=``None``)', '', []),
 ('constraint_type',
  '``str``, optional (default=``None``)',
  ['If provided, the CRF will be constrained at decoding time',
   'to produce valid labels based on the specified type (e.g. "BIO", or "BIOUL").']),
 ('include_start_end_transitions',
  '``bool``, optional (default=``True``)',
  ['Whether to include start and end transition parameters in the CRF.']),
 ('initializer',
  '``InitializerApplicator``, optional (default=``InitializerApplicator()``)',
  ['Used to initialize the model parameters.']),
 ('regularizer',
  '``RegularizerApplicator``, optional (default=``None``)',
  ['If provided, will be used to calculate the regularization penalty during training.'])]

In [21]: inspect_params
[...]

PR #2655, opened 2019-03-27
Fix docstrings after inspection

bryant1410:
This PR starts to tackle #1410

Tasks:

* [x] Fix docstrings.
* [ ] ~Add inspection to CI, testing that it works fine.~ (Edit: left for another PR)

There are some docs changes which I'm not 100% sure about it, so pay special attention when reviewing them. Also, I've made two `metadata` arguments required cause I saw they were in fact.

### DIFF
--- a/allennlp/common/params.py
+++ b/allennlp/common/params.py
@@ -344,7 +344,7 @@ class Params(MutableMapping):
         quiet: bool, optional (default = False)
             Whether to log the parameters before returning them as a dict.
         infer_type_and_cast : bool, optional (default = False)
-            If True, we infer types and cast (e.g. things look like floats to floats).
+            If True, we infer types and cast (e.g. things that look like floats to floats).
         """
         if infer_type_and_cast:
             params_as_dict = infer_and_cast(self.params)
--- a/allennlp/common/params.py
+++ b/allennlp/common/params.py
@@ -343,6 +343,8 @@ class Params(MutableMapping):
         ----------
         quiet: bool, optional (default = False)
             Whether to log the parameters before returning them as a dict.
+        infer_type_and_cast : bool, optional (default = False)
+            If True, we infer types and cast (e.g. things look like floats to floats).
         """
         if infer_type_and_cast:
             params_as_dict = infer_and_cast(self.params)
--- a/allennlp/models/coreference_resolution/coref.py
+++ b/allennlp/models/coreference_resolution/coref.py
@@ -111,8 +111,8 @@ class CoreferenceResolver(Model):
     def forward(self,  # type: ignore
                 text: Dict[str, torch.LongTensor],
                 spans: torch.IntTensor,
-                span_labels: torch.IntTensor = None,
-                metadata: List[Dict[str, Any]] = None) -> Dict[str, torch.Tensor]:
+                metadata: List[Dict[str, Any]],
+                span_labels: torch.IntTensor = None) -> Dict[str, torch.Tensor]:
         # pylint: disable=arguments-differ
         """
         Parameters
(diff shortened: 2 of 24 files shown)


allennlp/common/params.py
```diff
@@ -343,6 +343,8 @@ def as_dict(self, quiet: bool = False, infer_type_and_cast: bool = False):
         ----------
         quiet: bool, optional (default = False)
             Whether to log the parameters before returning them as a dict.
+        infer_type_and_cast : bool, optional (default = False)
+            If True, we infer types and cast (e.g. things look like floats to floats).
```
matt-gardner: s/things look like/things that look like/

allennlp/models/coreference_resolution/coref.py
```diff
@@ -111,8 +111,8 @@ def __init__(self,
     def forward(self,  # type: ignore
                 text: Dict[str, torch.LongTensor],
                 spans: torch.IntTensor,
-                span_labels: torch.IntTensor = None,
-                metadata: List[Dict[str, Any]] = None) -> Dict[str, torch.Tensor]:
+                metadata: List[Dict[str, Any]],
```
matt-gardner: My first thought was that changing the order of this was a problem for backwards compatibility, but because it's a `Model.forward` method, we'll pass a dictionary of parameters, anyway, so this should probably be ok.

matt-gardner: Actually, looking at this again, `metadata` is in fact optional (e.g., it's not supplied in our demo), so there is no need to change this.

allennlp/models/graph_parser.py
```diff
@@ -128,16 +128,20 @@ def __init__(self,
     @overrides
     def forward(self,  # type: ignore
                 tokens: Dict[str, torch.LongTensor],
+                metadata: List[Dict[str, Any]],
```
matt-gardner: I'd probably default to making this optional instead of moving the parameter, but I could go either way.

allennlp/models/coreference_resolution/coref.py
bryant1410: Yeah, I thought about the same. Maybe warn in the next release, cause someone may unlikely but possibly pass params to a model like this one in order.

bryant1410: So you're saying that I should change `metadata` back to be optional? I see that's needed when `span_labels` is supplied (see the code at the end of this method). What should I do then? Just keep it optional, as it was?

allennlp/models/graph_parser.py
bryant1410: How? By returning the `"tokens"` entry only if `metadata` is provided?

allennlp/models/coreference_resolution/coref.py
matt-gardner: The intent of most of these models is that you have a "training" setting where you have labels provided, and a "testing" setting where there are no labels.  `metadata` is only needed in the "training" setting, where you have `span_labels` given.  So, yes, it is optional.

allennlp/models/graph_parser.py
matt-gardner: Yes.

matt-gardner:
@bryant1410, do you want to have this merged now, before it gets too stale, and have a separate PR for figuring out the CI check?  You've marked this as a draft.

bryant1410:
> @bryant1410, do you want to have this merged now, before it gets too stale, and have a separate PR for figuring out the CI check? You've marked this as a draft.

:+1: we can do that

PR #2666, opened 2019-03-29
Upgrade pip inside Dockerfiles.

brendan-ai2:
- Torch was failing to install in Docker when I attempted to release.
  - It's pretty quiet about this, incidentally.
  - If you log into a running image and use the verbose flag, something like`pip -v install /allennlp-0.8.3-py3-none-any.whl`, you'll get better logging.
- Fixed with `pip install --upgrade pip`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,8 @@ RUN apt-get update --fix-missing && apt-get install -y \
 # Copy select files needed for installing requirements.
 # We only copy what we need here so small changes to the repository does not trigger re-installation of the requirements.
 COPY requirements.txt .
+# Ensure pip is new enough to install AllenNLP's dependencies.
+RUN pip install --upgrade pip
 RUN pip install -r requirements.txt
 
 COPY scripts/ scripts/
--- a/Dockerfile.pip
+++ b/Dockerfile.pip
@@ -22,6 +22,9 @@ WORKDIR /stage/allennlp
 ARG VERSION
 ARG SOURCE_COMMIT
 
+# Ensure pip is new enough to install AllenNLP's dependencies.
+RUN pip install --upgrade pip
+
 # Install the specified version of AllenNLP.
 RUN if [ ! -z "$VERSION" ]; \
     then echo "Installing allennlp==$VERSION."; pip install allennlp==$VERSION; \


brendan-ai2:
@schmmd, I hit this while attempting to release. Please take a look!

brendan-ai2:
Actually, this isn't the correct fix. It looks like the release script runs on a vanilla docker image. https://github.com/allenai/allennlp-internal/blob/master/scripts/release/release.py#L137

Does this mean we need to bump our minimum python version?

brendan-ai2:
Bump our minimum python version and use a corresponding docker image, that is. Like python:3.6.8, say?

brendan-ai2:
Okay, using the `python:3.6.8` image works as well. So two questions:

1. Should I have us use the `python:3.6.8` image or should we run `pip install --upgrade pip` on the current slightly older image?
2. Separately, does this mean we should bump our required python version in `setup.py`? https://github.com/allenai/allennlp/blob/master/setup.py#L151

@joelgrus, perhaps you have some advice here?

joelgrus:
my inclination is that bumping the python version is a better solution. what was the exact error we were getting?

schmmd:
I would bump the Python version.

schmmd:
I don't think we need to pump the python version in setup.py since that's not actually connected to the version of pip used and our code should run on 3.6.1 (I think, @joelgrus?).  Bumping the python version on the Docker image might work because newer python Docker images typically use newer versions of pip.

However--in this case--I don't think you can bump up the Docker image since we're already on the latest version of 3.6.  @joelgrus knows best, but I don't think we're ready to start releasing on 3.7.

In other words, this PR seems like a reasonable way forward as is.

brendan-ai2:
@joelgrus, here is the full trace from `pip -v install /allennlp-0.8.3-py3-none-any.whl`.  
[old_pip_error.txt](https://github.com/allenai/allennlp/files/3023987/old_pip_error.txt)

@schmmd, sorry, I'm afraid my multiple messages above were confusing. To clarify, this PR as-is _does not work_. I have two verified working solutions:

1. Use `python:3.6.8` instead of `python:3.6.1` [here](https://github.com/allenai/allennlp-internal/blob/master/scripts/release/release.py#L137).
2. Continue using `python:3.6.1` in the line above, but run `pip install --upgrade pip` before installing AllenNLP.

And the followup question: If we choose 1), should we should bump our required python version in [setup.py](https://github.com/allenai/allennlp/blob/master/setup.py#L151)?

joelgrus:
is it really that 3.6.8 is required, or just that the docker images prior to 3.6.8 have an old version of pip?

like, if I were to install 3.6.1 the usual way, would I run into this error, or is it caused by what's in that docker image?

brendan-ai2:
Talked to Joel in person. 3.6.1 is not strictly required, so we're going with option 2.

PR #2676, opened 2019-03-31
Fix ArrayField.to_tensor's dtype when it's a scalar

bryant1410:


bryant1410:
I see that the CI only runs the last commit, but fyi the test I provided fails just with the first commit.

bryant1410:
@DeNeutoy this is ready to be merged

PR #2680, opened 2019-04-01
CRF Long Mask Bug

### LINKED ISSUE #2658: The Conditional Random Field is giving me a negative loss
 - OS: Linux
 - Python version: 3.6.8
 - AllenNLP version: v0.8.2
 - PyTorch version: 1.0.0

**Question**
I am building my own NER tool using your conditional random field. In general, it all works splendidly, but occasionally, I find that my cost goes negative. I did not have this problem just doing softmax-based sequence labeling. Since the cost is calculated as -1 times the log joint likelihood minus the log input likelihood, negative cost suggests the input is less likely than the input _and_ the underlying distribution, which is impossible. I am aware that the CRF itself does not apply the (*-1) and have applied it myself.

I know that doesn't mean that some small approximation error couldn't cause this behavior, but needless to say, it breaks my model. Is this expected behavior in edge cases where the model has drastically overfit? Should I just apply more regularization?

I've made sure that I'm using the mask properly (ones for the length of each sentence, followed by zeros for padding).

vidurj:
Fixes #2658. The issue is that the `ConditionalRandomField` assumes the mask is the same length as the input, but does not assert this. This PR both adds an assertion for this, and also stops assuming that the mask is the same length.


vidurj:
Since we no longer assume the `mask` and the `input` have the same length, should I be asserting it? It just seems like the user of the module probably has a bug in their code if these don't have the same length.

vidurj:
@joelgrus brought up the point that we assume a mask has the right shape for its inputs in a lot of places in our code without asserting it.

PR #2681, opened 2019-04-01
Bugfix: stop BasicTextFieldEmbedder from crashing from mismatched token numbers

mikerossgithub:
When different token indexers return differing numbers of tokens, BasicTextFieldEmbedder crashes when concatenating their embeddings. 

For me this occurred because WordPieceIndexer truncated a text that was not truncated by another indexer. The WordPieceIndexer truncation happens here (note that this line is a TODO):
https://github.com/allenai/allennlp/blob/0f4175034f5d79f9415e47d915603cbba6af01a6/allennlp/data/token_indexers/wordpiece_indexer.py#L160

This PR prevents the crash in BasicTextFieldEmbedder by truncating to the shortest embedding. This solved my use case. But I'm not sure if it's the best solution, given that the original truncation is caused by the unresolved TODO above.

### DIFF
--- a/allennlp/models/model.py
+++ b/allennlp/models/model.py
@@ -273,7 +273,7 @@ class Model(torch.nn.Module, Registrable):
         model.extend_embedder_vocab()
 
         model_state = torch.load(weights_file, map_location=util.device_mapping(cuda_device))
-        model.load_state_dict(model_state, strict=False)
+        model.load_state_dict(model_state)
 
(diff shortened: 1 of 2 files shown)


mikerossgithub:
For the specific WordPiece indexer, this is addressed by #2537. But maybe there is still a use for truncating within the BasicTextFieldEmbedder in case other Indexers have similar behavior in the future?

matt-gardner:
Thanks for the PR, but we feel like the solution in #2537 is the right one here.  Closing this one.

PR #2683, opened 2019-04-02
Correct The Pearson Correlation Formula In Docs

rangwani-harsh:


nelson-liu:
Thanks @rangwani-harsh !

PR #2698, opened 2019-04-07
Update pylint to 1.9.4

bryant1410:
Pinned pylint to the last 1.x version. Upgrading to 2.0 means much more changes.

This was motivated mostly because I realize that it was failing to check `allennlp/semparse/domain_languages/domain_language.py`, so it was ignoring checks there. I did the correspoding changes now that this file is checked. I also removed an unnecessary type ignore.

And I refactored `scripts/verify.py` a bit. This may be another PR, but I was wondering if the overhead it's worth it.

### DIFF
--- a/allennlp/semparse/domain_languages/domain_language.py
+++ b/allennlp/semparse/domain_languages/domain_language.py
@@ -27,13 +27,14 @@ def is_callable(type_: Type) -> bool:
         return getattr(type_, '_name', None) == 'Callable'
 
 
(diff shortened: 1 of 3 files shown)


bryant1410:
Do I have to do something because of the code coverage failing check? I don't understand it pretty well.

DeNeutoy:
No problem about the codecov, it just let's us catch actual additions which should be covered by tests. This looks great, thanks!

PR #2709, opened 2019-04-11
Number predictions

schmmd:
I was running some speed tests and it was difficult to tell how much progress my runs were making.  I thought that numbering the predictions would be a good way to add some measure of progress.


allennlp/commands/predict.py
```diff
@@ -171,14 +172,17 @@ def _get_instance_data(self) -> Iterator[Instance]:
 
     def run(self) -> None:
         has_reader = self._dataset_reader is not None
+        index = 0
```
joelgrus: you may find this too cute, but one thing you could do to avoid having to track and increment the indices would be

```
index = itertools.count()
```

which generates an infinite sequence 0, 1, 2, ...

and then grab its value like

```
self._maybe_print_to_console_and_file(next(index), result, ...)
```

HarshTrivedi:
@schmmd, this will be useful. Can you update and merge this?

allennlp/commands/predict.py
joelgrus: seems fine to me, as long as it's clear what's happening

schmmd:
@HarshTrivedi I'll merge after CI.

HarshTrivedi:
@schmmd  Thanks!  👍

PR #2716, opened 2019-04-14
 Add option in PytorchSeq2VecWrapper to return the state of all layers

### LINKED ISSUE #2411: How to obtain the final hidden states for all layers of a seq2vec encoder?
This is something I think it cannot be done with AllenNLP currently, so it can turn into a feature request. Or maybe it's just that I don't know how to do it. It's about getting the final hidden states for all layers of a seqvec encoder.

I'm trying to implement a model such as the [one from TGIF-QA](https://arxiv.org/abs/1604.02748). The thing is that, if I encode the video with a 2-layer LSTM with a `Seq2VecEncoder` (such as `PytorchSeq2VecWrapper`), I don't know how can I obtain all hidden states instead of just the one from the last layer. I need them to then pass them as initial hidden states of another seq2vec encoder. The architecture is the following:

TGIF-QA architecture

This neither can be done with a seq2seq encoder such as `PytorchSeq2SeqWrapper` because it returns the sequence of the last layer.

From my understanding, this cannot be done with AllenNLP out-of-the-box. I'd be happy to send a patch, but what could be the best way to do it? I believe this should be provided as a seq2vec model (cause conceptually it's what it is) but I'm reluctant to modify `PytorchSeq2VecWrapper` cause it would be a breaking change. Maybe adding a flag to it, to enable this behavior? Another class?

bryant1410:
Add option in PytorchSeq2VecWrapper to return the state of all layers. Fixes #2411. Also, it adds an option to return all the states (e.g., in LSTMs also return the memory).

I addressed some comments from https://github.com/allenai/allennlp/commit/1d9464b11dc7c597a7a13b4cb36cfa54e55eb2d7

Pending stuff:

* [x] Implement the main thing.
* [ ] Review the case where `mask is None`
* [ ] Document
* [ ] Try to cover with tests the cases: {w/ mask, w/o mask} x {lstm, no lstm} x {all layers, last layer} x {return all hidden states, don't return all hidden states}

Where should the *params* `return_all_layers` and `return_all_hidden_states` be documented? Not the `PytorchSeq2VecWrapper` args but the supported params for RNNs and etc. I mean, where people should learn about them?

### DIFF
--- a/allennlp/modules/seq2vec_encoders/__init__.py
+++ b/allennlp/modules/seq2vec_encoders/__init__.py
@@ -71,10 +71,10 @@ class _Seq2VecWrapper:
         if self._module_class in self.PYTORCH_MODELS:
             params['batch_first'] = True
         module = self._module_class(**params.as_dict())
-        return_only_last = params.pop('return_only_last', True)
-        return_tuple_if_state_is_tuple = params.pop('return_tuple_if_state_is_tuple', False)
-        return PytorchSeq2VecWrapper(module, return_only_last=return_only_last,
-                                     return_tuple_if_state_is_tuple=return_tuple_if_state_is_tuple)
+        return_all_layers = params.pop('return_all_layers', False)
+        return_all_hidden_states = params.pop('return_all_hidden_states', False)
+        return PytorchSeq2VecWrapper(module, return_all_layers=return_all_layers,
+                                     return_all_hidden_states=return_all_hidden_states)
 
 # pylint: disable=protected-access
 Seq2VecEncoder.register("gru")(_Seq2VecWrapper(torch.nn.GRU))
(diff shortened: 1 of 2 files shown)


bryant1410:
Btw, I think the `mask is None` case had a bug: it didn't return the 2 directions if it's bidirectional.

bryant1410:
I'm realizing that if the output tensor shape changes (or if the output is a tuple), it breaks the use of `TimeDistributed`. Thoughts?

matt-gardner:
It breaks more than just `TimeDistributed`, it breaks compatibility with the `Seq2VecEncoder` API.  If you want this functionality, you'll be instantiating the `PytorchSeq2VecWrapper` directly yourself, knowing that you can't use this as a `Seq2VecEncoder`, and you'll have full control over what you do with it (e.g., how many times you wrap it in a `TimeDistributed`, if you need that).  (We do similar things in at least one other place, so I'm not too worried about allowing this, as long as there are clear warnings around these parameters explaining what they do and why.)

matt-gardner:
@bryant1410, any intention to follow up on this?

bryant1410:
> @bryant1410, any intention to follow up on this?

Yeah. I've been busy with other stuff. I don't have an ETA though.

DeNeutoy:
closing due to inactivity, it seemed like we were a little stuck on how to do this also.

Issue #2717, opened 2019-04-15
Metrics: "training_duration" does not include day count


scarecrow1123:
When the training duration gets longer than a day, the `metrics.json` misses to include the day count and only has "hh:mm:ss". It should rather have "dd:hh:mm:ss".

PR #2718, opened 2019-04-15
Fix to include day count in training duration for metrics.json 

scarecrow1123:
This is a tiny fix to have day count to be included in `training_duration` in the metrics json. #2717


joelgrus:
technically this isn't quite right either, since it does the wrong thing if your training takes more than 31 days (which is unlikely, but still, might as well fix it for good). 

I suspect that `strftime` isn't the right general solution for this (or at least you'll have to extract the number of days manually)

scarecrow1123:
> technically this isn't quite right either, since it does the wrong thing if your training takes more than 31 days (which is unlikely, but still, might as well fix it for good).
> 
> I suspect that `strftime` isn't the right general solution for this (or at least you'll have to extract the number of days manually)

Thanks for the nudge. Yes, `strftime` may not be a good fit and I've fixed it to use `datetime.timedelta`. Let me know if that would do.

joelgrus:
thanks

PR #2726, opened 2019-04-18
Add caching functionality to the base DatasetReader class

matt-gardner:
This adds a parameter that lets you specify a path to a cache file, where we will save already-constructed `Instances` for faster loading on subsequent reads of the data.  If you do a lot of processing, like running a tagger or expensive tokenizer or other things, this should save you a lot of time.  If you have only minimal pre-processing, this probably won't buy you much.

### DIFF
--- a/allennlp/data/dataset_readers/dataset_reader.py
+++ b/allennlp/data/dataset_readers/dataset_reader.py
@@ -86,7 +86,7 @@ class DatasetReader(Registrable):
         unique cache directories.  If you don't use our commands, that is your responsibility.
         """
         self._cache_directory = pathlib.Path(cache_directory)
-        self._cache_directory.mkdir(exist_ok=True)
+        os.makedirs(self._cache_directory, exist_ok=True)
 
     def read(self, file_path: str) -> Iterable[Instance]:
         """
--- a/allennlp/data/dataset_readers/dataset_reader.py
+++ b/allennlp/data/dataset_readers/dataset_reader.py
@@ -30,15 +30,15 @@ class _LazyInstances(Iterable):
     def __iter__(self) -> Iterator[Instance]:
         # Case 1: Use cached instances
         if self.cache_file is not None and os.path.exists(self.cache_file):
-            with open(self.cache_file) as f:
-                for line in f:
+            with open(self.cache_file) as data_file:
+                for line in data_file:
                     yield self.deserialize(line)
         # Case 2: Need to cache instances
         elif self.cache_file is not None:
-            with open(self.cache_file, 'w') as f:
+            with open(self.cache_file, 'w') as data_file:
                 for instance in self.instance_generator():
-                    f.write(self.serialize(instance))
-                    f.write("\n")
+                    data_file.write(self.serialize(instance))
+                    data_file.write("\n")
                     yield instance
         # Case 3: No cache
         else:
--- a/allennlp/data/dataset_readers/dataset_reader.py
+++ b/allennlp/data/dataset_readers/dataset_reader.py
@@ -77,7 +77,7 @@ class DatasetReader(Registrable):
     """
     def __init__(self, lazy: bool = False) -> None:
         self.lazy = lazy
-        self._cache_directory = None
+        self._cache_directory: str = None
 
     def cache_data(self, cache_directory: str) -> None:
         """
--- a/allennlp/data/dataset_readers/dataset_reader.py
+++ b/allennlp/data/dataset_readers/dataset_reader.py
@@ -1,6 +1,7 @@
 from typing import Iterable, Iterator, Callable
 import logging
 import os
+import pathlib
 
 import jsonpickle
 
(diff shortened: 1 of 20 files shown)


allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -23,6 +26,39 @@ def __iter__(self) -> Iterator[Instance]:
...
+        super().__init__()
+        self.read_from_file_method = read_from_file_method
+        self.read_from_cache_method = read_from_cache_method
+        self.serialize_instance_method = serialize_instance_method
+        self.file_path = file_path
+        self.cache_file = cache_file
+
+    def __iter__(self) -> Iterator[Instance]:
+        if os.path.exists(self.cache_file):
+            return self.read_from_cache_method(self.cache_file)
```
matt-gardner: I'm getting a mypy complaint on this line that I'm not sure about.  I think I maybe need to wrap this somehow, but I'm not certain.

matt-gardner:
Yes, this looks much better, thanks for the suggestions.  I was struggling with how to make the `_LazyInstances` do what I wanted, which is why I ended up with a separate class.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -13,15 +16,37 @@ class _LazyInstances(Iterable):
...
+        elif self.cache_file is not None:
+            with open(self.cache_file, 'w') as f:
+                for instance in self.instance_generator():
+                    f.write(self.serialize(instance))
+                    f.write("\n")
+                    yield instance
+        # Case 3: No cache
+        else:
+            instances = self.instance_generator()
+            if isinstance(instances, list):
```
joelgrus: now that we're using LazyInstances in both cases, this check is no longer correct (and is causing tests to fail).

I'm not sure how to fix it

matt-gardner: Hmm, I thought I checked the tests...  Looks like it's a couple I didn't run locally.  I'll figure it out.

matt-gardner: I believe this is because your `LazyFakeReader` class isn't actually lazy, even though it says it is: https://github.com/allenai/allennlp/blob/9dec020281ee9521e7f1ffd696bcbb102c399703/allennlp/tests/commands/train_test.py#L173-L184

So this new logic appears to be more correct than the old logic, because this test should have failed earlier.

Similarly, the DROP reader test is marked as being both lazy and non-lazy, but the reader did not `yield`, it returned a list.  I'm not sure why the previous logic didn't crash in this case.

allennlp/training/util.py
```diff
@@ -122,18 +123,85 @@ def str_to_time(time_str: str) -> datetime.datetime:
...
+        # We need to compute the parameter hash before the parameters get destroyed when they're
+        # passed to `DatasetReader.from_params`.
+        if not cache_prefix:
+            cache_prefix = _dataset_reader_param_hash(dataset_reader_params)
+            if validation_dataset_reader_params:
+                validation_cache_prefix = _dataset_reader_param_hash(validation_dataset_reader_params)
+            else:
+                validation_cache_prefix = cache_prefix
+        else:
+            validation_cache_prefix = cache_prefix
```
joelgrus: 1. I feel like the details of how the cache gets setup are ancillary to this function, so probably most of this "set up the cache" logic should be pulled out into its own function?
2. all this manual adding of slashes feels really clunky, I wonder if you could avoid it using either `os.path.join` or `pathlib.Path`?
3. similarly, all of the path creation using `+` feels clunky. 

Couldn't you do something like

```
cache_path = pathlib.Path(cache_directory) / cache_prefix
validation_cache_path = ...
```

and then you don't have to do all the addition?

matt-gardner: I believe I fixed all of this.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -84,10 +85,8 @@ def cache_data(self, cache_directory: str) -> None:
         is responsible for calling this method and ensuring that unique parameters correspond to
         unique cache directories.  If you don't use our commands, that is your responsibility.
         """
-        if not cache_directory.endswith('/'):
-            cache_directory += '/'
-        self._cache_directory = cache_directory
-        os.makedirs(self._cache_directory, exist_ok=True)
+        self._cache_directory = pathlib.Path(cache_directory)
+        self._cache_directory.mkdir(exist_ok=True)
```
matt-gardner: This is not recursive, so it breaks in the normal use case.  You have to use `os.makedirs()` instead.  I _thought_ that `os.makedirs` would require the trailing slash, which is why I didn't want to remove it, but it appears to be working with `pathlib.Path(cache_directory)`.  I just pushed a change here; if it passes I'll merge this.

joelgrus: oops, it looks like `Path.mkdir` takes a `parents=True` parameter that makes it recursive.

PR #2731, opened 2019-04-21
Added NAQANet to pretrained.py and DEFAULT_PREDICTORS

### LINKED ISSUE #2730: I could find predictor models for all demo examples except NAQANet.
Can you add that one as well? 

https://allennlp.org/models

It is missing NAQANet model used on the demo page:
https://demo.allennlp.org/reading-comprehension

matt-gardner:
Fixes #2730.

### DIFF
--- a/allennlp/pretrained.py
+++ b/allennlp/pretrained.py
@@ -37,6 +37,13 @@ def bidirectional_attention_flow_seo_2017() -> predictors.BidafPredictor:
                                 'machine-comprehension')
         return model.predictor() # type: ignore
 
+def naqanet_dua_2019() -> predictors.BidafPredictor:
+    with warnings.catch_warnings():
+        warnings.simplefilter(action="ignore", category=DeprecationWarning)
+        model = PretrainedModel('https://s3-us-west-2.amazonaws.com/allennlp/models/naqanet-2019.03.01.tar.gz',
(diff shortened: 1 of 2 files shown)


allennlp/pretrained.py
```diff
@@ -37,6 +37,13 @@ def bidirectional_attention_flow_seo_2017() -> predictors.BidafPredictor:
                                 'machine-comprehension')
         return model.predictor() # type: ignore
 
+def naqanet_dua_2019() -> predictors.BidafPredictor:
+    with warnings.catch_warnings():
+        warnings.simplefilter(action="ignore", category=DeprecationWarning)
+        model = PretrainedModel('https://s3-us-west-2.amazonaws.com/allennlp/models/naqanet-2019.03.01.tar.gz',
+                                'machine-comprehension')
```
DeNeutoy: This is almost certainly a deficit in my python knowledge but should this return statement be inside the `catch_warnings()` scope? Does that scope survive outside of the function?

joelgrus: I don't think a scope can ever survive outside the function it's defined in

matt-gardner: And the warning that's being caught is getting thrown by the constructor, so we don't need to worry about it, anyway.

PR #2734, opened 2019-04-22
Allow FromParams to construct Union types

matt-gardner:
This removes the need for custom `from_params` for `FeedForward` (and lets me more easily add some other modules that have similar behavior).

### DIFF
--- a/allennlp/common/from_params.py
+++ b/allennlp/common/from_params.py
@@ -335,8 +335,7 @@ class FromParams:
                 # you've done the right thing in passing your parameters, and nothing else needs to
                 # be recursively constructed.
                 if not takes_arg(subclass, 'extras'):
-                    # Necessarily subclass.from_params is a custom implementation, so we need to
-                    # pass it only the args it's expecting.
+                    # We should only pass on the extras that the constructor actually expects.
                     extras = {k: v for k, v in extras.items() if takes_arg(subclass, k)}
                 constructor_args = {**params, **extras}
                 return subclass(**constructor_args)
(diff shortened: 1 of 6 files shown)


allennlp/common/from_params.py
```diff
@@ -113,125 +111,169 @@ def create_kwargs(cls: Type[T], params: Params, **extras) -> Dict[str, Any]:
...
 
-        else:
-            # Pass it on as is and hope for the best.   ¯\_(ツ)_/¯
-            if optional:
-                kwargs[name] = params.pop(name, default)
-            else:
-                kwargs[name] = params.pop(name)
+    elif origin == Union:
+        # Storing this so we can recover it later if we need to.
+        param_value = params.get(name, Params({}))
```
joelgrus: shouldn't you use `.duplicate()` here? what if e.g. the first union type consumes some of the params before failing?

allennlp/common/from_params.py
```diff
@@ -113,125 +111,169 @@ def create_kwargs(cls: Type[T], params: Params, **extras) -> Dict[str, Any]:
...
-    return kwargs
+        # We'll try each of the given types in the union sequentially, returning the first one that
+        # succeeds.
+        for arg in args:
+            try:
+                return construct_arg(cls, name, arg, default, params, **extras)
+            except (ValueError, TypeError, ConfigurationError, AttributeError) as e:
+                # Our attempt to construct the argument may have popped `params[name]`, so we
+                # restore it here.
+                params[name] = param_value
```
joelgrus: and then similarly you should use `.duplicate()` here

allennlp/common/from_params.py
matt-gardner: Hmm...  I think this will never happen if you're dealing with `Union[X, List[X]]`, as I was thinking about (so the tests I wrote pass just fine), but it could easily happen with `Union[X, Dict[X]]`.  Thanks for pointing it out.

allennlp/common/from_params.py
matt-gardner: Done (and I believe I fixed the CI issues also).

PR #2735, opened 2019-04-23
Enable span-level F1 metric in SimpleTagger

nelson-liu:
This PR enables calculation of the span-level F1 metrics in SimpleTagger


allennlp/models/simple_tagger.py
```diff
@@ -39,23 +55,43 @@ class SimpleTagger(Model):
...
         self.encoder = encoder
+        self._verbose_metrics = verbose_metrics
         self.tag_projection_layer = TimeDistributed(Linear(self.encoder.get_output_dim(),
                                                            self.num_classes))
 
         check_dimensions_match(text_field_embedder.get_output_dim(), encoder.get_input_dim(),
                                "text field embedding dim", "encoder input dim")
+
+        if calculate_span_f1 is None:
+            calculate_span_f1 = label_encoding is not None
```
joelgrus: as I understand this, if you specify a label encoding, calculate_span_f1 is implicitly true, and if you don't then it's implicitly false? if that's the case, then why have a calculate_span_f1 parameter at all?

nelson-liu: 🤦‍♂️ sorry, was just copy-pasting from CRF tagger. If we remove it, not sure that the names `label_encoding` and `label_namespace` together suggest that they're used for calculating span F1---seems prudent to rename them? Any suggestions? I was thinking `span_f1_label_encoding` and `span_f1_label_namespace`, but maybe that's a bit verbose?

joelgrus: I guess the difference over there is that `label_encoding` does something useful even if `calculate_span_f1` is false, whereas here it doesn't.

I guess it's probably for the best to keep the APIs of these classes as similar as possible

joelgrus: but I would add a comment here explaining the reasoning, because it's not obvious

allennlp/models/simple_tagger.py
```diff
@@ -39,23 +55,43 @@ class SimpleTagger(Model):
...
+        self.label_encoding = label_encoding
+
         self.metrics = {
                 "accuracy": CategoricalAccuracy(),
                 "accuracy3": CategoricalAccuracy(top_k=3)
         }
 
+        self.calculate_span_f1 = calculate_span_f1
```
joelgrus: do we need to track `self.calculate_span_f1`? can't we just check whether `self._f1_metric` is None?

nelson-liu: In this case, yes.

allennlp/models/simple_tagger.py
```diff
@@ -39,23 +55,47 @@ class SimpleTagger(Model):
...
         self.metrics = {
                 "accuracy": CategoricalAccuracy(),
                 "accuracy3": CategoricalAccuracy(top_k=3)
         }
 
+        if calculate_span_f1:
+            if not label_encoding:
+                raise ConfigurationError("calculate_span_f1 is True, but "
+                                         "no label_encoding was specified.")
+            self._f1_metric = SpanBasedF1Measure(vocab,
```
joelgrus: ok, so having thought about this more, I think it would be slightly cleaner ("flat is better than nested") if you did the following. first in its own block, do

```
if calculate_span_f1 and not label_encoding:
    raise ConfigurationError(...)
```

Then in a second block do

```
if calculate_span_f1 or label_encoding:
    self._f1_metric = ...
else:
    self._f1_metric = None
```

I feel like the logic is a lot clearer than with the nested ifs

allennlp/tests/models/simple_tagger_test.py
```diff
@@ -71,6 +71,20 @@ def test_regularization(self):
         numpy.testing.assert_almost_equal(training_loss, validation_loss)
 
 
+class SimpleTaggerSpanF1Test(ModelTestCase):
+    def setUp(self):
+        super(SimpleTaggerSpanF1Test, self).setUp()
```
brendan-ai2: @nelson-liu, quick tip, you can just say `super().setUp()` here and get the same effect.

Issue #2404, opened 2019-01-20
Feature request: metric for official CoNLL SRL evaluation


nelson-liu:
Related to #667 , context is that the `SpanBasedF1Measure` does not exactly match the official conll evaluation perl script's output.

Since we wrap the `evalb` C program in an metric, why can't we do the same for the [`srl-eval.pl`](https://github.com/allenai/allennlp/blob/master/allennlp/tools/srl-eval.pl) script? We can turn it into another metric, and just have it run at validation time (just like evalb does now).

This isn't a solution to the problem, but it at least makes working with the srl model less annoying.

joelgrus:
it has 0 github stars, but here's someone who's written a python version

https://github.com/jgung/srleval

if you trust it, that's even easier

nelson-liu:
"Note, the official evaluation script should be used for any published results." in the README has me worried 😅

ducalpha:
The `_handle_continued_spans` function in `SpanBasedF1Measure` refers to **CoNLL 2012** evaluation script (on [this line](https://github.com/allenai/allennlp/blob/9f87aa56177ef3dc1b0c58ebb49d15584c500e29/allennlp/training/metrics/span_based_f1_measure.py#L189) but I think it should be **CoNLL 2005** instead. CoNLL 2012 shared task evaluates coreference predictions while CoNLL 2005 uses the [`srl-eval.pl`](https://github.com/allenai/allennlp/blob/master/allennlp/tools/srl-eval.pl).

nelson-liu:
@ducalpha sorry for the very belated response on this, you're right! But also worth noting that many SRL papers also evaluate on CoNLL 2005 and CoNLL 2012 (e.g., the paper that the AllenNLP SRL model is based off of: https://www.aclweb.org/anthology/P17-1044)

PR #2743, opened 2019-04-24
Add SrlEvalMetric

nelson-liu:
This PR adds an SrlEvalMetric that wraps the `srl-eval.pl` script and produces SRL F1 scores that are actually accurate, unlike the slightly-off `SpanBasedF1Measure`.

Fixes https://github.com/allenai/allennlp/issues/2404
Also an intermediate step toward fixing https://github.com/allenai/allennlp/issues/667

~Opening PR now because I want some feedback about whether this is the right way to go. In particular, I have the logic for converting from BIO output to CoNLL output in the metric right now. However, this essentially constrains the metric to be used for models that treat the SRL task like BIO tagging, while some don't (e.g., this recent SOTA paper https://arxiv.org/abs/1805.04787).~

~As a result, I'm thinking of moving the logic to go from BIO -> CoNLL format into the model. Then, the metric would then take as input a sequence of CoNLL-formatted SRL tags (instead of the BIO tags), and then handle the heavy lifting of writing the gold / predicted labels to file, running the perl script, and parsing the output.~

*Edit: I went and made the changes above.*

Thoughts? Not sure which of @DeNeutoy , @brendan-ai2 , @joelgrus want to take this...

TODO:

- [ ] improve tests


nelson-liu:
Also not too sure how to test this (the test I have right now is pretty trivial)---any suggestions? Was thinking of just copying the test from `SpanBasedF1Measure`, but I'd prefer to independently verify if possible.

allennlp/training/metrics/srl_eval_scorer.py
```diff
@@ -0,0 +1,159 @@
...
+                    batch_conll_formatted_predicted_tags,
+                    batch_conll_formatted_gold_tags):
+                write_conll_formatted_tags_to_file(predicted_file,
+                                                   gold_file,
+                                                   verb_index,
+                                                   sentence,
+                                                   predicted_tag_sequence,
+                                                   gold_tag_sequence)
+        perl_script_command = (f"perl {self._srl_eval_path} {predicted_path} {gold_path} "
+                               f"> {output_path}")
```
brendan-ai2: Rather than redirecting to a file you can directly capture the stdout. Should be something like:

```
completed_process = subprocess.run(["perl", self._srl_eval_path, predicted_path, gold_path], check=True, text=True)
for line in completed_process.stdout.split("\n"):
...
```

(Untested, so do check the docs: https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess)

This may help address your comment above about writing to disk often. In general it's also nice to avoid shell expansion for security's sake. (Though I think all the paths here should be controlled by the developer, so they should be safe...)

nelson-liu: ok, good to know! We should probably also fix this in the evalb metric, then, but that's a separate issue: https://github.com/allenai/allennlp/blob/master/allennlp/training/metrics/evalb_bracketing_scorer.py#L96-L98

brendan-ai2:
To the extent that the Perl script is already tested/trusted, I think it's reasonable to test this metric fairly lightly. I'll give this PR a closer pass tomorrow.

allennlp/training/metrics/srl_eval_scorer.py
```diff
@@ -0,0 +1,157 @@
...
+            the gold labels are the arguments for, or None if the sentence
+            contains no verbal predicate.
+        batch_sentences : ``List[List[str]]``, required.
+            The word tokens for each instance in the batch.
+        batch_conll_formatted_predicted_tags : ``List[str]``, required.
+            A list of predicted CoNLL-formatted SRL tags (itself a list) to compute score for.
+            Use allennlp.models.semantic_role_labeler.convert_bio_tags_to_conll_format
+            to convert from BIO to CoNLL format before passing the tags into the metric,
+            if applicable.
+        batch_conll_formatted_gold_tags : ``List[str]``, required.
```
brendan-ai2: There are some type mismatches between this docstring and the signature.

allennlp/training/metrics/srl_eval_scorer.py
```diff
@@ -0,0 +1,157 @@
...
+        all_tags: Set[str] = set()
+        all_tags.update(self._true_positives.keys())
+        all_tags.update(self._false_positives.keys())
+        all_tags.update(self._false_negatives.keys())
+        all_metrics = {}
+        for tag in all_tags:
+            precision, recall, f1_measure = self._compute_metrics(self._true_positives[tag],
+                                                                  self._false_positives[tag],
+                                                                  self._false_negatives[tag])
+            precision_key = "precision" + "-" + tag
```
brendan-ai2: Can the user possibly provide tags such that `tag == "overall"` or is that entirely decided by the Perl script? Please assert against it if needed.

allennlp/tests/training/metrics/srl_eval_scorer_test.py
```diff
@@ -0,0 +1,44 @@
...
+                                 tags in batch_bio_gold_tags]
+
+        srl_scorer = SrlEvalScorer()
+        srl_scorer(batch_verb_indices,
+                   batch_sentences,
+                   batch_conll_predicted_tags,
+                   batch_conll_gold_tags)
+        metrics = srl_scorer.get_metric()
+        for metric in metrics.values():
+            assert_allclose(metric, 1.0)
```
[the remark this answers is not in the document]
brendan-ai2: Thanks for breaking those out. However, I'll stand by my statement that having the values differ somewhat would be preferable. At least enough to show that precision/recall/f1 didn't accidentally get mixed up.

DeNeutoy:
@nelson-liu i'd love to get this merged, it would help me out for some modelling stuff. Looks like it's nearly there?

nelson-liu:
Oops, I didn't realize that Brendan responded---must have gotten buried in my inbox. I'll finish this today @DeNeutoy .

PR #2744, opened 2019-04-24
add print-results command

DeNeutoy:
After training models, the next step for me is often copy and pasting results into some kind of Death Star Tracking Google Doc. I always write a script which does this, but doing that is super annoying and somehow I always  manage to loose the script as I transition between research projects.

This command just makes this process much more flexible, as you can choose the keys of the metrics dictionary you want to print out, and which metrics json file you want to look at (e.g the result of running `allennlp evaluate` might be `test.json`, or you might want to look at metrics half way through training or something)


allennlp/commands/print_results.py
```diff
@@ -0,0 +1,88 @@
...
+    metrics_name = args.metrics_filename
+    keys = args.keys
+
+    results_dict = {}
+    for root, _, files in os.walk(path):
+
+        if metrics_name in files:
+            full_name = os.path.join(root, metrics_name)
+            metrics = json.load(open(full_name))
+            results_dict[full_name] = metrics
```
matt-gardner: Because `metrics_name` is necessarily the same across all files, I'd expect the table to be nicer if you just used `root` instead of including `metrics_name` here.

allennlp/commands/print_results.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+    sorted_keys = sorted(list(results_dict.keys()))
+    print(f"model_run, {', '.join(keys)}")
```
matt-gardner: Might be nice to add some sort of `separator` flag here, so that you could easily use this to make a latex table for a paper.  You'd probably also need some more configuration stuff for that, too, though, so it might be easier to specify a `format` flag instead of a `separator`...

It's fine to leave this for a separate PR when somebody wants this functionality.

allennlp/commands/print_results.py
DeNeutoy: That's true for a single run of this command, but for multiple runs with different metrics names (which might exist from running `allennlp evaluate`), I thought it would be nice to be able to see exactly which they came from.

allennlp/commands/print_results.py
DeNeutoy: Right, there are probably extensions to this which are helpful, e.g `--format [json|csv|tsv]` but also probably backward compatible. I'll leave those for now.

PR #2749, opened 2019-04-24
lowercase stopword set in word filter

### LINKED ISSUE #2748: StopwordFilter's additional tokens should be lowered at __init__ time
**Describe the bug**
When creating a StopwordFilter (from allennlp/data/tokenizers/word_filter.py) and specifying additional tokens to be considered as stopwords, through the parameter "tokens_to_add", these new tokens should be lowered (calling the string method ".lower()).
Otherwise, when calling the function "filter_words()", these new tokens won't be filtered, as the function only excludes lowered tokens:

`def filter_words(self, words: List[Token]) -> List[Token]:`
`return [word for word in words if word.text.lower() not in self.stopwords]`

and this is not how they are added to the set of tokens to exclude:

`for token in self._tokens_to_add:`
`self.stopwords.add(token)`

**Expected behavior**
Added tokens should be excluded when applying the filter_words method.

DeNeutoy:
fixes #2748


allennlp/data/tokenizers/word_filter.py
```diff
@@ -74,33 +75,34 @@ def __init__(self,
...
+            self.stopwords = set([token.lower() for token in
+                                  ['I', 'a', 'aboard', 'about', 'above', 'accordance', 'according',
+                                   'across', 'after', 'against', 'along', 'alongside', 'also', 'am',
+                                   'amid', 'amidst', 'an', 'and', 'apart', 'are', 'around', 'as',
+                                   'aside', 'astride', 'at', 'atop', 'back', 'be', 'because', 'before',
+                                   'behind', 'below', 'beneath', 'beside', 'besides', 'between',
+                                   'beyond', 'but', 'by', 'concerning', 'do', 'down', 'due', 'during',
+                                   'either', 'except', 'exclusive', 'false', 'for', 'from', 'happen',
+                                   'he', 'her', 'hers', 'herself', 'him', 'himself', 'his', 'how',
+                                   'how many', 'how much', 'i', 'if', 'in', 'including', 'inside',
```
brendan-ai2: Interestingly lowercase `'i'` already appears here. Any chance this filter was intended to be case sensitive originally and the `.lower()` call in `filter_words` is the bug?

DeNeutoy: Not sure, but lowercasing seems correct in the sense that I would expect stopwords which were capitalized to be filtered, so I think i'll leave it as is.

PR #2754, opened 2019-04-25
Extras checks

### LINKED ISSUE #2665: Custom FromParams need to also accept **extras
For classes with custom from_params (eg: FeedForward), the from_params method needs to accept extras. Otherwise, if being instantiated from a class with extras (Eg: Model), the current FromParams implementation passes them to the class, which results it an error.

Consider the following code snippet
```
@Model.register("tmp_model")`
class TmpModel(Model):
    def __init__(self, vocab: Vocabulary, feed_forward_list: List[FeedForward]):
        super(TmpModel, self).__init__(vocab=vocab)
        self.feed_forward = nn.ModuleList(feed_forward_list)

params = Params({
    "type": "tmp_model",
    "feed_forward": [{"input_dim": 512, "num_layers": 1, "hidden_dims": [2], "activations": ["linear"]}]
})

vocab = Vocabulary()
model = Model.from_params(vocab=vocab, params=params)
```

**Error**
from_params() got an unexpected keyword argument 'vocab'

**System:**
 - OS: [e.g. OSX, Linux]
 - AllenNLP version: [e.g. v0.8.2, or "I installed from master"]

codedecde:
Fixes #2665.

This ensures only extras relevant to a class are passed to its from_params constructor for classes with custom from_params. Additionally, also handles the hacky scenario when the class does not have a from_params method, the assumption there being that we want to pass extras for the class itself (instead if its from_params method)

### DIFF
--- a/allennlp/common/from_params.py
+++ b/allennlp/common/from_params.py
@@ -73,7 +73,6 @@ def takes_arg(obj, arg: str) -> bool:
 
 
 def takes_kwargs(obj) -> bool:
-    # pylint: disable=protected-access,simplifiable-if-statement
     """
     Checks whether a provided object takes in any positional arguments.
     Similar to takes_arg, we do this for both the __init__ function of
@@ -86,7 +85,7 @@ def takes_kwargs(obj) -> bool:
         signature = inspect.signature(obj)
     else:
         raise ConfigurationError(f"object {obj} is not callable")
-    return bool(any([p.kind == inspect._ParameterKind.VAR_KEYWORD  # type: ignore
+    return bool(any([p.kind == inspect.Parameter.VAR_KEYWORD  # type: ignore
                      for p in signature.parameters.values()]))
 
 
--- a/allennlp/common/from_params.py
+++ b/allennlp/common/from_params.py
@@ -73,6 +73,7 @@ def takes_arg(obj, arg: str) -> bool:
 
 
 def takes_kwargs(obj) -> bool:
+    # pylint: disable=protected-access,simplifiable-if-statement
     """
     Checks whether a provided object takes in any positional arguments.
     Similar to takes_arg, we do this for both the __init__ function of
@@ -85,11 +86,8 @@ def takes_kwargs(obj) -> bool:
         signature = inspect.signature(obj)
     else:
         raise ConfigurationError(f"object {obj} is not callable")
-    if any([p.kind == inspect._ParameterKind.VAR_KEYWORD  # type: ignore
-            for p in signature.parameters.values()]):
-        return True
-    else:
-        return False
+    return bool(any([p.kind == inspect._ParameterKind.VAR_KEYWORD  # type: ignore
+                     for p in signature.parameters.values()]))
 
 
 def remove_optional(annotation: type):
--- a/allennlp/common/from_params.py
+++ b/allennlp/common/from_params.py
@@ -76,7 +76,7 @@ def takes_kwargs(obj) -> bool:
     """
     Checks whether a provided object takes in any positional arguments.
     Similar to takes_arg, we do this for both the __init__ function of
-    the classp or a function / method
+    the class or a function / method
     Otherwise, we raise an error
     """
     if inspect.isclass(obj):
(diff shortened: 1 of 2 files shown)


allennlp/common/from_params.py
```diff
@@ -116,6 +116,37 @@ def create_kwargs(cls: Type[T], params: Params, **extras) -> Dict[str, Any]:
...
+    if hasattr(cls, "from_params"):
+        from_params_method = cls.from_params
+    else:
+        # In some rare cases, we get a registered subclass that does _not_ have a
+        # from_params method (this happens with Activations, for instance, where we
+        # register pytorch modules directly).  This is a bit of a hack to make those work,
+        # instead of adding a `from_params` method for them somehow. Then the extras
+        # in the class constructor are what we are looking for, to pass on.
+        from_params_method = cls
+    if takes_arg(from_params_method, "extras"):
```
matt-gardner: I think we actually want a check for something like `**extras`, whatever its name, instead of just hard-coding the _name_ `extras`.  This possible by checking for the `kind` of a parameter with `inspect`:

```python
import inspect

def f(a: str, *inputs, **kwargs):
    pass

s = inspect.signature(f)
for p in s.parameters.values():
    print(p.name, p.kind)
```

output:
```
a POSITIONAL_OR_KEYWORD
inputs VAR_POSITIONAL  
extras VAR_KEYWORD     
```

So, a function like `takes_kwargs(method)` that checks for the presence of a `VAR_KEYWORD` argument would be better here.

matt-gardner: Looking at the original code, it looks like you're inheriting this logic instead of adding it, but it'd still be good to fix it while we're touching this code.

allennlp/common/from_params.py
```diff
@@ -116,6 +116,37 @@ def create_kwargs(cls: Type[T], params: Params, **extras) -> Dict[str, Any]:
...
+def create_extras(cls: Type[T],
+                  extras: Dict[str, Any]) -> Dict[str, Any]:
+    """
+    Given a dictionary of extra arguments, returns a dictionary of
+    kwargs that actually are a part of the signature of the cls.from_params
+    (or cls) method.
+    """
+    subextras: Dict[str, Any] = {}
+    if hasattr(cls, "from_params"):
+        from_params_method = cls.from_params
```
matt-gardner: mypy is complaining about this line; you can just add `  # type: ignore` at the end, because mypy is wrong.

allennlp/common/from_params.py
codedecde: The latest commit fixes this issue.

allennlp/common/from_params.py
```diff
@@ -70,6 +71,27 @@ def takes_arg(obj, arg: str) -> bool:
...
+    the classp or a function / method
+    Otherwise, we raise an error
+    """
+    if inspect.isclass(obj):
+        signature = inspect.signature(obj.__init__)
+    elif inspect.ismethod(obj) or inspect.isfunction(obj):
+        signature = inspect.signature(obj)
+    else:
+        raise ConfigurationError(f"object {obj} is not callable")
+    if any([p.kind == inspect._ParameterKind.VAR_KEYWORD  # type: ignore
```
matt-gardner: pylint has two complaints about this line, and I think it's reasonable to just ignore both of them (with `# pylint: disable=protected-access,simplifiable-if-statement` inside the scope of this function).

allennlp/tests/common/from_params_test.py
```diff
@@ -132,6 +132,111 @@ def from_params(cls, params: Params, size: int) -> 'C':  # type: ignore
         assert c.name == "extra_c"
         assert c.size == 20
 
+    def test_extras_for_custom_classes(self):
```
matt-gardner: This file still has some mypy and pylint errors.

allennlp/common/from_params.py
codedecde: Changed the inspect to not access the protected variable, and changed the return type to match the simplifiable-if-statement part.

allennlp/tests/common/from_params_test.py
codedecde: Fixed.

Issue #2691, opened 2019-04-04
Install jre instead of jdk


schmmd:
In our `Dockerfile` we can probably install something like `apt install openjdk-8-jre-headless` instead of the full jdk and save a ton of space.

schmmd:
https://github.com/allenai/allennlp/pull/2764 removed our dependency on the JDK.  Once it's merged we can remove both the JDK and the JRE from our Dockerfile (or even better, we can put it in the PR itself!)

PR #2764, opened 2019-04-26
WTQ dataset reader and models moved from iterative-search-semparse

pdasigi:
Fixes https://github.com/allenai/allennlp/issues/2691

Most of the tests are passing locally. I'll merge this after they all pass and after I evaluate the trained models with the new code and get the same numbers. Also, I suppose there are a bunch of things that can now be deleted (type declarations, SEMPRE related stuff...). I'll add those deletions to this PR as well.


pdasigi:
@matt-gardner This PR is finally ready to be reviewed!

allennlp/predictors/wikitables_parser.py
```diff
@@ -59,12 +22,9 @@ def _json_to_instance(self, json_dict: JsonDict) -> Instance:
         question_text = json_dict["question"]
         table_rows = json_dict["table"].split('\n')
 
-        # pylint: disable=protected-access
-        tokenized_question = self._dataset_reader._tokenizer.tokenize(question_text.lower())  # type: ignore
         # pylint: enable=protected-access
         instance = self._dataset_reader.text_to_instance(question_text,  # type: ignore
-                                                         table_rows,
-                                                         tokenized_question=tokenized_question)
+                                                         table_rows)
```
matt-gardner: Does this need additional processing?  Does this have to be formatted as a corenlp-tagged table?

allennlp/models/semantic_parsing/wikitables/wikitables_erm_semantic_parser.py
```diff
@@ -229,14 +214,9 @@ def forward(self,  # type: ignore
             ``ProductionRule`` using a ``ProductionRuleField``.  We will embed all of these
             and use the embeddings to determine which action to take at each timestep in the
             decoder.
-        agenda : ``torch.LongTensor``
```
matt-gardner: I don't think you meant to remove this one.

matt-gardner: Also, up above, where I can't comment, you need to change the `WikiTablesWorld` type in the docstring (searching for that name in the whole repo would probably be a good idea).

allennlp/models/semantic_parsing/wikitables/wikitables_erm_semantic_parser.py
```diff
@@ -390,9 +371,14 @@ def _get_state_cost(self, worlds: List[WikiTablesWorld], state: CoverageState) -
         action_history = state.action_history[0]
         batch_index = state.batch_indices[0]
         action_strings = [state.possible_actions[batch_index][i][0] for i in action_history]
-        logical_form = world.get_logical_form(action_strings)
-        lisp_string = state.extras[batch_index]
-        if self._executor.evaluate_logical_form(logical_form, lisp_string):
+        logical_form = world.action_sequence_to_logical_form(action_strings)
```
matt-gardner: We can execute action sequences directly, which is definitely recommended here, so you don't have to go back and forth between strings.  It'll make things at least a little faster.

allennlp/models/semantic_parsing/wikitables/wikitables_semantic_parser.py
```diff
@@ -643,31 +635,54 @@ def _compute_validation_outputs(self,
...
-                outputs['best_action_sequence'].append(action_strings)
-                outputs['logical_form'].append(logical_form)
+                all_action_indices = [best_final_states[i][j].action_history[0] for j in
+                                      range(len(best_final_states[i]))]
+                found_denotation = False
+                for action_indices in all_action_indices:
+                    action_strings = [action_mapping[(i, action_index)] for action_index in action_indices]
+                    has_logical_form = False
+                    try:
+                        logical_form = world[i].action_sequence_to_logical_form(action_strings)
```
matt-gardner: Here, too, just execute the action sequence directly, instead of converting to a logical form first.  If we need to add a `world.evaluate_action_sequence(actions, targets)`, that's fine.

It'll simplify the logic below, also, as you don't need to worry about a `ParsingError`.

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -163,17 +163,14 @@ def get_table_knowledge_graph(self) -> KnowledgeGraph:
             self._table_knowledge_graph = KnowledgeGraph(entities, dict(neighbors), entity_text)
         return self._table_knowledge_graph
 
+    # TODO (pradeep): Make a ``read_from_json`` method similar to what we had in ``TableQuestionKnowledgeGraph``
```
matt-gardner: Do we need that for a demo?

allennlp/semparse/contexts/table_question_context.py
```diff
@@ -213,6 +210,88 @@ def read_from_lines(cls,
...
             last_row_index = row_index
+
+        return table_data, column_name_type_mapping
+
+    @classmethod
+    def get_table_data_from_untagged_lines(cls,
+                                           lines: List[List[str]]) -> Tuple[List[Dict[str, Dict[str, str]]],
+                                                                            Dict[str, Set[str]]]:
+        """
+        This method will be called only when we do not have tagged information from CoreNLP. That is, when we are
```
matt-gardner: Is it possible to just call corenlp's python API here instead of this?  Or would that require a call to java, which we're trying to avoid?  It probably would...

allennlp/semparse/domain_languages/wikitables_language.py
```diff
@@ -320,6 +320,21 @@ def get_agenda(self,
                 agenda.append(f"Number -> {number}")
         return agenda
 
+    @staticmethod
+    def is_instance_specific_entity(entity_name: str) -> bool:
```
matt-gardner: Hmm, seems like we should have a function for this on the base class, but maybe it doesn't fit...  I'll think about this some more.

allennlp/data/dataset_readers/semantic_parsing/wikitables/wikitables.py
```diff
@@ -32,28 +32,19 @@
 class WikiTablesDatasetReader(DatasetReader):
     """
     This ``DatasetReader`` takes WikiTableQuestions ``*.examples`` files and converts them into
-    ``Instances`` suitable for use with the ``WikiTablesSemanticParser``.  This reader also accepts
-    pre-processed JSONL files produced by ``scripts/preprocess_wikitables.py``.  Processing the
```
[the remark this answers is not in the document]
pdasigi: I was not aware of the caching functionality. That sounds great! I removed this because I thought reading in the preprocessed files was not too slow.

allennlp/predictors/wikitables_parser.py
pdasigi: It is just the raw table content. I added a comment here and updated the docstring of `text_to_instance` to make that clear.

allennlp/models/semantic_parsing/wikitables/wikitables_erm_semantic_parser.py
pdasigi: Done. Thanks!

allennlp/models/semantic_parsing/wikitables/wikitables_erm_semantic_parser.py
pdasigi: Done. Thanks! Added a  method`WikiTablesLanguage.evaluate_action_sequence`

allennlp/models/semantic_parsing/wikitables/wikitables_semantic_parser.py
pdasigi: But we're also measuring whether the parser is producing valid logical forms (`has_logical_form`), for which we need to make the logical forms. The value of that metric is usually `1.0`, but its nice to have that additional check.

allennlp/semparse/contexts/table_question_context.py
pdasigi: I don't think so. Removing that TODO.

allennlp/semparse/contexts/table_question_context.py
pdasigi: I don't think we're missing out on a lot by not using CoreNLP here.

allennlp/models/semantic_parsing/wikitables/wikitables_semantic_parser.py
matt-gardner: That should already be guaranteed by construction - we only allow actions that produce valid logical forms.  I don't think it's worth slowing down training to test something that should have been tested in a unit test.

matt-gardner: We'll also get `ExecutionErrors` if something was up with the action sequence, which is another place you could catch this.  This would catch two kinds of issues, which are the only ones that we should run into, anyway: (1) cases where we were loose in our type system, but have stricter execution requirements, like around lists vs. single items; and (2) cases where the action sequence didn't terminate, but we're trying to evaluate it anyway (keeping unfinished sequences at the end of beam search).

pdasigi: I was referring to action sequences that got terminated early hence producing incomplete logical forms. Do you have an estimate of how much slow down this could cause?

matt-gardner: No, but anything you do on the CPU is time that's not utilizing the GPU for training.  And those action sequences will be caught with execution errors anyway, so there's no reason to go to a logical form.

matt-gardner: Or you could just instruct the beam search to not include unfinished sequences, which is probably the right thing to do, anyway...

pdasigi: Okay, removing the `has_logical_form` metric.

pdasigi: Hmm.. one more thing. We do need to show the logical form in the demo, and to be able to convert the action sequence into a logical form, we need the corresponding world objects. I think this is the best place to make a logical form and store it in the output dict. Also, this method is only called during validation, so should not significantly slow down training.

matt-gardner: Hmm, yeah, fair enough, that's a good point.  You could pass the world object in the output dictionary, and create the logical form in `decode`, but it's probably fine to just put it here during validation.

schmmd:
@pdasigi would like to train a model with this code before merging this PR.

pdasigi:
I spent a lot of time trying to reproduce results from the best model we got using the code in [iterative-search-semparse](https://github.com/allenai/iterative-search-semparse) (reported in [this paper](https://www.aclweb.org/anthology/N19-1273)). I made some progress in getting there, but it's not fully done. I think the biggest discrepancies still remaining are:

1. In the iterative-search repo, we [add functions to the grammar depending on the content in the table](https://github.com/allenai/iterative-search-semparse/blob/master/weak_supervision/semparse/worlds/wikitables_variable_free_world.py#L65-L88), while we do not do that with the current domain language inference logic.

2. Generic types (like `ComparableColumn`) are handled differently here while inducing a grammar. In the current logic in this repo, we add terminals of multiple types, whereas there I had hierarchical types. This is probably not a huge difference.

I may find other minor differences after I fix these two issues.

That said, I think reproducing the result is beyond the scope of this PR. Given that this PR is already huge, and it has important changes (like removing jdk dependency), I've decided to merge this now and open another PR with changes that ensure we get the same result. Any objections @matt-gardner?

matt-gardner:
Is the result close?

pdasigi:
@matt-gardner Update: I trained an MML model, followed by an ERM model initialized from it using this code. The dev accuracies are both only 1 point behind those from comparable models using the old code. I think that's close enough, and I'm going to merge this PR (after resolving conflicts). I'll spend some time running experiments to try out the things we discussed offline yesterday, and if they give better results, I'll open a new PR.

PR #2765, opened 2019-04-27
fix batched_index_select

JianLiu91:
There is a bug in the code of `allennlp.nn.util.batched_index_select`. For example:
```
target = torch.rand(3, 4, 5) # (batch_size=3, sequence_length=4, embedding_size=5)
indices = torch.LongTensor([
    [0, 1],
    [1, 7], # Any elem in indices should not exceed sequence_length. (I.e., 7 > 4 should raise an error).
    [2, -1] # Any elem in indices should not be less than zero.
])

selected = batched_index_select(target, indices)
print(selected) 

# However, as shown in the following, the current code still outputs a value. 
# The reason is that the subroutine `flatten_and_batch_shift_indices` computes a wrong shifted index.
```


allennlp/nn/util.py
```diff
@@ -983,6 +983,8 @@ def flatten_and_batch_shift_indices(indices: torch.Tensor,
     offset_indices : ``torch.LongTensor``
     """
     # Shape: (batch_size)
+    if torch.max(indices) >= sequence_length or torch.min(indices) < 0:
+        raise ConfigurationError("Any elem in indices should be in a range of [0, " + str(sequence_length-1) + "].")
```
DeNeutoy: You can write this using python 3.6 f-string formatting like this:
```
ConfigurationError(f"All elements in indices should be in range (0, {sequence_length - 1})")
```

JianLiu91: OK! Have changed to this format.

JianLiu91:
OK! Have changed the string format to fit Python3.6, and added the unit test.

PR #2766, opened 2019-04-27
make mask arg optional

Shuailong:
The implementation of the module implies that the `mask` parameter is optional, however it is not default to `none`.  It makes the `mask` compulsory, which is not intended.


matt-gardner:
This actually is intended.  If you're not thinking about a mask when you're encoding a sequence, you are probably doing things wrong.  We want to encourage you to not have bugs, by making this a required argument.  If you're sure you don't need the mask for whatever reason, we handle this case by letting you pass `mask=None`.

PR #2775, opened 2019-04-29
Update auc.py

guoquan:
To allow, in some case, there is no positive or negative sample in a batch.


guoquan:
> I don't think this check does what you think it does
> 
> ```
> In [16]: [1] <= [0, 1]                                                         
> Out[16]: False
> ```

Oh sorry for that. I meant to have `set([1]) <= set([0, 1])`.

guoquan:
Hi, sorry for disturbing but how can I improve the PR now?

allennlp/training/metrics/auc.py
```diff
@@ -52,7 +52,7 @@ def __call__(self,
             raise ConfigurationError("AUC can be used for binary tasks only. gold_labels has {} unique labels, "
                                      "expected at maximum 2.".format(unique_gold_labels.numel()))
 
-        gold_labels_is_binary = list(torch.sort(unique_gold_labels)[0].numpy()) == [0, 1]
```
joelgrus: I think this would be cleaner as

```
set(unique_gold_labels.tolist()) <= {0, 1}
```

joelgrus: sorry, for the delay, this slipped through the cracks

guoquan: I thought I forgot to push updated codes.
Sorry, why there is index [0] in the original code?

joelgrus: `torch.sort` returns a tuple (sorted_values, sorted_indices), but the original code only needed the value.

if we're using `set` we don't need to sort at all

guoquan: Sorry, I forget the `sort`. Was thinking about `torch.unique`. I updated with `set` and leave `torch.unique` there since reducing with CUDA and interfacing with only a small tuple seems more efficient.

guoquan:
Hi, is this resolved, until 839503a?

joelgrus:
there you go, sorry it slipped it through the cracks

PR #2777, opened 2019-04-30
Make FBetaMeasure work with batch size of 1.

brendan-ai2:
- A stray squeeze caused batches of size one to be zero dimensional. Indexing into them was thus invalid and crashed.
- See https://github.com/allenai/allennlp/pull/2275#issuecomment-487486905


brendan-ai2:
Appreciate it!

Issue #2772, opened 2019-04-29
naqanet_dua_2019 pretrained model crashes on master


Phirefly9:
**Describe the bug**
Attempting to use the naqanet_dua_2019 model results in a crash on the master branch

**To Reproduce**
run the following code
```
from allennlp.pretrained import naqanet_dua_2019
predictor = naqanet_dua_2019()
```

**Expected behavior**
It should not crash

**System (please complete the following information):**
 - OS: [Linux] ubuntu 16.04
 - Python version: [ 3.7.3 ]
 - AllenNLP version: [ "I installed from master"]
 - PyTorch version: (came with pip install of allennlp)

**Additional context**
the error message is
```
RuntimeError: Error(s) in loading state_dict for NumericallyAugmentedQaNet:
	Missing key(s) in state_dict:
```

followed by what appears to be pretty much the entire network description

nelson-liu:
It'd be helpful if you could include the whole trace, particularly the bit right after `Missing key(s) in state_dict:` so we can see which keys are lacking.

Phirefly9:
```
untimeError: Error(s) in loading state_dict for NumericallyAugmentedQaNet:
[...]

PR #2692, opened 2019-04-04
QaNetEncoder Multi-GPU

kl2806:
Fixes #2605, allows `QaNetEncoder` for multiple GPUs.

PR #2780, opened 2019-04-30
Change lists to ModuleList to solve parallelism issue

### LINKED ISSUE #2776: "arguments are located on different GPUs" when training stacked_self_attention as encoder
Please first search our GitHub repository for similar questions.  If you don't find a similar example you can use the following template:

 - OS:  Linux
 - Python version: 3.6.5
 - AllenNLP version: "I installed from master"
 - PyTorch version: Pytorch 1.01

**Question**

I apply the ``stacked_self_attention`` module as the Seq2SeqEncoder over a sequence of embeddings and encounter the error "arguments are located on different GPUs". If I switch it to ``multi_head_attention`` or ``lstm``, then everything is fine.

Here is the whole trace:
`Traceback (most recent call last):
  File "train.py", line 156, in <module>
    output_dict = model(**batch)
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/torch/nn/modules/module.py", line 489, in __call__
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/torch/nn/parallel/data_parallel.py", line 143, in forward
    outputs = self.parallel_apply(replicas, inputs, kwargs)
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/torch/nn/parallel/data_parallel.py", line 153, in parallel_apply
    return parallel_apply(replicas, inputs, kwargs, self.device_ids[:len(replicas)])
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/torch/nn/parallel/parallel_apply.py", line 83, in parallel_apply
    raise output
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/torch/nn/parallel/parallel_apply.py", line 59, in _worker
    output = module(*input, **kwargs)
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/torch/nn/modules/module.py", line 489, in __call__
  File "/raid/caojize/microsoft_r2c/models/transformer/model.py", line 135, in forward
    qa_rep = self.span_encoder(qa_rep, qa_mask)
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/torch/nn/modules/module.py", line 489, in __call__
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/allennlp/modules/time_distributed.py", line 35, in forward
    reshaped_outputs = self._module(*reshaped_inputs)
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/torch/nn/modules/module.py", line 489, in __call__
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/allennlp/modules/seq2seq_encoders/stacked_self_attention.py", line 143, in forward
    feedforward_output = feedforward(output)
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/torch/nn/modules/module.py", line 489, in __call__
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/allennlp/modules/feedforward.py", line 80, in forward
    output = dropout(activation(layer(output)))
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/torch/nn/modules/module.py", line 489, in __call__
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/torch/nn/modules/linear.py", line 67, in forward
    return F.linear(input, self.weight, self.bias)
  File "/home/caojize/anaconda3/envs/r2c/lib/python3.6/site-packages/torch/nn/functional.py", line 1354, in linear
    output = input.matmul(weight.t())
RuntimeError: arguments are located on different GPUs at /opt/conda/conda-bld/pytorch_1549628766161/work/aten/src/THC/generic/THCTensorMathBlas.cu:253`

JizeCao:
The original ``stacked_self_attention`` cannot run across multiple GPUS through Data Parallel. Change all the lists to ModuleList to resolve this issue.

Fixes #2776.

### DIFF
--- a/allennlp/modules/seq2seq_encoders/stacked_self_attention.py
+++ b/allennlp/modules/seq2seq_encoders/stacked_self_attention.py
@@ -1,5 +1,3 @@
-from typing import List
-
 from overrides import overrides
 import torch
 from torch.nn import Dropout
@@ -73,13 +71,13 @@ class StackedSelfAttentionEncoder(Seq2SeqEncoder):
         super(StackedSelfAttentionEncoder, self).__init__()
 
         self._use_positional_encoding = use_positional_encoding
-        self._attention_layers = ModuleList([]) 
+        self._attention_layers = ModuleList([])
         self._feedfoward_layers = ModuleList([])
         self._layer_norm_layers = ModuleList([])
         self._feed_forward_layer_norm_layers = ModuleList([])
 
         feedfoward_input_dim = input_dim
-        for i in range(num_layers):
+        for _ in range(num_layers):
             feedfoward = FeedForward(feedfoward_input_dim,
                                      activations=[Activation.by_name('relu')(),
                                                   Activation.by_name('linear')()],
--- a/allennlp/modules/seq2seq_encoders/stacked_self_attention.py
+++ b/allennlp/modules/seq2seq_encoders/stacked_self_attention.py
@@ -3,6 +3,7 @@ from typing import List
 from overrides import overrides
 import torch
 from torch.nn import Dropout
+from torch.nn import ModuleList
 
 from allennlp.modules.feedforward import FeedForward
 from allennlp.modules.layer_norm import LayerNorm
@@ -72,10 +73,10 @@ class StackedSelfAttentionEncoder(Seq2SeqEncoder):
         super(StackedSelfAttentionEncoder, self).__init__()
 
         self._use_positional_encoding = use_positional_encoding
-        self._attention_layers: List[MultiHeadSelfAttention] = []
-        self._feedfoward_layers: List[FeedForward] = []
-        self._layer_norm_layers: List[LayerNorm] = []
-        self._feed_forward_layer_norm_layers: List[LayerNorm] = []
+        self._attention_layers = ModuleList([]) 
+        self._feedfoward_layers = ModuleList([])
+        self._layer_norm_layers = ModuleList([])
+        self._feed_forward_layer_norm_layers = ModuleList([])
 
         feedfoward_input_dim = input_dim
         for i in range(num_layers):
--- a/allennlp/modules/seq2seq_encoders/stacked_self_attention.py
+++ b/allennlp/modules/seq2seq_encoders/stacked_self_attention.py
@@ -87,11 +87,9 @@ class StackedSelfAttentionEncoder(Seq2SeqEncoder):
                                      num_layers=2,
                                      dropout=dropout_prob)
 
-            self.add_module(f"feedforward_{i}", feedfoward)
             self._feedfoward_layers.append(feedfoward)
 
             feedforward_layer_norm = LayerNorm(feedfoward.get_output_dim())
-            self.add_module(f"feedforward_layer_norm_{i}", feedforward_layer_norm)
             self._feed_forward_layer_norm_layers.append(feedforward_layer_norm)
 
             self_attention = MultiHeadSelfAttention(num_heads=num_attention_heads,
@@ -99,11 +97,9 @@ class StackedSelfAttentionEncoder(Seq2SeqEncoder):
                                                     attention_dim=projection_dim,
                                                     values_dim=projection_dim,
                                                     attention_dropout_prob=attention_dropout_prob)
-            self.add_module(f"self_attention_{i}", self_attention)
             self._attention_layers.append(self_attention)
 
             layer_norm = LayerNorm(self_attention.get_output_dim())
-            self.add_module(f"layer_norm_{i}", layer_norm)
             self._layer_norm_layers.append(layer_norm)
 
             feedfoward_input_dim = hidden_dim
--- a/allennlp/tests/models/language_model_test.py
+++ b/allennlp/tests/models/language_model_test.py
@@ -77,8 +77,8 @@ class TestUnidirectionalLanguageModelTransformer(TestUnidirectionalLanguageModel
         # they are not used.
         self.ensure_model_can_train_save_and_load(
                 self.param_file, gradients_to_ignore={
-                        "_contextualizer._feedforward_layer_norm_layers.0.gamma",
-                        "_contextualizer._feedforward_layer_norm_layers.0.beta"})
+                        "_contextualizer._feed_forward_layer_norm_layers.0.gamma",
+                        "_contextualizer._feed_forward_layer_norm_layers.0.beta"})
 
 class TestBidirectionalLanguageModel(TestUnidirectionalLanguageModel):
     def setUp(self):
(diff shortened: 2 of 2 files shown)


matt-gardner:
@brendan-ai2, @DeNeutoy, what do you think of this PR?  The issue is that this changes the parameter names inside the module, so any saved model that uses this has to modify its weights file to have the updated names, or it will crash (see #2772).  On the other hand, the code is clearly broken right now for multiple GPUs (see #2692), so something is broken either way.  I don't think there's a reasonable way to handle backwards compatibility in the code, but there is a way to modify saved models to be compatible with this change.  My proposal is to merge this and leave a note about breaking saved models that use this module in the release notes.  What do you think?

allennlp/modules/seq2seq_encoders/stacked_self_attention.py
```diff
@@ -72,37 +71,33 @@ def __init__(self,
...
-        for i in range(num_layers):
+        for _ in range(num_layers):
             feedfoward = FeedForward(feedfoward_input_dim,
                                      activations=[Activation.by_name('relu')(),
                                                   Activation.by_name('linear')()],
                                      hidden_dims=[feedforward_hidden_dim, hidden_dim],
                                      num_layers=2,
                                      dropout=dropout_prob)
 
-            self.add_module(f"feedforward_{i}", feedfoward)
```
brendan-ai2: @matt-gardner, Ah, I had forgotten that `ModuleList` would change the names. Somewhat surprising that `add_module` isn't sufficient. Unfortunate. However, I think we can work around this and preserve backwards compatibility by using, e.g.:

`setattr(self, f"feedforward_{i}", feedforward)`

@JizeCao, could you please give that a try?

matt-gardner: Pytorch does crazy stuff with `setattr`; I don't think this is a good idea.  It would make interactions with pytorch magic very hard to reason about.

brendan-ai2: @matt-gardner , I'm not sure I understand your concern. We use Pytorch's `__setattr__` every single time we modify any instance variable of a `Module`. The only difference here is that the name is dynamic. Example:

```
In [1]: class Foo:
   ...:     def __setattr__(self, name, value):
   ...:         print("FOO")
   ...:         object.__setattr__(self, name, value)
   ...:

In [2]: f = Foo()

In [3]: f.x = 7
FOO
```

brendan-ai2:
@JizeCao, would it be possible to make a simple test for this? You can for whether multiple GPUs are available like this: https://github.com/allenai/allennlp/blob/master/allennlp/tests/training/trainer_test.py#L114 Thank you!

allennlp/modules/seq2seq_encoders/stacked_self_attention.py
matt-gardner: Are you suggesting adding a call to `setattr` in addition to appending to the module list?  That's what I was expecting.  Having two calls will definitely break the pytorch module, because pytorch will list the parameters twice (it registers stuff at every call to `setattr`).  I think this is a case where we should just break backwards compatibility, because there's a workaround and because it's fixing a bug.

brendan-ai2: No, I'm suggesting using `setattr(self, f"feedforward_{i}", feedforward)` alone. This is very similar to the original code. Its dispatches to the exact same `__setattr__` that any assignment to `self` does. Preserving backwards compatibility while fixing the bug seems quite simple here.

matt-gardner: I'd be very surprised if that actually worked, because you'd expect pytorch to have the same logic there as it does when you call `add_module`, which we know doesn't work.  But, having a test would tell us one way or the other.

DeNeutoy: another solution would be to override `load_state_dict` in the following way:

```python

@overrides
def load_state_dict(self, state_dict, strict=True):
    new_state = {}
    for name, param in state_dict.items():
        # Depreciation warning here
        if name == ....
            name = new_name
        new_state[name] = param
    super().load_state_dict(new_state, strict=strict)

```

I think that would work?

matt-gardner: I dug around in pytorch's [load_state_dict](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/module.py) code, I don't think it's recursively implemented.  It was hard to tell, so I could be wrong, but I don't _think_ this would actually work.

DeNeutoy: it is recursively implemented:
https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/module.py#L759

brendan-ai2:
@JizeCao , @matt-gardner I am unable to repro this bug. Using `add_module` appears sufficient. Please see https://github.com/allenai/allennlp/pull/2784. @JizeCao , how did you obtain the trace you gave?

JizeCao:
I use pytorch DataParallel Wrapper over the model to parallelize it. I didn't use the json file to control the used cuda devices. 
`model = DataParallel(model).cuda()`
By doing so, other encoder structures, i.e: ``multi_heads_attention``, ``lstm``, can run parallelly across GPUs, whereas ``stacked_self_attention`` fails to do it without this PR.

brendan-ai2:
Thanks for the help, @JizeCao. I updated https://github.com/allenai/allennlp/pull/2784 and was able to repro the bug. @matt-gardner, using `setattr` along with `getattr` fixes the bug.

I'm currently struggling to understand why this bug doesn't manifest with AllenNLP's hand-rolled parallelism and `add_module`. It appears we dispatch to the same `torch/nn/parallel/replicate.py` as `DataParallel` to move the model. Ideas anyone?

brendan-ai2:
Actually, we don't even need `setattr`. We just need to read the attributes that `add_module` created in the first place. Updated https://github.com/allenai/allennlp/pull/2784.

brendan-ai2:
(Ugh, I couldn't repro earlier because the batch size was larger than the test dataset meaning we never had enough data for the second GPU. This bug definitely affects `trainer.py` too.)

matt-gardner:
Thanks for figuring this out @brendan-ai2, and thanks for the original PR @JizeCao.  I'm closing this in favor of #2784.

allennlp/modules/seq2seq_encoders/stacked_self_attention.py
matt-gardner: Except that's recursing on this private method, _not_ `load_state_dict()` - I don't think it would do anything unless you overrode the private method.  That's what I was confused about.  But, yeah, if you do it in the private method it'd probably work.

PR #2787, opened 2019-05-01
Bert for classification

joelgrus:
this PR contains (in essence) two different ways to accomplish classification using BERT (which you can see from the two test fixtures).

one is a full `Model`, which runs BERT, grabs the pooled output, and tacks on a linear layer. as mentioned in the docstring, there's something not-allennlp-ish about this design; nonetheless, I suspect that people will find it extremely useful.

another is a "Bert Pooler" `Seq2Vec` module, which when combined with the Bert Token Embedder reproduces the pooled output. This allows e.g the use of the `BasicClassifier` model, as well as (presumably) some hypothetical more-complex classifier. this means that it gives an easier way to experiment with swapping out the "pooled BERT" representation with other representations, which is good for science (but less good if you just want to train a BERT classifier)


joelgrus:
stupid codecov, the coverage is "really" like 100%, but it's "implicit" coverage via configuration files

allennlp/models/bert_for_classification.py
```diff
@@ -0,0 +1,141 @@
...
+                 bert_model: Union[str, BertModel],
+                 num_labels: int = None,
+                 index: str = "bert",
+                 label_namespace: str = "labels",
+                 trainable: bool = True,
+                 initializer: InitializerApplicator = InitializerApplicator()) -> None:
+        super().__init__(vocab)
+
+        if isinstance(bert_model, str):
+            self.bert_model = BertModel.from_pretrained(bert_model)
```
matt-gardner: Unrelated to this PR, but this made me worried about coercing a json dict to a str in the `from_params` code.  Looks like we don't coerce, so this should be fine, but we also don't do explicit type checking, which it might be worth adding.

allennlp/modules/seq2vec_encoders/bert_pooler.py
```diff
@@ -0,0 +1,50 @@
...
+    ----------
+    pretrained_model: ``Union[str, BertModel]``
+        The pretrained BERT model to use. If this is a string,
+        we will call ``BertModel.from_pretrained(pretrained_model)``
+        and use that.
+    """
+    def __init__(self, pretrained_model: Union[str, BertModel]) -> None:
+        super().__init__()
+
+        # TODO(joelgrus): it's inefficient to load the model here and (presumably) also in the
```
matt-gardner: You could take the approach we use for spacy models, where we basically have a static dictionary of loaded models.  This could get a little hairy if you're fine tuning (lots of potential crazy bugs if you use it wrong), but if it's private and the only use for it is sharing the token embedder and the pooler, it might be ok, because we'll do the right thing.

allennlp/models/bert_for_classification.py
```diff
@@ -0,0 +1,141 @@
...
+from pytorch_pretrained_bert.modeling import BertModel
+
+from allennlp.data.vocabulary import Vocabulary
+from allennlp.models.model import Model
+from allennlp.nn.initializers import InitializerApplicator
+from allennlp.training.metrics import CategoricalAccuracy
+
+
+@Model.register("bert_for_classification")
+class BertForClassification(Model):
```
matt-gardner: So, when we talked about this earlier, I was imagining a `Module`, not a `Model`.  But, this is probably better, because it's simpler and still gives someone an easy way to do this with allennlp if they want to.  It's just an example model that does classification with BERT, in a BERT-specific way.

allennlp/tests/fixtures/bert/bert_pooler.jsonnet
```diff
@@ -0,0 +1,53 @@
...
+    "model": {
+        "type": "basic_classifier",
+        "text_field_embedder": {
+            "allow_unmatched_keys": true,
+            "embedder_to_indexer_map": {
+                "bert": ["bert", "bert-offsets"],
+            },
+            "token_embedders": {
+                "bert": {
+                    "type": "bert-pretrained",
```
matt-peters: It's worthwhile to include the options to configure training in the same manner as `bert_for_classification.jsonnet`.  To fine tune from the top layer it's necessary to also add

```
"requires_grad": false,
"top_layer_only": true,
```

allennlp/models/bert_for_classification.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    An AllenNLP Model that runs pretrained BERT,
+    takes the pooled output, and adds a Linear layer on top.
+    If you want an easy way to use BERT for classification, this is it.
+    Note that this is a somewhat non-AllenNLP-ish model architecture,
+    in the sense that it goes straight from indexed tokens to pooled output
```
matt-gardner: I would say _why_ it goes against our philosophy, not _how_ it goes against it.  That is, "it's not configurable", not, "it doesn't use these abstractions".

allennlp/models/bert_for_classification.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        input_ids = tokens[self._index]
+        token_type_ids = tokens[f"{self._index}-type-ids"]
+        input_mask = (input_ids != 0).long()
+
+        _, pooled = self.bert_model(input_ids=input_ids,
+                                    token_type_ids=token_type_ids,
+                                    attention_mask=input_mask)
+
+        # apply classification layer
```
OyvindTafjord: It's customary to add dropout here, not sure how important that is: https://github.com/huggingface/pytorch-pretrained-BERT/blob/3ae8c8be1e3fc770968cd3fdb3b643e0b166e540/pytorch_pretrained_bert/modeling.py#L1058

allennlp/tests/fixtures/bert/bert_pooler.jsonnet
joelgrus: :+1:

allennlp/models/bert_for_classification.py
joelgrus: done, thanks

allennlp/modules/seq2vec_encoders/bert_pooler.py
joelgrus: done

allennlp/tests/models/bert_for_classification_test.py
```diff
@@ -0,0 +1,86 @@
...
+                 tokenizer: WordTokenizer = None) -> None:
+        super().__init__(lazy)
+        self._token_indexers = token_indexers or {}
+        self.tokenizer = tokenizer or WordTokenizer()
+
+    def _read(self, file_path: str):
+        #            2   3    4   3     5     6   8      9    2   14   12
+        sentence1 = "the quickest quick brown fox jumped over the lazy dog"
+        tokens1 = self.tokenizer.tokenize(sentence1)
+        label1 = "positive"
```
OyvindTafjord: Btw, realistic use of this model should add a `[CLS]` token at the start of the token list (and `[SEP]` at the end), not sure to what extent a test like this should also serve as an example of proper usage.

allennlp/tests/models/bert_for_classification_test.py
```diff
@@ -0,0 +1,86 @@
...
+                 lazy: bool = False,
+                 token_indexers: Dict[str, TokenIndexer] = None,
+                 tokenizer: WordTokenizer = None) -> None:
+        super().__init__(lazy)
+        self._token_indexers = token_indexers or {}
+        self.tokenizer = tokenizer or WordTokenizer()
+
+    def _read(self, file_path: str):
+        #            16    2   3    4   3     5     6   8      9    2   14   12  17
+        sentence1 = "[CLS] the quickest quick brown fox jumped over the lazy dog [SEP]"
```
OyvindTafjord: I don't think this will work as intended as the word splitter will mangle "[CLS]" here, I think it needs to be added after tokenization.

allennlp/tests/models/bert_for_classification_test.py
joelgrus: actually, the PretrainedBertIndexer adds [CLS] and [SEP] automatically

allennlp/tests/models/bert_for_classification_test.py
joelgrus: I reverted this change, see above (but I tested it and it didn't mangle, I'm pretty sure the Bert-specific word splitter handles these cases correctly)

allennlp/tests/models/bert_for_classification_test.py
joelgrus: see https://github.com/allenai/allennlp/blob/master/allennlp/tests/data/token_indexers/bert_indexer_test.py#L24

OyvindTafjord: Ah yes, I see now it has default start-tokens/end-tokens when instantiated from PretrainedBertIndexer.

joelgrus:
I think I addressed everyone's feedback, can someone review review this? thanks

allennlp/models/bert_for_classification.py
```diff
@@ -0,0 +1,153 @@
...
+
+        in_features = self.bert_model.config.hidden_size
+
+        if num_labels:
+            out_features = num_labels
+        else:
+            out_features = vocab.get_vocab_size(label_namespace)
+
+        if dropout is not None:
+            self._dropout = torch.nn.Dropout(p=dropout)
```
matt-gardner: I'm pretty sure we don't need to do this pattern anymore; I'm not sure why we did it originally, but if you set the default dropout value to `0.0` instead of `None`, you don't need these conditionals (here or in `forward`), and @yizhongw found there was no memory or speed hit for doing so.

PR #2796, opened 2019-05-02
fixing regression in drop script and adding a test case for the same

### LINKED ISSUE #2795: indentation of Line210~213 in drop_eval.py is incorrect
It seems to only append the em/f1 of the last question of every passage to instance_exact_match/instance_f1 and then compute the average as the final metrics.
https://github.com/allenai/allennlp/blob/master/allennlp/tools/drop_eval.py#L210

dDua:
Fixes #2795


matt-gardner:
Though, I'm guessing that pylint is going to complain about your spacing...

PR #2799, opened 2019-05-04
Adding NamedTuple case for Tensors to be moved to GPU.

### LINKED ISSUE #2792: Problem with Text2sql parser model (with grammar_based_text2sql dataset reader)
**Describe the bug**
I have a json config file that runs the text2sql_parser model with the grammar_based_text2sql dataset reader (see config below). It runs fine on CPU, but whenever I am running it on GPU I get the error: 

  File "/anaconda3/envs/myENV/lib/python3.6/site-packages/allennlp/models/semantic_parsing/text2sql_parser.py", line 327, in _create_grammar_state
    if action.rule == "":
AttributeError: 'tuple' object has no attribute 'rule'

What could be the problem? This error only happens on GPU. I couldn't find any example json experiment file for this model, so it could also be that my json file is the problem.

**To Reproduce**
Launch the following json experiment on GPU, using the GeoQuery data from `"Improving Text to SQL Evaluation Methodology" <https://arxiv.org/abs/1806.09029>`:

{
  "dataset_reader": {
    "type": "grammar_based_text2sql",
    "schema_path": "data/geography-schema.csv",
    "token_indexers": {
      "tokens": {
        "type": "single_id" }}
  },
  "train_data_path": "train.json",
  "validation_data_path": "dev.json",
  "model": {
    "type": "text2sql_parser",
    "utterance_embedder": {
      "tokens": {
        "type": "embedding",
        "embedding_dim": 400,
        "trainable": true
      }   },
    "action_embedding_dim": 200,
    "encoder": {
      "type": "lstm",
      "input_size": 400,
      "hidden_size": 800,
      "bidirectional": true,
      "num_layers": 1
    },
    "decoder_beam_search": {
    "beam_size": 10
    },
    "max_decoding_steps": 200,
    "input_attention": {"type": "dot_product"},
    "dropout": 0.5
  },
  "iterator": {
    "type": "basic",
    "batch_size" : 32
  },
  "trainer": {
    "num_epochs": 90,
    "patience": 3,
    "cuda_device": 1,
    "validation_metric": "+exact_match",
    "optimizer": {
      "type": "adam",
      "lr": 0.001
    }  } }

**Expected behavior**
The model trains on GPU.

 - OS: Linux
 - Python version: 3.6.1
 - AllenNLP version:  v0.8.3 (not completely sure, could also be v0.8.2)

ValentinaPy:
Fixes #2792 that occurred when converting semantic parsing actions into tensors on GPU. Since in an earlier commit the production rules have been changed to NamedTuples #
([360c3e1#diff-25fbc5b96eaa1ec951718b877078673b])(https://github.com/allenai/allennlp/commit/360c3e10d40a1e9e64085be571eeed4725c53095#diff-25fbc5b96eaa1ec951718b877078673b), a case for NamedTuples had to be added to the move_to_device function.


matt-gardner:
Can you add a test for this to `tests/nn/util_test.py`?  It should look something like this:

```python
def test_move_to_device(self):
    class FakeTensor(torch.Tensor):
        def __init__(self):
            self.device = None
        def cuda(self, device):
            self.device = device

    class A(NamedTuple):
        a: int
        b: FakeTensor

    structured_obj = {'a': [A(1, FakeTensor()), A(2, FakeTensor)], 'b': FakeTensor, 'c': (1, FakeTensor)}
    new_device = 4
    moved_obj = util.move_to_device(structured_obj, new_device)
    assert moved_obj['a'][0].a == 1
    assert moved_obj['a'][0].b.device == new_device
    assert moved_obj['a'][1].b.device == new_device
    assert moved_obj['b'].device == new_device
    assert moved_obj['c'][0] == 1
    assert moved_obj['c'][1].device == new_device
```

ValentinaPy:
While writing the test I realised that you cannot check whether an instance is a NamedTuple and that it only fixed my problem because I accidentally replaced the check for tuple with the check for NamedTuple (in my code), which then resulted in the else statement to be executed (e.g. returning the object as is). The suggested work-around for checking whether an object is a NamedTuple is checking whether it is a tuple and whether it has the attribute '_fields'. I realise that this is quite 'hacky', so if you have any better suggestions I would be happy to implement them. 

I also added a test for the move_to_device function.

matt-gardner:
Thanks for the test, and for fixing the `NamedTuple` check!  The reason I wanted the `FakeTensor` was so that we could run the test on our CI machine, which doesn't use a GPU.  I tried running my version and realized why you changed it, so I fixed it.  Looks good, thanks for doing this!

matt-gardner:
The build passed on CI, not sure why it's not getting reported (maybe because of my force push). I'm going to merge this; I'll fix the build if for some reason this is actually broken. Thanks @ValentinaPy!

ValentinaPy:
Great, thanks @matt-gardner !

PR #2805, opened 2019-05-06
Fix Model.load fails if model_params is str

bryant1410:


allennlp/models/model.py
```diff
@@ -319,7 +319,7 @@ def load(cls,
         """
 
         # Peak at the class of the model.
-        model_type = config["model"]["type"]
+        model_type = config["model"] if isinstance(config["model"], str) else config["model"]["type"]
```
DeNeutoy: In what circumstance is this needed? I think you can just do `{"type": "model_type"}` rather than passing `model_type` as a string.

matt-gardner: I'm pretty sure `from_params` allows it to be a string: https://github.com/allenai/allennlp/blob/aeafe1c58b6041e654c7fb645d6aa98f2fac9b9a/allennlp/common/from_params.py#L224-L227

So (I believe) @bryant1410 is just bringing this other code inline with what `from_params` already accepts.

bryant1410: @matt-gardner that's correct.

By the way, is it accepting a string instead of a dict desired everywhere, or maybe even you want to deprecate it? Because I see that ``Model`` and ``DataIterator`` accept it, but not ``Trainer`` (e.g., when I want to use "no_op").

PR #2817, opened 2019-05-08
callback based trainer

joelgrus:
think of this as a prototype for what things could look like rather than a concrete proposal as to what they should look like.

if you look at the `EventBasedTrainer` 

https://github.com/allenai/allennlp/compare/master...joelgrus:event-based-trainer?expand=1#diff-df40229c8a0a8844cfb56c0599b6e32b

it contains almost nothing, you just give it a model and a way of generating epochs of tensordicts, and it iterates over them and fires events. That's it. You want to do forward / backward / step? there's a `PROCESS_BATCH` event that you can hook that up to. Want to track progress on a validation set? There's a `END_EPOCH` event that you can hook that up to. It's almost certain that there probably needs to be more stuff in this skeletal base trainer, but I'll figure that out as I go.

I took a lot of inspiration from pytorch/ignite, but they had "process batch" as an argument to the trainer, whereas I made it just another callback. rather than passing state around, I just made it so that any state that a callback might need to read from / write to needed to belong to the trainer instance.

to see how this would look in (simple) practice, then look at the test for it:

https://github.com/allenai/allennlp/compare/master...joelgrus:event-based-trainer?expand=1#diff-fed704c8b9888ba5b1a4e22ac0721a82

it defines three callbacks (1) to do the forward / backward / step (2) to evaluate on a dataset (3) to mark that the training is done (the implementation is dumb, but imagine it posts to Slack or something).

my thinking is that

(1) if this is a good design, I would next implement based on this a trainer that has feature-parity with our current trainer but with all the functionality (tensorboard, checkpointing, etc) using callbacks. (The design goal would be for it to pass all of our current Trainer tests, so that there's no difference to the end user.)

(2) at which point, if you want to do e.g. multi-task training, you can just use those same pieces but with a different `PROCESS_BATCH`

(3) if you want to e.g. train with a slack notification, you can use that same trainer but add an extra `END_TRAINING` event

(4) if you want to save out the model.tar.gz every epoch (like a recent issue), you can just configure an extra `END_EPOCH` event to do that

(5) if you want to do gradient accumulation, similarly.

(6) OR, if you want to do something entirely different, implement your own subclass, but then still reuse all of the checkpointing / logging / etc logic that we have (almost) out of the box

Anyway, I would love to get people's feedback on this basic design, especially if you see any gaps where the trainer needs to be able to do something but this will make it really hard.


allennlp/training/event_based_trainer/event_based_trainer.py
```diff
@@ -0,0 +1,87 @@
...
+class EventBasedTrainer(TrainerBase):
+    def __init__(self,
+                 model: Model,
+                 epochs_generator: Iterable[Iterable[TensorDict]],
+                 serialization_dir: str,
+                 cuda_device: Union[int, List] = -1) -> None:
+        super().__init__(serialization_dir, cuda_device)
+        self.model = self._move_to_gpu(model)
+        self.epochs_generator = epochs_generator
+        self.callbacks: Dict[str, List[Callback]] = defaultdict(list)
```
matt-gardner: You should be able to pass a dictionary of callbacks to the constructor, so you can specify all of this `from_params`.

allennlp/tests/training/event_based_trainer_test.py
```diff
@@ -0,0 +1,131 @@
...
+class TrainOneBatch(Callback):
+    """
+    Trains one batch by running
+        optimizer.zero_grad -> model.forward -> loss.backward -> optimizer.step
+    """
+    def __init__(self, optimizer):
+        self.optimizer = optimizer
+
+    def call(self, trainer: EventBasedTrainer) -> None:
+        self.optimizer.zero_grad()
```
matt-gardner: Pretty sure this state is going to have to live in the trainer, not here.  Otherwise the `from_params` stuff doesn't work out very well.  It's possible that the `extras` stuff just works and passes it to here just fine, but then you'd still break existing config files if you want to replace our existing trainer with this.

allennlp/tests/training/event_based_trainer_test.py
```diff
@@ -0,0 +1,131 @@
...
+        loss = trainer.output["loss"]
+        loss.backward()
+        trainer.tqdm.set_description(f"loss: {loss}")
+
+        self.optimizer.step()
+
+
+class Evaluate(Callback):
+    def __init__(self,
+                 epoch_generator: Iterable[TensorDict],
```
matt-gardner: This one, also; getting the data into here `from_params` is probably going to tricky, but quite easy if it just lives in the trainer state.

allennlp/training/event_based_trainer/event_based_trainer.py
```diff
@@ -0,0 +1,87 @@
...
+        for epoch in self.epochs_generator:
+            self.fire_event(Events.EPOCH_START)
+
+            self.tqdm = Tqdm.tqdm(epoch)
+            for batch in self.tqdm:
+                # TODO(move to cuda device)
+
+                self.batch = batch
+                self.fire_event(Events.BATCH_START)
+                self.fire_event(Events.PROCESS_BATCH)
```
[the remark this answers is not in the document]
joelgrus: one obvious alternative is to do it the way pytorch/ignite does it, which is that PROCESS_BATCH is not an event, instead you pass in something like a `Callable[[TensorDict], Any]` to the constructor

allennlp/training/event_based_trainer/event_based_trainer.py
joelgrus: :+1:

allennlp/tests/training/event_based_trainer_test.py
joelgrus: I don't want to replace our existing trainer with this, I want to replace it with a subclass of this (that has a constructor that looks like the current trainer's constructor), and that subclass instance would own the optimizer. My thinking is that maybe in some other trainer variant I might have multiple optimizers, or something weird like that.

allennlp/tests/training/event_based_trainer_test.py
joelgrus: same as above, I don't think you'd ever want to instantiate `EventBasedTrainer` from_params (maybe I should call it `EventBasedTrainerBase`? yuck), I wanted to build the minimal base class that you could easily build any variety of trainers on top of.

allennlp/training/event_based_trainer/event_based_trainer.py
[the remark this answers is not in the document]
matt-gardner: Yeah, but if you go down that path, where do you stop?  And really, to do things `from_params`, you'd want that `Callable` to be `registered` somewhere, and so you add a new type for that callback, and things just get really messy (because how many callback types do you have, etc.).  I don't really like saving the state to `self`, but I haven't seen (and can't come up with) a better alternative.

allennlp/tests/training/event_based_trainer_test.py
matt-gardner: Hmm, yeah, I can see that.  If you require the `Trainer` to own all of that state, you make it hard to implement callbacks that require other kinds of state.  That's a really good point.  You're right here, and I retract what I said.  We just need to be sure that passing arguments through to the callbacks works as it should.

allennlp/tests/training/event_based_trainer_test.py
matt-gardner: No, I think you do want to instantiate this class `FromParams`, because all of the fun stuff is just happening in the callbacks, anyway.  You're right, though, that these some of these things do need to have some particularly special magic to make their way to the right callbacks (particularly the optimizers, which need the model parameters).

allennlp/training/event_based_trainer/event_based_trainer.py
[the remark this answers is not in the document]
bryant1410: An alternative I see is having different functions, instead of dynamic events (which btw I don't see as a great advantage, but please let me know your opinion on this :smile:), like `on_epoch_start`, etc. In this case, every function can receive its own params (e.g., `batch` in `process_batch`).

This is something similar to what Keras does, but they define all functions under the same class (so, `Callback` instead of having `call`, would have: `on_training_start`, `on_epoch_start`, etc.).

I don't know how this would interact with the `from_params` thing, or how do you imagine the `from_params` should be with the implementation as it is now.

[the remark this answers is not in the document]
joelgrus: right, the main reason for doing it this way is that you can do `Callback.from_params`, so that you could configure your whole trainer using config files, something like

```
trainer: {
    callbacks: {
        "TRAINING_START": [
            {"type": "set_up_tensorboard", ... }
        ],
        "EPOCH_END": [
            {"type": "check_validation_metrics", ... }
        ], ...
}
```

but you're right that the tradeoff that way is that it's hard to have a single interface that works for every kind of callback

joelgrus:
ok, this second iteration is very different, at a high level it does the following:

1. callbacks are generic over state
2. here every callback state is a protocol (i.e. structural subtyping) over the fields it expects the calling trainer to have. That is, if it's going to refer to trainer.model, the prototype for its state reflects that. (The version of mypy we're pinned to right now doesn't check these types.)
3. the trainer (see the class called `Trainer2`, which is a callback-based implementation of the current trainer) owns a `CallbackHandler` object, which it fires events to using `handler.fire_event`
4. the events ended up being extremely granular (which I'm not thrilled about) because of the need to do things like "set parameters to moving average -> save checkpoint -> restore parameters from moving average".
5. this all means that all state is mediated through the trainer object, which is not ideal, but it's hard to think of a generic-enough way to avoid this.

--

recommended way of understanding this PR:

1. look at `callback.py`, to see what a callback looks like
2. look at `callback_handler.py`, to see how they're called ("the `State` in `CallbackHandler` is really your trainer subclass)
3. look at `callback_handler_test.py` to see a very simple example of using these
4. look at `trainer2.py` to see what the stripped down trainer looks like (it's much shorter and cleaner)
5. look at `callbacks.py` to see how I shoved all the functionality into callbacks

--

I don't know if people will like this (I don't know if I like it), but I've been banging my head against this for a while, and this was the most sensible design I could come up with. Feedback encouraged!

Evpok:
Why wouldn't you just use ignite?

joelgrus:
I guess at the end of the day, this ended up looking pretty similar to Ignite. The biggest difference is that there's nothing analogous to their `Engine` abstraction, I was never able to come up with a non-awkward way to fit our training loop into that paradigm, and I needed a much more granular event space. (I know they allow custom events, but that makes things even more complicated on their side.)

joelgrus:
for (1) my thinking was to be able to say (in essence) "if you want to use the `Validation` callback, your Trainer must expose these fields / methods". That's literally the only thing the structural subtyping is doing. I agree that it's complicated, but I couldn't think of another way to get mypy to check that your callbacks are compatible with your trainer. If you were implementing your own trainer, you shouldn't have to think about Protocols at all, other than making sure that your trainer has the right fields. But I agree that it looks daunting and that's not great.

for (2) it's possible / likely that I should have moved even more of the training loop into the callbacks and/or that I didn't get the events exactly right. Let me think about this more.

for (3) right now they're impossible to override, I think as a next step you'd inject them and construct them from_params, I was just trying to get this to the point where it had feature parity with the existing trainer

allennlp/training/trainer2.py
```diff
@@ -0,0 +1,673 @@
...
+        train_generator = lazy_groups_of(raw_train_generator, num_gpus)
+        num_training_batches = math.ceil(self.iterator.get_num_batches(self.train_data)/num_gpus)
+        self.last_log = time.time()
+        last_save_time = time.time()
+
+        logger.info("Training")
+        train_generator_tqdm = Tqdm.tqdm(train_generator,
+                                         total=num_training_batches)
+        self.handler.state.cumulative_batch_size = 0
+        for batch_group in train_generator_tqdm:
```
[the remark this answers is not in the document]
joelgrus: I am really torn on this, whether the goal should be

(1) any training regime you want should be achievable using callbacks, or
(2) it should be easy to write your own trainer, because most of the add-on functionality lives in callbacks

should you be able to train a GAN (or do multitask training, or RL) just by changing the callbacks in the trainer? or should you have to write a GAN trainer, but then you can use callbacks to get checkpointing, early stopping, etc without having to implement it all yourself.

as you can see, the latter is already pretty complicated, I think the former would be even more complicated.

for instance, the `MovingAverage` functionality made things really complex, because at several points it needs to be done before an operation (e.g. saving weights) and then undone right afterward.

allennlp/training/callbacks/callback_handler.py
```diff
@@ -0,0 +1,32 @@
...
+    def add_callback(self, callback: Callback) -> None:
+        self.callbacks.append(callback)
+        self.callbacks.sort(key=lambda cb: cb.priority)
+
+    def remove_callback(self, callback: Callback) -> None:
+        self.callbacks = [cb for cb in self.callbacks if cb != callback]
+
+    def fire_event(self, event: str) -> None:
+        for callback in self.callbacks:
+            callback(event, self.state)
```
matt-gardner: I don't really get why every callback is handling every event. That seems like not a great idea. Why don't you add callbacks to specific events?

allennlp/training/trainer2.py
```diff
@@ -0,0 +1,673 @@
...
+        self.num_epochs = num_epochs
+
+        self.checkpointer = checkpointer
+        self.checkpoint_epoch: Union[int, str] = 0
+        self.model_save_interval = model_save_interval
+
+        self.grad_norm = grad_norm
+        self.grad_clipping = grad_clipping
+        self.learning_rate_scheduler = learning_rate_scheduler
+        self.momentum_scheduler = momentum_scheduler
```
matt-gardner: Shouldn't all of these parameters be passed to the callbacks now, and disappear from here?

allennlp/training/trainer2.py
```diff
@@ -0,0 +1,673 @@
...
+        callbacks: List[Callback] = [
+                LogTensorboard(log_batch_size_period),
+                LogTensorboardHistograms(),
+                LrsCallback(),
+                CheckpointCallback(),
+                MovingAverageCallback(),
+                Validate()
+        ]
+
+        self.handler = CallbackHandler(callbacks, self)
```
matt-gardner: So, I'm not sure what we get by using `Protocols` instead of just typing all of the `State` objects that you have just as `Trainer`. You (might) gain some flexibility in how you use the callbacks, but it comes at a pretty high cognitive cost, as @DeNeutoy points out.  I'd vote for just using `Trainer` everywhere, and pushing as much of the necessary state into the constructor parameters of the individual callbacks as possible.

allennlp/training/callbacks/callback_handler.py
matt-gardner: I guess this does simplify specifying the callbacks, because you could just take a list, instead of a dictionary...

joelgrus:
OK, a brand new iteration, I took all of your feedback into account, thanks for that. I totally got rid of the Protocol stuff and I moved almost everything out of the Trainer into the callbacks. It's possible I moved too much stuff out; it's also possible I could move even more stuff out. 

You can see that in this version the `Trainer` takes very few parameters (although possibly still too many). The only thing I really don't like about this design is that there's a lot of cognitive load involved in reasoning about the callbacks, each of which does things at several points in the training lifecycle (which itself is extremely granular).

Anyway, take a look and let me know what you love / hate.

epwalsh:
@joelgrus I like this a lot! I had one thought to address @matt-gardner's comment about how currently every callback receives every event:

What if we use a decorator-approach (similar to `Registrable.register`) to define which events a callback handles. Here is one possible solution:

```python
def event_handler(event: str, priority: int = None):
    """Mark a Callback method for handling a certain event."""
    def decorator(function):
        function._event = event
        function._priority = priority
        return function
    return decorator


class Callback(Registrable):

    # Priority is specified in each event handler (i.e. each method decorated with @ event_handler),
    # not at the class level.
    # No __call__ function needed.

    def get_training_state(self) -> dict:
        # Same as before
        return {}

    def restore_training_state(self, training_state: dict) -> None:
        # Same as before
        pass


@Callback.register("learning_rate_scheduler")
class LrsCallback(Callback):

    @event_handler("AFTER_BACKWARD")  # mark this method for handling the AFTER_BACKWARD event
    def handle_after_backward(self, trainer) -> None:
        print("LrsCallback AFTER_BACKWARD")


class CallbackHandler(Registrable):

    def __init__(self, callbacks, state) -> None:
        self._event_handlers = defaultdict(list)
        self.state = state
        for callback in callbacks:
            self.add_callback(callback)

    def add_callback(self, callback):
        for method in callback.__class__.__dict__.values():
            if hasattr(method, "_event"):
                self._event_handlers[method._event].append((callback, method))
                self._event_handlers[method._event].sort(key=lambda tup: tup[1]._priority)

    def remove_callback(self, callback):
        for event, handlers in self._event_handlers:
            self._event_handlers[event] = [(cb, method) for cb, method in handlers if cb != callback]

    def fire_event(self, event: str) -> None:
        for callback, method in self._event_handlers[event]:
            method(callback, self.state)
```

For example:

```python
handler = CallbackHandler([LrsCallback()], {})
handler._event_handlers
handler.fire_event("AFTER_BACKWARD")
```
yields

```
LrsCallback AFTER_BACKWARD
```

A couple notes on this approach:
- each callback only receives events for which it has a marked method for, i.e. for which it has a method decorated with `@ event_handler(event_name, priority=None)`.
- the order in which callback methods are fired for each event depends on the priority assigned to the method (through the `event_handler` decorator).
- a callback could have multiple handlers (marked methods) for each event. This might be useful if you want one callback to fire, and then another one, and then the first one to fire again with a different method after the second one has changed some state (I can't think of a specific example for this, but it definitely adds flexibility).

allennlp/training/callbacks/callback.py
```diff
@@ -0,0 +1,29 @@
...
+
+from allennlp.common.registrable import Registrable
+
+TrainerType = TypeVar('TrainerType')  # pylint: disable=invalid-name
+
+class Callback(Registrable, Generic[TrainerType]):
+    # Lower priority comes first
+    priority = 0
+
+    def __call__(self, event: str, trainer: TrainerType) -> None:
```
DeNeutoy: More of a question as i'm sure you've thought about this: Why should this be generic over a `TrainerType` as opposed to being `Trainer`? Is the idea that people will write their own trainers which _do not_ subclass `Trainer`?

allennlp/training/callbacks/callback.py
```diff
@@ -0,0 +1,29 @@
...
+        raise NotImplementedError
+
+    def get_training_state(self) -> dict:
```
DeNeutoy: Suggestion: mirroring Pytorch and calling these `get_state_dict` and `set_state_dict` will mean people immediately know what they do

allennlp/training/callbacks/callback.py
```diff
@@ -0,0 +1,29 @@
+from typing import TypeVar, Generic
+
+from allennlp.common.registrable import Registrable
+
+TrainerType = TypeVar('TrainerType')  # pylint: disable=invalid-name
+
+class Callback(Registrable, Generic[TrainerType]):
```
DeNeutoy: This class could do with some doccumentation describing the process of implementing a new callback, registering it, what events are exposed within the body of calling it, when they happen etc

allennlp/training/callbacks/callback.py
```diff
@@ -0,0 +1,29 @@
...
+    # Lower priority comes first
+    priority = 0
```
DeNeutoy: Having this be a class attribute is a little opaque for a user implementing their own subclass - why not just have it be a constructor argument with a default value?

allennlp/training/callbacks/callback.py
matt-gardner: Yeah, I also think that you could just remove the generics entirely.  If you want the types to actually work out, you can't mix and match callbacks that require different types of trainers, so I'm not really sure what the generics buy you at all, they just add a bunch of cognitive overhead.

allennlp/training/callbacks/callback_handler.py
```diff
@@ -0,0 +1,33 @@
...
+blah
+"""
+from typing import Iterable, TypeVar, Generic
+
+from allennlp.common.registrable import Registrable
+from allennlp.training.callbacks.callback import Callback
+
+State = TypeVar('State')  # pylint: disable=invalid-name
+
+class CallbackHandler(Registrable, Generic[State]):
```
matt-gardner: The entire need for generics in this class also goes away if you just use `Trainer`.

allennlp/training/callbacks/callbacks.py
```diff
@@ -0,0 +1,512 @@
...
+        self.tensorboard = tensorboard
+
+        self.cumulative_batch_size = 0
+
+        # For logging histograms
+        self.histogram_parameters: Set[str] = set()
+        self.param_updates: Dict[str, torch.Tensor] = {}
+
+    def __call__(self, event: str, trainer: 'trainer2.Trainer') -> None:
+        if event == Events.TRAINING_START:
```
matt-gardner: I much prefer @epwalsh's idea of using decorators on methods instead of this giant `__call__` method.  At the very least you should split out each block inside this `if/else` into its own method.  And once you do that, `__call__` becomes unnecessary if we can just decorate those methods directly - this simplifies the implementation of every callback, because the handler has logic to do the routing.

allennlp/training/callbacks/callbacks.py
```diff
@@ -0,0 +1,512 @@
...
+                self.tensorboard.log_histograms(trainer.model, self.histogram_parameters)
+
+        elif event == Events.EPOCH_END:
+            self.tensorboard.log_metrics(trainer.train_metrics,
+                                         val_metrics=trainer.val_metrics,
+                                         log_to_console=True,
+                                         epoch=trainer.epoch_number + 1)
+
+    @classmethod
+    def from_params(cls, serialization_dir: str, params: Params) -> 'LogToTensorboard':  # type: ignore
```
matt-gardner: Why do you need this?  Oh, because `serialization_dir` isn't a `TensorboardWriter`?  It'd be nice to remove the need for this custom `from_params` somehow.

allennlp/training/callbacks/callbacks.py
```diff
@@ -0,0 +1,512 @@
...
+    def restore_training_state(self, training_trainer: dict) -> None:
+        state_dict = training_trainer.pop("learning_rate_scheduler", None)
+
+        if state_dict:
+            self.learning_rate_scheduler.load_state_dict(state_dict)
+
+
+
+    @classmethod
+    def from_params(cls, params: Params, optimizer: Optimizer) -> 'LrsCallback':  # type: ignore
```
matt-gardner: You really shouldn't need this one - doesn't the optimizer just get passed through as an extra?

allennlp/training/callbacks/callbacks.py
```diff
@@ -0,0 +1,512 @@
...
+                trainer.epoch_number = int(training_state["epoch"].split('.')[0]) + 1
+
+        elif event == Events.TRAINING_END:
+            # Load the best model state before returning
+            best_model_state = self.checkpointer.best_model_state()
+            if best_model_state:
+                trainer.model.load_state_dict(best_model_state)
+
+    @classmethod
+    def from_params(cls, params: Params, serialization_dir: str) -> 'CheckpointCallback':  # type: ignore
```
matt-gardner: Same here; can't we get rid of this?  Maybe by changing the way `Checkpointer.from_params` works somehow?

joelgrus:
ok, next iteration. things to pay attention to:

1. I took @epwalsh 's suggestion of using a `@handle_event(event, priority)` decorator, so the code now reflects that

2. the `CallbackTrainer` class is now incredibly minimal, almost everything happens in callbacks

3. the CallbackTrainer tests are 1:1 with the TrainerTests, so it can do all the same things (at least, it can do all the things that are tested for :innocent: )

4. the callback trainer still assumes some things that might not be true: there is a single model, there is a single optimizer. its from_params also assumes there is a dataset reader and that it can load instances

5. if you look at the GanTrainerCallback test, I train a GAN that violates all of those assumptions. The model part is pretty easy, for the optimizer part I had to monkeypatch `Optimizer.from_params`, for the dataset_reader part I had to create a do-nothing dataset reader (but returning no instances is an error, so it returns a single empty instance :grimacing: )

anyway, take a look and let me know what you think, if you have any ideas for further improvements or for how to make my GAN example less awful

allennlp/training/callbacks/callback_handler.py
```diff
@@ -0,0 +1,49 @@
+"""
+blah
```
epwalsh: :wink:

allennlp/training/callbacks/callback_handler.py
```diff
@@ -0,0 +1,49 @@
...
+from collections import defaultdict
+import inspect
+
+from allennlp.training.callbacks.callback import Callback
+
+
+class CallbackMember(NamedTuple):
+    callback: Callback
+    name: str
+    member: Callable
```
epwalsh: Would it be better to call this something more specific like `event_handler`  instead of just `member`? Or maybe the named tuple could be called `CallbackEventHandler`? That way the naming would correspond to the `@event_handler` decorator

allennlp/training/callbacks/callback_handler.py
```diff
@@ -0,0 +1,49 @@
...
+        for callback in callbacks:
+            self.add_callback(callback)
+
+    def callbacks(self) -> Iterable[Callback]:
+        for callback_list in self._callbacks.values():
+            for callback in callback_list:
+                yield callback.callback
+
+    def add_callback(self, callback: Callback) -> None:
+        for name, method in inspect.getmembers(callback, _is_event_handler):
```
epwalsh: Nice!

allennlp/tests/training/gan_callback_trainer_test.py
```diff
@@ -0,0 +1,439 @@
...
+            "stdev": self.fake_stdev / max(self.count, 1)
+        }
+
+# This is bad, I have to monkeypatch Optimizer.from_params
+_original_optimizer_from_params = Optimizer.from_params  # pylint: disable=invalid-name
+
+def _optimizer_from_params(model_parameters: List, params: Params) -> Optimizer:
+    typ3 = params.get("type")
+    if typ3 == "gan":
+        generator_parameters = [[n, p] for n, p in model_parameters if n.startswith("generator.")]
```
matt-gardner: Can't you just `return GanOptimizer.from_params`?

allennlp/tests/training/gan_callback_trainer_test.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+def config(sample_size: int = 500,
+           batches_per_epoch: int = 40,
+           num_epochs: int = 50,
+           learning_rate: float = 0.05) -> Params:
+    return Params({
+        # These are unnecessary but the TrainerPieces.from_params expects them
+        "iterator": {"type": "basic"},
+        "dataset_reader": {"type": "nil"},
```
matt-gardner: Could you get around this by defining a `GanTrainer(CallbackTrainer)` that basically just defines a new constructor and calls `super().__init__`?  No other overrides necessary?  That way you could probably also have the constructor take two optimizers and two models, and not need the other hacks that you have.  Except the `**extras` stuff would depend on having a single named `model` argument that gets passed to each optimizer, so you would still need a custom `from_params`...  But that would at least let you isolate the problem to `GanTrainer.from_params`, without the need for a kludgey model and optimizer grouping.

The issue is that the types get a little off in the callback.  But it looks like they already are, because you don't have type annotations on `trainer` in your `TrainGan` callback.  This way, you could annotate it as `GanTrainer`, which is a subclass of `CallbackTrainer`, and things should be fine (except for the `handle_event` decorator, but I don't think mypy is sophisticated enough to check that anyway).

allennlp/tests/training/gan_callback_trainer_test.py
```diff
@@ -0,0 +1,439 @@
...
+            ]
+
+        }
+    })
+
+
+class GanCallbackTrainerTest(AllenNlpTestCase):
+    def setUp(self):
+        super().setUp()
+        MonkeyPatch().setattr(Optimizer, 'from_params', _optimizer_from_params)
```
matt-gardner: When does this undo itself?  Presumably not when this goes out of scope, because that's before the test runs.

allennlp/tests/training/gan_callback_trainer_test.py
```diff
@@ -0,0 +1,439 @@
...
+    # Run it yourself, it's fun!
+    #
+    # python -m allennlp.tests.training.gan_callback_trainer_test
+    #
+    # pylint: disable=invalid-name
+    MonkeyPatch().setattr(Optimizer, 'from_params', _optimizer_from_params)
+
+
+    from allennlp.training.trainer_base import TrainerBase
+    serialization_dir_ = tempfile.mkdtemp()
```
matt-gardner: Why the underscores?  Because of pylint?  But you already disabled `invalid-name`...

allennlp/training/callback_trainer.py
```diff
@@ -0,0 +1,250 @@
...
+        self.training_start_time = 0.0
+        self.checkpoint_epoch: Union[int, str] = 0
+        self.model_save_interval = model_save_interval
+
+        self.last_log = 0.0
+        self.epoch_number = 0
+        self.batch_grad_norm: Optional[float] = None
+        self.handler = CallbackHandler(callbacks, self)
+
+    def batch_loss(self, batch_group: List[TensorDict], for_training: bool) -> torch.Tensor:
```
matt-gardner: Why does this function exist here?  It's not called anywhere internally to this class.  Shouldn't it live in a callback somewhere?

allennlp/training/callback_trainer.py
```diff
@@ -0,0 +1,250 @@
...
+        self.training_start_time = time.time()
+
+        for self.epoch_number in range(starting_epoch, self.num_epochs):
+            epoch_start_time = time.time()
+            ####
+            self.handler.fire_event(Events.EPOCH_START)
+
+            self.train_loss = 0.0
+            # Set the model to "train" mode.
+            self.model.train()
```
matt-gardner: I feel like this should be in the `Events.EPOCH_START` handler.  Context: I'm looking for where things would break if you make a `GanTrainer` that does `super().__init__(model=None, optimizer=None)`.  I think this is the only thing that needs to be fixed, and then my suggestion should just work.

allennlp/training/callbacks/callback.py
matt-gardner: I agree with this.

allennlp/training/callbacks/callback.py
matt-gardner: I agree with this.

allennlp/training/callbacks/callback_handler.py
matt-gardner: I agree with this.

allennlp/training/callbacks/callback_handler.py
```diff
@@ -0,0 +1,49 @@
...
+    name: str
+    member: Callable
+    priority: int
+
+def _is_event_handler(member) -> bool:
+    return inspect.ismethod(member) and hasattr(member, '_event') and hasattr(member, '_priority')
+
+
+class CallbackHandler:
+    def __init__(self, callbacks: Iterable[Callback], state) -> None:
```
matt-gardner: I assume mypy doesn't like it if you give `state` a type?

allennlp/training/callbacks/callback_handler.py
matt-gardner: Can't you be more specific here, though?  `event_handler: Callable[[TrainerBase], None]`

allennlp/training/callbacks/callbacks.py
```diff
@@ -0,0 +1,555 @@
+# pylint: disable=unused-variable,arguments-differ
+
+# These should probably all live in separate files
```
matt-gardner: Yeah, I'd probably agree with this.

allennlp/training/callbacks/callbacks.py
```diff
@@ -0,0 +1,555 @@
...
+
+        if state_dict:
+            self.learning_rate_scheduler.load_state_dict(state_dict)
+
+
+@Callback.register("momentum_scheduler")
+class MomentumSchedulerCallback(Callback):
+    """
+    Callback that runs a Momentum Scheduler.
+    TODO: this looks almost exactly like the LrsCallback, could we share code?
```
matt-gardner: Probably not, but either way, the TODO should be in a comment, not in the docstring (which goes into our docs).

allennlp/tests/training/callback_trainer_test.py
```diff
@@ -0,0 +1,787 @@
...
+        next_epoch = trainer.epoch_number
+        best_epoch = trainer.metric_tracker.best_epoch
+
+        # Loss decreases in 3 epochs, but because we hard fed the val metrics as above:
+        assert next_epoch == 3
+        assert best_epoch == 1
+        assert trainer.metric_tracker._best_so_far == 0.1
+        assert trainer.metric_tracker._epochs_with_no_improvement == 1
+
+class TestSparseClipGrad(AllenNlpTestCase):
```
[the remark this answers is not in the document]
joelgrus: I copied over all the tests from `test_trainer.py`, they were there, I deleted them now

allennlp/tests/training/callback_trainer_test.py
```diff
@@ -0,0 +1,787 @@
...
+        from allennlp.commands.train import train_model
+
+        params = Params({
+                "trainer": {
+                    "type": "callback",
+                    "optimizer": {"type": "sgd", "lr": 0.01, "momentum": 0.9},
+                    "num_epochs": 2,
+                    "callbacks": [
+                        {"type": "log_to_tensorboard", "log_batch_size_period": 10},
+                        {"type": "checkpoint"},
```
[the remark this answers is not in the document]
joelgrus: changed

allennlp/tests/training/callback_trainer_test.py
```diff
@@ -0,0 +1,787 @@
...
+        restore_trainer.handler.fire_event(Events.RESTORE_CHECKPOINT)
+        assert restore_trainer.epoch_number == 2
+        # One batch per epoch.
+        assert restore_trainer.batch_num_total == 2
+
+    def test_trainer_from_base_class_params(self):
+        params = Params.from_file(self.FIXTURES_ROOT / 'simple_tagger' / 'experiment.json')
+
+        # Can instantiate from base class params
+        TrainerBase.from_params(params, self.TEST_DIR)
```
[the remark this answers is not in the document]
joelgrus: the `from_params` example at the top uses almost all the callbacks, I think?

allennlp/training/callback_trainer.py
joelgrus: it lives here so that both the "training" and "validation" callbacks (as well as anyone else who wants it, I guess) can use it

matt-gardner: Slight preference for having it be a utility function, so that the `CallbackTrainer` can more cleanly get away with `self.model = None`.

joelgrus:
ok, I think this is in pretty good shape, i was able to remove most of the unpleasantness around the GAN test, the only "cheating" left is that I use `setattr` in the model constructor to mark which parameters came from which model and then check that in the custom `GanOptimizer.from_params`

matt-gardner:
Huh, that was not the fix I expected, but I like it.  Nice work!

williamFalcon:
@matt-gardner @joelgrus. why not just use PyTorch Lightning as the trainer backend and AllenNLP? less code for you to maintain and all the beat features already built in...

no need to duplicate work. 
https://github.com/williamFalcon/pytorch-lightning

It’s an official PyTorch library

PR #2827, opened 2019-05-10
Add support for cross-validation

### LINKED ISSUE #1912: K fold cross validation
It would be great if we could have [K-fold cross validation](https://en.wikipedia.org/wiki/Cross-validation_(statistics)) . For input parameters we can specify the training data, no validation data, and a  new parameter called k-fold validation, with any k>0 indicating we must perform k-fold validation.

Perhaps one can use the KFold function from sklearn.

bryant1410:
Fixes #1912. It takes the 2nd option described there.

TODO:

* [ ] Make more sense of the "recover"
* [ ] Archive the model in every fold (and handle interruptions).
* [ ] Document
* [ ] Add tests

### DIFF
--- a/allennlp/training/cv_trainer.py
+++ b/allennlp/training/cv_trainer.py
@@ -3,10 +3,10 @@ import logging
 import os
 from typing import Any, Dict, Iterable, List, Optional, Tuple, Type
 
-import numpy as np
 from overrides import overrides
 from sklearn.model_selection import BaseCrossValidator, GroupKFold, KFold, LeaveOneGroupOut, LeaveOneOut, \
     LeavePGroupsOut, LeavePOut, StratifiedKFold, TimeSeriesSplit
+from sklearn.model_selection._split import _BaseKFold
 import torch
 
 from allennlp.common import Params, Registrable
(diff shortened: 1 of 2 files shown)


allennlp/training/cv_trainer.py
```diff
@@ -0,0 +1,204 @@
...
+        return self.cross_validator.get_n_splits(instances, groups=groups)
+
+
+class _CrossValidationSplitterWrapper:
+    def __init__(self, cross_validator_class: Type[BaseCrossValidator]) -> None:
+        super().__init__()
+        self.cross_validator_class = cross_validator_class
+
+    def from_params(self, params: Params) -> CrossValidationSplitter:
+        cross_validator = self.cross_validator_class(**params.as_dict())
```
bryant1410: [According to the Python docs](https://docs.python.org/3/library/typing.html#typing.Type), `cross_validator_score` should have as type all the classes that have different constructors. In this case, the base constructor doesn't accept args while the subclasses do. So, strictly speaking, type checking will fail with some tools. However, I preferred to keep it simple (the change would be to accept multiple types, probably to list every subclass).

bryant1410:
By the way, some context. This trainer supposes everything fits in memory. It now uses the validation data as well if available (logging so). And there's an option to leave everything trained with the training data and validate with the validation data (as usual) at the end in case there's an intention to evaluate on test data after it.

As a side comment, in my opinion, the training API overall needs some re-thinking (apart from this PR). `Checkpointer` is really tied to the `Trainer` implementation, for example supporting a definition of "best" (which doesn't make sense to `NoOpTrainer`) or having the concept of "epoch". The way to call `from_params` differs depending on the subtype of `TrainerBase`. Also, `TrainerPieces` seems to be specialized to `Trainer` implementation itself.

joelgrus:
the trainer API is currently being rethought, see https://github.com/allenai/allennlp/pull/2817

matt-gardner:
Yeah, @bryant1410, it'd be great for you to look at that PR and give some thought to whether it would be possible (or even easier) to do this in a callback-based trainer.

bryant1410:
By the way, the main functionality is ready for review. I post it still as a draft because everything is not ready to be merged (essentially because of the things missing in the TODO list above).

bryant1410:
> Yeah, @bryant1410, it'd be great for you to look at that PR and give some thought to whether it would be possible (or even easier) to do this in a callback-based trainer.

Yeah, it seems to me that it'd be easier to do it with the changes of #2817.

matt-gardner:
I'm going to close this because it'd be better to do this with the callback trainer.  I think we'd be very happy to have cross validation easily available in the library, it just should wait and use the new stuff, which should be merged in the next few days.

PR #2830, opened 2019-05-12
WIP: Add training and val metrics to no-op trainer

### LINKED ISSUE #2824: Return training and val metrics in NoOpTrainer
Wouldn't it make sense for `NoOpTrainer` to return training and validation (and "best_validation") metrics (but not loss)?

For example, if I train several models and consume the `TrainerBase` API, I'd expect to get the training and validation metrics returned by the `train` method (e.g., accuracy, but not "loss" because it could possibly not make sense), seamlessly, regardless it's a `Trainer` or a `NoOpTrainer` implementation.

An alternative would be to always call `evaluate` on every set I want it to. But I think it's more practical if I could get the metrics of a no-op trainer right away, without calling `evaluate`.

Thoughts? I'm willing to send a PR if you folks think it's a good idea.

bryant1410:
Fixes #2824.

### DIFF
--- a/allennlp/training/no_op_trainer.py
+++ b/allennlp/training/no_op_trainer.py
@@ -1,31 +1,75 @@
+import math
 import os
-from typing import Dict, Any
+from typing import Any, Dict, List, Union
 
 from allennlp.common import Params
+from allennlp.common.tqdm import Tqdm
+from allennlp.common.util import lazy_groups_of
 from allennlp.models import Model
+from allennlp.training import util as training_util
 from allennlp.training.checkpointer import Checkpointer
 from allennlp.training.trainer import TrainerPieces
 from allennlp.training.trainer_base import TrainerBase
 
+
 @TrainerBase.register("no_op")
 class NoOpTrainer(TrainerBase):
-    def __init__(self, serialization_dir: str, model: Model) -> None:
+    def __init__(self, serialization_dir: str, model: Model, cuda_device: Union[int, List] = -1) -> None:
         """
         A trivial trainer to assist in making model archives for models that do not actually
         require training. For instance, a majority class baseline.
         """
 
-        super().__init__(serialization_dir, cuda_device=-1)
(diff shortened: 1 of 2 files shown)


allennlp/training/no_op_trainer.py
```diff
@@ -1,32 +1,76 @@
...
-    def from_params(cls,   # type: ignore
+    def from_params(cls,  # type: ignore
                     params: Params,
                     serialization_dir: str,
                     recover: bool = False):
         # pylint: disable=arguments-differ
         pieces = TrainerPieces.from_params(params, serialization_dir, recover)  # pylint: disable=no-member
         return NoOpTrainer(serialization_dir, pieces.model)
 
+    def _run_on_validation(self) -> Dict[str, float]:
```
bryant1410: This basically copies:

https://github.com/allenai/allennlp/blob/9171766b148fb4e37576c20b2285611354a3a686/allennlp/training/trainer.py#L406-L443

bryant1410:
I have a related question @joelgrus : what is the problem of handling the evaluation iterator and dataset in the same way as in the "default" case?

https://github.com/allenai/allennlp/blob/b90722188a1ff670f5b1cb8a8f2bdb57c42efe1b/allennlp/commands/train.py#L227-L233

bryant1410:
> I have a related question @joelgrus: what is the problem of handling the evaluation iterator and dataset in the same way as in the "default" case?

Oh, I think I understand now. The problem is how to obtain them in the general case because the `TrainerPieces` object is obtained implicitly.

matt-gardner:
This also actually looks like it'd be easier with callbacks - you just don't specify a training callback, and you get a no-op trainer.  Is that right?

bryant1410:
> This also actually looks like it'd be easier with callbacks - you just don't specify a training callback, and you get a no-op trainer. Is that right?

I guess so. But in any case, it'd be good to be taken into account for the new Trainer.

bryant1410:
I can close it because the API is changing anyway.

PR #2835, opened 2019-05-13
Fix test evaluation after training in the general case

bryant1410:
Following https://github.com/allenai/allennlp/pull/2830#issuecomment-491868200

@joelgrus this patch will make test evaluation work until #2817 is merged.


joelgrus:
thanks

PR #2837, opened 2019-05-13
Remove multiple GPU warning

brendan-ai2:
- This code is relatively stable.
- However, our support isn't ideal as it's based on `torch.nn.DataParallel`.
- We should really investigate using `torch.nn.parallel.DistributedDataParallel`.


brendan-ai2:
Thanks, @schmmd !

Issue #1591, opened 2018-08-08
Informative error message for empty strings


maartensap:
I had a dataset reader that was reading from a CSV, one of my lines was just a bunch of "\n\n".
When the TextField containing the string calls `as_tensor()`, the `padding_length` are missing keys (upon debugging it seems the token attribute of the text field was an empty list).

(co written with Nelson Liu by my side)

matt-gardner:
I'm a bit confused as to the issue here.  Can you give a full stack trace?  I'm not sure that the `TextField` can catch these errors, as we need to be able to handle padding, so we can't just check for empty token lists.  This looks like it's just something that has to be handled in your `DatasetReader`, but if you give a full stack trace, it'll be more clear whether there's anything we can do here.

nelson-liu:
(I was helping Maarten debug this yesterday)

Stack trace:

‘’’
2018-08-07 10:27:09,803 - INFO - allennlp.training.trainer - GPU 1 memory usage MB: 2976
  0%|          | 0/30479 [00:00<?, ?it/s]2018-08-07 10:27:09,804 - INFO - allennlp.training.trainer - Training
accuracy: 0.8014, f1: 0.8783, loss: 0.4108 ||:  63%|######2   | 19128/30479 [18:08<10:46, 17.57it/s]Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/msap/venv36/lib/python3.6/site-packages/allennlp/run.py", line 18, in <module>
    main(prog="allennlp")
  File "/home/msap/venv36/lib/python3.6/site-packages/allennlp/commands/__init__.py", line 70, in main
    args.func(args)
  File "/home/msap/venv36/lib/python3.6/site-packages/allennlp/commands/train.py", line 101, in train_model_from_args
    args.recover)
  File "/home/msap/venv36/lib/python3.6/site-packages/allennlp/commands/train.py", line 131, in train_model_from_file
    return train_model(params, serialization_dir, file_friendly_logging, recover)
  File "/home/msap/venv36/lib/python3.6/site-packages/allennlp/commands/train.py", line 296, in train_model
    metrics = trainer.train()
  File "/home/msap/venv36/lib/python3.6/site-packages/allennlp/training/trainer.py", line 679, in train
[...]

PR #2843, opened 2019-05-15
Fix TextField padding when there are no tokens

bryant1410:
Related to (fixes?) #1591

The bug is about some `TextField` in some instance of the dataset that has no tokens and, at the same time, a `TokenCharactersIndexer` is used. When `to_tensor` is called it fails.


schmmd:
Thanks!  Seems like a good improvement but I requested a review from someone whose a bit deeper in the codebase.

PR #2857, opened 2019-05-17
Uptick dependencies.

schmmd:
We don't have to merge this--but these seem like safe upgrades.  I was playing around with dependency versions to find out how brittle the error I had yesterday was.  Today I cannot reproduce that problem at all.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,12 +7,10 @@
 
 # This installs Pytorch for CUDA 8 only. If you are using a newer version,
 # please visit http://pytorch.org/ and install the relevant version.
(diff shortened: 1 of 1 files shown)


joelgrus:
we should be as lenient as possible with our dependencies, at ICLR someone actually complained to me that we were too strict about them and that it made it hard to install allennlp alongside other libraries.

unless we absolutely need the newer versions of things we should leave the bounds as loose as we can

schmmd:
🤦‍♂ I'm not even going to go into what I thought I was doing in this PR.  Thank you for closing ;-)

PR #2860, opened 2019-05-17
Add output path to drop eval

gabrielStanovsky:
Leaderboard expects a json file with the evaluation metrics.
Having the evaluator code output a file that conforms to that format will simplify the maintenance of the DROP leaderboard, as the backend can just call 
`python -m allennlp.tools.drop_eval --gold_path path/to/gold --prediction_path path/to/predictions --output_path path/to/metrics`

### DIFF
--- a/allennlp/tools/drop_eval.py
+++ b/allennlp/tools/drop_eval.py
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 
 from collections import defaultdict
-from typing import Any, Dict, List, Set, Tuple, Union
+from typing import Any, Dict, List, Set, Tuple, Union, Optional
 import json
 import argparse
 import string
@@ -217,7 +217,7 @@ def evaluate_json(annotations: Dict[str, Any], predicted_answers: Dict[str, Any]
     return global_em, global_f1
 
 
-def evaluate_prediction_file(prediction_path: str, gold_path: str, output_path: str) -> Tuple[float, float]:
+def evaluate_prediction_file(prediction_path: str, gold_path: str, output_path: Optional[str]) -> Tuple[float, float]:
     """
     Takes a prediction file and a gold file and evaluates the predictions for each question in the
     gold file.  Both files must be json formatted and must have query_id keys, which are used to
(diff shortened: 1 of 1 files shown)


allennlp/tools/drop_eval.py
```diff
@@ -216,18 +217,28 @@ def evaluate_json(annotations: Dict[str, Any], predicted_answers: Dict[str, Any]
...
     predicted_answers = json.load(open(prediction_path, encoding='utf-8'))
     annotations = json.load(open(gold_path, encoding='utf-8'))
-    return evaluate_json(annotations, predicted_answers)
+    global_em, global_f1 = evaluate_json(annotations, predicted_answers)
+
+    # Output predictions to file
+    output_dict = {"global_em": global_em,
+                   "global_f1": global_f1}
+
+    with open(output_path, "w", encoding = "utf8") as fout:
```
matt-gardner: `outfile` would better follow our typical style.

allennlp/tools/drop_eval.py
```diff
@@ -243,5 +254,11 @@ def evaluate_prediction_file(prediction_path: str, gold_path: str) -> Tuple[floa
                         required=False,
                         default="sample_predictions.json",
                         help='location of the prediction file')
+    parser.add_argument("--output_path",
+                        type=str,
+                        required=False,
+                        default=os.devnull,
```
matt-gardner: I think you should hide this behind an `if` in the main logic, instead of trying to write to devnull.

allennlp/tools/drop_eval.py
gabrielStanovsky: fixed

allennlp/tools/drop_eval.py
gabrielStanovsky: I'm not sure I understand, what's the problem with writing to devnull?
Another option is to add a default output file, would that be better?

matt-gardner: It's just better to avoid it, I think. I don't know how it interacts with docker or other things. It's an unnecessary complexity that we can avoid just with a default value of `None`.

gabrielStanovsky: Got it, makes sense.
I changed this to use None instead.

gabrielStanovsky:
Thanks! I'll merge when build passes.

PR #2866, opened 2019-05-19
Make DROP EM and F1 calculation length aware

nelson-liu:
In short, the problems this fixes:

1) EM wasn't length-aware, since only the bags were being compared.
2) The F1 calculation was taking the average of the optimal alignments between the gold and the predicted. However, if there are more predicted answers than gold answers, the precision isn't penalized. In the case where there are more predicted answers, the excess answers are all given an F1 of 0.

## Examples:

Before this patch:

```
In [1]: from allennlp.tools.drop_eval import get_metrics

In [2]: get_metrics(predicted=["td"], gold=["td", "td"])
Out[2]: (1.0, 0.5)

In [3]: get_metrics(predicted=["td", "td"], gold=["td"])
Out[3]: (1.0, 1.0)

# Suppose the passage is ["the", "fat", "cat"]
In [4]: get_metrics(predicted=["the", "fat", "cat", "the fat", "fat cat", "the fat cat"], gold=["cat"])
Out[4]: (0, 1.0)
```

After this patch:

```
In [1]: from allennlp.tools.drop_eval import get_metrics

In [2]: get_metrics(predicted=["td"], gold=["td", "td"])
Out[2]: (0.0, 0.5)

In [3]: get_metrics(predicted=["td", "td"], gold=["td"])
Out[3]: (0.0, 0.5)

In [4]: get_metrics(predicted=["the", "fat", "cat", "the fat", "fat cat", "the fat cat"], gold=["cat"])
Out[4]: (0.0, 0.17)
```

### DIFF
--- a/allennlp/tests/tools/drop_eval_test.py
+++ b/allennlp/tests/tools/drop_eval_test.py
@@ -19,17 +19,21 @@ class TestDropEvalGetMetrics:
         assert get_metrics(["78"], ["78.0"]) == (1.0, 1.0)
 
     def test_metric_is_length_aware(self):
-        # EM is 0.0 here.
-        # F1 between gold[0] and predicted[0] is 1.0
-        # F1 between gold[1] and predicted[1] is 0.0 (since predicted[1] doesn't exist).
-        # So, overall F1 is 0.5 (the mean of [1.0, 0.0]).
+        # Overall F1 should be mean([1.0, 0.0])
         assert get_metrics(predicted=["td"], gold=["td", "td"]) == (0.0, 0.5)
         assert get_metrics("td", ["td", "td"]) == (0.0, 0.5)
-        # EM is 0.0 here.
-        # F1 between gold[0] and predicted[0] is 1.0.
-        # So, overall F1 is 1.0 (mean of [1.0]).
-        assert get_metrics(predicted=["td", "td"], gold=["td"]) == (0.0, 1.0)
-        assert get_metrics(predicted=["td", "td"], gold="td") == (0.0, 1.0)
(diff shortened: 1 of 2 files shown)


dDua:
Thanks for the PR! Just curious about the second test case,

get_metrics(predicted=["td", "td"], gold=["td"])

does it really matter if the output contains multiple same valued spans. The only reason why this would happen is when say same entity name appears at multiple places in the context and since we don't have the index of the right instance, would it matter if the output contains same span multiple times.  Can you give an example where this could potentially be bad?

allennlp/tests/tools/drop_eval_test.py
```diff
@@ -58,10 +75,12 @@ def test_simple_overlap_in_incorrect_cases(self):
 
     def test_multi_span_overlap_in_incorrect_cases(self):
         # only consider bags with matching numbers if they are present
+        # F1 scores of:     1.0        2/3   0.0   0.0   0.0   0.0
```
matt-gardner: The last one is zero because of a number mismatch?

nelson-liu: yes

dDua:
> LGTM, thanks @nelson-liu. @dDua, we definitely want to penalize a system that produces multiple spans for over-producing them. Otherwise you could just always output every possible span and maximize F1.

I agree with the over-generated spans. I was just asking about the third test case which requires changing from set to list. I was wondering why is it wrong for the system to generate the same answer span multiple times?

matt-gardner:
For the same reason. Why should you want the system to be able to give a wrong answer and still get credit?

Issue #2099, opened 2018-11-26
Any plan to seperate non-NLP part to its own library to benefit other deep learning research field


wheatdog:
I found the way that allennlp handle deep learning experiments quite elegant. And the quality of code is so great. I am mainly doing research on computer vision, and didn't find something similar. I think it would be good to reuse some part of allennlp for varies kind of deep learning research.

schmmd:
@wheatdog we don't presently have a plan to separate out the non-NLP components as our group is mainly focused on NLP problems.  What problems would you run into if you used the library as is, but avoided the NLP components?

wheatdog:
I am not sure how can I use allennlp to build a tool for deep learning experiment in other field, says computer vision. In #1877, I understand that I have to change the logic in Trainer, and maybe the interface of Model. So maybe the important part I can reuse will be Registrable class? The main idea will be reuse most of the part in allennlp if I can. I will try to play around with this idea and see what I can get.

matt-gardner:
@wheatdog - I'm guessing you can use much more than just that.  If you want to use this for computer vision, you'd basically just need to add in some `Fields` that handle image / video inputs.  Everything else should just work.  There might be things that are slow, because we haven't dealt with such large files as you get in vision problems, but that's where I'd start.

If you get this to work well, we'd also be very much interested in having you contribute back to the library, as there are plenty of places where vision and language intersect (e.g., captioning, VQA), and we've love to have the ability to do those tasks in AllenNLP.

sethah:
I've successfully converted a project of mine to use the Allen NLP training tools in a computer vision application, with no modifications to source code. Overall, the current tools, though designed for NLP, are general enough to work well for CV. This conversion simplified my hand-written PyTorch code *a lot*.

I must admit I am still learning the Allen NLP APIs. I used a collection of `ArrayField`s for my inputs and outputs, and a `BasicIterator` to batch them. My initial impressions are that there are a couple of leaky abstractions and then some missing features (missing might be a misnomer, since this is an NLP library so it is a bit weird to call CV features "missing"). For example, image augmentation (there are other libraries to do this, though). 

As far as the leaky abstractions, on example would be the padding methods like `get_padding_lengths` on the `Field` class. Another is having to initialize models with empty vocabularies.

```python
class AnchorBoxModel(allennlp.models.Model)

    def __init__(self):
        super(AnchorBoxModel, self).__init__(Vocabulary())
```

[...]

PR #2892, opened 2019-05-25
Get gradients for model predictions 

jens321:
- Adds a generic `get_gradients` function to the base predictor class to get the gradients of the loss of the prediction with respect to the input embeddings by adding a hook on the embedding layer of the model. 
- The gradients can be used for interpretation and model attack methods, which will be implemented on the specific predictor class of a model via functions called `interpret_from_json` and `attack_from_json`. 
- A normalize function was added to normalize the gradients.


allennlp/predictors/predictor.py
```diff
@@ -57,17 +63,157 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
...
+        def hook_layers(module, grad_in, grad_out):
+            self.extracted_grads.append(grad_out[0])
+
+        # Register the hooks
+        for module in self._model.modules():
+            if isinstance(module, BasicTextFieldEmbedder):
+                backward_hook = module.register_backward_hook(hook_layers)
+                self.hooks.append(backward_hook)
+
+    def _normalize(self, grads: Dict[str, np.ndarray]) -> Dict[str, np.ndarray]:
```
Eric-Wallace: this should probably live somewhere else, not sure where.

allennlp/predictors/predictor.py
```diff
@@ -57,17 +63,157 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
...
+        Uses the gradients from :func:`get_gradients` to provide 
+        normalized interpretations for specific models. 
+
+        Raises
+        ------
+        NotImplementedError
+            This method should be overriden by the specific predictor class of each
+            predictor.
+        """
+        raise NotImplementedError
```
matt-gardner: This needs to raise a `RuntimeError`, not a `NotImplementedError`.  Same for the other methods you added.  This is because we don't want to force all predictors to implement this - we're getting pylint errors about abstract methods not being overridden.  The `RuntimeError` should say something like "you need to implement this method if you want to give model interpretations".

allennlp/predictors/predictor.py
```diff
@@ -57,17 +63,157 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
         instance = self._json_to_instance(inputs)
         return self.predict_instance(instance)
 
+    def interpret_from_json(self, inputs: JsonDict) -> JsonDict:
+        """
+        Uses the gradients from :func:`get_gradients` to provide 
+        normalized interpretations for specific models. 
+
+        Raises
```
matt-gardner: We haven't included these warnings in our docs, particularly for something like this.  I'd just remove it (same for other methods).  You could put a note in the docstring that says subclasses must implement this method if they want this functionality (without the separate `Raises` block).

allennlp/predictors/predictor.py
```diff
@@ -57,17 +63,157 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
...
+        # so we append the incoming gradients to a list
+        self.extracted_grads = []
+        self.hooks = []
+
+        def hook_layers(module, grad_in, grad_out):
+            self.extracted_grads.append(grad_out[0])
+
+        # Register the hooks
+        for module in self._model.modules():
+            if isinstance(module, BasicTextFieldEmbedder):
```
matt-gardner: Can't you just call this on `TextFieldEmbedder` instead of `BasicTextFieldEmbedder`?  It would make this more general.

allennlp/predictors/predictor.py
```diff
@@ -57,17 +63,157 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
         instance = self._json_to_instance(inputs)
         return self.predict_instance(instance)
 
+    def interpret_from_json(self, inputs: JsonDict) -> JsonDict:
```
matt-gardner: Why can't this be implemented in the base class?  Same with `attack_from_json`.

allennlp/predictors/predictor.py
matt-gardner: Probably in `common/util.py`

allennlp/predictors/predictor.py
```diff
@@ -57,17 +63,157 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
...
+        literature and the interpretation is actually quite sensitive to it. 
+        In the code below, we normalize scores across the words, doing positive 
+        and negatives seperately.
+        """
+
+        for key, grad in grads.items():
+            grad = np.sum(grad, axis=1)
+            total_score_pos = 0
+            total_score_neg = 0
+            for idx, s in enumerate(grad):
```
matt-gardner: Use better names than `s` and `n`.

allennlp/predictors/predictor.py
```diff
@@ -57,17 +63,157 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
...
+            grads[key] = grad 
+        return grads 
+
+
     @contextmanager
     def capture_model_internals(self) -> Iterator[dict]:
         """
         Context manager that captures the internal-module outputs of
         this predictor's model. The idea is that you could use it as follows:
-
```
matt-gardner: Why remove these blank lines?  I think this actually breaks sphinx.

allennlp/predictors/predictor.py
```diff
@@ -129,14 +287,11 @@ def _batch_json_to_instances(self, json_dicts: List[JsonDict]) -> List[Instance]
     def from_path(cls, archive_path: str, predictor_name: str = None) -> 'Predictor':
         """
         Instantiate a :class:`Predictor` from an archive path.
-
         If you need more detailed configuration options, such as running the predictor on the GPU,
         please use `from_archive`.
-
```
matt-gardner: These blank lines are also necessary, I believe.

allennlp/predictors/predictor.py
```diff
@@ -166,4 +321,4 @@ def from_archive(cls, archive: Archive, predictor_name: str = None) -> 'Predicto
         model = archive.model
         model.eval()
 
-        return Predictor.by_name(predictor_name)(model, dataset_reader)
+        return Predictor.by_name(predictor_name)(model, dataset_reader)
```
matt-gardner: Add newline at end of file.

allennlp/predictors/predictor.py
```diff
@@ -57,17 +63,157 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
...
+            if isinstance(module, BasicTextFieldEmbedder):
+                backward_hook = module.register_backward_hook(hook_layers)
+                self.hooks.append(backward_hook)
+
+    def _normalize(self, grads: Dict[str, np.ndarray]) -> Dict[str, np.ndarray]:
+        """
+        Normalize the gradients into the range [0,1] for visualization. 
+        Notes
+        -----
+        0.0 is dark red and 1.0 is dark blue. 0.5 is no highlight.
```
matt-gardner: This line isn't applicable anymore.  How the [0,1] values translate to colors happens elsewhere.

allennlp/predictors/predictor.py
```diff
@@ -57,17 +63,157 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
...
+
+    def _normalize(self, grads: Dict[str, np.ndarray]) -> Dict[str, np.ndarray]:
+        """
+        Normalize the gradients into the range [0,1] for visualization. 
+        Notes
+        -----
+        0.0 is dark red and 1.0 is dark blue. 0.5 is no highlight. 
+        This normalization step is very rarely discussed in the 
+        literature and the interpretation is actually quite sensitive to it. 
+        In the code below, we normalize scores across the words, doing positive
```
matt-gardner: So, if I have scores of [-100, -70, 0, 0.1, 0.2], this will say that -100 is just as "negative" as 0.2 is "positive"?  Why is this the right thing to do?

allennlp/predictors/predictor.py
```diff
@@ -103,6 +249,18 @@ def _json_to_instance(self, json_dict: JsonDict) -> Instance:
         """
         raise NotImplementedError
 
+    def predictions_to_labels(self, instance: Instance, outputs: Dict[str, np.ndarray]) -> List[Instance]:
```
matt-gardner: This needs a better name.  I picked this name originally before we decided that the return value should just be a list of `Instances`.  Maybe `predictions_to_labeled_instances`?

allennlp/predictors/predictor.py
```diff
@@ -57,17 +63,157 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
...
+
+        Raises
+        ------
+        NotImplementedError
+            This method should be overriden by the specific predictor class of each
+            predictor.
+        """
+        raise NotImplementedError
+
+    def get_model_predictions(self, inputs: JsonDict) -> List[Instance]:
```
matt-gardner: This should probably be called `inputs_to_labeled_instances`.

allennlp/predictors/predictor.py
```diff
@@ -57,17 +63,157 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
...
+        """
+        Converts incoming json to a :class:`~allennlp.data.instance.Instance`
+        and adds labels to the :class:`~allennlp.data.instance.Instance`s given
```
matt-gardner: This would be clearer if you explicitly said that you run the model as the second step, before "and adds labels".

joelgrus:
sorry, I'm coming in late here, but what's the genesis of this PR? why does the base Predictor class need all these methods? what's the issue this is solving? 

(apologies if this was discussed elsewhere, I couldn't find it)

matt-gardner:
I've mentioned this a couple of times in our weekly meetings, I think - this is adding functionality to the demo that lets you see gradient importance information for the predictions made by any model (and adversarial attacks).  Work being done in Irvine by Eric and a couple of UCI students.  We have demos ready for some of the models, and are working on getting more of them.  We're adding the code incrementally, to avoid one huge PR.

Sameer (and Eric) has done a lot of work in this area (interpreting and attacking models using gradient techniques), and, as Sameer puts it, this ability to have gradient-based interpretation for any model in allennlp is one of the reasons it makes sense to have the two of us in the same room =).

matt-gardner:
Also, though we haven't actually used your "model internals" visualizations for anything yet, seeing how you did it was really helpful in getting this gradient interpretation stuff to work, so thanks for that.

joelgrus:
So I guess part of my concern is that up to this point we've mostly framed `Predictor` as a simple JSON wrapper for your model. But now it has a bunch more methods, and when I look at `attack_from_json` and `interpret_from_json` it's not clear at all to me what they're supposed to do or how I should implement them, and so suddenly the bar to implementing your own Predictor seems a lot higher.

DeNeutoy:
I'm with Joel on this one - it departs pretty substantially from the scope of the original `Predictor` (a json in json out model interface) and is functionality most people won't use. Why not make a mixin `GradientInspector` with these methods? 

Then you could have:

`class NewPredictor(Predictor, GradientInspector)`

for new models, and just register a new subclass with the mixin for existing models.

matt-gardner:
So these are the main things that need to happen:

1. The demo UI hits some rest API served by our demo server, asking for model interpretations.
2. The demo calls `forward` on the model to get the model's predictions.
3. The demo then calls some method that constructs specially-crafted `Instances` that let us compute gradients to see which parts of the input had the most influence on the model's predictions.
4. We call the model again on these new inputs, get the gradient importance stuff.
5. The demo server returns the importance weights to the demo UI, which visualizes it in some way.

Where would you recommend each of these pieces of code living?  Let's consider 2-4 as a single method for now.  Clearly 1, 5, and the call to method 2-4 all need to live in the demo server.  The only object the demo interacts with is a `Predictor`, so unless we change how that works, the method 2-4 needs to be exposed on the `Predictor`.  Are you suggesting making the demo construct something more complicated, instead of just `Predictor.from_archive`?  How does the demo know how to construct this for the individual models?

If we have a mixin on only some of these models, then mypy on the server code gets really complicated.  I'm also not sure how the typing on the mixin works, either, because the methods need to call things on `Predictor` (like `_json_to_instance`, and `_model`).  We could conceivably do a subclass of `Predictor` that adds these things, but I think that unnecessarily complicates things.

As I understood the implementation of this from prior discussions, there should only be one method that a `Predictor` needs to implement if you want gradient-based interpretations, and that's `predictions_to_labeled_instances`.  This method just gives a `RuntimeError` in the base class, so it's optional to implement, and the complexity of implementing a `Predictor` doesn't change.

As far as the original goal of the `Predictor`, I think we've come to the conclusion that its only real use is in the demo - we haven't said this explicitly anywhere, but I think we all agree that the `predict` command should really just use the dataset reader in almost all cases, and we should be able to get by without a `Predictor` at all, except in the demo.  So I'm personally not too bothered by adding demo stuff to the `Predictor`, because that's where it's used.

So, what about the method 2-4?  It has to be called by the demo server, but where should the code live?  We could conceivably have some piece of code call `_json_to_instance` on the `Predictor` in the demo (though we'd want to make it public), then pass off the `Instance` and the `Model` to an encapsulated piece of code that does the gradient interpretation and stuff.  I'm not opposed to this, but it complicates how the server works, or requires the `Predictor` to have a separate object that it constructs somehow.  I'm open to suggestions, but I can't think of a better alternative to what we already came up with.  If the only objection is that this broadens the scope of a `Predictor`, I vote that we should do it, because `Predictor` is really only used for the demo anyway, and the other options that I can think of add significant complexity in other places just to preserve the limited scope of an object that has already lost some of its original purpose.

joelgrus:
I feel like this description (the demo needs one extra method) is at odds with the PR, which introduces 5 new public methods on `Predictor`.

I also am not a fan of having a public method on the `Predictor` API that might just give you a `RuntimeError` and _you have no way of knowing if that's the case_. I take your point that Predictors aren't that useful in practice for `allennlp predict`, but we also use them for `allennlp.pretrained` and the simple server.

matt-gardner:
I commented on the review that I was expecting most of those methods to have implementations in the base class.  Only one new method *has to be implemented* if you want this functionality, which is what I was claiming.  And it can give a simple error saying that "You need to implement X method if you want this functionality", and we make this clear in the docstring.

Do you have an alternative proposal?

DeNeutoy:
Concrete alternative would be to have a registrable `AdversarialAttack` which took as input a (specific?)`Predictor` and had to implement `attack_from_json` and `interpret_from_json`. Then you could register attacks for different models (and also allows multiple attacks for different models which I could certainly imagine you wanting to do at some point in the future).

I think joel's point about the complexity is not that you have to do more work to implement a predictor, but on a first glance _it looks like you do_.

matt-gardner:
Also, @joelgrus, the use in the simple server is exactly my point, and the use in `pretrained` has little to do with json, which again I think argues in favor of my point.

@DeNeutoy, you're saying that the server should introduce a new object that handles the API requests in 1 and 5, and interacts with the predictor for 2-4?  That could work.  I like this a lot better than a mixin, because the types are a lot more clear.

matt-gardner:
Though, we still need `predictions_to_labeled_instances`, and it's really the `Predictor` that should know about how to do this, and so we haven't solved the original complaint.

DeNeutoy:
Yeah, the mixin was not a good idea. I actually don't see why the `predictions_to_labeled_instances` function should live in the predictor. If it was in this outside class, you could just call `predict_instance` with the new `Instance` you create from the labels.

Can you be more specific about what the "original complaint" was? It's hard to discuss if we're accidentally referencing the wrong things. Here's a sketch of what I meant:

```python
class AdversarialAttack(Registrable):

	def __init__(self, predictor: Predictor):
		self.predictor = predictor
	
	def interpret_from_json(self, inputs: JsonDict) -> JsonDict:
		raise NotImplementedError
	def attack_from_json(self, inputs: JsonDict) -> JsonDict:
	
		raise NotImplementedError
	def predictions_to_labels(self,
							  instance: Instance,
							  outputs: Dict[str, np.ndarray]) -> List[Instance]:
		raise NotImplementedError

	+ the other implemented functions in the PR.

AdversarialAttack.register("simple-tagger")
class SimpleTaggerAttack:

	def __init__(self, predictor: SimpleTaggerPredictor):

		super().__init__(predictor)

	# overrides particular methods as needed in a way which is specific
	# to the SimpleTagger Predictor and outputs.

```

I guess one downside to this is that it would require the predictor passed to `AdversarialAttack` to reference a model in the demo for the model.forward calls, rather than hitting one of the existing model apis, now they all run in separate processes in skiff.

matt-gardner:
By "the original complaint" I meant adding a method to the Predictor that needs to be (optionally) implemented by subclasses.

The reason I think the method belongs on the `Predictor` is that it logically fits there - there is one `Predictor` per task, and how you go from predictions on a task to labels for that task is something you only need one of per task (and it's about what you do with predictions, which fits with the intent).  Having a separate registry of objects that are in a one-to-one correspondence with `Predictors` seems odd to me.

We are imagining having several different interpretation and attack methods, which are all gradient based, and independent of the model / predictor being used.  So, e.g., you could imagine a class structure that looks like this (ignore the naming specifics):

```python
class Interpreter:
    def __init__(self, predictor):
        ...

class SimpleGradient(Interpreter):
    ...

class IntegratedGradient(Interpreter):
    ...

class Lime(Interpreter):
    ...
...

class Attacker:
    def __init__(self, predictor):
        ...

class Pathologies(Attacker):
    ...

class Hotflip(Attacker):
    ...
```

Then the `/interpret` endpoint in the demo calls an `Interpreter` (which must live on the host that the model lives on), and the `/attack` endpoint calls an `Attacker`, both of which use some base methods on the `Predictor`.

To make this work, there are two things that need to be added to the `Predictor`: (1) `predictions_to_labeled_instances`, which is task-specific, and (2) `get_gradients`, which can live on the base class and is very similar to the [capture_model_internals](https://github.com/allenai/allennlp/blob/master/allennlp/predictors/predictor.py#L60-L92) method that is already there.  We could possibly even implement it as a context manager like that method is, though I'm not sure if it's worth it or not.

joelgrus:
given what you describe here, shouldn't `attack_from_json` be a method on `Attacker` and `interpret_from_json` a method on `Interpreter`? or am I still missing something?

matt-gardner:
@joelgrus, yes, that was what I was suggesting in my last comment, sorry it wasn't clear enough.

matt-gardner:
@joelgrus, @DeNeutoy, what do you think of my latest comment?  Should we proceed with that design?

joelgrus:
I still am not sure I have the big picture correct, but if what you're imagining is that there is a variety of possible `Attacker`s and a variety of possible `Interpreter`s, and each operates on a `Predictor` using the predictor interface, then what you describe in your last comment sounds like the right design.

DeNeutoy:
Yep I reckon that would be better Also, what do you think about this being something  something that could live in its own library, if it's planning on becoming a major thing for multiple models? There's no reason this precludes it from being part of the demo, as the repo is separate and installs its own packages. I'm not sold on this actually working so if you don't want to then I don't have a strong opinion.

matt-gardner:
Ok, @jens321, can you separate out the pieces of this like I outlined above?  If you have questions about any of it, feel free to ping me on slack.

As far as splitting this out into a separate piece, I think there's reason to consider splitting things out into several different libraries (c.f. #2099), but until we get to that point, I think this code should live here.

jens321:
@matt-gardner Yes, thanks for the feedback, I'll start working on separating it out. Should I close this PR for now (and then later open a new one), or leave it open and just push once the updated pieces are in place?

matt-gardner:
Just leave this one open and update it, I think.

Eric-Wallace:
This can be closed as https://github.com/allenai/allennlp/pull/3032 is taking over as the relevant PR.

PR #2902, opened 2019-05-29
fix dependency conflicts (by removing moto + disabled tests that use it)

### LINKED ISSUE #2901: fix incompatible dependencies
```
ERROR: aws-sam-translator 1.11.0 has requirement jsonschema~=2.6, but you'll have jsonschema 3.0.1 which is incompatible.
ERROR: cfn-lint 0.21.3 has requirement jsonschema~=2.6, but you'll have jsonschema 3.0.1 which is incompatible.
ERROR: cfn-lint 0.21.3 has requirement requests<=2.21.0,>=2.15.0, but you'll have requests 2.22.0 which is incompatible.
```

it seems like all of these are caused by `moto`, which we use in some tests (although right now all of those tests are disabled because at some point moto was just broken 🤔)

joelgrus:
fixes #2901 

they were all caused by `moto`, the mock s3 library, which was used in a few of our tests; HOWEVER, those tests have all been disabled for many months, as the library broke. When I installed the latest version and enabled the tests, one was still broken, for reasons that were not obvious. 

Anyway, given that we don't run these tests anyway, the simplest way to fix the dependency conflict is by getting rid of the tests and getting rid of the dependency.


DeNeutoy:
+1 i'm pretty sure this was used for something pretty niche anyway

PR #2905, opened 2019-05-30
Lr scheduler bug

### LINKED ISSUE #2895: ReduceLROnPlateau mode be faithful to the validation metric passed
Currently, for the ReduceLROnPlateau (or any class subclassing _PyTorchLearningRateSchedulerWithMetricsWrapper), the mode (whether to take a step when a metric increases or decreases) has to be specified independent of the validation metric ("-loss" / "+acc" etc.). This can potentially lead to bugs with scheduling. Eg: the pytorch default for ReduceLROnPleateau uses a min for scheduling. So if you want to schedule based on accuracy, that leads to a bad schedule, if not specified in the learning rate scheduler.

Make sure that the from_params of the trainer specifies the mode for a learning rate scheduler if not specified, that is faithful to the "+/-" value of the validation metric, if specified. If the mode is not specified, then set it appropriately. 
Somewhere after
https://github.com/allenai/allennlp/blob/d6bb6c87581bc3a37ec7cccbb1778364344b1a7f/allennlp/training/trainer.py#L674
would be nice.
Optionally, a check in the \_\_init\_\_() function would also be nice, to make sure someone not using the from_params structure gets warned if the mode and the metric are not faithful to each other.

Another alternative would be that __PyTorchLearningRateSchedulerWrapper requires the mode param to be specified, i.e the default is never used.

codedecde:
Fixes #2895 
Couple of things that I would like to discuss

- Currently, the design makes it necessary to specify the mode while using the reduce_on_plateau scheduler, unless its specified in a trainer (in which case the mode is set automatically, based on the validation metric)
- If the metric and the mode do not match, currently the code uses a logger.warning instead of an exception. Can change that to be an exception.


allennlp/training/util.py
```diff
@@ -437,3 +437,22 @@ def description_from_metrics(metrics: Dict[str, float]) -> str:
     return ', '.join(["%s: %.4f" % (name, value)
                       for name, value in
                       metrics.items() if not name.startswith("_")]) + " ||"
+
```
DeNeutoy: This shouldn't be a private function because you're using it in another file, just remove the preceeding underscore

allennlp/training/trainer.py
```diff
@@ -663,16 +663,28 @@ def from_params(cls,  # type: ignore
                     validation_data: Optional[Iterable[Instance]],
                     params: Params,
                     validation_iterator: DataIterator = None) -> 'Trainer':
-        # pylint: disable=arguments-differ
+        # pylint: disable=arguments-differ,protected-access
```
DeNeutoy: Remove this once you've made the function public

codedecde:
Is it okay for it to be a logger warning when the metrics and mode don't agree in  update_scheduler_params (allennlp/training/util.py, line 455), or should that be a configuration Error  ?

matt-gardner:
@DeNeutoy, is this good to merge?

brendan-ai2:
@DeNeutoy ping. Thanks.

brendan-ai2:
@codedecde , sorry for letting this slip through the cracks. I've given this another skim and based on that plus @DeNeutoy's previous approval I'm going to merge.

Regarding your specific question on the configuration error/warning distinction (allennlp/training/util.py, line 455), I'd be partial to a hard error, but I don't think that needs to hold up merging this PR. Feel free to add that in a followup PR if you feel strongly.

Thanks for your contribution!

PR #2913, opened 2019-05-31
Composed Sequence to Sequence Abstraction

sai-prasanna:
This completes the work started by @generall https://github.com/allenai/allennlp/pull/2517 to decompose seq2seq encoder_decoder models.

## Explanation
* `ComposedSeq2Seq` is the class which composes older `Seq2SeqEncoder` and the newly defined `SeqDecoder`.
* `SeqDecoder` is an abstract class upon which Decoder implementations are extended upon 
* `AutoRegressiveSeqDecoder` is the default implementation for `SeqDecoder`. It composes `DecoderNet`. The decoder net can implemented by anything from a transformer to a LSTMCell. In case of implementations like transformer which support parallel decoding (for training), we have a `decodes_parallel` flag. The `AutoRegressiveSeqDecoder` uses this flag to exploit single forward pass decoding in case of teacher forcing.

## Questions
* Can the current items such as the embedder  in `SeqDecoder` can be moved to the default autoregressive decoder implementation?
* Are the current tests sufficient?
* I am reusing the harvard Transformer parts in the private API to implement `StackedSelfAttentionDecoderNet`

## Help
* ~Need help testing the transformer module on actual data. I am getting NAN errors during validation phase, while running beam search.~ Fixed

Solves https://github.com/allenai/allennlp/issues/2097


sai-prasanna:
@brendan-ai2 @matt-gardner

matt-gardner:
@brendan-ai2, can I get you to take this one?

brendan-ai2:
@matt-gardner, @sai-prasanna I can take a look, but I'm out from today until next Thursday (6/13). So it would be somewhat delayed.

sai-prasanna:
@matt-gardner @brendan-ai2 Thought of wrapping this up this weekend. Thursday will be fine I guess. Since y'all are busy doing cool research (a bit jealous :p).

A bit of good news, the bug in my transformer decoder is fixed. 
During training mutating the embeddings didn't increase the loss too much, but nans popped up only during validation set inference using the beam search. Turned out that NaNs were due to "*=" in-place operator mutating the embeddings directly :unamused: . During beam search, since the same embeddings are called repeatedly in the transformer decoder, it resulted in the bug having effect. 


Only the review of abstractions is left, I guess.

sai-prasanna:
@epwalsh Can you take a look into this to check whether implementations like CopyNet can be done using these abstractions without any hacks ?

@brendan-ai2 - Any update?

epwalsh:
Hi @sai-prasanna, I think one issue with CopyNet-compatibility is that `CopyNet.forward` takes a few other arguments in addition to `source_tokens` and `target_tokens`. But those just end up just being passed around in a dictionary, so maybe the methods layed out in your new API could just take a dictionary as input to be more flexible? Idk though.. personally I don't think that's a great solution 🙃

brendan-ai2:
Hey @sai-prasanna, I'm giving this a closer look now. In the meantime, could you please ensure that `scripts/verify.py` passes? There are a number of linter errors that need to be handled. Thanks!

sai-prasanna:
@brendan-ai2 Fixed the tests. Some guidance on what to unit test would be great. 
And the way I have done weight tying is very unclean. I have accessed private variable containing the  token embedder (can't change it to public without a big mess, it would break deserialization of older models). 

If there is some other generic mechanism, maybe in`IntializorApplicator` to tie two parameters after model construction it would be better I guess.

allennlp/tests/predictors/seq2seq_test.py
```diff
@@ -22,6 +22,22 @@ def test_uses_named_inputs_with_simple_seq2seq(self):
         assert isinstance(predicted_tokens, list)
         assert all(isinstance(x, str) for x in predicted_tokens)
 
+    def test_uses_named_inputs_with_composed_seq2seq(self):
+        inputs = {
+                "source": "What kind of test succeeded on its first attempt?",
```
brendan-ai2: 😄

allennlp/modules/seq2seq_decoders/stacked_self_attention_decoder_net.py
```diff
@@ -0,0 +1,149 @@
...
+                 positional_encoding_max_steps: int = 5000,
+                 dropout_prob: float = 0.1,
+                 residual_dropout_prob: float = 0.2,
+                 attention_dropout_prob: float = 0.1,) -> None:
+
+        super().__init__(decoding_dim=decoding_dim,
+                         target_embedding_dim=target_embedding_dim,
+                         decodes_parallel=True)
+
+        deep_copy = copy.deepcopy
```
brendan-ai2: I think saying `from copy import deepcopy` would be more typical unless I'm missing something.

allennlp/models/encoder_decoders/composed_seq2seq.py
```diff
@@ -0,0 +1,130 @@
...
+        return self._decoder(state, target_tokens)
+
+    @overrides
+    def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
+        """
+        Finalize predictions.
+        """
+        return self._decoder.post_process(output_dict)
+
+    def _encode(self, source_tokens: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
```
brendan-ai2: This function deserves a comment.

allennlp/modules/seq2seq_decoders/decoder_net.py
```diff
@@ -0,0 +1,88 @@
+from typing import Tuple, Dict, Optional
+import torch
+from allennlp.common import Registrable
+
+class DecoderNet(torch.nn.Module, Registrable):
+    # pylint: disable=abstract-method
+    """
+    A ``DecoderNet`` is a ``Module`` that takes as input a result of ``Seq2SeqEncoder`` and returns a
+    new sequence of vectors.
+    Decoder should implement function to perform a decoding step.
```
brendan-ai2: Please say which functions explicitly or just leave this line off. (A bit confusing as-is.)

allennlp/modules/seq2seq_decoders/stacked_self_attention_decoder_net.py
```diff
@@ -0,0 +1,149 @@
...
+import copy
+import math
+from overrides import overrides
+
+import torch
+from torch import nn
+from torch.autograd import Variable
+import torch.nn.functional as F
+
+from allennlp.modules.seq2seq_encoders.bidirectional_language_model_transformer import (subsequent_mask,
```
brendan-ai2: Thanks for reusing this. Could you please credit the Annotated Transformer tutorial for the rest? http://nlp.seas.harvard.edu/2018/04/03/attention.html (If that's indeed where this is from.)

allennlp/modules/seq2seq_decoders/seq_decoder.py
```diff
@@ -0,0 +1,38 @@
...
+        super().__init__()
+        self.target_embedder = target_embedder
+
+    def get_output_dim(self):
+        raise NotImplementedError()
+
+    def get_metrics(self, reset: bool = False) -> Dict[str, float]:
+        raise NotImplementedError()
+
+    def forward(self,
```
brendan-ai2: Since this is an abstract class it's very important that functions have comments describing their purpose.

allennlp/modules/seq2seq_decoders/seq_decoder.py
```diff
@@ -0,0 +1,38 @@
...
+import torch
+from torch.nn import Module
+
+from allennlp.common import Registrable
+from allennlp.modules import Embedding
+
+
+class SeqDecoder(Module, Registrable):
+    """
+    A ``SeqDecoder`` is a base class for different types of Seq decoding modules
```
brendan-ai2: Could you elaborate on this a bit? I think it's important that users reading this and `DecoderNet` be able to immediately understand the purpose of each. You could even some something like, "In contrast SeqDecoder, DecoderNet is intended for..."

allennlp/models/encoder_decoders/composed_seq2seq.py
```diff
@@ -0,0 +1,130 @@
...
+            if not isinstance(source_embedder, Embedding):
+                raise ConfigurationError("Unable to tie embeddings,"
+                                         "Selected source embedder is not an instance of `Embedding`.")
+            source_weight_shape = source_embedder.weight.shape
+            target_weight_shape = self._decoder.target_embedder.weight.shape
+            if  source_weight_shape != target_weight_shape:
+                raise ConfigurationError(f"Unable to tie embeddings due to mismatch in shared weights."
+                                         f"source  -  {source_weight_shape}"
+                                         f"target  -  {target_weight_shape}")
+            source_embedder.weight = self._decoder.target_embedder.weight
```
brendan-ai2: Hmm, this seems potentially broken given that an `Embedding` also can have a `self._projection`. Is it possible to just share the exact same `TokenEmbedder`? Then we wouldn't have to make this specific to `Embedding` either.

allennlp/models/encoder_decoders/composed_seq2seq.py
```diff
@@ -0,0 +1,130 @@
...
+
+        self._source_text_embedder = source_text_embedder
+        self._encoder = encoder
+        self._decoder = decoder
+
+        if self._encoder.get_output_dim() != self._decoder.get_output_dim():
+            raise ConfigurationError(f"Encoder output dimension {self._encoder.get_output_dim()} should be"
+                                     f" equal to decoder dimension {self._decoder.get_output_dim()}.")
+        if tied_source_embedder_key:
+            # Very ugly way to tie Embeddings
```
brendan-ai2: Yeah, this isn't ideal, but it seems worth it regardless. Could you explain this limitation in the comment above for `tied_source_embedder_key`?

allennlp/data/dataset_readers/seq2seq.py
```diff
@@ -56,6 +56,7 @@ def __init__(self,
                  target_token_indexers: Dict[str, TokenIndexer] = None,
                  source_add_start_token: bool = True,
                  delimiter: str = "\t",
+                 max_tokens: Tuple[int, int] = (1000, 1000),
```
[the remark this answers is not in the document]
sai-prasanna: Should there be a flag to choose whether to truncate or skip long sentences.

allennlp/models/encoder_decoders/composed_seq2seq.py
sai-prasanna: @brendan-ai2 Can we in future explore whether tying weights with something like Initializor's regex selection be possible, or is it too niche and generally not applicable?

allennlp/models/encoder_decoders/composed_seq2seq.py
sai-prasanna: I can only set the decoder's embedder to the encoder's embedder  because the decoder embedder might have weights tied already with its output layer. Hence it cannot be overwritten.

Shall I  check if the `get_output_dim()` of both the embedders are same, and then just set the decoder embedder to the encoder key?

allennlp/data/dataset_readers/seq2seq.py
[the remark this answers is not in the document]
brendan-ai2: Having such a flag sounds reasonable. I'd suggest keeping a counter of how often this occurs (either truncating or skipping) and logging it at the end of `_read` so that it doesn't happen silently. (Also, I'd suggest resetting these at the top of `_read` otherwise the logging is confusing when a dataset reader is used multiple times, e.g. on both training and validation sets.) And regardless, there should be some way to disable the truncating or skipping entirely as I mentioned above.

allennlp/models/encoder_decoders/composed_seq2seq.py
brendan-ai2: Weight tying via config does sound interesting, but I can't claim much insight into how generally useful it would be. I'd defer to other maintainers on that. But we can certainly discuss it in another PR or issue! :)

allennlp/models/encoder_decoders/composed_seq2seq.py
brendan-ai2: Right, I didn't mean to reverse the direction that they're tied. My point was just that `self._projection` is currently being ignored which on the surface looks like a bug. Please clarify if I'm wrong!

> Shall I check if the `get_output_dim()` of both the embedders are same, and then just set the decoder embedder to the encoder key?

By that do you mean that you'd remove `tied_source_embedder_key`? I think we should keep it actually. There's no reason why the weights should always be tied just because the output dimension is the same.

sai-prasanna:
@brendan-ai2  Sorry that it took a while to get back to this.
I have addressed the points in your review  and added few more tests.

schmmd:
@brendan-ai2 would you please take another look?

allennlp/data/dataset_readers/seq2seq.py
```diff
@@ -64,27 +66,46 @@ def __init__(self,
...
         with open(cached_path(file_path), "r") as data_file:
             logger.info("Reading instances from lines in file at: %s", file_path)
             for line_num, row in enumerate(csv.reader(data_file, delimiter=self._delimiter)):
                 if len(row) != 2:
                     raise ConfigurationError("Invalid line format: %s (line number %d)" % (row, line_num + 1))
                 source_sequence, target_sequence = row
                 yield self.text_to_instance(source_sequence, target_sequence)
+        if self._source_max_tokens and self._source_max_exceeded:
+            logger.info("In %d instances the source string exceeded the max token limit (%d) got truncated.",
+                        self._source_max_tokens, self._source_max_exceeded)
```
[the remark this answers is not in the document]
sai-prasanna: Nice catch, Fixed it.

sai-prasanna:
@brendan-ai2 Thanks for the review. Getting it into user's eyes would really help us get more clarity.

TBH I am currently using fairseq directly for my tasks because it supports fp16, accumulation, DDP, and dataset reader/tensorization is fast. I couldn't battle test  these abstractions as much as I wanted to.

brendan-ai2:
Merged! Thanks for the feedback about fairseq too. It's good for us to know what features people find important...

PR #2918, opened 2019-06-03
Fix bug in run_with_beaker.py script

kl2806:
Fix bug introduced in #2903, where docker image was not defined.


schmmd:
@kl2806 if you run into many more problems, we can revert #2903 (but you probably worked through all your problems already).

kl2806:
We do need to update this since beaker changed `blueprint` to `image`. I tested it with the newest version of beaker `Beaker v20190521 ('819a0749452c76ad60df93f14e8b6b46129908af')`. Thanks, @schmmd!

PR #2923, opened 2019-06-04
Fixed minor error while calculating span accuracy

abhishek0318:
Inputs to BooleanAccuracy must have the same dimensions. Here, the first argument has dimensions `(batch size, 2)`, while the second argument has dimensions `(batch size, 1, 2)`. Changing `torch.stack` to `torch.cat` makes the dimension of the second argument `(batch size, 2)`.

Note that span accuracy obtained earlier was not incorrect because of the way accuracy was being computed internally.

### DIFF
--- a/allennlp/models/reading_comprehension/bidaf.py
+++ b/allennlp/models/reading_comprehension/bidaf.py
@@ -262,7 +262,7 @@ class BidirectionalAttentionFlow(Model):
             self._span_start_accuracy(span_start_logits, span_start.squeeze(-1))
             loss += nll_loss(util.masked_log_softmax(span_end_logits, passage_mask), span_end.squeeze(-1))
             self._span_end_accuracy(span_end_logits, span_end.squeeze(-1))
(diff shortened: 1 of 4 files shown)


kl2806:
Thanks! Would you mind raising a config error checking that they are the same dimension similar to: https://github.com/allenai/allennlp/blob/master/allennlp/training/metrics/categorical_accuracy.py#L48
and a test that checks for this similar to https://github.com/allenai/allennlp/blob/1938a5af6d1ea3a2d8dccbb831f82990800fa3e1/allennlp/tests/training/metrics/categorical_accuracy_test.py#L79

matt-gardner:
Minor point: config errors should only be used when the error reasonably came from a configuration file problem. That's probably not the case here, and a value error would be more appropriate.

abhishek0318:
> Thanks! Would you mind raising a config error checking that they are the same dimension similar to and a test that checks for this similar to.

Sure. I will do it over the weekend.

abhishek0318:
I have added the checks and the corresponding tests. Please take a look.

PR #2940, opened 2019-06-10
Replace current default stopword list with spaCy's.

davatk:
The current default stopword list has some odd words like "prepatory", "irrespective", "accordance", "subsequent", etc. It's also missing words like "did", "so", "too", "can", and so on. This just replaces the hard-coded default list with one imported from spaCy, using the logic that since the default tokenizer uses spaCy, so should the default stopword list.

The biggest differences, beyond better word choices, are 1) punctuation is no longer filtered, 2) clitics like "'re" and "'ve" start getting filtered, and 3) potentially, a spaCy update could cause behavior to change. 

A regex filter could address (1); it seems odd to combine that behavior with stopword removal. (2), I think, is straightforwardly correct. Re: (3), the list looks like it's been pretty stable. The only changes in the past couple years have involved fixing issues with different kinds of apostrophes. But possibly people would prefer a hardcoded list? Not sure.

### DIFF
--- a/allennlp/data/tokenizers/word_filter.py
+++ b/allennlp/data/tokenizers/word_filter.py
@@ -2,6 +2,8 @@ from typing import List
 import re
 from overrides import overrides
 
+from spacy.lang.en.stop_words import STOP_WORDS
+
 from allennlp.common import Registrable
 from allennlp.data.tokenizers.token import Token
 from allennlp.common.file_utils import read_set_from_file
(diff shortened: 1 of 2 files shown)


matt-gardner:
I'm hesitant to change defaults, because that has the potential to break backwards compatibility.  In this case, though, I don't imagine many people are actually using this at all, so I'm not sure what harm it would do to change it.  It also isn't that hard to just override what's there if you want to...  Does anyone else have any strong opinions on this change?

joelgrus:
I'm sort of in the same place as you -- I don't like breaking backward compatibility, but also I doubt very many people are using this.

matt-gardner:
Thanks @davatk!

PR #2954, opened 2019-06-15
Deprecation warning fix

### LINKED ISSUE #2252: Timeline for deprecating top-level token_embedders?
Specifying token embedders for a BasicTextFieldEmbedder as top-level key-value pairs has been deprecated for awhile now---when are we thinking of removing it for good? Perhaps soon, since the config wizard is out?

Regardless, if we aren't going to immediately remove it, we should probably update the warning https://github.com/allenai/allennlp/blob/3f0953d19de3676ea82e642659fc96d90690e34d/allennlp/modules/text_field_embedders/basic_text_field_embedder.py#L141-L143 to indicate the version it was deprecated, and the version it will be removed.

pidugusundeep:
removed the DeprecationWarning


pidugusundeep:
closing this PR as it has many unwanted commits, **NO** Unwanted commits in the new PR

PR #2955, opened 2019-06-15
DeprecationWarning removed for op-level token_embedders

### LINKED ISSUE #2252: Timeline for deprecating top-level token_embedders?

pidugusundeep:
Fixes #2252


matt-gardner:
Thanks! You need to also remove the check for the `DeprecationWarning` in the test that's failing.

pidugusundeep:
> Thanks! You need to also remove the check for the `DeprecationWarning` in the test that's failing.

Got it, Removed the test and added a PR

pidugusundeep:
> Thanks! You need to also remove the check for the `DeprecationWarning` in the test that's failing.

Got it, Removed the test and added a PR

matt-gardner:
Sorry I wasn't clear - we still want the test for the old format, because we still handle the old format.  We just need to remove the check for the _warning_ from the test, because the warning doesn't appear anymore.  So it's just removing what was line 215, and un-indenting the line below it.

pidugusundeep:
> Sorry I wasn't clear - we still want the test for the old format, because we still handle the old format. We just need to remove the check for the _warning_ from the test, because the warning doesn't appear anymore. So it's just removing what was line 215, and un-indenting the line below it.

Got it attached a PR

matt-gardner:
Thanks!

PR #2964, opened 2019-06-18
Change registered names of scheduler callbacks

### LINKED ISSUE #2963: Nit-picking proposition: ensure all callback names are in imperative mood
This may be a little late, but I thought I should throw this out there before an official release of the callback trainer. It looks like all of the callback registered names are imperative mood (e.g. `log_to_tensorboard`), which I like, except for two: `learning_rate_scheduler` and `momentum_scheduler`. I propose changing their registered names to `schedule_learning_rate` and `schedule_momentum`, respectively. This would make them consistent with the other callback names and also disambiguate the callbacks from the schedulers.

epwalsh:
fixes #2963 

I didn't want to create a bigger diff than necessary, so to start I only changed the registered names and not the names of the `Callback` subclasses and files. Let me know if you want me to change those as well, such as `learning_rate_scheduler.py -> update_learning_rate.py`.


epwalsh:
^ Looks like the failing test is unrelated.. something to do with an HTTP connection error

matt-gardner:
@epwalsh we're seeing that with other PRs too; not sure what the issue is yet.

joelgrus:
I would change the class names and file names too

joelgrus:
you'll also need to update the API docs, in particular this file:

https://github.com/allenai/allennlp/blob/master/doc/api/allennlp.training.callbacks.rst

--

I'm also wondering if we should standardize the names even more

checkpoint -> save_checkpoints
moving_average -> compute_moving_average

as well as get `Callback` out of those class names.

you don't have to do that in this PR, I can do it in another one

epwalsh:
@joelgrus sounds good, I'll just update the docs

PR #2967, opened 2019-06-18
Allow option to only reset some states in _EncoderBase

### LINKED ISSUE #2828: [Feature Request] Splitting Long Sequences and Potential Modifications to Stateful Encoders

Language models are typically trained on copora consisting of long sequences of text. During training, these long sequences typically need to be split into smaller chunks to prevent memory issues (e.g. we need to manually truncate backpropagation), and model state needs to be propagated to subsequent chunks. _Support for splitting long sequences should be added to AllenNLP_. 

Originally, when I spoke to @matt-gardner about this, he recommended to perform this splitting inside a `DatasetReader`. However, to my understanding this approach will cause each chunk to become a seperate ``Instance``, which is problematic since: a. shuffling the data will not preserve the order of the chunks, and b. even if you don't shuffle the data sequences may be split across elements of the batch. _Ideally, data should be shuffleable at the sequence level before it is split into chunks._

As for propagating state, this can currently be achieved using `_EncoderBase` with `stateful=True`, so long as a signal can be emitted to call `reset_states()` at the start of a new sequence. The drawback of this approach is that it requires that all sequences start at the same time, e.g.
```
|-A-|-A-|-A-|-A-|-B-|...
|-C-|pad|pad|pad|-D-|...
                 ^Start of new sequences
```
In this case there will be many chunks which consist mainly of padding, and training will be biased to overfit the long sequence (e.g. sequence A above).

To avoid excess padding, it would be best to start a new sequence as soon as the previous one terminates, e.g.
```
|-A-|-A-|-A-|-A-|-B-|...
|-C-|-D-|...
    ^ D starts as soon as C ends
```
However, there is no appropriate place to call `reset_states()` in this approach. _Ideally, `reset_states()` should be able to reset only some of the internal states for a batch._

The `reset_states()` issue can be handled by allowing the user to pass an optional `ByteTensor` with dimension `(batch_size,)` which specifies which parts of the internal hidden state to reset, e.g. in the example above, you could pass `[0, 1]` when D begins and `[1, 0]` when B begins. Since the parameter is optional, this would not break backwards compatibility.

As for how/where to split sequences, I don't know that I have a good solution. The way I've handled this issue in my own work is to create a special `SplitIterator` that takes as an additional input a list of fields in the instance to split. You can refer to the implementation here for details: https://gist.github.com/rloganiv/bdfa7c18e004b5cd8c04638a6b4a9c70

My proposed approach has a few issues.
1. It is not very `DataIterator`-esque. Instead of implementing a `_create_batches` method, it overrides `__call__`. Maybe it just needs to be its own kind of `Iterator` and this is okay.
2. It requires specific instructions for how to split each kind of `Field`. This is not easy to maintain, and will not work if a user creates their own custom `Field`. Maybe `Field.__getitem__` methods should be added which allows slicing and returns a `Field` of the same type?

I'd love to hear if anyone has better ideas/alternative approaches for how to handle this issue before I start working on a PR.

rloganiv:
This PR adds the ability for only some states to be reset when `_EncoderBase.reset_states()` is called, which can be useful when not all sequences in a batch terminate. Closes #2828.


matt-gardner:
@DeNeutoy, who do you think is the right person to review this?  I'm happy to do it, but I'm not the most familiar with this part of the code.

PR #2968, opened 2019-06-18
experimental: add backoff

joelgrus:
Our tests have been consistently failing all day because of intermittent failed `requests.get` and `requests.head` calls to S3. It seems like these failures are rooted in AWS denying the requests (on account of volume / frequency?), so this PR implements retries with backoffs for all of our `requests` calls


joelgrus:
success!

brendan-ai2:
Sweet!

PR #2969, opened 2019-06-18
Add in-memory etag cache

brendan-ai2:
- We make HEAD requests on every call to `cached_file` to obtain etags.
- It looks as if S3 is throttling these now.
- Cache etags for duration of process to mitigate.

### DIFF
--- a/allennlp/common/file_utils.py
+++ b/allennlp/common/file_utils.py
@@ -7,7 +7,6 @@ import logging
 import shutil
 import tempfile
 import json
-import time
 from urllib.parse import urlparse
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


epwalsh:
Do you think it's worth caching the etags for `s3://`-style paths as well?

brendan-ai2:
Good call @epwalsh, done. I've also followed @joelgrus 's offline advice to drop the expiry period. We may not really need this now that Joel's https://github.com/allenai/allennlp/pull/2968 is in, but I'll keep it around for a few days in case the failures return sporadically.

schmmd:
Closing--since Joel had a different fix and so this isn't needed presently.

PR #2972, opened 2019-06-19
use starting offsets in the srl model so output is wellformed

DeNeutoy:
This PR does 3 mildly inter-related things:

1. Removes some unnecessary index shifting (randomly adding and minusing one everywhere I was using wordpieces)

2. Fixes a problem with decoding using end indices in the SRL model, which caused tags for a word which was split into multiple word pieces to produce an ill formed BIO sequence.

3. Fixes a bug/missing functionality in `viterbi_decode` where it was impossible to specify start/end transitions.


allennlp/models/semantic_role_labeler.py
```diff
@@ -226,6 +228,27 @@ def get_viterbi_pairwise_potentials(self):
...
+            The pairwise potentials between a START token and
+            the first token of the sequence.
+        """
+        all_labels = self.vocab.get_index_to_token_vocabulary("labels")
+        num_labels = len(all_labels)
+
+        start_transitions = torch.zeros(num_labels)
+
+        for i, label in all_labels.items():
+            if label[0] == "I":
```
matt-gardner: Are we guaranteed to always be using BIO here?

allennlp/models/srl_bert.py
```diff
@@ -136,6 +137,17 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
         Does constrained viterbi decoding on class probabilities output in :func:`forward`.  The
         constraint simply specifies that the output tags must be a valid BIO sequence.  We add a
         ``"tags"`` key to the dictionary with the result.
+
+        NOTE: First, we decode a BIO sequence on top of the wordpieces. This is important; viterbi
+        decoding produces low quality output if you decode on top of word representations directly,
+        because the model already learns very strong preferences for the BIO tag type.
```
matt-gardner: This last phrase isn't clear to me.

allennlp/data/dataset_readers/semantic_role_labeling.py
```diff
@@ -147,26 +140,49 @@ def __init__(self,
...
         BERT CLS and SEP tokens to the begining and end of the sentence.
+
+        A slight oddity with this function is that it also returns the wordpiece offsets
+        corresponding to the _start_ of words as well as the end.
+
+        We need both of these offsets (or at least, it's easiest to use both), because we need
+        to convert the labels to tags using the end_offsets. However, when we are decoding a
+        BIO sequence inside the SRL model itself, it's important that we use the start_offsets,
+        because otherwise we might select an ill-formed BIO sequence from the BIO sequence on top of
+        wordpieces (this happens in the case that a word is split into multiple word pieces,
```
matt-gardner: I think this comment would be more clear with a simple example.

allennlp/models/srl_bert.py
```diff
@@ -136,6 +137,17 @@ def decode(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
...
+
+        Secondly, it's important that the indices we use to recover words from the wordpieces are the
+        start_offsets (i.e offsets which correspond to using the first wordpiece of words which are
+        tokenized into multiple wordpieces) as otherwise, we might get an ill-formed BIO sequence
```
matt-gardner: I'm a bit confused about what's going on here, probably because I'm lacking context on how word tags are split into wordpiece tags in the first place.  If you get a bunch of `I-` tags from taking the last wordpiece, wouldn't you get a bunch of `B-` tags from taking the first one?  Or is it that all continuation wordpieces always have `I-`, but the first wordpiece will have the original word tag?  Then this makes more sense.

Unless you're using BIOUL, then this still gets messed up, because you might have `L-`, `I-`...

allennlp/nn/util.py
```diff
@@ -421,6 +423,14 @@ def viterbi_decode(tag_sequence: torch.Tensor,
         other, or those transitions are extremely unlikely. In this situation we log a
         warning, but the responsibility for providing self-consistent evidence ultimately
         lies with the user.
+    allowed_start_transitions : torch.Tensor, optional, (default = None)
+        An optional tensor of shape (num_tags,) describing which tags the START token
+        may transition TO. If provided, additional transition constraints will be used for
```
matt-gardner: Probably clearer to use `*to*` instead of `TO`.  Same below.

allennlp/models/semantic_role_labeler.py
DeNeutoy: Yes 👍

allennlp/models/srl_bert.py
DeNeutoy: Yes, it's basically that we need the `B-XXX` tag and not the `I-XXX` tag for words that have been split into multiple wordpieces.

allennlp/nn/util.py
DeNeutoy: done

allennlp/models/srl_bert.py
DeNeutoy: Basically if you select out the distributions corresponding to words from the wordpiece sequence and then run inference, the model gets really messed up because it's trained to do BIO tagging on wordpieces - not words.

matt-gardner: I see, ok.

DeNeutoy: I clarified the comment

PR #2973, opened 2019-06-19
gradient accumulation in the `train_supervised` callback

joelgrus:
this is the simplest possible approach, with fixed number of batches.

I think I got everything correct.


DeNeutoy:
Actually I can see this is a real spanner in the works for the callback based trainer, I can't think of an elegant way to deal with this

joelgrus:
hmm, let me think on it some

PR #2979, opened 2019-06-20
Fixes bug in Checkpointer when _num_serialized_models_to_keep is 0 

abhishek0318:
When _num_serialized_models_to_keep is 0, first condition evaluates to false and the checkpointer doesn't remove any previous serialised models.


DeNeutoy:
This is the same as #2880, let's choose one

abhishek0318:
> the default value is 20, so how would the value ever be None?
> I think it's probably OK to just delete that part of the check?

Looks correct to me.
 
> This is the same as #2880, let's choose one

#2880 has relevant test as well. You can close this one.

PR #2983, opened 2019-06-20
Add ERROR callback event

### LINKED ISSUE #2978: Feature request: callback event for training failure
It's nice to get notifications when training completes, which is handled nicely by the `post-to-url` callback (https://github.com/allenai/allennlp/blob/master/allennlp/training/callbacks/post_to_url.py#L36), but it would also be nice to get notifications when training fails for whatever reason (RuntimeError, OOM error, etc). We could do so by adding a callback like `TRAINING_FAIL` and keeping track of the exception and stack trace in the `Trainer` object, and then have a callback like `post-to-url` handle that event. The one downside I can think of there is that the `TRAINING_FAIL` event could only be fired after the `Trainer` has been initialized, which is after the dataset has been read, vocab has been indexed, etc., so any errors occurring during that setup phase would not trigger that callback.

I think it would be still useful though since it's good to know immediately when a long running training process fails for some reason (for me, it's usually due to some bug that only pops up in a certain batch that takes a while to get to, or an OOM error from bigger inputs that occur later in the epoch). I usually observe the logs directly for a few minutes, at least until the first epoch starts.

Currently I just wrap the `allennlp train` command in a bash script with a "try-catch" block and use `curl` to slack myself when training fails.

epwalsh:
Would close #2978.

This is one way to do it.... the downside is that any errors that occur before the training loop begin (or after) would not be caught.


joelgrus:
two immediate thoughts:

(1) I'm not sure how much I like the name `TRAINING_FAIL`, would it be better if the event was called `ERROR` or something like that?

(2) I definitely don't like that the `train` method has a try/catch block shoved into it. what about the following (cute) approach using decorators (I am not 100% sure this would work, but I think it would)

```
class CallbackTrainer:
   ...

    def handle_errors(method):
        @functools.wraps(method)
        def handled(self, *args, **kwargs):
            try:
                method(self, *args, **kwargs)
            except Exception as exc:
                ...
                self.handler.fire_event(Events.ERROR)
        return handled                

    @handle_errors
    def train(self) -> Dict[str, float]:
        ...
```

joelgrus:
I mean, a less cutesy way of doing the same would be to rename `train` to `_train` or something and then do

```
def train(self) -> Dict[str, Any]:
    try:
        return self._train()
    except Exception as exc:
        ...
        raise
```

epwalsh:
I like the idea of the cutesy approach but wasn't sure about a good way to implement it. I started going down a rabbit hole then just decided to keep it simple for now with a private `self._train()`.

joelgrus:
@epwalsh actually I made more changes than that, I implemented the decorator way and I added a test. if you take a look and give me a lgtm I'll merge this.

joelgrus:
@DeNeutoy check out this new version, it doesn't affect the stack traces at all now

epwalsh:
@joelgrus am a little concerned about using the `__del__` method to reset `sys.excepthook`, since it's not obvious when or even if `Trainer.__del__` will be called. I ran two quick experiments, one which raises an error during the training loop, and one which doesn't. In both cases `Trainer.__del__` was never called.

Maybe we could use the `Trainer` as a context manager instead:

```python
class CallbackTrainer(TrainerBase):
    ...
    def __enter__(self):
        sys.excepthook = self._excepthood

    def __exit__(self):
        sys.excepthook = sys.__excepthook__
```

And then in `allennlp.commands.train.train_model` we would do

```python
with Trainer.from_params(...) as trainer:
    ...
```

There's also the caveat with this approach that if someone were to write their own CLI / script to train a model and put `trainer.train()` within a broad `try-catch` block for some reason, the `ERROR` event would never be called unless they called it manually.

I do like the idea of not modifying the stack trace at all, but even with the previous approach (with the private `._train()` method) the `self.exception` object actually holds the original stack trace.

joelgrus:
ok, here's another bad solution, CallbackTrainer.train is unchanged, the constructor checks whether any of its callbacks are listening for an ERROR event, and *only if that's the case* it replaces `train()` with an error-handling version.

on one hand this is kind of nasty, on the other hand if you're not handling errors there's no difference

epwalsh:
:white_check_mark: I would submit an approval but I can't since it's my own PR

joelgrus:
@DeNeutoy is the other person who expressed reservations, I'd like to get his opinion as well

schmmd:
@DeNeutoy what's your opinion on this PR?  Pinging to follow up on @joelgrus's last comment.

schmmd:
@epwalsh I can marge if you give me the thumbs up (just wanted to check in since this has hung around for a while).  It'd be great if you could resolve the conflicts--they are very minor but you would have more perspective on the right resolution.  Sorry for the delay here!

epwalsh:
@schmmd I'll take a look now

epwalsh:
Ok should be good now :+1:

schmmd:
Thanks!

PR #2991, opened 2019-06-22
Fix error in BooleanAccuracy when total count is 0

### LINKED ISSUE #2822: Avoid divide by zero in Boolean Accuracy
**Describe the bug**
When you use a boolean accuracy metric. If you call reset but have not run the model yet, there is a float division by 0 error here https://github.com/allenai/allennlp/blob/master/allennlp/training/metrics/boolean_accuracy.py#L78. 
Similar issues have been fixed in past PR's, e.g., here https://github.com/allenai/allennlp/commit/59132d2be0e871b512c03a08fe80eb41ef7418a9

**Expected behavior**
If you call reset() it should do nothing if the metrics are still 0.

abhishek0318:
Fixes #2822

### DIFF
--- a/allennlp/training/metrics/boolean_accuracy.py
+++ b/allennlp/training/metrics/boolean_accuracy.py
@@ -84,7 +84,10 @@ class BooleanAccuracy(Metric):
         -------
         The accumulated accuracy.
         """
-        accuracy = float(self._correct_count) / float(self._total_count)
+        if self._total_count > 1e-12:
+            accuracy = float(self._correct_count) / float(self._total_count)
+        else:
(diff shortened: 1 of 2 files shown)


allennlp/training/metrics/boolean_accuracy.py
```diff
@@ -84,7 +84,10 @@ def get_metric(self, reset: bool = False):
         -------
         The accumulated accuracy.
         """
-        accuracy = float(self._correct_count) / float(self._total_count)
+        if self._total_count > 1e-12:
```
brendan-ai2: Can't you just test this against 0?

PR #3002, opened 2019-06-25
Fixes inconsistent resetting of metrics with Validate and TrackMetrics callbacks  (issue #3001)

### LINKED ISSUE #3001: Validate callback and Track Metrics callback not resetting metrics consistently
**Describe the bug**
When using the callback trainer with validate and track_metrics callbacks, the metrics are not getting reset properly.

**To Reproduce**
1. Use callback trainer with validate and track_metrics callback
2. compute some metric like boolean_accuracy
3. add a log or print message in the reset function of the metric
4. the reset happens twice, both immediately after validation. The 2nd time reset happens, since the validation metrics have already been reset boolean accuracy throws a division by zero error

**Expected behavior**
Metrics should have been reset once after training finishes and before validation begins and once after validation finishes

 - OS: Linux
 - Python version: 3.7
 - AllenNLP version: I installed from master

**Additional context**
After some debugging I noticed that the execution flows like this:
callback_trainer fires VALIDATE event
validate function is one of the handlers of this event which resets the validation metrics after validation is finished
https://github.com/allenai/allennlp/blob/9a13ab570025a0c1659986009d2abddb2e652020/allennlp/training/callbacks/validate.py#L83-L86

collect_metrics is another handler for the VALIDATE event and has priority 100, so it seems to be getting executed after validate. It also tries to reset the metrics, but right now the model is still in eval() mode, so it resets the same validation metrics again.
https://github.com/allenai/allennlp/blob/9a13ab570025a0c1659986009d2abddb2e652020/allennlp/training/callbacks/track_metrics.py#L93-L96

I don't even see where the training metrics are supposed to be reset. Ideally before the validation starts, the training metrics should already have been collected and reset, so that they don't interfere with val_metrics. That is the behaviour in the default trainer as well, but here it is inconsistent.

vikigenius:
This is a workaround that I am currently using for issue #3001.
Just changed the priorities a bit to make it work.

Another minor fix that I added is to assign trainer.train_metrics instead of update in train_supervised.py.
Originally that was causing cpu memory information to appear in tqdm which made it too long

### DIFF
--- a/allennlp/tests/training/callback_trainer_test.py
+++ b/allennlp/tests/training/callback_trainer_test.py
@@ -10,6 +10,7 @@ from typing import Dict, Iterable
 import torch
 import responses
 import pytest
+import numpy as np
 
 from allennlp.common.checks import ConfigurationError
 
@@ -348,7 +349,8 @@ class TestCallbackTrainer(ModelTestCase):
 
     def test_training_metrics_consistent_with_and_without_validation(self):
         default_callbacks = self.default_callbacks(serialization_dir=None)
-        default_callbacks_without_validation = [callback for callback in default_callbacks if type(callback).__name__ != 'Validate']
+        default_callbacks_without_validation = [callback for callback in default_callbacks
+                                                if not isinstance(callback, Validate)]
         trainer1 = CallbackTrainer(copy.deepcopy(self.model), copy.deepcopy(self.optimizer),
                                    callbacks=default_callbacks_without_validation,
                                    num_epochs=1, serialization_dir=None)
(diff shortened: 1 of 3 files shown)


joelgrus:
thanks, a couple of comments

(1) can you give the PR a more descriptive name, our convention is that the PR title should describe what it does, and then in the description it can say "fixes #3001" or whatever
(2) can you add a test that checks that this is fixed (and that prevents against reversions)
(3) I'm not super thrilled with the 0 -> 50 -> 100 priorities, I think it would be better (and still work) if the `collect_train_metrics` had priority -100 (e.g. "before validation") and the validation itself had priority 0 (which is much less surprising)

vikigenius:
@joelgrus Thanks for your comments, I can definitely change the priorities as you suggested and change the commit message. However, I a, struggling to test the behaviour in a reliable way. 

The idea is that train_metrics have to be collected and reset before validation starts, however both these handlers are handling the same event, just with different priorities, how can I write a test to ensure this behavior any suggestions ?

vikigenius:
Done. I just added two tests to check if the training metrics agree with and without the Validate callback and if the validation metrics agree with and without the TrackMetrics callback.

PR #3003, opened 2019-06-25
Add two optimize parameters for BeamSearch

wlhgtc:
###  1. stochastic_beam_search
The basic version of beam search can get stuck in local maximum as well. To help avoid this, stochastic beam search picks children with probability relative to their values. 
I "rewrite" a new class method `_sample_topk` and confirm the style same as `torch.topk`, if this parameter is True, use `_sample_topk` to replace `torch.topk`

### 2. score_normalize 
Original beam search tends to choose shorter sequence ( cause the score will drop as the length rise).So devide the score by their length may help avoid this. To avoid the devide by 0, we set the intial length as 1

### DIFF
--- a/allennlp/nn/beam_search.py
+++ b/allennlp/nn/beam_search.py
@@ -1,16 +1,14 @@
 from typing import List, Callable, Tuple, Dict
 import warnings
-
 import torch
 
 from allennlp.common.checks import ConfigurationError
-
 
 StateType = Dict[str, torch.Tensor]  # pylint: disable=invalid-name
 StepFunctionType = Callable[[torch.Tensor, StateType], Tuple[torch.Tensor, StateType]]  # pylint: disable=invalid-name
 
 
-class BeamSearch:
+class BeamSearchX:
     """
     Implements the beam search algorithm for decoding the most likely sequences.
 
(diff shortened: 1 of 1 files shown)


epwalsh:
Hey @wlhgtc, could you provide a reference for stochastic beam search? And maybe add it to the docstring for that parameter (similar to how there's a reference for the `per_node_beam_size` parameter)? Also it looks like you introduced a few linting errors by adding and removing whitespace :wink:

wlhgtc:
@epwalsh  Thanks for your reply.
But unfortunately, I couldn't find(or provide) any materials that confirm it's useful(my implementation).
I try to find this concept by keywords, there are 2 different sources:
1. "Stochastic Beams and Where to Find Them:The Gumbel-Top-k Trick for Sampling Sequences Without Replacement " It's accepted by ICML 2019, but different from my implication. It looks more  complicated. They add a Gumbel noise to the log prob and fit ”Gumbel-Max trick" to Beam Search.
2. My implication is based on Wiki (https://en.wikipedia.org/wiki/Beam_search).
 And the details about it could be find in:
[http://www-users.cselabs.umn.edu/classes/Fall-2017/csci4511/slides/week4/9.28.17.pdf](url)(P17).
It simply sample the item according to their prob(or log prob). Though, there are no compelling experiments, intuitively it works!

By the way, what do you mean by "linting errors" , redundant blank line ?

epwalsh:
@wlhgtc if you go to the TeamCity build and log in as a guest you can see where the build failed

wlhgtc:
@epwalsh 
After reading the paper below, I find my implement is worse than their method. But it's hard to rewrite this method according to their paper(even with their code).
Maybe you could finish it :<
So I close this PR

PR #3015, opened 2019-06-27
PassThroughIterator

rloganiv:
This PR adds a new `PassThroughIterator` which tensorizes `Instances` one at a time, and returns them in the exact order that they are created in the `DatasetReader`. It generalizes the `LanguageModelIterator` written by @nelson-liu in #2414, which is specifically designed for the task of contiguous language modeling. Since it seems like this approach is generally useful for problems which apply stateful models to encode long sequences (see discussion #2828), and #2414 is currently blocked by the tangential (and rather thorny) issue #2373, I think it is worth adding this seperately.

The only non-trivial aspect of this iterator is that it needs to remove the batch dimension introduced when calling `Batch.as_tensor_dict()` (since this iterator is intended to be used in situations where batching performed ahead of time within the `DatasetReader`). To do this, I've written a function which recursively squeezes the first dimension of tensors in a `TensorDict` (see [here](https://github.com/rloganiv/allennlp/blob/551575f536b31f9289b1e8b18501d0dde30e304d/allennlp/data/iterators/pass_through_iterator.py#L15-L24)). While I think the function behaves sensibly on tensors and dictionaries, I am not so sure about non-tensor fields like `MetadataField` or `ProductionRuleField`. Right now I am just returning them exactly as is, but maybe if they are singleton lists it would be better to return the single element?

### DIFF
--- a/allennlp/data/iterators/pass_through_iterator.py
+++ b/allennlp/data/iterators/pass_through_iterator.py
@@ -1,0 +1,81 @@
+from typing import Dict, Iterable, Iterator, Union
+import logging
+
+import torch
+
+from allennlp.data.dataset import Batch
+from allennlp.data.instance import Instance
+from allennlp.data.iterators import BasicIterator
+from allennlp.data.iterators.data_iterator import DataIterator, TensorDict
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+def _remove_batch_dim(singleton: Union[TensorDict, torch.Tensor]) -> TensorDict:
+    """Recursively removes the batch dimension from tensors in a TensorDict."""
+    if isinstance(singleton, dict):
+        return {key: _remove_batch_dim(value) for key, value in singleton.items()}
+    elif isinstance(singleton, torch.Tensor):
+        return singleton.squeeze(0)
+    # TODO(rloganiv): Not sure if this is appropriate for Fields whose as_tensor and batch_tensor
+    # methods do not return DataArrays (e.g. MetadataField and ProductionRuleField).
+    else:
+        return singleton
+
+
+@DataIterator.register("pass_through")
+class PassThroughIterator(BasicIterator):
+    """
+    An iterator which performs no batching or shuffling of instances, only tensorization. E.g,
+    instances are effectively passed 'straight through' the iterator.
+
+    This is essentially the same as a BasicIterator with shuffling disabled, the batch size set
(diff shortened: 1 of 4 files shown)


allennlp/data/iterators/pass_through_iterator.py
```diff
@@ -0,0 +1,84 @@
...
+    @overrides
+    def __call__(self,
+                 instances: Iterable[Instance],
+                 num_epochs: int = None,
+                 shuffle: bool = False) -> Iterator[TensorDict]:
+        if shuffle:
+            logger.warning("PassThroughIterator does not shuffle instances. If shuffling is "
+                           "required, please implement in your DatasetReader.")
+            shuffle = False
+        for tensor_dict in super().__call__(instances, num_epochs, shuffle):
```
matt-gardner: You could make this even simpler and remove the need for `_remove_batch_dim` by just doing:

```python
def __call__(self, instances, num_epochs, shuffle):
    for epoch in num_epochs:  # handle num_epochs == None here
        for instance in instances:
            yield instance.as_tensor_dict()
```

This means you don't get caching or epoch tracking, but it simplifies a lot of other things.  I'm not sure whether we should do it this way or not, just something to think about.

rloganiv: I like this approach. It handles the issue `_remove_batch_dim` has with non-tensor inputs. Also caching will probably not be needed in most use cases since it is expected that the dataset reader will perform actions like shuffling, perturbing sequence lengths, etc.

allennlp/data/iterators/pass_through_iterator.py
```diff
@@ -0,0 +1,48 @@
...
+    This is essentially the same as a BasicIterator with shuffling disabled, the batch size set
+    to 1, and maximum samples per batch disabled. The only difference is that this iterator
+    removes the batch dimension. This can be useful for rare situations where batching is best
+    performed within the dataset reader (e.g. for contiguous language modeling, or for other
+    problems where state is shared across batches).
+    """
+    def __init__(self):
+        super().__init__(batch_size=1)
+
+    def _create_batches(self, instances: Iterable[Instance], shuffle: bool) -> Iterable[Batch]:
```
matt-gardner: Keeping an `@overrides` tag here would be nice.  The point is that it makes it obvious to the reader why this method exists.

nelson-liu:
@rloganiv thanks for this! Would love your thoughts on the (thorny indeed) #2373 / unblocking #2414 , if you have any

PR #3021, opened 2019-06-29
Multilabel bug

### LINKED ISSUE #2415: MultiLabelField's `empty_field` method always fails
**Describe the bug**
MultiLabelField's `empty` method always fails, preventing it from being used with padding.

**To Reproduce**
```python
from allennlp.data.fields import MultiLabelField
f = MultiLabelField([])
f.empty_field()
```
Prints error:
```
  File "/Users/julian/.dotfiles/local/miniconda3/lib/python3.6/site-packages/allennlp/data/fields/multilabel_field.py", line 120, in empty_field
    return MultiLabelField([], self._label_namespace, skip_indexing=True)
  File "/Users/julian/.dotfiles/local/miniconda3/lib/python3.6/site-packages/allennlp/data/fields/multilabel_field.py", line 68, in __init__
    raise ConfigurationError("In order to skip indexing, num_labels can't be None.")
allennlp.common.checks.ConfigurationError: "In order to skip indexing, num_labels can't be None."
```
Also,
```python
f = MultiLabelField([], skip_indexing = True, num_labels = 4)
f.empty_field()
```
throws the same error.

**Expected behavior**
Basically I think the implementation of `empty_field(self)` should be something like:
```python
return MultiLabelField([], self._label_namespace, skip_indexing = self._num_labels is not None, self._num_labels)
```
This way it would bypass the indexing step when possible (i.e., `num_labels` is already known, for example if we're producing an empty field from an already-indexed one), but always do it when necessary (i.e., when `num_labels` is not already known).

 - OS: OSX
 - Python version: 3.6.5
 - AllenNLP version: v0.8.0

**Additional context**
This makes it impossible to use MultiLabelField inside ListField. Surprising that nobody has come across this before but I guess MultiLabelField is not often used...

codedecde:
Fixes #2415
Basically we don't check the configuration errors if labels is empty

### DIFF
--- a/allennlp/data/fields/multilabel_field.py
+++ b/allennlp/data/fields/multilabel_field.py
@@ -60,7 +60,7 @@ class MultiLabelField(Field[torch.Tensor]):
         self._maybe_warn_for_namespace(label_namespace)
         self._num_labels = num_labels
 
-        if skip_indexing:
+        if skip_indexing and self.labels:
             if not all(isinstance(label, int) for label in labels):
                 raise ConfigurationError("In order to skip indexing, your labels must be integers. "
                                          "Found labels = {}".format(labels))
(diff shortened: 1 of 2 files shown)


schmmd:
LGTM.  @brendan-ai2 would you take a look as well and merge if it looks good?

brendan-ai2:
Looks good to me. Thanks, @codedecde!

PR #3029, opened 2019-07-02
simplify callback trainer

joelgrus:
Last week @DeNeutoy and I had several discussions about the CallbackTrainer being way too complicated and difficult to reason about. I thought about this for a while and decided it was largely because I was trying to make the callbacks do too much work and the trainer too little. Accordingly, here's a pretty substantial proposed revision:

(1) eliminate the `TrainSupervised` and `GenerateTrainingBatches` callbacks and move those functionalities back into the trainer, in a modular, easy-to-override way:

```
def generate_training_batches
def train_one_batch_group
def train_one_epoch
def train
```

These contain enough built-in functionality so that you should be able to train a model without using any callbacks; now callbacks are really for making extra things happen at various points in the training loop. (Notably, `Validation` is still a callback.)

2. make the `MovingAverage` object a first-class part of the trainer, since any callback may potentially want to access it (to load and then restore parameters). This allowed me to

3. eliminate the `SAVE_CHECKPOINT` and `RESTORE_CHECKPOINT` events

I feel like this version is much cleaner and easier to reason about, but I am open to feedback about how it could be simplified further while keeping the callback-based flexibility.


allennlp/training/callback_trainer.py
```diff
@@ -234,7 +300,10 @@ def from_params(cls,  # type: ignore
         parameters = [[n, p] for n, p in model.named_parameters() if p.requires_grad]
         optimizer = Optimizer.from_params(parameters, params.pop("optimizer"))
 
-        model_save_interval = params.pop_float("model_save_interval", None)
+        if "moving_average" in params:
```
DeNeutoy: It's weird to me that `MovingAverage` can't be a callback.

What about if all the callbacks were keyed by name and set as attributes (or even the trainer just has a "_has_moving_average" flag), so you could do the same kind of check in the other `Callbacks` which need to know if there is also a `MovingAverage` callback? Then they could just call the relevant callback and not have to have this in the Trainer. It's not a big deal but it just seems a little out of place.

allennlp/training/callbacks/callback_handler.py
```diff
@@ -75,5 +75,5 @@ def fire_event(self, event: str) -> None:
         """
         for event_handler in self._callbacks.get(event, []):
             if self.verbose:
-                logger.info(f"event {event} -> {event_handler}")
```
DeNeutoy: Switch back to logger

allennlp/training/callback_trainer.py
joelgrus: thanks for the prod, I figured out a way to do this where the trainer doesn't have to know about this at all.

basically the CallbackHandler maintains an index of callbacks by _type_, and then the Validate and Checkpoint callbacks can just ask the handler "do you have any moving average callbacks"

take a look

joelgrus: actually, I changed my mind about that too, now any callback that wants to know about `MovingAverage` objects just iterates over the callbacks looking for them once at `TRAINING_START`

Issue #2949, opened 2019-06-14
Allow to specify dataset reader in Predictor.from_path


mojesty:
**Is your feature request related to a problem? Please describe.**
When training dataset reader uses on-the-fly augmentations and validation dataset reader does not, `Predictor.from_path` does not allow to select right `DatasetReader` from the archive.

**Describe the solution you'd like**
Add string parameter to the `from_path` and `from_archive` functions (default to training dataset reader), search for it in the archive and instantiate selected reader or throw an error if not found.

**Describe alternatives you've considered**
I've written my own wrapper that instantiated DatasetReader and Predictor

matt-gardner:
Hmm, I actually think the right thing to do might be to always load the validation dataset reader if it's there, and fall back to the training dataset reader if not.  You _probably_ want whatever you did for validation when you're loading a predictor.  I've had to modify model archives plenty of times when putting things in the demo because of this.

We could also add a flag that toggles this behavior if you really did want the training dataset reader.  If you want something totally separate, you can just override `predictor._dataset_reader` after it's loaded.

This seems like a reasonable thing to me; I'd say a PR is welcome here.  If anyone else has differing opinions, say so.

nelson-liu:
> Hmm, I actually think the right thing to do might be to always load the validation dataset reader if it's there, and fall back to the training dataset reader if not.

Agreed, I actually thought this was already the default.

matt-gardner:
This is the default for the [`evaluate` command](https://github.com/allenai/allennlp/blob/e7b0013002d3f3272dcd470b34cd686e6ef1c52b/allennlp/commands/evaluate.py#L137-L143), which is probably what you were thinking of.  It probably should also be the default for `Predictors` in general.

danieldeutsch:
This was addressed in https://github.com/allenai/allennlp/pull/3033

nelson-liu:
Thanks @danieldeutsch !

PR #3033, opened 2019-07-04
Adding ability to choose validation DatasetReader with "predict"

danieldeutsch:
Here is my proposal for #2949. It is a non-breaking change. It changes the `--use-dataset-reader` flag from a boolean to a string with an optional value which specifies which reader to use and defaults to "train".

I am not thrilled with the testing, but short of creating a custom model archive just for this change, I could not think of a better solution.

If you all are happy with this, I will make it style-compliant, etc.


matt-gardner:
If we want this to be backwards compatible for the behavior of that flag, I'm ok with this implementation.  I would probably go a bit farther, though, and make the default value be `validation` instead of `train` (and change the order that you check things when loading, to fall back to train if validation is not present) - it just makes more sense, and it's a very minor point for backwards incompatibility.

schmmd:
@danieldeutsch what do you think about @matt-gardner 's question?

danieldeutsch:
I am ok with whatever you all think is best.

One issue with my implementation that I found in my own scripts is that `argparse` can't correctly parse the command if the `--use-dataset-reader` flag comes just before the config file because it interprets the config file name as the argument for `--use-dataset-reader`. For example `allennlp predict --use-dataset-reader config.jsonnet input.jsonl`

I can think of a few ways around this. (1) We make the `train` or `validation` value to `--use-dataset-reader` required, but this would be a breaking change to anyone who uses the flag. (2) We make `--use-dataset-reader` use the validation reader by default and introduce a second argument that can override this decision. (3) We introduce a new flag `--use-validation-dataset-reader` and leave the original flag alone.

What do you all think?

matt-gardner:
My preference would be to just make the behavior of `predict` identical to the behavior of `evaluate`: https://github.com/allenai/allennlp/blob/083f3430b8cee77e3c1c553f4cba70ad8f3594b1/allennlp/commands/evaluate.py#L137-L146

That is, always use the validation dataset reader if it's available, and fall back to the training one, with no option to override that.

danieldeutsch:
I disagree with no option to override the choice. The specific scenario that I need this for is transfer learning in which the training and testing tasks are different. I still want to be able to run `predict` in the training scenario as well as the testing scenario so I can measure how well the model does on the training task on a held-out dataset.

matt-gardner:
Ok, fair enough.  I was going to wait for someone specifically requesting the ability to change it to add the feature; looks like that's now.  Then I would vote for your option (2).

danieldeutsch:
Sorry for the delay.

I switched the default dataset reader to the validation one and added a new flag to override that choice. One thing which could be nice (but not implemented) is to not quietly fall back to the training dataset reader if the user explicitly requests to use a validation dataset reader which does not exist.

PR #3035, opened 2019-07-06
Avoid divide by zero in boolean_accuracy.py

Eric-Wallace:
If you call model.get_metrics(reset=True) before saving any metric this function throws an error.


Eric-Wallace:
@brendan-ai2 here is that PR finally.

matt-gardner:
Someone beat you to it, but with a test: #2991.  But you also used `1e-12` instead of just comparing to 0.  Why?

Eric-Wallace:
People had done that in other accuracy files, so I copied them. Probably is not needed.

PR #3036, opened 2019-07-06
Add options for focal loss

guoquan:
Add focal loss to deal with class imbalance

### DIFF
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -8,6 +8,7 @@ import logging
 import copy
 import math
 import json
+import numpy
 
 import torch
 
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -701,9 +701,9 @@ def sequence_cross_entropy_with_logits(logits: torch.FloatTensor,
         probs_flat = torch.gather(probs_flat, dim=1, index=targets_flat)
         # shape : (batch * sequence_length,)
         focal_factor = (1. - probs_flat) ** gamma
-        focal_factor = focal_factor.view(-1, 1)
-    else:
-        focal_factor = 1.
+        focal_factor = focal_factor.view(*targets.size())
+        weights = weights * focal_factor
+
     if alpha is not None:
         # shape : () / (num_classes,)
         if isinstance(alpha, (float, int)):
@@ -728,8 +728,6 @@ def sequence_cross_entropy_with_logits(logits: torch.FloatTensor,
         # shape : (batch, max_len)
         alpha_factor = torch.gather(alpha_factor, dim=0, index=targets_flat.view(-1)).view(*targets.size())
         weights = weights * alpha_factor
-    else:
-        alpha_factor = 1.
 
     if label_smoothing is not None and label_smoothing > 0.0:
         num_classes = logits.size(-1)
@@ -745,8 +743,7 @@ def sequence_cross_entropy_with_logits(logits: torch.FloatTensor,
         # to extract the indices of the num_classes dimension which contribute to the loss.
         # shape : (batch * sequence_length, 1)
         negative_log_likelihood_flat = - torch.gather(log_probs_flat, dim=1, index=targets_flat)
-    # shape : (batch * sequence_length, 1)
-    negative_log_likelihood_flat = negative_log_likelihood_flat * focal_factor
+
     # shape : (batch, sequence_length)
     negative_log_likelihood = negative_log_likelihood_flat.view(*targets.size())
     # shape : (batch, sequence_length)
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -703,11 +703,15 @@ def sequence_cross_entropy_with_logits(logits: torch.FloatTensor,
     if alpha is not None:
         # shape : () / (num_classes,)
         if isinstance(alpha, float):
+            # pylint: disable=not-callable
             # shape : (2,)
             alpha_factor = torch.tensor([1. - alpha, alpha], device=targets_flat.device)
+            # pylint: enable=not-callable
         elif isinstance(alpha, list):
+            # pylint: disable=not-callable
             # shape : (c,)
             alpha_factor = torch.tensor(alpha, device=targets_flat.device)
+            # pylint: enable=not-callable
         elif isinstance(alpha, torch.Tensor):
             # shape : () / (num_classes,)
             alpha = alpha.clone().detach().to(device=targets_flat.device)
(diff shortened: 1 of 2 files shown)


guoquan:
Sorry, it seems there are some problems. But it works fine in my local copy.
Can anyone send me the error message from CI?

matt-gardner:
You can see it yourself here: http://build.allennlp.org/viewLog.html?buildId=17043&buildTypeId=AllenNLP_AllenNLPPullRequests&tab=buildLog.  Just log in as a guest.

guoquan:
Thanks. Did not see the guest option. Will get back when I figure this out.

guoquan:
Finally here. Not an easy job to make these checkers happy.

allennlp/nn/util.py
```diff
@@ -667,12 +685,51 @@ def sequence_cross_entropy_with_logits(logits: torch.FloatTensor,
...
+                alpha_factor = alpha_factor.view(1)
+                # shape : (2,)
+                alpha_factor = torch.cat([1 - alpha_factor, alpha_factor])
+        else:
+            raise TypeError(('alpha must be float, list of float, or torch.FloatTensor, '
+                             '{} provided.').format(type(alpha)))
+        # shape : (batch, max_len)
+        alpha_factor = torch.gather(alpha_factor, dim=0, index=targets_flat.view(-1)).view(*targets.size())
+        weights = weights * alpha_factor
+    else:
```
matt-gardner: I don't think you need this `else`.

allennlp/nn/util.py
```diff
@@ -667,12 +685,51 @@ def sequence_cross_entropy_with_logits(logits: torch.FloatTensor,
...
+    # focal loss coefficient
+    if gamma:
+        # shape : (batch * sequence_length, num_classes)
+        probs_flat = log_probs_flat.exp()
+        # shape : (batch * sequence_length,)
+        probs_flat = torch.gather(probs_flat, dim=1, index=targets_flat)
+        # shape : (batch * sequence_length,)
+        focal_factor = (1. - probs_flat) ** gamma
+        focal_factor = focal_factor.view(-1, 1)
+    else:
```
matt-gardner: I think you can similarly modify `weights` as the last line of the `if` above, and not need this `else`, or need to reference `focal_factor` down below.

allennlp/tests/nn/util_test.py
```diff
@@ -661,6 +662,7 @@ def test_sequence_cross_entropy_with_logits_averages_batch_correctly(self):
         # Batch has one completely padded row, so divide by 4.
         assert loss.data.numpy() == vector_loss.sum().item() / 4
 
+    @flaky # some time "not almost equal to 7 decimals"
```
matt-gardner: No need for the comment here; that's already the definition of `flaky`.  I'm not sure why the change you made would make this test flaky, though, while the ones below aren't.

allennlp/tests/nn/util_test.py
```diff
@@ -681,6 +683,106 @@ def test_sequence_cross_entropy_with_logits_averages_token_correctly(self):
...
+
+        loss = util.sequence_cross_entropy_with_logits(tensor, targets, weights, gamma=gamma)
+
+        correct_loss = 0.
+        for logit, label in zip(tensor.squeeze(0), targets.squeeze(0)):
+            p = torch.nn.functional.softmax(logit, dim=-1)
+            pt = p[label]
+            correct_loss += pt.log() * (1 - pt) ** gamma
+        # Average over sequence.
+        correct_loss = - correct_loss / length
```
matt-gardner: Slightly clearer to subtract inside the loop and just average here.  Same for all tests below.

allennlp/tests/nn/util_test.py
guoquan: It just happened to fail at a very early stage I try to modify the loss. Did not notice if it is related to my changes. It happens once out of 10 rounds. It really looks irrelevant. I just add this to stop the unit test complaining.
My test is using token-average, as in smooth label test. So it is not affected by the unstable problem when using batch-average, in the above flaky test.

However, I also see once test with the token-average complaint "not almost equal to 7 decimals", out of maybe 50+ times local and TC tests. Please refer to [# 7006](http://build.allennlp.org/viewLog.html?buildId=17137) in TC. Need `@flaky` for it also? This may imply all other tests rely on token-average could be flaky. There are 8-9 calls to token-average which is the default. So I will estimate the chance of flaky is ~1/500.

allennlp/tests/nn/util_test.py
guoquan: Great. Thanks.

allennlp/nn/util.py
guoquan: Yes. I used to have something like `focal_factor` but combining into `weights` looks better. Removed this `else`.

allennlp/nn/util.py
guoquan: Great. Update.

allennlp/tests/nn/util_test.py
guoquan: Sarcasm is that I get flaky immediately with this build http://build.allennlp.org/viewLog.html?buildId=17245. Alright, I will add `@flaky` to all of those.

guoquan:
It seems moving where to combine the weight brings in new problem. Will look at it later.

allennlp/nn/util.py
guoquan: Hi @matt-gardner. I realize there is a problem here because when calculating the average, "token", "batch", or even "none", it relies on `weights.sum()`. Do you think `gamma` (and also `alpha`) should go into the denominator of average `weights.sum()`? It is a kind of design decision, rather than correctness, of how to calculate the average here. What do you think?

allennlp/tests/nn/util_test.py
guoquan: It seems no problem this part as I discuss in the above conversation. The question here is if you want a minimum `@flaky` flags? If so I remove not necessary ones.

matt-gardner: I don't see a problem with having a `@flaky` decorator, I was just surprised it's necessary.

allennlp/nn/util.py
matt-gardner: `alpha` and `gamma` have the effect of changing the weighting on each of the instances / tags in the batch, so it will affect the direction of the gradients.  Changing the denominator in `weights.sum()` changes all directions of the gradient equally, so including `alpha` and `gamma` in the sum effectively just lowers the learning rate.  Right?  If I'm understanding this correctly, I'd probably compute the sum before multiplying by the `alpha` and `gamma` factors, so the effective learning rate doesn't change as much.

guoquan: Right. The outcome would be changing the learning rage if they are included. If this is the case, I can calculate the `sum()` before involving them.

PR #3040, opened 2019-07-08
Spacy token indexer

DeNeutoy:


allennlp/data/fields/text_field.py
```diff
@@ -154,7 +154,8 @@ def as_tensor(self, padding_lengths: Dict[str, int]) -> Dict[str, torch.Tensor]:
             # than a LongTensor here, and it's not clear how to signal that.  Maybe we'll need to
             # add a class method to TokenIndexer to tell us the type?  But we can worry about that
             # when there's a compelling use case for it.
```
matt-gardner: You'll want to remove the TODO above, as you're addressing it here.

DeNeutoy:
@matt-gardner ready for a proper look now, if you want.

allennlp/data/token_indexers/spacy_indexer.py
```diff
@@ -0,0 +1,83 @@
...
+import numpy
+
+from allennlp.common.util import pad_sequence_to_length
+from allennlp.data.vocabulary import Vocabulary
+from allennlp.data.tokenizers.token import Token
+from allennlp.data.token_indexers.token_indexer import TokenIndexer
+
+
+@TokenIndexer.register("spacy")
+class SpacyTokenIndexer(TokenIndexer[numpy.ndarray]):
```
matt-gardner: Might be nice to signal in the name that this is giving you a vector from spacy - `SpacyTokenIndexer` just says I get _something_ from spacy, not what I get, and vectors aren't the first thing I think of with spacy.

allennlp/data/token_indexers/spacy_indexer.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @overrides
+    def tokens_to_indices(self,
+                          tokens: List[SpacyToken],
+                          vocabulary: Vocabulary, # pylint: disable=unused-argument
+                          index_name: str) -> Dict[str, List[numpy.ndarray]]:
+
+        if not all([isinstance(x, SpacyToken) for x in tokens]):
+            raise ValueError("The spacy indexer requires you to use a Tokenizer which produces SpacyTokens.")
+        indices: List[int] = []
```
matt-gardner: Type here is wrong, isn't it?

allennlp/data/token_indexers/spacy_indexer.py
```diff
@@ -0,0 +1,83 @@
...
+        for token in tokens:
+            indices.append(token.vector)
+
+        return {index_name: indices}
+
+    @overrides
+    def get_padding_token(self) -> int:
+        return 0
```
matt-gardner: Shouldn't this be a numpy array with the right size?  Looks like you need to add a test for having some kind of padding with this indexer.  Also, the type annotation is wrong.

allennlp/data/token_indexers/spacy_indexer.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @overrides
+    def get_padding_lengths(self, token: int) -> Dict[str, int]:  # pylint: disable=unused-argument
```
matt-gardner: Type is wrong on `token`.  Looks like mypy isn't actually checking that the overridden methods match the correct instantiation of the type variable...

allennlp/data/token_indexers/spacy_indexer.py
```diff
@@ -0,0 +1,83 @@
...
+        return {}
+
+    @overrides
+    def array_type(self):
+        return torch.FloatTensor
+
+    def _zeros(self):
```
matt-gardner: Type annotation.  Also, looks like this is what you want for `get_padding_token`.  You could move this logic there, then replace the reference to `self._zeros` below with `self.get_padding_token`.

allennlp/data/token_indexers/spacy_indexer.py
```diff
@@ -0,0 +1,83 @@
...
+    @overrides
+    def get_padding_token(self) -> int:
+        return 0
+
+    @overrides
+    def get_padding_lengths(self, token: int) -> Dict[str, int]:  # pylint: disable=unused-argument
+        return {}
+
+    @overrides
+    def array_type(self):
```
matt-gardner: Return type annotation.

allennlp/data/tokenizers/word_splitter.py
```diff
@@ -134,6 +135,26 @@ def _remove_spaces(tokens: List[spacy.tokens.Token]) -> List[spacy.tokens.Token]
     return [token for token in tokens if not token.is_space]
 
 
+class WhitespaceTokenizer:
```
matt-gardner: Looks like this is only for the spacy word splitter?  I'd make it private then, and maybe put it inside the `SpacyWordSplitter` class.  Having it here means it will show up in the docs, which would be confusing.

joelgrus:
I think that in light of yesterday's discussion we should update the docstring for `TokenIndexer` to make clear the distinction:

`TokenIndexer` turns tokens into arrays suitable for feeding into your model (which could be the indexes of the tokens as per some vocabulary, could be as the indexes of the token-characters as per some vocabulary, could be the indexes of subword-pieces as per some vocabulary, or could even be as pretrained low-dimensional embeddings. If you want to learn (or fine-tune) embeddings, that should be handled by a `TokenEmbedder` inside your model that takes indices as inputs. (Even for non-trainable embeddings you could still generate indices here and use a frozen `TokenEmbedder` inside your model, and historically this it how we've done things.)

or something like that

doesn't have to be in this PR though

allennlp/data/token_indexers/token_indexer.py
```diff
@@ -96,6 +100,14 @@ def pad_token_sequence(self,
         """
         raise NotImplementedError
 
+    def array_type(self) -> Type[torch.Tensor]: # pylint: disable=no-self-use
```
joelgrus: why this (indirect) design rather than something more explicit like

```
def make_tensor(self, pre_tensor) -> torch.Tensor:
    """override if you don't want long tensors"""
    return torch.LongTensor(pre_tensor)
```

?

matt-gardner: I was wondering about all of the separate operations done in the interaction between `TextField` and `TokenIndexer`, also.  We could just change `pad_token_sequence` into an `as_padded_tensor` function, or something, and there may be other places we can similarly simplify.  That's probably for a separate PR, though.  Except that we don't really want to introduce a function here that we're going to remove later...

allennlp/data/token_indexers/spacy_indexer.py
DeNeutoy: I think `get_padding_token` is a relic of before we had `pad_token_sequence`, because I don't think any of the `get_padding_token` methods on any of the indexers are used outside of the class themselves

allennlp/data/token_indexers/spacy_indexer.py
DeNeutoy: removed

allennlp/data/token_indexers/spacy_indexer.py
DeNeutoy: fixed

allennlp/data/token_indexers/spacy_indexer.py
DeNeutoy: removed

allennlp/data/token_indexers/spacy_indexer.py
DeNeutoy: thanks!

allennlp/data/token_indexers/token_indexer.py
```diff
@@ -81,18 +93,36 @@ def get_token_min_padding_length(self) -> int:
...
+        then it will be truncated.
 
         ``padding_lengths`` is used to provide supplemental padding parameters which are needed
         in some cases.  For example, it contains the widths to pad characters to when doing
         character-level padding.
+
+        Note that this method should be abstract, but it is implemented to allow backward compatability.
+        """
+        if self.has_warned_for_as_padded_tensor:
+            warnings.warn("Using a Field with pad_token_sequence, which will be depreciated in 1.0.0."
```
matt-gardner: Nit: "deprecated", not "depreciated".  Same in other places.

allennlp/data/token_indexers/token_indexer.py
```diff
@@ -81,18 +93,36 @@ def get_token_min_padding_length(self) -> int:
...
+        of input tokens as a torch Tensor. If the input token list is longer than ``desired_num_tokens``
+        then it will be truncated.
 
         ``padding_lengths`` is used to provide supplemental padding parameters which are needed
         in some cases.  For example, it contains the widths to pad characters to when doing
         character-level padding.
+
+        Note that this method should be abstract, but it is implemented to allow backward compatability.
+        """
+        if self.has_warned_for_as_padded_tensor:
```
matt-gardner: I think you want a `not` here.

Issue #3039, opened 2019-07-07
BidirectionalEndpointSpanExtractor.forward "exclusive_span_ends > sequence_lengths.unsqueeze(-1)" seems bug.


cjopengler:
**Describe the bug**
BidirectionalEndpointSpanExtractor.forward 

```
 if (exclusive_span_starts < 0).any() or (exclusive_span_ends > sequence_lengths.unsqueeze(-1)).any():
            raise ValueError(f"Adjusted span indices must lie inside the length of the sequence tensor, "
                             f"but found: exclusive_span_starts: {exclusive_span_starts}, "
                             f"exclusive_span_ends: {exclusive_span_ends} for a sequence tensor with lengths "
                             f"{sequence_lengths}.")
```

```(exclusive_span_ends > sequence_lengths.unsqueeze(-1)``` should be fixed as ```(exclusive_span_ends >= sequence_lengths.unsqueeze(-1)```.  Because, when ```exclusive_span_ends == sequence_lengths``` will be out of boundary.

schmmd:
@cjopengler makes sense to me.  I created https://github.com/allenai/allennlp/pull/3043, please let me know if that looks good.

In general, pull requests are very welcome.

cjopengler:
> @cjopengler makes sense to me. I created #3043, please let me know if that looks good.
> 
> In general, pull requests are very welcome.

Yeah. Thanks for your encourage. If I got other, I will try to pull requests.

PR #3043, opened 2019-07-08
Fix bug in bidirectional_endpoint_span_extractor.

schmmd:
Fixes https://github.com/allenai/allennlp/issues/3039


schmmd:
@cjopengler the build fails with this change.  Perhaps there should be a special case for 0.  Ideally you would create a PR with a new test that fails before the change and succeeds afterwards.

schmmd:
Thanks for the explanation @DeNeutoy.  I figured it wasn't a bug when this change broke the tests.  I'll modify this PR to add some documentation instead.

PR #3051, opened 2019-07-11
gradient accumulation, try #2

joelgrus:
based on the latest iteration of the callback trainer. I think it's comparatively pretty clean.

unlike the first version, this one simply discards final incomplete batches (with the new design it would be pretty messy to make it behave otherwise)


allennlp/training/callback_trainer.py
```diff
@@ -117,6 +117,15 @@ def __init__(self,
         self.epoch_number = 0
         self.batch_grad_norm: Optional[float] = None
 
+        # If you're doing something like gradient accumulation, you might want
+        # to modify this during training to treat multiple forward / backward
+        # passes as a single "batch".
+        self.is_start_of_batch = True
```
DeNeutoy: These variable names are still pretty obnoxious because they can both be true simultaneously... maybe something like `should_run_epoch_start` and `should_wait_for_grad_step` or something?

HarshTrivedi:
@joelgrus This pr seems almost good to merge. When should it get in? By the time, do you see any problem using this branch for gradient accumulation?

schmmd:
Closing in favor of https://github.com/allenai/allennlp/pull/3512

Issue #2980, opened 2019-06-20
configuration logic for custom from_params methods should live with them


joelgrus:
it's not a good idea (and not sustainable) for `configuration.py` to contain a record of one-off logic for things like `Vocabulary` whose configuration schema can't be inferred from their constructors.

instead, this custom logic should live with the custom from_params methods itself.

one somewhat straightforward idea would be to introduce a decorator `@configuration(config)` for the custom from_params methods that takes in the custom configuration spec and attaches it as a `_configuration` property on the `from_params` method, so that then the config wizard can check if that exists and uses it

matt-gardner:
I like this idea.

DeNeutoy:
+ 1

matt-gardner:
We're instead trying to remove the need for custom `from_params` entirely, instead of doing it this way.  Closing this issue.

PR #3054, opened 2019-07-11
make custom from_params config live with from_params method via a decorator

joelgrus:
a step toward fixing #2980 

this looks like a bigger change than it is, I had to split `configuration.py` into two files to avoid circular references, and most of the +/- is just that.

the primary actual change was to add a `@configuration` decorator and to replace the hardcoded VOCAB_CONFIG, with a use of the decorator on `Vocabulary.from_params`. If people like this approach we can do it for the rest of the classes with custom from_params.

One choice I made that might be objectionable: the configs consist of typed `ConfigItem` NamedTuples, but I decided to allow the decorator annotations to also just be plain tuples, just so it's less verbose.

### DIFF
--- a/allennlp/common/autoconfigure.py
+++ b/allennlp/common/autoconfigure.py
@@ -1,0 +1,203 @@
+# pylint: disable=protected-access,too-many-return-statements
+
+from typing import List, TypeVar, Type, Dict
+import inspect
+import importlib
+
+import torch
+
+from allennlp.common.configuration import (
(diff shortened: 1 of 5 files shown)


matt-gardner:
Didn't look at the code, just the usage in `Vocabulary`, and I like it.

matt-gardner:
I think the approach I outlined in the comments in #3389 is preferable to this approach.  I'm planning on doing that in the near future; I'm going to close this.

PR #3055, opened 2019-07-11
infer allow_unmatched_keys from embedder_to_indexer_map

### LINKED ISSUE #3034: BasicTextFieldEmbedder.allow_unmatched_keys should default to `True` if embedder_to_indexer_map is provided
99.9% of the time, if you supply `embedder_to_indexer_map` you want to allow unmatched keys, so why not just make it default to True if that's the case so you don't have to remember to specify it manually?

joelgrus:
fixes #3034

### DIFF
--- a/allennlp/modules/text_field_embedders/basic_text_field_embedder.py
+++ b/allennlp/modules/text_field_embedders/basic_text_field_embedder.py
@@ -1,4 +1,4 @@
-from typing import Dict, List, Union, Any
+from typing import Dict, List, Union, Any, Optional
 import inspect
 
 import torch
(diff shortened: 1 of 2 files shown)


matt-gardner:
Can we go farther than this and just have each `Indexer` add something to the output of `TextField`?  Seems like there are two options for that:

1. Have the actual output type of `TextField.as_tensor` be `Dict[str, Union[Tensor, Dict[str, Tensor]]]` (or really, `Dict[str, Any]`).  Then the `Indexer` can produce whatever nesting it wants, and the `Embedder` still just needs a single matching key.
2. Have a separate `mapping` key (or something) that's added to the `Dict[str, Tensor]` (so now it's also more complicated...), which gets read by the `BasicTextFieldEmbedder`.

Both of these options suggest that defining an object called `TextFieldTensor` might make sense, instead of having these crazy nested types...

joelgrus:
I think we can do that, it seems like there might be some weird edge cases where that might break backward compatibility, but probably it's not worth worrying about

joelgrus:
the other challenge with doing this is that without an `embedder_to_indexer_map` in the multi-index case the BasicTextFieldEmbedder has not idea how to run each embedder. That is, the bert indexer returns a dict with four keys {index_name}, {index_name}-offsets, {index_name}-type-ids, mask, and then the BertTokenEmbedder wants three inputs called "input_ids" and "offsets" and "token_type_ids", and we need some way of communicating that.

Now it is the case that if each indexer generates a full dict we no longer have to worry about key collisions, and so we could insist the bert indexer generates a dict with the right names and then use that to decide what goes where. I actually like that idea, but it's much more of a breaking change.

joelgrus:
I guess you could still allow an embedder_to_indexer map and only insist on the correct names if it's not provided

matt-gardner:
If we go with the first option, the backwards compatibility breakage is relatively minor and mostly contained to our own code.  I'd say it's probably worth it.

joelgrus:
I will take a stab at this, it's a more wide-ranging change than it first sounds (but I think it's the right design, flattening the token indexer outputs into a single dict was probably not the right design, and some of the ugliness like having to pass in `index_name` and use that to construct the various output names are an unfortunate consequence of that; if everything were namespaced you wouldn't have to bother)

schmmd:
We want to change how the embedded indexer works entirely, so closing this.

PR #3058, opened 2019-07-12
Cache data using pickle instead of jsonpickle

dwadden:
I've found that large datasets cached using jsonpickle can take quite a bit of time load back in. Using Python's pickle is much faster. It may not be portable, but it may nevertheless be useful for faster turnaround during model development. I've added a `--cache-method` flag to `allennlp train` to allow the user to cache the data using pickle. I've also added a `--cache-directory` method to `allennlp fine-tune`, which previously didn't handle cached data.

Unit tests have been added to `dataset_reader_test.py` to make sure everything works OK for both caching methods.


allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -4,6 +4,7 @@
 import pathlib
 
 import jsonpickle
+import pickle as pkl
```
matt-gardner: Please don't shorten packages on imports like this.  Just use `pickle`.

allennlp/commands/train.py
```diff
@@ -191,6 +205,9 @@ def train_model(params: Params,
         For caching data pre-processing.  See :func:`allennlp.training.util.datasets_from_params`.
     cache_prefix : ``str``, optional
         For caching data pre-processing.  See :func:`allennlp.training.util.datasets_from_params`.
+    cache_method : ``str``, optional (default='jsonpickle')
+        Data caching method. By default use `jsonpickle`, which is slow but portable. Alternatively,
+        `pickle` can be much faster but may not be portable across different Python versions.
```
matt-gardner: You should also note here that `pickle` is incompatible with lazy dataset readers.  I tried this originally, but it doesn't work.

allennlp/commands/train.py
```diff
@@ -191,6 +205,9 @@ def train_model(params: Params,
         For caching data pre-processing.  See :func:`allennlp.training.util.datasets_from_params`.
     cache_prefix : ``str``, optional
         For caching data pre-processing.  See :func:`allennlp.training.util.datasets_from_params`.
+    cache_method : ``str``, optional (default='jsonpickle')
```
matt-gardner: I know that @joelgrus won't be happy with using strings here...

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -157,9 +173,29 @@ def _read(self, file_path: str) -> Iterable[Instance]:
...
+        elif self._cache_method == 'pickle':
+            # Need to yield the entries one at a time do provide a consistent interface.
+            with open(cache_filename, 'rb') as cache_file:
+                cached = pkl.load(cache_file)
+                for entry in cached:
+                    yield entry
+        else:
+            raise ConfigurationError('Available caching methods are `jsonpickle` and `pickle`.')
+
+    def _instances_to_cache_file(self, cache_file, instances):
```
matt-gardner: Here's a potential solution that would make it so you don't need to add the `cache_method` flag, or the stringly-typed argument in various places.  Just add this method, and this method only, to `DatasetReader`, and override it in your `DatasetReader` to use pickle instead of jsonpickle.  This kind of overriding was the original intent of the design here, we just assumed that everything would be line-by-line, so we could handle lazy datasets also.  I think it's fine to add this overridable method, and let people change it to not line-by-line if they want, with the understanding that it'll break lazy caching if they want to do that.

matt-gardner:
If we do what I suggested, I'd also be quite comfortable adding a parameter to `DatasetReader.__init__` called something like `cache_with_pickle`, which is much faster but incompatible with lazy reading, and having the functionality just live here.  No need to pass arguments around.

Oh, except this would require every concrete `DatasetReader` to add that flag and pass it up to the superclass if they want this...  That's annoying.  @joelgrus, here's another sticky issue that could use a nice engineering solution: can we make `FromParams` somehow get parameters from the super class and pass them through the concrete class without needing to list them explicitly?  Using some `**kwargs` magic on the subclasses, perhaps?

dwadden:
Thanks for the feedback. Agreed that it's painful to add another command line argument. If I understand correctly, you're suggesting that we specify the caching method in the `dataset_reader` object of the `jsonnet` config file, rather than as a command line flag?

matt-gardner:
I'm suggesting that you add a method that lets you override the cache behavior in a way that lets you serialize more than one instance at a time, so that you can use pickle in your own dataset reader if you want to.  That's the minimal change here (also adding the existing cache flags to `fine-tune` is great).  Then you can create your own dataset reader, override the right methods, and get caching with pickle, or with anything else, if that's what you want.

Because caching with pickle is very convenient, it would also be nice to add it as an option on the base `DatasetReader`, so you don't have to override the methods yourself.  This would _still_ require that you write your own `DatasetReader` so you can pass the flag to the super class (or we need to modify all of our existing dataset readers to pass that flag), but at least you don't have to implement the serialization / deserialization yourself.

dwadden:
Hmm OK I'll have to think about this a little more. I've gotta do EMNLP rebuttals, I'll take a look early next week.

mbugert:
Rolling your own caching IO with different formats etc. is quite the effort. There is the [DiskCache library](http://www.grantjenks.com/docs/diskcache/index.html) which might be a good drop-in solution. @dwadden @matt-gardner WDYT?

matt-gardner:
I gave DiskCache a brief look, and it looks way more complex than what we're talking about here.  Is pickle not good enough for what you want to do @mbugert?

mbugert:
Pickle would be good enough for my purposes, yes.
DiskCache is quite flexible w.r.t. caching formats but I agree that it's somewhat overkill here.

dwadden:
Sorry I dropped the ball on this. I'll take a look today or tomorrow and try to implement your suggestion @matt-gardner.

dwadden:
Superseded by #3131, which is the minimal change suggested by @matt-gardner. Now, classes inheriting from `DatasetReader` can simply override the cache read and write methods to use a protocol other than jsonpickle.

PR #3060, opened 2019-07-13
Speedups timing

brendan-ai2:
wip


training_config/benchmark_iter.jsonnet
```diff
@@ -0,0 +1,145 @@
...
+
+local BASE_ITERATOR = {
+  "type": "bucket",
+  "max_instances_in_memory": 16384 * NUM_GPUS,
+  // Larger than we really desire for a batch. Since we set
+  // maximum_samples_per_batch below we will pack approximately that many
+  // samples in every batch.
+  "batch_size": 512 * NUM_GPUS,
+  "sorting_keys": [["source", "num_tokens"]],
+  "maximum_samples_per_batch": ["num_tokens", 2000]
```
DeNeutoy: FYI this `maximum_samples_per_batch` makes looking at the perf here confusing, because within a single batch, it chops it up into potentially N batches and then yields them all really fast (because they have been tensorised already) so it means you get a stop-start type iteration, which makes it look like one of the queues is not being filled fast enough - when I think actually it's just the speed of yielding the chunked single batch.

brendan-ai2:
Superseded by https://github.com/allenai/allennlp/pull/3119

PR #3068, opened 2019-07-16
Minor WTQ ERM model and  dataset reader fixes for demo

pdasigi:
Earlier we used the final states from the ERM trainer both for training and prediction, but that meant that we could not use interactive beam search on an ERM trained model. The main change in this PR is to use beam search in the ERM model while testing. I confirmed that the dev accuracy of a trained model remains the same after this change.

### DIFF
--- a/allennlp/models/semantic_parsing/wikitables/wikitables_erm_semantic_parser.py
+++ b/allennlp/models/semantic_parsing/wikitables/wikitables_erm_semantic_parser.py
@@ -128,9 +128,9 @@ class WikiTablesErmSemanticParser(WikiTablesSemanticParser):
                                                                dropout=dropout)
         self._checklist_cost_weight = checklist_cost_weight
         self._agenda_coverage = Average()
-        # We don't need a separate beam search since the trainer does that already. But we're defining once just to
+        # We don't need a separate beam search since the trainer does that already. But we're defining one just to
         # be able to use interactive beam search (a functionality that's only implemented in the ``BeamSearch``
-        # class) in the demo.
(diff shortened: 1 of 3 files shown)


allennlp/models/semantic_parsing/wikitables/wikitables_erm_semantic_parser.py
```diff
@@ -127,6 +128,10 @@ def __init__(self,
                                                                dropout=dropout)
         self._checklist_cost_weight = checklist_cost_weight
         self._agenda_coverage = Average()
+        # We don't need a separate beam search since the trainer does that already. But we're defining one just to
+        # be able to use interactive beam search (a functionality that's only implemented in the ``BeamSearch``
+        # class) in the demo. We'll use this only at test time.
+        self._beam_search = BeamSearch(beam_size=decoder_beam_size)  # type: ignore
```
matt-gardner: Why do you need `type: ignore` here?

pdasigi: Mypy complained that it needs type annotation for that variable. I assumed that's a mypy bug since the variable is actually being defined there.

matt-gardner: Maybe try `self._beam_search: BeamSearch = ...`?

pdasigi: Looks like that works. Thanks!

Issue #2669, opened 2019-03-29
Non-lazy iterator with epoch tracking sometimes tries to create non-homogeneous batch


danieldeutsch:
I am using a non-lazy iterator which also tracks the epoch and uses a maximum number of instances per epoch. On the first batch on the second full pass of the data, the iterator crashes at this line
https://github.com/allenai/allennlp/blob/0abefe2566c843af063dd0e4608002321cc4aa28/allennlp/data/dataset.py#L43-L44

I think I understand what the problem is. The size of my dataset is not evenly divisible by the batch size or the number of instances per epoch, so in the final batch of the first pass, some instances are getting used a second time. The instances being used for a second time already have a `MetadataField` with the epoch number. The final instances from the first pass don't yet, and so when they are all used to create a `Batch`, they don't have the same fields.

I was able to repro it by changing this unit test from
https://github.com/allenai/allennlp/blob/0abefe2566c843af063dd0e4608002321cc4aa28/allennlp/tests/data/iterators/basic_iterator_test.py#L188-L189

to `iterator = BasicIterator(batch_size=2, track_epoch=True, instances_per_epoch=2)`

joelgrus:
sounds like a bug, PR welcome, or else we'll get to it eventually

DeNeutoy:
We fixed this by deleting the code entirely and replacing it with `DataLoaders`

PR #3076, opened 2019-07-17
Fix issue #2669

hawkeoni:
The problem was described pretty well by @danieldeutsch, but I'll have to repeat its explaination in more details step by step.
1) `BasicIterator` is created with `instances_per_epoch` argument. Next `DataIterator.__call__` is used upon list of instances.
2) `BasicIterator._create_batches` is called, which in turns uses `DataIterator._memory_sized_lists`.
3) Finally control goes to `DataIterator._take_instances`, which yields amount of instances equal to  `instances_per_epoch` specified at step 1.
https://github.com/allenai/allennlp/blob/014fe3187d7314f2de3970943db5a8870ad5881b/allennlp/data/iterators/data_iterator.py#L173-L189
4) After that control comes back to `DataIterator.__call__`, epoch number metadata is added and the batch is yielded.
https://github.com/allenai/allennlp/blob/014fe3187d7314f2de3970943db5a8870ad5881b/allennlp/data/iterators/data_iterator.py#L143-L158
5) Next we return to `BasicIterator._create_batches` and create a Batch object from list of yielded instances https://github.com/allenai/allennlp/blob/014fe3187d7314f2de3970943db5a8870ad5881b/allennlp/data/iterators/basic_iterator.py#L31
6) During object creation a check conducted: https://github.com/allenai/allennlp/blob/014fe3187d7314f2de3970943db5a8870ad5881b/allennlp/data/dataset.py#L35-L44
That is precisely where the error occurs. In case without `instances_per_epoch` all instances get metadata field and thus they have no metadata field during first epoch, and have that field afterwards.
In case with `instances_per_epoch` some objects were not used during the first epoch and when they got mixed with objects that were used the check for equal fields fails.

I decided to call `add_epoch_batches` before Batch object is constructed and added a special unit test for it.

### DIFF
--- a/allennlp/data/iterators/data_iterator.py
+++ b/allennlp/data/iterators/data_iterator.py
@@ -19,13 +19,12 @@ logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 TensorDict = Dict[str, Union[torch.Tensor, Dict[str, torch.Tensor]]]  # pylint: disable=invalid-name
 
 
-def add_epoch_number(batch: Batch, epoch: int) -> Batch:
+def add_epoch_number(batch: Iterable[Instance], epoch: int) -> Iterable[Instance]:
     """
     Add the epoch number to the batch instances as a MetadataField.
     """
-    for instance in batch.instances:
+    for instance in batch:
         instance.fields['epoch_num'] = MetadataField(epoch)
-    return batch
 
 
 class DataIterator(Registrable):
@@ -135,14 +134,16 @@ class DataIterator(Registrable):
                         epoch_tensor.fill_(epoch)
                     yield tensor_dict
             else:
+
+                if self._track_epoch:
+                    add_epoch_number(instances, epoch)
+
                 batches = self._create_batches(instances, shuffle)
 
                 # Should we add the instances to the cache this epoch?
                 add_to_cache = self._cache_instances and key not in self._cache
 
                 for batch in batches:
-                    if self._track_epoch:
-                        add_epoch_number(batch, epoch)
 
                     if self.vocab is not None:
                         batch.index_instances(self.vocab)
--- a/allennlp/data/iterators/data_iterator.py
+++ b/allennlp/data/iterators/data_iterator.py
@@ -19,7 +19,7 @@ logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 TensorDict = Dict[str, Union[torch.Tensor, Dict[str, torch.Tensor]]]  # pylint: disable=invalid-name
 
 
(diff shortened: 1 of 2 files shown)


hawkeoni:
https://github.com/allenai/allennlp/issues/2669 is the original issue

hawkeoni:
Actually, I think my solution is not optimal, as each epoch I update metadata field for all instances as opposed to updating it in each batch as it used to be (This is obviously a performance downgrade in case we use limited amount of instances each epoch). I think `add_epoch_num` should be moved to `DataIterator._memory_sized_lists` , but I'd also like to hear some alternatives

matt-gardner:
@hawkeoni, sorry we dropped the ball here.  If you want to pick this back up, we'll look at it.

matt-gardner:
Closing for now as stale, but let us know if you (or anyone) wants to revisit this.

hawkeoni:
@matt-gardner Hello, Matt!
I'd like to finish this issue.
My previous messages / investigation got messy, so I'll try to repeat it without digging too much in the code to save us all a headache.

During batch construction a check occurs: all instances must have the same fields, including "epoch_num". "epoch_num" is actually added after instances are packed into the batch in `add_epoch_number`. This all should run fine, but if we limit the amount of samples per epoch, some instances never get into batches, thus they do not have "epoch_num" field and when they get into batch with instances that actually got that field an error occurs, because their fields mismatch.

My solution was to update epoch num right before each epoch in all instances, but that adds a whole data run before each epoch.

My new idea is passing a python `map` object of instances and function `add_epoch_num` in `_create_batches`, so only instances that are taken into the batch will be updated. This basically allows us not to waste computational time.

Let me know, what you think. If it is fine in your opinion, I'll submit a PR in a day.

matt-gardner:
Wouldn't it be a whole lot simpler to just not modify the instances at all and add `{'epoch_num': epoch}` to the tensor dictionary after it's created?  Somewhere around here: https://github.com/allenai/allennlp/blob/32e37f0027edf4231570e0db5466cc94188094b1/allennlp/data/iterators/data_iterator.py#L155-L160

Anyone see any problems with this?  We'd basically rip out the existing `add_epoch_num` stuff, because it's way more complicated than it needs to be, and just add this.  Slight change for existing users of this functionality, because you'd get an `int` in `forward()` instead of a tensor of shape `(batch_size,)`, but I think you'd really want the `int` anyway.

joelgrus:
couldn't you even add the tensor there, just not to break anyone's code ?

matt-gardner:
Yes, you could, though I'd probably take the opportunity of the 1.0 release to fix this design flaw.

PR #3078, opened 2019-07-18
Beam search for input reduction.

Eric-Wallace:
Hey @matt-gardner could you take a look at the changes to input reduction here?

It adds a beam search to input reduction (drastically improves results).


matt-gardner:
I'll wait for the build to pass before looking at this more closely.

Eric-Wallace:
@matt-gardner This is passing now.

allennlp/interpret/attackers/input_reduction.py
```diff
@@ -22,6 +24,10 @@ class InputReduction(Attacker):
     This check is brittle, i.e., the code could break if the name of this field has changed, or if
     a non-NER model has a field called "tags".
     """
+    def __init__(self, predictor: Predictor) -> None:
+        super().__init__(predictor)
+        self.beam_size = 3
```
matt-gardner: Why not make this a constructor argument with a default value of 3?

allennlp/interpret/attackers/input_reduction.py
```diff
@@ -48,36 +54,57 @@ def attack_from_json(self, inputs: JsonDict = None,
 
             current_text_field: TextField = current_instance[input_field_to_attack]  # type: ignore
             current_tokens = deepcopy(current_text_field.tokens)
-            smallest_idx = -1
+            # new_
```
matt-gardner: Remove commented out code.

allennlp/interpret/attackers/input_reduction.py
```diff
@@ -48,36 +54,57 @@ def attack_from_json(self, inputs: JsonDict = None,
...
+                            outputs[output] = outputs[output].detach().cpu().numpy().squeeze().squeeze()
+                        elif isinstance(outputs[output], list):
+                            outputs[output] = outputs[output][0]
+
+                    # Check if any fields have changed, if so, next beam
+                    if "tags" not in current_instance:
+                        # relabel beam_instance since last iteration removed an input token
+                        beam_instance = self.predictor.predictions_to_labeled_instances(beam_instance, outputs)[0]
+                        if any(beam_instance[field] != fields_to_compare[field] for field in fields_to_compare):
+                            continue
```
matt-gardner: This method is pretty nested and long.  Can you break it up into several methods?  E.g., an `attack_instance()` method that gets called in the top-level for loop at line 40, and maybe a `beam_step` method, or something.  And you could pull out this check into a method like `has_changed` (and that method could potentially be shared with hotflip).

allennlp/interpret/attackers/input_reduction.py
```diff
@@ -86,7 +113,8 @@ def attack_from_json(self, inputs: JsonDict = None,
 def _remove_one_token(instance: Instance,
                       input_field_to_attack: str,
                       grads: np.ndarray,
-                      ignore_tokens: List[str]) -> Tuple[Instance, int]:
+                      ignore_tokens: List[str],
+                      beam_size: int = 1) -> List[Tuple[Instance, int]]:
```
matt-gardner: This is a private method.  You don't need the default value here (and it just gives another place for defaults to mismatch, as they currently do).

allennlp/interpret/attackers/input_reduction.py
Eric-Wallace: Ok i added a those three suggestions, and shared the has_changed with hotflip

Eric-Wallace: actually, I don't think the `beam_step` function is necessary, it has 8 arguments and makes things harder to follow in my opinion.

matt-gardner: That's fine, it was just a suggestion.  Probably `attack_instance` and `has_changed` are sufficient to make this reasonable.

matt-gardner: Oh, I meant taking this whole outer block, including the special case for tagging.

matt-gardner: Probably easiest to structure it like this:
```python
beam_instance = self.predictor.predictions_to_labeled_instances(beam_instance, outputs)[0]
if util.instance_has_changed(beam_instance, fields_to_compare):
    continue

current_text_field: TextField ...
```
And the special case for NER lives inside `instance_has_changed`.

PR #3084, opened 2019-07-19
Fix MetadataField.batch_tensors

epwalsh:
Unless I'm missing something, I think this was a typo: the overridden method `MetadataField.batch_tensors` was as annotated as a `classmethod` even though it should be a instance-bound method.


matt-gardner:
It wasn't a typo, it's just a quirk of history.  That method used to be a class method (and still could be for most `Fields`), but was changed after the implementation for `MetadataField` was added, and we overlooked changing that, because it just worked (if you care, see, e.g., [here](https://github.com/allenai/allennlp/commit/4bf4294c947913569b5856e1cfa95805b39efb70)).

epwalsh:
Ahh I see, thanks for the explanation @matt-gardner!

Issue #3082, opened 2019-07-19
Pending scalars not flushed to Tensorboard's event file for short training runs


dcfidalgo:
**Describe the bug**
When training for a very short time (and few epochs), the event files for Tensorboard remain empty. For longer training runs it can also happen that a few pending scalars/histograms are not flushed to disk and values are missing in the event files.

**To Reproduce**
Steps to reproduce the behavior
1. Execute:
```
allennlp train -s output ./allennlp/tests/fixtures/basic_classifier/experiment_seq2vec.jsonnet
```
2. Look at the empty events file in
```
ls -l output/log/*
```

**Expected behavior**
The event files should contain the logged scalars/histograms.

**System (please complete the following information):**
 - OS: OSX
 - Python version: 3.7.3
 - AllenNLP version: I installed from master
 - PyTorch version: 1.1.0

**Additional context**
By default TensorboardX flushes pending events to disk every 120 seconds or when the queue is longer than 10. For very short training runs, neither condition is met and the event files remain empty. 

The `SummaryWriter` in TensorboardX have a `close` method, that flushes pending events to disk and closes the file properly. 
A solution to this issue here would be to call this method for the two `SummaryWriter`s (`self._train_log` and `self._validation_log`) in the `TensorboardWriter` at the end of a training.

If you like I would be more than happy to provide a PR in this direction.

joelgrus:
a PR would be welcome, thanks

DeNeutoy:
closed by #3085

PR #3085, opened 2019-07-19
Close tensorboard's event files properly at the end of the training

dcfidalgo:
This PR addresses #3082 .
Issue: At the end of a training run, pending events are not flushed to disk and tensorboard's event files are not closed properly.

Solution: Call the `close` method of the `SummaryWriter` at the end of the training. I added this call to the "normal" trainer as well as to the callback trainer.

Not sure if the introduced `close` method in the `TensorboardWriter` needs a unit test though.


joelgrus:
I think we can live without a unit test for this in particular if the current tests all pass

joelgrus:
both pylint and sphinx seem unhappy, I'll take a stab at making them happy. (the pylint fix is obvious, the sphinx one is a mystery, like all sphinx issues)

PR #3101, opened 2019-07-27
Fix pearson correlation.py

### LINKED ISSUE #3102: ZeroDivisionError: float division by zero [pearson_correlation.py]
**Describe the bug**
Since the input tensor may be, for example(a batch of label), a tensor ([[0.,0.,0.,0.],
[0.,0.,0.,0. ]), there will be a case where (math.sqrt(predictions_variance) or math.sqrt(labels_variance)) is zero, so a judgment is added here to prevent the denominator from being zero. If it is zero, the denominator is assigned a value of 1.

so:
i fix it by:
https://github.com/allenai/allennlp/pull/3101

allennlp/training/metrics/pearson_correlation.py
```
        pearson_r = covariance / (math.sqrt(predictions_variance) * math.sqrt(labels_variance))
```
to
```
        denominator = (math.sqrt(predictions_variance) * math.sqrt(labels_variance))
        denominator = 1 if denominator == 0 else denominator
        pearson_r = covariance / denominator
```

JackKuo666:
Fixes #3102. Since the input tensor may be, for example(a batch of label), a tensor ([[0.,0.,0.,0.],
[0.,0.,0.,0. ]), there will be a case where (math.sqrt(predictions_variance) or math.sqrt(labels_variance)) is zero, so a judgment is added here to prevent the denominator from being zero. If it is zero, the denominator is assigned a value of 1.

### DIFF
--- a/allennlp/tests/training/metrics/pearson_correlation_test.py
+++ b/allennlp/tests/training/metrics/pearson_correlation_test.py
@@ -6,55 +6,82 @@ from numpy.testing import assert_allclose
 from allennlp.common.testing import AllenNlpTestCase
 from allennlp.training.metrics import PearsonCorrelation
 
+def pearson_corrcoef(predictions, labels, fweights=None):
+    covariance_matrices = np.cov(predictions, labels, fweights=fweights)
+    denominator = np.sqrt(covariance_matrices[0, 0] * covariance_matrices[1, 1])
+    if np.around(denominator, decimals=5) == 0:
+        expected_pearson_correlation = 0
+    else:
+        expected_pearson_correlation = covariance_matrices[0, 1] / denominator
+    return expected_pearson_correlation
+
 
 class PearsonCorrelationTest(AllenNlpTestCase):
     def test_pearson_correlation_unmasked_computation(self):
         pearson_correlation = PearsonCorrelation()
         batch_size = 100
         num_labels = 10
-        predictions = np.random.randn(batch_size, num_labels).astype("float32")
-        labels = 0.5 * predictions + np.random.randn(batch_size, num_labels).astype("float32")
+        predictions_1 = np.random.randn(batch_size, num_labels).astype("float32")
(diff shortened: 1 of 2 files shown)


allennlp/tests/training/metrics/pearson_correlation_test.py
```diff
@@ -6,55 +6,82 @@
...
     def test_pearson_correlation_unmasked_computation(self):
         pearson_correlation = PearsonCorrelation()
         batch_size = 100
         num_labels = 10
-        predictions = np.random.randn(batch_size, num_labels).astype("float32")
-        labels = 0.5 * predictions + np.random.randn(batch_size, num_labels).astype("float32")
+        predictions_1 = np.random.randn(batch_size, num_labels).astype("float32")
+        labels_1 = 0.5 * predictions_1 + np.random.randn(batch_size, num_labels).astype("float32")
+
+        predictions_2 = np.random.randn(1).repeat(num_labels).astype("float32")
```
matt-gardner: Can you give a comment above this for why you are doing two separate tests?

JackKuo666: I am very happy to see your review. The reason for fixing this bug is that in some input data, a batch's `predictions` or `labels` may be exactly the same. Then when the pearson_correlation.py is called, it will appear that the `predictions_variance` or `labels_variance` of the batch is 0 when calculating the `variance`, resulting in:

`ZeroDivisionError: float division by zero`

The two tests are separated because: in most cases, the data is constructed like predictions_1; but in a few cases, for example: predictions_2, the data of such a batch is exactly the same.

matt-gardner: Yes, I get the error, I was just wanting some more description _in the code_, so that if someone comes back to this several months down the line, it'll be more obvious why there are two sets of predictions and labels.

JackKuo666:
@matt-gardner please review

DeNeutoy:
@matt-gardner can you take another look at this?

PR #3106, opened 2019-07-29
Compute metrics during validation and minor language changes

pdasigi:
My last PR caused the metrics to not be computed during validation. This PR fixes it. I also made minor changes to the wikitables language to return None instead of default values like empty lists and zeros.


matt-gardner:
@pdasigi, if this is ready for review, can you request a review from the right person?

matt-gardner:
Actually, can you move this to the new semparse repo?

PR #3108, opened 2019-07-31
Allow --recover without existing serialization dir.

brendan-ai2:
For the Beaker preemptible work it's a bit simpler to always run with the same command, but as-is AllenNLP will fail with `--recover` and an empty serialization dir. This PR instead simply warns.

The alternative, of course, is to have a wrapper script that checks for the existence of the serialization dir and dispatches to `allennlp` either with or without `--recover`. This may be a cleaner solution.

### DIFF
--- a/allennlp/training/util.py
+++ b/allennlp/training/util.py
@@ -255,8 +255,9 @@ def create_serialization_dir(
     serialization_dir: ``str``
         The directory in which to save results and logs.
     recover: ``bool``
-        If ``True``, we will try to recover from an existing serialization directory, and crash if
-        the directory doesn't exist, or doesn't match the configuration we're given.
+        If ``True``, we will try to recover from an existing serialization directory.
+        If the configuration there doesn't match that given, we'll crash.  If the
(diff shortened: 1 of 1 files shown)


brendan-ai2:
Hey @matt-gardner, this is a very simple PR, but I thought you might have concerns regarding cleanliness (masking an error could confuse users essentially). If you like this approach, I'll add a test for the behavior. Otherwise we can just use a wrapper script. Thanks!

matt-gardner:
Why not just have the original command use the `-f` flag?

brendan-ai2:
I believe `-f` doesn't work as-is due to https://github.com/allenai/allennlp/blob/master/allennlp/training/util.py#L263. Presumably that check exists because conceptually force and recover are opposites. I could overload the meaning of force, but that might be confusing too. Maybe a wrapper script really is best?

matt-gardner:
Oh, I was confusing myself about the semantics of `-f`.  Yeah, I'd prefer a simple wrapper script over this.

brendan-ai2:
Sounds good. Thanks for the feedback!

PR #3109, opened 2019-07-31
Add dropout option for BERT Pooler

luckytoilet:
This adds an optional dropout to BERT after the pooling operation but before the final feedforward. This mirrors `BertForClassification`, which has a dropout at this step.


allennlp/modules/seq2vec_encoders/bert_pooler.py
```diff
@@ -26,18 +27,25 @@ class BertPooler(Seq2VecEncoder):
...
+    dropout : ``float``, optional, (default = 0.0)
+        Amount of dropout to apply after pooling
     requires_grad : ``bool``, optional, (default = True)
         If True, the weights of the pooler will be updated during training.
         Otherwise they will not.
     """
-    def __init__(self, pretrained_model: Union[str, BertModel], requires_grad: bool = True) -> None:
+    def __init__(self,
+                 pretrained_model: Union[str, BertModel],
+                 dropout: float = 0.0,
```
DeNeutoy: Please could you change the position of this argument to be last, so that it doesn't introduce a breaking change?

luckytoilet: Ok, done

Issue #2496, opened 2019-02-11
conllu library can't be unpinned


joelgrus:
if you try to use `conllu==1.0` or later (we're pinned to 0.11), a bunch of tests fail with errors like

```
>   fields["head_indices"] = SequenceLabelField([int(x[1]) for x in dependencies],
                                                tokens,
                                                label_namespace="head_index_tags")
E   TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'

allennlp/data/dataset_readers/universal_dependencies.py:105: TypeError
```

it would be ideal (but low-priority) to update this code so we can unpin

PR #3115, opened 2019-08-05
Upgrade conllu from 0.11 to 1.3.1

EmilStenstrom:
Hi! Author of the conllu library you are using here. As you are the largest user of conllu, and have it pinned at a very old version, I thought I should help you out with an upgrade.

This PR includes three separate changes. They are structured as separate commits, with comments with references to relevant documentation.

1. Elided tokens are no longer returned as None. Your code relied on this, and the None check has simply been replaced with `isinstance(x, int)`.
2. IDs are no longer returned as strings, so casting to `int` is no longer needed.
3. You had a home-baked version of incremental parsing of conllu files, I replaced it with the one included in the latest version of conllu, `parse_incr`.

When I upgraded the universal dependencies tests failed, but with the fixes included in this PR they work. To test this PR:

1. Check out this branch
2. Run `pip install -r requirements.txt` (this upgrades conllu)
3. Run `pytest allennlp/tests/data/dataset_readers/universal_dependencies_*`

Let me know if there's something else you need in place to merge this.


EmilStenstrom:
Question: I had to change both `requirements.txt` and `setup.py`. Would you be interested in a (separate) PR that reads all packages from `requirements.txt` into `setup.py` to avoid mistakes due to duplication?

DeNeutoy:
fixes #2496

Issue #2962, opened 2019-06-18
How can I make batch instances(converting input text fields to tensor objects) faster or efficiently?


ahnjaewoo:
**System**
 - OS: Linux
 - Python version: 3.6.8
 - AllenNLP version: v0.8.4
 - PyTorch version: 1.1.0

**Question**

I have 2 questions(=problems) in terms of making batch instances.

First of all, generating batch instances with any iterator is too slow, when the input tokens are quite large. Specifically, I am using 80 sentences with 25 words per sentence in a single batch with the batch size of 32 as input tokens. While training the seq2seq attention model, the average speed of training is about 3-4 iterations/sec. And it is quite slower than the result executed from the same Tensorflow codes.

Second, I found that there is a caching option for batch generation. Actually, when I turned on the cache mode, the speed increased to about 20 iterations/sec. However, I realized that the order of batch instances is different in each order, but the configuration of the batches is always the same. I think it is a serious problem, because constant batch instances would make the model converge faster even if it's not the case.

For these reasons, I am asking for methods of making batch instances efficiently.

DeNeutoy:
Hi, this is an annoying problem in allennlp. The reason this happens is that the code which generates batches is very general and generic, so we cannot specialise to particular types of indexing which might be faster.

Firstly, one thing you can do to make this problem not so severe is to use a

https://allenai.github.io/allennlp-docs/api/allennlp.data.iterators.html#multiprocess-iterator

in combination with a https://allenai.github.io/allennlp-docs/api/allennlp.data.dataset_readers.multiprocess_dataset_reader.html

with num_processes == 1, just to make the batching happen in a separate process.

However, in general we would like to speed up the iterators wherever possible. It seems you've done some speed investigations - if you have any short scripts which you have used to diagnose these problems, we would very much like to see them!

ahnjaewoo:
Thanks for the useful information.

I think that the problem arises from the iterator, not the dataset reader.
And actually, I had tried the multiprocess iterator but it wasn't stable as I expected.

By the way, instead of sharing the short scripts, I could explain the details.
First of all, I am using the wizard-of-wikipedia dataset (https://github.com/facebookresearch/ParlAI/tree/master/projects/wizard_of_wikipedia) 
and during "read_" process in dataset reader, there are several text fields as below.

```python
fields = {
    'context': TextField(context, self._token_indexers),
    'response': TextField(response, self._token_indexers),
[...]

Issue #1890, opened 2018-10-10
Unable to see the gains by using `MultiProcessDatasetReader`


HarshTrivedi:
@joelgrus, I am unable to see the gains by using the `MultiProcessDatasetReader`. In fact it's taking 7x more time to `dry-run` in this particular case. I could be doing something wrong here, so please let me know if so.

I have created a minimally reproducible branch [here](https://github.com/HarshTrivedi/allennlp/tree/check-multiprocess-reader). It runs ESIM textual entailment (one already present in `allennlp`) by reading SNLI train dataset
1. top 50 K pairs (single process) vs
2. top 50 K pairs (multi process where each file contains 10K splits)

Edit: Actually esim is only for borrowing config file, I am only doing `dry-run` here.

DatasetReader config is as follows:
```
    "dataset_reader": {
        "type": "multiprocess",
        "base_reader" : {
            "type": "snli",
            "token_indexers": {
                "tokens": {
                    "type": "single_id",
                    "lowercase_tokens": true
                }
            }
        },
        "num_workers": 10,
        "epochs_per_read": 3
    },
```

To reproduce, just run following on that branch:
```
time python allennlp/run.py dry-run experiment_single_process.jsonnet -s single_process_serialization

114.71s user 8.85s system 85% cpu 2:24.73 total
```
vs 
```
time python allennlp/run.py dry-run experiment_multi_process.jsonnet -s multi_process_serialization

734.79s user 119.35s system 134% cpu 10:33.51 total
```

Btw, I did change snli reader to use allennlp `Token` instead of spacy tokens as in #1887. You can check that in last commit on that branch.

schmmd:
@HarshTrivedi how large was your original dataset?  We have not tested this code thoroughly, but we're going through it soon so, if there are issues, we might sort them out soon.

HarshTrivedi:
@schmmd My original dataset is about ~2M pairs of sentences. I wasn't able to see gain in that so I made this small case with only 50K entailment pairs (from snli) - with and without multiprocess-reader enabled. This is easily reproducible w/o any part of my code ... It might help debug the problem.

brendan-ai2:
I'm hitting this now with the Calypso port. Investigating.

brendan-ai2:
One issue is that you'll also likely need the multiprocess iterator. Add something like the following to your config:
```
  "iterator": {
    "type": "multiprocess",
    "iterator": {
      "type": "basic",
      "batch_size": 32
    },
[...]

PR #3119, opened 2019-08-07
Benchmark iterator, avoid redundant queue, remove managers.

brendan-ai2:
- Adds a script to benchmark iterators.
  - Average speed
  - Introspects queues
- Removes a bottleneck when `MultiprocessDatasetReader` and `MultiprocessIterator` are used in conjunction.
  - Specifically, removes a redundant queue that was populated by a single process.
- Removes managers which have significant overhead.
- Results on training_config/bidirectional_language_model.jsonnet:
  - original code, no multiprocessing: 0.047 s/b over 10000 batches
  - original code, workers = 1: 0.073 s/b over 10000 batches
  - original code, workers = 10: 0.078 s/b over 10000 batches
  - this PR (-queue), workers = 1: 0.073 s/b over 10000 batches
  - this PR (-queue), workers = 10: 0.046 s/b over 10000 batches
  - this PR (-queue, - manager), workers = 1: 0.063 s/b over 10000 batches
  - this PR (-queue, - manager), workers = 10: 0.020 s/b over 10000 batches
  - Notably, previously we did not see any benefit from scaling to multiple workers. Now we do, albeit worse than linearly. More work required there.
- Related issues: https://github.com/allenai/allennlp/issues/2962, https://github.com/allenai/allennlp/issues/1890


brendan-ai2:
@matt-peters, any interest in reviewing this PR?

It would also be great to do a debugging session with you at some point. There's an additional bottleneck that I'm having trouble pinning down. In short, we're not swapping and the input queue is kept full, but CPU utilization is poor. Generating a batch takes about an order of magnitude more than your target of several milliseconds. My best guess at the moment is contention, but I'm not sure how to measure that in Python. Compile CPython with suitable flags for running Valgrind?

brendan-ai2:
@joelgrus, could you take a look as well? Thanks!

allennlp/data/dataset_readers/multiprocess_dataset_reader.py
```diff
@@ -30,26 +33,119 @@ def info(cls, message: str) -> None:
 def _worker(reader: DatasetReader,
             input_queue: Queue,
             output_queue: Queue,
-            index: int) -> None:
+            active_workers: Value,
+            inflight_items: Value,
```
joelgrus: should these be called `num_active_workers` and `num_inflight_items` so that it's clearer what they're doing?

allennlp/tests/data/dataset_readers/multiprocess_dataset_reader_test.py
```diff
@@ -72,6 +75,50 @@ def test_multiprocess_read(self):
...
         assert counts[("birds", "are", "animals", ".", "N", "V", "N", "N")] == 100
 
+    def test_multiprocess_read_partial_does_not_hang(self):
+        reader = MultiprocessDatasetReader(base_reader=self.base_reader, num_workers=4, output_queue_size=10)
+
+        all_instances = []
+
+        # Half of 100 files * 4 sentences / file
+        for _ in range(200):
+            instance = reader.read(self.identical_files_glob)
```
joelgrus: this doesn't affect the test per se, but doesn't `reader.read` return an Iterable of instances? maybe just the variable name is misleading

training_config/bidirectional_language_model.jsonnet
```diff
@@ -37,12 +37,18 @@ local BASE_ITERATOR = {
   "maximum_samples_per_batch": ["num_tokens", 2000]
 };
 
+//local BASE_ITERATOR = {
+//  "type": "basic",
+//  "batch_size": 66,
+//};
+
 {
+  #"dataset_reader": BASE_READER,
```
joelgrus: delete commented out code?

allennlp/data/dataset_readers/multiprocess_dataset_reader.py
brendan-ai2: Good call, done.

allennlp/tests/data/dataset_readers/multiprocess_dataset_reader_test.py
brendan-ai2: This was actually a bug. Nice catch!

training_config/bidirectional_language_model.jsonnet
brendan-ai2: Fixed

brendan-ai2:
Thanks, feedback addressed!

allennlp/data/dataset_readers/multiprocess_dataset_reader.py
brendan-ai2: Sorry about that missing that reference. Updated and grepped for any more stray instances. Should be good now.

allennlp/tests/data/dataset_readers/multiprocess_dataset_reader_test.py
```diff
@@ -72,6 +75,53 @@ def test_multiprocess_read(self):
...
 
+    def test_multiprocess_read_partial_does_not_hang(self):
+        reader = MultiprocessDatasetReader(base_reader=self.base_reader, num_workers=4, output_queue_size=10)
+
+        all_instances = []
+
+        # Half of 100 files * 4 sentences / file
+        i = 0
+        for instance in reader.read(self.identical_files_glob):
+            if i == 200:
```
joelgrus: I would probably just add a comment here, along the lines of "i == 200 means this is the 201st instance, which is too many, so break out and fail the test"

brendan-ai2:
Thanks for the review! (I'm dragging my heels on merging due to some test discrepancies between TC and my local machine. I'll ping for another pass if the fix is anything major.)

brendan-ai2:
Thanks for the review!

PR #3124, opened 2019-08-08
Support tqdm in interactive shells

ksurya:


schmmd:
@ksurya would you add a comment that describes why this is necessary?  What exactly is an InteractiveShell?

ksurya:
In Jupyter notebooks, sometimes `tqdm.tqdm` module creates new progress bars at each iteration. https://github.com/tqdm/tqdm/issues/375

Screen Shot 2019-09-11 at 1 37 51 PM

tqdm supports native progress bar widgets in Jupyter notebooks. https://pypi.org/project/tqdm/#ipython-jupyter-integration

This code loads `tqdm.tqdm_notebook` module, if the code is executing in a Jupyter notebook. 

So, the progress bars would look like this:

Screen Shot 2019-09-11 at 1 14 10 PM

schmmd:
I'd like to merge this, but `pylint` is giving an error that seems quite reasonable:

```
allennlp/common/tqdm.py:7: error: Name 'get_ipython' is not defined
```

How does this work since `get_ipython` isn't defined?

ksurya:
Sorry for the late reply, 

`get_ipython` is available only on `ipython` interpreter.

PR #3140, opened 2019-08-12
switch for DataIterators whether smaller batches should be skipped

### LINKED ISSUE #3118: Feature request: Switch for DataIterators whether batches smaller than batch_size should be skipped
In scenarios where the number of samples are not dividable by the batch size, the last batch cannot be filled completely.
If one needs fixed-length batches, one might want to discard those "incomplete" batches during training.

I haven't been able to find such a switch in the documentation although it seems to be a rather simple feature.

Use cases are scenarios where one needs a guaranteed minimal batch size, such as batch normalization where batches should always be bigger than 1, or models that use a buffer depending on a fixed batch size.

Usage example:
```
buck_iter = BucketIterator([("input", "num_tokens")], skip_smaller_batches=True)
```

mnschmit:
fixes #3118


mnschmit:
Yes, of course, I added them. Thanks for the hint!

DeNeutoy:
Great, thanks a lot!

PR #3141, opened 2019-08-12
Improve check_links.py CI script

epwalsh:
It looks like this check script isn't used in the CI pipeline anymore, but I have a similar script in some of my other projects that I recently made improvements to, so I thought I'd pass them along in case we use it again (and by the way, it looks there are about 15 broken links, most of them local links from tutorials to files that no longer exist).

The updates to the script include:
- using a single http session to make all requests, which speeds up the script by an order of magnitude.
- using a HEAD request instead of a GET request to avoid downloading content, which also provides a significant speed-up.
- an improved way of checking whether an external link is valid.


DeNeutoy:
Thanks, I think we had some concern about making too many requests from our CI domain or something - I think @schmmd had the most context on this.

schmmd:
@epwalsh thanks!  Sorry for the delay.  We don't run this as part of CI because we don't want to be spamming the internet automatically, but this is a great improvement for running it locally quickly.

epwalsh:
@schmmd no problem! Makes sense

PR #3142, opened 2019-08-12
Update HTTP links to HTTPS where possible

bryant1410:
I changed the HTTP links to HTTPS where I saw the domain accepted it (I checked them). I also updated some sites' links that changed address such as Flask. And I changed the http://docs.allennlp.org/en/latest to https://allenai.github.io/allennlp-docs that I saw it changed, in which for some I needed to change the URL to add a missing module name.

Btw, I saw build.allennlp.org doesn't support HTTPS and that http://labs.semanticscholar.org/corpus/ is down.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
                                  Apache License
                            Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+                        https://www.apache.org/licenses/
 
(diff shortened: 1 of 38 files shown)


bryant1410:
And I changed a `Dataset` mention in a tutorial file for `Batch`.

PR #3143, opened 2019-08-13
Add support for running preemptible workloads on beaker

viveklak:
Initial cut at enabling preemptible training on beaker


resumable_train.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
brendan-ai2: Could you please place this in the `scripts/ai2-internal/` directory as well? Our external users won't generally need it.

viveklak: Done.

PR #3145, opened 2019-08-13
Adding a PretrainedTransformerTokenizer

matt-gardner:
First (of many) PRs merging our hackathon project into the main repo.  This one adds a dependency on the new `pytorch-transformers` repo, and grabs a tokenizer from there.

This also fills a gap that we had in our data processing API, where there was no good way to get native BERT (or GPT, etc.) tokenization as input to your model - you had to use words, then split them further into wordpieces or byte pairs.  With this tokenizer, you can have wordpieces or byte pairs be your base tokens, instead of words.

### DIFF
--- a/allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
+++ b/allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
@@ -1,0 +1,12 @@
+# pylint: disable=no-self-use,invalid-name
+
+from allennlp.common.testing import AllenNlpTestCase
+from allennlp.data.tokenizers import PretrainedTransformerTokenizer
+
+class TestPretrainedTransformerTokenizer(AllenNlpTestCase):
+    def test_splits_into_wordpieces(self):
+        tokenizer = PretrainedTransformerTokenizer('bert-base-cased', do_lowercase=False)
+        sentence = "A, [MASK] AllenNLP sentence."
+        tokens = [t.text for t in tokenizer.tokenize(sentence)]
+        expected_tokens = ["[CLS]", "A", ",", "[MASK]", "Allen", "##NL", "##P", "sentence", ".", "[SEP]"]
+        assert tokens == expected_tokens
--- a/setup.py
+++ b/setup.py
@@ -131,6 +131,7 @@ setup(name='allennlp',
           'sqlparse>=0.2.4',
           'word2number>=1.1',
           'pytorch-pretrained-bert>=0.6.0',
+          'pytorch-transformers @ https://api.github.com/repos/huggingface/pytorch-transformers/tarball/a7b4cfe9194bf93c7044a42c9f1281260ce6279e',
           'jsonpickle',
       ],
       entry_points={
--- a/allennlp/data/tokenizers/__init__.py
+++ b/allennlp/data/tokenizers/__init__.py
@@ -5,5 +5,6 @@ tokenization, stemming, and filtering.
 
 from allennlp.data.tokenizers.tokenizer import Token, Tokenizer
 from allennlp.data.tokenizers.word_tokenizer import WordTokenizer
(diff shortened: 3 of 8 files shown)


allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
```diff
@@ -0,0 +1,12 @@
+# pylint: disable=no-self-use,invalid-name
+
+from allennlp.common.testing import AllenNlpTestCase
+from allennlp.data.tokenizers import PretrainedTransformerTokenizer
+
+class TestPretrainedTransformerTokenizer(AllenNlpTestCase):
+    def test_splits_into_wordpieces(self):
+        tokenizer = PretrainedTransformerTokenizer('bert-base-cased', do_lowercase=False)
```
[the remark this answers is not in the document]
matt-gardner: It would be good to not have to the dependency on the web for the test, but I'm pretty sure it wouldn't work, because we rely on pytorch-transformers' `AutoTokenizer`.

Well, *maybe* it would work if I make sure the filename that I give for the fixture passes their heuristics for which actual tokenizer to use, but (1) I'm not sure that I can, because their heuristics might not allow it, and (2) that would make the test too dependent on the internals of their implementation.

matt-gardner:
The BertBasicWordSplitter is for a different use case, where you have mismatched tokenization and indexing.  This can't actually replace that code.  We *should* rename all of that mismatched code eventually, though, to make it more obvious what's going on and when you should use each option.

setup.py
```diff
@@ -131,6 +131,7 @@
           'sqlparse>=0.2.4',
           'word2number>=1.1',
           'pytorch-pretrained-bert>=0.6.0',
+          'pytorch-transformers @ https://api.github.com/repos/huggingface/pytorch-transformers/tarball/a7b4cfe9194bf93c7044a42c9f1281260ce6279e',
```
huntzhan: Hi @matt-gardner, why not use `pytorch-transformers==1.1.0`?

PyPI won't allow direct URL like this one, as discussed in https://github.com/pypa/pip/issues/6301
And error msg should be raised as followed when running `twine upload`
```
HTTPError: 400 Client Error: Invalid value for requires_dist. Error: Can't have direct dependency: 'pytorch-transformers @ https://api.github.com/repos/huggingface/pytorch-transformers/tarball/a7b4cfe9194bf93c7044a42c9f1281260ce6279e' for url: https://upload.pypi.org/legacy/
```

matt-gardner: Because the auto stuff hadn't been released when I wrote this. It has been now, so this can be updated. PR welcome.

brendan-ai2:
@matt-gardner, I'm getting the following error when attempting to install the library which I believe is caused by this change.

```
brendanr.local ➜  entity_salience git:(integrate-bert) ✗ pip install --editable ../allennlp
Obtaining file:///Users/brendanr/repos/brendan-ai2/allennlp
Direct url requirement (like pytorch-transformers@ https://api.github.com/repos/huggingface/pytorch-transformers/tarball/a7b4cfe9194bf93c7044a42c9f1281260ce6279e) are not allowed for dependencies
You are using pip version 10.0.1, however version 19.2.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
brendanr.local ➜  entity_salience git:(integrate-bert) ✗ echo $?
1
```

Perhaps we can fix this by upgrading to the recently released pytorch-transformers v1.1.0? https://pypi.org/project/pytorch-transformers/1.1.0/

joelgrus:
there is a PR for this:

https://github.com/allenai/allennlp/pull/3171#pullrequestreview-277467263

I'm just waiting for it to update and then I'll merge

brendan-ai2:
Thanks, @joelgrus !

joelgrus:
ok, it's merged

PR #3146, opened 2019-08-13
Pretrained transformer indexer

matt-gardner:
Second PR getting the hackathon project into the main repo.  This one adds a `TokenIndexer` for pretrained transformers.  This indexer assumes that you're using the corresponding pretrained tokenizer, so we don't have a mismatch between our tokenization and indexing, and things are easy.  There aren't any bells and whistles around windowing or anything, but this should be good enough for now.  This PR depends on #3145, so it shouldn't be reviewed just yet, but I'm getting CI going for it.

### DIFF
--- a/allennlp/data/token_indexers/pretrained_transformer_indexer.py
+++ b/allennlp/data/token_indexers/pretrained_transformer_indexer.py
@@ -1,0 +1,87 @@
+from typing import Dict, List
+import itertools
+
+from overrides import overrides
+import pytorch_transformers
+from pytorch_transformers.tokenization_utils import PreTrainedTokenizer
+import torch
+
+from allennlp.common.util import pad_sequence_to_length
+from allennlp.data.vocabulary import Vocabulary
+from allennlp.data.tokenizers.token import Token
+from allennlp.data.token_indexers.token_indexer import TokenIndexer
+
+
+@TokenIndexer.register("pretrained_transformer")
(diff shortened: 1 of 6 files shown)


allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -0,0 +1,89 @@
...
+                          tokens: List[Token],
+                          vocabulary: Vocabulary,
+                          index_name: str) -> Dict[str, List[int]]:
+        if not self._added_to_vocabulary:
+            self._add_encoding_to_vocabulary(vocabulary)
+            self._added_to_vocabulary = True
+        token_text = [token.text for token in tokens]
+        indices = self.tokenizer.convert_tokens_to_ids(token_text)
+
+        return {index_name: indices}
```
joelgrus: how do you imagine we'll get the offsets in here? or does that become the dataset reader's job to stick them in a separate `ArrayField`? I'm having trouble seeing how that would work.

matt-gardner: We don't need them, because this indexer doesn't have a mismatch with its corresponding tokenizer.

joelgrus: but imagine you're trying to predict sequence labels that correspond to the actual tokens? or are you saying that in that case you'd need to "expand" them to correspond to the wordpieces?

PR #3150, opened 2019-08-13
Comparison operations return dtype torch.bool

davebulaval:
From [release note](https://github.com/pytorch/pytorch/releases) of torch:

> Comparison operations (lt (<), le (<=), gt (>), ge (>=), eq (==), ne, (!=) ) return dtype has changed from torch.uint8 to torch.bool (21113)

* Note:
Since I haven't made the correction for the other torch modification all the test faillure where there before my modification.

So feel free to reject it if this is unaceptable, just whant to help solve this problem since I got this error and managed to solved it.

### DIFF
--- a/allennlp/tests/data/fields/array_field_test.py
+++ b/allennlp/tests/data/fields/array_field_test.py
@@ -80,16 +80,16 @@ class TestArrayField(AllenNlpTestCase):
         returned_tensor1 = array_field1.as_tensor(array_field1.get_padding_lengths())
         assert returned_tensor1.dtype == torch.int64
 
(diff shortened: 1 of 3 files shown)


DeNeutoy:
Hi, thanks for this PR, but typically upgrading a pytorch version is something we do ourselves, so we can do it once across the entire codebase. Sorry if that wasn't clear! The new pytorch release has several breaking changes, so we will have to force users to upgrade, which means we need to wait a little bit before we do so. 

Thanks though!

davebulaval:
@DeNeutoy no worries. I understand.

PR #3153, opened 2019-08-13
quoref metric and evaluator

pdasigi:


pdasigi:
@matt-gardner this is ready to be reviewed. Should be pretty straightforward as the quoref metric depends heavily on the drop metric, with just the data formats being different. I think this is the last needed piece for the leaderboard.

allennlp/training/metrics/quoref_em_and_f1.py
```diff
@@ -0,0 +1,31 @@
+from typing import List, Union
+
+from overrides import overrides
+
+from allennlp.training.metrics.metric import Metric
+from allennlp.training.metrics.drop_em_and_f1 import DropEmAndF1
+
+
+@Metric.register("quoref")
+class QuorefEmAndF1(DropEmAndF1):
```
matt-gardner: Probably better to just instantiate the `DropEmAndF1` inside of `__init__` and call it in `__call__`, instead of inheriting.  The end result is the same, and you have more flexibility later.

allennlp/tools/quoref_eval.py
```diff
@@ -0,0 +1,108 @@
...
+    answers_dict: Dict[str, List[str]] = {}
+    for article_info in annotations["data"]:
+        for paragraph_info in article_info["paragraphs"]:
+            for qa_pair in paragraph_info["qas"]:
+                query_id = qa_pair["id"]
+                candidate_answers = [answer["text"] for answer in qa_pair["answers"]]
+                answers_dict[query_id] = candidate_answers
+    return answers_dict
+
+def evaluate_json(annotations: Dict[str, Any], predicted_answers: Dict[str, Any]) -> Tuple[float, float]:
```
matt-gardner: Why can't you use the drop script for this function (and the one above) also?

allennlp/tests/training/metrics/quoref_em_and_f1_test.py
```diff
@@ -0,0 +1,17 @@
+# pylint: disable=no-self-use,invalid-name
+from allennlp.common.testing import AllenNlpTestCase
+from allennlp.training.metrics import QuorefEmAndF1
+
+
+class QuorefEmAndF1Test(AllenNlpTestCase):
```
matt-gardner: I'd put a comment in here saying that this is intentionally very light on tests, because we are just passing off to the drop metric, which has a lot of tests.

allennlp/training/metrics/quoref_em_and_f1.py
matt-gardner: Even better, can you just have the model format the ground truth in the way that the other metric expects, and remove the need for this class entirely?

pdasigi: Agreed. Removing this class.

allennlp/tools/quoref_eval.py
pdasigi: Quoref's json format is the same as that of SQuAD, and DROP's is different. So `evaluate_json` needs to be different, and `evaluate_prediction_file` calls that function, so that needed to be rewritten as well. Mentioned this in the docstring.

allennlp/tests/training/metrics/quoref_em_and_f1_test.py
pdasigi: I removed this class along with `QuorefEmAndF1`. I added a comment similar to the one you suggested in the tests I have for the evaluation script.

allennlp/training/metrics/__init__.py
```diff
@@ -26,3 +26,4 @@
 from allennlp.training.metrics.srl_eval_scorer import SrlEvalScorer, DEFAULT_SRL_EVAL_PATH
 from allennlp.training.metrics.unigram_recall import UnigramRecall
 from allennlp.training.metrics.auc import Auc
+from allennlp.training.metrics.quoref_em_and_f1 import QuorefEmAndF1
```
matt-gardner: Remove this line.

PR #3164, opened 2019-08-16
Switch SemanticRoleLabeler metric to SrlEvalScorer.

nelson-liu:
#2743 added the SrlEvalScorer, but I didn't actually change the metric used in the `SemanticRoleLabeler` 😅 

This PR has a few things:
- Fixes a circular import between SrlEvalMetric and SemanticRoleLabeler by moving the utility functions for converting between BIO / CoNLL tags to a separate `srl_utils.py` file. Let me know if you want them somewhere else.
- Switch `SemanticRoleLabeler.span_metric` from `SpanBasedF1` to `SrlEvalScorer`.

I'll try actually training this model in a bit, to make sure that nothing broke disastrously. The fact that the tests still work is somewhat promising, though...

### DIFF
--- a/allennlp/models/semantic_role_labeler.py
+++ b/allennlp/models/semantic_role_labeler.py
@@ -159,7 +159,7 @@ class SemanticRoleLabeler(Model):
                                                       tags,
                                                       mask,
                                                       label_smoothing=self._label_smoothing)
-            if not self.ignore_span_metric:
+            if not self.ignore_span_metric and self.span_metric is not None and not self.training:
                 batch_verb_indices = [example_metadata["verb_index"] for example_metadata in metadata]
                 batch_sentences = [example_metadata["words"] for example_metadata in metadata]
                 # TODO (nfliu): This is kind of a hack, consider splitting out part
--- a/allennlp/models/semantic_role_labeler.py
+++ b/allennlp/models/semantic_role_labeler.py
@@ -8,14 +8,14 @@ import torch.nn.functional as F
 
 from allennlp.common.checks import check_dimensions_match
 from allennlp.data import Vocabulary
+from allennlp.models.srl_util import convert_bio_tags_to_conll_format, write_bio_formatted_tags_to_file
 from allennlp.modules import Seq2SeqEncoder, TimeDistributed, TextFieldEmbedder
 from allennlp.modules.token_embedders import Embedding
 from allennlp.models.model import Model
 from allennlp.nn import InitializerApplicator, RegularizerApplicator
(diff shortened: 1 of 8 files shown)


allennlp/models/semantic_role_labeler.py
```diff
@@ -144,20 +142,34 @@ def forward(self,  # type: ignore
...
                                                       tags,
                                                       mask,
                                                       label_smoothing=self._label_smoothing)
             if not self.ignore_span_metric:
-                self.span_metric(class_probabilities, tags, mask)
+                batch_verb_indices = [example_metadata["verb_index"] for example_metadata in metadata]
+                batch_sentences = [example_metadata["words"] for example_metadata in metadata]
+                # TODO (nfliu): This is kind of a hack, consider splitting out part
+                # of decode() to a separate function.
+                batch_bio_predicted_tags = self.decode(output_dict).pop("tags")
```
nelson-liu: I need a sanity check here...is this cheating? Using `decode()` and passing in the tags is actually "better" than just passing in the `class_probabilities` (since you get the best global tag sequence). The F1 score that I'm getting after running `allennlp evaluate` seems higher than it should be...`89.46 F1`. I figure that using `decode()` should help, but I don't expect a jump of like 4 F1...

nelson-liu: oops, I'm not ignoring the `V` tag, so the score is incredibly inflated. fix incoming, i verified that the results i get from `allennlp evaluate` matches manually using `allennlp predict`->converting to conll->running the perl file.

nelson-liu:
Ok, if I run the following on this branch:

```
allennlp evaluate https://allennlp.s3.amazonaws.com/models/srl-model-2018.05.25.tar.gz conll-formatted-ontonotes-5.0/data/test/ --cuda-device 0
```

```
2019-08-16 18:29:30,744 - INFO - allennlp.commands.evaluate - Finished evaluating.
2019-08-16 18:29:30,744 - INFO - allennlp.commands.evaluate - Metrics:
2019-08-16 18:29:30,744 - INFO - allennlp.commands.evaluate - precision-overall: 0.8448776845283509
2019-08-16 18:29:30,744 - INFO - allennlp.commands.evaluate - recall-overall: 0.8537753645773527
2019-08-16 18:29:30,744 - INFO - allennlp.commands.evaluate - f1-measure-overall: 0.8493032211741937
2019-08-16 18:29:30,744 - INFO - allennlp.commands.evaluate - loss: 0.5136776601058662
```

The P, R, and F1 are exactly what I get from the previous method for evaluating the SRL model (with `scripts/write_srl_predictions_to_conll_format.py` and then `srl-eval.pl`).

PR #3169, opened 2019-08-17
Modify bioul_tags_to_spans for ill-formed spans

Maybewuss:
If a span is ill-formed, it will not be append to spans list instead of raise InvalidTagSequence.


brendan-ai2:
Thanks for this. Could you please add a test as well? These metrics are quite important and delicate.

Maybewuss:
> Thanks for this. Could you please add a test as well? These metrics are quite important and delicate.
Sorry, I find a bug that I lose sight of the spans unlabeled, so I will pull request agin.

PR #3172, opened 2019-08-17
Add a test for the subcommands docstring help outputs

bryant1410:


allennlp/tests/commands/docstring_help_test.py
```diff
@@ -0,0 +1,32 @@
...
+                                               check=True, shell=True, stdout=subprocess.PIPE).stdout.decode()
+
+                pytorch_pretrained_bert_warning = "Better speed can be achieved with apex installed from" \
+                    " https://www.github.com/nvidia/apex.\n"
+                if actual_output.startswith(pytorch_pretrained_bert_warning):
+                    actual_output = actual_output[len(pytorch_pretrained_bert_warning):]
+
+                self.assertEqual(expected_output, actual_output)
+            else:
+                self.assertIn(module_info.name, [parent_module.__name__ + '.subcommand'],
```
DeNeutoy: This looks good but can you make this assert error more helpful - it's more likely that documentation has changed, rather than not being present, so this error might be a bit confusing.

Something like:
"The documentation for command {command} in file {file} does not match the output of running {command}. Please update the docstring to match the command help string."

bryant1410: Yeah. This error, in particular, is that the pattern `allennlp something --help` was not found. The previous assert is the one in which it doesn't match.

DeNeutoy: Ah right, I see - new messages look better anyway, thanks!

joelgrus:
this test takes > 20 seconds to run on my machine, it is really so valuable to run every time, or it should it be more like a sniff test that gets run rarely

bryant1410:
> this test takes > 20 seconds to run on my machine, it is really so valuable to run every time, or it should it be more like a sniff test that gets run rarely

There's a way around it, which is getting the argument parser object (or subparser?) associated with the module and calling the `print_help` method, preferably with a file object (as opposed to just stdout) to check the output. That probably runs faster cause doesn't spawn new processes.

bryant1410:
But this doesn't seem to be that valuable to run every time, as you say.

DeNeutoy:
Reverting - this test is too slow and and is a little fragile.

bryant1410:
> Reverting - this test is too slow and and is a little fragile.

Why is it fragile? We could leave at least the fixed files.

joelgrus:
it's fragile because it's failing on this PR

https://github.com/allenai/allennlp/pull/3182

even though I didn't touch any of the commands / docstrings

nelson-liu:
The docstring test seems to pass on CI, though? But I agree it's a bit slow, and that it might be better as a sniff test. It does seem like a good thing to test---the docstrings for the commands are frequently out of date.

bryant1410:
Yeah, I don't see it as failing, just slow.

joelgrus:
huh, it's not failing on teamcity, but (on the same branch) it fails for me locally

```
E               AssertionError: 'usag[1066 chars]tps://\n                        allennlp.s3.am[1452 chars]de\n' != 'usag[1066 chars]tps:/\n                        /allennlp.s3.am[1452 chars]de\n'
E               Diff is 4277 characters long. Set self.maxDiff to None to see it. : The documentation for the subcommand usage in the module allennlp.commands.elmo does not match the output of running `allennlp elmo --help`. Please update the docstring to match the output.
```

bryant1410:
> huh, it's not failing on teamcity, but (on the same branch) it fails for me locally
> 
> ```
> E               AssertionError: 'usag[1066 chars]tps://\n                        allennlp.s3.am[1452 chars]de\n' != 'usag[1066 chars]tps:/\n                        /allennlp.s3.am[1452 chars]de\n'
> E               Diff is 4277 characters long. Set self.maxDiff to None to see it. : The documentation for the subcommand usage in the module allennlp.commands.elmo does not match the output of running `allennlp elmo --help`. Please update the docstring to match the output.
> ```

@joelgrus can you check if the one from #3185 runs okay?

bryant1410:
To avoid this to happen again, what if [a short timeout (e.g., 5s) is set to every unit test (maybe with some case-by-case exceptions)](https://pypi.org/project/pytest-timeout/) so that future slows tests in PRs are automatically detected?

PR #3176, opened 2019-08-20
Use the pytorch base image.

schmmd:
The reason I attempted this was to get an image that could run as-is on our on-premise hardware (such as `allennlp-server-3`).  Currently `allennlp/allennlp:v0.8.4` will fail with a `CUDNN_STATUS_EXECUTION_FAILED` but this change will force `allennlp` to use the right version of CUDA (10.0).

I'm not sure this is a good idea.  It does increase the base image size, but only slightly.

```
pytorch/pytorch                                           1.1.0-cuda10.0-cudnn7.5-runtime            299bfb9e54db        3 months ago        4.17GB
```

For size (and build times) this is probably a net win because more is done (and cached) in the base image.  That said, we do have less control on what version of Python we're using (currently Python 3.6.8 from Anaconda) although we could regain that functionality by re-implementing https://github.com/pytorch/pytorch/blob/master/docker/pytorch/Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,15 +3,6 @@ FROM pytorch/pytorch:1.1.0-cuda10.0-cudnn7.5-runtime
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 
-ENV PATH /usr/local/nvidia/bin/:$PATH
-ENV LD_LIBRARY_PATH /usr/local/nvidia/lib:/usr/local/nvidia/lib64
-
-# Tell nvidia-docker the driver spec that we need as well as to
-# use all available devices, which are mounted at /usr/local/nvidia.
-# The LABEL supports an older version of nvidia-docker, the env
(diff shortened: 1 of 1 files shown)


schmmd:
@dplessas @splittingfield would you find this change helpful?

schmmd:
Looks like we cannot quite do this.  They do not commit to updating this registry.  I also found this discussion: https://discuss.pytorch.org/t/official-docker-image/4159/6

I might try basing off the nvidia base image.

schmmd:
I'm going to close this PR and open a new one that uses the `nvidia` base image only for `Dockerfile.pip`.

splittingfield:
This makes sense.  I would recommend running some perf comparisons through beaker first for such changes.  I was shocked to see such a significant decrease in speed when trying the Google Official Images. (https://beaker.org/gr/gr_eomt82zi4vna/compare)

PR #3177, opened 2019-08-20
Use an NVIDIA base image.

schmmd:
This yields a smaller base image with reduced complexity around NVIDIA/CUDA (we simply use the base image) but increased complexity around what version of Python we're installing (we need to manage it more manually rather than just using the base image).

This change would make it easier to run the `allennlp` Docker image on a GPU because it wouldn't require that the CUDA 10.0 drivers are installed--the image packages the drivers itself.  For example, `allennlp-server-3` has the 10.1 drivers installed.  With the current Docker image we have a `CUDNN_STATUS_EXECUTION_FAILED` but training works successfully on a GPU with this image.

```
allennlp                                                  nvidia-base-18.04                          b8b191a7a08d        2 hours ago         3.66GB
```


schmmd:
Another way I could have set this up is by relying more on `conda`--both managing pytorch and cuda versions with it.

I could easily apply the same change here on our `Dockerfile` we use for CI, if desired.

Dockerfile.pip
```diff
@@ -1,33 +1,35 @@
...
 WORKDIR /stage/allennlp
 
 ARG VERSION
 ARG SOURCE_COMMIT
 
+RUN apt-get update && apt-get install -y --no-install-recommends \
+         build-essential \
+         cmake \
+         git \
+         python3-dev \
```
joelgrus: can we be more specific about what version of python to install? off the top of my head I have no idea what version this would result in

schmmd: I'd have to use Conda.  I previously looked into that.

```
# python --version
Python 3.6.8
```

schmmd: I pushed a commit that forces 3.6 at least.

schmmd:
@joelgrus does this look good or should we try Conda?  We're now pinning to `3.6.x`.

PR #3182, opened 2019-08-21
upgrade to pytorch 1.2

joelgrus:
as you can see, this ended up not being that much work. 

there are still two tests failing for me. 

the first is the new "docstring" test, which I don't entirely understand:

```
E               AssertionError: 'usag[1066 chars]tps://\n                        allennlp.s3.am[1452 chars]de\n' != 'usag[1066 chars]tps:/\n                        /allennlp.s3.am[1452 chars]de\n'
E               Diff is 4277 characters long. Set self.maxDiff to None to see it. : The documentation for the subcommand usage in the module allennlp.commands.elmo does not match the output of running `allennlp elmo --help`. Please update the docstring to match the output.
```

the second is `naqanet_test`:

```
Parameter: _question_weights_predictor.bias had incorrect gradient: zeros with shape ((1,))
```

I'll keep digging into these, but if anyone has any ideas, let me know

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -101,7 +101,7 @@ setup(name='allennlp',
       packages=find_packages(exclude=["*.tests", "*.tests.*",
                                       "tests.*", "tests"]),
       install_requires=[
-          'torch>=0.4.1,<1.2',
+          'torch>=1.2.0',
           "jsonnet>=0.10.0 ; sys.platform != 'win32'",
           'overrides',
           'nltk',
--- a/allennlp/common/testing/model_test_case.py
+++ b/allennlp/common/testing/model_test_case.py
@@ -179,7 +179,7 @@ class ModelTestCase(AllenNlpTestCase):
         original_dropouts: Dict[str, float] = {}
 
         if disable_dropout:
-            # clone model
+            # Remember original dropouts so we can restore them.
             for name, module in model.named_modules():
                 if isinstance(module, torch.nn.Dropout):
                     original_dropouts[name] = getattr(module, 'p')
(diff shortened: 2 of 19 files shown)


matt-gardner:
For the naqanet test, first things to try are checking for dropout in the test fixture, and seeing if the test is just flaky.  Other than that, not so many good ideas.

DeNeutoy:
Hmm i'll revert that docstring test. It looks broken and takes 60 seconds...

joelgrus:
ok, this is ready for review, I guess. the one part I'm not happy with (and that you won't be happy with) is that I couldn't figure out a way to get the biaffine_dependency_parser_multilang_test to not trigger the zero gradients, so I had to add a `gradients_to_ignore` (which is just a single bias parameter) to those tests. 

https://github.com/allenai/allennlp/pull/3182/files#diff-83e1bde8a01afa0c96f68ca171f3ba8bR11

if anyone has a better way to deal with that, I'm open to it.

setup.py
```diff
@@ -101,7 +101,7 @@
       packages=find_packages(exclude=["*.tests", "*.tests.*",
                                       "tests.*", "tests"]),
       install_requires=[
-          'torch>=0.4.1,<1.2',
+          'torch>=1.2.0',
```
brendan-ai2: Do we strictly need features from 1.2.0 or is 1.0 sufficient?

joelgrus: this PR is in response to breaking changes that happened between 1.1 and 1.2, but also going forward we'd like to use PyTorch's dataloader infrastructure, and for that to support allennlp datasets we need torch>=1.2

brendan-ai2: Got it. Thanks for explaining.

brendan-ai2:
@TalSchuster, I believe you wrote the biaffine_dependency_parser_multilang code. Strangely one of the tests for that code is producing zero gradients when we upgrade to PyTorch 1.2. Any insight on why that might be? We're tempted to ignore, but it's a bit unsettling.

PR #3183, opened 2019-08-21
Revert "Add a test for the subcommands docstring help outputs"

DeNeutoy:
Reverts allenai/allennlp#3172


DeNeutoy:
It's a bit brittle though, because if any library adds something that appears on import, these tests will break - and they change infrequently anyway. I didn't think about that at the time.

bryant1410:
> It's a bit brittle though, because if any library adds something that appears on import, these tests will break - and they change infrequently anyway. I didn't think about that at the time.

They shouldn't print to the standard output. pytorch-transformers changed this behavior, and now logs messages (stderr).

DeNeutoy:
Yeah, it doesn't matter that they shouldn't - they do, and many libraries will do this in the future.

bryant1410:
> Yeah, it doesn't matter that they shouldn't - they do, and many libraries will do this in the future.

Okay. But I don't think it's a big deal. There's a solution also, which is to ignore some prefix in the output and maybe also a suffix (basically to search for a substring).

PR #3190, opened 2019-08-22
Add exist_ok parameter to registrable.register decorator.

nelson-liu:


nelson-liu:
> I just made up this idea off the top of my head 😬

Should I not be trusting ideas that you make up off the top of your head? 😄 

> anyway, you should (1) add a docstring to register explaining what exist_ok does, and (2) add logging in the exist_ok case that basically says something like f"{name} has already been used to register {existing_class} but exist_ok is True so overwriting with {new_class}"

Good ideas, added.

allennlp/common/registrable.py
```diff
@@ -39,14 +39,31 @@ class Registrable(FromParams):
...
             if name in registry:
-                message = "Cannot register %s as %s; name already in use for %s" % (
-                        name, cls.__name__, registry[name].__name__)
-                raise ConfigurationError(message)
+                if exist_ok:
+                    message = ("%s has already been registered as %s, but "
+                               "exist_ok=True, so overwriting with %s" % (
+                                       name, registry[name].__name__, cls.__name__))
+                    logger.debug(message)
+                if not exist_ok:
```
joelgrus: "else"?

allennlp/common/registrable.py
```diff
@@ -39,14 +39,31 @@ class Registrable(FromParams):
...
         def add_subclass_to_registry(subclass: Type[T]):
             # Add to registry, raise an error if key has already been used.
             if name in registry:
-                message = "Cannot register %s as %s; name already in use for %s" % (
-                        name, cls.__name__, registry[name].__name__)
-                raise ConfigurationError(message)
+                if exist_ok:
+                    message = ("%s has already been registered as %s, but "
+                               "exist_ok=True, so overwriting with %s" % (
+                                       name, registry[name].__name__, cls.__name__))
```
joelgrus: can you change this to use f-strings (it would be more readable) (I know it was already like this)

allennlp/common/registrable.py
```diff
@@ -39,14 +39,31 @@ class Registrable(FromParams):
...
+                    logger.debug(message)
```
joelgrus: I think this should really be an `info` message (if not a `warning`), if it happens and you're not expecting it to, it could cause you a lot of grief

allennlp/common/registrable.py
```diff
@@ -39,14 +39,31 @@ class Registrable(FromParams):
...
+                if not exist_ok:
+                    message = "Cannot register %s as %s; name already in use for %s" % (
+                            name, cls.__name__, registry[name].__name__)
```
joelgrus: again, this would be more readable as an f-string

allennlp/common/registrable.py
nelson-liu: want me to change this one to info too? https://github.com/allenai/allennlp/blob/b1b19edcfa1d0e80e4004ed0a24da9790a42547d/allennlp/common/registrable.py#L73

allennlp/common/registrable.py
```diff
@@ -39,21 +39,37 @@ class Registrable(FromParams):
...
+                               f"name already in use for {registry[name].__name__}")
+                    raise ConfigurationError(message)
             registry[name] = subclass
             return subclass
         return add_subclass_to_registry
 
     @classmethod
     def by_name(cls: Type[T], name: str) -> Type[T]:
-        logger.debug(f"instantiating registered subclass {name} of {cls}")
+        logger.info(f"instantiating registered subclass {name} of {cls}")
```
bryant1410: @nelson-liu I wonder why this was changed, which is reverting #2004

matt-gardner: Looks like a bug to me.  PR welcome!

nelson-liu: Sorry yes, this is a bug, thanks for pointing it out.

PR #3194, opened 2019-08-24
 Modify bioul_tags_to_spans for ill-formed spans 

Maybewuss:
If a span is ill-formed, it will not be append to spans list instead of raise InvalidTagSequence.


dirkgr:
Is this still active? Can we close it, or do we need this change?

PR #3201, opened 2019-08-26
Language model classes for making predictions (both masked LM and next token LM)

matt-gardner:
These models are only tested for predictions, not for training.  The masked LM _might_ be useful for training, but also might just be too slow.  The next token LM _definitely_ is not for training.  These are noted in the class docstrings.


pidugusundeep:
@DeNeutoy is there a tutorial explaining the usage of the same?

PR #3206, opened 2019-08-27
Targeted hotflip attacks and beam search for input reduction

matt-gardner:
Depends on #3202 and #3201.  Implements targeted hotflip attacks, beam search for input reduction, and other minor cleanup.


matt-gardner:
The code was written by Eric and then reviewed / cleaned up by me, so as long as you don't see any glaring issues, we're probably good.

PR #3208, opened 2019-08-28
Replace tensorboardX with torch.utils.tensorboard

nelson-liu:
PyTorch 1.2.0 includes native Tensorboard support, so I switched over the `TensorboardWriter` to use that and removed the `tensorboardx` dependency.


matt-gardner:
Closing this. @nelson-liu (or anyone else), feel free to give it another go when it's ready.

PR #3223, opened 2019-09-06
checking in wtq training config

pdasigi:


matt-gardner:
Can you submit this as a PR to allennlp-semparse instead of here?  There's not a lot of point in adding new semparse code to this repo, given that we're going to be splitting it out, and I've already moved stuff over; trying to sync changes between the two repos would be a pain.

pdasigi:
Sure. Opened a PR there. Closing this one.

Issue #3217, opened 2019-09-04
Get top k tag sequences


mikerossgithub:
It would be nice to have the ranked choices for sequence decoding. I've done this locally, and can submit a PR if there is interest in my approach (below)

For my local repo, I used this code (which is in turn based on AllenNLP): https://gist.github.com/PetrochukM/afaa3613a99a8e7213d2efdd02ae4762

To use it, you modify a model to take a `top_k=1` parameter in its `__init__`, and then call top-k viterbi during forward if `self.top_k>1` . Then, with the predict command you add `--overrides '{"model": {"top_k": 10}}'` to get the top 10 sequences in your `output_dict`

This does not affect default behavior, and can be run on weight files created before the code change. Would a PR be useful?

schmmd:
@mikerossgithub we'd love a PR!

PR #3226, opened 2019-09-08
Add top-k sequence decoding

mikerossgithub:
Adds a top_k option to `viterbi_decode`, `ConditionalRandomField` and `CrfTagger` 

To maintain backwards compatibility the returned value types are different depending on whether `top_k` is supplied. Another option would be to implement `viterbi_decode` as a thin wrapper around a separate `top_k_viterbi_decode` function. I wasn't sure which was the preferred method.

This can be used for prediction on models serialized before this feature, e.g. by calling `allennlp predict --overrides {"model": "top_k": 10}` I wasn't sure where to document this use case.

The code is based on the implementation at https://gist.github.com/PetrochukM/afaa3613a99a8e7213d2efdd02ae4762 (which is in turn based on AllenNLPs implementation)

### DIFF
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -557,9 +557,9 @@ def viterbi_decode(
         path_indices.append(paths.squeeze())
 
     # Construct the most likely sequence backwards.
-    path_scores = path_scores[-1].view(-1)
-    max_k = min(path_scores.size()[0], top_k)
-    viterbi_scores, best_paths = torch.topk(path_scores, k=max_k, dim=0)
+    path_scores_v = path_scores[-1].view(-1)
+    max_k = min(path_scores_v.size()[0], top_k)
+    viterbi_scores, best_paths = torch.topk(path_scores_v, k=max_k, dim=0)
     viterbi_paths = []
     for i in range(max_k):
         viterbi_path = [best_paths[i]]
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -408,6 +408,7 @@ def viterbi_decode(
     tag_observations: Optional[List[int]] = None,
     allowed_start_transitions: torch.Tensor = None,
     allowed_end_transitions: torch.Tensor = None,
+    top_k: int = None,
 ):
     """
     Perform Viterbi decoding in log space over a sequence given a transition matrix
@@ -439,6 +440,10 @@ def viterbi_decode(
         An optional tensor of shape (num_tags,) describing which tags may transition *to* the
         end tag. If provided, additional transition constraints will be used for determining
         the end element of the sequence.
+    top_k : int, optional, (default = None)
+        Optional integer specifying how many of the top paths to return. For top_k>=1, returns
+        a tuple of two lists: top_k_paths, top_k_scores, For top_k==None, returns a flattened
+        tuple with just the top path and its score (not in lists, for backwards compatibility).
 
     Returns
     -------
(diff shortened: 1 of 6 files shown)


mikerossgithub:
Addresses #3217

mikerossgithub:
ready for review

allennlp/tests/nn/util_test.py
```diff
@@ -604,6 +606,94 @@ def test_viterbi_decode(self):
                                          observations)
         assert indices == [2, 3, 3, 0, 4, 3]
 
+    def test_viterbi_decode_top_k(self):
+        # Test cases taken from: https://gist.github.com/PetrochukM/afaa3613a99a8e7213d2efdd02ae4762
+
+        # Test Viterbi decoding is equal to greedy decoding with no pairwise potentials.
```
brendan-ai2: Nit: Each of these comments seems to indicate a particular property to test. It would be somewhat cleaner to break them out into separate test cases.

allennlp/nn/util.py
```diff
@@ -433,6 +434,10 @@ def viterbi_decode(tag_sequence: torch.Tensor,
         An optional tensor of shape (num_tags,) describing which tags may transition *to* the
         end tag. If provided, additional transition constraints will be used for determining
         the end element of the sequence.
+    top_k : int, optional, (default = None)
+        Optional integer specifying how many of the top paths to return. For top_k>=1, returns
+        a tuple of two lists: top_k_paths, top_k_scores, For top_k==None, returns a flattened
+        tuple with just the top path and its score (not in lists, for backwards compatibility).
```
brendan-ai2: Thanks for the backwards compatibility! :)

allennlp/tests/nn/util_test.py
mikerossgithub: breaking out into multi would break the similarity with the previous test and others in the file.

mikerossgithub:
@brendan-ai2 I've made the requested changes (except for the nit -- see comment).

My code passes `black (v19.3b0 as in requirements)` locally, but fails on team city, and will not report the diff changes so I can't tell what to change.  Can you accept as is, or provide further guidance?

Also, it might be good to change black in scripts/verify to use the --diff option, so users can see what the suggested changes are in the team city log.

mikerossgithub:
@brendan-ai2 All tests pass now. Ready for review.

allennlp/tests/models/crf_tagger_test.py
```diff
@@ -41,6 +41,18 @@ def test_forward_pass_runs_correctly(self):
...
+        self.model.top_k = 5
+        output_dict = self.model.decode(self.model(**training_tensors))
+        top_k_tags = [[x["tags"] for x in item_topk] for item_topk in output_dict['top_k_tags']]
+        first_choices = [x[0] for x in top_k_tags]
+        assert first_choices == output_dict['tags']
+        lengths = [len(x) for x in top_k_tags]
+        assert set(lengths) == {5}
+        tags_used = set(tag for item_top_k in top_k_tags for tag_seq in item_top_k for tag in tag_seq)
+        assert all(tag in {'O', 'I-ORG', 'I-PER', 'I-LOC'} for tag in tags_used)
+
```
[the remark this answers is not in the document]
mikerossgithub: done

mikerossgithub:
@brendan-ai2 Any chance for a final review?

allennlp/tests/nn/util_test.py
brendan-ai2: Fair point about the previous test.

Issue #3218, opened 2019-09-05
Predict during training


mikerossgithub:
**Is your feature request related to a problem? Please describe.**
I would like to log model predictions for every epoch, even though I am only keeping checkpoints of the latest 3 epochs. This enables post-analysis on the predictions without the disk overhead of logging model weights every epoch. One use case is to generate training curves for new metrics on old (expensive) training runs.

**Describe the solution you'd like**
Either a Predict callback or a "predict" field added to the Validate callback

I looked into subclassing or modifying the Validate callback, but since I need the `output_dict`, I would also need to alter `CallbackTrainer.batch_loss()` to return the `output_dict` instead of just returning loss. For now, I have written a hack that duplicates CallbackTrainer.batch_loss() functionality inside the callback.

matt-gardner:
This is an argument in favor of your original design @joelgrus, right?  Where basically everything was in callbacks?

joelgrus:
that design was too complex, so I'm not sure this is an argument for it.

here's I think a possible solution: modify `CallbackTrainer.batch_loss` so that it saves `output_dict` as part of the trainer state, then any callback can use it for whatever it wants:

https://github.com/allenai/allennlp/blob/master/allennlp/training/callback_trainer.py#L152

would that solve your problem?

mikerossgithub:
So CallbackTrainer would have a `most_recent_batch_output` variable? That would solve my particular problem! But two other changes would be nice. Neither of these are necessary, so if you don't think they are worth it, I would be happy with the solution you proposed!

1.
https://github.com/allenai/allennlp/blob/5e2206df306d3e11640032b523dc1a6f24a1c95b/allennlp/training/callbacks/validate.py#L61
This could be refactored to separate the somewhat non-trivial code for validation data traversal from the metric accumulation. Then the PredictCallback can just subclass and override the metric accumulation with prediction accumulation. As it stands, I need to copy this entire method into my PredictCallback and monitor it for future AllenNLP changes. 

2. 
Instead of `most_recent_batch_output`, one could just refactor batch_loss into `calc_loss(batch_forward(batch), for_training)` -- the PredictCallback can then get the output_dict with `batch_forward` without adding a transient state variable to the CallbackTrainer.

sai-prasanna:
[...]

PR #3230, opened 2019-09-09
potential callback modifications

mikerossgithub:
Some potential slight refactoring of CallbackTrainer to make it a bit more flexible for different kinds of callbacks (see issue #3218). The motivation for these changes are:

* simplifies callback's access to model’s output_dict (via `batch_group_forward()`), allowing (for example) serializing *validation* predictions during training from a callback

* most recent batch_output and batch_loss are now stored as trainer state, allowing (for example) serializing *training* predictions during training from a callback

* Simplifies the logic of `train_one_batch_group()` to just a sequence of setting trainer state variables without any `if training` special logic

* adds additional callback events before loss calculation and before backward pass (allowing advanced callbacks to modify loss or output_dict) — Not sure if this adds too many events, but the simplified logic of train_one_batch_group made it easy to add this.

### DIFF
--- a/allennlp/training/callback_trainer.py
+++ b/allennlp/training/callback_trainer.py
@@ -17,7 +17,6 @@ from allennlp.common.util import lazy_groups_of
 from allennlp.data.instance import Instance
 from allennlp.data.iterators.data_iterator import DataIterator, TensorDict
 from allennlp.models.model import Model
-from allennlp.nn import util as nn_util
 from allennlp.training import util as training_util
 from allennlp.training.callbacks.callback import Callback
 from allennlp.training.callbacks.callback_handler import CallbackHandler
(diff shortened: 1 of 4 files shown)


mikerossgithub:
@joelgrus I've been using the CallbackTrainer extensively and implementing a number of custom callbacks. Overall the callback framework is fantastic, and has made it easy to implement some use cases which would have been more awkward with the default trainer. But there have been a few minor pain points. I saw in the release notes that you requested feedback on the Callback framework. Should I add comments here, or is there a better way to submit feedback?

joelgrus:
comments here is fine, apologies for not getting to this PR earlier, I'll try to look at it today

joelgrus:
actually, I take that back, comments here is not good, open an issue for "callback trainer enhancements" and put them there

mikerossgithub:
Relevant: #3420

schmmd:
We're blocked on merging this until we figure out https://github.com/allenai/allennlp/issues/3519.  We're not sure what we're doing but we don't want to continue to support two trainers.

DeNeutoy:
I'm going to close this because it's getting stale, and we're still not sure about how we're going to manage the trainer. Thanks for the fixes though!

PR #3236, opened 2019-09-10
Use a NVIDIA base image

schmmd:
This is an attempt to re-add https://github.com/allenai/allennlp/pull/3177.


schmmd:
I tested this and confirmed that the build works on my machine.  In addition to fixing the error in #3177 I also previously merged a commit that causes a particular test to fail in a system dependent way (which was making it hard to test this change).

Here's the original motivation:

This yields a smaller base image with reduced complexity around NVIDIA/CUDA (we simply use the base image) but increased complexity around what version of Python we're installing (we need to manage it more manually rather than just using the base image).

This change would make it easier to run the `allennlp` Docker image on a GPU because it wouldn't require that the CUDA 10.0 drivers are installed--the image packages the drivers itself.  For example, `allennlp-server-3` has the 10.1 drivers installed.  With the current Docker image we have a `CUDNN_STATUS_EXECUTION_FAILED` but training works successfully on a GPU with this image.

```
allennlp                                                  nvidia-base-18.04                          b8b191a7a08d        2 hours ago         3.66GB
```

Dockerfile.pip
```diff
@@ -1,27 +1,30 @@
 # This Dockerfile creates an environment suitable for downstream usage of AllenNLP.
 # It creates an environment that includes a pip installation of allennlp.
 
-FROM python:3.6.8-stretch
+FROM nvidia/cuda:10.0-base-ubuntu18.04
```
brendan-ai2: Doesn't the other dockerfile need these same changes?

Dockerfile.pip
```diff
@@ -1,27 +1,30 @@
...
-ENV NVIDIA_VISIBLE_DEVICES all
-ENV NVIDIA_DRIVER_CAPABILITIES compute,utility
-LABEL com.nvidia.volumes.needed="nvidia_driver"
-
 WORKDIR /stage/allennlp
 
 ARG VERSION
 ARG SOURCE_COMMIT
 
+RUN apt-get update && apt-get install -y --no-install-recommends \
```
brendan-ai2: And these?

brendan-ai2:
So what will be the behavior now if there's a mismatch in the drivers between the host and the image? No problem as the host drivers are ignored completely or something else?

schmmd:
@brendan-ai2 if there's a mismatch then the job will run instead of crashing.  I can't guarantee this for all combinations.

I can make a similar change for our `Dockerfile`, but it doesn't need them as it's only used for our CI.  It might be a good idea for consistency--and I'd be happy to do that if you thought it were a good idea.

brendan-ai2:
It's necessary to make the same changes for `Dockerfile` as we have GPU unit tests: http://build.allennlp.org/viewType.html?buildTypeId=AllenNLP_GpuUnitTests

Please also update the documentation at https://github.com/allenai/allennlp-internal/blob/master/docs/TeamCity.md#gpu-setup! :)

schmmd:
@brendan-ai2 the GPU unit tests will continue to work without additional changes.  You can see they did work as this PR passed CI.  Why do you think it's necessary to make additional changes?

That said--I'm happy to if you think it's a good idea.

Dockerfile
```diff
@@ -3,15 +3,6 @@ FROM python:3.6.8-stretch
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 
-ENV PATH /usr/local/nvidia/bin/:$PATH
```
brendan-ai2: So none of these lines are required to make GPUs on Docker work?

brendan-ai2:
The GPU unit tests only run on master as we don't have enough GPUs on that box to test every PR. The docs I linked mention that fine point. We should definitely keep these Dockerfiles in sync. Otherwise we'll have to maintain two ways of doing things which will be more work.

schmmd:
Happy to put them in sync.

Dockerfile.pip
```diff
@@ -1,27 +1,30 @@
...
+         git \
+         python3.6-dev \
+         python3-setuptools \
+         python3-pip && \
+     rm -rf /var/lib/apt/lists/*
+
+RUN ln -s /usr/bin/python3 /usr/bin/python
+RUN ln -s /usr/bin/pip3 /usr/bin/pip
+
+RUN pip install wheel
```
brendan-ai2: Should this be in the other Dockerfile too? (Sorry if I missed it...)

schmmd: I was wondering about that too.  I pushed a commit to run a build on the server and see if it succeeded, but now I'm getting some errors about Black ;-)

schmmd: I think it's only be required for `Dockerfile.pip` for `pip install`-ing from a GitHub repository.

schmmd:
@viveklak pointed out that Google recommends this base image for GPU jobs on GKE.

https://cloud.google.com/kubernetes-engine/docs/how-to/gpus
```
About the CUDA libraries
CUDA® is NVIDIA's parallel computing platform and programming model for GPUs. The NVIDIA device drivers you install in your cluster include the CUDA libraries.

CUDA libraries and debug utilities are made available inside the container at /usr/local/nvidia/lib64 and /usr/local/nvidia/bin, respectively.

CUDA applications running in Pods consuming NVIDIA GPUs need to dynamically discover CUDA libraries. This requires including /usr/local/nvidia/lib64 in the LD_LIBRARY_PATH environment variable.

You should use Ubuntu-based CUDA Docker base images for CUDA applications in GKE, where LD_LIBRARY_PATH is already set appropriately. The latest supported CUDA version is 10.0 on both COS (1.11.5+) and Ubuntu (1.11.8-gke.4+, 1.12.6-gke.6+).
```

Dockerfile
schmmd: They are--but they are in the `nvidia/cuda:10.0-base-ubuntu18.04` base image.

PR #3238, opened 2019-09-11
Add additive attention & unittest

hawkeoni:
Added Additive Attention module (Also known as Concat attention) from  https://arxiv.org/pdf/1409.0473.pdf

### DIFF
--- a/allennlp/common/params.py
+++ b/allennlp/common/params.py
@@ -5,7 +5,8 @@ logging and validation.
 """
 
 from typing import Any, Dict, List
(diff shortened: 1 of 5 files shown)


hawkeoni:
There is one major difference in the second commit: pytest 5.1.2 on python 3.7.4 fails every tests because of Deprecation warning. The reason is that MutableMapping should be imported from collections.abc, and not from collections. I fixed it in allennlp/common/params.py. Hopefully all tests will pass now.

PR #3239, opened 2019-09-11
Make Instance in charge of when to re-index

matt-gardner:
The `Instance` has a flag for whether or not its fields have been indexed, and our fields are mixed regarding whether or not they have their own check (e.g., `TextFields` always re-index when `index()` is called, while the ones in this PR don't).  I think it makes the most sense to have `Instance` own this, as it gives a user one place to reset the `Instance`, without having to check for all kinds of crazy places to know how to reset things (we need to do this in the interpret code).

This PR makes all `Fields` consistent: when `index()` is called, the field is re-indexed.  It's the `Instance`'s job to make sure this only happens when necessary.

### DIFF
--- a/allennlp/data/fields/adjacency_field.py
+++ b/allennlp/data/fields/adjacency_field.py
@@ -90,8 +90,9 @@ class AdjacencyField(Field[torch.Tensor]):
 
     @overrides
     def index(self, vocab: Vocabulary):
-        self._indexed_labels = [vocab.get_token_index(label, self._label_namespace)
-                                for label in self.labels]
+        if self.labels is not None:
+            self._indexed_labels = [vocab.get_token_index(label, self._label_namespace)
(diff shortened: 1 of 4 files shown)


matt-gardner:
Ok, it's good to look at now.

allennlp/data/fields/adjacency_field.py
```diff
@@ -90,9 +90,8 @@ def count_vocab_items(self, counter: Dict[str, Dict[str, int]]):
 
     @overrides
     def index(self, vocab: Vocabulary):
-        if self._indexed_labels is None and self.labels is not None:
```
matt-gardner: I'm actually not sure about this - @DeNeutoy, you know more about what this field does.  It looks like I should probably keep a check here for `self.labels is not None`?

PR #3248, opened 2019-09-14
Assorted fixes for run_with_beaker.py

brendan-ai2:
- Compile Jsonnet config locally so that `import` works.
- Add `--include-package` option so we can use the script with other repos.
- Add `--overrides`.

### DIFF
--- a/scripts/ai2-internal/run_with_beaker.py
+++ b/scripts/ai2-internal/run_with_beaker.py
@@ -22,7 +22,7 @@ from allennlp.common.params import Params
 def main(param_file: str, args: argparse.Namespace):
     commit = subprocess.check_output(["git", "rev-parse", "HEAD"], universal_newlines=True).strip()
     docker_image = f"allennlp/allennlp:{commit}"
-    overrides = ""
+    overrides = args.overrides
 
     # Reads params and sets environment.
     ext_vars = {}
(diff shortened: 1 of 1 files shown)


scripts/ai2-internal/run_with_beaker.py
```diff
@@ -56,21 +62,31 @@ def main(param_file: str, args: argparse.Namespace):
...
-            "-m",
-            "allennlp.run",
-            "train",
-            "/config.json",
-            "-s",
-            "/output",
-            "--file-friendly-logging"
+    # Arguments that differ between preemptible and regular machine execution.
+    if args.preemptible:
+        allennlp_prefix = ["/stage/allennlp/resumable_train.sh", "/output", "/config.json"]
```
DeNeutoy: Currently the base docker image doesn't have this script I think, only the pip Dockerfile. Is that a problem? I think this script supports building your own image - I don't know if you can raise an error or something if `premptible` is passed without specifying a particular image that contains that script or something ...

brendan-ai2: Maybe I'm misunderstanding, but I see it in both `Dockerfile` and `Dockerfile.pip`.
https://github.com/allenai/allennlp/blob/master/Dockerfile#L68
https://github.com/allenai/allennlp/blob/master/Dockerfile.pip#L34

If someone does define a Dockerfile without `resumable_train.sh` and they use `--preemptible`, they'll get an error as soon as they run. This seems reasonable to me. My rationale is that there are many other dependencies that someone could forget to include. We could add checks one by one, but this will be error prone and extra code. Just running the image is simpler and more robust.

brendan-ai2:
Thanks for the review!

scripts/ai2-internal/run_with_beaker.py
DeNeutoy: Oh! completely missed that when I checked the Dockerfile. LGTM.

PR #3253, opened 2019-09-16
allow implicit package imports

joelgrus:
this was an idea that came out of the allennlp summit, basically you can specify types as fully qualified classnames "mypackage.mymodule.readers.MyReader" and then not need to use include-package.

### DIFF
--- a/allennlp/common/registrable.py
+++ b/allennlp/common/registrable.py
@@ -71,12 +71,12 @@ class Registrable(FromParams):
     @classmethod
     def by_name(cls: Type[T], name: str) -> Type[T]:
         logger.info(f"instantiating registered subclass {name} of {cls}")
-        parts = name.split(".")
         if name in Registrable._registry[cls]:
             return Registrable._registry[cls].get(name)
-        elif len(parts) > 1:
+        elif "." in name:
             # This might be a fully qualified class name, so we'll try importing its "module"
             # and finding it there.
+            parts = name.split(".")
             submodule = ".".join(parts[:-1])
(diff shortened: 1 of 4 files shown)


allennlp/common/registrable.py
```diff
@@ -70,9 +71,31 @@ def add_subclass_to_registry(subclass: Type[T]):
...
+                raise ConfigurationError(f"tried to interpret {name} as a path to a class "
+                                         f"but unable to import module {submodule}")
+
+            try:
+                return getattr(module, class_name)
+            except AttributeError:
+                raise ConfigurationError(f"tried to interpret {name} as a path to a class "
+                                         f"but unable to find class {class_name} in {submodule}")
+
+        else:
```
DeNeutoy: This error should provide a way for people to discover how to do this, e.g:

```
ConfigurationError(f"{name} is not a registered name for {cls.__name__}. Alternatively you can register Models etc with fully qualified paths, e.g my_module.my_submodule.ClassName so allennlp can find and import them automatically.")
```

Additionally, if someone has tried to register their model but failed, this is going to show up most here:

https://github.com/allenai/allennlp/blob/0bd331942d824f5072bf7b6adffb1172f6ffe507/allennlp/common/params.py#L349

I wonder if it's worth adding a note to that error that you can do this as well?

joelgrus: in fact, I had to change this function, because otherwise there's no way for a fully-qualified class name to successfully pop_choice. good catch

Issue #3255, opened 2019-09-17
Unidirectional LM incompatible with Predictor


viking-sudo-rm:
**Describe the bug**
When I wrap the default unidirectional `LanguageModel`, `predict_instance` throws an error because the backward loss returned by `LanguageModel` is `None`.

For example:
```python
predictor = Predictor(model, reader)
for inst in iter(dataset):
    results = predictor.predict_instance(inst)
```
Yields:
```
  File "/home/willm/code/lm-distill/distill.py", line 12, in sample_trajectories
    results = predictor.predict_instance(inst)
  File "/home/willm/anaconda3/envs/bert-parsing/lib/python3.7/site-packages/allennlp/predictors/predictor.py", line 181, in predict_instance
    outputs = self._model.forward_on_instance(instance)
  File "/home/willm/anaconda3/envs/bert-parsing/lib/python3.7/site-packages/allennlp/models/model.py", line 124, in forward_on_instance
    return self.forward_on_instances([instance])[0]
  File "/home/willm/anaconda3/envs/bert-parsing/lib/python3.7/site-packages/allennlp/models/model.py", line 167, in forward_on_instances
    elif len(output) != batch_size:
TypeError: object of type 'NoneType' has no len()
```
I expected the `Predictor` to work normally in these conditions; perhaps the unidirectional language model should not add backward loss to the output dict, rather than adding a `None` entry?

See https://discourse.allennlp.org/t/predictor-for-languagemodel/38 for more info.

**To Reproduce**
Steps to reproduce the behavior
1. Initialize a `LanguageModel` with a unidirectional contextualizer.
2. Wrap `predictor = Predictor(model, reader)`
3. `predictor.predict_instance(inst)` will fail

**Expected behavior**
I expect to be able to use the default `Predictor` to get predictions from my unidrectional language model. My humble suggestion is that the unidirectional language model should not add backward loss to the output dict, rather than adding a `None` entry?

**System (please complete the following information):**
 - OS: Linux
 - Python version: 3.7
 - AllenNLP version: I installed from master
 - PyTorch version: 1.1.0

**Additional context**
Current work-around is:

```python
from overrides import overrides
import torch

from allennlp.models.language_model import LanguageModel


class SanitizedLM(LanguageModel):

    """Remove null keys from the output of a language model."""

    @overrides
    def decode(self, output_dict):
        return {key: value for key, value in output_dict.items()
                if value is not None}
```

viking-sudo-rm:
[...]

PR #3256, opened 2019-09-17
Unidirectional LM doesn't return backward loss.

viking-sudo-rm:
Changed `LanguageModel` so that, if there is no backward loss, then no backward loss key is added to the state dict. This was causing an issue when using `Predictor` with trained language models.

Fixes https://github.com/allenai/allennlp/issues/3255.

### DIFF
--- a/allennlp/tests/models/language_model_test.py
+++ b/allennlp/tests/models/language_model_test.py
@@ -6,7 +6,6 @@ from allennlp.common.testing import ModelTestCase
 from allennlp.common.checks import ConfigurationError
 from allennlp.common.params import Params
 from allennlp.models import Model
-from allennlp.predictors import Predictor
 
 
 class TestUnidirectionalLanguageModel(ModelTestCase):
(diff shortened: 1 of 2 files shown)


brendan-ai2:
@viking-sudo-rm , looks like there were some test failures. Could you address those? Let me know if you can't access TeamCity (via the "Details" link above) and I'll help you get set up.

allennlp/tests/models/language_model_test.py
```diff
@@ -86,6 +87,8 @@ def setUp(self):
 
         self.expected_embedding_shape = (2, 8, 14)
         self.bidirectional = True
+        self.result_keys = {"loss", "forward_loss", "backward_loss", "lm_embeddings",
```
brendan-ai2: `self.result_keys.add("backward_loss")` might call attention to the difference a bit better, but this is fine too.

allennlp/tests/models/language_model_test.py
```diff
@@ -44,7 +46,6 @@ def test_forward_pass_runs_correctly(self):
                                            decimal=3)
         else:
             np.testing.assert_almost_equal(loss, forward_loss, decimal=3)
-            assert result["backward_loss"] is None
```
brendan-ai2: How about we add a test case here that calls `forward_on_instances` and verifies that it doesn't crash? This matches the bug that you're fixing with this PR and I believe will address the PR test coverage requirement. In general it's good practice to add a test when you fix a bug to prevent regressions. It also makes it clear to people reading the code that the bug was really addressed.

viking-sudo-rm:
I added a test case for the issue I was facing with `forward_on_instance`. The diff test coverage is still low though, because the `average_loss` case (half of my changed lines) is not tested. This case was not tested before my changes, and it's not obvious to me how to add test coverage for it.

viking-sudo-rm:
"Merging is blocked"

Seems you need to override it

brendan-ai2:
Huh, looks like it's going through one more CI run (likely due to merging master). Could you please ping me if the problem persists after the final run?

Issue #3262, opened 2019-09-19
DROP Evaluation - Wrong Partition to Types


eladsegal:
In https://github.com/allenai/allennlp/blob/master/allennlp/tools/drop_eval.py#L193, max_type can be updated even though it might not be the best match.

For example, if `candidate_answers` has `number: 5` and `spans: ["7-meters"]` in that order and the prediction is `"5-yard"` then we get the following scores per each answer:
- `number: 5`: `em: 0`, `f1: 0.67`
- `spans: ["7-meters"]`: `em: 0`, `f1: 0`

and the max_type will be `span` instead of `number`.

I assume that the original intent for the condition was to have `and` instead of  `or`.
https://github.com/allenai/allennlp/pull/3263 fixes the issue.

schmmd:
Fixed by #3263.  Thanks!

PR #3263, opened 2019-09-19
Fix wrong partition to types in DROP evaluation

eladsegal:
This pull request is to address https://github.com/allenai/allennlp/issues/3262.

The problem was that max_type was updated even if the candidate answer didn't have metrics that maximized both em and f1, due to use of `or` instead of `and`.

### DIFF
--- a/allennlp/tests/tools/drop_eval_test.py
+++ b/allennlp/tests/tools/drop_eval_test.py
@@ -147,10 +147,17 @@ class TestDropEvalFunctional:
         annotation = {"pid1": {"qa_pairs":[{"answer": {"number": "5"}, "validated_answers": \
                                                         [{"spans": ["7-meters"]}], "query_id":"qid1"}]}}
         prediction = {"qid1": "5-yard"}
-
         with io.StringIO() as buf, redirect_stdout(buf):
             evaluate_json(annotation, prediction)
             output = buf.getvalue()
+        lines = output.strip().split("\n")
+        assert lines[4] == 'number: 1 (100.00%)'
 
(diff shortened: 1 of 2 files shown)


matt-gardner:
Hey, thanks for this!  Can you add a test case that fails before this change, and passes afterward?

eladsegal:
No problem!
Added simple tests, let me know if something more comprehensive is required

allennlp/tests/tools/drop_eval_test.py
```diff
@@ -139,3 +142,22 @@ def test_json_loader(self):
                                         {"answer": {"spans": ["answer2"]}, "query_id":"qid2"}]}}
         prediction = {"qid1": "answer", "qid2": "answer2"}
         assert evaluate_json(annotation, prediction) == (0.5, 0.5)
+
+    def test_type_partition_output(self):
+        annotation = {"pid1": {"qa_pairs":[{"answer": {"number": "5"}, "validated_answers": \
+                                                        [{"spans": ["7-meters"]}], "query_id":"qid1"}]}}
+        prediction = {"qid1": "5-yard"}
+        with io.StringIO() as buf, redirect_stdout(buf):
```
matt-gardner: I'm a bit confused about why you did this this way, instead of just doing `assert evaluate_json(annotation, prediction) == (...)`, as above.  Why doesn't that kind of test work?

eladsegal: The output of `evaluate_json` is just `(global_em, global_f1)`. 
The change for `max_type` only affects the dictionaries `type_to_em` and `type_to_f1` which are only used in the printed output.

matt-gardner: Ok, great, thanks, I understand this better now.

PR #3266, opened 2019-09-19
upgrade to latest pylint

joelgrus:
most of the changes I needed to make were replacing `set([...])` with set comprehensions or literals. there are a few various other small things here and there.

I ended up disabling three of the new rules:

1. `logging-fstring-interpolation`

apparently the latest pylint doesn't like f-strings in logging calls. as far as I can tell there's no good reason for this, and it seems like a stupid rule, so I disabled it.

2. `unnecessary-pass`

this caused spurious warnings when e.g. we had a do-nothing method in a base class that subclasses could override to do something.

3. `no-else-raise`

the latest pylint doesn't like `else` statements after `raise` statements, e.g.

```
if something_bad:
    raise Exception("something bad happened")
else:
    do_something_good()
```

I get where it's coming from ("flat is better than nested"), but many times these else's make the code more readable. I got rid of some of them where I thought it made sense, but I also disabled the rule.

--

I put only a lower bound on the new pylint, so at some point it will upgrade and we'll be forced to fix whatever new problems arise, rather than waiting 2 years until someone feels motivated to do it.

--

In addition, it turns out the latest pylint has incompatible dependencies with the old version of mypy (which for mysterious reasons puts an upper bound on `typed-ast`), which is why the first build failed. I added a lower bound on `typed-ast`, which is on paper incompatible with our old mypy but in practice still works. Hopefully we can upgrade mypy next.


allennlp/semparse/domain_languages/nlvr_language.py
```diff
@@ -486,163 +486,163 @@ def object_count_lesser_equals(self, objects: Set[Object], count: int) -> bool:
 
     @predicate
     def object_color_count_equals(self, objects: Set[Object], count: int) -> bool:
-        return len(set([obj.color for obj in objects])) == count
+        return len({obj.color for obj in objects}) == count
 
     @predicate
     def object_color_count_not_equals(self, objects: Set[Object], count: int) -> bool:
-        return len(set([obj.color for obj in objects])) != count
+        return len({obj.color for obj in objects}) != count
```
DeNeutoy: RIP

PR #3267, opened 2019-09-19
Advertise Discourse on the issue template

schmmd:
Very open to suggestions on the wording ;-)


schmmd:
I don't think we're entirely clear within our group--and what we want to do hinges on whether Discourse is successful in building more community engagement--which is TBD.

PR #3273, opened 2019-09-22
Reduce number of samples in smoothgrad

Eric-Wallace:
Smoothgrad is really slow with 25 samples, speed it up with only 10.


Eric-Wallace:
@matt-gardner

matt-gardner:
This looks like it really should have been made a constructor parameter, by the way.

PR #3277, opened 2019-09-24
Fix crash when hotflip gets OOV input

matt-gardner:


allennlp/interpret/attackers/hotflip.py
```diff
@@ -306,14 +308,17 @@ def _first_order_taylor(self, grad: numpy.ndarray,
         first-order taylor approximation of the loss.
         """
         grad = torch.from_numpy(grad)
-        embedding_matrix = embedding_matrix.cpu()
-        word_embeds = torch.nn.functional.embedding(torch.LongTensor([token_idx]),
-                                                    embedding_matrix)
-        word_embeds = word_embeds.detach().unsqueeze(0)
+        if token_idx >= self.embedding_matrix.size(0):
```
Eric-Wallace: maybe add some comments here on what and why we are doing this.

PR #3281, opened 2019-09-24
upgrade to latest mypy + pylint

joelgrus:
after I merged my last PR, a new version of pylint dropped, so I had to make a bunch of changes for that.

regarding mypy, the biggest thing was that I had to add the `--no-site-packages` flag; otherwise I got a *ton* of errors related to mypy partially understanding pytorch code. you can see all of them here:

https://gist.github.com/joelgrus/601621717f2fb5b2b7d03f29f47926ff

(and in particular you can see that they're not helpful)

I am still trying to understand if there are any further consequences of using that flag, but I'm opening a PR to make sure that CI doesn't catch anything

in theory I should have been able to add a section to mypy.ini that did something like

```
[mypy-torch]
ignore_missing_imports = True
```
but I couldn't get it to work. (maybe it only works for modules that we own?)


joelgrus:
as discussed online, I think that adding the new flags to mypy is ok, it still seems to catch the errors you'd want it to catch.

also, I'm very sour on pylint, in this new version they introduced random rules literally because someone read someone's blog post saying "don't do X". once this is merged I'm going to take a look at switching to flake8, which seems more highly regarded (and faster)

PR #3289, opened 2019-09-26
Use Python3-style super call

bryant1410:


matt-gardner:
Oh, looks like there are some spacing issues to fix after this.

Issue #3287, opened 2019-09-26
ValueError on 'sanitize' True of type <class 'numpy.bool_'> (util.py)


12seetharaman:
**Description**
"numpy.bool_" type was not included on Sanitize to convert it into python type. 
```
x = True   ## type(x) ->  numpy.bool_ 

    def sanitize(x: Any) -> Any:  # pylint: disable=invalid-name,too-many-return-statements
        """
        Sanitize turns PyTorch and Numpy types into basic Python types so they
        can be serialized into JSON.
        """
        if isinstance(x, (str, float, int, bool)):
            # x is already serializable
            return x
        elif isinstance(x, torch.Tensor):
            # tensor needs to be converted to a list (and moved to cpu if necessary)
            return x.cpu().tolist()
        elif isinstance(x, numpy.ndarray):
            # array needs to be converted to a list
            return x.tolist()
        elif isinstance(x, numpy.number):  # pylint: disable=no-member
            # NumPy numbers need to be converted to Python numbers
            return x.item()
        elif isinstance(x, dict):
            # Dicts need their values sanitized
            return {key: sanitize(value) for key, value in x.items()}
        elif isinstance(x, (spacy.tokens.Token, allennlp.data.Token)):
            # Tokens get sanitized to just their text.
            return x.text
        elif isinstance(x, (list, tuple)):
            # Lists and Tuples need their values sanitized
            return [sanitize(x_i) for x_i in x]
        elif x is None:
            return "None"
        elif hasattr(x, 'to_json'):
            return x.to_json()
        else:
>           raise ValueError(f"Cannot sanitize {x} of type {type(x)}. "
                             "If this is your own custom class, add a `to_json(self)` method "
                             "that returns a JSON-like object.")
E           ValueError: Cannot sanitize True of type <class 'numpy.bool_'>. If this is your own custom class, add a `to_json(self)` method that returns a JSON-like object.
```

**Call log**
```
env/lib/python3.7/site-packages/allennlp/predictors/predictor.py:213: in predict_batch_instance
    return sanitize(outputs)
env/lib/python3.7/site-packages/allennlp/common/util.py:73: in sanitize
    return [sanitize(x_i) for x_i in x]
env/lib/python3.7/site-packages/allennlp/common/util.py:73: in <listcomp>
    return [sanitize(x_i) for x_i in x]
env/lib/python3.7/site-packages/allennlp/common/util.py:67: in sanitize
    return {key: sanitize(value) for key, value in x.items()}
[...]

PR #3290, opened 2019-09-27
sanitize function fix to convert numpy.bool_ to bool

12seetharaman:
Added check, to convert numpy.bool_ to python bool. 
Test passed.
#3287


joelgrus:
thanks for the fix!

PR #3294, opened 2019-09-28
fix an annotation issue in basic_transition_function.py

entslscheia:
I also update self._input_projection_layer = Linear(output_dim + action_embedding_dim, input_dim) to self._input_projection_layer = Linear(encoder_output_dim + action_embedding_dim, input_dim). Though as a matter of fact, output_dim equals encoder_output_dim, using encoder_output_dim is more consistent in this context.


matt-gardner:
Though, we're moving the semantic parsing code to [allennlp-semparse](https://github.com/allenai/allennlp-semparse).  So I'm going to merge this code over there after merging it here.  All future PRs for the semantic parsing code should go to that repo.

PR #3301, opened 2019-09-30
`source activate` -> `conda activate` in README

bryant1410:


README.md
```diff
@@ -143,7 +143,7 @@ the full suite of tests used by our continuous build environment.
 Once you've installed AllenNLP, you can run the command-line interface either
 with the `allennlp` command (if you installed via `pip`) or `allennlp` (if you installed via source).
 
-```bash
+```
```
bryant1410: This code block is not Bash itself as it contains a command and its output.

README.md
```diff
@@ -80,7 +80,7 @@ environment you want to use, you can skip to the 'installing via pip' section.
 3.  Activate the Conda environment. You will need to activate the Conda environment in each terminal in which you want to use AllenNLP.
 
     ```bash
-    source activate allennlp
+    conda activate allennlp
```
DeNeutoy: I actually don't prefer this, although it is conda's recommendation. The reason is that in some shells, you cannot use `conda activate allennlp` as a command. `source` is also more backward compatible with older versions of conda. The other changes in the PR are fine though!

bryant1410: It's the conda recommended way tho, I don't think allennlp should give opinionated recommendations to people on conda as it's kind of unrelated to the purpose of it. I believe it's better to stick to whatever it's the standard and recommended option, but as you guys prefer.

They claim to support at least bash, fish, powershell, tcsh, xonsh and zsh for the init script (see `conda init --help`), which I think it covers a lot.

As a reference, it was also discussed in allenai/allennlp-demo#271. If a shell script gives some problems, such as within Slurm, `eval "$(conda shell.bash hook)"` has been a solution to me.

Should I change it then?

DeNeutoy: sure, if we've already merged something related that's fine. It's personal preference.

PR #3307, opened 2019-09-30
fix flake8 from #3092

DeNeutoy:


bryant1410:
I was creating a PR right now :laughing: 

#3092 (and the changes from #3288) made me think that maybe the merge to master should be blocked unless the PR is up to date? Or maybe not, as this type of errors are uncommon? Because this happens after a non-up-to-date PR is passing the checks but when it's merged (or rebased) without conflicts by GitHub, it actually doesn't pass the tests.

Also, shouldn't all "pylint" comments be removed by now?

bryant1410:
(GitHub allows you to enforce the check I mentioning)

As a drawback, I've seen that if you want to merge a bunch of PRs, you need to merge each one first (which can be done from GitHub), wait for the checks to pass after each one (again, wait for each one) and then merge. It makes it slower if you want to merge multiple stuff, but safer in terms of not breaking master.

DeNeutoy:
Yes - we have it like this because often there are genuinely unrelated changes, and typically the development team on allennlp are aware enough to only do this for truly unrelated changes, like readmes etc. We'll leave it as is for now, because our build takes a long time and sometimes being able to merge changes like this is helpful.

I think you're right about pylint though, let me update this.

bryant1410:
> Yes - we have it like this because often there are genuinely unrelated changes, and typically the development team on allennlp are aware enough to only do this for truly unrelated changes, like readmes etc. We'll leave it as is for now, because our build takes a long time and sometimes being able to merge changes like this is helpful.

Nice! It makes sense.

> I think you're right about pylint though, let me update this.

Maybe all pylint comments should be removed now?

DeNeutoy:
Oh - actually I think we were keeping these in case we need to re-enable a subset of the pylint tests. @joelgrus was that right?

joelgrus:
no, I thought I removed all the pylint comments, but it's possible I missed some?

matt-gardner:
Sorry!  I merged some PRs today without updating them, forgetting about the flake8 change that made it so that was no longer safe.

bryant1410:
@joelgrus these come from #3092, after the flake8 change.

PR #3308, opened 2019-09-30
for discussion: incorporate black code formatter

joelgrus:
1. add `black` dependency + appropriate checks to verify scripts

as discussed, I set the max line length for black as 100 (so that it will try to make lines shorter than this when it can) but I left flake8 set to 115 (since we have a lot of lines that are longer than this that black can't touch)

2. apply `black` to allennlp codebase 

> 708 changed files with 35,697 additions and 23,073 deletions.

3. add `pre-commit` dev dependency
4. add `.pre-commit-config.yaml` with pre-commit hooks for black and flake8
    
This doesn't do anything unless/until you run `pre-commit install`. If you do that it will install pre-commit hooks for black and flake8. What this means (I think) is that when you `git commit` it will run black and flake8 against your code before committing, and if either one errors out, it will abort the commit. In general black shouldn't error out, it will just format your code, but flake8 will error out if your code doesn't pass the flake8 check. 

If you only want one of these checks, you can modify the yaml file before running `pre-commit install`. If you don't want either of these checks, then do nothing; they have to be installed manually by you.

Both of these checks should only run on the files that were changed in the commit, so it should be fast most of the time.


matt-gardner:
Is there a way to have `black` run automatically on all PRs?

joelgrus:
yes, we would just add another step to the build on teamcity

DeNeutoy:
FYI we should be careful about when this automatically runs. It could confuse new contributors if it runs on all new PRs, as you'd push code to a PR, black would run, then you'd try to push again and git would say that the remote had changed. I'm in favour if it can run when we merge into master, but hopefully it won't run after every commit. Does that make sense?

joelgrus:
teamcity would just run `black --check`, we don't want teamcity making changes to files (I'm not even sure if that's possible)

DeNeutoy:
Oh - i'd assumed that was what Matt meant. We should defo have it as a teamcity check.

joelgrus:
oh no, black moved all the `# type: ignore`s to the wrong lines, mypy is not happy

matt-gardner:
A failure case I'd like to avoid is people failing CI because of flake8 errors that would have been fixed by running black.  If we can have black run automatically so that they don't even see the flake8 errors, that would be ideal.  I'm not sure how to accomplish that, though.

Also, yeah, that's too bad about the mypy stuff.  I was afraid that would happen.

PR #3314, opened 2019-10-02
Add question id to MetadataField

mkardas:
According to docstring of SquadReader the instance's ID should be
accessible as metadata['id'].


matt-gardner:
Thanks for this!  We're in the process of moving the reading comprehension data and model code into a separate repository, however.  Can I get you to open this PR on the [new repository](https://github.com/allenai/allennlp-reading-comprehension) instead?

matt-gardner:
Well, I'm going to merge this here now.  We need to remember to also merge it in the other repository, however; it'd be really helpful if you could do it.

PR #3323, opened 2019-10-04
Upgrade Dockerfiles to Python 3.7.

schmmd:
I don't particularly need to merge this as is, but I was curious what work it would take to use Python 3.7 as our base distribution.  While this change doesn't prevent using AllenNLP with Python 3.6, with it we would eventually probably introduce incompatible features that would force 3.7.

If we don't merge this, I'll at least merge my documentation changes and additional mypy ignore, which are good additions even if we continue to use Python 3.6 as our base.


schmmd:
We're probably going to hold off on upgrading to Python 3.7--but do chime in on why you'd be excited about this @12seetharaman.  I am going to make sure we make the documentation changes to reference 3.7 instead of 3.6.

PR #3328, opened 2019-10-06
Expose -e flag in EVALB scorer

nelson-liu:


nelson-liu:
(looks like this is broken since 91d684580042abfbf271f27f6775c4093f119d41 is broken)

PR #3344, opened 2019-10-10
fix(from_params): load fullnamed clases from type parameter

frascuchon:
This PR includes capabilities of implicit package imports to `type` definitions. It's a great functionality but, for `type` parameters is not working.

This minimal change fixes that problem an allow avoid decorators for `allennlp.models.Model` or `allennlp.data.DatasetReader`


schmmd:
@frascuchon can you add a test so we can clearly see what this fixes?  Presently we can only speculate...

brendan-ai2:
According to @matt-gardner, this is fixed in master, so I'm going to go ahead and close this PR. But thank you regardless!

PR #3346, opened 2019-10-11
Pin to pytorch 1.3 in setup.py

matt-gardner:


matt-gardner:
Just merging, as I know that this is the problem and it's a one line fix.

PR #3348, opened 2019-10-11
Removing the reading comprehension code

matt-gardner:
We have created [a separate repository](https://github.com/allenai/allennlp-reading-comprehension) to house the reading comprehension data and model code.  This PR removes the code from the main repository, in favor of the new repo.


schmmd:
@matt-gardner how do you propose that we update the usage documentation for the reading comprehension model once it's removed from this repository?

https://demo.allennlp.org/reading-comprehension

schmmd:
Sounds greart.

dirkgr:
I tried to update this branch to make it mergeable, but it's not that easy. I'll put this on the back burner right now, in favor of getting BERT QA merged first.

matt-gardner:
I'll take care of this.

dirkgr:
This has a lot of test failures that I am working on now.

matt-gardner:
There was just one and I just fixed it.

matt-gardner:
I don't know why these conflicts are showing up again; I fixed them already.

matt-gardner:
I guess those were from the docstring fix.  Another thing that maybe we should port over to the other repo.  But it was a simple one-liner that can be found in Michael's PR.

schmmd:
PR: https://github.com/allenai/allennlp/pull/3588

Sorry to cause conflicts.  If they are too severe it's not the most precious change--although it does grant us some consistency.

matt-gardner:
Ok, I see now a whole bunch of test failures.  These are probably the ones you were talking about @dirkgr, and I haven't worked on them at all.  I need to be done with this for today, so I'll leave it to you, and I won't work on this again unless you give me a go-ahead, so I don't duplicate any work.

dirkgr:
I got these tests to the point where it fails because Python throws a `SIGABRT`. This might be a problem on my machine, so I wanted TC to run it and see if it happens there too.

dirkgr:
Looks like a problem with my environment. I will check here in the morning, and if I don't see any note to stop me, I'll finish it then.

dirkgr:
I moved the remaining DialogQA stuff to https://github.com/allenai/allennlp-reading-comprehension/pull/11.

dirkgr:
✔️

dirkgr:
@matt-gardner, now that the tests work, care to look it over again?

allennlp/tests/commands/evaluate_test.py
```diff
@@ -66,22 +66,33 @@ def test_evaluate_calculates_average_loss_with_weights(self):
     def test_evaluate_from_args(self):
         kebab_args = [
             "evaluate",
-            str(self.FIXTURES_ROOT / "bidaf" / "serialization" / "model.tar.gz"),
-            str(self.FIXTURES_ROOT / "data" / "squad.json"),
+            str(
+                self.FIXTURES_ROOT / "simple_tagger_with_span_f1" / "serialization" / "model.tar.gz"
```
matt-gardner: Sorry, I should have said something earlier; when picking models to replace things with in tests, we should probably just use a simple classifier.  I think all other models are going to be removed, so replacing a removed model with another model that's going to be removed just means we have more work when we go to remove the tagger.  I don't think you should worry about fixing this in this PR, it's just something to keep in mind for when you (or anyone else) does similar things in the future.

allennlp/tests/interpret/hotflip_test.py
```diff
@@ -30,53 +30,6 @@ def test_hotflip(self):
             attack["original"]
         )  # hotflip replaces words without removing
 
-        # test using SQuAD model (tests different equals method)
```
matt-gardner: Can you add this to a test somewhere in the RC repo?  This is an argument for pulling the interpret code into its own repo, so we can actually do integration tests like this, but there are things on the `Predictor` that are only there for the interpret code, so we can't pull everything out...  That's a decision for later, though.

allennlp/tests/commands/evaluate_test.py
dirkgr: I figured the simple tagger might stay, given its name, so I tried to put it in wherever I could.

allennlp/tests/interpret/hotflip_test.py
dirkgr: https://github.com/allenai/allennlp-reading-comprehension/pull/11/commits/133c74160034779025438f319c46642c6f045053

dirkgr: This kind of thing makes me worry about the whole project of splitting everything into separate repos. It's one of the pathologies of software development teams. Teams don't like the problems with their monorepo, and so they want to split it all up. But splitting it all up creates a mess of dependencies that wants to be managed.
- We have no way of running tests across repos. If I change something in the root `allennlp` library, I want the automated tests for my dependencies to run. Google has a system for this. We don't. We already struggled with this in Scala for S2, and never solved the problem.
- We can't guarantee that one version of `allennlp-reading-comprehension` always runs with a corresponding version of `allennlp`. So now we have to write backwards-compatible code every step of the way, or think hard about version pins for our own code. In practice, many projects will require work in multiple repos, so I now have a compatibility problem within a single project, and I have to carefully manage what gets merged when. This is why you had to have `allennlp-reading-comprehension` depend on a SHA. This is why the tests in that repo don't run right now.
- Circular dependencies, like we have in this case, always mean you want to split things up even more, making the problem worse. When you work on Hotflip2, do you want to coordinate changes in `allennlp`, `allennlp-reading-comprehension`, and `allennlp-interpret`?
- Back to tests: With lots of little repos, the ideal test is isolated to running in just one of them. It is hard to make a cross-repo test. So where do we stick integration tests? In my experience, unit tests are 90% useless, but integration tests are the ones that find bugs.

matt-gardner: 1. We do have a way of running tests across repos: we run `allennlp-hub` tests on master commits, which @brendan-ai2 set up.  We run unit tests on all commits in a PR, integration tests on merge, in case our unit tests didn't catch something.  And when that happens, we make a new unit test that would have caught the thing that failed during integration tests.
2. Yeah, the sub-repos might sometimes have to pin things.  I think you're seeing issues now, though, because we're making breaking changes on purpose and we are in the middle of making things work.  Once 1.0 is released, the library should be stable and we will not make breaking changes very often, if at all, and this will be much less of an issue.
3. Yeah, the interpret stuff is more coupled with tasks than I would like it to be, so that particular one is a bit of an issue (and again is an argument for splitting it out, but that could easily make things more circular).  I don't think we will have this problem much in practice, though.  In general, splitting out the repos is also an attempt to force ourselves to think like users, where we have to use allennlp as a third-party library.  This should lead to better code that's easier for users, as we'll try to solve pain points when we run into them.
4. See answer to 1.

brendan-ai2:
This change broke the build for allennlp-hub. See http://build.allennlp.org/viewLog.html?buildId=23873&buildTypeId=AllenNLPHub_Master. Could you please take a look?

PR #3349, opened 2019-10-11
Update docs to refer to 3.7 instead of 3.6

schmmd:


allennlp/semparse/domain_languages/domain_language.py
```diff
@@ -37,7 +37,7 @@ def is_generic(type_: Type) -> bool:
         return isinstance(type_, GenericMeta)  # type: ignore
     else:
 
-        from typing import _GenericAlias
+        from typing import _GenericAlias  # type: ignore
```
schmmd: This causes mypy errors if you are using 3.7 so we should include it for users doing development there.

schmmd: This was removed from the repository so it's no longer in this PR.

DeNeutoy: Just FYI I think that mypy error is actually valid (in the sense you actually can't import `_GenericAlias` from typing anymore in 3.7), so it should be fixed in the semparse repo, still.

matt-gardner: I use python 3.7.4 on my machine, and this code runs.  It also doesn't cause a mypy error when I run it, again using python 3.7.4.  If you look above, this already has an explicit check for python 3.7; I added this if/else when I switched to using 3.7 locally.

matt-gardner: It's possible that I don't get an error here because of the `--ignore-missing-imports` flag that @joelgrus had us use with newer versions of mypy.

matt-gardner: Ok, I finally got past the other CI errors in allennlp-semparse and made it to mypy, and found that with an updated mypy version, I get this mypy error (even though tests pass when using the code).  Fixing it there.

PR #3350, opened 2019-10-11
Add resume daemon for Beaker experiments.

brendan-ai2:
- Compatible with basic experiments created with `scripts/ai2_internal/run_with_beaker.py`.
- Intended to be run locally, e.g. via cron, to ensure experiments are resumed.


brendan-ai2:
Hi @viveklak, I've written a simple tool to let AllenNLP users automatically resume preempted experiments created with our `run_with_beaker.py` script? Would you mind reviewing? I'd like to make sure I'm not doing anything gratuitously wrong with how I use the command line tools. Thanks!

scripts/ai2_internal/resume_daemon.py
```diff
@@ -0,0 +1,219 @@
...
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+logger.setLevel(logging.DEBUG)
+handler = RotatingFileHandler(
+        f"{os.environ['HOME']}/.allennlp/resume.log",
+        maxBytes=1024*1024,
+        backupCount=10)
+logger.addHandler(handler)
+
+BEAKER_QUERY_INTERVAL_SECONDS = 1
```
viveklak: Unused?

scripts/ai2_internal/resume_daemon.py
```diff
@@ -0,0 +1,219 @@
...
+        #        "archived": false
+        #    }
+        #]
+
+        command = ["beaker", "experiment", "inspect", experiment_id]
+        experiment_json = subprocess.check_output(command)
+        experiment_data = json.loads(experiment_json)
+        # TODO(brendanr): Are these constraints ever actually violated?
+        assert len(experiment_data) == 1
+        assert len(experiment_data[0]["nodes"]) == 1
```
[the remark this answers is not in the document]
brendan-ai2: Thanks, that makes sense. I've updated the comment appropriately and documented above that this script is intended to be used with run_with_beaker.py specifically.

scripts/ai2_internal/resume_daemon.py
brendan-ai2: Thanks for catching this. I've added some sleeps to smooth the load to beaker. I don't we'll ever have the number of users for this to really matter, but just as a matter of good practice I've added this in.

brendan-ai2:
Glad to hear that automatic resume will be baked in! That'll definitely be a better solution long term. I've made a few fixes per your comments and fleshed out the install feature.

viveklak:
LGTM

brendan-ai2:
Thanks for the review!

Issue #2793, opened 2019-05-02
flatten_parameters automatically for multi-GPU RNNs


vin-ivar:
I'm not entirely sure whether this is intentional, but perhaps adding a `flatten_parameters` call to the underlying PyTorch RNN might be handy? Right now, multi-GPU RNNs throw warnings, and I'm not sure what the impact on performance is - #2294 references this, the author solved it by manually calling it.

```RuntimeWarning: RNN module weights are not part of single contiguous chunk of memory. This means they need to be compacted at every call, possibly greatly increasing memory usage. To compact weights again call flatten_parameters().```

joelgrus:
we hope to get to this soon, but it's not an explicit priority right now. PR welcome.

PR #3357, opened 2019-10-15
Add workarounds to avoid _flat_weights issues for DropConnect.

tianjianjiang:
I hesitate to open a pull request for it because it involves workarounds. However, since #3092 has been merged, I figure it might be better to submit early and submit often as long as it doesn't break tests.

I'm working on using AWD-BiLSTM-CRFs for sequence labeling. Two issues require workarounds for my implementation:
1. PyTorch raises `AttributeError`: '`LSTM`' object has no attribute '`_flat_weights`'.
2. If duplicated parameters are deleted, `allennlp fine-tune` won't work because it will try to load parameters based on the default model config and the deleted parameters won't be in the `state_dict` anymore.

This PR is for the first issue. The second issue may need further discussions and I can submit another PR for review, with a simple model based on the tutorial of POS tagging.

One of the diffs in this PR was included in #3092 until e439c64. I'm not so sure why it happened. Also, I suspect the test is running on CPU such that it won't encounter `AttributeError` of `_flat_weights` because it will be a no-op (See [torch/nn/modules/rnn.py#L102](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/rnn.py#L102)), and [salesforce's test may be using cuda()](https://github.com/salesforce/awd-lstm-lm/blob/master/weight_drop.py#L85) so it can cover this situation. To my best knowledge, it seems only `ModelTestCase` supports testing on GPU for the time being. Therefore it may require a complementary PR of a demo model to ensure that `DropConnect` works properly. 

Another part of the diffs is a workaround for #2793.

Please advise. Thank you!

----
I gather related discussions about fastai implementation here: https://forums.fast.ai/t/multilingual-ulmfit/28117/142

### DIFF
--- a/allennlp/modules/drop_connect.py
+++ b/allennlp/modules/drop_connect.py
@@ -103,7 +103,7 @@ class DropConnect(torch.nn.Module):
         # This means they need to be compacted at every call, possibly greatly increasing memory
         # usage. To compact weights again call `flatten_parameters()`.
         with warnings.catch_warnings():
-            warnings.simplefilter('ignore')
+            warnings.simplefilter("ignore")
             # Call the top-level module on the inputs.
             output = self._module.forward(*args)
         # Lastly, remove the dropped out parameters from their parent module's _parameter
--- a/allennlp/modules/drop_connect.py
+++ b/allennlp/modules/drop_connect.py
@@ -1,5 +1,6 @@
 from typing import Iterator
 import re
+import warnings
 
 import torch
 import torch.nn.functional as F
@@ -64,6 +65,7 @@ class DropConnect(torch.nn.Module):
         for match in self._matches:
             self.register_parameter(match.raw_parameter_name, match.parameter)
             delattr(match.module, match.parameter_name)
+            self._prepare_matched_parameters(match)
 
     def _search_parameters(self, regex: str) -> Iterator[_Match]:
         """
@@ -75,6 +77,18 @@ class DropConnect(torch.nn.Module):
                 if re.search(regex, parameter_name):
                     yield _Match(submodule_name, submodule, parameter_name, parameter)
 
+    def _prepare_matched_parameters(self, match: _Match):
+        """
+        Apply non-training dropout to all of matched parameters otherwise PyTorch will raise
+        `AttributeError`: '`LSTM`' object has no attribute '`_flat_weights`'.
+        The `forward()` method, however, will still warn against fragmented memory usages of
+        non-flatten parameters. A corresponding workaround is in `forward()`.
+        """
+        raw_parameter = getattr(self, match.raw_parameter_name)
+        match.module._parameters[match.parameter_name] = F.dropout(
+            raw_parameter, p=self._dropout, training=False
+        )
+
     def forward(self, *args):
         # Apply dropout to all of the matching parameters, and force them into their original
         # parent module's _parameter dictionary
@@ -83,8 +97,15 @@ class DropConnect(torch.nn.Module):
             match.module._parameters[match.parameter_name] = F.dropout(
                 raw_parameter, p=self._dropout, training=self.training
             )
-        # Call the top-level module on the inputs.
-        output = self._module(*args)
+        # Avoid the warning of
+        # `/pytorch/aten/src/ATen/native/cudnn/RNN.cpp`:1238:
+        # `UserWarning`: RNN module weights are not part of single contiguous chunk of memory.
+        # This means they need to be compacted at every call, possibly greatly increasing memory
+        # usage. To compact weights again call `flatten_parameters()`.
+        with warnings.catch_warnings():
+            warnings.simplefilter('ignore')
+            # Call the top-level module on the inputs.
+            output = self._module.forward(*args)
         # Lastly, remove the dropped out parameters from their parent module's _parameter
         # dictionary. If not PyTorch might raise errors due to using non-leaf tensors as
         # parameters.
@@ -101,3 +122,4 @@ class DropConnect(torch.nn.Module):
         for match in self._matches:
             self.register_parameter(match.raw_parameter_name, match.parameter)
             delattr(match.module, match.parameter_name)
+            self._prepare_matched_parameters(match)
--- a/allennlp/modules/drop_connect.py
+++ b/allennlp/modules/drop_connect.py
@@ -64,13 +64,8 @@ class DropConnect(torch.nn.Module):
         # updated twice when calling Optimizer.step().
         for match in self._matches:
             self.register_parameter(match.raw_parameter_name, match.parameter)
-            delattr(match.module, match.parameter_name)
-        # Skip `RNNBase._apply()`'s last step otherwise it will run into mismatched `_all_weights`.`
-        # Use `_called_no_op_f_called_no_op_flatten_parameters` as a dummy flag for test coverage.
-        self._called_no_op_flatten_parameters = None
-        if issubclass(type(self._module), torch.nn.RNNBase):
-            self._module.flatten_parameters = self._no_op_flatten_parameters
-            self._called_no_op_flatten_parameters = 0
+        # Don't remove the dropped out parameters from their parent module's _parameter dictionary.
+        # Otherwise a model's `state_dict` won't be reusable for fine-tuning.
 
     def _search_parameters(self, regex: str) -> Iterator[_Match]:
         """
@@ -81,15 +76,6 @@ class DropConnect(torch.nn.Module):
             for parameter_name, parameter in submodule.named_parameters(recurse=False):
                 if re.search(regex, parameter_name):
                     yield _Match(submodule_name, submodule, parameter_name, parameter)
-
-    def _no_op_flatten_parameters(self):
-        """
-        We need to replace flatten_parameters with a no-op function.
-        It must be a function rather than a lambda as otherwise pickling explodes.
-        """
-        if self._called_no_op_flatten_parameters is not None:
-            self._called_no_op_flatten_parameters += 1
-        return
 
     def forward(self, *args):
         # Apply dropout to all of the matching parameters, and force them into their original
--- a/allennlp/modules/drop_connect.py
+++ b/allennlp/modules/drop_connect.py
@@ -65,9 +65,12 @@ class DropConnect(torch.nn.Module):
         for match in self._matches:
             self.register_parameter(match.raw_parameter_name, match.parameter)
             delattr(match.module, match.parameter_name)
-        # Skip `RNNBase._apply()`'s last step otherwise it will run into mismatched `_all_weights`.
+        # Skip `RNNBase._apply()`'s last step otherwise it will run into mismatched `_all_weights`.`
+        # Use `_called_no_op_f_called_no_op_flatten_parameters` as a dummy flag for test coverage.
+        self._called_no_op_flatten_parameters = None
         if issubclass(type(self._module), torch.nn.RNNBase):
             self._module.flatten_parameters = self._no_op_flatten_parameters
+            self._called_no_op_flatten_parameters = 0
 
     def _search_parameters(self, regex: str) -> Iterator[_Match]:
         """
@@ -84,6 +87,7 @@ class DropConnect(torch.nn.Module):
         We need to replace flatten_parameters with a no-op function.
         It must be a function rather than a lambda as otherwise pickling explodes.
         """
+        self._called_no_op_flatten_parameters += 1
         return
 
     def forward(self, *args):
--- a/allennlp/modules/drop_connect.py
+++ b/allennlp/modules/drop_connect.py
@@ -65,7 +65,9 @@ class DropConnect(torch.nn.Module):
         for match in self._matches:
             self.register_parameter(match.raw_parameter_name, match.parameter)
             delattr(match.module, match.parameter_name)
-            self._prepare_matched_parameters(match)
+        # Skip `RNNBase._apply()`'s last step otherwise it will run into mismatched `_all_weights`.
+        if issubclass(type(self._module), torch.nn.RNNBase):
+            self._module.flatten_parameters = self._no_op_flatten_parameters
 
     def _search_parameters(self, regex: str) -> Iterator[_Match]:
         """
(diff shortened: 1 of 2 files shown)


matt-gardner:
@rloganiv, are you able to take a look at this?

allennlp/modules/drop_connect.py
```diff
@@ -75,6 +77,18 @@ def _search_parameters(self, regex: str) -> Iterator[_Match]:
...
 
+    def _prepare_matched_parameters(self, match: _Match):
+        """
+        Apply non-training dropout to all of matched parameters otherwise PyTorch will raise
+        `AttributeError`: '`LSTM`' object has no attribute '`_flat_weights`'.
+        The `forward()` method, however, will still warn against fragmented memory usages of
+        non-flatten parameters. A corresponding workaround is in `forward()`.
+        """
+        raw_parameter = getattr(self, match.raw_parameter_name)
+        match.module._parameters[match.parameter_name] = F.dropout(
```
rloganiv: Why apply dropout with training=False instead of just assigning the parameter?

allennlp/modules/drop_connect.py
```diff
@@ -83,8 +97,15 @@ def forward(self, *args):
...
-        output = self._module(*args)
+        # Avoid the warning of
+        # `/pytorch/aten/src/ATen/native/cudnn/RNN.cpp`:1238:
+        # `UserWarning`: RNN module weights are not part of single contiguous chunk of memory.
+        # This means they need to be compacted at every call, possibly greatly increasing memory
+        # usage. To compact weights again call `flatten_parameters()`.
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            # Call the top-level module on the inputs.
+            output = self._module.forward(*args)
```
rloganiv: Why use .forward() instead of calling the module? Does hook processing cause an issue.

allennlp/tests/modules/drop_connect_test.py
```diff
@@ -47,6 +48,22 @@ def test_lstm_outputs(self):
         output_b, _ = dropped_lstm(input_tensor)
         assert torch.allclose(output_a, output_b)
 
+    @unittest.skipIf(not torch.cuda.is_available(), reason="No CUDA device registered.")
+    @flaky(max_runs=10, min_passes=1)
```
rloganiv: Why run this test multiple times?

rloganiv:
@tianjianjiang thanks for this PR. I agree both of the issues you raised above need to be addressed.

> One of the diffs in this PR was included in #3092 until e439c64. I'm not so sure why it happened.

The motivation for deleting the dropped out parameters was to avoid double counting. In the code you've posted `optimizer.step()` will update the parameters twice (1x for the _raw weights in DropConnect and 1x for the module weights).

[SalesForce's implementation also includes code which replaces the module's flatten_parameters method with a no-op](https://github.com/salesforce/awd-lstm-lm/blob/32fcb42562aeb5c7e6c9dec3f2a3baaaf68a5cb5/weight_drop.py#L21-L24). Would this solve the `_flat_weights` issue without requiring that the dropped out weights be saved in the parameter dict?

allennlp/modules/drop_connect.py
tianjianjiang: As https://github.com/fastai/fastai/commit/b842586e9b080ed83afb251d4236ec6843d823de described, it is probably a workaround for and only for QRNN with discriminative fine-tuning. That was the main reason why I was reluctant to submit this PR. Another related concern is that cloning the weight by the identity function of dropout somewhat contradicts the intention of `delattr()`. By the way, while `delattr()` seems to destroy the `all_weights`, the original Salesforce version uses a less convenient yet seemingly more surgical way of `del self.module._parameters[name_w]`.

allennlp/modules/drop_connect.py
tianjianjiang: May be not. I haven't really made up my mind on this. On one hand it feels consistent to have `forward()`, but on the other hand it is a composition rather than an inheritance, so it's confusing for me whether leaving all extra steps of the hook for the outermost module or making it happen here.

allennlp/tests/modules/drop_connect_test.py
tianjianjiang: My bad. I tried to parametrize the original `test_lstm_outputs()` and then realized that it probably won't work well with `@flaky` and won't make sense anyway.
Will remove the `@flaky`.

tianjianjiang:
> @tianjianjiang thanks for this PR. I agree both of the issues you raised above need to be addressed.
> 
> > One of the diffs in this PR was included in #3092 until [e439c64](https://github.com/allenai/allennlp/commit/e439c649e957de7b5e501efedcca1d0f1ac26ba3). I'm not so sure why it happened.
> 
> The motivation for deleting the dropped out parameters was to avoid double counting. In the code you've posted `optimizer.step()` will update the parameters twice (1x for the _raw weights in DropConnect and 1x for the module weights).
> 
> [SalesForce's implementation also includes code which replaces the module's flatten_parameters method with a no-op](https://github.com/salesforce/awd-lstm-lm/blob/32fcb42562aeb5c7e6c9dec3f2a3baaaf68a5cb5/weight_drop.py#L21-L24). Would this solve the `_flat_weights` issue without requiring that the dropped out weights be saved in the parameter dict?

@rloganiv Thank you for reviewing the PR.
I understand that it will probably double-count parameters if not deleting the source, just not sure whether it is incorrect or just slower. To my limited understanding, Salesforce version uses no-op `flatten_parameters` to bypass the `_flat_weights` error and warning, yet since it only deletes the source once in `__init__`, the `state_dict` will still have the source parameter name right after `forward`. What I haven't figured out is the trade-offs of `del` vs `delattr()` and `setattr()` vs. `match.module._parameters[match.parameter_name] = ...`. They have different assumptions and side-effects that may concern AllenNLP's conventions.

rloganiv:
> I understand that it will probably double-count parameters if not deleting the source, just not sure whether it is incorrect or just slower.

Double-counting parameters results in incorrect behavior. Essentially, `optimizer.step` will be called 2x effectively doubling the learning rate for the dropped out parameters.

In addition, storing the dropped out weights as parameters has historically caused issues. PyTorch has routinely changed whether or not `F.dropout(..., training=False)` outputs a leaf tensor. In the versions where the dropped out tensors are not leaf nodes, PyTorch will raise an error when trying to initialize an `Optimizer` using `WeightDrop.parameters()`. This is not an issue for the current version, but may become one again in the future.

> What I haven't figured out is the trade-offs of del vs delattr() and setattr() vs. match.module._parameters[match.parameter_name] = ....

I'll need to look into the point you've raised above about `del` vs `delattr()`. I can provide some insight into the difference between `setattr()` and directly assigning to `module._parameters`. In short, `Module.setattr()` calls `Module.register_parameter()` (see [here](https://github.com/pytorch/pytorch/blob/d04973beda96bd83b6a5ba2f8c81a3576e4ddcfb/torch/nn/modules/module.py#L594-L599)), which will [raise an error when the argument is not a leaf tensor](https://github.com/pytorch/pytorch/blob/d04973beda96bd83b6a5ba2f8c81a3576e4ddcfb/torch/nn/modules/module.py#L167-L172). Since applying dropout results in a non-leaf tensor, you need to assign `Module._parameters` manually in order to avoid raising the error.

tianjianjiang:
> Double-counting parameters results in incorrect behavior. Essentially, `optimizer.step` will be called 2x effectively doubling the learning rate for the dropped out parameters.

Please correct me if I'm misinformed. Wouldn't the optimize only initializes and obtains the parameter groups once? Perhaps that was the reason why Salesforce version deletes hidden weight parameters once in `__init__()` but not in `forward()`. Unless optimizer will retain the parameter groups after every `forward()`.

> In addition, storing the dropped out weights as parameters has historically caused issues. PyTorch has routinely changed whether or not `F.dropout(..., training=False)` outputs a leaf tensor. In the versions where the dropped out tensors are not leaf nodes, PyTorch will raise an error when trying to initialize an `Optimizer` using `WeightDrop.parameters()`. This is not an issue for the current version, but may become one again in the future.

I see, thank you for the information. It seems to circle back to my previous question.

> I'll need to look into the point you've raised above about `del` vs `delattr()`. I can provide some insight into the difference between `setattr()` and directly assigning to `module._parameters`. In short, `Module.setattr()` calls `Module.register_parameter()` (see [here](https://github.com/pytorch/pytorch/blob/d04973beda96bd83b6a5ba2f8c81a3576e4ddcfb/torch/nn/modules/module.py#L594-L599)), which will [raise an error when the argument is not a leaf tensor](https://github.com/pytorch/pytorch/blob/d04973beda96bd83b6a5ba2f8c81a3576e4ddcfb/torch/nn/modules/module.py#L167-L172). Since applying dropout results in a non-leaf tensor, you need to assign `Module._parameters` manually in order to avoid raising the error.

Thank you for the explanation. I've checked and it seems the differences are unrelated. My apologies. Although the parameter deletion renders `_all_weights` invalid and then causes issues of `_flat_weights`, Salesforce's workaround should fix it.

rloganiv:
> Thank you for the explanation. I've checked and it seems the differences are unrelated. My apologies. Although the parameter deletion renders _all_weights invalid and then causes issues of _flat_weights, Salesforce's workaround should fix it.

:+1:

So is there any need to keep `prepare_matched_parameters()`?

tianjianjiang:
> So is there any need to keep `prepare_matched_parameters()`?

No. I was trying to intervene `_all_weights` and `_apply()` and even swapping weights in it, but it turned out just not as convenient as the no-op `flatten_parameters`.)
So the latest version now deletes the weights only once in `__init__()` and revises the leaf tensor test case to cover optimizer situation.

Please advise, thank you!

rloganiv:
> Not sure how to deal with the "codecov/patch — 80% of diff hit (target 90%)" failure though.

My best guess is that `_no_op_flatten_parameters` needs to be tested  (as ridiculous as that seems :stuck_out_tongue:).

tianjianjiang:
> > Not sure how to deal with the "codecov/patch — 80% of diff hit (target 90%)" failure though.
> 
> My best guess is that `_no_op_flatten_parameters` needs to be tested (as ridiculous as that seems 😛).

I thought so, but now it seems more about the unused `reset()` since it's a fastai convention, not an AllenNLP one. This situation also reveals that the test case of `eval()` should fail. It isn't failing now only because it follows the test case of `reset()`. I suspect that `eval()`'s `forward()` may require another integration test that resembles the case of `train()` with an optimizer.

matt-gardner:
The test coverage target can be ignored if the lines of code that aren't being tested would be a huge pain to test and are likely to be correct.  We aren't always strict about that check, if there are at least some tests.  Do you both think that this is ready to merge?  @rloganiv, I'll trust your judgment.

tianjianjiang:
Please give me one more day or so to figure out whether the `eval()` mode is correct or not. Thank you both for your patience.

tianjianjiang:
It seems that `delattr()` in `__init__()` still causes problems of a model's `state_dict`. However, as long as we initialize an optimizer with non-recursive `module.parameters()` by default, the optimizer won't run into any non-leaf tensor. If `module.parameters()` always uses `recursive=False`, we wont' even need the trick of no-op `flatten_parameters()`.

tianjianjiang:
> It seems fine by default.

1. After the weight-dropped layer and the whole model initialized, someone creates an optimizer with parameter value IDs from:
    1. `.parameters()` (or `.named_parameters()`) plus **some preprocessing** such as filtering by `.requires_grad` for frozen layers;
    2. `._named_members()` subsequently. It returns **unique value IDs**. Either the top module's parameter takes precedence, or a random one pops up if duplicated parameters are at the same module (but probably no one will register it at the same module anyway).
2. During the first iteration:
    1. Before `.forward()`, parameters inside the optimizer should all have `is_leaf=True` otherwise the optimizer will complain. Luckily [1.2] takes care of it.
    2. After `.forward()`, the original parameter will have a different value of weight-dropped `Tensor` (without a wrapper of type `Parameter`), so **anyone uses `.parameters()` will have an additional parameter from now on**.
    3. `.backward()` accumulates `grad`'s to the `_raw` one, but not the original one that has `.is_leaf=False`.
    4. `.step()` updates the `_raw` one's weight.
3. If somebody runs `.eval()` and `.forward()` for validation, it will remove the weight-dropped `Tensor` from `.parameters()`, because `F.dropout(training=False)` makes the duplicated parameters share the same value again.

matt-gardner:
Sorry to have not responded on this for a while; @tianjianjiang, what's the state of this?

tianjianjiang:
@matt-gardner My apologies too for not being assertive in my latest reply. It is pending for code review.

matt-gardner:
@rloganiv, when you get a chance, it'd be great to hear whether you think this is ready to merge.

allennlp/tests/modules/drop_connect_test.py
```diff
@@ -90,14 +121,38 @@ def test_parameters_are_leaf_tensors(self):
...
         weight_dropped_linear.eval()
-        assert all(parameter.is_leaf for parameter in weight_dropped_linear.parameters())
+        # The duplicated-non-leaf weight still exists.
+        pre_eval_parameters = list(weight_dropped_linear.parameters())
+        assert len(pre_eval_parameters) == _n_params + 1
+        assert not all(parameter.is_leaf for parameter in pre_eval_parameters)
+        weight_dropped_linear(input_tensor)
+        # But only the raw weight applies.
+        eval_parameters = list(weight_dropped_linear.parameters())
+        assert len(eval_parameters) == _n_params
```
rloganiv: This snippet illustrates my key concern - the length of the parameter list changes depending on whether the module is in eval mode, and whether or not the forward method has been invoked. Presumably if the model state was saved after one of these steps we could run into another issue loading the state dict due to unexpected/missing parameters.

Note: I am not 100% sure whether this issue would occur in practice. My point is that parameters are usually fixed once a module is instantiated, however that does not appear to be the case here, which is difficult to reason about.

tianjianjiang: Hi @rloganiv and @matt-gardner,
Happy new year!
Yes, I'm aware of the side-effect. In my opinion, because the current revision before this PR has another issue of parameter list and model state, so `DropConnect` implementation here hasn't been a pure function, and the trading one side-effect with another seems an acceptable deal.
That being said, I also really want `DropConnect` to be a pure function. Although admittedly I haven't figured out what is an ideal implementation, [the approach of torchnlp](https://github.com/PetrochukM/PyTorch-NLP/blob/master/torchnlp/nn/weight_drop.py) is intriguing. If you guys have other ideal designs in mind, please let me know. I can try to spend some more time on making `DropConnect` pure.

rloganiv:
Happy new year to you as well!

I think I've finally come up with a simple solution. The problem with the current code is that when the module is in training mode and forward is called, the original module's weights are replaced by the dropped out weights, effectively adding a a new (erroneous) parameter to the module (note: erroneous since parameters need to be leafs). At all other points in time (e.g., after initialization, and after the module is called in eval mode) the behavior of the module is correct (all parameters are leaf nodes). So what if we just change forward so that the module's state is always the same after it is called? E.g., after computing `output` we add:
```{python}
        for match in self._matches:
            raw_parameter = getattr(self, match.raw_parameter_name)
            match.module._parameters[match.parameter_name] = F.dropout(
                raw_parameter, p=self._dropout, training=False
            )
```

DeNeutoy:
@tianjianjiang Thanks for the contribution, but we've actually decided to remove this code, as it's a bit tricky to maintain properly. Sorry about that! Hopefully @rloganiv's ideas have helped you fix the issue in your own codebase.

tianjianjiang:
@DeNeutoy I see. Too bad there's another issue with PyTorch 1.4.
In my code base for the time being, I think @rloganiv 's idea will work, because I had a similar step before without problems.
Thank you guys for going through all these troubles. I sincerely hope it will be better next time. :)

PR #3358, opened 2019-10-15
Allow passing a Path in ElmoEmbedder

konstin:
`ElmoEmbedder` did accept `Path`s instead of strings, but the typings didn't acknowledge that.


konstin:
The problem is that I'm using mypy in a downstream project and get errors when trying to pass a `Path` when the type annotation says `str`, even though the underlying `cached_path` works fine with `Path`.

Issue #3345, opened 2019-10-10
Remove WordSplitter


matt-gardner:
Another candidate idea for simplifying API stuff for a 1.0 release: remove the whole notion of `WordSplitter`, and just call everything `Tokenizers`.  Basically no one uses the extra stuff that the WordTokenizer has, and it just adds a level of indirection that's unnecessary.  If you want the extra filtering and stemming, implement it in a standalone `Tokenizer`.

sai-prasanna:
Can I do the following changes?and remove the code for stemmer and lemattizer (which we don't really use anywhere) ?

1.  Change the existing wordsplitters to individual tokenizers. SpacyWordSplitter -> SpacyTokenizer , BertWordSplitter -> BertBasicTokenizer etc.
2. Remove word_stemmer and word_filter code completely. Maybe other users rely on this, but breaking change in 1.0.0 is permitted?
3. Make SpacyTokenizer the default tokenizer instead of the WordTokenizer.

On a side note can we can get replace the  default spacy tokenizer with something that does only tokenization faster (like https://github.com/microsoft/BlingFire). But we do need spacy for pos tagging for POS indexer.

matt-gardner:
Yes, you can do those.  This will break sniff tests, though, as it will change config file requirements.  A few thoughts:

1. We probably don't need the OpenAI or BERT tokenizers as separate objects, because we now have the pretrained transformer tokenizer that should let you get the same functionality.
2. We can probably put some special logic in a custom `Tokenizer.from_params` method, that checks for `"type": "word"`, pulls out the word splitter, and redirects.  Does this make sense?  I can give more detail if you need it.  This should handle most of the backwards compatibility issues with config files, and it _should_ be possible to make this pass our sniff tests without modifying any configs.
3. It's fine to remove the stemmer and filter stuff.  If someone wants that functionality, they can open a PR to put it back in a new `Tokenizer`.  I really doubt many people actually want it.
4. No strong opinion on the default tokenizer.  I'd guess that most people these days are using pretrained transformers for most things, so they should be using a tokenizer that matches.

matt-gardner:
Well, spacy should probably stay the default, so we don't break a whole lot of existing saved models.

sai-prasanna:
1. We are using "bert-basic" splitter in the "bert_for_classification" and "bert_pooler" examples. 
[...]

PR #3361, opened 2019-10-15
Replace existing WordSplitter with Tokenizers

sai-prasanna:
Removes the notion of word splitter and replaces it with tokenizer.

Fixes https://github.com/allenai/allennlp/issues/3345


allennlp/data/dataset_readers/copynet_seq2seq.py
```diff
@@ -10,7 +10,7 @@
 from allennlp.data.dataset_readers.dataset_reader import DatasetReader
 from allennlp.data.fields import TextField, ArrayField, MetadataField, NamespaceSwappingField
 from allennlp.data.instance import Instance
-from allennlp.data.tokenizers import Token, Tokenizer, WordTokenizer
+from allennlp.data.tokenizers import Token, Tokenizer, SpacyWordTokenizer
```
matt-gardner: Can we call this `SpacyTokenizer` instead?

allennlp/data/dataset_readers/masked_language_modeling.py
```diff
@@ -5,8 +5,8 @@
 
 from allennlp.data.instance import Instance
 from allennlp.data.tokenizers.tokenizer import Tokenizer
-from allennlp.data.tokenizers import Token, WordTokenizer
-from allennlp.data.tokenizers.word_splitter import JustSpacesWordSplitter
+from allennlp.data.tokenizers import Token
+from allennlp.data.tokenizers.word_tokenizer import JustSpacesTokenizer
```
matt-gardner: Can we call this `WhitespaceTokenizer` instead?  And it looks like these tokenizers should be moved to separate files, instead of keeping them all in `word_tokenizer`.

allennlp/data/tokenizers/word_tokenizer.py
```diff
@@ -1,19 +1,27 @@
...
+from allennlp.common.util import get_spacy_model
 from allennlp.data.tokenizers.token import Token
 from allennlp.data.tokenizers.tokenizer import Tokenizer
-from allennlp.data.tokenizers.word_filter import WordFilter, PassThroughWordFilter
-from allennlp.data.tokenizers.word_splitter import WordSplitter, SpacyWordSplitter
-from allennlp.data.tokenizers.word_stemmer import WordStemmer, PassThroughWordStemmer
+from allennlp.data.token_indexers.openai_transformer_byte_pair_indexer import text_standardize
 
 
 @Tokenizer.register("word")
```
matt-gardner: This should be registered as `spacy`.

allennlp/data/tokenizers/word_tokenizer.py
```diff
@@ -38,42 +55,272 @@ class WordTokenizer(Tokenizer):
...
+
+    @overrides
+    def tokenize(self, text: str) -> List[Token]:
+        # This works because our Token class matches spacy's.
+        return self._sanitize(_remove_spaces(self.spacy(text)))
+
+    @classmethod
+    def from_params(cls, params: Params) -> "SpacyWordTokenizer":  # type: ignore
+
+        # Backwards compatibility for legacy "word" Tokenizer
```
matt-gardner: I think this should instead go on `Tokenizer.from_params`.  Check for the `type` key, and if it's `word`, pull out the `word_splitter` and then do `Tokenizer.by_name(word_splitter).from_params(params)`.  Should be simple.

allennlp/data/tokenizers/sentence_splitter.py
```diff
@@ -12,7 +12,7 @@ class SentenceSplitter(Registrable):
     A ``SentenceSplitter`` splits strings into sentences.
     """
 
-    default_implementation = "spacy"
+    default_implementation = "word"
```
matt-gardner: Did you mean to change this?

allennlp/data/tokenizers/word_tokenizer.py
```diff
@@ -38,42 +55,272 @@ class WordTokenizer(Tokenizer):
...
+            ner=ner,
+            keep_spacy_tokens=keep_spacy_tokens,
+            split_on_spaces=split_on_spaces,
+            start_tokens=start_tokens,
+            end_tokens=end_tokens,
+        )
+
+
+@Tokenizer.register("simple")
+class SimpleSpacyWordTokenizer(Tokenizer):
```
matt-gardner: Should just be `SimpleTokenizer`.  Probably also ok to just remove it.  You should use spacy or bling instead.

allennlp/data/tokenizers/word_tokenizer.py
```diff
@@ -38,42 +55,272 @@ class WordTokenizer(Tokenizer):
...
+                tokens.append(Token(field))
+            tokens.extend(add_at_end)
+        return tokens
+
+    def _can_split(self, token: str):
+        return token and token.lower() not in self.special_cases
+
+
+@Tokenizer.register("letters_digits")
+class LettersDigitsSpacyWordTokenizer(Tokenizer):
```
matt-gardner: `LettersDigitsTokenizer`.  And in its own file.

allennlp/data/tokenizers/word_tokenizer.py
```diff
@@ -1,19 +1,27 @@
-from typing import List
```
matt-gardner: There shouldn't be a `word_tokenizer.py` file.  Just a `tokenizer.py`, and one file per tokenizer.  (If they are really simple, like the LettersDigitsTokenizer, they could possibly just be put in `tokenizer.py` under the initial class definition.  But then they are hard to find, so even if they're really small, it's still probably best to put them in their own file.)

allennlp/data/tokenizers/word_tokenizer.py
```diff
@@ -38,42 +55,272 @@ class WordTokenizer(Tokenizer):
...
+    tokens does not matter.  This will never result in spaces being included as tokens.
+    """
+
+    @overrides
+    def tokenize(self, text: str) -> List[Token]:
+        return [Token(t) for t in text.split()]
+
+
+@Tokenizer.register("openai")
+class OpenAISpacyWordTokenizer(Tokenizer):
```
matt-gardner: `OpenAiTokenizer`.  We probably will want to just remove this later, in favor of the transformer tokenizer, but we can keep it for now, I suppose.

allennlp/data/tokenizers/word_tokenizer.py
```diff
@@ -38,42 +55,272 @@ class WordTokenizer(Tokenizer):
...
+    @overrides
+    def tokenize(self, text: str) -> List[Token]:
+        return [Token(text) for text in self.basic_tokenizer.tokenize(text)]
+
+
+def _remove_spaces(tokens: List[spacy.tokens.Token]) -> List[spacy.tokens.Token]:
+    return [token for token in tokens if not token.is_space]
+
+
+class _WhitespaceSpacyTokenizer:
```
matt-gardner: This should end up in the `spacy_tokenizer.py` file.

allennlp/data/tokenizers/word_tokenizer.py
sai-prasanna: There is some more to handle various ways "type" can be given. It could be let out or the params themselves can be a string instead of json etc. 

I have handled all the cases, only problem would be if anyone had used non spacy tokenizer with start_tokens and end_tokens. (The only one using this option in our tests is a elmo one in sniffer tests). Should I add "start_tokens" and "end_tokens" to all the existing Tokenizers ? (might be unneeded)

allennlp/data/tokenizers/white_space_tokenizer.py
```diff
@@ -0,0 +1,23 @@
+from typing import List
+
+from overrides import overrides
+
+from allennlp.data.tokenizers.token import Token
+from allennlp.data.tokenizers.tokenizer import Tokenizer
+
+
+@Tokenizer.register("just_spaces")
```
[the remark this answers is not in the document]
sai-prasanna: Thanks for the quick review. 

https://github.com/allenai/allennlp/blob/6373bbba690cf8a33f0320ae9a1e43259317d12e/allennlp/tests/models/sniff_test.py#L247

This pretrained model has ` {'end_tokens': ['@@NULL@@']}` in its config. So I added start and end to `SpacyTokenizer` alone.

matt-gardner:
Thanks again!

PR #3364, opened 2019-10-15
Remove ignored do_lowercase argument to transformer tokenizers

### LINKED ISSUE #3360: do_lowercase should be smart about which model is being used
In the `pretrained_transformer` tokenizer, we take two required arguments: a model name and `do_lowercase`. https://github.com/allenai/allennlp/blob/6d5bcaf11e12938c4d6da1f04c05bf17bf2f2c92/allennlp/data/tokenizers/pretrained_transformer_tokenizer.py#L40-L53

The thing is, the model name already uniquely specifies `do_lowercase`, so the extra parameter is redundant.  We should keep a list of valid model names and infer the right value based on the model name, possibly leaving the parameter as an option in case a new model comes out that isn't in our list.

hawkeoni:
Hello, @matt-gardner!
There is no easy way to import all model names from pytorch transformers as dictionaries with names and links are named *MODELNAME*_PRETRAINED_CONFIG_ARCHIVE_MAP, where MODELNAME is a constantly growing list. That's the reason why taking modelnames from https://huggingface.co/transformers/pretrained_models.html is not a good idea - someone will have to update the list each time a new model is released. Dictionary of cased/uncased models https://github.com/huggingface/transformers/blob/be916cb3fb4579e278ceeaec11a6524662797d7f/transformers/tokenization_bert.py#L70-L86 is only available for bert, so I updated a simple heuristic.

I decided to add an automatic inference of cased parameter. Minor difference is that I changed the condition from `endswith` to `in`, because there are models with names like `distilbert-base-uncased-distilled-squad`.

Fixes #3360


allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -37,20 +37,25 @@ class PretrainedTransformerTokenizer(Tokenizer):
...
-            logger.warning(
-                "Your pretrained model appears to be cased, "
-                "but your tokenizer is lowercasing tokens."
-            )
-        elif model_name.endswith("-uncased") and not do_lowercase:
-            logger.warning(
-                "Your pretrained model appears to be uncased, "
-                "but your tokenizer is not lowercasing tokens."
-            )
+        name_do_lowercase = "-uncased" in model_name
```
matt-gardner: I think `guessed_do_lowercase` is a little bit more readable here.  "name" is too ambiguous to be immediately obvious what this variable name means.

Also, can you move this into a separate function?  It'll be more self-contained and easier to improve later that way.  Something like `do_lowercase = _get_casing(model_name, do_lowercase)`.  And in that function, you can have something like `if do_lowercase is None: return guessed_do_lowercase`, to simplify the if/else blocks a bit.

hawkeoni: I'm, currently working on it. I added some tests, because that do_lowercase parameter does not matter at all, since tokenizer ignores it, if it doesn't match model casing.

hawkeoni:
I do realize we want some backward compatibility, but I believe we either have to choose one of two ways:
1) We leave everything as is, possibly remove do_lowercase parameter, because tokenizer ignores it if it is wrong (I added a test just for that).
2) We somehow force do_lowercase on text. It is quite simple, when we want to lowercase in a cased model, but probably impossible to keep case in uncased model.

hawkeoni:
@matt-gardner Can you help me with failed build? Flake8 and tests all seem to pass.

matt-gardner:
Wait, the huggingface code _ignores_ it if it's wrong?  What's the point of the parameter, then?  If it's just ignored, we should just remove it.  We could possibly keep it as optional (but ignored) for backwards compatibility, but we're planning on making some breaking changes for the 1.0 release anyway, and I think we should get this right.

The build is failing because you need to use `black` to format your code.  You should be able to do this with `pip install black; black .`.

matt-gardner:
Thanks for finding this out, by the way!  I just added that parameter because it looked like huggingface code expected it (it was during a hackathon and I didn't slow down to carefully read much of their documentation or try things out).

hawkeoni:
There are a lot of code (like indexers) which use the same `do_lowercase` parameter. I'll remove it, where necessary.

hawkeoni:
I guess we are good to merge, if everything looks fine.

ZhaofengWu:
@hawkeoni Why is this option removed, and where does the Huggingface code ignore this option? It seems to me that without specifying this argument, the Huggingface code simply lowercases everything? (https://github.com/huggingface/transformers/blob/master/src/transformers/tokenization_bert.py#L187, https://github.com/huggingface/transformers/blob/master/src/transformers/tokenization_utils.py#L675)

hawkeoni:
@ZhaofengWu 
There is a special dictionary of `do_lowercase` parameter for a model https://github.com/huggingface/transformers/blob/90b7df444fc30d5f476e5ab32d1f89340998a28d/src/transformers/tokenization_bert.py#L74-L92 which is automatically used in `from_pretrained` constructor.
Also this code snippet can be run to test this.
```python
import transformers
tokenizer_cased = transformers.BertTokenizer.from_pretrained("bert-base-cased")
tokenizer_uncased = transformers.BertTokenizer.from_pretrained("bert-base-uncased")
s = "Hello There!"
print(tokenizer_cased.tokenize(s))
# ['Hello', 'There', '!']
print(tokenizer_uncased.tokenize(s))
#['hello', 'there', '!']
```

Allennlp does not create tokenizers through `__init__`, it uses `from_pretrained`. I believe that is the reason for confusion.

ZhaofengWu:
Thanks for the pointer! That's really quite obscure.

ZhaofengWu:
@hawkeoni Actually, what if the user is using a custom _cased_ model that is not in that dictionary? Say a local file. Then there's currently no way to specify that we want to maintain the casing through AllenNLP after this PR right?

hawkeoni:
@ZhaofengWu 
That is true, though model casing can not be changed through __init__ anymore, it can still be changed thorugh:
```python
allennlp_tokenizer_object.tokenizer.do_lower_case=True
```
Also as you've said adding to the class configuration to AutoTokenizer is another soultion.

The idea behind this was that the user should not specify this parameter, because he can set the wrong value and default implementations had the right parameter themselves anyway.

I believe that the case with custom cased model is not a very common case, as it requires a lot of time and computational power to do so. Most people are happy with just the pretrained models.

But you rise a very interesting question: what if we really do need this custom parameter? But what about all the other custom parameters we never even considered? As you can see there are more parameters than just `do_lower_case`. There are also `never_split`, `tokenize_chinese_chars`, `do_basic_tokenize`, `pad_token` and some others which you cold find depending on the model.

A better solution should be adding a vocabulary with parameters, which would be in turn passed to the `from_pretrained` method through **.

@matt-gardner I believe we need your opinion on this issue. I could implement my proposed solution if you think it is a good one.

matt-gardner:
If there's an easy way to have **kwargs passed to the autotokenizer, seems like an obvious thing to add.  It solves all of these issues, right?

ZhaofengWu:
I don’t think this is a very uncommon issue.
It’s isn’t always user-trained models that are local files. It could be
model weights downloaded elsewhere, e.g. SpanBERT which is what I’m playing
with right now. I think given the rapid development of pertained
transformers, Huggingface of course wouldn’t be able to instantly
internalize all of them, so I think it could actually be quite common. Also
Huggingface has a dedicated custom model hub which I believe is also not
associated with predefined configurations.

And yes, I vote for **kwargs.

hawkeoni:
Great, I'll submit a new PR soon.

Issue #2536, opened 2019-02-22
Using DistributedDataParallel for multi GPU training


scarecrow1123:
I'm looking into the codebase to possibly use torch's `DistributedDataParallel` for multi GPU training. Based on the [docs](https://pytorch.org/docs/stable/nn.html#torch.nn.parallel.DistributedDataParallel), it certainly would improve the training speed when compared with `DataParallel`. I'm only looking into single node - multi GPU case for now. I would like to get a heads up on possible caveats that I could face when integrating it into the current `Trainer` class. I am only worried about goofing up any stateful parts of the code that could result in leaks. Would be great if someone could give pointers to such instances if any.

joelgrus:
we recently refactored the `Trainer` class to make it easier to implement different trainer variants.

so probably my recommendation would be to implement the simplest possible `ParallelTrainer` subclass, see how well it works, and then after that we could decide if/how to incorporate it into the library.

does that make sense?

scarecrow1123:
Thanks for the heads up @joelgrus . I did a preliminary analysis on how something like a `ParallelTrainer` could be achieved. Here are a few observations and questions based on what I understood:

- Essentially, a single worker function, say `ParallelTrainer.train` shall be spawned as separate processes
- Optimizer will run independently and hence the speed up when compared to `DataParallel`
- The main caveat here could be sampling subsets of mini-batches across the workers. In `DataParallel` the mini-batch gets split and sent from a single place. However, in this case the different processes themselves have to sample them and hence things like shuffling and the indices to sample must be deterministic across processes. The pytorch way is to pass `DistributedSampler` to the `DataLoader` which uses the local rank of the GPU process(simply a process ID) to choose the index of a data point. However in AllenNLP's case, I assume that there is no such provision. I'm still thinking of what could be a potential solution for that. Would it be good to introduce a `Sampler` notion in the current data iterators? Or have I missed anything?

scarecrow1123:
So I've been fiddling with the code and hacked around to make `DistributedDataParallel`(DDP) work with a standard AllenNLP setup. The interesting thing is that I'm able to see only small improvements from DDP when compared with the DataParallel(DP) one. The speedup obtained when compared with a single GPU run is good. Here are some numbers that I've got so far for a [DeepSpeech](https://arxiv.org/abs/1412.5567) model that I have implemented.

Dataset: [Librispeech 100 hour](http://www.openslr.org/12/)
GPU: 2080Ti on Ubuntu
[...]

PR #3372, opened 2019-10-17
Logging and metrics changes for distributed training

scarecrow1123:
As discussed at #2536 , this is the first PR to bring in `DistributedDataParallel` support. To start with this contains refactored implementations of `prepare_global_logging` and `training_util.get_metrics` methods.

* Both changes involve adding `rank` and `world_size` as optional attributes.

* Wrt logging, in the distributed case, each worker will log to their respective files with names `stdout_{rank}.log`/`stderr_{rank}.log`. 

* Logs to stdout/stderr are only from the master(rank=0) process. `tqdm` gets updated with metrics aggregated across all workers

* With these changes to logging, `TeeLogger` becomes unnecessary IMO and hence I took the liberty to remove it.

* `get_metrics` does a `dist.all_reduce` to aggregate all the provided metrics

P.S: pytest failures are yet to be fixed


allennlp/common/tee_logger.py
```diff
@@ -1,63 +0,0 @@
...
-        sys.stderr = TeeLogger("stdout.log", sys.stderr)
-    """
-
-    def __init__(
-        self, filename: str, terminal: TextIO, file_friendly_terminal_output: bool
-    ) -> None:
-        self.terminal = terminal
-        self.file_friendly_terminal_output = file_friendly_terminal_output
-        parent_directory = os.path.dirname(filename)
-        os.makedirs(parent_directory, exist_ok=True)
```
DeNeutoy: This call is missing in your replacement function, which is why all the tests are failing. Just add this and then I think that should kill most of the CI output :)

allennlp/training/util.py
```diff
@@ -378,17 +379,31 @@ def rescale_gradients(model: Model, grad_norm: Optional[float] = None) -> Option
...
     """
     Gets the metrics but sets ``"loss"`` to
     the total loss divided by the ``num_batches`` so that
     the ``"loss"`` metric is "average loss per batch".
     """
     metrics = model.get_metrics(reset=reset)
     metrics["loss"] = float(total_loss / num_batches) if num_batches > 0 else 0.0
-    return metrics
+
+    if world_size > 1:
```
DeNeutoy: I think this needs to be before the bit above where we replace the `metrics["loss"]` with a float, as it won't have been all_reduced (and then subsequently won't be all reduced correctly, as we've lost the reference to the tensor)

DeNeutoy:
Also, as you're going to be submitting several PRs, here are a couple of development tips for allennlp which I find useful:

- You can run the CI checks locally using the scripts in the `scripts` directory, e.g `bash scripts/black.sh`, `bash scripts/flake8.sh`, `bash scripts/flake8.sh`. 
- Typically I run only tests I think I might have broken locally, e.g `pytest allennlp/tests/the/one/I/broke` otherwise it takes _super_ long.
- In our CI we have checks for the allennlp docs. In general don't worry about these, as we will always be able to tell you how to fix it, but basically we check that we have docs for all public modules, and that we don't have docs for modules which don't exist. For example, this PR is failing the docs check because you deleted the `TeeLogger`, but the docs are still present here: 

https://github.com/allenai/allennlp/blob/master/doc/api/allennlp.common.tee_logger.rst

Hopefully adding/removing them is self-explanatory, but getting them to build correctly can be a pain (sphinx is very fiddly). If you ever have trouble with that, let us know, because there's lots of sphinx rules i've internalised inside my head haha.

brendan-ai2:
One clarification, if we literally `git revert 7afe971` that revert will be carried into `master` when we later merge all of this work. So I suspect @DeNeutoy was really suggesting something like:

1. Make a new branch `fix` based off of `torch-distributed`.
2. `git cherry-pick` in just the desired commits from `scarecrow1123:ddp-1`.
3. Then check out `scarecrow1123:ddp-1` and `git reset --hard fix`.

That will ensure `scarecrow1123:ddp-1` contains just the desired commits without the harmful revert. I suspect there's a better way to do step 2 than cherry picking in the commits individually, but this is what I came up with off the top of my head.

allennlp/common/tee_logger.py
scarecrow1123: The actual issue was with calling `logging.shutdown`. According to the [docs](https://docs.python.org/3/library/logging.html?highlight=logging%20shutdown#logging.shutdown) the log handlers are properly flushed/closed by default at module exit. Hence it is not needed and the tests pass now. 

`TeeLogger's` API received a filepath and hence it was necessary to create necessary directories(`TeeLogger` doesn't assume `parent_directory` to be the `serialization_dir`) . But in the current change, everything is taken care in `prepare_global_logging` which assumes the directory is always `serialization_dir` and hence this need not be done.

scarecrow1123:
> it will make rebasing/merging against master easier if we don't pull in diffs from there as we go

> That will ensure scarecrow1123:ddp-1 contains just the desired commits without the harmful revert. I suspect there's a better way to do step 2 than cherry picking in the commits individually, but this is what I came up with off the top of my head.

Apologies for the goof up. I've modified the tree to be clean thanks to @brendan-ai2 's suggestions.

allennlp/training/util.py
scarecrow1123: Sorry, I don't quite get your point here. `total_loss` is a float and I'm not sure what reference we'd be losing. Essentially, `total_loss / num_batches` give the effective loss in a worker and we need to log the average loss across workers. Aggregation happens for all the metrics in the `dict` and hence the effective loss is being added there before the `for` loop. 
Could you elaborate a bit here?

DeNeutoy: Ah I see - the loss has already been all-reduced for the backward pass so here we only care about the value.

scarecrow1123: Strictly speaking, it is the gradients that have been all-reduced during the backward pass.

scarecrow1123:
@DeNeutoy Let me know if there is anything else to be changed here and if this is good to go into the branch. The changes for the next PR is underway.

PR #3375, opened 2019-10-18
Specify batch size through CLI flag in allennlp evaluate

nelson-liu:
This is mostly to match `allennlp predict`


nelson-liu:
thoughts on merging this?

matt-gardner:
Sorry this got dropped!  If it's matching the behavior of `predict`, I see no issue with merging it.  We need to overhaul how all of this works at some point, though, so we have some unified processing between the commands...

Issue #3367, opened 2019-10-16
test-install freeze


chsuong:
**System:**
 - OS: OSX (High Sierra)
 - Python version: 3.7
 - AllenNLP version: 0.9.0

**Question**
The terminal freezes for hours after I run the command "allennlp test-install" after installing allenlp via pip in a conda environment as instructed [here](https://github.com/allenai/allennlp). Why is this happening? Is there another way to run unit tests or check whether it was installed correctly?: 

```
(allennlp) MacBook-Pro:code_data$ allennlp test-install
2019-10-16 10:37:03,783 - INFO - pytorch_pretrained_bert.modeling - Better speed can be achieved with apex installed from https://www.github.com/nvidia/apex .
2019-10-16 10:37:05,651 - INFO - pytorch_transformers.modeling_bert - Better speed can be achieved with apex installed from https://www.github.com/nvidia/apex .
2019-10-16 10:37:05,665 - INFO - pytorch_transformers.modeling_xlnet - Better speed can be achieved with apex installed from https://www.github.com/nvidia/apex .
2019-10-16 10:37:06,233 - INFO - allennlp.common.registrable - instantiating registered subclass relu of <class 'allennlp.nn.activations.Activation'>
2019-10-16 10:37:06,234 - INFO - allennlp.common.registrable - instantiating registered subclass relu of <class 'allennlp.nn.activations.Activation'>
2019-10-16 10:37:06,236 - INFO - allennlp.common.registrable - instantiating registered subclass relu of <class 'allennlp.nn.activations.Activation'>
2019-10-16 10:37:06,237 - INFO - allennlp.common.registrable - instantiating registered subclass relu of <class 'allennlp.nn.activations.Activation'>
2019-10-16 10:37:06,379 - INFO - allennlp.commands.test_install - Changing directory to /anaconda3/envs/allennlp/lib/python3.7/site-packages
2019-10-16 10:37:06,380 - INFO - allennlp.commands.test_install - Running tests at /anaconda3/envs/allennlp/lib/python3.7/site-packages/allennlp
============================= test session starts ==============================
platform darwin -- Python 3.7.4, pytest-5.2.1, py-1.8.0, pluggy-0.13.0
rootdir: /anaconda3/envs/allennlp/lib/python3.7/site-packages
plugins: flaky-3.6.1
collected 1482 items / 11 deselected / 1471 selected                           

allennlp/tests/commands/configure_test.py .                              [  0%]
[...]

PR #3381, opened 2019-10-18
Pin away from torch 1.3.0.

brendan-ai2:
- For https://github.com/allenai/allennlp/issues/3367
- I've verified that this fixes the aforementioned bug using a fresh conda environment in Mac OS 10.13.6.


brendan-ai2:
Thanks!

PR #3384, opened 2019-10-19
fix epoch tracking in cache_instances feature of the iterator

nitishgupta:
The `epoch_num` field is a MetadataField and hence tensorized as a List[int]. 
The cache_instances branch of the data_iterator was treating it as a Tensor and hence throwing an error.


matt-gardner:
Merging this without a test because this code is being replaced, anyway.  This is just fixing a small bug.

PR #3385, opened 2019-10-20
Add support for the trainer param to be a str

bryant1410:
As well as many (all?) params that support both `{"type": "param_type"}` and `"param_type"`, this adds support for the trainer param to be a `str`.

### DIFF
--- a/allennlp/commands/find_learning_rate.py
+++ b/allennlp/commands/find_learning_rate.py
@@ -223,6 +223,9 @@ def find_learning_rate_model(
     train_data = all_datasets["train"]
 
     trainer_params = params.pop("trainer")
+    if isinstance(trainer_params, str):
+        trainer_params = Params({"type": trainer_params})
+
     no_grad_regexes = trainer_params.pop("no_grad", ())
     for name, parameter in model.named_parameters():
         if any(re.search(regex, name) for regex in no_grad_regexes):
--- a/allennlp/commands/fine_tune.py
+++ b/allennlp/commands/fine_tune.py
@@ -333,6 +333,9 @@ def fine_tune_model(
     test_data = all_datasets.get("test")
 
     trainer_params = params.pop("trainer")
+    if isinstance(trainer_params, str):
+        trainer_params = Params({"type": trainer_params})
+
     no_grad_regexes = trainer_params.pop("no_grad", ())
     for name, parameter in model.named_parameters():
         if any(re.search(regex, name) for regex in no_grad_regexes):
--- a/allennlp/tests/training/no_op_trainer_test.py
+++ b/allennlp/tests/training/no_op_trainer_test.py
@@ -3,11 +3,12 @@ from typing import Dict
 
 import torch
 
+from allennlp.common import Params
 from allennlp.common.testing import AllenNlpTestCase
 from allennlp.data import Vocabulary
 from allennlp.data.dataset_readers import SequenceTaggingDatasetReader
 from allennlp.models.model import Model
-from allennlp.training import NoOpTrainer
+from allennlp.training import NoOpTrainer, TrainerBase
 
 
 class ConstantModel(Model):
@@ -32,3 +33,10 @@ class TestNoOpTrainer(AllenNlpTestCase):
         assert metrics == {}
         assert os.path.exists(serialization_dir / "best.th")
         assert os.path.exists(serialization_dir / "vocabulary")
+
+    def test_trainer_from_base_class_params_trainer_str(self):
+        params = Params.from_file(self.FIXTURES_ROOT / "simple_tagger" / "experiment.json")
+        params["trainer"] = "no_op"
+
+        # Can instantiate from base class params
+        TrainerBase.from_params(params, self.TEST_DIR)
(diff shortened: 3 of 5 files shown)


bryant1410:
> Do you have a concrete use case for this that you could share?
> 
> Not a fan currently as it doesn't make sense to define a trainer from a string only, whereas it does for other params, such as activation functions.

`no_trainer`

bryant1410:
I see this issue happens in other places. What if `Params` had a method `pop_params` (or similar) like the following:

```python
def pop_params(self, key: str) -> Params:
    params = self.pop(key)
    if isinstance(params, str):
        params = Params({"type": params})
    else:
        assert isinstance(self, Params)
    return params
```

So it's like `pop_bool`, `pop_int` and so on, but to get `Params`. The rationale is that when you get `trainer`, `model`, `lr_scheduler`, `regularizer` or whatever thing, it could be a string that represents the type or it could be a `Params` like `{"type": ...}`. So you always get a `Params`, which is what you get (that's why `pop_params`).

DeNeutoy:
It happens in exactly 2 places:

https://github.com/allenai/allennlp/blob/3dda5ac9fcf390d8b83eb855249360711707c11c/allennlp/training/optimizers.py#L42

https://github.com/allenai/allennlp/blob/3dda5ac9fcf390d8b83eb855249360711707c11c/allennlp/nn/regularizers/regularizer_applicator.py#L79

I don't know why I did this in the first place.... Basically it is not good but it has to be there for backward compatibility. Sorry!

bryant1410:
It also happens in:

https://github.com/allenai/allennlp/blob/3dda5ac9fcf390d8b83eb855249360711707c11c/allennlp/models/model.py#L320

https://github.com/allenai/allennlp/blob/457a85bde57ba58289d32e4e1d07df1f94c813b0/allennlp/data/tokenizers/tokenizer.py#L61

https://github.com/allenai/allennlp/blob/457a85bde57ba58289d32e4e1d07df1f94c813b0/allennlp/common/from_params.py#L357

https://github.com/allenai/allennlp/blob/457a85bde57ba58289d32e4e1d07df1f94c813b0/allennlp/common/from_params.py#L234

And for consistency, it should also work with `trainer` and `lr_scheduler`, so I would count them as places. Otherwise, you need to remember in what place is okay to have a string and in what places you need to put `{"type": ...}`. I think that's pretty inconsistent.

I believe the most elegant and compatible solution is to have a function similar to the one I mention.

matt-gardner:
So, this is something that we support natively inside of `FromParams`, right?  We call a no-argument constructor if we just get a string for a registrable type.  It doesn't work here because we're not actually calling `Trainer.from_params`, right?  I agree with @DeNeutoy that it's not great to have custom logic all over the place to handle this, but if there's a centralized place to fix this, I see no problem with it.

matt-gardner:
And to be clear, I think that `from_params` is the right place to be handling this, not any `pop` method.  Having that method modify the dictionary seems too confusing.

bryant1410:
> So, this is something that we support natively inside of `FromParams`, right? We call a no-argument constructor if we just get a string for a registrable type. It doesn't work here because we're not actually calling `Trainer.from_params`, right? I agree with @DeNeutoy that it's not great to have custom logic all over the place to handle this, but if there's a centralized place to fix this, I see no problem with it.

Yeah, that's the reason.

> And to be clear, I think that `from_params` is the right place to be handling this, not any `pop` method. Having that method modify the dictionary seems too confusing.

Makes sense.

The issue is that all these places need to be changed because actually they have duplicate code, it looks like they need a refactor. `fine_tune`, `find_learning_rate` and `dry_run` (I forgot to modify `dry_run` in this PR) all have similar logic to what `TrainerPieces.from_params` does, the problem is that they do something slightly different and that's likely why they don't call it. The piece of code they all do that concern this change is the following (which to me should be done by `Trainer`, but there's likely a reason):

https://github.com/allenai/allennlp/blob/3dda5ac9fcf390d8b83eb855249360711707c11c/allennlp/training/trainer_pieces.py#L97-L110

(btw, the 2nd block is not in `find_learning_rate`, I don't know why)

Plus `TrainerBase.from_params` would need to be changed as well. So, thinking this way, it's only 2 places.

Now, the actual problem, in my opinion, is that there's this logic repeated all over. Maybe, at some point, there will be no more `Trainer` (btw, is there an ETA?) and this can be more easily solved? In this case, I think `TrainerBase.from_params` wouldn't even exist. And it would be a centralized place to change this, I guess.

So, I guess this change can either be done in all these places or deferred until all this is refactored (so it's a centralized thing).

matt-gardner:
My vote would be to not put an ugly patch on a very minor problem with an easy work around, but instead fix the underlying issue (the refactoring / duplicated code), which will then make it easy to add a not-ugly patch to fix the problem.  If you want to help refactor things so there isn't so much duplicated code, that would be awesome.  I'm not sure what you mean by there being no more `Trainer`, though.

bryant1410:
I mean, isn't it the idea to only keep `CallbackTrainer` in the future?

matt-gardner:
I don't think that changes this issue at all, right?

bryant1410:
Yeah, I think it makes removing the duplicate code easier.

bryant1410:
Also, do you know why `TrainerPieces` is currently not being called from
`Trainer.from_params` (like the other trainers)?

matt-gardner:
I haven't looked at the trainer code in quite a while, so I'm not sure.  It's probably historical, or keeping things for backwards compatibility, or something.

PR #3386, opened 2019-10-21
New Iterators

DeNeutoy:
This PR adds the new allennlp data iteration API.
It's quite large, but conceptually it contains only a few pieces.

### Moving registered class names and adding stubs

- `BucketIterator` -> "bucket_old"
- `BasicIterator` -> "basic_old"
- `SameLanguageIterator` -> "same_language_old"
- `HomogeneousBatchIterator` -> "homogenous_batch_old"

All the above classes now have a `Stub` class which constructs the new `TransformIterator` with the correct arguments so as to have the exact same functionality. Doing it like this fulfills two objectives:

1. People who have inherited from any of the iterators can continue to use their classes as normal.
2. People using the iterators from a configuration file automatically start using the new iterators.

### New things

- `transforms.py`
- `transform_iterator.py`

The new `TransformIterator` is also compatible with both torch datasets, `Dataset` and `Iterable` dataset.

- Includes a duplicate set of tests, https://github.com/allenai/allennlp/pull/3386/files#diff-86909b900b648b31406858bb955e2d6c and https://github.com/allenai/allennlp/pull/3386/files#diff-0f1d14bf3326c70991f15d2d6f8c76a0 which check that all the functionality of the old `BasicIterator` and `BucketIterator` is maintained by the new `TransformIterator`. For the other two iterators, the tests are included in the original test files, because they were pretty small.


allennlp/data/transforms.py
```diff
@@ -0,0 +1,466 @@
...
+from allennlp.common.util import lazy_groups_of
+
+
+Batched = Iterable[Instance]
+InstanceOrBatch = TypeVar("InstanceOrBatch", Iterable[Instance], Instance)
+
+
+class DatasetFromList(TorchDataset):
+    def __init__(self, instances: Iterable[InstanceOrBatch]):
+        self.instances
```
sai-prasanna: self.instances = instances ?

DeNeutoy: Good job this class is not used anywhere 😅

sai-prasanna: please show TorchDataset some ♥️ 😂

allennlp/data/transforms.py
```diff
@@ -0,0 +1,490 @@
...
+
+from allennlp.common.util import lazy_groups_of
+
+
+Batched = Iterable[Instance]
+InstanceOrBatch = TypeVar("InstanceOrBatch", Iterable[Instance], Instance)
+
+
+class DatasetFromGenerator(IterableTorchDataset):
+    def __init__(self, generator: Iterable[Union[Iterable[Instance], Instance]]):
```
bryant1410: `generator: Iterable[InstanceOrBatch]`?

allennlp/data/transforms.py
```diff
@@ -0,0 +1,490 @@
...
+        self.generator = generator
+
+    def __iter__(self) -> Iterable[Union[Iterable[Instance], Instance]]:
```
bryant1410: `-> Iterable[InstanceOrBatch]:`?

allennlp/data/transforms.py
```diff
@@ -0,0 +1,531 @@
...
+from allennlp.common.registrable import Registrable
+from allennlp.data.instance import Instance
+from allennlp.data.vocabulary import Vocabulary
+
+from allennlp.data.fields import MetadataField
+
+from allennlp.common.util import lazy_groups_of
+
+
+Batched = Iterable[Instance]
```
brendan-ai2: Nit: Why not `Batch`?

allennlp/data/transforms.py
```diff
@@ -0,0 +1,531 @@
...
+        raise NotImplementedError
+
+    def transform_batch(self, batches: Iterable[Batched]) -> Iterable[Batched]:
+
+        # TODO Maybe switch this so that the base class just yields and the
+        # subclasses which are generic over Batched yield from
+        for batch in batches:
+            yield from self.transform(batch)  # type: ignore
+
+    def __call__(self, dataset: Iterable[Instance]) -> Iterable[InstanceOrBatch]:
```
brendan-ai2: I understand why this functionality is nice for writers of transforms, but I can't shake the feeling that we have an opportunity here to make the `Transform` abstraction a bit cleaner. Essentially, the top-level `Transform` could be completely generic. It maps A -> B. One use case could be that a transform could take in URL strings and output iterables of instances. Of course, for the bulk of the transforms this dispatching magic is still useful, so we would have an `InstanceTransform` (or something) superclass that preserved it.

allennlp/data/transforms.py
```diff
@@ -0,0 +1,531 @@
...
+            instances, self.sorting_keys, self.vocab, self.padding_noise
+        )
+
+        yield instances
+
+
+@Transform.register("epoch_tracker")
+class EpochTracker(Transform[Instance]):
+    """
+    Adds a allennlp Field to each Instance which specifies how many
```
brendan-ai2: Nit: "Adds an"

allennlp/data/transforms.py
```diff
@@ -0,0 +1,531 @@
...
+
+    def transform(self, dataset: Iterable[Instance]) -> Iterable[Instance]:
+        i = 0
+        for instance in dataset:
+            yield instance
+            i += 1
+            if i >= self.max:
+                break
+
+    def transform_batch(self, batches: Iterable[Batched]) -> Iterable[Batched]:
```
brendan-ai2: I think I'm missing something here. Why does `StopAfter` operate differently in the `transform` and `transform_batch` functions?

allennlp/tests/data/iterators/bucket_iterator_test.py
```diff
@@ -60,9 +60,9 @@ def test_biggest_batch_first_works(self):
     def test_from_params(self):
 
         params = Params({})
-
-        with pytest.raises(ConfigurationError):
-            iterator = BucketIterator.from_params(params)
+        # No padding keys is explicitly allowed.
+        iterator = BucketIterator.from_params(params)
+        assert iterator._sorting_keys is None
```
brendan-ai2: Per my comment above, I think we should preserve this.

allennlp/tests/data/transforms/basic_iterator_transforms_compat_test.py
```diff
@@ -0,0 +1,295 @@
+from typing import List, Iterable, Dict, Union
```
brendan-ai2: This is pretty much just a copy of the existing test, right? As a reader, it's a bit hard to see that as-is. And we'll also be liable to drift. Would it be possible to reduce this duplication by changing the existing tests to call a factory function that either uses the old class or the new stub? Then we could parameterize the test to use the factory function in one way or the other.

allennlp/tests/fixtures/basic_classifier/common.jsonnet
```diff
@@ -19,7 +19,8 @@
     "iterator": {
         "type": "bucket",
         "sorting_keys": [["tokens", "num_tokens"]],
-        "batch_size": 5
+        "batch_size": 5,
+        "padding_noise": 0.0
```
brendan-ai2: Is `"padding_noise": 0.0` necessary now? I'd like to avoid breaking existing config if at all possible.

allennlp/tests/models/reading_comprehension/naqanet_test.py
```diff
@@ -14,4 +14,7 @@ def setUp(self):
 
     @flaky(max_runs=3, min_passes=1)
     def test_model_can_train_save_and_load(self):
-        self.ensure_model_can_train_save_and_load(self.param_file)
+        self.ensure_model_can_train_save_and_load(
+            self.param_file,
+            gradients_to_ignore={"_passage_span_end_predictor._linear_layers.1.bias"},
```
brendan-ai2: I'm always paranoid about test changes in this sort of PR. ;) What's up?

allennlp/tests/fixtures/basic_classifier/common.jsonnet
DeNeutoy: These were just spurious, non-deterministic test failures which I fixed (they were failing in model test where we compare 2 datasets read from file are the same), so I removed the non-determinism in the bucketing process here. Same below.

allennlp/data/iterators/basic_iterator.py
```diff
@@ -1,24 +1,68 @@
...
 from allennlp.data.instance import Instance
-from allennlp.data.iterators.data_iterator import DataIterator
 from allennlp.data.dataset import Batch
 
 logger = logging.getLogger(__name__)
 
 
 @DataIterator.register("basic")
-class BasicIterator(DataIterator):
+class BasicIteratorStub(DataIterator):
```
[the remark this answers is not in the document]
DeNeutoy: Gotcha 👍

allennlp/data/iterators/bucket_iterator.py
```diff
@@ -108,8 +217,6 @@ def __init__(
         maximum_samples_per_batch: Tuple[str, int] = None,
         skip_smaller_batches: bool = False,
     ) -> None:
-        if not sorting_keys:
```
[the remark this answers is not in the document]
DeNeutoy: Hmm - i've actually come to a slightly different conclusion, which is that users assume that `BasicIterator` is a good starting point and do not discover that BucketIterator would be better. I'm generally against having arguments which duplicate checks like whether the sorting keys are none, because it's not super intuitive. Perhaps a warning is good here, if you don't pass any sorting keys?

allennlp/data/iterators/basic_iterator.py
```diff
@@ -1,24 +1,68 @@
...
+        from allennlp.data.iterators.bucket_iterator import BucketIteratorStub
+
+        return BucketIteratorStub(
+            batch_size=batch_size,
+            instances_per_epoch=instances_per_epoch,
+            max_instances_in_memory=max_instances_in_memory,
+            cache_instances=cache_instances,
+            track_epoch=track_epoch,
+            maximum_samples_per_batch=maximum_samples_per_batch,
+        )
```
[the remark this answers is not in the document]
DeNeutoy: Ah, yep that's better. Thanks.

allennlp/data/iterators/transform_iterator.py
```diff
@@ -0,0 +1,143 @@
...
+            key = id(instances)
+
+            iterator = self._cursors.get(key, itertools.cycle(instances))
+            data = transforms.Compose(dataset_transforms)(iterator)
+            batch_generator = iter(DataLoader(data, batch_size=1, collate_fn=self._collocate))
+
+            while True:
+                try:
+                    # If there are instances left on this iterator,
+                    # yield one and decrement max_instances.
```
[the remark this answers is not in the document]
DeNeutoy: No longer applies, removed.

allennlp/data/transforms.py
DeNeutoy: The default implementation of the `transform_batch` is to apply `transform` to each batch passed to `transform_batch` and `yield from`, resulting in a flattened stream (which might still be batched, because some transforms map from `Instances -> Batched`). So transform_batch in this case is actually implementing `StopAfter` at the batch level, which is what would be expected. 

I agree it's a bit odd that this is the default implementation (rather than assuming that `transform_batch` would apply `transform` and just `yield`) - see [this comment](https://github.com/allenai/allennlp/pull/3386/files#diff-41d08074b6dcde4374683117cb26a2cbR46). I did it this way around just because more of the `Transforms` use this than don't, which resulted in less of them needing to override the method. I can switch it around if you want.

allennlp/data/transforms.py
DeNeutoy: We already have a `Batch` class which actually does some work, so I was trying to avoid confusion. I changed it to `Batch`.

DeNeutoy: Oh wait, we already have a `Batch` transform in the same file. That's why this needed to be batched...

allennlp/data/transforms.py
DeNeutoy: I added this - I think it will be good for people who know what they are doing (i.e they know that the types of the transforms that get chained together have to match up exactly). The main reason for this dispatching logic is so that any of the `Transforms` can be chained together in arbitrary order and not break. But it would be great if we could just have a dataset of url strings, and then have transforms which convert them into instances. It doesn't exactly play nicely with the rest of the library (e.g we use `text_to_instance` from the dataset reader in other places, such as the predictor, and it's not clear how that would work if the thing that produces instances is now a Transform) but definitely it's cleaner.

allennlp/tests/data/transforms/basic_iterator_transforms_compat_test.py
DeNeutoy: I'm going to leave this as is, because 1) It's not too much duplication, 2) immediately after merging this PR i'll open a PR to depreciate the old iterators, and it will be easier to completely remove.

There is additionally a functional difference, which is that a lot of the iterator tests use the private method `_create_batches` to generate the batches in the tests - the new iterator actually yields `TensorDicts` here, because the operation is expensive and should be pushed inside the dataloader workers. So to test, I monkeypatched that function to _not_ call `batch.as_tensor_dict()`, so I could test the new iterator fulfilled the spec of the old iterator. Some of the tests require that patch and some don't, which made parameterising the tests a bit messy. Because of this I think it's easier to have them separate.

allennlp/data/iterators/bucket_iterator.py
```diff
@@ -95,9 +96,117 @@ class BucketIterator(DataIterator):
...
+
+        if track_epoch:
+            dataset_transforms.append(transforms.EpochTracker())
+
+        if max_instances_in_memory is not None:
+            dataset_transforms.append(transforms.MaxInstancesInMemory(max_instances_in_memory))
+
+        if sorting_keys is not None:
+            # To sort the dataset, it must be indexed.
+            # currently this happens when we call index_with, slightly odd
```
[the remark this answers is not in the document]
DeNeutoy: Oh - no, it's not different from the old way, it's just a bit more odd in the new functional iterator style. E.g this transform would break, apart from we add in the `Index` transform automatically when we call `iterator.index_with(vocab)`.

DeNeutoy:
@brendan-ai2 i've addressed most of your comments, let me know what you think!

allennlp/data/transforms.py
matt-gardner: It's possible that you might want to base classes here, one for instance -> instance transforms, and one for instance -> batch transforms, that have different default implementations of `transform_batch`.

Note that this is not in conflict with my saying that we should not have separate `Transform` and `InstanceTransform` classes.  That complaint is about the *type system* and how it interacts with `FromParams` and such - it's really annoying and adds a bunch of cognitive load to see (and use) `List[InstanceTransform]` in my type annotations for the data iterators, instead of just `List[Transform]` (i.e., I have to think about what the difference is, why there's a difference, and which one I need to use).   I think using `List[InstanceTransform]` in the data iterators even precludes the advanced usage you were talking about, like passing in strings and transforming them to lists of instances.

allennlp/data/transforms.py
```diff
@@ -0,0 +1,558 @@
...
+            "Add a transform which batches your instances to your pipeline."
+        )
+
+    def transform_batch(self, batches: Iterable[Batched]) -> Iterable[Batched]:
+
+        batches = list(batches)
+        if len(batches) <= 2:
+            yield from batches
+        else:
+            last = batches.pop()
```
matt-gardner: I don't think this does what's advertised.  Why will the last batches be biggest here?

allennlp/data/transforms.py
```diff
@@ -0,0 +1,558 @@
...
+class Shuffle(InstanceTransform[Batched]):
+    """
+    Shuffles a set of instances and returns them.
+
+    NOTE: This could be generic over batches and instances, i.e
+    for the batch case it just shuffles the batches. This is purposely NOT
+    like that, because it makes it easy to accidentally shoot yourself
+    in the foot if you have a dataset which is infinite, because in the batched
+    case, we still try to read all batches into memory. This way,
+    if this is called on a dataset which is already batched, it shuffles within
```
matt-gardner: Shuffling within a batch accomplishes nothing.  If I sort by padding, how am I supposed to ensure that the batches are shuffled?

allennlp/data/iterators/basic_iterator.py
```diff
@@ -1,24 +1,52 @@
...
+
+
+@DataIterator.register("basic_old")
+class BasicIterator(DataIterator):
```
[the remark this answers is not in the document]
DeNeutoy: I'm in favour of stripping these old ones out, but it would break backward compatibility for anyone who has subclassed the iterators.

allennlp/data/iterators/transform_iterator.py
```diff
@@ -0,0 +1,136 @@
...
+from allennlp.data.vocabulary import Vocabulary
+from allennlp.data import transforms
+
+logger = logging.getLogger(__name__)
+
+TensorDict = Dict[str, Union[torch.Tensor, Dict[str, torch.Tensor]]]
+
+
+@DataIterator.register("transform")
+class TransformIterator:
```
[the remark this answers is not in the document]
DeNeutoy: Yeah... It can subclass it, but the interface for the private `_create_batches` method is different. The objective is to remove all the other iterators anyway, so idk what's best/if this distinction is important.

allennlp/data/iterators/same_language_iterator.py
```diff
@@ -22,7 +25,7 @@ def split_by_language(instance_list):
 
 
 @DataIterator.register("same_language")
-class SameLanguageIterator(BucketIterator):
+class SameLanguageIteratorStub:
```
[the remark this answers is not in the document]
DeNeutoy: For backward compatibility, yeah I don't know why both these iterators exist in the first place, I guess we just missed the similarity in #2628.

Thanks, I missed that stub vs shim thing.

allennlp/data/iterators/basic_iterator.py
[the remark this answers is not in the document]
matt-gardner: I vote for just removing them. The exposure is probably small, and the reduction in complexity those people would gain by upgrading is worth it.

allennlp/tests/data/transforms/basic_iterator_transforms_compat_test.py
brendan-ai2: My concern isn't one of code purity or anything. It's a practical question, "How do we avoid regressions when using different tests?" Even though my prior is that you did a great job on these tests, it's really hard to actually help as a reviewer with a big diff like this. (And personally, I'd have difficulty ensuring I didn't make some mistake when converting the tests in this manner.) Another issue is that given that we want to support people who may have subclassed the old iterators, the deprecation may not move particularly quickly. I'd argue that this sort of thing is really worth spending some time on.

Thanks for pointing out the monkey patch issue, that's definitely complicates things. Could you add some comments around why that's necessary? I'll take a stab at parameterizing things. Have a rough idea of something that might work...

allennlp/tests/fixtures/basic_classifier/common.jsonnet
brendan-ai2: Sorry if I'm being dense, but why did this only start now? I don't see spurious failures on http://build.allennlp.org/viewType.html?buildTypeId=AllenNLP_AllenNLPPullRequests&branch_AllenNLP=%3Cdefault%3E&tab=buildTypeStatusDiv.

allennlp/data/iterators/bucket_iterator.py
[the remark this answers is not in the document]
brendan-ai2: Maybe reword to something like: "Adding the SortByPadding transform requires that the instances will be indexed by the time the transform is called. Currently this happens when we call index_with in XYZ." I'd drop the part about it being odd unless you want to explain why or suggest an alternative.

brendan-ai2:
Sorry, that last pass was concurrent with @matt-gardner's. May be duplicate/conflicting advice.

allennlp/data/iterators/basic_iterator.py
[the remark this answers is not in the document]
brendan-ai2: @matt-gardner, I've been struggling with statements like "The exposure is probably small". It seems very difficult to actually measure this. At the same time, "reduction in complexity" is quite nebulous while "the code still runs" is very concrete. Given that we've been able to engineer this PR such that standard users of the jsonnet API will get the desired benefit it seems a shame to drop support for the more advanced subclass use-case.

allennlp/data/transforms.py
```diff
@@ -0,0 +1,558 @@
...
+    def __init__(self, generator: Iterable[Union[Iterable[Instance], Instance]]):
+        self.generator = generator
+
+    def __iter__(self) -> Iterable[Union[Iterable[Instance], Instance]]:
+
+        for x in self.generator:
+            yield x
+
+
+class Transform(IterableTorchDataset, Registrable):
```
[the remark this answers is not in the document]
brendan-ai2: Maybe this is due to my background in other languages, but I'd like to push back a bit on playing loose with types. Though mypy doesn't always do the best job of enforcing type constraints, the type annotations are still very useful for communicating with the reader of the code. I've been tripped up in Python due to type mismatches and I think that sort of hack should be reserved for extraordinary cases.

allennlp/data/transforms.py
brendan-ai2: Honestly, I'm okay with leaving this out for now too. My initial worry was that if we don't do it now, we'll paint ourselves into a corner, but we have until at least 1.0 to undo that. Waiting to generalize is of course eminently reasonable. My hunch is that we will want to generalize because having `DatasetReaders` fit the `Transform` API will allow us to move reading into the `DataLoader` which gives us much better multiprocess support than our homegrown `MultiprocessDatasetReader`.

allennlp/data/transforms.py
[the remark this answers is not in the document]
matt-gardner: Using `Any` as a type annotation doesn't get you any type safety, anyway, so I'm really not sure what you gain with this separation. It seems strictly better to just have one type, `Transform`, and parameterizing both its input and output if you really care.

The only thing I'm really complaining about is the external use of `InstanceTransform`. If the external type annotations all use `List[Transform]`, I'd be fine with this. I basically suggested doing it in another comment. But I'd still make `Transform` generic, instead of using Any.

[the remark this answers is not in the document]
brendan-ai2: +1 for generics over `Any`!

allennlp/data/iterators/basic_iterator.py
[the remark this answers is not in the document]
matt-gardner: I guess I don't think we should be total slaves to backwards compatibility, at the expense of good, clean code. I agree with you on trying really hard to preserve compatibility on most releases, but we have an opportunity with 1.0 to significantly clean things up; let's not hamstring ourselves in the name of backwards compatibility.

allennlp/data/transforms.py
matt-gardner: If I missed something about why this is necessary for integrating with `DatasetReaders`, then by all means, tell me I'm wrong, that's fine.  But, to be more clear, here is a sketch of a proposal that I think we can all agree on:

```python
class Transform(Registrable, Generic[Input, Output]):
    def transform(self, dataset: Iterable[Input]) -> Union[Iterable[Output], Iterable[Iterable[Output]]]:
        raise NotImplementedError
    def transform_batches(self, batches: Iterable[Iterable[Input]]) -> Iterable[Iterable[Output]]:
        raise NotImplementedError
    def __call__(self, dataset: Union[Iterable[Input], Iterable[Iterable[Input]]]) -> ...:
        # base implementation of __call__ goes here

class InstanceTransform(Transform[Instance, Instance]):
    def transform_batches(self, ...):
        # handle case for instance -> instance transforms here

class InstanceToBatchTransform(Transform[Instance, Instance]):
    def transform_batches(self, ...):
        # handle case for instance -> batch transforms here
```

And all type annotations (for `Registrable` / `FromParams` purposes) use `Transform` directly, not either of the bottom two classes.

We're not going to get complete type safety when composing lists of transforms, especially when we read those transforms from a config file.  But that's ok; that's a benefit of using python instead of scala, where this wouldn't actually compile.  And this will be pretty close.

matt-gardner: So, I tried for a while to get the types to actually work out with what I suggested above, and I couldn't figure it out.  So, here's a better proposal with types that actually work:

```python
class Transform(Registrable, Generic[Input, Output]):
    def transform(self, dataset: Iterable[Input]) -> Iterable[Output]:
        raise NotImplementedError

    def __call__(self, dataset: Union[Iterable[Input], Iterable[Iterable[Input]]) -> Union[...]:
        # This is where we'll hide our type checks to handle the union of possible inputs
        raise NotImplementedError

    # no transform_batches; they types don't really work out to do this right, and it's not needed

class InstanceTransform(Transform[Instance, Instance]):
    # transform() is still abstract, but has type Iterable[Instance] -> Iterable[Instance]

    def __call__(self, Union[...]) -> Union[...]:
        # handle case for instance -> instance transforms here

class InstanceToBatchTransform(Transform[Instance, Iterable[Instance]]):
    # transform() is still abstract, but has type Iterable[Instance] -> Iterable[Iterable[Instance]]

    def __call__(self, ...):
        # handle case for instance -> batch transforms here

class BatchToBatchTransform(Transform[Instance, Iterable[Instance]]):
    # transform() is still abstract, but has type Iterable[Iterable[Instance]] -> Iterable[Iterable[Instance]]

    def __call__(self, ...):
        # handle case for batch -> batch transforms, including throwing an error if not already batched

class FileToInstancesTransform(Transform[str, Instance]):
    # transform() is abstract, but is basically a dataset reader, or calls one

    def __call__(self, dataset: Iterable[str]) -> Iterable[Instance]:
        yield from self.transform(dataset)
```

allennlp/tests/fixtures/basic_classifier/common.jsonnet
DeNeutoy: The build began failing at  https://github.com/allenai/allennlp/pull/3386/commits/28335cde7f2179eaaa2b2fbc99a757608f989da9:

http://build.allennlp.org/viewLog.html?buildId=21427&buildTypeId=AllenNLP_AllenNLPPullRequests&tab=buildLog&_focus=1481

3 tests failed: all of them were because differences not in the model but checking that the iterators load an identical first batch, which happens here:

https://github.com/allenai/allennlp/blob/master/allennlp/common/testing/model_test_case.py#L122


The batch elements are identical but in a different order (I checked). This test was essentially equivalent to checking the order of a list which with 0.1 prob has it's order changed.

The next commit https://github.com/allenai/allennlp/pull/3386/commits/285d60a13beffa70a7fac82ee96341f4a44c6ad3, also failed:

http://build.allennlp.org/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=21428&_focus=1635

This time it was the NaQaNet test, which is already flaky - the reason this test fails so much is because it has a lot of layers with relus in it, see here: 
https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2seq_encoders/qanet_encoder.py#L202

allennlp/tests/data/transforms/basic_iterator_transforms_compat_test.py
DeNeutoy: Thanks - I don't need you to take a stab at parameterizing it, I actually had it similar to what you are suggesting before: https://github.com/allenai/allennlp/pull/3386/commits/770999c98b34dfec51cb8f0ea360373cb630e9b5

I did it like this because I actually think it's better. Regardless, we need to iron out whether we are removing these old iterators before I change this.

Feel a bit stuck in between a rock and a hard place here - if i'd have changed the tests for the current iterators, your comment would have been that it's unclear if backward compatibility is maintained.

allennlp/data/transforms.py
brendan-ai2: @matt-gardner, thanks a lot for sketching that out! That looks quite good. @DeNeutoy, I think we talked about something like this, but you were concerned that it would cause some duplication because some transforms need to operate on instances directly as well as instances that were already batched. If I understand Matt's code correctly, we would still have some special dispatch in `InstanceTransform.__call__` to handle that case. Alternatively, I think we could have `Transform.__call__` take just `Iterable[Input]` (no `Union`) and provide an explicit `UnpackTransform`. Something like:

```
class UnpackTransform(Transform[Iterable[Instance], Iterable[Instance]]):
    def __init__(self, base_transform: InstanceTransform):
        self._base_transform = base_transform

    def __call__(self, input):
        for batch in input:
            yield self._base_transform(batch)

# In user code
pipeline = [CreateBatches(), ShuffleBatches(), UnpackTransform(SomeInstanceTransform())]
...
```

I slightly prefer being explicit in that manner, but the automatic dispatch seems okay too.

brendan-ai2: Slight correction:

```
class UnpackTransform(BatchTransform):
    def __init__(self, base_transform: InstanceTransform):
        self._base_transform = base_transform

    def transform(self, input: Iterable[Iterable[Instance]]) -> Iterable[Iterable[Instance]]:
        for batch in input:
            yield self._base_transform(batch)
```

allennlp/data/iterators/basic_iterator.py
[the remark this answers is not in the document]
DeNeutoy: I agree with Matt here - I don't think the reduction in complexity is nebulous - it allows us to make changes to the codebase significantly faster. The diff on this PR would be half the size (in terms of lines added) if we removed the old iterators.

Additionally, it's at least clear in my mind that backward compatibility itself doesn't drive usage - look at pytorch transformers. 

Regardless, what we decide here makes a big difference to how I proceed with other parts of this PR, so...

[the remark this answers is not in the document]
brendan-ai2: I'm very confused by the claim that the PR would be half the size in terms of lines added. Are you talking about not having the shims at all? If so, I think that would be a major mistake. You did a great job on the shims and I think it will be very helpful to maintain that for people.

The only additional lines of code to maintain the old iterators as-is is some duplicated comments and function/class prototypes. Maybe the tests would have been modified instead of duplicated, but that the work to upgrade them would be the same, so I'm not sure about moving faster...

allennlp/data/transforms.py
DeNeutoy: hmm, @matt-gardner how does this handle the following case:

```
pipeline = [Index(), MaxInstancesInMemory(), Batch()]
pipeline = [Index(), Batch()]
```
Is the idea that there would be two types of batch transforms, one inheriting from `BatchTransform` and one inheriting from `InstanceToBatchTransform`? This seems not great as it means that in a config, you have to know which one to use, which is something I had thought we were explicitly trying to avoid? I'm not sure I see how having the type dispatching in the base `Transform` class would actually avoid this, because if there was only a single class, `Batch[Instance, Iterable[Instance]]`, and it was called with `Iterable[Instance]`, we still don't have strict typing anyway. I think the part i'm missing is how the dispatching would happen without having to have multiple classes?

allennlp/tests/data/transforms/basic_iterator_transforms_compat_test.py
brendan-ai2: Sorry, my intent is definitely not to make you feel stuck. Hence why I should try to parameterize to demonstrate what I'm talking about. My goal is to do something more surgical such that it's clear that backwards compatibility is maintained. (By the way, was https://github.com/allenai/allennlp/commit/770999c98b34dfec51cb8f0ea360373cb630e9b5 really the one you meant? Doesn't seem parameterized.)

allennlp/data/transforms.py
DeNeutoy: @brendan-ai2 i'm not a massive fan of `UnpackTransform` - it seems better to just have duplicate classes in this case, because I don't think it would be clear to a user which transforms you can unpack. Also, I don't remember having that objection in a conversation - my original conception of this was to have two classes, `Transform` and `BatchTransform`, but Matt convinced me that it was possible to make them order agnostic with this dispatching stuff.

brendan-ai2: Thanks for the correction regarding the motivation. `UnpackTransform` would work on any transform that is an `InstanceTransform` which is clear from the docs/class declaration. If a user is building their own transform pipeline, they'll presumably need to look at the docs a bit regardless. But like I said, automatic dispatch sounds fine to me too.

matt-gardner: `CreateBatches` is an `InstanceToBatchTransform`, and that class has logic inside of `__call__` to handle the unwrapping that needs to be done when it's given batches as input instead of `Iterable[Instance]`, exactly as you have logic for it in `transform_batch()` now.  We only need one class for this, just as you only need one class in what you have.

There are really just a few simple differences between what you have and what I'm suggesting:
- I'm taking the separate base implementations of `transform_batch` (where you have one of them implemented by default, and the other two overridden manually), and putting them into the `__call__` method of separate base classes.  The concrete transforms all inherit from the appropriate base class, and you basically never have to override `__call__` (except maybe for `StopAfter`, which probably doesn't even need to implement `transform()`).
- I've made the types on the base class explicit with generics, and recommend just annotating uses of these transforms as `List[Transform]` instead of `List[InstanceTransform]`.

I agree with @DeNeutoy on not being a fan of `UnpackTransform` - we should just design the transforms to do the right thing, or crash with a meaningful error if the order you gave doesn't work.

I also would make one small tweak to my proposed API: `__call__` on `Transform` should just have type `Iterable[Input] -> Iterable[Output]`.  We'll break this with `Union` types on some of the subclasses, but that's ok, I think.  The point of separating out `__call__` instead of having everything be in `transform()` is to have a place where we can play loose with the types where it's helpful, while still having `transform()` make sense and be easy to reason about.

allennlp/data/iterators/basic_iterator.py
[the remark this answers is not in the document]
matt-gardner: No, it's not about the shims - we need those for backwards compatibility of config files (which I definitely approve of).  It's about retaining the old classes themselves, registered as `"old_*"`.  I don't think we need to do that, the classes should just be removed.

What's the worst case that happens to someone who has subclassed one of the classes we remove?  They copy the code from an earlier release of allennlp into their repo, change an import line or two, and everything still works.  I don't think that's a very bad outcome, and it saves us a significant maintenance and code complexity burden.  Backwards compatibility at all costs is far too extreme; this is a case where the costs are worth breaking compatibility, in my opinion, as the costs are low and the gain is high.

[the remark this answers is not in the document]
brendan-ai2: Absolutely agreed that backwards compatibility at all costs is too extreme. My claim is that the ongoing maintenance costs will actually be very low. Essentially we'd just slap a deprecation warning on the classes, a comment saying that users should use `TransformIterator` and never look at them again.

To clarify one point about the impact, this change doesn't just affect users that subclassed our iterators, it also affects those that wrote new iterator classes and dispatched to `_create_batches` (as its contract has changed). For instance, https://github.com/allenai/ernie/blob/master/ernie/self_attn_bucket_iterator.py.

At any rate, if we want to break backwards compatibility now is a good time, so consider my objection lodged and carry on if that's your decision.

allennlp/data/iterators/same_language_iterator.py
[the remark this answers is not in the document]
maksym-del: We have two of them because they were landed at about the same time and we did not manage to synchronize.

While ``homogenous batch`` is more general, ``same lang`` supports bucketing. ``homogenous batch`` with bucketing would be the best of two worlds.

sai-prasanna:
Any help needed on this one? Would be glad to.

dirkgr:
I had a comment here questioning why there needs to be such a heavy-handed treatment of something that's essentially just a generator, or a system of generators that can be composed.

But I think I get it now. It's because they all have to be `Registrable` and configured from Jsonnet, right?

matt-gardner:
The need to instantiate stuff from a config file should not factor into any API decision that we make. The API should make sense whether you use it by creating python objects or by using config files.  I'm trying to fix `FromParams` so that it is entirely behind-the-scenes magic that you don't have to worry about if you don't want to use it, and that never shows up in your code other than the call to `@register()`.

The reason this API is "heavy-handed", I believe, is because we want to be able to compose arbitrary transforms.

dirkgr:
Raw generators compose too. Couldn't the code read like this?

```Python
from allennlp import transforms

instances = reader.read()
instances = transforms.index(instances)
instances = transforms.sort_by_padding(instances)
batches = transforms.batch(instances, batch_size=32)
batches = transforms.skip_smaller_than(batches, min_size=32)
batches = transforms.shuffle(batches)
```

Since all of these are generators, they also compose with ad-hoc generator expressions, and everything in [`itertools`](https://docs.python.org/3.7/library/itertools.html) and packages like [`more_itertools`](https://more-itertools.readthedocs.io/en/latest/api.html).

matt-gardner:
Except if the code looks like that, you have to repeat it for training and for validation.  We want one function to call to perform all of the operations that you want.  It should look more like this:

```python
instances = reader.read(train_file)
batches = train_transforms.apply(instances)

dev_instances = reader.read(dev_file)
dev_batches = dev_transforms.apply(instances)
```

How do we construct the `train_transforms` object / method / whatever it is?  If it's literally a python function that the user writes, then it means that in order to change it we need to change python code, which makes it harder to keep track of controlled experiments.  So, yes, having configurable dependencies is a fundamental design principle of the library.  That's different from "config files" necessarily - you can configure things in more ways than with our `FromParams` stuff.  But you're right that it's our approach to dependency injection, requiring our objects to take explicit dependencies when we expect there to be some variation in behavior, that's driving the design here.

dirkgr:
If I don't worry about driving everything with configuration, I would do this:
```Python
def make_batches(input_file: str):
   instances = reader.read()
   instances = transforms.index(instances)
   batches = transforms.batch(instances, batch_size=32)
   return transforms.shuffle(batches)

batches = make_batches(train_file)
dev_batches = make_batches(dev_file)
```

If I do worry about it, this case becomes a bit weaker, because of all the functional machinery we have to introduce to make it work:
```Python
def compose_generators(left: Generator, right: Generator) -> Generator:
   def inner(items: Iterable):
      return right(left(items))
   return inner

def curry(fn: Callable, **default_kwargs):
    def inner(*args, **kwargs):
        return fn(*args, **{**default_kwargs, **kwargs})
    return inner

def identity_generator(items: Iterable):
    yield from items

def make_transform(params):
    g = identity_generator
    for transform_params in params["transforms"]:
        step = find_transform_by_name(transform_params["type"]) # Using Registrable or something?
        step = curry(step, **transform_params)
        g = compose_generators(g, step)
    return g

transform = make_transform(config_params)
instances = reader.read(train_file)
batches = transforms.apply(instances)
```
I didn't actually run this, but the intent should be clear.

If we do this, all the functionality itself is still contained in neat, readable generators that can be used as an API without surprises, like any other function. The downside is all the functional magic required to make it work. Some of this functional magic exists in Python standard libraries already, but not all. I could not find a `compose()` function.

DeNeutoy:
I don't actually think this approach works (I tested it and it was not faster than what we have currently), but not for the reasons you are describing. 

@dirkgr your points:

1. The design was inspired by `torchvision.transforms`, which use object oriented generators which are composed. 

2. The gross stuff in this PR is trying to write transforms which are generic over an input and output type _and_ trying to make them work with both batches and individual instances. This is likely what you are talking about by "heavy handed".

Your pseudocode is exactly what `transforms.Compose` in this PR does.


FWIW I think I did this the wrong way, because I tried to move away from pytorch's `Dataset` and `Sampler` abstraction. This means I had to hack around the batching that `DataLoader` does and do it myself (we have batching logic which is not stateless, which made it hard to use the DataLoader's `collate_fn`). The reason I did this is because Matt has a fairly strong opinion on having `Datasets` be a class which reads data, not which represents data itself, which made this hard to deal with. 

It's possible that we can fix this current implementation we have here, but I don't know how to do it. Everyone has very strong opinions about this PR and it's pretty unrewarding to work on. I'm all ears to other solutions.

matt-gardner:
I don't remember what you mean by strong opinions on `Datasets` as a class that reads data instead of representing data itself.  Did you mean `DatasetReader`?  Can you remind me of the alternative you suggested?

matt-gardner:
My opinion: I have never looked at pytorch's native data processing pipelines, so I have absolutely no insight into what would be best to integrate what we have with what pytorch provides.  I'm totally open to whatever others think is the best way to go to get the benefit of what's already in pytorch.  There are a few design considerations that I could give feedback on when I see a proposal (like, for example, being able to specify a processing / reading pipeline once and using the same one on multiple files / datasets), but probably any concerns I have could be figured out.

scarecrow1123:
In the case of multi-GPU training, sticking with PyTorch's native `DataLoader` and `Sampler` semantics would bring in huge benefits in terms of splitting the dataset across processes (refer [DistributedSampler](https://pytorch.org/docs/stable/data.html#torch.utils.data.distributed.DistributedSampler)). With the current `DatasetReader` abstraction without any `DataLoader` stuff, users need to have checks in `_read()` to filter out data so that each trainer worker only gets a subset without duplicates(as in [here](https://github.com/scarecrow1123/allennlp-distributed-training/blob/master/distributed-training/data/dataset_readers/snli_dist.py#L52)) which should also be documented somewhere.

DeNeutoy:
@scarecrow1123 - I actually investigated the possibility of this yesterday.

Unfortunately, pytorch don't support samplers for `IterableDatasets`, which is kind of a requirement. One possibility is to rethink how we've done lazy datasets, which would alleviate this problem. Actually, maybe that's the right idea. Let me think about that for a bit.

matt-gardner:
It seems totally reasonable to me to provide new functionality that is dramatically more efficient, but drops support for laziness, with a promise of trying to add it back in later.

sai-prasanna:
@DeNeutoy Take a look into pytext for inspiration, if you haven't. They have pretty much similar structure as allennlp. But they seem to use torch datasets. I am looking through their code currently. 

On a side note, if you all have any ideas of easy export to torch script, I will be more than happy to work with you and contribute.

DeNeutoy:
Ok, closing in favour of radically new approach - thanks for feedback!

PR #3390, opened 2019-10-23
Changes to `train_model` for distributed training support

scarecrow1123:
This is the second PR after #3372 to bring in distributed training support. Following are the high level changes done:

* Extra params to `TrainerBase`, `Trainer` and `CallbackTrainer` classes to include rank and world_size options
* In a distributed setup, the plan is to run training in sub processes by using `multiprocessing.spawn`. Hence parts of `train_model` function in `train.py` is isolated to a new `_train_worker` function which is invoked as a process
* In this PR, changes for distributed training are still not in. Only the isolation of the above mentioned function is done, to mostly test its correctness when run with existing tests/experiments

~~P.S: Since this is a dependent PR, the changes done for the previous PR is being shown until the older one is merged. I'd be happy to know if there is a better way to do this :)~~

@DeNeutoy Once this reviewed, the next PR candidate would be the actual distributed training related code.

### DIFF
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -44,11 +44,21 @@ which to write the results.
 import argparse
 import logging
 import os
+from typing import Optional
+
+import torch
+import torch.distributed as dist
+import torch.multiprocessing as mp
 
 from allennlp.commands.subcommand import Subcommand
-from allennlp.common.checks import check_for_gpu
 from allennlp.common import Params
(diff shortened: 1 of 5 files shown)


DeNeutoy:
@scarecrow1123 could you rebase this please? Then we'll review 👍

scarecrow1123:
@DeNeutoy Rebase is done. Please proceed with the review.

DeNeutoy:
Thanks!

PR #3393, opened 2019-10-24
Fix several resume daemon bugs.

brendan-ai2:
- Execute w/ `python3 resume_daemon.py` in case the exec bit isn't set.
- Create `~/.allennlp` if it wasn't already in case `allennlp` hadn't been run locally.
- Handle empty crontab case where `crontab -l` has a non-zero exit code.


brendan-ai2:
@sanjayss34, could you please take a look? Thanks for finding these issues!

brendan-ai2:
Thanks!

PR #3396, opened 2019-10-24
Tighten pin to avoid only broken binaries and allow 1.3.0.

brendan-ai2:
Resolution of https://github.com/allenai/allennlp/issues/3367. Pytorch has fixed their binaries for 1.3.0. Now we'll use 1.3.0.post2.

I've tested this locally on macOS 10.13.6 and both `pip install -r requirements.txt` and `pip install --editable allennlp` fix broken installations.


brendan-ai2:
Thanks!

PR #3400, opened 2019-10-27
Language modeling dataset loader and model are inconsistent

nicola-decao:
I propose to modify the data structure of `allennlp.data.dataset_readers.LanguageModelingReader` since it is inconsistent with the input expected by `allennlp.models.LanguageModel`.

With this new data structure, one can train a `LanguageModel` model using the `LanguageModelingReader` directly when before it would have raised an error.

### DIFF
--- a/allennlp/common/params.py
+++ b/allennlp/common/params.py
@@ -21,12 +21,16 @@ try:
 except ImportError:
 
     def evaluate_file(filename: str, **_kwargs) -> str:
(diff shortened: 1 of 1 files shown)


nicola-decao:
Hello, all checks are ok except the 66% code diff with 90% target of `allennlp.data.dataset_readers.LanguageModelingReader`. The problem is we need those edits if we want to make it compatible with `allennlp.models.LanguageModel`.

Issue #3401, opened 2019-10-27
Issue in overrides 


dDua:
**Describe the bug**
The same code I have been using for past week started giving the following error today 

Traceback (most recent call last):
  File "/opt/conda/lib/python3.6/site-packages/overrides/overrides.py", line 165, in _get_base_class
    obj = namespace[components[0]]
KeyError: 'str'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/conda/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/opt/conda/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/opt/conda/lib/python3.6/site-packages/allennlp/run.py", line 14, in <module>
    from allennlp.commands import main  # noqa
  File "/opt/conda/lib/python3.6/site-packages/allennlp/commands/__init__.py", line 8, in <module>
    from allennlp.commands.configure import Configure
  File "/opt/conda/lib/python3.6/site-packages/allennlp/commands/configure.py", line 26, in <module>
    from allennlp.service.config_explorer import make_app
  File "/opt/conda/lib/python3.6/site-packages/allennlp/service/config_explorer.py", line 24, in <module>
    from allennlp.common.configuration import configure, choices
  File "/opt/conda/lib/python3.6/site-packages/allennlp/common/configuration.py", line 27, in <module>
    from allennlp.data.dataset_readers import DatasetReader
  File "/opt/conda/lib/python3.6/site-packages/allennlp/data/__init__.py", line 1, in <module>
    from allennlp.data.dataset_readers.dataset_reader import DatasetReader
  File "/opt/conda/lib/python3.6/site-packages/allennlp/data/dataset_readers/__init__.py", line 10, in <module>
    from allennlp.data.dataset_readers.ccgbank import CcgBankDatasetReader
  File "/opt/conda/lib/python3.6/site-packages/allennlp/data/dataset_readers/ccgbank.py", line 9, in <module>
    from allennlp.data.dataset_readers.dataset_reader import DatasetReader
  File "/opt/conda/lib/python3.6/site-packages/allennlp/data/dataset_readers/dataset_reader.py", line 8, in <module>
    from allennlp.data.instance import Instance
  File "/opt/conda/lib/python3.6/site-packages/allennlp/data/instance.py", line 3, in <module>
    from allennlp.data.fields.field import DataArray, Field
[...]

PR #3402, opened 2019-10-27
adding version to overrides packages

dDua:
Adding the version number for overrides package in requirements to avoid the error reported in issue #3401


nelson-liu:
Thanks!

PR #3404, opened 2019-10-28
Refactor out fine-tune command

bryant1410:
Refactor the subcommands main logic after the comments in #3385. @matt-gardner can you give me some feedback before I go on?

I feel at this point that `fine-tune` and `dry-run` should be options of `train`. The former is basically specifying a model to start from and possibly extending a vocabulary. I realized it when changing `fine_tuning.py` little by little and then arrived at almost exactly the same code. The latter is like Trainer but not doing `trainer.train()` and reporting some dataset stats. I also feel like the vocab creation from all these methods plus from `find-lr` and `make-vocab` should be refactored out.

So, I made `fine_tuning.py` use `trainer.py`, but up to what degree do you think it's okay? 1) Delete it completely? 2) Keep the subcommand but use `trainer.py` at all points? 3) Re-use `trainer.train_model` only (as the code is now). 4) Don't re-use it? What about `dry-run`? For `fine-tune`, the main logic for vocabulary and model instantiation is now in `TrainerPieces.from_params`, which was changed to be more generic to support loading an existing model.

Note there was (and still is) a lot of repeated code here and there. Some places where `create_serialization_dir` and some places where it isn't but feels like it should have been. Some places that have code to dump the metrics and some places that call `dump_metrics`. Some that `check_for_gpu` and some places that don't but it feels like they should. Multiple places that check the trainer type, have conditional logic based on it and then evaluate on test conditionally. And so on. This makes me think subcommands need even more refactoring overall.


allennlp/common/tee_logger.py
```diff
@@ -61,3 +61,7 @@ def isatty(self):
     def cleanup(self) -> TextIO:
         self.log.close()
         return self.terminal
+
+    @staticmethod
+    def getvalue() -> str:
```
bryant1410: This is just something I did so some tests not fail in PyCharm.

bryant1410:
And I hope this doesn't give a lot of conflicts with [torch-distributed branch](allenai/allennlp/tree/torch-distributed).

bryant1410:
Next changes also would include puting `TrainerPieces` inside `Trainer.from_params` (as the rest of the trainers) and make the `evaluate_on_test` only for Trainer, to be able to just do `TrainerBase.from_params` regardless of it's the default or not.

allennlp/commands/dry_run.py
```diff
@@ -142,10 +142,4 @@ def dry_run_from_params(params: Params, serialization_dir: str) -> None:
...
             parameter.requires_grad_(False)
 
-    frozen_parameter_names, tunable_parameter_names = get_frozen_and_tunable_parameter_names(model)
-    logger.info("Following parameters are Frozen  (without gradient):")
-    for name in frozen_parameter_names:
-        logger.info(name)
-    logger.info("Following parameters are Tunable (with gradient):")
-    for name in tunable_parameter_names:
-        logger.info(name)
+    log_and_get_frozen_and_tunable_parameter_names(model)
```
matt-gardner: Instead of changing this name, I'd add a new function that is something like `log_parameter_tuning`, that calls the existing function.

allennlp/common/tee_logger.py
matt-gardner: What failed?

allennlp/tests/commands/fine_tune_test.py
```diff
@@ -47,7 +47,7 @@ def test_fine_tune_does_not_expand_vocab_by_default(self):
         model = load_archive(self.model_archive).model
 
         # By default, no vocab expansion.
-        fine_tune_model(model, params, self.serialization_dir)
+        train_model(params, self.serialization_dir, model=model)
```
matt-gardner: We shouldn't be calling `train_model` here directly in a `fine_tune_test.py` file.  If there are tests for functionality that got moved to `train.py`, those tests should be moved into `train_test.py`.  In general, the tests in a file should only call methods in the corresponding implementation (so, `fine_tune_model_from_file_paths`).

allennlp/training/trainer_pieces.py
```diff
@@ -41,44 +41,76 @@ def from_params(
...
-                    instance
-                    for key, dataset in all_datasets.items()
-                    if key in datasets_for_vocab_creation
-                    for instance in dataset
-                ),
+            vocab = None
+
+        vocabulary_path = os.path.join(serialization_dir, "vocabulary")
+
+        if not vocab or extend_vocab:
```
matt-gardner: This whole block looks like it should be a separate method, with its own tests.

allennlp/training/trainer_pieces.py
```diff
@@ -41,44 +41,76 @@ def from_params(
...
 
-        logger.info(
-            "From dataset instances, %s will be considered for vocabulary creation.",
-            ", ".join(datasets_for_vocab_creation),
-        )
+        if model:
+            if params.pop("model", None):
+                logger.warning(
+                    "You passed parameters for the model in your configuration file, but we "
+                    "are ignoring them, using instead the model parameters in the archive."
```
matt-gardner: At this point we don't have an "archive", so this wording is a bit confusing.

allennlp/training/trainer_pieces.py
```diff
@@ -41,44 +41,76 @@ def from_params(
...
+                )
+
+            if vocabulary_params.get("directory_path", None):
+                logger.warning(
+                    "You passed `directory_path` in parameters for the vocabulary in "
+                    "your configuration file, but it will be ignored. "
```
matt-gardner: Why?  `s/ignored. "/ignored, because we already have a model with a vocabulary."`

allennlp/common/tee_logger.py
bryant1410: PyCharm has its own ways of running the tests, and it seems that it supposes there exists a `getvalue` method in `sys.stdout`:

```
  File "/snap/pycharm-professional/167/helpers/pycharm/_jb_unittest_runner.py", line 35, in <module>
    main(argv=args, module=None, testRunner=unittestpy.TeamcityTestRunner, buffer=not JB_DISABLE_BUFFERING)
  File "/home/sacastro/.pyenv/versions/3.6.7/lib/python3.6/unittest/main.py", line 95, in __init__
    self.runTests()
  File "/home/sacastro/.pyenv/versions/3.6.7/lib/python3.6/unittest/main.py", line 256, in runTests
    self.result = testRunner.run(self.test)
  File "/snap/pycharm-professional/167/helpers/pycharm/teamcity/unittestpy.py", line 308, in run
    return super(TeamcityTestRunner, self).run(test)
  File "/home/sacastro/.pyenv/versions/3.6.7/lib/python3.6/unittest/runner.py", line 176, in run
    test(result)
  File "/home/sacastro/.pyenv/versions/3.6.7/lib/python3.6/unittest/suite.py", line 84, in __call__
    return self.run(*args, **kwds)
  File "/home/sacastro/.pyenv/versions/3.6.7/lib/python3.6/unittest/suite.py", line 122, in run
    test(result)
  File "/home/sacastro/.pyenv/versions/3.6.7/lib/python3.6/unittest/suite.py", line 84, in __call__
    return self.run(*args, **kwds)
  File "/home/sacastro/.pyenv/versions/3.6.7/lib/python3.6/unittest/suite.py", line 122, in run
    test(result)
  File "/home/sacastro/.pyenv/versions/3.6.7/lib/python3.6/unittest/case.py", line 653, in __call__
    return self.run(*args, **kwds)
  File "/home/sacastro/.pyenv/versions/3.6.7/lib/python3.6/unittest/case.py", line 624, in run
    result.stopTest(self)
  File "/snap/pycharm-professional/167/helpers/pycharm/teamcity/unittestpy.py", line 261, in stopTest
    output = sys.stdout.getvalue()
AttributeError: 'TeeLogger' object has no attribute 'getvalue'
```

bryant1410:
Btw, shouldn't black's and flake8's max line length be the same?

https://github.com/allenai/allennlp/blob/f3083c8fb9150f07e3ca98bb3ea9368a081df028/pyproject.toml#L4-L5

https://github.com/allenai/allennlp/blob/f3083c8fb9150f07e3ca98bb3ea9368a081df028/.flake8#L1-L2

allennlp/training/trainer_pieces.py
bryant1410: What about now?

allennlp/common/tee_logger.py
matt-gardner: I would feel more comfortable with this change if I knew that the return value was correct.  Can you just keep this change local to your dev environment?

allennlp/tests/commands/train_test.py
```diff
@@ -318,3 +318,143 @@ def test_vocab_extended_model_with_transferred_embedder_is_loadable(self):
         # Check that such a dumped model is loadable
         # self.serialization_dir = self.TEST_DIR / 'fine_tune'
         load_archive(str(self.TEST_DIR / "train" / "model.tar.gz"))
+
+
+class TestFineTune(AllenNlpTestCase):
```
matt-gardner: Why is this a separate class?  Just have these be additional tests on the main class, unless this is because you really want different setup for these tests.  If that's true, then add a comment saying so.

allennlp/commands/print_results.py
```diff
@@ -85,15 +85,15 @@ def print_results_from_args(args: argparse.Namespace):
 
     results_dict = {}
     for root, _, files in os.walk(path):
-
         if metrics_name in files:
             full_name = os.path.join(root, metrics_name)
-            metrics = json.load(open(full_name))
+            with open(full_name) as file:
```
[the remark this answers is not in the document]
bryant1410: `file` is not a reserved word in Python 3 (it is in Python 2).

I was likely revising the commands I saw this file not being closed and changed it, which is usually a recommended change.

allennlp/common/tee_logger.py
bryant1410: Yes, that's the idea. Maybe I shouldn't have committed it.

bryant1410:
> Thanks, this looks pretty close to ready to merge.

Yeah, after the pending additions/changes in testing, I also feel like there's something to merge, which is re-factoring `fine-tune`. However, note there's still a bunch of repeated code between subcommands that I believe should be addressed at some point (other PRs).

allennlp/commands/print_results.py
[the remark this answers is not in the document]
matt-gardner: Fair enough, the changes seem fine.  Except I would still avoid the use of `file`, as it doesn't interact well with python tooling at this point, even though it's no longer reserved.

matt-gardner:
Smaller PRs are basically always preferred when possible, as they are easier to review and reason about.  It's fine to leave the rest of the refactoring work to future PRs.

bryant1410:
> Smaller PRs are basically always preferred when possible, as they are easier to review and reason about. It's fine to leave the rest of the refactoring work to future PRs.

Yeah, I agree!

allennlp/commands/print_results.py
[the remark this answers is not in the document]
bryant1410: I can change the name.

But can you point me out where python tooling gives a problem? Is it just GitHub? [The code highlighting tool they use just treats Python 2 and 3 the same way](https://github.com/MagicStack/MagicPython/blob/master/grammars/MagicPython.cson#L1785), they have it as a deprecated one though, but I don't think they can do much without complicating stuff. I haven't seen other tools affected by it.

bryant1410:
I'm gonna shoot for next week to do the test additions and changes (and other minor changes as necessary).

bryant1410:
> Btw, shouldn't black's and flake8's max line length be the same?
> 
> https://github.com/allenai/allennlp/blob/f3083c8fb9150f07e3ca98bb3ea9368a081df028/pyproject.toml#L4-L5
> 
> https://github.com/allenai/allennlp/blob/f3083c8fb9150f07e3ca98bb3ea9368a081df028/.flake8#L1-L2

And what about this?

matt-gardner:
Flake is intentionally more permissive than black, for awkward situations that black can't fix, but we don't want to count as errors.  For tooling, I was thinking about syntax highlighting in general.  If you have a syntax highlighter like github's in your IDE, it's confusing to see `file`.

matt-gardner:
@bryant1410, sorry to be slow.  Can you remind me of the state of this PR? Is it ready to merge?

bryant1410:
You're good, the ball is on my side. My bad, I've been busy. I have to make
some tests and it'll be ready to go. Hopefully I'll do it within the next
week or so.

bryant1410:
Regarding the last commit/fix (695b494), is it okay? It fixes a test but isn't it the expected behavior to fail to recover a vocabulary from a path if `recover==True` and the path doesn't exist?

matt-gardner:
On that test, I changed how vocab creation works in #3550, which is about to be merged.  Can you update this PR and see if there's a better solution now that there are better ways of constructing the vocabulary?

bryant1410:
> On that test, I changed how vocab creation works in #3550, which is about to be merged. Can you update this PR and see if there's a better solution now that there are better ways of constructing the vocabulary?

Update. Now that I look again, maybe it's okay to both check `recover` and if the path exists, as the recovery is not only for the vocabulary but also for the training. So everything should be good now.

bryant1410:
Two more things.

Now I see the optional arguments `cache-directory`, `cache-prefix`, and `node-rank` are in `train` but not in `fine-tune`, which makes me have stronger thoughts on combining them (e.g., fine-tuning being an option of `train` somehow). Thoughts?

We should warn somewhere that `fine-tune` only works for the default trainer.

bryant1410:
And now that I'm thinking again about it, `dry-run` should be also refactored out. The only different part from the rest is the one about the dataset and vocabulary stats:

https://github.com/allenai/allennlp/blob/56a3966053b6ebc47b23942b8ccbd9cc5193eecd/allennlp/commands/dry_run.py#L130-L133

Moreover, a dry run is supposed to be like a normal run but without actually running it, so it makes sense to me that's integrated with `train`.

matt-gardner:
Can you have a look at #3627 and see if you think that implementation would change anything you're doing here?  Sorry to keep throwing these things at you, a bunch of parts of the library are in flux at the moment as we're trying to improve thing for a 1.0 release.

bryant1410:
> Can you have a look at #3627 and see if you think that implementation would change anything you're doing here? Sorry to keep throwing these things at you, a bunch of parts of the library are in flux at the moment as we're trying to improve thing for a 1.0 release.

It's okay, I understand. I was actually looking right now.

bryant1410:
@matt-gardner I think the changes in this PR are quite orthogonal to those in #3627. I mean, they are useful (because they remove redundancy) and, even though there are related parts such as `TrainerPieces`, the changes here are somewhat different. Maybe part of the changed code (even if unchanged) is gonna be discarded in the future because of other changes, but parts such as those of `fine_tune.py` I believe are gonna stay. So I'd continue reviewing this PR regardless of #3627.

Issue #3407, opened 2019-10-29
AutoRegressiveSeqDecoder has no decode


dasguptar:
I came across a missing method `decode` in AutoRegressiveSeqDecoder when trying to pass in a token_based_metric. The culprit seems to be [this line](https://github.com/allenai/allennlp/blob/master/allennlp/modules/seq2seq_decoders/auto_regressive_seq_decoder.py#L429).

Probably replacing the call to `decode` with a call to `post_process` would work, but I am not 100% sure. Can raise a PR if that is the fix.

schmmd:
@brendan-ai2 does @dasguptar's proposal sound like the right fix?

brendan-ai2:
@dasguptar, thanks for catching that! I do believe that's the correct fix and we'd be happy to take a PR. If you do create one, could you please add a small test case? Depending on whether it's easier to trigger this failure as a unit test or not I'd create an `auto_regressive_seq_decoder_test.py` or use the existing https://github.com/allenai/allennlp/blob/master/allennlp/tests/models/encoder_decoders/composed_seq2seq_test.py. Thank you!

dasguptar:
Hi @brendan-ai2, I've submitted PR #3411.
Could not add test cases, explained more in the PR.

DeNeutoy:
fixed by #3433

Issue #2919, opened 2019-06-03
Copynet_seq2seq does not correctly call SequenceAccuracy metric


tomsherborne:
The implementation of Copynet_seq2seq appears to incorrectly call the SequenceAccuracy metric as a tensor_based_metric. 

SequenceAccuracy expects at least the tensors `predictions (batch_size, k, sequence_length)`, `gold_labels (batch_size, sequence_length)` and copynet_seq2seq passes 
tensors of size `(batch_size, max_predicted_sequence_length)` and `(batch_size, target_sequence_length)`. The Metric expects inputs to be the same size in dimension `-1` and the predictions for all `k` beams. 

**To Reproduce**
Config pasted below. I'm working with GeoQuery data, but this also applies to other .TSV data files.

**Expected behavior**
Model does not crash during validation reporting.

**System (please complete the following information):**
 - OS: Ubuntu 18.04
 - Python version: Python 3.6.3 :: Anaconda custom (64-bit)
 - AllenNLP version: v0.8.3
 - PyTorch version: 1.0.1.post2

**Additional context**
I am happy to work on a fix/PR if this is worth fixing, I'm just not sure where the fix should go. I'm guessing there needs to be a modification to copynet_seq2seq to truncate top_k_predictions to size (batch_size, beam_size, target_sequence_length) and possibly some other adjustments.

Config
```
{
  "dataset_reader": {
    "type": "copynet_seq2seq",
    "source_tokenizer": {
      "type": "word",
      "word_splitter": {
        "type": "just_spaces"
      }
    },
    "target_tokenizer": {
      "type": "word",
      "word_splitter": {
        "type": "just_spaces"
      }
    },
    "source_token_indexers": {
      "tokens": {
        "type": "single_id",
        "namespace": "tokens"
      }
    },
    "target_namespace": "target_tokens"
  },
  "train_data_path": "data/geoquery/funql/geo.en.train",
  "validation_data_path": "data/geoquery/funql/geo.en.val",
  "vocabulary": {
    "directory_path": "data/geoquery/funql/en_vocab_copy"
  },
  "model": {
    "type": "copynet_seq2seq",
    "source_embedder": {
      "type": "basic",
      "token_embedders": {
        "tokens": {
          "type": "embedding",
          "embedding_dim": 150,
          "trainable": true,
          "vocab_namespace": "tokens"
        }
      }
    },
    "encoder": {
      "type": "lstm",
      "input_size": 150,
      "hidden_size": 150,
[...]

PR #3411, opened 2019-10-30
Use correct method in auto_regressive_seq_decoder when computing token based metric

dasguptar:
Fix for #3407.

There are currently no existing tests for the whole `auto_regressive_seq_decoder` module, and I was not sure of what fixtures to create to test the module.

To test just this PR, I needed to use a `token_based_metric`. From a cursory glance of the available metrics, `BLEU` seemed to be a good choice. However, `BLEU` works on tensors, whereas this class is passing `List[List[str]]`.

As noted in #2919, even `tensor_based_metric` is not being called correctly, as `SequenceAccuracy` expects a different format. Maybe the metrics can be removed from this class?

### DIFF
--- a/allennlp/modules/seq2seq_decoders/auto_regressive_seq_decoder.py
+++ b/allennlp/modules/seq2seq_decoders/auto_regressive_seq_decoder.py
@@ -426,7 +426,7 @@ class AutoRegressiveSeqDecoder(SeqDecoder):
                     )
 
                 if self._token_based_metric is not None:
-                    output_dict = self.decode(output_dict)
+                    output_dict = self.post_process(output_dict)
                     predicted_tokens = output_dict["predicted_tokens"]
 
                     self._token_based_metric(  # type: ignore


brendan-ai2:
@sai-prasanna, do you have any suggestions for how to add tests for this module? I was a bit surprised we missed that previously. My only quick idea is to add an integration level test to https://github.com/allenai/allennlp/blob/master/allennlp/tests/models/encoder_decoders/composed_seq2seq_test.py that covers this specific case.

@dasguptar, I don't think the choice of metric is terribly important for verifying this change. I would suggest just making a simple fake one in the test that verifies the shape of the passed data and returns a constant.

As for #2919 and metrics generally, it's not that `tensor_based_metric` is broken for `SequenceAccuracy`, it's that different metrics have different interfaces and there's not an obvious way to satisfy all the metrics simultaneously. Perhaps we should encourage people to instead subclass when they need to add a metric.

sai-prasanna:
@brendan-ai2 My bad, I think I missed the tests relying on coverage.
Will look into it and add the necessary tests by the next weekend.

dasguptar:
Hi @sai-prasanna , @brendan-ai2 

I can help write tests for this class if you can guide me a bit on what the fixtures should be like.

sai-prasanna:
@dasguptar From the top of my head, you can just create a AutoRegressiveSeqDecoder object programmatically, passing an implementation of DecoderNet (LSTM or SelfAttention) to it. And then write test cases for all the methods in the class.

nicola-decao:
Look here https://github.com/allenai/allennlp/pull/3433. Your catc is correct but we need to also fix the target token variable.

nicola-decao:
https://github.com/allenai/allennlp/pull/3433 is merged in master so this pull request should be closed.

brendan-ai2:
Thanks, all. Closing.

PR #3417, opened 2019-11-01
Fix char_span_to_token_span breaking from under tokenization

mbugert:
This fixes an issue in `allennlp.data.dataset_readers.reading_comprehension.util.char_span_to_token_span` and adds unit tests.
If the last two tokens in the sentence were under tokenized (1 token instead of 2) and the character span pointed to the end of this under tokenized token, the method would break with an IndexError.

The method also breaks if the end offset of `character_span` exceeds the sentence length. I feel like this would warrant a ValueError, but the rest of the method is written in a way that it fails silently. How would you want to handle these cases?


mbugert:
@DeNeutoy I decided to fix the IndexError as well and added corresponding unit tests. Hope that's okay.

PR #3441, opened 2019-11-09
Adding source_add_end_token to Seq2SeqDatasetReader

nicola-decao:
`Seq2SeqDatasetReader` has `source_add_start_token` parameter to control whether to add a start token to the source. I added `source_add_end_token` to control whether to add the end token to the source. The default is `True` so it is compatible with previous versions.

### DIFF
--- a/allennlp/data/dataset_readers/seq2seq.py
+++ b/allennlp/data/dataset_readers/seq2seq.py
@@ -46,6 +46,8 @@ class Seq2SeqDatasetReader(DatasetReader):
         ``source_token_indexers``.
     source_add_start_token : bool, (optional, default=True)
         Whether or not to add `START_SYMBOL` to the beginning of the source sequence.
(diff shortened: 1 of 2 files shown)


allennlp/data/dataset_readers/seq2seq.py
```diff
@@ -46,6 +46,8 @@ class Seq2SeqDatasetReader(DatasetReader):
         ``source_token_indexers``.
     source_add_start_token : bool, (optional, default=True)
         Whether or not to add `START_SYMBOL` to the beginning of the source sequence.
+    source_add_end_token : bool, (optional, default=True)
+        Whether or not to add `END_SYMBOL` to the beginning of the source sequence.
```
matt-gardner: `s/beginning/end/`

matt-gardner:
Thanks!

Issue #3448, opened 2019-11-13
Upgrade pytorch_transformers (v1) to transformers (v2)


maksym-del:
Latest [transformers](https://github.com/huggingface/transformers/releases) package is 2.1.1 already.

It was renamed to just "transformers" which affects imports in the code.

It might be worth upgrading before the upcoming allennlp release.

PR #3449, opened 2019-11-13
pytorch_transformers -> transformers

maksym-del:
Fixes https://github.com/allenai/allennlp/issues/3448.

Breaking changes that were introduced in the new "transofrmers" release should not affect existing allennlp codebase.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -53,6 +53,7 @@ word2number>=1.1
 # To use the BERT model
 pytorch-pretrained-bert>=0.6.0
 pytorch-transformers==1.1.0
+transformers==2.1.1
 
 # For caching processed data
 jsonpickle
--- a/requirements.txt
+++ b/requirements.txt
@@ -52,7 +52,6 @@ word2number>=1.1
 
 # To use the BERT model
 pytorch-pretrained-bert>=0.6.0
-pytorch-transformers==1.1.0
 transformers==2.1.1
 
 # For caching processed data
--- a/setup.py
+++ b/setup.py
@@ -132,6 +132,7 @@ setup(
         "word2number>=1.1",
         "pytorch-pretrained-bert>=0.6.0",
         "pytorch-transformers==1.1.0",
+        "transformers=2.1.1",
         "jsonpickle",
     ],
     entry_points={"console_scripts": ["allennlp=allennlp.run:run"]},
--- a/setup.py
+++ b/setup.py
@@ -132,7 +132,7 @@ setup(
         "word2number>=1.1",
         "pytorch-pretrained-bert>=0.6.0",
         "pytorch-transformers==1.1.0",
-        "transformers=2.1.1",
+        "transformers==2.1.1",
         "jsonpickle",
     ],
     entry_points={"console_scripts": ["allennlp=allennlp.run:run"]},
--- a/setup.py
+++ b/setup.py
@@ -131,7 +131,6 @@ setup(
         "sqlparse>=0.2.4",
         "word2number>=1.1",
         "pytorch-pretrained-bert>=0.6.0",
-        "pytorch-transformers==1.1.0",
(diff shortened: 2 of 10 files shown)


requirements.txt
```diff
@@ -53,6 +53,7 @@ word2number>=1.1
 # To use the BERT model
 pytorch-pretrained-bert>=0.6.0
 pytorch-transformers==1.1.0
```
matt-gardner: Can't you remove this one?

setup.py
```diff
@@ -132,6 +132,7 @@
         "word2number>=1.1",
         "pytorch-pretrained-bert>=0.6.0",
         "pytorch-transformers==1.1.0",
```
matt-gardner: And this.

nicola-decao:
This is another place:
https://github.com/allenai/allennlp/blob/5238561412d1f63fa47512b7d6726dee9ed5ae88/allennlp/modules/token_embedders/bert_token_embedder.py#L16
However, this needs a more careful edit since if you just change that it brakes. Indeed, `BertModel.forward` works differently in the new version.

maksym-del:
Yes, ``pytorch_pretrained_bert`` -> ``transformers`` switch is harder then ``pytorch_transformers``->``transformers``. Existing bert_srl model depends on it. 

Let this PR be just ``pytorch_transformers``->``transformers`` migration and upgrading from ``pytorch_pretrained_bert`` can have a separate PR.

matt-gardner:
I agree, I was just commenting on removing the `pytorch-transformers` dependency, not the `pretrained-bert` one.

Evpok:
Regarding this, is the hard-pin to 2.1.1 really necessary?

maksym-del:
@Evpok unless we want to constantly monitor upcoming releases for breaking changes, I believe so.

Evpok:
Well, I respect your choices but surely at some point you will have to monitor this anyway, won't you? So why not just specify a minimum version and only restrict the maximum version if/when tests break?

In any case I am mostly asking because due to python packaging idiosyncrasies, it is basically impossible to override allennlp requirements, so this kind of hard pin is really inconvenient for downstream consumers of both allennlp and transformers. I completely understand if this is inconvenient for you, though.

maksym-del:
@Evpok I see your point regarding specifying a range of versions.

For the transformers library however, there are basically 2 versions: v2.0.0 and v2.1.1. https://github.com/allenai/allennlp/pull/3453 accounts for 2.1.1 and breaks in case of v2.0.0 because of breaking changes. I believe that users should just upgrade to the new version and update their code accordingly (wrt changes in transformers).

Regarding the dependencies issue, running ``pip install transformers=2.1.1`` from allennlp's conda env successfully uninstalls the old version and installs the specified one in my case! You should open a new issue if it is not the case for your setup for some reason.

matt-gardner:
Actually, I think @Evpok is right - we try to be as loose as possible in our bounds, for the reasons that @Evpok brings up.  I thought about this during my review, but ignored it because the previous `pytorch-transformers` dependency was pinned.  But both of them should probably have been `>=`, not `==`.  We get notified very quickly when an updated dependency breaks our build, and it's easy to catch it then and pin it.  @maksym-del, can you change this in #3453?

maksym-del:
@Evpok @matt-gardner sure, I made a commit.

Evpok:
Thanks @matt-gardner, sorry @maksym-del :-( and it was working fine in my setup (which uses virtualenv instead of conda but it doesn't matter), I was just anticipating the future releases as I have been burned by that kind of things in the past. In any case, thank you both for your great work on this library :-)

julien-c:
By the way, if we break your build when releasing a new minor, we also want to know it as fast as possible 🤗 

So feel free to open an issue or send an email if that's the case!

Issue #3356, opened 2019-10-15
Improve getting start, end and intermediate tokens in pretrained_transformer_tokenizer


amarasovic:
I’m classifying a pair of two sentences using RoBERTa following the [huggingface’s input format](https://github.com/huggingface/transformers/blob/80889a0226b8f8022fd9ff65ed6bce71b60ba800/transformers/tokenization_roberta.py#L93):

`start = [Token('<s>')]`
`sep = [Token('</s>')] `
`seq = start + self._tokenizer.tokenize(sentence1) + sep + sep + self._tokenizer.tokenize(sentence2) + sep`

Tokenizer here is `pretrained_transformer`. I assumed that omitting `start_tokens` and `end_tokens` from the config file will prevent the tokenizer from automatically adding them, but they were added because `roberta-base` contains `bert`. Eventually I prevented adding them by passing an empty list for those parameters.

I don’t think `_guess_start_and_end_token_defaults` is necessary and since it is easy to make a mistake (that won’t result in an error), I think it's better to remove it altogether. 

Btw. since `roberta-base` contains `bert`, `[CLS], [SEP]` will be used for RoBERTa, although huggingface now uses `<s>, </s>`. Not sure how important that is.

matt-gardner:
I would have suggested putting those start and end tokens in the tokenizer, as it gives you flexibility to switch to something new later, without hard-coding the separator tokens.  I think we *should* by default try to do the right thing for any particular input text, we just need to fix it so it doesn't do the wrong thing for some models.  I didn't know what the right sequence was for RoBERTa, and I thought you had an error in your code above at first; it would be nice if it just worked without me having to look up what the right thing to do for RoBERTa was.

One issue in what I see in your code is the in-between separator tokens.  We need to handle this better.  There should be a method on the `tokenizer` that takes a list of texts and adds the right separator tokens (including initial and final tokens).  Not sure what the right name is for this.  But then you don't have to hard-code specific values for these separator tokens, which defeats the whole point of the abstraction.

If we do this right, all you should have to do is `seq = self._tokenizer.tokenize_sentences([sentence1, sentence2])`, and you get `[CLS] SENTENCE1 [SEP] SENTENCE2 [SEP]` for BERT, and `<s> SENTENCE1 </s></s> SENTENCE2 </s>` for RoBERTa.

So, concrete things to fix:
1. Add a `tokenize_sentences` (or similarly named) method that adds separator tokens in between sentences.
[...]

Issue #3097, opened 2019-07-25
Rethink vocabulary API


matt-gardner:
The way we do vocab stuff with BERT / other pretrained wordpiece tokenizers is a little crazy and leads to subtle bugs.  We need to get the BERT wordpiece vocab into our `Vocabulary` object so that we can pass it to the model and use it there if necessary.  But if you save a vocab object with this vocabulary, it will give this error on load:

```
  File "tmp/tmp.py", line 30, in predict
    predictor = Predictor.from_path(path, 'masked_lm_predictor')
  File "/Users/mattg/clone/allennlp/allennlp/predictors/predictor.py", line 253, in from_path
    return Predictor.from_archive(load_archive(archive_path, cuda_device=cuda_device), predictor_name)
  File "/Users/mattg/clone/allennlp/allennlp/models/archival.py", line 230, in load_archive
    cuda_device=cuda_device)
  File "/Users/mattg/clone/allennlp/allennlp/models/model.py", line 327, in load
    return cls.by_name(model_type)._load(config, serialization_dir, weights_file, cuda_device)
  File "/Users/mattg/clone/allennlp/allennlp/models/model.py", line 256, in _load
    vocab = Vocabulary.by_name(vocab_choice).from_files(vocab_dir)
  File "/Users/mattg/clone/allennlp/allennlp/data/vocabulary.py", line 324, in from_files
    vocab.set_from_file(filename, is_padded, namespace=namespace)
  File "/Users/mattg/clone/allennlp/allennlp/data/vocabulary.py", line 378, in set_from_file
    assert self._oov_token in self._token_to_index[namespace], "OOV token not found!"
AssertionError: OOV token not found!
```

Somehow our existing BERT models are not actually saving the BERT vocabulary, even though it appears in the vocab object after indexing data.  If I'm understanding things right, this is because we only count vocab items on our first pass through the data, and don't actually index anything when constructing the vocab.  It's only after we've saved the vocab that we index stuff: https://github.com/allenai/allennlp/blob/417a75727d4604e1da0eda461b03b29a73f0bf50/allennlp/training/trainer_pieces.py#L67-L77

This means that we are basically only avoiding the above bug because of an accident in when we are saving the vocab.  We should probably figure out a better solution for this.

matt-gardner:
Oh, actually, this means that the model won't be getting the right vocab sizes if you use our standard training script.  I was hacking my own thing, so I got the right vocab size, but I ran into this OOV error...

[...]

Issue #3224, opened 2019-09-07
How to set the token_type_ids when use  the pytorch-transformers Bert


bestbzw:
**Question**
I want to use the Robert model, but i don't know how to set the token_type_ids.
It seems that only feed the token_ids 

```python
@TokenEmbedder.register("pretrained_transformer")
class PretrainedTransformerEmbedder(TokenEmbedder):
    """
    Uses a pretrained model from ``pytorch-transformers`` as a ``TokenEmbedder``.
    """
    def __init__(self, model_name: str) -> None:
        super().__init__()
        self.transformer_model = AutoModel.from_pretrained(model_name)
        # I'm not sure if this works for all models; open an issue on github if you find a case
        # where it doesn't work.
        self.output_dim = self.transformer_model.config.hidden_size

    @overrides
    def get_output_dim(self):
        return self.output_dim

    def forward(self, token_ids: torch.LongTensor) -> torch.Tensor:  # type: ignore
        # pylint: disable=arguments-differ
        return self.transformer_model(token_ids)[0]
```

joelgrus:
it looks like you provide it here:

https://github.com/huggingface/pytorch-transformers/blob/master/pytorch_transformers/modeling_bert.py#L584

?

matt-gardner:
Yeah, I know that, what I don't know is where the token type ids are computed.

maksym-del:
Hi @matt-gardner, 

looks like it happens here: [segment_ids](https://github.com/huggingface/pytorch-transformers/blob/5a49b793d95042ca1a371cda6f8e2ed489eb319a/examples/utils_glue.py#L456).

mikerossgithub:
In case it is helpful for anyone else, here is the interim fix I am currently using (for Bert only):

Replace this line:
https://github.com/allenai/allennlp/blob/3e0bad4a0a17647bb4f4276d1ee7512918c61753/allennlp/data/token_indexers/pretrained_transformer_indexer.py#L80
With this:
```
result = {index_name: indices}
result.update(_guess_additional_indices(index_name, tokens))
return result
```
And add this function in the same file:
```
def _guess_additional_indices(index_name: str, tokens) -> Dict[str, List[str]]:
    if 'bert' in index_name:
        type_ids = []
        type_id = 0
        for token in tokens:
            type_ids.append(type_id)
            if token == '[SEP]':
                type_id += 1
        return {f'{index_name}-type-ids': type_ids}
    else:
        return {}
```

[...]

PR #3453, opened 2019-11-14
Rework Pretrained Tokenizer and Indexer

maksym-del:
The idea is to both tokenize and index tokens in the tokenizer while adding token ids to the Token data structure. In this setup, PretrainedTokenIndexer only needs to pick up correct token id as well as create allennlp vocabulary from the pretrained one.
 Thus we subclass  SingleIdTokenIndexer. 
To get special tokens handled, we first index the string with huggingface method that adds special token ids as well and then converting indexes back to string tokens (to be used by AllenNLP tokens).
This feels to me the cleanest overall approach.

We also now have token_type_id field added to the token. It can now be computed automatically with huggingface's code. I do not use it in any places yet since it is not used by recent transformers like XLM or RoBERTa. However, future PR can better utilize it.

Fixes https://github.com/allenai/allennlp/issues/3356 and answers the question in https://github.com/allenai/allennlp/issues/3224.


maksym-del:
This one is ready to be reviewed.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -22,36 +24,98 @@ class PretrainedTransformerTokenizer(Tokenizer):
     We take a model name as an input parameter, which we will pass to
     ``AutoTokenizer.from_pretrained``.
 
+    We also add special tokens relative to the pretrained model and trunkate the sequences.
```
matt-gardner: `s/trunkate/truncate/`

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -22,36 +24,98 @@ class PretrainedTransformerTokenizer(Tokenizer):
...
+
+    This tokenizer also indexes tokens and adds the indexes as the ``Token`` fields so that
```
matt-gardner: `s/as the/to the/`, and a period at the end of the sentence.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -22,36 +24,98 @@ class PretrainedTransformerTokenizer(Tokenizer):
...
+
     Parameters
     ----------
     model_name : ``str``
         The name of the pretrained wordpiece tokenizer to use.
     start_tokens : ``List[str]``, optional
-        If given, these tokens will be added to the beginning of every string we tokenize.  We try
-        to be a little bit smart about defaults here - e.g., if your model name contains ``bert``,
-        we by default add ``[CLS]`` at the beginning and ``[SEP]`` at the end.
+        If given, these tokens will be added to the beginning of every string we tokenize.
```
matt-gardner: I'd probably just remove these parameters, as `add_special_tokens` covers the intended use for `start_tokens` and `end_tokens`.  If we really want to keep them, we need to note that these get added _before_ adding special tokens.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -22,36 +24,98 @@ class PretrainedTransformerTokenizer(Tokenizer):
...
+            max_length=self._max_length,
+            stride=self._stride,
+            truncation_strategy=self._truncation_strategy,
+            return_tensors=None,
+        )
+        # token_ids containes final list with ids for both regualr and special tokens
+        token_ids = encoded_tokens["input_ids"]
+        token_type_ids = encoded_tokens["token_type_ids"]
 
+        token_strings = self._tokenizer.convert_ids_to_tokens(token_ids, skip_special_tokens=False)
```
matt-gardner: Seems like this is an extra iteration over the tokens that we don't need, given the iteration you're doing below.  Why not remove this line and do a simple vocab lookup inside the loop?

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -22,36 +24,98 @@ class PretrainedTransformerTokenizer(Tokenizer):
...
     def tokenize(self, text: str) -> List[Token]:
+        """
+        This can also handle sentence pairs. In this case 1st and 2nd sentences in ``text``
+        are expected to be separated with ``ALLENNLP_SENTENCE_PAIR_SEP`` symbol.
+        """
         # TODO(mattg): track character offsets.  Might be too challenging to do it here, given that
         # ``transformers``` is dealing with the whitespace...
-        token_strings = self._start_tokens + self._tokenizer.tokenize(text) + self._end_tokens
-        return [Token(t) for t in token_strings]
+        if ALLENNLP_SENTENCE_PAIR_SEP in text:
```
matt-gardner: I'd be happier if we added a `tokenize_sentences(sentences: List[str]) -> List[Token]` method to `Tokenizer`, instead of adding this special constant.  But I suppose this could do for now...  I'm just wary of adding a recommended public constant that we might want to remove later.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -9,6 +9,8 @@
 
 logger = logging.getLogger(__name__)
 
+ALLENNLP_SENTENCE_PAIR_SEP = "@@SEP_SENTPAIR@@"
```
matt-gardner: As I said in another comment, I'm not thrilled with having this.  But if we do want it, it should not belong here, it should be in something like `data/common.py`.  And I would name it `SENTENCE_SEPARATOR`, instead of what you have.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
maksym-del: Yeah, but the downside of it is that we cannot handle a sentence pair with an inherited ``tokenize`` method in this setup. So I introduced this symbol to tell this method that it gets a sentence pair. Optional argument to the ``tokenize`` might be a cleaner solution, but this breaks API. 

I am not sure however how important it is to handle a sentence pair in the main tokenize method and I followed your suggestion.

Regarding 3+ sentences, to the best of my knowledge, it will be a custom logic that is not available in transformers repo. Probably for another PR.

matt-gardner: Well, either way the `DatasetReader` has to do something differently to handle this.  In your version, you have to manually add the separator token, then have the tokenizer immediately undo this concatenation and process things accordingly.  This also means that we'd have to modify *every* tokenizer to handle this token, because the `DatasetReader` would be adding it, and our other tokenizers would see it if you switched.

In my way, the default behavior is just to concatenate the sentences (probably with a space) and pass it to `tokenize()`.  No other tokenizers break, and so the `DatasetReader` doesn't have to have custom logic that's aware of which `tokenizer` it's using.

maksym-del: I see your point regarding the special token, it totally makes sense to omit it since we might want to change the tokenizer type we are using.

Regarding concatenating sentence pair with space, how will you then decide where to split it in ``tokenize``? Huggingface's ``encode_plus`` method expects sentences to come separated.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -109,13 +82,26 @@ def tokenize(self, text: str) -> List[Token]:
...
 
         tokens = []
-        for token_str, token_id, token_type_id in zip(token_strings, token_ids, token_type_ids):
+        for token_id, token_type_id in zip(token_ids, token_type_ids):
+            token_str = self._tokenizer.convert_ids_to_tokens(token_id, skip_special_tokens=False)
             tokens.append(Token(text=token_str, text_id=token_id, type_id=token_type_id))
+
         return tokens
+
+    def tokenize_sentence_pair(self, sentence_1: str, sentence_2: str):
```
matt-gardner: I was suggesting putting this on the base class, but I suppose putting it here for now is fine, and we can add it higher up later if necessary.  This just means that the dataset reader has to instantiate a `PretrainedTransformerTokenizer`, instead of just a `Tokenizer`, for the types to work out.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
matt-gardner: I was talking about the default implementation in `Tokenizer.tokenize_sentences`, which should be easy.  `PretrainedTransformerTokenizer` would override `tokenize_sentences` with the logic you have here.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
maksym-del: I added the commit representing this approach.

allennlp/data/tokenizers/tokenizer.py
```diff
@@ -47,6 +47,18 @@ def tokenize(self, text: str) -> List[Token]:
         """
         raise NotImplementedError
 
+    def tokenize_sentences(self, sentences: List[str]) -> List[Token]:
```
DeNeutoy: Why does this need to be on the base class? Seems like it should be a private method on the `pretrained_transformer_tokenizer` class. It has confusing behaviour that is only desirable for tokenising text for a language model.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -22,36 +22,101 @@ class PretrainedTransformerTokenizer(Tokenizer):
...
 
-def _guess_start_and_end_token_defaults(model_name: str) -> Tuple[List[str], List[str]]:
-    if "bert" in model_name:
-        return (["[CLS]"], ["[SEP]"])
-    else:
-        return ([], [])
+        return tokens
+
+    @overrides
+    def tokenize_sentences(self, sentences: List[str]) -> List[Token]:
```
DeNeutoy: This seems to be too general, even as a method on this specific class - it's not possible to tokenise more than 2 sentences, so why allow it to be passed more than two sentences?

allennlp/tests/data/token_indexers/pretrained_transformer_indexer_test.py
```diff
@@ -3,24 +3,74 @@
...
 from allennlp.data import Token, Vocabulary
 from allennlp.data.token_indexers import PretrainedTransformerIndexer
+from allennlp.data.tokenizers import PretrainedTransformerTokenizer
 
 
 class TestPretrainedTransformerIndexer(AllenNlpTestCase):
-    def test_as_array_produces_token_sequence(self):
+    def test_as_array_produces_token_sequence_bert(self):
         tokenizer = AutoTokenizer.from_pretrained("bert-base-uncased")
+        allennlp_tokenier = PretrainedTransformerTokenizer("bert-base-uncased")
```
DeNeutoy: allennlp_tokenier should be `allennlp_tokenizer`

allennlp/data/tokenizers/tokenizer.py
maksym-del: If it is a private method and it is used somewhere in DatasetReader, it will mean that this DatasetReader will only work with this type of tokenizer.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
maksym-del: I thought in the future we could implement automatic addition of special tokens in case of 3+ sentences. This might be useful for BERT based reading comprehension or summarization.

allennlp/data/tokenizers/tokenizer.py
maksym-del: However, I am not against making it private again. It will still be useful in many use cases,
and as @matt-gardner pointed out, we can always add it higher up later if necessary.

allennlp/tests/data/token_indexers/pretrained_transformer_indexer_test.py
```diff
@@ -3,24 +3,74 @@
...
+
+        # uncased
```
brendan-ai2: A block level comment like this suggests that the block should be broken out into its own test case. (Especially since the blocks don't appear to rely on much shared state.)

allennlp/tests/data/token_indexers/pretrained_transformer_indexer_test.py
```diff
@@ -3,24 +3,74 @@
...
+        # cased
         indexer = PretrainedTransformerIndexer(model_name="bert-base-cased")
-        tokens = tokenizer.tokenize("AllenNLP is great")
+        string_specials = "[CLS] AllenNLP is great [SEP]"
+        string_no_specials = "AllenNLP is great"
+        tokens = tokenizer.tokenize(string_specials)
+        expected_ids = tokenizer.convert_tokens_to_ids(tokens)
+        # tokens tokenized with our pretrained tokenizer have indices in them
+        allennlp_tokens = allennlp_tokenier.tokenize(string_no_specials)
+        vocab = Vocabulary()
```
brendan-ai2: I think we might want some expectations on the vocabulary produced. (In a separate test case.) Nothing too heavyweight. Perhaps:

1. Sanity check that just verifies that a short simple sentence produces an expected vocabulary.
2. If it's possible to have tokens that aren't indexed by the tokenizer, introduce some. Verify that the indices of the other tokens don't change.

allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
```diff
@@ -17,11 +50,32 @@ def test_splits_cased(self):
             ".",
             "[SEP]",
         ]
-        tokenizer = PretrainedTransformerTokenizer("bert-base-cased")
         tokens = [t.text for t in tokenizer.tokenize(sentence)]
         assert tokens == expected_tokens
 
-    def test_splits_uncased(self):
+        # sentence pair
```
brendan-ai2: Similarly, this would be cleaner as a separate test case.

allennlp/data/tokenizers/tokenizer.py
DeNeutoy: Yes - my point is simply that currently, if you change the tokenizer to one which is not a pretrained lm tokenizer, what happens is pretty undesirable (your two sentences are tokenised as though they are a single sentence). Is this something you can handle in a `DatasetReader` without detecting that you aren't using one of the LM specific tokenizers in the first place? 

We can wait for Matt's opinion, but i'd suggest removing this.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -9,16 +9,17 @@
...
 logger = logging.getLogger(__name__)
 
 
 @TokenIndexer.register("pretrained_transformer")
-class PretrainedTransformerIndexer(TokenIndexer[int]):
+class PretrainedTransformerIndexer(SingleIdTokenIndexer):
     """
-    This :class:`TokenIndexer` uses a tokenizer from the ``transformers`` repository to
-    index tokens.  This ``Indexer`` is only really appropriate to use if you've also used a
+    This ``TokenIndexer`` assumes that Tokens already have their indexes in them.
```
brendan-ai2: Is it possible to assert the indices are present in the tokens already? If not, see my comment about a test to distinguish between tokens that already have indices and those that don't.

allennlp/tests/data/token_indexers/pretrained_transformer_indexer_test.py
maksym-del: Thanks @brendan-ai2 ! Your first sanity check made me discover https://github.com/allenai/allennlp/issues/3456

allennlp/data/token_indexers/pretrained_transformer_indexer.py
maksym-del: Added a check to the indexer

allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
maksym-del: Done

allennlp/tests/data/token_indexers/pretrained_transformer_indexer_test.py
maksym-del: Done

allennlp/data/tokenizers/tokenizer.py
maksym-del: @DeNeutoy do you mean removing it completely or just making it private?

matt-gardner: We talked about this internally at a review meeting today, and I agree with @DeNeutoy's strong counterexample about the generality of this approach.  You can't just change tokenizers to go from BERT-style encoding to something else, you'd have to add new `TextFields` and such.  So it doesn't really make sense to have the method on the super class as I recommended.  Sorry.  We should move it back to a (public) method just on `PretrainedTransformerTokenizer`.  And given that, we probably don't need to handle more than two sentences, and just giving a sentence pair method for now is fine.  Sorry for the extra work that I recommended.

maksym-del: Done!

allennlp/data/token_indexers/pretrained_transformer_indexer.py
maksym-del: Adding this check also spoiled potential bugs in masked_lm and casual_lm readers.

maksym-del:
I addressed feedback and also moved BERT vocab creation inside the ```Vocabulary``` constructor. It fixes https://github.com/allenai/allennlp/issues/3097.

There are 2 tests failing that I do not see to be related to the code I changed... Also, all my local pytest checks pass.

So I believe this PR is ready for review.

allennlp/data/vocabulary.py
```diff
@@ -662,6 +681,32 @@ def extend_from_instances(
             min_pretrained_embeddings=min_pretrained_embeddings,
         )
 
+    def _extend_from_transformers_vocabulary(
```
matt-gardner: Can you move this back inside the token indexer call?  I don't think this is the right fix to #3097.  I'd like to avoid extra parameters that you have to remember to set in the config file; it just makes everything harder to use.  I think there's a different solution that we can come up with, probably just in the timing of when we save the vocab in the trainer code.

allennlp/data/dataset_readers/next_token_lm.py
```diff
@@ -78,5 +85,7 @@ def text_to_instance(
         input_field = TextField(tokens, self._token_indexers)
         fields: Dict[str, Field] = {"tokens": input_field}
         if target:
-            fields["target_ids"] = TextField([Token(target)], self._token_indexers)
+            wordpiece = self._targets_tokenizer.tokenize(target)[0]
```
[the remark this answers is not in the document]
maksym-del: Done

allennlp/data/vocabulary.py
maksym-del: Done

maksym-del:
I addressed the feedback. All the relevant tests pass!

maksym-del:
Thanks!

PR #3432, opened 2019-11-06
Add support for spacy >= 2.2 

tamuhey:
from #3424

- changed requirements: `spacy < 2.2` => `spacy < 2.3`
- modified some tests. They fail due to chages of spacy's statistical models

PR #3468, opened 2019-11-19
Fix and simplify spacy tests

DeNeutoy:
#3432 broke our build somehow, even though it passed CI 🤔 . I fixed the spacy tests by deleting the test that failed, because we shouldn't have tests for statistical models which we don't directly control.


brendan-ai2:
For the record, Mark and I suspect that #3432 passed because Spacy released models (not code) more recently. In particular, see https://github.com/explosion/spacy-models/releases/tag/en_core_web_sm-2.2.5.

PR #3477, opened 2019-11-22
Remove DEPRECATION.md

brendan-ai2:
- By and large we haven't been following this proposal. I'd like remove it to avoid confusion.
- Further, I'm not sure it's the approach we'd like to take going forward.


brendan-ai2:
@matt-gardner, @DeNeutoy, I'm becoming increasingly concerned that my views on backward compatibility are not relevant to the type of use cases we wish to support. My background in building systems makes me view it as a key good, but I recognize that a standard research flow has different needs and that I am not a canonical user. Given this, I intend to stop advocating for backward compatible changes. Slowing down or blocking improvements to the library is something I'd greatly like to avoid. My request to you two is that you guide our project's views around backward compatibility to the extent that you find it important. I'm happy to help with either implementation or advice if requested, but I intend to stop commenting on this matter in PRs beyond trivialities like defaults for new arguments. Hopefully in the long run this will be best for our users. Thank you.

brendan-ai2:
Your concrete ideas sound very reasonable. Noting breaking changes while having a lightweight process sounds like a good balance. I can certainly operate in that framework. @DeNeutoy, do you think Matt's proposal will address your concerns?

(Regarding your side note, I just tried serializing a model with `dill` and it seemed to roundtrip correctly. I make no judgements as to whether that's actually a good idea, but it's a possible tool for shipping artifacts should we need another.)

brendan-ai2:
@DeNeutoy verbally LGTMed this after the triage. (I thought there was an LGTM via Github too, but it seems to have disappeared.)

Thank you both.

PR #3478, opened 2019-11-22
Remove pretrained models and sniff tests.

brendan-ai2:
- Pretrained models are now available from `allennlp-hub`. See https://github.com/allenai/allennlp-hub/blob/master/allennlp_hub/pretrained/__init__.py.
  - So if you were previously using them with `from allennlp import pretrained`, now you should use `from allennlp_hub import pretrained`.
- Sniff tests for same are at http://build.allennlp.org/viewType.html?buildTypeId=AllenNLPHub_Master.
- There's a bit of a race condition with respect to actually releasing this change. `allennlp-hub` should ideally depend on `allennlp` 1.0 -- indeed, the release build for `allennlp-hub` is broken because the expected outputs changed with the new Spacy models -- but if we release `allennlp` 1.0 first, then users will briefly not have access to `pretrained.py`. Perhaps releasing an alpha `allennlp-hub` dependent on `allennlp` via a github sha will be sufficient to iron out any release issues. Then we could quickly cut over to `allennlp` 1.0 and re-release.
- This is a breaking change. Noted in the v1.0 draft release notes. Link that may become stale: https://github.com/allenai/allennlp/releases/tag/untagged-9ecb65d3368d2841e938.


brendan-ai2:
This is of course failing TC on the sniff tests. I'll remove those from the TC config after approval and re-run the tests.

schmmd:
@brendan-ai2 can we merge this now or are we still waiting on something?

brendan-ai2:
Test failures. There were a few I had to work through and it got dropped. Looks like the last one is just removing the sniff tests from CI. I'll do that now.

brendan-ai2:
Done and tests re-running. If it passes, we can merge.

tedgoddard:
Where is the best place to look for instructions on upgrading to allennlp-hub and installing the allennlp-reading-comprehension package? (Sorry, this question isn't entirely relevant to this issue.)

schmmd:
@tedgoddard it's best to open a new issue if you'd like a response ;-)

tedgoddard:
https://github.com/allenai/allennlp/issues/3822

PR #3483, opened 2019-11-26
Fix an f-formatted string

JohnGiorgi:
There is a `KeyError` message in `pretrained_transformer_indexer.py` that appears to be an f-formatted string with broken syntax. This resulted in the literal string `token.text` being printed when the error is raised, instead of the value of the `text` attribute of `token`.

I simply fixed the syntax, so now this error message will print the value of `token.text` to screen when the error is raised.

### DIFF
--- a/allennlp/data/token_indexers/pretrained_transformer_indexer.py
+++ b/allennlp/data/token_indexers/pretrained_transformer_indexer.py
@@ -86,9 +86,8 @@ class PretrainedTransformerIndexer(TokenIndexer[int]):
                 # # It computed in PretrainedTransformerTokenizer.
                 indices.append(token.text_id)
             else:
-                raise KeyError(
-                    f"Using PretrainedTransformerIndexer but field text_id is not set for the following token: {token.text}"
-                )
+                raise KeyError(("Using PretrainedTransformerIndexer but field text_id is not set"
+                                f" for the following token: {token.text}"))
 
         return {index_name: indices}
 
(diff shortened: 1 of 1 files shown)


brendan-ai2:
Thanks for finding and fixing this! Would you mind also breaking up the long line? Our CI gives the following error:
```
[Step 4/11] ./allennlp/data/token_indexers/pretrained_transformer_indexer.py:90:116: E501 line too long (124 > 115 characters)
```

JohnGiorgi:
> Thanks for finding and fixing this! Would you mind also breaking up the long line? Our CI gives the following error:
> 
> ```
> [Step 4/11] ./allennlp/data/token_indexers/pretrained_transformer_indexer.py:90:116: E501 line too long (124 > 115 characters)
> ```

My pleasure! Hows that?

brendan-ai2:
Ugh, sorry, we use `black` as an auto-formatter and now that's complaining. If you switch to:

```
                raise KeyError(
                    "Using PretrainedTransformerIndexer but field text_id is not set"
                )
```

that should work. For a longer PR I'd ask you to give me permission to push to `JohnGiorgi:patch-1` so I could help deal with some of the minutiae. But for this quick bug-fix I'm really hoping we don't need to!

JohnGiorgi:
> Ugh, sorry, we use `black` as an auto-formatter and now that's complaining. If you switch to:
> 
> ```
>                 raise KeyError(
>                     "Using PretrainedTransformerIndexer but field text_id is not set"
>                     f" for the following token: {token.text}"
>                 )
> ```
> 
> that should work. For a longer PR I'd ask you to give me permission to push to `JohnGiorgi:patch-1` so I could help deal with some of the minutiae. But for this quick bug-fix I'm really hoping we don't need to!

No problem! I copy-pasted your solution. Third times the charm?

PR #3491, opened 2019-11-28
Add bleu_ngram_weights option to SimpleSeq2Seq

shaleenx:
This PR adds an option to specify n-gram weights for calculating
the bleu score weighted for each n-gram size. SimpleSeq2Seq passes
these weights to the BLEU class constructor as is.
Default value is set to (0.25, 0.25, 0.25, 0.25).

modified: models/encoder_decoders/simple_seq2seq.py


schmmd:
@shaleenx some of the tests are failing so we cannot yet merge this.

PR #3503, opened 2019-12-04
import spaCy model module directly instead of forcing a link after download

nanowizard:
I was dealing with a similar problem of downloading spaCy models programmatically in an internal python package I'm working on. Seeing the way you handled the issue presented in https://github.com/explosion/spaCy/issues/3435 was invaluable in coding up our solution. Since spaCy supports [native imports of their models](https://spacy.io/usage/models#usage-import) (the team even recommends native imports for large projects) and since this method seems to work in the same python session I think using native imports to load the model vs the `link` workaround is a better solution. In fact, I've seen some issues with the `link` workaround on some Windows machines, whereas the native import always seems to work.

Just figured I'd throw this out there since looking at your solution was invaluable to our project :).

### DIFF
--- a/allennlp/common/util.py
+++ b/allennlp/common/util.py
@@ -317,7 +317,7 @@ def get_spacy_model(
 
             # Import the downloaded model module directly and load from there
             spacy_model_module = __import__(spacy_model_name)
-            spacy_model = spacy_model_module.load(disable=disable)
+            spacy_model = spacy_model_module.load(disable=disable)  # type: ignore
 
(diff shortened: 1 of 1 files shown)


schmmd:
I'm open to this--but this change causes a `mypy` failure and this can't be merged until it's fixed.

`allennlp/common/util.py:320: error: Module has no attribute "load"`

matt-gardner:
Thanks @nanowizard!

PR #3506, opened 2019-12-07
Fix OOM while loading best model state

### LINKED ISSUE #3499: Potential OOM while loading best model state?
At the end of a training, `trainer.model` is updated with the best model state by calling `checkpointer.best_model_state()` ([here](https://github.com/allenai/allennlp/blob/master/allennlp/training/checkpointer.py#L172)) . This method returns the state as follows:

```python
best_model_state_path = os.path.join(self._serialization_dir, "best.th")
return torch.load(best_model_state_path)
```

Wouldn't it be better to load the state onto the CPU and return it? Since a `model.load_state_dict` would be called on this returned data, possibilities of any OOMs on GPU is less if done this way. In fact, `checkpointer.restore_checkpoint` [documents exactly this point](https://github.com/allenai/allennlp/blob/master/allennlp/training/checkpointer.py#L160) which is where I picked this up. I faced an OOM in one of my training runs and changing it to load to CPU first avoided the error.

scarecrow1123:
Closes #3499

@brendan-ai2

### DIFF
--- a/allennlp/training/checkpointer.py
+++ b/allennlp/training/checkpointer.py
@@ -169,7 +169,7 @@ class Checkpointer(Registrable):
         if self._serialization_dir:
             logger.info("loading best weights")
             best_model_state_path = os.path.join(self._serialization_dir, "best.th")
-            return torch.load(best_model_state_path)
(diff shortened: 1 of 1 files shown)


brendan-ai2:
Thank you!

Issue #3504, opened 2019-12-05
Unable to install the master branch with poetry due to incomplete pyproject.toml


willfrey:
**Description of the bug**
I'm unable to install the master branch directly from GitHub with [poetry](https://github.com/sdispater/poetry) due to an incomplete [pyproject.toml](https://github.com/allenai/allennlp/blob/master/pyproject.toml) file. According to https://github.com/sdispater/poetry/issues/1678, the `[tool.poetry]` section lacks the mandatory keys `version`, `description`, and `authors`.

Additionally, the `pyproject.toml` file should have the project's dependencies declared in the `[tool.poetry.dependencies]` section.

I tried a couple of solutions and one worked. All of that information is down in the **Additional context** section, as to not clutter this section anymore than I already have.

**To Reproduce**
1. Install `poetry`.
2. Run the following snippet

```console
cd /tmp
poetry new pkg
cd pkg
poetry env use python3  # Assuming python3 is Python 3.7.5
poetry add -vvv "git+https://github.com/allenai/allennlp.git@master"
```

3. Witness the `RuntimeError` that is displayed in the terminal complaining about `version` and `description`.

**Expected behavior**
I expect `poetry` to be able to install `allennlp` from the master branch on GitHub successfully, including parsing the declared project dependencies.

**System (please complete the following information):**
 - OS: macOS Catalina Version 10.15.1
 - Python version: 3.7.5
 - AllenNLP version: master

**Additional context**
I can get this to work if I edit `pyproject.toml` by removing the `[tool.poetry]` section _entirely_ and specify `setuptools` as the build system like such

```toml
[tool.black]
line-length = 100

include = '\.pyi?$'

exclude = '''
(
      __pycache__
    | \btutorials\b
    | \bbuild\b
    | \.git
    | \.mypy_cache
    | \.pytest_cache
    | \.vscode
    | \.venv
    | \bdist\b
    | \bdoc\b
)
'''

[build-system]
requires = ["setuptools", "wheel"]
build-backend = "setuptools.build_meta"
```

This tells `poetry` that we're going to build this with `setuptools` and `wheel`. Dependencies get parsed correctly and everything installs just like I want it to!

I also tried adding the minimum fields to `[tool.poetry]` to see if that'd work

```toml
[tool.poetry]
name = "allennlp"
version = ""
description = ""
authors = []

[tool.black]
line-length = 100

include = '\.pyi?$'

exclude = '''
(
      __pycache__
    | \btutorials\b
[...]

PR #3508, opened 2019-12-09
Remove [tool.poetry] and add [build-system]

willfrey:
`pyproject.toml` had a `[tool.poetry]` section that was incomplete. This removes that section and adds a `[build-system]` section that properly instructs any build system using `pyproject.toml` that it should use `setuptools` and `wheel` to build AllenNLP.

This should solve #3504

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,3 @@
-[tool.poetry]
-name = "allennlp"
-
 [tool.black]
 line-length = 100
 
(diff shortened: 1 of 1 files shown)


willfrey:
You can read more about this in [PEP 517](https://www.python.org/dev/peps/pep-0517/). Although the PEP is marked as provisional at the moment, this should have no impact on traditional anything that's calling `setup.py` directly.

schmmd:
I don't know anything about poetry--but this passes tests and looks good to me.

PR #3515, opened 2019-12-12
Dist tests

DeNeutoy:
Add some tests, fix a couple of minor bugs:

- Include packages was the wrong type, so we crashed
- `torch.multiprocessing.set_start_method` needs to be called with a `force` flag, otherwise in some environments it will raise an error.

I tested this on my multi-gpu machine, will run on CI too.

### DIFF
--- a/allennlp/run.py
+++ b/allennlp/run.py
@@ -23,6 +23,10 @@ def run():
 if __name__ == "__main__":
     # First, let Pytorch's multiprocessing module know how to create child processes.
     # Refer https://docs.python.org/3.7/library/multiprocessing.html#multiprocessing.set_start_method
-    torch.multiprocessing.set_start_method("spawn")
+    # We use force=True because this is guarded by the if __name__ ... block, which
+    # means that it only runs if allennlp is being run as a binary.
+    # Without this, if another library that allennlp uses has set the start method,
+    # this line would crash.
+    torch.multiprocessing.set_start_method("spawn", force=True)
 
     run()
--- a/allennlp/run.py
+++ b/allennlp/run.py
@@ -3,8 +3,6 @@ import logging
 import os
 import sys
 
-import torch
-
 if os.environ.get("ALLENNLP_DEBUG"):
     LEVEL = logging.DEBUG
 else:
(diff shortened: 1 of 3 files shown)


allennlp/run.py
```diff
@@ -23,6 +23,10 @@ def run():
 if __name__ == "__main__":
     # First, let Pytorch's multiprocessing module know how to create child processes.
     # Refer https://docs.python.org/3.7/library/multiprocessing.html#multiprocessing.set_start_method
-    torch.multiprocessing.set_start_method("spawn")
+    # We use force=True because this is guarded by the if __name__ ... block, which
+    # means that it only runs if allennlp is being run as a binary.
+    # Without this, if another library that allennlp uses has set the start method,
+    # this line would crash.
+    torch.multiprocessing.set_start_method("spawn", force=True)
```
brendan-ai2: Hmm, this is interesting. What are the implications if another library has set a different start method and we force past that? I'm having trouble finding the docs for `force`. Pytorch seems to import it from Python which is here: https://github.com/python/cpython/blob/master/Lib/multiprocessing/context.py#L241 But the docs don't mention it and the code doesn't help much either... (Docs : https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_start_method doesn't mention it.)

brendan-ai2: I'm guessing you have some other source that explains why it's necessary?

DeNeutoy: It appears that setting this is some kind of global python state, which I don't understand - e.g running this sequentially in two separate python interpreters one after another, closing inbetween, results in it crashing. I'm not super sure about the implications, but using distributed with pytorch does require us to use this spawn method.

DeNeutoy: Actually on my linux machine, even opening a python interpreter causes it to already be set.

brendan-ai2: Oof, this is nasty. What it sounds like is that some orphaned processes are keeping alive the process that Python creates to spawn the workers (or something in that vein). See https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods.

brendan-ai2:
Fairseq links for reference:
https://github.com/pytorch/fairseq/blob/b5f41f828b0ec9b67fa60aceb0778073d1b368b2/fairseq/distributed_utils.py#L71
https://github.com/pytorch/fairseq/blob/0d03aa889cf6f8363067a854cd40cbe5bd2bfb29/train.py#L299

PR #3516, opened 2019-12-13
strip out old DP stuff, ensure multiple cuda devices raises errors

DeNeutoy:
- `DataParallel` is removed
- `Trainer` and `CallbackTrainer` now take a single cuda device
- `batch_loss` now takes only a single batch, rather than a list (prev required for DataParallel training).
- `trainer._cuda_devices` removed and replaced with `trainer.cuda_device`
- `trainer._multi_gpu` is removed
- Config for distributed training is now read from the top level of the config, rather than the `Trainer`, because it is not explicitly used to construct the `Trainer` object. This means we have to do less config inspection in `allennlp train`, as well as ensuring that `Trainer` configs work with all other commands.

### DIFF
--- a/allennlp/commands/find_learning_rate.py
+++ b/allennlp/commands/find_learning_rate.py
@@ -316,7 +316,7 @@ def search_learning_rate(
     else:
         lr_update_factor = (end_lr / start_lr) ** (1.0 / num_batches)
 
-    for i, batch_group in enumerate(train_generator_tqdm):
+    for i, batch in enumerate(train_generator_tqdm):
 
         if linear_steps:
             current_lr = start_lr + (lr_update_factor * i)
@@ -327,7 +327,7 @@ def search_learning_rate(
             param_group["lr"] = current_lr
 
         trainer.optimizer.zero_grad()
-        loss = trainer.batch_loss(batch_group, for_training=True)
+        loss = trainer.batch_loss(batch, for_training=True)
         loss.backward()
         loss = loss.detach().cpu().item()
 
--- a/allennlp/commands/find_learning_rate.py
+++ b/allennlp/commands/find_learning_rate.py
@@ -52,7 +52,7 @@ import logging
 import shutil
 
 from allennlp.commands.subcommand import Subcommand
-from allennlp.common.checks import ConfigurationError, check_for_gpu
+from allennlp.common.checks import ConfigurationError, check_for_gpu, parse_cuda_device
 from allennlp.common import Params, Tqdm
 from allennlp.common.util import prepare_environment, lazy_groups_of
 from allennlp.data import Vocabulary, DataIterator
@@ -193,6 +193,14 @@ def find_learning_rate_model(
     prepare_environment(params)
 
     cuda_device = params.params.get("trainer").get("cuda_device", -1)
+    devices = parse_cuda_device(cuda_device)
+
+    # HACK: The trainer can not be constructed with multiple gpus.
+    # TODO(Mark): rework this so that cuda devices for distributed training are passed
+    # somewhere else, so configs are always valid.
+    if isinstance(devices, list):
+        cuda_device = devices[0]
+        params.params["trainer"]["cuda_device"] = cuda_device
     check_for_gpu(cuda_device)
 
     all_datasets = datasets_from_params(params)
@@ -223,6 +231,7 @@ def find_learning_rate_model(
     train_data = all_datasets["train"]
 
     trainer_params = params.pop("trainer")
+
     no_grad_regexes = trainer_params.pop("no_grad", ())
     for name, parameter in model.named_parameters():
         if any(re.search(regex, name) for regex in no_grad_regexes):
@@ -296,10 +305,7 @@ def search_learning_rate(
 
     trainer.model.train()
 
-    num_gpus = len(trainer._cuda_devices)
-
-    raw_train_generator = trainer.iterator(trainer.train_data, shuffle=trainer.shuffle)
-    train_generator = lazy_groups_of(raw_train_generator, num_gpus)
+    train_generator = trainer.iterator(trainer.train_data, shuffle=trainer.shuffle)
     train_generator_tqdm = Tqdm.tqdm(train_generator, total=num_batches)
 
     learning_rates = []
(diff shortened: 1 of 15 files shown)


allennlp/commands/find_learning_rate.py
```diff
@@ -193,6 +193,14 @@ def find_learning_rate_model(
     prepare_environment(params)
 
     cuda_device = params.params.get("trainer").get("cuda_device", -1)
+    devices = parse_cuda_device(cuda_device)
+
+    # HACK: The trainer can not be constructed with multiple gpus.
```
DeNeutoy: @brendan-ai2 - this PR makes the `Trainer` take strictly a single GPU. This presents a small problem, which is that we still want configs which have a `"cuda_device": [1,2,3]` line to work with other commands which are not `train` (like in this find_learning_rate command). 

I put this hack in here to demonstrate the problem (we have to make sure to manually parse the cuda devices every time we try to build the trainer, which is kind of gross), but I think a good way to resolve this is to have a separate `"distributed_gpus"` argument, either in the trainer or the top level of the config, which we can use in `train` to determine and launch the workers. This way, `Trainer.from_params` would be guaranteed to work in all cases without manually parsing out the cuda device.

Also, maybe it makes sense to move the `distributed`, `master_address` and `master_port` config arguments to the top level of the config - they are not actually arguments to the `Trainer`, they just affect how it is constructed within the `train` command. Does that make sense? Do you have any other ideas?

DeNeutoy: Actually I think moving the distributed config up to the top level just solves all the problems, I think i'll do that.

allennlp/tests/commands/train_test.py
```diff
@@ -99,12 +99,9 @@ def test_train_model_distributed(self):
...
                 "validation_data_path": SEQUENCE_TAGGING_DATA_PATH,
                 "iterator": {"type": "basic", "batch_size": 2},
-                "trainer": {
-                    "num_epochs": 2,
-                    "optimizer": "adam",
-                    "distributed": True,
-                    "cuda_device": [0, 1],
-                },
+                "trainer": {"num_epochs": 2, "optimizer": "adam"},
+                "distributed": True,
```
DeNeutoy: This flag is a bit pointless now - maybe we can just rely on distributed cuda_devices as the single flag?

allennlp/training/trainer.py
```diff
@@ -775,14 +763,15 @@ def from_params(  # type: ignore
         lr_scheduler_params = params.pop("learning_rate_scheduler", None)
         momentum_scheduler_params = params.pop("momentum_scheduler", None)
 
+        check_for_gpu(cuda_device)
         if isinstance(cuda_device, list):
-            model_device = cuda_device[0]
-        else:
-            model_device = cuda_device
-        if model_device >= 0:
+            raise ConfigurationError(
```
DeNeutoy: I would like to move this check inside `parse_cuda_devices`, as the two places we pass cuda devices are strictly typed - `cuda_device` inside the trainer config must be an `int` and `distributed_cuda_devices` at the top level must be a `List[int]`. However, we should probably still raise a nice warning for people who are converting their configs. Thoughts?

allennlp/commands/find_learning_rate.py
brendan-ai2: Yeah, that makes total sense. One suggestion though, I suspect that we may need further distributed options soon. (E.g., various multi-node settings) What about having config like:

```
"trainer": {"num_epochs": 2, "optimizer": "adam"},
"distributed": {"cuda_devices": [0, 1, 2]},
```

This gives us an obvious place to add them.

allennlp/tests/commands/train_test.py
brendan-ai2: Less redundancy sounds good. Though see my point above about having a top-level `distributed: {...` stanza.

allennlp/training/trainer.py
brendan-ai2: I'm definitely in favor of raising a nice warning to assist in changing the configs. To that end, could we add a snippet in the exception? Something like:

```
"""In allennlp 1.0, the Trainer cannot be passed multiple cuda devices. Instead, use the faster Distributed Data Parallel. For instance, if you previously had config like:
{
  "trainer": {
    "cuda_device": [0, 1, 2, 3],
    "num_epochs": 20,
    ...
  }
}

simply change it to:

{
  "distributed": {
    "cuda_devices": [0, 1, 2, 3],
  },
  "trainer": {
    "num_epochs": 20,
    ...
  }
}
```

brendan-ai2: As for where to put the warning, it seems that `parse_cuda_devices` is outdated now given the strict typing you mention. Maybe make a `parse_cuda_device` that gives an error if you pass it a list and similarly a `parse_cuda_devices` that mandates it has a list?

allennlp/training/trainer_base.py
```diff
@@ -36,12 +36,18 @@ def __init__(
         rank: int = 0,
```
brendan-ai2: Type of cuda_device a few lines up should be just `int` no`.

allennlp/training/util.py
```diff
@@ -328,36 +325,6 @@ def create_serialization_dir(
         os.makedirs(serialization_dir, exist_ok=True)
 
 
-def data_parallel(
```
brendan-ai2: Hooray! 🎉

allennlp/training/trainer_base.py
```diff
@@ -36,12 +36,18 @@ def __init__(
...
+        if isinstance(cuda_device, list):
+            raise ConfigurationError(
+                "In allennlp 1.0, the Trainer can only be assigned a single `cuda_device`. "
+                "Instead, we use torch's DistributedDataParallel at the command level, meaning "
+                "our Trainer always uses a single GPU per process."
+            )
+
+        if not isinstance(cuda_device, int):
             raise ConfigurationError(
                 "Expected an int or list for cuda_device, got {}".format(cuda_device)
```
brendan-ai2: This error still mentions list.

DeNeutoy: Thanks, fixed

allennlp/training/trainer_base.py
DeNeutoy: Thanks, fixed

allennlp/training/trainer.py
DeNeutoy: Added this inside `parse_cuda_device`, which now strictly returns an `int`. We ended up not needing the version which parses multiple of them, as we only do it in one place.

PR #3522, opened 2019-12-13
Fix mask usage in CRF decoding

eladsegal:
Before the fix, viterbi_tags takes a mask as an argument, but it doesn't really use it as a mask - it just truncates the sequence to the number of masked positions: https://github.com/allenai/allennlp/blob/9fa0bde5ce99abf2d5ec12a36366e662241f7099/allennlp/modules/conditional_random_field.py#L392-L400

This commit fixes it and also makes the mask optional.
Tests are modified accordingly.
Fixes https://github.com/allenai/allennlp/issues/3523.

### DIFF
--- a/allennlp/modules/conditional_random_field.py
+++ b/allennlp/modules/conditional_random_field.py
@@ -331,7 +331,7 @@ class ConditionalRandomField(torch.nn.Module):
         return torch.sum(log_numerator - log_denominator)
 
     def viterbi_tags(
-        self, logits: torch.Tensor, mask: torch.Tensor, top_k: int = None
+        self, logits: torch.Tensor, mask: torch.Tensor = None, top_k: int = None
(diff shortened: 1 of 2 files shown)


eladsegal:
I first encountered the problem when I wanted to ignore both in training and decoding tokens that are non-initial wordpieces (e.g. ["##NL", "##P"] when "AllenNLP" is tokenized to ["Allen", "##NL", "##P]). This is probably not a common use case, but I was still mislead by the naming of the argument as `mask`. 
A more common example would be ignoring special tokens ([CLS], [SEP]).

Even though the examples I mentioned can be handled by modifying the input before calling `viterbi_tags`, the current mask usage in it is inconsistent with `_input_likelihood` and `_joint_likelihood` that do treat the mask as a mask and not just as an indicator to the length of the sequence.

DeNeutoy:
Makes perfect sense, thanks a lot for the contribution!

PR #3525, opened 2019-12-14
Found reason/fix for skipped test in text classification dataset reader

apmoore1:
The text classification dataset reader test to test for sentence splitting was failing due to the sentence splitter changing when the version of Spacy changes an example of the different sentence splits based on different version of Spacy can be seen below:

```python
from allennlp.data.tokenizers.sentence_splitter import SpacySentenceSplitter
splitter= SpacySentenceSplitter()
text = "Memphis Rout Still Stings for No. 14 Louisville; Coach Petrino Vows to Have Team Better Prepared. NASHVILLE, Tenn. Nov 3, 2004 - Louisville #39;s 30-point loss at home to Memphis last season is still a painful memory for the Cardinals."
sentences = splitter.split_sentences(text)
sentences
['Memphis Rout Still Stings for No. 14 Louisville;', 
'Coach Petrino Vows to Have', 
'Team Better Prepared.', 
'NASHVILLE,', 
'Tenn. Nov 3, 2004 - Louisville #39;s 30-point loss at home to Memphis last season is still a painful memory for the Cardinals.']
import spacy
spacy.__version__
'2.1.9'
```

```python
from allennlp.data.tokenizers.sentence_splitter import SpacySentenceSplitter
splitter= SpacySentenceSplitter()
text = "Memphis Rout Still Stings for No. 14 Louisville; Coach Petrino Vows to Have Team Better Prepared. NASHVILLE, Tenn. Nov 3, 2004 - Louisville #39;s 30-point loss at home to Memphis last season is still a painful memory for the Cardinals."
sentences = splitter.split_sentences(text)
sentences
['Memphis Rout Still Stings for No. 14 Louisville; Coach Petrino Vows to Have Team Better Prepared.', 
'NASHVILLE,', 
'Tenn. Nov 3, 2004', 
'- Louisville #39;s 30-point loss at home to Memphis', 
'last season is still a painful memory for the Cardinals.']
import spacy
spacy.__version__
'2.2.3'
```

Therefore I created a test that uses the SpacySentenceSplitter and Spacy Tokenizer to test that the text read in from the files into the dataset readers are being processed correctly. It also test that if the sentence splitter is being used in conjunction with the `max_sequence_length` argument then the sentences after they have been split only contain at most  `max_sequence_length` number of tokens.


apmoore1:
I agree, I never thought about pinning a version of spacy to create consistent results, but I think that would be a better approach.

matt-gardner:
We had a discussion about this on a different issue not very long ago. We don't want to pin spacy. We want to be as permissive as possible in our dependencies, so users can do what they want with those other libraries. Pinning spacy just for a test is pretty extreme; much better to have a test that won't break if spacy's models change.  The point of the test is not to ensure a particular splitting, but to test that the component runs. We don't want to test spacy's models, just our integration with them.

(I haven't looked at the actual code here, just commenting on the discussion.)

apmoore1:
I think the test written here satisfies those requirements.

PR #3527, opened 2019-12-14
remove make vocab

DeNeutoy:
`make-vocab` is identical to `dry-run` apart from it does less stuff. This removes it.


allennlp/commands/__init__.py
```diff
@@ -71,7 +70,6 @@ def create_parser(
         "train": Train(),
         "evaluate": Evaluate(),
         "predict": Predict(),
-        "make-vocab": MakeVocab(),
```
brendan-ai2: How about `"make-vocab": DryRun()` here?

DeNeutoy: sure, done

DeNeutoy:
Dry run was added after make-vocab, I guess at the time I didn't think about how similar they were because dry-run was more focused on printing out some vocab stats and doing a few other checks.

PR #3528, opened 2019-12-15
Tokenize with offsets: Hugging Face's transformers

eladsegal:
This pull request suggests a new tokenizer that uses the tokenizers from https://github.com/huggingface/transformers.
In constrat to `PretrainedTransformerTokenizer`, the suggestion for this tokenizer is to be used with Hugging Face's API of `encode_plus`, and without TokenIndexer.

The main appeal of this tokenizer it is the addition of `tokenize_with_offsets`, which uses the same code as in https://github.com/huggingface/transformers/pull/2178 and for which thorough testing was performed. If Hugging Face will merge it, then the duplicate parts can be removed.

### DIFF
--- a/allennlp/data/fields/labels_field.py
+++ b/allennlp/data/fields/labels_field.py
@@ -1,0 +1,133 @@
+from typing import Dict, List, Union, Set, Iterator
+import logging
+import textwrap
+
+from overrides import overrides
+import torch
+
+from allennlp.common.checks import ConfigurationError
+from allennlp.common.util import pad_sequence_to_length
+from allennlp.data.fields.sequence_field import SequenceField
+from allennlp.data.vocabulary import Vocabulary
+
+logger = logging.getLogger(__name__)
+
+
+class LabelsField(SequenceField):
+    """
+    A ``LabelsField`` is a sequence of labels.
+
+    This field will get converted into a list of integer class ids, representing the correct class
+    for each element in the sequence.
+
+    Parameters
+    ----------
+    labels : ``Union[List[str], List[int]]``
+        A sequence of categorical labels, encoded as strings or integers.  These could be POS tags like [NN,
+        JJ, ...], BIO tags like [B-PERS, I-PERS, O, O, ...], or any other categorical tag sequence. If the
+        labels are encoded as integers, they will not be indexed using a vocab.
+    label_namespace : ``str``, optional (default='labels')
+        The namespace to use for converting tag strings into integers.  We convert tag strings to
+        integers for you, and this parameter tells the ``Vocabulary`` object which mapping from
+        strings to integers to use (so that "O" as a tag doesn't get the same id as "O" as a word).
+    """
+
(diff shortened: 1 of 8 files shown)


allennlp/data/fields/labels_field.py
```diff
@@ -0,0 +1,135 @@
...
+
+from allennlp.common.checks import ConfigurationError
+from allennlp.common.util import pad_sequence_to_length
+from allennlp.data.fields.sequence_field import SequenceField
+from allennlp.data.vocabulary import Vocabulary
+
+logger = logging.getLogger(__name__)
+
+
+class LabelsField(SequenceField):
```
matt-gardner: What does this provide on top of our [`SequenceLabelField`](https://github.com/allenai/allennlp/blob/master/allennlp/data/fields/sequence_label_field.py)?

allennlp/data/tokenizers/huggingface_transformers_tokenizer.py
```diff
@@ -0,0 +1,430 @@
...
+    4. Optioanlly, add fields such as `token_type_ids` and `special_tokens_mask`
+
+    Example:
+        tokens = tokenizer.tokenize(text)
+        encoded_inputs = self._tokenizer.encode_plus([token.text for token in tokens],
+                                    add_special_tokens=True,
+                                    return_token_type_ids=True,
+                                    return_special_tokens_mask=True)
+        fields: Dict[str, Field] = {}
+        fields['tokens'] = LabelsField(encoded_inputs['input_ids'])
```
matt-gardner: This goes pretty strongly against the way that AllenNLP is designed to work with text inputs.  I think the better solution is to improve our `PretrainedTransformerTokenizer` and related classes with whatever functionality is missing, instead of adding an entirely separate, largely duplicate class that doesn't keep to the design principles of the library.

allennlp/data/tokenizers/huggingface_transformers_tokenizer.py
```diff
@@ -0,0 +1,430 @@
...
+        self._pretrained_model = pretrained_model
+        self._kwargs = kwargs
+        self._tokenizer = AutoTokenizer.from_pretrained(pretrained_model, **init_kwargs)
+        self._tokenization_cache: Dict[str, str] = {}
+
+    @overrides
+    def tokenize(self, text: str) -> List[Token]:
+        return [Token(text=token) for token in self._tokenizer.tokenize(text, **self._kwargs)]
+
+    def tokenize_with_offsets(self, text: str) -> List[Token]:
```
matt-gardner: Having this functionality would be great, but if it's going to go upstream, we should just wait for it to be merged there, instead of duplicating the code here.

eladsegal:
I agree with everything you wrote. 
I'll close it and make some changes so it adheres to the principles of AllenNLP,
and then I think this PR can be useful for anyone who is eager for token offsets (as I was).

PR #3529, opened 2019-12-16
Torch distributed

DeNeutoy:
Includes the following PRs from the `torch-distributed` branch:
https://github.com/allenai/allennlp/pull/3516
https://github.com/allenai/allennlp/pull/3515
https://github.com/allenai/allennlp/pull/3414
https://github.com/allenai/allennlp/pull/3390
https://github.com/allenai/allennlp/pull/3372


DeNeutoy:
@matt-gardner I assigned you just in case you had any Qs/ red flags.

matt-gardner:
If you two are happy with it, no concerns from me.  It'd be nice to see a brief writeup somewhere with what's new and different here (maybe in release notes?)

DeNeutoy:
Yep, will do 👍

brendan-ai2:
🚀

scarecrow1123:
Hey @DeNeutoy , thanks for taking this up! Just adding a word of caution since I noticed that `torch.set_start_method("spawn")` has been removed in the final version of `run.py`. The default start method that PyTorch uses is `fork` which is prone to deadlocks in various scenarios. Hence the recommended ways are to set it to either of `spawn` or `forkserver`. This would ensure that any multi-threading happening in the distributed context doesn't lead to any deadlocks. There may be no mutli-threading code in AllenNLP, but things could go wrong in the underlying `NCCL`/`OpenMP` layers as well (refer [this open PyTorch issue](https://github.com/pytorch/pytorch/issues/2245)). I was not sure if I had to raise an issue for this.

brendan-ai2:
@scarecrow1123 we dug into this a bit and it's not clear that this is necessary. In particular, we're using `multiprocessing.spawn` from https://github.com/pytorch/pytorch/blob/f9010d764894f3ba9bc6a129213a0faf42af9b65/torch/multiprocessing/spawn.py#L138 to create processes which uses `get_context` with `'spawn'`, i.e. it doesn't appear to rely on the default `start_method` you suggest setting. That said, my understanding here is far from crystal clear, so if you can elaborate on the issue and why it's necessary, we'd definitely appreciate it. :)

scarecrow1123:
@brendan-ai2 You're spot on with `mp.spawn`. Just want to make sure that the places such as data reader/iterator, where multi-processing is used would not be affected. I don't remember them using a separate context, please correct me if I'm wrong. Neither am I too sure about whether something would go wrong, but just being careful here.

brendan-ai2:
Thanks for the extra info, @scarecrow1123! Fortunately we're planning on ripping out the old `MultiprocessDatasetReader` and `MultiprocessIterator` shortly, so they shouldn't be an issue for long. Point well taken about being careful!

PR #3530, opened 2019-12-16
Improve docs and delete MODELS.md

schmmd:
We wanted to deprecate and remove `MODELS.md` and instead put that information in `pretrained.py` so it's in fewer locations.


schmmd:
I'll improve the docs formatting, but since i upgraded to catalina I first need to install xcode to run make-docs...

allennlp/pretrained.py
```diff
@@ -25,7 +25,13 @@ def predictor(self) -> Predictor:
```
brendan-ai2: This file is actually moving to https://github.com/allenai/allennlp-hub/blob/master/allennlp_hub/pretrained/allennlp_pretrained.py and allennlp_semparse_pretrained.py. You should might be able to generate a `diff` and `patch` `allennlp_pretrained.py` in the hub repo. Sorry for the trouble :(

schmmd: Not a big deal--the hard part was fighting with the doc formatting...

Shouldn't I just leave it in here for now though and merge forward?  I can apply a separate patch on `allennlp_hub`.  Alternatively I can revert my changes to `pretrained.py` with `git checkout master -- allennlp/pretrained.py`.

schmmd: NM--I see I have a conflict...

schmmd:
I'm pretty confused by the build failures I got (`Found a total of 768 violations and reported 18`) but maybe they had something to do with the overlapping changes.  I just merged master back in.

schmmd:
@brendan-ai2 I removed my changes to `pretrained.py` and can apply them to the hub separately.

allennlp/pretrained.py
brendan-ai2: Yep, that's my fault. Sorry about that. `allennlp-hub` is now the sole source for pretrained artifacts so this shouldn't be an issue going forward.

schmmd:
https://github.com/allenai/allennlp-hub/pull/9

PR #3531, opened 2019-12-16
Fix for infinite iterator

dirkgr:
I changed this to use some code I've been using for a long time.

If we want a more minimal fix, I think this would do it as well:
```
def lazy_groups_of(iterable: Iterable[A], group_size: int) -> Iterable[List[A]]:
    iterable = iter(iterable)
    return iter(lambda: list(islice(iterable, 0, group_size)), [])
```

I find the long form more readable, but I'm happy either way.


allennlp/common/util.py
```diff
@@ -106,12 +106,18 @@ def group_by_count(iterable: List[Any], count: int, default_value: Any) -> List[
 A = TypeVar("A")
 
 
-def lazy_groups_of(iterator: Iterator[A], group_size: int) -> Iterator[List[A]]:
+def lazy_groups_of(iterable: Iterable[A], group_size: int) -> Iterable[List[A]]:
     """
     Takes an iterator and batches the individual instances into lists of the
```
brendan-ai2: s/iterator/iterable/

allennlp/common/util.py
```diff
@@ -106,12 +106,18 @@ def group_by_count(iterable: List[Any], count: int, default_value: Any) -> List[
 A = TypeVar("A")
 
 
-def lazy_groups_of(iterator: Iterator[A], group_size: int) -> Iterator[List[A]]:
+def lazy_groups_of(iterable: Iterable[A], group_size: int) -> Iterable[List[A]]:
```
brendan-ai2: The return type here shouldn't be `Iterable` given that we're aiming for laziness. I think you want to use `Generator`.

brendan-ai2: (Or just stick with `Iterator`. The fact that it's a generator probably shouldn't matter.)

allennlp/common/util.py
```diff
@@ -106,12 +106,18 @@ def group_by_count(iterable: List[Any], count: int, default_value: Any) -> List[
...
     """
     Takes an iterator and batches the individual instances into lists of the
     specified size. The last list may be smaller if there are instances left over.
     """
-    return iter(lambda: list(islice(iterator, 0, group_size)), [])
+    iterable = iter(iterable)
```
brendan-ai2: Nit: `iterator = iter(iterable)` The `islice` call below looks like a bug when operating on an arbitrary iterable.

dirkgr: I like to keep the number of local variables down for debugging, but I see your point. Will make the change.

allennlp/common/util.py
dirkgr: A `Generator` is an `Iterator`. So to make the minimum number of promises to the caller, I'll change it back to `Iterator`.

PR #3537, opened 2019-12-17
Gradient accumulation with the distributed trainer

dirkgr:
I re-did gradient accumulation with the distributed trainer.

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -330,10 +330,22 @@ class Trainer(TrainerBase):
         self._pytorch_model.train()
 
         # Get tqdm for the training batches
-        train_generator = self.iterator(self.train_data, num_epochs=1, shuffle=self.shuffle)
+        batch_generator = self.iterator(self.train_data, num_epochs=1, shuffle=self.shuffle)
+        batch_group_generator = lazy_groups_of(
+            batch_generator, self._num_gradient_accumulation_steps
+        )
         num_training_batches = math.ceil(
             self.iterator.get_num_batches(self.train_data) / self._num_gradient_accumulation_steps
         )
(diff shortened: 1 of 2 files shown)


allennlp/training/trainer.py
```diff
@@ -343,21 +353,20 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
             train_generator_tqdm = train_generator
 
         cumulative_batch_size = 0
-        for batch in train_generator_tqdm:
+        for batches in lazy_groups_of(train_generator_tqdm, self._num_gradient_accumulation_steps):
```
brendan-ai2: Nit: Maybe `accumulation_groups` (or `batch_group` or whatever) instead of `batches`? Might be confusing to a reader see lines like `cur_batch = sum(training_util.get_batch_size(batch) for batch in batches)`.

allennlp/training/trainer.py
```diff
@@ -323,7 +331,9 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
 
         # Get tqdm for the training batches
         train_generator = self.iterator(self.train_data, num_epochs=1, shuffle=self.shuffle)
```
brendan-ai2: The `lazy_groups_of` call should be here to create a new generator and `tqdm` should be applied to that. As in https://github.com/allenai/allennlp/blob/3dda5ac9fcf390d8b83eb855249360711707c11c/allennlp/training/trainer.py#L306 previously. Otherwise tqdm will give an inaccurate count of the batches.

dirkgr:
@brendan-ai2, I took some license with the naming, and fixed the `tqdm` issue while I was at it.

PR #3543, opened 2019-12-18
Remove unidecode as it's GPLv2 and isn't in requirements.txt.

schmmd:


schmmd:
See https://discourse.allennlp.org/t/license-of-allennlp/154/2

schmmd:
`unidecode` was introduced in 121d3e8cc

PR #3546, opened 2019-12-19
Update names of environment variables for consistency

himkt:
**(This PR includes changes to break backward compatibility.)**

I update some environment variables for naming consistency.
I use `*_DATA_PATH` following [coref config](https://github.com/allenai/allennlp/blob/master/training_config/coref.jsonnet).


DeNeutoy:
Thanks but i'm not sure this is important enough to possibly break people/our commands on our website.

PR #3549, opened 2019-12-19
Add ability to compose Seq2SeqEncoders.

viking-sudo-rm:
## Problem

I have often faced the problem of wanting to add a "pre-contextualizer" before another kind of `Seq2SeqEncoder`. Currently, it seems the best way to accomplish this is to change a `Model` to have two encoder slots, or to add a nested sub-encoder to the `Seq2SeqEncoder` you are using. Both of these are ad hoc solutions, requiring code modifications as opposed to a simple fix in the configuration file.

## Solution

In order to solve this problem easily, I created a `ComposeEncoder` that allows you to create a new `Seq2SeqEncoder` out of a sequence of sub-encoders. If you are creating a `LanguageModel` and want to add an LSTM as a pre-contextualizer before `ENCODER`, you can simply write:

```json
    "contextualizer": {
      "type": "compose",
      "encoders": [
        {
          "type": "lstm",
          "input_size": 100,
          "hidden_size": 64,
          "bidirectional": false,
          "num_layers": 1,
        },
        ENCODER,
      ],
```

## Details

The `ComposeEncoder` applies its list of sub-encoders in sequence. It enforces the following basic constraints:
1. There must be at least one sub-encoder.
2. All sub-encoders must match in bidirectionality.
3. The output dimension of each sub-encoder is equal to the input dimension of the following sub-encoder.

### DIFF
--- a/allennlp/modules/seq2seq_encoders/compose_encoder.py
+++ b/allennlp/modules/seq2seq_encoders/compose_encoder.py
@@ -1,0 +1,75 @@
+from overrides import overrides
+import torch
+from typing import List
+
+from allennlp.modules.seq2seq_encoders.seq2seq_encoder import Seq2SeqEncoder
+
+
+@Seq2SeqEncoder.register("compose")
+class ComposeEncoder(Seq2SeqEncoder):
+
+    """This class can be used to compose several encoders in sequence.
+
+    Among other things, this can be used to add a "pre-contextualizer" before a Seq2SeqEncoder.
+
+    Parameters
+    ----------
+    encoders : ``List[Seq2SeqEncoder]``, required.
+        A non-empty list of encoders to compose. The encoders must match in bidirectionality.
+    """
+
+    def __init__(self, encoders: List[Seq2SeqEncoder]):
+        super().__init__()
+        self.encoders = encoders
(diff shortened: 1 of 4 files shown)


allennlp/modules/seq2seq_encoders/compose_encoder.py
```diff
@@ -0,0 +1,76 @@
...
+        # Compute bidirectionality.
+        all_bidirectional = all(encoder.is_bidirectional() for encoder in encoders)
+        any_bidirectional = any(encoder.is_bidirectional() for encoder in encoders)
+        self.bidirectional = all_bidirectional
+
+        if all_bidirectional != any_bidirectional:
+            raise ValueError("All encoders need to match in bidirectionality.")
+
+        if len(self.encoders) < 1:
+            raise ValueError("Need at least one encoder.")
```
dirkgr: It would be neat if this just does pass-through in the case of zero encoders, but I assume that doesn't work because then you can't compute input and output dimensionality?

viking-sudo-rm: Yeah, exactly. The pass through requires an “embedding dim” in order to fulfill that part of the API.

Issue #3389, opened 2019-10-23
Completely remove custom from_params methods


sai-prasanna:
**Is your feature request related to a problem? Please describe.**

Every registrable class has a classmethod from_params. We override it in cases where we want an alternative way to construct a object for the class.

Since we also use from_params as a dependency resolver, we have to inspect params object to satisfy dependencies of our main constructor.

Each class gets its dependency configuration and uses it to construct the object.

I think exposing the params to the class directly is kind of anti pattern. We will not be able to share  same dependency object across the object sub graph. We already face this with vocabulary object, which we now handle by explicitly passing it down the subgraph along with configuration. But for cases like sharing layers/module etc, we need a better approach.

**Describe the solution you'd like**

Remove from_params from Registrable. 
We still retain how we write configuration jsonnet since it's convinient.
But to construct the graph, we do it by constructing the object graph from leaf, up to the root, explicitly.
While possibly warning if we hit a cycle.

Now, we can allow dependencies in the configuration to be satisfied by specifying a key to shared dependency in the object graph. Now we can even allow global singleton, shared layers etc.
 
And overridden from_params can be converted to alternate constructor that is used by default to construct the object nodes if its present.

**Describe alternatives you've considered**

Haven't considered the full complexitiy of implementing such a full blown dependency resolver.

DeNeutoy:
Interesting idea, but it has some problems:

1. If we allow shared references, `from_params` get's _hugely_ complicated and very easy to shoot yourself in the foot if we implicitly share some references (i.e what we deem to be global state, such as the `vocab`) but don't share others, as at some point this will almost certainly clash with someone's intuition.

2. We don't really want to be in the business of managing a dependency resolver.

3. I'd like to encourage more scripted use of allennlp so that the first thought when you want to pass an object to multiple classes isn't "how can I make this work with `from_params`", but "this is easy to do if I write my own script". Making `from_params` more complicated does not help with this.

Having said that, we've had previous very good improvements to `from_params`, such as making it automatic for most classes which have type hints. So if you think the idea's good, a prototype version would be most welcome (but absolutely not guaranteed to be accepted as a contribution).

Thanks Sai!

sai-prasanna:
To think out aloud, from_params would be replaced with a load() classmethod, where you declare your dependencies. say Embedding would provide 'pre_trained_file' option there, which it doesn't expose in the ordinary constructor. And it might also have Vocabulary there. So the classes will not be aware of what is global and what is not.

[...]

PR #3550, opened 2019-12-19
Allow registering specific methods to use as constructors

matt-gardner:
Fixes one of the two main problems in #3389. This makes it so that you can specify a particular method to use as a constructor when registering a class, removing one of the main cases where we needed to write a custom `from_params` method.  Those methods are not self-documenting, as you don't know what's inside the `params` argument, and they make it very hard to use our code without also buying into our config file architecture.

This PR isn't actually as big as it looks.  I took the opportunity to re-order some of the methods in `Vocabulary` (grouping together the various constructors), which is the bulk of the actual diff here.

There are two main classes that were the target for this change: `Vocabulary` and `Embedding`.  `Vocabulary` has three different constructors, depending on how you want to create the vocab, and so is registered three times.  That's the place to look to see the benefit of this change.

The place to look to see the main logic that changed is in `FromParams`, and a little bit in `Registrable`.  Everything else is either using the new functionality (`Vocabulary` and `Embedding`) or minor things to account for how calling the classes changed (and this is mostly in tests, but also in a few other places).

A side effect of how I did this was that I needed to add `**extras` to every remaining custom `from_params` method.  The plan is to remove those entirely, eventually, so I didn't worry about it too much.  Some of them will need a bit of redesign of how things are constructed and passed around, though (mostly in the callbacks).


DeNeutoy:
@matt-gardner sorry this caught me just as I was leaving for Christmas. Is this in a reviewable state or...? Looks like quite a few conflicts.

matt-gardner:
Hmm, I thought it was in a reviewable state; let me look at it again.  I'll let you know when it's ready.

matt-gardner:
If there are still failure at this point, they are minor and shouldn't affect a code review.  So I'd say it's good to look at now.

allennlp/common/registrable.py
```diff
@@ -38,18 +38,22 @@ class Registrable(FromParams):
...
 
-    _registry: Dict[Type, Dict[str, Type]] = defaultdict(dict)
+    _registry: Dict[Type, Dict[str, Tuple[Type, str]]] = defaultdict(dict)
     default_implementation: str = None
 
     @classmethod
-    def register(cls: Type[T], name: str, exist_ok=False):
+    def register(cls: Type[T], name: str, constructor: str = None, exist_ok: bool = False):
         """
         Register a class under a particular name.
```
DeNeutoy: Now we can have markdown in docstrings, can you add a couple of examples of registering different constructors and their corresponding jsonnet config?

brendan-ai2:
@matt-gardner, looks like this caused some issues in `allennlp-hub`. See http://build.allennlp.org/viewLog.html?buildId=24152&buildTypeId=AllenNLPHub_Master&tab=buildLog&branch_AllenNLPHub=%3Cdefault%3E.

matt-gardner:
Can we get a notification for these breakages in our slack channel?

brendan-ai2:
Absolutely. I'll figure out how to do that.

matt-gardner:
Also, FYI, the tests are broken because I removed the old, deprecated way of passing arguments to a `BasicTextFieldEmbedder` (part of my effort to squash custom `from_params` methods).  We just need to modify the config files inside all of those `model.tar.gz` files, to include a `token_embedders` wrapper around their embedder arguments.  I'll add a note to the breaking changes in the release notes.  I'm not in the best position to fix that issue at the moment, as I'm on a slow internet connection at home, and downloading and re-uploading those models would be a pain.  I'll open an issue on the hub repo to fix this, and hopefully get to it on Friday when I'm in the office.

brendan-ai2:
https://github.com/allenai/allennlp-hub/issues/13

PR #3551, opened 2019-12-19
Ignore errors when cleaning up temp files.

schmmd:
Our builds are full of errors that are not very helpful.  My hope is that this flag will cause our build system to omit these exceptions, which are cluttering up our logs.

```
[11:36:38][Step 6/6] --- Logging error ---
[11:36:38][Step 6/6] Traceback (most recent call last):
[11:36:38][Step 6/6]   File "/usr/local/lib/python3.7/logging/__init__.py", line 1037, in emit
[11:36:38][Step 6/6]     stream.write(msg + self.terminator)
[11:36:38][Step 6/6]   File "/usr/local/lib/python3.7/site-packages/_pytest/capture.py", line 427, in write
[11:36:38][Step 6/6]     self.buffer.write(obj)
[11:36:38][Step 6/6] ValueError: I/O operation on closed file
[11:36:38][Step 6/6] Call stack:
[11:36:38][Step 6/6]   File "/usr/local/lib/python3.7/site-packages/allennlp/models/archival.py", line 251, in _cleanup_archive_dir
[11:36:38][Step 6/6]     logger.info("removing temporary unarchived model dir at %s", path)
```

### DIFF
--- a/allennlp/models/archival.py
+++ b/allennlp/models/archival.py
@@ -249,4 +249,4 @@ def load_archive(
 def _cleanup_archive_dir(path: str):
     if os.path.exists(path):
         logger.info("removing temporary unarchived model dir at %s", path)
-        shutil.rmtree(path)
+        shutil.rmtree(path, ignore_errors=True)


schmmd:
@dirkgr if I knew I'd solve the root problem... I don't understand why `shutil.rmtree` would ever throw `I/O operation on closed file`.

This certainly removes most of the errors in the logs, but as I said I'm not certain whether it's a good idea.

dirkgr:
`I/O operation on closed file` comes from when it tries to write to the logs. The root error here is that something closes the log file.

I'd feel better about this change if we knew why it works.

schmmd:
Gotcha.  So an error with `shutil.rmtree` is causing an error in the logger (because the logging file is closed).

Issue #3542, opened 2019-12-18
Is it possible to do HotFlip on GPU?


c4n:
* Is it possible to do HotFlip on GPU?

 I'm working on adversarial training. Is it possible to use HotFlip to create adversarial examples using GPU?

matt-gardner:
Have you tried it and it's not running on the GPU?

c4n:
Yes, I have tried it. It didn't work on GPU, It only worked when I transferred the model to the CPU.
`tagger = NewTaggerPredictor(model.cpu(), conll_read)
hotflipper = Hotflip(tagger, vocab_namespace="tokens")`

I noticed a few spots on [hotflip.py](https://github.com/allenai/allennlp/blob/master/allennlp/interpret/attackers/hotflip.py) where it moves things to the cpu e.g.:
line 79
`self.embedding_matrix = self._construct_embedding_matrix().cpu()`

matt-gardner:
I believe we did that because our primary use was having the code run in our demo, which runs on CPUs.  If you remove all of the calls to `.cpu()` (or add explicit calls to `.to()`), does it work?  If it does, a PR to add a `cuda_device` parameter to the constructor of `Hotflip` would be very welcome.

c4n:
Thank you for your reply. I will try to work on it this weekend.

c4n:
It worked! I will submit a PR soon.
It's about 4 times faster on GPU on one input sentence. (initialization time included)
![ภาพ](https://user-images.githubusercontent.com/4990317/71369067-34f92f00-25dc-11ea-94d6-b2705b96a896.png)

PR #3555, opened 2019-12-23
Add GPU capability to HotFlip

c4n:
Now it is possible to do HotFlip with GPU acceleration . 
 
see issue: #3542


allennlp/predictors/predictor.py
```diff
@@ -47,6 +48,7 @@ class Predictor(Registrable):
     def __init__(self, model: Model, dataset_reader: DatasetReader) -> None:
         self._model = model
         self._dataset_reader = dataset_reader
+        self.cuda_device = next(self._model.named_parameters())[1].get_device()
```
dirkgr: Why `[1]` and not `[0]`?

allennlp/predictors/predictor.py
```diff
@@ -110,13 +112,17 @@ def get_gradients(self, instances: List[Instance]) -> Tuple[Dict[str, Any], Dict
...
         dataset.index_instances(self._model.vocab)
-        outputs = self._model.decode(
-            self._model.forward(**dataset.as_tensor_dict())  # type: ignore
-        )
-
-        loss = outputs["loss"]
-        self._model.zero_grad()
-        loss.backward()
+        # To bypass "RuntimeError: cudnn RNN backward can only be called in training mode"
+        with backends.cudnn.flags(enabled=False):
```
dirkgr: This seems sketchy to me. Doesn't that disable running on the GPU altogether?

I assume you want to do this on a model that's in evaluation mode because you care about getting the gradients, but you don't care about making an update to the model. Doesn't PyTorch have a prescribed solution to this problem? It seems to me that this is just a safety check?

dirkgr: Apparently this is an open issue in PyTorch: https://github.com/pytorch/pytorch/issues/10006. That thread suggests a different workaround, but yours seems safer.

c4n:
> Can you run `flake8` and `black .` on this until neither has any complaints anymore? I think that'll get the tests to succeed.

All checks have passed!

Issue #3465, opened 2019-11-18
PretrainedTransformerEmbedder does not pass masks to *transformer models


DeNeutoy:
The huggingface repo accepts masks, we should definitely be passing them

https://github.com/huggingface/transformers/blob/master/transformers/modeling_bert.py#L629

https://github.com/allenai/allennlp/blob/master/allennlp/modules/token_embedders/pretrained_transformer_embedder.py#L25

PR #3560, opened 2019-12-27
Use attention_mask in pretrained transformers

RyujiTamaki:
Fix this issue https://github.com/allenai/allennlp/issues/3465

I also experimented with the MRPC dataset.
https://github.com/RyujiTamaki/mrpc-allennlp

Results is bellow.
```
2019-12-27 22:08:17,420 - INFO - allennlp.common.util - Metrics: {
  "best_epoch": 2,
  "peak_cpu_memory_MB": 3267.492,
  "peak_gpu_0_memory_MB": 10369,
  "peak_gpu_1_memory_MB": 19,
  "peak_gpu_2_memory_MB": 19,
  "training_duration": "0:01:17.479140",
  "training_start_epoch": 0,
  "training_epochs": 2,
  "epoch": 2,
  "training_accuracy": 0.9175637393767705,
  "training_loss": 0.20614140655274862,
  "training_cpu_memory_MB": 3267.492,
  "training_gpu_0_memory_MB": 10369,
  "training_gpu_1_memory_MB": 19,
  "training_gpu_2_memory_MB": 19,
  "validation_accuracy": 0.8505154639175257,
  "validation_loss": 0.4683948732339419,
  "best_validation_accuracy": 0.8505154639175257,
  "best_validation_loss": 0.4683948732339419
}
```

### DIFF
--- a/allennlp/data/token_indexers/pretrained_transformer_indexer.py
+++ b/allennlp/data/token_indexers/pretrained_transformer_indexer.py
@@ -95,7 +95,7 @@ class PretrainedTransformerIndexer(TokenIndexer[int]):
         # tokens are attended to.
         attention_mask = [1] * len(indices)
 
-        return {index_name: indices, "attention_mask": attention_mask}
+        return {index_name: indices, "mask": attention_mask}
 
     @overrides
     def get_padding_lengths(self, token: int) -> Dict[str, int]:
--- a/allennlp/data/token_indexers/pretrained_transformer_indexer.py
+++ b/allennlp/data/token_indexers/pretrained_transformer_indexer.py
@@ -31,10 +31,18 @@ class PretrainedTransformerIndexer(TokenIndexer[int]):
         We use a somewhat confusing default value of ``tags`` so that we do not add padding or UNK
         tokens to this namespace, which would break on loading because we wouldn't find our default
         OOV token.
+    mask_padding_with_zero:  ``bool``, optional (default=True)
+        If set to ``True``, the attention mask will be filled by ``1`` for actual values
+        and by ``0`` for padded values. If set to ``False``, inverts it (``1`` for padded values, ``0`` for
+        actual values)
     """
 
     def __init__(
-        self, model_name: str, namespace: str = "tags", token_min_padding_length: int = 0
+        self,
+        model_name: str,
+        namespace: str = "tags",
+        token_min_padding_length: int = 0,
+        mask_padding_with_zero: bool = True,
     ) -> None:
         super().__init__(token_min_padding_length)
         self._namespace = namespace
@@ -42,6 +50,8 @@ class PretrainedTransformerIndexer(TokenIndexer[int]):
         self._padding_value = self._tokenizer.convert_tokens_to_ids([self._tokenizer.pad_token])[0]
         logger.info(f"Using token indexer padding value of {self._padding_value}")
         self._added_to_vocabulary = False
+        self._mask_padding_with_zero = mask_padding_with_zero
+        self._mask_padding_value = 0 if mask_padding_with_zero else 1
 
     def _add_encoding_to_vocabulary(self, vocab: Vocabulary) -> None:
         """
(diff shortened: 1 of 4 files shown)


allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -31,17 +31,27 @@ class PretrainedTransformerIndexer(TokenIndexer[int]):
...
+        actual values)
     """
 
     def __init__(
-        self, model_name: str, namespace: str = "tags", token_min_padding_length: int = 0
+        self,
+        model_name: str,
+        namespace: str = "tags",
+        token_min_padding_length: int = 0,
+        mask_padding_with_zero: bool = True,
```
sai-prasanna: Any particular reason for this option?

RyujiTamaki: I implemented this pull request by referring to the following code.

https://github.com/huggingface/transformers/blob/master/src/transformers/data/processors/glue.py

When I looked for a case where mask_padding_with_zero would be False, XLNet hit it.

https://github.com/zihangdai/xlnet/blob/master/classifier_utils.py#L113-L115

I can remove this mask_padding_with_zero if I specify Input_mask. I fixed it that way because it seemed simpler.
https://github.com/huggingface/transformers/blob/master/src/transformers/modeling_xlnet.py#L581-L587

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -16,12 +16,17 @@ def __init__(self, model_name: str) -> None:
...
     @overrides
     def get_output_dim(self):
         return self.output_dim
 
-    def forward(self, token_ids: torch.LongTensor) -> torch.Tensor:  # type: ignore
+    def forward(
+        self, token_ids: torch.LongTensor, attention_mask: torch.LongTensor
+    ) -> torch.Tensor:  # type: ignore
+        if self.model_type == "xlnet":
+            return self.transformer_model(input_ids=token_ids, input_mask=attention_mask)[0]
```
matt-gardner: Looking at the code, I'm pretty sure this will do the wrong thing - if you want to use `input_mask`, you need to compute the inverse of what you computed above.  We should just remove this case.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -91,7 +91,11 @@ def tokens_to_indices(
                     f" for the following token: {token.text}"
                 )
 
-        return {index_name: indices}
+        # The mask has 1 for real tokens and 0 for padding tokens. Only real
+        # tokens are attended to.
+        attention_mask = [1] * len(indices)
+
+        return {index_name: indices, "attention_mask": attention_mask}
```
dirkgr: At https://github.com/allenai/allennlp/blob/master/allennlp/nn/util.py#L626 (and apparently many other places), we assume that the mask is stored under the key `"mask"`.

dirkgr: We also assume it's a `LongTensor`, not a `List`.

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
RyujiTamaki: Thank you for your advice.
I read huggingface's xlnet implementation and seem to have misunderstood. I will delete this case.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
RyujiTamaki: The `tokens_to_indices` method is expected to return `Dict [str, List [int]]`, so I think it's better not to return a `LongTensor`.  Is it correct?

matt-gardner: Yes, @RyujiTamaki, you're right.  The conversion to a tensor happens in `as_padded_tensor`, not here.

PR #3563, opened 2019-12-31
add double backquotes on doc strings 

hayata-yamamoto:
I formated docstrings so that sphinx can recognize arguments and types separately. 

```python 

class ClassName: 
"""
arg1: ``int``, required 
    explanation of argument 1 
arg2: ``str``, optional (default = None)
    explanation of argument2
"""

     def __init__(self, arg1: int, arg2: str = None): 
         pass
```

On documentation, this modification will be reflected like [this](https://allenai.github.io/allennlp-docs/api/allennlp.modules.elmo.html#allennlp.modules.elmo.Elmo).

## Background
Previously, some classes had troubles on documentations like 

```
arg1int
arg2str
```

These were caused by below docstrings

```python
class ClassName
"""
Parameters
----------
arg1 : int 
arg2 : str
"""
    def __init__(self, arg1: int, arg2: str = None): 
        pass
```

## Solutions 

1. Don't add whitespace between argument name and type 
2. Wrap type name by double backquotes like ``type``

Looking at existing codes, codes using the second is more than the first, so I've modified by the second. Also, since some codes lacked information of whether this argument is required or optional, I added such information on docstring based on initializers. 

## Remains

Since I checked only allennlp.modules, the others might have the same trouble.

### DIFF
--- a/allennlp/modules/attention/additive_attention.py
+++ b/allennlp/modules/attention/additive_attention.py
@@ -17,11 +17,11 @@ class AdditiveAttention(Attention):
 
     Parameters
     ----------
-    vector_dim : ``int``
+    vector_dim : ``int``, required
         The dimension of the vector, ``x``, described above.  This is ``x.size()[-1]`` - the length
         of the vector that will go into the similarity computation.  We need this so we can build
         the weight matrix correctly.
-    matrix_dim : ``int``
+    matrix_dim : ``int``, required
         The dimension of the matrix, ``y``, described above.  This is ``y.size()[-1]`` - the length
         of the vector that will go into the similarity computation.  We need this so we can build
         the weight matrix correctly.
--- a/allennlp/modules/attention/bilinear_attention.py
+++ b/allennlp/modules/attention/bilinear_attention.py
@@ -15,11 +15,11 @@ class BilinearAttention(Attention):
 
     Parameters
     ----------
-    vector_dim : ``int``
+    vector_dim : ``int``, required
         The dimension of the vector, ``x``, described above.  This is ``x.size()[-1]`` - the length
         of the vector that will go into the similarity computation.  We need this so we can build
         the weight matrix correctly.
-    matrix_dim : ``int``
+    matrix_dim : ``int``, required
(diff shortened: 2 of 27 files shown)


allennlp/modules/attention/additive_attention.py
```diff
@@ -17,11 +17,11 @@ class AdditiveAttention(Attention):
 
     Parameters
     ----------
-    vector_dim : ``int``
+    vector_dim : ``int``, required
```
schmmd: I'm confused because some of the lines contain a space before the `:` and others don't.  Should we have a space here or not?

matt-gardner: A long time ago, we found that we needed to have a space before and after the `:` in order for things to render properly.  I haven't looked at this in a *long* time, so I don't know if something changed from under us, or if I misunderstood something back then.  I'd like to know the answer, too =).

hayata-yamamoto: @schmmd @matt-gardner 
I understand your confusion. I was also confused. 
As I wrote earlier, 

> Don't add whitespace between argument name and type

is another solution. This is the sphinx style. To be honest, it's better to remove spaces before a colon in terms of the sphinx and docstring readability. 

However, most docstrings are styled by double backquotes. As long as I know, this style is also rendered correctly because sphinx recognizes backquotes as inline code.

Did you resolve a concern? 
If you want to remove white space, I'm willing to. Let me know your preferences!

schmmd: Ah--you're right.  We're entirely inconsistent in our codebase (I just looked at the output of `git grep -A 5 Parameters`).  Because we're already inconsistent that issue is independent of this PR.  Thanks for taking care of this!

schmmd:
@hayata-yamamoto just an FYI we made some radical changes to our docs so hopefully they are easier to manage going further! https://allenai.github.io/allennlp-docs/

hayata-yamamoto:
@schmmd 
Excellent News! Thank you for sharing me!

PR #3573, opened 2020-01-01
Update conllu to 2.2

EmilStenstrom:
This update has no backwards compatible changes, and all tests run clean ([meaning the same four tests fail as without this change](3570)) after this change.


DeNeutoy:
Thanks!

PR #3580, opened 2020-01-02
Remove requirements.txt as we install via setup.py.

schmmd:


schmmd:
Great--adding Brendan and Matt in case they catch something I missed.  If I don't hear more concerns I'll merge this before EOD.

README.md
```diff
@@ -121,6 +121,7 @@ Create a Python 3.7 virtual environment, and install AllenNLP in `editable` mode
 
   ```bash
   pip install --editable .
+  pip install -r dev-requirements.txt
```
brendan-ai2: I'd note that this is only if you want to develop on the library. Might not be obvious to a casual user.

schmmd: I think that it's best that, if users are cloning the master version of our GitHub repository locally, they simply install the developer requirements.  Most users will `pip install allennlp` (which comes earlier in the README) and will avoid the `dev-requirements.txt` (because it's not mentioned in that section at all).

I think that most people who install from source will end up needing `dev-requirements.txt` and for those who don't they shouldn't get in the way.

dev-requirements.txt
```diff
@@ -0,0 +1,40 @@
+#### TESTING-RELATED PACKAGES ####
+
+# Checks style, syntax, and other useful errors.
+flake8==3.7.8
```
matt-gardner: Do we need to pin these?

schmmd: I don't know--but these are just copied from the old `requirements.txt`.  It'd be nice not to pin `black` and `flake8` (which shouldn't change much).  I'm going to merge this PR and unpin `black` and `flake8` in a follow up PR.

My guess is that we need to pin `mypy`.

PR #3587, opened 2020-01-06
Unpin black and flake8 for development

schmmd:
These tools shouldn't change their behavior as they have rather strict definitions.  Ideally we would be able to use the latest bugfixes.


schmmd:
I just created this because I was curious.  While I think it's fine to merge (the changes are fairly minimal) I'm not particularly passionate about it either.

schmmd:
We're seeing user confusion due to our outdated version of `black` so I think we should just merge this.

PR #3593, opened 2020-01-07
No longer requires a predictor to be supplied when doing prediction

### LINKED ISSUE #1675: make `allennlp predict` work without a predictor (if you use a dataset reader)
We now have a `--use-dataset-reader` flag.  It seems like this is pretty much always what someone would want to use if they are running things from the command line.  If you look at the code, using `--use-dataset-reader` means you go to `Predictor.predict_instances`, which has a default implementation in the base class (the only subclass implementations add stuff that's specific to the demo).  We don't need a `Predictor` at all to handle this use case.

#1674 implements a `Predictor` for the sole purpose of making JSON predictions with `allennlp predict`, but the code just re-implements dataset reader logic, and it would almost certainly be used with a `--use-dataset-reader` flag anyway.  So why do we require a `Predictor` for running `allennlp predict`?  It seems like we only really need it for demos.

ZhaofengWu:
Fixes #1675. When a predictor is not given, we first try to find a predictor in the DEFAULT_PREDICTORS list, and if one is not found, we use the base class' default implementation (i.e. `Predictor.predict_instance` or `Predictor.predict_json`).


ZhaofengWu:
That is correct.

PR #3594, opened 2020-01-08
Be robust to not having torch.distributed.

brendan-ai2:
- One can't call `distributed.is_initialized()` if `not distributed.is_available()`.
- This caused test failures on macOS High Sierra.


brendan-ai2:
Found this after you asked me to run `pytest` locally. Everything is green for me now, so the issues you hit are probably some annoying combination of OS/hardware/dependency differences. :/

brendan-ai2:
Yep, 3.6 for now. Ensuring everything is golden with 3.7 should definitely be in that 1.0 milestone.

brendan-ai2:
Thanks for the review!

PR #3595, opened 2020-01-08
Fix HotFlip padding error in ``_make_embedder_input" module 

c4n:
- Sometime if all the tokens are very short you might get this error from CNN encoder:
RuntimeError: Calculated padded input size per channel: (2). Kernel size: (3). Kernel size can't be greater than actual input size allennlp
- this PR should be able to fix it.


c4n:
I have no idea why it failed codecov/patch. :(

schmmd:
@c4n don't worry about the code coverage test.  I has a lot of false positives which are confusing.

PR #3596, opened 2020-01-09
Factor out a single function I need in allennlp_reading_comprehension

dirkgr:
I thought I needed more changes, but right now this is all.


brendan-ai2:
Incidentally, I've been seeing that nonsensical change too, so maybe we should just give up and merge it...

PR #3603, opened 2020-01-10
Guess sorting keys when none are given to BucketIterator

matt-gardner:
In refactoring the TokenIndexer stuff, I had to change how padding keys are determined, to make things a bit more sane.  That change makes figuring out which keys to use even more annoying than it was.  As I was thinking of how to fix that, it occurred to me that we could just guess which key to use and be right basically every time, by just finding the padding key with the longest value.

One potential failure case is when you have a large embedding matrix that's preprocessed (e.g., spacy vectors) and we use that length instead of the number of tokens, or something.  That's a rare case, and we could still handle it by taking the top k keys instead of the top 1, as the top one should be equal for all instances.

I guess we could detect variance of the keys and use that instead of just the max length...  But I'd vote to worry about that if we run into cases where this simpler logic fails.  My guess is that this does the right thing basically all of the time.

### DIFF
--- a/allennlp/data/iterators/bucket_iterator.py
+++ b/allennlp/data/iterators/bucket_iterator.py
@@ -159,8 +159,8 @@ class BucketIterator(DataIterator):
         instances_with_lengths.sort(key=lambda x: x[0])
         return [instance_with_lengths[-1] for instance_with_lengths in instances_with_lengths]
 
-    def _guess_sorting_key(self, instances: List[Instance]) -> None:
-        max_length = 0
+    def _guess_sorting_keys(self, instances: List[Instance]) -> None:
+        max_length = 0.0
         longest_padding_key: Tuple[str, str] = None
         for instance in instances:
             instance.index_fields(self.vocab)
--- a/allennlp/data/iterators/bucket_iterator.py
+++ b/allennlp/data/iterators/bucket_iterator.py
@@ -135,7 +135,9 @@ class BucketIterator(DataIterator):
         ``(field_name, padding_key)`` tuples.
         """
         if not self._sorting_keys:
+            logger.info("No sorting keys given; trying to guess a good one")
             self._guess_sorting_keys(instances)
+            logger.info(f"Using {self._sorting_keys} as the sorting keys")
         instances_with_lengths = []
         for instance in instances:
             # Make sure instance is indexed before calling .get_padding
@@ -171,7 +173,10 @@ class BucketIterator(DataIterator):
                         max_length = length
                         longest_padding_key = (field_name, padding_key)
         if not longest_padding_key:
-            raise ValueError(
-                "Found no field that needed padding; why are you using a bucket iterator?"
+            # This shouldn't ever happen (you basically have to have an empty instance list), but
+            # just in case...
+            raise AssertionError(
+                "Found no field that needed padding; we are surprised you got this error, please "
+                "open an issue on github"
             )
         self._sorting_keys = [longest_padding_key]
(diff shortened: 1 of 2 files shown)


allennlp/data/iterators/bucket_iterator.py
```diff
@@ -164,3 +127,51 @@ def _create_batches(self, instances: Iterable[Instance], shuffle: bool) -> Itera
...
+            instance.index_fields(self.vocab)
+            padding_lengths = cast(Dict[str, Dict[str, float]], instance.get_padding_lengths())
+            for field_name, field_padding in padding_lengths.items():
+                for padding_key, length in field_padding.items():
+                    if length > max_length:
+                        max_length = length
+                        longest_padding_key = (field_name, padding_key)
+        if not longest_padding_key:
+            raise ValueError(
+                "Found no field that needed padding; why are you using a bucket iterator?"
```
DeNeutoy: This should probably be a `ConfgigurationError`, as it explicitly relates to a poorly configured config. Also, make the error message better. This one assumes knowledge - instead, advise what someone should use, rather than just telling them they're doing something wrong.

allennlp/data/iterators/bucket_iterator.py
```diff
@@ -97,7 +65,7 @@ class BucketIterator(DataIterator):
 
     def __init__(
         self,
-        sorting_keys: List[Tuple[str, str]],
+        sorting_keys: List[Tuple[str, str]] = None,
```
DeNeutoy: FYI this now means that the `BasicIterator` can be deprecated and just subclassed with this.

allennlp/data/iterators/bucket_iterator.py
```diff
@@ -164,3 +127,51 @@ def _create_batches(self, instances: Iterable[Instance], shuffle: bool) -> Itera
                 batches.insert(0, last_batch)
 
             yield from batches
+
+    def _sort_by_padding(self, instances: List[Instance]) -> List[Instance]:
+        """
+        Sorts the instances by their padding lengths, using the keys in
+        ``sorting_keys`` (in the order in which they are provided).  ``sorting_keys`` is a list of
+        ``(field_name, padding_key)`` tuples.
+        """
```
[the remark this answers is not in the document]
matt-gardner: Done.

allennlp/data/iterators/bucket_iterator.py
matt-gardner: I made it an assertion error instead, with a direction to open an issue on github; it's basically impossible to get there, and it means I made a wrong assumption somewhere in the logic.

allennlp/data/iterators/bucket_iterator.py
matt-gardner: Yes, agreed, we can remove the BasicIterator entirely for 1.0.

matt-gardner: Well, we might want to add an option to this one to skip sorting, just in case you have data that really doesn't need it and don't want to waste computation with a useless sort.

brendan-ai2: Yes, I think we should retain an option to not sort. (Whether it's a separate iterator or a flag.) This is a bit of an edge case, but when debugging perf issues the bucket iterator really complicated understanding what was going on by introducing long periods of sorting. I suspect there are other cases where not sorting will be desired -- perhaps when your model is very sensitive to input lengths. (One could increase padding_noise excessively, but then we'll just be wasting compute sorting the noise.)

PR #3611, opened 2020-01-14
fixed the fix-endpoint_span_extractor issue

### LINKED ISSUE #3416: EndpointSpanExtractor should check that input matches input_dim
Currently, the EndpointSpanExtractor will happily take input that doesn't match its passed-in `input_dim` if the exclusive span indices are not being used. Feels like there should maybe be a check somewhere for this...https://github.com/allenai/allennlp/blob/master/allennlp/modules/span_extractors/endpoint_span_extractor.py

nadgeri14:
Fixes #3416. Check for dimensions in EndpointSpanExtractor when use_exclusive_start_indices = False.

### DIFF
--- a/allennlp/modules/span_extractors/endpoint_span_extractor.py
+++ b/allennlp/modules/span_extractors/endpoint_span_extractor.py
@@ -107,7 +107,11 @@ class EndpointSpanExtractor(SpanExtractor):
             span_ends = span_ends * span_indices_mask
 
         if not self._use_exclusive_start_indices:
-            assert sequence_tensor.size(-1) == self._input_dim,"dimension mismatch expected ({e}) received ({r})".format(e=sequence_tensor.size(-1), r=self._input_dim)
+            if sequence_tensor.size(-1) != self._input_dim:
+                raise ValueError(
+                    f"Dimension mismatch expected ({sequence_tensor.size(-1)}) "
+                    f"received ({self._input_dim})."
+                )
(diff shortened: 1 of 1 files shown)


matt-gardner:
@nelson-liu, do you mind if I assign this to you?  You have the most context on this PR.  It looks to me like the logic is fine, but you'd know for sure.

@nadgeri14, this is failing CI because the line you added is too long.  Can you break it up into multiple lines?  You'll also want to run `black` on the file, to be sure it meets our formatting requirements.  I'd also recommend using `f""` strings instead of `"".format()`.

nadgeri14:
@matt-gardner Thank you for reviewing the PR :) I will make the necessary changes as suggested. 

@matt-gardner @nelson-liu  **What is the best method**: Should I update the existing PR or create a new PR

Thanks for the help :D

matt-gardner:
Just update this PR.

nelson-liu:
instead of using assert, can you use an `if` block and raise a `ValueError`?

nadgeri14:
@nelson-liu  Yea, I will change it for sure :) 
What should be an idea error message, for now, I have written it as: 
**Dimension mismatch expected ({input_size}) received ({input_dimension}).**

nelson-liu:
Thanks @nadgeri14 , much appreciated!!

nadgeri14:
@nelson-liu Thankyou, would not have been possible without your help :D

PR #3616, opened 2020-01-14
make matplotlib optional

### LINKED ISSUE #2459: remove matplotlib dependency (just don't produce plots if it isn't available)
I heard from someone who would like to run allennlp but without taking the flask dependency. Unfortunately, now that `allennlp configure` is a top-level command, it's impossible to run the `allennlp` command without flask being imported.

It should be a simple (if slightly ugly) fix to make this import conditional on flask being available.

DeNeutoy:
Fixes #2459

### DIFF
--- a/allennlp/commands/find_learning_rate.py
+++ b/allennlp/commands/find_learning_rate.py
@@ -353,10 +353,16 @@ def _smooth(values: List[float], beta: float) -> List[float]:
 
 
 def _save_plot(learning_rates: List[float], losses: List[float], save_path: str):
-    import matplotlib
 
-    matplotlib.use("Agg")  # noqa
-    import matplotlib.pyplot as plt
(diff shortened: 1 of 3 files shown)


brendan-ai2:
This change appears to have broken the "Library Master" build. Specifically the test that checks whether we can install from pip. See http://build.allennlp.org/viewLog.html?buildId=24194&buildTypeId=AllenNLP_AllenNLPCommits&tab=buildLog.

I'm not sure what the right fix is. Just rolling back seems okay to me, but it's not clear to me how annoying `matplotlib` is for people. Otherwise we could parameterize `setup.py` to optionally install `matplotlib`. Installing `dev-requirements.txt` in the Docker image is probably not ideal because then we're not really testing the `pip install`.

Thoughts?

Issue #3570, opened 2020-01-01
Tests don't run on clean master checkout


EmilStenstrom:
**Describe the bug**
When cloning the allennlp repository an running `pytest` i get test errors. This makes it harder to contribute since I'm not sure if I broke something or not. 

<details>
<summary>[Full output of `pytest -vv`]</summary>

```
================================================================================= FAILURES ==================================================================================
__________________________________________________________ TestPosTagIndexer.test_count_vocab_items_uses_pos_tags ___________________________________________________________

self = <allennlp.tests.data.token_indexers.pos_tag_indexer_test.TestPosTagIndexer testMethod=test_count_vocab_items_uses_pos_tags>

    def test_count_vocab_items_uses_pos_tags(self):
        tokens = self.tokenizer.tokenize("This is a sentence.")
        tokens = [Token("<S>")] + [t for t in tokens] + [Token("</S>")]
        indexer = PosTagIndexer()
        counter = defaultdict(lambda: defaultdict(int))
        for token in tokens:
            indexer.count_vocab_items(token, counter)
        assert counter["pos_tokens"] == {"DT": 2, "VBZ": 1, ".": 1, "NN": 1, "NONE": 2}
    
        indexer._coarse_tags = True
        counter = defaultdict(lambda: defaultdict(int))
        for token in tokens:
            indexer.count_vocab_items(token, counter)
>       assert counter["pos_tokens"] == {"AUX": 1, "PUNCT": 1, "DET": 2, "NOUN": 1, "NONE": 2}
E       AssertionError: assert defaultdict(<..., 'PUNCT': 1}) == {'AUX': 1, 'DE...NOUN': 1, ...}
E         Common items:
E         {'DET': 2, 'NONE': 2, 'NOUN': 1, 'PUNCT': 1}
E         Left contains 1 more item:
E         {'VERB': 1}
E         Right contains 1 more item:
E         {'AUX': 1}
E         Full diff:
E         + {'AUX': 1, 'DET': 2, 'NONE': 2, 'NOUN': 1, 'PUNCT': 1}
E         - defaultdict(<class 'int'>,
E         -             {'DET': 2,
E         -              'NONE': 2,
E         -              'NOUN': 1,
E         -              'PUNCT': 1,
E         -              'VERB': 1})

allennlp/tests/data/token_indexers/pos_tag_indexer_test.py:27: AssertionError
--------------------------------------------------------------------------- Captured stdout call ----------------------------------------------------------------------------
[...]

PR #3619, opened 2020-01-14
Skip a test that we know will fail

dirkgr:
This should fix, or at least go somewhat towards fixing, #3570.


schmmd:
Makes sense to skip tests that are reliably failing.  @brendan-ai2 should we cut an issue to fix this test?

brendan-ai2:
We'll be deleting this code soon. I wouldn't worry about making an issue for it.

schmmd:
👍

brendan-ai2:
Somewhat tangentially, we don't reliably support macOS since we don't have CI for it. This is embarrassing. At some point it would be great to look into getting a macOS machine in the cloud to run our tests.

DeNeutoy:
@dirkgr - for PRs like this feel free to just squash and merge, we all need that extra spice in our lives

schmmd:
👍 to @DeNeutoy 's comment.  We only brought the master merge back because--without it--GitHub doesn't give you the option to merge in master (which is useful sometimes).

PR #3627, opened 2020-01-15
Add a Lazy[] annotation for simpler FromParams construction

matt-gardner:
High-level idea: we often have a custom `from_params` method because of sequential dependencies in the arguments to a function: e.g., the `Trainer` needs to have constructed the `Model` before it can pass its parameters to an `Optimizer`, and the `TrainerBase` (which runs the whole training loop) does a ton of complicated things with data processing before passing things to the actual `Trainer` object.

To remove the need for custom `from_params` methods in these cases, this PR introduces a `Lazy` type annotation, which you can use as follows:

```python
@TrainerBase.register("default", constructor="from_partial_objects")
class Trainer(TrainerBase):
    @classmethod
    def from_partial_objects(cls, model: Model, ..., optimizer: Lazy[Optimizer]) -> Trainer:
        optimizer = optimizer.construct(model=model)
        return cls(model, optimizer, ...)
```

The benefit is that now we have a method that has arguments we can document, instead of a cryptic `Params` argument where you have to look at the code to figure out what's supposed to be inside it.  The actual logic inside of `from_partial_objects` is incredibly minimal in this case (you can see it in this PR).

-----------------------

I was going to stop there with this initial PR, as this is a minimal demonstration and (I think) an obvious improvement over what was there before.  The trouble is that because of how `TrainerBase` interacts with `Trainer` to run the training loop, this doesn't work in its current state.  I have to also rework how `TrainerBase` does its own custom `from_params` in order for this to work.  This raises a larger question: I'm pretty sure I can dramatically simplify how the whole `TrainerBase` works with this - no need for `TrainerPieces`, no cryptic popping of arguments from a `Params` object anywhere, and much easier to follow training logic, that's also easier to replicate and modify in your own code if you want to.  But that's a somewhat large project, and before I started on it I wanted to get feedback on this idea.

The gist would be that I can make a `Train` class, or something, that has a constructor with lazy parameters that exactly match all of the top-level keys we accept in a config file:

```python
class Train(FromParams):
    @classmethod
    def from_partial_objects(
        cls,
        dataset_reader: DatasetReader,
        iterator: DataIterator,
        model: Lazy[Model],
        trainer: Lazy[Trainer],
        validation_dataset_reader: DatasetReader = None,
        validation_iterator: DataIterator = None,
    ) -> Train:
        ...
```

This method has the logic that's currently buried in `TrainerPieces.from_params`, and will push as much code into methods that don't have `Lazy` arguments as possible, so that code is re-usable for people who do not choose to use our `FromParams` pipeline.  And the `train` command would instantiate this object `from_params`, then call `train.run()`.  Someone who's writing their own script would just create their objects, pass them to `Train.__init__`, and call `train.run()`.

Does this make sense?  Any objections?

------------------------

Things I broke in this implementation, which we need to be sure work (probably there will be a failing test that forces me to do this, or remove the test and open an issue):

- [x] The --recover flag for training
- [ ] Extending vocabularies.  I think the right thing to do here is have a `Model` type that is `"from_archive"`, which gets passed the vocabulary.  It can use the vocabulary to extend itself if it wants.  We should also have a `Vocabulary` type that is `"empty"` or `"none"`, for cases where you're just loading a model and don't want to iterate through the data and compute a vocab for no reason.


allennlp/training/trainer.py
```diff
@@ -788,57 +811,26 @@ def from_params(  # type: ignore
...
-            momentum_scheduler = None
+            learning_rate_scheduler_ = None
 
-        if "checkpointer" in params:
-            if (
-                "keep_serialized_model_every_num_seconds" in params
-                or "num_serialized_models_to_keep" in params
-            ):
-                raise ConfigurationError(
-                    "Checkpointer may be initialized either from the 'checkpointer' key or from the "
```
matt-gardner: Note that I removed this duplicate means of specifying the parameters - you now have to pass the arguments to the checkpointer.

DeNeutoy:
I like the idea - one question is how many constructors do we have that actually have this sequential dependency problem? Also I think `LazilyConstructed` might be a better name for the class.

matt-gardner:
I just went through all of our custom `from_params` methods, and there are two main cases where this comes up:

1. Things related to training, almost always because of interactions with an optimizer or needing access to model parameters.
2. Top-level scripts that pop things from a config file directly, instead of just implementing plain python logic.

I was aiming to just solve the first case first, and this simple PR does it, it's just that the starting place I picked also requires solving the second case.

On the name, `LazilyConstructed` feels overly verbose to me, and that's basically the definition of `Lazy` anyway - the thing that always makes things lazy is delaying execution of something, and if it's on a type it implies it's the construction of an object.  What else might you think of if you see `Lazy[Optimizer]` as a type annotation?

matt-gardner:
And thanks, unless anyone else has some objections to this idea, I'm going to move forward on handling the rest of what I mentioned, probably this afternoon or tomorrow morning.

bryant1410:
> On the name, `LazilyConstructed` feels overly verbose to me, and that's basically the definition of `Lazy` anyway - the thing that always makes things lazy is delaying execution of something, and if it's on a type it implies it's the construction of an object. What else might you think of if you see `Lazy[Optimizer]` as a type annotation?

Agree, that's the only definition that comes to my mind of "lazy" in this context.

I think this change does improve the state of things, so I like it and approve it. However, I think there's an underlying issue going on here, which I'm gonna elaborate on shortly.

brendan-ai2:
I like this change as well. Strikes me as a nice cleanup.

bryant1410:
> I think this change does improve the state of things, so I like it and approve it. However, I think there's an underlying issue going on here, which I'm gonna elaborate on shortly.

`FromParams` (and its subclasses' constructors) is playing a double duty here: it checks what are the valid children elements for the configuration file of a given element from a function args and constructs a class instance from them. The problem is that the class dependencies may go beyond those specified within the configuration (e.g., `Optimizer` needs a `Model`, `Trainer` needs a `Model`, `Optimizer` needs a `LearningRateScheduler`). So, we're kinda forcing the construction of the objects to follow the hierarchy established in the syntax of the configuration file, which may be different from the actual dependency graph. In consequence, there's a need to pass along either `Params` objects with the args that are gonna be necessary later on or lazy objects indicating other things that are needed.

It makes sense with the change that there's now a `constructor_to_call` and a `constructor_to_inspect`. It allows to separate the actual dependencies from the configuration file hierarchy. However, it only partially solves it: the actual dependencies need to be obtained somewhere. This makes me remember Dependency Injection in some ways. I believe a neater solution to this problem, albeit maybe more complicated (because the dependencies and the configuration file hierarchy most of the times are the same), is to separate these two things further. I'm not sure how, but the general idea would be to parse the configuration file on one side and separately construct the objects only based on their dependencies in whatever order (maybe based on the dependency graph), using data from the configuration file.

matt-gardner:
@bryant1410, what you're saying is similar to what was at the top of #3389. I don't think we need to go as far as that, because I don't think we have actual problems that require us to go that far.

Thanks for the comments everyone, I will start work on the rest of this tomorrow morning.

allennlp/tests/training/multi_task_trainer_test.py
```diff
@@ -240,40 +240,33 @@ def train(self) -> Dict:
...
-        train_file_paths = params.pop("train_file_paths").as_dict()
 
-        datasets = {name: reader.read(train_file_paths[name]) for name, reader in readers.items()}
+        datasets = {
+            name: reader.read(train_file_paths[name])
+            for name, reader in train_dataset_readers.items()
+        }
 
         instances = (instance for dataset in datasets.values() for instance in dataset)
         vocab = Vocabulary.from_params(Params({}), instances=instances)
```
bryant1410: `Vocabulary.from_instances`?

allennlp/commands/train.py
```diff
@@ -485,26 +445,136 @@ def _train_worker(
...
+                    raise ConfigurationError(f"invalid 'dataset_for_vocab_creation' {key}")
+
+        instance_generator = (
+            instance
+            for key, dataset in datasets.items()
+            if not datasets_for_vocab_creation or key in datasets_for_vocab_creation
+            for instance in dataset
+        )
+
+        vocabulary_ = vocabulary.construct(instances=instance_generator)
```
bryant1410: I think having both the variable and the variable with the same name but with the underscore suffix is a bit confusing. What about renaming the former to `partial_vocabulary`?

matt-gardner: I agree that this isn't optimal. But we can't rename the method argument, because it has to match what we want to see in the config file.  We can only rename the constructed one.  Suggestions welcome.

bryant1410: Oh, I see. What about renaming the other variable? To something like `constructed_vocabulary`?

matt-gardner:
Ok, I think this is ready for review.  @DeNeutoy, can you take a look?

What I see as the key benefit of this PR is that logic and parameters that were buried inside of `from_params` methods are now documented.  I'm not tied to the `from_partial_objects` name that I used pretty consistently; suggestions for a better name are welcome.  But those methods would show up in our docs, and we can have docstrings (probably the main class docstring?) that clearly says that this is the function that we will look at for building things `from_params`, so those are the keys that you need / are available in a config file.

One thing that came up was a bit of an awkward separation between the top-level train command and the `Trainer` object.  There are kind of two separate "trainers" here - one is the thing that defines the script being run (`TrainModel`), and one is the thing that runs the actual training loop (`TrainerBase` and `Trainer`).  Especially on the GAN / multi-task stuff, I think we really want to modify *both* of these, not just the `Trainer`.  It's a lot more sane.

You could imagine pretty easily using this to rework how we do all of our commands.  Not that we necessarily *should*, it was just something I was thinking about as I moved logic into the `TrainModel` object - it seemed like it was only going halfway to making things cleaner.  I definitely think this is an improvement over what was there before, it just could go farther than I did here.

PR #3628, opened 2020-01-15
[WIP] remove sorting in encoders

DeNeutoy:
`torch.nn.utils.rnn.pack_padded_sequence` now accepts unsorted sequence tensors, which means we no-longer have to sort them upfront, meaning we can remove loads of hideously complicated indexing code 👍 

Changes:

- we no longer sort input tensors by sequence length
- we no longer remove tensors which are completely padded. Instead, we (slightly hackily) set their sequence length to 1, and then mask out the output.
- we no longer have to manually append padding onto the sequence length dimension, because this can now be specified in the `total_length` argument to `pad_packed_sequence`.

### DIFF
--- a/allennlp/modules/encoder_base.py
+++ b/allennlp/modules/encoder_base.py
@@ -1,8 +1,8 @@
-from typing import Tuple, Union, Optional, Callable, Any
+from typing import Tuple, Union, Optional, Callable
 import torch
 from torch.nn.utils.rnn import pack_padded_sequence, PackedSequence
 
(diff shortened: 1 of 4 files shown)


matt-gardner:
Does not sorting have any performance implications?  Presumably the logic on our side is faster, because we don't do as much. What about on the pytorch side?

DeNeutoy:
Good point, I am checking on the pytorch slack now

DeNeutoy:
Turns out this is infeasible because we have several classes which assume the packed sequence is sorted. Damn.

PR #3632, opened 2020-01-16
Remove pypandoc

bryant1410:
It's not used anymore.


matt-gardner:
Are you sure? What do we do with our README getting onto pypi now?

bryant1410:
Yeah. I used to do the same. PyPI supports Markdown since some time ago. See #1201, and see:

https://github.com/allenai/allennlp/blob/master/setup.py#L88

And also see the changed line was the only occurrence of it (`grep -Ire pandoc`).

PR #3633, opened 2020-01-16
Allow FromParams to handle **kwargs passed to superclass constructors

### LINKED ISSUE #3614: FromParams: Look at superclass constructor for additional arguments when **kwargs is given
It should be possible to have `FromParams` get arguments from the superclass, to make it easier to add parameters in a base class and not require modifying all subclasses to pass it through (e.g., the `lazy` parameter for `DatasetReaders`).  I think we can do this just by looking for a base class constructor, checking its arguments, and adding any unique ones to the list of things we're trying to construct.  We can have this behavior trigger only when there's a `**kwargs` argument to the object we're trying to construct, and only when it's an `__init__` method (instead of a `@classmethod` that's been registered as a constructor).  Shouldn't be too hard to implement this, and it would make the code easier and more extensible in a few places.

matt-gardner:
Fixes #3614. In implementing #3627, I ran into an issue with how dataset caching is handled.  To fix it the right way, I needed to implement this first.  It was easier than I thought it might be.

The implication of this change in the dataset reader caching is that we will no longer auto-create a cache directory for you, but you also won't have to have special parameters passed to the `train` command - you just specify a cache directory as an argument to your dataset reader, and that's it.  I'm guessing @DeNeutoy will prefer this way, as he didn't like the command-line arguments to the train command in the first place.


allennlp/common/from_params.py
```diff
@@ -128,23 +128,46 @@ def create_kwargs(
...
+    # constructor.
+    has_kwargs = False
+    for param in parameters.values():
+        if param.kind == param.VAR_KEYWORD:
+            has_kwargs = True
+    if has_kwargs:
+        # "mro" is "method resolution order".  The first one is the current class, the next is the
+        # first superclass, and so on.  Taking the first superclass should work in all cases that
+        # we're looking for here.
+        superclass = cls.mro()[1]
```
DeNeutoy: Is this fragile if someone does:

```
class(OtherType, RegisterableType):
```

And `OtherType` also has a constructor? Equally I don't know a great way to improve it, maybe checking if the superclass inherits from Registrable or something?

matt-gardner: Yes, that would probably break this.  My vote is to wait to figure out how to handle that case until someone actually runs into it; seems pretty unlikely that that would happen.

allennlp/tools/EVALB/evalb.dSYM/Contents/Info.plist
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "https://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
```
bryant1410: Is this file generated automatically?

matt-gardner: Ugh, that's an annoying file.  I didn't realize I changed it.  I'm going to remove it and add it to .gitignore.

allennlp/common/from_params.py
```diff
@@ -128,23 +128,46 @@ def create_kwargs(
...
+        superclass_signature = inspect.signature(superclass.__init__)  # type: ignore
+        for param_name, param in superclass_signature.parameters.items():
+            if param_name == "self":
+                continue
+            parameters[param_name] = param
```
mahnerak: I think overwriting is not a good idea here. The correct way is to infer only those parameters that couldn't be inferred at the first step.
An example:
```python
class Some(DatasetReader):
    def __init__(self, some_param: str, **kwargs):
        super().__init__(**kwargs)
        self._some_param = some_param

class B(A):
    def __init__(self, some_param: str = 'default_value', **kwargs):
        super().__init__(some_param=some_param, **kwargs)
```

allennlp/common/from_params.py
mahnerak: Maybe Multiple Inheritance is not what you want to handle here, but are you sure this inference shouldn't be recursive? What if the superclass constructor has `kwargs`?

allennlp/common/from_params.py
mahnerak: PR #3653 tries to fix this issue.

PR #3634, opened 2020-01-16
Add PretrainedTransformerMismatched{Indexer, Embedder}

ZhaofengWu:
These classes are useful when the dataset is pre-tokenized and we want to compute loss and metrics using the original tokens. We tokenize each word separately, concatenate the wordpieces, feed them to the transformer model, and then pool wordpiece representations to get word representations.


allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -29,27 +31,50 @@ class PretrainedTransformerIndexer(TokenIndexer):
...
+        Should be set to the same value as the ``intra_word_tokenized`` option on the
+        :class:`PretrainedTransformerEmbedder`.
     """
 
     def __init__(
-        self, model_name: str, namespace: str = "tags", token_min_padding_length: int = 0
+        self,
+        model_name: str,
+        namespace: str = "tags",
+        num_added_start_tokens: int = 1,
```
brendan-ai2: Earlier you mentioned that you could auto-detect these, but since they were needed in the embedder we decided to specify them explicitly. However, I think you managed to avoid that in this version of the code. Is it possible to leave this out now?

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -23,7 +34,59 @@ def get_output_dim(self):
...
+        wordpiece_mask: torch.LongTensor = None,
+        offsets: torch.LongTensor = None,
     ) -> torch.Tensor:  # type: ignore
+        """
+        # Parameters
+
+        token_ids: torch.LongTensor
+            Shape: [batch_size, num_wordpieces].
+        mask: torch.LongTensor
+            Shape: [batch_size, num_orig_tokens if self._intra_word_tokenized else num_wordpieces].
```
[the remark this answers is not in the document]
matt-gardner: This is one reason why I might have split this into two classes instead of having it be a flag on one class.  I can see arguments both ways, but I'm somewhat partial to having separate collections of (Tokenizer, Indexer, Embedder) for each of the three cases we discussed, instead of different flags.

(Sorry, didn't make it to reviewing this PR today, just saw this discussion.  I'll try to get to it tomorrow, but no promises; Fridays are really busy for me.)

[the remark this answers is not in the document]
ZhaofengWu: In the beginning I didn't think it warrants a standalone class, not foreseeing a lot of complications. But it looks like that's necessary for this to be clean, so I will try to make that change now. I will take you both off the reviewer list for now and re-request review when I'm done.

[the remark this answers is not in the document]
ZhaofengWu: (If I can figure out how to move people off the reviewer list)

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -23,7 +34,59 @@ def get_output_dim(self):
...
+
+        transformer_mask = wordpiece_mask if self._intra_word_tokenized else mask
+        # Shape: [batch_size, num_wordpieces, embedding_size].
+        embeddings = self.transformer_model(input_ids=token_ids, attention_mask=transformer_mask)[0]
+
+        if self._intra_word_tokenized:
+            batch_size, _, embedding_size = embeddings.size()
+            num_orig_tokens = offsets.size(1)
+
+            orig_embeddings = torch.FloatTensor(batch_size, num_orig_tokens, embedding_size)
```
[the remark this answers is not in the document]
ZhaofengWu: I don't have any existing tensor with the same shape, but yes I did forget about device. Will change.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
ZhaofengWu: The embedder will need those variables when I submit the long sequence wrapping PR. It'd be a little troublesome to add back that logic now and remove it in that PR. With that said, if we think the auto-detection logic is better, we can always add that back, perhaps as a standalone PR, to replace config options.

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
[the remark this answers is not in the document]
brendan-ai2: Ah, the issue with `get_text_field_mask` makes sense. Thanks for explaining. I might wait for Matt to give his first pass before attempting to break apart the code. But if you can avoid duplicating the other logic, I certainly don't object.

And you can just ping us explicitly. No need to modify the reviewer list.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
matt-gardner: The interface between the indexer and the embedder can be arbitrary - you can pass whatever you need.  Based on the `assert` below, it looks like you can compute this automatically.  I would definitely say you should remove these as parameters here.

We sometimes have things like this be parameters on the _modeling_ side, where we need to know an embedding dimension up front, for example.  So that would mean having a parameter for it in the `Embedder`.  But there's no reason to have it as a parameter here, even if you need it in the `Embedder`.  But even better than making it a parameter in the `Embedder`, you can just compute it here and pass it through in your tensor dictionary, and the embedder can get the right value that way.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -29,27 +31,52 @@ class PretrainedTransformerIndexer(TokenIndexer):
         We use a somewhat confusing default value of ``tags`` so that we do not add padding or UNK
         tokens to this namespace, which would break on loading because we wouldn't find our default
         OOV token.
+    intra_word_tokenization: ``bool``, optional (default=False)
```
matt-gardner: This parameter is duplicated in the docstring.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -89,15 +127,42 @@ def tokens_to_indices(
...
+            offsets = [
+                (start + self._num_added_start_tokens, end + self._num_added_start_tokens)
+                for start, end in offsets
+            ]
 
-        return {"token_ids": indices, "mask": attention_mask}
+        output = {"token_ids": indices, "mask": mask}
+        if self._intra_word_tokenization:
+            output["wordpiece_mask"] = [1] * len(indices)
+            output["offsets"] = offsets  # type: ignore
```
matt-gardner: Why do you need a `# type: ignore` here?  You probably just need a more permissive type annotation on line 138 - mypy probably inferred a too-specific type from the inline dictionary instantiation.

matt-gardner: Oh, nevermind, is this because of my cheating with `IndexedTokenList`?

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -29,27 +31,52 @@ class PretrainedTransformerIndexer(TokenIndexer):
...
+        num_added_end_tokens: int = 1,
+        intra_word_tokenization: bool = False,
     ) -> None:
         super().__init__(token_min_padding_length)
         self._namespace = namespace
-        self._tokenizer = AutoTokenizer.from_pretrained(model_name)
-        self._padding_value = self._tokenizer.convert_tokens_to_ids([self._tokenizer.pad_token])[0]
-        logger.info(f"Using token indexer padding value of {self._padding_value}")
+        # add_special_tokens=False for intra_word_tokenization
+        self._tokenizer = PretrainedTransformerTokenizer(model_name, add_special_tokens=False)
```
matt-gardner: Why do you need our tokenizer, instead of just the huggingface one?  What does the allennlp one get you?

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -23,7 +34,64 @@ def get_output_dim(self):
...
+        embeddings = self.transformer_model(input_ids=token_ids, attention_mask=transformer_mask)[0]
+
+        if self._intra_word_tokenized:
+            batch_size, _, embedding_size = embeddings.size()
+            num_orig_tokens = offsets.size(1)
+
+            orig_embeddings = torch.FloatTensor(
+                batch_size, num_orig_tokens, embedding_size
+            ).to(device)
+            for batch_idx in range(batch_size):  # TODO: do we have to use loops?
```
matt-gardner: I'm going to wait to comment on this TODO until I see it split into a separate file.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
ZhaofengWu: Oops typo.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
brendan-ai2: @matt-gardner, wouldn't we need to pass along a full batch of these? I suppose that's likely not significant in terms memory, but it does seem a bit hacky. Maybe something we can use to inform future API improvements.

matt-gardner: You can make the interface between the indexer and the embedder whatever you want, including controlling [how things get batched](https://github.com/allenai/allennlp/blob/52bece613672d8b289b722ddc86731ded07cca3f/allennlp/data/token_indexers/token_indexer.py#L89).  You can just not batch this field.

matt-gardner: Hmm, I was too hasty, that just does padding, not batching.  But, I think the additional overhead is pretty minor here, relative to requiring a user to figure out the specifics of the what the underlying tokenizer is doing just to use the new model.

ZhaofengWu: Huggingface gives us `num_added_tokens` which a sum of my `num_added_start_tokens` and `num_added_end_tokens`. Those values are not provided. The `assert` only acts as a checksum. Yes there is a way (read: hack) I can auto-determine these two values by feeding a dummy sentence to the tokenizer. However, both the indexer and embedder need to know these two values (not right now, but in a next PR coming very soon). I was not aware of a simple way to pass this information across, and I really don't want to repeat this hack in two different places.

matt-gardner: If we really care, we could allow the indexer to control batching itself; that would be a non-user-facing change that would give us a very slight gain in efficiency.  The `TextField` would just delegate batching to the underlying indexers.  I don't really think that's worth it, but we could if we wanted.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
ZhaofengWu: The indexer takes `Token`s as input, but (all) tokenizers take raw text. When doing subword tokenization I have to extract raw text from the tokens, tokenize, and then convert back to `Token`s. This last step is what our tokenizer is for. See `_intra_word_tokenize()` at the bottom of the file.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
ZhaofengWu: Ok, I will add back the auto-determination logic back in, and then pass that information to the embedder via `tokens_to_indices` output. On a related note, in the future if similar issues come up again, it might be nice to give the embedder a reference to the indexer in its `__init__`.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
matt-gardner: Ok, yeah, that makes sense.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
ZhaofengWu: I think mypy just infers the dict value type to be `List[int]`, either from the creation statement or using the return type, and is unhappy when I insert a value with type `List[Tuple[int, int]]`. I'm not sure if it has something to do with `IndexedTokenList`, although I wonder if it'd be nicer to be a little more generic with `IndexedTokenList` and allow it to be `Dict[str, List[Any]]`?

allennlp/data/token_indexers/pretrained_transformer_indexer.py
matt-gardner: In the original design of the precursor to this library, the indexer and the embedder were a single object.  There clearly is a strong relationship between these two, and having them share state can be helpful in some cases, like this one.  We decided, though, that decoupling the data processing from the model is more important and in the end cleaner overall.

On the auto-determination logic, it's probably worth at least a little thought on if this can be added upstream in the huggingface library, so that we can just call methods there that give us what we need.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
matt-gardner: Yeah, if it doesn't break other things, feel free to loosen the definition of `IndexedTokenList`, or use a `TypeVar`, or something.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
ZhaofengWu: Sounds good. I will first add that as a function in AllenNLP, and will submit a PR to huggingface if I got a chance.

ZhaofengWu:
@matt-gardner @brendan-ai2 ready for re-review.

allennlp/data/token_indexers/pretrained_transformer_pretokenized_indexer.py
```diff
@@ -0,0 +1,95 @@
...
+    `PretrainedTransformerPretokenizedEmbedder`.
+    """
+
+    def __init__(
+        self, model_name: str, namespace: str = "tags", token_min_padding_length: int = 0
+    ) -> None:
+        super().__init__(model_name, namespace, token_min_padding_length)
+        # add_special_tokens=False sicne we don't want wordpieces to be surrounded by special tokens
+        self._allennlp_tokenizer = PretrainedTransformerTokenizer(
+            model_name, add_special_tokens=False
```
ZhaofengWu: Do we want to support tokenizer options here, e.g. lowercasing? We'd have to add them to the __init__ args, which might increase tokenizer/indexer coupling.

allennlp/modules/token_embedders/pretrained_transformer_pretokenized_embedder.py
```diff
@@ -0,0 +1,58 @@
+from overrides import overrides
+import torch
+
+from allennlp.modules.token_embedders import PretrainedTransformerEmbedder, TokenEmbedder
+from allennlp.nn import util
+
+
+@TokenEmbedder.register("pretrained_transformer_pretokenized")
+class PretrainedTransformerPretokenizedEmbedder(PretrainedTransformerEmbedder):
```
matt-gardner: Again here, don't use inheritance.

allennlp/tests/data/token_indexers/pretrained_transformer_pretokenized_indexer_test.py
```diff
@@ -0,0 +1,21 @@
+from transformers.tokenization_auto import AutoTokenizer
+
+from allennlp.common.testing import AllenNlpTestCase
+from allennlp.data import Token, Vocabulary
+from allennlp.data.token_indexers import PretrainedTransformerPretokenizedIndexer
+
+
+class TestPretrainedTransformerIndexer(AllenNlpTestCase):
```
matt-gardner: The name of this class needs changing.

allennlp/tests/data/token_indexers/pretrained_transformer_pretokenized_indexer_test.py
```diff
@@ -0,0 +1,21 @@
...
+    def test_pretokenized_behavior(self):
```
matt-gardner: I don't think you're going to get enough coverage with just this test.

allennlp/tests/modules/token_embedders/pretrained_transformer_pretokenized_embedder_test.py
```diff
@@ -0,0 +1,54 @@
...
+        instance2 = Instance({"tokens": TextField(tokens2, {"bert": token_indexer})})
+
+        batch = Batch([instance1, instance2])
+        batch.index_instances(vocab)
+
+        padding_lengths = batch.get_padding_lengths()
+        tensor_dict = batch.as_tensor_dict(padding_lengths)
+        tokens = tensor_dict["tokens"]
+
+        assert tokens["bert"]["offsets"].tolist() == [
```
matt-gardner: Oh, you're testing it here.  This assert should live in the indexer test, shouldn't it?

allennlp/data/token_indexers/pretrained_transformer_pretokenized_indexer.py
```diff
@@ -0,0 +1,95 @@
...
+from allennlp.data.vocabulary import Vocabulary
+from allennlp.data.tokenizers.token import Token
+from allennlp.data.token_indexers import PretrainedTransformerIndexer, TokenIndexer
+from allennlp.data.token_indexers.token_indexer import IndexedTokenList
+from allennlp.data.tokenizers import PretrainedTransformerTokenizer
+
+logger = logging.getLogger(__name__)
+
+
+@TokenIndexer.register("pretrained_transformer_pretokenized")
```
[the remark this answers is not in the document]
ZhaofengWu: Ok, I was thinking "pretokenized" as in already tokenized into words but not wordpieces. I agree that it's not very accurate. But I think `Mismatched` might be a little generic and confusing. What is mismatched here? The indexer input and the transformer model input? Many other things could cause this mismatch though, e.g. I'm implementing a long sequence folding logic which splits (indexer input) [CLS] A B C D E F [SEP] into (embedder input) [ [CLS] A B C [SEP], [CLS] D E F [SEP] ]. This is also "mismatched" in a sense.

allennlp/data/token_indexers/pretrained_transformer_pretokenized_indexer.py
```diff
@@ -0,0 +1,95 @@
...
+    into wordpieces independently and concatenate them back together. Use it along with
+    `PretrainedTransformerPretokenizedEmbedder`.
+    """
+
+    def __init__(
+        self, model_name: str, namespace: str = "tags", token_min_padding_length: int = 0
+    ) -> None:
+        super().__init__(model_name, namespace, token_min_padding_length)
+        # add_special_tokens=False sicne we don't want wordpieces to be surrounded by special tokens
+        self._allennlp_tokenizer = PretrainedTransformerTokenizer(
```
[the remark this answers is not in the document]
ZhaofengWu: This is because the superclass has a `self._tokenizer` for the huggingface tokenizer, which is also used in this class to `build_inputs_with_special_tokens`, etc.

allennlp/data/token_indexers/pretrained_transformer_pretokenized_indexer.py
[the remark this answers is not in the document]
matt-gardner: "mismatched" here refers to mismatched tokenization and modeling. I think that accurately describes the problem and gives a hint to the reader about why this is more complicated case with harder logic, and that they might want to think twice about using it (because most use cases probably shouldn't use this anymore).  If we use that name, I would state this explicitly in the docstring.

allennlp/data/token_indexers/pretrained_transformer_pretokenized_indexer.py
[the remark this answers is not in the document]
matt-gardner: Ah, ok, so, getting rid of the inheritance will clarify that issue quite a bit.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -37,10 +39,13 @@ def __init__(
         super().__init__(token_min_padding_length)
         self._namespace = namespace
         self._tokenizer = AutoTokenizer.from_pretrained(model_name)
-        self._padding_value = self._tokenizer.convert_tokens_to_ids([self._tokenizer.pad_token])[0]
-        logger.info(f"Using token indexer padding value of {self._padding_value}")
         self._added_to_vocabulary = False
 
+        (
+            self._num_added_start_tokens,
```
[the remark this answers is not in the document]
ZhaofengWu: This is because they will very soon be used in this class when I submit the PR for the long sequence folding logic. If I put it in the subclass now, I'd have to move it here then. I'm fine with that though since you raised this point.

brendan-ai2:
(I'm going to give this a final pass once Zhaofeng has incorporated your feedback, Matt.)

allennlp/data/token_indexers/pretrained_transformer_pretokenized_indexer.py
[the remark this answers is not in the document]
ZhaofengWu: Ok let's used "mismatched" for now. I don't know whether "most use cases probably shouldn't use this" is true though. This should be applicable to many span- or token-based tasks (as opposed to sentence-based), e.g. NER, coref, QA, parsing, etc.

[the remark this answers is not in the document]
matt-gardner: The reason I say most use cases probably shouldn't use this is because you're very likely better off distributing your labels instead of having mismatched tokenization.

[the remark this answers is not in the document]
matt-gardner: Though I admit that at this point this is just intuition, not something with hard empirical evidence.

allennlp/data/token_indexers/pretrained_transformer_pretokenized_indexer.py
```diff
@@ -0,0 +1,95 @@
...
+        return output
+
+    @overrides
+    def as_padded_tensor_dict(
+        self, tokens: IndexedTokenList, padding_lengths: Dict[str, int]
+    ) -> Dict[str, torch.Tensor]:
+        offsets_tokens = tokens.pop("offsets")
+        offsets_padding_lengths = padding_lengths.pop("offsets")
+
+        tensor_dict = super().as_padded_tensor_dict(tokens, padding_lengths)
```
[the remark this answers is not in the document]
ZhaofengWu: What kind of tests are you proposing that I should add here?

allennlp/tests/modules/token_embedders/pretrained_transformer_pretokenized_embedder_test.py
ZhaofengWu: Can you elaborate on this and the coverage issue you mentioned above? I do check the `offsets` key in the indexer test. I'm just checking here again, just like `TestPretrainedTransformerEmbedder.test_end_to_end` also tests some indexer input, e.g. it has `assert tokens["bert"]["mask"].tolist() == X`.

allennlp/data/token_indexers/pretrained_transformer_pretokenized_indexer.py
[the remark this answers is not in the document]
matt-gardner: I hadn't looked at the tests when I wrote this, I was just thinking out loud.  The one end-to-end test that you have is maybe sufficient, though it could use some more detail.  I'll comment there.

allennlp/tests/modules/token_embedders/pretrained_transformer_pretokenized_embedder_test.py
matt-gardner: The difference between this test and the last one is that this is checking after padding has been added.  That's particularly the line that you're relying on the superclass for.  I'd prefer a test that checks for padding on all keys that are output by this class.

allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py
```diff
@@ -0,0 +1,160 @@
...
+        tokens to this namespace, which would break on loading because we wouldn't find our default
+        OOV token.
+    """
+
+    def __init__(self, model_name: str, namespace: str = "tags", **kwargs) -> None:
+        super().__init__(**kwargs)
+        # The matched version v.s. mismatched
+        self._matched_indexer = PretrainedTransformerIndexer(model_name, namespace, **kwargs)
+
+        # add_special_tokens=False sicne we don't want wordpieces to be surrounded by special tokens
```
brendan-ai2: typo: s/sicne/since/

allennlp/tests/modules/token_embedders/pretrained_transformer_mismatched_embedder_test.py
```diff
@@ -0,0 +1,54 @@
...
+from allennlp.data.batch import Batch
+from allennlp.data.fields import TextField
+from allennlp.data.instance import Instance
+from allennlp.data.token_indexers import PretrainedTransformerMismatchedIndexer
+from allennlp.modules.text_field_embedders import BasicTextFieldEmbedder
+from allennlp.common.testing import AllenNlpTestCase
+
+
+class TestPretrainedTransformerMismatchedEmbedder(AllenNlpTestCase):
+    def test_end_to_end(self):
```
brendan-ai2: This is a good test. One issue, however, is that given that it's using BERT we don't really have interpretable embeddings to make assertions against. So the logic in `forward` itself isn't really tested. One way around this would be to add an extra test and parameterize things such that you can pass in a fake `PretrainedTransformerEmbedder` that would return simple embeddings. This would be more work and I won't say it's strictly required, but if there happens to be a bug, we're going to catch it much later (and at greater cost) when our embeddings aren't giving us the performance we expect. So it's a tradeoff. I leave it to your discretion!

allennlp/modules/token_embedders/pretrained_transformer_mismatched_embedder.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@TokenEmbedder.register("pretrained_transformer_mismatched")
+class PretrainedTransformerMismatchedEmbedder(TokenEmbedder):
+    """
+    Use this embedder to embed wordpieces given by `PretrainedTransformerMismatchedIndexer`
+    and to pool the resulting vectors to get word-level representations.
+    """
+
+    def __init__(self, model_name: str) -> None:
```
DeNeutoy: Could you add a docstring to this class?

allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py
```diff
@@ -0,0 +1,160 @@
...
+            wordpieces.extend(subword_wordpieces)
+
+            start_offset = cumulative
+            cumulative += len(subword_wordpieces)
+            end_offset = cumulative - 1  # inclusive
+            offsets.append((start_offset, end_offset))
+
+        return wordpieces, offsets
+
+    def _determine_num_special_tokens_added(self):
```
DeNeutoy: This method could use a return type :)

allennlp/modules/token_embedders/pretrained_transformer_mismatched_embedder.py
ZhaofengWu: You mean the `# Parameters` section?

DeNeutoy: Yes - someone new to the library will not necessarily know what `model_name` means. It also helps the classes render nicely in the docs.

ZhaofengWu: Done

allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py
ZhaofengWu: Done

allennlp/tests/modules/token_embedders/pretrained_transformer_mismatched_embedder_test.py
ZhaofengWu: I think one test that we can add similar to this purpose is when I verify that all this works in a real model, I can assert the output logits for a fixed input (assuming fixed seed, etc.). This helps prevent following changes accidentally breaking it.

brendan-ai2: Yes, this is an option too. The benefit is that it's easy to implement. The downside is that it's brittle (an updated model from huggingface is liable to break it) and it's not interpretable, i.e., it won't necessarily be clear that the test shows that the embedder is working.

matt-gardner: Yeah, I would advise against hard-coding specific output embeddings or logits in a test, as that's really brittle.  Normally we haven't tried to test the specific predictions out of `forward` methods, as they are hard to test, we just test that it doesn't crash, and we rely on real-world usage to find those kinds of problems.  It's just too hard to catch in a unit test.

PR #3642, opened 2020-01-17
Remove some Sphynx artifacts from the docs

schmmd:
* Remove occurrences of `.. codeblock:`
* Replace double backquotes with a single backquote.


schmmd:
@DeNeutoy I'd like to test these changes locally but when I run `scripts/build_docs.py` locally I don't get any html files.  What do I need to do to build the docs as they are served?

schmmd:
Ah... I just found `build_docs.sh`.

schmmd:
@DeNeutoy changes look good locally.

DeNeutoy:
@schmmd Feel free to remove that failing docstring test, it's too fragile - it's broken every single time we've changed the command line docstrings

matt-gardner:
I disagree with removing that test - you're admitting that what shows up in the docs will be perpetually out of date.  If we want to remove the test, we should remove the docstrings that have the help text as well.

schmmd:
@matt-gardner I removed them and cut https://github.com/allenai/allennlp/issues/3645 so I can take a look later.

DeNeutoy:
Why not leave them in for now (so they appear in the docs) and make that issue something like "write a script which automatically replaces all the docstring comments in `commands`". I don't like that test because it's fiddly to fix, which means it should be a script which does it for you. We shouldn't check things we can automate instead.

schmmd:
@DeNeutoy good call.  I'll update the issue.

schmmd:
I'm ambivalent about whether we leave the doc strings in for the time being.

schmmd:
I'm just going to revert the commands directory to avoid any debate ;-)

DeNeutoy:
haha, this is what our tests should avoid - preventing us from making positive changes! But sounds good.

schmmd:
Now I'm getting conflicts... I'll likely just merge this manually after resolving.

DeNeutoy:
Sorry that was my fault - I tried to fix them but your fork is protected, so I can't push to it.

DeNeutoy:
@schmmd FYI I deployed this, should be live shortly

PR #3646, opened 2020-01-17
Changes I need for the TransformerQA model

### LINKED ISSUE #3224: How to set the token_type_ids when use  the pytorch-transformers Bert

dirkgr:
This adds a few odds and ends I need for the TransformerQA model at https://github.com/dirkgr/allennlp-reading-comprehension/tree/TransformerQA.

Virtually all of it is about the addition of type ids. Fixes #3224.


allennlp/common/util.py
```diff
@@ -577,3 +590,44 @@ def sanitize_wordpiece(wordpiece: str) -> str:
         return wordpiece[1:]
     else:
         return wordpiece
+
+
+def threaded_generator(g: Iterable, maxsize: int = 8) -> Generator:
```
brendan-ai2: I'm not crazy about adding this to the core library for several reasons:

1) There are no usages in the main library. Are there actually more than 2 sub-repos where you wish to use this?
2) It's simple enough that duplication may be preferable. Application code might want more options as well and it will be easier to edit the code directly than expand this interface.
3) There are no tests. If it's so simple it doesn't need tests, see 2.
4) It uses threads which won't help if your bottleneck is the CPU. Are you actually IO bound?

I'd be curious to hear more about your use case.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -35,6 +37,7 @@ def __init__(self, model_name: str, namespace: str = "tags", **kwargs) -> None:
         super().__init__(**kwargs)
         self._namespace = namespace
         self._tokenizer = AutoTokenizer.from_pretrained(model_name)
+        self._padding_value = self._tokenizer.convert_tokens_to_ids([self._tokenizer.pad_token])[0]
```
brendan-ai2: Paranoia: Could you assert that you only get back a single item?

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -83,14 +87,39 @@ def tokens_to_indices(self, tokens: List[Token], vocabulary: Vocabulary) -> Inde
...
+        self, tokens: IndexedTokenList, padding_lengths: Dict[str, int],
+    ) -> Dict[str, torch.Tensor]:
+        # Different transformers use different padding values for tokens, but for mask and type id, the padding
+        # value is always 0.
+        return {
+            key: torch.LongTensor(
+                pad_sequence_to_length(
+                    val,
+                    padding_lengths[key],
+                    default_value=lambda: 0 if key in {"mask", "type_ids"} else self._padding_value,
```
brendan-ai2: Wait so were we broken for the mask case?

allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py
```diff
@@ -64,6 +64,9 @@ def tokens_to_indices(self, tokens: List[Token], vocabulary: Vocabulary) -> Inde
         output = self._matched_indexer.tokens_to_indices(tokens, vocabulary)
 
         # self._intra_word_tokenize() does not insert special tokens, so we need to do it here
```
brendan-ai2: This comment refers to the line below your diff. Could you move it below?

allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py
```diff
@@ -64,6 +64,9 @@ def tokens_to_indices(self, tokens: List[Token], vocabulary: Vocabulary) -> Inde
...
+        output["type_ids"] = self._tokenizer.create_token_type_ids_from_sequences(
```
brendan-ai2: Perhaps we should move this behind a flag? I'm not sure if people always want type ids. Will take up extra space, for instance.

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -38,4 +43,8 @@ def forward(
 
         Shape: [batch_size, num_wordpieces, embedding_size].
         """
-        return self.transformer_model(input_ids=token_ids, attention_mask=mask)[0]
+        if self.transformer_model.embeddings.token_type_embeddings.num_embeddings == 1:
```
brendan-ai2: Could you add a comment explaining this condition? At a glance it looks like the model does in fact have type ids given that it has `num_embeddings` > 0.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
DeNeutoy: Yeah, could you add a test for this because it's pretty critical.

allennlp/common/util.py
DeNeutoy: +1 to Brendan's comment

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
DeNeutoy: It seems like a token indexer which uses a model which does not have `type_ids` should not be producing them, rather than catching it here in the token embedder. I.e this `if` statement should be unnecessary if the token indexer just didn't produce token_type_ids for relevant models.

allennlp/common/util.py
dirkgr: Happy to add tests, but let's make sure it should stay first.

You think this is simple enough that it should be copy-and-pasted wherever we need it? It seems long for that.

This is a very general utility that I've used in a variety of contexts before. Either producer or consumer being IO bound is one, but the one that's more interesting for us is producer or consumer being GPU bound. Inside of torch operations, the GIL is released, so the producer can work on getting the next batch ready while the previous batch is being run. If you have this:
```Python
for batch in batches:
    loss = model(batch)
    loss.backward()
    optimizer.step()
```
you can replace it with
```Python
for batch in threaded_generator(batches):
    loss = model(batch)
    loss.backward()
    optimizer.step()
```
and now you can do a fair bit of batch construction work quite late in the process without starving the GPU. That's always handy, but especially so when your dataset is so large it needs to be processed lazily.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
dirkgr: Actually, that's no longer needed. You can just call `tokenizer.pad_token_id`. I changed it that way.

dirkgr: e132b61ec3672f56fb06eb1cb8a80ae3cdff7426

allennlp/data/token_indexers/pretrained_transformer_indexer.py
dirkgr: RoBERTa pads with 1s. We probably have other places too where we assume the padding value is always 0.

dirkgr: 89cd484d17affbeda989db2939da57c21db3471e

allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py
dirkgr: Both of those lines refer ti the insertion of special tokens. Those functions are unfortunately named, but they do the same thing. One does it for token ids, and the other does it for type ids.

allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py
dirkgr: This is how type ids are handled inside the huggingface library:
```Python
    def forward(self, input_ids=None, token_type_ids=None, position_ids=None, inputs_embeds=None):
        if token_type_ids is None:
            token_type_ids = torch.zeros(input_shape, dtype=torch.long, device=device)
```
I can let them add the zero tensor instead of doing it here, but then I just have to drag around a special case.

The other reason we should have them is that there might be a transformer that uses type ids to mark special tokens. I don't know of any, but there are many in the library that I'm not familiar with.

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
dirkgr: `type_ids` doesn't have to come from the indexer. In my model, it comes from the dataset reader. That model uses `type_ids` to distinguish the context (which may contain the answer) from question and special tokens (which may not).

I could buy the argument that then the reader should not produce these, but it makes the code much more convenient if I can handle it in one place. Most of the time, `type_ids` are just a "bonus" feature that the model may or may not take advantage of, and RoBERTa does not.

For some models though, `type_ids` carry critical information. In those cases, it would be an error to use RoBERTa at all. I've built such a model before. To get it to work, I trained the type embeddings from scratch during fine-tuning (as they are discussing here: https://github.com/huggingface/transformers/issues/1234). It works, and beats the other transformers on the task. It would be a neat feature if we could offer RoBERTa with type ids, but that's out of scope for this PR.

On balance, I'm leaning towards moving this into the reader, but let me know what you think.

allennlp/common/util.py
brendan-ai2: Right, so for the specific case of producing data to be fed into the GPU we have the `MultiprocessDatasetReader` and the `MultiprocessIterator`. I worked a bit on improving these last quarter, but the quality still isn't great. Specifically, they don't scale linearly and can be tricky to configure (e.g. bad ulimit leads to mysterious errors), so I wouldn't recommend using them unless a) you really need them and b) you've benchmarked that they help in your case. Mark has worked on improving our iterators as well and has a PR that should replace both of those classes eventually. See https://github.com/allenai/allennlp/pull/3386. So if you want to work more on this issue, that PR is probably the right place to do it.

Is there another place where `threaded_generator` would be used?

allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py
brendan-ai2: I'm pretty confused by this... The docs seem to say that this does exactly what it sounds like, create type ids. See https://huggingface.co/transformers/model_doc/bert.html#transformers.BertTokenizer.create_token_type_ids_from_sequences What special tokens are inserted?

brendan-ai2: Okay, I ran the following and see what you mean, but the comment is still really confusing/broken now.

```
In [1]: import transformers

In [2]: tokenizer = transformers.tokenization_auto.AutoTokenizer.from_pretrained("bert-base-cased")

In [3]: tokenizer.build_inputs_with_special_tokens([1,2,3,4])
Out[3]: [101, 1, 2, 3, 4, 102]

In [4]: tokenizer.create_token_type_ids_from_sequences([1,2,3,4])
Out[4]: [0, 0, 0, 0, 0, 0]
```

How about:

```
self._intra_word_tokenize() does not insert special tokens, so we need to do it here. Before this we need to create the type ids as create_token_type_ids_from_sequences expects no special tokens -- though it does insert their corresponding ids.
```

matt-gardner:
Does this fix #3224?  It looks like it probably does.  If so, can you add "fixes #3224" to the description?

allennlp/common/util.py
dirkgr: I did not benchmark this. `threaded_generator()` is low overhead so I just throw it in whenever I think it might be needed. But for this thread, I just did. For the use case I'm putting it to, it saves 1.4 seconds out of a 188 second runtime. So I guess I'll just take it out.

That said, I don't understand your comment. `Multiprocess*` are flaky and hundreds of lines of code. The tests crash the Python interpreter on my Mac. `multiprocess` and `torch` have never been a stable combination. Even if it worked flawlessly, as soon as you use `multiprocess`, you have to serialize your data in at least one direction, and often in both directions. To make this cost worth it, you have to do a lot of processing in the reader.

And yet, you are suggesting it as an alternative to this thread thing, which is so small you're suggesting to copy and paste it where it's needed, and so reliable you can just toss it in and not worry about it, like you would just throw in `enumerate()` or one of the `itertools` functions. Also, it is so flexible that I can use it in [this use case](https://github.com/allenai/allennlp-reading-comprehension/pull/19/files#diff-6eeb463412e3fe022b3660b3a432e449R67), which is actually not a clear cut case of running the dataset reader in another thread.

I think this gets to the heart of the difference here. I don't think of `threaded_generator()` as something related to dataset readers. I think of it as a missing function in `itertools`.

I have some thoughts on #3386, but I'll put them there.

brendan-ai2: Sorry, sounds like I haven't made my concerns here clear. Let me take a step back. Fundamentally my reasoning is as follows:

1) The only use case we've identified so far for `threaded_generator` is data loading. Your [new example](https://github.com/allenai/allennlp-reading-comprehension/pull/19/files#diff-6eeb463412e3fe022b3660b3a432e449R67) is also data loading. 
2) We should provide a clear way for users to improve the throughput of their data loading. If your example didn't fit neatly match the `Iterator` abstraction (I think it does?), we should fix that abstraction.
3) `threaded_generator` does not interoperate with our existing tools (e.g., `allennlp train`) and according to your benchmarking is of negligible help.

Given all this, I'd like to ensure we don't confuse our users by introducing a new tool and instead redirect our attention to fixing https://github.com/allenai/allennlp/pull/3386.

(As for `Multiprocess*`, it is unfortunate that we're broken on your version of macOS and Python. Ideally at some point we would have CI covering macOS. However, for heavy workloads you're going to want a big GPU anyway -- typically on Linux -- which CI does cover. For small workloads, none of this matters.)

allennlp/data/token_indexers/pretrained_transformer_indexer.py
brendan-ai2: 👍

allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py
brendan-ai2: Sounds good.

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
brendan-ai2: I'd defer to you and Mark on the reader question. But if you do want to leave this check here, please comment the condition. It doesn't make any sense to me.

allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py
dirkgr: I added two comments. f04588327419209610795288815b49d983d7bf06

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
dirkgr: I changed this so that it throws an exception if your type ids are too big for the embeddings. It's a runtime check that will have a minuscule impact on performance, but when you get it wrong, the error message is really cryptic. e1be452cc12a38034ce19e3654a36c0f9b03058b

matt-gardner: I think the prior approach is actually better, as it uses fewer tensor operations.  There's no need for the additional operation here.  I'd agree with "minuscule" if it were one-time, but this is at every call to `forward`.

@brendan-ai2, the prior check worked because it's checking the number of embeddings in an embedding matrix (1, meaning that everything has the same, non-informative type id), not the number of embedding matrices.

brendan-ai2: Ah, thanks @matt-gardner !

dirkgr: I'm happy to take it out. Consider though that the error message you get when this is wrong is `RuntimeError: cublas runtime error : library not initialized at /tmp/pip-req-build-ocx5vxk7/aten/src/THC/THCGeneral.cpp:216`.

dirkgr: I did a little benchmarking for this. For `roberta-base`, the `forward()` call will take 316 ms (batch size 16). The `max()` call will take 68.5 µs. In other words, it adds 0.02% to the runtime of the `forward()` call.

matt-gardner: Fair enough, it's up to you. I wasn't trying to stop you from doing something, just express the general ideal that minimizing tensor operations is good, both for runtime and for memory. If you're confident it doesn't matter in this case, great.

PR #3648, opened 2020-01-17
Factor out batched_span_select code

ZhaofengWu:
I want to re-use this logic in a different place.

### DIFF
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -1260,6 +1260,73 @@ def batched_index_select(
     return selected_targets
 
 
+def batched_span_select(target: torch.Tensor, spans: torch.LongTensor) -> torch.Tensor:
+    """
+    The given ``spans`` of size ``(batch_size, num_spans, 2)`` indexes into the sequence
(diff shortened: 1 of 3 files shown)


allennlp/nn/util.py
```diff
@@ -1260,6 +1260,73 @@ def batched_index_select(
     return selected_targets
 
 
+def batched_span_select(target: torch.Tensor, spans: torch.LongTensor) -> torch.Tensor:
+    """
+    The given ``spans`` of size ``(batch_size, num_spans, 2)`` indexes into the sequence
```
DeNeutoy: Our docs only need markdown formatting now, so you can just use single backticks around code/dimensions

allennlp/tests/nn/util_test.py
```diff
@@ -1053,6 +1053,41 @@ def test_batched_index_select(self):
         with pytest.raises(ConfigurationError):
             util.batched_index_select(targets, indices)
 
+    def test_batched_span_select(self):
+        # Each element is a vector of it's index.
```
brendan-ai2: typo: s/it's/its/

PR #3656, opened 2020-01-20
Update run_with_beaker.py to avoid dropped single-file dataset support

viveklak:
Beaker recently dropped single-file dataset support and the mechanism used by the script would result in failing to accurately mount the jsonnet configuration file.


scripts/ai2_internal/run_with_beaker.py
```diff
@@ -152,7 +153,7 @@ def resume_command(experiment_id):
         experiment_id = subprocess.check_output(experiment_command, universal_newlines=True).strip()
         print(
             f"Experiment {experiment_id} submitted. "
-            f"See progress at https://allenai.beaker.org/ex/{experiment_id}"
+            f"See progress at https://beaker.org/ex/{experiment_id}"
```
viveklak: Switching since allenai.beaker.org is being migrated to beaker.org as we speak. This hardcoded reference is probably best avoided though?

scripts/ai2_internal/run_with_beaker.py
```diff
@@ -36,7 +36,8 @@ def main(param_file: str, args: argparse.Namespace):
     params = Params.from_file(param_file, overrides, ext_vars)
 
     # Write params as json. Otherwise Jsonnet's import feature breaks.
-    compiled_params_path = tempfile.mkstemp(".json", "config")[1]
+    params_dir = tempfile.mkdtemp(prefix="config")
```
viveklak: This does leak this directory but the script seems to leak temporary files generally, I assume this is to allow users look at the config for debugging etc. Please let me know if we would like to clean up the directory at the end instead.

scripts/ai2_internal/run_with_beaker.py
schmmd: This is a very good switch ;-)

I don't understand your question.

viveklak: Not particularly important: `beaker configure` can point to multiple URLs but since allenai.beaker.org is going away, seems fine to emit the hardcoded URL...

brendan-ai2: Thanks for catching this. And yes, if beaker.org is the only one in practice, let's just stick with the hardcoded URL.

scripts/ai2_internal/run_with_beaker.py
brendan-ai2: Yes, this is handy for debugging. Should be fine as-is. Someone would have to create an insane amount of beaker jobs before the periodic cleaning of the temp directory wouldn't be enough.

schmmd:
We could run a short-lived job to test this script--but honestly I might rather prioritize getting people off this script.  My key concern is that it doesn't encourage using AllenNLP as a dependency and thus requires a significant amount of building / network traffic.

brendan-ai2:
@schmmd, I'm not sure I understand your point about this script meaning we can't use AllenNLP as a dependency. I've done this and it worked fine after a small fix or two. See https://github.com/allenai/allennlp/pull/3248. The issue with network traffic is more a function of the Dockerfile than this script. And regardless of that, we have some heavy dependencies like torch. Finally, I'd note that we're already addressing the issue of unnecessary dependencies by factoring apart our repo.

In the absence of some clearly superior alternative I don't think having a goal of getting of off this script is compelling.

schmmd:
@brendan-ai2 it's probably easier to discuss in person.  I did not say with this script people can't use AllenNLP as a dependency.  My point was that this script builds the Dockerfile from scratch rather than depending on a particular AllenNLP release (e.g. `allennlp/allennlp:v0.9.0`).  I'd love to encourage the latter.  I've also seen it encourage people to fork AllenNLP in a way that makes it difficult to update when there's a new release.

Perhaps this script is a fine approach for people working on the AllenNLP library itself (which is what I wrote it for) but I wouldn't highlight it as the preferred way AllenNLP users should use Beaker.

brendan-ai2:
Happy to elaborate on offline as well, but, in short, I'm strongly against using Docker for dependency management. The way I used this script on the salience project was with 1) AllenNLP as a dependency and 2) our own Dockerfile that we managed. No weird forking involved. We can certainly add some explanation to encourage this.

schmmd:
I look forward to learning why you are strongly against using Docker for dependency management when building a Docker image.  Let's discuss in person.

Issue #3658, opened 2020-01-21
Make find_learning_rate compatible with DistributedDataParallel.


brendan-ai2:
**Describe the bug**
`find-lr` cannot use multiple GPUs. This is a regression. Multi-GPU compat was intentionally added in https://github.com/allenai/allennlp/pull/2045. More specifically, allennlp/tests/commands/find_learning_rate_test.py::TestFindLearningRate::test_find_learning_rate_multi_gpu currently fails. In 79601fa6f09d3fb6fedc3965850e5df907b716ea, for instance. 

**To Reproduce**
Run `pytest allennlp/tests/commands/find_learning_rate_test.py` on a multi-GPU machine. Be sure to use a commit at or before 79601fa6f09d3fb6fedc3965850e5df907b716ea as I intend to flip the assert to unbreak our tests.

**Expected behavior**
Test passes.

**System (please complete the following information):**
 - OS: Ubuntu 16.04.4 LTS
 - Python version: Python 3.6.7 :: Anaconda, Inc.
 - AllenNLP version: 79601fa6f09d3fb6fedc3965850e5df907b716ea
 - PyTorch version: 1.2.0

**Additional context**
There are other commands that could benefit from DistributedDataParallel, e.g., `evaluate`. We should try to reuse code between them.

DeNeutoy:
This is fine in it's current state, it seems unlikely that making this work in a more robust fashion would be valuable for the work it would take.

PR #3659, opened 2020-01-21
Assert that distributed config is not used with find-lr.

brendan-ai2:
- It would have no effect currently. We should fix this. See https://github.com/allenai/allennlp/issues/3658.
- Unbreaks multi-GPU tests in the interim by correcting the config, but asserting failure.


brendan-ai2:
Thanks!

PR #3662, opened 2020-01-21
Replace Sphynx class/function syntax with Markdown

schmmd:
1.  I manually cleaned up `common/`.
2.  I ran `find . -name '*.py' -exec gsed -i 's/:[a-z][a-z]*:`\~*/`/g' \{\} \;` to change syntax like ```:class:`~``` to simply a backtick.


brendan-ai2:
@schmmd, ping :)

schmmd:
@DeNeutoy seems like a manual process to add these links (I don't think I can just use a regex) so I might not get to this for quite a bit.

DeNeutoy:
Replaced by #3724

PR #3666, opened 2020-01-21
Support splitting long sequences into multiple segments for transformers

ZhaofengWu:
Split, independently embed, and finally concatenate back to get to original sequence representation.

### DIFF
--- a/allennlp/data/token_indexers/pretrained_transformer_indexer.py
+++ b/allennlp/data/token_indexers/pretrained_transformer_indexer.py
@@ -138,10 +138,10 @@ class PretrainedTransformerIndexer(TokenIndexer):
             # TODO(zhaofengw): we aren't respecting word boundaries when segmenting wordpieces.
 
             # Strips original special tokens
-            indices = indices[self._num_added_start_tokens:-self._num_added_end_tokens]
+            indices = indices[self._num_added_start_tokens : -self._num_added_end_tokens]
             # Folds indices
             wrapped_indices = [
-                indices[i:i + self._effective_max_len]
+                indices[i : i + self._effective_max_len]
                 for i in range(0, len(indices), self._effective_max_len)
             ]
             # Adds special tokens to each segment
--- a/allennlp/data/token_indexers/pretrained_transformer_indexer.py
+++ b/allennlp/data/token_indexers/pretrained_transformer_indexer.py
@@ -159,7 +159,7 @@ class PretrainedTransformerIndexer(TokenIndexer):
 
     @overrides
     def get_empty_token_list(self) -> IndexedTokenList:
-        result = {"token_ids": [], "mask": [], "type_ids": []}
+        result: IndexedTokenList = {"token_ids": [], "mask": [], "type_ids": []}
         if self._max_len > 0:
             result["segment_concat_mask"] = []
         return result
--- a/allennlp/data/token_indexers/pretrained_transformer_indexer.py
+++ b/allennlp/data/token_indexers/pretrained_transformer_indexer.py
@@ -107,7 +107,8 @@ class PretrainedTransformerIndexer(TokenIndexer):
 
     def _extract_token_and_type_ids(self, tokens: List[Token]) -> List[int]:
         """
-        Essentially returns `[token.text_id for token in tokens]`, with some checks.
+        Roughly equivalent to `zip(*[(token.text_id, token.type_id) for token in tokens])`,
+        with some checks.
         """
         indices: List[int] = []
         type_ids: List[int] = []
--- a/allennlp/data/token_indexers/pretrained_transformer_indexer.py
+++ b/allennlp/data/token_indexers/pretrained_transformer_indexer.py
@@ -94,7 +94,7 @@ class PretrainedTransformerIndexer(TokenIndexer):
     ) -> IndexedTokenList:
         """
         `tokens` may already be indices, in which case the token -> index step is a no-op, but other
-        logic still takes place, e.g. long sequence wrapping.
+        logic still takes place, e.g. long sequence splitting.
 
         `tokens` should have special tokens already inserted.
         """
@@ -140,17 +140,17 @@ class PretrainedTransformerIndexer(TokenIndexer):
             # Strips original special tokens
             indices = indices[self._num_added_start_tokens : -self._num_added_end_tokens]
             # Folds indices
-            wrapped_indices = [
+            folded_indices = [
                 indices[i : i + self._effective_max_len]
                 for i in range(0, len(indices), self._effective_max_len)
             ]
             # Adds special tokens to each segment
-            wrapped_indices = [
+            folded_indices = [
                 self._tokenizer.build_inputs_with_special_tokens(segment)
-                for segment in wrapped_indices
+                for segment in folded_indices
             ]
             # Flattens
-            indices = [i for segment in wrapped_indices for i in segment]
+            indices = [i for segment in folded_indices for i in segment]
 
             result["token_ids"] = indices
             result["segment_concat_mask"] = [1] * len(indices)
--- a/allennlp/data/token_indexers/pretrained_transformer_indexer.py
+++ b/allennlp/data/token_indexers/pretrained_transformer_indexer.py
@@ -1,4 +1,4 @@
-from typing import Dict, List, Tuple
+from typing import Dict, List, Optional, Tuple
 import logging
 import torch
 from allennlp.common.util import pad_sequence_to_length
@@ -105,7 +105,9 @@ class PretrainedTransformerIndexer(TokenIndexer):
 
         return self._postprocess_output(output)
 
-    def _extract_token_and_type_ids(self, tokens: List[Token]) -> List[int]:
+    def _extract_token_and_type_ids(
+        self, tokens: List[Token]
+    ) -> Tuple[List[int], Optional[List[int]]]:
         """
         Roughly equivalent to `zip(*[(token.text_id, token.type_id) for token in tokens])`,
         with some checks.
--- a/allennlp/data/token_indexers/pretrained_transformer_indexer.py
+++ b/allennlp/data/token_indexers/pretrained_transformer_indexer.py
@@ -1,4 +1,4 @@
-from typing import Dict, List
+from typing import Dict, List, Union
 import logging
 import torch
 from allennlp.common.util import pad_sequence_to_length
@@ -31,13 +31,35 @@ class PretrainedTransformerIndexer(TokenIndexer):
         We use a somewhat confusing default value of `tags` so that we do not add padding or UNK
         tokens to this namespace, which would break on loading because we wouldn't find our default
         OOV token.
+    max_len: `int`, optional (default = -1)
+        If positive, split the document into segments of this many tokens (including special tokens)
+        before feeding into the embedder. The embedder embeds these segments independently and
+        concatenate the results to get the original document representation. Should be set to
+        the same value as the `max_len` option on the `PretrainedTransformerEmbedder`.
     """
 
-    def __init__(self, model_name: str, namespace: str = "tags", **kwargs) -> None:
+    def __init__(
+        self, model_name: str, namespace: str = "tags", max_len: int = -1, **kwargs
+    ) -> None:
         super().__init__(**kwargs)
         self._namespace = namespace
         self._tokenizer = AutoTokenizer.from_pretrained(model_name)
         self._added_to_vocabulary = False
+
+        (
+            self._num_added_start_tokens,
+            self._num_added_end_tokens,
+        ) = self._determine_num_special_tokens_added()
+
+        self._max_len = max_len
+        if self._max_len > 0:
+            self._effective_max_len = (  # we need to take into account special tokens
+                self._max_len - self._tokenizer.num_added_tokens()
+            )
+            if self._effective_max_len <= 0:
+                raise ValueError(
+                    "max_len needs to be greater than the number of special tokens inserted."
+                )
 
     def _add_encoding_to_vocabulary(self, vocab: Vocabulary) -> None:
         """
@@ -67,7 +89,15 @@ class PretrainedTransformerIndexer(TokenIndexer):
         pass
 
     @overrides
-    def tokens_to_indices(self, tokens: List[Token], vocabulary: Vocabulary) -> IndexedTokenList:
+    def tokens_to_indices(
+        self, tokens: List[Union[Token, int]], vocabulary: Vocabulary
+    ) -> IndexedTokenList:
+        """
+        `tokens` may already be indices, in which case the token -> index step is a no-op, but other
+        logic still takes place, e.g. long sequence wrapping.
+
+        `tokens` should have special tokens already inserted.
+        """
         if not self._added_to_vocabulary:
             self._add_encoding_to_vocabulary(vocabulary)
             self._added_to_vocabulary = True
@@ -75,7 +105,9 @@ class PretrainedTransformerIndexer(TokenIndexer):
         indices: List[int] = []
         type_ids: List[int] = []
         for token in tokens:
-            if getattr(token, "text_id", None) is not None:
+            if isinstance(token, int):
+                indices.append(token)
+            elif getattr(token, "text_id", None) is not None:
                 # `text_id` being set on the token means that we aren't using the vocab, we just use
                 # this id instead. Id comes from the pretrained vocab.
                 # It is computed in PretrainedTransformerTokenizer.
@@ -97,11 +129,40 @@ class PretrainedTransformerIndexer(TokenIndexer):
         result = {"token_ids": indices, "mask": mask}
         if type_ids is not None:
             result["type_ids"] = type_ids
+
+        if self._max_len > 0:
+            # We prepare long indices by converting them to (assuming max_len == 5)
+            # [CLS] A B C [SEP] [CLS] D E F [SEP] ...
+            # Embedder is responsible for folding this 1-d sequence to 2-d and feed to the
+            # transformer model.
+            # TODO(zhaofengw): we aren't respecting word boundaries when segmenting wordpieces.
+
+            # Strips original special tokens
+            indices = indices[self._num_added_start_tokens:-self._num_added_end_tokens]
+            # Folds indices
+            wrapped_indices = [
+                indices[i:i + self._effective_max_len]
+                for i in range(0, len(indices), self._effective_max_len)
+            ]
+            # Adds special tokens to each segment
+            wrapped_indices = [
+                self._tokenizer.build_inputs_with_special_tokens(segment)
+                for segment in wrapped_indices
+            ]
+            # Flattens
+            indices = [i for segment in wrapped_indices for i in segment]
+
+            result["token_ids"] = indices
+            result["segment_concat_mask"] = [1] * len(indices)
+
         return result
 
     @overrides
     def get_empty_token_list(self) -> IndexedTokenList:
-        return {"token_ids": [], "mask": [], "type_ids": []}
+        result = {"token_ids": [], "mask": [], "type_ids": []}
+        if self._max_len > 0:
+            result["segment_concat_mask"] = []
+        return result
 
     @overrides
     def as_padded_tensor_dict(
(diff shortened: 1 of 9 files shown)


allennlp/modules/token_embedders/pretrained_transformer_mismatched_embedder.py
```diff
@@ -14,13 +14,19 @@ class PretrainedTransformerMismatchedEmbedder(TokenEmbedder):
     # Parameters
 
     model_name : `str`
-        The name of the `transformers` model to use.
+        The name of the `transformers` model to use. Should be the same as the corresponding
+        `PretrainedTransformerMismatchedIndexer`.
+    max_len : `int`, optional (default = -1)
```
matt-gardner: Full word here, too.

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -1,41 +1,132 @@
...
 class PretrainedTransformerEmbedder(TokenEmbedder):
     """
     Uses a pretrained model from `transformers` as a `TokenEmbedder`.
+
+    # Parameters
+
+    model_name : `str`
+        The name of the `transformers` model to use. Should be the same as the corresponding
+        `PretrainedTransformerIndexer`.
+    max_len : `int`, optional (default = -1)
```
matt-gardner: Full name.  Also, `None` might be a slightly better default value for all of these parameters - it's hard to do the wrong thing on accident that way, because the default value isn't comparable to other ints.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -65,14 +87,24 @@ def count_vocab_items(self, token: Token, counter: Dict[str, Dict[str, int]]):
         pass
 
     @overrides
-    def tokens_to_indices(self, tokens: List[Token], vocabulary: Vocabulary) -> IndexedTokenList:
+    def tokens_to_indices(
+        self, tokens: List[Union[Token, int]], vocabulary: Vocabulary
+    ) -> IndexedTokenList:
+        """
+        `tokens` may already be indices, in which case the token -> index step is a no-op, but other
```
matt-gardner: When will `tokens` already be indices?

ZhaofengWu: This is basically making the mismatched case a little easier. With some changes introduced in this PR, now the mismatched indexer has to do subword tokenization AND special token insertion before [calling](https://github.com/allenai/allennlp/blob/master/allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py#L64) the matched, i.e. default, indexer's `tokens_to_indices()`. Because we use huggingface's `_build_input_with_special_tokens(List[int]) -> List[int]` method to insert special tokens, we already have a list of `int` indices before calling the matched indexer's `tokens_to_indices()`. Technically we can convert back to `Token`s and then leave this place unchanged, but I think it's just a little easier this way.

matt-gardner: In that case you should use a `# type: ignore` on the caller, not change the type annotation here.  What you did here should actually make mypy complain, because now the type signature doesn't match the superclass.

matt-gardner: I'm not sure why mypy isn't complaining, actually.  Maybe it was that pylint would have complained, but we got rid of pylint?

ZhaofengWu: I _am_ using a `# type: ignore` on the caller -- you can search for the string `)  # type: ignore` on the `Files Changed` page.

Yeah I didn't realize the mismatch with superclass issue. Are you suggesting to leave the type signature here unchanged (i.e. back to `List[Token]`) but actually expects that `tokens` might be `List[int]` in the function body?

ZhaofengWu: Ideally though I think we should break apart the `Token -> Index` logic and any potential postprocessing logic (e.g. the long sequence splitting one here) into two separate functions. And if we already have indices, we can skip the former function and directly call the second. But here we only have this one function so it is a little weird that a function called `tokens_to_indices()` actually accepts indices as input.

allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py
```diff
@@ -58,19 +58,15 @@ def count_vocab_items(self, token: Token, counter: Dict[str, Dict[str, int]]):
     @overrides
     def tokens_to_indices(self, tokens: List[Token], vocabulary: Vocabulary) -> IndexedTokenList:
         orig_token_mask = [1] * len(tokens)
-        tokens, offsets = self._intra_word_tokenize(tokens)
+        tokens, offsets = self._intra_word_tokenize(tokens)  # type: ignore
```
matt-gardner: This is a private method only called by this class.  Why do you need to ignore types?  You can just give the right type signature to the private method.  Is it because you're changing the type of `tokens` potentially?

allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py
```diff
@@ -58,19 +58,15 @@ def count_vocab_items(self, token: Token, counter: Dict[str, Dict[str, int]]):
...
+        wordpiece_mask = [1] * len(tokens)
 
         # {"token_ids": ..., "mask": ...}
         output = self._matched_indexer.tokens_to_indices(tokens, vocabulary)
```
matt-gardner: This is the line where I would have expected a `# type: ignore`, from our conversation above.  You changed the type signature of this method in the called class, instead of just add an ignore here.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
ZhaofengWu: Oh nvm I wasn't using a `# type: ignore` for this function. Sorry about the confusion.

allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py
ZhaofengWu: Yes I think so -- sorry this wasn't what we were talking about above.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
matt-gardner: I think the issue is entirely because you're using one class to implement part of the functionality of the other class, in a way that isn't how the API was designed.  It doesn't look like you're proposing an external API change - there's nothing really to break apart with `Token -> Index` and postprocessing, because all of that happens internal to the `TokenIndexer` in what you're doing here.

matt-gardner: To be clear, I'm not necessarily complaining about this usage of the API, as long as you have good tests, and comments explaining whatever type wrangling you have to do to make this work.  I was just saying that this looks to me like an internal implementation decision for how to handle long sequences, instead of something that requires rethinking of our external API.  If you think there is actually something larger here that should be external facing, we can talk about that.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -82,15 +117,43 @@ def tokens_to_indices(self, tokens: List[Token], vocabulary: Vocabulary) -> Inde
...
                     f" for the following token: {token.text}"
                 )
-
-        # The mask has 1 for real tokens and 0 for padding tokens. Only real tokens are attended to.
-        mask = [1] * len(indices)
-
-        return {"token_ids": indices, "mask": mask}
+        return indices
+
+    def _postprocess_output(self, output: IndexedTokenList) -> IndexedTokenList:
```
brendan-ai2: Please add a docstring here. Even though this is a protected method it is much easier to understand with documentation.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -59,17 +84,27 @@ def _add_encoding_to_vocabulary(self, vocab: Vocabulary) -> None:
...
 
+        indices = self._extract_token_ids(tokens)
+        # The mask has 1 for real tokens and 0 for padding tokens. Only real tokens are attended to.
+        output = {"token_ids": indices, "mask": [1] * len(indices)}
+
+        return self._postprocess_output(output)
+
+    def _extract_token_ids(self, tokens: List[Token]) -> List[int]:
+        """
+        `tokens` should have special tokens already inserted.
```
brendan-ai2: Could you elaborate on what "extract" means here?

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -1,41 +1,137 @@
...
+            segment_concat_mask = fold(segment_concat_mask)
+
+        transformer_mask = segment_concat_mask if self._max_length is not None else mask
+        # Shape: [batch_size, num_wordpieces, embedding_size],
+        # or if self._max_length is not None:
+        # [batch_size * num_segments, self._max_length, embedding_size]
+        embeddings = self.transformer_model(input_ids=token_ids, attention_mask=transformer_mask)[0]
+        embedding_size = embeddings.size(2)
+
+        if self._max_length is not None:
```
brendan-ai2: Please break this block out into a separate method and add a unit test for it.

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -1,41 +1,137 @@
...
 
         # Returns:
 
         Shape: [batch_size, num_wordpieces, embedding_size].
         """
-        return self.transformer_model(input_ids=token_ids, attention_mask=mask)[0]
+        batch_size = token_ids.size(0)
+
+        if self._max_length is not None:
+            # [ [CLS] A B C [SEP] [CLS] D E F [SEP] ] ->
```
brendan-ai2: This also looks amenable to unit testing.

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -1,41 +1,191 @@
...
+            Shape: [batch_size * num_segments, self._max_length].
+        """
+        num_segment_concat_wordpieces = token_ids.size(1)
+        num_segments = math.ceil(num_segment_concat_wordpieces / self._max_length)
+        padded_length = num_segments * self._max_length
+        length_to_pad = padded_length - num_segment_concat_wordpieces
+
+        def fold(tensor):  # Shape: [batch_size, num_segment_concat_wordpieces]
+            # Shape: [batch_size, num_segments * self._max_length]
+            tensor = F.pad(tensor, [0, length_to_pad], value=0)
```
brendan-ai2: Is 0 really always the padding value? See this line in @dirkgr's outstanding PR: https://github.com/allenai/allennlp/pull/3646/files#diff-90fb6650b75551e526647ad854fa5998R40

allennlp/tests/modules/token_embedders/pretrained_transformer_embedder_test.py
```diff
@@ -69,3 +71,124 @@ def test_end_to_end(self):
...
+        assert tokens["bert"]["segment_concat_mask"].tolist() == [
+            [1] * segment_concat_length,
+            [1] * (segment_concat_length - 4) + [0] * 4,  # 4 is hard-coded length difference
+        ]
+
+        # Attention mask
+        bert_vectors = token_embedder(tokens)
+        assert bert_vectors.size() == (2, 9, 768)
+
+    def test_fold_long_sequences(self):
```
brendan-ai2: Thanks for adding these! Makes it very easy to see that the logic is solid :)

allennlp/tests/modules/token_embedders/pretrained_transformer_embedder_test.py
```diff
@@ -69,3 +71,124 @@ def test_end_to_end(self):
...
+            ]
+        )
+        folded_segment_concat_mask = (folded_token_ids > 0).long()
+
+        token_embedder = PretrainedTransformerEmbedder("bert-base-uncased", max_length=6)
+
+        folded_token_ids_out, folded_segment_concat_mask_out = token_embedder._fold_long_sequences(
+            token_ids, segment_concat_mask
+        )
+        assert (folded_token_ids_out == folded_token_ids).all()
```
brendan-ai2: We'll often use `numpy.testing.assert_array_equal`. There's also `numpy.testing.assert_array_almost_equal` for cases where floating point math is involved.

allennlp/tests/modules/token_embedders/pretrained_transformer_embedder_test.py
```diff
@@ -69,3 +71,124 @@ def test_end_to_end(self):
...
+        assert (folded_segment_concat_mask_out == folded_segment_concat_mask).all()
+
+    def test_unfold_long_sequences(self):
+        # Let's just say [PAD] is 0, [CLS] is 1, and [SEP] is 2
+        # We assume embeddings are 1-dim and are the same as indices
+        embeddings = torch.LongTensor(
+            [
+                [1, 101, 102, 103, 104, 2],
```
brendan-ai2: Could you give the [CLS] and [SEP] embeddings here different values in each segment so we can see which one is selected?

allennlp/tests/modules/token_embedders/pretrained_transformer_embedder_test.py
```diff
@@ -69,3 +71,124 @@ def test_end_to_end(self):
...
+                [1, 105, 106, 107, 108, 2],
+                [1, 109, 2, 0, 0, 0],
+                [1, 201, 202, 203, 204, 2],
```
brendan-ai2: Could you please make a test case where the first segment has padding at the end? I think there may be a bug here.

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -1,41 +1,191 @@
...
+        We truncate the start and end tokens for all segments, recombine the segments,
+        and manually add back the start tokens. We generally don't need to add back
+        the end tokens -- because the last segment is usually shorter than self._max_length,
+        the final end tokens usually won't be touched in our truncation process.
+
+        There are special cases where sequences have a full last segment. The only issue
+        in this case is that the [SEP] embedding will be wrong. Nevertheless, [SEP] embeddings
+        are rarely, if ever, used, so it shouldn't be a huge problem. However, we do remedy
+        this case when such sequences with a full last segment are the longest ones in a batch
+        by adding back the last token representations.
```
brendan-ai2: I think the reason why there are some issues here is because you're not using the mask associated with the embeddings. If you have the mask it's actually possible to avoid these bugs. For example, if you have the mask, then you know exactly where the [SEP] is for each segment. I'd say it's important to cover these cases robustly. Is there a reason why you avoided using the mask?

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -1,41 +1,191 @@
...
+        # We want to remove all segment-level special tokens but maintain sequence-level ones
+        num_wordpieces = num_segment_concat_wordpieces - (num_segments - 1) * self._num_added_tokens
+
+        embeddings = embeddings.reshape(batch_size, num_segments * self._max_length, embedding_size)
+        # Shape: (batch_size, self._num_added_start_tokens, embedding_size)
+        start_token_embeddings = embeddings[:, : self._num_added_start_tokens, :]
+        # See comment above -- remedy when the longest sequences have full last segments.
+        # This is not necessarily the end token embeddings when sequences aren't full.
+        # Shape: (batch_size, self._num_added_end_tokens, embedding_size)
+        last_token_embeddings = embeddings[:, -self._num_added_end_tokens :, :]
```
brendan-ai2: This is not correct. You need to use something like https://github.com/allenai/allennlp/blob/master/allennlp/nn/util.py#L1207 where the end indices can be computed from the mask.

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
ZhaofengWu: For the token ids, the padding value doesn't matter (we are not attending to paddings, etc.). For the mask, it _has_ to be 0 (which is why I'm explicitly specifying here).

allennlp/tests/modules/token_embedders/pretrained_transformer_embedder_test.py
ZhaofengWu: These are ints, so it's fine. I'm using regular assert here because people were using these in this file before me, and I want to be consistent. If we don't like these in the future, doing a sed wouldn't be hard.

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
brendan-ai2: Right, of course. Sorry about that!

allennlp/tests/modules/token_embedders/pretrained_transformer_embedder_test.py
brendan-ai2: 👍 for consistency.

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
brendan-ai2: Of course. Sorry about that!

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
ZhaofengWu: Discussed offline; will try to do this with fancy index manipulations.

allennlp/tests/modules/token_embedders/pretrained_transformer_embedder_test.py
ZhaofengWu: Discussed offline; resolving this.

brendan-ai2: For the record, what I asked for never occurs by construction, so no need for the test.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -82,15 +117,50 @@ def tokens_to_indices(self, tokens: List[Token], vocabulary: Vocabulary) -> Inde
...
+            # Adds special tokens to each segment
+            folded_indices = [
+                self._tokenizer.build_inputs_with_special_tokens(segment)
+                for segment in folded_indices
+            ]
+            # Flattens
+            indices = [i for segment in folded_indices for i in segment]
+
+            output["token_ids"] = indices
+            output["segment_concat_mask"] = [1] * len(indices)
```
brendan-ai2: Shouldn't this just be the `"mask"` and we can remove `"segment_concat_mask"`? You've already modified `output["token_ids"]` to be a concatenation of segments, so it seems like the mask should correspond to that. Further, in the embedder it seems like the usage of "mask" and "segment_concat_mask" are mutually exclusive.

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -1,41 +1,232 @@
...
+            tensor = F.pad(tensor, [0, length_to_pad], value=0)
+            # Shape: [batch_size * num_segments, self._max_length]
+            return tensor.reshape(-1, self._max_length)
+
+        return fold(token_ids), fold(segment_concat_mask)
+
+    def _unfold_long_sequences(
+        self,
+        embeddings: torch.FloatTensor,
+        segment_concat_mask: torch.LongTensor,
```
brendan-ai2: I'd just call this the mask.

allennlp/nn/util.py
```diff
@@ -1250,7 +1250,7 @@ def batched_index_select(
         flattened_indices = flatten_and_batch_shift_indices(indices, target.size(1))
 
     # Shape: (batch_size * sequence_length, embedding_size)
-    flattened_target = target.view(-1, target.size(-1))
+    flattened_target = target.reshape(-1, target.size(-1))
```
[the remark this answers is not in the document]
ZhaofengWu: Yes, but when a copy is necessary, it's necessary, right? And `reshape` can decide if it is indeed necessary, which in my understanding is just a more intelligent version of `view`. In this case, `view` didn't work, probably as a result of my fancy segmentations, so I had to change to `reshape`.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
ZhaofengWu: No because the `"mask"` key is reserved by whoever wants to call `get_text_field_mask`, and `"segment_concat_mask"` is for ourself to consume internally and to discard after the unfolding. See https://github.com/allenai/allennlp/pull/3666/files#diff-19a237a1412853c5e07d541a4e2ba518R58-R63

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
ZhaofengWu: See above

ZhaofengWu: I want to make thinks absolutely clear by always using different terminology for different things.

matt-gardner:
@dirkgr, I'll just note that this is a relatively common practice, whether or not it's intuitively the best option.  No comment on the specifics of the code here.

ZhaofengWu:
@dirkgr Re. whether or not it should be in the `Pretrained*` classes: this logic was in the old `WordpieceIndexer` and `PretrainedBertEmbedder`: https://github.com/allenai/allennlp/blob/master/allennlp/data/token_indexers/wordpiece_indexer.py#L57-L60

Re. whether we should _always_ use an overlapping window: [experiments](https://arxiv.org/pdf/1908.09091.pdf) have shown that overlapping window doesn't help for some cases compared to simply using non-overlapping ones. This was also how SpanBERT did coref (same team and same code as the paper linked above). If we want to support overlapping windows in the future we could build on top of this code.

PR #3667, opened 2020-01-21
update relative links in interpret and predictor folders

DeNeutoy:


DeNeutoy:
I didn't, as the paths are relative to the directory the file is in. I just did it by hand.

PR #3671, opened 2020-01-23
Add a plugin discovery mechanism

bryant1410:
Follows allenai/allennlp-server#3.

@matt-gardner sorry to be insisting, but what do you think of this way of doing it? It uses a namespace package. According to [PEP 420](https://www.python.org/dev/peps/pep-0420/), a namespace package is "a mechanism for splitting a single Python package across multiple directories on disk".

* To discover plugins: for whatever is inside `allennlp_plugins`.
* To add a plugin, as a user: create a folder `allennlp_plugins`, then another folder inside of it and put the Python files there. That's it. The person shouldn't put an `__init__.py` file inside of the first-mentioned folder, otherwise it won't work.

I created a test to show that it works.

What do you think? I think is dead simple. Still, if you don't like it, I can do the file approach.

Next steps:

* [x] docstrings.
* [x] load the plugins when `allennlp` command is run (or another PR?).
* [x] load the available subcommands (or another PR?).


allennlp/common/util.py
```diff
@@ -9,30 +9,28 @@
...
+import torch.distributed as dist
 from spacy.cli.download import download as spacy_download
 from spacy.language import Language as SpacyModelType
 
-# This base import is so we can refer to allennlp.data.Token in `sanitize()` without creating
-# circular dependencies.
+import allennlp_plugins
 from allennlp.common.checks import log_pytorch_version_info
 from allennlp.common.params import Params
 from allennlp.common.tqdm import Tqdm
```
bryant1410: I moved these imports before the `try` block so the code is neater.

allennlp/common/util.py
```diff
@@ -9,30 +9,28 @@
...
-import torch
 import numpy
 import spacy
+import torch
+import torch.distributed as dist
 from spacy.cli.download import download as spacy_download
 from spacy.language import Language as SpacyModelType
 
-# This base import is so we can refer to allennlp.data.Token in `sanitize()` without creating
-# circular dependencies.
```
bryant1410: This comment seems not to apply anymore, that's why I removed it.

allennlp/common/util.py
```diff
@@ -577,3 +575,12 @@ def sanitize_wordpiece(wordpiece: str) -> str:
...
         return wordpiece
+
+
+def discover_plugins(namespace: Type = allennlp_plugins) -> Iterable[pkgutil.ModuleInfo]:
+    return pkgutil.iter_modules(namespace.__path__, namespace.__name__ + ".")
+
+
+def import_plugins() -> None:
+    for module_info in discover_plugins():
+        importlib.import_module(module_info.name)
```
bryant1410: This is the main code.

setup.py
```diff
@@ -99,7 +100,8 @@
     author="Allen Institute for Artificial Intelligence",
     author_email="allennlp@allenai.org",
     license="Apache",
-    packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"]),
+    packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"])
+    + find_namespace_packages(include=["allennlp_plugins.*"]),
```
bryant1410: Note that having the folder `allennlp_plugins` in this project is not necessary. However, I added it so doing `import allennlp_plugins` in `allennlp/commons/util.py` doesn't fail if there are no plugins, and to avoid to do `try`-`except ImportError`, etc. And so it's included in the pip package, I added this line.

And I added the `.keep` file under `allennlp_plugins` so the folder isn't empty and can be added to git.

allennlp/tests/common/test_util.py
```diff
@@ -88,3 +103,30 @@ def test_get_frozen_and_tunable_parameter_names(self):
...
+            # We make one plugin available in the path, in another directory as if it were another project.
+            sys.path.append(temp_dir_a)
+
+            # We move to another directory with a different plugin, as if it were other project.
+            with pushd(temp_dir_b):
+                # In general when we run scripts or commands in a project, the current directory is the root of it
+                # and is part of the path. So we emulate this here.
+                sys.path.append(".")
+
+                available_plugins = list(util.discover_plugins())
```
matt-gardner: I don't understand why we need these two different methods.

allennlp/tests/common/test_util.py
```diff
@@ -88,3 +103,30 @@ def test_get_frozen_and_tunable_parameter_names(self):
...
         assert set(frozen_parameter_names) == {"linear.weight", "linear.bias"}
         assert set(tunable_parameter_names) == {"conv.weight", "conv.bias"}
+
+    def test_plugins_are_discovered_and_imported(self):
+        plugins_root = self.FIXTURES_ROOT / "plugins"
+        project_a_fixtures_root = plugins_root / "project_a"
+        project_b_fixtures_root = plugins_root / "project_b"
+
+        with tempfile.TemporaryDirectory() as temp_dir_a, tempfile.TemporaryDirectory() as temp_dir_b:
+            distutils.dir_util.copy_tree(project_a_fixtures_root, temp_dir_a)
```
matt-gardner: So, the way this works, you are required to name your project `allennlp_plugins`?  That seems suboptimal to me.  All of our example repos, and our recommended third-party setup, says "name your package whatever you want, then use `--include-package=[my_package_name]`".  Now you're saying that your package name *has* to be `allennlp_plugins`.  Right?

I agree that this is pretty easy on our side, but I don't really like the user side of it much.

bryant1410: No, you're not required to name your project `allennlp_plugins`. Just having that folder.

allennlp/tests/common/test_util.py
bryant1410: There's no need. In case someone wants to see the available plugins and check if a particular one is available. I can merge them anyway.

allennlp/tests/common/test_util.py
matt-gardner: But only things in that folder will be registered.  And basically all of the code that you write in a typical simple project is stuff that you want to be registered.  So, basically, yes, your project.

bryant1410: You can import your own project from that folder.

bryant1410: I'll add it to the test to make it clearer.

bryant1410: Check the new commit, the test shows how it works.

Essentially, the user has to:

    1. Add an `allennlp_plugins` folder.
    2. Add a folder inside of it, with the name of his project preferably.
    3. Add an `__init__.py` file that imports what he/she wants AllenNLP to see. He can add more files and Python symbols if she/he wants.

Maybe the `tree` output can make it clearer:

```
$ tree allennlp/tests/fixtures/plugins/project_a/
allennlp/tests/fixtures/plugins/project_a/
├── a
│   ├── a.py
│   └── __init__.py
└── allennlp_plugins
    └── a
        └── __init__.py

3 directories, 3 files
```

matt-gardner: Still seems a lot simpler to just not require the user to make that separate directory, and instead just add one line to a `.allennlp_plugins` file in the same directory (we could have our code check both the current directory, for local plugins, and a global `~/.allennlp/plugins` file for globally-installed stuff).  Or, if it's easy to make our pip packages add things to this namespace package, use that for first-party libraries, and use `.allennlp_plugins` for user libraries.

bryant1410: I'd tend to choose what Python recommends for this, but what you propose sounds a bit simpler for the user for his local project, yeah. For the global scope, I'm not convinced of having that global file. Maybe we can support both if it works fine for first-party pip-installed libraries. Let me do more testing and get back to you.

matt-gardner: There are two problems here: (1) having things like `pip install allennlp-server` just work with our normal commands when you install them.  And (2) letting users remove the need for constant `--include-package` statements everywhere.  These problems seem very related to me, but it's possible that they both should have separate solutions, because of python environment issues.  That's ok.

bryant1410:
Idk what happened that the tests take so long. Doesn't seem to me that I'd related to the new test, but maybe I'm wrong. Shouldn't there be a timeout?

bryant1410:
Okay, I left several more tests and fixed the reloading.

[To do plugin unloading seems to be hard](https://stackoverflow.com/a/487718/1165181) (i.e., when you want to make a plugin stop being available anymore at runtime, in the middle of a run). It seems like a far-fetched use case, so I don't worry about it. I skipped 2/9 tests because of it. I tried many things, including deleting from `sys.modules`, invalidating caches and so no.

Also, I do a `try`-`except` for discovering namespace plugins, even when the case that the namespace module does not exist is not an exception but something expected because [checking if a module is importable and then importing it involves more overhead than the necessary](https://docs.python.org/3/library/importlib.html#checking-if-a-module-can-be-imported).

@matt-gardner can you review when you can? If it's all fine, then the next steps would be to add docstrings and also to actually load the plugins in the main command somewhere (or maybe for another PR?).

matt-gardner:
Yes, I will look at this. Fridays are hard, though; might take me a few days.

allennlp/commands/__init__.py
```diff
@@ -82,16 +82,13 @@ def create_parser(
...
-                default=[],
-                help="additional packages to include",
-            )
+        subparser.add_argument(
+            "--include-package",
+            type=str,
+            action="append",
+            default=[],
+            help="additional packages to include",
+        )
```
bryant1410: It could be an option of the main parser if it's always used.

allennlp/commands/__init__.py
```diff
@@ -51,37 +50,33 @@ def add_argument(self, *args, **kwargs):
...
+    prog: str = None, subcommand_overrides: Iterable[Subcommand] = None
 ) -> argparse.ArgumentParser:
     """
     Creates the argument parser for the main program.
     """
-    if subcommand_overrides is None:
-        subcommand_overrides = {}
+    subcommand_overrides = subcommand_overrides or []
 
-    parser = ArgumentParserWithDefaults(description="Run AllenNLP", usage="%(prog)s", prog=prog)
```
bryant1410: I removed `usage` because just the output "usage: allennlp\n" doesn't make sense to me.

allennlp/commands/__init__.py
```diff
@@ -51,37 +50,33 @@ def add_argument(self, *args, **kwargs):
...
 
-    subcommands = {
-        # Default commands
-        "train": Train(),
-        "evaluate": Evaluate(),
-        "predict": Predict(),
-        "elmo": Elmo(),
-        "fine-tune": FineTune(),
-        "dry-run": DryRun(),
-        "make-vocab": DryRun(),  # deprecated, but keeping for backward compatibility.
```
bryant1410: I removed "make-vocab" implicitly. Please, lmk if you wanted it to be added again.

allennlp/commands/__init__.py
```diff
@@ -51,37 +50,33 @@ def add_argument(self, *args, **kwargs):
...
+
+    subcommands = set(subcommand_overrides)
+
+    # The class of each instance in the overrides iterable is gonna be in the subclasses
+    # iterable, by definition. So we only add those `Subcommand` subclasses that don't have an
+    # instance in the overrides iterable.
+    for subcommand_class in Subcommand.__subclasses__():
+        subcommand = subcommand_class()
+        if subcommand.name not in subcommand_overrides_names:
+            subcommands.add(subcommand)
```
bryant1410: Note that the overridden subcommands have preferences over the rest. However, there's no preference between the plugin's ones and the ones declared in allennlp. Is this okay?

allennlp/commands/__init__.py
```diff
@@ -50,59 +50,53 @@ def add_argument(self, *args, **kwargs):
...
-                help="additional packages to include",
-            )
+    subcommand_overrides_names = {subcommand.name for subcommand in subcommand_overrides}
+
+    subcommands = set(subcommand_overrides)
+
+    # The class of each instance in the overrides iterable is gonna be in the subclasses
+    # iterable, by definition. So we only add those `Subcommand` subclasses that don't have an
+    # instance in the overrides iterable.
+    for subcommand_class in Subcommand.__subclasses__():
```
matt-gardner: Checking for `__subclasses__` is not great, and it's brittle (it only looks for one level down, so grandchildren are not included).  Instead, can we just use the `Registry`?  Make all `Subcommands` register themselves with a name, and just look in the registry to get all registered `Subcommands`.  Then you can remove the `overrides` parameter, also, because there's a trivial mechanism to get your command in - just register it.  Also, this would let you remove the `name` property that you added to `Subcommand` - you just use the registered name as the command name, and that's it.

bryant1410: > Checking for `__subclasses__` is not great, and it's brittle (it only looks for one level down, so grandchildren are not included).

We could do it recursively.

> Instead, can we just use the `Registry`? Make all `Subcommands` register themselves with a name, and just look in the registry to get all registered `Subcommands`. Then you can remove the overrides parameter, also, because there's a trivial mechanism to get your command in - just register it. Also, this would let you remove the name property that you added to Subcommand - you just use the registered name as the command name, and that's it.

Okay, I'll try with the `Registry`. Makes sense.

Related, I think subclasses (and recursively) could be registered automatically within this kind of `Registrable` API, maybe optionally. See epwalsh/python-registrable#2. I'm not saying to do so here but just saying that, if it's abstracted away then it'd be nice to have that option for the API consumers.

allennlp/common/plugins.py
```diff
@@ -0,0 +1,74 @@
...
+by AllenNLP with the methods provided in this module.
+
+There are two ways of declaring plugins for discovery:
+
+    * Writing the package name to import in a file (typically ".allennlp_plugins", in the current
+    directory from which the command "allennlp" is run). This is the simplest approach.
+
+    * Creating a folder called "allennlp_plugins" that's in the Python path when you run the
+    "allennlp" command (typically under your project's root directory), then creating a subfolder
+    with the name you want and creating an "__init.py" file that imports the code you want (e.g.,
```
matt-gardner: `__init__.py`?

allennlp/common/plugins.py
```diff
@@ -0,0 +1,74 @@
...
+    your Python package). This option is preferred when you want to create a pip-installable
+    package and you want to make your AllenNLP plugin available when users install your package.
```
matt-gardner: Can you give an example of this?

allennlp/commands/__init__.py
```diff
@@ -38,81 +39,58 @@ def _is_empty_default(default):
...
 
 
-def main(prog: str = None, subcommand_overrides: Dict[str, Subcommand] = None) -> None:
+def main(prog: Optional[str] = None) -> None:
     """
     The :mod:`~allennlp.run` command only knows about the registered classes in the ``allennlp``
     codebase. In particular, once you start creating your own ``Model`` s and so forth, it won't
-    work for them, unless you use the ``--include-package`` flag.
+    work for them, unless you use the ``--include-package`` flag or you make your code available
+    as a plugin.
```
matt-gardner: Maybe have a reference here to how to make this happen?

allennlp/tests/commands/main_test.py
```diff
@@ -54,79 +56,103 @@ def test_other_modules(self):
...
+            # Rewrite out config file, but change a value.
+            with open(config_path, "w") as new_config_file:
+                new_config_file.write(config_json.replace('"num_epochs": 2,', '"num_epochs": 4,'))
 
-        main()
+            # This should fail because the config.json does not match that in the serialization directory.
+            with pytest.raises(ConfigurationError):
+                main()
+
+    def test_plugins_loaded(self):
```
matt-gardner: Can you have a similar test here for the `.allennlp_plugins` option?

allennlp/tests/common/plugins_test.py
```diff
@@ -0,0 +1,190 @@
...
+        self.assertSetEqual(set(), available_plugins)
+
+    def test_global_namespace_plugin(self):
+        with pip_install(self.project_a_fixtures_root, "a"):
+            available_plugins = set(discover_plugins())
+            self.assertSetEqual({"allennlp_plugins.a"}, available_plugins)
+
+            import_plugins()
+            subcommands_available = {t.__name__ for t in Subcommand.__subclasses__()}
+            self.assertIn("A", subcommands_available)
```
matt-gardner: Can you also test that A is not in the available subcommands before you do the pip install?  Same for all of the other tests.

matt-gardner: Hmm, I guess the plugin unloading bit makes me think that won't actually work?

allennlp/tests/common/plugins_test.py
```diff
@@ -0,0 +1,190 @@
...
+        available_plugins = set(discover_file_plugins("dummy_file"))
+        self.assertSetEqual(set(), available_plugins)
+
+    def test_global_namespace_plugin(self):
+        with pip_install(self.project_a_fixtures_root, "a"):
+            available_plugins = set(discover_plugins())
+            self.assertSetEqual({"allennlp_plugins.a"}, available_plugins)
+
+            import_plugins()
+            subcommands_available = {t.__name__ for t in Subcommand.__subclasses__()}
```
matt-gardner: Just use `Subcommand.list_available()` here also?

allennlp/tests/common/plugins_test.py
bryant1410: I think it should work because it's before, not after.

bryant1410: Oh, you mean because it was installed in another test?

matt-gardner: Yeah.

allennlp/tests/commands/main_test.py
```diff
@@ -130,7 +130,20 @@ def test_other_modules(self):
...
+        plugins_root = self.FIXTURES_ROOT / "plugins"
+        # "b" sets a "local" namespace plugin, because it's supposed to be run from that directory.
+        project_b_fixtures_root = plugins_root / "project_b"
+
+        sys.argv = ["allennlp"]
+
+        with push_python_project(project_b_fixtures_root):
+            main()
+
+        subcommands_available = {t.__name__ for t in Subcommand.__subclasses__()}
```
matt-gardner: `list_available()`

allennlp/common/plugins.py
matt-gardner: Oh, what I meant was an example of *how* to do this.  If that example is a pointer to a repo that does it, that's fine too. (and in this case it's ok to put a pointer to a repo that doesn't do it yet, but will very soon)

allennlp/common/testing/plugins_test_util.py
```diff
@@ -0,0 +1,35 @@
+from contextlib import contextmanager
```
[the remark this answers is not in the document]
matt-gardner: pytest quirk: name the file `plugins_util.py`, not `plugins_test_util.py`, otherwise pytest will think it's a file that has tests.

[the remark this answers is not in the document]
matt-gardner: Also, can you put it as close as possible to the thing it's testing?  I think that's probably `allennlp/tests/common/plugins_util.py`.

allennlp/common/plugins.py
bryant1410: Haha, I wonder why don't you just say the repo name.

allennlp/tests/common/plugins_test.py
bryant1410: The issue is also that the subcommands keep registered because `Subcommand` is imported only once for the whole test suite. It's not easy to solve because it could potentially be implicitly and recursively imported even if you put an explicit import inside each test function.

So what I did is just to see if the plugins are discovered before each test function, which is a proxy of it.

PR #3674, opened 2020-01-23
Remove files_to_archive from Params

### LINKED ISSUE #3657: Remove files_to_add functionality
Our ability to use and cache URLs should now be preferred over `files_to_archive`.  That old functionality is a main remaining cause of needing a custom `from_params` method, and it's causing headaches as I'm trying to update configs for https://github.com/allenai/allennlp-hub/pull/14.  It looks like if you untar the model archive, modify the config, then retar the archive, things that used to work in the "files_to_archive" bit no longer work, for reasons I don't yet understand.

matt-gardner:
Fixes #3657.  This was only ever used for ELMo, and we should just rely on `cached_path` instead of this.  It greatly simplifies our model archives and removes the need for some custom `from_params` methods.


brendan-ai2:
I'm not sure I see how we can remove this without some replacement. The ability to have additional files added to the archive seems critical not just for fine tuning ELMo (which seems enough...) but for composing models more generally. This strikes me as a very important use case.

matt-gardner:
The only thing this was ever used for was to put ELMo weight and option files into the model archive. We could have used this for BERT (etc.), but instead we just used `cached_path`, which accomplishes the same goal (have a file you had at training time available on some other machine when you want to load a model), but *much* simpler.  `cached_path` can do everything that `files_to_archive` did.

matt-gardner:
In other words: the replacement is already in place, and we are using it for everything except some old, outdated code.

allennlp/models/archival.py
```diff
@@ -148,14 +126,6 @@ def archive_model(
         archive.add(weights_file, arcname=_WEIGHTS_NAME)
         archive.add(os.path.join(serialization_dir, "vocabulary"), arcname="vocabulary")
 
-        # If there are supplemental files to archive:
-        if files_to_archive:
```
brendan-ai2: How does `cached_path` handle this situation? That is, some model uses ELMo on a downstream task and fine tunes the ELMo weights. This code will ensure that the fine-tuned weights are saved. This is a very common use case, no? Certainly `cached_path` works in the read-only case. Sorry if I'm missing something obvious...

brendan-ai2: Actually, sorry. Let me look at this more closely.

matt-gardner: The model always has its weights that are learned / fine-tuned in `weights.th`, the main component of our `model.tar.gz` files.  When using ELMo, the actual ELMo weights always end up as part of the model, because it's part of some `TokenEmbedder` that the model keeps as a `Module`.  The "weight_file" parameter that I removed from the file archive [here](https://github.com/allenai/allennlp/pull/3674/files#diff-941a4360ef08efef82e2b72210a7ab93L228) only has the *original* ELMo weights.  It's only actually used at initialization, and gets immediately overwritten when you load the saved weights in the model archive (this is an annoying aspect of how our model loading works, which I don't know how to get around) (also note that whether you fine-tuned the ELMo weights or not doesn't matter here).  There's also an "options_file" in the line above the one that I linked, and the contents of that one is more important at model loading time - it defines sizes and things, and without it we couldn't load a model.  But we can just specify a URL to that file, instead of a local path, and everything works exactly as it does now, without the need for `files_to_archive` and the accompanying `from_params` logic.

brendan-ai2: On a more careful reading it now looks to me like with the old code (in the fine-tuning ELMo case) we would copy the ELMo weight file (before fine-tuning!) into the archive. Then, when loading, we would load those files while constructing the `Elmo` module only to overwrite them with the actually fine-tuned weights stored separately in the archive. Is that correct?

brendan-ai2: We need a mutex on these comments ;)

brendan-ai2:
Okay, thanks for correcting my confusion on fine-tuning. Backing up a bit, I am still worried that this feels like a significant loss in functionality. I don't think `cached_path` with a URL is a full replacement for several reasons.

1) An average user might not have readily accessible hosting. (`file://` requires an absolute path, so that's not useful for distributing the model.)
2) Even if they have hosting, this might require them to make the file publicly accessible. This will be a significant security issue for certain users and use cases.
3) Prioritizing URLs on the net over local files seems pretty contrary to UNIX design.

brendan-ai2:
@DeNeutoy could you weigh in here too? I'm worried my context for the ELMo stuff is minimal.

matt-gardner:
To be clear, *this functionality is not currently used*. We should not be arguing about keeping it.

matt-gardner:
To your specific points:

1. Dropbox.
2. The only place we *ever* used this was with ELMo, where the files are already available online.
3. You can still use local files just fine. It's only when you want to move the file to another machine that things get hairy.  We should move that hairiness to the user, instead of dramatically complicating our own code to handle a *really* infrequent use case (where you have some external file you depend on that you can't put online somewhere).

matt-gardner:
We definitely still can do this. I think you're just missing a bunch of context on why this particular piece of functionality was added and how it's currently used. I can tell you for certain that all of its current usage is directly replaceable with `cached_path`, and that other more complex uses are things that we probably shouldn't be trying to support anyway.  It's only a *really* crazy scenario where you would not be able to just use `cached_path` to solve your problem.

matt-gardner:
Some additional context: a key goal for 1.0 is to have the library be easily usable without needing `from_params`.  The way `files_to_archive` worked made it impossible to remove custom `from_params` methods, and stops me from dramatically simplifying how `from_params` works.  If we want to make the library usable from scripts, we need to remove this.  It turns out that `cached_path` is already a drop-in replacement, so no harm done, anyway.

Sorry that I didn't give more context to start out, and that I lost my patience in the discussion.

brendan-ai2:
Thanks for the additional context, Matt. No worries at all. I know it's frustrating when a change that's obvious to the author (who has actually done the real work!) isn't to the reviewer.

As an aside, for now I think we can suggest to any users that need (specifically) local files that they simply place them in their current working directory and use relative paths in their config. If they're a bit careful, this will be portable across machines.

PR #3676, opened 2020-01-25
change where we block in the distributed case

### LINKED ISSUE #3663: Possible race condition in metric calculation during multi-gpu training
**Describe the bug**
Allennlp models normally calculate the metrics in the model forward methods. However, this may cause synchronization issues between gpus while doing multi-gpu training using training_util.data_parallel. Regarding this, I noticed problems while I was training a BERT model on a classification task. Specifically, the accuracy recorded after each epoch was different among multiple runs of the same model with the same training configuration, although I used the prepare_environment to set the same seed. The stranger thing is that the accuracy sometimes exceeds 1.0, which was very confusing. I noticed that jiant users face a similar problem during multi-gpu training as can be seen from https://github.com/nyu-mll/jiant/issues/947#issue-516769220.
As I understand, their current solution is to take out metric calculation from model's forward method. Instead, they do the calculation on a single gpu after forward call returns in the trainer.

**To Reproduce**
Take a powerful classification model (e.g. BERT) and train it with multiple-gpus on a very small toy dataset so that it can overfit it. If you disable the patience and wait for some time, the model might exceed 1.0 accuracy because of this possible bug.

DeNeutoy:
Fixes #3663 

@scarecrow1123 would love to get your point of view on this, I think this is correct? We want the metrics returned at the end of each epoch to only be computed once we are sure that all the workers have finished their epoch.

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -453,6 +453,12 @@ class Trainer(TrainerBase):
                 self._save_checkpoint(
                     "{0}.{1}".format(epoch, training_util.time_to_str(int(last_save_time)))
                 )
+
+        # Let all workers finish their epoch before computing
+        # the final statistics for the epoch.
+        if self._distributed:
+            dist.barrier()
+
         metrics = training_util.get_metrics(
             self.model,
             train_loss,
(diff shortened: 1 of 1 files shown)


scarecrow1123:
@DeNeutoy Apologies for the delayed reply.

> We want the metrics returned at the end of each epoch to only be computed once we are sure that all the workers have finished their epoch.

Yes, having the check before computing the metrics for sure looks correct to me. Thanks!

PR #3680, opened 2020-01-26
remove generated mkdocs

DeNeutoy:
#3642 accidentally included the generated mkdocs section for the API. It should not be checked in, this removes it.


schmmd:
@DeNeutoy thanks--sorry for missing the comment.

PR #3683, opened 2020-01-27
Remove spurious part of condition

brendan-ai2:


brendan-ai2:
Thanks for catching this!

PR #3684, opened 2020-01-27
Add label distribution logic to coref

ZhaofengWu:
Add label distribution logic to Conll dataset reader and move some logic to `PretrainedTransformerTokenizer` for better reuse.

### DIFF
--- a/allennlp/data/dataset_readers/coreference_resolution/conll.py
+++ b/allennlp/data/dataset_readers/coreference_resolution/conll.py
@@ -168,6 +168,13 @@ class ConllCorefReader(DatasetReader):
 
         cluster_dict = {}
         if gold_clusters is not None:
+            if self._wordpiece_modeling_tokenizer is not None:
+                for cluster in gold_clusters:
+                    for mention_id, mention in enumerate(cluster):
+                        start = offsets[mention[0]][0]
+                        end = offsets[mention[1]][1]
+                        cluster[mention_id] = (start, end)
+
             for cluster_id, cluster in enumerate(gold_clusters):
                 for mention in cluster:
                     cluster_dict[tuple(mention)] = cluster_id
@@ -180,15 +187,17 @@ class ConllCorefReader(DatasetReader):
             for start, end in enumerate_spans(
                 sentence, offset=sentence_offset, max_span_width=self._max_span_width
             ):
+                if self._wordpiece_modeling_tokenizer is not None:
+                    start = offsets[start][0]
+                    end = offsets[end][1]
+                    if end - start + 1 > self._max_span_width:
+                        continue
+
                 if span_labels is not None:
                     if (start, end) in cluster_dict:
                         span_labels.append(cluster_dict[(start, end)])
                     else:
                         span_labels.append(-1)
-
-                if self._wordpiece_modeling_tokenizer is not None:
-                    start = offsets[start][0]
-                    end = offsets[end][1]
 
                 spans.append(SpanField(start, end, text_field))
             sentence_offset += len(sentence)
@@ -197,12 +206,6 @@ class ConllCorefReader(DatasetReader):
 
         metadata: Dict[str, Any] = {"original_text": flattened_sentences}
         if gold_clusters is not None:
-            if self._wordpiece_modeling_tokenizer is not None:
-                for cluster in gold_clusters:
-                    for mention_id, mention in enumerate(cluster):
-                        start = offsets[mention[0]][0]
-                        end = offsets[mention[1]][1]
-                        cluster[mention_id] = (start, end)
             metadata["clusters"] = gold_clusters
         metadata_field = MetadataField(metadata)
 
--- a/allennlp/data/dataset_readers/coreference_resolution/conll.py
+++ b/allennlp/data/dataset_readers/coreference_resolution/conll.py
@@ -192,6 +192,10 @@ class ConllCorefReader(DatasetReader):
                     end = offsets[end][1]
 
                     # `enumerate_spans` uses word-level width limit; here we apply it to wordpieces
+                    # We have to do this check here because we use a span width embedding that has
+                    # only `self._max_span_width` entries, and since we are doing wordpiece
+                    # modeling, the span width embedding operates on wordpiece lengths. So a check
+                    # here is necessary or else we wouldn't know how many entries there would be.
                     if end - start + 1 > self._max_span_width:
                         continue
                     # We also don't generate spans that contain special tokens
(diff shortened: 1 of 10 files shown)


DeNeutoy:
@ZhaofengWu - before we merge this, we should know that it improves performance (I think it's a little unlikely, from previous experiments with wordpiece vs pooled word representations). It's quite a complicated change and ideally we don't want to have options for things that aren't important. Obviously, if it works well, then it's great to add it.

matt-gardner:
@DeNeutoy, didn't you do the same thing for BERT-SRL?  Did it work there?

ZhaofengWu:
@DeNeutoy 

1. Based on one coref experiment that I ran, wordpiece modeling has similar performance as word-level (pooled) modeling. I ran two identical word-level modeling experiments, the first had 72.14 F1 but I accidentally killed it before fully converging, but a re-run gave just 71.68. Wordpiece-level modeling has 71.89 F1. @matt-gardner 's intuition is that wordpiece modeling would perform better. But regardless I think it could be nice to allow people to try both?

2. This PR does not add any new option to the "core" part (i.e. not dataset readers and not models) of the library. Besides adding a single argument to the `ConllCorefReader`, I only moved things around among the `PretrainedTransformerX` classes in a way that I think would make sense anyway regardless of whether or not we want the label distribution logic. E.g. I'm moving the `intra_word_tokenize()` function from `PretrainedTransformerIndexer` to `PretrainedTransformerTokenizer` which now is parallel with `tokenize()` and `tokenize_sentence_pair()`. That function also now returns a list of `Token`s rather than `int`s. So if our ultimate decision is that we don't want the label distribution logic, I'm willing to remove that part (i.e. changes to `ConllCorefReader`) but keeping the other changes.

DeNeutoy:
@matt-gardner I did that because I wanted to manually set the token type ids to indicate predicates, which is not possible with our token indexers. As a side effect of that, it was easier to just use wordpiece labels, so I didn't have to maintain duplicates of all the inputs in terms of both wordpieces and words. 

I don't particularly think it's nice to let people try both, really. I don't have an opinion on the rest of the PR, i'm sure it's probably great - I just don't think we should add options which are not effective.

matt-gardner:
Do we have any evidence one way or the other about whether it's better to do modeling with some kind of word pooling vs. distributing labels?  There's one data point from @ZhaofengWu above that seems to say it doesn't matter.

The high-level point is how to handle mismatched tokenization and modeling.  Either you can do pooling in your model, or you can distribute your labels.  My intuition was that distributing labels would be better, but I could very well be wrong.  If it doesn't matter, we could just recommend that people use pooling, and not worry about it.  I don't have a strong opinion about whether to include the option or not.  I have a bias towards letting people try things, but the more we can push that experimentation to user code instead of code that we maintain, the better.

ZhaofengWu:
As for results, I'm not aware of any work that does this comparison. I know SpanBERT uses wordpiece-modeling, but that doesn't definitively say it works better.

On whether or not we include this option in this PR (and just to be clear, we are talking about the `wordpiece_modeling_tokenizer` arg in the `ConllCorefReader`), I think ultimately it's just whether or not we want to give users an example on how to do this. Some users will try wordpiece modeling regardless, but it might just be a little easier for them if we provide an example here.

allennlp/data/dataset_readers/coreference_resolution/conll.py
```diff
@@ -165,6 +187,23 @@ def text_to_instance(
             for start, end in enumerate_spans(
                 sentence, offset=sentence_offset, max_span_width=self._max_span_width
             ):
+                if self._wordpiece_modeling_tokenizer is not None:
+                    start = offsets[start][0]
+                    end = offsets[end][1]
+
+                    # `enumerate_spans` uses word-level width limit; we need to be stricter here
```
brendan-ai2: This comment confuses me a bit. IIUC, the meaning is that the next statement ensures `_max_span_width` operates on word pieces instead of words. I might say, "`enumerate_spans` uses a word-level width limit; here we apply it to word pieces." The phrasing "we need to be stricter here" sounds almost like a todo comment.

As an aside, did you use the same `_max_span_width` in your experiments with and without the label distribution logic? It seems like that might affect the perf. Actually, dropping this check might allow for a fairer comparison. That said, I don't have much domain knowledge w.r.t. coref, so please let me know if this is a bad/irrelevant idea for some reason.

allennlp/tests/data/dataset_readers/coreference/coref_reader_test.py
```diff
@@ -184,6 +185,104 @@ def test_read_from_file(self, lazy):
...
+            "will",
+            "soon",
+            "open",
+            "to",
+            "the",
+            "public",
+            "here",
+            ".",
+            "[SEP]",
+        ]
```
brendan-ai2: Could you also include asserts against a few mentions explicitly like the other test case? Ideally in a way that shows how they're different with wordpieces. One issue with the test data currently is that all the words are so common we don't actually see the influence of wordpieces, i.e. no tokens are being split. I'd edit the test data to include some very long words (nonsense is okay) including where the mentions are.

allennlp/data/dataset_readers/coreference_resolution/conll.py
ZhaofengWu: We can't drop it here because we use a span width embedding that has only `_max_span_width` entries, and since we are doing wordpiece modeling, the span width embedding operates on wordpiece lengths. So a check here is necessary or else we wouldn't know how many entries that would be.

And yes I agree that controlling this across word vs. wordpiece modeling experiments is not entirely trivial, but that's a config choice and shouldn't be affected on the code.

allennlp/tests/data/dataset_readers/coreference/coref_reader_test.py
```diff
@@ -184,6 +185,81 @@ def test_read_from_file(self, lazy):
...
+
+        gold_span_labels = fields["span_labels"]
+        gold_indices_with_ids = [(i, x) for i, x in enumerate(gold_span_labels.labels) if x != -1]
+        gold_mentions_with_ids: List[Tuple[List[str], int]] = [
+            (candidate_mentions[i], x) for i, x in gold_indices_with_ids
+        ]
+
+        assert (["Hong", "Kong"], 0) in gold_mentions_with_ids
+        # Within span_width before wordpiece splitting but exceeds afterwards
+        assert (["the", "Hong", "Kong", "SA", "##R", "government"], 0) not in gold_mentions_with_ids
```
brendan-ai2: This is a fine assertion to include too, but my point from before is that we should have a positive test case here, e.g.

```
assert (["foo", "##bar"], 0) in gold_mentions_with_ids
```

As written one can't tell whether the code is actually including any mentions that were split which is fundamentally what this change is for.

allennlp/data/dataset_readers/coreference_resolution/conll.py
brendan-ai2: Thanks. Could you please update the comment in the code with that explanation?

PR #3693, opened 2020-01-29
Check for exceptions when importing plugins

bryant1410:


bryant1410:
Bulldozer is not gonna merge because of codecov unsuccessful check, right>

matt-gardner:
Not sure how bulldozer works.  If it won't do the coverage check right, that's too bad.

bryant1410:
> Not sure how bulldozer works. If it won't do the coverage check right, that's too bad.

My understanding is that it'll only merge if "the checks pass", whatever they are. I find weird this codecov behavior and hard to debug, I don't get what's going on. Now it passed :man_shrugging:

matt-gardner:
Bulldozer was supposed to merge with master and merge if checks pass.  Maybe it only does the merge with master if checks pass, and the "checks" that is uses for deciding whether to merge with master are different from the "checks" it uses for deciding whether to merge the PR?  @schmmd, any idea?

schmmd:
@matt-gardner I don't quite understand what you wrote.  All three checks passed and it wouldn't merge?

image

I believe those three need to pass and the PR needs to be approved--but I'm also still learning.

matt-gardner:
Now I'm really confused.  This originally had a 66% codecov patch, which wasn't passing.  Now it says "coverage not affected".  I have no idea what's going on there.

@schmmd, it's not worth the distraction to keep talking about this.  If I see the same thing ever come up again, I'll let you know.

schmmd:
Sounds good.  What you highlighted about codecov changing explains my confusion as well.

PR #3697, opened 2020-01-30
Roll forward PR #3691

brendan-ai2:
- There was a key error here: https://github.com/allenai/allennlp/commit/d23c0cdd7412e4dfe1b88343fef29645728bcaa4#diff-d4e8f335bf383d8d2b18287949003bbeR292
- GPU Unit Tests run only on commits to master which is why we didn't catch this sooner.


allennlp/commands/train.py
```diff
@@ -284,17 +284,19 @@ def train_model(
...
-            "directory": os.path.join(serialization_dir, "vocabulary"),
-            "padding_token": vocab._padding_token,
-            "oov_token": vocab._oov_token,
-        }
+        # Creating `Vocabulary` objects from workers could be problematic since
+        # the data iterators in each worker will yield only `rank` specific
+        # instances. Hence it is safe to construct the vocabulary and write it
+        # to disk before initializing the distributed context. The workers will
+        # load the vocabulary from the path specified.
+        if params.get("vocabulary", Params({})).get("type", "") != "from_files":
```
ZhaofengWu: In the future when Python 3.8 becomes a requirement we could be using the walrus operator for this :p

PR #3700, opened 2020-01-30
Data V2

DeNeutoy:
Attempt 2 at better data loading:

- `Trainer` now takes a `DataLoader` rather than the data itself and an iterator.
- `DatasetReaders` now return a subclass of pytorch `Dataset` which contains instances, and calls `index` on them before returning them. Note that this means the responsibility of indexing the instances has moved from `Iterator` to the `Dataset`, so correspondingly it now has a `def index_with(self, vocab: Vocabulary)` method. If the vocab is None, the instances are not indexed before returning.
- Adds some stubs so we can construct pytorch `DataLoader`, `Sampler` and `BatchSampler` objects `from_params`.
- Lazy datasets no longer support bucketing. Instead, the dataset reader should order instances in a way such that they are naturally bucketed - e.g read from a shard, sort raw strings by length, and just use the basic `DataLoader` functionality. This may change in the future if the pytorch `DataLoader` accepts a `BucketSampler` with an `IterableDataset`, see https://github.com/pytorch/pytorch/issues/28743 and https://github.com/pytorch/pytorch/pull/26547
- The `Trainer` is now essentially independent of how a user chooses to create their model inputs - a user doesn't have to use allennlp's data piece at all, because the trainer just accepts a dataloader, which passes it's batches to the model. 

**Before:**

```
    "iterator": {
      "type": "bucket",
      "sorting_keys": [["tokens", "num_tokens"]],
      "batch_size" : 32
    },
```
**New config using a batch sampler which buckets instances by length:**

```
    "data_loader": {
	  "batch_sampler"{
      "type": "bucket",
      "sorting_keys": [["tokens", "num_tokens"]],
      "batch_size" : 32
    }
},
```
**Vanilla config which does not use a sampler - in this case the dataloader will just use a sequential sampler by default.**
```
    "data_loader": {
	  "num_workers": 5,
      "timeout": 4,
      "batch_size" : 32
},
```

Ok, I did some rough benchmarks on ESIM + SNLI and it seems that this method is a bit faster than previously, but i'm not seeing the speedup with the number of workers that I was expecting. My current hypothesis for that is that the data loading process is not slow for the ESIM config, as it doesn't do any fancy indexing etc. I will try out some slower data loading models, but perhaps we can do that after merging, as a bunch of other stuff is waiting on this PR. 

Screen Shot 2020-02-25 at 3 31 58 PM


torch_datasets.py
```diff
@@ -0,0 +1,194 @@
...
+    def __getitem__(self) -> Instance:
+
+        raise NotImplementedError
+
+    def index_with(self, vocab: Vocabulary):
+        self.vocab = vocab
+
+
+"""
+Here we have two SNLI readers in both of the different styles.
```
matt-gardner: This is old?  I only see one here.

torch_datasets.py
```diff
@@ -0,0 +1,194 @@
...
+They are only slightly different.
+"""
+
+
+class SnliDataset(Dataset):
+    def __init__(
+        self, file_path: str, token_indexers: Dict[str, TokenIndexer] = None, lazy: bool = False
```
matt-gardner: As long as there's some other object that separates the `file_path` argument from the `token_indexers` argument, I'm ok with this.  The thing that's separated should be the one that's `Registrable`, though.

torch_datasets.py
```diff
@@ -0,0 +1,194 @@
...
+        file_path = cached_path(file_path)
+        self.examples = []
+
+        for line in open(file_path, "r"):
+            example = json.loads(line)
+            if example["gold_label"] == "-":
+                # These were cases where the annotators disagreed; we'll just skip them.  It's
+                # like 800 out of 500k examples in the training data.
+                continue
+            self.examples.append(example)
```
matt-gardner: Probably better to have `self.examples` actually be `self.instances`, with actual `Instance` objects, because then we can cache them very easily.

torch_datasets.py
```diff
@@ -0,0 +1,194 @@
...
+class Dataset(TorchDataset, Registrable):
+    def __init__(self):
+        self.vocab: Vocabulary = None
+
+    def text_to_instance(self, *inputs) -> Instance:
+
+        raise NotImplementedError
+
+    def __getitem__(self) -> Instance:
+
```
matt-gardner: I'd vote for having a default implementation here, like:

```python
def __getitem__(self, index: int) -> Instance:
    if not self._instances:
        self.load_instances()  # or something
    return self._instances[index]
```

torch_datasets.py
```diff
@@ -0,0 +1,194 @@
...
+            "premise_tokens": [x.text for x in premise_tokens],
+            "hypothesis_tokens": [x.text for x in hypothesis_tokens],
+        }
+        fields["metadata"] = MetadataField(metadata)
+        return Instance(fields)
+
+
+class BatchInstanceSampler(BatchSampler):
+    
+    def __init__(self, data, batch_size: int, sorting_keys: List[Tuple[str, str]] = None, padding_noise: float = 0.1):
```
matt-gardner: Also better here if there's something that separates the configuration from the data, so we can easily use the same configuration on multiple datasets (e.g., train and dev).

torch_datasets.py
```diff
@@ -0,0 +1,194 @@
...
+
+        self.vocab = data.vocab
+        self._sorting_keys = sorting_keys
+        self._padding_noise = padding_noise
+        self._batch_size = batch_size
+        self.data = data
+
+    def _argsort_by_padding(self, instances: List[Instance]) -> List[int]:
```
matt-gardner: Not sure why it's `argsort` instead of just `sort`.

torch_datasets.py
```diff
@@ -0,0 +1,194 @@
...
+data = SnliDataset("snli_20.jsonl")
+vocab = Vocabulary.from_instances(data)
+data.index_with(vocab)
+
+
+sampler = SequentialSampler(data)
+batch_sampler = BatchInstanceSampler(data, 4)
+
+
+def allennlp_collocate(batch):
```
matt-gardner: `collate`, not `collocate`, if you want to match the pytorch name.

torch_datasets.py
DeNeutoy: The torch sampler and batchsampler classes return indices into your dataset. So this method is returning the positions of indices in the original dataset such that they would be nicely bucketed together.

DeNeutoy:
I think that does make sense, apart from [pytorch really restricts the `DataLoader` when you are using a `IterableDataset`](https://github.com/pytorch/pytorch/blob/master/torch/utils/data/dataloader.py#L148)

which would make doing anything but the most naive dataset batching with an iterable dataset very difficult. Maybe this just means that for iterable datasets, you need to do this kind of thing in the dataset reader itself for now (for example, reading a shard of language modelling dataset, sort the sentences by length and yielding them will  give you a reasonable approximation to bucketing, whilst not requiring a sampler). 

Related issues:

https://github.com/pytorch/pytorch/issues/28743

https://github.com/pytorch/pytorch/issues/24915

https://github.com/pytorch/pytorch/pull/26547

matt-gardner:
Yeah, if we have a mechanism to support laziness, even if it's a bit more cumbersome than it currently is, I think it's ok.  I'm pretty sure laziness is much less common than being able to store all of the data in memory, so we should optimize for the common use case, as long as there are workarounds available for the less common case.

The practical language modeling dataset readers that I've seen in allennlp already basically bypass our data iterators, anyway.  So if that's the typical use case for laziness, we definitely should not be designing our iterators around them - they need something more than we can reasonably provide, anyway.

allennlp/common/testing/model_test_case.py
```diff
@@ -93,24 +94,27 @@ def ensure_model_can_train_save_and_load(
...
+        data_loader_params2 = Params(copy.deepcopy(data_loader_params.as_dict()))
+
+        data_loader = DataLoader.from_params(dataset=model_dataset, params=data_loader_params)
+        data_loader2 = DataLoader.from_params(dataset=loaded_dataset, params=data_loader_params2)
+
+        # We'll check that even if we index the dataset with each model separately, we still get
+        # the same result out.
+        model_batch = next(iter(data_loader))
+
+        loaded_batch = next((iter(data_loader2)))
```
bryant1410: A little thing :sweat_smile: 

```suggestion
        loaded_batch = next(iter(data_loader2))
```

DeNeutoy: lol thanks, i've worked on this too much today

bryant1410:
Why is the GAN test pointless? Doesn't it prove/test you can do GAN training with AllenNLP?

torch_datasets.py
```diff
@@ -0,0 +1,201 @@
+from typing import Dict, List, cast, Tuple
```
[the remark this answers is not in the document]
DeNeutoy: yup, thanks

allennlp/tests/training/trainer_test.py
```diff
@@ -42,16 +43,18 @@ def setUp(self):
         )
         self.model = SimpleTagger.from_params(vocab=self.vocab, params=self.model_params)
         self.optimizer = torch.optim.SGD(self.model.parameters(), 0.01, momentum=0.9)
-        self.iterator = BasicIterator(batch_size=2)
-        self.iterator.index_with(vocab)
+        self.data_loader = DataLoader(self.instances, batch_size=2, collate_fn=allennlp_collocate)
```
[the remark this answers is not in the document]
DeNeutoy: You can either pass a batch size, in which case it will use a uniform random sampler, or you can pass a `Sampler` (which do not take a batch size) with a specified `batch_size` passed here, _or_ you can specify only a `BatchSampler` - which is mutually exclusive with the other two options.

allennlp/data/samplers/samplers.py
```diff
@@ -0,0 +1,263 @@
...
+    """
+
+    def __iter__(self) -> Iterable[List[int]]:
+
+        raise NotImplementedError
+
+
+@Sampler.register("sequential")
+class SequentialSampler(Sampler, data.SequentialSampler):
+    def __init__(self, data_source: data.Dataset, **kwargs):
```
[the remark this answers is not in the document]
DeNeutoy: I did this because I thought that the way we construct samplers below, where we always pass a `data_source` argument even if the specific class doesn't take that argument, requires it? E.g we can't know apriori if a sampler takes a data_source or not.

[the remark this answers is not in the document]
matt-gardner: No, the `from_params` code figures that out. You can safely remove the `kwargs`.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -53,7 +80,16 @@ def __iter__(self) -> Iterator[Instance]:
...
+            for instance in instances:
+                if self.vocab is not None:
+                    instance.index_fields(self.vocab)
+                yield instance
+
+    def index_with(self, vocab: Vocabulary):
+        self.vocab = vocab
+
+    def __len__(self):
+        return 1
```
bryant1410: Why does it return 1?

DeNeutoy: Yeah.... This is just moving previous behaviour from the iterator onto the dataset instead:

https://github.com/allenai/allennlp/blob/master/allennlp/data/iterators/data_iterator.py#L305

We rely in a couple of places that calling `len` on the iterator (or now, the dataloader) doesn't raise an error. In the case that you have an `IterableDataset` and you call `len`, the pytorch dataloader actually spits out a warning - but we need actually calling it to not crash.

bryant1410: https://github.com/allenai/allennlp/blob/465224f867701f83ec3866b5913cfb96e74a4e80/allennlp/data/iterators/data_iterator.py#L314

I see. Maybe you can add a comment there?

allennlp/data/samplers/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+    return batch.as_tensor_dict(batch.get_padding_lengths())
+
+
+class DataLoader(Registrable, data.DataLoader):
+    def __init__(
+        self,
+        dataset: data.Dataset,
+        batch_size: int = 1,
+        shuffle: bool = False,
+        sampler: Lazy[Sampler] = None,
```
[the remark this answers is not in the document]
DeNeutoy: I see what you mean - we should probably even make this class private, because you should never use it - you should just use the pytorch dataloader.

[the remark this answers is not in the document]
matt-gardner: Making it private means changing type annotations in other code to a private object, which is also unfortunate.  Probably better to just be really clear in the class docstring why this object exists (purely to get type annotations that will let us construct it using our `FromParams` pipeline), and that you can safely ignore it and use pytorch's DataLoader directly if you don't need the `FromParams` stuff.

allennlp/training/trainer.py
```diff
@@ -512,14 +508,16 @@ def _validation_loss(self) -> Tuple[float, int]:
...
-            val_iterator = self.iterator
+            raise ConfigurationError(
+                "Validation results cannot be calculated without a validation_data_loader"
+            )
 
-        val_generator = val_iterator(self._validation_data, num_epochs=1, shuffle=False)
-        num_validation_batches = val_iterator.get_num_batches(self._validation_data)
-        val_generator_tqdm = Tqdm.tqdm(val_generator, total=num_validation_batches)
+        val_generator_tqdm = Tqdm.tqdm(
+            iter(validation_data_loader), total=len(validation_data_loader)
```
dirkgr: Does it not work to just say `Tqdm.tqdm(validation_data_loader)`, and then it will also work if `__len__` is not available?

allennlp/training/trainer.py
```diff
@@ -881,25 +877,27 @@ def from_partial_objects(
         if not optimizer_:
             optimizer_ = Optimizer.default(parameters)
 
-        batches_per_epoch = iterator.get_num_batches(train_data)
-        if batches_per_epoch == 1:  # get_num_batches returns 1 when it can't determine the answer
+        try:
```
matt-gardner: I thought you gave an implementation.  Oh, this is the data loader, not the dataset...  Ok.  But if we never call `len()` on the dataset itself, we don't need a default implementation there anymore, do we?  Or does one of the samplers call `len()` on the dataset?

allennlp/data/samplers/samplers.py
```diff
@@ -0,0 +1,137 @@
...
+class BatchSampler(Registrable):
+    """
+    A copy of the pytorch
+    [BatchSampler](https://pytorch.org/docs/stable/data.html#torch.utils.data.BatchSampler)
+    which allows us to register it with `Registrable.`
+    """
+
+    def __iter__(self) -> Iterable[List[int]]:
+
+        raise NotImplementedError
```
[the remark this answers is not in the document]
DeNeutoy: Because they're abstract base classes, I guess? Should I change it to something else?

allennlp/training/trainer.py
DeNeutoy: oh yep, nice 👍

allennlp/training/trainer.py
```diff
@@ -5,18 +5,21 @@
...
+
 from allennlp.common import Lazy, Tqdm
 from allennlp.common.checks import ConfigurationError, check_for_gpu
 from allennlp.common import util as common_util
-from allennlp.data.instance import Instance
-from allennlp.data.iterators.data_iterator import DataIterator, TensorDict
+
+from allennlp.data import DataLoader
+
+from allennlp.data.iterators.data_iterator import TensorDict
```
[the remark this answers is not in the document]
DeNeutoy: I will, in the next PR which removes the iterator stuff.

allennlp/training/trainer.py
DeNeutoy: The dataloader unfortunately calls len (and warns if you call len on a lazy dataset, but still requires it to be implemented, somewhat bizarrely).

allennlp/data/samplers/samplers.py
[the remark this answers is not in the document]
DeNeutoy: Oh I see, you're asking why none of the subclasses implement them - it's because they all also inherit from the pytorch ones which do implement `__iter__` .

allennlp/data/samplers/bucket_batch_sampler.py
```diff
@@ -0,0 +1,146 @@
...
+        The pytorch `Dataset` of allennlp Instances to bucket.
+    sorting_keys : List[Tuple[str, str]], optional
+        To bucket inputs into batches, we want to group the instances by padding length, so that we
+        minimize the amount of padding necessary per batch. In order to do this, we need to know
+        which fields need what type of padding, and in what order.
+
+        Specifying the right keys for this is a bit cryptic, so if this is not given we try to
+        auto-detect the right keys by iterating once through the data up front, reading all of the
+        padding keys and seeing which one has the longest length.  We use that one for padding.
+        This should give reasonable results in most cases.
```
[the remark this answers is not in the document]
DeNeutoy: I can add that in if you say it's true, but I haven't thought about this deeply 😄

DeNeutoy:
Ok, i'm going to merge this and follow up with two PRs, one removing the iterator code and one updating the training configs.

davidstap:
Is is possible to define batch size with maximum number of tokens (instead of a fixed batch size) in this setup? If so, where does this magic happen (I am unable to locate it)?

matt-gardner:
@davidstap, no, we didn't implement that functionality in the new data code, as it didn't seem like it was widely used.  We do have a `BucketBatchSampler`, which is significantly more efficient than other samplers, but we don't cap by number of tokens.  If you would really like to see it added, please open a new issue for it.

dirkgr:
Tobi already has this feature in his Bart work. If it's needed urgently, we
can split it out of his pull request and get it in now.

davidstap:
That would be awesome @dirkgr

dirkgr:
I just merged that change: https://github.com/allenai/allennlp/blob/master/allennlp/data/samplers/max_tokens_batch_sampler.py

Thanks to @Tobias-Rohde for getting this done!

PR #3703, opened 2020-01-31
Support tokenizer_kwargs and fix lowercasing determination heuristics

ZhaofengWu:


allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -63,16 +63,17 @@ def __init__(
         stride: int = 0,
         truncation_strategy: str = "longest_first",
         calculate_character_offsets: bool = False,
+        **tokenizer_kwargs,
```
matt-gardner: This should be `tokenizer_kwargs: Dict[str, Any] = None`.  The reason for that would be apparent if you tried to construct this using `Tokenizer.from_params`.  Actually, adding a test for that would be good; if it works using this way of doing things, I might be ok with keeping it like it is.  I really don't think it'll work, though.  It'll tell you "extra parameters passed".

ZhaofengWu: Done

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -50,6 +50,7 @@ class PretrainedTransformerTokenizer(Tokenizer):
         - 'do_not_truncate': Do not truncate (raise an error if the input sequence is longer than max_length)
     calculate_character_offsets : `bool`, optional (default=False)
         Attempts to reconstruct character offsets for the instances of Token that this tokenizer produces.
+    tokenizer_kwargs: 'Dict[str, Any]' - dictionary with additional arguments for AutoTokenizer.from_pretrained.
```
matt-gardner: Does this format work with our docs?  Shouldn't the description be on the next line?

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -63,16 +64,19 @@ def __init__(
         stride: int = 0,
         truncation_strategy: str = "longest_first",
         calculate_character_offsets: bool = False,
+        tokenizer_kwargs: Dict[str, Any] = None,
     ) -> None:
-        self.tokenizer = AutoTokenizer.from_pretrained(model_name)
+        self.tokenizer = AutoTokenizer.from_pretrained(
+            model_name, **(tokenizer_kwargs if tokenizer_kwargs is not None else {})
```
matt-gardner: Readability nit: slightly more aesthetically pleasing to just do this:

```python
tokenizer_kwargs = tokenizer_kwargs or {}
self.tokenizer = AutoTokenizer.from_pretrained(model_name, **tokenizer_kwargs)
```

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
ZhaofengWu: Oh just copy pasted from that previous PR. Will update.

PR #3705, opened 2020-01-31
Remove RegularizerApplicator.from_params, move models to **kwargs

matt-gardner:
Getting this right required fixing some of the recursion that happens in `FromParams`, so there are more changes there than you would have expected.  The main changes are some logic fixes in `FromParams` and removing `RegularizerApplicator.from_params`.  The rest of the files that were touched are just updating our models to use newer, better functionality.

This requires a minor change to config files, where for regularizers you need to specify it as `"model": {"regularizer": {"regexes": [("conv", "l1", ...]}}`.  There's a new "regexes" key needed.  This seems reasonable to me, but if you disagree, we can maybe see if there's a way around this (there probably isn't a reasonable one, if we want to remove custom `from_params` methods).

### DIFF
--- a/allennlp/common/from_params.py
+++ b/allennlp/common/from_params.py
@@ -125,6 +125,32 @@ def remove_optional(annotation: type):
         return annotation
 
 
+def infer_params(cls: Type[T],
+                 constructor: Callable[..., T] = None,
+                 ):
+    if constructor is None:
+        constructor = cls.__init__
+
+    signature = inspect.signature(constructor)
+    parameters = dict(signature.parameters)
+
+    has_kwargs = False
+    for param in parameters.values():
+        if param.kind == param.VAR_KEYWORD:
+            has_kwargs = True
+
+    if not has_kwargs:
+        return parameters
+
+    super_class = cls.mro()[1]
+    super_parameters = infer_params(super_class)
+
+    return {
+        **super_parameters,
+        **parameters  # Subclass parameters overwrite superclass ones
+    }
+
+
 def create_kwargs(
     constructor: Callable[..., T], cls: Type[T], params: Params, **extras
 ) -> Dict[str, Any]:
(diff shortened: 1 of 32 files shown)


allennlp/common/from_params.py
```diff
@@ -118,6 +125,27 @@ def remove_optional(annotation: type):
...
+
+    has_kwargs = False
+    for param in parameters.values():
+        if param.kind == param.VAR_KEYWORD:
+            has_kwargs = True
+
+    if not has_kwargs:
+        return parameters
+
+    super_class = cls.mro()[1]
```
DeNeutoy: Is this correct? Don't we need to check which super class inherits from `Registrable` to be agnostic of inheritance order?

DeNeutoy: I.e I thought this was what @mahnerak had fixed in his PR which this supersedes.

allennlp/common/from_params.py
```diff
@@ -171,10 +181,17 @@ def create_kwargs(
...
         # and an __args__ field indicating `(str, int)`. We capture both.
         annotation = remove_optional(param.annotation)
-        kwargs[param_name] = construct_arg(
+
+        constructed_arg = pop_and_construct_arg(
             cls.__name__, param_name, annotation, param.default, params, **extras
         )
 
+        # If we just ended up constructing the default value for the parameter, we can just omit it.
+        # Leaving it in can cause issues with **kwargs in some corner cases, where you might end up
```
DeNeutoy: Could you give an example of such a corner case? I can't think of how this could occur.

allennlp/tests/fixtures/crf_tagger/experiment.json
```diff
@@ -51,9 +51,11 @@
       "dropout": 0.0,
       "bidirectional": true
     },
-    "regularizer": [
-      ["transitions$", {"type": "l2", "alpha": 0.01}]
-    ]
+    "regularizer": {
+      "regexes": [
```
DeNeutoy: We need to start keeping track of breaking config changes if we aren't going to update them in the models as we go, because otherwise doing them all at once is going to be a nightmare. Can you add this to the release notes also?

allennlp/common/from_params.py
matt-gardner: @mahnerak was dealing with two problems, which can be seen in the added tests: (1) deeper inheritance trees, where you need to check the superclass's superclass, and (2) when you (for some reason) override the type of an argument in a subclass; in this case the subclass's type annotation should be used for inference, not the superclass's.  That PR did not deal with multiple inheritance, which is what you're worried about here.  That's still a potential problem.

matt-gardner: Looks like there was a comment that got dropped in that PR, though, which I'm adding back in, and giving a note that this could fail in certain cases of multiple inheritance.

allennlp/common/from_params.py
matt-gardner: Added an example to the comment.  If you do something like `super().__init__(lazy=True, **kwargs)` in a dataset reader, without this `if` you'd end up with both `lazy=True` (from your first argument) and `lazy=False` (inside of **kwargs, because we constructed the default value) in `DatasetReader.__init__`.

allennlp/tests/fixtures/crf_tagger/experiment.json
matt-gardner: Yeah, thanks for the reminder, I added it.  I am volunteering to make all of the changes to our saved model archives.  You're right, it will be a pain, but I have the most context on what all of the changes are and how to fix them.

allennlp/common/from_params.py
DeNeutoy: would it make sense to just fix the multiple inheritance case here, because we actually do use this frequently with `nn.Module`? 

e.g 

```python
super_class = None
super_classes = cls.mro()
while super_class is None:
	candidate = super_classes.pop(0)

	if issubclass(candidate, Registrable):
		super_class = candidate

```
or something?

matt-gardner: Hmm, yeah, I see your point.  Let me open a new issue for that.  It should be a pretty easy, self-contained PR to get that to work.

matt-gardner:
Something is up with our TC integration (and thus also bulldozer).  CI passes (http://build.allennlp.org/viewLog.html?buildId=25286&tab=buildResultsDiv&buildTypeId=AllenNLP_AllenNLPPullRequests), so I'm merging this.

PR #3708, opened 2020-02-01
Fix `allennlp test-install`

brendan-ai2:
- https://github.com/allenai/allennlp/pull/3616 made matplotlib optional, but this broke our test that verified pip installs.
- http://build.allennlp.org/viewLog.html?buildId=24194&buildTypeId=AllenNLP_AllenNLPCommits&tab=buildChangesDiv
- Simply skip the test if matplotlib isn't installed.


brendan-ai2:
I considered just always installing `dev-requirements.txt` for testing, but I worry we'd quickly break pip installs. The downside with this method is that people have to either install dependencies manually when alerted via an exception or know about an obscure environment variable. But given that the original change relied on the exception alerts I think this is okay. Happy to hear better ideas...

DeNeutoy:
* We install dev requirements in PR CI, and presumably we use the Dockerfile.pip for the pip builds, so I think we'd be able to catch breakages still?

brendan-ai2:
@DeNeutoy, good call, done. Please take another look.

Issue #3711, opened 2020-02-01
Implementation of RMS Normalization


avinashsai:
**Is your feature request related to a problem? Please describe.**
RMS Normalization https://openreview.net/pdf?id=SygkZ3MTJE is a normalization technique proposed for normalizing CNN and LSTM.

**Describe the solution you'd like**
I have experimented Layer Normalization and RMS Normalization using LSTM on IMDB dataset. Experiment results can be found here: https://github.com/avinashsai/All-you-need-to-know-about-Normalization/blob/master/Experiments/Normalization_IMDB.ipynb. I would like to include this normalization technique in the framework. Please share your thoughts on this.  

Thank you

avinashsai:
@matt-gardner please share your thoughts on this
Thank you.

avinashsai:
@matt-gardner Please share your thoughts on this.
Thank you

schmmd:
@avinashsai what's your motivation for adding this to the library?  I don't know much about RMS normalization.

avinashsai:
RMS Normalization is computationally simpler to Layer Normalization. There are very few normalization layers proposed for RNN. RMS Normazation is one of them.

DeNeutoy:
Hi @avinashsai, 

Thanks for the contribution. Unfortunately this paper does not demonstrate statistically significant performance improvements over LayerNorm, and the speedup also seems to be marginal. It's great code to experiment with, but I don't think it needs to be in allennlp.

schmmd:
Closing as we are not convinced this feature should be in AllenNLP.

PR #3713, opened 2020-02-03
RMS Normalization

avinashsai:
Added code for RMS Normalization as mentioned in the issue #3711.

### DIFF
--- a/allennlp/modules/rms_norm.py
+++ b/allennlp/modules/rms_norm.py
@@ -1,0 +1,30 @@
+import torch
+
+
+class RMSNorm(nn.Module):
+    """
+    An implementation of [RMS Normalization](
+    https://openreview.net/pdf?id=SygkZ3MTJE.
+    RMS Normalization stabilises the training of deep neural networks by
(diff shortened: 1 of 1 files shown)


schmmd:
Thanks for the contribution!  I removed @matt-gardner as a reviewer since we typically decide who should review PRs internally.

It'd be helpful to have an example where this normalization is used effectively.  Without that, it's hard to know whether this contribution is implemented correctly.

Also, this PR isn't passing code review.

avinashsai:
I have compared Layer Normalization and RMS Normalization on IMDB Text Classification. The link to notebook https://github.com/avinashsai/All-you-need-to-know-about-Normalization/blob/master/Experiments/Normalization_IMDB.ipynb

DeNeutoy:
See comments in #3711 - code looks great, but probably better off in a github repo that you oversee! Thanks for the idea though. In the future, if you give us time to give some feedback on the issue, we'll be able to let you know if we want the contribution or not. Thanks again!

PR #3717, opened 2020-02-03
Configure Bulldozer to update PRs when tagged with 'Update Me'.

schmmd:


.bulldozer.yml
```diff
@@ -53,14 +53,14 @@ merge:
 
 ## "update" defines how and when to update pull request branches. Unlike with
 ## merges, if this section is missing, bulldozer will not update any pull requests.
-#update:
+update:
 #  # "whitelist" defines the set of pull requests that should be updated by
 #  # bulldozer. It accepts the same keys as the whitelist in the "merge" block.
 #  whitelist:
-#    labels: ["WIP", "Update Me"]
+    labels: ["Update Me"]
```
brendan-ai2: If I understand the semantics correctly, I think this should include "merge when ready" so that we don't have to apply multiple labels.

schmmd: I kinda agree but Vivek thought this wasn't a good idea.  I thought I'd try this as a first step and possibly take another once we've had the change to play around with it a bit more.

Is it possible we'll want "Merge When Ready" without "Update Me"?  I don't know...

brendan-ai2: @viveklak , what was the issue here?

I don't think we want "Merge When Ready" without updating. If a user wants find grained control over what gets merged in, they should just merge manually.

matt-gardner: I second @brendan-ai2's comments here.

viveklak: I just haven't tried using the same label for the two actions before. @schmmd I can try it out on a smaller repo and report back.

schmmd: We can just try it on AllenNLP and revert if we have issues.  I pushed that change just now.

viveklak: I just tried it out. It seems to work. 

Just a caveat that the way bulldozer is written, the only time updates are evaluated is when there is a push to the remote origin branch (e.g. master). So once you set the whitelisted label on the PR, any new merges to master from that point on will result in an attempted update (unless there are conflicts).  Setting the whitelisted label after the push to master will not result in an update of the PR automatically.

.bulldozer.yml
```diff
@@ -53,14 +53,14 @@ merge:
 
 ## "update" defines how and when to update pull request branches. Unlike with
 ## merges, if this section is missing, bulldozer will not update any pull requests.
-#update:
+update:
 #  # "whitelist" defines the set of pull requests that should be updated by
 #  # bulldozer. It accepts the same keys as the whitelist in the "merge" block.
 #  whitelist:
```
bryant1410: @schmmd shouldn't the elements `whitelist:` and `blacklist:` be uncommented as well? Otherwise, it shouldn't work well IMHO.

schmmd: Hmmm... not quite sure how I made that mistake... thanks.

schmmd: See https://github.com/allenai/allennlp/pull/3727

PR #3718, opened 2020-02-03
Automatic batches per epoch

dirkgr:
This is about feeding `num_epochs` and `batches_per_epoch` to the learning rate schedulers automatically.

### DIFF
--- a/allennlp/modules/token_embedders/pretrained_transformer_embedder.py
+++ b/allennlp/modules/token_embedders/pretrained_transformer_embedder.py
@@ -46,6 +46,13 @@ class PretrainedTransformerEmbedder(TokenEmbedder):
     def get_output_dim(self):
         return self.output_dim
 
+    def _number_of_token_type_embeddings(self):
+        config = self.transformer_model.config
+        if hasattr(config, "type_vocab_size"):
+            return config.type_vocab_size
+        else:
+            return 0
+
     @overrides
     def forward(
         self,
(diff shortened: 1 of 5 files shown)


dirkgr:
At this point, this uses `inspect` to find out whether the learning rate scheduler constructor takes two parameters, `num_epochs` and `batches_per_epoch`. If I commit to this approach, I can make a more minimal PR, where any learning rate scheduler that has parameters with that name will see them filled in with magic. I like that it's minimal, but I don't like that it's magic.

The alternative is to treat `num_epochs` and `batches_per_epoch` exactly like `optimizer`. Every learning rate scheduler has to take those parameters, and can then ignore them if it doesn't need them. I like that this is explicit, no magic, but it touches all the learning rate schedulers.

matt-gardner:
I *think* we can handle this just using `from_params`.  The way it handles `extras` in the code should allow us to pass in the extra parameters even if they aren't used by a particular scheduler.  If you're willing to wait a couple of days for me to get to this, I'm happy to figure out how to get this working right.  I think `inspect` is a bit unfortunate to use outside of `FromParams`.

matt-gardner:
What's your vote @dirkgr? If you want me to try to look at this, I can do it starting in about half an hour.

dirkgr:
I'm for it. If you can fold this into the general `from_params` re-work, that would be better than either of the hacks I outlined above.

matt-gardner:
Closing in favor of #3734.

PR #3725, opened 2020-02-04
New transformers version

dirkgr:
`roberta-large` in the `transformers` library has some bugs involving token type ids. They are fixed in 2.4.0 and above.

I'm getting the feeling I'm the only person in the world using token type ids for anything.


bryant1410:
Related to https://github.com/allenai/allennlp/pull/3692#issuecomment-579941448, I think the transformers library could be pinned, so dependabot sends a PR, we see if tests pass or not and the changelog and merge. This way we're always sure allennlp works well with the latest version, semi-automatically.

PR #3726, opened 2020-02-04
update docs for nn

DeNeutoy:


schmmd:
So my understanding is that "merge when ready" won't merge in master until a new commit is pushed.  We can test this out on this PR if we'd like to...

Issue #2167, opened 2018-12-11
Change "unable to check gpu_memory_mb()" error to a warning, remove uninformative stack trace


magic282:
**Describe the bug**

```
2018-12-10 23:59:23,547 - ERROR - allennlp.common.util - unable to check gpu_memory_mb(), continuing
Traceback (most recent call last):
  File "/home/v-qizhou/anaconda3/envs/allennlp/lib/python3.6/site-packages/allennlp/common/util.py", line 343, in gpu_memory_mb
    encoding='utf-8')
  File "/home/v-qizhou/anaconda3/envs/allennlp/lib/python3.6/subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "/home/v-qizhou/anaconda3/envs/allennlp/lib/python3.6/subprocess.py", line 403, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/home/v-qizhou/anaconda3/envs/allennlp/lib/python3.6/subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
  File "/home/v-qizhou/anaconda3/envs/allennlp/lib/python3.6/subprocess.py", line 1275, in _execute_child
    restore_signals, start_new_session, preexec_fn)
OSError: [Errno 12] Cannot allocate memory
2018-12-10 23:59:23,547 - INFO - allennlp.training.trainer - Training
```

**To Reproduce**
No idea how to reproduce.

**Expected behavior**

**System (please complete the following information):**
 - OS: Linux
 - Python version: 3.6.7
 - AllenNLP version: 0.7.2
 - PyTorch version: (if you installed it yourself)

**Additional context**
Under linux `screen`

DeNeutoy:
This isn't critical functionality and we actually expect it to fail occasionally, which is why we just continue when it happens. We probably won't be able to fix this, sorry!

WindChimeRan:
Same error.
Thanks.
I think it'd be better if you tell us this is not an error in the log, not here.

guoquan:
Maybe better to just put a warning instead of an error at L461.
And document the exact potential exception there.
https://github.com/allenai/allennlp/blob/55147761ca3b46a9ce994532e2451af95088258c/allennlp/common/util.py#L436-L462

matt-gardner:
@guoquan (and @WindChimeRan), I agree with this. I'm re-opening this issue, and renaming it.  Should be an easy fix; PR welcome!

epwalsh:
This should no longer be an issue

PR #3729, opened 2020-02-05
Update gpu_memory_mb() not to log at error level

guoquan:
According to the discussion in #2167 , there are known occasional failures in this non-critical function.
Log as a warning with the stack trace and error information.


guoquan:
Hi, it seems it is impossible to pass this coverage rate test because of the commit is 100% in an `except` clause that is rarely reached. Even if I test this function, all the runs would be trapped in the first except in the testing environment (if there is no GPU). Any idea?

DeNeutoy:
@guoquan no need to pass the coverage tests for such a change, thanks for the contribution!

Issue #3579, opened 2020-01-02
add documentation on for new LearningRateScheduler interface


DSLituiev:
I am refactoring old code and see that `LearningRateWithoutMetricsWrapper` has been removed, and no documentation on how to use new interface available.

I have a following code:
```
from torch.optim.lr_scheduler import MultiStepLR
from allennlp.training.learning_rate_schedulers import LearningRateWithoutMetricsWrapper

learning_rate_scheduler = LearningRateWithoutMetricsWrapper(MultiStepLR(optimizer,
                                                                        [10, 20, 40],
                                                                        gamma=0.25, last_epoch=-1))
```

 Looking into the [commits](https://github.com/allenai/allennlp/pull/2345/commits/12085f62ee0253dca224b77f9640d1f411ed1273), I guessed it must be something like:

```
from torch.optim.lr_scheduler import MultiStepLR
from allennlp.training.learning_rate_schedulers import LearningRateScheduler

optimizer = optim.SGD(model.parameters(), lr=0.1)
learning_rate_scheduler = LearningRateScheduler(optimizer, dict(type="MultiStepLR",
                                                                milestones=[10, 20, 40],
                                                                # initial_lr=0.1,
                                                                gamma=0.25, last_epoch=-1))
```

But that fails with error (whether `initial_lr=0.1` is included or not):
```
/Applications/anaconda3/envs/nlp/lib/python3.7/site-packages/allennlp/training/learning_rate_schedulers/learning_rate_scheduler.py in __init__(self, optimizer, last_epoch)
     12 class LearningRateScheduler(Scheduler, Registrable):
     13     def __init__(self, optimizer: torch.optim.Optimizer, last_epoch: int = -1) -> None:
---> 14         super().__init__(optimizer, "lr", last_epoch)
     15 
     16     def get_values(self) -> None:

/Applications/anaconda3/envs/nlp/lib/python3.7/site-packages/allennlp/training/scheduler.py in __init__(self, optimizer, param_group_field, last_epoch)
     42                 if self._initial_param_group_field not in group:
     43                     raise KeyError(
---> 44                         f"{self._initial_param_group_field} missing from param_groups[{i}]"
     45                     )
     46         self.base_values = [

KeyError: 'initial_lr missing from param_groups[0]'
```

dirkgr:
Try creating the learning rate scheduler like this:
```
LearningRateScheduler.from_params(
    optimizer, dict(
        type="MultiStepLR",
[...]

PR #3734, opened 2020-02-05
Removed LearningRateScheduler.from_params

### LINKED ISSUE #3707: Learning rate scheduler should not need to be told about the number of steps per epoch
Some of the learning rate schedulers take a `num_steps_per_epoch` parameter. It breaks several abstractions to provide it this way, and in the config. At least when we have a non-lazy reader, we should determine that number automatically.

matt-gardner:
Supersedes #3718.  Fixes #3707.

In removing the custom `from_params` method, it turned out to be easier to also remove the abuse of the registry that we were doing with using pytorch classes - it's not hard to just write our own typed wrappers for these classes that we can properly inspect.  Maintenance burden on those classes should be extremely low, as they are just constructors and that's it.  A side effect of this is that it's also much easier to use these classes in your own script without needing to use `from_params` (cf. #3579; this would have made that a non-issue).

I might end up doing the same thing for optimizers and initializers, though there are more of those than there are schedulers, so it would be more annoying.  We'll see, when I get to those classes (maybe tomorrow).


allennlp/common/from_params.py
```diff
@@ -442,7 +442,12 @@ def construct_arg(
         subextras = create_extras(value_cls, extras)
 
         def constructor(**kwargs):
-            return value_cls.from_params(params=popped_params, **kwargs, **subextras)
+            # If there are duplicate keys between subextras and kwargs, this will overwrite the ones
+            # in subextras with what's in kwargs.  If an argument shows up twice, we should take it
+            # from what's passed to Lazy.construct() instead of what we got from create_extras().
+            # Almost certainly these will be identical objects, anyway.
+            subextras.update(kwargs)
```
dirkgr: This will modify `subextras`. That makes me nervous. We should never call this twice, but if we do it'll apply left-over arguments from the previous time we called it.

In the context of this function, it makes sense to think that `kwargs` should take precedence, but in the context of the whole system, it should be the other way around. If I explicitly specify `num_epochs` in the config for the learning rate scheduler, I expect that value to take effect. As it is, it will be overridden by the values the trainer determines to be right.

At least, if we override an explicit value, we should print a warning.

matt-gardner: I thought about the fact that we're modifying `subextras`, but it's created as a new dictionary in `create_extras`, so modifying it is not a problem.  It's pretty clearly outside the intended usage of the `Lazy` object to call `construct()` multiple times, and I'd rather not program so defensively that I worry about misuse of `Lazy` in here.

Which one should take precedence / should we give a warning is a separate issue.  I think you're misunderstanding what comes from where in this method.  Things that you specify in a config file will end up in `popped_params`, not `subextras` or `kwargs`.  These will take precedence, yes.

The things in `subextras` are things that came from `**kwargs` higher up in the object hierarchy.  This is like, `Model.from_params(params=params["model"], vocab=vocab)`.  `vocab` here ends up in `subextras` as an available argument if a particular constructor wants it, but is ignored if the constructor doesn't need it (this is how, e.g., the `Embedding` gets the `Vocabulary` - the `TextFieldEmbedder` passes it on from the model using this mechanism).

The things in `**kwargs` in this context are things that are passed directly to `lazy_object.construct()`.  You shouldn't need to pass things along that are already passed along in `subextras`, but if you do on accident, you would get a crash because the same key showed up twice.  I'm pretty sure I ran into this when I was doing `learning_rate_scheduler.construct(num_epochs=num_epochs)`.  `num_epochs` in that setting comes from `**kwargs`.  But it also was in `subextras`, because it gets passed along as an available argument to subclasses if they request it, using the mechanism I described above.  So, I should probably have just removed the `num_epochs` argument to the `construct()` method that I called, but that actually makes the code a little less clear, I think, and it's easy to handle that case here, anyway.

If I had *also* directly specified `num_epochs` in the parameters for the learning rate scheduler, I'm not sure how that's handled.  It might do the right thing, and it might crash.  I'll add a case for that tomorrow in my test and make sure that it does the right thing.

matt-gardner: Turns out that if a key was passed separately in the config file, we would silently only take what's passed from above, though that had nothing to do with the logic here.  I changed it in 10f0a43 anyway.

dirkgr: Makes sense, thanks for the explanation!

I would want to suppress the new warning when the values are the same, but for some possible values equality might be hard to establish consistently, so this is fine as it is.

matt-gardner: Yeah, I thought about checking for equality.  The trouble is that at the place we have that check, we haven't yet constructed the parameter, so we don't have what we need to do that check.  If we run into an issue later, we can maybe do the restructuring that's needed to make that better.

PR #3735, opened 2020-02-05
Removing more custom from_params methods

matt-gardner:
Some of these ones break backwards compatibility slightly, but in ways that we aren't very concerned about.  The functionality is still available (either through a config file change, with the `Tokenizer`, or by just not using config files, with the `Elmo` ones), just moved.

The added lines are from a vocabulary test fixture that got automatically generated.  This is really a small PR that is almost entirely removing code.


matt-gardner:
Ack, thanks for the reminder.  Doing it now.

PR #3736, opened 2020-02-06
max_instances for DatasetReader

dirkgr:
This is from a comment on the TransformerQA PR. Rather than having a one-off limitation on reader activity, this gives the ability to all the readers.

I had to change how the dataset reader tests are run to make `@pytest.mark.parametrize` work.


allennlp/tests/data/dataset_readers/dataset_reader_test.py
```diff
@@ -1,30 +1,31 @@
...
-        super().setUp()
-        self.cache_directory = str(self.FIXTURES_ROOT / "data_cache" / "with_prefix")
+class TestDatasetReader:
+    cache_directory = str(AllenNlpTestCase.FIXTURES_ROOT / "data_cache" / "with_prefix")
 
-    def tearDown(self):
-        super().tearDown()
-        shutil.rmtree(self.cache_directory)
+    @pytest.yield_fixture(autouse=True)
+    def cache_directory_fixture(self):
```
DeNeutoy: https://docs.pytest.org/en/latest/yieldfixture.html

FYI, looks like this is deprecated for pytest > 3.0 (I have 4.3.1 installed) and you can just `yield` from a `pytest.fixture()` instead.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -84,8 +85,14 @@ class DatasetReader(Registrable):
         parameters you set for this DatasetReader!_
     """
 
-    def __init__(self, lazy: bool = False, cache_directory: str = None) -> None:
+    def __init__(
+        self,
+        lazy: bool = False,
+        cache_directory: Optional[str] = None,
+        max_instances: Optional[int] = None,
```
matt-gardner: Can you add this to the docstring?

allennlp/tests/data/dataset_readers/dataset_reader_test.py
dirkgr: fa7b32498974f04bf841f3185b6c9fdf34799990

allennlp/data/dataset_readers/dataset_reader.py
dirkgr: bb753f02638cabaf03a207272d49647179fd6ba4

PR #3739, opened 2020-02-06
Removing custom from_params from initializers

matt-gardner:
This PR adds a level of nesting to configuration files, similar to what I did with regularizers in #3705.  In this case, that simplifies quite a bit how the "prevent" stuff worked, as we don't have to do any special detection for it anymore.  This PR also adds actual classes for initialization functions that we can inspect and instantiate, similar to what was done in #3734.  Some justification for these classes from that PR:

> In removing the custom from_params method, it turned out to be easier to also remove the abuse of the registry that we were doing with using pytorch classes - it's not hard to just write our own typed wrappers for these classes that we can properly inspect. Maintenance burden on those classes should be extremely low, as they are just constructors and that's it. A side effect of this is that it's also much easier to use these classes in your own script without needing to use from_params (cf. #3579; this would have made that a non-issue).


matt-gardner:
Just updated release notes to account for this change.  Merging.

PR #3743, opened 2020-02-07
Removed custom Optimizer.from_params

### LINKED ISSUE #3389: Completely remove custom from_params methods

matt-gardner:
Fixes #3389.  This is the last of them, except for some custom methods in the callbacks, which I'm not going to deal with until we decide what to do about the trainer.

### DIFF
--- a/allennlp/training/optimizers.py
+++ b/allennlp/training/optimizers.py
@@ -21,7 +21,7 @@ The available optimizers are
 import logging
 import re
 import math
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Tuple, Union
 
 import torch
 import transformers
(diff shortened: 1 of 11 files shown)


matt-gardner:
The diff looks larger than it actually is, because I took a large block of code from a class method and moved it into a function in the module.

allennlp/training/optimizers.py
```diff
@@ -31,143 +31,332 @@
 logger = logging.getLogger(__name__)
 
 
+def make_parameter_groups(
+    model_parameters: List[torch.nn.Parameter],
+    groups: List[Tuple[List[str], Dict[str, Any]]] = None,
+):
```
brendan-ai2: Return type and brief doc string would be nice here.

allennlp/training/optimizers.py
```diff
@@ -31,143 +31,332 @@
...
-        # Log the number of parameters to optimize
-        num_parameters = 0
-        for parameter_group in parameter_groups:
-            if isinstance(parameter_group, dict):
-                num_parameters += sum(parameter.numel() for parameter in parameter_group["params"])
-            else:
-                num_parameters += parameter_group.numel()
-        logger.info("Number of trainable parameters: %s", num_parameters)
-
-        # By default we cast things that e.g. look like floats to floats before handing them
```
brendan-ai2: So if I understand correctly, this just always happens now? I don't see any usages of it in the negative (even with `git log -S infer_type_and_cast -p`), so it's unclear to me why the option existed in the first place. Guessing that was your thinking too... :)

matt-gardner: Basically, that was added because of craziness with using `std.extvar` (or something) in jsonnet / environment variables with json.  Now that we have explicit constructors, it is entirely unnecessary.

allennlp/training/optimizers.py
matt-gardner: In moving some comments around to make this better, I noticed that I may have messed up some of the implementation.  Our tests should have caught it, though, so either the comment is wrong or I missed something.  Going to double check that (and add a new test if it is in fact broken) before merging this.

brendan-ai2: 👍

matt-gardner: [This test](https://github.com/allenai/allennlp/blob/ff3ed1c89ed10e55da5532ad3fcde9d89747d431/allennlp/tests/training/optimizer_test.py#L39-L62) checks for what I thought was off in the comment.  The test passes, so the comment is wrong.  Merging after adding an updated docstring.

PR #3746, opened 2020-02-09
Removing old BERT files

### LINKED ISSUE #3305: Update BERT-specific embedders to new pretrained transformer indexers and embedders
* Drop dependencies on pytorch-pretrained-bert and pytorch-transformers (depend on transformers instead).
* SOTA models fine-tune transformers so we should make this easy.  Specifically we should review transformers tutorials, verify that those tasks are easily accomplishable in AllenNLP, and write equivalent tutorials for AllenNLP.

matt-gardner:
Fixes #3305.


allennlp/common/util.py
```diff
@@ -62,6 +64,27 @@
 ContextManagerFunctionReturnType = Generator[T, None, None]
 
 
+class PretrainedTransformer:
```
DeNeutoy: I know this wasn't changed in this PR, but I think having this class is a _super_ bad idea - it has reallly gross side effects if you modify the model, e.g setting `requires_grad`. We should just eat the cost of loading the model twice. It takes < 10 seconds and uses max ~2GB memory.

matt-gardner: Yeah, fair enough, there aren't a lot of places where we'd actually want this caching.  I just removed it.

matt-gardner:
TC and github are not playing nicely again, not really sure why.  But the build is passing if you go to build.allennlp.org and find this PR, so I'm going to merge this.

bryant1410:
> TC and github are not playing nicely again, not really sure why. But the build is passing if you go to build.allennlp.org and find this PR, so I'm going to merge this.

Have you folks considered GH Actions (at least for most tests)? I'm surprised by how fast it runs things.

matt-gardner:
We use it for all of our new repos; we set up teamcity before GH actions was available.  I don't deal much with the build system, but there might be complexities that we have that make it harder for GH actions (e.g., I think we now run multi-GPU tests on team city - will GH actions let us actually run on a machine with 2 GPUs?).

bryant1410:
I see. Yeah, I think that for multi-GPU it'd be hard, but you can keep TC for those and GH actions for most, which I bet will fail less frequently.

allennlp/modules/seq2vec_encoders/cls_pooler.py
```diff
@@ -0,0 +1,41 @@
...
+
+    @overrides
+    def get_output_dim(self) -> int:
+        return self._embedding_dim
+
+    @overrides
+    def forward(self, tokens: torch.Tensor, mask: torch.Tensor = None):
+        # tokens is assumed to have shape (batch_size, sequence_length, embedding_dim).  We just
+        # want the first token for each instance in the batch.
+        return tokens[:, 0]
```
ZhaofengWu: Except for XLNet which puts [cls] at the _end_ of the sequence. Maybe it's better to call this class `FirstTokenPooler` or something?

matt-gardner: Ugh, yeah, that's annoying to handle right in something like this, because you want the original text field tensors.  Care to submit a PR to change the name?

PR #3747, opened 2020-02-09
Removing fine-tune command by enabling train to handle fine tuning

### LINKED ISSUE #3521: Consider removing the fine-tune command
Users are often confused about how to use this, it can be conflated with BERT fine-tuning, and it's only applicable to a narrow use case.  It's brittle.

### LINKED ISSUE #3690: Re-enable vocabulary extension
#3627 removed the ability to extend vocabulary when training or fine-tuning a model, because it greatly complicated the logic, and there are now easier ways of accomplishing the same thing.  I think it's pretty easy to add back in by having a `from_archive` `Model` type (which we should have anyway, to replace the `fine-tune` command), which gets passed a `Vocabulary` that it can optionally use to extend its vocab.  We should also have a Vocabulary type that is "empty" or "none", for cases where you're just loading a model and don't want to iterate through the data and compute a vocab for no reason.

matt-gardner:
Fixes #3521. Fixes #3690. 

With the improvements to the `FromParams` pipeline I've been doing, it became very easy to entirely replace the `fine-tune` command with a `from_archive` `Model` type.  This will load a model from a given archive and train it on the data in your config file, no separate command needed.  While I was doing this, I also made it possible to extend the vocabulary in the model that you load, which is now done without all of the crazy command line arguments that were needed in previous iterations of the vocab extension code.


allennlp/tests/commands/train_test.py
```diff
@@ -251,6 +251,21 @@ def test_train_model_can_instantiate_from_params(self):
...
+        train_loop = TrainModel.from_params(
+            params=params, serialization_dir=self.TEST_DIR, local_rank=0, batch_weight_key=""
+        )
+        train_loop.run()
+
+        # This is checking that the vocabulary actually got extended.  This token was not in the
+        # original training data for the model that we are fine-tuning, but is in the new training
+        # data, and the vocabulary settings in the experiment config are such that this token should
+        # have been added.
+        assert "journalism" in train_loop.model.vocab.get_token_to_index_vocabulary()
```
DeNeutoy: Instead of this test relying on the data, could you just compare the vocab size or something? Is it guaranteed to be larger?

allennlp/models/model.py
```diff
@@ -365,9 +365,33 @@ def extend_embedder_vocab(self, embedding_sources_mapping: Dict[str, str] = None
                     self.vocab, extension_pretrained_file=pretrained_file, model_path=model_path
                 )
 
+    @classmethod
+    def from_archive(cls, archive_file: str, vocab: Vocabulary = None) -> "Model":
```
DeNeutoy: Probably a dumb question, but how does this work from a config file without the `Lazy`  parameter? Is `Vocabulary` an exception to that rule?

allennlp/tests/commands/train_test.py
matt-gardner: Hmm, ok, yeah, I could do that.  I'll have to load the original model again separately, which I hadn't thought of doing before, but it should work just fine.

allennlp/models/model.py
matt-gardner: This does make use of `Lazy`, but it's [these `Lazy`s](https://github.com/allenai/allennlp/blob/eb9abcf370fed33dc0e47b8420c2f5b96de0de70/allennlp/commands/train.py#L532-L535) that get leveraged.  By the time we get to this class, the `Vocabulary` has already been constructed, so we don't need another `Lazy` annotation (for the same reason that `Model.__init__` also doesn't need `Lazy[Vocabulary]` - that gets passed in using the `**extras` stuff in `FromParams`, seen in this code in [this line](https://github.com/allenai/allennlp/blob/eb9abcf370fed33dc0e47b8420c2f5b96de0de70/allennlp/commands/train.py#L629)).  `Lazy` is only necessary if you have multiple arguments to a *single function* that have sequential dependencies, like a `Vocabulary` that has its own parameters but depends on training data for its construction.

vin-ivar:
Is it possible to have this extend a model without explicitly loading it from an archive? I already have the model that `train_model` returned, I'd have used something like `fine_tune_model(model, config)` earlier to avoid explicitly having to think about paths.

PR #3748, opened 2020-02-10
Use mathy_pydoc for better docs type formatting

### LINKED ISSUE #3731: Optional[] annotations show up poorly in our docs
Something in the docs script converts `Optional[X]` into `Union[X, NoneType]`, which is unfortunate and distracting in the docs:

Screen Shot 2020-02-05 at 7 12 43 AM

This makes me even more in favor of just removing `Optional[]` annotations from our code.

### LINKED ISSUE #3738: Figure out how to put signatures on multiple lines in docs, instead of having them scroll vertically
Issue title should be self-explanatory.  For more evidence that this is an issue, see the end of this comment: https://github.com/allenai/allennlp/issues/3731#issuecomment-582859101, which I agree with.

justindujardin:
I saw @DeNeutoy's note in #3738 and it was easy to publish my custom version of pydoc-markdown to pypi. 

I've also been bothered by the ugly printing of things like `Union[str, NoneType]` instead of `Optional[str]` and `_ForwardRef('TheType')` instead of `TheType`, so I used a few regex's to prettify them.

It looks like AllenNLP uses a mathy-inspired API docs build script (:heart:) so it was an easy drop-in to use my custom version. 

Before:
Screen Shot 2020-02-09 at 4 20 37 PM

After:
Screen Shot 2020-02-09 at 5 08 36 PM

This should fix #3731 and fix #3738


DeNeutoy:
Merging, looks like our CI is not reporting to github 🤷‍♂

PR #3750, opened 2020-02-10
Remove unused args from fine-tune

bryant1410:


allennlp/commands/train.py
```diff
@@ -336,10 +318,6 @@ def _train_worker(
     master_port: int = 29500,
     world_size: int = 1,
     distributed_device_ids: List[str] = None,
```
bryant1410: `recover` is unused here. I wonder why.

allennlp/commands/train.py
```diff
@@ -564,6 +550,7 @@ def from_partial_objects(
             The directory where logs and model archives will be saved.
         local_rank: `int`
```
bryant1410: `local_rank` is unused. Can it be removed from here?

allennlp/modules/token_embedders/embedding.py
```diff
@@ -194,7 +194,7 @@ def extend_vocab(
         model_path : `str`, (optional, default=None)
             Path traversing the model attributes upto this embedding module.
             Eg. "_text_field_embedder.token_embedder_tokens". This is only useful
-            to give helpful error message when extend_vocab is implicitly called
+            to give a helpful error message when extend_vocab is implicitly called
             by fine-tune or any other command.
```
bryant1410: I noticed that `Embedding.extend_vocab` is used only in tests now. Should it be removed?

allennlp/commands/train.py
matt-gardner: It's not actually unused, it gets passed along behind the scenes to the `Trainer`.

allennlp/modules/token_embedders/embedding.py
matt-gardner: It gets called here: https://github.com/allenai/allennlp/blob/c4559f3751775aa8bc018db417edc119d29d8051/allennlp/models/model.py#L364-L366.

allennlp/commands/train.py
matt-gardner: I traced the logic as I was doing the `Lazy[]` refactoring, and I was pretty sure it was only used when creating the serialization directory (we check to see if it already exists).  There was a use in vocab extension, but the refactoring made that unnecessary.  I didn't do a thorough check of the `recover` pipeline, because I'm pretty sure we have tests that check it.

At this point I *think* this parameter is unnecessary for this function and can be removed.  If you want to check that and remove it in a subsequent PR, that would be awesome.

matt-gardner:
Speaking of multi-GPU tests, it looks like this PR broke one of them: http://build.allennlp.org/viewLog.html?buildTypeId=AllenNLP_GpuUnitTests&buildId=25644&tab=buildLog.  Would you mind digging into this error?  If you're not going to have time to look at it soon, let us know.

bryant1410:
> Speaking of multi-GPU tests, it looks like this PR broke one of them: http://build.allennlp.org/viewLog.html?buildTypeId=AllenNLP_GpuUnitTests&buildId=25644&tab=buildLog. Would you mind digging into this error? If you're not going to have time to look at it soon, let us know.

Oh, looking at it now.

bryant1410:
It's passing here:

```
$ pytest -k "test_train_model_distributed"
======================================================================================================== test session starts ========================================================================================================
platform linux -- Python 3.7.4, pytest-4.4.1, py-1.8.0, pluggy-0.9.0
rootdir: /***/allennlp, inifile: pytest.ini, testpaths: allennlp/tests/
plugins: cov-2.8.1, flaky-3.5.3
collected 1171 items / 1170 deselected / 1 selected                                                                                                                                                                                 

allennlp/tests/commands/train_test.py .                                                                                                                                                                                       [100%]
===Flaky Test Report===


===End Flaky Test Report===

============================================================================================ 1 passed, 1170 deselected in 15.64 seconds =============================================================================================
```

It's not being skipped:

```
$ python -c "import torch; print(torch.cuda.device_count())"
4
```

I think it's just flaky, and I don't find a correlation with the change.

matt-gardner:
Thanks for checking; I'm running the build again.

matt-gardner:
Re-running it succeeded.

bryant1410:
> Re-running it succeeded.

There seems to be a race condition there then.

PR #3751, opened 2020-02-10
Remove dry-run command

bryant1410:
I'm okay with this change being rejected, but I believe there's no reason to keep `dry-run` as a separate command.


allennlp/data/batch.py
```diff
@@ -5,9 +5,9 @@
 
 import logging
 from collections import defaultdict
-from typing import Dict, List, Union, Iterator, Iterable
+from typing import Dict, Iterable, Iterator, List, Union
 
-import numpy
+import numpy as np
```
matt-gardner: Please don't do this.  If you're writing code that uses numpy, I won't complain if you import it like this.  But don't go changing all of the imports in the code to use this contraction.

allennlp/data/batch.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, instances: Iterable[Instance]) -> None:
         """
         super().__init__()
 
-        self.instances: List[Instance] = ensure_list(instances)
```
matt-gardner: When there are type annotations like this, it's generally because mypy complained about something.  Make sure that mypy doesn't complain if you remove them.

allennlp/commands/train.py
```diff
@@ -418,23 +453,24 @@ def _train_worker(
...
-            logging.info(
-                "Training interrupted by the user. Attempting to create "
-                "a model archive using the current best epoch weights."
-            )
-            archive_model(serialization_dir)
-        raise
-
-    if master:
-        train_loop.finish(metrics)
+    if not dry_run:
```
matt-gardner: If you're removing `dry-run`, I expected your recommendation to be to just use a no-op trainer, instead of adding an argument here.  I'm not sure I'm a fan of adding the argument, especially in a place like this block, where the logic gets quite nested.

In general, jack-of-all-trades functions are not good software design, because they lead to spaghetti code.  Merging `fine-tune` and `train` made sense because there was so much duplication, and when I actually removed the fine-tune command, there were no extra parameters needed anywhere (in fact you removed the ones that were added previously).  The amount of duplication with `dry-run` is much smaller (because the code in `dry-run` is much smaller), and I'm not convinced that the small reduction in duplication is worth the added complexity here.

@DeNeutoy, any opinion?

bryant1410: See the 1st commit as a reference of how `dry-run` looks refactored out. I could change it to use a no-op trainer if you want.

matt-gardner: The refactoring in that commit seems like a strict improvement over what was there before, and it would be pretty obvious to accept a PR that just did that.  It's the further combining it with `train` that I'm less inclined about.

bryant1410: Yeah, I understand. From a practical point of view, a dry run should be like a normal run but without running. If they're kept separate, then it's not gonna be an actual run (e.g., there can be changes in `train` that are by mistake not made in `dry-run`).

allennlp/data/batch.py
```diff
@@ -112,20 +110,17 @@ def as_tensor_dict(
...
         if verbose:
-            logger.info(
-                "Padding batch of size %d to lengths %s", len(self.instances), str(padding_lengths)
-            )
+            logger.info(f"Padding batch of size {len(self.instances)} to lengths {padding_lengths}")
             logger.info("Getting max lengths from instances")
         instance_padding_lengths = self.get_padding_lengths()
         if verbose:
-            logger.info("Instance max lengths: %s", str(instance_padding_lengths))
+            logger.info(f"Instance max lengths: {instance_padding_lengths}")
```
DeNeutoy: This is a bit sad that python can't handle this, but if you format logging using the %s way, it is lazily evaluated - whereas if you do it in the new `f` way, it is evaluated immediately. 

See e.g:
https://stackoverflow.com/questions/34619790/pylint-message-logging-format-interpolation

matt-gardner: Do you think it's worth following pylint's suggestion here?  My vote would be to just use f strings, unless it's something we know would be expensive.

DeNeutoy: Yeah, not sure. It does mean that we could have pretty slow code running for someone who has explicitly disabled it, but I don't know whether it matters in the grand scheme of things.

allennlp/commands/train.py
```diff
@@ -418,23 +452,24 @@ def _train_worker(
...
```
DeNeutoy: No need to make this entire block nested - just do:

```
if dry_run:
	return None
```
above it.

allennlp/commands/train.py
```diff
@@ -169,6 +180,14 @@ def train_model_from_file(
         Rank of the current node in distributed training
     include_package : ``str``, optional
         In distributed mode, extra packages mentioned will be imported in trainer workers.
+    dry_run : ``bool``, optional (default=False)
+        Do not train a model, but create a vocabulary, show dataset statistics and other training
+        information.
+
+    # Returns
+
+    best_model : ``Model``
```
DeNeutoy: Double backticks aren't needed anymore, just single ones will do

allennlp/commands/train.py
bryant1410: But if `dry_run` is `True`, I want the rest of the code to be executed (return the model if it's not distributed training):

https://github.com/allenai/allennlp/pull/3751/files/8c47ff48a1fbca1fde99940270e2ad0ff60227e4#diff-d4e8f335bf383d8d2b18287949003bbeR471-R472

allennlp/commands/train.py
```diff
@@ -1,5 +1,5 @@
 """
-The ``train`` subcommand can be used to train a model.
```
brendan-ai2: It's best not to mix in formatting cleanups with functional changes. Makes the diff harder to read.

allennlp/commands/train.py
```diff
@@ -311,6 +341,7 @@ def _train_worker(
     file_friendly_logging: bool = False,
     include_package: List[str] = None,
     batch_weight_key: str = "",
+    dry_run: bool = False,
```
brendan-ai2: In general add new args to the end of the parameter list.

bryant1410: I put it here because the rest are for distributed training.

allennlp/commands/train.py
DeNeutoy: I'm not sure there is there a need to return the model? The functions above don't do anything with it, so I think we can return `None` and remove this nesting.

bryant1410: It's transitively used here, for example:

https://github.com/allenai/allennlp/blob/57cdffb45205f1eeeff0737e9f3bb8f57103a29f/tutorials/tagger/config_allennlp.py#L91

Or you mean just if `dry_run == True`? If that's what you mean, I'm not sure of breaking the function contract in that case only, but we should state so in the docstring if we do.

allennlp/tests/commands/find_learning_rate_test.py
```diff
@@ -20,11 +21,7 @@
 
 
 def is_matplotlib_installed():
-    try:
-        import matplotlib  # noqa: F401 - Matplotlib is optional.
-    except:  # noqa: E722. Any exception means we don't have a working matplotlib.
-        return False
-    return True
+    return importlib.util.find_spec("matplotlib") is not None
```
bryant1410: I think this is neater and avoids adding exceptions to the linters.

allennlp/commands/train.py
DeNeutoy: The thing is, we probably don't ever really want to return an untrained model - so this is actually saving people time (e.g if they forget to set dry-run to false again in some script, it's probably better that it returns None than an untrained model).

allennlp/tests/commands/find_learning_rate_test.py
DeNeutoy: It would be great in the future to isolate these changes to ultra tiny PRs that we can just approve. But sure, this is fine.

bryant1410: Yeah, this started cause I had a bug after the merge, and then got fixed in another PR when I've already fixed it. But also sometimes it annoys me to open many little PRs :man_shrugging:

brendan-ai2: I'd disagree that this is actually neater. It requires the reader to go figure out `importlib.util.find_spec` does. On the other hand, a try/except is absolutely clear and precisely the behavior we want. Bypassing a linter is not a major concern if we have good reason to do so.

allennlp/commands/train.py
brendan-ai2: Fine here given that the function is internal, but in general that sort of grouping wouldn't take precedence over keeping code working.

allennlp/tests/commands/find_learning_rate_test.py
bryant1410: Should I change it back?

brendan-ai2: Yes, please.

DeNeutoy:
Thanks @bryant1410!

PR #3752, opened 2020-02-10
Remove unnecessary list creations

bryant1410:


allennlp/training/metrics/conll_coref_scores.py
```diff
@@ -227,7 +227,7 @@ def phi4(gold_clustering, predicted_clustering):
         """
         return (
             2
-            * len([mention for mention in gold_clustering if mention in predicted_clustering])
```
matt-gardner: This one I would probably keep, for readability.  Does just using `len()` without an internal list not work?

scripts/convert_openie_to_conll.py
```diff
@@ -245,7 +243,7 @@ def pad_line_to_ontonotes(line, domain) -> List[str]:
     word_ind, word = line[:2]
     pos = "XX"
     oie_tags = line[2:]
-    line_num = 0
+    line_num = "0"
```
matt-gardner: I probably wouldn't change this.  Not sure why mypy isn't complaining about it, but this doesn't seem like the right fix if there's a type complaint (unless you traced the logic and are *sure* that downstream stuff doesn't care that this is a string, then it at least deserves a comment).

allennlp/training/metrics/conll_coref_scores.py
bryant1410:
```python
len(mention for mention in gold_clustering if mention in predicted_clustering)
```

When you omit the parentheses and have only one argument that is a comprehension expression, it's a generator (with more arguments you need the parentheses). A generator doesn't define `__len__`, so it'd crash.

matt-gardner: Yeah, I guess python doesn't want to silently try to call `len()` on an infinite generator.  Fair enough.  Yeah, I'd keep this list.

bryant1410:
The failed tests were an `append` I confused with an `extend`, and another code which I also thought I reviewed and fixed (or maybe there's a duplicate line somewhere else?).

allennlp/training/metrics/conll_coref_scores.py
bryant1410: I also now removed a `float` conversion there that is unnecessary in Python 3.

scripts/convert_openie_to_conll.py
bryant1410: Why would mypy complain about it? I changed it because it's supposed to return strings (and it wasn't), and the callers convert everything to string, so it's like there's something odd here. But I changed it back.

matt-gardner: Mypy would complain because it's not returning a str, even though it says it is.  If the only caller is in this file, and it just converts everything to strings anyway, then sure, it's ok to remove it.  I was just looking at that change without context and thinking it was a scary change.

bryant1410: Oh, okay. I changed it again.

allennlp/tests/training/learning_rate_schedulers/slanted_triangular_test.py
```diff
@@ -112,7 +112,7 @@ def test_from_params_in_trainer(self):
         )
         # The method called in the logic below only checks the length of this list, not its
         # contents, so this should be safe.
-        instances = [1,] * 40  # noqa: E231, flake doesn't like what black does with this list
+        instances = [1] * 40  # noqa: E231, flake doesn't like what black does with this list
```
matt-gardner: Can you remove the comment here, also?

scripts/convert_openie_to_conll.py
```diff
@@ -34,55 +37,45 @@
 
 def main(inp_fn: str, domain: str, out_fn: str) -> None:
     """
-    inp_fn: str, required.
+    inp_fn: ``str``
```
matt-gardner: We use markdown now, so single backticks.

matt-gardner:
Thanks!

allennlp/training/metrics/conll_coref_scores.py
dirkgr: FWIW, I like to use `sum(1 for x in y)` in these cases, but keeping the list might be more straightforward.

PR #3755, opened 2020-02-10
Remove default None in dict get calls

bryant1410:
Remove the `, None)` part when calling `.get(..., None)` in a dictionary or `Params`.

It improves readability and it's the default value.

### DIFF
--- a/allennlp/common/params.py
+++ b/allennlp/common/params.py
@@ -242,9 +242,7 @@ class Params(MutableMapping):
             try:
                 value = self.params.pop(key)
             except KeyError:
-                raise ConfigurationError(
-                    'key "{}" is required at location "{}"'.format(key, self.history)
-                )
+                raise ConfigurationError(f'key "{key}" is required at location "{self.history}"')
         else:
             value = self.params.pop(key, default)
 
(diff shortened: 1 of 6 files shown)


allennlp/common/params.py
```diff
@@ -296,15 +294,8 @@ def get(self, key: str, default: Any = DEFAULT):
         Performs the functionality associated with dict.get(key) but also checks for returned
         dicts and returns a Params object in their place with an updated history.
         """
-        if default is self.DEFAULT:
-            try:
-                value = self.params.get(key)
-            except KeyError:
-                raise ConfigurationError(
-                    'key "{}" is required at location "{}"'.format(key, self.history)
-                )
```
bryant1410: This will never get called. `get` doesn't throw a `KeyError`. Am I missing something? It does apply to `pop` tho.

allennlp/models/archival.py
```diff
@@ -89,7 +89,7 @@ def extract_module(self, path: str, freeze: bool = True) -> Module:
 
 
 def archive_model(
-    serialization_dir: str, weights: str = _DEFAULT_WEIGHTS, archive_path: str = None,
+    serialization_dir: str, weights: str = _DEFAULT_WEIGHTS, archive_path: str = None
```
bryant1410: This was black :man_shrugging:

allennlp/common/params.py
matt-gardner: No, I think you're right to remove this.  It was probably there because of a copy-paste a long time ago.

PR #3759, opened 2020-02-10
Remove unused variable recover

bryant1410:


allennlp/tests/commands/train_test.py
```diff
@@ -13,10 +13,9 @@
 from allennlp.common import Params
 from allennlp.common.checks import ConfigurationError
 from allennlp.common.testing import AllenNlpTestCase
-from allennlp.data import DatasetReader, Instance, Vocabulary
```
matt-gardner: Hmm, not sure why flake wasn't complaining about this...

bryant1410: https://github.com/allenai/allennlp/blob/ddb886b889232b9c28ee9389039067d6af6fa5c6/.flake8#L19-L24

matt-gardner: Oh, right, thanks.  Maybe that's so that we can get things in the registry?  Though I think all of our registrable classes are supposed to be imported in our `__init__` files...  Not totally sure why we need that rule.

PR #3760, opened 2020-02-10
Add missing plugin import in distributed training

bryant1410:


matt-gardner:
@bryant1410, is this ready for review?  Can you remind me `import_plugins` is necessary on top of what's already there?

bryant1410:
> @bryant1410, is this ready for review? Can you remind me why `import_plugins` is necessary on top of what's already there?

Yes, it's ready.

Loading a plugin there is necessary depending on the plugin. If it's something that each process should have then yes. For example, if someone uses a plugin instead of `--include-package` to use some model, then yes.

matt-gardner:
Ok, got it, I think: the `include_package` stuff that was there was written specifically for `--include-package` arguments, and only handled things that were passed as an argument that way.  `import_plugins()` handles plugins specified in other ways, and it was missing.

The fact that we call the `include_package` function `import_submodules`, and why we don't recurse for plugins, is a little confusing.

bryant1410:
> Ok, got it, I think: the `include_package` stuff that was there was written specifically for `--include-package` arguments, and only handled things that were passed as an argument that way. `import_plugins()` handles plugins specified in other ways, and it was missing.
> 
> The fact that we call the `include_package` function `import_submodules`, and why we don't recurse for plugins, is a little confusing.

Yeah, I agree. Also, remember they are separately called because the plugins need to be imported before the arguments are parsed (because there may be new subcommands), while the `--include-package` ones need to be after (because they need to be parsed with the arguments).

bryant1410:
Maybe we can rename `import_submodules` to `import_submodules_recursively`?

matt-gardner:
I don't think we need to do that particular renaming.  The reason it was confusing me was because it says _sub_ modules, not just `import_modules`, which probably wouldn't have me wondering what was going on.  If we do a renaming, I'd go for something like `import_modules`, `import_module_and_submodules`, or something that actually says you're going to import the top-level module.  The naming as it is made me wonder if the `import_plugins` and the `import_submodules` were interacting in some way, when they are just related but totally separate pieces of logic.

matt-gardner:
A better comment right there might also just be enough.

PR #3762, opened 2020-02-11
Remove unnecessary flake8 exceptions

bryant1410:
Also tackles https://github.com/allenai/allennlp/pull/3760#issuecomment-584709526 and https://github.com/allenai/allennlp/pull/3759#discussion_r377721012

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -13,40 +13,14 @@ exclude =
 
 per-file-ignores =
     # __init__.py files are allowed to have unused imports and lines-too-long
-    allennlp/__init__.py:F401
-    allennlp/**/__init__.py:F401,E501
+    **/__init__.py:F401
+    # for some reason ("fixtures" dir not being a Python package?) we need to add this line
+    allennlp/tests/fixtures/**/__init__.py:F401
 
     # tests don't have to respect
-    #  E501: line length
-    #  E261: 2 spaces before comments (it would be ideal if they did, though)
     #  E731: do not assign a lambda expression, use a def
-    #  F401: unused imports
-    allennlp/tests/**:E501,E261,E731,F401
+    allennlp/tests/**:E731
 
     # scripts don't have to respect
-    #  E501: line length
     #  E402: imports not at top of file (because we mess with sys.path)
-    #  E302 expected 2 blank lines, found 1 (it would be ideal if they did, though)
-    scripts/**:E501,E402,E302
-
-    # here are a few files that get to ignore line-too-long, lucky them
-    allennlp/pretrained.py:E501
-
-    # here are some files that get to ignore E402 module level import not at top of file
-    allennlp/modules/elmo.py:E402
-    allennlp/modules/elmo_lstm.py:E402
-    allennlp/modules/token_embedders/embedding.py:E402
-    allennlp/commands/elmo.py:E402
-    allennlp/tests/modules/elmo_test.py:E402
-    allennlp/tests/commands/elmo_test.py:E402
-    allennlp/tests/modules/token_embedders/embedding_test.py:E402
-
-    # Should probably fix this, but there's a lot of them.
-    # E127 continuation line over-indented for visual indent
-    allennlp/tests/semparse/worlds/atis_world_test.py:E127
-
-    # This one test is much more readable with nonstandard comments
-    # E114 indentation is not a multiple of four (comment)
-    # E116 unexpected indentation (comment)
-    # E128 continuation line under-indented for visual indent
-    allennlp/tests/data/token_indexers/bert_indexer_test.py:E114,E116,E128
+    scripts/**:E402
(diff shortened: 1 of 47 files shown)


matt-gardner:
@DeNeutoy, just using `AllenNlpTestCase` should do the necessary imports, right?  I'm not sure why we need the flake exception.  There are only 6 failing tests here, and I think they are failing because of missing imports in the corresponding `__init__.py` files, so this is actually catching bugs.  Also, I think we should avoid using `from_params` to construct things in tests going forward, except for unusual cases, or integration-style tests like our model tests.

On reshuffling of imports: when I touch a file, I will often sort the imports in that file - makes it easier to find things later.  I didn't look closely at all of the shuffling in this PR, but it looks like that's all that's going on here.

DeNeutoy:
Oh ok, happy to take your lead on both issues

bryant1410:
> @DeNeutoy, just using `AllenNlpTestCase` should do the necessary imports, right? I'm not sure why we need the flake exception. There are only 6 failing tests here, and I think they are failing because of missing imports in the corresponding `__init__.py` files, so this is actually catching bugs. Also, I think we should avoid using `from_params` to construct things in tests going forward, except for unusual cases, or integration-style tests like our model tests.

I'm trying to add the missing imports.

The `from_params` is used in a script. In any case, IIUC, @DeNeutoy you say you're gonna fix this?

> On reshuffling of imports: when I touch a file, I will often sort the imports in that file - makes it easier to find things later. I didn't look closely at all of the shuffling in this PR, but it looks like that's all that's going on here.

I think we can add a linter to optimize the imports, like in the transformers library: https://github.com/huggingface/transformers/blob/ee5de0ba449d638da704e1c03ffcc20a930f5589/.circleci/config.yml#L105

.flake8
```diff
@@ -12,41 +12,14 @@ exclude =
     tutorials/tagger/**
 
 per-file-ignores =
-    # __init__.py files are allowed to have unused imports and lines-too-long
-    allennlp/__init__.py:F401
-    allennlp/**/__init__.py:F401,E501
+    # __init__.py files are allowed to have unused imports
+    */__init__.py:F401
+    */**/**/__init__.py:F401
```
bryant1410: Apparently this works. If you can come up with a way to grab all with one line, that'd be cool.

allennlp/commands/elmo.py
```diff
@@ -68,25 +68,24 @@
...
-from allennlp.common.checks import ConfigurationError
 from allennlp.data.token_indexers.elmo_indexer import ELMoTokenCharactersIndexer
-from allennlp.nn.util import remove_sentence_boundaries
 from allennlp.modules.elmo import _ElmoBiLm, batch_to_ids
-from allennlp.commands.subcommand import Subcommand
+from allennlp.nn.util import remove_sentence_boundaries
+
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", category=FutureWarning)
+    import h5py
```
bryant1410: I put it at the end so the regular imports are sorted.

allennlp/common/util.py
```diff
@@ -434,7 +434,7 @@ def push_python_path(path: PathType) -> ContextManagerFunctionReturnType[None]:
         sys.path.remove(path)
 
 
-def import_submodules(package_name: str) -> None:
+def import_module_and_submodules(package_name: str) -> None:
```
bryant1410: Note I make this change mentioned in another PR here.

scripts/convert_openie_to_conll.py
```diff
@@ -217,7 +217,7 @@ def read(fn: str) -> Iterable[List[Extraction]]:
         yield prev_sent
 
 
-def convert_sent_to_conll(sent_ls: List[Extraction]) -> Iterable[Tuple[str, str, ...]]:
+def convert_sent_to_conll(sent_ls: List[Extraction]) -> Iterable[Tuple[str, ...]]:
```
bryant1410: I changed this because in the end you can only put ellipsis in the second arg.

bryant1410:
Tests pass now, but I'm not quite sure where what was missing gets imported exactly.

Also, I believe that tests that load archives with config files can locally import what's necessary inside the test function (or `setUp`).

.flake8
matt-gardner: Please leave the `lines-too-long`; black will catch anything that's fixable, so we don't need flake to complain at us for these files.  Even if there aren't any files where it matters now, it'd be annoying to have this come up later.

bryant1410:
About messing with `sys.path` within the scripts, is it really necessary? If they're called from the root dir they should work well.

matt-gardner:
It may not be necessary, but I and others have run into it enough times that it's convenient to have, and it doesn't cause problems.

PR #3763, opened 2020-02-11
Embedding api

### LINKED ISSUE #2898: Documentation/Method unclear for Pretrained File in Embedding
TLDR; The code for programmatically using pretrained embedding file in the Embedding object is hard to understand and use. 

The [official tutorial](https://allennlp.org/tutorials) uses AllenNLP programmatically, and includes embeddings, but not from pretrained. I tried implementing my own model programmatically, and used Embedding(pretrained_file="/path/to/glove", ...) but found that this does nothing. 

AFAIK, the way to do it is to use a private method ("_read_pretrained_embeddings_file") and pass the resultant weights to the Embedding object (argument "weight"). 

This is counterintuitive. I expected that the "pretrained_file" argument would work, because that's how the config files work.

DeNeutoy:
Fixes #2898


allennlp/modules/token_embedders/embedding.py
```diff
@@ -74,9 +74,12 @@ class Embedding(TokenEmbedder, Registrable):
         construct it in the original training. We store vocab_namespace used during the original
         training as an attribute, so that it can be retrieved during fine-tuning.
     pretrained_file : `str`, (optional, default=None)
-        Used to keep track of what is the source of the weights and loading more embeddings at test time.
-        **It does not load the weights from this pretrained_file.** For that purpose, use
-        `Embedding.from_params`.
+        Path to a file of word vectors to intialize the embedding matrix. It can be the
+        path to a local file or an URL of a (cached) remote file. Two formats are supported:
```
matt-gardner: a URL instead of an URL?  Or is that a British / American difference?

allennlp/modules/token_embedders/embedding.py
```diff
@@ -99,6 +102,7 @@ def __init__(
         sparse: bool = False,
         vocab_namespace: str = None,
         pretrained_file: str = None,
+        vocabulary: Vocabulary = None,
```
matt-gardner: If you're taking this parameter in the constructor, I think you can just remove `from_vocab_or_files` entirely, and remove the `constructor=` parameter from the registry, and the duplicate registration at the bottom of the class.  I'm in favor of this, it simplifies things quite a bit.

DeNeutoy: Yes, I thought a bit about this. However:

1. programmatic usage should still support `Embedding(num_embeddings, embedding_dim)`. If I make num_embeddings optional (as it can now be inferred if a vocab is also passed), I will have to break Embedding in a really insidious way, as it is the first argument. 

2. There is some interplay with `vocab_namespace` being set to `None` if `num_embeddings` is not passed and how the vocabulary is extended which I don't really understand.

allennlp/modules/token_embedders/embedding.py
DeNeutoy: oh sure, I just copied this comment from below. I'll update it.

allennlp/modules/token_embedders/embedding.py
matt-gardner: 1. You'll only be breaking it for people who called it directly, which is likely a minority of users.  We can have a big warning about this in our release notes.  It does seem like a pretty big reduction in complexity whose only cost is a breaking change.

2. I think the main complexity is that the default in `from_vocab_or_files` is `"tokens"`, while it's `None` here.  That code block was trying to override the default value.  It'd be pretty easy to just do `num_embeddings = num_embeddings or vocab.get_vocab_size(vocab_namespace or "tokens")` in the constructor (maybe not so one-line, but that's the idea), and there's no issue.

You can pick whether you think it's worth the breaking change or not.  I think it probably is, but I'm ok if you think otherwise.

bratao: Please, only one request. Do not forget the people who use the lib directly, not using the config. It is a great use case for production apps who need to process in-memory data. The Flair library do an awesome work on this area. But I really love the AllenNLP maturity. ❤️

matt-gardner: A major effort in our upcoming 1.0 release is to make everything easier to use for people who don't want to use our config files. So, no, we haven't forgotten you :). In fact, this recommended breakage is specifically to make it easier in the long run to use this module without config files, it just requires a one-time breaking change to make it better.

DeNeutoy:
@matt-gardner do you have a grip on why these tests are failing so spectacularly? I can dig into it but i'm not really sure where to start ...

DeNeutoy:
Ok I think the problem is that if `Embedding` needs to be constructed with a `Vocab`, it creates a chain of dependencies of having `Lazy[Vocab]` from_params constructors for everything above it in the hierarchy (or maybe just that they all have to pass it as extra `**kwargs`).  This effects like half the codebase, and it's also a breaking change, so i'm just going to revert the above commit and keep the small, custom from_params. If we figure out how to remove it later, then great.

matt-gardner:
I can look at this, but not until Wednesday.  ACL rebuttals today, meetings all day tomorrow.  Seems fine to just merge without that and revisit the fix later.

PR #3765, opened 2020-02-11
test out github actions for automated nightly releases

DeNeutoy:
This actually turned out to be not so hard, but it requires a couple of changes:

1. I need to change `allennlp/version.py` to look for an environment variable when setting the version, so I can change it to `MAJOR.MINOR.GITSHA` for the nightly versions

2. It requires adding our release details for pypi to github secrets

I didn't use TC for this because I thought it might be helpful to be able to generalise this across all of the allennlp* packages eventually, and it's easier for outside maintainers to see what is happening. 

I'll need to merge this before I actually know whether it works or not, but equally the nightly releases are a good place to test this functionality before we try using it for our proper releases. LMK what you think.

For our proper releases, we can configure these to run when new github tags matching a regex are pushed, which would be really nice - we just update the version in a PR, check the tests in TC run, and then push a tag once we've verified that. Then the whole packaging, testing and release step happens automatically.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -2,7 +2,7 @@ name: Nightly Release
 
 on:
   schedule:
-    - cron:  '0 9 * * *'
+    - cron:  '0 11 * * *'
 
 jobs:
   build:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,31 @@
+name: Nightly Release
+
+on:
+  schedule:
+    - cron:  '0 9 * * *'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Install requirements
+      run: |
+        pip install -r dev-requirements.txt
+    - name: Build Package
+      env: 
+        ALLENNLP_VERSION: $(git rev-parse --short "$GITHUB_SHA")
+      run: |
+        python setup.py bdist_wheel sdist
+    - name: Test Packages
+      run: |
+        pip install $(ls dist/*.whl)
+        allennlp test-install
+    - name: Deploy
+      run: |
+        twine upload -u ${{ secrets.PYPI_NIGHTLY_USERNAME }} -p ${{ secrets.PYPI_NIGHTLY_PASSWORD }} dist/*
+        rm -rf dist/
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -18,7 +18,7 @@ jobs:
         pip install -r dev-requirements.txt
     - name: Build Package
       env:
-        ALLENNLP_REVISION: $(git rev-parse --short "$GITHUB_SHA")
+        ALLENNLP_REVISION: dev.$(git rev-parse --short "$GITHUB_SHA")
       run: |
         python setup.py bdist_wheel sdist
     - name: Test Packages
@@ -26,6 +26,7 @@ jobs:
         pip install $(ls dist/*.whl)
         allennlp test-install
     - name: Check diff from 24hr ago
+      if: success()
       run: scripts/24hr_diff.sh
     - name: Deploy
       if: success()
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -17,15 +17,18 @@ jobs:
       run: |
         pip install -r dev-requirements.txt
     - name: Build Package
-      env: 
-        ALLENNLP_VERSION: $(git rev-parse --short "$GITHUB_SHA")
+      env:
+        ALLENNLP_REVISION: $(git rev-parse --short "$GITHUB_SHA")
       run: |
         python setup.py bdist_wheel sdist
     - name: Test Packages
       run: |
         pip install $(ls dist/*.whl)
         allennlp test-install
+    - name: Check diff from 24hr ago
+      run: scripts/24hr_diff.sh
     - name: Deploy
+      if: success()
       run: |
         twine upload -u ${{ secrets.PYPI_NIGHTLY_USERNAME }} -p ${{ secrets.PYPI_NIGHTLY_PASSWORD }} dist/*
         rm -rf dist/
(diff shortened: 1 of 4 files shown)


brendan-ai2:
I could be missing something, but don't we need the `allennlp/version.py` changes in this PR?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,31 @@
+name: Nightly Release
+
+on:
+  schedule:
+    - cron:  '0 9 * * *'
```
bryant1410: Does it make sense to release in the case there were no releases?

bryant1410:
> I could be missing something, but don't we need the allennlp/version.py changes in this PR?

To upload different versions, otherwise the upload fails.

.github/workflows/nightly.yml
bryant1410: If there weren't changes in a day, seems to me like `twine upload` is gonna fail along with the whole workflow.

DeNeutoy: Oh good point, thanks - i'll figure that out :)

DeNeutoy:
Yes, I hadn't finished the changes - this was just a "do we hate this approach" kind of review request 👍 . I will finish this off, given people seem positive

brendan-ai2:
@bryant1410, I understand why we need `version.py`. I was asking why the changed file was not in this PR. :)

Thanks for explaining, Mark!

bryant1410:
> @bryant1410, I understand why we need `version.py`. I was asking why the changed file was not in this PR. :)

Oh, I was on the bus with the phone, and I think I misread "don't" with "do" :sweat_smile:

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,31 @@
...
+    - name: Setup Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Install requirements
+      run: |
+        pip install -r dev-requirements.txt
+    - name: Build Package
+      env: 
+        ALLENNLP_VERSION: $(git rev-parse --short "$GITHUB_SHA")
```
bryant1410: I think `git describe` is better for this.

brendan-ai2: I'd argue that the SHA is precisely how we should identify git commits. Let's not introduce another identifier with different semantics.

bryant1410: Yeah, but, for example, which version is the latest, `ac19db2` or 5fc3719`?

bryant1410: What I mean is that PyPI/pip has to determine an order for the versions, and it gets hard with just the SHA.

bryant1410: `git describe` gives you the latest tag from the parents, the number of commits there are from it and the SHA; in that order.

brendan-ai2: Ah, sure. I would expect that a nightly release would be named (in part) off of the date it was made.

brendan-ai2: Here's what I get when I install the torch nightlies, for instance:

```
$ pip list | grep torch
torch       1.5.0.dev20200211
torchvision 0.6.0a0+d8d36f0
```

(The most recent tag for `torchvision` is v0.5.0, incidentally. So this is forward looking in contrast to `git describe`.)

brendan-ai2:
```
(tmp) brendanr.local ➜  allennlp git:(master) ✗ git describe
v0.9.0-215-g473db763
(tmp) brendanr.local ➜  allennlp git:(master) ✗ git show g473db763
fatal: ambiguous argument 'g473db763': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

That's not great behavior...

bryant1410: I think the leading `g` is extra. Try with the rest.

bryant1410: (I think it goes for "git")

brendan-ai2: Good catch, thank you. That works.

brendan-ai2: But I still see no reason to use this. Date is standard for naming nightlies. SHA is standard for identifying git commits. No need for a new thing.

bryant1410:
```bash
echo $(git describe --abbrev=0).dev$(date +%Y%m%d)
```

bryant1410: `date -u +%Y%m%d` to use UTC 0.

allennlp/version.py
```diff
@@ -1,6 +1,11 @@
+import os
+
 _MAJOR = "0"
 _MINOR = "9"
-_REVISION = "1-unreleased"
+
+# We check the environment for the revision here, because it
+# enables us to set it to a git sha when doing automated nightly releases.
+_REVISION = os.environ("ALLENNLP_REVISON", None) or "1-unreleased"
```
bryant1410: typo: REVISION

.github/workflows/nightly.yml
brendan-ai2: `echo $(date +%Y%m%d)-$(git rev-parse --short master)`

bryant1410: But this version is gonna be greater than the stable releases. Also, for SemVer, you need to specify somewhere that it's "dev" or something like that, otherwise systems will think it's stable.

allennlp/version.py
```diff
@@ -1,6 +1,11 @@
+import os
+
 _MAJOR = "0"
 _MINOR = "9"
-_REVISION = "1-unreleased"
+
+# We check the environment for the revision here, because it
+# enables us to set it to a git sha when doing automated nightly releases.
+_REVISION = os.environ.get("ALLENNLP_REVISION", None) or "1-unreleased"
```
bryant1410:
```suggestion
_REVISION = os.environ.get("ALLENNLP_REVISION", "1-unreleased")
```

.github/workflows/nightly.yml
brendan-ai2: This is just the revision. 0.9.0 gets added inside `version.py`. Thanks for the heads up about "dev".

bryant1410:
I'm still concerned that the version would be like `0.9.0.dev.200948d`, but PyPI/pip won't be able to determine which one is the latest.

bryant1410:
And btw, related, I found this website useful: https://cron.help/#0_11_*_*_*

PR #3768, opened 2020-02-12
Add date to the nightly build.

brendan-ai2:
- Adds a test to ensure we're compatible with semantic versioning.


brendan-ai2:
@bryant1410, could you please let me know if I missed anything? Thanks for the pointer to semantic versioning!

setup.py
```diff
@@ -124,7 +124,7 @@
     ],
     entry_points={"console_scripts": ["allennlp=allennlp.run:run"]},
     setup_requires=setup_requirements,
-    tests_require=["pytest", "flaky", "responses>=0.7"],
+    tests_require=["pytest", "flaky", "responses>=0.7", "semantic_version"],
```
DeNeutoy: Think you either need to add this to dev-requirements, or above. I'm not actually sure why we maintain tests_require given that one of our allennlp commands run the tests...

brendan-ai2: Thanks for catching this, fixed. Yeah, I don't know enough about `setuptools` to say. Added a TODO for when someone has the time to grok that.

bryant1410: Yeah, actually `tests_require` is only necessary when using `python setup.py test`, which is actually deprecated. So this entry can be removed.

brendan-ai2: Thanks I modified the comment to mention that, but left the line in case someone is still depending on it.

bryant1410: I don't think something depends on it because I have never seen a reference to `python setup.py test` in this codebase. But I think it's okay if you leave it, yeah.

brendan-ai2: Because AllenNLP is a library determining whether someone else has a dependency on some feature is not tractable. For this reason I try to avoid unnecessary breaking changes.

bryant1410: I see, makes sense. However, `python setup.py test` doesn't apply to people that do `pip install` because there are no tests in the package unless they do `--editable`, or clone/copy the codebase and explicitly run the tests this way, which has been deprecated by setuptools a while ago.

bryant1410: Also, we're preparing for 1.0.0, so I'd take advantage and break compatibility now rather than after its release.

brendan-ai2: I'm not sure what deleting that line gains. But it could well break a script somebody out there wrote. I'm fine with breaking backward compatibility when there is a gain, but I don't see that here. Just because we're preparing for 1.0.0 doesn't mean we should break backward compatibility without a clear benefit.

bryant1410: Just that we have a duplicate definition of test reqs, and this one is actually inaccurate, cause also the rest of the things in `dev-requirements.txt` should be necessary as well.

setup.py
```diff
@@ -121,10 +121,12 @@
         "conllu==2.2.2",
         "transformers>=2.4.0",
         "jsonpickle",
+        "semantic_version",
```
bryant1410: Wait, I just realized that `semantic_version` shouldn't be in the regular dependencies, and also that it's missing from `dev-requirements.txt` :grimacing:

PR #3769, opened 2020-02-12
Enable other features in Token

### LINKED ISSUE #3591: Make `Token` class extensible.
I'd like to add additional task-specific tags(similar to pos/ner tags) which are not included in the attributes of the current `Token` class  to enrich the embeddings. Thus, I have to extend the `Token` class in AllenNLP.
However, it seems impossible to creating `MyToken` class that subclass the `Token` class (inheriance seems a must to not breaking type checking in other classes) according to [this link on StackOverflow](https://stackoverflow.com/questions/44287623/a-way-to-subclass-namedtuple-for-purposes-of-typechecking`)

Replace the `NamedTuple` with the new `dataclass`, which introduced in Python 3.7 based on [PEP557](https://www.python.org/dev/peps/pep-0557/). 
You can also refer to the discussion in the above StackOverflow link.

bratao:
Hello, 

I´m working on a dataset that have layout features. For example, a bold word have a different meaning.

One problem of AllenNLP is that I cannot easily extend to index other features, as the Token class is a NamedTuple with a limited set of options. To support my use case I would have to extend the Token, the SingleIdTokenIndexer and the TextField ( as it require a Token or SpacyToken).

With this change I could easily add new features to a token, and create embeddings that use them.

Fixes #3591.


dirkgr:
Have you thought about using `type_id` for these features? You'll get a `type_id` tensor in your model that you can pass directly to an embedder.

bratao:
@dirkgr I tried, but to use the type_id would be very painful. I should have to manage the ID, use a personalized indexer...

With this change, one can easily use multiple text features such as:
```
token_list = []
for token in tokens:
    token_list.append(Token(tokens['text'], other_features={
        "feature1": tokens['feature1'],
        "feature2": tokens['featurew']}))

token_indexer = {"tokens": SingleIdTokenIndexer(),
                 "feature1": SingleIdTokenIndexer(namespace="feature1",
                                                  feature_name="feature1"),
                 "feature2": SingleIdTokenIndexer(namespace="feature2",
                                                  feature_name="feature2")}
sentence_field = TextField(token_list, token_indexer)

```

matt-gardner:
I'd prefer a different solution to this problem: just make our `Token` a `dataclass` (see #3591), so you can extend it and have everything still work.

bratao:
@matt-gardner I can try to use a dataclass.
It is ok to support only >=3.7? Or should I import a dataclass in 3.6?

matt-gardner:
I think this is the only thing that would require 3.7; if it's easy to keep support for 3.6 with an import, I would do that.

bratao:
@matt-gardner could you please check it?

I feel that the API is more elegant with a dataclass. One minor problem is that Token is mutable now.

allennlp/data/token_indexers/single_id_token_indexer.py
```diff
@@ -34,6 +34,7 @@ def __init__(
         lowercase_tokens: bool = False,
         start_tokens: List[str] = None,
         end_tokens: List[str] = None,
+        feature_name: str = "text",
```
matt-gardner: Hmm, I think this would actually let us entirely remove the `Ner` and `PosTag` indexers.  Probably should have done this a long time ago...

Can you add a note to the docstring about this new parameter?

bratao:
@matt-gardner Done. Yeah, for sure now we can replace those other taggers. I will do in another PR

matt-gardner:
TC isn't reporting for some reason, but it's passing, so I'm merging this.

matt-gardner:
Thanks @bratao!

Issue #3017, opened 2019-06-28
Why tensors are moved to CPU when calculating metrics?


shtratos:
**System:**
 - OS: Ubuntu 18.04.2
 - Python version: 3.6.7
 - AllenNLP version: v0.8.4
 - PyTorch version: 1.1.0

**Question**

When training a simple language model on GPU I get a big computational performance hit when calculating accuracy metrics.
I've tracked it down to this method:
https://github.com/allenai/allennlp/blob/70fa4aac3762c2de48e1123938ead2c50c0bb99c/allennlp/training/metrics/metric.py#L42-L49

I understand, that `detach()` is required to avoid connection to computational graph and thus memory leaks. But why tensors are transferred to CPU?

When I remove `.cpu()`, metrics are calculated much faster and I don't see any significant increase in GPU memory consumption.

matt-gardner:
I think some of our metrics (including some of our earliest ones, when this functionality was originally written) did some loops and stuff that required being on the CPU, and so this got added to the generic method without realizing that it wasn't necessary everywhere.  You're right that it should be removed, but we need to be careful to add it back in whatever place it's actually necessary.

This method also made more sense back when pytorch had `Variables`, and now it's largely unnecessary, or at least misnamed - `detach_tensors` would be a better name, or a simple call to `.detach()` inline where this is called.

PR welcome!

shtratos:
If I make the change, will tests in CI build pickup any problems where tensors must be on CPU? Or is more complex analysis required.

matt-gardner:
Our CI only runs on CPU, so this probably won't find all potential problems, unfortunately.  I'm not sure what the best way to find this is.

DeNeutoy:
This definitely won't work for all of the metrics, because some of them rely on converting tensors in to string labels, which requires the tensors to be on the CPU, e,g:
https://github.com/allenai/allennlp/blob/master/allennlp/training/metrics/span_based_f1_measure.py

The easiest, least overhead way to do this is to make converting the tensors to CPU be optional and have it true by default, with a way to make a metric know that it can run on the GPU.

matt-gardner:
And when we talk about default options for such a simple function, it definitely makes me think the function should just go.

brendan-ai2:
@matt-gardner, @shtratos we actually do have some limited GPU testing (albeit not on PRs due to having a limited number of GPUs on our TC box). It runs after PRs are merged to master with the intention being to catch issues before we release.

[...]

PR #3770, opened 2020-02-12
Add isort to sort the imports automatically

bryant1410:
This patch adds [`isort`](https://timothycrosley.github.io/isort/) to sort the imports, so they're more readable and there's consistency across the codebase. Now the linter pipeline is black+isort+flake8+mypy.

isort is generally used in Python for this (e.g., [transformers](https://github.com/huggingface/transformers) uses black+isort+flake8 as the linters). They're apparently close to release a new major version (v5), and what they have stable now doesn't have some changes that are convenient here, that's why I use the latest in the "develop" branch (I put some comments about it in dev-requirements.txt).

This is something boring to review, but it's mostly about imports, so if you see imports changing then it's fine. I reviewed it myself before sending it (I always do that with PRs).

Some comments (including non-trivial changes):

* `isort` works like black: `isort .` re-formats the files, following `pyproject.toml` (also supports custom files, but I think this is the best for this codebase), `isort . --interactive` is interactive. `isort . --check --diff` just checks and shows the diff, which is what I put in scripts/verify.py.
* `isort` plays nice with `black`. It has a `profile` option which I set to `black`, so it avoids some conflicts. With the "develop" version, I didn't have to add any `# isort:skip` comment to avoid anything, which is cool. `black`, `isort`, `flake8`, and `mypy` pass successfully. I put the same line length as in `black` (which IMO it could be increased and would play well with flake8).
* I updated the `pre-commit` configuration. Remember `pre-commit` it's a tool to run fast lints over the modified files when you do `git commit`, and can be easily configured by just doing `pre-commit install`. So now it runs isort+black+flake8, in that order. I had to change its configuration a bit. Now it uses local tools instead of remote ones (see the file .pre-commit-config.yaml). The main reason of this change is that if it uses remote repos to obtain the tools then it uses its own virtualenv, and then `isort` isn't able to tell which imports are from 3rd parties (e.g., `import torch`) and which ones are from the repo itself (i.e., `import allennlp`), which are organized differently. Also, [pre-commit recommends using the repo local hooks if the scripts are tightly coupled to the codebase and also for the virtualenv thing](https://pre-commit.com/#repository-local-hooks). For some context, `pre-commit`, which is written in Python, is also used in non-Python projects, that's why it uses it's own venv in general.
* I fixed some imports in scripts that were placed after the `sys.path` thing, which I screwed up in another PR.
* I removed a few pylint references that I found.
* I added a `scripts/isort.sh`, but I have no idea where those are used :man_shrugging:
* I changed a paragraph in STYLE.md to show the new pipeline.
* I changed some comments and docstring that had line lengths >100.
* Some script I run changed CRLF to LF in 2 files about spearman_correlation. In those files, there're also just import changes, and that's it.
* dev-requirements.txt, .pre-commit-config.yaml pyproject.toml was changed.
* I changed the usage of `regex` for `re` in `scripts/convert_openie_to_conll.py` (which was being an undeclared dependency, and some unnecessary extra backslashes that my IDE identified.
* I changed an f-string in `scripts/benchmark_iter.py`.
* I changed a deprecated function in `scripts/convert_openie_to_conll.py`.
* I sorted deps in setup.py alphabetically and added "botocore", which I saw was imported somewhere (meaning it's a high-level dependency).
* I made an accidental pass in the tutorials folder, but the changes make sense to me.

Btw, you should change something in TC, right? Can't TC configuration be versioned?


.pre-commit-config.yaml
```diff
@@ -1,10 +1,18 @@
...
-    rev: v1.2.3
-    hooks:
-    - id: flake8
+- repo: local
+  hooks:
+  - id: isort
+    name: isort
+    entry: isort
+    language: python
+    types: [python]
```
bryant1410: Not sure if `name` is required in these, actually, also `language` for a local repo. `entry` is the command to be run. `types` is the types of changed files it'll pass to the program as argument.

allennlp/__init__.py
```diff
@@ -2,27 +2,26 @@
...
 
-if sys.version_info < (3, 6, 1):
-    raise RuntimeError("AllenNLP requires Python 3.6.1 or later")
-
 # We get a lot of these spurious warnings,
 # see https://github.com/ContinuumIO/anaconda-issues/issues/6678
-import warnings  # noqa
+import warnings
+
+from allennlp.version import VERSION as __version__  # noqa
```
bryant1410: Does this one necessarily need to be at the end of the file? It doesn't seem so to me, but not sure.

DeNeutoy:
Hi Santiago, 

Thanks for the detailed description of the PR - we aren't going to accept this change, for the following reasons:

* We want our checks to focus on correctness, and style in the case that it is likely to correlate to correctness (for example, this is why we switched from pylint to flake8.).
* Import order and alphabetical ordering do not functionally address the correctness requirement that we want to check.
* We're worried that allennlp already has extremely high standards for contributions, and that we don't want to add additional checks such as this which will be difficult to understand in CI for new contributors.

There are a couple of parts of this PR that we'd love to accept if you want to strip them out:

```
I removed a few pylint references that I found.
I changed a paragraph in STYLE.md to show the new pipeline.
```

Broadly we love your contributions, and your attention to detail has got us wondering if you'd be interested in taking a look at any of these harder issues (which have more concrete benefits for allennlp/which you might find interesting to work on):

* #2962 - several people have commented that our indexing/tensorisation code is somewhat slow. What parts are slow specifically? Are there any easy wins to speed them up?

* #3017 - Quite a lot of our metrics calculation happens on CPU unnecessarily. Converting the relevant metrics to run on the GPU would be very helpful.

Finally, as we push toward 1.0, we have a bunch of stuff in our milestone which we need to kill off. If you wanted to pick out any of these issues, that would be fantastic, and very helpful. 

https://github.com/allenai/allennlp/milestone/10

bryant1410:
Yeah, Matt also pointed me out to the issues in that milestone, that's why I helped with the plugins thing to make the AllenNLP server pluggable and installable. Then I came across things that have bugged me at different moments in the past and send some PRs, hoping to share a bit of Python experience. For context, when I use 3rd party libraries, I like to learn how to use them by looking at their code (PyCharm makes it really easy), then fix/change them. So then I got familiarized with this codebase over time, have changed stuff on my side and have sent patches when I had time (kinda "if you don't like something, fix it"). And I like being active on GitHub :man_shrugging: 

I understand and it makes sense to me that some of these changes can give you little functional value (but maybe some maintenance value, albeit not discussed), and you prefer that I would concentrate efforts on things that give more value to you, moreover with the upcoming 1.0 release. So I'll try to do that.

My available time comes and goes. But I'm gonna look into those when I have time. The first one seems complex. BTW, I appreciate that you guys put labels with effort estimates on the issues (e.g., "day"), that's helpful.

DeNeutoy:
@bryant1410 Thanks for the feedback, we'll definitely keep using the day/week/month labels if they are helpful to you. 

FWIW we do occasionally welcome big ideas like this - it just saves both of us time if you open an issue with the idea first - hopefully you've found us reasonably responsive in this way.

Just to re-iterate, we absolutely love your contributions to allennlp, and you'll be getting a big shout out when we finally get to 1.0 - you've been overwhelmingly the most productive OS contributor that we've had in the history of the project. So that's great - just sometimes we want to focus the energy canon of Santiago Castro in a particular direction 😉

PR #3772, opened 2020-02-13
Rewrite the augmented LSTM to use the improvments from PyText

### LINKED ISSUE #2482: Fuse operations in AugmentedLSTM
Pytext reimplemented our AugmentedLSTM but made it faster by doing the gate equations as a single add op. We should check if the speedup is real and if so switch to using it.

https://github.com/facebookresearch/pytext/pull/266

bratao:
The PyText project rewrote the augmented LSTM and made it faster by doing the gate equations as a single add op.

It is faster than the old one, and also include a drop-in replacement for the torch.LSTM. I had some work to keep the behavior the same as the old one, including the tests to pass.

 Fixes #2482


bratao:
Thank you @DeNeutoy . I will refactor to be compatible with the older code. I will also check all tests. I will remove the WIP and tag you when I´m done

DeNeutoy:
@bratao great! Also I forgot one test, sorry - you can run it with `bash scripts/mypy.sh`. It is a python type checker.

bratao:
@DeNeutoy could you please review it for me?

allennlp/modules/augmented_lstm.py
```diff
@@ -1,147 +1,228 @@
...
+        a single layer unidirectional AugmentedLSTM representation of the sequential
+        input and new state tensors.
+
+        Args:
+            inputs (PackedSequence): Input tensor of shape
+                (bsize x seq_len x input_dim).
+            states (Tuple[torch.Tensor, torch.Tensor]): Tuple of tensors containing
+                the initial hidden state and the cell state of each element in
+                the batch. Each of these tensors have a dimension of
+                (1 x bsize x num_directions * nhid). Defaults to `None`.
```
dirkgr: There is only one direction in `AugmentedLstm`, isn't there? It looks like `num_directions` belongs in the class below?

allennlp/modules/augmented_lstm.py
```diff
@@ -1,147 +1,228 @@
...
+        """
+        Warning: Would be better to use the BiAugmentedLstm class in a regular model
+
+        Given an input batch of sequential data such as word embeddings, produces
+        a single layer unidirectional AugmentedLSTM representation of the sequential
+        input and new state tensors.
+
+        Args:
+            inputs (PackedSequence): Input tensor of shape
+                (bsize x seq_len x input_dim).
```
dirkgr: It's not really a `Tensor`, is it?

Issue #3761, opened 2020-02-11
Continue training on different data


vin-ivar:
**System (please complete the following information):**
 - OS: Linux
 - Python version: 3.6.8
 - AllenNLP version: v0.9.0
 - PyTorch version: v1.2.0

**Question**
Is there any easy way to (sequentially) train on multiple datasets? The documentation says I shouldn't use `recover` - "For continuing training a model on new data, see `Model.from_archive`.", but `Model.from_archive` doesn't actually have any information about how I do this.

I assume this is not what AllenNLP refers to as fine-tuning, by the way - I'm fitting a dependency parser on top of a BERT encoder. Running `fine_tune_model` throws an out-of-vocab error for unknown tokens. Am I right in assuming that what I am looking for is indeed just training?

matt-gardner:
Are you writing your own training script, or using our commands?  It looks like you're writing your own script (this detail is important for what I recommend you do).  Also, can you give the specific error that you're seeing?  If you're just using BERT, you shouldn't be computing a vocabulary anyway, so my first thought at what was going on here is probably wrong.  We need more detail to be able to help you.

vin-ivar:
I'm writing my own training script, here's what it looks like:

```
config = Params.from_file(args.config, ext_vars={'train_path': args.train, 'val_path': args.val_train})
model = train_model(config, args.save, force=True)

finetune_config = Params.from_file(args.config, ext_vars={'train_path': args.finetune, 'val_path': args.val_finetune})
model = fine_tune_model(model, finetune_config, args.save + '_ft')
```

Here's the full dump from the error, it looks it's trying to look up the default unknown identifier in BERT's vocab:

```
Traceback (most recent call last):
  File "/Users/vinit/dev/bert-lang-distance/dep-parse/runner.py", line 60, in <module>
    main()
  File "/Users/vinit/dev/bert-lang-distance/dep-parse/runner.py", line 35, in main
    model = fine_tune_model(model, finetune_config, os.path.join(args.save, 'ft'))
  File "/Users/vinit/.miniconda3/miniconda3/envs/nlp/lib/python3.6/site-packages/allennlp/commands/fine_tune.py", line 318, in fine_tune_model
    metrics = trainer.train()
  File "/Users/vinit/.miniconda3/miniconda3/envs/nlp/lib/python3.6/site-packages/allennlp/training/trainer.py", line 478, in train
    train_metrics = self._train_epoch(epoch)
[...]

PR #3774, opened 2020-02-13
Add util to extend layer sizes

vin-ivar:
Helps with fine-tuning on larger output spaces. Addresses #3761

### DIFF
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -1680,19 +1680,18 @@ def find_embedding_layer(model: torch.nn.Module) -> torch.nn.Module:
     raise RuntimeError("No embedding module found!")
 
 
-def extend_layer(layer: torch.nn.Module, new_dim: int, layer_type: str="linear") -> torch.nn.Module:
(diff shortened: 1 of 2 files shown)


allennlp/nn/util.py
```diff
@@ -1678,3 +1678,32 @@ def find_embedding_layer(model: torch.nn.Module) -> torch.nn.Module:
                             return embedder
             return module
     raise RuntimeError("No embedding module found!")
+
+
+def extend_layer(layer: torch.nn.Module, new_dim: int, layer_type: str="linear") -> torch.nn.Module:
+    valid_layer_types = ["linear", "bilinear"]
```
matt-gardner: Maybe instead of this parameter you can just check `isinstance(layer, (torch.nn.Linear, torch.nn.Bilinear))`?  You're depending on specific field names in the module, so it's probably better to just check types.

vin-ivar: Yeah good idea - fixed that up a bit and added a test. Turns out the operation's in-place, by the way, hope that's not an issue.

PR #3775, opened 2020-02-13
add missing dependencies for build

DeNeutoy:
Turns out that you need the `wheel` package installed to build wheels, which we never knew because we did our builds in non-automated python environments 🤷‍♂


schmmd:
Sorry--I might have caused this regression.

PR #3778, opened 2020-02-13
More restrictive argument parsing

dirkgr:


allennlp/common/from_params.py
```diff
@@ -334,14 +334,17 @@ def construct_arg(
...
     # If the parameter type is a Python primitive, just pop it off
     # using the correct casting pop_xyz operation.
-    elif annotation == str:
-        return popped_params
-    elif annotation == int:
-        return int(popped_params)  # type: ignore
-    elif annotation == bool:
-        return bool(popped_params)
+    elif annotation in {str, int, bool}:
+        if type(popped_params) in {str, int, bool}:
```
brendan-ai2: Won't the popped param always be a float since it's coming from json? Highly suggest a test making it clear how this works.

dirkgr: I checked it in the debugger just now. Ints come in as ints.

dirkgr:
```
In [14]: type(json.loads(json.dumps({'x':1}))['x'])
Out[14]: int
```

brendan-ai2: Cool, thanks for correcting me! Still wouldn't mind that test just to prevent regressions.

dirkgr: Added in 4462721.

allennlp/tests/training/optimizer_test.py
```diff
@@ -68,7 +68,7 @@ def test_optimizer_parameter_groups(self):
 
     def test_parameter_type_inference(self):
         # Should work ok even with lr as a string
```
brendan-ai2: This appears to be a regression, but perhaps it's no longer relevant. @matt-gardner, does your recent work make this unnecessary? If so, this test should just be deleted.

dirkgr: Argh, I didn't even read the comment above. Clearly something has to change.

matt-gardner: I think the ability to handle this was added because of `std.extvar` substitution (or an earlier version of that) gave strings instead of floats in the json.  I'm ok removing it, but we may need to reconsider this at some point if we get people complaining about things not working with environment variables.  Should be easy to handle this again if we need to, though.

brendan-ai2: Ah, right. That does seem useful. But we can suggest people use `std.parseJson` (there is no std.parseFloat, see https://github.com/google/jsonnet/issues/759) as an alternative.

matt-gardner: To be clear, I'm really not sure if this is still necessary or not; I think we did this before we moved to jsonnet, and it may no longer be necessary.

brendan-ai2: For context, this change was added about a year ago when jsonnet was in place. See https://github.com/allenai/allennlp/pull/2324. (I hold no strong opinions on the functionality either way.)

dirkgr:
Codeconv is confused. I'm merging anyways.

PR #3781, opened 2020-02-14
Fix hang, make sharded reader rank aware

brendan-ai2:
- Previously imbalanced data would lead to workers hanging.
- I fixed this by adding code to detect this and stop the workers with excess data. This doesn't seem ideal, but I'm not sure of cleaner fix.
  - A user can mitigate this by ensuring that the number of shards is divisible by the number of workers. If the shards are also equally sized then no (or little -- depending on your batching mechanism) data will be lost.
- Makes the `ShardedDatasetReader` return data based on its rank.


brendan-ai2:
@dirkgr, with this PR the distributed training setup should be working for Transformer ELMo. However, I haven't validated it extensively and would like to call out a few issues that could affect both efficiency and perplexity:

1. It currently uses a full softmax with a token-based vocab. This is just wrong from an efficiency perspective. See the TODO here https://github.com/allenai/allennlp/pull/3781/files#diff-7e806e3220e0087294151062ab876fa2R34. That will almost certainly need to be fixed before we can beat the old method (which used a sampled softmax with sparse weights -- not compatible with `DistributedDataParallel`). Should be straightforward.

2. As mentioned in the description, be sure to use a worker count that divides the number of shards. Dropping training data could be an easy way to hurt the perplexity.

3. I basically threw together the mechanism to detect data imbalances. It requires extra synchronization between workers, so it could slow things down. `Model.forward` already implicitly synchronizes inside the same loop, so I don't suspect it will be drastically worse, but that's a hunch. Not something I've measured.

DeNeutoy:
@brendan-ai2 could you more precisely describe when this problem occurs? Is it only happening if we have more entire shards than workers, or does this also occur if we have the same number of workers as shards, but the shards are differently sized/produce different numbers of batches? 

Broadly I think we will fix this eventually by moving to the `Dataset/Sampler` model of loading data, which would mean that we can use [pytorch's explict DDP Sampler class](https://pytorch.org/docs/stable/data.html#torch.utils.data.distributed.DistributedSampler). Obviously in the mean time this seems like an appropriate fix.

brendan-ai2:
Mark, the hang was occurring whenever one worker had a batch to process (during train or validation) when the other did not. Having more shards than workers was fine, but the number of workers needed to divide the number of shards. Further, this hang could even occur when the number of shards was correct, but one shard had at least one extra batch. (This would affect a non-sharded reader too.) My understanding is that this is because `Model.forward` when the model is distributed results in a barrier. Since one of the workers has stopped processing data already, the other worker will then wait at that barrier indefinitely.

Agreed that this is likely fixed by the `Dataset/Sampler` mechanism. In particular, `drop_last` seems to be the relevant option.

I wouldn't say I'm super happy with the code as-is, but it's something to unblock further experimentation.

DeNeutoy:
I see, thanks. This does seem difficult to accomplish at the moment because of the way the workers batch data independently from each other, where the batches are possibly going to be differently sized.

allennlp/training/trainer.py
```diff
@@ -360,7 +360,27 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
         logger.info("Training")
 
         cumulative_batch_group_size = 0
+        stopped_early = False
```
dirkgr: This is a nit to pick, but I don't like the name `stopped_early`. "Early stopping" is something else.

allennlp/training/trainer.py
```diff
@@ -360,7 +360,27 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
...
         for batch_group in batch_group_generator_tqdm:
+            if self._distributed:
+                # Check whether the other workers have stopped already (due to differing amounts of
+                # data in each). If so, we can't proceed because we would hang when we hit the
+                # barrier implicit in Model.forward. We use a IntTensor instead a BoolTensor
+                # here because NCCL process groups apparently don't support BoolTensor.
+                done = torch.tensor(0, device=self.cuda_device)
+                torch.distributed.all_reduce(done, torch.distributed.ReduceOp.SUM)
+                if done.item() > 0:
```
dirkgr: Why do they all have to stop at the same time? Is there no way of getting each worker to finish their share of the data, and then be done?

allennlp/training/trainer.py
brendan-ai2: Heh, I originally called it `exit_early`, but that seemed worse. I will happily name it anything. I'll try `done_early`, but let me know if that's also bad...

allennlp/training/trainer.py
brendan-ai2: Not that I know of, easily... I could imagine finishing early, grabbing the weights from the DDP model, creating a new model restricted to the workers that still had data and restarting, but that's going to be pretty gnarly. Definitely out of scope for this PR.

allennlp/training/trainer.py
brendan-ai2: And fixed.

allennlp/training/trainer.py
dirkgr: The worker without data could just call `barrier()` a bunch until all workers signal that they're done. But let's fix that next time.

brendan-ai2:
Thanks!

gladuo:
@brendan-ai2 I found the "hang" still happened now in 1.3.0 when I use allen-models/rc during validation process

dirkgr:
What exactly are you running? Just TransformerQA on multiple GPUs?

epwalsh:
Probably fixed by https://github.com/allenai/allennlp-models/pull/193

allennlp/training/trainer.py
BladeSun: > The worker without data could just call `barrier()` a bunch until all workers signal that they're done. But let's fix that next time.

I suggest, at least in the `_validation_loss` function of the trainer, we can use the `barrier()` logic instead of this `done_early` logic? Since if we drop some samples during the validation phase, the val result will be incorrect.

PR #3789, opened 2020-02-17
Fix the nightly build version number

bryant1410:
**TL;DR:** the nightly build GH action is failing because PyPI doesn't accept the suffix "+commit-sha". This patch fixes it.

<details>
 <summary>Longer explanation</summary>

If you do `python setup.py check` with the nightly build version with the `ALLENNLP_VERSION_SUFFIX` used in the file, it's gonna say it's correct. If you check it with the [2nd regex in PEP 440 Appendix B](https://www.python.org/dev/peps/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions), it will also be fine. However, if you check it with the first one, you'll see it's not gonna be "canonical". There it says:

> As noted earlier in the _Public version identifiers_ section, published version identifiers SHOULD use the canonical format.

Versions that have a `+something` are called in [PEP 440 "local versions"](https://www.python.org/dev/peps/pep-0440/#local-version-identifiers). It says:

> Local version identifiers SHOULD NOT be used when publishing upstream projects to a public index server, but MAY be used to identify private builds created directly from the project source. Local version identifiers SHOULD be used by downstream projects when releasing a version that is API compatible with the version of the upstream project identified by the public version identifier, but contains additional changes (such as bug fixes). As the Python Package Index is intended solely for indexing and hosting upstream projects, it MUST NOT allow the use of local version identifiers.

Also, [TF uploads nightlies like the ones here but without that suffix](https://pypi.org/project/tf-nightly/#history), so that will work.
</details>

If keeping one commit per day then there won't be any problems with this change. Btw, to have in mind, I saw once a package called [versioneer](https://pypi.org/project/versioneer/) that helps you with this stuff.

Also, the "-" is converted to "." by setuptools. So I changed to reflect what it really is.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -3,7 +3,7 @@ name: Nightly Release
 on:
   # This is 11am in Seattle time.
   schedule:
-    - cron:  '0 19 * * *'
+    - cron: '0 19 * * *'
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


bryant1410:
You mean outside PyPI, right? What about keeping the `+commit_sha`? Or to keep the order between versions, you can use the output from `$(git describe)`.

Btw, [there was an authorization issue in the last nightly release attempt](https://github.com/allenai/allennlp/runs/451230890?check_suite_focus=true#step:9:16), in case you didn't see it.

PR #3797, opened 2020-02-17
Remove test which depends on Python 3.7.

schmmd:
This test uses pip internals which have changed between 3.6 and 3.7.

### DIFF
--- a/allennlp/tests/commands/main_test.py
+++ b/allennlp/tests/commands/main_test.py
@@ -12,7 +12,7 @@ from allennlp.common.checks import ConfigurationError
 from allennlp.common.plugins import discover_plugins
 from allennlp.common.testing import AllenNlpTestCase
 from allennlp.common.util import push_python_path
-from allennlp.tests.common.plugins_util import pip_install, push_python_project
+from allennlp.tests.common.plugins_util import push_python_project
 
 
 class TestMain(AllenNlpTestCase):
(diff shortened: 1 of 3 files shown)


bryant1410:
@schmmd I believe this depends on `pip>=20.0.0` (which supports Python 3.6), as defined in `dev-requirements.txt`, not in the Python version.

Also, removing them means to stop testing the plugins' behavior in most cases.

bryant1410:
I tested this class on Python 3 with `pip==20.0.2` and the tests work fine.

DeNeutoy:
@bryant1410 Is it only the test itself that relies on this `pip` version? If so I think we should just wrap this in a try except block, just so we can raise a coherent error. A user probably doesn't update their pip version very much, and it is not in the main dependencies. But I agree we should keep this test.

bryant1410:
It's actually only the import:

https://github.com/allenai/allennlp/pull/3797/files#diff-c4bfe2fce8a99b3bdbc17cc02bc8a022L3

We could add a try-except, and try with another one for a previous pip version (I think from pip 10 it's `pip._internal.main`).

Note this won't update the user pip, it just used for tests and it's only in `dev-requirements.txt` (not in setup.py).

DeNeutoy:
Ok, that sounds like a great fix - yep, the problem is that if someone runs `allennlp test-install`, this will almost certainly break, because no-one keeps their pip versions up to date. And it's very unclear from this error that the solution is to update the pip version - so in summary, I think your try/catch idea is great.

PR #3798, opened 2020-02-18
Add support for previous pip versions

bryant1410:
Fix https://github.com/allenai/allennlp/pull/3797#issuecomment-587226717


bryant1410:
I tested it on `pip==19.3.1`.

I _think_ the min version is 10, based on https://stackoverflow.com/a/49853826/1165181 (note that in pip<10 it's different also, but that's too old).

allennlp/tests/common/plugins_util.py
```diff
@@ -1,9 +1,12 @@
 from contextlib import contextmanager
 
-from pip._internal.cli.main import main as pip_main
-
 from allennlp.common.util import ContextManagerFunctionReturnType, PathType, push_python_path, pushd
 
+try:
```
DeNeutoy: Please can you add a comment about why this try except block is here? Just something like "pip moved this entrypoint somewhere between version 19-20, so we check both places, because it's likely that a user does not fix their version of pip". 

Also, are these the only two possible locations for this function? I'm trying to understand if it would be sensible to have the second import (in the except block) also with a try/catch, so we can recommend that someone upgrade their pip version if it does crash.

bryant1410: > Also, are these the only two possible locations for this function? I'm trying to understand if it would be sensible to have the second import (in the except block) also with a try/catch, so we can recommend that someone upgrade their pip version if it does crash.

For pip<10, yes. But that's kinda old. Do you think we should still do so?

bryant1410:
It's not working for 19.3.1, it doesn't more than one `pip install` in a row, don't know why, gonna check it out.

bryant1410:
I changed completely in the end. It's explained in the code.

allennlp/tests/common/plugins_util.py
```diff
@@ -13,11 +13,17 @@ def pip_install(path: PathType, package_name: str) -> ContextManagerFunctionRetu
     This method is intended to use with `with`, so after its usage, the package will be
     uninstalled.
     """
-    pip_main(["install", str(path)])
+    # Not using the function `main` from pip._internal because it assumes that once it finished,
```
DeNeutoy: Great explanation, thanks!

dev-requirements.txt
```diff
@@ -26,7 +26,7 @@ matplotlib>=2.2.3
 aiohttp
 
 # Required to run plugin tests.
-pip>=20.0.1
+pip
```
DeNeutoy: Is it relevant to have this here at all anymore, as pip will certainly be installed in a user's environment?

bryant1410: I'll remove it. It could happen in Docker or Conda for example, but if they use this file, they likely have it installed.

schmmd:
Love it, thanks @bryant1410 !

allennlp/tests/common/plugins_util.py
DeNeutoy: nah, that's fine then 👍

PR #3799, opened 2020-02-18
Remove unused old BERT test files and references

bryant1410:
And these config files in fixtures don't work anymore.


bryant1410:
Actually, `bert_pooler.jsonnet` could be kept, and this test could be recovered: https://github.com/allenai/allennlp/pull/3746/files#diff-b33604bc29905b3fdf0a3a5febd60d15L76-L80

bryant1410:
@DeNeutoy I think that actually the `bert_pooler.jsonnet` test can be recovered. What are your thoughts? It can give an example of how to do a BERT classifier now that `BertForClassification` is gone.

PR #3800, opened 2020-02-18
Exclude .DS_Store files from the Pip package

bryant1410:
I found one in the latest stable version.

Btw, "scripts", "training_config" and "tutorials" aren't being part of the packaging. Should those be removed as well?


schmmd:
@bryant1410 they probably should be excluded--but I'm not particularly expert on what should be included in `pip` packages.

PR #3806, opened 2020-02-18
Lets start_token and end_token be ints as well

dirkgr:


matt-gardner:
To be clear in what I mean for that comment (as I haven't reviewed many of your PRs yet): if I were doing this, I would probably add a simple test for loading this thing using `from_params` with an `int`, just to be sure we don't have a regression.  But it doesn't bother me if you want to merge without it.  Up to you.

dirkgr:
The test for models that use this fails, but that doesn't run in this repo. Also, the new behavior (since last week) is that it throws an exception when you get it wrong, instead of failing silently and giving you slightly worse performance. On balance, I think it's fine without an extra test.

PR #3807, opened 2020-02-18
Add support for weighted F1, Precision and Recall

bryant1410:


allennlp/training/metrics/fbeta_measure.py
```diff
@@ -198,6 +203,12 @@ def get_metric(self, reset: bool = False):
             precision = precision.mean()
             recall = recall.mean()
             fscore = fscore.mean()
+        elif self._average == "weighted":
+            weights = true_sum
+            weights_sum = true_sum.sum()
```
DeNeutoy: It's possible for this value to be zero, if the metric is called with a completely zero mask - you can use the `_prf_divide` function to guard against this 👍

bryant1410: Good catch!

PR #3808, opened 2020-02-19
Add PreCo dataset reader

ZhaofengWu:
Also fix `PretrainedTransformerTokenizer` to better handle tokens that are removed by the HuggingFace tokenizer. These tokens exist in the PreCo dataset.


allennlp/data/dataset_readers/coreference_resolution/preco.py
```diff
@@ -0,0 +1,89 @@
...
+from allennlp.data.tokenizers import PretrainedTransformerTokenizer
+from allennlp.data.token_indexers import TokenIndexer
+
+logger = logging.getLogger(__name__)
+
+
+@DatasetReader.register("preco")
+class PrecoReader(DatasetReader):
+    """
+    Reads a single JSON-lines file. Each line contains a "sentences" key for a list of sentences
```
DeNeutoy: Could you add a markdown link to the paper/dataset url as well?

allennlp/data/dataset_readers/coreference_resolution/conll.py
```diff
@@ -135,8 +132,8 @@ def text_to_instance(
             A list of lists representing the tokenised words and sentences in the document.
         gold_clusters : `Optional[List[List[Tuple[int, int]]]]`, optional (default = None)
             A list of all clusters in the document, represented as word spans. Each cluster
-            contains some number of spans, which can be nested and overlap, but will never
-            exactly match between clusters.
+            contains some number of spans, which can be nested and overlap. If there are
+            exactly matches between clusters, they will be resolved using `canonicalize_clusters`.
```
DeNeutoy: "If there are exact matches between clusters"

allennlp/data/dataset_readers/coreference_resolution/preco.py
```diff
@@ -0,0 +1,89 @@
...
+    def __init__(
+        self,
+        max_span_width: int,
+        token_indexers: Dict[str, TokenIndexer] = None,
+        wordpiece_modeling_tokenizer: Optional[PretrainedTransformerTokenizer] = None,
+        **kwargs,
+    ) -> None:
+        super().__init__(**kwargs)
+        self._conll_reader = ConllCorefReader(
+            max_span_width, token_indexers, wordpiece_modeling_tokenizer, **kwargs
```
DeNeutoy: rather than doing this funny composition, it might be nice to pull out the `text_to_instance` method of the `ConllCorefReader` into a `coreference_resolution/util.py` file, [like we have done for reading comprehension](https://github.com/allenai/allennlp-reading-comprehension/blob/master/allennlp_rc/dataset_readers/util.py#L163).

This might be good, because I know there is a new dataset that lots of people are evaluating on - [GAP](https://github.com/google-research-datasets/gap-coreference). If we wanted to add that as well, it would make it really easy.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -232,6 +233,9 @@ def intra_word_tokenize_in_id(
         cumulative = starting_offset
         for token in tokens:
             subword_wordpieces = self.tokenizer.encode(token, add_special_tokens=False)
+            if len(subword_wordpieces) == 0:
```
DeNeutoy: How can this happen? When it is a whitespace character?

So I think this is what was happening with the BERT model in the demo. Because if this was returning an empty list, all the offsets into the original tokens would be incorrect (I think). 

This needs a test, and also it would be great if you could verify that this was what was going wrong with the model in the demo, because we need to be absolutely sure of what's happening here and have it very well documented in the docstring of this class.

DeNeutoy:
Oh, also, the entire class could do with a test like this one: https://github.com/allenai/allennlp/blob/master/allennlp/data/dataset_readers/coreference_resolution/conll.py

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
ZhaofengWu: The most trivial case would be a whitespace character, but there could also some other unicode characters like `"\u007f"`. Agreed that this should be the cause of the demo problem. It's not that the offsets will be wrong -- they still are correct. The problem is that in the mismatched embedder, the length of these spans are 0. And when I needed the average span representation, there will be divide-by-zero's that cause Nan's which mess up all following tensor operations. So it's very likely that this is the cause, but I can verify later.

https://github.com/allenai/allennlp/blob/master/allennlp/modules/token_embedders/pretrained_transformer_mismatched_embedder.py#L82-L83

PR #3814, opened 2020-02-19
Allow specifying max_sentences in coref readers

ZhaofengWu:
Sometimes the documents are very long which creates memory issues when using large models and large representations. This option allows sampling a subset of sentences for each document. It should be used during training only.

### DIFF
--- a/allennlp/data/dataset_readers/coreference_resolution/util.py
+++ b/allennlp/data/dataset_readers/coreference_resolution/util.py
@@ -29,11 +29,23 @@ def make_coref_instance(
 
     sentences : `List[List[str]]`, required.
         A list of lists representing the tokenised words and sentences in the document.
+    token_indexers : `Dict[str, TokenIndexer]`
+        This is used to index the words in the document.  See :class:`TokenIndexer`.
+    max_span_width : `int`, required.
+        The maximum width of candidate spans to consider.
     gold_clusters : `Optional[List[List[Tuple[int, int]]]]`, optional (default = None)
         A list of all clusters in the document, represented as word spans with absolute indices
         in the entire document. Each cluster contains some number of spans, which can be nested
         and overlap. If there are exact matches between clusters, they will be resolved
         using `_canonicalize_clusters`.
+    wordpiece_modeling_tokenizer: `PretrainedTransformerTokenizer`, optional (default = None)
+        If not None, this dataset reader does subword tokenization using the supplied tokenizer
+        and distribute the labels to the resulting wordpieces. All the modeling will be based on
+        wordpieces. If this is set to `False` (default), the user is expected to use
+        `PretrainedTransformerMismatchedIndexer` and `PretrainedTransformerMismatchedEmbedder`,
+        and the modeling will be on the word-level.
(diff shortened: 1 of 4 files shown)


allennlp/data/dataset_readers/coreference_resolution/util.py
```diff
@@ -47,6 +49,31 @@ def make_coref_instance(
                 how many spans we are considering), we represent this a as a `SequenceLabelField`
                 with respect to the `spans `ListField`.
     """
+    if max_sentences is not None and len(sentences) > max_sentences:
```
DeNeutoy: Ah! This makes the `max_sentences` description a little off, because really this is randomly selecting a `max_sentences` size chunk from the document. Could you rename the parameter so it is more appropriate? something like `sample_to_length` or something.

allennlp/tests/data/dataset_readers/coreference/coref_reader_test.py
```diff
@@ -288,3 +288,25 @@ def check_candidate_mentions_are_well_defined(self, span_starts, span_ends, text
...
+        limited_instances = ensure_list(
+            limited_conll_reader.read(
+                str(AllenNlpTestCase.FIXTURES_ROOT / "coref" / "coref.gold_conll")
+            )
+        )
+
+        assert len(limited_instances) == len(instances) == 4
+
+        # Some are truncated, some short ones are not
+        length_of = lambda instance: len(instance.fields["text"].tokens)
```
DeNeutoy: Can you add a check that there are no spans from other sentences as well?

allennlp/data/dataset_readers/coreference_resolution/util.py
ZhaofengWu: Why do you think the `max_sentences` description is a little off? Because we are sampling instead of truncating? If so what about `max_sentences_to_sample`? I think `sample_to_length` is a little underspecified as to whether it's the number of sentences or tokens.

DeNeutoy: yeah, `max_sentences_to_sample` is better 👍  Basically just that if something says `max_sentences`, I expect it to be strictly a limit, and not an indication that some sampling is happening.

allennlp/tests/data/dataset_readers/coreference/coref_reader_test.py
ZhaofengWu: Because of the randomness it's a bit hard to test, e.g. `'X' in instance_a and 'X' not in instance_b` (if that's what you're referring to?). Alternatively I could just make the implementation to be truncation instead of sampling -- I don't think that makes too much of a difference anyway.

ZhaofengWu:
Ok I changed to sampling to truncation which is simpler and allows more detailed tests.

allennlp/tests/data/dataset_readers/coreference/coref_reader_test.py
ZhaofengWu: EDIT: changed to truncation.

PR #3815, opened 2020-02-19
Remove embedding from params

matt-gardner:
Following up on #3763.  Turns out the problem was just a mismatch between expected names.  The model was passing something called `vocab`, and the argument name in your original commit was `vocabulary`.  It's a bit magical to find this error, and unfortunate that we don't have better error handling here, but I'm not sure how to make this better.  The good news is that it's not magical if you're not using config files, and this change makes it easier to do that.

### DIFF
--- a/allennlp/models/coreference_resolution/coref.py
+++ b/allennlp/models/coreference_resolution/coref.py
@@ -101,7 +101,9 @@ class CoreferenceResolver(Model):
 
         # 10 possible distance buckets.
         self._num_distance_buckets = 10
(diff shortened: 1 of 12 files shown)


matt-gardner:
There are some other failures that are coming up, so this isn't quite ready yet, but that was the biggest issue.

matt-gardner:
I don't think there was anything surprising or controversial in what I did after you approved this (it was just fixing all calls to `Embedding()`), so I'm merging this.

PR #3817, opened 2020-02-19
Looks like the way to call tokens_to_indices changed

dirkgr:
This makes the demo run again, but I am not sure this is the right way to fix it.


dirkgr:
Looking at this (https://github.com/allenai/allennlp/pull/3597/files#diff-18ed7be54d4a8faa7ff0311fd1c073fbR141), I think this is the right way to fix the issue.

PR #3818, opened 2020-02-19
Removing multi-lang code

matt-gardner:
@TalSchuster, we appreciate this contribution, but we're moving all model code out of the main repository, to reduce our maintenance burden and make the core library itself more lightweight and easier to work with.  We're putting some of the models into other repositories that we'll maintain (e.g., [this one](https://github.com/allenai/allennlp-reading-comprehension)), but this code isn't going into one of our repos.  If you'd like it to still be available, just put it in a public repo, and anyone that wants to can use these components pretty easily.


TalSchuster:
@matt-gardner, that's unfortunate. I know quite a few that are using it for multilingual models with Elmo or Bert. 

I'm keeping a public version [here](https://github.com/TalSchuster/allennlp-MultiLang) but hope that allennlp will return this support in the future to encourage multilingual models.

matt-gardner:
@TalSchuster, we're moving *everything* out, not just this.  Only some of that is done so far, but it's going to ramp up very quickly over the next couple of weeks.  It's really easy to add components to the library, so this shouldn't be a hinderance to anyone wanting to use these multi-lingual elmo models.  It's even possible to make a pip package that adds an allennlp plugin that's immediately available in config files, if you want.  We can give you pointers on how to do that, we just don't have the resources to do it for you (we're a *really* small team).

PR #3825, opened 2020-02-20
Remove Pruner module and replace it with a masked_topk util function

ZhaofengWu:
This makes the utility more general in that:
(1) the client can self-generate scores arbitrarily without being forced to encapsulate all scoring logic in a single `torch.nn.Module` that just takes the embeddings as input;
(2) allows input with an arbitrary number of dimensions;
(3) allows selection on any dimension.


docs/tutorials/how_to/span_representations.md
```diff
@@ -110,54 +110,6 @@ assert spans == [(0, 1), (0, 2), (1, 2), (1, 3), (2, 3)]
 There are other helpful functions in `allennlp.data.dataset_readers.dataset_utils.span_utils`,
 such as a function to convert between BIO labelings and span-based representations.
 
-### Use a Pruner
```
ZhaofengWu: I removed this section in the tutorial because it doesn't feel that closely related to span representations, especially after this change. But please lmk if we want to keep it but change it to use the new `masked_topk`.

allennlp/models/coreference_resolution/coref.py
```diff
@@ -78,12 +78,11 @@ def __init__(
...
         self._context_layer = context_layer
-        self._antecedent_feedforward = TimeDistributed(antecedent_feedforward)
-        feedforward_scorer = torch.nn.Sequential(
-            TimeDistributed(mention_feedforward),
-            TimeDistributed(torch.nn.Linear(mention_feedforward.get_output_dim(), 1)),
+        self._mention_feedforward = TimeDistributed(mention_feedforward)
+        self._mention_scorer = TimeDistributed(
+            torch.nn.Linear(mention_feedforward.get_output_dim(), 1)
         )
-        self._mention_pruner = Pruner(feedforward_scorer)
```
DeNeutoy: I think this is going to break the serialised coref models, because it will have the parameters in a different place now?

ZhaofengWu: Yes. I already have a model trained with `maked_topk`, and anticipate to get a final model ~tomorrow. I will update the model then.

ZhaofengWu:
@DeNeutoy Then the Pruner will merely look something like
```
class Pruner(...):
  def forward(self, scores, mask, k):
    return masked_topk(scores, mask, k)
```

Not sure if that provides much value. If clarity is an issue maybe I can add some usage example to the function doc, or keep the section in the tutorial but just change `Pruner` to `masked_topk`?

PR #3826, opened 2020-02-21
Removing unnecessary indexers

### LINKED ISSUE #3788: Remove unnecessary token indexers
#3769 adds a flag to `SingleIdTokenIndexer` which would let us very easily replace several of our other indexers with that one, just using that flag.  I think this is an improvement.  The particular indexers are the dependency label, NER, and pos tag indexers.

We should probably do this before the 1.0 release, as removing those is a breaking change.

matt-gardner:
Fixes #3788. The `text_id` behavior is a relatively large change, but I think it makes things less magical, and the original behavior was probably never used, anyway.


scripts/train_fixtures.py
```diff
@@ -66,17 +66,17 @@ def train_fixture_gpu(config_prefix: str) -> None:
...
+                #"decomposable_attention",
                 "encoder_decoder/composed_seq2seq",
                 "encoder_decoder/simple_seq2seq",
                 "encoder_decoder/copynet_seq2seq",
-                "esim",
-                "simple_tagger_with_span_f1",
-                "srl",
+                #"esim",
+                #"simple_tagger_with_span_f1",
+                #"srl",
```
dirkgr: Are you still working on this? This looks like you're in the middle of things.

matt-gardner: Thanks for catching this one, this was not an intentional commit.  It's mostly done, just getting CI to pass at this point.

allennlp/data/token_indexers/single_id_token_indexer.py
```diff
@@ -16,7 +19,10 @@ class SingleIdTokenIndexer(TokenIndexer):
     # Parameters
 
     namespace : `str`, optional (default=`tokens`)
-        We will use this namespace in the :class:`Vocabulary` to map strings to indices.
+        We will use this namespace in the :class:`Vocabulary` to map strings to indices.  If you
+        explicitly pass in `None` here, we will skip indexing and vocabulary lookups.  This means
```
dirkgr: The type annotation for `namespace` should be `Optional[str]` then.

allennlp/data/token_indexers/single_id_token_indexer.py
```diff
@@ -8,6 +8,9 @@
 from allennlp.data.token_indexers.token_indexer import TokenIndexer, IndexedTokenList
 
 
+_DEFAULT_VALUE = "THIS IS A REALLY UNLIKELY VALUE THAT HAS TO BE A STRING"
```
dirkgr: The way `_DEFAULT_VALUE` is used now, it could just be `None`, with no surprising strings like this one.

matt-gardner: The reason I did this was to allow `None` to be explicitly passed as a default value, if you wanted that for some reason.  I originally just had `default_value = None` in the constructor, but then you can't actually have `None` as a default value.

matt-gardner: And then I had `_DEFAULT_VALUE = object()`, but then I got type complaints because `object()` isn't a `str`, which is what the type of `default_value` is annotated as.

dirkgr: It's a fair point. I don't like the solution, but I don't have a better idea.

allennlp/data/token_indexers/single_id_token_indexer.py
matt-gardner: Done.

matt-gardner:
Just FYI on process that we've converged on @dirkgr; if your remaining feedback is all minor, we most often will approve the PR, trusting the developer to clean things up at their discretion before merging.  This is mostly just to save reviewer time, so you don't feel like you have to look at it again, and also iteration time, so the dev doesn't have to wait for another round of review.

It doesn't bother me if you want to do something different, but this is what others on the team will most often do.

dirkgr:
I'm down with that process.

Issue #3519, opened 2019-12-13
Consider removing CallbackTrainer


DeNeutoy:
The callback trainer hasn't really worked, because callbacks that we've tried to add have required setting state on the `CallbackTrainer` itself, which makes them hard to add. Given this, we are just maintaining 2 Trainers unnecessarily, which slows us down.

Happy to hear reasons why we should keep the callback trainer/ if people have found it particularly useful!

matt-gardner:
I would have voted the other way, that we should remove the original trainer.

DeNeutoy:
I'm opposed, because:

- It contains code which is hard to reason about:
https://github.com/allenai/allennlp/blob/master/allennlp/training/callbacks/log_to_tensorboard.py#L45

- It's an unusual programming style to have an object pass itself to handlers, which I think is confusing to readers/users

- It requires setting state on the trainer which would otherwise not be necessary: https://github.com/allenai/allennlp/blob/master/allennlp/training/callback_trainer.py#L113-L136

- It still contains hard coded batch updates etc. Even the toy GAN trainer has to override the base class - at that point, I think we should be recommending that users write their own scripts. (FYI I also think this is true of `fine-tune` - we should just make allennlp easier to use in a scripting fashion and remove this, because it's impossible to cover all use cases)

matt-gardner:
To summarize some in-person discussion: because training GANs and typically training are so different, we explicitly *don't* want to have one class that can do both, because it means that it's too general.  What we want instead is to have re-usable components that make it easy to do normal training, GAN-style training, and customization in your training loop.  We're not certain yet exactly what form that will take, but we don't think either the original trainer or the callback trainer really give us that.

epwalsh:
FWIW I love the callback trainer - even if it's implementation / API isn't perfect - because it makes it really easy to customize logging and add some other useful stuff to the training loop.

mojesty:
I like CallbackTrainer too, because it makes extending training process _far_ easier then the in original Trainer. Actiually, setting state of the CallbackTrainer is a bit painful, however it's still better than re-writing the whole code.

epwalsh:
If anyone has thoughts on how to improve it though I wouldn't mind putting in some work there.

matt-gardner:
[...]

PR #3828, opened 2020-02-21
remove callback trainer

DeNeutoy:
Removing this so that #3700  is a bit easier, but we are planning to integrate the callbacks into the original trainer.


bryant1410:
(related to #3519)

DeNeutoy:
Thanks!

viking-sudo-rm:
Curious since I have been using custom callbacks to implement magnitude pruning: is the plan to reintegrate all existing callbacks, or only a subset of them? My use case in particular depends on end of epoch and end of training

matt-gardner:
Please comment on #3519 with some more detail on what exactly you would like to keep, and we'll see what we can do.  I had a list of things I was going to do, and modifying `grad_clipping` and `grad_norm` didn't make the list, but if there's demand for generalizing that functionality, we can maybe figure that out.

PR #3829, opened 2020-02-21
 Fix plugins were not discovered from the current directory

bryant1410:
The first commit shows this behavior is broken, as the tests fail with it.

The thing was that I thought the Python path (`sys.path` + `PYTHONPATH`) adds by default the current directory, but in fact, [it adds the script directory by default](https://docs.python.org/3/library/sys.html#sys.path).

It's okay to load plugins from the current directory, right? (Note that invocations of `allennlp` from the user in their subdirectories won't work).

The code change is essentially moving the `push_python_path(".")` call from the tests to inside of `discover_plugins()`.


allennlp/common/plugins.py
```diff
@@ -66,9 +67,10 @@ def discover_plugins() -> Iterable[str]:
     """
     Returns an iterable of the plugins found.
     """
-    for module_info in discover_namespace_plugins():
-        yield module_info.name
-    yield from discover_file_plugins()
+    with push_python_path("."):
```
dirkgr: What does this do when I run something from some unexpected working directory?

bryant1410: What do you mean? This pushes a path string into `sys.path`.

dirkgr: Yes, I get it now.

PR #3830, opened 2020-02-21
Rename run.py to __main__.py

bryant1410:
I'm proposing this change because `__main__.py` is the standard location for this (executing the main use case of a python package). So it makes `python -m allennlp` work (as opposed to `python -m allennlp.run`, which you have to discover that filename by yourself. Apart from helping the discoverability from the command line and following a standard, this eases the discoverability of running it as a module from Python as well.

It's okay if you guys don't accept this change. I send a PR (as opposed as an issue for discussion) cause it was easy to do.


matt-gardner:
No complaints from me on this one.

PR #3835, opened 2020-02-22
Make instantiation logging in FromParams use debug instead of info

matt-gardner:
We don't need this much verbose logging, especially now that we don't have custom `from_params` methods anymore.


DeNeutoy:
Maybe add something to the release notes about this, as it's quite a visual difference when running any of our commands?

Issue #3562, opened 2019-12-30
Use of invariant vs. covariant types in function/method signature annotations


willfrey:
**System (please complete the following information):**
 - OS: macOS Catalina Version 10.15.2
 - Python version: 3.7.5
 - AllenNLP version: v0.9.1-unreleased (master)
 - PyTorch version: 1.3.1

**Question**
I've encountered function/method signature annotations in the codebase that are invariant but could easily be relaxed to be a covariant equivalent. For example the (signature of block_orthogonal)[https://github.com/allenai/allennlp/blob/master/allennlp/nn/initializers.py#L101] could be changed from

```python
def block_orthogonal(tensor: torch.Tensor, split_sizes: List[int], gain: float=1.0) -> None:
    ...
```

to

```python
def block_orthogonal(tensor: torch.Tensor, split_sizes: Sequence[int], gain: float=1.0) -> None:
    ...
```

Would it be helpful if I changed these annotations as I encounter them and submit PRs?

I wanted to ask before instead of just diving into it because there may be some underlying reason you have invariant argument annotations that I'm not aware of, perhaps something to do with the `FromParams` mixin.

Thanks!

matt-gardner:
Is there an actual problem that you encounter with those type annotations?  E.g., a mypy failure, or something?

Our reasoning initially was that the type annotations exist almost entirely for readability.  And for the average reader, seeing `List` instead of `Sequence` (and similar for other types, like `Dict` vs. `Mapping`) is more familiar.  So while `Sequence` would be more technically "correct", for our purposes we just went with the simpler ones.

We now also actively inspect the type annotations, and so there might also be some logic in `FromParams` or other places that depends on those particular types, but I *think* it's already handled correctly, and failing tests should tell us if it doesn't.

My vote is to stick with the simpler types, unless there's a compelling reason, like something actually broken, to switch them.

willfrey:
I've found myself using `types.MappingProxyType` from time to time to provide immutable views for some sort of default mapping of objects such as a default mapping of token indexers. `mypy` will yell at me for this.

Additionally, annotating with `Dict` gives me no information about whether or not that dictionary is going to be mutated by the method. If it was annotated with `Mapping`, then I know it shouldn't mutate my dictionary, whereas `MutableMapping` tells me it could mutate.

`mypy` will also give you warning notes recommending that you use `Mapping` instead of `Dict` or `Sequence` instead of `List`.

matt-gardner:
[...]

PR #3838, opened 2020-02-22
Fix class creation `from_params` with `Iterable`s and `Mapping`s

bryant1410:


allennlp/tests/common/from_params_test.py
```diff
@@ -680,3 +680,68 @@ def __init__(self, lazy_object: Lazy[ConstructedObject]):
...
+        @A.register("b")
+        class B(A):
+            def __init__(self, size: int) -> None:
+                self.size = size
+
+        class C(Registrable):
+            pass
+
+        @C.register("d")
+        class D(C):  # noqa
```
matt-gardner: Why do you need all of these `# noqa` statements?

bryant1410: I can remove them. It was just that I have warnings of unused class or wrong types.

matt-gardner: Due to local IDE stuff that isn't reflected in our CI?

allennlp/common/from_params.py
```diff
@@ -468,6 +467,31 @@ def constructor(**kwargs):
             return value_cls.from_params(params=deepcopy(popped_params), **constructor_extras)
 
         return Lazy(constructor)  # type: ignore
+
+    # This condition has to be at the end because other types can also be iterables
+    # (e.g., `str`, `tuple`, and `dict`).
+    elif (
+        origin in {collections.abc.Iterable, Iterable, List, list}
```
matt-gardner: I'm a bit conflicted on using `Iterable` here, mostly because of the need to move it to the end - is this really something we should be handling?  It's much more obvious to me to accept `Sequence` than `Iterable` (see also https://github.com/allenai/allennlp/issues/3562).

allennlp/tests/common/from_params_test.py
bryant1410: Yes.

matt-gardner: I think our policy should be that directives like this that get checked in should strictly be for things that our CI does.  Otherwise it could eventually become a jumbled mess with people adding things for their favorite checker that we're not actually using.

allennlp/common/from_params.py
matt-gardner: (Also, after resolving this issue one way or the other, I think this looks good and is ready to merge; thanks for the PR!)

bryant1410: > I'm a bit conflicted on using `Iterable` here, mostly because of the need to move it to the end - is this really something we should be handling? It's much more obvious to me to accept `Sequence` than `Iterable` (see also #3562).

Why? I think the author of #3562 suggests `Sequence` instead of `Iterable` there because `split_sizes` within the function `block_orthogonal` gets iterated many times. `Iterable` is the most general way of defining something you can iterate on.

We would still need to move it to the end even with `Sequence`. Could you explain why is this a problem?

allennlp/tests/common/from_params_test.py
bryant1410: Makes sense :+1:

allennlp/common/from_params.py
bryant1410: (`Sequence` is gonna be supported automatically by supporting `Iterable`)

bryant1410: (reference: https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes)

matt-gardner: I think the distinction is that `Sequence` is not a super-type of `dict`, while `Iterable` is.  I don't think you would have to move this to the end if you just use `Sequence` (though you might need to move it after `Tuple`).  It's not the moving that bothers me, it's what the fact that you have to move it implies.  `Iterable` is a super general class; should our default behavior really be to create a list for it?  I'm more tempted to just punt on that case, and not try to construct something so general.  Though I can see an argument that constructing a list when you ask for an iterable isn't such a bad thing to do.  It just makes me pause, where adding `Sequence` here wouldn't make me pause at all.

bryant1410: I think it makes sense for our default behavior (when many if-conditions aren't met) to create a list for it (or any iterable) provided it's `Iterable`.

Also, if we don't support it, users could use `Iterable` in their constructors and it'd fail.

matt-gardner: Well, yes, the question is whether it's reasonable to fail if someone chooses to use `Iterable` - what exactly did they want, and is a list sufficient?  It probably is.

Maybe just change the comment to say something like, "for any other kind of iterable, we will just assume that a list is good enough, and treat it that same as `List`.  This condition needs to be at the end, so we don't catch other kinds of `Iterables` with this branch."

PR #3840, opened 2020-02-24
Expand documentation of modules.conditional_random_field

DNGros:
This PR attempts to expand the documentation the ConditionalRandomField code. Specifically I found the details around the mask and start-end tokens to be confusing on first reading. Hopefully this makes it bit more explicit. 

This is a simple docs-only change. In future might come back and try clean up the code a bit / improve performance (The CRF layer seems to add a pretty non-trivial overhead to training time. Given the sequential loops and lots of pointwise-ops, it seems like a prime use case of TorchScript, but currently it seems the code would need some changes to be compatible.)

Feedback welcome.


allennlp/modules/conditional_random_field.py
```diff
@@ -319,7 +322,25 @@ def forward(
...
+            A tensor of logit values representing the prediction for each tag. It
+            should have shape `(batch_size, seq_len, num_tags)`.
+        tags : `torch.Tensor`
+            A tensor which is the expected tag labels. It should have shape of
+            `(batch_size, seq_len)`
+        mask : `torch.Tensor`
+            An optional mask which is True where we care about the labeling. So
+            for example if one has pad's in the sequence it should be False there.
+            It can optionally be false in the middle of the sequence, which
+            corresponds to pretending that token is not in the sequence. When
```
matt-gardner: "pretending that token is not in the sequence" is not quite accurate - the mask only affects the loss computation.  We still include the token when computing transitions.  That is, we don't, like, connect the previous token with the next non-masked token, as you might think with "pretending that token is not in the sequence".  You might change this sentence to, "It can optionally be false in the middle of the sequence, which is used, e.g., to compute the loss for a single predicted span, so we can get input gradients of just that span's loss."

matt-gardner: It *does* look like that's what's happening in `viterbi_decode`, though, which is surprising to me.  That makes me wonder if there's a bug in our interpret code.  It wouldn't affect the loss, though, so the input gradients would still be correct.

matt-gardner: No, it's probably ok, because the logits still come from the unmasked sequence, and the span should be contiguous.  So you're getting out the same result you would, other than the START and END transitions...

DNGros: Thanks for the feedback! Sorry about the formatting/CI issues.

It does seem that with the current code my statement was incorrect. From
https://github.com/allenai/allennlp/blob/4b467ec2882c49c9162d37f40fb55e08d7f5fee0/allennlp/modules/conditional_random_field.py#L249-L253
I got the impression that "skipping over as if it doesn't exist" might have been intended in `forward()` like what is done in `viterbi_decode`. However, I don't think this is actually what the code does, as if there are multiple mask False values in a row, alpha will go to zero, not retain the previous token. This means if we had a mask like [[0, 1, 1, 0, 0, 1, 1]] we would "forget" about the likelihood of the first span.

Additionally it seems like https://github.com/allenai/allennlp/blob/4b467ec2882c49c9162d37f40fb55e08d7f5fee0/allennlp/modules/conditional_random_field.py#L298-L301
will potentially treat the middle of the sequence as connecting to END if there are False values in the middle of the mask. In order to be able to distinguish between actual ends, and masks that end early, I think the user would need to pass in sequence lengths into `forward()`. 

Similarly masking for the START is not being handled.

Given the loss values are not quite correct, it seems like masks which do not represent pads at the end should be disallowed. 9b5c9c0 explains this and adds a check for unsupported masks.

If this seems ok, I can make an issue explaining how the calculation might be wrong, and add a commit to this PR linking to it in the docstring / error message. Having False values in places besides the end of the mask seems like a rather niche/low-priority use case for most models (I might be missing something though).

I feel a little guilty committing something that introduces a new exception, so I could also change this to raise a warning explaining the loss might not be correct for non-pad masks, rather than exception, if we think there are people who might depend the current (if slightly incorrect) behavior.

matt-gardner: Thanks for digging into this!  I don't think we can have this check, though, because we rely on being able to put masks on arbitrary tokens in our interpret code (see [here](https://github.com/allenai/allennlp/blob/master/allennlp/interpret/attackers/input_reduction.py#L197-L222)).  I'm surprised this is passing tests, actually, because we have a [test that looks at input reduction for NER](https://github.com/allenai/allennlp/blob/4b467ec2882c49c9162d37f40fb55e08d7f5fee0/allennlp/tests/interpret/input_reduction_test.py#L32-L49).  Maybe that test, because it's using a tiny fixture as its prediction model, isn't actually testing this sufficiently.

On your first point, I don't see how alpha goes to zero if there are multiple 0s in the mask.  I don't think it does, because at every step you're carrying forward the previous value of alpha.  Otherwise having multiple 0 values at the end would also zero out the whole loss, which isn't what's happening.

So, I still think the loss is doing the right thing, it's just `viterbi_decode` that is wrong.  I'm fine with having a warning (but not a crash) in `decode` that the logic isn't quite right, as long as it's not overly slow to run the check, but we don't need one in the loss computation.

DNGros: Ok, didn't know about that use case with interpret. Seems interesting. Thanks for pointing that out.

Hm, yeah, that's right. Not sure why I thought alpha was looking at the previous mask value, instead current. Sorry about that.

Should `forward()` accept non-pad masks when include_start_end_transitions is True? It does seem that the START/END transitions and start input_likelihood logits are not being handled correctly. A warning could be raised unless the user either gives `forward()` a possibly newly introduced "force_disable_start_end_transitions" arg or an "original_seq_lens" arg so the END transitions could be correct. (I understand this does make much difference in loss value unless get unlucky and have `mask.sum(0).long() - 1` point to something that typically never transitions to END).

matt-gardner: Ah, sorry, I somehow missed the bit about the end transitions in your original comment (not sure how, it's pretty clear... I guess I dug into the alpha values and then came back and skipped to the wrong part of your comment).  Yeah, that's wrong...  I don't see an error in the start transitions, but the line you highlighted where we compute `last_tag_index` is definitely wrong.

I think the right solution here, though, is to put a warning in the docstring and open a new issue to fix that calculation (and the logic in `viterbi_decode`).  I don't think the solution is to crash.  This is something we want to support, it's currently at least close, and it was good enough to give reasonable output when we were developing the interpret code.

schmmd:
@DNGros thanks!  We will need the tests to pass before we can merge.

DNGros:
@schmmd ok, sorry to leave this hanging. I got caught up in some other things. 
I'll try to address it on Sunday or early next week as at very least a mergeable docs-update with corresponding issue for any identified issues with loss/viterbi for later PR's if it effects others.

dirkgr:
@DNGros, do you think you'll have time to finish this? It's possible the code now lives in the allennlp-models repo, but it's still the same code.

Issue #3824, opened 2020-02-20
`Metric`s should be related to the task/dataset and not to the `Model`, maybe?


bryant1410:
**Is your feature request related to a problem? Please describe.**

I find myself creating instances of the same `Metric`s when I build `Model`s for the same `DatasetReader`, so I wonder if they should be closer to the "task" or "dataset" than to the model.

I wanted to hear opinions on this at least.

**Describe the solution you'd like**

I know this is a breaking change, but somehow `Metric`, IMHO should be closer to the data. Maybe to `DatasetReader` (and `Model.forward` can be expected to return `scores` as well as the `loss`)? I'm not convinced, because it should just read the data. However, it knows about the kind of data. Like in BiDAF, it says `squad_metrics`: https://github.com/allenai/allennlp-reading-comprehension/blob/e4f8e5df4f9fa35287d44e94fc8b26b9cabed0a5/allennlp_rc/models/bidaf.py#L329-L337

**Describe alternatives you've considered**
An alternative could be to have a "base" model that defines the metrics, but still, I think we metrics are related to the task/problem. Like in SQuAD, you care about F1 and ExactMatch, regardless of the model.

bryant1410:
Maybe the `Trainer` (which already does a lot of stuff..)?

matt-gardner:
Let's say the `Metric` lives with the `DatasetReader`.  How does it actually get used?

bryant1410:
An iterable or dict of `Metric`s are obtained from the dataset from `TrainerModel` and passed to the `Trainer`, which does gradient descent, calling `Model.forward` for each batch, getting the `predictions` (a tensor; or the raw `scores`) as well as the `loss`, passing the former to each `Metric`, and then computing the value from each `Metric` (which preferably IMHO should return a `Dict[str, float]` so it's a homogeneous API).

matt-gardner:
I guess I don't really see what's gained by making this change.  One thing that's lost is that the `Model` now has more constraints on its output, instead of just having to return a `loss`.  How do you know which key to pass predictions in?  And what if you want different metrics that depend on different predictions?  How do you handle the mapping?  What if you want metrics on intermediate model computations?  Seems strictly less flexible to do what you propose, and we do have concrete uses for all of what I said above, which would be hard in this setting.

bryant1410:
[...]

PR #3846, opened 2020-02-25
Make `BasicClassifier` accept any metric

bryant1410:
Make `BasicClassifier` accept any metric, passed as args.

I didn't add tests yet because I want to receive feedback first. I'm leaving some comments on the code.

### DIFF
--- a/allennlp/training/metrics/metric.py
+++ b/allennlp/training/metrics/metric.py
@@ -1,4 +1,5 @@
-from typing import Dict, Optional, Tuple, Union, List
+from typing import Dict, Iterable, List, Optional, Tuple, Union
+import collections.abc
 import torch
 
 from allennlp.common.registrable import Registrable
@@ -33,6 +34,26 @@ class Metric(Registrable):
         Compute and return the metric. Optionally also call `self.reset`.
         """
         raise NotImplementedError
+
+    def get_metric_name_value_pairs(
+        self, default_name: str, reset: bool = False
+    ) -> Iterable[Tuple[str, float]]:
+        """
+        Return the metric as in `self.get_metric` but as an iterable of string-float pairs.
+        """
+        value = self.get_metric(reset)
+        if isinstance(value, collections.abc.Mapping):
+            for sub_name, sub_value in value.items():
+                if isinstance(sub_value, collections.abc.Iterable):
+                    for i, sub_value_i in enumerate(sub_value):
+                        yield f"{sub_name}_{i}", sub_value_i
+                else:
+                    yield sub_name, sub_value
+        elif isinstance(value, collections.abc.Iterable):
+            for i, sub_value in enumerate(value):
+                yield f"{default_name}_{i}", sub_value
+        else:
+            yield default_name, value
 
     def reset(self) -> None:
         """
--- a/allennlp/training/metrics/metric.py
+++ b/allennlp/training/metrics/metric.py
@@ -12,7 +12,10 @@ class Metric(Registrable):
     """
 
     def __call__(
-        self, predictions: torch.Tensor, gold_labels: torch.Tensor, mask: Optional[torch.Tensor]
+        self,
+        predictions: torch.Tensor,
+        gold_labels: torch.Tensor,
+        mask: Optional[torch.Tensor] = None,
     ):
         """
         # Parameters
@@ -50,8 +53,8 @@ class Metric(Registrable):
                 else:
                     yield sub_name, sub_value
         elif isinstance(value, collections.abc.Iterable):
-            for i, sub_value in enumerate(value):
-                yield f"{default_name}_{i}", sub_value
+            for i, sub_value in enumerate(value):  # type: ignore
+                yield f"{default_name}_{i}", sub_value  # type: ignore
         else:
             yield default_name, value
 
--- a/allennlp/models/basic_classifier.py
+++ b/allennlp/models/basic_classifier.py
@@ -173,6 +173,6 @@ class BasicClassifier(Model):
     def get_metrics(self, reset: bool = False) -> Dict[str, float]:
         return {
             name: value
-            for default_name, metric in self.metrics.items()
+            for default_name, metric in self._metrics.items()
             for name, value in metric.get_metric_name_value_pairs(default_name, reset)
         }
--- a/allennlp/models/basic_classifier.py
+++ b/allennlp/models/basic_classifier.py
@@ -8,7 +8,7 @@ from allennlp.models.model import Model
 from allennlp.modules import FeedForward, Seq2SeqEncoder, Seq2VecEncoder, TextFieldEmbedder
 from allennlp.nn import InitializerApplicator
 from allennlp.nn.util import get_text_field_mask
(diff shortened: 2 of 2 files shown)


allennlp/training/metrics/metric.py
```diff
@@ -11,7 +12,10 @@ class Metric(Registrable):
     """
 
     def __call__(
-        self, predictions: torch.Tensor, gold_labels: torch.Tensor, mask: Optional[torch.Tensor]
+        self,
+        predictions: torch.Tensor,
+        gold_labels: torch.Tensor,
+        mask: Optional[torch.Tensor] = None,
```
bryant1410: I changed this because mypy was complaining about the line that says `metric(logits, label)`, stating there's a missing param, which is actually true. I could have set it to `None`, but then saw the documentation and it says "default = None", which is actually false, so I changed it.

allennlp/training/metrics/metric.py
```diff
@@ -34,6 +38,26 @@ def get_metric(
         """
         raise NotImplementedError
 
+    def get_metric_name_value_pairs(
```
bryant1410: This function is necessary to have a homogeneous output out of `get_metric`. It can be here, or it could be a util function maybe.

allennlp/training/metrics/metric.py
```diff
@@ -34,6 +38,26 @@ def get_metric(
...
+        value = self.get_metric(reset)
+        if isinstance(value, collections.abc.Mapping):
+            for sub_name, sub_value in value.items():
+                if isinstance(sub_value, collections.abc.Iterable):
+                    for i, sub_value_i in enumerate(sub_value):
+                        yield f"{sub_name}_{i}", sub_value_i
+                else:
+                    yield sub_name, sub_value
+        elif isinstance(value, collections.abc.Iterable):
+            for i, sub_value in enumerate(value):  # type: ignore
```
bryant1410: mypy associates the previous variable (`sub_value`) with the same name the same type as this one, so I added this ignore.

allennlp/training/metrics/metric.py
```diff
@@ -34,6 +38,26 @@ def get_metric(
...
+                yield f"{default_name}_{i}", sub_value  # type: ignore
```
bryant1410: Even when changing the variable name, mypy says `sub_value` is of type `object` instead of `float`.

allennlp/training/metrics/metric.py
```diff
@@ -34,6 +38,26 @@ def get_metric(
         """
         raise NotImplementedError
 
+    def get_metric_name_value_pairs(
+        self, default_name: str, reset: bool = False
```
bryant1410: Maybe a default default name can be created from the underlying class name :man_shrugging:

allennlp/training/metrics/metric.py
bryant1410: Another option is to change `get_metric`.

matt-gardner:
I think the changes to the basic classifier are fine, but I don't really see the value in the changes to `Metric`.  You're trying to make a uniform API for `Metric`, but it doesn't actually exist, unless you do something like `*args`.  Mypy is just going to complain.  We could try to make a uniform API by adding subclasses like `ClassificationMetric`, but that's just confusing and not worth the complexity.

bryant1410:
> You're trying to make a uniform API for Metric, but it doesn't actually exist, unless you do something like *args. Mypy is just going to complain.

Do you mean when calling `metric(logits, label)`?

bryant1410:
> You're trying to make a uniform API for Metric, but it doesn't actually exist, unless you do something like *args. Mypy is just going to complain.

Do you mean when calling `metric(logits, label)`?

bryant1410:
Mypy doesn't after adding the `= None` to `mask` because `metric(logits, label)` meets the contract, while some subclasses break it.

What if I do `metric(predictions=..., gold_labels=...)` in a try-except, and say to the user that it supports metrics that only support that in case of exception?

An option that goes further (non-exclusive with the previous one): change `__call__` to receives `**kwargs` and state in the docstrings the typical behavior, asking the subclasses to follow a common naming (`predicitons` and `gold_labels`)?

DeNeutoy:
I'm less sure whether these changes are fine - you cannot pass any metric, so the type annotation as it stands is not correct. Why do you need to parametrise this? This is not a pattern we want people using our models to replicate, because they will get confused about which metrics can be passed to which models. I don't think we should accept this change, the Metrics are actually a great part of the library already, and their usage has never raised any questions.

DeNeutoy:
> What if I do `metric(predictions=..., gold_labels=...)` in a try-except, and say to the user that it supports metrics that only support that in case of exception?
> 
Instead, you should just have fixed metrics in your model

> An option that goes further (non-exclusive with the previous one): change `__call__` to receive `**kwargs` and state in the docstrings the typical behavior, asking the subclasses to follow a common naming (`predicitons` and `gold_labels`)?

But in the case you pass the wrong arguments in kwargs, you will definitely be also passing the wrong arguments as args.

matt-gardner:
@DeNeutoy, additional context [here](https://github.com/allenai/allennlp/issues/3824#issuecomment-589710136) (in the whole issue, but particularly that comment).  And yes, I agree with Mark that the model changes are too general.  I think it's fine to make it accept more metrics, but if we are going to do that, it should be more targeted than this.  As I said at the bottom of the referenced comment, the classification model could "have a list of allowable metrics, and a configuration option that allows a user to select them".  Allowing any metric like this is just asking for errors.

matt-gardner:
And in general, if you want to make your model general enough to allow taking arbitrary metrics, it's really easy to do - just change the constructor arguments, as you do here.  The question is what our example models in the library should do.  And the way they are is probably good enough, though I could be convinced to allow, e.g., F1 or other things as options.

matt-gardner:
But even then, we really want to be encouraging people to make their own models, not just use the ones we have out of the box.  Copying our model code and modifying it is just fine, and a major intended usage.  And so, given that, just having a comment in the original model saying that you can change this to F1 or something else if that's what you want in your model is maybe even good enough by itself.

bryant1410:
I still think the underlying problem is that models are coupled with metrics, as discussed in #3824. They have certain inputs (e.g., `tokens` and `label`) and outputs (e.g., `logits`, `probs` and `loss`), but they also have metrics, instead of the metrics being computed from their outputs. But, anyway.

Okay, so I change it to instead just accept F1 as well as Accuracy? With options for the beta value and the type of average?

matt-gardner:
So, with talking about beta and average parameters, it's starting to get very messy, which makes me agree even more with @DeNeutoy that we don't need this change.  There's no reason for the example model we have in here to be a super-flexible, I-can-do-everything classifier.  That's probably not what any user would want to be writing, and having that be our example would scare people off as too confusing.

bryant1410:
So maybe just a comment saying to copy and paste if you want flexibility?

matt-gardner:
If we go that route, then we should basically be annotating all kinds of things in that example code.  That might be better suited to a template project, and leave the fact that you can modify stuff implicit in the stuff in here.

PR #3848, opened 2020-02-25
Add coarse-to-fine and second order inference for coref. Also a new coref config

ZhaofengWu:
Based on https://arxiv.org/pdf/1804.05392.pdf. 79.17 dev 78.87 test. URL: https://storage.googleapis.com/allennlp-public-models/coref-spanbert-large-2020.02.27.tar.gz

SpanBERT reports 79.6 on test. According to the e2e-coref paper, speaker and genre metadata, which we aren't using, contributes 1.4 F1. So I assume that's where the difference is coming from.

### DIFF
--- a/allennlp/models/coreference_resolution/coref.py
+++ b/allennlp/models/coreference_resolution/coref.py
@@ -177,7 +177,6 @@ class CoreferenceResolver(Model):
         # Shape: (batch_size, document_length, embedding_size)
         text_embeddings = self._lexical_dropout(self._text_field_embedder(text))
 
-        device = util.get_device_of(spans)
         batch_size = spans.size(0)
         document_length = text_embeddings.size(1)
         num_spans = spans.size(1)
--- a/allennlp/models/coreference_resolution/coref.py
+++ b/allennlp/models/coreference_resolution/coref.py
@@ -9,7 +9,7 @@ from overrides import overrides
 from allennlp.data import TextFieldTensors, Vocabulary
 from allennlp.models.model import Model
 from allennlp.modules.token_embedders import Embedding
-from allennlp.modules import FeedForward
+from allennlp.modules import FeedForward, GatedSum
 from allennlp.modules import Seq2SeqEncoder, TimeDistributed, TextFieldEmbedder
 from allennlp.modules.span_extractors import SelfAttentiveSpanExtractor, EndpointSpanExtractor
 from allennlp.nn import util, InitializerApplicator
@@ -21,10 +21,10 @@ logger = logging.getLogger(__name__)
 @Model.register("coref")
 class CoreferenceResolver(Model):
     """
-    This `Model` implements the coreference resolution model described "End-to-end Neural
-    Coreference Resolution"
-    <https://www.semanticscholar.org/paper/End-to-end-Neural-Coreference-Resolution-Lee-He/3f2114893dc44eacac951f148fbff142ca200e83>
-    by Lee et al., 2017.
+    This `Model` implements the coreference resolution model described
+    "Higher-order Coreference Resolution with Coarse-to-fine Inference"
+    <https://arxiv.org/pdf/1804.05392.pdf>
+    by Lee et al., 2018.
     The basic outline of this model is to get an embedded representation of each span in the
     document. These span representations are scored and used to prune away spans that are unlikely
     to occur in a coreference cluster. For the remaining spans, the model decides which antecedent
@@ -53,6 +53,10 @@ class CoreferenceResolver(Model):
         spans we retain with respect to the number of words in the document.
     max_antecedents: int, required.
         For each mention which survives the pruning stage, we consider this many antecedents.
+    coarse_to_fine: bool, optional (default = False)
+        Whether or not to apply the coarse-to-fine filtering.
+    num_coref_layers: int, optional (default = 1)
+        The number of span representation refinement layers.
     lexical_dropout : `int`
         The probability of dropping out dimensions of the embedded text.
     initializer : `InitializerApplicator`, optional (default=`InitializerApplicator()`)
@@ -71,6 +75,7 @@ class CoreferenceResolver(Model):
         spans_per_word: float,
         max_antecedents: int,
         coarse_to_fine: bool = False,
+        num_coref_layers: int = 1,
         lexical_dropout: float = 0.2,
         initializer: InitializerApplicator = InitializerApplicator(),
         **kwargs
@@ -110,8 +115,11 @@ class CoreferenceResolver(Model):
         self._max_antecedents = max_antecedents
 
         self._coarse_to_fine = coarse_to_fine
-        if self._coarse_to_fine:  # TODO: make this a blinear module
+        if self._coarse_to_fine:
             self._coarse2fine_scorer = torch.nn.Linear(mention_feedforward.get_input_dim(), mention_feedforward.get_input_dim())
+        self._num_coref_layers = num_coref_layers
+        if self._num_coref_layers > 1:
+            self._span_updating_gated_sum = GatedSum(mention_feedforward.get_input_dim())
 
         self._mention_recall = MentionRecall()
         self._conll_coref_scores = ConllCorefScores()
@@ -260,11 +268,6 @@ class CoreferenceResolver(Model):
                 valid_antecedent_offsets.unsqueeze(0).expand(batch_size, num_spans_to_keep, num_spans_to_keep).reshape(batch_size * num_spans_to_keep, num_spans_to_keep, 1),
                 top_antecedent_indices.view(-1, max_antecedents),
             ).reshape(batch_size, num_spans_to_keep, max_antecedents)
-
-            # Compute antecedent scores.
-            top_antecedent_embeddings = util.batched_index_select(
-                top_span_embeddings, top_antecedent_indices
-            )
         else:
             # Now that we have our variables in terms of num_spans_to_keep, we need to
             # compare span pairs to decide each span's antecedent. Each span can only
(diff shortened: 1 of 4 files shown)


allennlp/models/coreference_resolution/coref.py
```diff
@@ -21,10 +21,10 @@
 @Model.register("coref")
 class CoreferenceResolver(Model):
     """
-    This `Model` implements the coreference resolution model described "End-to-end Neural
-    Coreference Resolution"
-    <https://www.semanticscholar.org/paper/End-to-end-Neural-Coreference-Resolution-Lee-He/3f2114893dc44eacac951f148fbff142ca200e83>
-    by Lee et al., 2017.
+    This `Model` implements the coreference resolution model described in
+    "Higher-order Coreference Resolution with Coarse-to-fine Inference"
```
DeNeutoy: Can you use a markdown link?

allennlp/models/coreference_resolution/coref.py
```diff
@@ -53,6 +53,10 @@ class CoreferenceResolver(Model):
         spans we retain with respect to the number of words in the document.
     max_antecedents: int, required.
         For each mention which survives the pruning stage, we consider this many antecedents.
+    coarse_to_fine: bool, optional (default = False)
+        Whether or not to apply the coarse-to-fine filtering.
+    num_coref_layers: int, optional (default = 1)
```
DeNeutoy: `num_span_refinement_layers` might be a better name, given your description 😄

allennlp/models/coreference_resolution/coref.py
```diff
@@ -472,15 +525,185 @@ def _generate_valid_antecedents(
...
+        return valid_antecedent_indices, valid_antecedent_offsets, valid_antecedent_mask
+
+    def _distance_pruning(
+        self,
+        top_span_embeddings: torch.FloatTensor,
+        top_span_mention_scores: torch.FloatTensor,
+        max_antecedents: int,
+    ) -> Tuple[torch.FloatTensor, torch.BoolTensor, torch.LongTensor, torch.LongTensor]:
+        """
+        Generates antecedents for each span and prunes down to `max_antecedents`.
```
DeNeutoy: Can you add a more comprehensive description of the pruning methods here and below?

allennlp/models/coreference_resolution/coref.py
```diff
@@ -472,15 +525,185 @@ def _generate_valid_antecedents(
...
+        """
+        batch_size, num_spans_to_keep = top_span_embeddings.size()[:2]
+        device = util.get_device_of(top_span_embeddings)
+
+        # Shape: (1, num_spans_to_keep, num_spans_to_keep)
+        _, _, valid_antecedent_mask = self._generate_valid_antecedents(
+            num_spans_to_keep, num_spans_to_keep, device
+        )
+
+        # Shape: (batch_size, num_spans_to_keep, num_spans_to_keep); broadcast op
```
DeNeutoy: Can you make this section a bit easier to read? I might write it like:
```
	mention_one_score = top_span_mention_scores.unsqueeze(0)
	mention_two_score = top_span_mention_scores.unsqueeze(1)
	coarse_to_fine_weights = self._coarse2fine_scorer(top_span_embeddings).transpose(1, 2)
	coarse_to_fine_scores = torch.matmul(top_span_embeddings, coarse_to_fine_weights)		
	
	partial_antecedent_scores = coarse_to_fine_scores + mention_one_score + mention_two_score
```

allennlp/models/coreference_resolution/coref.py
```diff
@@ -472,15 +525,185 @@ def _generate_valid_antecedents(
...
+        ) + top_span_mention_scores.unsqueeze(2)
+        partial_antecedent_scores += torch.matmul(
+            top_span_embeddings, self._coarse2fine_scorer(top_span_embeddings).transpose(1, 2)
+        )
+
+        # Shape: (batch_size, num_spans_to_keep, num_spans_to_keep); broadcast op
+        span_pair_mask = top_span_mask.unsqueeze(-1) & valid_antecedent_mask
+
+        # Shape:
+        # (batch_size, num_spans_to_keep, max_antecedents) * 3
```
DeNeutoy: I was confused by your `* 3`, can you just say "all of size: "

allennlp/models/coreference_resolution/coref.py
```diff
@@ -505,7 +728,7 @@ def _compute_span_pair_embeddings(
             (batch_size, num_spans_to_keep, max_antecedents, embedding_size).
         antecedent_offsets : `torch.IntTensor`, required.
             The offsets between each top span and its antecedent spans in terms
-            of spans we are considering. Has shape (1, max_antecedents).
+            of spans we are considering. Has shape (batch_size, num_spans_to_keep, max_antecedents).
```
DeNeutoy: I don't understand this change - was this comment wrong? Why do we need batch size here again? Is it because the span pair embeddings are no longer symmetric or something?

allennlp/models/coreference_resolution/coref.py
```diff
@@ -605,17 +815,23 @@ def _compute_coreference_scores(
...
-            (batch_size, num_spans_to_keep, max_antecedents, encoding_dim)
-        top_span_mention_scores : `torch.FloatTensor`, required.
-            Mention scores for every span. Has shape
+        top_span_embeddings : `torch.FloatTensor`, required.
+            Embedding representations of the kept spans. Has shape
+            (batch_size, num_spans_to_keep, embedding_size)
+        top_antecedent_embeddings: `torch.FloatTensor`, required.
+            The embeddings of antecedents for each span candidate. Has shape
+            (batch_size, num_spans_to_keep, max_antecedents, embedding_size)
+        top_partial_antecedents_scores : `torch.FloatTensor`, required.
```
DeNeutoy: Can you explain why this is partial, it's not clear right now what it is missing

allennlp/tests/models/coreference_resolution/coref_test.py
```diff
@@ -12,7 +12,24 @@ def setUp(self):
         )
 
     def test_coref_model_can_train_save_and_load(self):
-        self.ensure_model_can_train_save_and_load(self.param_file)
+        for coarse_to_fine in (True, False):
+            for num_coref_layes in (1, 3):
+                self._test_coref_model_can_train_save_and_load(coarse_to_fine, num_coref_layes)
```
DeNeutoy: \s\num_coref_layes\num_coref_layers

allennlp/models/coreference_resolution/coref.py
```diff
@@ -361,12 +412,14 @@ def make_output_human_readable(self, output_dict: Dict[str, torch.Tensor]):
         # A tensor of shape (num_spans_to_keep, max_antecedents), representing the indices
         # of the predicted antecedents with respect to the 2nd dimension of `batch_top_spans`
         # for each antecedent we considered.
-        antecedent_indices = output_dict["antecedent_indices"].detach().cpu()
+        batch_antecedent_indices = output_dict["antecedent_indices"].detach().cpu()
```
DeNeutoy: Ah, I think I understand why we need the batch dim - you want to have different possible numbers of antecedent indices per batch?

training_config/coref_spanbert_large.jsonnet
```diff
@@ -0,0 +1,109 @@
...
+    "antecedent_feedforward": {
+        "input_dim": span_pair_embedding_dim,
+        "num_layers": 2,
+        "hidden_dims": 1500,
+        "activations": "relu",
+        "dropout": 0.3
+    },
+    "initializer": {
+      "regexes": [
+        [".*linear_layers.*weight", {"type": "xavier_normal"}],
```
DeNeutoy: You may want to initialize the gated sum module for the refinement, just as an FYI.

ZhaofengWu:
@DeNeutoy When `max_antecedents` is 100, the delta with respect to coarse-to-fine is 2.24 F1 on dev. We are using small `max_antecedents` values due to the large models using more memory, so the difference is a lot more significant than in the original paper. I had to tune `max_antecedents` further down to 50 to avoid OOM when incorporating 2nd order inference, but that's fine because 50 was also the value used by the SpanBERT people. I didn't rerun the experiments, but I image coarse-to-fine helps even more at 50 antecedents.

allennlp/models/coreference_resolution/coref.py
ZhaofengWu: Yes, for the non-coarse-to-fine case, we always select the antecedents with the same offsets, i.e. `[max_antecedents, 0)`. But the coarse-to-fine case allows more flexibility so that each span can select arbitrary antecedents with a scoring function.

DeNeutoy:
Awesome that's great 👍 Consider most of my points above superficial in that case/just questions.

allennlp/models/coreference_resolution/coref.py
ZhaofengWu: Not just different numbers of antecedent indices, but really different values of antecedent indices, due to the flexibility of the coarse-to-fine scorer.

allennlp/models/coreference_resolution/coref.py
ZhaofengWu: Yeah my description wasn't 100% accurate either. Updated both.

allennlp/models/coreference_resolution/coref.py
ZhaofengWu: Explained in the doc.

rahular:
Hi @ZhaofengWu, thank you for the awesome work! I want to know which GPU(s) you use to train the bert coref models? A single 12gb GPU is running OOM.

ZhaofengWu:
@rahular I used a 48GB GPU. You can adjust [the config file](https://github.com/allenai/allennlp/blob/master/training_config/coref_spanbert_large.jsonnet) to your GPU spec accordingly, e.g. by using the SpanBERT base model.

PR #3851, opened 2020-02-26
Make most metrics work on GPU

### LINKED ISSUE #3017: Why tensors are moved to CPU when calculating metrics?

bryant1410:
Fix #3017.

I did the following: renamed `unwrap_to_tensors` to `detach_tensors`, and replaced it to send to `cuda()` instead of `cpu()`. Then ran all the tests with 2 GPUs (only 4 tests were skipped). It failed on:

* spearman correlation
* fbeta
* entropy
* boolean accuracy
* bleu
* auc
* allennlp/tests/predictors/srl_test.py:82 (TestSrlPredictor.test_prediction_with_no_verbs)
* allennlp/tests/predictors/coref_test.py:86 (TestCorefPredictor.test_replace_corefs)
* simpleseq2seqtest
* allennlp/tests/models/encoder_decoders/copynet_seq2seq_test.py:15 (CopyNetTest.test_model_can_train_save_load_predict)
* coreftest
* graphparsertest
* allennlp/tests/interpret/simple_gradient_test.py:30 (TestSimpleGradient.test_simple_gradient_coref)

Then went through all `detach_tensors()` callers and all metrics, and check they are compatible with GPU usage, and calling `.cpu()` if necessary.

The following should work fine on GPU:

* [x] attachment score
* [ ] auc - Should work fine for `__call__` but `get_metric` needs CPU because it uses SciPy stuff.
* [x] average
* [x] bleu
* [x] boolean accuracy
* [x] categorial accuracy
* [ ] CoNLL Coref Scores - uses SciPy stuff, so I just converted to CPU.
* [x] covariance
* [x] entropy
* [ ] evalb - doesn't use tensors, not even does `import torch`.
* [x] f1
* [x] fbeta
* [x] MAE
* [ ] mention recall - the only input tensor is converted to a list right away, so it's all CPU in the end.
* [x] pearson
* [x] perplexity
* [x] seq accuracy
* [x] span based
* [ ] spearman - Should work fine for `__call__` but `get_metric` needs CPU because it uses SciPy stuff.
* [ ] srl eval - doesn't use tensors, not even does `import torch`.
* [x] unigram recall

Then I ran the tests again, and fix one test that does `.numpy()` without doing `.cpu()` first. Then I ran all the tests again (2 GPUs; only 4 skipped tests) and it works fine. Then I removed the `.cuda()` call from `detach_tensors` and ran the tests again (still with 2 GPUs available; only 4 skipped tests). They were successful. Feel free to try it yourselves because the CI won't do (multi-)GPU testing (remember to change `detach_tensors` so it does `.cuda()`).

(Note there are 4 unconditionally skipped tests.)

I realized all metrics have tests except for `Average`, `Perplexity`, and `MentionRecall`. However, if you look at what I changed on those files you'll realize is really minor and it should still work. Note I didn't modify `Perplexity` but it subclasses `Average`. It'd be good to still have tests for those though.

Btw, I saw a bunch of `nominator / (denominator + 1e-13)` (and other similar values), which makes me think those won't work well on FP16. I believe those ideally should be `eps` args that could be changed if using FP16 for example.


allennlp/training/metrics/bleu.py
```diff
@@ -108,7 +108,7 @@ def _get_brevity_penalty(self) -> float:
         return math.exp(1.0 - self._reference_lengths / self._prediction_lengths)
 
     def _get_valid_tokens_mask(self, tensor: torch.LongTensor) -> torch.ByteTensor:
-        valid_tokens_mask = torch.ones(tensor.size(), dtype=torch.bool)
+        valid_tokens_mask = torch.ones_like(tensor, dtype=torch.bool)
```
bryant1410: Using `_like` uses the same device as the input tensor also.

allennlp/training/metrics/conll_coref_scores.py
```diff
@@ -150,29 +153,21 @@ def update(self, predicted, gold, mention_to_predicted, mention_to_gold):
...
-            if self.recall_denominator == 0
-            else self.recall_numerator / float(self.recall_denominator)
-        )
+        precision = self.get_precision()
+        recall = self.get_recall()
         return 0 if precision + recall == 0 else 2 * precision * recall / (precision + recall)
 
     def get_recall(self):
-        if self.recall_numerator == 0:
+        if self.recall_denominator == 0:
```
bryant1410: This and the next one were bugs IIRC.

allennlp/training/metrics/conll_coref_scores.py
```diff
@@ -150,29 +153,21 @@ def update(self, predicted, gold, mention_to_predicted, mention_to_gold):
...
             return 0
         else:
-            return self.recall_numerator / float(self.recall_denominator)
+            return self.recall_numerator / self.recall_denominator
```
bryant1410: Not necessary anymore in Python 3 these float castings. I tried checking if they were tensors or not before removing them so as not to remove them if they were.

allennlp/training/metrics/mention_recall.py
```diff
@@ -18,7 +18,7 @@ def __call__(
         batched_top_spans: torch.Tensor,
         batched_metadata: List[Dict[str, Any]],
     ):
-        for top_spans, metadata in zip(batched_top_spans.data.tolist(), batched_metadata):
+        for top_spans, metadata in zip(batched_top_spans.tolist(), batched_metadata):
```
bryant1410: I think in previous PyTorch versions this was necessary, but not anymore.

allennlp/training/metrics/spearman_correlation.py
```diff
@@ -40,13 +40,16 @@ def __call__(
...
-        predictions, gold_labels, mask = self.unwrap_to_tensors(predictions, gold_labels, mask)
-        # Flatten predictions, gold_labels, and mask. We calculate the spearman correlation between
+        predictions, gold_labels, mask = self.detach_tensors(predictions, gold_labels, mask)
+        # Flatten predictions, gold_labels, and mask. We calculate the Spearman correlation between
         # the vectors, since each element in the predictions and gold_labels tensor is assumed
         # to be a separate observation.
         predictions = predictions.view(-1)
         gold_labels = gold_labels.view(-1)
 
+        self.total_predictions = self.total_predictions.to(predictions.device)
```
bryant1410: Note that at initialization time we don't know the device we should use, but here we move it. If it was already in that device, it's a no-op so it's fine.

bryant1410:
> One question - we have GPU tests which we run occasionally in our CI - is it easy to parametrise all of the metrics tests in a conditional way such that they run on the cpu when available, and cpu and gpu when both are?

The approach I can come up with is using [`pytest.param`](https://docs.pytest.org/en/latest/reference.html#pytest-param) for every test function that should support both GPU and CPU, like (I guess it works with `self`):

```python
@pytest.mark.parametrize("device", [
    torch.cpu,
    pytest.param(
        torch.cuda,
        marks=pytest.mark.skipif(not torch.cuda.is_available(), reason="requires cuda")
    ),
])
def test_func(self, device):
    ...
```

Does it make sense?

bryant1410:
> The approach I can come up with is using [`pytest.param`](https://docs.pytest.org/en/latest/reference.html#pytest-param) for every test function that should support both GPU and CPU, like (I guess it works with `self`):
> 
> ```python
> @pytest.mark.parametrize("device", [
>     torch.cpu,
>     pytest.param(
>         torch.cuda,
>         marks=pytest.mark.skipif(not torch.cuda.is_available(), reason="requires cuda")
>     ),
> ])
> def test_func(self, device):
>     ...
> ```
> 
> Does it make sense?

[In jax they do something similar.](https://github.com/google/jax/blob/master/tests/multibackend_test.py)

DeNeutoy:
Nice, is there a way to make that decorator itself a decorator? It's a bit verbose and it would be nice to be able to do:

```
@multi_device
def test_func(self, device):
```

This might not be possible with pytest, I know it's a bit finicky about how it uses those decorators.

allennlp/tests/training/metrics/span_based_f1_measure_test.py
```diff
@@ -235,24 +244,24 @@ def test_span_f1_matches_perl_script_for_continued_arguments(self):
...
+        assert_allclose(metric_dict["precision-overall"], 1.0)
+        assert_allclose(metric_dict["f1-measure-overall"], 1.0)
 
         # Check that the number of true positive ARG1 labels is the same as the perl script's output:
         gold_file_path = os.path.join(self.TEST_DIR, "gold_conll_eval.txt")
         prediction_file_path = os.path.join(self.TEST_DIR, "prediction_conll_eval.txt")
-        with open(gold_file_path, "a+") as gold_file, open(
-            prediction_file_path, "a+"
+        with open(gold_file_path, "w") as gold_file, open(
+            prediction_file_path, "w"
```
bryant1410: I changed this because the test failed (function called twice), and also the append wasn't necessary.

bryant1410:
I added that utility. I discovered that pytest and unittest don't go along well together, and it's hard to parametrize in that context with the utils they provide.

I had to change the test to actually use the device. That implied to avoid using NumPy if I could as well. In the end, `torch` provides a `testing` module with `assert_allclose` that's convenient. It has good defaults, and when you specify `rtol` you also have to specify `atol` (either both or none), don't know why. I had also to change some `FloatTensor` creations and similar ones for `tensor`, because of some exceptions (and I saw they are recommended to not use anymore).

I think as a general practice, we should specify the device when we create new tensors with `ones`, `zeros`, `randn`, `rand`, and `tensor` (or any constructor; not with like `ones_like` because the device is copied).

allennlp/tests/training/metrics/fbeta_measure_test.py
```diff
@@ -77,17 +87,21 @@ def test_fbeta_multiclass_metric(self):
         fscores = metric["fscore"]
 
         # check value
-        numpy.testing.assert_almost_equal(precisions, self.desired_precisions, decimal=2)
-        numpy.testing.assert_almost_equal(recalls, self.desired_recalls, decimal=2)
-        numpy.testing.assert_almost_equal(fscores, self.desired_fscores, decimal=2)
+        assert_allclose(precisions, self.desired_precisions)
+        assert_allclose(recalls, self.desired_recalls)
+        assert_allclose(fscores, self.desired_fscores)
```
bryant1410: In many of these changes, the expected precision is actually stronger. It works because I changed `0.33` as the expected value to `1 / 3`.

allennlp/training/metrics/spearman_correlation.py
```diff
@@ -44,14 +44,14 @@ def __call__(
         # Flatten predictions, gold_labels, and mask. We calculate the Spearman correlation between
         # the vectors, since each element in the predictions and gold_labels tensor is assumed
         # to be a separate observation.
-        predictions = predictions.view(-1)
-        gold_labels = gold_labels.view(-1)
+        predictions = predictions.reshape(-1)
```
bryant1410: For some reason, on GPU sometimes this `view` fails. We can use `reshape` anyway, that tries to do a `view` but if it can't will reshape.

allennlp/tests/training/metrics/span_based_f1_measure_test.py
```diff
@@ -198,24 +203,28 @@ def test_bmes_span_metrics_are_computed_correctly(self):
...
 
-    def test_span_f1_can_build_from_params(self):
+    @multi_device
+    def test_span_f1_can_build_from_params(self, device: str):
         params = Params({"type": "span_f1", "tag_namespace": "tags", "ignore_classes": ["V"]})
         metric = Metric.from_params(params=params, vocabulary=self.vocab)
-        assert metric._ignore_classes == ["V"]
-        assert metric._label_vocabulary == self.vocab.get_index_to_token_vocabulary("tags")
+        assert metric._ignore_classes == ["V"]  # type: ignore
+        assert metric._label_vocabulary == self.vocab.get_index_to_token_vocabulary(  # type: ignore
```
bryant1410: Idk why this mypy errors didn't appear before.

allennlp/common/testing/test_case.py
```diff
@@ -40,3 +43,34 @@ def setUp(self):
...
+    """
+    Decorator to create parameterized tests.
+
+    # Parameters
+
+    arg_names : `Iterable[str]`, required.
+        Argument names to pass to the test function.
+    arg_values : `Iterable[Iterable[Any]]`, required.
+        Iterable of values to pass to each of the args.
+        A function call is gonna be made for each inner iterable.
```
DeNeutoy:
```suggestion
		The decorated test will be run for each inner iterable.
```

allennlp/common/testing/test_case.py
```diff
@@ -40,3 +43,34 @@ def setUp(self):
 
     def tearDown(self):
         shutil.rmtree(self.TEST_DIR)
+
+
+def parametrize(arg_names: Iterable[str], arg_values: Iterable[Iterable[Any]]):
```
DeNeutoy: Very cute, this is really nice!

allennlp/tests/common/testing.py
```diff
@@ -0,0 +1,19 @@
+import torch
+
+from allennlp.common.testing import AllenNlpTestCase, multi_device
+
+
+class TestFromParams(AllenNlpTestCase):
```
matt-gardner: Class name here needs updating.

(I came to see what Mark thought looked cute, noticed a copy-paste bug.)

bryant1410: Thanks! Good catch.

bryant1410: I put `TestTesting` after the module name, hope it's fine.

bryant1410:
@DeNeutoy related to this and to be on the safe side, in `Trainer`, before calling `get_metrics()`, shouldn't we do `with torch.no_grad():` (in training; in validation it's already there)?

PR #3855, opened 2020-02-26
Fixes XLNet, and adds a test for it

### LINKED ISSUE #3795: Type ID heuristics failing on XLNet
The `PretrainedTransformerEmbedder` class currently checks the maximum type ID is no greater than what is specified in the config which is determined heuristically, checking for the presence of the `"type_vocab_size"` key in the config. For XLNet, this key is absent while the maximum type ID is 2 (https://github.com/huggingface/transformers/blob/0dbddba6d2c5b2c6fc08866358c1994a00d6a1ff/src/transformers/tokenization_xlnet.py#L229), failing the check https://github.com/allenai/allennlp/blob/8532dc5d906e29e055f36b7d76ec0a84453181f6/allennlp/modules/token_embedders/pretrained_transformer_embedder.py#L95-L96

A simple fix would be to remove this check.

dirkgr:
Fixes #3795


dirkgr:
`create_token_type_ids_from_sequences` is sometimes wrong (for example, in the case of DistilBERT, https://github.com/huggingface/transformers/issues/2702). We'd just have a different set of special cases.

PR #3865, opened 2020-02-27
Add NLI models

ZhaofengWu:
SNLI: dev 92.47, test 91.65, URL https://storage.googleapis.com/allennlp-public-models/snli-roberta-large-2020.02.27.tar.gz

MNLI: dev-matched 89.10, dev-mismatched 88.71, URL https://storage.googleapis.com/allennlp-public-models/mnli-roberta-large-2020.02.27.tar.gz

### DIFF
--- a/allennlp/data/dataset_readers/snli.py
+++ b/allennlp/data/dataset_readers/snli.py
@@ -9,7 +9,7 @@ from allennlp.data.dataset_readers.dataset_reader import DatasetReader
 from allennlp.data.fields import Field, TextField, LabelField, MetadataField
 from allennlp.data.instance import Instance
 from allennlp.data.token_indexers import SingleIdTokenIndexer, TokenIndexer
-from allennlp.data.tokenizers import Tokenizer, SpacyTokenizer
+from allennlp.data.tokenizers import Tokenizer, SpacyTokenizer, PretrainedTransformerTokenizer
 
 logger = logging.getLogger(__name__)
 
@@ -68,16 +68,23 @@ class SnliReader(DatasetReader):
     ) -> Instance:
 
         fields: Dict[str, Field] = {}
-        premise_tokens = self._tokenizer.tokenize(premise)
-        hypothesis_tokens = self._tokenizer.tokenize(hypothesis)
-        fields["premise"] = TextField(premise_tokens, self._token_indexers)
-        fields["hypothesis"] = TextField(hypothesis_tokens, self._token_indexers)
+
+        if isinstance(self._tokenizer, PretrainedTransformerTokenizer):
+            tokens = self._tokenizer.tokenize_sentence_pair(premise, hypothesis)
+            fields["tokens"] = TextField(tokens, self._token_indexers)
+        else:
+            premise_tokens = self._tokenizer.tokenize(premise)
+            hypothesis_tokens = self._tokenizer.tokenize(hypothesis)
+            fields["premise"] = TextField(premise_tokens, self._token_indexers)
+            fields["hypothesis"] = TextField(hypothesis_tokens, self._token_indexers)
+
+            metadata = {
+                "premise_tokens": [x.text for x in premise_tokens],
+                "hypothesis_tokens": [x.text for x in hypothesis_tokens],
+            }
+            fields["metadata"] = MetadataField(metadata)
+
         if label:
             fields["label"] = LabelField(label)
 
-        metadata = {
-            "premise_tokens": [x.text for x in premise_tokens],
-            "hypothesis_tokens": [x.text for x in hypothesis_tokens],
-        }
-        fields["metadata"] = MetadataField(metadata)
         return Instance(fields)
--- a/allennlp/data/dataset_readers/snli.py
+++ b/allennlp/data/dataset_readers/snli.py
@@ -30,7 +30,7 @@ class SnliReader(DatasetReader):
     token_indexers : `Dict[str, TokenIndexer]`, optional (default=`{"tokens": SingleIdTokenIndexer()}`)
         We similarly use this for both the premise and the hypothesis.  See :class:`TokenIndexer`.
     combine_input_fields : `bool`, optional
-            (default=`not isinstance(tokenizer, PretrainedTransformerTokenizer)`)
+            (default=`isinstance(tokenizer, PretrainedTransformerTokenizer)`)
         If False, represent the premise and the hypothesis as separate fields in the instance.
         If True, tokenize them together using `tokenizer.tokenize_sentence_pair()`
         and provide a single `tokens` field in the instance.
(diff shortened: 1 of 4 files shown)


allennlp/data/dataset_readers/snli.py
```diff
@@ -68,16 +68,23 @@ def text_to_instance(
     ) -> Instance:
 
         fields: Dict[str, Field] = {}
-        premise_tokens = self._tokenizer.tokenize(premise)
-        hypothesis_tokens = self._tokenizer.tokenize(hypothesis)
-        fields["premise"] = TextField(premise_tokens, self._token_indexers)
-        fields["hypothesis"] = TextField(hypothesis_tokens, self._token_indexers)
+
+        if isinstance(self._tokenizer, PretrainedTransformerTokenizer):
```
matt-gardner: Slight preference for having this be a configuration option instead of relying on this type check.

allennlp/data/dataset_readers/snli.py
```diff
@@ -29,14 +29,29 @@ class SnliReader(DatasetReader):
         We use this `Tokenizer` for both the premise and the hypothesis.  See :class:`Tokenizer`.
     token_indexers : `Dict[str, TokenIndexer]`, optional (default=`{"tokens": SingleIdTokenIndexer()}`)
         We similarly use this for both the premise and the hypothesis.  See :class:`TokenIndexer`.
+    tokenize_separately : `bool`, optional
```
matt-gardner: I like having the default be a best guess like this, though the name is a bit confusing to me.  Maybe `combine_fields` or `use_single_field`, or something, with the default being `False`?

matt-gardner: The issue is this isn't necessarily about tokenization, it's more about how the data is represented.

ZhaofengWu: Do you mean the default being `None`? Because if it's `False` we can no longer take a guess.

matt-gardner: Right, sorry, I guess I just meant negating the logic in the guess.

allennlp/data/dataset_readers/snli.py
```diff
@@ -29,14 +29,27 @@ class SnliReader(DatasetReader):
         We use this `Tokenizer` for both the premise and the hypothesis.  See :class:`Tokenizer`.
     token_indexers : `Dict[str, TokenIndexer]`, optional (default=`{"tokens": SingleIdTokenIndexer()}`)
         We similarly use this for both the premise and the hypothesis.  See :class:`TokenIndexer`.
+    combine_input_fields : `bool`, optional
+            (default=`not isinstance(tokenizer, PretrainedTransformerTokenizer)`)
```
matt-gardner: This line is now wrong.  
```suggestion
            (default=`isinstance(tokenizer, PretrainedTransformerTokenizer)`)
```
After this, looks great!

training_config/mnli_roberta.jsonnet
```diff
@@ -0,0 +1,66 @@
...
+    },
+    "feedforward": {
+      "input_dim": transformer_dim,
+      "num_layers": 1,
+      "hidden_dims": transformer_dim,
+      "activations": "tanh"
+    },
+    "dropout": 0.1
+  },
+  "iterator": {
```
matt-gardner: Sorry I didn't notice this earlier - this needs to be updated to match #3700.  We can probably also just remove the `sorting_keys` key here.

training_config/mnli_roberta.jsonnet
```diff
@@ -0,0 +1,66 @@
...
+          "max_length": 512
+        }
+      }
+    },
+    "seq2vec_encoder": {
+       "type": "cls_pooler",
+       "embedding_dim": transformer_dim,
+       "cls_is_last_token": cls_is_last_token
+    },
+    "feedforward": {
```
matt-gardner: I was just checking Anthony's results versus the numbers you're reporting.  You're about 1.5 points below Anthony's MNLI result, and one point below the published number.  I looked at Anthony's code, and it looks like the only difference is that he doesn't have this feedforward layer in there.  What happens if you remove this?

training_config/mnli_roberta.jsonnet
```diff
@@ -0,0 +1,66 @@
...
+  "trainer": {
+    "num_epochs": 10,
+    "cuda_device" : 0,
+    "validation_metric": "+accuracy",
+    "learning_rate_scheduler": {
+      "type": "slanted_triangular",
+      "cut_frac": 0.06
+    },
+    "optimizer": {
+      "type": "huggingface_adamw",
```
matt-gardner: This is another difference with Anthony's code, though I'm not sure how much difference it makes: https://github.com/anthonywchen/generative-nli/blob/b81d4e3e7f635fcc1e5ca09305707e5df983ec88/configs/roberta_large_config.json#L37-L43

training_config/mnli_roberta.jsonnet
ZhaofengWu: Ah, thanks for the catch. Will update this (and a new coref config as well).

training_config/mnli_roberta.jsonnet
ZhaofengWu: RoBERTa actually has this feedforward layer (using the same model as BERT) https://github.com/huggingface/transformers/blob/908fa43b543cf52a3238129624f502240725a6a6/src/transformers/modeling_bert.py#L426-L438

ZhaofengWu: And I tried w/ vs. w/o this on SST -- didn't make a difference.

training_config/mnli_roberta.jsonnet
ZhaofengWu: Weight decay is different but I took the value from the RoBERTa paper (https://arxiv.org/pdf/1907.11692.pdf, last page last table last column). I doubt 1e-5 vs. 2e-5 LR is making 1.5 difference in F1? Anthony also didn't regularize bias which is probably the right thing to do.

anthonywchen: Yeah in experiments I tried, the differences in different learning rates are basically negligible.

anthonywchen: One other difference I see is the betas in ADAMW optimizer. I set mine to be [0.9, 0.98], which is recommended from their paper. 

> Similarly, we found setting β2 = 0.98 to improve stability when training with large batch sizes.

anthonywchen: One other difference is in the beta parameters in the adamw optimizer. I set mine to be [0.9, 0.98], which is recommended on their examples.

https://github.com/pytorch/fairseq/blob/master/examples/roberta/README.glue.md

Issue #2149, opened 2018-12-06
Mixed Precision Training


matthew-z:
It seems that currently there is no fp16 support in AllenNLP, but I think it is quite straightforward to do in a clean way with [Apex](https://github.com/NVIDIA/apex).

Advantages of FP16:
1. Larger batch size, especially useful for large embedding models like ELMO and BERT
2. Faster training: the latest GPUs (P100, V100, RTX) can achieve 60-80% speed up 
3. Many experiments have showed that FP16 does not hurt accuracy for CV models.

Challenge:
Some functions in AllenNLP are not FP16 stable, such as `masked_softmax`

matt-gardner:
Thanks!  For now, this is pretty low on our priority list, but contributions would be welcome here.

matt-peters:
Would love to see fp16 support in allennlp!

Hyperparticle:
I played around with apex/amp and found that integrating it into allennlp is straightforward, with a measurable performance boost on a large CRF tagger model (RTX GPU).

After cloning and installing [apex](https://github.com/NVIDIA/apex), you only need to modify [allennlp/training/trainer.py](https://github.com/allenai/allennlp/blob/master/allennlp/training/trainer.py) (or create your own trainer).

Near the top after the imports, import amp and initialize it.

```python
from apex import amp
amp_handle = amp.init()
```

Replace the `self.optimizer = optimizer` assignment, wrapping amp in the `__init__()` function.

```python
self.optimizer = amp_handle.wrap_optimizer(optimizer)
```

Finally, replace `loss.backward()` in `_train_epoch()` with scaled loss.

```python
with self.optimizer.scale_loss(loss) as scaled_loss:
    scaled_loss.backward()
```

I tested this with a larger version of the [crf_tagger.json](https://github.com/allenai/allennlp/blob/master/tutorials/getting_started/walk_through_allennlp/crf_tagger.json) tutorial config on [conll 2003](https://www.clips.uantwerpen.be/conll2003/ner/). With such a small model, performance was actually slower with `fp16` enabled. However, once I boosted the embedding/RNN size to something like 256/4096, I could see ~17% faster training times per epoch on a Titan RTX (over 10 epochs). It's not close to 2x like I hoped, but there definitely may be optimizations in order that can improve this. I also notice a small regularization effect on this model in that `fp32` training overfits early on where `fp16` does not. 

[...]

PR #3866, opened 2020-02-27
Enable AMP with Apex

JohnGiorgi:
# Overview

This PR enables automatic mixed precision training with NVIDIA's [Apex](https://nvidia.github.io/apex/index.html) which has been much discussed in #2149 and is one of the action items on the [roadmap](https://github.com/allenai/allennlp/blob/master/ROADMAP.md).

The modifications were very straightforward. Here is the high-level:

1. Add argument `opt_level` to `Trainer`. This coincides with the [opt levels of Apex](https://nvidia.github.io/apex/amp.html#opt-levels).
2. In the constructor of `Trainer`, if `opt_level is not None`, wrap `self.model` and `self.optimizer` with `amp.initialize`.
3. `rescale_gradients` needs to be modified slightly when using Apex. See [here](https://nvidia.github.io/apex/advanced.html#gradient-clipping) for more info.
4. In `_train_epoch`, `if opt_level is not None`, call `loss.backward()` in the `amp.scale_loss` context manager.

I got all of this straight from the [Apex docs](https://nvidia.github.io/apex/index.html).

## Benchmark

I tested it on [my model](https://github.com/JohnGiorgi/t2t) and found a ~40% speedup in time per epoch during training. It also allowed me to train with a 40% larger mini-batch. The majority of my models parameters exist in a pre-trained transformer (`distilroberta-base` from [Transformers](https://huggingface.co/transformers/index.html)).

Simple benchmark on a [16GB-V100](https://www.nvidia.com/en-us/data-center/v100/) using a random 1K documents from [wikitext-103](https://blog.einstein.ai/the-wikitext-long-term-dependency-language-modeling-dataset/):

| `opt_level`   |      Batch Size      |  Time per epoch |
|----------|:-------------:|------:|
| `"O0"` |  12* | ~58s |
| `"O1"` |  12 | ~35s |
| `"O1"` |  20* | ~39s |

\*Maximum batch size that fits in memory at this `opt_level`.

Obviously, mileage will vary but this gives a flavour of the magnitude of speedup you can get with Apex.

## Things I need help with

- The way I have written this, there will be an `ImportError` if the user does not have `amp` installed. This is probably not what we want.
- I just shoehorned in the required modification to `rescale_gradients`. There is likely a cleaner way to do this.

### DIFF
--- a/allennlp/tests/training/trainer_test.py
+++ b/allennlp/tests/training/trainer_test.py
@@ -120,6 +120,21 @@ class TestTrainer(AllenNlpTestCase):
             Trainer(
                 self.model, self.optimizer, self.data_loader, num_epochs=2, cuda_device=[0, 1],
             )
+    
+    @pytest.mark.skipif(not torch.cuda.is_available(), reason="No CUDA device registered.")
+    @pytest.mark.skipif(importlib.util.find_spec("apex") is None, reason="Apex is not installed.")
+    def test_trainer_can_run_amp(self):
+        from apex import amp
+        self.model.cuda()
+        trainer = Trainer(
+            self.model, self.optimizer, self.data_loader, num_epochs=2, cuda_device=0, opt_level="O1"
+        )
+        metrics = trainer.train()
+        assert "peak_cpu_memory_MB" in metrics
+        assert isinstance(metrics["peak_cpu_memory_MB"], float)
+        assert metrics["peak_cpu_memory_MB"] > 0
+        assert "peak_gpu_0_memory_MB" in metrics
+        assert isinstance(metrics["peak_gpu_0_memory_MB"], int)
 
     def test_trainer_can_resume_training(self):
         trainer = Trainer(
--- a/allennlp/tests/training/trainer_test.py
+++ b/allennlp/tests/training/trainer_test.py
@@ -1,6 +1,5 @@
 import copy
 import glob
-import importlib
 import json
 import os
 import re
@@ -8,6 +7,11 @@ import time
 
 import math
 import pytest
+
+try:
+    from apex import amp
+except ImportError:
+    amp = None
 import torch
 from torch.utils.data import DataLoader
 
@@ -123,7 +127,7 @@ class TestTrainer(AllenNlpTestCase):
             )
 
     @pytest.mark.skipif(not torch.cuda.is_available(), reason="No CUDA device registered.")
-    @pytest.mark.skipif(importlib.util.find_spec("apex") is None, reason="Apex is not installed.")
+    @pytest.mark.skipif(amp is None, reason="Apex is not installed.")
     def test_trainer_can_run_amp(self):
 
         self.model.cuda()
@@ -135,12 +139,7 @@ class TestTrainer(AllenNlpTestCase):
             cuda_device=0,
             opt_level="O1",
         )
-        metrics = trainer.train()
-        assert "peak_cpu_memory_MB" in metrics
-        assert isinstance(metrics["peak_cpu_memory_MB"], float)
-        assert metrics["peak_cpu_memory_MB"] > 0
-        assert "peak_gpu_0_memory_MB" in metrics
-        assert isinstance(metrics["peak_gpu_0_memory_MB"], int)
+        _ = trainer.train()
 
     def test_trainer_can_resume_training(self):
         trainer = Trainer(
--- a/allennlp/tests/training/trainer_test.py
+++ b/allennlp/tests/training/trainer_test.py
@@ -121,14 +121,20 @@ class TestTrainer(AllenNlpTestCase):
             Trainer(
                 self.model, self.optimizer, self.data_loader, num_epochs=2, cuda_device=[0, 1],
             )
-    
+
     @pytest.mark.skipif(not torch.cuda.is_available(), reason="No CUDA device registered.")
     @pytest.mark.skipif(importlib.util.find_spec("apex") is None, reason="Apex is not installed.")
     def test_trainer_can_run_amp(self):
         from apex import amp
+
         self.model.cuda()
         trainer = Trainer(
-            self.model, self.optimizer, self.data_loader, num_epochs=2, cuda_device=0, opt_level="O1"
+            self.model,
+            self.optimizer,
+            self.data_loader,
+            num_epochs=2,
+            cuda_device=0,
+            opt_level="O1",
         )
         metrics = trainer.train()
         assert "peak_cpu_memory_MB" in metrics
--- a/allennlp/tests/training/trainer_test.py
+++ b/allennlp/tests/training/trainer_test.py
@@ -125,7 +125,6 @@ class TestTrainer(AllenNlpTestCase):
     @pytest.mark.skipif(not torch.cuda.is_available(), reason="No CUDA device registered.")
     @pytest.mark.skipif(importlib.util.find_spec("apex") is None, reason="Apex is not installed.")
     def test_trainer_can_run_amp(self):
-        from apex import amp
 
         self.model.cuda()
         trainer = Trainer(
--- a/allennlp/tests/training/trainer_test.py
+++ b/allennlp/tests/training/trainer_test.py
@@ -1,5 +1,6 @@
 import copy
 import glob
+import importlib
 import json
 import os
 import re
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -188,6 +188,11 @@ class Trainer(TrainerBase):
             Gradients are accumulated for the given number of steps before doing an optimizer step. This can
             be useful to accommodate batches that are larger than the RAM size. Refer Thomas Wolf's
             [post](https://tinyurl.com/y5mv44fw) for details on Gradient Accumulation.
+        opt_level : `str`, optional, (default = `None`)
+            Each opt_level establishes a set of properties that govern Amp’s implementation of pure or mixed
+            precision training. Must be a choice of `"O0"`, `"O1"`, `"O2"`, or `"O3"`.
+            See the Apex [documentation](https://nvidia.github.io/apex/amp.html#opt-levels-and-properties) for
+            more details. If `None`, Amp is not used. Defaults to `None`.
         """
         super().__init__(serialization_dir, cuda_device, distributed, local_rank, world_size)
 
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -282,11 +282,15 @@ class Trainer(TrainerBase):
         if self._opt_level is not None:
             if amp is not None:
                 raise ConfigurationError(
-                    ("Apex not installed but opt_level was provided. Please install NVIDIA's Apex to enable"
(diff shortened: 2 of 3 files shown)


allennlp/training/trainer.py
```diff
@@ -66,6 +68,7 @@ def __init__(
         local_rank: int = 0,
         world_size: int = 1,
         num_gradient_accumulation_steps: int = 1,
+        opt_level: Optional[str] = None,
```
matt-gardner: Can you add this to the docstring?

allennlp/training/trainer.py
```diff
@@ -32,6 +32,8 @@
 from allennlp.training.tensorboard_writer import TensorboardWriter
 from allennlp.training.trainer_base import TrainerBase
 
+from apex import amp
```
matt-gardner: I'm ok with just adding this to our list of dependencies, unless it's super heavy.  Though it should be above, in the same block as the `torch` imports.

Also, can you add a test with a not-`None` `opt_level`, showing that this works?

allennlp/training/trainer.py
```diff
@@ -282,7 +290,14 @@ def __init__(
...
     def rescale_gradients(self) -> Optional[float]:
-        return training_util.rescale_gradients(self.model, self._grad_norm)
+        if self._opt_level is not None:
+            if self._grad_norm:
+                # See: https://nvidia.github.io/apex/advanced.html#gradient-clipping
+                parameters_to_clip = [p for p in amp.master_params(self.optimizer) if p.grad is not None]
+                return training_util.sparse_clip_norm(parameters_to_clip, self._grad_norm)
+            return None
+        else:
+            return training_util.rescale_gradients(self.model, self._grad_norm)
```
matt-gardner: Another option here is to add a `use_amp` flag to `rescale_gradients`.  Oh, except that then also requires passing in the optimizer...  Yeah, I think what you have here is good enough.

allennlp/training/trainer.py
```diff
@@ -188,6 +188,11 @@ def __init__(
             Gradients are accumulated for the given number of steps before doing an optimizer step. This can
             be useful to accommodate batches that are larger than the RAM size. Refer Thomas Wolf's
             [post](https://tinyurl.com/y5mv44fw) for details on Gradient Accumulation.
+        opt_level : `str`, optional, (default = `None`)
+            Each opt_level establishes a set of properties that govern Amp’s implementation of pure or mixed
+            precision training. Must be a choice of `"O0"`, `"O1"`, `"O2"`, or `"O3"`.
+            See the Apex [documentation](https://nvidia.github.io/apex/amp.html#opt-levels-and-properties) for
+            more details. If `None`, Amp is not used. Defaults to `None`.
```
JohnGiorgi: @matt-gardner How's this?

allennlp/training/trainer.py
JohnGiorgi: Yes exactly my thinking. Othewise, we could have a second util function named something like `rescale_gradients_amp` which accepts an optimizer.

allennlp/training/trainer.py
JohnGiorgi: Done, added a comment below.

allennlp/training/trainer.py
JohnGiorgi: Okay, moved apex import to torch block. Put it at the top because its convention for them to be alphabetically sorted (I think?)

[Here](https://github.com/NVIDIA/apex#linux) is the install process for apex. I will leave it you as to whether this is too heavy or not!

Sure, will work on adding a test either tonight or tomorrow.

matt-gardner: I guess we can add it to our `setup.py`, and if someone wants to install their own optimized version, that's fine too?  The point is we want to not crash when someone just does `pip install allennlp`.

allennlp/training/trainer.py
matt-gardner: I think adding that function would be overkill.  If we want to refactor things, we should have `rescale_gradients` take a list of parameters.  But then we can just remove it and call `sparse_clip_norm` directly.  I'm somewhat in favor of that option, as it would simplify the above code.

allennlp/training/trainer.py
JohnGiorgi: Right, that seems perfectly reasonable to me.

allennlp/training/trainer.py
JohnGiorgi: Okay, I refactored `Trainer.rescale_gradients` to use `training_utils.sparse_clip_norm` directly. Otherwise the function is the same.

allennlp/training/trainer.py
JohnGiorgi: Also, I need a little time to get my head around how the testing works in AllenNLP. If it is very easy for one of the AllenNLP devs to add a test for when `opt_level is not None` I could also pass the buck so this can get merged sooner.

dev-requirements.txt
```diff
@@ -25,6 +25,9 @@ matplotlib>=2.2.3
 # Required to run sanic tests
 aiohttp
 
+# Required for automatic mixed precision (AMP) training
+git+https://github.com/NVIDIA/apex.git@master
+
```
JohnGiorgi: Okay, as discussed, I have added apex as a dev requirement

allennlp/tests/training/trainer_test.py
```diff
@@ -120,6 +120,21 @@ def test_passing_trainer_multiple_gpus_raises_error(self):
...
+        self.model.cuda()
+        trainer = Trainer(
+            self.model, self.optimizer, self.data_loader, num_epochs=2, cuda_device=0, opt_level="O1"
+        )
+        metrics = trainer.train()
+        assert "peak_cpu_memory_MB" in metrics
+        assert isinstance(metrics["peak_cpu_memory_MB"], float)
+        assert metrics["peak_cpu_memory_MB"] > 0
+        assert "peak_gpu_0_memory_MB" in metrics
+        assert isinstance(metrics["peak_gpu_0_memory_MB"], int)
```
JohnGiorgi: Added a test to make sure the model runs with `amp`. There are two `skipif` checks, one to make sure a cuda device is avaible and one to check if `apex` is importable.

JohnGiorgi:
@DeNeutoy Thanks a lot for the guidance. That made it easy to add a test. Also added `apex` to the dev requirements.

One thing: If I put `from apex import amp` into the `if self._opt_level is not None` check, I will get an import error in `rescale_gradients` when I try to call `amp.master_params`. Is there a way to import apex within the `if` statement such that I can reference it outside the constructor's scope?

Good catch on the checkpointing thing. I guess it is up to AllenNLP devs, I could open another PR after this is merged to get checkpointing with an amp-enabled model working. Or try to add it to this PR.

allennlp/training/trainer.py
```diff
@@ -7,6 +7,7 @@
 import traceback
 from typing import Dict, List, Optional, Tuple, Union, Any
 
+from apex import amp
```
DeNeutoy: I would suggest this:

```
try:
	from apex import amp
except ImportError:
	amp = None

```

and then below, check that `amp is not None` and raise a `ConfigurationError` if it is and the user has passed an opt level. Does that make sense?

allennlp/training/trainer.py
bryant1410: What about `amp_level`?

JohnGiorgi: Hmm, the also call this arg `opt_level` in `apex` (see [here](https://nvidia.github.io/apex/amp.html#opt-levels-and-properties)). Is there a good argument for calling it something different in AllenNLP?

bryant1410: Yeah, cause amp is the main thing in apex. As far as I understand, `opt_level` means "option level". In AllenNLP there are many options. So, `amp_level` would mean "automatic mixed precision level"

bryant1410: [Oh, it's actually "optimization level"](https://nvidia.github.io/apex/amp.html#opt-levels-and-properties). Still, I'd do `amp_opt_level`, cause there are many different options in `Trainer` and it can get confusing. What do you think?

matt-gardner: I would lean towards sticking with `opt_level` here, or perhaps `amp_opt_level`.  Switching the name of a parameter makes it harder for someone familiar with apex to know what's going on here.

bryant1410:
As mentioned in #3851, there are some things that IMHO wouldn't play well with FP16 and I have had problems in the past, such as having some `+ 1e-13`, but I guess those will be fixed later when people detect and report them, I guess when they have radically different results with and without some specific AMP `opt_level`.

allennlp/training/trainer.py
```diff
@@ -277,6 +280,12 @@ def __init__(
         # Enable automatic mixed precision training with NVIDIA Apex.
         self._opt_level = opt_level
         if self._opt_level is not None:
+            if amp is not None:
+                raise ConfigurationError(
+                    ("Apex not installed but opt_level was provided. Please install NVIDIA's Apex to enable"
+                     " automatic mixed precision (AMP) training. See: https://github.com/NVIDIA/apex.")
+                )
+
```
JohnGiorgi: As discussed!

allennlp/training/trainer.py
JohnGiorgi: Done.

allennlp/tests/training/trainer_test.py
```diff
@@ -121,6 +122,26 @@ def test_passing_trainer_multiple_gpus_raises_error(self):
...
+        trainer = Trainer(
+            self.model,
+            self.optimizer,
+            self.data_loader,
+            num_epochs=2,
+            cuda_device=0,
+            opt_level="O1",
+        )
+        metrics = trainer.train()
+        assert "peak_cpu_memory_MB" in metrics
```
DeNeutoy: You can delete these asserts, as they aren't related to this test.

allennlp/tests/training/trainer_test.py
```diff
@@ -121,6 +122,26 @@ def test_passing_trainer_multiple_gpus_raises_error(self):
                 self.model, self.optimizer, self.data_loader, num_epochs=2, cuda_device=[0, 1],
             )
 
+    @pytest.mark.skipif(not torch.cuda.is_available(), reason="No CUDA device registered.")
+    @pytest.mark.skipif(importlib.util.find_spec("apex") is None, reason="Apex is not installed.")
```
DeNeutoy: This `importlib` thing is doing something funny to our CI, can you just change it to how we imported it in the trainer itself?

```
try:
	import apex
accept:
	apex = None

...

@pytest.mark.skipif(apex is None, ...)
```

JohnGiorgi: Done!

allennlp/tests/training/trainer_test.py
JohnGiorgi: Done!

bryant1410:
I saw the GPU tests failed after this PR. Maybe there's something to review?

Can the bulldozer run the GPU tests and report if they fail?

PR #3856, opened 2020-02-26
update training configs

DeNeutoy:

PR #3867, opened 2020-02-28
Update some training configs to match Data V2

ZhaofengWu:


training_config/coref_spanbert_large.jsonnet
```diff
@@ -85,10 +85,12 @@ local span_pair_embedding_dim = 3 * span_embedding_dim + feature_size;
...
     "inference_order": 2
   },
-  "iterator": {
-    "type": "bucket",
-    "sorting_keys": [["text", "tokens___token_ids"]],
-    "batch_size": 1
+  "data_loader": {
+    "batch_sampler": {
+      "type": "bucket",
+      "sorting_keys": [["text", "tokens___token_ids"]],
```
matt-gardner: Remove this line.

ZhaofengWu: https://github.com/allenai/allennlp/pull/3856#issuecomment-592147896

matt-gardner: Oh, the span field might actually be longer than the text field here, so yeah, ok, we should keep this.

matt-gardner: Putting in a comment about that would be good.

PR #3868, opened 2020-02-28
Use the new tokenizers

dirkgr:


dirkgr:
This is currently blocked on https://github.com/huggingface/transformers/issues/3057.

dirkgr:
Also https://github.com/huggingface/transformers/issues/2917.

dirkgr:
And https://github.com/huggingface/transformers/issues/3058.

bryant1410:
Do you think support for using both the fast and slow versions should be provided? Maybe the fast version is buggy for some cases, so then the user can choose slow one.

allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
```diff
@@ -50,8 +45,8 @@ def test_splits_cased_bert(self):
             ",",
             "[MASK]",
             "Allen",
-            "##NL",
-            "##P",
+            "NL",
+            "P",
```
bryant1410: Why these were wrong?

allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
```diff
@@ -169,99 +160,59 @@ def test_token_idx_roberta(self):
             "<s>",
             "A",
             ",",
-            "ĠnaÃ¯ve",  # RoBERTa has a funny way of encoding combining characters.
+            "naïve",
```
bryant1410: I wonder if this breaks RoBERTa usage, because the tokenizer is what changed but the model is the same.

mfuntowicz:
@bryant1410 , @dirkgr We're (@hugginface) actively looking at all these blocking issues, thanks for reporting 👍

dirkgr:
> Do you think support for using both the fast and slow versions should be provided? Maybe the fast version is buggy for some cases, so then the user can choose slow one.

We could certainly back up to that possibility, but I was really excited about getting rid of the hacky token offset code we had there. If we also allow the slow tokenizers, we have to have that back.

allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
dirkgr: The new code sets `token.text` as a slice from the original text, using the offsets provided by the new tokenizers. The original test will never contain the `##` characters that we had there before.

allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
dirkgr: RoBERTa usage is broken because of the leading spaces. The special characters are now handled the right way (except where they break the offsets).

dirkgr:
Found some more: https://github.com/huggingface/transformers/issues/3091, https://github.com/huggingface/transformers/issues/3088.

I'll go do something else now until the next huggingface release. No sense in filing 100 nearly identical issues :-)

dirkgr:
@ZhaofengWu, you might be interested in this, because it re-writes your version of `intra_word_tokenize()`.

ZhaofengWu:
Wait, there is `add_special_tokens` in `encode_plus` _and_ the constructor? Why is that necessary?

ZhaofengWu:
I'm a bit confused by the current approach -- do you only use the dummies to get the type IDs? What makes the previous approach no longer work?

dirkgr:
> Wait, there is add_special_tokens in encode_plus and the constructor? Why is that necessary?

The new fast tokenizers need to know at `__init__` time whether they produce special tokens or not. Because we need both behaviors, I create one of each and then use the right one when we need them.

dirkgr:
> do you only use the dummies to get the type IDs? 

No, it uses the dummies to find out where the special tokens go. I stole that code from your code below.

The only reason I changed the dummy numbers is that we have to pass strings to the tokenizer now, not integers anymore. Token `1000` is a Chinese character (at least in bert-base) that'll be difficult to read for our mostly English speaking developers and users.

dirkgr:
> What makes the previous approach no longer work?

There were a few things, but the main one is this: In this version, `token.text` is always a slice from the original text, as provided by the offsets that the new tokenizers give us.

ZhaofengWu:
> > Wait, there is add_special_tokens in encode_plus and the constructor? Why is that necessary?
> 
> The new fast tokenizers need to know at `__init__` time whether they produce special tokens or not. Because we need both behaviors, I create one of each and then use the right one when we need them.

And we also need to supply that option in `encode_plus`? Or is it just for compatibility with the old tokenizers?

dirkgr:
> And we also need to supply that option in encode_plus? Or is it just for compatibility with the old tokenizers?

I put it there just to be safe.

The new tokenizers actually print a warning when you turn off special characters. I would have expected them to print a warning only when you pass the wrong thing into `encode()`, but it actually prints it all the time. I'm not sure what huggingface is intending here, but I felt this was the safest thing to do.

dirkgr:
I added https://github.com/huggingface/transformers/pull/3112, porting some of our tests to huggingface, so we have a better chance at getting compatible updates.

dirkgr:
Turns out the new tokenizers are still broken: https://github.com/huggingface/transformers/issues/3796

dirkgr:
Since we can't get this into the RC in time, these are the changes to AllenNLP we'll see when we finally do:
 * Character offsets will be either present or absent. Right now, you can have some tokens that have them and some that don't, depending on your input text.
   * We have that setting on the tokenizer, `calculate_character_offsets`, which defaults to `False` right now. This setting won't do anything in the new version, but we have to keep it for backwards compatibility.
 * Tokenization gets faster
 * The mismatched tokenizer logic will have to change. At the very least, the new huggingface tokenizers actually have their own version of re-tokenizing pre-tokenized text, so maybe we can use it. I am not 100% certain we can keep the existing API.

dirkgr:
If we merge this, `token.text` will no longer map to a unique `token.token_id`. @matt-gardner, @Tobias-Rohde, @ZhaofengWu, do you see a problem with that?

matt-gardner:
In our API, the `Tokens` produced by the tokenizer are intended to be consumed by a `TokenIndexer`.  As long as the tokenizer and indexer agree on the meaning of the `Tokens`, it doesn't really matter what's actually in the fields.

bryant1410:
Are all the blocking issues fixed now?

dirkgr:
> Are all the blocking issues fixed now?

All the ones I know about are. One strange thing is that in an intermediate version, spaces were part of RoBERTa tokens. But in this version, they aren't anymore. I asked the @mfuntowicz about it. I don't really care which way it goes, but I want to make sure it's intentionally this way.

ZhaofengWu:
Haven't looked at this in much depth yet, but if the text -> ID mapping is no longer unique, what would happen if we wanted to do things like `convert_tokens_to_ids`?

dirkgr:
`convert_tokens_to_ids()` is a huggingface function, not an AllenNLP function. When you call it with an ambiguous `token.text` from AllenNLP, it will either pick one of the options at random, or not find the token at all.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -66,67 +66,167 @@ def __init__(
...
-            if tokenizer_kwargs["use_fast"]:
-                logger.warning(
-                    "Fast huggingface tokenizers are known to break in certain scenarios."
-                )
-        else:
-            tokenizer_kwargs["use_fast"] = False
-        # As of transformers==2.8.0, fast tokenizers are broken.
-        self.tokenizer = AutoTokenizer.from_pretrained(model_name, **tokenizer_kwargs)
+        if "use_fast" not in tokenizer_kwargs:
+            tokenizer_kwargs["use_fast"] = True
```
bryant1410, line 76: This is a nice readability trick:

```suggestion
        tokenizer_kwargs.setdefault("use_fast", True)
```

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -66,67 +66,167 @@ def __init__(
...
+        )
+
+        self._add_special_tokens = add_special_tokens
+        self._max_length = max_length
+        self._stride = stride
+        self._truncation_strategy = truncation_strategy
 
         # Huggingface tokenizers have different ways of remembering whether they lowercase or not. Detecting it
-        # this way seems like the least brittle way to do it.
+        # this way seems like the least brittle way to do it.		         )
```
bryant1410, line 89: ?

setup.py
```diff
@@ -67,7 +67,7 @@
         "flaky",
         "responses>=0.7",
         "conllu==2.3.2",
-        "transformers>=2.8.0,<2.9.0",
+        "transformers==2.9",
```
bryant1410, line 70: `>=2.9,<2.10`?

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
dirkgr: Not sure it's more readable, but I like concise :-)

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
dirkgr: fixed

setup.py
dirkgr: fixed

dirkgr:
This has gotten way out of control.

In huggingface, the tokenizers are the only source for how to properly format input for some transformer model. They are called "tokenizers", but they do all manner of preprocessing. This version basically reverse engineers a given huggingface tokenizer (with regards to special tokens and token type ids at least), so that we can perform those steps separately. I want to move to a system where tokenizers tokenize, and other parts of the system add special tokens, concatenate text fields, set type ids, split long sequences, and so on. This is not that system, but this is a step in that direction.

Practical benefits of this PR are faster tokenizers, and reliable offsets into the original text as long as the HF tokenizer supports it. Not all do, so sometimes we have to fall back to my crappy version of calculating offsets.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -91,42 +92,140 @@ def __init__(
         detokenized = " ".join(tokenized)
         self._tokenizer_lowercases = "a" in detokenized
 
-        self._add_special_tokens = add_special_tokens
-        self._max_length = max_length
-        self._stride = stride
-        self._truncation_strategy = truncation_strategy
-        self._calculate_character_offsets = calculate_character_offsets
+        # Reverse-engineer the tokenizer for two sequences
```
matt-gardner, line 95: Maybe split this into a separate method?

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -91,42 +92,140 @@ def __init__(
...
-            tokens.append(Token(text=token_str, text_id=token_id, type_id=token_type_id))
+        token_ids, token_type_ids, token_offsets = (
+            encoded_tokens["input_ids"],
+            encoded_tokens["token_type_ids"],
+            encoded_tokens.get("offset_mapping"),
+        )
 
-        if self._calculate_character_offsets:
+        # If we don't have token offsets, try to calculate them ourselves.
+        if token_offsets is None:
```
matt-gardner, line 228: Maybe make this a separate method?

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -177,139 +276,153 @@ def _tokenize(self, sentence_1: str, sentence_2: str = None):
...
+            else:
+                start, end = offsets
+                token_str = text[start:end]
 
-    def tokenize_sentence_pair(self, sentence_1: str, sentence_2: str) -> List[Token]:
-        """
-        This methods properly handles a pair of sentences.
-        """
-        return self._tokenize(sentence_1, sentence_2)
+            tokens.append(Token(text=token_str, text_id=token_id, type_id=token_type_id, idx=start))
```
matt-gardner, line 297: Sorry if I've missed or forgotten an explanation you already gave, but I don't really follow why this needs to be `token_str`, instead of just doing a reverse lookup on `token_id`, as (I believe) we had before.  What do we gain from this?  What do we lose?  (I gave one of the latter in other comments on this PR.)

I'm _pretty sure_ that nothing else in the indexing pipeline will look at `token.text`, because the pretrained transformer indexer can just look at `text_id` and `type_id`, and ignore `token.text`.  So what's the point of this change?  Is it to make offsets easier?  So you can do `idx + len(text)` in order to get character offsets?  If that's the problem that's being solved here, seems easier to just change `idx` to `start_index` and `end_index`, and we can keep the semantics of `token.text` the same.

And, sorry that I didn't realize this issue earlier.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -177,139 +276,153 @@ def _tokenize(self, sentence_1: str, sentence_2: str = None):
...
+            if offsets is None or offsets[0] >= offsets[1]:
+                token_str = self.tokenizer.convert_ids_to_tokens(
+                    token_id, skip_special_tokens=False
+                )
+                start = None
+            else:
+                start, end = offsets
+                token_str = text[start:end]
 
-    def tokenize_sentence_pair(self, sentence_1: str, sentence_2: str) -> List[Token]:
```
matt-gardner, line 186: I don't see this added back anywhere.  Is the functionality just removed, or did I miss it?

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -177,139 +276,153 @@ def _tokenize(self, sentence_1: str, sentence_2: str = None):
...
-
-            start_offset = cumulative
-            cumulative += len(subword_wordpieces)
-            end_offset = cumulative - 1  # inclusive
-            offsets.append((start_offset, end_offset))
-        return wordpieces, offsets, cumulative
-
-    def ids_to_tokens(
-        self, token_ids_a: List[int], token_ids_b: Optional[List[int]] = None
+        def tokens_from_string_tokens(
```
matt-gardner, line 327: Do you really want to be defining functions inside the call to `intra_word_tokenize`?

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
dirkgr: Yes, that function is removed. You should now call `tokenize()` and `add_special_tokens()` separately.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
dirkgr: Why not? If I don't need the function anywhere else?

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
matt-gardner: Ah, I get it now, thanks.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
matt-gardner: It just seems like unfortunate overhead that happens in middle of a pre-processing loop.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
dirkgr: The original purpose was so that the length is correct. But I also like the intuition that you can concatenate tokens and get real text back.

On the other hand, I like the intuition that there is a 1:1 mapping between token string and token id. We can't have both.

allennlp/common/util.py
```diff
@@ -519,9 +519,9 @@ def sanitize_wordpiece(wordpiece: str) -> str:
     if wordpiece.startswith("##"):
         return wordpiece[2:]
     elif wordpiece.startswith("Ġ"):
-        return wordpiece[1:]
+        return wordpiece.replace("Ġ", " ")
```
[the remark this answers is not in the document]
dirkgr: Huggingface have also changed their tack on this. At some point the argument was that `'Ġ'` is just the encoding for space, and the space is part of the token, so I made it this way. But in the latest huggingface, the space is no longer part of the token. I'm hoping for clarification from @mfuntowicz on the matter, but I guess if in doubt, I'll make this match the current behavior.

I don't know if the demo is using `sanitize_wordpiece()` though. It's not a commonly used call as far as I know.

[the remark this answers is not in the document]
matt-gardner: Oh, sorry, you're right.  I thought `sanitize_wordpiece` was called by [`sanitize`](https://github.com/allenai/allennlp/blob/b916720b7e07f2a1aa0273b42309fc2fe5a2148f/allennlp/common/util.py#L63) (which is [used in predictors](https://github.com/allenai/allennlp/blob/b916720b7e07f2a1aa0273b42309fc2fe5a2148f/allennlp/predictors/predictor.py#L172)), but it's not.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
matt-gardner: Well, we can have both by just adding another (optional) field to `Token`, like `display_text`, or something.  And beyond intuition is the fact that we have parts of our code that rely on the 1:1 mapping, and it's not clear to me how I could modify that logic to handle the case where we lose the 1:1 mapping.  It's pretty easy to think of how to modify whatever code depends on the length being right, if there is any.

allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
```diff
@@ -114,33 +56,19 @@ def test_token_idx_bert_uncased(self):
...
-            21,
-            23,
-            25,
-            33,
-            None,
-        ]
-        tokenizer = PretrainedTransformerTokenizer(
-            "bert-base-uncased", calculate_character_offsets=True
-        )
+        expected_idxs = [None, 0, 1, 3, 9, 16, 21, 23, 25, 33, None]
```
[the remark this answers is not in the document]
dirkgr: No, `idx` never does. `idx` is the offset into the original text.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
dirkgr: The whole `Token` class doesn't make sense for transformers. What's `pos_` or `lemma_` on a word piece?

After 1.0 ships we need to do some real clean-up of this API. I'm beginning to formulate ideas for 2.0.

Meanwhile, for 1.0, I'll add an `idx_end` field to track this.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
dirkgr: At some point, I was tempted to remove the `_add_special_tokens` flag completely, assume it's always true, and then deal with special tokens in a `concatenate()` call that does the right thing. But I decided that is too big a change right now.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
dirkgr: I refactored this, and I like the new version better. However, I always thought Python would JIT those functions once and never touch them again, so there is no overhead. So I tested it:

Before: `758 µs ± 943 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)`
After: `734 µs ± 2.21 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)`

So I guess it is 3% faster afterwards. Not a lot, but could matter in some cases. Interesting.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
dirkgr: I changed this back so that `token.text` is the word piece, and I added `token.idx_end`, which is one past the last character of the token.

allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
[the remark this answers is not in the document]
matt-gardner: Oh, sorry, misread the test.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
matt-gardner: I agree with that change, and I don't think it's very large (not saying that it should be done in this PR, though).  We'll it removes an argument, so it's a breaking change, ok...

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
matt-gardner: Huh, thanks for measuring.  I did not assume anything would get JITed, so I thought there would be overhead, but I had no intuition for how large of a difference it would make.  Pretty small in the end, but if it's in the inner-loop of data processing, even small gains that increase GPU utilization are probably worth it.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
dirkgr: Yes, it's not big in terms of lines-of-code, but in downstream impact. I'd have to check and fix all of the readers to be compatible with that change.

allennlp/common/util.py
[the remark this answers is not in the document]
dirkgr: I've changed this back.

allennlp/data/tokenizers/tokenizer.py
```diff
@@ -45,3 +45,37 @@ def tokenize(self, text: str) -> List[Token]:
...
+            An optional second list of tokens. This will be concatenated with `tokens1`. Special tokens will be
+            added as appropriate.
+
+        # Returns
+        tokens : `List[Token]`
+            The combined list of tokens, with special tokens added.
+        """
+        return tokens1 + (tokens2 or [])
+
+    def special_tokens_for_sequence(self) -> int:
```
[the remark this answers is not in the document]
dirkgr: I changed the name.

allennlp/tests/data/token_indexers/pretrained_transformer_indexer_test.py
```diff
@@ -35,20 +35,6 @@ def test_as_array_produces_token_sequence_bert_cased(self):
         indexed = indexer.tokens_to_indices(allennlp_tokens, vocab)
         assert indexed["token_ids"] == expected_ids
 
-    def test_as_array_produces_token_sequence_bert_cased_sentence_pair(self):
```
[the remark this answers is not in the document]
dirkgr: They are testing `tokenize_sentence_pair()`, which I removed. I resurrected them to test `add_special_tokens()`.

allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
```diff
@@ -178,149 +102,36 @@ def test_token_idx_roberta(self):
...
-
-        gc.collect()
-        start = time.monotonic()
-        for i in range(200):
-            tokenizer_with_idx.tokenize(text)
-        with_idx_time = time.monotonic() - start
-
-        assert with_idx_time <= 2 * without_idx_time
-
-    def test_token_idx_sentence_pairs(self):
```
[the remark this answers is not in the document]
dirkgr: The `tokenize_sentence_pair()` call was removed, so there is nothing to test.

allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
```diff
@@ -380,48 +191,60 @@ def test_intra_word_tokenize_whitespaces(self):
...
+
         tokenizer = PretrainedTransformerTokenizer("xlnet-base-cased")
-        assert tokenizer._determine_num_special_tokens_added() == (0, 1, 2)
+        assert get_token_ids(tokenizer.sequence_pair_start_tokens) == []
+        assert get_token_ids(tokenizer.sequence_pair_mid_tokens) == [4]
+        assert get_token_ids(tokenizer.sequence_pair_end_tokens) == [4, 3]
+        assert get_token_ids(tokenizer.single_sequence_start_tokens) == []
+        assert get_token_ids(tokenizer.single_sequence_end_tokens) == [4, 3]
 
-    def test_tokenizer_kwargs_forced_lowercase(self):
```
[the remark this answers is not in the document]
dirkgr: The `do_lower_case` parameter doesn't exist anymore. It is now autodetected whether we need to lowercase or not. There was never a reason to have this setting, as you always had to match it to what your huggingface tokenizer expected anyways. It was just one more thing you could get wrong. Huggingface no longer requires you to give this parameter, and so neither do we.

allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
```diff
@@ -178,149 +102,36 @@ def test_token_idx_roberta(self):
...
-            tokenizer = PretrainedTransformerTokenizer(
-                tokenizer_name, calculate_character_offsets=True
-            )
+            tokenizer = PretrainedTransformerTokenizer(tokenizer_name)
             tokenized = tokenizer.tokenize(sentence)
             assert tokenized[-2].text == "."
             assert tokenized[-2].idx == len(sentence) - 1
 
-    @flaky(max_runs=3)  # This test relies on elapsed wall time, so it's inherently flaky.
-    def test_token_idx_performance(self):
```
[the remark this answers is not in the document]
dirkgr: You could argue that we should still have it, but the situation is now different. We only fall back to the slow code now if Huggingface doesn't have a fast tokenizer for the model yet. So it's a much more limited case. Do we still care whether it is slow in that case? Would we refuse to merge a PR if that was the case?

Also, the `calculate_character_offset` option doesn't exist anymore. It now always calculates them (if needed). So this test would have to compare a fast HF tokenizer with a non-fast one, and then it's unclear what we're actually testing.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
dirkgr: I looked into doing this, and I don't like what it becomes. We'd have to return 5 values from the method, and pass in two tokenizers. There is no way to get a tokenizer with special tokens from one without special tokens, and the function needs both. It's all solvable, but not worth it.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
dirkgr: OdTc3Ce

dirkgr:
Given that the model tests at https://github.com/allenai/allennlp-models/pull/44 are passing, I'm merging this.

PR #3873, opened 2020-02-28
Fix for --recover with distributed training.

### LINKED ISSUE #3872: Configuration error while using --recover with distributed training.
**Describe the bug**
While trying to recover training a model (which was being trained using distributed training), _make_vocab_from_param_ function throws the following configuration error : "The 'vocabulary' directory in the provided serialization directory is non-empty". When tried the same with a single GPU for training, a warning stating "vocabulary serialization directory (provided serialization directory) is not empty" was raised and the vocabulary directory was over-written.

**To Reproduce**
Start training a simple_tagger model provided in docs. The provided json file does not contain distributed training params, hence make the following changes to the json file.

> local NUM_GPUS = 2;
> 
>   "distributed": {
>     "cuda_devices": if NUM_GPUS > 1 then std.range(0, NUM_GPUS - 1) else 0,
>   },
>   "trainer": {
>     "optimizer": "adam",
>     "num_epochs": 40,
>     "patience": 10,
>     "distributed": true,
>   }

Start the training with following command:
> allennlp train tutorials/getting_started/walk_through_allennlp/simple_tagger.json --serialization-dir /home/simple_tagger

Stop after 3 or 4 epochs and restart the training using the below command:
> allennlp train tutorials/getting_started/walk_through_allennlp/simple_tagger.json --serialization-dir /home/simple_tagger --recover

**Expected behavior**
On running the last command, the training should resume from the epoch where we previously stop. But instead it raises and configuration error and the training does not resume.

 - OS: Linux
 - Python version: 3.7
 - AllenNLP version: master

meghana1697:
Closes #3872

### DIFF
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -301,16 +301,19 @@ def train_model(
         # instances. Hence it is safe to construct the vocabulary and write it
         # to disk before initializing the distributed context. The workers will
         # load the vocabulary from the path specified.
-        if params.get("vocabulary", Params({})).get("type", "") != "from_files":
+        vocab_dir = os.path.join(serialization_dir, "vocabulary")
+        if recover:
+            vocab = Vocabulary.from_files(vocab_dir)
+        else:
             vocab = training_util.make_vocab_from_params(
                 params.duplicate(), serialization_dir, print_statistics=dry_run
             )
-            params["vocabulary"] = {
-                "type": "from_files",
-                "directory": os.path.join(serialization_dir, "vocabulary"),
-                "padding_token": vocab._padding_token,
-                "oov_token": vocab._oov_token,
-            }
+        params["vocabulary"] = {
+            "type": "from_files",
+            "directory": vocab_dir,
+            "padding_token": vocab._padding_token,
+            "oov_token": vocab._oov_token,
+        }
 
         mp.spawn(
             _train_worker,
--- a/allennlp/training/util.py
+++ b/allennlp/training/util.py
@@ -453,11 +453,6 @@ def make_vocab_from_params(
     os.makedirs(serialization_dir, exist_ok=True)
     vocab_dir = os.path.join(serialization_dir, "vocabulary")
 
-    if os.path.isdir(vocab_dir) and os.listdir(vocab_dir) is not None:
-        raise ConfigurationError(
-            "The 'vocabulary' directory in the provided serialization directory is non-empty"
-        )
-
     all_datasets = datasets_from_params(params)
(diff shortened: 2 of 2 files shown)


matt-gardner:
Thanks for the PR!  I'm not sure this is the right fix for that issue, though.  I think a better fix would be to just wrap these lines in a `if not recover` block:
https://github.com/allenai/allennlp/blob/4b467ec2882c49c9162d37f40fb55e08d7f5fee0/allennlp/commands/train.py#L305-L307

meghana1697:
@matt-gardner We already have a check for serialization directory, which creates a new directory if recover is false. But in case we have recover as "False" and the serialization directory provided by us is already present, it throws an error stating "Serialization directory ({serialization_dir}) already exists and is not empty. Specify --recover to recover from an existing output folder.". Hence there will not be a case when we have an existing vocabulary directory and recover as "False".

Following is the if statement within _create_serialization_dir_ function:
> if os.path.exists(serialization_dir) and os.listdir(serialization_dir):
        if not recover:
            raise ConfigurationError(
                f"Serialization directory ({serialization_dir}) already exists and is "
                f"not empty. Specify --recover to recover from an existing output folder."
            )

meghana1697:
@matt-gardner There are two reasons for not choosing the above mentioned fix. 

1.  If we block those lines, we need to have a new function that reads the vocabulary params from the existing directory.
https://github.com/allenai/allennlp/blob/4b467ec2882c49c9162d37f40fb55e08d7f5fee0/allennlp/commands/train.py#L308-L313
In the above lines "vocab" is the Vocabulary object returned by "make_vocab_from_params" function.
2. In case of non-distributed training, (i.e. in case of single gpu), the vocabulary gets over-written. "save_to_files" function just throws a warning stating "vocabulary serialization directory %s is not empty".
https://github.com/allenai/allennlp/blob/c4559f3751775aa8bc018db417edc119d29d8051/allennlp/data/vocabulary.py#L576-L577

matt-gardner:
Hmm, right, I forgot about the `vocab._padding_token` and `vocab._oov_token` lines...  It doesn't really matter if the vocab is overwritten if we're recovering, because it will be overwritten with the same thing.  So probably the right thing to do is to have something like:

```python
if recover:
    vocab = Vocabulary.from_files(serialization_dir)  # may not be exactly right
else:
   vocab = training_util.make_vocab_from_params(...)
params["vocabulary"] = ...  # as is currently there
```

The key point is that if we're recovering, then we will already have a vocabulary that we can just read.

meghana1697:
@matt-gardner Have made the required changes.

PR #3876, opened 2020-02-28
Default sort key for token indexers

### LINKED ISSUE #3664: unintuitive `sorting_keys` scoping in new token indexers

viking-sudo-rm:
Gave each `TokenIndexer` a default sorting key, so that if the user wants to manually specify sorting keys for a `BucketBatchSampler`, they do not need to know the internal key names used by the `TokenIndexer`.

For example, say you have a `TextField` called "text" with a `PretrainedTransformerIndexer` called "roberta". The sorting keys would now be `(text, roberta)` instead of `(text, roberta___token_ids)`.

I gave each `Field` an `expand_sort_key` method, which can be used to rewrite default user-specified sort keys into the underlying internal format. Currently, the only field that utilizes this is `TextField`, although it would be straightforward to extend the same pattern to other fields.

Fixes #3664


viking-sudo-rm:
Also worth mentioning that this should be fully backwards compatible. In other words, old-style `roberta___tokens_ids` sort keys should still work.

allennlp/data/samplers/bucket_batch_sampler.py
```diff
@@ -144,5 +153,22 @@ def _guess_sorting_keys(self, instances: Iterable[Instance], num_instances: int
             )
         self.sorting_keys = [longest_padding_key]
 
+    def _expand_sorting_keys(self, schema: Instance) -> None:
```
DeNeutoy: The fact that we need this to exist on the `Sampler` class means this design needs to be revisited.

viking-sudo-rm:
Closing since #3902 is potentially a better fix.

PR #3881, opened 2020-03-02
Add missing param `Trainer.num_serialized_models_to_keep`

bryant1410:


matt-gardner:
This is not a missing param, it's a duplicate parameter that should be passed to the checkpointer, instead of directly to the trainer: https://github.com/allenai/allennlp/blob/4b467ec2882c49c9162d37f40fb55e08d7f5fee0/allennlp/training/checkpointer.py#L30

matt-gardner:
This is also mentioned in the release notes already: https://github.com/allenai/allennlp/releases/tag/v1.0-prerelease

PR #3889, opened 2020-03-02
Use training metrics if validation data is not provided

bratao:
I have an ultra-small dataset that I cannot have the luxury of having a validation dataset. 

A training using only the loss is enough to archive good results on this dataset. However, now the Allennlp do not allow to use a learning_rate_scheduler without a validation dataset.

I could pass the training dataset as the validation during the training, but in this way I have a useless slowdown because you have to double check the training dataset.

In the proposed PR, without the validation_data_loader, we utilize the training metrics to the learning date scheduler.


bratao:
@DeNeutoy I do understand, thanks!

PR #3890, opened 2020-03-02
Change all masks to type torch.bool

### LINKED ISSUE #3745: mask with type torch.bool do not support (1.0 - mask)
**Describe the bug**
I find that many opperations like `(1.0 - mask).to(torch.bool)` in the library, but some functions add 
`mask = mask.float()` while some not, and if mask's type is torch.bool, opreations like `1.0 - mask` will cause error.
> Subtraction, the `-` operator, with a bool tensor is not supported. If you are trying to invert a mask, use the `~` or `bitwise_not()` operator instead.

**To Reproduce**
When use function like allennlp.nn.util.masked_max/mean

**Additional context**
I think docs should illustrate fine type of parameter mask instead of torch.Tensor.
Since Allennlp had upgraded to pytorch 1.2 , operations like this should be modified, I suggest all functions
that need mask should receive parameter with torch.bool.

ZhaofengWu:
Fixes #3745. Also change relevant type signatures to use the more specific `torch.BoolTensor`. I tried to be as comprehensive as possible but there might be ones I missed since there's no single regex that can catch all occurrences.

And please use `torch.BoolTensor` for all your masks in the future for consistency.

### DIFF
--- a/allennlp/models/basic_classifier.py
+++ b/allennlp/models/basic_classifier.py
@@ -115,7 +115,7 @@ class BasicClassifier(Model):
             A scalar loss to be optimised.
         """
         embedded_text = self._text_field_embedder(tokens)
-        mask = get_text_field_mask(tokens).float()
+        mask = get_text_field_mask(tokens)
 
         if self._seq2seq_encoder:
             embedded_text = self._seq2seq_encoder(embedded_text, mask=mask)
--- a/allennlp/models/biaffine_dependency_parser.py
+++ b/allennlp/models/biaffine_dependency_parser.py
@@ -363,7 +363,7 @@ class BiaffineDependencyParser(Model):
         attended_arcs: torch.Tensor,
         head_indices: torch.Tensor,
         head_tags: torch.Tensor,
-        mask: torch.Tensor,
+        mask: torch.BoolTensor,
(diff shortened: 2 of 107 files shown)


bryant1410:
I'm fine with using `BoolTensor` for typing annotation but not for construction. [It's deprecated and doesn't work with CUDA.](https://github.com/pytorch/pytorch/blob/4b3ae7e0af17cfaab832fd7954f5bfad07cee0ba/torch/csrc/utils/tensor_new.cpp#L361) So the GPU tests will fail. Try the following:

```python-console
>>> import torch
>>> torch.BoolTensor([False], device='cuda')
RuntimeError: legacy constructor for device type: cpu was passed device type: cuda, but device type must be: cpu
```

We should use `torch.tensor(..., dtype=torch.bool)` instead.

See also https://github.com/pytorch/pytorch/issues/20122#issuecomment-489293326

ZhaofengWu:
@bryant1410 Thanks for the info! Will change.

bryant1410:
See http://build.allennlp.org/viewLog.html?buildId=27005&buildTypeId=AllenNLP_GpuUnitTests&tab=buildLog&_focus=3101 as an example of a GPU test that failed.

PR #3891, opened 2020-03-02
Remove legacy attention

### LINKED ISSUE #3836: Remove LegacyAttention and LegacyMatrixAttention and similarity_functions
This is old code that can just be removed.  Some of the old RC models depend on it, so we'll either want to re-add it back in there, or update those models to use the new code (which means either retraining them, or hacking their stored weights).

I'm giving this a label of "hour" for removing the code.  Adding it back to the RC repo is maybe another hour.  If we want to update models or code it turns into a day or more.

DeNeutoy:
- Removes `LegacyAttention`
- Removes `LegacyMatrixAttention`
- Removes `similarity_functions`
- Removes `IntraSentenceAttentionEncoder`

- Updates ESIM and DecomposableAttention models to use the non-legacy versions of matrix attention
- Removes the option in `basic_seq2seq_encoder` of passing a similarity function instead of the `Attention` itself.

This is a breaking change for the esim and decomposable attention models.

Fixes #3836


matt-gardner:
It's also a breaking change for BiDAF in the RC repo - did you update any actual trained models here, or just the code?  If you did update saved models, can you document how you did it, so we can do it for BiDAF too?

DeNeutoy:
@matt-gardner see https://github.com/allenai/allennlp-reading-comprehension/pull/26

I've updated the code, but not the models. We should not retrain any models until we've finished all the breaking changes anyway. But you're right that Bidaf, esim and decomposable attention will need to be retrained. Thankfully, they are relatively easy to retrain.

PR #3893, opened 2020-03-03
Return BoolTensor from get_mask_from_sequence_lengths

willfrey:
I saw that this function didn't get updated along with the other masks to be a bool tensor.


willfrey:
I also just noticed that `get_lengths_from_binary_sequence_mask(...)` was missing a return annotation, so I added it.

willfrey:
Thanks for fixing the formatting. I made the edits on here and I probably exceeded the max line length on the annotation!

PR #3894, opened 2020-03-03
More changes: torch.BoolTensor(...) -> torch.tensor(...)

ZhaofengWu:
Same as #3892 but add some missing ones. Apologies that I used a regex that didn't catch all occurrences in #3892.

In the future it might make sense to run some GPU tests before merging PRs to avoid similar situations.


ZhaofengWu:
For this PR I'm essentially doing `s/torch.BoolTensor(/torch.tensor(`, so there shouldn't be more for the constructors. Hopefully we don't have crazy things like:
```
(
  torch.BoolTensor
  (
  ...
  )
)
```

As for the general change to make masks `bool`s, there might be more because it's harder to catch with regexes. See #3893 for an example.

PR #3898, opened 2020-03-04
Wrap the pytorch transformer

dirkgr:
Should go some way towards wrapping up #2270.

Note: This is _not_ pytorch-pretrained-transformers. It has nothing to do with huggingface.

### DIFF
--- a/allennlp/modules/seq2seq_encoders/pytorch_transformer_wrapper.py
+++ b/allennlp/modules/seq2seq_encoders/pytorch_transformer_wrapper.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from overrides import overrides
 import torch
 from torch import nn
@@ -13,7 +15,7 @@ class PytorchTransformer(Seq2SeqEncoder):
     architecture in [Attention is all you Need]
     (https://www.semanticscholar.org/paper/Attention-Is-All-You-Need-Vaswani-Shazeer/0737da0767d77606169cbf4187b83e1ab62f6077).
 
-    This class adapts the Transformer from torch.nn for use in AllenNLP.
+    This class adapts the Transformer from torch.nn for use in AllenNLP. Optionally, it adds positional encodings.
 
     # Parameters
 
@@ -41,7 +43,8 @@ class PytorchTransformer(Seq2SeqEncoder):
         num_layers: int,
         feedforward_hidden_dim: int = 2048,
         num_attention_heads: int = 8,
-        use_positional_encoding: bool = True,
+        positional_encoding: Optional[str] = "sinusoidal",
+        positional_embedding_size: int = 512,
         dropout_prob: float = 0.1,
         activation: str = "relu",
     ) -> None:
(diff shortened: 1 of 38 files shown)


allennlp/modules/seq2seq_encoders/pytorch_transformer_wrapper.py
```diff
@@ -0,0 +1,103 @@
...
+        if self._positional_embedding is not None:
+            position_ids = torch.arange(inputs.size(1), dtype=torch.long, device=output.device)
+            position_ids = position_ids.unsqueeze(0).expand(inputs.shape[:-1])
+            output = output + self._positional_embedding(position_ids)
+
+        # For some reason the torch transformer expects the shape (sequence, batch, features), not the more
+        # familiar (batch, sequence, features), so we have to fix it.
+        output = output.permute(1, 0, 2)
+        output = self._transformer(output, src_key_padding_mask=mask)
+        output = output.permute(1, 0, 2)
```
dirkgr: Stuff like this makes me think I'm barking up the wrong tree with this thing. Why do I have to permute the input, but not the mask? What kind of API is this?

I got the shape annotations from https://pytorch.org/docs/stable/nn.html#torch.nn.Transformer.forward, and it clearly says that the dimensions for the input are different from the dimensions for the mask. But the PyTorch people are not idiots, so what am I missing?

zhangguanheng66: That's true. Unfortunately, the input dimension (seq, batch, feature) used in pytorch transform followed the convention of fairseq.

allennlp/modules/seq2seq_encoders/pytorch_transformer_wrapper.py
```diff
@@ -0,0 +1,103 @@
...
+            d_model=input_dim,
+            nhead=num_attention_heads,
+            dim_feedforward=feedforward_hidden_dim,
+            dropout=dropout_prob,
+            activation=activation,
+        )
+        self._transformer = nn.TransformerEncoder(layer, num_layers)
+        self._input_dim = input_dim
+
+        if positional_encoding is None:
```
zhangguanheng66: Correct me if I'm wrong. IMO, the regular/positional/type embedding layers should stay outside the transformer architecture, although I agree most transformer models need them anyway.

dirkgr: I'm trying to get this as close as possible to our old `StackedSelfAttentionEncoder`, and that one had sinusoidal encodings. I think it's fine to have it here. There is no penalty to turning it off (though you have to remember to do it).

allennlp/modules/seq2seq_encoders/pytorch_transformer_wrapper.py
dirkgr: Thanks for chiming in! Compatibility makes sense as a reason.

allennlp/modules/seq2seq_encoders/pytorch_transformer_wrapper.py
dirkgr: Another reason: In the AllenNLP context, some people will want to use this in via configuration, not code. Those people don't have the option to add in an embedding unless I put it in here.

zhangguanheng66: Make sense to me. It would be great to document those configs and remain some people to turn it off if necessary.

dirkgr:
I would like to convert at least one of our training configs from `stacked_self_attention` to this one and see if it still works, but they don't live in this repo anymore, so this PR is complete.

dirkgr:
Looks like we don't have those models anymore?

DeNeutoy:
Oh, I think the only model that used this was one of the NA-QA net ones from reading comprehension, none of the models in core use it

PR #3899, opened 2020-03-04
Predictor eval by default

### LINKED ISSUE #3888: Non-determinism in some Predictors with dropout
**Describe the bug**
When loading a Predictor directly from its constructor, the underlying model is kept in training mode. This causes models with dropout to apply it during inference, which of course leads to non-deterministic predictions.

**Steps to reproduce**
Creating some custom Predictor and creating it directly from the constructor (`predictor = MyPredictor(model, dataset_reader)`) will yield True for `predictor._model.training`. This does _not_ apply when loading from_path or from_archive etc.

Not sure if this is something the team feels would need to be fixed or is on the user to be aware of (or if it was documented somewhere and I've missed it). For anyone experiencing the same issue, this can be fixed by using `super().__init__(model.eval(), dataset_reader)` in your Predictor's init.

jacobdanovitch:
Fixes #3888, which explains how Predictors don't freeze / set a model to evaluation mode by default when building a Predictor directly from the constructor (instead of `from_archive` or `from_path`). This can lead to non-deterministic predictions due to things like dropout. 

@matt-gardner I added the parameter you requested to the constructor of Predictor, and updated `from_archive` and `from_path` to use it rather than manually calling `model.eval()`. I set it to `True` by default because it seemed like the intended behavior (given that both other methods of loading a Predictor function this way as is). Also added a couple notes in the docstrings. Let me know if there's anything I should change.


bratao:
@jacobdanovitch could you please check the tests. Multiple tests broke with with change:
http://build.allennlp.org/viewLog.html?buildId=27093&buildTypeId=AllenNLP_AllenNLPPullRequests&tab=buildLog

jacobdanovitch:
@bratao Sorry about that, not sure why tests broke. I'll look into it and try again.

jacobdanovitch:
I fixed the tests that were breaking but it seems like there's an unrelated bug in `SrlPredictor`. Tests fail for me on master without my changes as well.

bratao:
@jacobdanovitch the tests pass on the buildbot now. Only the black fomatter test is failing.
Just run "black allennlp" and commit your changes with it.

allennlp/predictors/predictor.py
```diff
@@ -239,6 +241,7 @@ def from_path(
         predictor_name: str = None,
         cuda_device: int = -1,
         dataset_reader_to_load: str = "validation",
+        frozen: bool = False,
```
matt-gardner: You have the default as `False` here, but `True` below.  I think setting this to `True` by default should be fine.

allennlp/predictors/predictor.py
```diff
@@ -276,6 +282,7 @@ def from_archive(
         archive: Archive,
         predictor_name: str = None,
         dataset_reader_to_load: str = "validation",
+        frozen: bool = False,
```
matt-gardner: Again here the default is `False`.

allennlp/predictors/predictor.py
```diff
@@ -40,7 +40,9 @@ class Predictor(Registrable):
     that can be used for serving models through the web API or making predictions in bulk.
     """
 
-    def __init__(self, model: Model, dataset_reader: DatasetReader) -> None:
+    def __init__(self, model: Model, dataset_reader: DatasetReader, frozen: bool = True) -> None:
+        if frozen:
+            model = model.eval()
```
matt-gardner: Just call `model.eval()` here, not `model = model.eval()`.  That's what you do below.

jacobdanovitch:
Sorry for going MIA on this for a bit (I blame the world crumbling around us!). I'll make those fixes in the next day or two.

PR #3901, opened 2020-03-04
Fix special values overflow/underflow for amp

ZhaofengWu:
We currently use a lot of extremely positive/negative/tiny for special values in masked operations. However these cause overflow/underflow with AMP. So this PR changes them to unified constants depending on the dtype.

### DIFF
--- a/allennlp/models/esim.py
+++ b/allennlp/models/esim.py
@@ -205,8 +205,8 @@ class ESIM(Model):
 
         # The pooling layer -- max and avg pooling.
         # (batch_size, model_dim)
-        v_a_max, _ = masked_max(v_ai, premise_mask.unsqueeze(-1), 1)
-        v_b_max, _ = masked_max(v_bi, hypothesis_mask.unsqueeze(-1), 1)
+        v_a_max = masked_max(v_ai, premise_mask.unsqueeze(-1), 1)
+        v_b_max = masked_max(v_bi, hypothesis_mask.unsqueeze(-1), 1)
 
         v_a_avg = torch.sum(v_ai * premise_mask.unsqueeze(-1), dim=1) / torch.sum(
             premise_mask, 1, keepdim=True
--- a/allennlp/models/esim.py
+++ b/allennlp/models/esim.py
@@ -205,8 +205,8 @@ class ESIM(Model):
 
         # The pooling layer -- max and avg pooling.
         # (batch_size, model_dim)
-        v_a_max = masked_max(v_ai, premise_mask.unsqueeze(-1), 1)
-        v_b_max = masked_max(v_bi, hypothesis_mask.unsqueeze(-1), 1)
+        v_a_max = masked_max(v_ai, premise_mask.unsqueeze(-1), dim=1)
+        v_b_max = masked_max(v_bi, hypothesis_mask.unsqueeze(-1), dim=1)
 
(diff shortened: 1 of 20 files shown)


allennlp/nn/util.py
```diff
@@ -1786,8 +1786,8 @@ def masked_topk(
     k = k.reshape(-1)
```
matt-gardner: There's a `1e-13` above that was chosen specifically for regular precision.  It should be addressed somehow for amp.  Maybe with an `EPSILON` value.

ZhaofengWu:
Where would be an ideal place for this constant?

And yes I think I'm missing some occurrences. I will take a look at all `*1e*`. That means we'd need a `VERY_POSITIVE` as well.

bryant1410:
What about introducing an `eps` param for each case? I think it's good for them to actually not be `1e-4` by default when using FP32 (e.g., `1e-8`; depending on the function). What about changing the eps if amp is on?

ZhaofengWu:
> Why is this one a literal?

I got some weird numerical issues when using the smallest representable number for floats. So I'm now hardcoding this in the util function.

ZhaofengWu:
@matt-gardner I made some quite significant changes so it'd be great if you can take another look

allennlp/models/esim.py
```diff
@@ -205,8 +205,8 @@ def forward(  # type: ignore
 
         # The pooling layer -- max and avg pooling.
         # (batch_size, model_dim)
-        v_a_max, _ = replace_masked_values(v_ai, premise_mask.unsqueeze(-1), -1e7).max(dim=1)
-        v_b_max, _ = replace_masked_values(v_bi, hypothesis_mask.unsqueeze(-1), -1e7).max(dim=1)
+        v_a_max = masked_max(v_ai, premise_mask.unsqueeze(-1), 1)
```
matt-gardner: Can you keep the `dim=1` here?  Having a magic value of 1 is not very readable - it leaves me guessing what that 1 means.

bryant1410:
What about adding a comment?

matt-gardner:
Adding a comment where?  For the `dim=1` line?  That's the thing with self-documenting code - you don't need comments if the code is already understandable without it.  Just adding `dim=1` is sufficient, so a comment would be overkill. https://google.github.io/styleguide/cppguide.html#Implementation_Comment_Donts

bryant1410:
Not there, with respect to @ZhaofengWu answer to my comment. I'm in the cell phone and can't hit "r" 😅

ZhaofengWu:
@bryant1410 See https://github.com/allenai/allennlp/pull/3901/files#diff-9cf374ec23a782eaa746b2bd1b389963R1851

bryant1410:
@ZhaofengWu I was referring to the change in `masked_softmax`, but I see you reverted it.

I wonder if you have comments on using `info_value_of_dtype(dtype).eps` instead.

Minor thing, which I arrived late :sweat_smile: maybe it could be `dtype_min`, instead of `min_value_of_dtype` (and similar for the rest)?

bryant1410:
> I wonder if you have comments on using `info_value_of_dtype(dtype).eps` instead.

Now that I think about it, when we do `num / (denom + eps)`, we care about being different from zero, which `info.tiny` is fine and `finfo.eps` is the min `eps` such that `1.0 + eps != 1.0`. So tiny should be better then :+1: The naming is a bit misleading.

PR #3915, opened 2020-03-06
Remove codeblocks from docstrings in commands package.

schmmd:


allennlp/commands/elmo.py
```diff
@@ -11,8 +11,6 @@
 For information, see "Deep contextualized word representations", Peters et al 2018.
 https://arxiv.org/abs/1802.05365
 
-.. code-block:: console
```
DeNeutoy: Did you check that these don't need to be replaced by triple backticks?

scripts/build_docs.sh
```diff
@@ -3,7 +3,7 @@ set -e
 
 cp README.md docs/README.md
 # Alter the relative path of the README image for the docs.
-sed -i '1s/docs/./' docs/README.md
+gsed -i '1s/docs/./' docs/README.md
```
DeNeutoy: What's the difference?

schmmd: I didn't mean to commit this, but on my mac I needed to use `gsed` because I got an error using `sed`.  This shouldn't be committed because it's system dependent and if you're on a Linux box then `sed` will be GNU `sed`.

allennlp/commands/elmo.py
schmmd: Partially.  I'm not done with this PR.  For some reason part of it formats correctly and part doesn't.

DeNeutoy: oh, sorry. To be clear I think these do need bacticks, as they are the output of a terminal help output.

schmmd: Indentation will also format a code block, but what I had didn't quite work because the command example sometimes only had three spaces but four are required to trigger a code block.  I probably should have just used backticks ;-)

PR #3919, opened 2020-03-06
Fix trainer tests when Apex installed

epwalsh:
Hey ya'll,

So I ran into a bunch of issues running the trainer tests after installing Apex. After debugging, I realized this was because the trainer test that uses Apex modifies the model itself, and so every non-Apex test that ran after the Apex test would be using the modified model, which would cause them to fail.

So this PR just refactors the trainer tests so that the Apex test runs with it's own model, vocab, data, etc.


bryant1410:
This was fixed in #3903

Issue #3917, opened 2020-03-06
Publish docs for each version of AllenNLP


schmmd:
Presently we publish a single set of documentation at docs.allennlp.org.  Presently these docs correspond to the tip of AllenNLP, which is confusing if people are using the 0.9 release.

Ideally we would publish docs per commit and for each release.  For example, we could have:

1.  http://docs.allennlp.org/master
2.  http://docs.allennlp.org/1.0.0
3.  http://docs.allennlp.org/1.0.1

To accomplish docs for master per commit, we might want to speed up the doc creation process.

schmmd:
@DeNeutoy @dirkgr @matt-gardner does the above sound reasonable?

dirkgr:
From a user perspective, that's very reasonable. In fact, I'd love to have a version selector at the top left corner, the way Python docs have.

But it's going to be a bit of work to go from what we have now to doing that.

epwalsh:
On the subject of speeding up the doc creation process: I had the same issue with another project I was working on, so ended up using a Makefile to manage creating the markdown API doc files with `mathy_pydoc` so it's only run on files that have changed. Basically something like this:

```make
# Makefile
SRC = allennlp
SRCS = $(shell find $(SRC) -type f -name '*.py' | grep -v -E 'tests|__init__|version')
DOCS = $(subst .py,.md,$(subst $(SRC)/,docs/api/,$(SRCS)))

docs/api/%.md : $(SRC)/%.py
	mkdir -p $(shell dirname $@)
	python -m mathy_pydoc $(subst .py,++,$(subst /,.,$<)) > $@

.PHONY : build-docs
build-docs : $(DOCS)
```

Of course for this to work in a CI environment we'd have to cache the source code and markdown doc output as well otherwise everything would be rebuilt everytime.

schmmd:
Well, it takes `0.48s` to full generate the docs if we avoid shelling out for each file.  Otherwise it takes minutes.

schmmd:
Ah shoot--nevermind I'm pretty sure those timings have a problem...

schmmd:
I think it's 6s vs 1165s with the following intermediate hack:

```
diff --git a/scripts/build_docs.py b/scripts/build_docs.py
index 05b62d5b..962f10bf 100644
--- a/scripts/build_docs.py
+++ b/scripts/build_docs.py
@@ -1,3 +1,5 @@
+from mathy_pydoc.__main__ import main
+import sys
 from typing import Dict
[...]

PR #3920, opened 2020-03-07
Speed up docs generation.

schmmd:
Related to https://github.com/allenai/allennlp/issues/3917

This speeds up docs generation from 1165s (20 minutes) to 6s for a 194x speedup.  The code has some gnarly workarounds for the fact that the code in `mathy_pydoc` is not very modular (the main method reads from the args directly and writes directly to stdout).

### DIFF
--- a/scripts/build_docs.py
+++ b/scripts/build_docs.py
@@ -1,3 +1,7 @@
+from mathy_pydoc.__main__ import main
+from contextlib import redirect_stdout
+import sys
+import io
 from typing import Dict
 import os
 from pathlib import Path
@@ -15,7 +19,6 @@ exclude_files = [
     "__main__.py",
 ]
 
-
 def render_file(relative_src_path: str, src_file: str, to_file: str, modifier="++") -> None:
     """
     Shells out to pydocmd, which creates a .md file from the docstrings of python functions and classes in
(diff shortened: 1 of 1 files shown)


scripts/build_docs.py
```diff
@@ -31,10 +34,13 @@ def render_file(relative_src_path: str, src_file: str, to_file: str, modifier="+
...
-    args = ["mathy_pydoc", namespace]
-    call_result = check_output(args, env=os.environ).decode("utf-8")
+    sys.argv = ["mathy_pydoc", namespace]
+    output = io.StringIO()
+    with redirect_stdout(output):
+        main()
+
     with open(to_file, "w") as f:
-        f.write(call_result)
+        f.write(output.getvalue())
```
dirkgr: If it just gets written to a file, why not do `with redirect_stdout(f):`?

scripts/build_docs.py
```diff
@@ -1,3 +1,7 @@
+from mathy_pydoc.__main__ import main
```
dirkgr: It's not super important, but it would make me happy if it was abundantly clear (through the name) that this `main()` function is not our own `main()` function, for example by saying `from mathy_pydoc.__main__ import main as pydoc_main` or something like that.

schmmd: Good point.  I had done that the first time ;-P

scripts/build_docs.py
schmmd: That seems to work ;-)

schmmd: We're down to 5s now...

dirkgr: Pretty soon this script will finish before it started.

Issue #3914, opened 2020-03-06
The "edit pencil" in the docs 404's


schmmd:
The pencil should go to the right file to edit the docstring, but instead if 404's.  We should either fix the link or remove the pencil.

![image](https://user-images.githubusercontent.com/954798/76103485-a19e8b80-5f86-11ea-8885-7287ff3ef3ed.png)

schmmd:
@DeNeutoy do you have an idea of how to fix this?  I remember you mentioning this would work once we merged down some of your initial changes.

DeNeutoy:
Oh, sorry, this slipped through my net for the docs. It's easy to configure here:

https://github.com/allenai/allennlp/blob/master/mkdocs.yml#L21

Generally everything that is configurable about the docs lives in this file.

schmmd:
We also need to change the edit button so it points to `py` files rather than `md` files.

DeNeutoy:
Actually, we should probably remove the button entirely, because it's a relative path within the docs, which we aren't going to be able to make point to our .py files. I'll send a PR.

PR #3929, opened 2020-03-09
Remove the edit link from the docs.

schmmd:
https://github.com/allenai/allennlp/pull/3916 didn't fix https://github.com/allenai/allennlp/issues/3914 because a default value is set when `repo_url` is set.  See https://www.mkdocs.org/user-guide/configuration/.

> On a few known hosts (specifically GitHub, Bitbucket and GitLab), the edit_uri is derived from the 'repo_url' and does not need to be set manually. Simply defining a repo_url will automatically populate the edit_uri configs setting.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -17,7 +17,6 @@ theme:
 
 
 repo_name: allenai/allennlp
-repo_url: https://github.com/allenai/allennlp
 # TODO(markn): Consider adding GA here, if we care about it.
 
 nav:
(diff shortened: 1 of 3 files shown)


mkdocs.yml
```diff
@@ -17,7 +17,6 @@ theme:
 
 
 repo_name: allenai/allennlp
-repo_url: https://github.com/allenai/allennlp
```
dirkgr: This is the magic?

schmmd: yes

PR #3932, opened 2020-03-09
Switch to the new PytorchTransformer

dirkgr:
`QanetEncoder` is the only thing that used the old `StackedSelfAttention` class. But shouldn't all the Qanet stuff live in the reading comprehension repo, or in our new models repo?


dirkgr:
Screen Shot 2020-03-09 at 15 27 01

I tried training Qanet with this (https://github.com/allenai/allennlp-reading-comprehension/pull/28), and these are the learning curves I'm getting.

It trains slower, the quality is lower, and it's less configurable. Also, the API is bizarre. Maybe we keep what we had after all.

dirkgr:
We're not doing this, since it turns out that the old `StackedAttentionEncoder` code is not a 1:1 implementation of https://www.semanticscholar.org/paper/Attention-is-All-you-Need-Vaswani-Shazeer/204e3073870fae3d05bcbc2f6a8e263d9b72e776. Instead, I will rename this to `QanetEncoder` and move it into the models repository.

PR #3933, opened 2020-03-09
Fix doc links in tutorials.

schmmd:


schmmd:
We decided to point to master docs.

PR #3946, opened 2020-03-14
Fix bare string from params error

### LINKED ISSUE #3864: `RandomSampler` creation by name fails
**Describe the bug**

Creating a `RandomSampler` in a config file by just the name fails.

**To Reproduce**

Define a config file having:

```jsonnet
{
  // ...
  data_loader: {
      sampler: 'random',
  },
  // ...
}
```

When trying to train, it'll fail to create the object.

The error seems to be in:

https://github.com/allenai/allennlp/blob/94d2f1511e0d46191e4f230615da3d5882474a41/allennlp/common/from_params.py#L338

**Expected behavior**

The config creation should happen successfully.

 - OS: [e.g. OSX, Linux] Linux
 - Python version: [if it's not 3.6.1 or later, that's probably your problem] 3.7.6
 - AllenNLP version: [e.g. v0.9.0, or "I installed from master"] I installed from master
 - PyTorch version: (if you installed it yourself) 1.4.0

matt-gardner:
This is an attempt to fix #3864, but I failed to reproduce the error.  Digging through the logic, my best guess is that @bryant1410 actually got the error from a different part of the code, because the sampler creation appears to work correctly.  @bryant1410, can you modify this to add a failing test case?  I guess that means either opening a PR to this branch, or just opening your own PR with a failing test case, because I don't think you'll be able to directly push to this branch.  The underlying issue that you brought up in #3864 should be very easy to fix, but I need a failing test case first, and I couldn't find one.

EDIT: Now fixes #3864.


bryant1410:
:+1: Thanks for opening the PR! See matt-gardner/allennlp#2

PR #3947, opened 2020-03-14
Make kwargs inference work with arbitrary multiple inheritance order

### LINKED ISSUE #3716: Make **kwargs inference robust to multiple inheritance
We somewhat frequently do things like `class MyModel(torch.nn.Module, FromParams)`, but our argument inference code is not robust to this kind of multiple inheritance (it is dependent on the order in which you specify the superclasses).  I don't think we actually run into issues from this right now, because we get lucky in how we typically use `Registrable`, but it's a potential problem.  See https://github.com/allenai/allennlp/pull/3705#discussion_r374321802 for some more context.

We need a simple test case using multiple inheritance, then a fix to the call to `cls.mro()[1]` in our **kwargs inference code.  Should be pretty easy.

matt-gardner:
Fixes #3716.


allennlp/common/from_params.py
```diff
@@ -152,11 +152,15 @@ def infer_params(cls: Type[T], constructor: Callable[..., T] = None):
...
-    # problem because of it.
-    super_class = cls.mro()[1]
+    # first superclass, and so on.  We take the first superclass we find that inherits from
+    # FromParams.
+    super_class = None
+    for super_class_candidate in cls.mro()[1:]:
+        if issubclass(super_class_candidate, FromParams):
+            super_class = super_class_candidate
+            break
+    if not super_class:
```
dirkgr: I prefer `if super_class is None:`. There are many things in Python that are falsy. It can be confusing.

Of course, that style is common in the library, and one more instance doesn't make much of a difference.

matt-gardner: Yeah, I do that in a bunch of cases, where there's a chance for error.  I don't think it's needed here - classes returned by `mro()` are not going to be falsy.  I'll also note that `if not X` is not just common in the library, it's recommended, idiomatic python usage.

Issue #3913, opened 2020-03-06
Make the trainer more configurable


schmmd:
From https://github.com/allenai/allennlp/issues/3519#issuecomment-588419410:

Ok, I'm scoping this out right now, taking the approach of adding a few targeted callbacks to the original trainer, and simplifying what's there / making it more modular.  Here's a list of specific items that I'm thinking of doing; I'm putting this here to get feedback before starting on this.  My process for deciding on these things to change was (1) see what parameters to `__init__` I can remove / push to other, configurable dependencies, (2) see what other places might want some customization, and propose specific callbacks for them.

1. ~Change the name of `TrainerBase` to `Trainer`, and `Trainer` to `GradientDescentTrainer`.  This is just a cosmetic change that makes the trainer code consistent with other naming conventions in the library, and makes it so we can use `Trainer` annotations in places where it's appropriate, instead of the more awkward-looking `TrainerBase`.  I could maybe be persuaded to just make it one class, even, that's still `Registrable` and can be subclassed (like `Embedding` is).~

2. ~Make `TensorboardWriter` an argument to `__init__`, and to `from_partial_objects` (with a `Lazy` annotation), so I can remove several of the `__init__` parameters that are specific to the tensorboard writer.  Specifically, that's these parameters: https://github.com/allenai/allennlp/blob/dd4ee62cf55e5fe50c870152501f909203a6b1d7/allennlp/training/trainer.py#L59-L62~

[...]

PR #3949, opened 2020-03-14
Simplifying some parameters in the trainer

matt-gardner:
This is a fix for some parts of #3913. Specifically, items 1 and 2 and the first part of 4.  Once this is merged, I'll cross out those parts of the issue, to better keep track of what's left.

There should not be any logic changes in here, this is just moving things around and changing a couple of names.


allennlp/training/tensorboard_writer.py
```diff
@@ -17,30 +17,45 @@ class TensorboardWriter(FromParams):
...
         If None, activation histograms will not be written out.
+        When this parameter is specified, the following additional logging is enabled:
+            * Histograms of model parameters
+            * The ratio of parameter update norm to parameter norm
+            * Histogram of layer activations
+        We log histograms of the parameters returned by
+        `model.get_parameters_for_histogram_tensorboard_logging`.
+        The layer activations are logged for any modules in the `Model` that have
+        the attribute `should_log_activations` set to `True`.  Logging
+        histograms requires a number of GPU-CPU copies during training and is typically
```
bryant1410: > requires a number of GPU-CPU copies during training

While this is true and completely fine to state, it makes me remember that using tqdm after every batch to show the training status has also the same effect.

bryant1410:
(LGTM)

allennlp/training/tensorboard_writer.py
dirkgr: Can you elaborate on that? Doesn't `tqdm` just copy a tiny tensor, instead of all the parameters?

matt-gardner:
It's removing lines of code because moving the logic let me remove some things that were no longer necessary.  This includes (1) removing a test that was checking for multiple ways of passing parameters to the checkpointer (~half of the line reduction), (2) removal of logic for dealing with (1), and (3) removing the lines in various places that specified the parameters that got moved into `Checkpointer` and `TensorboardWriter`.

(Also, please don't feel any pressure to work on the weekend because I chose to work on the weekend.)

allennlp/training/tensorboard_writer.py
matt-gardner: This didn't use to be true of metrics, but now metrics can live entirely on the GPU.  If they do, then you'll be doing GPU/CPU copies of the metric values when reporting metrics during training / validation.  This is *much* lower bandwidth transfer than doing the same for weights and activations.  Not sure how much impact it has on runtimes.

bryant1410: > Can you elaborate on that? Doesn't `tqdm` just copy a tiny tensor, instead of all the parameters?

Yes. When the GPU operations are executed with the current data batch (supposing GPU usage), they live in GPU memory and never do GPU-CPU sync, and are executed async actually because they're enqueued by PyTorch transparently. When any GPU variable is read and needs to be used in CPU memory, the CUDA device is gonna be synchronized (waiting for the necessary operations to finish execution) and transferred to CPU memory, to then continue with the execution. This happens when displaying the metrics with tqdm:

1. The batch is executed on GPU. The metrics are computed on GPU.
2. The value is copied to CPU to show in tqdm.
3. Next batch...

The next batch waits for the transfer in the middle. Which is different from not waiting for GPU-CPU transfers (which are expensive relative to GPU execution) between many batch executions. It could be solved with some `log_every` option or similar.

Btw, it'd be good to measure this somehow (e.g, show there are no transfers with some profiling, doing transfers on purpose to see how slow it is).

As a side note, I'm not 100% confident this is how it works, but I'm sure enough to raise the concern.

Relevant documentation: [Asynchronous execution - CUDA Semantics - PyTorch master documentation](https://pytorch.org/docs/stable/notes/cuda.html#asynchronous-execution).

dirkgr:
(I don't feel that pressure. I just don't set strict boundaries.)

PR #3956, opened 2020-03-17
pads in a more pythonic way which is significantly faster.

qiuwei:
This commit does the padding in a more pythonic way which significantly improves the speed.

A preliminary benchmark on my own dataset and model:
The original version:
image

new version:
image

The second column is the call count. The third is the execution time and the percentage. The forth is the execution time of `pad_sequence_to_length` excluding the functions it calls 

I think the difference is mainly due to that doing  lots of insertion and append is expensive.

### DIFF
--- a/allennlp/common/util.py
+++ b/allennlp/common/util.py
@@ -180,6 +180,8 @@ def pad_sequence_to_length(
         padded_sequence = sequence[-desired_length:]
     # Continues to pad with default_value() until we reach the desired length.
     pad_length = desired_length - len(padded_sequence)
+    # This just creates the default value once, so if it's a list, and if it gets mutated
+    # later, it could cause subtle bugs. But the risk there is low, and this is much faster.
     values_to_pad = [default_value()] * pad_length
     if padding_on_right:
         padded_sequence = padded_sequence + values_to_pad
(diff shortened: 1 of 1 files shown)


allennlp/common/util.py
```diff
@@ -179,11 +179,12 @@ def pad_sequence_to_length(
     else:
         padded_sequence = sequence[-desired_length:]
     # Continues to pad with default_value() until we reach the desired length.
-    for _ in range(desired_length - len(padded_sequence)):
-        if padding_on_right:
-            padded_sequence.append(default_value())
-        else:
-            padded_sequence.insert(0, default_value())
+    pad_length = desired_length - len(padded_sequence)
+    values_to_pad = [default_value()] * pad_length
```
matt-gardner: My guess is that a major speedup is coming from only calling `default_value()` once.  This is probably ok, but it does make me *slightly* nervous, because that default value is a list in some cases, so you'll sometimes have the *same list* show up lots of times as the padding value.  Because we don't ever mutate these lists, it should be fine, though.  But it at the very least deserves a comment noting that this could be dangerous in some uses.
```suggestion
    # This just creates the default value once, so if it's a list, and if it gets mutated
    # later, it could cause subtle bugs. But the risk there is low, and this is much faster.
    values_to_pad = [default_value()] * pad_length
```

PR #3957, opened 2020-03-17
Removes language models and seq2seq models

dirkgr:
These are moved to https://github.com/allenai/allennlp-models/pull/6.


bryant1410:
What about ELMo?

dirkgr:
ELMo is a module, not a model, in our code. I am only moving models. Also, ELMo is too general to be stuck into the allennlp-models repo. All kinds of other models rely on it.

PR #3958, opened 2020-03-17
gitignore doc artifacts so you can easily build docs locally

epwalsh:
This just makes it so you can build the docs locally without having to clean up after yourself. The biggest change is renaming `mkdocs.yml` to `mkdocs-skeleton.yml`. I'm not sure if that's the best name though.


schmmd:
image

Our build system seems to depend on this file existing.  I agree with your change though--it's annoying that you need to carefully not check in this file.  Someone on our side would probably need to look into how to modify our build system to get this changed though.

epwalsh:
@schmmd I think I might have fixed it

schmmd:
👍

epwalsh:
No worries!

Issue #3100, opened 2019-07-26
Add support for significance testing, random splits, and other pro-reproducibility features


julianmichael:
**Is your feature request related to a problem? Please describe.**
There's been a lot of talk in the community about reproducibility issues, incl. with effects of maxing over hyperparameter settings and overfitting to the test set.

**Describe the solution you'd like**
There are some recommendations that have come out of the literature and water cooler: standard hyperparameter search procedures, reporting mean/stdev of multiple training runs, [significance testing with bonferroni procedure](https://www.aclweb.org/anthology/P18-1128) on [random splits](https://wellformedness.com/papers/gorman-bedrick-2019.pdf), etc. and I think adoption of these techniques is usually driven by ease of use. So AllenNLP seems like a perfect place to implement tools to do this: significance testing relates to the choice of metric, data splits fit naturally into the trainer abstraction, and hyperparameter optimization is a feature of the training procedure. All of these are extant abstractions in AllenNLP and any implementation of these features on top of AllenNLP would likely duplicate work. More concretely:

* An additional `Trainer` (or config options) that constructs random train/dev/test splits and stores metadata to reconstruct the split in the serialization directory.
* Options for `Trainer` that support multiple training runs (same model hyperparams, optionally different data splits) and aggregating metrics across runs; perhaps with a new serialization directory structure that contains each individual run as a subdirectory, with aggregated metrics in the superdirectory (matrix directory?).
* A module for significance testing, with appropriate subcommands including interop with the new trainers: imagine typing something like `allennlp sigtest students-t save/model1_multi save/model2_multi` and immediately getting results---or getting warnings/errors if the underlying datasets/training configs don't match.
* Possibly additions to the `Metric` abstraction to (optionally) support canonical aggregation methods and significance tests; for example, if the final metric is the mean of the metric value on examples, it is appropriate to perform a paired Student's t-test (as long as _n_ is great enough), so out-of-the-box support for this can be documented and tested as a one-liner.

Hyperparameter optimization may still be out of scope for now / more appropriate for an outer layer like Beaker.

**Describe alternatives you've considered**
One alternative is to leave this stuff as an "outer layer" for a separate library. However, the tools I'm describing seem too specific for Beaker and are conceptually closely tied to AllenNLP. Easy interop (e.g. with one-line significance testing) would require relying on the AllenNLP workflow anyway.

**Additional context**
[...]

PR #3960, opened 2020-03-17
Add Cross Validation

bryant1410:
Related to #2827 and #3100.

Can you give me feedback? It's ready to review in abstraction and code working fine (not for documentation and testing).

I'm gonna leave some comments as a review.


allennlp/training/cross_validation.py
```diff
@@ -0,0 +1,317 @@
...
+    @overrides
+    def get_n_splits(self, instances: Sequence[Instance]) -> int:
+        groups, labels = self._labels_groups(instances)
+        return super().get_n_splits(instances, labels, groups=groups)
+
+    @staticmethod
+    def _labels_groups(
+        instances: Sequence[Instance],
+    ) -> Tuple[Optional[Sequence[Union[str, int]]], Optional[Sequence[Any]]]:
+        labels = [instance["label"] for instance in instances] if "label" in instances[0] else None
```
bryant1410: I'm thinking that the key names "label" and "_group" may be defined when the `CrossValidator` object is defined, or be the dataset `Sequence[Instance]`. Thoughts on this? Also, what if it's a nested value (e.g., `instance["metadata"]["_group"]`)?

allennlp/training/cross_validation.py
```diff
@@ -0,0 +1,317 @@
...
+    def __init__(
+        self,
+        serialization_dir: str,
+        dataset: Dataset,
+        data_loader_builder: Lazy[DataLoader],
+        model: Model,
+        trainer_builder: Lazy[Trainer],
+        cross_validator: CrossValidator,
+        batch_weight_key: str = "",
+        retrain: bool = False,
```
bryant1410: In scikit-learn defaults to true, but I think it makes more sense here to default to false.

bryant1410:
@matt-gardner would you mind reviewing when you have time?

allennlp/training/cross_validation.py
```diff
@@ -0,0 +1,317 @@
...
+    @overrides
+    def run(self) -> Dict[str, Any]:
+        metrics_by_fold = []
+
+        n_splits = self.cross_validator.get_n_splits(self.dataset)
+
+        for fold_index, (train_indices, test_indices) in enumerate(
+            self.cross_validator(self.dataset)
+        ):
+            logger.info(f"Fold {fold_index}/{n_splits - 1}")
```
bryant1410: I do zero-based, as in the epoch number.

allennlp/training/cross_validation.py
```diff
@@ -0,0 +1,317 @@
...
+
+            serialization_dir = os.path.join(self.serialization_dir, f"fold_{fold_index}")
+            os.makedirs(serialization_dir, exist_ok=True)
+
+            train_dataset = Subset(self.dataset, train_indices)
```
bryant1410: This fails if there's a bucket sampler, because it needs the `Vocabulary` from the `Dataset`, but `Subset` doesn't have one. Thoughts? I see 2 options: the sampler having a `Vocabulary` arg instead of a `Dataset` (it only uses it for that), or implementing our own `Subset`.

allennlp/training/cross_validation.py
```diff
@@ -0,0 +1,317 @@
...
+
+    @overrides
+    def finish(self, metrics: Dict[str, Any]):
+        common_util.dump_metrics(
+            os.path.join(self.serialization_dir, "metrics.json"), metrics, log=True
+        )
+
+    @classmethod
+    @overrides
+    def from_partial_objects(
```
bryant1410: Is there a way some of the content of this function is refactored out, along with `TrainModel`?

allennlp/training/cross_validation.py
```diff
@@ -0,0 +1,317 @@
...
+class TimeSeriesSplit(CrossValidator, sklearn.model_selection.TimeSeriesSplit):
+    def __init__(self, n_splits: int = 5, max_train_size: Optional[int] = None) -> None:
+        super().__init__(n_splits=n_splits, max_train_size=max_train_size)
+
+
+logger = logging.getLogger(__name__)
+
+
+@TrainModel.register("cross_validation", constructor="from_partial_objects")
+class CrossValidateModel(TrainModel):
```
bryant1410: I don't think I need to subclass `TrainModel` but it felt like the best thing to do though. Maybe there should be an abstract class for both (and the other one maybe name it `HoldoutTrain`?), which just defines the contract?

allennlp/training/cross_validation.py
```diff
@@ -0,0 +1,317 @@
...
+
+            train_data_loader = self.data_loader_builder.construct(dataset=train_dataset)
+            test_data_loader = self.data_loader_builder.construct(dataset=test_dataset)
+
+            model = copy.deepcopy(self.model)
+            subtrainer = self.trainer_builder.construct(
+                serialization_dir=serialization_dir, data_loader=train_data_loader, model=model
+            )
+
+            fold_metrics = subtrainer.train()
```
bryant1410: We could add support for nested cross-validation, which should be as easy as supporting another `CrossValidator` and using it.

bryant1410:
[Mlextend](https://rasbt.github.io/mlxtend/) offers scikit-learn-like more statistical testing, which could also be incorporated, after #3100.

allennlp/training/cross_validation.py
bryant1410: There's also just doing: `train_dataset.vocab = self.dataset.vocab`

allennlp/training/cross_validation.py
matt-gardner: If we do this, the base class should be `Command`.  That's basically what's going on here; we're defining a command and a config file specification.  Doing this should let us remove a bunch of boilerplate in our command code.

allennlp/training/cross_validation.py
matt-gardner: It's not obvious to me what this is doing or why it's needed.  Can you give some more detail?

allennlp/training/cross_validation.py
matt-gardner: I don't understand this enough to answer this question.  I'd need more detail / to read more, and I don't have time for that right now.

allennlp/training/cross_validation.py
matt-gardner: You're only talking about like 8 lines of logic.  The function looks like it takes a lot of lines, but that's just because of all of the argument passing.  Not sure it's worth factoring out those 8 lines.  Though, actually, those lines could be simplified.  Comments below.

allennlp/training/cross_validation.py
```diff
@@ -0,0 +1,317 @@
...
+        local_rank: int,  # It's passed transparently directly to the trainer.
+        vocabulary: Lazy[Vocabulary] = None,
+        retrain: bool = False,
+    ) -> "CrossValidateModel":
+        logger.info(f"Reading data from {data_path}")
+        dataset = dataset_reader.read(data_path)
+
+        vocabulary_ = vocabulary.construct(instances=dataset)
+        if not vocabulary_:
+            vocabulary_ = Vocabulary.from_instances(dataset)
```
matt-gardner:
```suggestion
        vocabulary_ = vocabulary.construct(instances=dataset) or Vocabulary.from_instances(dataset)
```

allennlp/training/cross_validation.py
```diff
@@ -0,0 +1,317 @@
...
+        model_ = model.construct(vocab=vocabulary_)
+
+        if common_util.is_master():
```
matt-gardner: Pretty sure you can remove these, unless you are actually planning on literally using the `train` script for this, including the distributed training.  Not sure how that interacts with how this does cross validation.

allennlp/training/cross_validation.py
dirkgr: I'd be happy to look at that as a separate PR, but let's not do it here.

allennlp/training/cross_validation.py
```diff
@@ -0,0 +1,317 @@
...
+            if isinstance(fold_0_metric_value, float):
+                average = Average()
+                for fold_index, fold_metrics in enumerate(metrics_by_fold):
+                    metric_value = fold_metrics[metric_key]
+                    metrics[f"fold{fold_index}_{metric_key}"] = metric_value
+                    average(metric_value)
+                metrics[f"average_{metric_key}"] = average.get_metric()
+            else:
+                for fold_index, fold_metrics in enumerate(metrics_by_fold):
+                    metrics[f"fold{fold_index}_{metric_key}"] = fold_metrics[metric_key]
```
dirkgr: You could do this for float values as well, right? Basically, do the `else` block for everything, but for floats, also report an average? Maybe add an stddev/min/max, so that it's easy to see if there us an outlier skewing the mean?

allennlp/training/cross_validation.py
bryant1410: You mean to refactor out `TrainModel` behavior first in another PR, then come back here and rebase it to continue here?

allennlp/training/cross_validation.py
bryant1410: There are 2 options in CV (and when randomly splitting) that are common: stratification and grouping.

The former is about making sure all class labels are represented with a similar balance in train and test. For that, you need to know about the target class (so, we need to know "label").

For the latter, it can be for example about doing leave-one-group-out CV or making sure train and test instances come from different "groups" (whatever groups mean in a given dataset). For that, you need to specify what the group is for each instance.

allennlp/training/cross_validation.py
bryant1410: `BucketBatchSampler` needs a special `Dataset`: one that has a `vocab` attribute. `Subset` doesn't.

https://github.com/allenai/allennlp/blob/644ef22e90c8493af8e14e81a1e10839a8f0b62a/allennlp/data/samplers/bucket_batch_sampler.py#L69

allennlp/training/cross_validation.py
bryant1410: I was referring to the fact of having to define this method, which is somewhat not the best thing to do overall IMHO, beyond its size.

allennlp/training/cross_validation.py
bryant1410: `Command` sounds confusing when also having `Subcommand`. What about `TrainingMethod` (because it's used only with `allennlp train`).

allennlp/training/cross_validation.py
dirkgr: The problem is that torch's `Dataset`s don't have a `vocab` field, and they don't have a `index_with` method either. So whenever you derive a new `Dataset` from an existing `AllennlpDataset`, using torch methods, those two things get lost.

I think what we should do is just use straight torch `Dataset` everywhere, move the vocab around manually (or maybe inside the instances?), and have a function `index_dataset_with_vocab(dataset, vocab)` which we call instead of `Dataset.index_with`.

I also think that should not be part of this PR. For this PR, `train_dataset.vocab = self.dataset.vocab` is fine. Let's open another issue for the follow-up work item. Many small PRs are better than one huge one.

allennlp/training/cross_validation.py
bryant1410: > unless you are actually planning on literally using the `train` script for this,

You mean the parts of `train.py` that are not in `TrainModel`?

allennlp/training/cross_validation.py
dirkgr: No, I mean let's not invent new infrastructure in this PR.

Let's also keep in mind that there are only two cases here, training and cross-validation. We don't need to invent a whole new infrastructure as if we'd have dozens of similar training procedures. There are just two, which (maybe) need to share a bit of code.

It was my understanding that you would put this into a `Subcommand`. Did I misunderstand that?

allennlp/training/cross_validation.py
dirkgr: Let's get the 90% case working first.

allennlp/training/cross_validation.py
dirkgr: Doesn't that produce weird output in some cases? "Fold 0/0", when you have only one? "Fold 1/1" when you have two?

allennlp/training/cross_validation.py
dirkgr: So this works only if the instance have a field called `"label"`?

allennlp/training/cross_validation.py
matt-gardner: I definitely agree with not doing anything like this in this PR.  Sorry, that wasn't what I was suggesting at all.

In this PR, the base class for `CrossValidateModel` should be `Registrable`, and it should be structured similarly to `TrainModel`.  The class should live in `commands`, because that's what it is.

*If* we want an abstract class for both, it should be something like `Command`.  That's a totally separate issue that would need a bunch of design work to get it right, and may or may not be worth the effort.

allennlp/training/cross_validation.py
matt-gardner: I'd be happier with this if we found a way to let a caller pass in a function to pull out the label or group.

allennlp/training/cross_validation.py
matt-gardner: I also definitely agree on not doing this in this PR.  I think what you're suggesting, @dirkgr, is having the sampler / dataloader own the vocab, instead of the dataset?  That would probably work, but I'd have to dig through the API to be sure.

allennlp/training/cross_validation.py
matt-gardner: This method is necessary to construct things from config files.  If you want to construct the `CrossValidationModel` object yourself, you don't need this.  There's no way around having this method, and it can't be shared.

allennlp/training/cross_validation.py
matt-gardner: Yes, that's what I mean.

allennlp/training/cross_validation.py
bryant1410: > So this works only if the instance have a field called `"label"`?

Yes. Note this is also assumed here (and if IIRC, in more places):

https://github.com/allenai/allennlp/blob/e20997059a6d7db32cdd56df77d710b564349957/allennlp/models/basic_classifier.py#L93

My idea is to have that as a default but to be customizable.

> I'd be happier with this if we found a way to let a caller pass in a function to pull out the label or group.

I think ideally this should come from the `Instance` or the `Dataset`. It seems simpler to be specified by the `CrossValidator` or the function caller. If it's the function caller (`CrossValidateModel`), it can get a function as arg in `__init__`, but how to specify a `Callable` when loading from params (from a json file)? If it's just the function name, what if it's more complex, like a metadata field?

allennlp/training/cross_validation.py
bryant1410: Yes, but the epoch number is already like that. I think it's better to be consistent.

allennlp/training/cross_validation.py
bryant1410: Agree on not doing it in this PR. Wanted to open up the discussion. What if `BucketBatchSampler` receives a `Vocabulary` as an arg? That'd solve it.

bryant1410: Also, note `BucketBatchSampler` doesn't just accept any `Dataset`, as internally it assumes it gives `Instance`s.

allennlp/training/cross_validation.py
bryant1410: The abstraction is already there by subclassing (and using `run` and `finish`). It's just making it cleaner by making an interface out of it. What if I do it here and we see what do we do?

allennlp/training/cross_validation.py
matt-gardner: One option is to give a field name and have a function that pull that field out of the instance.  You could also take that a little bit farther, to handle metadata fields, or something.

Another option is to introduce an abstraction over these functions, which you can then create `FromParams`.  That seems way too heavy for this problem, though.

allennlp/training/cross_validation.py
matt-gardner: No, subclassing is definitely not the right thing to do.  You can implement the same methods, if you want, to make it easier to make an abstraction out of this in the future, but inheritance is the wrong relationship between these two objects.

matt-gardner: In general: inheritance is useful for polymorphism.  We make use of that quite a bit in allennlp.  But that's about the only place that inheritance should be used.  If you are trying to share code outside of polymorphism, there is basically always a better option.

bryant1410: Yeah, I completely agree, just describing how it is now.

bryant1410: See 52f0387

matt-gardner: Please don't add an abstraction in this PR.  Just create a `CrossValidateModel(Registrable)` class.  What abstraction, if any, should subsume those two commands needs a bunch of design thought, and we shouldn't just introduce something here that we very well might want to remove later.

allennlp/training/cross_validation.py
bryant1410: I think it *could* work in distributed training as it is with some changes I just added. Though I haven't tried it. So I'll leave it.

allennlp/commands/cross_validator.py
```diff
@@ -0,0 +1,164 @@
+from typing import Any, Iterator, Optional, Sequence, Tuple, Union
```
bryant1410: Not sure if this file should be here.

allennlp/training/cross_validation.py
bryant1410: I changed this. It just accepts a key name from params, but more flexible in the constructor.

As a separate issue, I was thinking that it'd be good for `Callable`s to be registered. They can be instantiated by creating a class that defines `__call__` (note that a class `C` that does so satisfies `isinstance(C, Callable)`). So, for example, here there could be a class that gets instantiated when a `Callable` arg gets assigned to a type like `instance_field_by_key` from params. Similarly for other `Callable` args.

Related, what about allowing any class to have a registry, not only `Registrable` subclasses? Instead of doing `Subclass.register(...)` to do like `Registrable.register(Class, ...)`. This can help `Callable`, and also maybe avoid intermediate classes such as `Optimizer`.

matt-gardner: It only really makes sense to register an abstract class if its subclasses are interchangeable when used downstream.  That's not true for `Callable`, because the arguments and return types of different `Callables` don't have to be the same.  So if you find yourself tempted to register something as general as a `Callable`, you should think about adding a more useful abstraction, or finding a different way to do what you want without adding so heavy as a new abstraction.

As for `Optimizer`, the main reason that exists is only so that we can use the type annotation in a constructor.  Doing `Registrable.register([something? "optimizer"?], ...)` defeats the entire purpose, because what do we put for the type annotation on the argument to the trainer?

bryant1410: > As for Optimizer, the main reason that exists is only so that we can use the type annotation in a constructor.

But we could, in theory, use PyTorch's `Optimizer` if we needn't subclass `Registrable`, right?

> because what do we put for the type annotation on the argument to the trainer?

We still put `Optimizer`. It'd be like `Registrable.register(Optimizer, ...)`.

matt-gardner: That was essentially what we did before I removed all of our custom `from_params` logic.  It wasn't exactly this, but it was close enough.  The trouble with that is that you have to do crazy magic in `from_params` that makes everything hard to use, because the objects we're trying to construct don't use `FromParams`, and don't have type annotations.

bryant1410: I see. If they had type annotations then it'd be easy? I think it'd be cool to add all subclasses of a certain type, supposing they have type annotations (e.g., all PyTorch optimizers under `Optimizer`, setting a default one, and setting names based on the class names).

matt-gardner: If they had type annotations on their argument, we'd have to change our `FromParams` logic a bit, but in principle we'd be able to accommodate constructing them using just `torch.nn.Optimizer` as a type annotation.

allennlp/training/cross_validation.py
dirkgr: Makes me think we should fix both, but not in this PR.

dirkgr:
So what's next for this thing? I think the changes look pretty good, but it's still missing tests and an implementation of `Subcommand`, right?

allennlp/training/cross_validation.py
bryant1410: Yeah. Consider it's also used when checkpointing (e.g., `metrics_epoch_N.json`). I'm mentioning it just to list more places.

And it's gonna be a breaking change.

bryant1410:
> So what's next for this thing? I think the changes look pretty good, but it's still missing tests and an implementation of `Subcommand`, right?

Still misses 1) tests, 2) docs, and 3) I also want to see the nested CV thing (and I wonder how to do CV on val but holdout test, another common config...).

About having an implementation of `Subcommand`, I don't think it's necessary. This is done as a `train` type now (by setting type: `cross_validation` in the config file, `train` recognizes it).

bryant1410:
I don't have an ETA. I'm gonna continue when I have some time.

dirkgr:
> About having an implementation of Subcommand, I don't think it's necessary. This is done as a train type now (by setting type: cross_validation in the config file, train recognizes it).

Oh, neat! I did not realize this.

matt-gardner:
@dirkgr, it's because of [this line](https://github.com/allenai/allennlp/pull/3960/files#diff-b9b3757fad64ec3657c5ce733ba0420bR23), which I'm a _little_ bit skeptical about, but is probably ok.  This was the argument for having a "training" abstraction.  Turns out you can get the same behavior for now, without actually adding the abstraction, just by registering the class in the right way.  This ability is because of how I refactored the train command to just call `TrainModel.from_params()`.

HarshTrivedi:
@bryant1410 This would be very useful! Are you planning to work on or finish this soon?

bryant1410:
@HarshTrivedi unfortunately I'm not planning on finishing it soon. But the core is working, you can use it.

Also, feel free to contribute to it.

PR #3963, opened 2020-03-20
Move Tensorboard logic into TensorboardWriter

matt-gardner:
Just moving things around, to have less logic inside of `Trainer`, and making it easier to customize what happens in tensorboard logging, if you want.

This fixes another part of #3913, specifically all of part 3 up to the "question".  I'm thinking at this point that the answer to the "question" should be to not worry about changing the tensorboard writer at all, but instead add two callbacks: a `BatchCallback`, and an `EpochCallback`, that are given appropriate objects that a person can do whatever they want with.  By default we do nothing.  This adds very little complexity to our code, but gives people a lot of flexibility to customize things.


matt-gardner:
Well, the point was to have them be _stated_ assumptions, not _unstated_ assumptions :).  So then in the unlikely event that something breaks in the future, the comments will hopefully give some help.

PR #3968, opened 2020-03-20
Move checkpointing logic into checkpointer, remove old callbacks

matt-gardner:
Fixes the second part of item 4 in #3913.  After this is merged, we just need to add a couple of simple callbacks (which are `None` by default) and we can close that issue.  When I made these changes, one of the callbacks that was still sitting around in the code broke, so I took the opportunity to remove all of the callbacks in this PR.  I'm going to add two simple callbacks, which don't need all of the baggage that the old callbacks had.

I'm not super happy that I had to add two methods to the `Trainer` to deal with the model/training state, instead of just one.  But because the moving average modifies model weights before and after checkpointing, I didn't see a better way around it.  If you have any good ideas, suggestions welcome.

This also adds back in some docstrings that got deleted in a PR a few weeks ago (I accidentally removed them from the trainer without adding them back to the checkpointer).


allennlp/training/checkpointer.py
```diff
@@ -21,29 +23,71 @@ class Checkpointer(Registrable):
     Dict[str, Any]), but they will be fed to `torch.save` so they should be serializable
     in that sense. They will also be restored as Dict[str, Any], which means the calling
     code is responsible for knowing what to do with them.
+
+    # Parameters
+
+    num_serialized_models_to_keep : `int`, optional (default=20)
+        Number of previous model checkpoints to retain.  Default is to keep 20 checkpoints.
```
dirkgr: Why is the default so high anyways? I've never been in a situation where I wanted the model from 19 iterations ago, but I have run out of disk space many times.

allennlp/training/checkpointer.py
```diff
@@ -21,29 +23,71 @@ class Checkpointer(Registrable):
...
+        epoch_str = f"{epoch}.{training_util.time_to_str(int(self._last_save_time))}"
+        self.save_checkpoint(epoch_str, trainer)
 
     def save_checkpoint(
         self,
         epoch: Union[int, str],
-        model_state: Dict[str, Any],
-        training_states: Dict[str, Any],
-        is_best_so_far: bool,
+        trainer: "allennlp.training.trainer.Trainer",
```
dirkgr: This type annotation is a string to avoid a circular dependency?

allennlp/training/trainer.py
```diff
@@ -716,22 +711,15 @@ def train(self) -> Dict[str, Any]:
...
-            The epoch of training.  If the checkpoint is saved in the middle
-            of an epoch, the parameter is a string with the epoch and timestamp.
-        """
-        # If moving averages are used for parameters, we save
-        # the moving average values into checkpoint, instead of the current values.
+    @overrides
+    def prep_state_for_checkpointing(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
         if self._moving_average is not None:
+            # Assigning average value to model parameters.  The checkpointer will call
+            # `restore_state_after_checkpointing` when it is done to put this back to what it was.
```
dirkgr: Should this be a context manager then?

allennlp/training/checkpointer.py
matt-gardner: Yeah, good point.  I'm going to change it to 2.

allennlp/training/checkpointer.py
matt-gardner: Yes.

allennlp/training/trainer.py
matt-gardner: I'm not sure how to actually do that.  Another wrinkle in here is that the checkpointer might decide to decline actually saving anything when the `save_checkpoint` method is called, and we don't want to do extra copies in that case.  Maybe you're talking about having something like this in the checkpointer?

```python
def save_checkpoint(..., trainer, ...):
    with trainer.get_checkpoint_state():
        # save stuff
    # all of the cleanup logic that's there
```

Yes, I think this could work, and I like it much better than what I did.  I've never actually written a context manager before, so I'll have to figure out how to do this, but it looks like it should work.

matt-gardner: Done, and yeah, I like this a whole lot better.  Thanks for the suggestion!

allennlp/training/checkpointer.py
```diff
@@ -45,7 +45,7 @@ def __init__(
         self,
         serialization_dir: str = None,
         keep_serialized_model_every_num_seconds: int = None,
-        num_serialized_models_to_keep: int = 20,
+        num_serialized_models_to_keep: int = 2,
```
dirkgr: Thanks for doing this :-)

Issue #3785, opened 2020-02-14
Fix file friendly logging


brendan-ai2:
@OyvindTafjord has identified a bug in our (formerly) file friendly logging that originated during the DistributedDataParallel additions. In short, `--file-friendly-logging` does not appear to work anymore.

To quote his Slack message:

> Regarding the lack of file-friendly logging: I'm guessing it's the distributed code (https://github.com/allenai/allennlp/pull/3529). As a workaround I hacked in the earlier code and that now works for my purposes (https://github.com/OyvindTafjord/allennlp/commit/8a1c7adaf78d3a1c2412312760b31192be936b0c). I also note that another change was that the stderr.log file for better or worse used to record these tqdm statuses, but with the new code that doesn't seem to happen (maybe that was intentional, but occasionally I have made use of those to get a post-hoc rough "training curve", although when using beaker those are also in the main beaker log).

schmmd:
@DeNeutoy any idea on this one?

dirkgr:
Should be fixed with #3519.

schmmd:
@dirkgr thanks--I have not had time to look at this yet...

matt-gardner:
@schmmd, we talked about this in our issue triage, and thought it should be covered by work I'm planning for #3519, so you shouldn't have to worry about this.

schmmd:
How would we validate that this is indeed fixed now that #3519 is merged?

DeNeutoy:
Now the issue that this is tracking with is #3913 , so it isn't fixed yet...

schmmd:
This is independent of the other PRs and needs attention.

matt-gardner:
Because there's some example code already that probably fixes the issue, I'm giving this one a "day" label.  Could be much less, if that example code just works without much effort, I just don't know.

PR #3970, opened 2020-03-21
Add batch and epoch callbacks to the trainer

### LINKED ISSUE #3218: Predict during training

### LINKED ISSUE #3913: Make the trainer more configurable

matt-gardner:
Fixes #3913 and fixes #3218.  This is NOT a fix for #3785, which I thought might be covered in the work for #3913, but it turns out was unrelated.

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -256,8 +256,8 @@ class GradientDescentTrainer(Trainer):
         momentum_scheduler: Optional[MomentumScheduler] = None,
         tensorboard_writer: TensorboardWriter = None,
         moving_average: Optional[MovingAverage] = None,
-        batch_callback: Optional[BatchCallback] = None,
-        epoch_callback: Optional[EpochCallback] = None,
+        batch_callbacks: List[BatchCallback] = None,
+        epoch_callbacks: List[EpochCallback] = None,
         distributed: bool = False,
         local_rank: int = 0,
         world_size: int = 1,
@@ -304,8 +304,8 @@ class GradientDescentTrainer(Trainer):
         self._learning_rate_scheduler = learning_rate_scheduler
         self._momentum_scheduler = momentum_scheduler
         self._moving_average = moving_average
-        self._batch_callback = batch_callback
-        self._epoch_callback = epoch_callback
+        self._batch_callbacks = batch_callbacks or []
+        self._epoch_callbacks = epoch_callbacks or []
 
         # We keep the total batch number as an instance variable because it
         # is used inside a closure for the hook which logs activations in
@@ -517,8 +517,8 @@ class GradientDescentTrainer(Trainer):
 
             if self._master:
                 self._checkpointer.maybe_save_checkpoint(self, epoch, batches_this_epoch)
-                if self._batch_callback:
-                    self._batch_callback(
+                for callback in self._batch_callbacks:
+                    callback(
                         self,
                         batch_group,
                         batch_group_outputs,
@@ -620,8 +620,8 @@ class GradientDescentTrainer(Trainer):
             val_generator_tqdm.set_description(description, refresh=False)
 
             if self._master:
-                if self._batch_callback:
-                    self._batch_callback(
+                for callback in self._batch_callbacks:
+                    callback(
                         self,
                         [batch],
                         [batch_outputs],
@@ -673,14 +673,12 @@ class GradientDescentTrainer(Trainer):
         for key, value in self._metric_tracker.best_epoch_metrics.items():
             metrics["best_validation_" + key] = value
 
-        if self._epoch_callback:
-            self._epoch_callback(self, metrics={}, epoch=-1)
+        for callback in self._epoch_callbacks:
+            callback(self, metrics={}, epoch=-1)
 
         for epoch in range(epoch_counter, self._num_epochs):
             epoch_start_time = time.time()
             train_metrics = self._train_epoch(epoch)
-            if self._epoch_callback:
-                self._epoch_callback(self, metrics=train_metrics, epoch=epoch)
 
             # get peak of memory usage
             if "cpu_memory_MB" in train_metrics:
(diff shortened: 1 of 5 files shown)


allennlp/training/trainer.py
```diff
@@ -126,6 +160,8 @@ def __init__(
         momentum_scheduler: Optional[MomentumScheduler] = None,
         tensorboard_writer: TensorboardWriter = None,
         moving_average: Optional[MovingAverage] = None,
+        batch_callback: Optional[BatchCallback] = None,
+        epoch_callback: Optional[EpochCallback] = None,
```
dirkgr: You could make these `List`s instead of `Optional`, and call all of them. If we're going the callback route, we might as well go all in.

matt-gardner: Note that we are not providing any implementation of a callback here.  Anything a person wants to do will be code they write.  If they want to have several things done, they can just put those things in a callback.  If they want to have composable things in their callbacks, they have modularize it however they want (including a callback that calls a list of functions).

dirkgr: I can imagine a few orthogonal, default callbacks we could provide. Can we implement `TensorboardWriter` as one of these? Writing predictions to a file? Automatic logging to W&B?

The advantage of using callbacks is configurability and composability. It would be a neat feature if you can just add what you need to this list of callbacks.

matt-gardner: Fair enough, you make a good point.  I don't think we can pull the tensorboard writer out into a callback, because of how computing gradient updates works, but you're right that you could imagine some built-in ones that get added later, and being able to pass in a list makes sense.  It was easy enough to change.

mahnerak:
Is there any reason why are there `BatchCallback` and `EpochCallback` instead of a single `Callback` class that could handle both? (+a callback at the end of the training, as it is done in Keras, PyTorch-Lightning, and was done for CallbackTrainer)

Say, I have a custom logger which tracks both epoch and batch. What's the recommended way to share an open connection/file (and when to close it when finished)?

dirkgr:
This is an interesting scenario. As a workaround, you can stick stuff into the trainer that you want to share between the two. It's not super elegant, but it will work.

`BatchCallback` and `EpochCallback` receive different parameters, so that's why they are separate. But I see your point. Maybe in version 2.0 we will add a `TrainerCallback` that can do both, to address this scenario better.

PR #3971, opened 2020-03-22
Allow CRF to compute loss just for single span prediction

matt-gardner:
I finally figured out why our NER interpret demo never worked as well as the original implementation that was used when @Eric-Wallace created his video.  Somehow the loss masking code never got checked in, so we were computing the loss for the whole sequence.  As that includes "O" tags that were intentionally changed away from predicted NER tags, the loss for those tags was quite high, and dominated the loss computation, so the result was basically the opposite of what we wanted.

This PR adds the ability to mask the loss computation in our CRF model, so that the loss is only calculated on non-O tags.  In order to set this flag at interpretation time, but not any other time, I added a `FlagField` that batches as a single (in this case boolean) value that gets passed to the model, and I added that field to the instances in the NER predictor.  I verified manually that this gives basically the same result with our current NER model and code as we got in our original paper and video.

This means that only models that accept this flag (`ignore_loss_on_o_tags`) now work with the `SentenceTaggerPredictor` (when doing interpretations; this doesn't affect anything else), but this is actually an improvement, because previously models that didn't use this flag were just silently broken.  Now you'll just crash.

### DIFF
--- a/allennlp/data/fields/flag_field.py
+++ b/allennlp/data/fields/flag_field.py
@@ -1,0 +1,43 @@
+from typing import Any, Dict, List
+
+from overrides import overrides
+
+from allennlp.data.fields.field import Field
+
+
+class FlagField(Field[Any]):
+    """
+    A class representing a flag, which must be constant across all instances in a batch.
+    This will be passed to a `forward` method as a single value of whatever type you pass in.
+
+    This is intended for use with boolean flags, but in principle it should also work with any other
+    type.
+    """
+
+    def __init__(self, flag_value: Any) -> None:
+        self.flag_value = flag_value
+
+    @overrides
+    def get_padding_lengths(self) -> Dict[str, int]:
+        return {}
+
+    @overrides
+    def as_tensor(self, padding_lengths: Dict[str, int]) -> Any:
+        return self.flag_value
+
+    @overrides
+    def empty_field(self):
+        # Because this has to be constant across all instances in a batch, we need to keep the same
+        # value.
+        return FlagField(self.flag_value)
+
+    def __str__(self) -> str:
+        return f"FlagField({self.flag_value})"
+
+    @overrides
+    def batch_tensors(self, tensor_list: List[Any]) -> Any:
+        if len(set(tensor_list)) != 1:
+            raise ValueError(
+                f"Got different values in a FlagField when trying to batch them: {tensor_list}"
+            )
+        return tensor_list[0]
--- a/allennlp/data/fields/flag_field.py
+++ b/allennlp/data/fields/flag_field.py
@@ -9,9 +9,6 @@ class FlagField(Field[Any]):
     """
     A class representing a flag, which must be constant across all instances in a batch.
     This will be passed to a `forward` method as a single value of whatever type you pass in.
-
-    This is intended for use with boolean flags, but in principle it should also work with any other
-    type.
     """
 
     def __init__(self, flag_value: Any) -> None:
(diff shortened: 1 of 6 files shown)


allennlp/data/fields/flag_field.py
```diff
@@ -0,0 +1,43 @@
...
+from allennlp.data.fields.field import Field
+
+
+class FlagField(Field[Any]):
+    """
+    A class representing a flag, which must be constant across all instances in a batch.
+    This will be passed to a `forward` method as a single value of whatever type you pass in.
+
+    This is intended for use with boolean flags, but in principle it should also work with any other
+    type.
```
dirkgr: This seems useful in more cases than just boolean flags.

Now, if you need a setting to be consistent between reader and model, you can put it into a `FlagField`. Before, you had to specify it twice in the config, giving opportunity to mess it up.

allennlp/models/crf_tagger.py
```diff
@@ -223,8 +228,13 @@ def forward(
             output["top_k_tags"] = best_paths
 
         if tags is not None:
+            if ignore_loss_on_o_tags:
+                o_tag_index = self.vocab.get_token_index("O", namespace=self.label_namespace)
```
dirkgr: Are `"O"` tags a general-enough concept to hard-code it like this?

I'd rather have a flag named `ignore_loss_tags` that's a set of strings. Then you can use it to tag anything.

matt-gardner: Yes, I believe "O" is general enough for this.  Any flavor of BIO tagging with have this tag, and you really only want this kind of behavior when you have a span prediction model that uses BIO tags.

allennlp/data/fields/flag_field.py
matt-gardner: Well, this specifically goes to `forward`, where things that you specify in a model config go to the model's constructor.

You're suggesting making the comment here less prescriptive?

dirkgr: Yes

allennlp/models/crf_tagger.py
dirkgr: You would never use this for POS tagging or some other tagging that's not span based?

matt-gardner: It's hard for me to imagine a situation where that's useful.  And even in a POS tagging scenario, a list of strings wouldn't solve the problem, because if there are two matching POS tags in the sentence, you couldn't exclude just one of them.  That approach needs a different solution, and it's a use case that I have a hard time imagining anyway, so it didn't seem like a good idea to design for it.

allennlp/data/fields/flag_field.py
matt-gardner: I couldn't think of anything useful to say that wasn't as prescriptive, so I just removed the line entirely.

allennlp/models/crf_tagger.py
dirkgr: The way I'm really coming at it from is that this component is called "CRF Tagger", but we're putting in code that only works when you use it as a "CRF Span Tagger" (using magic strings that you just have to know about, no less). The examples are just me trying to boil down my unease to something specific.

It seems pretty arbitrary to give the string `"O"` different handling from all other strings, doesn't it?

I guess I can agree to it by thinking of it as something that may help you in a common case, and you can otherwise ignore it 🤷. I just worry that it becomes a maintenance problem when we discover people actually do want to tag things that aren't spans.

matt-gardner: https://github.com/allenai/allennlp/blob/master/allennlp/models/crf_tagger.py#L39-L51

All of this code is specific to span tagging.  I'm just adding one more flag that's specific to span tagging, which you can ignore if you're not doing span tagging.  And all of the supported label encodings assign specific meaning to "O", so this isn't exactly arbitrary.  The flag itself also gives the string that's used in its variable name.

dirkgr: :+1:

PR #3975, opened 2020-03-22
A few misc doc fixes

matt-gardner:
I was poking around at the docs this morning, and found a ton of things that aren't rendering great.  I opened issues for a couple of them.  This PR has one-off fixes for a few files.  In particular, I found that for things to render nicely, we need to reformat all of our `Parameters` blocks to what you see in the `params.py` changes in this PR.  Here's a visual comparison:

Before:
Screen Shot 2020-03-22 at 9 27 13 AM

After:
Screen Shot 2020-03-22 at 9 27 19 AM

Unfortunately, it looks to me like we really should just manually go through all of our files and fix these little formatting issues.  I'm not sure there's a better way.


matt-gardner:
The module-level docs are not currently rendering very nicely (they mess things up in the UI), and these particular ones either did not seem to add anything useful (`params.py`) or were just woefully out of date (`from_params.py`), so I just removed them.

PR #3976, opened 2020-03-22
Fix mistake on tutorial

urigoren:
Fix `ValueError: too many values to unpack (expected 2)` when trying to run the example on:

https://allennlp.org/tutorials


matt-gardner:
The whole tutorial is very out of date, and will be replaced soon with something a lot better.  I'd accept the PR, but it's actually still wrong if you're using master code, so there isn't a lot of point.  Thanks for finding it, though!

PR #3981, opened 2020-03-24
Add __len__ to FlagField

matt-gardner:
I didn't add this method when I first implemented `FlagField`, because I didn't know what it was for and thought it was unnecessary.  Then I looked at the bucket sampler code and realized that we needed this.  I'm a bit unhappy that our CI didn't warn me about an unimplemented abstract method - pylint would have done that.  Oh well.


bryant1410:
> pylint would have done that.

What about adding it just for a few things?

PR #3985, opened 2020-03-25
Shuffle batches in bucket sampler

matt-gardner:
In writing a chapter for the allennlp course, I trained a model and discovered that the batches were not shuffled when using the `BucketBatchSampler`, so the model was seeing shortest instances first, then longer instances.  This isn't great from a learning standpoint.  This PR fixes that issue.


matt-gardner:
Probably only in tests.

PR #3992, opened 2020-03-26
Persist/load amp state during training

### LINKED ISSUE #3875: Properly checkpoint models trained with Apex

#3866 enables mixed precision training with Apex, but pushes the changes required to properly save and load a model trained with `amp` down the line. As discussed, I am simply opening this issue so I don't forget and will (hopefully) make a PR in the next couple of days.

Follow the instructions in the Apex docs to checkpoint a model trained with `amp`.

**Additional context**

[Apex docs](https://github.com/NVIDIA/apex#checkpointing).

JohnGiorgi:
# Overview

This is a small modification to `GradientDescentTrainer` that persists the `amp.state_dict` when checkpointing the model and loads it when recovering training state from a checkpoint.

WIP.

## TODO

- [ ] Add tests (help?)
- [x] `Model._load()` should probably call `amp.initialize`, otherwise, a model trained with `amp` might be loaded without it (say, during prediction) which will hurt performance (I think? there would be information loss afaik).
- [x] Related to the point above, it is possible that a user will try to load a model that was trained with `amp`, but `amp` is not installed (i.e. `amp = None`). Should there be a warning or even error in this case?

## Closes

Closes #3875.

### DIFF
--- a/allennlp/__main__.py
+++ b/allennlp/__main__.py
@@ -6,7 +6,8 @@ import sys
 if os.environ.get("ALLENNLP_DEBUG"):
     LEVEL = logging.DEBUG
 else:
-    LEVEL = logging.INFO
+    level_name = os.environ.get("ALLENNLP_LOG_LEVEL")
+    LEVEL = logging._nameToLevel.get(level_name, logging.INFO)
 
 sys.path.insert(0, os.path.dirname(os.path.abspath(os.path.join(__file__, os.pardir))))
 logging.basicConfig(format="%(asctime)s - %(levelname)s - %(name)s - %(message)s", level=LEVEL)
--- a/allennlp/common/util.py
+++ b/allennlp/common/util.py
@@ -330,7 +330,8 @@ def prepare_global_logging(
     if os.environ.get("ALLENNLP_DEBUG"):
         LEVEL = logging.DEBUG
     else:
-        LEVEL = logging.INFO
+        level_name = os.environ.get("ALLENNLP_LOG_LEVEL")
+        LEVEL = logging._nameToLevel.get(level_name, logging.INFO)
 
     if rank == 0:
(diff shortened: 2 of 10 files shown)


dirkgr:
> Related to the point above, it is possible that a user will try to load a model that was trained with `amp`, but `amp` is not installed (i.e. `amp = None`). Should there be a warning or even error in this case?

Is that a problem? It will just ignore the `amp` state and train anyways, right? If it's not a problem, we don't have to do anything about it.

JohnGiorgi:
> > Related to the point above, it is possible that a user will try to load a model that was trained with `amp`, but `amp` is not installed (i.e. `amp = None`). Should there be a warning or even error in this case?
> 
> Is that a problem? It will just ignore the `amp` state and train anyways, right? If it's not a problem, we don't have to do anything about it.

I guess it is only a problem if, when loading a model trained with mixed precision, _not_ calling `amp.initialize()` impacts model performance in some way. I actually don't know the answer to this

dirkgr:
If it impacts model performance, there should be at least a warning, and maybe more. This is the kind of thing that you might never discover and your model silently underperforms. That's important to avoid.

JohnGiorgi:
> If it impacts model performance, there should be at least a warning, and maybe more. This is the kind of thing that you might never discover and your model silently underperforms. That's important to avoid.

Right, that is my worry as well. I just don't know when or if we expect a model trained with amp but loaded without it (or vice verse) to underperform -- maybe someone who knows more about mixed precision training can chime in.

Should `Model._load()` at least log a warning of some sort so this PR can get merged?

dirkgr:
This is going to be a rare case, so let's not hold up the merge with it. Let's just put a warning in, and we can remove it later if we find it's not important.

PR #3993, opened 2020-03-26
Find the vocab even in Albert

dirkgr:
Albert uses a different way of storing the vocab, so we need a different way of getting it out and filling our own `Vocab` object.

### DIFF
--- a/allennlp/data/token_indexers/pretrained_transformer_indexer.py
+++ b/allennlp/data/token_indexers/pretrained_transformer_indexer.py
@@ -70,22 +70,30 @@ class PretrainedTransformerIndexer(TokenIndexer):
         if self._added_to_vocabulary:
             return
 
-        vocab_field_name = None
-        if hasattr(self._tokenizer, "vocab"):
-            vocab_field_name = "vocab"
-        elif hasattr(self._tokenizer, "encoder"):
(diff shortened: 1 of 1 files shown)


allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -70,22 +70,30 @@ def _add_encoding_to_vocabulary_if_needed(self, vocab: Vocabulary) -> None:
...
-        vocab_field_name = None
-        if hasattr(self._tokenizer, "vocab"):
-            vocab_field_name = "vocab"
-        elif hasattr(self._tokenizer, "encoder"):
-            vocab_field_name = "encoder"
+        # Find the key under which the vocab is hiding.
+        for vocab_field_name in ["vocab", "encoder", "sp_model"]:
+            if hasattr(self._tokenizer, vocab_field_name):
+                break
+        else:
```
ZhaofengWu: This for-else might be a little weird. Maybe something like
```
a = None
for b in list:
  if cond:
    a = x
```

dirkgr: Then I'd need an extra variable.

PR #3999, opened 2020-03-28
Pad empty text fields

dirkgr:


allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -108,9 +108,11 @@ def tokens_to_indices(self, tokens: List[Token], vocabulary: Vocabulary) -> Inde
 
         indices, type_ids = self._extract_token_and_type_ids(tokens)
         # The mask has 1 for real tokens and 0 for padding tokens. Only real tokens are attended to.
-        output: IndexedTokenList = {"token_ids": indices, "mask": [True] * len(indices)}
-        if type_ids is not None:
-            output["type_ids"] = type_ids
+        output: IndexedTokenList = {
+            "token_ids": indices,
+            "mask": [True] * len(indices),
+            "type_ids": type_ids,
```
matt-gardner: I assume you checked that this still works with models that don't use type ids?

dirkgr: I didn't explicitly check it. I trust the tests. How do you think it could break?

matt-gardner: Are there huggingface models that don't accept this argument?  I don't know.  If there are, then this would break.  We don't test all huggingface models, so the tests wouldn't necessarily catch this.

dirkgr: That's a fair point, but we have this line in the embedder: https://github.com/allenai/allennlp/blob/master/allennlp/modules/token_embedders/pretrained_transformer_embedder.py#L95

If the type ids are all zero, it doesn't supply them to the model.

Then why put in all zeros here? Because this happens pre-batching, and just because they are all zero here doesn't mean they won't get batched together with some other instances where they are not all zero.

matt-gardner: :+1:

Issue #3630, opened 2020-01-15
Updates to the docs


DeNeutoy:
#3602 switched our docs from restructured text to markdown, which is a big improvement. However, there are some left over traces of rst formatting in the docstrings. It would be great if we could comb through these and update them.

- [ ] commands
- [x] common
- [ ] data
- [x] interpret
- [x] models
- [x] modules
- [x] nn
- [x] predictors
- [x] tools
- [x] training

# How to help

If you're interested in helping out, just do the below steps and send a PR, mentioning this issue with the section you have completed.

## Class and Func references

`:class:` and `:func:` references need to be changed to relative links. e.g, we previously had:

`:class:~allennlp.training.trainer.Trainer`, which needs to be updated to a link to a `.md` file, relative to the allennlp directory structure. So if this reference was in `allennlp/modules/encoder_base.py`, the correct replacement link would look like:

```
[Trainer](../training/trainer.md)
```

## External URLS

External URLS need to be switched to the markdown format, e.g `[link text](url)`. I've done a lot of these, but not all of them.

## Types in function/class parameters

Types in function parameters now only need to be surrounded by single backticks, rather than two. e.g ` ``int`` ` -> ` `int` `

## Adding code examples

For complicated or hard to use code, it would be nice to have some examples added to docstrings using the triple backtick formatting. These will render nicely in our docs.

## Admonition

[Admonition](https://squidfunk.github.io/mkdocs-material/extensions/admonition/) is a markdown extension which the docs support, which allow fancy rendering of important blocks/notes/warnings in docs.

The syntax for this is:

```
!!! note
	this is a note
```

```
??? note
	this is a note with the main text hidden in a dropdown
```

Anything relevant, like TODOs, important notes, or just anything that is critical to _not miss_ whilst reading a docstring could use these.

DeNeutoy:
#3642 has done most of the double backtick cleanup for all directories.

RyoMazda:
I love the documentation style a lot, especially the API part where we can see the source codes for the parameters of the functions and the classes!

I'm wondering if you guys have any plans for making this docs-functionality into a mkdocs plugin or a python package independent of AllenNLP?

If no, am I allowed to take the code like `scripts/py2md.py` and put it on the public repository of mine (of course I'm gonna refer appropriately in that case)?

epwalsh:
Hi @RyoMazda, I think at some point it might make sense to pull `py2md` out into it's own repo, but it's not a priority at the moment.

[...]

PR #4004, opened 2020-03-30
Fix some external url formatting.

mklimasz:
Addresses some parts of #3630.

External urls in parameters sections are not rendering correctly (i.e. http://docs.allennlp.org/master/api/modules/stacked_alternating_lstm/). I have used reference style links (https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links) to avoid that.

Also fixed some broken links in "regular" places.

### DIFF
--- a/allennlp/modules/stacked_alternating_lstm.py
+++ b/allennlp/modules/stacked_alternating_lstm.py
@@ -18,7 +18,7 @@ class StackedAlternatingLstm(torch.nn.Module):
     the sequence and going backwards. This implementation is based on the
     description in [Deep Semantic Role Labelling - What works and what's next][0].
 
-    [0]: https://homes.cs.washington.edu/~luheng/files/acl2017_hllz.pdf
+    [0]: https://www.aclweb.org/anthology/P17-1044.pdf
     [1]: https://arxiv.org/abs/1512.05287
 
     # Parameters
--- a/allennlp/modules/stacked_alternating_lstm.py
+++ b/allennlp/modules/stacked_alternating_lstm.py
@@ -16,8 +16,10 @@ class StackedAlternatingLstm(torch.nn.Module):
     """
     A stacked LSTM with LSTM layers which alternate between going forwards over
     the sequence and going backwards. This implementation is based on the
-    description in [Deep Semantic Role Labelling - What works and what's next]
-    (https://homes.cs.washington.edu/~luheng/files/acl2017_hllz.pdf).
+    description in [Deep Semantic Role Labelling - What works and what's next][0].
+
+    [0]: https://homes.cs.washington.edu/~luheng/files/acl2017_hllz.pdf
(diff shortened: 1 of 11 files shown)


allennlp/modules/stacked_alternating_lstm.py
```diff
@@ -16,8 +16,10 @@ class StackedAlternatingLstm(torch.nn.Module):
     """
     A stacked LSTM with LSTM layers which alternate between going forwards over
     the sequence and going backwards. This implementation is based on the
-    description in [Deep Semantic Role Labelling - What works and what's next]
-    (https://homes.cs.washington.edu/~luheng/files/acl2017_hllz.pdf).
+    description in [Deep Semantic Role Labelling - What works and what's next][0].
+
+    [0]: https://homes.cs.washington.edu/~luheng/files/acl2017_hllz.pdf
```
dirkgr, line 21: Can you update this link to point to https://www.aclweb.org/anthology/P17-1044.pdf? The current one is dead.

mklimasz: Done, but doesn't it mean that script for url checking is not working (scripts/check_links.py) - masking it is worth making an issue?

dirkgr: Hm, that script actually fails when I run it, but due to a different error. If it's something small, I'll just fix it right here.

dirkgr: It fails because of other cases of the same problem. I fixed that just now, and added a commit.

dirkgr: Now, the `check_links.py` script fails because GitHub is sending out invalid headers. The `urllib` people think they can [wage a Twitter war to get it fixed on the GitHub end](https://github.com/urllib3/urllib3/issues/1822). Best not to get involved.

dirkgr:
The stupid linter is going to be the end of me.

allennlp/modules/stacked_alternating_lstm.py
mklimasz: Wow, that's very optimistic I guess.

pquentin: > Now, the `check_links.py` script fails because GitHub is sending out invalid headers. The `urllib` people think they can [wage a Twitter war to get it fixed on the GitHub end](https://github.com/urllib3/urllib3/issues/1822). Best not to get involved.

For what it's worth the person that suggested taking it to Twitter is not a maintainer, and never mentioned a war, just asked for a fix. Looks like it worked according to @andriyor.

Thanks for AllenNLP!

dirkgr: Looks like I should give GitHub (and @ssbarnea) more credit :-)

Thanks for urllib!

Issue #2300, opened 2019-01-07
Feature request: separate logging for model computed loss and regularization loss in tensorboard


matt-peters:
It would be nice to separately log model computed loss from regularization loss in tensorboard. Involves minor changes to the Trainer.

tabshaikh:
@matt-gardner would like to get started on this issue any insights on where to begin :)

mklimasz:
#4008 PR adding regularization loss as separate metric (:

dirkgr:
This is done, no?

PR #4008, opened 2020-03-31
Add separate logging for regularization loss.

mklimasz:
As #2300 was tagged with ``Good First Issue``  - there is PR extracting regularization loss to separate metric (also checked in unit test).

I wasn't able to come up with better idea than wrapping ``get_regularization_penalty`` value in ``batch_outputs`` (as there isn't any device information in model).

I'm happy to address issues.

### DIFF
--- a/allennlp/tests/models/simple_tagger_test.py
+++ b/allennlp/tests/models/simple_tagger_test.py
@@ -129,7 +129,11 @@ class SimpleTaggerRegularizationTest(ModelTestCase):
         training_batch = next(iter(self.data_loader))
         validation_batch = next(iter(self.data_loader))
 
-        training_loss = self.trainer.batch_outputs(training_batch, for_training=True)["loss"].data
+        training_batch_outputs = self.trainer.batch_outputs(training_batch, for_training=True)
+        training_loss = training_batch_outputs["loss"].data
+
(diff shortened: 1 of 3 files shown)


maksym-del:
Hi @mklimasz and @dirkgr, thanks for the PR!

I am training a model with the latest master, and now I get "reg_loss" as an additional metric automatically with the value equals 0.0 which is saved to my metrics and is shown with tqdn. (see screenshot below)

Probably we should only print ```reg_loss``` metric when it is actually computing something...

Another issue, probably unrelated, (but it is also on the image below), is that the training progress for some reason is not printed to file. The screenshot below shows stdout.log output:
image

maksym-del:
Hi @mklimasz and @dirkgr, thanks for the PR!

I am training a model with the latest master, and now I get "reg_loss" as an additional metric automatically with the value equals 0.0 which is saved to my metrics and is shown with tqdn.

Probably we should only print reg_loss metric when it actually computes something...

Another issue (probably unrelated) is that the training progress for some reason is not printed to file.

dirkgr:
> Another issue (probably unrelated) is that the training progress for some reason is not printed to file.

I hope that #3785 will fix that?

PR #4018, opened 2020-04-04
New version of transformers

dirkgr:


allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
```diff
@@ -258,7 +269,7 @@ def test_token_idx_sentence_pairs(self):
             ".",
             "</s>",
             "</s>",
-            "It",
+            "ĠIt",
```
matt-gardner, line 272: Do you know why it's adding spaces to the first token?  Is that expected?

setup.py
```diff
@@ -119,7 +119,7 @@
         "flaky",
         "responses>=0.7",
         "conllu==2.3.2",
-        "transformers>=2.4.0,<2.5.0",
+        "transformers>=2.6.0",
```
matt-gardner, line 122: Are you sure you don't want to keep an upper bound?

allennlp/tests/data/tokenizers/pretrained_transformer_tokenizer_test.py
dirkgr: I think it's a change in defaults.

At first, people thought that `Ġ` means that this is the beginning of the word (opposite of `##` in BERT). So then people were complaining about the "bug" where the `Ġ` wasn't there at the beginning of the first token, and they made a fix. But now it means "space", so not having it makes sense ... I want to not get involved and just stick with the huggingface default.

setup.py
dirkgr: Probably a good idea. Added in 3264dcc.

Issue #4012, opened 2020-04-01
`batches per epoch` feature


maksym-del:
**Is your feature request related to a problem? Please describe.**
Since allennlp's data iterators are gone, a useful feature to specify the epoch size is gone as well.

**Describe the solution you'd like**
Possibility to specify this parameter somewhere.

dirkgr:
All the `DatasetReader`s take a `max_instances` parameter that makes them read only the first `n` instances. Does that do what you want?

maksym-del:
@dirkgr the idea is to still pass through all the instances in the dataset. We just take a chunk of instances and decide that the epoch ends after this many instances.

DataIterator did exactly that:
https://github.com/allenai/allennlp/blob/8dc3a5b45f5b6441ca0acbf225a8d259522f02e3/allennlp/data/iterators/data_iterator.py#L64 
https://github.com/allenai/allennlp/blob/8dc3a5b45f5b6441ca0acbf225a8d259522f02e3/allennlp/data/iterators/data_iterator.py#L209
https://github.com/allenai/allennlp/blob/8dc3a5b45f5b6441ca0acbf225a8d259522f02e3/allennlp/data/iterators/data_iterator.py#L165

dirkgr:
Do you want to do this so that you can run validation more often, and maybe early-stop in the middle of an epoch?

matt-gardner:
I think the primary use case is to have the notion of an "epoch" (and whatever learning rate updates that entails) be decoupled from the actual size of your data, perhaps because your dataset is very large.  I'm not totally sure how to accomplish this in the new paradigm.  If we want to avoid modifying the trainer to handle this, then we'd need some kind of sampler that retains state.

dirkgr:
I would want to stick with the definition of "epoch" being one run through the data. But it makes sense to me that you might want to do certain things more frequently than once per epoch.

matt-gardner:
That definition really doesn't make sense in some cases, particularly for language modeling.  See, e.g., section 3.1.2 of the T5 paper: https://arxiv.org/pdf/1910.10683.pdf.  We use "epoch" to determine when checkpoints and validation happen, which really doesn't have to be after one pass through the data.  Enforcing that those have to align is pretty restrictive, and unnecessarily so.

maksym-del:
[...]

PR #4021, opened 2020-04-04
Add ability to specify custom epoch size

maksym-del:
Should fix https://github.com/allenai/allennlp/issues/4012


allennlp/training/trainer.py
```diff
@@ -242,6 +266,10 @@ class GradientDescentTrainer(Trainer):
         precision training. Must be a choice of `"O0"`, `"O1"`, `"O2"`, or `"O3"`.
         See the Apex [documentation](https://nvidia.github.io/apex/amp.html#opt-levels-and-properties) for
         more details. If `None`, Amp is not used. Defaults to `None`.
+    batches_per_epoch: `int`, optional, (default = `None`)
```
matt-gardner, line 269: We are trying to minimize the parameters that need to passed directly to the trainer.  I'm pretty sure you don't even need this, as you can just specify a `DataLoader`.  Can you make your data loader class inherit from our `DataLoader`, and register it as something?  In this PR, I'd like to see only modifications to data code, with no changes necessary to the trainer at all.

allennlp/training/trainer.py
```diff
@@ -140,6 +140,30 @@ def __call__(
         raise NotImplementedError
 
 
+class CustomlySizedEpochDataLoader:
```
matt-gardner, line 143: Can you move this inside `dataloader.py`?  Also, maybe name it `CustomEpochSizeDataLoader`?

allennlp/training/trainer.py
maksym-del: How do you like if I just add this optional parameter to existing Allennlp ```DataLoader```, so that default custom ```collate_fn``` is not the only reason we subclass it?

matt-gardner: Sure, that seems reasonable enough to me.

maksym-del: In this case, instead of specifying the new data loader with "type" keyword, just in order to be able to add a single parameter to the config file, the users will just get it as an additional field for free.

allennlp/data/dataloader.py
```diff
@@ -36,6 +36,10 @@ class DataLoader(Registrable, data.DataLoader):
     # Instances which have _already_ been indexed.
     my_loader = DataLoader(dataset, batch_size=32, collate_fn=allennlp_collate)
     ```
+
+    Secondly, this class adds `batches_per_epoch` parameter which defines the epoch size in GPU batches
+    (as opposite to the effective batches obtained with Gradient Accumulation).
+    `None` value (default) means that the epoch size will equal the full pass through the data.
```
matt-gardner, line 42:
```suggestion
    Secondly, this class adds a `batches_per_epoch` parameter which, if given, determines the number
    of batches after which an epoch ends.  If this is `None`, then an epoch is set to be one full pass
    through your data.  You might use this if you have a very large dataset and want more frequent
    checkpoints and evaluations on validation data, for instance.
```

allennlp/tests/training/trainer_test.py
```diff
@@ -139,6 +152,59 @@ def test_passing_trainer_multiple_gpus_raises_error(self):
                 self.model, self.optimizer, self.data_loader, num_epochs=2, cuda_device=[0, 1],
             )
 
+    def test_trainer_respects_epoch_size_equals_total(self):
+        num_epochs = 3
+        batches_per_epoch = len(self.data_loader_equal_epoch)
```
matt-gardner, line 157: All of these tests are identical and aren't really testing what you want, because there's nothing actually enforcing that the data loader did what you expected.  You're just reading a property here and are using this value in an assert below, in a way that should always be true of the trainer and is already tested.  I'd expect something like `assert batches_per_epoch == 3` here (or whatever the value should be).  Same for the tests below.

allennlp/training/trainer.py
```diff
@@ -277,6 +277,7 @@ def __init__(
         self.model = model
 
         self.data_loader = data_loader
+
```
matt-gardner, line 280: Can you just revert all changes to this file?

PR #4022, opened 2020-04-04
Fix amp initialize issues in Model._load

JohnGiorgi:
There were to issues in my last PR that threw errors when loading a model trained with amp:

-  a misspelling of the `amp.initialize` method in my last PR (#3992). This caused an error when trying to load a model trained with `amp` (I also found the word initialize misspelled in a docstring so I fixed that as well).
- the code that moves the loaded model to GPU was moved to before the call to `amp.initialize`, otherwise, it will throw an error.

### DIFF
--- a/allennlp/models/model.py
+++ b/allennlp/models/model.py
@@ -284,6 +284,13 @@ class Model(torch.nn.Module, Registrable):
         remove_pretrained_embedding_params(model_params)
         model = Model.from_params(vocab=vocab, params=model_params)
 
+        # Force model to cpu or gpu, as appropriate, to make sure that the embeddings are
+        # in sync with the weights
+        if cuda_device >= 0:
+            model.cuda(cuda_device)
+        else:
+            model.cpu()
+
         # If the model was trained with amp and amp is available, we should re-initialize it with
         # the opt_level that was used. If the model was trained with amp but amp is not availble, log a warning
         # so this doesn't pass silently.
@@ -292,7 +299,7 @@ class Model(torch.nn.Module, Registrable):
                 logger.warning(
                     (
                         f"This model was trained with amp (opt_level: {opt_level}) but amp is not available."
-                        " Any further training will be full-precision."
+                        " Any further training or inference will happen at full-precision."
                     )
                 )
             else:
@@ -308,13 +315,6 @@ class Model(torch.nn.Module, Registrable):
 
         model_state = torch.load(weights_file, map_location=util.device_mapping(cuda_device))
         model.load_state_dict(model_state)
-
-        # Force model to cpu or gpu, as appropriate, to make sure that the embeddings are
-        # in sync with the weights
-        if cuda_device >= 0:
-            model.cuda(cuda_device)
-        else:
-            model.cpu()
 
         return model
 
--- a/allennlp/models/model.py
+++ b/allennlp/models/model.py
@@ -284,7 +284,7 @@ class Model(torch.nn.Module, Registrable):
         remove_pretrained_embedding_params(model_params)
         model = Model.from_params(vocab=vocab, params=model_params)
 
-        # If the model was trained with amp and amp is available, we should re-intialize it with
+        # If the model was trained with amp and amp is available, we should re-initialize it with
         # the opt_level that was used. If the model was trained with amp but amp is not availble, log a warning
         # so this doesn't pass silently.
         if opt_level is not None:
@@ -296,7 +296,7 @@ class Model(torch.nn.Module, Registrable):
                     )
                 )
             else:
-                model = amp.intialize(model, opt_level=opt_level)
+                model = amp.initialize(model, opt_level=opt_level)
 
         # If vocab+embedding extension was done, the model initialized from from_params
         # and one defined by state dict in weights_file might not have same embedding shapes.
--- a/allennlp/modules/token_embedders/embedding.py
+++ b/allennlp/modules/token_embedders/embedding.py
@@ -75,7 +75,7 @@ class Embedding(TokenEmbedder):
         construct it in the original training. We store vocab_namespace used during the original
         training as an attribute, so that it can be retrieved during fine-tuning.
     pretrained_file : `str`, (optional, default=None)
-        Path to a file of word vectors to intialize the embedding matrix. It can be the
+        Path to a file of word vectors to initialize the embedding matrix. It can be the
         path to a local file or a URL of a (cached) remote file. Two formats are supported:
             * hdf5 file - containing an embedding matrix in the form of a torch.Tensor;
             * text file - an utf-8 encoded text file with space separated fields.


JohnGiorgi:
Sorry, should have noted that this is WIP. I still can't get my setup to load a model trained with amp.

Here is the relevant code in `Model._load`

```python
        remove_pretrained_embedding_params(model_params)
        model = Model.from_params(vocab=vocab, params=model_params)

        model.extend_embedder_vocab()

        if cuda_device >= 0:
            model.cuda(cuda_device)
        else:
            model.cpu()

        if opt_level is not None:
            if amp is None:
                logger.warning(
                    (
                        f"This model was trained with amp (opt_level: {opt_level}) but amp is not available."
                        " Any further training or inference will happen at full-precision."
                    )
                )
            else:
                model = amp.initialize(model, opt_level=opt_level)

        model_state = torch.load(weights_file, map_location=util.device_mapping(cuda_device))
        model.load_state_dict(model_state)

        return model
```

and here is the error `amp` throws:

```
  File "scripts/run_senteval.py", line 456, in <module>
    app()
  File "/home/johnmg/t2t/lib/python3.7/site-packages/typer/main.py", line 215, in __call__
    return get_command(self)()
  File "/home/johnmg/t2t/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/johnmg/t2t/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/johnmg/t2t/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/johnmg/t2t/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/johnmg/t2t/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/johnmg/t2t/lib/python3.7/site-packages/typer/main.py", line 496, in wrapper
    return callback(**use_params)  # type: ignore
  File "scripts/run_senteval.py", line 436, in allennlp
    archive = load_archive(path_to_allennlp_archive)
  File "/scratch/johnmg/t2t/allennlp/allennlp/models/archival.py", line 186, in load_archive
    cuda_device=cuda_device,
  File "/scratch/johnmg/t2t/allennlp/allennlp/models/model.py", line 363, in load
    return model_class._load(config, serialization_dir, weights_file, cuda_device)
  File "/scratch/johnmg/t2t/allennlp/allennlp/models/model.py", line 314, in _load
    model = amp.initialize(model, opt_level=opt_level)
  File "/home/johnmg/t2t/lib/python3.7/site-packages/apex/amp/frontend.py", line 358, in initialize
    return _initialize(models, optimizers, _amp_state.opt_properties, num_losses, cast_model_outputs)
  File "/home/johnmg/t2t/lib/python3.7/site-packages/apex/amp/_initialize.py", line 171, in _initialize
    check_params_fp32(models)
  File "/home/johnmg/t2t/lib/python3.7/site-packages/apex/amp/_initialize.py", line 93, in check_params_fp32
    name, param.type()))
  File "/home/johnmg/t2t/lib/python3.7/site-packages/apex/amp/_amp_state.py", line 32, in warn_or_err
    raise RuntimeError(msg)
RuntimeError: Found param _text_field_embedder.token_embedder_tokens.transformer_model.roberta.embeddings.word_embeddings.weight with type torch.FloatTensor, expected torch.cuda.FloatTensor.
When using amp.initialize, you need to provide a model with parameters
located on a CUDA device before passing it no matter what optimization level
you chose. Use model.to('cuda') to use the default device.
```

Any idea why the tensor `_text_field_embedder.token_embedder_tokens.transformer_model.roberta.embeddings.word_embeddings.weight` is not moved to the GPU even though `model.cuda(cuda_device)` is called?

matt-gardner:
Is it because you're not specifying a cuda device when trying to load the model?

```
  File "scripts/run_senteval.py", line 436, in allennlp
    archive = load_archive(path_to_allennlp_archive)
```

matt-gardner:
If that's the right error, then it suggests you should either (1) try to detect something better for `cuda_device`, somehow (probably not reasonable), or (2) throw a more informative error `if cuda_device == -1 and opt_level is not None`.

JohnGiorgi:
Yup, that was the issue. I forgot to pass `cuda_device` to `load_archive`. Thanks a lot.

I will add another check to `Model._load` that raises a `ConfigurationError`, with a more informative message (unless there is a better type of error to raise here?)

matt-gardner:
I think we should move away from `ConfigurationError` in most cases.  This is probably a `ValueError`?

matt-gardner:
Oh, one pretty big issue: if you train your model with `amp`, how do you load it if you don't have a GPU?  I guess our current workaround would be specify `overrides` when calling `load_archive`.  That's pretty annoying and opaque.  Maybe it'd be better to add an `opt_level = None` parameter to `load_archive`, parallel to `cuda_device`.  Then you can call `opt_level = opt_level or config.get("trainer", {}).get("opt_level")`.

matt-gardner:
Or, actually, what's the right behavior when loading on a CPU?  Does `amp` do anything if the model is on the cpu?  If it doesn't, then maybe the right fix is to just do `if cuda_device >= 0 and opt_level` as our check before calling `amp.initialize`.  Does that make sense?

JohnGiorgi:
> Or, actually, what's the right behavior when loading on a CPU? Does `amp` do anything if the model is on the cpu? If it doesn't, then maybe the right fix is to just do `if cuda_device >= 0 and opt_level` as our check before calling `amp.initialize`. Does that make sense?

Good question, I am not sure but I do know that if you try to call `amp.initialize` on a model that is on the CPU you get the error I mentioned above.

The `if cuda_device >= 0 and opt_level` seems reasonable to me. There are basically 2 scenarios for a loading a model trained with amp:

- `amp is not None` and `cuda_device > 0` --> call `model = amp.initialize(model, opt_level=opt_level)`.
- _Either_ `amp is None` or `cuda_device == -1` --> log a (different) warning for each case?

This seems the least restrictive and prevents the latter two scenarios passing silently

matt-gardner:
Yeah, seems reasonable to just log a warning.  Maybe it's unnecessary and we can remove it, but until we know for sure, the warning is safer.  Is it easy to run an experiment and see if you get the same result during prediction if you load an amp-trained model on the CPU?  If you do, then we can probably just skip the warning?  Especially if you can also still get gradients just fine.

JohnGiorgi:
Okay, my setup is a little complicated as I evaluated on [SentEval](https://github.com/facebookresearch/SentEval), but I get near identical results when I train a model __with__ amp and then evaluate:

- On a GPU, __without__ `amp.initialize`
- On a CPU, (__without__ `amp.initialize` as this doesn't make sense)

The performance takes a big hit when I try to evaluate on a GPU _with_ `amp.initialize`. I think this is because SentEval slaps its own PyTorch classifiers on top of my encoder, which would __not__ have been wrapped with `amp.initialize`, so it's likely this is unique to my evaluation setup.

So, I think we can drop the warnings, but we may want to go with your earlier suggestion of allowing the user to explicitly __not__ use amp when loading a model (even if it was trained with `amp`) via an `opt_level` param to `load_archive`.

matt-gardner:
Yeah, totally fine to add that parameter and remove warnings in a new PR.  The fact that `None` has semantic meaning here makes the parameter slightly tricky (you want the default value to be to read from the config, not to set to `None`), but still easy to do.

PR #4023, opened 2020-04-04
Fix transformers vocab not being saved to files

maksym-del:
This is a local fix for https://github.com/allenai/allennlp/issues/3097. Here we add a method to vocab to populate its namespace from transformers encoding (which is fetched in indexer) with an option to resave vocab to files (which is the main problem of https://github.com/allenai/allennlp/issues/3097).

I reserved another ```non_padded_namespace``` family: ```*from_transformers``` to use it by default for transformers instead of ```tags```. 

Not saving transformers vocab to disc prevents ```from_archive``` model loading (because vocab folder can be empty) with all the implications.

I know this fix is not ideal, but I think it is an improvement over what we have currently. We can also open a separate issue dedicated to figuring out transformers <-> allennlp relationships.

### DIFF
--- a/allennlp/data/token_indexers/pretrained_transformer_indexer.py
+++ b/allennlp/data/token_indexers/pretrained_transformer_indexer.py
@@ -28,7 +28,7 @@ class PretrainedTransformerIndexer(TokenIndexer):
 
     model_name : `str`
         The name of the `transformers` model to use.
-    namespace : `str`, optional (default=`tags`)
+    namespace : `str`, optional (default=`from_transformeres`)
         We will add the tokens in the pytorch_transformer vocabulary to this vocabulary namespace.
         We use a somewhat confusing default value of `tags` so that we do not add padding or UNK
         tokens to this namespace, which would break on loading because we wouldn't find our default
@@ -41,7 +41,11 @@ class PretrainedTransformerIndexer(TokenIndexer):
     """
 
     def __init__(
-        self, model_name: str, namespace: str = "tags", max_length: int = None, **kwargs
+        self,
+        model_name: str,
+        namespace: str = "from_transformeres",
+        max_length: int = None,
+        **kwargs,
     ) -> None:
         super().__init__(**kwargs)
         self._namespace = namespace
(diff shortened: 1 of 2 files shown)


allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -65,35 +69,13 @@ def __init__(
...
-                for word, idx in pretrained_vocab.items():
-                    vocab._token_to_index[self._namespace][word] = idx
-                    vocab._index_to_token[self._namespace][idx] = word
-        else:
-            logger.warning(
-                """Wasn't able to fetch vocabulary from pretrained transformers lib.
-                Neither <vocab> nor <encoder> are the valid fields for vocab.
-                Your tokens will still be correctly indexed, but vocabulary file will not be saved."""
-            )
+        vocab.extend_from_precomputed_encoding(
```
matt-gardner, line 76: This will need to be rebased / merged with the latest master, because your other PR was just merged.

allennlp/data/vocabulary.py
```diff
@@ -223,6 +223,7 @@ def __init__(
         self._padding_token = padding_token if padding_token is not None else DEFAULT_PADDING_TOKEN
         self._oov_token = oov_token if oov_token is not None else DEFAULT_OOV_TOKEN
 
+        self._last_save_directory = None
```
matt-gardner, line 226: This needs a type annotation (`self._last_save_directory: str = None`) to make mypy happy.

allennlp/data/vocabulary.py
```diff
@@ -627,6 +630,29 @@ def add_token_to_namespace(self, token: str, namespace: str = "tokens") -> int:
         else:
             return self._token_to_index[namespace][token]
 
+    def extend_from_precomputed_encoding(
```
matt-gardner, line 633: Can you rename this to `extend_from_dictionary`, and put it up by the other `extend*` methods?

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -28,7 +28,7 @@ class PretrainedTransformerIndexer(TokenIndexer):
 
     model_name : `str`
         The name of the `transformers` model to use.
-    namespace : `str`, optional (default=`tags`)
+    namespace : `str`, optional (default=`from_transformers`)
         We will add the tokens in the pytorch_transformer vocabulary to this vocabulary namespace.
         We use a somewhat confusing default value of `tags` so that we do not add padding or UNK
```
matt-gardner, line 33:
```suggestion
        We use `from_transformers` as the default namespace, which gets special treatment in our
        `Vocabulary` class to not get padding or UNK tokens.
```

Actually, how are `UNK` and padding handled here?  Is it possible to set them using the transformer?  We probably *do* want UNK handling and padding tokens, we just need them set in the right way, both here and during loading.  This might need some additional changes to how we save and load vocabularies.

And, while we're at it, we probably want the padding and unk tokens to be namespace-dependent.

allennlp/data/vocabulary.py
```diff
@@ -455,6 +456,29 @@ def extend_from_vocab(self, vocab: "Vocabulary") -> None:
             for token in vocab.get_token_to_index_vocabulary(namespace):
                 self.add_token_to_namespace(token, namespace)
 
+    def extend_from_dictionary(
+        self,
+        encoding_dictionary: Dict[str, int],
+        namespace: str = "from_transformers",
+        resave_to_files: bool = False,
```
matt-gardner, line 463: This is getting to be a bigger change, but it's probably necessary to really fix this the right way.  But I *think* this function should take arguments for setting padding and oov tokens, and those should set them in a namespace-specific way.  And the padding and oov tokens then would need to be serialized with the vocabulary and read from the serialized files.  Then we wouldn't need the `from_transformers` namespace added to the non-padded namespaces.

I would wait to do any work on this, though, until @dirkgr has had a chance to glance at it and see if he agrees.  I think this is the right approach, but I could be missing something.

dirkgr:
It's a minimal fix, but I am very unhappy about the amount of magic going on. The way we do vocabularies already involves an untenable amount of magic, and this adds more. Saving state like this, _especially_ the name of a file that gets overwritten, is just asking for trouble the next time this is used in a scenario we didn't think of.

I sketched an alternative here: #4034. I'll add some discussion over there. It is not as minimal as this is, but it has some other advantages.

PR #4026, opened 2020-04-05
Decouple generator and filepath to fix pickle bug

### LINKED ISSUE #3924: Lazy datasets are not compatible with distributed training
**Describe the bug**

I ran into an error when trying to train my model in a distributed setting using `lazy=True`

```bash
-- Process 1 terminated with the following error:
  File "/home/johnmg/t2t/lib/python3.7/site-packages/torch/multiprocessing/spawn.py", line 19, in _wrap
    fn(i, *args)
  File "/scratch/johnmg/t2t/allennlp/allennlp/commands/train.py", line 464, in _train_worker
    metrics = train_loop.run()
  File "/scratch/johnmg/t2t/allennlp/allennlp/commands/train.py", line 523, in run
    return self.trainer.train()
  File "/scratch/johnmg/t2t/allennlp/allennlp/training/trainer.py", line 646, in train
    train_metrics = self._train_epoch(epoch)
  File "/scratch/johnmg/t2t/allennlp/allennlp/training/trainer.py", line 364, in _train_epoch
    batch_generator = iter(self.data_loader)
  File "/home/johnmg/t2t/lib/python3.7/site-packages/torch/utils/data/dataloader.py", line 279, in __iter__
    return _MultiProcessingDataLoaderIter(self)
  File "/home/johnmg/t2t/lib/python3.7/site-packages/torch/utils/data/dataloader.py", line 719, in __init__
    w.start()
  File "/cvmfs/soft.computecanada.ca/easybuild/software/2017/Core/python/3.7.4/lib/python3.7/multiprocessing/process.py", line 112, in start
    self._popen = self._Popen(self)
  File "/cvmfs/soft.computecanada.ca/easybuild/software/2017/Core/python/3.7.4/lib/python3.7/multiprocessing/context.py", line 223, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "/cvmfs/soft.computecanada.ca/easybuild/software/2017/Core/python/3.7.4/lib/python3.7/multiprocessing/context.py", line 284, in _Popen
    return Popen(process_obj)
  File "/cvmfs/soft.computecanada.ca/easybuild/software/2017/Core/python/3.7.4/lib/python3.7/multiprocessing/popen_spawn_posix.py", line 32, in __init__
    super().__init__(process_obj)
  File "/cvmfs/soft.computecanada.ca/easybuild/software/2017/Core/python/3.7.4/lib/python3.7/multiprocessing/popen_fork.py", line 20, in __init__
    self._launch(process_obj)
  File "/cvmfs/soft.computecanada.ca/easybuild/software/2017/Core/python/3.7.4/lib/python3.7/multiprocessing/popen_spawn_posix.py", line 47, in _launch
    reduction.dump(process_obj, fp)
  File "/cvmfs/soft.computecanada.ca/easybuild/software/2017/Core/python/3.7.4/lib/python3.7/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
AttributeError: Can't pickle local object 'DatasetReader.read.<locals>.<lambda>'
```

and traced it down to this block in `DatasetReader`

https://github.com/allenai/allennlp/blob/bdb29a831ed68cb948b18b42fa61646b9ec11bf8/allennlp/data/dataset_readers/dataset_reader.py#L177-L183

My understanding is that in the distributed setting, the `DataLoader` must be pickled before being copied to each replica, and this pickling fails because of the `lambda` (I may be getting some of the details wrong here, but that's the gist).

I was able to produce a minimal example below that triggers the error and seems to suggest this is in fact the case.

**To Reproduce**

```python
import json
import pickle

from allennlp.data.dataset_readers import TextClassificationJsonReader

# Create a dummy dataset
data = {"text": "blah blah", "label": 0}
with open("data.jsonl", "w") as f: json.dump(data, f)

# Create a reader and read some data
reader = TextClassificationJsonReader(lazy=True)
dataset = reader.read("data.jsonl")

# Try to pickle the resulting dataset
with open("tmp.p", "wb") as f: pickle.dump(dataset, f)
```

Which gives you

```bash
AttributeError                            Traceback (most recent call last)
<ipython-input-22-59d848162c6c> in <module>
     13
     14 # Try to pickle the resulting dataset
---> 15 with open("tmp.p", "wb") as f: pickle.dump(dataset, f)

AttributeError: Can't pickle local object 'DatasetReader.read.<locals>.<lambda>'
```

the same error I reached when trying to train my own model and dataset reader.

I can avoid this error in the minimal example by importing dill as pickle (as per [this SO post](https://stackoverflow.com/questions/25348532/can-python-pickle-lambda-functions))

```python
import dill as pickle

# No error, (seemingly) pickles succesfully
with open("tmp.p", "wb") as f: pickle.dump(dataset, f) 
```

To be perfectly honest, I don't know why this works but am adding it here in case it is useful.

**Expected behavior**

I expect a model with a dataset reader for which `lazy=True` to train without error in the distributed case.

 - OS: Linux
 - Python version: 3.7.4
 - AllenNLP version: "I installed from master"
 - PyTorch version: 1.4.0

**Additional context**

I was using my own custom dataset reader when I first bumped into the error, not `TextClassificationJsonReader`, so this is not just some issue with `TextClassificationJsonReader`. I only used it for the simplified example.

JohnGiorgi:
# Overview

This PR addresses #3924 by decoupling the `instance_generator` argument to `_LazyInstances` into two arguments: `instance_generator` and `file_path`. Intead of passing a `lambda: self._read(file_path),` to `_LazyInstances` in `DatasetReader`, we now pass `self._read` and `file_path`, and `_LazyInstances` just calls: `self._read(filepath)`.

I tested that this fixes #3924 by training in distributed mode with a lazy dataset reader.

Thanks to @matt-gardner for the solution.

## Closes

Closes #3924.

### DIFF
--- a/allennlp/data/dataset_readers/dataset_reader.py
+++ b/allennlp/data/dataset_readers/dataset_reader.py
@@ -41,7 +41,7 @@ class _LazyInstances(IterableDataset):
 
     def __init__(
         self,
-        instance_generator: Callable[[], Iterable[Instance]],
+        instance_generator: Callable[[str], Iterable[Instance]],
         file_path: str,
         cache_file: str = None,
         deserialize: Callable[[str], Instance] = None,
--- a/allennlp/data/dataset_readers/dataset_reader.py
+++ b/allennlp/data/dataset_readers/dataset_reader.py
@@ -42,6 +42,7 @@ class _LazyInstances(IterableDataset):
     def __init__(
         self,
         instance_generator: Callable[[], Iterable[Instance]],
+        file_path: str,
         cache_file: str = None,
         deserialize: Callable[[str], Instance] = None,
         serialize: Callable[[Instance], str] = None,
@@ -49,6 +50,7 @@ class _LazyInstances(IterableDataset):
     ) -> None:
         super().__init__()
         self.instance_generator = instance_generator
+        self.file_path = file_path
         self.cache_file = cache_file
         self.deserialize = deserialize
         self.serialize = serialize
@@ -67,7 +69,7 @@ class _LazyInstances(IterableDataset):
         # Case 2: Need to cache instances
         elif self.cache_file is not None:
             with open(self.cache_file, "w") as data_file:
-                for instance in self.instance_generator():
+                for instance in self.instance_generator(self.file_path):
                     data_file.write(self.serialize(instance))
                     data_file.write("\n")
                     if self.vocab is not None:
@@ -75,7 +77,7 @@ class _LazyInstances(IterableDataset):
                     yield instance
         # Case 3: No cache
         else:
-            instances = self.instance_generator()
+            instances = self.instance_generator(self.file_path)
             if isinstance(instances, list):
                 raise ConfigurationError(
                     "For a lazy dataset reader, _read() must return a generator"
@@ -176,7 +178,8 @@ class DatasetReader(Registrable):
 
         if lazy:
             instances: Iterable[Instance] = _LazyInstances(
-                lambda: self._read(file_path),
+                self._read,
+                file_path,
                 cache_file,
                 self.deserialize_instance,
                 self.serialize_instance,


allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -42,13 +42,15 @@ class _LazyInstances(IterableDataset):
     def __init__(
         self,
         instance_generator: Callable[[], Iterable[Instance]],
```
matt-gardner, line 44:
```suggestion
        instance_generator: Callable[[str], Iterable[Instance]],
```

matt-gardner:
Our integrations are having a hard time at the moment, it looks like, but I can see that the build passed when looking directly at TC, so I'm merging this.

matt-gardner:
Many thanks, @JohnGiorgi!

JohnGiorgi:
Awesome, my pleasure!!

matt-gardner:
It looks like this broke our multi-GPU tests: http://build.allennlp.org/viewLog.html?buildTypeId=AllenNLP_GpuUnitTests&buildId=28549&tab=buildLog.  I'm re-running it, to see if that was just a flaky test, but it looks like it [failed again](http://build.allennlp.org/viewLog.html?buildId=28560&buildTypeId=AllenNLP_GpuUnitTests&tab=buildLog&branch_AllenNLP=%3Cdefault%3E).  Can you run that failing test locally and see if you can figure out what's up?  You have more context here than I do.  If it's working for you locally, it's possible the test is just bad.

JohnGiorgi:
Weird, okay sure I can do that. How do I login to see that failed test?

matt-gardner:
There should be an option to proceed as a guest, or something.

JohnGiorgi:
Hmm, it fails locally as well, although I can train my model in the distributed setting, with `lazy=True` and it completes 1 epoch/serializes to disk.

The test just complains that "finishing training early!" is not in the `worker0_log`. It will probably take me some time to debug this as I have no idea what the issue is.

matt-gardner:
This could just be an assumption of the old code.  It comes from this: https://github.com/allenai/allennlp/blob/d66db44dfe8072f0b8bac2770968d996664e683b/allennlp/training/trainer.py#L445-L464.  Maybe with the new data loader stuff, we don't run into this as much?  Or maybe the iterable iterator is actually running through all batches on all GPUs?  I'm not really sure.

JohnGiorgi:
Actually, this might be indicative of an actual problem. When I try to train my model in the distributed setup, while I don't get any explicit errors, my epoch completes in an unreasonably short amount of time (~2 sec, I expect it to take 10s of minutes). Still trying to debug

maksym-del:
I think I figured it out. I will make a fix in a moment and respond.

bryant1410:
> Our integrations are having a hard time at the moment, it looks like, but I can see that the build passed when looking directly at TC, so I'm merging this.

Doesn't it fail every Sunday (or weekend)?

PR #4028, opened 2020-04-06
Fix for loading IterableDatasets with undefined length

maksym-del:


allennlp/training/trainer.py
```diff
@@ -973,11 +973,7 @@ def from_partial_objects(
         if not optimizer_:
             optimizer_ = Optimizer.default(parameters)
 
-        try:
-            batches_per_epoch = len(data_loader)
-        except TypeError:
-            # If the dataset is lazy, it won't have a length.
-            batches_per_epoch = None
+        batches_per_epoch = len(data_loader)  # returns "1" instead of TypeError for _LazyInstances
```
maksym-del, line 976: The ```TypeError``` exception here is never actually thrown since the ``__len__()`` method is overridden in AllenNLP's ```_LazyInstances``` and always returns ```1```. 
In case someone somehow passes an instance of plain ```IterableDataset```, it is better to actually let it throw TypeError then pass a custom ```None``` to the scheduler which it cannot handle it anyway (it accepts ```int```).

matt-gardner: I agree with your analysis of the behavior here, but does having a length of 1 make sense for the learning rate schedulers?

allennlp/data/dataloader.py
```diff
@@ -76,19 +76,24 @@ def __init__(
             worker_init_fn=worker_init_fn,
             multiprocessing_context=multiprocessing_context,
         )
-        self._batches_per_epoch = batches_per_epoch or super().__len__()
```
matt-gardner, line 79: I'm a bit confused about why this is different.  The only thing I can come up with is that there's something going on in the super class that makes it so that querying the length here doesn't work.  Can you give me some pointers on why this is fixing the issue?

Also, it'd be nice to add a test with lazy datasets that fails before this change and passes after the change, just to be sure we've actually captured the issue, and don't have a regression later.

matt-gardner: And @JohnGiorgi, can you confirm that this code solves your issue?

JohnGiorgi: @matt-gardner Looks like it does. Training now take the expected amount of time.

maksym-del: @matt-gardner in short, the output of ```__len__``` method in case of ```_Lazy_Instances``` is not meaningful while ```self._batches_per_epoch``` has to be meaningful in order for the code to work.

Specifically, the problem with this line was that in the case of ```batches_per_epoch is None``` and ```_LazyInstances``` dataset, ```self._batches_per_epoch``` would be equal to ```1```. (```super().__len__()```  would return its overridden value of ```1```). 
And if ```self._batches_per_epoch``` is ```1``` then the code yields one batch per epoch which was the reason @JohnGiorgi had so short epochs.

matt-gardner: Ah, got it, thanks for the additional explanation.  To paraphrase, the problem is that `__iter__` will have the wrong behavior when you combine lazy datasets and no `_batches_per_epoch`.  `__len__` will still do the same thing, but `__iter__` only gives you one batch per epoch.

allennlp/training/trainer.py
maksym-del: If you ask me, having the length on ```1``` does not make sense in all cases apart from the case where we have one batch per epoch. 

Here is our doc for the method:
https://github.com/allenai/allennlp/blob/d66db44dfe8072f0b8bac2770968d996664e683b/allennlp/data/dataset_readers/dataset_reader.py#L93-L100
I would suggest to actually do crash instead and make clients of the method handle the exception in the proper way.
  
---
Regarding Schedulers, the only one which asks for this parameter is currently is ```SlantedTriangular```. I didn't look closely, but it seems it has some proper response to cases like this:
https://github.com/allenai/allennlp/blob/d66db44dfe8072f0b8bac2770968d996664e683b/allennlp/training/learning_rate_schedulers/slanted_triangular.py#L139
but it might be by chance. 
---

dirkgr: That like is not "by chance", but it is a bit of defensive coding that should not be relied on. I believe if `actual_num_steps_per_epoch` ends up smaller than the real number of steps, your learning rate would go negative. I wanted to avoid that with this line.

dirkgr:
`batches_per_epoch` is only used for `SlantedTriangular`, right? There is no other place that needs that number?

If this is true, I think the correct behavior is this:
 * Eager dataset in all cases: No problem anyways
 * Lazy dataset, not using `SlantedTriangular`: Runs as normal. Does not print an accurate ETA during training, unless you specify `batches_per_epoch` in the config.
 * Lazy dataset, using `SlantedTriangular`: Only works if you specify `batches_per_epoch` in the config. Crashes otherwise.

Did I capture this correctly?

allennlp/training/trainer.py
maksym-del: Thanks for the note.
> That line is not "by chance", but it is a bit of defensive coding

Turns out we discovered that unfortunately, IterableDatasets with undefined length breach the defense :)

Anyway, it existed before and this PR does not change the interaction between _LazyInstances and the scheduler, so we could have a separate issue.

maksym-del:
@dirkgr yes, it is also how I see it regarding SlantedTriangularScheduler (it is the only Scheduler that seem to use this parameter).
However, each time we call ```len(dataset)``` or ```len(dataloader)``` we should expect that the result could be ```1``` because the dataset could be lazy and have epoch size unspecified. So we should inspect all existing code in allennlp which asks for the length of these objects. And also new data-related code should account for this (that is why it is better to throw an error IMO).

PR #4032, opened 2020-04-06
Moving NER models to allennlp_models

dirkgr:
See https://github.com/allenai/allennlp-models/pull/14.


dirkgr:
Woha, tests succeed on the first try. 😌

PR #4038, opened 2020-04-08
modified the make_output_human_readable method in basic_classifier for allennlp-demo

IsThatYou:
In https://github.com/allenai/allennlp-demo/issues/339, I suggested that allennlp-demo uses tokens returned by `predict_json` rather than directly from requestedData. This PR changes the basic_classifier.py, so that the output_dict can contain a tokens field for allennlp-demo. This fix is only for BERT and may not be compatiable to other models. @matt-gardner 
What do you think of this solution? Or do you think maybe changing dataset_reader is more generalizable?


allennlp/models/basic_classifier.py
```diff
@@ -163,6 +163,17 @@ def make_output_human_readable(
             )
             classes.append(label_str)
         output_dict["label"] = classes
+        tokens = []
+        for instance_tokens in output_dict["token_ids"]:
+            tokens.append(
+                [
+                    self.vocab.get_token_from_index(
+                        token_id.item(),namespace="tags"
```
matt-gardner, line 171: Yeah, this looks like the right approach.  The only problem is this namespace.  We don't want to hard-code this here, but let it be specified in the model constructor.  Can you add this as a parameter above?

This is also why the tests are failing, which you can see here: http://build.allennlp.org/viewLog.html?buildId=28680&buildTypeId=AllenNLP_AllenNLPPullRequests&tab=buildLog (log in as guest).  The default value for the constructor parameter should be whatever makes those tests pass, which is probably "tokens".

IsThatYou: I added the namespace to the model constructor. But the test still fails? Can you double-check I understand you correctly?

dirkgr:
@IsThatYou, how come you are submitting someone else's branch as a PR?

IsThatYou:
> @IsThatYou, how come you are submitting someone else's branch as a PR?
> 
> Edit: You made all the commits though. I guess the real question is, how come you're using Eric's repo to store your work?

That's because I have been working on Eric's branch for the project. Sorry if that causes any confusion.

PR #4043, opened 2020-04-08
Moves NLI components

dirkgr:
In spirit, this moves NLI components. Unfortunately we use them for tests, so the components stay and only the configs move. There are a few other fixes rolled into this PR though.

See https://github.com/allenai/allennlp-models/pull/16.


matt-gardner:
My preference on the ensemble would be to just remove the bidaf ensemble code, or put whatever tiny bit of logic was needed from `ensemble.py` into the bidaf ensemble code, instead of bringing back this largely pointless base class.

Is there a big problem with changing the tests such that they use the simple classifier instead of an NLI model?

dirkgr:
> My preference on the ensemble would be to just remove the bidaf ensemble code, or put whatever tiny bit of logic was needed from ensemble.py into the bidaf ensemble code, instead of bringing back this largely pointless base class.

This is good. I will do this.

> Is there a big problem with changing the tests such that they use the simple classifier instead of an NLI model?

I fear it would be a lot of work.

matt-gardner:
I agree that it would be a lot of work, but it's kind of necessary if we really want to split the repos.  Do you have a coherent picture of which models get to stay in the main repo, and why?  My vote would be that exactly one model stays in the repo, a simple classifier, so that we can test integration of various components.

matt-gardner:
I even think there's a good argument for having that one model (and associated dataset reader) live in `tests` instead of the core library.

dirkgr:
I think it will delay the release candidate by two days if nothing else goes wrong.

matt-gardner:
I'm looking into this now; I'll update the PR to fix the tests after removing the three NLI models.

matt-gardner:
@dirkgr, take a look.

dirkgr:
That's impressively fast. One test is broken, but apparently it was broken before.

How do you make the `parameters_inspection.json` files? Is there a tool for that, or do you just have to know how to write them?

matt-gardner:
Apparently there's an `inspect_parameters` function in `util`, which dumps that.  I just assumed the implementation was correct and made the file pass the test.  And if you actually check that against the `config.json` inside the model, it looks right.

The vocab extension code broke when I was doing from_params refactoring, I believe.  I'm not sure whether it's worth actually fixing that, but it's a separate issue.  I'm going to remove the fixtures that aren't needed anymore, also; I forgot that the first time.

matt-gardner:
Looks like I forgot the SNLI dataset reader; removing that also required some simple changes to the base dataset reader tests.

dirkgr:
Did you add these back in https://github.com/allenai/allennlp-models/ somewhere? I can't find them.

I have a PR open for this: https://github.com/allenai/allennlp-models/pull/16. I'm happy to do it, but I don't want to create a conflict if you've already done it somewhere.

matt-gardner:
No, I didn't touch the models repo, I was only fixing tests in here.

PR #4045, opened 2020-04-09
Display activation functions as modules.

mklimasz:
Currently some of the activation functions (and therefore list of activation functions in FeedForward or any custom model with list / lambda based one) wouldn't be displayed in model textual summary (by summary I mean ``print(model)``).

I provide wrapper around lambda based implementations of activation functions (non PyTorch ones) and make corresponding change in FeedForward model.

Example result can be seen in test file.


mklimasz:
I'm not sure what is failing (cannot see details). I guess tests are failing (test that broke on my branch also seems to be broken on master when I ran that locally).

matt-gardner:
Thanks for the PR!  You can see the failures by logging into team city as a guest.  The failures all come from tests that currently use non-modules to specify the activation.  Looks like it's two one-line fixes, here: https://github.com/allenai/allennlp/blob/c497103542096037dc21c24466127bc6c820e75b/allennlp/tests/modules/seq2seq_encoders/compose_encoder_test.py#L37 and here: 
https://github.com/allenai/allennlp/blob/c497103542096037dc21c24466127bc6c820e75b/allennlp/tests/modules/seq2seq_encoders/feedforward_encoder_test.py#L12

While you're at it, can you change the name of this class to be `TestComposeEncoder`?
https://github.com/allenai/allennlp/blob/c497103542096037dc21c24466127bc6c820e75b/allennlp/tests/modules/seq2seq_encoders/compose_encoder_test.py#L42

You'll also need to run `black` to pass some of our code formatting CI checks.

matt-gardner:
Also, can you update the docstring to reflect that the input must be a module, not just a function, and give an example; `e.g., torch.nn.ReLU instead of torch.relu`.

allennlp/nn/activations.py
```diff
@@ -53,13 +54,32 @@ def __call__(self, tensor: torch.Tensor) -> torch.Tensor:
...
+class _ActivationLambda(torch.nn.Module):
+    """Wrapper around non PyTorch, lambda based activations to display them as modules whenever printing model."""
+
+    def __init__(self, func: Callable[[torch.Tensor], torch.Tensor], name: str):
+        super().__init__()
+        # Override class name to match activation name as PyTorch uses that field to create the textual
+        # representation.
+        self.__class__.__name__ = name
+        self.func = func
+
```
matt-gardner, line 66: This actually doesn't work:

```python
>>> class A:
>>>   def __init__(self, name):
>>>     self.__class__.__name__ = name
>>> 
>>> a = A('test')
>>> b = A('test2')
>>> a.__class__.__name__
'test2'
```

Looks like there's a `_get_name` function that we need to override instead.  How does this look?

```suggestion
        self._name = name
        self.func = func

    @overrides
    def _get_name(self):
        return self._name
```

mklimasz: Yep, u're right.
https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/module.py#L1129

I'll change that and write better test (since we override there private class method - to address some pytorch changes in the future).

bryant1410:
Oh, wow, nice! I opened #4101 and I just see this. Nice that someone's working on it. DIfferent motivation but same patch!

mklimasz:
Hey guys, can we get this merged?

Issue #3606, opened 2020-01-13
Add a "log once" function to our loggers.


matt-gardner:
It would be really nice if we could call something like `logger.warn_once(message)`.  There are *lots* of times where we want to do something like this.

If there isn't already a pre-baked solution for this, to implement this we probably need to have our own wrapper around around the calls to `logging.getLogger`, or instead of `import logging` do `from allennlp.common import logging`, and have our wrapper there.

bellecarrell:
I created a PR for this issue. I made a few changes to comply with black that required removing star imports, so using the custom logger currently requires `from allennlp.common.logging import getLogger`.

PR #4062, opened 2020-04-13
Add custom 'logging.Logger' that has warn_once(), debug_once(), etc m…

bellecarrell:
Addresses #3606 by creating custom Logger and Filter classes.


bellecarrell:
Yes, that is simpler. I made the changes.

allennlp/common/logging.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+
+    def __init__(self, name):
+        super().__init__(name)
+        self.msgs = set()
+
+    def debug_once(self, msg, *args, **kwargs):
+        if msg not in self.msgs:
+            self.debug(msg, *args, **kwargs)
+        self.msgs.add(msg)
```
dirkgr, line 21: You can indent the `set.add` call as well, for a minuscule performance boost.

allennlp/common/logging.py
```diff
@@ -0,0 +1,44 @@
...
+
+class AllenNlpLogger(logging.Logger):
+    """
+    A custom subclass of 'logging.Logger' that keeps a set of messages to
+    implement {debug,info,etc.}_once() methods.
+    """
+
+    def __init__(self, name):
+        super().__init__(name)
+        self.msgs = set()
```
dirkgr, line 16: Can you call this variable `self._seen_msgs`, or something with an underscore? We generally use underscores to mark "private" variables.

allennlp/tests/common/logging_test.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def setUp(self):
+        super().setUp()
+        logger = logging.getLogger(str(self.__class__.i))
+        self.test_log_file = os.path.join(self.TEST_DIR, "test.log")
+        hdlr = logging.FileHandler(self.test_log_file)
+        logger.addHandler(hdlr)
+        self.logger = logger
+        self.msg = "test message"
+        self.__class__.i += 1
```
dirkgr, line 18: What is this for?

allennlp/tests/common/logging_test.py
```diff
@@ -0,0 +1,58 @@
+import os
+import logging
+from allennlp.common.logging import AllenNlpLogger
+from allennlp.common.testing import AllenNlpTestCase
+
+
+class TestLogging(AllenNlpTestCase):
```
dirkgr, line 7: Can you add a test that makes sure that this sequence only produces one log entry:
```Python
logger.debug_once("There are %d lights.", 4)
logger.debug_once("There are %d lights.", 5)
```
The reason I am concerned about it is that if this doesn't work, we might end up storing millions of messages in `self.msg`.

dirkgr: In fact, maybe add that to the test, i.e., make sure that after that sequence, `sizeof(logger.msg)` is 1.

PR #4064, opened 2020-04-13
Added method and test for sanitizing text that was tokenized with Sta…

Tobias-Rohde:
…nfords PTB tokenzier.


Tobias-Rohde:
> This also lowercases all text, right? Does it have to do that?
> 
> Can you run linter, formatter, and type checker? We have some contribution guidelines here: https://github.com/allenai/allennlp/blob/master/CONTRIBUTING.md. The short of it is, run `black .; and flake8; and mypy allennlp --ignore-missing-imports --no-strict-optional --no-site-packages --cache-dir=/dev/null` until there are no more errors.

It does not lowercase everything. I am just getting a lowercased version of the current token for looking it up in the dictionaries/sets so that tokens with different casing also get parsed. I added a test case to ensure that casing is preserved. This also helped me find a related bug that I fixed.

I ran the linter etc. and fixed all issues, it should pass everything now.

Issue #3715, opened 2020-02-03
Parameters in docs aren't rendering properly


matt-gardner:
This is for the CONLL 2000 dataset reader:
<img width="1174" alt="Screen Shot 2020-02-03 at 10 32 14 AM" src="https://user-images.githubusercontent.com/3291951/73680177-a8c53780-4670-11ea-84a6-33feac7c9e2e.png">

I haven't looked at the code to know what's causing it to render like this, but it should be fixed.

epwalsh:
I believe pydocmarkdown doesn't support the `:class:ClassName` reference syntax, and in general everything after a colon is meant to be treated as the parameter description. So when the description contains something like `:class:TokenIndexer` it breaks because pydocmarkdown thinks you are defining a new parameter there.

So I think we'd just need to remove all `:class:` and other references of that form.

matt-gardner:
I'm giving this a "week" label, because it's a ton of small changes that have to be done manually all over the docs.  Some related issues / PRs on docs stuff: #3972, #3973, #3974, #3975.

matt-gardner:
Also: #3978.

epwalsh:
A couple updates on this:

## Cross-referencing

It seems like `mathy_pydoc` doesn't handle automatic cross-referencing yet (see https://github.com/justindujardin/mathy/blob/master/libraries/mathy_pydoc/mathy_pydoc/preprocessor.py#L73), so in order to cross-reference we'd have to replace all `:class:Foo` type things with an actual markdown link like `[Foo](./foo.md)`. This would be annoying.

However [pydoc-markdown](https://github.com/NiklasRosenstein/pydoc-markdown) (which `mathy_pydoc` is based on) does do cross-referencing automatically. So we if we switch to `pydoc-markdown` we could replace `:class:Foo` with `#Foo`.

Is there any reason why we went with `mathy_pydoc` instead of `pydoc-markdown`? @DeNeutoy 

## Parameter descriptions that contain colons

Both `mathy_pydoc` and `pydoc-markdown` have trouble when a parameter description contains a colon. For example:

```python
def hello(name: str) -> str:
    """
    Says hello to the person specified with `name`.

    This link will render fine [some_link](https://example.com).

    # Parameters

    name : `str`
        The name to say hello to. This link will break
        this example [some_link](https://example.com).

    """
    return f"Hello, {name}!"
```

Will render as
<details><summary>Expand to see the generated Markdown</summary>

# a

## hello
```python
[...]

PR #4066, opened 2020-04-14
move doc building logic to Makefile

epwalsh:
This is step 1 to addressing https://github.com/allenai/allennlp/issues/3715 since
- (A) it will let us easily swap out the "docstring to markdown" tool (currently `mathy_pydoc`), and
- (B) it makes working on the docs locally much easier since only files that changed need to be re-built.

To serve the docs locally use the command `make serve-docs --jobs 16` and then open your browser to `localhost:8000`. When you change a source file, run `make update-docs` in a separate terminal to re-build the relevant files, then refresh the page in your browser.

The trade-off is that building the docs from scratch takes a little longer. For example, on my computer:

```bash
make clean
time make update-docs --jobs 16
```

takes just under 30 seconds, whereas before it would take about 9 seconds. But if you're changing just one file then re-building takes a split second:

```bash
touch allennlp/common/file_utils.py
time make update-docs
```


schmmd:
Seems great!  But I added @dirkgr to the code review since he deals with things like this more day to day... I just happened to have touched docs recently...

scripts/build_docs_config.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python
```
schmmd, line 1: Did anything change here or did it just move?

epwalsh: This is basically the old `scripts/build_docs.py` simplified. Now it's only responsible for creating the `mkdocs.yml` file instead of actually calling the "docstring to markdown" converter (`mathy_pydoc`).

schmmd:
Looks good.  It's annoying that we're still wasting time shelling out (what the heck is Python doing each time it's loaded!) but this seems pretty nifty!

Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+ifeq ($(shell uname),Darwin)
+ifeq ($(shell which gsed),)
+$(error Please install GNU sed with 'brew install gnu-sed')
+else
+SED = gsed
+endif
+else
+SED = sed
+endif
```
dirkgr, line 21: I'm assuming Makefile syntax doesn't allow indenting this?

scripts/build_docs_config.py
```diff
@@ -0,0 +1,62 @@
...
+
+API_TOC_KEY = "API"
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("target_yaml")
+    parser.add_argument("source_yaml")
+    parser.add_argument("docs_root")
+    parser.add_argument("api_docs_path")
```
dirkgr, line 22: Can you throw some meat at these bones, so that `build_docs_config.py -h` prints something useful?

PR #4068, opened 2020-04-14
Some changes pulled from #3868

dirkgr:
Since we're not merging #3868 before the RC, I pulled some changes from it into here.


allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -71,7 +71,10 @@ def __init__(
         tokenizer_kwargs: Dict[str, Any] = None,
     ) -> None:
         tokenizer_kwargs = tokenizer_kwargs or {}
-        self.tokenizer = AutoTokenizer.from_pretrained(model_name, **tokenizer_kwargs)
+        # As of transformers==2.8.0, fast tokenizers are broken.
+        self.tokenizer = AutoTokenizer.from_pretrained(
+            model_name, **tokenizer_kwargs, use_fast=False
```
epwalsh, line 76: Maybe we should check if `use_fast` is in `tokenizer_kwargs` and throw a warning if it is.

dirkgr: That's a good idea.

PR #3978, opened 2020-03-23
Replace sphynx syntax with markdown.

schmmd:
`find . -name '*.py' -exec gsed -i 's/:[a-z][a-z]*:~*//g' \{\} \;`

PR #4072, opened 2020-04-15
Fix doc formatting and improve doc styling

### LINKED ISSUE #3715: Parameters in docs aren't rendering properly

### LINKED ISSUE #3973: Module-level docstrings appear to mess with our docs
http://docs.allennlp.org/master/api/commands/find_learning_rate/

Compare to https://github.com/allenai/allennlp/blob/master/allennlp/commands/find_learning_rate.py

The docs page doesn't have the `FindLearningRate` subcommand.  Not sure at this point where else this happens, but my guess is a bunch of places.

### LINKED ISSUE #3974: Class attributes render quite poorly in our docs
E.g., `default_implementation`: http://docs.allennlp.org/master/api/commands/train/#default_implementation.  It's getting a default docstring for `str` from somewhere, and my attempts to override it with a docstring above or below the class attribute did not work.

epwalsh:
Fixes #3715, fixes #3973, fixes #3978, fixes #3974

This PR does a few things:
- Replaces `mathy_pydoc` with `pydoc-markdown` and uses some custom processors to handle our specific docstring syntax. `pydoc-markdown` has less bugs, more features, and is more customizable than `mathy_pydoc`.
- Adds some custom css to the docs.
- Makes a bunch of little fixes in docstrings.


allennlp/models/model.py
```diff
@@ -348,20 +348,20 @@ def load(
 
         # Parameters
 
-        config: Params
+        config : `Params`
             The configuration that was used to train the model. It should definitely
             have a `model` section, and should probably have a `trainer` section
             as well.
-        serialization_dir: str = None
+        serialization_dir: `str = None`
```
schmmd, line 355: Do you need a space before the colon? `:` I'm not sure.

dev-requirements.txt
```diff
@@ -33,7 +33,7 @@ git+https://github.com/NVIDIA/apex.git@master
 # YAML manipulation
 ruamel.yaml
 
-mathy_pydoc>=0.6.7,<0.8.0
+git+https://github.com/NiklasRosenstein/pydoc-markdown.git@master
```
schmmd, line 36: Should we pin to a particular commit so we have a deterministic dependency?

schmmd:
Looks great!

allennlp/models/model.py
epwalsh: Either way works

dev-requirements.txt
epwalsh: Good idea

scripts/py2md.py
```diff
@@ -0,0 +1,226 @@
...
+    def _format_function_signature(
+        self, func: Function, override_name: str = None, add_method_bar: bool = True
+    ) -> str:
+        parts = []
+        for dec in func.decorators:
+            parts.append("@{}{}\n".format(dec.name, dec.args or ""))
+        if self.signature_python_help_style and not func.is_method():
+            parts.append("{} = ".format(func.path()))
+        if func.is_async:
+            parts.append("async ")
```
dirkgr, line 120: Do we have `async` functions anywhere? How did you decide to add this?

epwalsh: This comes from here https://github.com/NiklasRosenstein/pydoc-markdown/blob/develop/src/pydoc_markdown/contrib/renderers/markdown.py#L36. Even though we don't have async functions anywhere at the moment I figured it would be good to keep, especially if we decide to spin this thing off into a more general tool in the future.

PR #4074, opened 2020-04-16
ensure all API docs rebuilt when script changes, and always build from scratch in CI

epwalsh:
A few things:
- tweaks the Makefile to make sure all of the API docs are re-built whenever the `scripts/py2md.py` file changes.
- runs `make clean` before `make build-docs` in `scripts/build-docs.sh` to ensure the docs are built from scratch in CI.
- avoids shelling out to Python multiple times when building docs from scratch. Several orders of magnitude faster now.

### DIFF
--- a/scripts/py2md.py
+++ b/scripts/py2md.py
@@ -5,7 +5,10 @@ Turn docstring from a single module into a markdown file.
 """
 
 import argparse
+import os
+from pathlib import Path
 import re
+from typing import Optional
 
 from nr.databind.core import Struct
 from nr.interface import implements, override
@@ -34,10 +37,18 @@ class AllenNlpDocstringProcessor(Struct):
         lines = []
         codeblock_opened = False
         current_section = None
+        consecutive_blank_line_count = 0
         for line in node.docstring.split("\n"):
             if line.startswith("```"):
                 codeblock_opened = not codeblock_opened
             if not codeblock_opened:
+                if not line.strip():
+                    consecutive_blank_line_count += 1
+                    # Two blank lines ends a section.
+                    if consecutive_blank_line_count >= 2:
+                        current_section = None
+                else:
+                    consecutive_blank_line_count = 0
                 line, current_section = self._preprocess_line(line, current_section)
             lines.append(line)
         node.docstring = "\n".join(lines)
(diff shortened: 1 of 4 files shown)


scripts/py2md.py
```diff
@@ -223,10 +230,56 @@ def main():
...
 
     pydocmd.load_modules()
     pydocmd.process()
     pydocmd.render()
+    logging.info("Processed %s", module)
+
+
+def _py2md_wrapper(x: Tuple[str, str]):
+    """
+    Used to wrap py2md since we can't pickle a lambda (needed for multiprocessing).
```
dirkgr, line 245: I wrote a library called "bettermap" at https://github.com/allenai/bettermap.

The gist of it is that normal Python multiprocessing foregoes some optimization and convenience when it comes to multiprocessing, and it does this to be compatible with Windows. If you don't care about Windows, you can just fork the process. That's what `bettermap` does.

You would say `bettermap.map_per_process(fn, sequence)`, and it takes care of the rest. `fn` can be any callable, and your Python source file doesn't have to be importable. You can do this from a console if you want. Due to the magic of forking, the elements of `sequence` don't have to be serialized (or even serializable).

So it has a bunch of advantages. On the other hand, it's an extra dependency (or you have to copy-and-paste the code in here). I'll leave it to your judgement whether you want to use it.

scripts/py2md.py
```diff
@@ -223,10 +230,56 @@ def main():
...
+    return parser.parse_args()
+
+
+def main():
+    opts = parse_args()
+    outputs = opts.out if opts.out else [None] * len(opts.modules)
+    if opts.concurrency > 1 and opts.out:
+        logging.info("Using %d threads", opts.concurrency)
+        with Pool(opts.concurrency) as p:
+            list(p.imap(_py2md_wrapper, zip(opts.modules, outputs), 10))
```
dirkgr, line 278: Apparently the idiomatic way of exhausting an iterator is `collections.deque(iterator, maxlen=0)`?

epwalsh: Ahhhhhh thank you!

scripts/py2md.py
epwalsh: This is neat! Unfortunately on my first try using it, it turned out to be slower here because the `py2md()` function has to do some module importing behind the scenes. With `multiprocessing.Pool` those imports only have to happen once per thread, but with `better.map_per_process` they are imported every time the `py2md()` function is called.

I'm guessing there's a way around this. Maybe by importing everything in the main process before `map_per_process` is called?

dirkgr: Yes, that would do it.

PR #4090, opened 2020-04-17
Fix `from_params` when the class has no registered impl

bryant1410:
This happened to me with `BatchCallback`, where it failed silently to create my callback because it was not found, and then when it was called it gave a `NotImplementedError`. I think the problem is that, if a `Registrable` class has no registered implementations (and shouldn't default to itself) then it shouldn't instantiate itself.

This PR adds a test to show this behavior. It's still pending to patch it. Does somebody want to make a patch? You can cherry-pick this commit if so. Or, if you're a maintainer, you can continue this PR.


dirkgr:
I put a fix on this, but it's not an elegant one. I tried an elegant solution, but it blew up.

allennlp/common/from_params.py
```diff
@@ -564,6 +564,17 @@ def from_params(
         else:
             # This is not a base class, so convert our params and extras into a dict of kwargs.
 
+            # If there was a "type" given, we probably didn't want to end up here.
+            if "type" in params:
+                value = params.pop("type")
+                message = (
+                    f"{value} not in acceptable choices for {params.history}type. "
```
bryant1410, line 571: Missing space before "type"?

bryant1410:
Thanks!

allennlp/common/from_params.py
dirkgr: No, `params.history` already ends in a period. I know, it looks weird ...

dirkgr:
@bryant1410 , just so you know, Matt is on vacation, but I really want his review on this, so this might have to sit for a few days.

bryant1410:
> @bryant1410 , just so you know, Matt is on vacation, but I really want his review on this, so this might have to sit for a few days.

Thanks for letting me know! Makes sense, there's no rush.

allennlp/common/from_params.py
```diff
@@ -564,6 +564,17 @@ def from_params(
         else:
             # This is not a base class, so convert our params and extras into a dict of kwargs.
 
+            # If there was a "type" given, we probably didn't want to end up here.
+            if "type" in params:
```
matt-gardner, line 568: This check isn't sufficient.  If the test had done `A.from_params({"some": "dict"})`, you wouldn't get here.  I'll look into getting a better fix for this.

PR #4100, opened 2020-04-20
Fix typo in `InitializerApplicator` docstring

bryant1410:


allennlp/nn/initializers.py
```diff
@@ -444,7 +444,7 @@ class InitializerApplicator(FromParams):
         A list mapping parameter regexes to initializers.  We will check each parameter against
         each regex in turn, and apply the initializer paired with the first matching regex, if
         any. If "prevent" is assigned to any regex, then it will override and prevent the matched
-        parameters to be initialzed.
+        parameters to be initialized.
```
epwalsh, line 447: I think this whole sentence is a little confusing and possibly redundant given the description of `prevent_regexes` on line 449. Maybe we should just remove it?

PR #4105, opened 2020-04-20
Fixes for GitHub Actions

epwalsh:


.github/workflows/nightly.yml
```diff
@@ -46,6 +46,6 @@ jobs:
     - name: Deploy
       if: success()
       env:
-        PYPI_USERNAME: ${{ secrets.PYPI_USERNAME }}
+        PYPI_USERNAME: allennlp
```
epwalsh, line 49: Using a secret for this causes any mention of "allennlp" to be starred out with "***" in the build logs which is annoying

PR #4107, opened 2020-04-21
Fix docs workflow

epwalsh:


.github/workflows/docs.yml
```diff
@@ -71,12 +71,17 @@ jobs:
         ssh-keyscan github.com >> ~/.ssh/known_hosts
         
         # Checkout allennlp-docs to /allennlp-docs
-        mkdir -p /ssh
-        echo $DOCS_DEPLOY_KEY > /ssh/allennlp-docs.key
-        ssh-add /ssh/allennlp-docs.key
-        GIT_SSH_COMMAND='ssh -i /ssh/allennlp-docs.key' git clone git@github.com:allenai/allennlp-docs.git /allennlp-docs
+        echo $DOCS_DEPLOY_KEY > ~/allennlp-docs.key
```
epwalsh, line 74: Would have had to add sudo to a bunch of these commands if we didn't put the key in the home directory

.github/workflows/docs.yml
```diff
@@ -71,12 +71,17 @@ jobs:
...
         
         # Copy the generated docs to the checked out docs repo
-        rm -rf /allennlp-docs/$DIRECTORY/
-        mkdir -p /allennlp-docs/$DIRECTORY
-        cp -r site/* /allennlp-docs/$DIRECTORY
+        rm -rf ~/allennlp-docs/$DIRECTORY/
+        mkdir -p ~/allennlp-docs/$DIRECTORY
+        cp -r site/* ~/allennlp-docs/$DIRECTORY
+
+        # And push them up to GitHub
```
epwalsh, line 83: Forgot to actually push the files 🤦

.github/workflows/docs.yml
dirkgr: Do you have to set the permissions on the key to `600`? Sometimes `ssh` is finicky about that.

Issue #4096, opened 2020-04-17
Improved release process


dirkgr:
Releasing the 1.0 release candidate was a little messy. We should clean it up before the big 1.0 release. Here are some requirements in no particular order:
 * We need to coordinate releases between [allennlp](https://github.com/allenai/allennlp) and [allennlp-models](https://github.com/allenai/allennlp-models) so they are released in quick succession.
 * We need to release both on pypi and make a GitHub release.
 * We want to have release notes in GitHub.
 * `allennlp-models==1.0.0` needs to depend on `allennlp==1.0.0`.
 * These versions have to have been tested together. There is a possibility that `-models` fails and the fix is in the core library, so we need to test before we push out any releases.
 * We need to release an updated docker image at the same time, so that the instructions in the README and at https://pypi.org/project/allennlp/1.0.0rc1/ work.
 * We need to update the demo to use the latest version. That means we need to run the demo with that version and smoke test it before the release.
 * We need to upload new docs to a directory that matches the version we just released.
 * Nightly release management: The nightly release version should always be `{latest released version + 0.0.1}.dev{date}`.

epwalsh:
Now that I think about, we could release models and core at the same time from within the same CI build. It could go like this:

*Release build on core is triggered from a GitHub release:*
- Pulls master branch of models into separate directory, runs all tests
- If all tests pass, a GitHub release of models is pushed and then both models and core are built and pushed to PyPI

schmmd:
I'd remove "We need to update the demo to use the latest version. That means we need to run the demo with that version and smoke test it before the release." as a requirement.  The demo has almost always been on a particular SHA, but it's nice to have it on a release.

dirkgr:
> The demo has almost always been on a particular SHA, but it's nice to have it on a release.

At the very least we should indicate what version of the library is running on the demo. It's a common use case to try something in the demo, and then expect it to happen the same way in your code.

epwalsh:
So my tentative plan for this is to add one more GitHub Actions workflow called "PyPI". This will look a lot like the current [nightly core release workflow](https://github.com/allenai/allennlp/blob/master/.github/workflows/nightly.yml) and the [models release workflow](https://github.com/allenai/allennlp-models/blob/master/.github/workflows/release.yml). Both of those workflows will be removed and absorbed into this workflow so that the core and models libraries are released together.

This workflow will only run when a GitHub release is published (and also nightly) and will include a few improvements over the existing release workflows mentioned above.

[...]

PR #4112, opened 2020-04-21
Adds release workflow

epwalsh:
Addresses #4096 and fixes other issues with the GitHub Actions workflows including the ssh problems with the docs deployment job. This also fixes the issue with the logging test.

## Changes

The PR checks remain the same. What's different is what happens on a release or master commit. When either of these events occur, 4 more jobs will be ran once the usual PR checks are successful:

**`Build Package`**: Builds the PyPI distribution files.

**`Test Package`**: Tests installing the core library from the distribution files.

**`Docker`**: Builds, tests, and pushes a Docker image by installing the core library from the distribution files. This only runs once the `Test Package` job has completed successfully. On a master commit, the image will be pushed as `allennnlp/commit:COMMIT_SHA`, and on a release it will be pushed as `allennlp/allennlp:RELEASE_TAG`.

**`Docs`**: Builds and pushes the documentation to the `allennlp-docs` repo. This only runs once the `Test Package` job has completed successfully. On a master commit, the docs are put in the `master/` folder, and on a release they are put into a folder with the name of the release tag.

Further, for a release (not for a master commit), one more job named `PyPI` runs after all of the other jobs mentioned complete successfully. This just uploads the distribution files to PyPI.

## Release process

With this updated workflow, the release process will look like this:

1. Create and publish a GitHub release for the core library.
2. Wait for the workflow to pass, which will result in the core library being published to PyPI.
3. Create and publish a GitHub release for the models library.
4. Wait for the workflow (in the models repo) to pass, which will result in the modesl library being published to PyPI.

## What this PR doesn't address

- Nightly builds. I'll follow up with a PR for that later.
- Automating the process of creating a GitHub release. This will be more involved.


schmmd:
Painful commit history!  The release process looks like a huge improvement over our past one!

epwalsh:
> Painful commit history! The release process looks like a huge improvement over our past one!

It really was 😭 😭 😭

epwalsh:
I've confirmed that the fixes to the docs workflow actually do the trick: https://github.com/epwalsh/allennlp/runs/606384562

epwalsh:
And this confirms that the logging test has been fixed: https://github.com/epwalsh/allennlp/runs/606506610

Dockerfile.whl
```diff
@@ -0,0 +1,32 @@
+# This Dockerfile creates an environment suitable for downstream usage of AllenNLP.
+# It creates an environment that includes a wheel installation of allennlp.
```
dirkgr, line 2: Should we install `-models` into this too?

dirkgr:
It's not great that it says "skipped" for all those tests, but not critical either.

.github/workflows/main.yml
```diff
@@ -0,0 +1,348 @@
+name: CI
+
+on:
+  # TODO: add nightly schedule.
```
dirkgr, line 4: So no nightly release right now?

epwalsh: Yea, although I could probably get that done by the end of the day

.github/workflows/main.yml
```diff
@@ -0,0 +1,348 @@
...
+  release:
+    types: [published]
+
+jobs:
+  check_core:
+    name: Check Core
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python: ['3.6', '3.7']
```
bryant1410, line 20: I know you just moved things from files, but maybe we should make the tests run on 3.6-8? As these are the 3 major Python versions we support.

.github/workflows/main.yml
```diff
@@ -0,0 +1,348 @@
...
+
+    - name: Lint
+      run: |
+        make lint
+
+    - name: Type check
+      run: |
+        make typecheck
+
+    - name: Run tests
```
bryant1410, line 55: What about also adding a `python setup.py check`? This performs checks package-wise, especially on the metadata.

.github/workflows/main.yml
```diff
@@ -0,0 +1,348 @@
...
+    - uses: actions/cache@v1
+      with:
+        path: ${{ env.pythonLocation }}
+        key: ${{ runner.os }}-pydeps-${{ matrix.python }}-${{ hashFiles('setup.py') }}-${{ hashFiles('dev-requirements.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pydeps-${{ matrix.python }}
+
+    - name: Install requirements
+      run: |
+        pip install --upgrade pip setuptools wheel
```
bryant1410, line 96: Is it necessary to specify "setuptools" and "wheel", given that they're here and will be installed if missing by the next `pip install -e .`?:

https://github.com/allenai/allennlp/blob/0638cbb249caff0174ee0b7a6f6c8c235443e798/pyproject.toml#L22

.github/workflows/main.yml
```diff
@@ -0,0 +1,348 @@
...
+        pip install -e .
+        pip install --upgrade -r dev-requirements.txt
```
bryant1410, line 98: It can be faster to do:

```python
pip install --upgrade -e . -r dev-requirements.txt
```

As it resolves the dependencies only once.

bryant1410:
Oh, my comments were 5m late.

.github/workflows/main.yml
epwalsh: I'm not sure if we should support 3.8 yet. Last time I tried to use 3.8 I ran into a butch of issues with dependencies not supporting it yet. That was a couple months ago so maybe those issues are fixed by now. In any case it would be worth trying but should probably hold off until we get the release process completely ironed out.

.github/workflows/main.yml
epwalsh: That's probably a good idea. It looks like it will fail though when we have a `-unreleased` version?

.github/workflows/main.yml
epwalsh: I did run into an issue once where some deps failed to install because wheel hadn't been installed yet. Maybe I was doing something weird at the time.

.github/workflows/main.yml
bryant1410: Yeah, it breaks pip and PyPI. There are some rules of what are valid suffixes.

.github/workflows/main.yml
bryant1410: I understand what you say. However, it should be installed now if missing because of that line I'm mentioning. See https://pip.pypa.io/en/stable/reference/pip/#pep-517-and-518-support

epwalsh: Gotcha, I didn't know about that until now so thank you. It's probably a good idea to keep them up-to-date though. Do you know if `pip` will automatically do that?

bryant1410: It won't, at least when I had problems with older Python versions. I'm not sure if it's necessary to keep them up-to-date but probably is necessary to set a minimum version. I had to set a minimum version for `setuptools` once, to support my features, because I saw the default in some configuration (Python 3.5?) wasn't working fine: https://github.com/pln-fing-udelar/fast-krippendorff/commit/e8ec06fa267ecaa691061e8331615f313c516566#diff-90829e76e906f1c73140c7ded7e1b268

.github/workflows/main.yml
dirkgr: @bryant1410, did you try it with 3.8? Does it work?

.github/workflows/main.yml
dirkgr: That's neat. It also gives it a chance to do a better job resolving dependencies.

.github/workflows/main.yml
bryant1410: I use a project with Python 3.8 with AllenNLP from master

.github/workflows/main.yml
epwalsh: @bryant1410 feel free to make a PR for this

PR #4116, opened 2020-04-22
Update README with new CI links / badge

epwalsh:
Also matches the formatting of the logo with what's in allennlp-models and adds a badge for PyPI.

Screen Shot 2020-04-21 at 5 30 10 PM


epwalsh:
The centering of the main text and badges is just personal preference. I think it looks better that way since the logo is centered. But feel free to disagree

PR #4117, opened 2020-04-22
Simplify pull request workflow

epwalsh:
This separates the workflow that is ran on pull requests from the one that is ran on master commits and releases.

The benefit of this is that **only the relevant jobs will show up on pull requests** instead of all jobs (most of which are skipped). The downside is that two of the jobs need to be duplicated across the workflows.

I definitely don't like duplicated configs, but I also don't like having so many jobs showing up on pull requests. I imagine that would be a source of confusion for contributors.

Also @bryant1410 this covers the `pip install -e . -r dev-requirements.txt` optimization you pointed out.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -101,7 +101,7 @@ jobs:
         EXCLUDE_ALLENNLP_IN_SETUP: "true"
       run: |
         git clone https://github.com/allenai/allennlp-models.git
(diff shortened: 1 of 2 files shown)


bryant1410:
Good, thanks! We're already seeing the benefit of versioning the config, as anybody can comment/improve on it.

PR #4123, opened 2020-04-22
Add nightly release schedule back

epwalsh:
This will run Sunday through Thursday night. It runs through the same workflow as a usual release except that the docs and docker images are not updated.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -1,12 +1,13 @@
 name: Master
 
 on:
-  # TODO: add nightly schedule.
   push:
     branches:
     - master
   release:
     types: [published]
+  schedule:
+  - cron: '37 11 * * 0,1,2,3'  # nightly at 11:37 UTC Sunday - Thursday
 
 jobs:
   check_core:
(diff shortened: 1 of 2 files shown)


.github/workflows/master.yml
```diff
@@ -1,12 +1,13 @@
...
 
 on:
-  # TODO: add nightly schedule.
   push:
     branches:
     - master
   release:
     types: [published]
+  schedule:
+  - cron: '37 11 * * 0,1,2,3'  # nightly at 11:37 UTC Sunday - Thursday
```
bryant1410, line 10: Just FYI, there was a preference before this on releasing 11 AM Seattle time: https://github.com/allenai/allennlp/commit/b4c7d76cc1f1e9a039c2d27ad7ecbfd828a36ec2#diff-0aebd1990a83156b6c14a24f3a89f76eL6

epwalsh: Was there a specific reason for 11 am? I made it earlier on purpose so that I would know if something went wrong first thing in the morning

bryant1410: I have no idea. I think @DeNeutoy put it that way, but idk the rationale :man_shrugging:

DeNeutoy: I think I was debugging it one day and I wanted it to run again that day, and it was about 10:45 😅 😅 😅 . Feel free to release whenever you like, it's not important.

PR #4137, opened 2020-04-24
Remove unused Dockerfiles

epwalsh:
closes #4118. I just moved `Dockerfile.whl` to `Dockerfile` and removed `Dockerfile.pip`.


schmmd:
@epwalsh previously there was a straightforward way to rebuild a Dockerfile from a commit, release, etc.  I'm not sure how to do this now that we need a `whl`.  How would you build a Dockerfile for--say--v0.9.0?

epwalsh:
@schmmd that came from `Dockerfile.pip`. Do you want to keep that? Only reason I wanted to remove it was for one less thing to maintain, since the Docker images for every commit and release are on Docker Hub.

schmmd:
Rather, I'd love to have some brief documentation in the `Dockerfile` about how you'd do this.  How can I download a wheel for a published version of AllenNLP for example?

epwalsh:
@schmmd gotcha, just added some instructions

epwalsh:
@schmmd actually I realize the Docker instructions in the README need to be updated as well, so I just moved the comments from the Dockerfile to the README

PR #4138, opened 2020-04-24
add shortcuts for 'stable' and 'latest' to docs

epwalsh:
This adds a step to the doc building CI job which will update the redirect links at [https://docs.allennlp.org/latest/](https://docs.allennlp.org/latest/) and [https://docs.allennlp.org/stable/](https://docs.allennlp.org/stable/).

For a stable release (release without any suffix or with a "post" suffix) both "latest" and "stable" will be updated to redirect to the new release version docs. For a non-stable release, only "latest" will be updated.


.github/workflows/master.yml
```diff
@@ -308,28 +308,67 @@ jobs:
...
+          <body>
+            <p>Please follow <a href="/${DOCS_FOLDER}/">this link</a>.</p>
+          </body>
+        </html>
+        EOL
+        cat latest/index.html
+
+        SUFFIX=$(echo $DOCS_FOLDER | sed -E 's/v[0-9]+\.[0-9]+\.[0-9]+//')
+        if [[ -z $SUFFIX ]] || [[ "$SUFFIX" == *"post"* ]]; then
+            # If there's no suffix to the version or the suffix is a 'post',
```
dirkgr, line 356: What does it mean for the suffix being a "post"? As in, a post-release? A hot fix?

`"$SUFFIX" == *"post"*`, does that actually work? Won't that just engage the file glob mechanism?

.github/workflows/master.yml
```diff
@@ -300,28 +300,67 @@ jobs:
...
+      # Only run this on main repo (not forks) for commits and releases but not for
+      # nightly builds.
+      if: github.repository == 'allenai/allennlp' && github.event_name != 'schedule'
+      run: |
         # And push them up to GitHub
-        cd ~/allennlp-docs/$DIRECTORY
-        git add .
+        cd ~/allennlp-docs/
+        git add -A
         git commit -m "automated update of the docs"
```
schmmd, line 364: You can also just do `git commit -a -m "automated update of the docs"`.

epwalsh: @schmmd that would miss any new files added I believe

schmmd: Yep--you're likely right!

epwalsh:
> Other than that I'm OK with shipping this, but I'm wondering if it would be better to detect the latest stable version every time through, rather than only catching it when this script runs on that version.

@dirkgr that's a good point. I just added a script to the other PR: https://github.com/allenai/allennlp/pull/4142/files that will make this easier. So when that's merged I'll update this.

.github/workflows/master.yml
epwalsh: I did double-check that it worked, but with the last commit I just pushed I removed this line anyway.

epwalsh: Weird bash stuff

PR #4152, opened 2020-04-27
clean up scripts dir

### LINKED ISSUE #4115: Clean up scripts folder
Since the CI steps are now in the Makefile we shouldn't need `scripts/verify.py` and some other stuff in that folder.

epwalsh:
closes #4115 

I only removed what I know for sure we don't need. There are some other scripts I wasn't sure about:
- `train_fixtures.py`
- `regenerate_archived_models.py`
- `cache_models.py`
- `benchmark_iter.py`
- `build_demo.py`


schmmd:
`train_fixtures` is crucial.  I have no idea what `build_demo` does though...

matt-gardner:
- train_fixtures.py: definitely keep this one
- regenerate_archived_models.py: I think this functionality got folded into `train_fixtures.py`, and this file can be removed.
- cache_models.py: I think this was used to pre-load a list of models in the demo.  Relies on `commands.serve`, which is gone.  Can definitely be removed.
- benchmark_iter.py: looks like this was for the old `Iterator` code, and is no longer working or needed.
- build_demo.py: very old (from before we split the demo into its own repo), and can be removed

PR #4153, opened 2020-04-27
Reduce number of warnings seen after running tests

epwalsh:
This addresses some of the warnings you see when running tests.

One major warning came from `pytest.mark.spawn` because, as it turns out, the `spawn` mark from https://github.com/allenai/allennlp/pull/3903 is actually not a pytest thing. It comes from [PyTorch Lightning](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/tests/conftest.py).


bryant1410:
I don't completely understand, can you elaborate? What do you mean `@pytest.mark.spawn` is actually not a pytest thing? And how does the module `conftest.py` gets loaded (and how things in it work)?

epwalsh:
@bryant1410 

> And how does the module conftest.py gets loaded (and how things in it work)?

`conftest.py` is something pytest always looks for in a test directory. It's used to provide fixtures and other things.

> What do you mean @pytest.mark.spawn is actually not a pytest thing? 

If you look at the output from the tests before this PR you'll see this:

Screen Shot 2020-04-27 at 5 29 34 PM

allennlp/training/optimizers.py
```diff
@@ -521,14 +521,14 @@ def make_sparse(values):
 
                 else:
                     # Decay the first and second moment running average coefficient
-                    exp_avg.mul_(beta1).add_(1 - beta1, grad)
-                    exp_avg_sq.mul_(beta2).addcmul_(1 - beta2, grad, grad)
+                    exp_avg.mul_(beta1).add_(grad, alpha=1 - beta1)
```
epwalsh, line 524: Someone may want to double-check these changes.

allennlp/training/optimizers.py
```diff
@@ -521,14 +521,14 @@ def make_sparse(values):
...
+                    exp_avg_sq.mul_(beta2).addcmul_(grad, grad, value=1 - beta2)
```
epwalsh, line 525: and this

allennlp/training/optimizers.py
```diff
@@ -521,14 +521,14 @@ def make_sparse(values):
...
                     denom = exp_avg_sq.sqrt().add_(group["eps"])
 
                     bias_correction1 = 1 - beta1 ** state["step"]
                     bias_correction2 = 1 - beta2 ** state["step"]
                     step_size = group["lr"] * math.sqrt(bias_correction2) / bias_correction1
 
-                    p.data.addcdiv_(-step_size, exp_avg, denom)
+                    p.data.addcdiv_(exp_avg, denom, value=-step_size)
```
epwalsh, line 532: and this

bryant1410:
Cool you're looking after the warnings! They reveal things many times, such as this one.

allennlp/training/optimizers.py
dirkgr: I did. The documentation is dangerously confusing about this, but when I try it in the console I get the same results before and after.

Same for the other two

epwalsh: Thanks for confirming this!

Issue #4122, opened 2020-04-22
`load_archive` not working with models fine-tuned with "type": "from_archive"


ZhaofengWu:
If I train a model A, and then fine-tune it with
```
{
  "model": {
    "type": "from_archive",
    "archive_file": path
  }
}
```
and get a model B, `load_archive('B/model.tar.gz')` will give the following error
```
  File ".....", line 120, in __init__
    archive = load_archive(archive_file)
  File "...../allennlp/allennlp/models/archival.py", line 186, in load_archive
    cuda_device=cuda_device,
  File "...../allennlp/allennlp/models/model.py", line 336, in load
    return model_class._load(config, serialization_dir, weights_file, cuda_device)
AttributeError: 'function' object has no attribute '_load'
```

matt-gardner:
@ZhaofengWu, if you'd be willing to make a simple minimal failing test, I can look into fixing this.

dirkgr:
I think I know what the problem is here. He's fine-tuning a model twice, using `from_archive` both times. The first time, the model config comes from the model archive. But the second time the model archive only contains `from_archive`, and has none of the important information.

matt-gardner:
I'm not so sure; the `path` has a config file inside of it, so it has the important information.  The issue is that this line is returning a function for some reason: https://github.com/allenai/allennlp/blob/27e374ac6e6aaae6565751970d5f592ff6b855de/allennlp/models/model.py#L386
I need more detail to diagnose exactly what's going on, but it looks like a simple untested error with a change that I made to how `by_name()` works, not a problem with missing configuration.  It also looks like @ZhaofengWu is running an older version of allennlp, because there is no `opt_level` in the call to `model_class._load`.  Can you update and make a failing test case on the current master?

ZhaofengWu:
Closed by #4156

PR #4156, opened 2020-04-28
Add failing from_archive test

ZhaofengWu:
For #4122

The fixture model was trained with `allennlp train allennlp/tests/fixtures/basic_classifier/experiment_from_archive.jsonnet -s from_archive_serialization`


matt-gardner:
So, I dug into this a bit more to see if there was a better way to fix this problem.  I decided that `_load` actually already implements the non-hacky thing that I was thinking should be done here (calling `Model.from_params`, with whatever necessary pre- and post-processing there was).  From there, I thought a decent solution was just to remove `load` and rename `_load` to `load`.  That breaks anyone who subclassed `_load`, but that seems extremely uncommon and not really useful or necessary.  Turns out, though that the bidaf ensemble does this.  I'm still in favor of just removing the bidaf ensemble and doing that renaming, but I implemented a different fix, which you can see in the code.

When I did that fix, I realized that the original test was inadequate, and I ran into an unrelated bug in `Embedding` that needed a one-line fix in order for the model to load correctly.  So I made the test better and fixed the bug.

Please take another look.

PR #4157, opened 2020-04-28
add test for version convention

epwalsh:
This will ensure that we are consistent with how we version future releases.

### DIFF
--- a/allennlp/tests/version_test.py
+++ b/allennlp/tests/version_test.py
@@ -1,0 +1,52 @@
+import re
+
+import pytest
+
+from allennlp.version import VERSION
+
+
+# Regex to check that the current version set in `allennlp.version` adheres to
+# PEP 440, as well as some of our own internal conventions, such as the `.dev`
+# suffix being used only for nightly builds.
+# 0.0.0rc0.post0.dev20200424
+VALID_VERSION_RE = re.compile(
+    r"^"
+    r"(0|[1-9]\d*)"  # major
+    r"\.(0|[1-9]\d*)"  # minor
+    r"\.(0|[1-9]\d*)"  # patch
+    r"(rc(0|[1-9]\d*))?"  # patch suffix
+    r"(\.post(0|[1-9]\d*))?"  # [.postN]
+    r"(\.dev2020[0-9]{4})?"  # [.devDATE]
+    r"$"
+)
+
+
+def is_valid(version: str) -> bool:
+    return VALID_VERSION_RE.match(version) is not None
+
+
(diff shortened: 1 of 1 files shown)


allennlp/tests/version_test.py
```diff
@@ -0,0 +1,52 @@
+import re
+
+import pytest
+
+from allennlp.version import VERSION
+
+
+# Regex to check that the current version set in `allennlp.version` adheres to
+# PEP 440, as well as some of our own internal conventions, such as the `.dev`
+# suffix being used only for nightly builds.
```
epwalsh, line 10: "`.dev` suffix being used only for nightly builds" might be too restrictive.

allennlp/tests/version_test.py
```diff
@@ -0,0 +1,52 @@
...
+# 0.0.0rc0.post0.dev20200424
+VALID_VERSION_RE = re.compile(
+    r"^"
+    r"(0|[1-9]\d*)"  # major
```
schmmd, line 14: Is this the right regex?  Isn't the `* just on the `\d` rather than on the whole expression?

epwalsh: It is just on the `\d`, which I think gives the correct behavior because it disallows a preceding "0", unless by itself. For example:

- ✅  "0"
- ✅  "1"
- ✅  "10"
- ❌  "01"

schmmd: sorry--brain fart

allennlp/tests/version_test.py
```diff
@@ -0,0 +1,52 @@
...
+    r"\.(0|[1-9]\d*)"  # minor
+    r"\.(0|[1-9]\d*)"  # patch
+    r"(rc(0|[1-9]\d*))?"  # patch suffix
+    r"(\.post(0|[1-9]\d*))?"  # [.postN]
+    r"(\.dev2020[0-9]{4})?"  # [.devDATE]
```
schmmd, line 19: You might comment on what `post` and `dev` are.  I don't know.

bryant1410:
Doesn't `python setup.py check` already check it

epwalsh:
Yes but if the version is invalid it will only emit a warning, yet still have a zero exit code. So not great for CI. We are also imposing some additional restrictions on what constitutes a valid version, which would not be checked with that.

bryant1410:
What if you take the output and pipe it so it's an error instead?

bryant1410:
I think that's better, because we're kinda re-inventing the wheel with this patch.

PR #4159, opened 2020-04-28
Remove allennlp sparse_clip_grad and replace with torch clip_grad_norm_.

### LINKED ISSUE #4075: Use pytorch's native clip_grad and clip_grad_norm functions
Pytorch now has its own[ implementations](https://github.com/pytorch/pytorch/blob/master/torch/nn/utils/clip_grad.py) of these functions and it looks they also support sparse gradients. We could use them instead of custom allennlp's alternatives.

bellecarrell:
Closes #4075. 

`sparse_clip_grad` appears to be, based on documentation and usage, the custom `allennlp` implementation of Torch's `clip_grad_norm_`.


epwalsh:
@bellecarrell will this only work for PyTorch >= 1.5?

allennlp/tests/training/trainer_test.py
```diff
@@ -1001,23 +1000,3 @@ def test_trainer_can_run_amp(self):
             opt_level="O1",
         )
         _ = trainer.train()
-
-
-class TestSparseClipGrad(AllenNlpTestCase):
```
DeNeutoy, line 1006: Does this test actually pass with the torch gradient clipping? The reason we implemented it ourselves is so that it works for sparse embedding layers - looking at the pytorch docs for `clip_grad_norm` i'm not sure that they've handled this case. Possibly it's fixed by them implementing `norm` for sparse tensors, but it doesn't look like they have from the docs: https://pytorch.org/docs/stable/sparse.html

epwalsh: @bellecarrell could you put that test back with torch's `clip_grad_norm` substituted in?

epwalsh:
Thanks @bellecarrell! This is good that the test passes, but I guess I'm still concerned that it will break for older versions of torch. I think we should try to keep backwards compatibility at least as far back as torch 1.4. Could you see if it works on 1.4?

PR #4162, opened 2020-04-28
ensure typing backport uninstalled first

epwalsh:
Ugh, somehow fix # 2 still didn't do it

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,8 @@ test-with-cov :
 
 .PHONY : install
 install :
+	# Making sure the typing backport isn't installed.
+	pip uninstall -y typing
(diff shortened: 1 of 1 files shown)


epwalsh:
https://github.com/allenai/allennlp/pull/4159/checks

Makefile
```diff
@@ -57,15 +57,18 @@ test-with-cov :
 
 .PHONY : install
 install :
+	# Making sure the typing backport isn't installed.
+	pip uninstall -y typing
```
epwalsh, line 61: Note: this will have exit code 0 even if typing is not installed

dirkgr: Yes, but it prints a warning. That's why I did the `grep` thing.

PR #4163, opened 2020-04-28
Ensure error messages aren't duplicated in log streams

epwalsh:
Currently if anything is logged at the error level (or higher) it will be written to both stdout and stderr, so those message will appear twice in the terminal. This PR ensures that errors will only be written to stderr (note that errors will still appear in both the `stdout.log` and `stderr.log` files).


epwalsh:
Closing in favor of https://github.com/allenai/allennlp/pull/4164

PR #4164, opened 2020-04-28
Fix logging

### LINKED ISSUE #3785: Fix file friendly logging

epwalsh:
Closes #3785 and also ensures tqdm logged to file.

The solution just patches `sys.stdout` and `sys.stderr` with our own `TeeHandler` class.

@OyvindTafjord pinging you so you can give this a review as well

## Example output

**With `--file-friendly-logging`**

image

**Without `--file-friendly-logging`**

image

And in both cases the `stderr.log` looks like this:

image

## Downsides

- Relies on patching `sys.stdout` and `sys.stderr`. This is scary.
- The tqdm progress bars are not as pretty. Compare the output above to the status bars before this PR:

image

We can partially fix this by setting `ascii=False` when initializing the Tqdm instance. With that, it looks like this:

image

But it's still not as pretty because it doesn't take up the entire terminal width.

### DIFF
--- a/allennlp/commands/elmo.py
+++ b/allennlp/commands/elmo.py
@@ -76,7 +76,8 @@ from overrides import overrides
 from allennlp.commands.subcommand import Subcommand
 from allennlp.common.checks import ConfigurationError
 from allennlp.common.tqdm import Tqdm
-from allennlp.common.util import lazy_groups_of, prepare_global_logging
+from allennlp.common.logging import prepare_global_logging
+from allennlp.common.util import lazy_groups_of
 from allennlp.data.token_indexers.elmo_indexer import ELMoTokenCharactersIndexer
 from allennlp.modules.elmo import _ElmoBiLm, batch_to_ids
 from allennlp.nn.util import remove_sentence_boundaries
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -49,6 +49,7 @@ from overrides import overrides
 from allennlp.commands.subcommand import Subcommand
 from allennlp.common import Params, Registrable, Lazy
 from allennlp.common.checks import check_for_gpu, ConfigurationError
+from allennlp.common.logging import prepare_global_logging
 from allennlp.common import util as common_util
 from allennlp.common.plugins import import_plugins
 from allennlp.data import DatasetReader, Vocabulary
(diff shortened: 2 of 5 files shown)


matt-gardner:
We used to have a tee logger; did that get removed at some point?  Was that the cause of #3785?

epwalsh:
I’m not sure when it was removed or why, but  I mostly copied it.

OyvindTafjord:
Thanks! Looks good to me, although I don't have an opinion on the scariness of the fix. I don't particularly care about the formatting of the progress bars in normal mode, although having it span the terminal width doesn't seem all that important (or even preferable?).

Looks like the tee logger was removed in the PR https://github.com/allenai/allennlp/pull/3529 quoted in the issue https://github.com/allenai/allennlp/issues/3785.

matt-gardner:
Oh, right.  @epwalsh, can you verify that this still works when using multiple GPUs?

epwalsh:
I don’t have an environment set up yet with multiple GPUs, but I’ll have time tomorrow to spin one up. Until then, if anyone else is able to test on multiple GPUs LMK

epwalsh:
Ok I gave it try on 2 GPUs on it looks like everything worked as expected. Here are some screen shots.

**Without `--file-friendly-logging` flag**

*Serialization directory*

image

*Terminal output during training*

image

*Worker 0 stderr log file*

image

*Worker 1 stderr log file*

image

*Worker 0 stdout log file*

image

*Worker 1 stdout log file*

image

**With `--file-friendly-logging` flag**

image

The log files looked the same as they did without the flag.

PR #4166, opened 2020-04-29
tweak torch requirement

epwalsh:
As it is the `setup.py` requires `torch>1.3.1,<=1.5.0`. But this excludes version `1.5.0+cu101` which is needed if you have cuda 10.1 and not cuda 10.2. So - as I found out today - if you have `1.5.0+cu101` installed and you try running an `allennlp` command it will completely fail with this:

image


setup.py
```diff
@@ -50,7 +50,7 @@
     license="Apache",
     packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"]),
     install_requires=[
-        "torch>1.3.1,<=1.5.0",
+        "torch>1.3.1,<1.5.1",
```
bryant1410, line 53: What about `<1.6.0`? PyTorch "patch" version changes shouldn't be breaking. But maybe you think it's better to have them CI-tested anyway?

epwalsh: Our CI will always test against the newest allowed package, so we'd know right away if a new patch version breaks anything. So yea, I agree with you, we should make it `<1.6.0`

PR #4169, opened 2020-04-29
Support for bart in allennlp-models

Tobias-Rohde:
- Added option to PretrainedTransformerEmbedder to allow usage with
encoder-decoder models and created unit test.
- Added ROUGE-N metric. ROUGE-L will follow soon
- Added indices to tokens abstract method to TokenIndexer. Implemented
for PretrainedTransformerIndexer. This is useful for turning decoded
sequences in seq2seq models into text.
- added timestep parameter to step function in beamsearch
- other minor changes


allennlp/training/metrics/bleu.py
```diff
@@ -88,13 +74,16 @@ def _get_modified_precision_counts(
         count of that ngram in the reference sentence. The denominator is just
         the total count of predicted ngrams.
         """
+        # TODO: fix not being able to import this normally
+        from allennlp.training.util import ngrams
```
dirkgr, line 78: What is this? A problem with circular imports?

allennlp/training/metrics/rouge.py
```diff
@@ -0,0 +1,120 @@
...
+        the total count of reference ngrams.
+        """
+        # TODO: fix not being able to import this normally
+        from allennlp.training.util import ngrams
+
+        matches = 0
+        total_reference = 0
+        for batch_num in range(predicted_tokens.size(0)):
+            predicted_row = predicted_tokens[batch_num, :]
+            reference_row = reference_tokens[batch_num, :]
```
dirkgr, line 65: You could just say `for predicted_row, reference_row in zip(predicted_tokens, reference_tokens):`.

dirkgr:
Oh, you already explained about ROUGE-L in the write-up. 👀

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -101,6 +101,43 @@ def tokens_to_indices(self, tokens: List[Token], vocabulary: Vocabulary) -> Inde
 
         return self._postprocess_output(output)
 
+    @overrides
+    def indices_to_tokens(
+        self, indexed_tokens: IndexedTokenList, vocabulary: Vocabulary
+    ) -> List[Token]:
+        token_ids = indexed_tokens["token_ids"]
+        type_ids = indexed_tokens["type_ids"] if "type_ids" in indexed_tokens else None
```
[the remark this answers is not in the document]
Tobias-Rohde: Done!

allennlp/nn/util.py
```diff
@@ -622,10 +624,10 @@ def get_text_field_mask(
     smallest_dim = tensor_dims[0][0] - num_wrapping_dims
     if smallest_dim == 2:
         token_tensor = tensor_dims[0][1]
-        return token_tensor != 0
+        return token_tensor != padding_id
     elif smallest_dim == 3:
         character_tensor = tensor_dims[0][1]
-        return (character_tensor > 0).any(dim=-1)
+        return (character_tensor != padding_id).any(dim=-1)
```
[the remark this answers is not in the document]
Tobias-Rohde: Do you mean specifically the change from > 0 to != or just the addition of a custom padding id in general? I think the two are equivalent since indices should never be negative.

I added two tests for using the custom padding id.

allennlp/training/metrics/bleu.py
Tobias-Rohde: It shows up as a generic ImportError, but it looks like a circular important. I'll keep looking into it.

allennlp/training/metrics/rouge.py
Tobias-Rohde: Thanks that looks better. I took this code from the bleu class, so I changed it in rouge and bleu.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -101,6 +101,43 @@ def tokens_to_indices(self, tokens: List[Token], vocabulary: Vocabulary) -> Inde
...
+        return [
+            Token(
+                text=vocabulary.get_token_from_index(token_ids[i], self._namespace),
+                text_id=token_ids[i],
+                type_id=type_ids[i] if type_ids is not None else None,
+            )
+            for i in range(len(token_ids))
+        ]
+
+    def _preprocess_segmented_tokens(self, tokens: List[Token]):
```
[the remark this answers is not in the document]
Tobias-Rohde: I agree that it's not a great solution, I rewrote it to simply remove all special tokens and add special tokens back in afterwards.

I think that in reality we won't ever get such sequences in `indices_to_tokens`. To my understanding max length is there to fold sequences that are too long for the given transformer into shorter segments. I think currently the only use case of max length is for the `PretrainedTransformerEmbedder`, which takes care of unfolding itself (besides that, those are embeddings and not ids anyways).

I am not sure yet how we should deal with models that BART that generate variable length output. What if we feed in a very long article that is split up into several segments and then text is generated for each segment? Each segment is missing context from other segments, so we can't expect the text generated from each segment to make sense, especially not if we would simply concatenate them together.

Anyways, the reason I included that code is to make `indices_to_tokens` a proper inverse
of `tokens_to_indices`. Otherwise, if you use `tokens_to_indices` with `max_length` and then use `indices_to_tokens`, you will get different output due to the special tokens.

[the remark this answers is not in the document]
dirkgr: Trying to make them perfect inverses of each other is a good idea, but I think it's too complex in this case. Let's remove that part of the code, and document that in the presence of `_max_length`, this is not a perfect inverse.

allennlp/nn/util.py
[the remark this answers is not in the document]
dirkgr: Thanks!

allennlp/training/metrics/rouge.py
```diff
@@ -32,43 +32,131 @@ class ROUGE(Metric):
...
+                else:
+                    cur_lcs[j] = max(cur_lcs[j + 1], prev_lcs[j])
+            prev_lcs = cur_lcs
+
+        return prev_lcs[0].item()
+
+    def _get_valid_tokens_mask(self, tensor: torch.LongTensor) -> torch.ByteTensor:
+        valid_tokens_mask = torch.ones_like(tensor, dtype=torch.bool)
+        for index in self._exclude_indices:
+            valid_tokens_mask = valid_tokens_mask & (tensor != index)
```
dirkgr, line 87: It'll be faster to say `valid_tokens_mask &= (tensor != index)`, because that avoids calculating an extra copy of `valid_tokens_mask`.

allennlp/training/metrics/rouge.py
```diff
@@ -32,43 +32,131 @@ class ROUGE(Metric):
...
+        prev_lcs = torch.zeros(n + 1, dtype=torch.long)
+
+        for i in range(m - 1, -1, -1):
+            # Make sure we don't count special tokens as part of the subsequences
+            if seq_1[i].item() in self._exclude_indices:
+                continue
+
+            cur_lcs = torch.zeros_like(prev_lcs)
+            for j in range(n - 1, -1, -1):
+                if seq_1[i] == seq_2[j]:
```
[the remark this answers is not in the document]
Tobias-Rohde: No, since the seq1 check assures that we won't have equality there if `seq2` contains one of the ids `_exclude_indices`. For the dynamic programming to work properly, we need still need to update the table `cur_lcs` at index `j`, even if `seq_2[j]` is an excluded id.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -106,9 +106,9 @@ def indices_to_tokens(
         self, indexed_tokens: IndexedTokenList, vocabulary: Vocabulary
     ) -> List[Token]:
         token_ids = indexed_tokens["token_ids"]
-        type_ids = indexed_tokens["type_ids"] if "type_ids" in indexed_tokens else None
+        type_ids = indexed_tokens.get("type_ids")
 
-        token_ids = self._preprocess_segmented_tokens(token_ids)
+        token_ids = self._preprocess_segmented_token_ids(token_ids)
```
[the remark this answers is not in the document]
Tobias-Rohde: Okay, I made this change and modified the unit test accordingly.

allennlp/training/metrics/rouge.py
Tobias-Rohde: Sounds good. I made the change and also moved the function to training util, since I originally copied it form our BLEU code.

allennlp/common/util.py
```diff
@@ -139,6 +139,32 @@ def lazy_groups_of(iterable: Iterable[A], group_size: int) -> Iterator[List[A]]:
...
+    iterable: Iterable[A], sizes: Iterable[int], max_size: int
+) -> Iterator[List[A]]:
+    cur_size = 0
+    group: List[A] = []
+
+    iterator = iter(iterable)
+    size_iter = iter(sizes)
+
+    while True:
+        item, size = next(iterator, None), next(size_iter, None)
```
dirkgr, line 152: Is this different from `for item, size in zip(iterator, size_iter):`?

allennlp/common/util.py
```diff
@@ -139,6 +139,32 @@ def lazy_groups_of(iterable: Iterable[A], group_size: int) -> Iterator[List[A]]:
             break
 
 
+def lazy_groups_of_max_size(
+    iterable: Iterable[A], sizes: Iterable[int], max_size: int
+) -> Iterator[List[A]]:
```
dirkgr, line 144: This will need documentation to live in `util.py`.

allennlp/data/samplers/bucket_batch_sampler.py
```diff
@@ -66,20 +70,24 @@ class BucketBatchSampler(BatchSampler):
...
+        batch_size: Optional[int] = None,
+        max_tokens: Optional[int] = None,
         sorting_keys: List[str] = None,
         padding_noise: float = 0.1,
         drop_last: bool = False,
     ):
-
+        assert (batch_size is None) != (max_tokens is None), (
+            "Need to provide exactly one of" "batch_size or max_tokens"
+        )
```
dirkgr, line 81: Something happened here with the formatting.

allennlp/data/samplers/bucket_batch_sampler.py
```diff
@@ -33,8 +33,12 @@ class BucketBatchSampler(BatchSampler):
...
         The pytorch `Dataset` of allennlp Instances to bucket.
 
-    batch_size : `int`, required.
-        The size of each batch of instances yielded when calling the dataloader.
+    batch_size : `int`, optional (default = None)
+        The size of each batch of instances yielded when calling the dataloader. Cannot be used with
+        `max_tokens`.
+
+    max_tokens : `int`, optional (default = None)
+        The maximum number of tokens to include in a batch. Cannot be used with `batch_size`.
```
dirkgr, line 41: You added this because huggingface uses this?

allennlp/data/samplers/bucket_batch_sampler.py
```diff
@@ -92,28 +100,41 @@ def _argsort_by_padding(self, instances: Iterable[Instance]) -> List[int]:
...
                         f'Sorting key "{field_name}" is not a field in instance. '
                         f"Available fields/keys are {list(instance.fields.keys())}."
                     )
-                lengths.append(
-                    add_noise_to_value(len(instance.fields[field_name]), self.padding_noise)
-                )
-            instances_with_lengths.append((lengths, instance))
+                true_lengths.append(len(instance.fields[field_name]))
+
+                lengths.append(add_noise_to_value(true_lengths[-1], self.padding_noise))
```
dirkgr, line 113: It's a small thing, but I'd rather `lengths` was called `noisy_length`, and the other one just `length`.

allennlp/data/samplers/bucket_batch_sampler.py
dirkgr: After having read the whole thing, I don't like that this is part of `BucketIterator`. I think it should be its own type of iterator.

Of course, for pure research code, adding this functionality like this would be great. But this is not that case. This code actually has many users, and most of them never care about the `max_tokens` scenario, so I want to keep the API clean and easy for those guys.

allennlp/nn/beam_search.py
```diff
@@ -111,7 +135,8 @@ def search(
         # beam to `beam_size`^2 candidates from which we will select the top
         # `beam_size` elements for the next iteration.
         # shape: (batch_size, num_classes)
-        start_class_log_probabilities, state = step(start_predictions, start_state, 0)
+        with torch.no_grad():
```
dirkgr, line 138: If you never want any grads in here, you can annotate the whole function with `no_grad`.

allennlp/training/learning_rate_schedulers/polynomial_decay.py
```diff
@@ -0,0 +1,78 @@
...
+        current_decay_steps = self.total_steps - self.steps
+        total_decay_steps = self.total_steps - self.warmup_steps
+        f = (current_decay_steps / total_decay_steps) ** self.power
+        return [
+            f * (lr - self.end_learning_rate) + self.end_learning_rate for lr in self.base_values
+        ]
+
+    @overrides
+    def step(self, metric: float = None) -> None:
+        pass
```
dirkgr, line 68: How come this function isn't needed?

allennlp/training/metrics/bleu.py
```diff
@@ -120,6 +114,9 @@ def __call__(
 
         None
         """
+        # TODO: fix not being able to import this normally
```
dirkgr, line 117: Any progress?

allennlp/training/learning_rate_schedulers/polynomial_decay.py
dirkgr: We cleared that up in our call just now.

allennlp/training/metrics/bleu.py
Tobias-Rohde: Finally fixed this. It turns out that using `import x as y` works much more reliably than `from x import z`. Originally I tracked down where exactly the imports are circular and I thought maybe it can be resolved by moving some functions to a different module. But from what I can tell the circularity makes sense here so I just used:

`import allennlp.training.util as training_util`

allennlp/nn/beam_search.py
Tobias-Rohde: Neat, I didn't know this.

allennlp/data/samplers/bucket_batch_sampler.py
Tobias-Rohde: I created a subclass of `BucketBatchSampler` called `MaxTokensBatchSampler` and reverted `BucketBatchSampler`. I kept the changes for the argsort `_argsort_by_padding`, since this only returns the lengths alongside the indices.

allennlp/data/samplers/bucket_batch_sampler.py
Tobias-Rohde: `black .` did this for me. `black .` has been doing some odd formatting changes for me recently. For instance, this:

```
            m = training_util.get_valid_tokens_mask(reference_seq, self._exclude_indices).sum().item()
            n = training_util.get_valid_tokens_mask(predicted_seq, self._exclude_indices).sum().item()
```

turned into this

```
            m = (
                training_util.get_valid_tokens_mask(reference_seq, self._exclude_indices)
                .sum()
                .item()
            )
            n = (
                training_util.get_valid_tokens_mask(predicted_seq, self._exclude_indices)
                .sum()
                .item()
            )
```

allennlp/common/util.py
Tobias-Rohde: The reason I did this was that I need to finalize the last batch after the iterators are done and I wanted to just have a single loop. I rewrote it with zip and ended up finding a much neater way to create the last batch.

allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -90,15 +90,15 @@ def __init__(
             model_name, add_special_tokens=True, **tokenizer_kwargs
         )
         dummy_output = tokenizer_with_special_tokens.encode_plus(
-            "1",  # "a" and "b" get tokenized into multiple word pieces, but for now, "1" and "2" do not.
-            "2",
+            " 1",  # "a" and "b" get tokenized into multiple word pieces, but for now, "1" and "2" do not.
+            " 2",
```
dirkgr, line 94: @Tobias-Rohde, what is this for?

allennlp/tests/data/token_indexers/pretrained_transformer_indexer_test.py
```diff
@@ -165,3 +165,36 @@ def test_long_sequence_splitting(self):
         assert indexed["token_ids"] == expected_ids
         assert indexed["segment_concat_mask"] == [True] * len(expected_ids)
         assert indexed["mask"] == [True] * 7  # original length
+
+    @staticmethod
+    def _assert_tokens_equal(expected_tokens, actual_tokens):
+        for expected, actual in zip(expected_tokens, actual_tokens):
+            assert expected.text == actual.text
+            assert expected.text_id == actual.text_id
+            assert expected.type_id == actual.type_id
```
dirkgr, line 174: Is that necessary? Just comparing the tokens doesn't work?

PR #4172, opened 2020-04-30
Use new env var in the allennlp-models build

dirkgr:


epwalsh:
Going to merge this now so the nightly workflow will have it

PR #4175, opened 2020-04-30
Another missing param `eps` in lr_scheduler

YKX-A:
Just now, I found the `eps` is also missed in the initialization of `ReduceOnPlateauLearningRateScheduler.lr_scheduler`. 
Now, I think all the parameters here are being used.


YKX-A:
Looks like something wrong, I'll retry.

PR #4179, opened 2020-04-30
Find the right embedding layer for mismatched cases

matt-gardner:
I was diagnosing an error in the interpret code for a scibert NER model, and found this.  Comments in the code should explain the problem and solution well enough, let me know if they don't.

### DIFF
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -1658,11 +1658,30 @@ def find_embedding_layer(model: torch.nn.Module) -> torch.nn.Module:
     )
     from allennlp.modules.token_embedders.embedding import Embedding
 
+    # The special case only works if we don't have mismatched embedding.  What we essentially want
+    # to do is grab a transformer's wordpiece embedding layer, because using that is a lot easier
+    # for something like hotflip than running a network to embed all tokens in a vocabulary.  If
+    # you've used a mismatched embedder, though, your input tokens are actually *words*, so that's
+    # what we'll be visualizing and attacking, even though you're modeling things at the wordpiece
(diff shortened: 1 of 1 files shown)


matt-gardner:
The model test failure here is actually a bug in the test that I need to fix.  It's a one-line change in the model repo.  What's the process for doing this, given that we're running model tests here, and I can't add a fix to this PR?

epwalsh:
@matt-gardner just make a separate PR in the models repo. We can merge that into master and then re-run this workflow

matt-gardner:
But that PR will fail tests without this change.

epwalsh:
You could temporarily update the workflow in models CI to use your patch branch by modifying the `requirements.txt` file

matt-gardner:
That seems pretty bad - I wouldn't want to merge such a PR into that repo.  I can see modifying the allennlp commit in the models repo to demonstrate that we have two PRs, one on each repo, that both work, before either one gets merged.  But I think we're going to have to live with the fact the we'll have a short window where one or the other CI is broken.  I'm not sure there's a reasonable alternative.  @dirkgr, do you see a way around this, either?

epwalsh:
Personally I'm okay with one repo being broken for a few minutes

matt-gardner:
Me too.  I'll make another PR for the model repo, to show that we have a fix for the breakage in this PR, then after this PR gets merged I can update the other PR to remove the CI hack, and merge the fix for the model repo tests.

matt-gardner:
Ok, corresponding model PR: https://github.com/allenai/allennlp-models/pull/38.

PR #4180, opened 2020-04-30
Add self-hosted runner for GPU checks

### LINKED ISSUE #4108: Set up self hosted GitHub Actions runner in GCP for GPU tests

epwalsh:
closes #4108

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -18,9 +18,24 @@ LABEL com.nvidia.volumes.needed="nvidia_driver"
 
 WORKDIR /stage/allennlp
 
+# Add minimal files necessary to install dependencies.
+COPY allennlp/version.py allennlp/version.py
+RUN touch allennlp/__init__.py
 COPY setup.py setup.py
 COPY dev-requirements.txt dev-requirements.txt
 
-RUN pip install torch
+# Install deps.
+RUN pip install --no-cache-dir -e . && \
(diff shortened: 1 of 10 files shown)


Dockerfile.test
```diff
@@ -0,0 +1,41 @@
...
+# use all available devices, which are mounted at /usr/local/nvidia.
+# The LABEL supports an older version of nvidia-docker, the env
+# variables a newer one.
+ENV NVIDIA_VISIBLE_DEVICES all
+ENV NVIDIA_DRIVER_CAPABILITIES compute,utility
+LABEL com.nvidia.volumes.needed="nvidia_driver"
+
+WORKDIR /stage/allennlp
+
+# Add minimal files necessary to install dependencies.
```
dirkgr, line 21: Why? So that we have a docker layer that captures all dependencies?

If that's the case, can we also use this trick in the other Dockerfiles?

ucalyptus:
@dirkgr is it possible to use google colab as a self hosted runner? Out of context, but just asking.

dirkgr:
I don't know much about Google colab. Maybe @epwalsh knows? Do you just need a Docker image to get going in colab?

epwalsh:
@ucalyptus I imagine it's possible. You could turn anything into a self hosted runner as long as you can download and execute the runner binary, and the firewall doesn't block outgoing connections.

PR #4184, opened 2020-05-01
Fix heuristic in util.get_token_ids_from_text_field_tensors

matt-gardner:
This looks like a much bigger change than it actually is.  The bulk of the line changes are a little bit of refactoring in `hotflip.py` that I did while trying to diagnose an error in the demo.  I didn't end up needing to change `Hotflip` or add any tests for it (though I originally had some new tests there), but I left the refactoring, as I think it's an improvement.

The problem in the demo boiled down to a failure of our heuristic in `util.get_token_ids_from_text_field_tensors`.  The point of that method is to take a potentially complex input and guess which part of it corresponds to actual single token ids, that you can pass to your vocabulary and convert into strings, for whatever reason you might want that (there are also some uses in seq2seq models).  The heuristic relied on some naming conventions, which it turns out were broken from the start when using elmo, though I didn't realize it.  This PR fixes the heuristic by changing the name of elmo-internal indexer / embedder API.  Instead of using the key "tokens" as an argument, it uses "elmo_tokens", as this doesn't overlap with what the single ID indexer uses, so it's easy to guess what the right token ids are.  I fixed all of the tests in the core repo that this broke, and there were no other consequences here for changing the name, other than a small handful of places where that name was hard-coded (mostly in tests).  It's possible that there are one or two places in the models repo that also need to change; we'll see if this breaks any of those tests.


matt-gardner:
Ok, so there are a few failures in the model repo, and I'll have to do something similar to what #4179 did, with two parallel PRs.  That PR is here: https://github.com/allenai/allennlp-models/pull/39.

scripts/train_fixtures.py
```diff
@@ -13,7 +13,6 @@
 from allennlp.commands.train import train_model_from_file, train_model
```
matt-gardner, line 13: I simplified this file a bit because most of the models have moved.  It looks like a corresponding script doesn't exist in the models repo yet, though.  Might be nice to move this over at some point.  Though I suppose if someone wants this, they can just find the old version of this file from a prior version, and copy it over when they need it...

matt-gardner:
This is not something that ever showed up in a config.  At least, not after the fix to sorting keys.

dirkgr:
I force-merged this. Will now check that https://github.com/allenai/allennlp-models/pull/39/files goes green.

PR #4188, opened 2020-05-01
remove namespace plugin mechanism

### LINKED ISSUE #4187: Change the plugin mechanism
We want two changes:
 * Remove the namespace plugin approach to loading allennlp-models. Keep the config file approach.
 * try-import `allennlp-models` in the core repo.

We will recommend to third-parties to change the config file as part of their `pip install`.

epwalsh:
Closes  #4187. Also see https://github.com/allenai/allennlp-models/pull/40. This should be merged first.

Most of the diff is just removing test fixtures we don't need anymore.


allennlp/common/plugins.py
```diff
@@ -1,52 +1,23 @@
...
-
-        if reload:
-            importlib.reload(namespace_module)
-
-        return pkgutil.iter_modules(
-            namespace_module.__path__, namespace_module.__name__ + "."  # type: ignore
-        )
-    except ModuleNotFoundError:
-        return []
+DEFAULT_PLUGINS = ("allennlp_models",)
```
dirkgr, line 20: Why is this a tuple? Isn't that more of a list kind of thing?

epwalsh: I just think tuples are better than lists for globals like this since tuples are immutable and take less memory

PR #4197, opened 2020-05-04
attach allennlp cache to Docker images and fix apex test

epwalsh:
About the apex test:

Calling `mp.spawn` fails because something is not pickle-able. But this isn't needed anyway because of https://github.com/allenai/allennlp/pull/3919.

This slipped by before because the Apex test wasn't ran anywhere. But adding the `@pytest.mark.gpu` annotation to that test ensures it's now run in the GPU Checks workflow.


Makefile
```diff
@@ -13,6 +13,7 @@ MD_DOCS_EXTRAS = $(addprefix $(MD_DOCS_ROOT),README.md LICENSE.md ROADMAP.md CON
 DOCKER_LABEL    = latest
 DOCKER_TAG      = allennlp/allennlp:$(DOCKER_LABEL)
 DOCKER_TEST_TAG = allennlp/test:$(DOCKER_LABEL)
+DOCKER_RUN_CMD  = docker run --rm -v $$HOME/.allennlp:/root/.allennlp
```
dirkgr, line 16: This works even on the GitHub runners?

epwalsh: Well only with our self-hosted runner, since otherwise `$HOME/.allennlp` would be empty

PR #4202, opened 2020-05-06
Separate linting from formatting in CI, always run all steps of workflow

epwalsh:
I think it's more clear if we separate flake8 linting from black formatting into two separate steps of the CI workflow. This also ensures all steps of the "Check Core" workflow run even if one fails.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -62,15 +62,23 @@ jobs:
       run: |
         pip freeze
 
+    - name: Format
+      if: always()
+      run: |
+        make format
+
     - name: Lint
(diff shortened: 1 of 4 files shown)


.github/workflows/master.yml
```diff
@@ -62,15 +62,23 @@ jobs:
       run: |
         pip freeze
 
+    - name: Format
+      if: always()
```
dirkgr, line 66: Why do you have to say that?

epwalsh: So that even if one step fails, the others will be still be ran

PR #4204, opened 2020-05-07
[PROPOSAL] Switch to pytest style test classes, use plain asserts

epwalsh:
As far as I can tell, since we're using `pytest` to run tests, there's no benefit of using `unittest.TestCase` as the base class for `AllenNlpTestCase` as opposed to plain [`pytest` style classes](https://docs.pytest.org/en/latest/xunit_setup.html?highlight=setup_class#class-level-setup-teardown).

There is, however, several disadvantages:
1.  [`@pytest.mark.parametrize` does not work with methods on a `unittest.TestCase` class](https://stackoverflow.com/questions/18182251/does-pytest-parametrized-test-work-with-unittest-class-based-tests). Given that parametrizing tests is such a powerful tool, I think this is a major issue.
2. `unittest.TestCase` assertion methods (`self.assertEqual`, `self.assertSetEqual`, etc) are less concise than plain `assert` statements, which are recommended by `pytest`.
3. The `setUp` and `tearDown` methods do not have [PEP-8](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables) compatible names, and it's not clear when these methods are run based on their names. On the other hand, the `pytest` equivalents (`setup_method` and `teardown_method` or `setup_class` and `teardown_class`) have snake-case names that are perfectly clear as to when they run.

---

I know this looks like a huge PR but it's mostly just replacing `unittest` assertion methods with the `pytest` equivalents.

### DIFF
--- a/allennlp/common/testing/test_case.py
+++ b/allennlp/common/testing/test_case.py
@@ -61,7 +61,7 @@ class AllenNlpTestCase:
 
 _available_devices = ["cpu"] + (["cuda"] if torch.cuda.is_available() else [])
 
-multi_device = pytest.mark.parametrize("device", _available_devices)
+multi_device = lambda f: pytest.mark.parametrize("device", _available_devices)(pytest.mark.gpu(f))
 """
 Decorator that provides an argument `device` of type `str` for each available PyTorch device.
 """
--- a/allennlp/common/testing/test_case.py
+++ b/allennlp/common/testing/test_case.py
@@ -3,17 +3,17 @@ import os
 import pathlib
 import shutil
 import tempfile
-from typing import Any, Iterable
-from unittest import TestCase, mock
+from unittest import mock
 
 import torch
+import pytest
 
 from allennlp.common.checks import log_pytorch_version_info
 
 TEST_DIR = tempfile.mkdtemp(prefix="allennlp_tests")
 
 
-class AllenNlpTestCase(TestCase):
+class AllenNlpTestCase:
     """
     A custom subclass of `unittest.TestCase` that disables some of the more verbose AllenNLP
     logging and that creates and destroys a temp directory as a test fixture.
@@ -25,7 +25,7 @@ class AllenNlpTestCase(TestCase):
     TESTS_ROOT = MODULE_ROOT / "tests"
     FIXTURES_ROOT = TESTS_ROOT / "fixtures"
 
-    def setUp(self):
+    def setup_method(self):
         logging.basicConfig(
             format="%(asctime)s - %(levelname)s - %(name)s - %(message)s", level=logging.DEBUG
         )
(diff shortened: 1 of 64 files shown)


epwalsh:
The "Check Models" workflow will fail until the `unittest` assertion methods in the models repo are replaced with the pytest equivalents.

allennlp/common/testing/test_case.py
```diff
@@ -54,37 +54,14 @@ def _cleanup_archive_dir_without_logging(path: str):
...
-
-        return wrapper
-
-    return decorator
-
-
 _available_devices = ["cpu"] + (["cuda"] if torch.cuda.is_available() else [])
-multi_device = parametrize(("device",), [(device,) for device in _available_devices])
+
+multi_device = lambda f: pytest.mark.parametrize("device", _available_devices)(pytest.mark.gpu(f))
```
epwalsh, line 64: The `pytest.mark.gpu(f)` ensures these tests are ran in the "GPU Checks" workflow

dirkgr:
I have been wanting to do this for a while. It's especially confusing in the models repo, because `AllenNlpTestCase` supplies some functions that don't work there.

allennlp/tests/commands/docstring_help_test.py
```diff
@@ -42,20 +42,16 @@ def test_docstring_help(self):
                 subcommand = match.group(2)
                 actual_output = _subcommand_help_output(subcommand)
 
-                self.assertEqual(
-                    expected_output,
-                    actual_output,
+                assert expected_output == actual_output, (
```
dirkgr, line 45: Will this give me those pretty comparisons, that show immediately where the difference is, even in large data?

epwalsh: Yes sir!

epwalsh:
> This is a bit different than what I thought you were doing. I thought you were changing everything to top-level functions named `test_*`, with common setup and tear-down in [pytest fixtures](https://docs.pytest.org/en/latest/fixture.html#fixture). The base class `AllenNlpTestCase` would then go away. Any reason not to do that?

I think having a base class is beneficial because it wraps up all of its fixtures. Without it, you would have to import the fixtures (the temp directory and whatever else the base class comes with) to their test module or to a `conftest.py` at the same directory level as their test module, because there is no way of declaring global fixtures that work for all test modules without explicitly importing them.

That said, if you don't need those fixtures then it's just unnecessary overhead. So we could probably just use `def test_*` functions in many places.

allennlp/tests/data/fields/sequence_label_field_test.py
```diff
@@ -65,25 +66,25 @@ def test_sequence_label_field_raises_on_incorrect_type(self):
         with pytest.raises(ConfigurationError):
             _ = SequenceLabelField([[], [], [], [], []], self.text)
 
-    def test_class_variables_for_namespace_warnings_work_correctly(self):
+    def test_class_variables_for_namespace_warnings_work_correctly(self, caplog):
```
matt-gardner, line 69: Nit: `caplog` is not a great variable name.

allennlp/tests/training/learning_rate_schedulers/learning_rate_scheduler_test.py
```diff
@@ -8,19 +9,19 @@
...
 class LearningRateSchedulersTest(AllenNlpTestCase):
-    def setUp(self):
-        super().setUp()
+    def setup_method(self):
+        super().setup_method()
         self.model = torch.nn.Sequential(torch.nn.Linear(10, 10))
 
     def test_reduce_on_plateau_error_throw_when_no_metrics_exist(self):
-        with self.assertRaises(ConfigurationError) as context:
+        with pytest.raises(ConfigurationError) as context:
```
matt-gardner, line 17: I think this should just be something like `match="learning rate ..."`, instead of `as context`, right?

allennlp/tests/data/fields/sequence_label_field_test.py
epwalsh: This is actually an out-of-the-box pytest fixture so I unfortunately I think we're stuck with that name

epwalsh: https://docs.pytest.org/en/latest/logging.html#caplog-fixture

allennlp/tests/training/learning_rate_schedulers/learning_rate_scheduler_test.py
epwalsh: Good point. Just updated

allennlp/tests/data/fields/sequence_label_field_test.py
matt-gardner: Oh, hmm, didn't realize that pytest did this magic. Ok 🤷‍♂️

PR #4210, opened 2020-05-08
improve Docker workflows

epwalsh:
- Uses correct naming convention. Image tag = "latest", image name = "allennlp/allennlp:latest", for example
- Doesn't push "latest" tag for commit images, since if a previous workflow was re-ran, it would override the "latest" image.


epwalsh:
Merging now. Should be good, but the only way to know for sure if the changes to the master workflow work is by.. well, merging.

PR #4240, opened 2020-05-14
ensure nightly builds don't run for forks

### LINKED ISSUE #4239: Master workflow is running on forks again
**Describe the bug**
I think the check that was put in place to fix #3931 has gotten removed at some point, because the master workflow is running on my fork again.

epwalsh:
closes #4239

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -36,6 +36,8 @@ jobs:
 
   check_core:
     name: Check Core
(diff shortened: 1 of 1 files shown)


epwalsh:
I wish there were a way to specify that the `schedule` event should only trigger runs on the main repo, but this currently isn't possible, so we have to throw in a check for each individual job in the workflow.

PR #4243, opened 2020-05-15
Make character tokenizer in token character indexer configurable.

mklimasz:
In the current form of TokenCharactersIndexer it is impossible to modify it's character_tokenizer to own, custom implementation as shown in the script below.
It ends with ```allennlp.common.checks.ConfigurationError: Extra parameters passed to CharacterTokenizer: {'type': 'custom_character'}```.

I believe that this not desired design so I have changed the type (however I kept default value).
```python
import json

from allennlp.data import Tokenizer, TokenIndexer
from allennlp.data.tokenizers import CharacterTokenizer


@Tokenizer.register("custom_character")
class CustomCharacterTokenizer(CharacterTokenizer):
    pass


char_token_indexer = TokenIndexer.from_params(Params(json.loads("""
{
    "type": "characters",
    "character_tokenizer": {
        "type": "custom_character",
        "start_tokens": ["__START__"],
        "end_tokens": ["__END__"],
        "lowercase_characters": false
    }
}
""")))
```

### DIFF
--- a/allennlp/data/token_indexers/token_characters_indexer.py
+++ b/allennlp/data/token_indexers/token_characters_indexer.py
@@ -8,6 +8,7 @@ import torch
 from allennlp.common.checks import ConfigurationError
 from allennlp.common.util import pad_sequence_to_length
 from allennlp.data.token_indexers.token_indexer import TokenIndexer, IndexedTokenList
+from allennlp.data.tokenizers import Tokenizer
 from allennlp.data.tokenizers.character_tokenizer import CharacterTokenizer
 from allennlp.data.tokenizers.token import Token
 from allennlp.data.vocabulary import Vocabulary
@@ -44,7 +45,7 @@ class TokenCharactersIndexer(TokenIndexer):
     def __init__(
         self,
         namespace: str = "token_characters",
-        character_tokenizer: CharacterTokenizer = CharacterTokenizer(),
+        character_tokenizer: Tokenizer = CharacterTokenizer(),
         start_tokens: List[str] = None,
         end_tokens: List[str] = None,
         min_padding_length: int = 0,


mklimasz:
I guess (by looking at errors from allennlp-models that this requires adding ```type: character``` in configuration).

I guess I requires some green light, but when given I can prepare PR there.

matt-gardner:
What if instead of changing the type annotation, you change your call to `register()` to be `CharacterTokenizer.register("custom_character")`?  I think that should work, without requiring any code changes to allennlp.

mklimasz:
I do not think it works.

```python
import json

from allennlp.data import TokenIndexer
from allennlp.data.tokenizers import CharacterTokenizer


@CharacterTokenizer.register("custom_character")
class CustomCharacterTokenizer(CharacterTokenizer):
    pass


char_token_indexer = TokenIndexer.from_params(Params(json.loads("""
{
    "type": "characters",
    "character_tokenizer": {
        "type": "custom_character",
        "start_tokens": ["__START__"],
        "end_tokens": ["__END__"],
        "lowercase_characters": false
    }
}
""")))
```

It will end with ```allennlp.common.checks.ConfigurationError: Default implementation spacy is not registered```

There is workaround if at the begging of my script I override ```CharacterTokenizer.default_implementation = "custom_character"``` however it doesn't seem correct to leave it this way.

matt-gardner:
Actually, that workaround is right, though I would probably change the default to "character", and I would put it right next to your `CharacterTokenizer.register()` call, with a comment about why it's there.

We could maybe take a PR to add that `default_implementation` line to the `CharacterTokenizer` class, but this is a rare enough use case that I'd probably instead recommend that you just resolve it locally in your code, because it's easy to do with that one line of code.

mklimasz:
Thanks for help :)

PR #4253, opened 2020-05-18
make 'cached_path' work offline

epwalsh:
This also does some additional clean up to `file_utils.py`, in particular renaming the internal functions of the module with a leading underscore so that they won't be included in the API docs.

PR #4254, opened 2020-05-19
Don't re-serialize vocab to file in train command

epwalsh:
Currently the train command does the following WRT the vocabulary in the distributed setting:
- initializes the vocab from the main process and serializes it to file,
- spawns the `_train_worker` command for each worker, which themselves re-initialize the vocab from file,
- then the master worker (or just the main worker if not distributed) re-serializes the vocab to file after initializing the model in case the model has modified the vocab in some way.

But this is really, *really* bad because a non-master worker could be initializing their vocab from the files that are currently being written to from the master worker. In fact, I believe this what caused this CI run to fail: https://github.com/allenai/allennlp/pull/4253/checks?check_run_id=687029330.

---

Update:

Just removing the second vocab serialization broke a lot of test cases. So I guess there is more of a need for this than I thought.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,3 +150,13 @@ docker-test-image :
 .PHONY : docker-test-run
 docker-test-run :
 	$(DOCKER_RUN_CMD) --gpus 2 $(DOCKER_TEST_IMAGE_NAME) $(ARGS)
+
+.PHONY : dist-test
+dist-test :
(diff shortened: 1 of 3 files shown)


epwalsh:
Ok, this was not the right solution. I'm closing this PR but making an issue to track the bug: https://github.com/allenai/allennlp/issues/4255

Issue #4255, opened 2020-05-19
Distributed training bug


epwalsh:
Currently the train command does the following WRT the vocabulary in the distributed setting:
- initializes the vocab from the main process and serializes it to file,
- spawns the `_train_worker` command for each worker, which themselves re-initialize the vocab from file,
- then the master worker re-serializes the vocab to file after initializing the model in case the model has modified the vocab in some way.

But this is potentially really bad because a non-master worker could be initializing their vocab from the files that are currently being written to from the master worker. In fact, I believe this what caused this CI run to fail: https://github.com/allenai/allennlp/pull/4253/checks?check_run_id=687029330.

---

Potential solutions:

1. Only serialize the vocab once, and do it *before* initializing the model.

    This has a quick fix. See https://github.com/allenai/allennlp/pull/4256 for a draft.

    The downside here is that model side-effects on the vocab won't be saved to the vocab files. Is this really a downside? I feel like models should not modify the vocab, and I only know of one that does: CopyNet. But it's really not necessary. See https://github.com/allenai/allennlp-models/pull/55.

2. Refactor the train command so that even in the distributed setting we only serialize the vocab once, but do it *after* initializing the model. This will definitely take more work.

PR #4256, opened 2020-05-19
Only serialize vocab once, and do it before model initialization

epwalsh:
See https://github.com/allenai/allennlp/issues/4255


epwalsh:
Closing in favor of #4274 for now

PR #4260, opened 2020-05-19
Add a CHANGELOG

epwalsh:


.github/workflows/pull_request.yml
```diff
@@ -6,6 +6,23 @@ on:
     - master
 
 jobs:
+  changelog:
+    name: CHANGELOG
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
```
bryant1410, line 14:
```suggestion
    - uses: actions/checkout@v2
```

:)

epwalsh: `git diff --name-only $(git merge-base origin/master HEAD)` will actually fail if we use v2.

epwalsh: https://github.com/allenai/allennlp/pull/4260/checks?sha=0b7d61bb9538094980769b1e34338909a7af778a

epwalsh: I didn't dig into this too much because I was being lazy. But if there's an easy way to make it work with v2 I'm all ears

bryant1410: Ohh :+1:  Do you think it's worth it to add a comment?

epwalsh: Definitely!

PR #4263, opened 2020-05-19
Clean up the command to create a commit list.

schmmd:
I'm not really sure what the previous command does.  It seems easier if we simplify the syntax to specify the version range.  I think it will make it less likely to make a mistake.

### DIFF
--- a/RELEASE_PROCESS.md
+++ b/RELEASE_PROCESS.md
@@ -38,16 +38,10 @@ Then you can copy and paste the commands below without worrying about mistyping 
     ```
 
 4. Find the tag you just pushed [on GitHub](https://github.com/allenai/allennlp/tags) and
-click edit. Now copy over the latest section from the `CHANGELOG.md`. And finally, add a section called "Commits" with the output from this:
+click edit. Now copy over the latest section from the `CHANGELOG.md`. And finally, add a section called "Commits" with the output a command like the following:
 
     ```bash
-    git log `git describe --always --tags --abbrev=0 HEAD^^`..HEAD^ --oneline
-    ```
-
-    Or, if you're using fish,
-
-    ```fish
-    git log (git describe --always --tags --abbrev=0 HEAD^^)..HEAD^ --oneline
+    git log v0.9.0..v1.0.0rc4 --oneline
     ```
 
     On a Mac, for example, you can just pipe the above command into `pbcopy`.
--- a/RELEASE_PROCESS.md
+++ b/RELEASE_PROCESS.md
@@ -44,6 +44,11 @@ click edit. Now copy over the latest section from the `CHANGELOG.md`. And finall
     OLD_TAG=$(git describe --always --tags --abbrev=0 $TAG^)
     git log $OLD_TAG..$TAG --oneline
     ```
+    
+    ```fish
+    set -x OLD_TAG $(git describe --always --tags --abbrev=0 $TAG^)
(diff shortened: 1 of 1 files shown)


RELEASE_PROCESS.md
```diff
@@ -38,16 +38,10 @@ Then you can copy and paste the commands below without worrying about mistyping
...
 
     ```bash
-    git log `git describe --always --tags --abbrev=0 HEAD^^`..HEAD^ --oneline
-    ```
-
-    Or, if you're using fish,
-
-    ```fish
-    git log (git describe --always --tags --abbrev=0 HEAD^^)..HEAD^ --oneline
+    git log v0.9.0..v1.0.0rc4 --oneline
```
epwalsh, line 44: I feel like this actually makes it more error prone since you can't just copy and paste this commad.. you'd have to edit it to change the tags.

And `git describe --always --tags --abbrev=0 HEAD^^` just prints the name of the previous tag.

Would doing

```suggestion
    git log `git describe --always --tags --abbrev=0 HEAD^^`..$TAG --oneline
```

make it more clear?

epwalsh, line 44: Or a two-liner:

```suggestion
    OLD_TAG=$(git describe --always --tags --abbrev=0 $TAG^)
    git log $OLD_TAG..$TAG --oneline
```

RELEASE_PROCESS.md
```diff
@@ -38,16 +38,10 @@ Then you can copy and paste the commands below without worrying about mistyping
...
+click edit. Now copy over the latest section from the `CHANGELOG.md`. And finally, add a section called "Commits" with the output a command like the following:
 
     ```bash
-    git log `git describe --always --tags --abbrev=0 HEAD^^`..HEAD^ --oneline
-    ```
-
-    Or, if you're using fish,
-
-    ```fish
-    git log (git describe --always --tags --abbrev=0 HEAD^^)..HEAD^ --oneline
```
epwalsh, line 50: I like having this because I'm a fisher, so I can just copy and paste it. That said, I'm not opposed to removing it. I could always just stuff this into a function in my dotfiles.

schmmd: I am too.  I only deleted it because my change made the command identical for fish and bash.

RELEASE_PROCESS.md
schmmd: Yes, substantially.  It would also reduce error, as the previous command depends on the commit you're at.

RELEASE_PROCESS.md
schmmd: Pete--I added a fish command back in.  I recommend taking a second look.

RELEASE_PROCESS.md
```diff
@@ -38,16 +38,16 @@ Then you can copy and paste the commands below without worrying about mistyping
     ```
 
 4. Find the tag you just pushed [on GitHub](https://github.com/allenai/allennlp/tags) and
-click edit. Now copy over the latest section from the `CHANGELOG.md`. And finally, add a section called "Commits" with the output from this:
+click edit. Now copy over the latest section from the `CHANGELOG.md`. And finally, add a section called "Commits" with the output a command like the following:
```
epwalsh, line 41:
```suggestion
click edit. Now copy over the latest section from the `CHANGELOG.md`. And finally, add a section called "Commits" with the output of a command like the following:
```

PR #4264, opened 2020-05-19
move test and fixtures to root level and simplify test-install command

### LINKED ISSUE #4252: Do we still need the test-install command?
`allennlp test-install` just runs `pytest` in a roundabout way, and adds extra maintenance work since we have to make sure [this function](https://github.com/allenai/allennlp/blob/master/allennlp/commands/test_install.py#L52) stays up-to-date with our pytest configuration in `Makefile` and `pytest.ini`.

It's also redundant (in our CI at least) to be running the unit tests again, and I don't how useful this command is to end users. In theory, if they (the users) can successfully install AllenNLP on an officially supported OS and Python version, they shouldn't need to run a bunch of unit tests to verify that their installation works. They should trust that we've already done that (and we have).

At most I would suggest they just run `allennlp --help` to make sure the binary has been installed and torch and what-not can be successfully loaded.

epwalsh:
Closes #4252. We keep the `test-install` comman, but it will just print out some useful info and ensure all of the core submodules can be imported.

There are a few places in our CI that actually use the `test-install` command, so for now I think this is the easiest solution. And it's important to have a check like this so that we can catch certain errors like forgetting to include a library dependency in `setup.py`.


allennlp/commands/test_install.py
```diff
@@ -50,23 +36,14 @@ def _get_module_root():
...
+    import_module_and_submodules("allennlp.common")
+    import_module_and_submodules("allennlp.data")
+    import_module_and_submodules("allennlp.interpret")
+    import_module_and_submodules("allennlp.models")
+    import_module_and_submodules("allennlp.modules")
+    import_module_and_submodules("allennlp.nn")
+    import_module_and_submodules("allennlp.predictors")
+    import_module_and_submodules("allennlp.training")
+    logger.info("AllenNLP version %s installed to %s", VERSION, _get_module_root())
+    logger.info("Cuda devices available: %s", torch.cuda.device_count())
```
schmmd, line 49: I like this ;-)

epwalsh:
@matt-gardner tests and fixtures are now at the root ✅

PR #4272, opened 2020-05-21
Enable multi-process training on CPU

dirkgr:
This is probably only useful during testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Additional CI checks to ensure docstrings are consistently formatted.
-- Ability to train on CPU with multiple processes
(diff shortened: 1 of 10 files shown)


dirkgr:
@epwalsh, can you review this? Especially the way the tests are configured?

epwalsh:
Yeup! I’ll take a look right after lunch

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Additional CI checks to ensure docstrings are consistently formatted.
+- Ability to train on CPU with multiple processes
```
epwalsh, line 20:
```suggestion
- Ability to train on CPU with multiple processes by setting `cuda_devices` to a list of negative integers in your training config. For example: `"distributed": {"cuda_devices": [-1, -1]}`. This is mainly to make it easier to test and debug distributed training code..
```

dirkgr:
> LGTM! Looks like there are some multi-gpu tests in tests/training/trainer_test.py that might make sense to mark as `@cpu_or_gpu` as well?

There is only one, and it's about raising an error, so I want to leave that one alone.

PR #4275, opened 2020-05-21
Document when parameters do not need an entry in a config file

matt-gardner:
The point of `FromParams` is to have a config file's keys exactly match constructor parameters.  There are a few cases where that isn't actually true, because we construct an object separately and pass it in, removing the need to specify that constructor argument in a config file.  In writing about how config files work for the allennlp course, I realized that this was a pretty big piece of magic that needed explanation and documentation.  This PR adds documentation for that behavior.

I additionally noticed that there was a `batch_weight_key` in `commands.train` that could be simplified quite a bit, so I simplified it.


matt-gardner:
In a few minutes I'll have the section of the course on this functionality finished; I'll send it to you to see what you think.

PR #4290, opened 2020-05-26
clean up dependencies

epwalsh:
- Removes `conllu` dependency (see https://github.com/allenai/allennlp-models/pull/61)
- Removes `semantic_version` dependency (we don't use this at all anymore)
- Removes the `tests_requires` and `setup_requries` section of `setup.py`, since we don't use the `python setup.py test` command. Also see the deprecation notice of `pytest-runner` for why we shouldn't be using this: https://github.com/pytest-dev/pytest-runner#deprecation-notice.
- Moves a few other requirements in the `install_requires` section of `setup.py` to `dev-requirements.txt`, since they are only imported in our `tests/`.
- Removes `aiohttp` from `dev-requirements.txt` since it isn't used anywhere.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -146,8 +146,6 @@ jobs:
       run: |
         git clone https://github.com/allenai/allennlp-models.git
         cd allennlp-models
-        # TODO: remove this line.
-        git checkout conllu
         pip install --upgrade --upgrade-strategy eager -e . -r dev-requirements.txt
 
     - name: Run models tests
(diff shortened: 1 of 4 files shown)


.github/workflows/pull_request.yml
```diff
@@ -146,6 +146,8 @@ jobs:
       run: |
         git clone https://github.com/allenai/allennlp-models.git
         cd allennlp-models
+        # TODO: remove this line.
+        git checkout conllu
```
epwalsh, line 150: This line can be removed after https://github.com/allenai/allennlp-models/pull/61 is merged.

dirkgr: I just merged that.

Issue #4281, opened 2020-05-25
"TypeError: not a sequence" on simple coreference resolution


arthurdeschamps:
**Describe the bug**

I get a `TypeError: not a sequence` when trying to predict this simple string: "Besides its prominence in sports, Notre Dame is also a large, four-year, highly residential research University, and is consistently ranked among the top twenty universities in the United States  and as a major global university."

Full stacktrace:

```
Traceback (most recent call last):
  File "/home/arthur/question-generation/models/sg_dqg.py", line 156, in <module>
    preprocess(args.ds)
  File "/home/arthur/question-generation/models/sg_dqg.py", line 124, in preprocess
    coreferences = coreference_resolution(evidences_list)
  File "/home/arthur/question-generation/models/sg_dqg.py", line 74, in coreference_resolution
    document="Besides its prominence in sports, Notre Dame is also a large, four-year, highly residential research University, and is consistently ranked among the top twenty universities in the United States  and as a major global university."
  File "/home/arthur/anaconda3/envs/tensorflow_env/lib/python3.6/site-packages/allennlp_models/coref/coref_predictor.py", line 65, in predict
    return self.predict_json({"document": document})
  File "/home/arthur/anaconda3/envs/tensorflow_env/lib/python3.6/site-packages/allennlp/predictors/predictor.py", line 48, in predict_json
    return self.predict_instance(instance)
  File "/home/arthur/anaconda3/envs/tensorflow_env/lib/python3.6/site-packages/allennlp/predictors/predictor.py", line 171, in predict_instance
    outputs = self._model.forward_on_instance(instance)
  File "/home/arthur/anaconda3/envs/tensorflow_env/lib/python3.6/site-packages/allennlp/models/model.py", line 142, in forward_on_instance
    return self.forward_on_instances([instance])[0]
  File "/home/arthur/anaconda3/envs/tensorflow_env/lib/python3.6/site-packages/allennlp/models/model.py", line 167, in forward_on_instances
    model_input = util.move_to_device(dataset.as_tensor_dict(), cuda_device)
  File "/home/arthur/anaconda3/envs/tensorflow_env/lib/python3.6/site-packages/allennlp/data/batch.py", line 139, in as_tensor_dict
    for field, tensors in instance.as_tensor_dict(lengths_to_use).items():
  File "/home/arthur/anaconda3/envs/tensorflow_env/lib/python3.6/site-packages/allennlp/data/instance.py", line 99, in as_tensor_dict
[...]

PR #4291, opened 2020-05-26
ensure intra_word_tokenize raises an exception when token cannot be split into wordpieces

epwalsh:
Closes https://github.com/allenai/allennlp/issues/4281.

I think when `intra_word_tokenize` cannot split a token into wordpieces we should raise an exception instead of just using `None` for the offsets. Especially considering there are several other downstream functions that depend on the offsets not being `None`. For example,
- https://github.com/allenai/allennlp/blob/master/allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py#L96
- https://github.com/allenai/allennlp-models/blob/master/allennlp_models/coref/util.py#L105


epwalsh:
Well, here is is an alternate solution then: https://github.com/allenai/allennlp/pull/4292

Issue #4270, opened 2020-05-21
deepcopy(instance) fails when using mismatched indexer.


dheerajrajagopal:
**Describe the bug**
deepcopy(instance) fails when using mismatched indexer. 

**Stack Trace** 
    <pre> 
    20         to the input and returns those gradients normalized and sanitized.
     21         """
---> 22         labeled_instances = self.predictor.json_to_labeled_instances(inputs)
     23
     24         # List of embedding inputs, used for multiplying gradient by the input for normalization
/usr1/home/dheeraj/anaconda3/envs/interpret-baseline/lib/python3.7/site-packages/allennlp/predictors/predictor.py in json_to_labeled_instances(self, inputs)
     62         instance = self._json_to_instance(inputs)
     63         outputs = self._model.forward_on_instance(instance)
---> 64         new_instances = self.predictions_to_labeled_instances(instance, outputs)
     65         return new_instances
     66
/usr1/home/dheeraj/anaconda3/envs/interpret-baseline/lib/python3.7/site-packages/allennlp/predictors/text_classifier.py in predictions_to_labeled_instances(self, instance, outputs)
     43         self, instance: Instance, outputs: Dict[str, numpy.ndarray]
     44     ) -> List[Instance]:
---> 45         new_instance = deepcopy(instance)
     46         label = numpy.argmax(outputs["probs"])
     47         new_instance.add_field("label", LabelField(int(label), skip_indexing=True))
/usr1/home/dheeraj/anaconda3/envs/interpret-baseline/lib/python3.7/copy.py in deepcopy(x, memo, _nil)
    178                     y = x
    179                 else:
--> 180                     y = _reconstruct(x, memo, *rv)
</pre>

**To Reproduce**
```
from allennlp.predictors.predictor import Predictor
from allennlp.data.tokenizers import SpacyTokenizer
import allennlp_models.sentiment
from allennlp.interpret.saliency_interpreters import SimpleGradient

predictor = Predictor.from_path("https://storage.googleapis.com/allennlp-public-models/sst-roberta-large-2020.05.05.tar.gz")
predictor.predict(
   sentence="a very well-made, funny and entertaining picture."
 )
tokenizer = SpacyTokenizer()
tokenized_sentence = tokenizer.tokenize("a very well-made, funny and entertaining picture.")
simple_grad_viz =  SimpleGradient(predictor).saliency_interpret_from_json({'sentence':"a very well-made, funny and entertaining picture."}) 
```

**Expected behavior**
[...]

PR #4294, opened 2020-05-27
Adds a "duplicate()" method on instances and fields

epwalsh:
Closes https://github.com/allenai/allennlp/issues/4270.

This new method should be used in place of `copy.deepcopy()`. It can be overridden on certain fields to avoid making deep copies of certain attributes when deep copying is unnecessary or not possible.

---

Note: I was lazy and haven't added tests yet, but I'll do so before merging if ya'll agree with this approach.


epwalsh:
> LGTM! Is there anything in the models repo that needs a similar change?

There is, I'll follow up with a PR for that.

PR #4298, opened 2020-05-28
ensure only one process initializes its model at a time to avoid race condition

epwalsh:
This should make distributed training more robust at least until we can improve our caching mechanism with lock files.


epwalsh:
Closing in favor of https://github.com/allenai/allennlp/pull/4299

PR #4299, opened 2020-05-28
add lock file within caching mechanism

epwalsh:
This should make our caching mechanism much more robust. It uses a lock file to ensure that if one process is caching a particular file, any other calls to `cached_path()` on the same file (from separate Python processes) will block until the writing process has finished.

I'm not sure if there's a good way to unit test this, or if it's even worth it because the logic is pretty simple. But you can give it a try by running a script like this from multiple terminals at the same time:

```python
import logging

logging.basicConfig(level=logging.INFO)

filename = "https://allennlp.s3.amazonaws.com/datasets/glove/glove.6B.100d.txt.gz"
print(cached_path(filename))
```

Also https://github.com/allenai/allennlp/pull/4298 may not be necessary if we merge this.

### DIFF
--- a/allennlp/data/vocabulary.py
+++ b/allennlp/data/vocabulary.py
@@ -10,6 +10,8 @@ import os
 from collections import defaultdict
 from typing import Any, Callable, Dict, Iterable, List, Optional, Set, Union, TYPE_CHECKING
 
+from filelock import FileLock
+
 from allennlp.common.util import namespace_match
 from allennlp.common import Registrable
 from allennlp.common.checks import ConfigurationError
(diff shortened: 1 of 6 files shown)


allennlp/data/vocabulary.py
```diff
@@ -604,21 +610,24 @@ def save_to_files(self, directory: str) -> None:
...
-            os.path.join(directory, NAMESPACE_PADDING_FILE), "w", "utf-8"
-        ) as namespace_file:
-            for namespace_str in self._non_padded_namespaces:
-                print(namespace_str, file=namespace_file)
-
-        for namespace, mapping in self._index_to_token.items():
-            # Each namespace gets written to its own file, in index order.
+        # We use a lock file to avoid race conditions where multiple processes
+        # might be reading/writing from/to the same vocab files at once.
+        with FileLock(os.path.join(directory, ".lock")):
```
matt-gardner, line 615: Does this mean you can remove the barrier in the training loop?

epwalsh: Yea I think we can

PR #4301, opened 2020-05-29
Deals with the case where a word doesn't have any word pieces assigned

dirkgr:
I have not tested this end-to-end yet, and I haven't written tests. But this should fix https://github.com/allenai/allennlp/issues/4281 without breaking other usage scenarios.


epwalsh:
FYI `None` offsets will still be an issue here: https://github.com/allenai/allennlp-models/blob/master/allennlp_models/coref/util.py#L105

dirkgr:
@ZhaofengWu, I don't think this is the same case. These are not character offsets. These are the offsets that `intra_word_tokenize()` produces to map tokens to word pieces.

ZhaofengWu:
Ah okay. Sorry about that

allennlp/data/token_indexers/pretrained_transformer_mismatched_indexer.py
```diff
@@ -63,6 +63,11 @@ def tokens_to_indices(self, tokens: List[Token], vocabulary: Vocabulary) -> Inde
         self._matched_indexer._add_encoding_to_vocabulary_if_needed(vocabulary)
 
         wordpieces, offsets = self._allennlp_tokenizer.intra_word_tokenize([t.text for t in tokens])
+
+        # For tokens that don't correspond to any word pieces, we put (-1, -1) into the offsets.
+        # That results in the embedding for the token to be all zeros.
+        offsets = [x if x is not None else (-1, -1) for x in offsets]
```
epwalsh, line 69: I don't love that we have to re-create this list. What if `intra_word_tokenize` could take a default value to use when a token doesn't result in any word pieces? By default this is `None`, but in this case we'd use `(-1, -1)`.

Or `intra_word_tokenize` could return a tuple of iterators instead of lists. This might add a little complexity though.

dirkgr: I looked at that option. I don't want to do it because there are actually many functions affected, not just one, and they would all have to gain an extra parameter. The function is complicated enough as it is.

But if we find that this causes performance problems, I'll change it immediately.

dirkgr: Actually, there is this: https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.unzip

With that we could return two iterators, and `more_itertools` takes care of the complexity of splitting an iterator of tuple into two iterators. I love it because it's clever, but I'm still not wild about the complexity for something that hasn't been proven to be slow...

dirkgr: From an API perspective it's a little weird to return offsets into an iterator though. You'd basically always have to convert it into a list before you can do anything with it.

dirkgr: Actually, nevermind. Those two lists aren't the same length anyways, so `more_itertools` doesn't solve the problem.

PR #4302, opened 2020-05-29
Stop calculating span-based F1 metric when `calculate_span_f1` is `False`.

himkt:
👋 from Japan.
First of all, Let me say thank you for developing this fantastic library, AllenNLP.
I found a small problem in `SimpleTagger` and created this PR.

## Motivation

[allennlp/models/simple_tagger.py:88-91](https://github.com/allenai/allennlp/blob/master/allennlp/models/simple_tagger.py#L88-L91)

```python
if calculate_span_f1 or label_encoding:  # remove `label_encoding`
  self._f1_metric = SpanBasedF1Measure(
    vocab, tag_namespace=label_namespace, label_encoding=label_encoding
  )
	else:
    self._f1_metric = None
```

Skip calculating `span_f1` when `calculate_span_f1` is `False` ignoring `label_encoding`.
This implementation is consistent with [crf_tagger.py](https://github.com/allenai/allennlp-models/blob/master/allennlp_models/tagging/models/crf_tagger.py#L138-L145).

## Why?

[allennlp/training/metrics/span_based_f1_measure.py:173-192](https://github.com/allenai/allennlp/blob/master/allennlp/training/metrics/span_based_f1_measure.py#L173-L191)

```python
tags_to_spans_function = None
# `label_encoding` is empty and `tags_to_spans_function` is provided.
if self._label_encoding is None and self._tags_to_spans_function:
    tags_to_spans_function = self._tags_to_spans_function
# Search by `label_encoding`.
elif self._label_encoding == "BIO":
    tags_to_spans_function = bio_tags_to_spans
elif self._label_encoding == "IOB1":
    tags_to_spans_function = iob1_tags_to_spans
elif self._label_encoding == "BIOUL":
    tags_to_spans_function = bioul_tags_to_spans
elif self._label_encoding == "BMES":
    tags_to_spans_function = bmes_tags_to_spans

predicted_spans = tags_to_spans_function(predicted_string_labels, self._ignore_classes)
gold_spans = tags_to_spans_function(gold_string_labels, self._ignore_classes)

predicted_spans = self._handle_continued_spans(predicted_spans)
gold_spans = self._handle_continued_spans(gold_spans)
```

`SimpleTagger` doesn't use `CRF`, which means its prediction isn't strongly constrained by label transitions. In other words, invalid tag sequences could be generated.

However, `SimpleTagger` calculates span-based F1 metrics if `label_encoding` is specified. This causes `InvalidTagSequence` error in training NER model.

## Reproduction

[bert-linear.jsonnet](https://github.com/himkt/allennlp-NER/blob/bert/config/bert-linear.jsonnet)

I think you can reproduce `InvalidTagSequence` by simply running `allennlp train bert-linear.jsonnet -s bert-linear`.

```python
  ...
  File ".../lib/python3.7/site-packages/allennlp/models/simple_tagger.py", line 173, in forward
    self._f1_metric(logits, tags, mask)
  File ".../lib/python3.7/site-packages/allennlp/training/metrics/span_based_f1_measure.py", line 187, in __call__
    predicted_spans = tags_to_spans_function(predicted_string_labels, self._ignore_classes)
  File ".../lib/python3.7/site-packages/allennlp/data/dataset_readers/dataset_utils/span_utils.py", line 268, in bioul_tags_to_spans
    raise InvalidTagSequence(tag_sequence)
allennlp.data.dataset_readers.dataset_utils.span_utils.InvalidTagSequence: I-ORG B-ORG B-ORG I-PER I-LOC B-ORG I-MISC B-ORG I-ORG
```

```python
-> predicted_spans = tags_to_spans_function(predicted_string_labels, self._ignore_classes)
(Pdb) predicted_string_labels
['I-ORG', 'B-ORG', 'B-ORG', 'I-PER', 'I-LOC', 'B-ORG', 'I-MISC', 'B-ORG', 'I-ORG']
(Pdb) gold_string_labels
['U-ORG', 'O', 'U-MISC', 'O', 'O', 'O', 'U-MISC', 'O', 'O']
```


himkt:
I think test fails since I don't update fixture.

himkt:
@epwalsh 
Thank you for guiding me, I updated JSON files and `CHANGELOG`.

himkt:
And I also have a proposal about `SpanBasedF1Measure`.
How about stop raising `InvalidTagSequence` for model predictions?
(This might be out of the score in this PR but I'm happy if I can discuss it here.)

```
diff --git a/allennlp/training/metrics/span_based_f1_measure.py b/allennlp/training/metrics/span_based_f1_measure.py
index d16a9a28..eb145db6 100644
--- a/allennlp/training/metrics/span_based_f1_measure.py
+++ b/allennlp/training/metrics/span_based_f1_measure.py
@@ -184,11 +184,15 @@ class SpanBasedF1Measure(Metric):
             elif self._label_encoding == "BMES":
                 tags_to_spans_function = bmes_tags_to_spans
 
-            predicted_spans = tags_to_spans_function(predicted_string_labels, self._ignore_classes)
             gold_spans = tags_to_spans_function(gold_string_labels, self._ignore_classes)
+            gold_spans = self._handle_continued_spans(gold_spans)
+
+            try:
+                predicted_spans = tags_to_spans_function(predicted_string_labels, self._ignore_classes)
+            except Exception:
+                predicted_spans = []
 
             predicted_spans = self._handle_continued_spans(predicted_spans)
-            gold_spans = self._handle_continued_spans(gold_spans)
 
             for span in predicted_spans:
                 if span in gold_spans:
```

epwalsh:
Hey @himkt, after looking at the `SimpleTagger` class a little more closely I realized we should probably make a slightly different change so that the `calculate_span_f1` logic is consistent with `CrfTagger`.

In the `CrfTagger`, the span f1 metric is used whenever `label_encoding` is specified and `calculate_span_f1` is either `True` or `None`. See
- https://github.com/allenai/allennlp-models/blob/master/allennlp_models/tagging/models/crf_tagger.py#L115, and
- https://github.com/allenai/allennlp-models/blob/master/allennlp_models/tagging/models/crf_tagger.py#L138

However, currently the `SimpleTagger` will use the span f1 metric whenever `label_encoding` is specified, regardless of the value of `calculate_span_f1` (even is `calculate_span_f1` is `False`). Though with your changes, `calculate_span_f1` would be required to be set to `True`.

But neither the current logic on master or the logic in your proposed change is consistent with the `CrfTagger`, which I believe has the most correct logic.

So would you mind modifying your PR so that `SimpleTagger` is consistent with `CrfTagger`? And that way using the span f1 metric could be avoided by just setting `calculate_span_f1` to `False`.

epwalsh:
> And I also have a proposal about `SpanBasedF1Measure`.
> How about stop raising `InvalidTagSequence` for model predictions?

I think we want to keep it how it is since the span based f1 metric is undefined when the tag sequence is invalid. The `InvalidTagSeqence` exception should be handled by the model, anyway. Of course the `SimpleTagger` doesn't handle this exception, but this model is really only meant to be used a an example model. The `CrfTagger` on the other hand should be a drop-in replacement for the `SimpleTagger`, and is much more suitable for application.

himkt:
> the span based f1 metric is undefined when the tag sequence is invalid. 

Thank you for the clarifying, it makes sense.

> Of course the SimpleTagger doesn't handle this exception, but this model is really only meant to be used a an example model. The CrfTagger on the other hand should be a drop-in replacement for the SimpleTagger, and is much more suitable for application.

Now I agree with you. I take down my proposal. Thank you for explanation.

himkt:
I revised my PR to make `SimpleTagger` consistent with `CrfTagger` in e219ebf.

allennlp/models/simple_tagger.py
```diff
@@ -82,24 +82,26 @@ def __init__(
...
+        self.calculate_span_f1 = calculate_span_f1
+        if calculate_span_f1:
+            if not label_encoding:
+                raise ConfigurationError(
+                    "calculate_span_f1 is True, but no label_encoding was specified."
+                )
             self._f1_metric = SpanBasedF1Measure(
                 vocab, tag_namespace=label_namespace, label_encoding=label_encoding
             )
-        else:
```
epwalsh, line 101: I think we should keep this "else" clause where `self._f1_metric` is set to `None` so that it's always defined.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Similar to our caching mechanism, we introduced a lock file to the vocab to avoid race
   conditions when saving/loading the vocab from/to the same serialization directory in different processes.
+- Stop calculating span-based F1 metric when `calculate_span_f1` is `False`.
```
epwalsh, line 35:
```suggestion
- SimpleTagger will no longer calculate span-based F1 metric when `calculate_span_f1` is `False`.
```

allennlp/models/simple_tagger.py
himkt: Oh sorry, I removed `else: self._f1_...` based on `CrfTagger`.

https://github.com/allenai/allennlp-models/blob/master/allennlp_models/tagging/models/crf_tagger.py#L138-L145

I reverted the change in 51012f8.

himkt:
Thank you for the review, please take a look again. :pray:

epwalsh:
Hey @himkt, don't worry too much about the codecov/patch check, it's not required.

test_fixtures/simple_tagger_without_span_f1/experiment.json
```diff
@@ -0,0 +1,65 @@
+{
```
epwalsh, line 1: I don't think we need to add another fixture. This case should already be covered by https://github.com/allenai/allennlp/blob/master/test_fixtures/simple_tagger/experiment.json, right?

himkt: It was intended to check the case `calculate_span_f1` is `false`. (In `experiment.json`, `calculate_span_f1` would be `None`, I think.)

But I also think it is too much, then removed in a6be2ae.

epwalsh: Gotcha! Yea, I don't think that really needs a test. Anyway, thank you! Looks good!

PR #4306, opened 2020-06-01
move tensorboard memory logging to BatchCallback

epwalsh:


epwalsh:
Confirmed performance is back up: https://beaker.org/ex/ex_zsue7v2w1zzm/tasks/tk_oz12b26ef3vp

Issue #4268, opened 2020-05-20
Will using slot classes for fields and other data abstractions significantly reduce memory usages?


epwalsh:
Some helpful info:
- https://www.python-course.eu/python3_slots.php

matt-gardner:
From that page:

> Since Python 3.3 this advantage is not as impressive any more. With Python 3.3 Key-Sharing Dictionaries are used for the storage of objects.

Seems worth trying, though, at least.  Should be easy to try.

PR #4312, opened 2020-06-02
Making Token class a "slots" class

epwalsh:
Partially address #4268.

This defines the `__slots__` attribute of the `Token` class, which is an optimization that dramatically reduces the size in memory of a `Token`  instance.

You can use the [`pympler`](https://pympler.readthedocs.io/en/latest/) module to inspect the size of a `Token` instance:

```python
from pympler import asizeof
from allennlp.data.tokenizers import Token
asizeof.asizeof(Token("hello"))
```

Before this change to the `Token` class, the size reported by `pympler` was 864 bytes. With this change, the size is 200 bytes.

This also has might have a big impact on the speed of our tokenizers. To help understand that impact, I added a benchmarking framework ([`pytest-benchmark`](https://github.com/ionelmc/pytest-benchmark) - actually a pytest plugin, but it's the best framework I could fine so far) and a simple benchmark of our character tokenizer.

You can run this benchmark with `make benchmarks`.

Before this change to the `Token` class, the results on my laptop showed a medium time of about 280us. With this change, the medium time was 200us.


Makefile
```diff
@@ -42,15 +42,15 @@ check-for-cuda :
 
 .PHONY : lint
 lint :
-	flake8 ./scripts ./tests $(SRC)
```
matt-gardner, line 45: Is this intentional? Don't we want these checks on tests?  Oh, I think I get it - this change just makes it find all python files under the current directory?  I didn't realize it would work with a mixed (with non-python) directory like this.

allennlp/data/tokenizers/token.py
```diff
@@ -40,16 +40,57 @@ class Token:
...
+        "pos_",
+        "tag_",
+        "dep_",
+        "ent_type_",
+        "text_id",
+        "type_id",
+    ]
+    # Defining the `__slots__` of this class is an optimization that dramatically reduces
+    # the size in memory of a `Token` instance. The downside of using `__slots__`
+    # with a dataclass is that you can assign default values at the class level,
```
matt-gardner, line 57:
```suggestion
    # with a dataclass is that you can't assign default values at the class level,
```
is this what you meant?

allennlp/data/tokenizers/token.py
```diff
@@ -40,16 +40,57 @@ class Token:
...
-    idx: Optional[int] = None
-    idx_end: Optional[int] = None
-    lemma_: Optional[str] = None
-    pos_: Optional[str] = None
-    tag_: Optional[str] = None
-    dep_: Optional[str] = None
-    ent_type_: Optional[str] = None
-    text_id: Optional[int] = None
-    type_id: Optional[int] = None
+    __slots__ = [
```
matt-gardner, line 43: Do you know if this impacts extensions to this class?  E.g., it would be sad to break this functionality: https://github.com/allenai/allennlp/pull/3769.

I just tried to answer this myself, and it looks like we won't break that functionality, which is good: https://stackoverflow.com/questions/1816483/how-does-inheritance-of-slots-in-subclasses-actually-work.

Makefile
epwalsh: Yeup, this should find every python file in the repo

allennlp/data/tokenizers/token.py
epwalsh: It might be good to have a test for this

epwalsh:
I ran the bidaf training experiment on this branch and so far it's reduce the CPU memory by about 400 MB. https://beaker.org/ex/ex_gptpbv7du0f7/tasks/tk_dnfwgkpe9ebf

epwalsh:
> Just throwing this out there: How small does it get if we have `Token` with only a small number of fields, and `SpacyToken`, which derives from `Token` and adds all those extra fields like `pos_` and `lemma_`, etc.?

If we just had "text", "idx", "idx_end", "text_id", and "type_id" as fields, the size of `Token("hello")` would be 160 bytes.

epwalsh:
> LGTM. I assume `dataclasses.replace()` still works? I think one of the tests would fail otherwise.

Yeup!

Makefile
bryant1410: Don't you wanna check only the changed files?

Issue #4039, opened 2020-04-08
Fields have large memory footprints


nelson-liu:
I've noticed that `Fields` can have quite a bit of memory overhead—it's very common that i'll have a dataset that I can load into memory (e.g., with `json.load`), but turning it into fields / instances blows up the memory usage.

For instance, I've been doing some stuff with span representations recently, so using a `ListField[SpanField]` in my `Instances`. Each of these `ListFields` is around ~3000 elements. If I run the following:

```
In [1]: from allennlp.data import Token
   ...: from allennlp.data.fields import TextField, SpanField
   ...: from allennlp.data.token_indexers import SingleIdTokenIndexer

In [2]: indexers = {"words": SingleIdTokenIndexer("words")}

In [3]: text = TextField(
   ...:             [Token(t) for t in ["here", "is", "a", "sentence", "for", "spans", "."]], indexers
   ...:         )

In [4]: l = []

In [5]: for i in range(10000*3500):
   ...:     l.append(SpanField(2, 3, text))
```

and check the memory usage, it's around 6.5 gigabytes for 10,000 simulated instances (when I have much more, ranging between 100,000 on some datasets to 800,000 on other, larger datasets). Maybe this size is reasonable, though, and I'm just being dumb here. Let me know if that's the case...

Where does all of this overhead come from? Is there anything we can do to reduce it?

nelson-liu:
Fwiw, replacing `ListField[SpanField]` with `ArrayField` (with padding value -1 and dtype np.int64) is _much_ faster and also has a _much_ smaller memory footprint.

matt-gardner:
This is almost certainly overhead in python object / list creation.  Maybe the solution is to just give a caveat in `ListField` that if you have very long lists, you might consider using `ArrayField` directly instead.  I don't think this is a huge problem for our other uses of this (e.g., the constituency parser, the coref model, etc.).

schmmd:
To close this ticket out we want to add a disclaimer to the docstring on ListField, as @matt-gardner suggests above.

dirkgr:
@Tobias-Rohde also ran into this problem when training BART.

dirkgr:
According to @nelson-liu's numbers above, instances take 650kb each. That is hard to explain with just `ListField[SpanField]` vs. `ArrayField`. I think this is a real problem and we need to follow up, not just update the documentation.

nelson-liu:
[...]

PR #4313, opened 2020-06-02
make Instance, Batch, and all field classes "slots" classes

epwalsh:
Addresses #4268.


allennlp/data/fields/field.py
```diff
@@ -116,7 +118,10 @@ def batch_tensors(self, tensor_list: List[DataArray]) -> DataArray:  # type: ign
 
     def __eq__(self, other) -> bool:
         if isinstance(self, other.__class__):
-            return self.__dict__ == other.__dict__
+            for attr in self.__slots__:
```
matt-gardner, line 121: What happens if we have a `Field` subclass the adds a new slot?  This would break, right?  I don't think we currently have this case, but it's something to think about.  I guess that class would have to also override `__eq__`?

epwalsh: This works fine for subclasses without overriding `__eq__`. And I believe we have tests that cover this (see the `IndexField` tests, for example).

epwalsh:
> Code changes seem fine to me. Any idea yet on how this affects performance?

The memory improvements are similar to what we saw with the `Token` class. For example, the size of an empty `TextField` goes from 704 bytes down to 544 bytes. An empty `Instance` goes from 568 bytes down to 336 bytes.

allennlp/data/fields/field.py
matt-gardner: I was imagining something like this:

```python
class Field:
    __slots__ = []

class IndexField(Field):
    __slots__ = ["index"]

class SpecialIndexField(IndexField):
    __slots__ = ["new_member_variable"]
```

From what I saw when I was reading about inheritance in slot classes, I'm pretty sure this would break without also overriding `__eq__`.  We don't have any such case right now, I was just wondering if this logic was actually safe in general.

epwalsh: That should still work, but I'll add a test to make sure.

matt-gardner: Oh, you're saying that `self.__slots__` gets updated in the right way?  I'd be surprised, but I guess all of the slots stuff is magic to me, so maybe there's more magic I didn't know about.

epwalsh: Ah, you're absolutely right. That does break it. I'm working on a fix.

matt-gardner: I mean, we have other cases where we have to override `__eq__`.  It's not the end of the world if it's impossible to make general logic here.  We should at least leave a note about the potential issue somewhere.

epwalsh: Ok just added a test like that with a fix.

matt-gardner: When I run this locally, I get to `object` in `__bases__`, and get an attribute error because `object` doesn't have `__slots__`.

epwalsh: Oh, hmm, are you using Python 2.*?

matt-gardner: 3.7.  Not sure what the deal is.

epwalsh: I'm not sure either, but try again now. The last commit should hopefully fix that for you

allennlp/data/fields/field.py
```diff
@@ -116,7 +118,21 @@ def batch_tensors(self, tensor_list: List[DataArray]) -> DataArray:  # type: ign
...
         if isinstance(self, other.__class__):
-            return self.__dict__ == other.__dict__
+            # With the way "slots" classes work, self.__slots__ only gives the slots defined
+            # by the current class, but not any of its base classes. Therefore to truly
+            # check for equality we have to check through all of the slots in all of the
+            # base classes as well.
+            for class_ in (self.__class__,) + self.__class__.__bases__:
+                slots = getattr(class_, "__slots__")
+                if slots:
+                    for attr in slots:
```
matt-gardner, line 128:
```suggestion
                for attr in getattr(class_, "__slots__", []):
```

epwalsh: Thanks, this is definitely better

allennlp/data/fields/field.py
matt-gardner: I was just playing around in an interpreter because I was curious about what happened if there was a class in the chain that didn't have slots.  I was wondering how your tests were passing given what I saw.  But the logic that's there now would definitely work.

matt-gardner: I'm still not sure how your tests passed given what I saw in the interpreter, but 🤷‍♂️

dirkgr:
Is this really worth it? HotpotQA has roughly 100k examples, with maybe 12 text fields each. That makes it a "large" dataset, though not so large you hav to use a lazy reader.

In that case, this change saves less than 200 MB.

It's worth it for the tokens because there are millions of them, but we don't have that many fields and instances.

epwalsh:
> Is this really worth it? HotpotQA has roughly 100k examples, with maybe 12 text fields each. That makes it a "large" dataset, though not so large you hav to use a lazy reader.
> 
> In that case, this change saves less than 200 MB.
> 
> It's worth it for the tokens because there are millions of them, but we don't have that many fields and instances.

IMO 200MB is huge. And what's the downside other than a little more complexity in `Field.__eq__`?

Also, custom fields don't need to be defined as slots classes, so nothing changes from a user's perspective.

matt-gardner:
If we have a simple change that gives us a 20% memory reduction in representing data across the board, I'm definitely in favor of it.

dirkgr:
I don't know exactly how big HotpotQA would be in our data format, but
making some assumptions (256 bytes per token, 256 tokens per article, 10
articles per question, 100k questions, all other data requirements
ignored), this comes out to about 61GB. 200 MB is 0.3%.

epwalsh:
Well, to provide some exact numbers, I loaded the squad v1.1 train dataset into memory (with the Squad dataset reader, i.e. the one BiDAF uses) and saw a 60 MB difference, which came out to 2%.

epwalsh:
It also took a few seconds less time to load (difference of 1-4% in loading time).

epwalsh:
You can see the results with BiDAF here:
- https://beaker.org/ex/ex_pkvgg5g6hhcc/tasks/tk_rjxbkdpzhu77 (using this branch)
- https://beaker.org/ex/ex_gptpbv7du0f7/tasks/tk_dnfwgkpe9ebf (current master branch, includes the Token slots classes)

epwalsh:
Some more results, this time for TransformerQA:
- https://beaker.org/ex/ex_185prfnmjkza/tasks/tk_g3on5j0wrddn (using this branch)
- https://beaker.org/ex/ex_az40he2vfjbk/tasks/tk_4qie7yu3mkzq (current master branch, includes the Token slots classes)

The memory usage is lower with this branch by about 48 MB (a 0.05% improvement), and the dataset loading time decreased by 75 seconds (a 7% improvement).

matt-gardner:
Ok, yeah, my 20% number was pretty off the mark :sweat_smile:

epwalsh:
I'd say those numbers are still significant though. Enough to justify the minimal amount of additional complexity.

bratao:
Just for one more data-point. I work with documents with billions of tokens. I already heavily rely on disk-cache for handling it, but this improvement would be very welcome to me!

epwalsh:
Ok, I'm moving forward with this. If we find a strong reason in the future to not use slots classes we can easily revert.

nelson-liu:
Any memory savings are useful for me, as well :) Also, not sure how much memory usage you'd save under this workload: https://github.com/allenai/allennlp/issues/4039

epwalsh:
@nelson-liu changing `Token` to a slots class made a huge impact. On SQuAD v1.1 with the `TransformerQA` model, the memory used during training decreased from ~13.2 GB to ~10 GB after these changes: https://docs.google.com/spreadsheets/d/1iIRyRFGhKr725a-WpDPSNjxP6gtN9rXoMGh0o1BQ9Ms/edit#gid=0

That said I think there's still a lot of room for improvement. Like how an `Instance` is just a wrapper around a `dict`.

nelson-liu:
Wow, that's a huge improvement—i appreciate you digging into this!

Those experimental results are really interesting, it's great to see that there's a substantial single-GPU speedup (at least for BiDAF) simply by moving off of `0.9.0`. Thanks for keeping such detailed notes, and the performance improvements look great!

PR #4323, opened 2020-06-04
report CPU memory usage for each worker

### LINKED ISSUE #4318: Peak CPU memory not reported correctly in distributed training
I noticed this bug while benchmarking some distributed experiments on beaker. I think `allennlp.common.util.peak_memory_mb()` is only called from the master process, and it only reports the memory usage for the master process. This should really report the combined memory used across all workers.

epwalsh:
Closes #4318

### DIFF
--- a/allennlp/common/util.py
+++ b/allennlp/common/util.py
@@ -353,12 +353,11 @@ def import_module_and_submodules(package_name: str) -> None:
 
 def peak_memory_mb() -> Dict[int, float]:
     """
-    Get peak memory usage for this process, as measured by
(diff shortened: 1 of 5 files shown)


allennlp/common/util.py
```diff
@@ -361,20 +361,37 @@ def peak_memory_mb() -> float:
     Only works on OSX and Linux, returns 0.0 otherwise.
```
matt-gardner, line 361: Might be nice to update this to better describe the return value.  It's not obvious why the return value is a dictionary.

PR #4340, opened 2020-06-08
Consistently use underscores in Predictor names

dirkgr:
This seems like a nit-picky thing, but we had situations where the model was using underscores, and the corresponding predictor was using the exact same name except with dashes.


dirkgr:
The others are all in allennlp-models.

PR #4342, opened 2020-06-08
Update vocabulary load to a system-agnostic newline

bratao:
Hello, 
I had a problem about training a model on a Linux machine and loading on a Windows machine.  The error was:
AssertionError: OOV token not found!

After some debugging I found out that during the vocabulary loading, it was splitting by '\n', where this can cause a difference between Linux and Windows. This PR change the split to OS agnostic method of new-line splitting.

### DIFF
--- a/allennlp/data/vocabulary.py
+++ b/allennlp/data/vocabulary.py
@@ -443,7 +443,7 @@ class Vocabulary(Registrable):
             self._token_to_index[namespace] = {}
             self._index_to_token[namespace] = {}
         with codecs.open(filename, "r", "utf-8") as input_file:
-            lines = input_file.read().split("\n")
+            lines = input_file.read().splitlines()
             # Be flexible about having final newline or not
             if lines and lines[-1] == "":
                 lines = lines[:-1]


allennlp/data/vocabulary.py
```diff
@@ -443,7 +444,7 @@ def set_from_file(
             self._token_to_index[namespace] = {}
             self._index_to_token[namespace] = {}
         with codecs.open(filename, "r", "utf-8") as input_file:
-            lines = input_file.read().split("\n")
+            lines = re.split(r"\n|\r\n?", input_file.read())
```
epwalsh, line 447: What about just

```suggestion
            lines = input_file.read().splitlines()
```

See https://docs.python.org/3/library/stdtypes.html#str.splitlines

bratao: That was the first idea, but it broke the tests. Because splitlines also split on some table separators, and broke this example:
`vocab_file.write("tricky\x0bchar\n")`

epwalsh: Ah, good point.

epwalsh, line 447: This should be more efficient if we pre-compile the regex. So somewhere at the top of this module you could have 

```python
_NEW_LINE_REGEX = re.compile(r"\n|\r\n")
```

And then change this line to

```suggestion
            lines = _NEW_LINE_REGEX.split(input_file.read())
```

bratao: Done!

Issue #4330, opened 2020-06-05
Transformer tokenizers cause deadlocks when dataset reader is lazy and dataloader num_workers > 0


epwalsh:
## Checklist

- [x] I have verified that the issue exists against the `master` branch of AllenNLP.
- [x] I have read the relevant section in the [contribution guide](https://github.com/allenai/allennlp/blob/master/CONTRIBUTING.md#bug-fixes-and-new-features) on reporting bugs.
- [x] I have checked the [issues list](https://github.com/allenai/allennlp/issues) for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/allenai/allennlp/pulls) for existing proposed fixes.
- [x] I have checked the [CHANGELOG](https://github.com/allenai/allennlp/blob/master/CHANGELOG.md) and the [commit log](https://github.com/allenai/allennlp/commits/master) to find out if the bug was already fixed in the master branch.
- [x] I have included in the "Description" section below a traceback from any exceptions related to this bug.
- [x] I have included in the "Related issues or possible duplicates" section beloew all related issues and possible duplicate issues (If there are none, check this box anyway).
- [x] I have included in the "Environment" section below the name of the operating system and Python version that I was using when I discovered this bug.
- [x] I have included in the "Environment" section below the output of `pip freeze`.
- [x] I have included in the "Steps to reproduce" section below a minimally reproducible example.

## Description

Dataset readers that use a `PretrainedTransformerTokenizer` can cause a deadlock when used lazily and with `num_workers > 0` in the `Dataloader`.

<details>
<summary><b>Python traceback:</b></summary>
<p>

```
...
2020-06-05 15:04:14,980 - INFO - allennlp.common.params - trainer.learning_rate_scheduler.last_epoch = -1
2020-06-05 15:04:14,980 - INFO - allennlp.common.params - trainer.learning_rate_scheduler.gradual_unfreezing = False
2020-06-05 15:04:14,980 - INFO - allennlp.common.params - trainer.learning_rate_scheduler.discriminative_fine_tuning = False
2020-06-05 15:04:14,980 - INFO - allennlp.common.params - trainer.learning_rate_scheduler.decay_factor = 0.38
2020-06-05 15:04:14,982 - WARNING - allennlp.training.trainer - You provided a validation dataset but patience was set to None, meaning that early stopping is disabled
2020-06-05 15:04:14,982 - INFO - allennlp.training.trainer - Beginning training.
[...]

PR #4343, opened 2020-06-08
issue warning when multi-process data loading used with lazy dataset

epwalsh:
There are several issues with using [PyTorch's native multi-process data loading](https://pytorch.org/docs/stable/data.html#multi-process-data-loading), which is enabled by setting the `DataLoader`'s `num_workers` to a positive number.

1. The main issue with PyTorch's multi-process data loading is that it's fundamentally incompatible with our `AllennlpLazyDataset`, unless using a `ShardedDatasetReader`. This is because our dataset readers don't take into account that there are other workers since they never check [`torch.data.get_worker_info()`](https://pytorch.org/docs/stable/data.html#torch.utils.data.get_worker_info). Hence each worker will read all of the instances, and so you'll get `num_workers` multiples of each instance.

    This is easy to see by running this example:

    ```python
    from allennlp.data import DataLoader, Vocabulary
    from allennlp_models.pair_classification.dataset_readers import SnliReader
    
    reader = SnliReader(lazy=True, max_instances=10)
    
    print("initialized dataset")
    ds = reader.read("https://allennlp.s3.amazonaws.com/datasets/snli/snli_1.0_dev.jsonl")
    
    print("creating vocab")
    vocab = Vocabulary.from_instances(ds)
    
    ds.index_with(vocab)
    
    print("reading instances")
    i = 0
    for instance in DataLoader(ds, num_workers=2):
        print(instance["hypothesis"])
        i += 1
    print(f"read {i} instances")
    ```

    The only way to fix this would be to handle it at the `DatasetReader` level, which might require significant changes.

2. Another issue is that using some `PretrainedTransformerTokenizer`s or, more generally, any tokenizer from the [tokenizers](https://github.com/huggingface/tokenizers) Python package will result in a deadlock if the tokenizer is used from the master process and then used again from a forked thread. So any time one of these tokenizers is used within a lazy dataset reader in a multi-process data loader, you'll get a deadlock. See https://github.com/allenai/allennlp/issues/4330.

2. ~~Lastly, even though multi-process data loading works correctly with non-lazy dataset readers, the only thing that the workers will be doing in parallel is calling `instance.index_fields(self.vocab)` (which comes from [`AllennlpDataset.__getitem__(self, idx)`](https://github.com/allenai/allennlp/blob/master/allennlp/data/dataset_readers/dataset_reader.py#L24)).~~

    ~~This is probably not worth the overhead.~~

~~So this PR just removes the `num_workers` parameter to effectively barricade off the multi-process data loading feature.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Previously it would always return 1.
 - Removed old tutorials, in favor of [the new AllenNLP Guide](https://guide.allennlp.org)
 
+### Removed
+
+- Removed the `num_workers` parameter from the `allennlp.data.DataLoader` since PyTorch's multi-process data loading doesn't play well with lazy dataset readers, and provides very little, if any, speed up with non-lazy readers.
+
 ## [v1.0.0rc5](https://github.com/allenai/allennlp/releases/tag/v1.0.0rc5) - 2020-05-26
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,10 +49,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Previously it would always return 1.
 - Removed old tutorials, in favor of [the new AllenNLP Guide](https://guide.allennlp.org)
 
-### Removed
-
-- Removed the `num_workers` parameter from the `allennlp.data.DataLoader` since PyTorch's multi-process data loading doesn't play well with lazy dataset readers, and provides very little, if any, speed up with non-lazy readers.
-
 ## [v1.0.0rc5](https://github.com/allenai/allennlp/releases/tag/v1.0.0rc5) - 2020-05-26
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


epwalsh:
And FWIW I've been thinking about ways to incorporate multi-processing into a data loading pipeline. It probably needs to happen within the `DatasetReader` class.

matt-gardner:
Are there cases where you would get significant speedup in a non-lazy setting by using `num_workers`?

epwalsh:
>     * This breaks the API compatibility with pytorch
> 
>     * Indexing some fields is not fast, e.g character representations for tokens.
> 
>     * Your last point I think is not quite correct, because the pytorch `DataLoader` also uses a separate process to batch data, as well as reading it.

Those are all fair points. Then how about we keep `num_workers` but raise a `ConfigurationError` when it's not compatible with the dataset?

DeNeutoy:
Sure - I haven't thought a huge amount about what the best way to do lazy reading is. I think it would involve having a small in memory dataset of indices which point to shards, which then read data. Also maybe looking into the huggingface ArrowDataset might be an interesting idea.

epwalsh:
Hmm actually, this should probably just issue a warning since you could fix this issue within the `_read` method of a dataset reader.

epwalsh:
Okay, ready for another look when you get a chance @matt-gardner and @DeNeutoy

epwalsh:
Hmm I don't think the sharded reader does deal with this issue. It never calls `torch.utils.data.get_worker_info()`, and so it wouldn't know if it's being used within a multi-process loader.

epwalsh:
Closing this in favor of https://github.com/allenai/allennlp/pull/4344.

PR #4344, opened 2020-06-09
simplify dataset classes, fix multi-process lazy loading

epwalsh:
- Substantially simplifies the logic in the `DatasetReader` and the data set classes. There is no longer a need for multiple lazy dataset classes. Now we only need `AllennlpDataset` and `AllennlpLazyDataset`, both of which have very minimal logic.
- Fixes multi-process data loading with lazy dataset readers (except for the deadlock issue caused by `tokenizers`, but we now issue a warning about that).
- Removes the potential for race-conditions on dataset cache files by using a file locks, and ensures that the caching system will work properly with distributed training and multi-process data loading.
- Adds a parameter `manual_multi_process_sharding` to the `DatasetReader` class (this is similar to the `manual_distributed_sharding` parameter).


epwalsh:
Ideally `_read()` should generate the rawest serialized form of the data so that the it does as little pre-processing and deserialization as possible.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -335,5 +269,54 @@ def deserialize_instance(self, string: str) -> Instance:
...
+            # So dividing self.max_instances by step_size is a good starting point,
+            # but we still might have a remainder.
+            per_process_max_instances = self.max_instances // step_size
+            remainder = self.max_instances % step_size
+            # Divide the remainder among the first `remainder` processes.
+            if remainder < start_index:
+                per_process_max_instances += 1
+
+        if transform:
+            return map(
```
DeNeutoy, line 311: IMO maps are not super readable/unusual in python, something maybe more pythonic:

```
slices = itertools.islice(iterable, start_index, per_process_max_instances, step_size)
return (self.ensure_instance(transform(x)) for x in slices)
```

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -36,101 +38,23 @@ def index_with(self, vocab: Vocabulary):
...
-        super().__init__()
-        self.deserialize = deserialize
-        self.serialize = serialize
-        self.vocab = vocab
-
-    def index_with(self, vocab: Vocabulary):
-        self.vocab = vocab
-
-
-class _LazyInstances(AllennlpLazyDataset):
```
DeNeutoy, line 52: 👏

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -335,5 +269,54 @@ def deserialize_instance(self, string: str) -> Instance:
...
         """
-
         return jsonpickle.loads(string)  # type: ignore
+
+    def ensure_instance(self, raw_instance) -> Instance:
+        if isinstance(raw_instance, Instance):
+            return raw_instance
+        return self.text_to_instance(**raw_instance)
+
+    def _auto_islice(
```
DeNeutoy, line 279: The name of this function could be clearer that it is uniquely applicable to distributed/multi-process data loading - maybe `multi_worker_islice`?

Also passing the transform is nice, but it could have a full docstring with an example, because it wasn't immediately clear what kind of transform I might pass here.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -335,5 +269,54 @@ def deserialize_instance(self, string: str) -> Instance:
         The default implementation is to use `jsonpickle`.  If you would like some other format
         for your pre-processed data, override this method.
         """
-
         return jsonpickle.loads(string)  # type: ignore
+
+    def ensure_instance(self, raw_instance) -> Instance:
```
DeNeutoy, line 274: Could this be a `Union[Dict[str, Any], Instance]`?

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -36,101 +38,23 @@ def index_with(self, vocab: Vocabulary):
...
-class _LazyInstances(AllennlpLazyDataset):
-    """
-    An `Iterable` that just wraps a thunk for generating instances and calls it for
-    each call to `__iter__`.
-    """
-
-    def __init__(
-        self,
-        instance_generator: Callable[[str], Iterable[Instance]],
+        instance_iterator_factory: Callable[[str], Iterable[Instance]],
```
DeNeutoy, line 41: This could use a docstring now it is more complicated, highlighting that you use this like:

`AllennlpLazyDataset(DatasetReader(lazy=True).read)`

epwalsh:
@DeNeutoy thanks for thorough review! I think I addressed all your comments, but I did make one more change for efficiency sake. Namely, I added a generic type variable to `DatasetReader` that represents the data type generated by `_read`:

https://github.com/allenai/allennlp/pull/4344/files#diff-e8efb5415e0922fb361952dd92d3d9e4R82

I also felt like `parse_raw_data` would be a better name for `ensure_instance` given this change.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -271,17 +295,50 @@ def deserialize_instance(self, string: str) -> Instance:
         """
         return jsonpickle.loads(string)  # type: ignore
 
-    def ensure_instance(self, raw_instance) -> Instance:
-        if isinstance(raw_instance, Instance):
-            return raw_instance
-        return self.text_to_instance(**raw_instance)
+    def parse_raw_data(self, raw_data: RawData) -> Instance:
```
DeNeutoy, line 305: Was this from someone else's feedback? What's the benefit of having additional raw data types here? I feel like `Dict[str, Any] || Instance` is ok.

epwalsh: The idea being that `_read` should do as little pre-processing as possible. See https://github.com/allenai/allennlp/blob/multi-process-loading/allennlp/data/dataset_readers/sequence_tagging.py#L71 for example.

DeNeutoy: Hmm, I think this is too extreme. You can still do this with a single Dict type in a _new_ dataset reader by doing `{"data": blob}` and having your `text_to_instance` method accept this. It isn't worth the API complexity of handling the different types and the errors that users will encounter from generically typed dataset readers.

epwalsh: Hm alright, well I think it's worth supporting a `Tuple[Any, ...]` as well, which would be treated as positional arguments to `text_to_instance`. It's less verbose to generate tuples as opposed to dictionaries, and probably a little more efficient.

DeNeutoy: I'd recommend you remove that for simplicities sake, as well as removing the ability for users to shoot themselves in the foot in the case that they re-order the arguments to `text_to_instance`. It's very likely that some args are compatible with each other, and this would be a really frustrating bug to run into, just because the argument order needs to be defined in two places in the `DatasetReader`.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -213,92 +163,103 @@ def read(self, file_path: str) -> Dataset:
...
-                else:
-                    instances = itertools.islice(instances, 0, self.max_instances)
-            if not self.manual_distributed_sharding and util.is_distributed():
-                from torch import distributed
-
-                logger.info(
-                    "Returning instances i%%%d==%d",
-                    distributed.get_world_size(),
-                    distributed.get_rank(),
+        if cache_file is not None:
```
dirkgr, line 183: Wouldn't it be easier (and faster) to have one cache file per worker, which contains only the instances for that one worker? Then you don't need locking or slicing in the cache case. The downside is that you can't re-use the cache when the number of workers changes.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -213,92 +163,103 @@ def read(self, file_path: str) -> Dataset:
...
-            return AllennlpDataset(instances)
+        Each object generated by this method is transformed into an instance
+        by the `DatasetReader.ensure_instance` method. The default implementation of
+        `ensure_instance` can handle several different data types, each of which are
+        interpreted differently.
 
-    def _get_cache_location_for_file_path(self, file_path: str) -> str:
-        return str(self._cache_directory / util.flatten_filename(str(file_path)))
+        Namely, `tuple`s are interpreted as positional arguments for `DatasetReader.text_to_instance`,
+        and `dict`s are treated as keyword arguments to `DatasetReader.text_to_instance`.
```
dirkgr, line 255: I'm with @DeNeutoy on this: Let's not have positional arguments at all.

allennlp/data/dataset_readers/dataset_reader.py
dirkgr: Can we just pass a dataset reader class instead? Or at least make that an option?

I feel strongly that the word "factory" should appear nowhere in our public API. Passing functions is a dodgy concept for many users. Passing members is worse. Passing members that return lazy generators is 🤯.  Sometimes these things are unavoidable, but I don't think this is one of those times.

epwalsh:
Ok, I pulled back the scope of this PR. This is focused only on bug fixes. The API of `DatasetReader` remains the same.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -335,5 +289,122 @@ def deserialize_instance(self, string: str) -> Instance:
...
+            start_index += worker_info.id
+            # Scale `step_size` by `num_workers`.
+            step_size *= worker_info.num_workers
+
+        islice = itertools.islice(iterable, start_index, self.max_instances, step_size)
+        if transform is not None:
+            return (transform(x) for x in islice)
+        return islice
+
+    def _instance_iterator(self, file_path: str) -> Iterable[Instance]:
```
epwalsh, line 332: Yea, there's a lot going on here.

Most of this comes from the now removes `_LazyInstances` class: https://github.com/allenai/allennlp/blob/master/allennlp/data/dataset_readers/dataset_reader.py#L72-L101

The additional complexity is from managing a file lock which is necessary to avoid bugs that might be very hard to catch.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -335,5 +289,122 @@ def deserialize_instance(self, string: str) -> Instance:
...
+            if util.is_distributed() or (get_worker_info() and get_worker_info().num_workers):
+                # But we can't write to the cache if there's more than one process loading
+                # instances since each worker only receives a partial share of the instances.
+                logger.warning(
+                    "Can't cache data instances when there are multiple processes loading data"
+                )
+                # Release the lock, then just read the instances as normal.
+                cache_file_lock.release()
+                for instance in self._multi_worker_islice(self._read(file_path)):
+                    yield instance
```
dirkgr, line 387: I think you can just say `yield from self._multi_worker_islice(...)`, or even `return self._multi_worker_islice(...)`. No loop necessary.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -335,5 +289,122 @@ def deserialize_instance(self, string: str) -> Instance:
...
+            else:
+                # Otherwise we can safely write to it.
+                try:
+                    with open(cache_file, "w") as data_file:
+                        for instance in self._multi_worker_islice(self._read(file_path)):
```
dirkgr, line 392: If you run this with `self._max_instances` first, it will create a cache file that also gets used later if you turn `self._max_instances` off.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -335,5 +289,122 @@ def deserialize_instance(self, string: str) -> Instance:
...
+                            data_file.write(self.serialize_instance(instance) + "\n")
+                            yield instance
+                except:  # noqa: E722
+                    # If anything went wrong, the cache file will be corrupted, so we should
+                    # remove it.
```
dirkgr, line 397: If the computer loses power while this is going on, we still end up with a half-written file. I think it's safer to write to a temporary file (`NamedTemporaryFile`?) in the same directory, and then rename it when the writing process finishes successfully.

I also don't know what happens if we yield a few instances, and then an exception gets thrown outside of here. Does that get propagated here? Does this `except` block run in that case?

I was curious, so I tried it. Looks like we don't see the exception:
```
In [1]: def f():
   ...:     try:
   ...:         yield 1
   ...:         yield 2
   ...:         yield 3
   ...:     except:
   ...:         print("We are excepting")
   ...:

In [2]: def g():
   ...:     g = f()
   ...:     print(next(g))
   ...:     print(next(g))
   ...:     raise ValueError(0)
   ...:     print(next(g))
   ...:

In [3]: g()
1
2
ValueError                                Traceback (most recent call last)
<ipython-input-3-5fd69ddb5074> in <module>
----> 1 g()

<ipython-input-2-a306c1db4465> in g()
      3     print(next(g))
      4     print(next(g))
----> 5     raise ValueError(0)
      6     print(next(g))
      7

ValueError: 0
```

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -169,6 +112,18 @@ class DatasetReader(Registrable):
         dataset in each worker, but filtering out the instances that are not needed. If you
         can implement a faster mechanism that only reads part of the data, set this to True,
         and do the sharding yourself.
+    manual_multi_process_sharding : `bool`, optional (default=`False`)
+        The is similar to the `manual_distributed_sharding` parameter, but applies to
+        multi-process data loading. By default, when this reader is used by a multi-process
+        data loader, each worker will filter out all but a subset of the instances
+        that are needed so that you don't end up with duplicates. If you
+        can implement a faster mechanism that only reads part of the data, set this to True,
+        and do the sharding yourself.
```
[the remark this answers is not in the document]
epwalsh: Hmm no I can't think of any reason to. Probably the only scenario where it makes sense to use is if you handle the filtering logic specially within `_read()`, in which case you'd want to set `manual_multi_process_sharding` to `True`.

Maybe we should make this more clear somehow.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -335,5 +289,122 @@ def deserialize_instance(self, string: str) -> Instance:
...
+            try:
+                cache_file_lock.acquire()
+            except Timeout:
+                # Couldn't acquire the lock within the timeout specified, so we'll
+                # just log a warning and ignore the cache.
+                logger.warning(
+                    "Failed to acquire lock on dataset cache file within {}s",
+                    self.CACHE_FILE_LOCK_TIMEOUT,
+                )
+                cache_file = None
```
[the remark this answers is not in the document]
epwalsh: Haha, yeup. Good catch.

allennlp/data/dataset_readers/dataset_reader.py
epwalsh: Good point.

allennlp/data/dataset_readers/dataset_reader.py
epwalsh: Hmm interesting. Yea I guess writing to a temp file would be better then 👌

allennlp/data/dataset_readers/dataset_reader.py
epwalsh: I like `yield from ..`. But I just found out that putting a `return` statement in a generator does some very strange things.

For example:

```
def foo():
    return (x for x in [1, 2])
    yield "hi"

list(foo())
```

The result is just an empty list 😕

dirkgr: Oh, interesting. So you can either `return my_iterator` or `yield from my_iterator`, but you better be consistent about it across different code paths?

Like this:
```
In [1]: def f():
   ...:     if True:
   ...:         return (x for x in [1, 2])
   ...:     else:
   ...:         yield "hi"
   ...:

In [2]: list(f())
Out[2]: []
```

That's ... something.

allennlp/data/dataset_readers/dataset_reader.py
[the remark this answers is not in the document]
matt-gardner: Maybe I misunderstood, but I think @dirkgr was saying something else: does this automatic sharding actually provide gains if every process is actually reading all of the data?  Would it be better to just disable this functionality until we have a better story?

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -289,14 +233,25 @@ def _read(self, file_path: str) -> Iterable[Instance]:
         raise NotImplementedError
 
     def _instances_from_cache_file(self, cache_filename: str) -> Iterable[Instance]:
+        # Try to acquire a lock just to make sure another process isn't in the middle
+        # of writing to the cache.
+        cache_file_lock = FileLock(cache_filename + ".lock", timeout=self.CACHE_FILE_LOCK_TIMEOUT)
```
matt-gardner, line 238: Putting the file locking logic inside this method makes it harder to override if you want better cache behavior.  Is it feasible to move the locking outside of this method?

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -289,14 +233,25 @@ def _read(self, file_path: str) -> Iterable[Instance]:
...
-            for line in cache_file:
-                yield self.deserialize_instance(line.strip())
+            for instance in self._multi_worker_islice(cache_file, self.deserialize_instance):
+                yield instance
 
     def _instances_to_cache_file(self, cache_filename, instances) -> None:
-        with open(cache_filename, "w") as cache:
-            for instance in Tqdm.tqdm(instances):
-                cache.write(self.serialize_instance(instance) + "\n")
+        with FileLock(cache_filename + ".lock", timeout=self.CACHE_FILE_LOCK_TIMEOUT):
```
matt-gardner, line 251: Same here with the lock file.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -335,5 +289,122 @@ def deserialize_instance(self, string: str) -> Instance:
...
+        node rank (for distributed training) and worker ID (for multi-process data loading).
+
+        # Parameters
+
+        iterable : `Iterable[Any]`
+            An iterable that yields raw data that can be transformed into `Instance`s
+            through the `transform` function.
+        transform : `Optional[Callable[[Any], Instance]]`, optional (default = `None`)
+            An optional function that will be applied to the raw data generated
+            by `iterable` to create `Instance`s.
```
matt-gardner, line 308: Might be nice to say what the intended use of this is.  The first time I read this code I was wondering what this could possibly be used for.  Then I saw it used when loading the cache file.
```suggestion
            by `iterable` to create `Instance`s. This is used, e.g., when reading cached data.
```

allennlp/data/dataset_readers/dataset_reader.py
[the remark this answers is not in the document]
epwalsh: I think we're all on the same page here. There are no gains unless you actually you implement a more efficient way of doing this within `_read()`.

I suggested we remove the `num_workers` parameter before, but @DeNeutoy thought it would be better to keep it so we at least are consistent with the PyTorch API, which I've come to agree with.

allennlp/data/dataset_readers/dataset_reader.py
epwalsh: Yea, that's feasible. I was just thinking that someone at some point might want to use a non-local cache like Redis, at which point the lockfile would be meaningless. But that would probably require them to override other functionality as well.

allennlp/data/dataset_readers/dataset_reader.py
epwalsh: Python encouraging people to shoot themselves in the foot.

epwalsh:
Ok, phew. I think I've addressed all of the comments so far and I added a TON of testing, especially around the caching+filelock system and the distributed/mutli-process slicing logic.

allennlp/data/dataset_readers/dataset_reader.py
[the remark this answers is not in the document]
dirkgr: I agree with being consistent. Let's clarify in the documentation, and maybe even with a warning, that setting `num_workers > 1` and also `manual_multi_process_sharding == False` is safe though useless most of the time. I mainly want to avoid people being confused about this.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -233,25 +267,28 @@ def _read(self, file_path: str) -> Iterable[Instance]:
...
+            # we are copying the file before closing it, so flush to avoid truncation
+            temp_file.flush()
+            # shutil.copyfileobj() starts at the current position, so go to the start
+            temp_file.seek(0)
+
+            logger.info("copying %s to cache at %s", temp_file.name, cache_filename)
             with open(cache_filename, "w") as cache:
-                for instance in Tqdm.tqdm(instances):
-                    cache.write(self.serialize_instance(instance) + "\n")
+                shutil.copyfileobj(temp_file, cache)
```
dirkgr, line 289: To do this right, you have to _move_/_rename_ the file. `copyfileobj` will copy the file one buffer at a time, and you could still end up with a truncated file if the power goes out.

This code does this: https://github.com/allenai/datastore/blob/master/python/datastore.py#L200
Though it's not very clear in that snippet, it makes sure that the temporary file and the final destination file are in the same directory (or at least on the same filesystem). Otherwise, that call to `Path().rename()` is not atomic (or it will fail, I don't remember which).

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -334,77 +388,66 @@ def _instance_iterator(self, file_path: str) -> Iterable[Instance]:
...
+                                yield instance
+
+                            # we are copying the file before closing it, so flush to avoid truncation
+                            temp_file.flush()
+                            # shutil.copyfileobj() starts at the current position, so go to the start
+                            temp_file.seek(0)
+
+                            logger.info("copying %s to cache at %s", temp_file.name, cache_file)
+                            with open(cache_file, "w") as cache:
+                                shutil.copyfileobj(temp_file, cache)
```
dirkgr, line 441: Is this here twice?

allennlp/data/dataset_readers/dataset_reader.py
epwalsh: That's a good point. We should update our `file_utils` caching system as well then.

allennlp/data/dataset_readers/dataset_reader.py
epwalsh: This handles lazy caching, the other place is non-lazy caching. I thought about trying to combine the logic but it didn't work out.

allennlp/common/file_utils.py
```diff
@@ -243,6 +242,46 @@ def _find_latest_cached(url: str, cache_dir: str) -> Optional[str]:
     return None
 
 
+class CacheFile:
```
dirkgr, line 245: That's some good stuff. I think the logging can be at the `debug` level, but other than that, this is the kind of abstraction we should have used in datastore or beakerstore.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -110,10 +110,11 @@ class DatasetReader(Registrable):
...
 
-        IMPORTANT CAVEAT: It is the _caller's_ responsibility to make sure that this directory is
-        unique for any combination of code and parameters that you use.  That is, if you pass a
-        directory here, we will use any existing cache files in that directory _regardless of the
-        parameters you set for this DatasetReader!_
+        !!! NOTE
+            It is the _caller's_ responsibility to make sure that this directory is
+            unique for any combination of code and parameters that you use.  That is, if you pass a
+            directory here, we will use any existing cache files in that directory _regardless of the
+            parameters you set for this DatasetReader!_
```
dirkgr, line 117: We could hash the parameters to the dataset reader and make the hash part of the filename. But that's for another PR.

allennlp/common/file_utils.py
dirkgr: I wish it had a better name, because you can use this for many things, not just caches. But `AtomicallyWrittenFile` is awkward, and I can't think of anything better.

allennlp/common/file_utils.py
```diff
@@ -243,6 +242,46 @@ def _find_latest_cached(url: str, cache_dir: str) -> Optional[str]:
...
+    This is a context manager that makes robust caching easier.
+
+    On `__enter__`, an IO handle to a temporarily file is returned, which can
+    be treated as if it's the actual cache file.
+
+    On `__exit__`, the temporarily file is renamed to the cache file. If anything
+    goes wrong while writing to the temporary file, it will be removed.
+    """
+
+    def __init__(self, cache_filename: Union[Path, str], mode="w+b") -> None:
```
dirkgr, line 256: The default file mode should probably be whatever the default is in `open()`, which is `t`, not `b`. Just to minimize surprise.

allennlp/common/file_utils.py
epwalsh: Yea... I struggled to come up with anything better

allennlp/data/dataset_readers/dataset_reader.py
epwalsh: That's a good thought.

allennlp/common/file_utils.py
epwalsh: `w+b` is the default for `NamedTemporaryFile` though. I'm fine with changing it, but if we're consistent with one we're inconsistent with the other

allennlp/data/dataset_readers/dataset_reader.py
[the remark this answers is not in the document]
epwalsh: I just added a warning

allennlp/common/file_utils.py
dirkgr: That seems like an oversight in Python then. But yeah, I agree, let's be consistent with `NamedTemporaryFile` then.

PR #4346, opened 2020-06-10
remove links to tutorials in API docs

epwalsh:


epwalsh:
Oh no worries, I just happened to see this in warnings. There might actually be a setting that forces the doc build step to fail when there a broken links. That would be good to have.

PR #4350, opened 2020-06-10
Fix handling of "datasets_for_vocab_creation" param

epwalsh:
This fixes a couple of bugs related to the `datasets_for_vocab_creation` config parameter.

The main bug is that `datasets_for_vocab_creation: []` ends up being treated like `datasets_for_vocab_creation: null` due to this statement in `commands.train.TrainModel.from_partial_objects`:

```python
        instance_generator = (
            instance
            for key, dataset in datasets.items()
            if not datasets_for_vocab_creation or key in datasets_for_vocab_creation
            for instance in dataset
        )
```

which should be

```python
        instance_generator = (
            instance
            for key, dataset in datasets.items()
            if datasets_for_vocab_creation is None or key in datasets_for_vocab_creation
            for instance in dataset
        )
```

The 2nd bug occurs when `training.util.vocab_from_params(...)` is called. Even if `datasets_for_vocab_creation` is set to `[]`, all of the datasets will be initialized. This means that with non-lazy readers, you read all of the data unnecessarily.

### DIFF
--- a/allennlp/training/util.py
+++ b/allennlp/training/util.py
@@ -450,6 +450,9 @@ def make_vocab_from_params(
         for instance in dataset
     )
 
+    if print_statistics:
+        instances = list(instances)
+
     vocab = Vocabulary.from_params(vocab_params, instances=instances)
 
     logger.info(f"writing the vocabulary to {vocab_dir}.")
--- a/allennlp/training/util.py
+++ b/allennlp/training/util.py
@@ -431,14 +431,15 @@ def make_vocab_from_params(
     all_datasets = datasets_from_params(params)
     datasets_for_vocab_creation = set(params.pop("datasets_for_vocab_creation", all_datasets))
 
-    for dataset in datasets_for_vocab_creation:
-        if dataset not in all_datasets:
-            raise ConfigurationError(f"invalid 'dataset_for_vocab_creation' {dataset}")
+    if datasets_for_vocab_creation:
+        for dataset in datasets_for_vocab_creation:
+            if dataset not in all_datasets:
+                raise ConfigurationError(f"invalid 'dataset_for_vocab_creation' {dataset}")
 
-    logger.info(
-        "From dataset instances, %s will be considered for vocabulary creation.",
-        ", ".join(datasets_for_vocab_creation),
-    )
+        logger.info(
+            "From dataset instances, %s will be considered for vocabulary creation.",
+            ", ".join(datasets_for_vocab_creation),
+        )
 
     instances: Iterable[Instance] = (
         instance
--- a/allennlp/training/util.py
+++ b/allennlp/training/util.py
@@ -123,6 +123,9 @@ def datasets_from_params(
     """
     datasets: Dict[str, Dataset] = {}
 
+    train = train and ("train_data_path" in params)
+    validation = validation and ("validation_data_path" in params)
+    test = test and ("test_data_path" in params)
     if not any((train, validation, test)):
         # Return early so don't unnecessarily initialize the train data reader.
         return datasets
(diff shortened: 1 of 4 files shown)


dirkgr:
This is exactly the kind of bug that makes me stay away from Python falsyness in virtually all cases.

allennlp/training/util.py
```diff
@@ -428,38 +415,43 @@ def make_vocab_from_params(
...
     instances: Iterable[Instance] = (
         instance
-        for key, dataset in all_datasets.items()
-        if key in datasets_for_vocab_creation
+        for key, dataset in datasets.items()
+        if datasets_for_vocab_creation is None or key in datasets_for_vocab_creation
         for instance in dataset
     )
 
-    if print_statistics:
```
matt-gardner, line 450: I'm pretty sure this was here because in a prior version of the code, `instances` was a generator that could get consumed when constructing the vocabulary, and then nothing would get printed.  I assume that's no longer an issue?  I've lost track of how the datasets work now.

epwalsh: Yea, idk. It seems like a hack to turn the dataset into a big batch just to print statistics.

allennlp/training/util.py
```diff
@@ -428,28 +418,38 @@ def make_vocab_from_params(
...
-
-    logger.info(
-        "From dataset instances, %s will be considered for vocabulary creation.",
-        ", ".join(datasets_for_vocab_creation),
+    datasets_for_vocab_creation: Optional[List[str]] = params.pop(
+        "datasets_for_vocab_creation", None
     )
+    # Do a quick sanity check here. There's no need to load any datasets if the vocab
+    # type is "empty".
+    if datasets_for_vocab_creation is None and vocab_params.get("type") == "empty":
```
epwalsh, line 426: I'm not crazy about this ad hoc solution here but I don't see another way. At least I added a test for this case to make sure we don't get a regression in the future.

matt-gardner: A possible alternative is to build a lazy generator here that doesn't actually call `.read()` on anything unless it's needed.  Not sure how feasible that is, though.

matt-gardner: One obvious concern is that it requires changes to `datasets_from_params` also in ways that change the return type of that method; I'm pretty sure, though that this is the only use of `datasets_from_params`, and this function is only ever called on the master process in a distributed setting.  In every other configuration, we go through a different code path.

epwalsh: It's also called from the `find_learning_rate` command, but that's it.

epwalsh: This is another reason why I think readers should always be lazy. Leave the lazy/non-lazy decision up to the dataset.

matt-gardner: Yeah, I can see that.  Where do you make the lazy / non-lazy decision, then?  You have to make a choice once you get to the `DataLoader`, because of how batching / sampling works.  You just base the decision of which dataset to use on how you've decided to do data loading?

epwalsh: I was either thinking we'd make that decision part of the `DataLoader` API, or have a separate configuration for the type of `Dataset` to use.

I like the first option because the lazy vs. non-lazy is literally an issue of data loading. The downside though is that we'd be breaking from the PyTorch API. But having to separately configure a `DatasetReader`, `Dataset`, and `DataLoader` seems overboard.

mateuszpieniak:
Lol, I discovered this some time ago, but I had thought it was on purpose. @epwalsh What is the best way to avoid building vocabulary when I use prebuilt transformers? Should I use ```datasets_for_vocab_creation: []``` or is there something else?

epwalsh:
@mateuszpieniak just using `{vocab: {type: empty}}` will work now.

PR #4371, opened 2020-06-17
ensure Vocab.from_files and ShardedDatasetReader can handle archives

epwalsh:
Note: the logic I added in `cached_path` comes from here https://github.com/huggingface/transformers/blob/master/src/transformers/file_utils.py#L289.


allennlp/common/file_utils.py
```diff
@@ -85,34 +88,96 @@ def filename_to_url(filename: str, cache_dir: str = None) -> Tuple[str, str]:
...
 
+    file_path: str
+    extraction_path: Optional[str] = None
+
     if parsed.scheme in ("http", "https", "s3"):
         # URL, so get it from the cache (downloading if necessary)
-        return get_from_cache(url_or_filename, cache_dir)
+        file_path = get_from_cache(url_or_filename, cache_dir)
+
+        if extract_archive and (is_zipfile(file_path) or tarfile.is_tarfile(file_path)):
```
matt-gardner, line 135: Seems like we should pull this out into a separate, small pip utility, that can be shared by both us and huggingface.

allennlp/data/vocabulary.py
```diff
@@ -316,12 +317,24 @@ def from_files(
         # Parameters
```
matt-gardner, line 317: It's high enough above this that I can't directly make a suggestion to change it, but can you modify the main text of the docstring to say something like `Loads a Vocabulary that was serialized either using save_to_files or inside a model archive file`?

epwalsh: Will do

Issue #4339, opened 2020-06-08
Make an AllenNLP Project Template


dirkgr:
We should have a template GitHub repo for people to get started with.

matt-gardner:
Is this good enough as a template? https://github.com/allenai/allennlp-guide-examples/tree/master/quick_start

schmmd:
@dirkgr great suggestion.  @matt-gardner possibly, but if we make it a GitHub template (e.g. https://github.com/allenai/skiff-template) it'd be even easier to get started...

schmmd:
We might consider three templates:

1 - python template
2 - configuration file
3 - pytorch lightening

schmmd:
Closinga s this is done.  @matt-gardner can you link to the other tempalte?

schmmd:
Templates are here: https://github.com/allenai/allennlp#getting-started-using-the-library

PR #4372, opened 2020-06-17
Add link to new template repo for config file development

matt-gardner:
Partial fix for #4339.  Not saying it closes that issue, because there are two more templates to make first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,13 @@
 - [Nightly Releases](https://pypi.org/project/allennlp/#history)
 - [Officially Supported Models](https://github.com/allenai/allennlp-models)
 
+## Getting Started Using the Library
+
+If you're interested in using AllenNLP for model development, we recommend you check out the
+[AllenNLP Guide](https://guide.allennlp.org).  When you're ready to start your project, we've
+created a template repository that you can use as a starting place:
+https://github.com/allenai/allennlp-template-config-files.
+
 ## Package Overview
 
 <table>


README.md
```diff
@@ -34,6 +34,13 @@
 - [Nightly Releases](https://pypi.org/project/allennlp/#history)
 - [Officially Supported Models](https://github.com/allenai/allennlp-models)
 
+## Getting Started Using the Library
+
+If you're interested in using AllenNLP for model development, we recommend you check out the
+[AllenNLP Guide](https://guide.allennlp.org).  When you're ready to start your project, we've
+created a template repository that you can use as a starting place:
+https://github.com/allenai/allennlp-template-config-files.
```
matt-gardner, line 42: I'm planning on making two more template repositories, one for development using your own script instead of `allennlp train`, and one for doing that along with using pytorch lightning as your trainer, instead of the allennlp trainer.  Then this will be a list, instead of a single link.

matt-gardner:
Did you look at the template repo at all?  Any comments on it?

epwalsh:
Do you think it's worth adding a `dev-requirements.txt` file with at least pytest in it?

matt-gardner:
Hmm, for a template like this, having two requirements files seems like overkill, but yeah, I'll add a bit more there about pytest.  I think I forgot to say how to run the tests, and I'll add pytest to `requirements.txt`.  Thanks!

PR #4377, opened 2020-06-18
ensure transformer params are frozen at initialization when train_parameters is false

epwalsh:
If we don't do this at initialization, then in the training logs it will list the transformer's parameters as trainable, and it will also cause model tests to fail on the `check_model_computes_gradient_correctly` method if your test model has `train_parameters` set to `False`.


allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -62,6 +62,10 @@ def __init__(
         self._num_added_end_tokens = len(tokenizer.single_sequence_end_tokens)
         self._num_added_tokens = self._num_added_start_tokens + self._num_added_end_tokens
 
+        if not self._train_parameters:
```
matt-gardner, line 65: Do you even need to keep this around as a member variable anymore?

epwalsh: Probably not

PR #4379, opened 2020-06-18
make transformer tests run on self-hosted runner

epwalsh:
Tests that involve pre-trained transformers are often longer-running and require big downloads that can't be cached on GitHub runners. Therefore it makes sense to run these on the self-hosted runner so we can utilize the runner's local cache and not have to continuously keep downloading the same files.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -32,7 +32,26 @@ jobs:
 
     - name: Run GPU tests
       run: |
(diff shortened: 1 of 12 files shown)


epwalsh:
@matt-gardner that's probably a better idea, although it seems there's some weird caveats with that particular model.

I'm going to look into training a custom super tiny transformer that we can use.

Besides, a downside of the solution in this PR is that the test coverage reported would be lower than it actually is.

PR #4381, opened 2020-06-18
Add a method to ModelTestCase for use without config files

matt-gardner:
This was necessary to have tests for the no-config-file template repo.  I also noticed a few other small typos in the documentation, which are fixed in here.

### DIFF
--- a/allennlp/common/testing/model_test_case.py
+++ b/allennlp/common/testing/model_test_case.py
@@ -12,6 +12,7 @@ from allennlp.data import DatasetReader, Vocabulary
 from allennlp.data import DataLoader
 from allennlp.data.batch import Batch
 from allennlp.models import load_archive, Model
+from allennlp.training import GradientDescentTrainer
 
 
 class ModelTestCase(AllenNlpTestCase):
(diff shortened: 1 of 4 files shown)


allennlp/common/testing/model_test_case.py
```diff
@@ -177,6 +178,63 @@ def ensure_model_can_train_save_and_load(
...
+        A simple test for model training behavior when you are not using configuration files. In
+        this case, we don't have a story around saving and loading models (you need to handle that
+        yourself), so we don't have tests for that.  We just test that the model can train, and that
+        it computes gradients for all parameters.
+
+        Because the `Trainer` already has a reference to a model and to a data loader, we just take
+        the `Trainer` object itself, and grab the `Model` and other necessary objects from there.
+
+        # Parameters
+
```
epwalsh, line 200: Missing `trainer` param here

matt-gardner: Thanks, good catch.

matt-gardner:
@epwalsh, this is now failing docs builds, and I am having a really hard time figuring out what the offending line is.  Any help?  Is there a way to have the scripts actually output the line that's the problem?

epwalsh:
That’s weird, I’ve never seen that before. I’ll try to debug that on Monday

Issue #4378, opened 2020-06-18
PDB++ arrow/tab keys don't work with allennlp


successar:
## Checklist

- [x] I have verified that the issue exists against the `master` branch of AllenNLP.
- [ ] I have read the relevant section in the [contribution guide](https://github.com/allenai/allennlp/blob/master/CONTRIBUTING.md#bug-fixes-and-new-features) on reporting bugs.
- [x] I have checked the [issues list](https://github.com/allenai/allennlp/issues) for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/allenai/allennlp/pulls) for existing proposed fixes.
- [x] I have checked the [CHANGELOG](https://github.com/allenai/allennlp/blob/master/CHANGELOG.md) and the [commit log](https://github.com/allenai/allennlp/commits/master) to find out if the bug was already fixed in the master branch.
- [x] I have included in the "Description" section below a traceback from any exceptions related to this bug.
- [x] I have included in the "Related issues or possible duplicates" section beloew all related issues and possible duplicate issues (If there are none, check this box anyway).
- [x] I have included in the "Environment" section below the name of the operating system and Python version that I was using when I discovered this bug.
- [x] I have included in the "Environment" section below the output of `pip freeze`.
- [x] I have included in the "Steps to reproduce" section below a minimally reproducible example.

## Description

I can't use pdb/pdb++ arrow keys or tab key for autocompletion when running `allennlp train` . Possibly related to this issue https://stackoverflow.com/a/56418204

</details>

## Related issues or possible duplicates

- Reopening this issue - https://github.com/allenai/allennlp/issues/2176

## Environment

OS: Linux Ubuntu 18.04

Python version: 3.7.7

<details>
<summary><b>Output of <code>pip freeze</code>:</b></summary>
<p>

```
allennlp==1.0.0
appdirs==1.4.3
astroid==2.4.1
attrs==19.3.0
black==19.10b0
blis==0.4.1
boto3==1.14.4
botocore==1.17.4
catalogue==1.0.0
certifi==2020.4.5.2
chardet==3.0.4
click==7.1.2
cymem==2.0.3
docutils==0.15.2
fancycompleter==0.9.1
filelock==3.0.12
future==0.18.2
gnureadline==8.0.0
h5py==2.10.0
idna==2.9
importlib-metadata==1.6.1
isort==4.3.21
jmespath==0.10.0
joblib==0.15.1
jsonnet==0.16.0
[...]

PR #4383, opened 2020-06-19
Automatic file-friendly logging

dirkgr:
This is a fix for https://github.com/allenai/allennlp/issues/4378.

@DeNeutoy, why do we redirect `sys.stdout` and `sys.stderr` in the first place? Was this a big problem at some point that lots of output went there instead of through the `logging` framework?


dirkgr:
@matt-gardner says we don't need to redirect `std*`. We can just use this solution all the time and not have a flag.

dirkgr:
Also, commands other than `train` should set up logging this way.

DeNeutoy:
@dirkgr I'm not 100% on this, but I don't think the `logging.FileHandlers` collect things which aren't logging statements? I.e when your code crashes, the stack trace does not end up in the file, unless it's wrapped in a try except block and you pass `exc_info=True`.

epwalsh:
I think Mark is right. We patched these to ensure everything is captured in the log files. See https://github.com/allenai/allennlp/pull/4164

allennlp/commands/train.py
```diff
@@ -91,6 +91,14 @@ def add_subparser(self, parser: argparse._SubParsersAction) -> argparse.Argument
             "other training information",
         )
 
+        subparser.add_argument(
+            "--no-redirect-std",
+            action="store_false",
+            dest="redirect_std",
+            default=True,  # This is confusing. `True` gets stored into the `redirect_std` variable.
+            help="do not redirect stdout and stderr to the log files. This can be helpful for debugging.",
```
epwalsh, line 99: I'm not sure why this would be helpful for debugging. You should see the same terminal output regardless of whether or not you use this flag. But without this flag, everything output to the terminal will also end up in `stdout.log` or `stderr.log`. So not using the flag seems like it would be more helpful for debugging.

matt-gardner: Maybe "when using a console-based debugger like pdb".

epwalsh: Oh right, good point Matt, I forgot this fixes (though I don't know why) the issue with pdb.

dirkgr:
I think we could merge this with a renamed flag, but I wanted to give this one more attempt at coming up with a solution that doesn't require any flags.

6b1eb77 overrides the unhandled exception handler so that those get written to the log files. We will still miss instances of `print()` and `sys.stderr.write()` now, but how often do those really happen?

dirkgr:
How about this?
* No more patching stderr or stdout
* Works in the debugger
* `tqdm` works on the console and in the log files.
  * The "100%" message shows up in the logs. This is a bit of a hack, but it works.
* Exceptions show up in the logs, including stack trace.
  * ... but Ctrl-C still works
* It's 56 lines less.
* No new flags
* No need for `--file-friendly-logging` now. All logging is file friendly, with no compromise in logging fidelity on the command line.

The downside is still the same as before:
 * `print()` and `sys.stderr.write()` don't show up in the logs anymore.

There is a way to make sure that log messages that are written to the console don't interfere with progress bars. That was already broken before, and this is complicated enough as it is, so I left it for a future PR.

allennlp/common/tqdm.py
```diff
@@ -22,29 +28,32 @@
...
-        else:
-            Tqdm.default_mininterval = 0.1
+class TqdmToLogsWriter(object):
+    def __init__(self):
+        self.last_message_written_time = 0
 
+    def write(self, message):
+        sys.stderr.write(message)
+        now = time()
+        if now - self.last_message_written_time >= 10 or "100%" in message:
```
dirkgr, line 42: The one weirdness with this "100%" hack is that `tqdm` will often write two messages that say "100%" in rapid succession. I don't know why this happens. In principle, we want the second one of the two, because that's what you see on the console, and that's the one with the accurate items-per-second count. I can filter so it shows only the first one of the two, but not so that it only shows the second one. So I decided it's better to have the message there twice than to have an inaccurate items-per-second number in the logs.

schmmd:
I believe `file-friendly-logging` was added to specifically address terrible output in Beaker when running AllenNLP jobs there.  Specifically Beaker would capture rather terrible TQDM output on stdout, which would be written to a file.  We should make sure we continue to support this use case after this change.

allennlp/common/tqdm.py
epwalsh: Could you add this as a comment here?

DeNeutoy:
IMO this is a hard requirement of logging?

> * `print()` and `sys.stderr.write()` don't show up in the logs anymore.

 Maybe others think differently but I would certainly want this appearing in logging, particularly if I am running a lot of experiments programmatically e.g using slurm.

dirkgr:
@DeNeutoy, virtually nothing writes to stderr/stdout now. Even huggingface cleaned up their messages. If capturing those is a hard requirement, I don't see a way of doing it without having both flags (or giving up on the debugger with `wontfix`).

I also don't think it's good software design to redirect stdout/stderr in general. This is the kind of thing that makes a library hard to use, because it produces unexpected interactions with other tools (such as pdb in this case).

DeNeutoy:
🤷  ok, I guess I disagree. For me 95% of the value of having logs redirected to files is capturing stack traces for errors in jobs run remotely. If we fail to capture these in the logs I think many people will be unhappy. Obviously we could just remove this entirely and force people to use `tee` to create logs themselves.

matt-gardner:
I would agree that having reasonable access to stack traces in common environments is a pretty hard requirement, and it's pretty bad if this makes it impossible / dramatically harder to get them.

dirkgr:
We have stack traces in this version. That's what https://github.com/allenai/allennlp/pull/4383/files#diff-8c42826087f72f894eaa54381512061dR98 is all about.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Added a way to turn of redirecting output to stderr and stdout
```
epwalsh, line 12: I think this needs to be updated, and would probably be better to put under the `## Changed` section

PR #4385, opened 2020-06-19
Add a section for external tutorials

schmmd:
We may need to rework this when we have many.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,10 @@ If you're interested in using AllenNLP for model development, we recommend you c
 created a template repository that you can use as a starting place:
 https://github.com/allenai/allennlp-template-config-files.
 
+In addition, there are external tutorials:
+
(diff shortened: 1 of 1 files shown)


dhruvdcoder:
You could also include this micro-library called [wandb-allennlp](https://pypi.org/project/wandb-allennlp/) which enables the use of Weight&Biases to perform hyperparameter search for AllenNLP models. I have been using W&B with AllenNLP models for quite some time now and hence decided to create a package for others to use the same setup. 

Example search sweep visualization [link](https://app.wandb.ai/dhruveshpate/wandb_allennlp_models_demo/sweeps/vwwu3sa0)

Usage details are in the readme-files in the [repo](https://github.com/dhruvdcoder/wandb-allennlp). A short tutorial can be found [here](http://dhruveshp.com/machinelearning/wandb-allennlp/). Although the tutorial is for AllenNLP v0.9, the library works with v1.0 as well. 

I would be happy to give more information if needed.

dirkgr:
I'm excited about W&B. I made a separate issue for it at #4387. If we can change the way something works in AllenNLP to make that integration work better, let us know!

PR #4400, opened 2020-06-25
Adds the ability to automatically detect whether we have a GPU

dirkgr:
~~With this PR, when you set `cuda_device: -2` in the config, it automatically uses one GPU if there is one.~~

When you don't set `cuda_device` at all, it automatically chooses one GPU of there is one.

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -51,7 +51,7 @@ class Trainer(Registrable):
     def __init__(
         self,
         serialization_dir: str,
-        cuda_device: Union[None, int, torch.device] = None,
+        cuda_device: Optional[Union[int, torch.device]] = None,
         distributed: bool = False,
         local_rank: int = 0,
         world_size: int = 1,
@@ -342,7 +342,7 @@ class GradientDescentTrainer(Trainer):
         num_epochs: int = 20,
         serialization_dir: Optional[str] = None,
         checkpointer: Checkpointer = None,
-        cuda_device: int = -1,
+        cuda_device: Optional[Union[int, torch.device]] = None,
         grad_norm: Optional[float] = None,
         grad_clipping: Optional[float] = None,
         learning_rate_scheduler: Optional[LearningRateScheduler] = None,
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -1023,7 +1023,7 @@ class GradientDescentTrainer(Trainer):
         patience: int = None,
         validation_metric: str = "-loss",
         num_epochs: int = 20,
-        cuda_device: int = -1,
+        cuda_device: Optional[Union[int, torch.device]] = None,
         grad_norm: float = None,
         grad_clipping: float = None,
         distributed: bool = None,
@@ -1055,6 +1055,13 @@ class GradientDescentTrainer(Trainer):
         If you're not using `FromParams`, you can just construct these arguments in the right order
         yourself in your code and call the constructor directly.
         """
+        if cuda_device is None:
+            from torch import cuda
+
+            if cuda.device_count() > 0:
+                cuda_device = 0
+            else:
+                cuda_device = -1
 
         check_for_gpu(cuda_device)
         if cuda_device >= 0:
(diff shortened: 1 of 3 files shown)


allennlp/training/trainer.py
```diff
@@ -56,6 +56,12 @@ def __init__(
         local_rank: int = 0,
         world_size: int = 1,
     ) -> None:
+        if cuda_device == -2:
```
matt-gardner, line 59: I like the feature that this is adding.  I'm not thrilled with using `-2` to specify it; it's too magic.  At the very least this needs better documentation.  I'd be happier if there were a better value to pass in to specify this.  E.g., you could do some string value that at least has a meaningful name, though I'm not super happy with mixing strings and ints here, either.

dirkgr: I'd probably be happiest if the default for this setting is `None`, and if you don't specify anything in the config it auto-detects.

matt-gardner: That seems totally reasonable to me.  I was going to suggest using `None` to mean that, but it was weird with a default value of `-1`.  If you switch the default to `None`, it works out.

schmmd:
Automatically choosing a GPU if there is one seems like a huge win!

allennlp/training/trainer.py
dirkgr: I made that change. Can you review again?

dirkgr:
@epwalsh, something is wrong with the GPU runner. Can you take a look?

This seems like a particularly important PR to have the GPU tests run.

allennlp/training/trainer.py
```diff
@@ -51,11 +51,18 @@ class Trainer(Registrable):
     def __init__(
         self,
         serialization_dir: str,
-        cuda_device: Union[int, torch.device] = -1,
+        cuda_device: Union[None, int, torch.device] = None,
```
matt-gardner, line 54: I think `None` here might mess with the `FromParams` logic. Can you make this the last type argument, at least?

Or maybe `None` gets stripped already because of how `Optional` is handled...?  Nope, the special case there wouldn't handle this: https://github.com/allenai/allennlp/blob/aa2943e50d8ee81ea86a907c72f5d56a83ddc2fc/allennlp/common/from_params.py#L108-L109

Can you either just remove the `None` here in the type annotation, or have a test that explicitly checks that this gets the right value with various configuration options?  I'm nervous that this won't actually work as it is.

tests/commands/train_test.py
```diff
@@ -50,6 +50,7 @@ def __call__(  # type: ignore
 
 
 class TestTrain(AllenNlpTestCase):
+    @cpu_or_gpu  # This is marked like this to test the automatic detection of GPU presence.
```
matt-gardner, line 53: But it doesn't actually test the automatic detection of GPU presence, does it?  We won't know if this runs on CPU on the GPU machine.  Or am I missing something?

dirkgr: I guess that's true. I was thinking of it from a code coverage perspective. It will run through all the lines. But I'll add an `assert` that makes sure it actually chooses the right one.

allennlp/training/trainer.py
dirkgr: So if I change it to `Optional[Union[int, torch.device]]`, `FromParams` will work?

matt-gardner: It might still work now, I just am not sure, and that makes me nervous.

dirkgr:
This turned out to be much harder to fix than I thought, so I beefed up the tests as well. @matt-gardner, can you take another look?

allennlp/training/trainer.py
```diff
@@ -1048,6 +1055,13 @@ def from_partial_objects(
         If you're not using `FromParams`, you can just construct these arguments in the right order
         yourself in your code and call the constructor directly.
         """
+        if cuda_device is None:
```
matt-gardner, line 1058: Unfortunate that this has to be in both places, but it looks like that's necessary.  It might work to put this into a shared method or something, though, like the existing `check_for_gpu` or `int_to_device`.

dirkgr: I'll do that the third time I need it :-)

PR #4403, opened 2020-06-26
Assign an empty mapping array to empty fields of `NamespaceSwappingField`

pcyin:
Assign an empty mapping array to empty fields created by `NamespaceSwappingField`. Leaving `._mapping_array ` unassigned (`None`) will cause issues in batching when `NamespaceSwappingField` is used as members in `ListField`.


matt-gardner:
Thanks @pcyin! Can you add a quick note to `CHANGELOG.md` saying something like `Fixed a bug where NamespaceSwappingField did not work correctly with .empty_field().`

pcyin:
Updated `CHANGELOG.md`. Thanks!

PR #4415, opened 2020-06-29
More multiple-choice changes

dirkgr:


allennlp/training/learning_rate_schedulers/linear_with_warmup.py
```diff
@@ -0,0 +1,46 @@
+from overrides import overrides
+import torch
+
+from allennlp.training.learning_rate_schedulers.learning_rate_scheduler import LearningRateScheduler
+
+
+@LearningRateScheduler.register("linear_with_warmup")
+class LinearWithWarmup(LearningRateScheduler):
+    """
+    Implements a learning rate scheduler that increases the learning rate to `lr` during the first
```
schmmd, line 10:
```suggestion
    Implements a learning rate scheduler that increases the learning rate to `lr` during the third
```

schmmd: This was just a test during the intern orientation.

allennlp/training/learning_rate_schedulers/linear_with_warmup.py
```diff
@@ -0,0 +1,46 @@
+from overrides import overrides
+import torch
+
+from allennlp.training.learning_rate_schedulers.learning_rate_scheduler import LearningRateScheduler
+
+
+@LearningRateScheduler.register("linear_with_warmup")
```
Tobias-Rohde, line 7: (I saw this by coincidence during the tech orientation)

Why do we need this class? It seems to be exactly the same as `PolynomialDecay` with `end_learning_rate=0.0` and `power=`1.0`. Perhaps we should fix up `PolynomialDecay` to take in `num_epochs` and `num_steps_per_epoch` instead of `total_steps` and then get rid of `LinearWithWarmup`?

dirkgr: I want to keep it as a class, because it's easier to remember when you're writing a config, but `LinearWithWarmup` could just be a sub-class of `PolynomialDecay`, and pass in the right parameters.

PR #4416, opened 2020-06-29
generalize DataLoader

epwalsh:
This makes `allennlp.data.DataLoader` an abstract, registrable class. The default implementation
remains the same, but was renamed to is `allennlp.data.PyTorchDataLoader`.


CHANGELOG.md
```diff
@@ -27,6 +27,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added the ability to pass an archive file instead of a local directory to `Vocab.from_files`.
 - Added the ability to pass an archive file instead of a glob to `ShardedDatasetReader`.
 
+### Changed
+
+- `allennlp.data.DataLoader` is now an abstract registrable class. The default implementation
+remains the same, but was renamed to is `allennlp.data.PyTorchDataLoader`.
```
matt-gardner, line 33:
```suggestion
remains the same, but was renamed to `allennlp.data.PyTorchDataLoader`.
```

allennlp/training/trainer.py
```diff
@@ -328,10 +328,10 @@ def __init__(
         self,
         model: Model,
         optimizer: torch.optim.Optimizer,
-        data_loader: torch.utils.data.DataLoader,
```
matt-gardner, line 331: There are small parts of the guide that should be updated with this change.  Not sure how to keep this in sync with the guide / when to release things.

epwalsh: I feel like the guide should be targeted at the latest stable version (1.0.0 right now), right? But it would definitely be good to have some CI for the guide. I don't know how hard that would be.

allennlp/data/dataloader.py
```diff
@@ -20,7 +20,28 @@ def allennlp_collate(instances: List[Instance]) -> TensorDict:
...
+    This class only has one required method, `__iter__()`, that creates an iterable
+    of `TensorDict`s. Additionally, this class comes with a `__len__()` method
+    that just raises a `TypeError` by default. When possible, this should be overriden
+    to return the number of batches that will be generated by the `__iter__()` method.
+    """
+
+    default_implementation = "pytorch_dataloader"
+
+    def __len__(self) -> int:
+        raise TypeError
```
dirkgr, line 37: Why `TypeError`? Is that what the torch one does?

allennlp/training/trainer.py
dirkgr: Does this still refer to the pytorch `DataLoader` class? It should if we want this trainer to work with arbitrary `DataLoader`s.

allennlp/data/dataloader.py
epwalsh: Yea, `TypeError` is raised when you use a lazy dataset, and it's also what's raised when `len()` is called on an object that doesn't implement `__len__`. So we'd get the same result by just not including `__len__` in this base class. I just thought it's better to be explicit so that it's obvious `__len__` should be implemented by subclasses whenever possible.

allennlp/training/trainer.py
epwalsh: This refers to the new general base class

dirkgr: So the trainer won't work with non-AllenNLP data loaders? That is a problem, right?

epwalsh: It will work with any class that implements `allennlp.data.DataLoader`, such as `allennlp.data.PyTorchDataLoader`. So this is more general than before.

PR #4420, opened 2020-06-30
Adds Detectron support

dirkgr:


allennlp/models/detectron.py
```diff
@@ -0,0 +1,23 @@
...
+        **kwargs,
+    ) -> None:
+        super().__init__(vocab, **kwargs)
+        from torch.nn import Embedding
+        self.embedding = Embedding(10, 10)  # dummy
+
+    def forward(  # type: ignore
+        self, image: Any
+    ) -> Dict[str, torch.Tensor]:
+        return {"loss": torch.tensor(0.0)}
```
dirkgr, line 23: On this line, `image` contains the detectron images in the correct format for the detectron models.

dirkgr:
This new version can now run a detectron model, though only in inference mode. I didn't take the time to figure out how to make sure that the dataset reader loads data in the same way that the model needs it, i.e., with the right number of classes, etc. I suspect it's a matter of matching up the `cfg`s.

The next obvious steps here are to make both the dataset reading and the model loading configurable through the jsonnet files.

matt-gardner:
Maybe make a new `vision` branch that we can merge things into for now?

allennlp/data/fields/detectron_field.py
```diff
@@ -0,0 +1,45 @@
...
+    @overrides
+    def as_tensor(self, padding_lengths: Dict[str, int]) -> DataArray:
+        return self.image_dict  # type: ignore
+
+    @overrides
+    def empty_field(self) -> "DetectronField":
+        return DetectronField({})
+
+    @overrides
+    def batch_tensors(self, tensor_list: List[DataArray]) -> List[DataArray]:  # type: ignore
```
jiasenlu, line 41: we can use this (https://github.com/facebookresearch/detectron2/blob/master/detectron2/structures/image_list.py#L8) from detectron2 to convert the instance to batched tensors.

dirkgr: I want to make sure that what we get in the `Model.forward()` method is exactly what we need for detectron's `forward()` and `inference()` methods. It seems inefficient to me, but it looks like they just take `List[Dict]`, not properly batched tensors. That's why this is the way it is.

dirkgr:
I'm closing this to make a PR that merges into the `vision` branch.

PR #4435, opened 2020-07-02
Detectron integration

dirkgr:


dirkgr:
The reader can now read Detectron's YAML-based configs, and also apply overrides from the AllenNLP jsonnet files. That was actually only a few lines of code.

allennlp/data/fields/tensor_field.py
```diff
@@ -0,0 +1,48 @@
+from typing import Dict, Any
+
+import torch
+from overrides import overrides
+
+from allennlp.data.fields.field import Field
+
+
+class TensorField(Field[torch.Tensor]):
```
dirkgr, line 9: @matt-gardner, I would really like your opinion on this. I was surprised that we didn't have such a field already. This is basically the same as `ArrayField`, but using torch tensors.

matt-gardner: I'd be happier if we could just allow `ArrayField` to take a tensor as input, instead of having basically identical fields, but if that's really undesirable for some reason I could live with this.

dirkgr: Torch tensors can live on GPUs, but numpy arrays can not. Also, it's an extra step to convert from one to the other.

I'd rather do it the other way around, i.e, change `ArrayField` to store the data as torch tensors, and have it automatically convert numpy input.

matt-gardner: That would also be totally fine with me; I'd just like to avoid having two almost identical classes, that only differ on whether they are passed numpy arrays or torch tensors.

dirkgr:
Closing this in favor of #4481.

PR #4448, opened 2020-07-07
Minor tqdm and logging clean up

epwalsh:
The main thing this does is add progress bar descriptions in a few places, so that when we have nested progress bars the output is a lot more clear.


epwalsh:
There will still be nested progress bars in a few places, but it looks much more readable with the description on each one. In other words, I don't think nested progress bars are actually an issue as long as you can tell which progress bar is which.

Issue #4436, opened 2020-07-02
Stop logging regularization loss of 0


dirkgr:
We now log (and show during training) the regularization loss. But in many configs, that loss is always 0. We should not show it in that case, at least not on the console.

mateuszpieniak:
Btw, @dirkgr When I add `weight_decay` to my optimizer (e.g. Adam), then do I need to use `RegularizerApplicator`? I've just discovered that I only used `weight_decay` for optimizer in my config file. Reported `reg_loss` == 0 drew my attention that perhaps I have a bug?

dirkgr:
Hmm, this is a little bit confusing. When you use the `weight_decay` option of Adam, the regularization is applied directly in the optimizer. It bypasses the loss calculation, so there is no `reg_loss` reported when you do this. This is correct and you don't need to use the applicator.

PR #4449, opened 2020-07-07
Fix regloss logging

AkshitaB:
Fixes https://github.com/allenai/allennlp/issues/4436 

reg_loss is only displayed if the model being trained is configured to have a regularization penalty.

### DIFF
--- a/allennlp/models/model.py
+++ b/allennlp/models/model.py
@@ -86,9 +86,13 @@ class Model(torch.nn.Module, Registrable):
         if self._regularizer is None:
             regularization_penalty = None
         else:
-            regularization_penalty = self._regularizer(self)
-            if isinstance(regularization_penalty, float):
-                regularization_penalty = torch.tensor(regularization_penalty)
+            try:
+                regularization_penalty = self._regularizer(self)
+                if isinstance(regularization_penalty, float):
+                    assert regularization_penalty == 0.0
+                    regularization_penalty = torch.tensor(regularization_penalty)
+            except AssertionError:
+                raise RuntimeError("The regularizer cannot be a non-zero float.")
         return regularization_penalty
 
     def get_parameters_for_histogram_tensorboard_logging(self) -> List[str]:
--- a/allennlp/models/model.py
+++ b/allennlp/models/model.py
@@ -78,13 +78,13 @@ class Model(torch.nn.Module, Registrable):
         self.vocab = vocab
         self._regularizer = regularizer
 
-    def get_regularization_penalty(self) -> Union[float, torch.Tensor]:
+    def get_regularization_penalty(self) -> Optional[Union[float, torch.Tensor]]:
         """
         Computes the regularization penalty for the model.
-        Returns 0 if the model was not configured to use regularization.
+        Returns None if the model was not configured to use regularization.
         """
         if self._regularizer is None:
-            return 0.0
+            return None
         else:
             return self._regularizer(self)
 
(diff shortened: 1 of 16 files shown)


allennlp/training/trainer.py
```diff
@@ -409,6 +409,11 @@ def __init__(
         self._tensorboard.get_batch_num_total = lambda: self._batch_num_total
         self._tensorboard.enable_activation_logging(self.model)
 
+        # Only display reg_loss if the model's configuration has regularization.
+        self._show_metrics = ["loss", "reg_loss"]
+        if self.model.get_regularization_penalty() == 0.0:
```
dirkgr, line 414: It's likely that we have to detect whether a regularization loss is configured by looking whether the number is `0`. But it would be better if we make that decision based on whether it's configured, not based on whether the result is zero. That way we can show it, for example, when it's misconfigured.

AkshitaB: I believe the `get_regularization_penalty()` function returns a 0 if the model is configured to not have regularization. This is what the docstring for that function says: `Returns 0 if the model was not configured to use regularization.`

matt-gardner:
I agree that what @dirkgr suggests sounds like a cleaner change.

allennlp/training/trainer.py
```diff
@@ -574,19 +578,20 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
             for batch in batch_group:
                 batch_outputs = self.batch_outputs(batch, for_training=True)
                 batch_group_outputs.append(batch_outputs)
-                loss = batch_outputs["loss"]
-                reg_loss = batch_outputs["reg_loss"]
+                loss = batch_outputs.get("loss")
```
dirkgr, line 581: I think `"loss"` is always there.

allennlp/models/model.py
```diff
@@ -78,15 +78,18 @@ def __init__(self, vocab: Vocabulary, regularizer: RegularizerApplicator = None)
...
+        Returns None if the model was not configured to use regularization.
         """
         if self._regularizer is None:
-            return 0.0
+            regularization_penalty = None
         else:
-            return self._regularizer(self)
+            regularization_penalty = self._regularizer(self)
+            if isinstance(regularization_penalty, float):
+                regularization_penalty = torch.tensor(regularization_penalty)
```
dirkgr, line 91: If we get a `float` from `self._regularizer`, how does that work with gradients? I don't see how regularization can work if we lose track of the computation graph here. `float` doesn't have any information about the computation graph.

I guess it works if the `float` is 0. Is that what this case is capturing?

allennlp/training/trainer.py
AkshitaB: Yes, it was more for uniformity's sake.

allennlp/models/model.py
AkshitaB: That's what I also thought. But this test case: `TestElmoTokenEmbedder.test_tagger_with_elmo_token_embedder_can_train_save_and_load` seems to require explicit conversion to tensor. The value is actually 0.0 .

dirkgr: I see. This happens in the applicator when none of the regexes match. That's probably not a great situation, since it might mean the config doesn't do what you think it does. But let's not invent a whole new problem area to solve in this PR.

In this case, let's throw an exception when the number isn't `0.0` and call it done?

PR #4453, opened 2020-07-08
fix boto3 dependency for package manager

### LINKED ISSUE #4450: Restrict the version of `boto3` for package manager
Now allennlp depends on `boto3==*`, which sometimes causes a package manager `poetry` to get stuck in dependency resolution (see https://github.com/python-poetry/poetry/issues/771#issuecomment-630677036).
Users can avoid this by specifying `boto3` dependency, but it is not trivial.

I think it's good to rewrite the below line to, for example, `boto3>=1.14`

https://github.com/allenai/allennlp/blob/5b988d637d9d2fde405eb0d0bc22b3bdb4012ad2/setup.py#L60

tamuhey:
closes #4450

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ setup(
         "spacy>=2.1.0,<2.4",
         "numpy",
         "tensorboardX>=1.2",
-        "boto3>=1.14",
+        "boto3>=1.14,<2.0",
         "requests>=2.18",
         "tqdm>=4.19",
         "h5py",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -57,7 +57,7 @@
         "spacy>=2.1.0,<2.4",
         "numpy",
         "tensorboardX>=1.2",
-        "boto3",
+        "boto3>=1.14",
```
epwalsh, line 60: At this point I think we should also pin this below `2.0`, since dependabot will automatically create a PR to update this requirement when 2.0 comes out.

```suggestion
        "boto3>=1.14,<2.0",
```

schmmd:
@tamuhey if you accept Pete's suggestion and update the CHANGELOG.md we can merge this down!

PR #4462, opened 2020-07-12
Update beam_search.py

wlhgtc:
1. support for multi-layer decoders


allennlp/nn/beam_search.py
```diff
@@ -191,18 +181,29 @@ def new_step(
...
                 continue
-            _, *last_dims = state_tensor.size()
+
             # shape: (batch_size * beam_size, *)
-            state[key] = (
-                state_tensor.unsqueeze(1)
-                .expand(batch_size, self.beam_size, *last_dims)
-                .reshape(batch_size * self.beam_size, *last_dims)
-            )
+            if state_tensor.size()[0] == batch_size:
```
dirkgr, line 190: This is going to fail if the batch size is exactly equal to the number of layers, isn't it? Is there a better way to detect this, such as the number of dimensions in `state_tensor`?

allennlp/nn/beam_search.py
```diff
@@ -293,19 +294,33 @@ def new_step(
...
-                ).expand(batch_size, self.beam_size, *last_dims)
-
-                # shape: (batch_size * beam_size, *)
-                state[key] = (
-                    state_tensor.reshape(batch_size, self.beam_size, *last_dims)
-                    .gather(1, expanded_backpointer)
-                    .reshape(batch_size * self.beam_size, *last_dims)
-                )
 
+                if state_tensor.size()[0] == batch_size * self.beam_size:
```
dirkgr, line 298: same question

allennlp/nn/beam_search.py
wlhgtc: Yes, I could use the code like `state_tensor.dim() > 3` . But we use `*last_dims` for the shape. 
I wonder are there any situation that we accept tensor has dimension more than 3 ?
 If true, I have to check whether the tensors in state have the same dimension. It's a rude way :<

wlhgtc:
@dirkgr  Sorry for the format questions, I use PyCharm to auto-format my code. I will recover it to the original version.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+
+### Added
+
+- Adjust beam search to support multi-layer decoder.
```
epwalsh, line 13: We already have an "Added" section

allennlp/nn/beam_search.py
```diff
@@ -14,11 +12,8 @@
 class BeamSearch:
     """
     Implements the beam search algorithm for decoding the most likely sequences.
-
```
epwalsh, line 17: All of the blank lines are here for a reason. The API docs won't render properly without them.

wlhgtc: Fix

allennlp/nn/beam_search.py
wlhgtc: fix

allennlp/nn/beam_search.py
wlhgtc: fix

CHANGELOG.md
wlhgtc: fix

allennlp/nn/beam_search.py
```diff
@@ -196,13 +195,23 @@ def new_step(
...
                 continue
-            _, *last_dims = state_tensor.size()
-            # shape: (batch_size * beam_size, *)
-            state[key] = (
-                state_tensor.unsqueeze(1)
-                .expand(batch_size, self.beam_size, *last_dims)
-                .reshape(batch_size * self.beam_size, *last_dims)
-            )
+
+            if state_tensor.dim() == 3 and key in ["decoder_hidden", "decoder_context"]:
```
wlhgtc, line 199: I use key name check here. 
As I mentioned: the `ComposedSeq2Seq` has `state['encoder_output'] with shape (batch, length, hidden)`.
It's hard to check whether it has the dimension of `num_of_layers`.
So the directly name check works. @dirkgr

wlhgtc:
@dirkgr  @epwalsh  I have passed all check. Hope it can be merged to the master.

allennlp/nn/beam_search.py
```diff
@@ -78,7 +79,6 @@ def search(
         """
         Given a starting state and a step function, apply beam search to find the
         most likely target sequences.
-
```
epwalsh, line 81: Why delete this line? I believe this blank line is needed for the docs to render properly.

allennlp/nn/beam_search.py
```diff
@@ -103,16 +103,15 @@ def search(
             A function that is responsible for computing the next most likely tokens,
             given the current state and the predictions from the last time step.
             The function should accept two arguments. The first being a tensor
-            of shape `(group_size,)`, representing the index of the predicted
```
epwalsh, line 106: Why change `group_size` to `batch_size`?

wlhgtc: I think it's same in conception but different in name. Unfortunately I am wrong.

allennlp/nn/beam_search.py
wlhgtc: fix

allennlp/nn/beam_search.py
wlhgtc: fix

wlhgtc:
@dirkgr @epwalsh I have fixed all problems, any other suggestions ?

allennlp/nn/beam_search.py
```diff
@@ -293,18 +304,34 @@ def new_step(
...
-                    batch_size, self.beam_size, *([1] * len(last_dims))
-                ).expand(batch_size, self.beam_size, *last_dims)
-
-                # shape: (batch_size * beam_size, *)
-                state[key] = (
-                    state_tensor.reshape(batch_size, self.beam_size, *last_dims)
-                    .gather(1, expanded_backpointer)
-                    .reshape(batch_size * self.beam_size, *last_dims)
-                )
+                if state_tensor.dim() == 3 and key in ["decoder_hidden", "decoder_context"]:
```
dirkgr, line 307:
```suggestion
                multilayer_decoder = state_tensor.dim() == 3 and key in {"decoder_hidden", "decoder_context"}
                if multilayer_decoder:
```
It's functionally the same, but it makes it a little clearer what this `if` statement does.

dirkgr: The same would work for the other place where you do this.

allennlp/nn/beam_search.py
dirkgr: @Tobias-Rohde, as the last person that has been deep inside this code, do you have an opinion on this?

dirkgr:
Also, thank you for cleaning up the formatting. The diff is now much easier to read!

allennlp/nn/beam_search.py
wlhgtc: @dirkgr  Thanks for your advice. 
Then main question is : only the beam_search need the state shape like `(batch,...)`.  
We could reshape state here or **before(and after) the decoder layer in all models** . I prefer the first one, slightly change the code rather than adjust all models' decoder.

allennlp/nn/beam_search.py
```diff
@@ -197,8 +196,12 @@ def new_step(
         for key, state_tensor in state.items():
             if state_tensor is None:
                 continue
+            multilayer_rnn_decoder = state_tensor.dim() == 3 and key in {
```
wlhgtc, line 199: @dirkgr  thanks for your advice, it's more clearly. But I use `multilayer_rnn_decoder `  rather than `multilayer_decoder `, cause the original code seems works on `Transformer`.

allennlp/nn/beam_search.py
```diff
@@ -293,18 +307,34 @@ def new_step(
...
-                    batch_size, self.beam_size, *([1] * len(last_dims))
-                ).expand(batch_size, self.beam_size, *last_dims)
-
-                # shape: (batch_size * beam_size, *)
-                state[key] = (
-                    state_tensor.reshape(batch_size, self.beam_size, *last_dims)
-                    .gather(1, expanded_backpointer)
-                    .reshape(batch_size * self.beam_size, *last_dims)
-                )
+                if multilayer_rnn_decoder:
```
epwalsh, line 310: Where does this variable `multilayer_rnn_decoder` come from? It would need to be set for each `(key, state_tensor)` pair like it is above, right? Otherwise the value of `multilayer_rnn_decoder` here will just be equal to whatever it was last assigned above.

dirkgr:
> only the beam_search need the state shape like (batch,...).
> We could reshape state here or before(and after) the decoder layer in all models . I prefer the first one, slightly change the code rather than adjust all models' decoder.

I saw your explanation in the other PR. I think you are right. It is better here.

allennlp/nn/beam_search.py
wlhgtc: Fix!

wlhgtc:
@dirkgr  @epwalsh  Any other suggestions ?

dirkgr:
I wanted to give @Tobias-Rohde a chance to say something, because he is more familiar with this area of the code. That's the only thing I'm waiting for.

allennlp/nn/beam_search.py
Tobias-Rohde: I am not sure if I like adding this decoder specific code to the beam search.

Couldn't you instead change the model to transpose the state tensors to have the num_layers dimension at the end? So for instance for the SimpleSeq2Seq model, here:

https://github.com/allenai/allennlp-models/blob/4d0e090514af273471555be08676bc02999fd0ff/allennlp_models/generation/models/simple_seq2seq.py#L415

you would get the state tensors and transpose them such that num_layers is the first dimension and then before putting them back into the state here:

https://github.com/allenai/allennlp-models/blob/4d0e090514af273471555be08676bc02999fd0ff/allennlp_models/generation/models/simple_seq2seq.py#L438

you would transpose them to have num_layers as the last dimension. That way we could leave the beam search unmodified and move the decoder specific to the model, where it belongs imo. Let me know if this would work, otherwise I think the way you implemented it is fine too.

wlhgtc: Thanks for your quickly response. 
There are two different ways to work. I have discussed with @dirkgr  in [here](https://github.com/allenai/allennlp-models/pull/90#issuecomment-659231259).
I prefer the second way( make changes in beam search rather than model decoders).
The reason is  as follows: It is the beam_search code that requires state shape like `(batch,...)`. If you implement it in decoder model, that means you should repeat same code in every (LSTM)decoder.

Tobias-Rohde: I see your point, I think both ways are fine. I'd say just keep it the way you've implemented it currently. The beam search code should be rewritten in the near future anyways, there are many very useful beam search features that huggingface and fairseq have (min length, max length, max repeated n-grams, ...) and I think we should have those too.

wlhgtc: Ok, hope for your future version~ 
@dirkgr  Can we merge it into master, then I can update my PR in `allennlp-model`

dirkgr:
@epwalsh, have your concerns been addressed?

PR #4468, opened 2020-07-13
More BART changes

dirkgr:
For some reason, that tokenizer change speeds up the BART dataset reader by 50%.


epwalsh:
> For some reason, that tokenizer change speeds up the BART dataset reader by 50%.

I'm guessing it was using a pure-Python tokenizer before

PR #4469, opened 2020-07-13
Skip ETag check if OSError

JohnGiorgi:
Based on the discussion [here](https://discourse.allennlp.org/t/use-cached-path-to-download-a-file-from-github-releases/352), attempts to fix an issue that was preventing `cached_path` from being used to download files from GitHub releases.

It looks like [ETags are optional](https://en.wikipedia.org/wiki/HTTP_ETag), so I elected to just catch the `OSError` that is thrown by `_http_etag` and set `etag = None` in this case (`url_to_filename` already supports an unspecified `etag`). This solution seems cleaner to me than asking the user to provide some flag to skip the ETag fetch.

Let me know if you are happy with this solution and then I can try to write a unit test for this scenario / update the `CHANGELOG.md`.


JohnGiorgi:
Done :)

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed a bug in `TextClassificationPredictor` so that it passes tokenized inputs to the `DatasetReader`
   in case it does not have a tokenizer.
 - reg_loss is only now returned for models that have some regularization penalty configured.
+- Fixed a bug that prevented cached_path from downloading assets from GitHub releases.
```
epwalsh, line 31:
```suggestion
- Fixed a bug that prevented `cached_path` from downloading assets from GitHub releases.
```

epwalsh:
Thank you very much!

PR #4477, opened 2020-07-14
Fix reported loss

### LINKED ISSUE #4354: Reported loss is confusing
The loss we write into tensorboard is the average loss over the epoch. So the first number we get is not averaged at all, while the last number in the epoch is averaged over the entire epoch. Then we start the second epoch, and the next value reported is just the loss of the first batch of the second epoch. As a result, we always see a big drop in loss when the epoch changes. This makes no sense.

AkshitaB:
Fixes #4354


HarshTrivedi:
It seems that the `batch_loss` is computing total loss up to that batch group update. So it starts from a small value and continuously increases throughout the epoch. Is this the intended behavior? From the name, I was expecting this to show loss of that specific batch update.

dirkgr:
Oh, I think you are right. I read that totally wrong.

@AkshitaB, can you confirm? I think we need another fix.

PR #4481, opened 2020-07-15
Detectron NLVR2

dirkgr:
TODO:

 - [x] Figure out detectron settings that produce the 2048-vectors that VilBERT needs
 - [x] Remove `DetectronField`
 - [x] Remove the model stub from the `DetectronField` experiments
 - [x] Unify `ArrayField` and `TensorField`
 - [x] Remove Matt's reader, or at least share common code between them
 - [x] Create a tiny dataset to run the test


test_fixtures/vilbert/experiment_detectron.json
```diff
@@ -0,0 +1,60 @@
+{
+  "dataset_reader": {
+    "type": "nlvr2",
+    "image_dir": "/Users/dirkg/Documents/data/vision/nlvr2/images",
```
jiasenlu, line 4: Is it possible to add those images to the github? and relative locations?

dirkgr: The way I have it right now, that's 11GB of images. I'll add a subset to GitHub so we can run the test.

dirkgr:
> @dirkgr For Detectron dataloader, do we want to return image tensor? I think we should return detectron input format.

dirkgr:
Sorry, I closed this by accident. Wanted to close the window, not the whole PR ...

allennlp/common/detectron.py
```diff
@@ -0,0 +1,111 @@
...
+def get_detectron_cfg(
+    builtin_config_file: Optional[str] = None,
+    yaml_config_file: Optional[str] = None,
+    overrides: Optional[Dict[str, Any]] = None,
+    freeze: bool = True,
+) -> CfgNode:
+    from detectron2.config import get_cfg
+
+    cfg = get_cfg()
+    # add_attribute_config(cfg)  # TODO: What is this? It's not general, not configurable.
```
dirkgr, line 53:
```suggestion
```

allennlp/data/image_loader.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def load(self, filenames: ManyPaths) -> List[FloatTensor]:
+        raise NotImplementedError()
+
+
+@ImageLoader.register("detectron")
+class DetectronImageLoader(ImageLoader):
+    def __init__(
+        self,
+        pixel_mean: List[float] = [103.530, 116.280, 123.675],
```
dirkgr, line 38: Where do these numbers come from?

allennlp/data/image_loader.py
```diff
@@ -37,20 +35,51 @@ def load(self, filenames: ManyPaths) -> List[FloatTensor]:
...
+        }
+        from allennlp.common.detectron import get_detectron_cfg
+        cfg = get_detectron_cfg(None, None, overrides)
         from detectron2.data import DatasetMapper
-
         self.mapper = DatasetMapper(cfg)
 
     def load(self, filenames: ManyPaths) -> List[FloatTensor]:
         images = [{"file_name": str(f)} for f in filenames]
+        # TODO: is this efficient enough, the detectron2 impelemntation ?
```
dirkgr, line 80: If it's not, do we have an alternative? I am under the impression that the pre-processing takes no time at all when compared to proposal generation and embedding boxes.

allennlp/modules/vision/proposal_generator.py
```diff
@@ -17,9 +28,300 @@ def forward(self, images: FloatTensor):
...
+    """
+    Download and cache a tarfile containing detectron checkpoint (_weights.th)
+    It might be better to replace this with native utility functions like load_archive (https://github.com/allenai/allennlp/blob/master/allennlp/models/archival.py#L132) later,
+    where the tarfile includes 1) config.json 2) _weights.th
+    """
+    gs_url = "https://storage.googleapis.com/allennlp-public-models"
+    detectron_checkpoint_url = f'{gs_url}/{checkpoint_name}.tar.gz'
+    extraction_path = cached_path(detectron_checkpoint_url, extract_archive=True)
+
+    folder_name = detectron_checkpoint_url.split('/')[-1].split('.')[0]  # ex) RCNN-X152-C4-2020-07-18
```
dirkgr, line 53: Isn't that just `checkpoint_name` again?

allennlp/modules/vision/proposal_generator.py
```diff
@@ -17,9 +28,300 @@ def forward(self, images: FloatTensor):
...
+    weights_path = f'{extraction_path}/{folder_name}/_weights.th'
```
dirkgr, line 54: You can get the path to a file inside of an extracted archive from `cached_path()`. For example:
```Python
cached_path("http://foo.com/bar.tar.gz!file/inside/archive.txt", extract_archive=True)
```

allennlp/common/detectron.py
```diff
@@ -9,13 +9,36 @@ class DetectronConfig(NamedTuple):
     overrides: Optional[Dict[str, Any]] = None
 
 
-def apply_dict_to_cfg(d: Dict[str, Any], c: CfgNode) -> None:
-    for key, value in d.items():
-        key = key.upper()
-        if isinstance(value, dict):
-            apply_dict_to_cfg(value, c[key])
+# def apply_dict_to_cfg(d: Dict[str, Any], c: CfgNode) -> None:
```
dirkgr, line 12: Why did you change this? The new code is also fine, no problem. I'm just wondering if I missed something when I wrote `apply_dict_to_cfg()`.

allennlp/data/image_loader.py
```diff
@@ -37,20 +35,51 @@ def load(self, filenames: ManyPaths) -> List[FloatTensor]:
...
         self.mapper = DatasetMapper(cfg)
 
     def load(self, filenames: ManyPaths) -> List[FloatTensor]:
         images = [{"file_name": str(f)} for f in filenames]
+        # TODO: is this efficient enough, the detectron2 impelemntation ? 
         images = [self.mapper(i) for i in images]
-        images = [i["image"] for i in images]
+
+        # TODO: for detectronImage loader, do we want to return image tensor or orginial detectron dict format?  
+        # images = [i["image"] for i in images]
```
dirkgr, line 84: I wanted to return tensors, so the output of this can be fed straight into `Image2ImageModule` and `ProposalGenerator`. Also, any other implementation of `ImageLoader` would have to return images in Detectron format.

Why would you want to change it? One thing that we don't handle well here is the extra fields for size that Detectron keeps in the dict.

allennlp/common/detectron.py
j-min: There are some attributes (ex. cfg.MODEL.ROI_ATTRIBUTE_HEAD) that are not in the original detectron2 CfgNode. Accessing those keys with c[key] causes KeyError.

allennlp/modules/vision/proposal_generator.py
j-min: You're right. I'll replace `folder name` with `checkpoint_name` later.

allennlp/modules/vision/proposal_generator.py
j-min: Oh I didn't know that.
This looks much better. I'll re-write in this way.

allennlp/common/detectron.py
dirkgr: Sure, but those new attributes could have been added with `applt_dict_to_cfg()`, right?

j-min: You might be right, I just couldn't find a simple way how to handle nested CfgNode classes.
CfgNode sometimes has CfgNodes as an attribute (see https://github.com/allenai/allennlp/pull/4481/files#diff-b2b05f69be08bf21366facbbe9abfc5dR94).

allennlp/common/detectron.py
```diff
@@ -269,14 +285,245 @@ def get_pipeline_from_flat_parameters(
...
-            }
+            },
         },
+        "TEST": {"DETECTIONS_PER_IMAGE": fp.test_detections_per_image},
     }
 
     return get_pipeline(None, None, overrides, make_copy)
+
+
+class AttributePredictor(nn.Module):
```
dirkgr, line 301: @jiasenlu, these classes are some extra stuff that we wish was part of Detectron, but isn't, correct? It has nothing directly to do with AllenNLP?

allennlp/modules/vision/region_detector.py
```diff
@@ -0,0 +1,137 @@
...
+        proposal_start = 0
+        for image_num, image_proposal_indices in enumerate(r_indices):
+            feature_cell = {}
+            feature_cell['boxes'] = proposals[image_num].proposal_boxes.tensor[image_proposal_indices]
+            feature_cell['probs'] = cls_probs[image_num][image_proposal_indices]
+            feature_cell['features'] = torch.narrow(pooled_features, 0, proposal_start, len(proposals[image_num]))[image_proposal_indices]
+            proposal_start += len(proposals[image_num])
+            features.append(feature_cell)
+            
+        return features
```
dirkgr, line 137: The output is supposed to be a `Dict[str, Tensor]`, but this output is a `List`.

dirkgr: I guess that's more of a question for @jiasenlu. What happened here?

allennlp/data/dataset_readers/nlvr2.py
```diff
@@ -29,146 +47,120 @@ def __init__(
...
+        right_image_path = self.images[f"{image_name_base}-img1.png"]
+
+        to_compute = []
+        if left_image_path not in self._feature_cache:
+            to_compute.append(left_image_path)
+        if right_image_path not in self._feature_cache:
+            to_compute.append(right_image_path)
+
+        if to_compute:
+            images, sizes = self.image_loader([left_image_path, right_image_path])
```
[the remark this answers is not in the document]
matt-gardner: Thanks for the catch.

allennlp/data/dataset_readers/nlvr2.py
```diff
@@ -29,146 +47,120 @@ def __init__(
...
+        left_image_path = self.images[f"{image_name_base}-img0.png"]
+        right_image_path = self.images[f"{image_name_base}-img1.png"]
+
+        to_compute = []
+        if left_image_path not in self._feature_cache:
+            to_compute.append(left_image_path)
+        if right_image_path not in self._feature_cache:
+            to_compute.append(right_image_path)
+
+        if to_compute:
```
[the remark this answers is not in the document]
matt-gardner: I just checked PEP8 (the standard python style guide) and it doesn't give an opinion on this point, though it does have a warning to be careful when doing this.  This is such a common usage in python that I expected to see something, and was surprised when it wasn't there.  Because it's so common, and because the list was created 6 lines above, I'm not too concerned with this particular case, but yes, I agree that avoid this except in really simple cases like this is probably a good idea.

allennlp/data/fields/array_field.py
```diff
@@ -62,5 +62,9 @@ def __str__(self) -> str:
     def __len__(self):
         return 1 if len(self.tensor.size()) <= 0 else self.tensor.size(0)
 
-    def __eq__(self, other: "ArrayField") -> bool:
-        return torch.equal(self.tensor, other.tensor)
+    def __eq__(self, other) -> bool:
+        if isinstance(self, other.__class__):
+            return (
+                torch.equal(self.tensor, other.tensor) and self.padding_value == other.padding_value
+            )
```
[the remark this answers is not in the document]
matt-gardner: Yeah, this is what that other code should have been.  If existing code relies on a bug, we should still fix the bug.  But I really doubt that anything was relying on that bug.

allennlp/data/image_loader.py
```diff
@@ -1,55 +1,67 @@
...
 
     def load(self, filenames: ManyPaths) -> ImagesWithSize:
         raise NotImplementedError()
 
 
+# These are in strings to avoid a costly / perhaps uninstalled import of detectron.  This typedef is
+# not inplace because it makes the line too long to include flake and mypy ignore statements inline.
+# TODO(mattg): Maybe better is to make some vision/ directory that doesn't get imported anywhere by
+# default, but allows for putting the detectron imports at the top of the file, because this isn't
+# the only place we have this issue.
```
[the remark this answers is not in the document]
matt-gardner: I'm pretty strongly against local imports, except in exceptional circumstances that involve heavy imports, like detectron or nltk or some optional dependency.  [PEP8 agrees](https://www.python.org/dev/peps/pep-0008/#imports), and that's the style guide that we follow.

I don't have a strong opinion on whether we should accept detectron as a required dependency of the library.  It was a pain for me to install on my mac, and it might be nice to not require users to install it.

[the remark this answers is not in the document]
matt-gardner: Oh, actually, I don't think we _can_ assume that it's always installed, because I don't think it's on pypi.  Maybe they'll change that someday?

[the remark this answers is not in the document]
matt-gardner: I also don't think we currently have a problem with circular imports.  Where are you encountering this problem?

allennlp/common/detectron.py
jiasenlu: yes. It has nothing directly to do with AllenNLP, but current detectron2 feature extraction model requires   "AttributeRes5ROIHeads", which we register this class blow. 

https://github.com/allenai/allennlp/pull/4481/files/8515be27a00aa1aebd3363630c89ad8c3d364535..5db6805b50fb1f0197c82daea2469687d5dc248a#diff-b2b05f69be08bf21366facbbe9abfc5dR174

allennlp/modules/vision/grid_embedder.py
```diff
@@ -23,13 +24,55 @@ def get_output_dim(self) -> int:
...
+    def get_stride(self) -> int:
+        """
+        Returns the overall stride of this `GridEmbedder`, which, when combined with the input image
+        size, will give you the height and width of the output grid.
+        """
+        raise NotImplementedError
+
 
-@GridEmbedder.register("null")
-class NullGridEmbedder(GridEmbedder):
```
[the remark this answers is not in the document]
matt-gardner: I thought they were just there to satisfy the bug where you have no concrete implementations of a registrable class.  I have no objection to having these available, other than I thought they were now useless.

allennlp/common/detectron.py
jiasenlu: yes. It has nothing directly to do with AllenNLP, but current detectron2 feature extraction model requires "AttributeRes5ROIHeads", which we register this class blow.

https://github.com/allenai/allennlp/pull/4481/files/8515be27a00aa1aebd3363630c89ad8c3d364535..5db6805b50fb1f0197c82daea2469687d5dc248a#diff-b2b05f69be08bf21366facbbe9abfc5dR174

allennlp/data/image_loader.py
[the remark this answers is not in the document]
jiasenlu: yes. I agree.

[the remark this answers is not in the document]
matt-gardner: Ok, I see the issue with circular imports, and it's because `dataset_readers` is inside of `data`.  The intended dependency graph goes something like `common -> data -> modules -> model -> training`.  Having the NLVR2 reader depend on a module is the cause of the circular dependency.  This would go away if we either (1) move the model and dataset reader into the models repo (which will probably happen eventually), or (2) move `dataset_readers` out of `data` and parallel to `model` in the dependency graph (which is a large breaking change that's unlikely to happen).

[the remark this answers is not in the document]
dirkgr: If there is any time it could happen, it's for a 2.0 release.

allennlp/common/file_utils.py
```diff
@@ -25,6 +27,7 @@
 from requests.adapters import HTTPAdapter
 from requests.exceptions import ConnectionError
 from requests.packages.urllib3.util.retry import Retry
+import lmdb # installtest_fixtures/vilbert/experiment_from_huggingface_vqa.jsonnet lmdb by "pip install lmdb"
```
dirkgr, line 30: Copy and paste error?

allennlp/common/file_utils.py
```diff
@@ -260,7 +263,7 @@ def _s3_get(url: str, temp_file: IO) -> None:
     """Pull a file directly from S3."""
     s3_resource = _get_s3_resource()
     bucket_name, s3_path = _split_s3_path(url)
-    s3_resource.Bucket(bucket_name).download_fileobj(s3_path, temp_file)
+    s3_resource.Bucket(bucketest_fixtures/vilbert/experiment_from_huggingface_vqa.jsonnett_name).download_fileobj(s3_path, temp_file)
```
dirkgr, line 266: ?

test_fixtures/vilbert/experiment_from_huggingface_vqa.jsonnet
```diff
@@ -3,16 +3,15 @@ local model_name = "bert-base-uncased";
   "dataset_reader": {
     "type": "vqav2",
     "lazy": true,
-    "image_dir": "/net/nfs2.corp/prior/datasets/coco/",
+    "image_dir": "/home/jiasenl/data/coco/",
+    "lmdb_cache_dir": "/home/jiasenl/data/coco.lmdb",
```
dirkgr, line 7: Please don't check your local paths into the shared repo.

allennlp/training/trainer.py
```diff
@@ -584,21 +584,22 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
...
-                    loss = loss / len(batch_group)
-
-                    batch_loss = loss.item()
-                    train_loss += batch_loss
-                    if reg_loss is not None:
-                        reg_loss = reg_loss / len(batch_group)
-                        batch_reg_loss = reg_loss.item()
-                        train_reg_loss += batch_reg_loss
+                # TODO(mattg): vision stuff is incompatible with 1.6, removing autocast for now.
+                # THIS SHOULD NOT BE COMMITTED TO MASTER!
```
dirkgr, line 588: @matt-gardner, what is this? The detectron modules fail when we use `amp.autocast`?

allennlp/data/dataset_readers/nlvr2.py
```diff
@@ -0,0 +1,200 @@
...
+from allennlp.data.token_indexers import TokenIndexer
+from allennlp.data.tokenizers import PretrainedTransformerTokenizer
+from allennlp.data.tokenizers import Tokenizer
+from allennlp.modules.vision.grid_embedder import GridEmbedder
+from allennlp.modules.vision.region_detector import RegionDetector
+
+
+@DatasetReader.register("nlvr2")
+class Nlvr2Reader(DatasetReader):
+    """
```
matt-gardner, line 25: This needs a description in the docstring.

allennlp/common/file_utils.py
```diff
@@ -339,6 +359,92 @@ def _find_latest_cached(url: str, cache_dir: Union[str, Path]) -> Optional[str]:
     return None
 
 
+def _serialize(data):
+    buffer = pickle.dumps(data, protocol=-1)
+    return np.frombuffer(buffer, dtype=np.uint8)
+
+
+class TensorCache(MutableMapping[str, Tensor], ABC):
```
matt-gardner, line 367: This class should have a docstring.

allennlp/data/fields/array_field.py
```diff
@@ -1,68 +1,78 @@
...
-import numpy
 import torch
+import numpy as np
 from overrides import overrides
 
 from allennlp.data.fields.field import Field
 
 
-class ArrayField(Field[numpy.ndarray]):
+class ArrayField(Field[torch.Tensor]):
```
matt-gardner, line 10: Given the changes you made, I'd be tempted to rename this and add something like `ArrayField = TensorField` for backwards compatibility.

allennlp/modules/vision/image2image.py
```diff
@@ -0,0 +1,39 @@
+import torch
+from torch import nn, FloatTensor, IntTensor
+from typing import List
+
+from allennlp.common.registrable import Registrable
+
+
+class Image2ImageModule(nn.Module, Registrable):
```
matt-gardner, line 8: I don't remember - do we actually use this anywhere?  Is this an abstraction that we want to keep, or is it an old one that we're happy to get rid of now?

matt-gardner (approved):

matt-gardner:
I think coverage is having a fit because this is on a non-master branch?  But CI is otherwise passing, and the model tests run in a few seconds.  The reader test takes a minute for some reason, perhaps because of some detectron initialization / caching, or something?  All we are doing in the reader test is creating a `DetectronImageLoader`, the actual processing is done with fake objects.  When I run the test locally it takes a few seconds.  So I'm guessing it's something that's cached on my local machine but not in the docker CI environment.  Overall, though, this PR still only adds a minute to CI.  I'd say it's good to merge when you're happy with it @dirkgr.

PR #4488, opened 2020-07-16
use longer tqdm intervals when output being redirected

epwalsh:


epwalsh:
> I broke this with the `--file-friendly-logging` stuff, didn't I. Did we believe at the time `tqdm` would do this automatically?

I think sometimes it does it automatically, but there are scenarios where `sys.stderr.isatty()` returns `False`, resulting in us replacing carriage returns with new lines, but `Tqdm` still prints in rapid intervals.

731935354:
Hi, I'm also wondering if I can slow down tqdm update frequency in jupyter notebook (when I build `trainer` in python code and run `trainer.train()`). I'm asking this because sometimes tqdm prints every line instead of updating the same line, in this case my browser crashes.

dirkgr:
@731935354, the best thing I can think of is that you could monkey-patch our class `Tqdm` to set a larger default interval. If you have a better idea though, including a way to use the notebook-specific implementations of `tqdm`, please submit a PR! That would be a great addition to the library.

epwalsh:
> Hi, I'm also wondering if I can slow down tqdm update frequency in jupyter notebook (when I build `trainer` in python code and run `trainer.train()`). I'm asking this because sometimes tqdm prints every line instead of updating the same line, in this case my browser crashes.

@731935354 if there's a way to detect that we're running in a notebook we could handle this automatically. A PR would be greatly appreciated here

PR #4495, opened 2020-07-20
Initializing a VilBERT model from a pre-trained transformer

matt-gardner:
Still a bunch of work left to do here:
- [x] I need to make sure I can load a range of different models and get the weight sizes to work out correctly (I may need to have vision sizes depend on text sizes, e.g.)
- [x] I need to actually initialize weights for the various components.  I'm not sure yet how hard this will be; hopefully it won't end up being a huge hack.  I might need to make a bunch of assumptions about the structure of the loaded model.

### DIFF
--- a/allennlp/data/dataset_readers/nlvr2_reader.py
+++ b/allennlp/data/dataset_readers/nlvr2_reader.py
@@ -8,7 +8,6 @@ import time
 
 from overrides import overrides
 import numpy as np
-import spacy
 
 from allennlp.data.token_indexers.token_indexer import TokenIndexer
 from allennlp.data.dataset_readers.dataset_reader import DatasetReader
@@ -84,20 +83,18 @@ class Nlvr2LxmertReader(DatasetReader):
     """
     Parameters
     ----------
-    text_path_prefix: ``str``
+    text_path_prefix: `str`
         Path to folder containing text files for each dataset split. These files contain
         the sentences and metadata for each task instance.
-    visual_path_prefix: ``str``
+    visual_path_prefix: `str`
         Path to folder containing `tsv` files with the extracted objects and visual
         features
-    topk_images: ``int``, optional (default=-1)
+    topk_images: `int`, optional (default=-1)
         Number of images to load from each split's visual features file. If -1, all
         images are loaded
-    mask_prepositions_verbs: ``bool``, optional (default=False)
-        Whether to mask prepositions and verbs in each sentence
-    drop_prepositions_verbs: ``bool``, optional (default=False)
-        Whether to drop (remove without replacement) prepositions and verbs in each sentence
-    lazy : ``bool``, optional
+    tokenizer: `Tokenizer`, optional
+    token_indexers: `Dict[str, TokenIndexer]`
+    lazy : `bool`, optional
         Whether to load data lazily.  Passed to super class.
     """
 
@@ -108,8 +105,6 @@ class Nlvr2LxmertReader(DatasetReader):
         topk_images: int = -1,
         tokenizer: Tokenizer = None,
         token_indexers: Dict[str, TokenIndexer] = None,
-        mask_prepositions_verbs: bool = False,
-        drop_prepositions_verbs: bool = False,
         lazy: bool = False,
     ) -> None:
         super().__init__(lazy)
@@ -122,11 +117,7 @@ class Nlvr2LxmertReader(DatasetReader):
             token_indexers = {"tokens": PretrainedTransformerIndexer("bert-base-uncased")}
         self._token_indexers = token_indexers
         self.topk_images = topk_images
-        self.mask_prepositions_verbs = mask_prepositions_verbs
-        self.drop_prepositions_verbs = drop_prepositions_verbs
         self.image_data: Dict[str, Dict[str, Any]] = {}
-        # Loading Spacy to find prepositions and verbs
-        self.spacy = spacy.load("en_core_web_sm")
 
     def get_all_grouped_instances(self, split: str):
         text_file_path = os.path.join(self.text_path_prefix, split + ".json")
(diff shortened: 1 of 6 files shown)


matt-gardner:
Using albert as the base model passes tests.  My current hypothesis for zero gradients with BERT is that after a random initialization basically all activations get zeroed out by relus, so the model can't compute gradients.  I'm going to try getting the weights to load correctly from the pre-trained model, then see if that zero gradients error goes away.

matt-gardner:
Initializing the weights from the pretrained model instead of default initializations worked, and the `"bert-base-uncased"` test now passes.  This is ready for review.

matt-gardner:
Test error in CI seems unrelated; not sure if it's flaky or not.

matt-gardner:
The test that's failing here passes in my local environment (and it's totally unrelated to this PR).  Maybe there's a new version of some dependency that's making the test fail...?

dirkgr:
I'm hoping that test will succeed with an update from `master`.

allennlp/data/dataset_readers/nlvr2_reader.py
```diff
@@ -106,17 +106,21 @@ def __init__(
         text_path_prefix: str,
         visual_path_prefix: str,
         topk_images: int = -1,
+        tokenizer: Tokenizer = None,
+        token_indexers: Dict[str, TokenIndexer] = None,
```
dirkgr, line 110: If the only correct choices here are of the `PretrainedTransformer*` variety, and they have to match, then I think this should only take `model_name: str = "bert-base-uncased"` parameter.

allennlp/models/vilbert.py
```diff
@@ -714,6 +789,95 @@ def __init__(
...
+        # to do if it fails.  We should probably pull this out into a central "transformers_util"
+        # module, or something, and just have a method that pulls out an initialized embedding layer
+        # from a huggingface model.
+        text_embeddings = deepcopy(transformer.embeddings)
+        if hasattr(transformer.config, "embedding_size"):
+            config = transformer.config
+
+            from transformers.modeling_albert import AlbertModel
+
+            if isinstance(transformer, AlbertModel):
```
dirkgr, line 827: The comment above says this is just for BERT, but here you have Albert?

allennlp/models/vilbert.py
```diff
@@ -714,6 +789,95 @@ def __init__(
...
+            # because Sequential doesn't accept *inputs, only a single argument.
+
+            class EmbeddingsShim(torch.nn.Module):
+                def __init__(self, embeddings: torch.nn.Module, linear_transform: torch.nn.Module):
+                    super().__init__()
+                    self.linear_transform = linear_transform
+                    self.embeddings = embeddings
+
+                def forward(self, *inputs, **kwargs):
+                    return self.linear_transform(self.embeddings(*inputs, **kwargs))
```
dirkgr, line 846: These embeddings are just a linear transform on top of existing pre-trained embeddings? What's the point of that?

Are the embeddings shared in some other place, and that's why? If these embedding weights are supposed to be tied to some other weights somewhere, I'm pretty sure that the `deepcopy` above breaks that.

dirkgr:
> I'm hoping that test will succeed with an update from `master`.

Looks like that worked!

allennlp/data/dataset_readers/nlvr2_reader.py
matt-gardner: Yeah, I thought about that, though I think there's only one place that we actually require this, and it's in logic that only exists because of a small research project that Sanjay did.  I'm not sure we need to keep that logic in this version of the class.

To be more concrete: a `PretrainedTransformerTokenizer` is only required if you use `mask_prepositions_verbs`.  Instead of restricting the types to only be `PretrainedTransformer*` objects here, I'd instead either:
1. Remove the `{drop,mask}_prepositions_verbs` functionality from this reference implementation as not needed
2. Raise an error if you don't have the right tokenizer when using `mask_prepositions_verbs`
3. Come up with a workaround for how to mask prepositions and verbs when you're not using a pretrained transformer (though this probably doesn't make much sense).

I put those in my order of preference.  I'm not feeling too well today, so I'm not going to be working on this, but I'll do option 1 the next time I write code, unless anyone has objections.

allennlp/models/vilbert.py
matt-gardner: Thanks, yep, I need to update that comment.

allennlp/models/vilbert.py
matt-gardner: The pipeline in a transformer is essentially:

```
hidden_states = self.embeddings(inputs)
for layer in self.layers:
    hidden_states = layer(hidden_states)
```

For this logic to work on its own, `embedding_dim` must equal `hidden_dim`, and this is what BERT (and most other transformers) do.  Albert decides to have a much smaller `embedding_dim`, with a linear transform after the embedding to get the hidden states up to the right size:

```
hidden_states = self.linear_transform(self.embeddings(inputs))
for layer in self.layers:
    hidden_states = layer(hidden_states)
```

This reduces the number of parameters in your model: embedding matrix goes from `(vocab_size, 768)` to `(vocab_size, 128)`, while you add a `(128, 768)` linear transform matrix.

There is nothing that's supposed to be tied anywhere; I used `deepcopy` here to get a new copy of these existing weight matrices (both the embeddings and the linear transform), to initialize the weights in our model correctly.  Duplicating the modules themselves is much easier than aligning all of the underlying `Parameters` and calling `.copy_()`.

matt-gardner: Hmm, it's possible there is an activation in between the embeddings and the transform in albert; I should double check that.

dirkgr: Makes sense :+1:

allennlp/data/dataset_readers/nlvr2_reader.py
dirkgr: I agree that this functionality might not be needed in this reference implementation, but that's orthogonal to how the tokenizer and indexer are specified, isn't it?

matt-gardner: Yes, but it's that functionality that is the only thing placing constraints on this.

Oh, unless you're saying that the model requires the use of a pretrained transformer?  Well, vilbert does, but you could use this dataset reader with other models that don't make that assumption.  There's nothing in the reader itself that requires that use, so I was keeping as much flexibility in the types as possible.

allennlp/models/vilbert.py
matt-gardner: I just checked, and there is no activation in between the embedding and the hidden layer in albert, so this logic is correct.

matt-gardner:
@dirkgr, this should be ready for another look.

allennlp/models/vilbert.py
```diff
@@ -814,11 +814,17 @@ def from_huggingface_model_name(
...
+        # not sure it works for all of them, or what to do if it fails.
+        # We should probably pull everything up until the instantiation of the image feature
+        # embedding out into a central "transformers_util" module, or something, and just have a
+        # method that pulls an initialized embedding layer out of a huggingface model.  One place
+        # for this somewhat hacky code to live, instead of having to duplicate it in various models.
         text_embeddings = deepcopy(transformer.embeddings)
+
+        # Albert (and maybe others?) has this "embedding_size", that's different from "hidden_size".
+        # To get them to the same dimensionality, it uses a linear transform after the embedding
+        # layer, which we need to pull out and copy here.
```
dirkgr, line 827: These seem like excellent concerns to handle in @AkshitaB's transformer toolkit? We could have a function that says "give me the embedding layer from "roberta-large" or "albert-v2", and it returns a `torch.nn.Module` that does the right thing?

matt-gardner: Yes, agreed, that was the point of the comment starting on line 819.

PR #4497, opened 2020-07-20
new data loading

epwalsh:
Whoa, I know this is a massive PR and I apologize in advance for that. Let me try to help the review process by showing where I would start:

1. First, read the rest of this comment to get a high-level overview of what's changed and why.
2. Then, take a look at [the shiny new `DataLoader`](https://github.com/epwalsh/allennlp/blob/data-loading/allennlp/data/data_loaders/multi_process_data_loader.py) along with [its tests](https://github.com/epwalsh/allennlp/blob/data-loading/tests/data/data_loaders/multi_process_data_loader_test.py) to see it in action,
3. ... and the simplified [`DatasetReader`](https://github.com/epwalsh/allennlp/blob/data-loading/allennlp/data/dataset_readers/dataset_reader.py).
4. And don't worry, because you can still use the [`PyTorch` DataLoader](https://github.com/epwalsh/allennlp/blob/data-loading/allennlp/data/data_loaders/pytorch_data_loader.py).

Most of the other file changes only involved a couple of lines each (like removing the `lazy` option from `DatasetReader` tests), with the exception of the [`samplers`](https://github.com/epwalsh/allennlp/tree/data-loading/allennlp/data/samplers) module, which involves some more substantial changes since I had to do a little finagling to make the existing `BatchSamplers` work with both the `PyTorchDataLoader` and the new `MulitProcessDataLoader`.

Ook, phew... deep breath...

## Overview

- Provides a new high-performance default data loader, [`MultiProcessDataLoader`](https://github.com/epwalsh/allennlp/blob/data-loading/allennlp/data/data_loaders/multi_process_data_loader.py).

- Adds an [`iter_instances()`](https://github.com/epwalsh/allennlp/blob/data-loading/allennlp/data/data_loaders/data_loader.py) method to the [`DataLoader`](https://github.com/epwalsh/allennlp/blob/data-loading/allennlp/data/data_loaders/data_loader.py) API so that data loaders can also be used to load instances, like, for example, when you're [creating a vocabulary](https://github.com/epwalsh/allennlp/blob/data-loading/allennlp/data/vocabulary.py#L265).

- Refactors the [`DatasetReader`](https://github.com/epwalsh/allennlp/blob/data-loading/allennlp/data/dataset_readers/dataset_reader.py) class and the [`data.samplers`](https://github.com/epwalsh/allennlp/tree/data-loading/allennlp/data/samplers) module to work with both the new data loader as well as the existing [`PyTorchDataLoader`](https://github.com/epwalsh/allennlp/blob/data-loading/allennlp/data/data_loaders/pytorch_data_loader.py).

## Benefits of the new data loader

- Can leverage multi-process loading in both the lazy and non-lazy context, and for both generating batches (like how we usually use data loaders) and generating instances.

- Can utilize batch samplers with lazy loading.

## Breaking changes

- **Dataset readers are always lazy**, so there is no longer a `lazy` parameter in the `DatasetReader` class. Instead, the `DataLoader` decides to load lazily or not.

    Why do this? From an API perspective, the decision whether to load data lazily or non-lazily seems like it should belong the `DataLoader`. And this makes even more sense when you realize that there are certain `DataLoader` options that depend on whether or not you're loading data lazily. So, if the `DataLoader` didn't get to decide if it wanted to be lazy, if would end up having to check with the `DatasetReader` to see if it can be lazy, anyway. But the `DatasetReader` really doesn't care if it's lazy or not. It just wants to give you `Instance`s, and if it can do so while being lazy all the time, great.

- I stripped out the current caching mechanism (❗ **temporarily** ❗ ) in the `DatasetReader` class since the behavior depended on the `lazy` parameter, which I took out. If we still want to have a caching mechanism we need to decide how it will work within this new API where `DatasetReader`s are inherently lazy.

    Here's a couple of ideas:

    *a.* Just let users do their own caching. Our caching current isn't great anyways. *Especially* for lazy readers. In fact, it's so slow for lazy readers that not using caching is probably almost always faster.

    *b.* Have the `DataLoader` handle caching in a similar way to how `DatasetReader` currently does.

    *c.* Or just try to keep the caching logic in the `DatasetReader` class. Let the user decide to cache lazily or non-lazily.

- The `dataloader` module moved to `data_loaders`.

- The class names of many samplers have changed.

## In the wild

Training TransformerQA with the new `MultiProcessDataLoader` vs the `PyTorchDataLoader`: https://beaker.org/ws/ai2/allennlp-new-data-loading/experiments?archived=false&sort=created:ascending

These benchmarks show the performance of the new `MultiProcessDataLoader` vs the current `PyTorchDataLoader`. They were all ran through Beaker on the AI2 on prem server2. Each benchmark is just a small training job for the TransformerQA model on 5000 instances from SQUAD v1.1 using a batch size of 8. The only thing that varies between the experiments is certain data loader settings.

### Caveats

Every experiment uses the same `BucketBatchSampler` except for the lazy PyTorch data loader experiments, since batch samplers are incompatible with PyTorch `IterableDataset`s.

#### Non lazy loading

First of all, it's worth noting that there is really no benefit of using `num_workers > 0` with the PyTorch data loader in non-lazy mode, and the overhead actually decreases performance (compare [this experiment](https://beaker.org/ex/ex_jllcbifudjc2) with 0 workers to [this experiment](https://beaker.org/ex/ex_6e649yj1afxw) with 4 workers).

Using the new data loader with multiple workers though improves the overall training time. Compare [this experiment](https://beaker.org/ex/ex_req1qxcdczyd) to the above two, for example.

#### Lazy loading

The performance of the new data loader in lazy mode is much better than the PyTorch data loader using the same number of workers, even though it's actually doing more work because it uses a `BatchSampler` when the PyTorch loader does not. Compare [this experiment](https://beaker.org/ex/ex_iag4j4fvwd4d/tasks/tk_qafff3o0hb3c) with the new loader to [this experiment](https://beaker.org/ex/ex_o4gqjjwvu53d/tasks/tk_00qj9a3jd86p) with the PyTorch loader.

## Still TODO

- [x] Add more documentation to the new `DataLoader` and the new `apply_token_indexers` method on the `DatasetReader`.
- [x] Make sure `drop_last` is handled correctly.


dirkgr:
> Training TransformerQA with the new DataLoader: https://beaker.org/ex/ex_iqs9pku30v78/tasks/tk_egxfk0upy0n9 (6 workers, non-lazy), https://beaker.org/ex/ex_2alm2yevgoin/tasks/tk_p388cntba3n6 (same but with lazy=True and max_batches_in_memory=100).

The lazy one has significantly worse results. But 100 instances is not a lot to keep in memory either. A long time ago when I wrote something similar, I think I kept 1024 instances? Or even 16384?

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -0,0 +1,304 @@
...
+            worker.join(1)
+            if worker.is_alive():
+                logger.warning("Worker is still alive, killing now")
+                worker.terminate()
+
+    def _gather_instances(self, queue: mp.JoinableQueue) -> Iterable[Instance]:
+        done_count: int = 0
+        while done_count < self.num_workers:
+            instances_chunk: Optional[List[Instance]]
+            for instances_chunk in iter(queue.get, None):
```
dirkgr, line 170: Woha. I have never seen this use of `iter()` before.

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -0,0 +1,304 @@
...
+    def _instance_worker(self, worker_id: int, queue: mp.JoinableQueue) -> None:
+        self.reader._set_worker_info(WorkerInfo(self.num_workers, worker_id))
+
+        def index_fields(instance: Instance) -> Instance:
+            self.reader.apply_token_indexers(instance)
+            if self._vocab is not None:
+                instance.index_fields(self._vocab)
+            return instance
+
+        instances = (index_fields(instance) for instance in self.reader.read(self.data_path))
```
dirkgr, line 186: This needs error handling, in case `read()` throws something. If you want to peek at how I did it in bettermap, the code is here: https://github.com/allenai/bettermap/blob/master/bettermap.py#L84

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -0,0 +1,304 @@
...
+        instance_chunks: Iterable[List[Instance]]
+        if self.max_batches_in_memory is not None:
+            chunk_size = self.batch_size * self.max_batches_in_memory
+            instance_chunks = lazy_groups_of(instance_iterator, chunk_size)
+        else:
+            instance_chunks = [list(instance_iterator)]
+
+        for instances in instance_chunks:
+            if self.shuffle:
+                random.shuffle(instances)
```
dirkgr, line 207: This is better than nothing, but it would be even better if we kept instances in a pool and then picked random subsets from that pool to make batches from. The way we have it now, early instances will never be shuffled together with late instances, but with a pool, that's an option.

It also occurs to me that the order of instances is now _always_ random, because it depends on the processing speed of the workers. In light of that, maybe we should set `self.shuffle` to `True` all the time and make it properly random all the time? It should also be called out in the documentation that you will not get exactly the same sequence twice.

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -0,0 +1,304 @@
...
+
+    def _iter_batches(self) -> Iterator[TensorDict]:
+        if self._instances is not None or self.num_workers <= 0:
+            for batch in self._instances_to_batches(self.iter_instances()):
+                yield batch
+        else:
+            # At this point self.max_batches_in_memory is not None since lazy must be False.
+            assert self.max_batches_in_memory is not None
+
+            # First we start "instance workers", which are in charge generating raw
```
dirkgr, line 230: "of"

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -0,0 +1,304 @@
...
+                target=self._batch_worker, args=(instance_queue, batch_queue,), daemon=True
+            )
+            batch_worker.start()
+
+            # We can now start consuming from the `batch_queue` as the `batch_worker`
+            # produces batches.
+            batch_group: Optional[List[TensorDict]]
+            for batch_group in iter(batch_queue.get, None):
+                yield from batch_group
+                batch_queue.task_done()
```
dirkgr, line 249: If all we're doing here is read from the batch queue, why can't this process, the parent process, be the `batch_worker`?

allennlp/data/data_loaders/multi_process_data_loader.py
dirkgr: This is what happens on the receiving end: https://github.com/allenai/bettermap/blob/master/bettermap.py#L107

dirkgr: Ah, and then there is this bit: https://github.com/allenai/bettermap/blob/master/bettermap.py#L132

That makes sure that all the other workers get killed when something goes wrong in one of them.

allennlp/data/data_loaders/pytorch_data_loader.py
```diff
@@ -0,0 +1,224 @@
...
+    def __len__(self):
+        return len(self.instances)
+
+    def __iter__(self) -> Iterator[Instance]:
+        """
+        Even though it's not necessary to implement this because Python can infer
+        this method from `__len__` and `__getitem__`, this helps with type-checking
+        since `AllennlpDataset` can be considered an `Iterable[Instance]`.
+        """
+        yield from self.instances
```
dirkgr, line 33: There is a semantic difference though. If you pretend to be an `Iterable` through `__getitem__`, it runs `index_fields()`. Here, it does not.

allennlp/data/data_loaders/pytorch_data_loader.py
```diff
@@ -0,0 +1,224 @@
...
+def allennlp_worker_init_fn(worker_id):
+    worker_info = data.get_worker_info()
+    dataset = worker_info.dataset
+    if isinstance(dataset, AllennlpLazyDataset):
+        dataset.reader._set_worker_info(
+            WorkerInfo(num_workers=worker_info.num_workers, worker_id=worker_id)
+        )
+
+
+@DataLoader.register("pytorch_data_loader", constructor="from_partial_objects")
```
dirkgr, line 77: We generally don't put the type into the registrable name. I think this should just be called `"pytorch"` then.

allennlp/data/samplers/pytorch_samplers.py
```diff
@@ -151,12 +156,63 @@ class BasicBatchSampler(data.BatchSampler, BatchSampler):
...
 
-    def __init__(self, sampler: Sampler, batch_size: int, drop_last: bool):
+    def __init__(self, sampler: PyTorchSampler, batch_size: int, drop_last: bool):
         super().__init__(sampler, batch_size, drop_last)
+
+
+@PyTorchBatchSampler.register("bucket")
+class PyTorchBucketBatchSampler(PyTorchBatchSampler):
+    """
+    A PyTorch-compatable version of `BucketBatchSampler`.
```
dirkgr, line 173: "compatible"

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -0,0 +1,304 @@
...
+            if self.drop_last or num_instances % self.batch_size == 0:
+                return num_instances // self.batch_size
+            else:
+                return 1 + num_instances // self.batch_size
+        elif self.batches_per_epoch is not None:
+            return self.batches_per_epoch
+        else:
+            # We can't know the number of batches for a lazy loader when batches_per_epoch
+            # is not specified.
+            raise TypeError
```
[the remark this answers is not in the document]
epwalsh: `TypeError` is consistent with what PyTorch does. They provide some rational here: https://github.com/pytorch/pytorch/blob/master/torch/utils/data/sampler.py#L27

epwalsh:
Thanks for the review @dirkgr! I'll try to answer your questions:

> The lazy one has significantly worse results. But 100 instances is not a lot to keep in memory either. A long time ago when I wrote something similar, I think I kept 1024 instances? Or even 16384?

That's 100 batches, not 100 instances. I also tried with 10 batches and 500 batches. 100 batches worked the best.

> What's the story with compatibility with PyTorch now? Can you use PyTorch data loaders? Lightning?

PyTorch data loaders are still supported. See https://github.com/epwalsh/allennlp/blob/data-loading/allennlp/data/data_loaders/pytorch_data_loader.py.

> What happens when a worker dies? What happens to the workers when the parent dies? What happens when the batching process dies?

I'm still testing error handling, but when a worker dies (including a batching worker) the main process will throw an exception. When the main process dies, workers are also killed because we start workers with [`deamon=True`](https://docs.python.org/3.7/library/multiprocessing.html#multiprocessing.Process.daemon).

allennlp/data/data_loaders/multi_process_data_loader.py
epwalsh: Thanks, I'll check that out

allennlp/data/data_loaders/multi_process_data_loader.py
epwalsh: For non-lazy loading, you're right, we don't gain anything from having `_batch_worker` run in another process. But for lazy loading this is important since `_batch_worker` handles the batch sampling. Doing the batch sampling in the main process would slow things down.

allennlp/data/data_loaders/pytorch_data_loader.py
epwalsh: Good catch!

allennlp/data/data_loaders/multi_process_data_loader.py
epwalsh: > The way we have it now, early instances will never be shuffled together with late instances....

That's true for lazy loading, but not for non-lazy loading.

tests/data/data_loaders/pytorch_dataloader_test.py
```diff
@@ -20,10 +20,13 @@ def _read(self, filename: str) -> Iterable[Instance]:
             for i in range(NUM_INSTANCES):
                 yield Instance({"index": LabelField(i, skip_indexing=True)})
 
-    reader = FakeDatasetReader(lazy=lazy)
-    dataset = reader.read("blah")
-
-    loader = PyTorchDataLoader(dataset, batch_size=BATCH_SIZE, batches_per_epoch=BATCHES_PER_EPOCH)
+    loader = PyTorchDataLoader.from_partial_objects(
```
matt-gardner, line 23: If you're not actually using any `Lazy` objects, you shouldn't need to call this function.  Why not just call the constructor?

tests/data/samplers/max_tokens_batch_sampler_test.py
```diff
@@ -64,34 +58,27 @@ def test_guess_sorting_key_picks_the_longest_key(self):
         assert sampler.sorting_keys == ["passage"]
 
     def test_from_params(self):
```
matt-gardner, line 60: We don't need these tests anymore, because no classes actually implement `from_params` themselves.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -134,157 +100,85 @@ class DatasetReader(Registrable):
         filter out all but a subset of the instances that are needed so that you
         don't end up with duplicates.
 
-        !!! NOTE
-            **There is really no benefit of using a multi-process
-            `DataLoader` unless you can specifically implement a faster sharding mechanism
-            within `_read()`**. In that case you should set `manual_multi_process_sharding`
-            to `True`.
+        However, there is really no benefit to using multiple workers in your `DataLoader`
+        unless you implement handle the sharding within your `_read()` method, in which
```
matt-gardner, line 104:
```suggestion
        unless you implement the sharding within your `_read()` method, in which
```

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -134,157 +100,85 @@ class DatasetReader(Registrable):
...
+            raise ValueError("If specified, max_instances should be a positive int")
+
         self.max_instances = max_instances
-        self._cache_directory: Optional[Path] = None
-        if cache_directory:
-            self._cache_directory = Path(cache_directory)
-            os.makedirs(self._cache_directory, exist_ok=True)
         self.manual_distributed_sharding = manual_distributed_sharding
         self.manual_multi_process_sharding = manual_multi_process_sharding
+        self.__worker_info: Optional[WorkerInfo] = None
```
matt-gardner, line 153: You really want this to be double underscores?  We've never used this in other places in our code.

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -134,157 +100,85 @@ class DatasetReader(Registrable):
...
+        self.__dist_info: Optional[DistributedInfo] = None
```
matt-gardner, line 154: `dist` is pretty ambiguous; I'd recommend avoiding this abbreviation. (I also think that about the already-existing shortened import at the top of the file...)

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -304,143 +198,135 @@ def text_to_instance(self, *inputs) -> Instance:
         """
         raise NotImplementedError
 
-    def serialize_instance(self, instance: Instance) -> str:
+    def apply_token_indexers(self, instance: Instance) -> None:
```
matt-gardner, line 201: I'm pretty sure we can implement this for you, both to remove them and to restore them.  It's trivial if the field is just a `TextField` (store a map from field name to dictionary, using `id()` to make sure it's the same object per field name and crashing if it's not).  The only other case to handle is `ListField[TextField]`, which also shouldn't be hard, you just need to store how many nested lists there are.  This could be done in the base class implementation, allowing people to override it if they have some crazy new field type that they implement, but the default should just work, to both remove and restore the indexers, so no one has to implement anything.

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -0,0 +1,352 @@
...
+        shuffle: bool = False,
+        batch_sampler: Lazy[BatchSampler] = None,
+        batches_per_epoch: int = None,
+        num_workers: int = 0,
+        lazy: bool = False,
+        max_batches_in_memory: int = 100,
+        start_method: str = "fork",
+    ) -> "MultiProcessDataLoader":
+        if batch_sampler is not None:
+            batch_sampler_ = batch_sampler.construct()
```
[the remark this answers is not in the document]
matt-gardner: This is the wrong way to change for `None`, also; if you have a test that uses this class with `FromParams`, it should be breaking.  You have to check for `None` after the call to `.construct()` (a bit confusing, I know).

matt-gardner:
On caching, I agree that the existing functionality is not very useful, and it should probably just be removed.  We could later expand this new data loader to include caching functionality, perhaps using some database that just stores tensors directly.  The data loader is already the one that is caching things, so it makes more sense to have it happen there.  Except then the data loader has to know when to call the reader and when not...

Also, I didn't look very closely at the multiprocessing logic, I was looking at API-level stuff mostly.

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -0,0 +1,352 @@
...
+    _BATCH_CHUNK_SIZE = 10
+
+    def __init__(
+        self,
+        reader: DatasetReader,
+        data_path: str,
+        batch_size: int = 1,
+        drop_last: bool = False,
+        shuffle: bool = False,
+        batch_sampler: BatchSampler = None,
```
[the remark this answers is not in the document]
epwalsh: The `BucketBatchSampler` requires the `batch_size` to be given though.

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -0,0 +1,352 @@
...
+        self.start_method = start_method
+
+        # When lazy = False, we'll keep a cache of all instances in this list.
+        self._instances: Optional[List[Instance]] = None
+        # Keeps track of state when `batches_per_epoch` is used.
+        self._batch_generator: Optional[Iterator[TensorDict]] = None
+        # For indexing instances.
+        self._vocab: Optional[Vocabulary] = None
+
+        if not self.lazy:
```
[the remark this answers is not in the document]
epwalsh: That seems reasonable

allennlp/data/data_loaders/pytorch_data_loader.py
```diff
@@ -0,0 +1,228 @@
+from typing import List, Iterator
```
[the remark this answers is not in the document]
epwalsh: The `from_partial_objects` method is pretty much the only that's changed

allennlp/data/data_loaders/multi_process_data_loader.py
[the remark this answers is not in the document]
epwalsh: Good point

allennlp/data/dataset_readers/dataset_reader.py
epwalsh: Meh, yea just a personal preference of mine, but I'll stick with our conventions

allennlp/data/dataset_readers/dataset_reader.py
epwalsh: Hmm the problem I see with this is that the "removing token indexers" and "restoring the token indexers" steps happen in separate processes. The "removing" step in a worker and the "restoring" step in the main process. So without some multi-process communication, only the worker process would know what the mapping is.

I'll put some more thought into this though.

matt-gardner: Hmm, ok.  Another potential solution, again as a default implementation that would work for most use cases, and can be overridden for special cases: just assume that there's a `self._token_indexers`, which is the only one that gets used.  Then you don't need any communication between processes.

allennlp/data/data_loaders/multi_process_data_loader.py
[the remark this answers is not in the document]
epwalsh: I went with your suggestion of just using `max_batches_in_memory` (so no more `lazy` and no `cache_instances` params).

allennlp/data/dataset_readers/dataset_reader.py
epwalsh: That seems reasonable. And we can also check to make sure that `self._token_indexers` is the only `TokenIndexer` attribute of the `DatasetReader`, if it exists. Throwing an error if we find other token indexers.

allennlp/data/data_loaders/multi_process_data_loader.py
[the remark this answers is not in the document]
matt-gardner: I guess the thought was that if you get `None`, you could take that to mean "instantiate a bucket sampler in the constructor, passing in the batch size".  No strong opinion from me on this, though, I just have a hard time imagining a case where you _don't_ want a bucket sampler, so making it the default option seems like a reasonable thing to do.

allennlp/data/data_loaders/pytorch_data_loader.py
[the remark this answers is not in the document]
matt-gardner: Ok, I just looked that over, and it looks ok to me.

allennlp/data/dataset_readers/dataset_reader.py
epwalsh: On second thought, I'm not sure this is the right way to go.

My main concern is that this default implementation of `apply_token_indexers` would not be very efficient since there's no guarantee what fields an `Instance` can contain, and therefore `apply_token_indexers` would have to inspect each `Instance` to some degree.

matt-gardner: Yeah, but the difference between train and test will typically not be in `TextFields`.  It's probably sufficient to just guess what fields need to be modified from the first `Instance`, log what the guess is, and proceed.  If something crashes, catch the error and give a nicer message saying that our guess was wrong, and you need to override this method for multi-processing to work.

matt-gardner: It's maybe even possible to make that guess be a configurable parameter to the base `DatasetReader` class, similar to how the sorting keys are optional parameters to the bucket sampler, and guessed if they aren't provided.  It's trickier in this case because specifying the guess is hard, if there are lists.

matt-gardner: I guess, part of my thinking is, if we are going to require people to implement a new method, why don't we just go farther and fix the `TextField` API to remove those arguments in the first place?  If we aren't willing to do that (and there are very good reasons not to), then we should try hard to make it so that no one has to modify their code in order to use this new functionality.

epwalsh: I'm still thinking about how to implement this efficiently, but a couple of thoughts:

> Yeah, but the difference between train and test will typically not be in `TextField`s.

Seq2seq models are one example where this isn't true.

> If we aren't willing to [change the `TextField` API] (and there are very good reasons not to), then we should try hard to make it so that no one has to modify their code in order to use this new functionality.

By "new functionality" are you referring to multi-process loading? Because they will already need to modify their `_read()` method to handle the sharding logic. Additionally overriding `apply_token_indexers` doesn't seem like a big ask.

matt-gardner: Hmm, ok, fair enough.  Still, though, you've given a helper method that makes the modification to `_read()` into a one-line change, unless I'm missing something.  The barrier to implementing this `apply_token_indexers` method feels a lot higher.

Also, having a default implementation that covers almost all use cases and an example of overriding it seems like a strict improvement over having no default implementation.

This is a pretty minor point for me, too, in the context of this PR.  Just do what you think is best.

epwalsh: Yea, true. Well, I'll put some more thought into it and come back to it tomorrow. Like, my main concern is that we could be making assumptions which, when wrong, would introduce very subtle bugs.

allennlp/data/data_loaders/multi_process_data_loader.py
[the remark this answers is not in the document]
epwalsh: I see. Yea that's a good point. So then "if `batch_size` size given, `shuffle` set to `False`, and `batch_sampler` is `None`, default to using `BucketBatchSampler`"?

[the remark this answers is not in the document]
epwalsh: Although, the problem with that logic is there would be no way to actually turn off both shuffling and batch sampling at the same time. And you might want to turn those off like, for example, when testing and you want the batches to be in a deterministic order.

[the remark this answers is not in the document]
matt-gardner: Hmm, yeah, fair point.  Maybe there's a way to have a non-`None` default option which produces a bucket sampler, but you can explicitly pass `None` to turn off batch sampling?  (I was thinking that whatever shuffle and batch size options you provided would be passed to the bucket sampler.)

Not sure if this is worth it or not. 🤷‍♂️

allennlp/data/data_loaders/multi_process_data_loader.py
dirkgr: Right, but with the pool approach, you can make it not-true for lazy loading as well.

dirkgr:
Is this ready for another look?

epwalsh:
@dirkgr yeup, this is ready for another look whenever you get a change

epwalsh:
New benchmarks: https://beaker.org/ws/ai2/allennlp-new-data-loading/experiments?archived=false&sort=created:ascending

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -0,0 +1,434 @@
...
+            if worker.is_alive():
+                worker.terminate()
+
+    def _gather_instances(self, queue: mp.JoinableQueue) -> Iterable[Instance]:
+        done_count: int = 0
+        while done_count < self.num_workers:
+            for instances_chunk, worker_error in iter(queue.get, (None, None)):
+                if worker_error is not None:
+                    e, tb = worker_error
+                    sys.stderr.write("".join(tb))
```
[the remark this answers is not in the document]
epwalsh: When the error `e` is sent across processes we lose the original traceback, and unfortunately traceback objects are not pickleable, so we can't send the traceback with the exception. This way at least we still get to see the original traceback.

PR #4499, opened 2020-07-21
adding a function to get model's description from docstring

AkshitaB:
Adding get_description to Model base class.


AkshitaB:
Converted to draft, as it may no longer be required; depending on approval of changes here: https://github.com/allenai/allennlp-models/pull/98

PR #4500, opened 2020-07-23
More BART changes

dirkgr:
This is mainly about being able to test models that don't return a loss in eval mode.

### DIFF
--- a/allennlp/common/testing/model_test_case.py
+++ b/allennlp/common/testing/model_test_case.py
@@ -166,16 +166,18 @@ class ModelTestCase(AllenNlpTestCase):
         print("Predicting with loaded model")
         loaded_model_predictions = loaded_model(**loaded_batch)
 
-        # Check loaded model's loss exists and we can compute gradients, for continuing training.
-        loaded_model_loss = loaded_model_predictions["loss"]
-        assert loaded_model_loss is not None
-        loaded_model_loss.backward()
-
         # Both outputs should have the same keys and the values for these keys should be close.
         for key in model_predictions.keys():
             self.assert_fields_equal(
                 model_predictions[key], loaded_model_predictions[key], name=key, tolerance=tolerance
             )
+
+        # Check loaded model's loss exists and we can compute gradients, for continuing training.
+        loaded_model.train()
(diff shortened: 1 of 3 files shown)


allennlp/data/token_indexers/pretrained_transformer_indexer.py
```diff
@@ -106,6 +106,8 @@ def tokens_to_indices(self, tokens: List[Token], vocabulary: Vocabulary) -> Inde
     def indices_to_tokens(
         self, indexed_tokens: IndexedTokenList, vocabulary: Vocabulary
     ) -> List[Token]:
+        self._add_encoding_to_vocabulary_if_needed(vocabulary)
```
matt-gardner, line 109: There was an issue that came up a little while ago that made me think we should also add this line to `count_vocab_items`.  What do you think?  I don't think it would hurt anything to have it there too, and it fixes a potential bug.

allennlp/common/testing/model_test_case.py
```diff
@@ -166,17 +166,19 @@ def ensure_model_can_train_save_and_load(
...
-        loaded_model_loss.backward()
-
         # Both outputs should have the same keys and the values for these keys should be close.
         for key in model_predictions.keys():
             self.assert_fields_equal(
                 model_predictions[key], loaded_model_predictions[key], name=key, tolerance=tolerance
             )
 
+        # Check loaded model's loss exists and we can compute gradients, for continuing training.
+        loaded_model.train()
```
matt-gardner, line 176: Yes, this order makes more sense, this is the way it should have been originally.

allennlp/data/token_indexers/pretrained_transformer_indexer.py
dirkgr: `count_vocab_items()` doesn't receive a `Vocabulary`, so it can't be done there.

We still have a problem with this whole setup anyways, because there is no guarantee that we pass in the same vocab object every time, but we remember whether we've already done this in the indexer class. This only happens to work, but it's not right.

But this is not the PR to fix it.

PR #4525, opened 2020-07-31
Metrics in distributed setting

### LINKED ISSUE #4050: Distributed evaluation when metrics cannot be averaged across processes.
 - OS: All
 - Python version: Any 
 - AllenNLP version: Master
 - PyTorch version: 1.4

**Question**

Shouldn't the aggregation of evaluation metrics in the case of distributed evaluation be delegated to the concrete metric classes, instead of taking an average of the metrics across processes? This might not be a problem for metrics like accuracy but for something like recall or precision this can lead to very different numbers in single process evaluation vs distributed evaluation. In particular, I am talking about the following line in code:

[https://github.com/allenai/allennlp/blob/c497103542096037dc21c24466127bc6c820e75b/allennlp/training/util.py#L351](https://github.com/allenai/allennlp/blob/c497103542096037dc21c24466127bc6c820e75b/allennlp/training/util.py#L351)

Am I missing something? 

Thanks!

AkshitaB:
Fixes #4050 

Currently work-in-progress.

* Added a DistributedTestContextManager in `allennlp.common.testing`. 
* Removed generic `dist.all_reduce` in `training.utils` that sums output from distributed processes.
* Changes to `get_metric` method for metric classes.
* All metrics will return either a single float value, or a Dict.

- [x] Attachment scores
- [x] AUC
- [x] Average (aggregated on CPU).
- [x] BLEU
- [x] Boolean accuracy
- [x] Categorical accuracy
- [x] Covariance
- [x] Entropy
- [x] Evalb_bracketing_scorer (aggregated on CPU).
- [x] f1_measure
- [x] fbeta_measure
- [x] Mean absolute error
- [x] Pearson correlation.
- [x] Perplexity (aggregated on CPU).
- [x] ROUGE
- [x] Sequence accuracy
- [ ] Span-based f1_measure
- [x] Spearman correlation
- [x] Unigram recall


allennlp/training/metrics/metric.py
```diff
@@ -50,3 +54,12 @@ def detach_tensors(*tensors: torch.Tensor) -> Iterable[torch.Tensor]:
         """
         # Check if it's actually a tensor in case something else was passed.
         return (x.detach() if isinstance(x, torch.Tensor) else x for x in tensors)
+
+    @staticmethod
+    def _aggregate_metric(metric_tensor: torch.Tensor) -> torch.Tensor:
```
AkshitaB, line 59: This is not required and will be removed.

allennlp/training/util.py
```diff
@@ -286,7 +287,7 @@ def get_metrics(
     the `"loss"` metric is "average loss per batch".
     Returns the `"batch_loss"` separately.
     """
-    metrics = model.get_metrics(reset=reset)
+    metrics = model.get_metrics(reset=reset, world_size=world_size)
```
AkshitaB, line 290: Needs another input - cuda_device.

allennlp/training/metrics/bleu.py
```diff
@@ -134,7 +135,29 @@ def __call__(
...
+        self,
+        reset: bool = False,
+        world_size: int = 1,
+        cuda_device: Union[int, torch.device] = torch.device("cpu"),
+    ) -> Dict[str, float]:
+
+        if world_size > 1:
+            _prediction_lengths = torch.tensor(self._prediction_lengths).to(cuda_device)
+            _reference_lengths = torch.tensor(self._reference_lengths).to(cuda_device)
+            for ngram in self._precision_matches:
```
AkshitaB, line 148: Concerned about efficiency here.

epwalsh: I could be wrong, but I don't think this will produce the right results.

`self._precision_matches` could have different keys (ngram indices) for each worker, right? So each time `all_reduce` is used, it might be reducing on mismatched ngrams

allennlp/training/metrics/bleu.py
```diff
@@ -134,7 +135,29 @@ def __call__(
...
+                _precision_matches = torch.tensor(self._precision_matches[ngram]).to(cuda_device)
+                _precision_totals = torch.tensor(self._precision_totals[ngram]).to(cuda_device)
+                dist.all_reduce(_precision_matches, op=dist.ReduceOp.SUM)
+                dist.all_reduce(_precision_totals, op=dist.ReduceOp.SUM)
+                self._precision_matches[ngram] = _precision_matches.item() / world_size
+                self._precision_totals[ngram] = _precision_totals.item() / world_size
```
dirkgr, line 154: Is that correct? What if one worker saw an ngram 100 times, and another saw it only once?

allennlp/training/metrics/bleu.py
dirkgr: The fact that this works at all makes it clear to me that I don't understand how `torch.dist` works. If one worker has seen 10 different ngrams, and another has seen 11, wouldn't this hang? How could it possibly match up the tensors from the worker process in the `all_reduce` call, let alone match them to the correct ngram (like @epwalsh already noted)?

Does `torch.dist` only work if all the workers take the exact same code path? That seems like an insanely dangerous programming model.

allennlp/training/metrics/boolean_accuracy.py
```diff
@@ -95,9 +101,13 @@ def get_metric(self, reset: bool = False):
             accuracy = float(self._correct_count) / float(self._total_count)
         else:
             accuracy = 0.0
+        if world_size > 1:
+            accuracy_tensor = torch.tensor(accuracy).to(cuda_device)
+            dist.all_reduce(accuracy_tensor, op=dist.ReduceOp.SUM)
+            accuracy = accuracy_tensor.item() / world_size
```
dirkgr, line 107: This is only correct if each worker has the same `self._total_count`, right? Can we rely on that?

allennlp/training/metrics/metric.py
```diff
@@ -50,3 +54,22 @@ def detach_tensors(*tensors: torch.Tensor) -> Iterable[torch.Tensor]:
...
+    def _aggregate_metrics(
+        metrics: Dict[str, Any],
+        world_size: int = 1,
+        cuda_device: Union[int, torch.device] = torch.device("cpu"),
+    ) -> Dict[str, Any]:
+        """
+        Aggregate metrics across different processes. The default is to
+        take the average.
+        """
+        # raise NotImplementedError
```
dirkgr, line 68: leftovers?

allennlp/models/basic_classifier.py
```diff
@@ -176,8 +176,13 @@ def make_output_human_readable(
         output_dict["tokens"] = tokens
         return output_dict
 
-    def get_metrics(self, reset: bool = False) -> Dict[str, float]:
-        metrics = {"accuracy": self._accuracy.get_metric(reset)}
+    def get_metrics(
+        self,
+        reset: bool = False,
+        world_size: int = 1,
+        cuda_device: Union[int, torch.device] = torch.device("cpu"),
```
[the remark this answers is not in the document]
AkshitaB: dist functions only work if dist.init_process_group has been called. We need a way to indicate that this is a distributed setting.

allennlp/training/metrics/metric.py
AkshitaB: Yes, the PR is still WIP.

allennlp/training/metrics/bleu.py
epwalsh: The way I understand it, anytime a node calls any sort of distributed syncronization/communication function (e.g. a reduce function, a barrier, etc.), that call will block until it can communicate with all "matching" calls from the other nodes. The other nodes don't need to take the same code path, but they do need to call the same function.

So if there's a mismatched number of reduce operation calls between nodes, I think the nodes with the most reduce calls would essentially freeze up indefinitely as they wait for a matching call from the other nodes, right?

allennlp/models/basic_classifier.py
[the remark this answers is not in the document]
epwalsh: What we usually do is something like this:

```python
import torch.distributed as dist
from allennlp.common.util import is_distributed

world_size = 1 if not is_distributed() else dist.get_world_size()
```

That said, one benefit of having the `get_metrics` method take the `world_size` as a parameter is that it makes it really clear when you're implementing your own model that you have to handle to distributed case.

allennlp/training/metrics/attachment_scores.py
```diff
@@ -95,6 +101,24 @@ def get_metric(self, reset: bool = False):
...
+
+        if world_size > 1:
+            self._unlabeled_correct = torch.tensor(self._unlabeled_correct).to(cuda_device)
+            self._exact_unlabeled_correct = torch.tensor(self._exact_unlabeled_correct).to(
+                cuda_device
+            )
+            self._labeled_correct = torch.tensor(self._labeled_correct).to(cuda_device)
+            self._exact_labeled_correct = torch.tensor(self._exact_labeled_correct).to(cuda_device)
+            self._total_sentences = torch.tensor(self._total_sentences).to(cuda_device)
+            self._total_words = torch.tensor(self._total_words).to(cuda_device)
```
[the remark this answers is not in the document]
epwalsh: With the `nccl` back-end, which is what we use for distributed GPU training, `all_reduce` is only supported on GPU: https://pytorch.org/docs/stable/distributed.html#backends

allennlp/models/basic_classifier.py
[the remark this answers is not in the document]
dirkgr: You mean when implementing a new metric?

allennlp/training/metrics/attachment_scores.py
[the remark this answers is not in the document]
dirkgr: `nccl` supports nothing on CPU. What a terrible API.

allennlp/models/basic_classifier.py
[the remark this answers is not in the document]
AkshitaB: @epwalsh Yes, I was actually changing to that just now (not committed yet). Perhaps we can do something else to indicate that people should handle the distributed case when adding their own metrics.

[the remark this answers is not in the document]
epwalsh: That or if implementing a new model I suppose.

allennlp/training/metrics/bleu.py
AkshitaB: I've moved the `all_reduce` now. Since `_ngram_weights` will be the same for all workers, the number of reduce calls should be the same.

allennlp/common/testing/__init__.py
```diff
@@ -45,3 +50,42 @@ def cpu_or_gpu(test_method):
...
+            assert_allclose(metrics[key], desired_values[key])
+
+
+def global_distributed_metric(
+    global_rank: int,
+    world_size: int,
+    gpu_id: Union[int, torch.device],
+    metric: Metric,
+    metric_kwargs: Dict[str, Any],
+    desired_values: Dict[str, Any],
```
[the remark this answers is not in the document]
AkshitaB: Not really. `metric_kwargs` can be any tensors, or float (in case of perplexity), or even lists of nltk trees. 
 Or do you mean `metric_kwargs` and `desired_values` should be a combined Dict argument to `global_distributed_metric`?

allennlp/common/testing/distributed_test_context_manager.py
```diff
@@ -0,0 +1,80 @@
...
+
+    device_ids: `List[int]`
+        List of cuda devices. There need to be at least 2 devices. Default is [-1, -1].
+
+    # Usage
+
+    ```
+    python
+    with DistributedTestContextManager(devices_id) as test_this_function:
+        test_this_function(your_distributed_func, name_of_variable_to_check, value_to_check_against)
```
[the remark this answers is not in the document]
AkshitaB: Initially there was a need to clean up stray processes. This is no longer required. That said, having a class instead of different functions seems neater.

allennlp/training/metrics/auc.py
```diff
@@ -82,7 +87,28 @@ def __call__(
...
+                world_size = dist.get_world_size()
+                _all_predictions = [
+                    torch.zeros(self._all_predictions.shape) for i in range(world_size)
+                ]
+                _all_gold_labels = [
+                    torch.zeros(self._all_gold_labels.shape, dtype=torch.long)
+                    for i in range(world_size)
+                ]
+                dist.all_gather(_all_predictions, self._all_predictions)
+                dist.all_gather(_all_gold_labels, self._all_gold_labels)
```
AkshitaB, line 102: Potential memory concern with concatenating tensors across processes.

dirkgr:
GitHub has sent me approximately 1000 email about this, but it looks like it's in progress. I will not look at it again until you tell me to, or until it leaves "Draft" status. Sound good?

allennlp/common/testing/distributed_test_context_manager.py
```diff
@@ -0,0 +1,88 @@
...
+
+    device_ids: `List[int]`
+        List of devices. There need to be at least 2 devices. Default is [-1, -1].
+
+    # Usage
+
+    ```
+    python
+    with DistributedTestContextManager(devices_id) as test_this_function:
+        test_this_function(your_distributed_func, *args)
```
dirkgr, line 24: I strongly disagree that this is cleaner as a context manager, even if there was cleanup. You could restructure this so you need only the second line from your example:
```Python
test_distributed_function(devices_id, your_distributed_func, *args)
```
That's all it takes! The context manager adds nothing.

allennlp/training/metrics/auc.py
```diff
@@ -82,7 +87,30 @@ def __call__(
             [self._all_gold_labels, torch.masked_select(gold_labels, mask).long()], dim=0
         )
 
+        if is_distributed():
+            try:
+                # The following logic will only work if the batches are of equal length.
```
dirkgr, line 92: Is that a safe assumption?

allennlp/training/metrics/auc.py
dirkgr: I think with one float per instance, we're OK here.

allennlp/training/metrics/bleu.py
```diff
@@ -116,10 +118,21 @@ def __call__(
...
             precision_matches, precision_totals = self._get_modified_precision_counts(
                 predictions, gold_targets, ngram_size
             )
+            if is_distributed():
+                _precision_matches = torch.tensor(precision_matches).to(device)
+                _precision_totals = torch.tensor(precision_totals).to(device)
+                dist.all_reduce(_precision_matches, op=dist.ReduceOp.SUM)
+                dist.all_reduce(_precision_totals, op=dist.ReduceOp.SUM)
+                precision_matches = _precision_matches.item() / world_size
+                precision_totals = _precision_totals.item() / world_size
```
dirkgr, line 135: Aren't `precision_matches` and `precision_totals` counts? Why is it OK to average them?

allennlp/training/metrics/categorical_accuracy.py
```diff
@@ -103,8 +114,11 @@ def get_metric(self, reset: bool = False):
             accuracy = float(self.correct_count) / float(self.total_count)
         else:
             accuracy = 0.0
+
         if reset:
             self.reset()
+
+        # return {"accuracy": accuracy}
```
dirkgr, line 121: ?

allennlp/training/metrics/pearson_correlation.py
```diff
@@ -69,9 +75,21 @@ def get_metric(self, reset: bool = False):
         covariance = self._predictions_labels_covariance.get_metric(reset=reset)
         predictions_variance = self._predictions_variance.get_metric(reset=reset)
         labels_variance = self._labels_variance.get_metric(reset=reset)
+        denominator = math.sqrt(predictions_variance) * math.sqrt(labels_variance)
+        if is_distributed():
+            # Note: this gives an approximate aggregation of the covariance.
+
+            # Aggregating on the CPU since these are single float values.
+            device = torch.device("cpu")
```
dirkgr, line 83: @epwalsh mentioned that this won't work when the backend is NCCL though?

allennlp/training/metrics/rouge.py
```diff
@@ -148,6 +156,18 @@ def _get_rouge_n_stats(
...
+            device = predicted_tokens.device
+            _total_recall = torch.tensor(total_recall).to(device)
+            _total_precision = torch.tensor(total_precision).to(device)
+            _total_f1 = torch.tensor(total_f1).to(device)
+            dist.all_reduce(_total_recall, op=dist.ReduceOp.SUM)
+            dist.all_reduce(_total_precision, op=dist.ReduceOp.SUM)
+            dist.all_reduce(_total_f1, op=dist.ReduceOp.SUM)
+            total_recall = _total_recall.item()
+            total_precision = _total_precision.item()
+            total_f1 = _total_f1.item()
```
dirkgr, line 169: This ends up returning the sum of the F1 scores? Maybe I'm missing context, but why is it OK to sum up the F1 scores?

allennlp/training/metrics/spearman_correlation.py
```diff
@@ -58,17 +63,40 @@ def __call__(
             self.total_predictions = torch.cat((self.total_predictions, predictions), 0)
             self.total_gold_labels = torch.cat((self.total_gold_labels, gold_labels), 0)
 
+        if is_distributed():
+            try:
+                # The following logic will only work if the batches are of equal length.
```
dirkgr, line 68: Is that a safe assumption?

allennlp/training/metrics/spearman_correlation.py
```diff
@@ -58,17 +63,40 @@ def __call__(
...
+                ]
+                _total_gold_labels = [
+                    torch.zeros(self.total_gold_labels.shape, device=device)
+                    for i in range(world_size)
+                ]
+                dist.all_gather(_total_predictions, self.total_predictions)
+                dist.all_gather(_total_gold_labels, self.total_gold_labels)
+                self.total_predictions = torch.cat(_total_predictions, dim=0)
+                self.total_gold_labels = torch.cat(_total_gold_labels, dim=0)
+            except Exception:
```
dirkgr, line 83: Can you catch something more specific? This will mask other errors too.

allennlp/training/metrics/spearman_correlation.py
```diff
@@ -58,17 +63,40 @@ def __call__(
...
+                logger.warning(
+                    "Distributed aggregation for SpearmanCorrelation is currently not supported "
+                    "for batches of unequal length."
+                )
```
dirkgr, line 87: I am happier with the "equal-length batches" restriction if we at least fail properly when the assumption is violated.

dirkgr: Oh, I see this is just a warning. I'm thinking we should throw an exception. Otherwise people will run this stuff, miss the warning, copy the number into their paper and publish it, and then get called out and it's our fault. @matt-gardner, do you have thoughts about this?

How hard is it to get rid of the equal-length limitation completely?

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed testing models that only return a loss when they are in training mode.
 - Fixed a bug in `FromParams` that caused silent failure in case of the parameter type being `Optional[Union[...]]`.
 - Fixed a bug where the program crashes if `evaluation_data_loader` is a `AllennlpLazyDataset`.
+- Changed the evaluation of metrics in the distributed setting to metric-specific logic rather than default averaging.
```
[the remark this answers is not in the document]
AkshitaB: Haha, alright!

allennlp/training/metrics/rouge.py
AkshitaB: The calculation for getting the precision, recall, f1-scores does the following in the for loop:

```
# Accumulate stats
total_recall += recall
total_precision += precision
total_f1 += f1
```

In the distributed setting, we sum the sums.

allennlp/training/metrics/spearman_correlation.py
AkshitaB: Unfortunately, the exception that's thrown when the batches are unequal is of the generic type `Exception`.

allennlp/training/metrics/spearman_correlation.py
AkshitaB: That makes sense. For backwards compatibility, I was returning what was already being returned. But perhaps, failing here, and making the model explicitly take care of handling the failure is a better way.

allennlp/common/testing/__init__.py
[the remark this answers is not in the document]
AkshitaB: Ah, I see what you're saying. As a matter of fact, `metric_kwargs` is supposed to be `Dict[str, List[Any]]`. I'll update the type.

allennlp/training/metrics/auc.py
AkshitaB: Batches will be of equal length in all cases except when using `max_tokens_batch_sampler`. Currently, we emit a warning in that case. Can change it to error.

allennlp/training/metrics/bleu.py
AkshitaB: That's correct. Fixing.

allennlp/training/metrics/pearson_correlation.py
AkshitaB: Yes, correct. Fixing.

AkshitaB:
@dirkgr @epwalsh The CI failure for Check Models is due to a required change in the models repository. Details here: https://github.com/allenai/allennlp-models/pull/109

allennlp/training/metrics/auc.py
dirkgr: I'd prefer an error. Downstream components may make choices based on the metrics (such as early stopping), and a warning after 30 minutes of training is easy to miss.

allennlp/training/metrics/rouge.py
dirkgr: I'll assume Tobi had his reasons for writing it this way then.

allennlp/training/metrics/spearman_correlation.py
dirkgr: What I have done in the past in these cases is pattern match on the string form of the exception. This is a particularly egregious example: https://github.com/allenai/pipette/blob/master/pipette/pipette.py#L656

If we can't change the kind of exception that's thrown, this is a last resort. Catching everything is too dangerous. Even `MemoryError` is a type of `Exception`, and we definitely don't want to catch that and pretend it's handled.

allennlp/training/metrics/auc.py
dirkgr: Oh, I see you've already done this. Just not in this here diff :-)

AkshitaB:
> I think the only thing left then is that `Exception` business.

I do throw a RuntimeException instead of a warning for cases when we don't handle distributed training. Is that what you mean?

PR #4526, opened 2020-08-01
add 'use_amp' option to transformer embedders

epwalsh:
This adds an option to enable AMP for just the transformer part of your model. I found this was helpful because AMP currently fails with certain core torch modules, like anything that's an RNN, and so in those cases enabling AMP at the trainer level will result in a crash.

See https://github.com/pytorch/pytorch/issues/36428, for example, and also the failure in [this build](https://github.com/allenai/allennlp-models/pull/104/checks?check_run_id=933149196) where I tried to get AMP working with CopyNet.

That said, it feels like this solution is a little too ad hoc. Maybe there's a clean way to make this more general so that you could easily configure any part of your model to use AMP individually. On the other hand, doing this to the transformer part of a model will probably be the most common use case, and so the complexity involved in making this more general may not be worth it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- A `use_amp` option to `PretrainedTransformerEmbedder` and `PretrainedTransformerMismatchedEmbedder`,
+  which you can set to `True` if you want automatic mixed precision enabled only for the transformer
+  part of your model.
+
 ## [v1.1.0rc2](https://github.com/allenai/allennlp/releases/tag/v1.1.0rc2) - 2020-07-31
 
 ### Changed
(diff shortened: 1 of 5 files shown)


epwalsh:
> Another option, maybe, because this looks so easy to use, is to just find the modules that are broken and explicitly disable it there, no matter what the global setting is.

True, but we'd also have to go through every model that uses an RNN cell directly, like CopyNet, and wrap those calls with `autocast(False)`. Maybe that wouldn't be too bad, and actually this worked fine with CopyNet: https://github.com/allenai/allennlp-models/pull/104/checks. But we'd have to remember to revert all of those changes as soon as AMP is working with RNNs.

epwalsh:
Huh, so AMP actually seems to work fine with our PyTorch Seq2Seq / Seq2Vec wrappers that use RNNs, at least with the tests that I ran, and I fixed the issues with LSTM cells in https://github.com/allenai/allennlp-models/pull/104. So I'm going to close this.

matt-gardner:
Sounds like a good outcome to me :)

PR #4537, opened 2020-08-05
improve the typing annotations in allennlp

wj-Mcat:
There are many type annotation in allennp which use old style. I think we should improve the params type annotation style to `__future__ annotations`. 

refer to : 
- [PEP 563 -- Postponed Evaluation of Annotations](https://www.python.org/dev/peps/pep-0563/)
- [example type annotation in python-wechaty](https://github.com/wechaty/python-wechaty/blob/master/src/wechaty/wechaty.py)

I have change a single module, and there are many modules that use the old style type annotation. How do you think about it ?  If it's ok, I will make next commits for type annotation.

### DIFF
--- a/allennlp/data/vocabulary.py
+++ b/allennlp/data/vocabulary.py
@@ -3,6 +3,8 @@ A Vocabulary maps strings to integers, allowing for strings to be mapped to an
 out-of-vocabulary token.
 """
 
+from __future__ import annotations
+
 import codecs
 import copy
 import logging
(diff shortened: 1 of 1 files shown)


wj-Mcat:
Although it doesn't support `python 3.6`, this feature is more humanize.

AkshitaB:
Hi, as of now, we would like to support python 3.6 . Once we upgrade to supporting 3.7 and 3.8, we will be happy to make these changes.

PR #4538, opened 2020-08-05
Fix logger being created twice

bryant1410:


AkshitaB:
Thank you for the fix!

PR #4540, opened 2020-08-05
EpochCallBack for tracking epoch

nitishgupta:
Added an EpochCallback in `trainer.py` to track the epoch as a member of the model class.

### DIFF
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -179,9 +179,10 @@ EpochCallback.register("null")(EpochCallback)
 @EpochCallback.register("track_epoch_callback")
 class TrackEpochCallback:
     """
-    A callback that you can pass to the `GradientDescentTrainer` to access the current epoch number in progress in the
-    model as a class member `model.epoch`. Since the EpochCallback passes `epoch=-1` at the start of the training,
-    we set `model.epoch = epoch + 1` which now denotes the number of completed epochs at a given training state.
+    A callback that you can pass to the `GradientDescentTrainer` to access the current epoch number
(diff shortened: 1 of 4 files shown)


nitishgupta:
I don't know why the codecov/patch test fails.

matt-gardner:
Thanks @nitishgupta!  That check is failing because there is no test for your code.  Can you add a really simple one?

nitishgupta:
Thanks @matt-gardner. When I run the tests locally, all my tests (`32 failed, 8 passed, 1 skipped in 15.80`) for `trainer_test.py` fail with `AttributeError: module 'torch.cuda.amp' has no attribute 'autocast'` but pass here.
Tests for `tests/nn/util_test.py` pass though for some reason.

allennlp/training/trainer.py
```diff
@@ -176,6 +176,28 @@ def __call__(
 EpochCallback.register("null")(EpochCallback)
 
 
+@EpochCallback.register("track_epoch_callback")
+class TrackEpochCallback:
+    """
+    A callback that you can pass to the `GradientDescentTrainer` to access the current epoch number
+    in progress in the model as a class member `model.epoch`. Since the EpochCallback passes `epoch=-1`
```
matt-gardner, line 183: Minor wording suggestion:

```suggestion
    A callback that you can pass to the `GradientDescentTrainer` to access the current epoch number
    in your model during training. This callback sets `model.epoch`, which can be read inside of
    `model.forward()`. Since the EpochCallback passes `epoch=-1`
```

matt-gardner: Github let me just commit this myself, so I did it.

Issue #4532, opened 2020-08-04
How can we tokenize sentence pair with 'PretrainedTransformerTokenizer'


wlhgtc:
@dirkgr  Seem the new version code remove  @ZhaofengWu 's changes to support sentence pairs in #3868 
How can we tokenizer sentence pair in 1.1.0 ?

wlhgtc:
@ZhaofengWu  I read your discussion in #3868, but still confused with the change.
What changes leads no support for sentence pair  or  any other ways to do it ? Maybe you could help me ?

dirkgr:
Use `PretrainedTransformerTokenizer.add_special_tokens()`. Make sure that the input you pass to that function doesn't have special tokens in it already.

The reason I split it this way is because we were on the path of creating a single `tokenize()` call that does everything, like huggingface has. I don't think that's a good design. I think we need to have separate functions for tokenization, adding special tokens, splitting long sequences, and so on. This was one step in that direction. Sadly, for backwards compatibility, we didn't fully commit to it, so all we have is the `add_special_tokens()` call. I am hoping to rectify that situation later.

wlhgtc:
@dirkgr   Thanks for your quickly reply. 

In my view, suppose the 2 part is  `tokens2` and `tokens2` code is as follows ?
```
def text_to_instance(tokens1:str,tokens2:str)
    ...
    tokens1 = self.tokenize(pair1)
    tokens2 = self.tokenize(pair2)
    concat_tokens = self.add_special_tokens(tokens1, tokens2)
    field['tokens'] = TextField(concat_tokens, self._token_indexers)
    ...
```

dirkgr:
That is correct!

wlhgtc:
Case solved, close this issue

PR #4542, opened 2020-08-06
Adjust tokenize sentence pair in 'pretrained_transformer_tokenizer'

wlhgtc:
The original method to tokenize sentence pairs looks wired, in #4532. 
So I adjust the tokenize method.


wlhgtc:
@dirkgr  
I adjust the original `tokenize ` function cause it's  a little bit wired. 
In my view: we should use the original `tokenize ` function (now named `_tokenize_text `) just for text. Then the new `tokenize` function could contains : `_tokenizer_text`,`add_special_tokens` functions.

dirkgr:
This is a well-crafted PR, but what is it for? What does it fix?

wlhgtc:
> This is a well-crafted PR, but what is it for? What does it fix?

This PR adjust `tokenize` function in `pretrained_transformer_tokenizer`.
In old version code, if you want tokenize sentence pair, you need code like:
```
    tokens1 = self.tokenize(pair1)
    tokens2 = self.tokenize(pair2)
    concat_tokens = self.add_special_tokens(tokens1, tokens2)
```
But now, we just need:
```
    tokens= self.tokenize(pair1, pair2)
``` 
The original `add_special_tokens` function support tokenize sentence pair, but it's covered by `tokenize` function which only support process single sentence. 
So I remove the call of  `add_special_tokens` function from `tokenize` function, and use a new function to  do it.

dirkgr:
I understand what it does, but why? Why would you expect `tokenize()` to do something about pairs of sentences? Why would you expect a function called `tokenize()` to do anything about special tokens?

wlhgtc:
> I understand what it does, but why? Why would you expect `tokenize()` to do something about pairs of sentences? Why would you expect a function called `tokenize()` to do anything about special tokens?


In some situations, we do need to tokenize sentence pair (e.g. in  multi-choice tasks, we want diff query and context).
But the original implement may be confused for those who don't read the Tokenizer class carefully. 

So I want the `tokenize` method become the only **outside method** for users,  these settings( like 'add special tokens', 'tokenize sentence pair'  and 'splitting long sequences' )could be controlled by parameters.

As you mentioned in #4532, these settings need separate functions, but they should be inside(or private) method and be called by `tokenize()` method.

It's convenient and user-friendly for some fresh users. They don't need to  combine these different methods outside. All they have to do is read function declaration and do what they wan by set different parameters.

Hope it's acceptable for you.

dirkgr:
I'm afraid I disagree. There are too many different concerns to handle them all in a single function. We have no way of knowing ahead of time what problems they want to solve with the tools we give them, and what issues people may run into. Even the huggingface `tokenize()` function can only handle what the huggingface people have already thought of. If you need something different, you will have a very hard time with it. We can do a better job for users by providing them with a set of functions that are easy to understand, and combine well.

It is unfortunate that `add_special_tokens` is `True` by default, because it breaks the goal of "combining well". For now, you have to remember to set it to `False`. It had to be `True` for backwards compatibility, and it is something I want to change for version 2.0 of AllenNLP (unless we have time for a more comprehensive fix I am working on in my head).

wlhgtc:
You've sold it to me. We can't meet all users' needs, provide separate functions is a good idea. 
And  hope the new version will come soon.

dirkgr:
I left a note at #4572.

PR #4544, opened 2020-08-07
add gradient checkpointing for transformer token embedders

ArneBinder:
Gradient checkpointing has been released with huggingface/transformers v3.0.0. Gradient checkpointing reduces memory by 5x which makes it possible to process longer sequences on smaller GPUs. This PR allows to fine-tune transformer token embedders on a more regular GPU with high(er) max_length values.

Sorry for not providing any tests, I have no clue how to create a simple test case and no resources at the moment to figure that out. But I tried it succesfully with Longformer (max_length: 4096) and SciBert (max_length: 512) on a single RTX 2080 Ti.


dirkgr:
Awesome. I just noticed that the support for checkpointing in Huggingface came originally from AI2's own @ibeltagy. Thanks for driving that!

PR #4545, opened 2020-08-07
Static type checking fixes

### LINKED ISSUE #4535: A "py.typed" marker
Hi all,

I was wondering if there's a specific reason that neither `allennlp`, nor `allennlp-models` include an empty marker file named `py.typed` in their respective python package root directories, as that file is required to mark a package as "PEP-561 compliant", which gives static code analysis tools like mypy the go-ahead to do type inference.

Otherwise, I am currently forced to clone the repo locally, add the marker file, and `pip install -e` to stop mypy from complaining that stubs are missing for `allennlp`.

davidatbu:
Small type annotation fixes. 

PS: I had `isort` runnning automatically which reordered some imports in the files I modified, hope that's ok(didn't see anything in CONTRIBUTING.md about import order).

Closes #4535 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Also, when `max_length` was set to a non-`None` value, several warnings would appear
   for certain transformer models around the use of the `truncation` parameter.
 - Fixed evaluation of all metrics when using distributed training.
+- Added a `py.typed` marker. Fixed type annotations in `allennlp.training.util`.
 
 ## [v1.1.0rc2](https://github.com/allenai/allennlp/releases/tag/v1.1.0rc2) - 2020-07-31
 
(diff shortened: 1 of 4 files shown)


davidatbu:
@AkshitaB , CI is failing because I didn't update  CHANGELOG.md, and because code coverage(of the diff?) is below 90%.

Lmk if you think I should update CHANGELOG.md. About code coverage, no idea what to do.

dirkgr:
What's going on with this PR? Can we get it in? I'm a big fan of having this!

epwalsh:
I think `py.typed` may need to be listed in the `package_data` section of `setup.py`. See https://mypy.readthedocs.io/en/latest/installed_packages.html#making-pep-561-compatible-packages

davidatbu:
@epwalsh , @dirkgr , sorry for the radio silence. There's a `include_package_data=True` arg passed to `setup()`, which does the same thing as `package_data = { "": [ 'py.typed'] }` in our case. I tested this by doing `pip install  .` in a fresh virtual env, and I found the `py.typed` marker in the installed dir.

Issue #4557, opened 2020-08-13
Metrics for Multi-label Classification Task


shounakpaul95:
Is there any built-in functionality to calculate metrics like Precision, Recall and F1-Score for the multi-label multi-class classification scenario (one or more target classes)?

The current implementation of FBeta Measure can only take inputs for the single-label scenario.

JohnGiorgi:
We are actually working on exactly that! Our implementation subclasses `FBetaMeasure` and has (almost) all the same unit tests. Still stuck on a couple.

You can take a look at it [here](https://github.com/semantic-health/allennlp-multi-label) where we also have multi-label dataset reader and a multi-label text classificatio model. The metric specifically is [here](https://github.com/semantic-health/allennlp-multi-label/blob/master/allennlp_multi_label/metrics.py).

foodiehack:
I have made my custom function for calculating Precision, Recall, and F1-Score for the multi-label classification. It's a generic function that can be used with any algo or framework you are using for multilabel classification.

dirkgr:
@JohnGiorgi, we'd be excited to have this in the core library. I'm happy to review a PR or two :-)

PR #4562, opened 2020-08-15
Multi-label F-beta metric

JohnGiorgi:
Discussed in #4557, this PR adds a multi-label f-beta metric `training.metrics.FBetaMultiLabelMeasure`. It subclasses `FBetaMeasure` and just implements the `__call__` function. I copied all the tests for `FBetaMeasure` and modified them for the multi-label case. As a sanity check, I compared it to `sklearn.metrics` modules.

```python
import torch

from allennlp.training.metrics import FBetaMultiLabelMeasure
from sklearn.preprocessing import MultiLabelBinarizer
from sklearn.metrics import precision_score, recall_score, f1_score
from numpy.testing import assert_almost_equal

num_examples = 10
num_classes = 20
threshold = 0.5

y_true = torch.randint(2, (num_examples, num_classes))
y_pred = torch.randint(2, (num_examples, num_classes))

fbeta = FBetaMultiLabelMeasure(average="micro")
fbeta(y_pred, y_true)
actual = fbeta.get_metric()

expected = {
    "fscore": f1_score(y_true, y_pred >= threshold, average="micro"),
    "precision": precision_score(y_true, y_pred >= threshold, average="micro"),
    "recall": recall_score(y_true, y_pred >= threshold, average="micro")
}

assert_almost_equal(actual["fscore"], expected["fscore"])
assert_almost_equal(actual["precision"], expected["precision"])
assert_almost_equal(actual["recall"], expected["recall"])
```

This is *almost* ready to go. The only thing I haven't figure out is how passing a `mask` to `__call__` should work in the multi-label case, and controlling for edge cases (batch size of one, no prediction for a given class etc.). The tests that need to be implemented (copied from `FBetaMeasure`) are:

- [X] `test_fbeta_multilabel_with_mask`
- [X] `test_fbeta_multilabel_handles_batch_size_of_one`
- [X] `test_fbeta_multilabel_handles_no_prediction_false_last_class`
- [X] `test_fbeta_multilabel_handles_no_prediction_true_last_class`
- [X] `test_fbeta_multilabel_handles_no_prediction_true_other_class`
- [X] `test_fbeta_multilabel_handles_no_prediction_true_all_class`
- [x] I also don't think `self.true_negative_sum` is being computed correctly for the multi-label case.

I am struggling with these but will continue to work on it. If the answer to any of these are obvious to anyone let me know or feel free to tack on to the PR 😄


dirkgr:
Screen Shot 2020-08-17 at 09 31 41 :eyes:

Most of it is tests though. That's a good thing :-)

dirkgr:
I don't have easy answers to your questions, but I suspect it should be done analogously to `FBetaMeasure` and `sklearn`'s implementation.

We already have a dependency on `sklearn`. I'm assuming there is a reason you can't implement ours in terms of theirs? I suppose it can't handle the distributed case properly?

Thank you for tackling this!

JohnGiorgi:
> We already have a dependency on `sklearn`. I'm assuming there is a reason you can't implement ours in terms of theirs? I suppose it can't handle the distributed case properly?

I started by looking at `FBetaMeasure` and `F1Measure` and noticed neither uses `sklearn` (correct me if I am wrong). I just figured there was some reason for this, so I didn't use `sklearn` in the `FBetaMeasureMultiLabel` subclass.

If there is no good reason _not_ to use sklearn let me know and I can re-work the implementation (but that would probably mean `FBetaMeasure` and `F1Measure` should be similary re-worked, right?)

dirkgr:
If you want to do the work, and we don't get a problem with performance (especially in the distributed case), I would welcome it. I'm in favor of anything that means we can have the same functionality with less code. But if you think the fastest way to the goal is to implement it ourselves, I won't question it.

JohnGiorgi:
> If you want to do the work, and we don't get a problem with performance (especially in the distributed case), I would welcome it. I'm in favor of anything that means we can have the same functionality with less code. But if you think the fastest way to the goal is to implement it ourselves, I won't question it.

I think it could work & greatly simplify things. AFAICT it would involve deleting a bunch of code, computing the metrics with `sklearn.metrics` and updating the unit tests accordingly.

I am mainly confused with what would happen to the `reset` method. If `sklearn.metrics` is being used to compute precision/recall/fbeta, there's no need to hold on to `_true_positive_sum`, `_pred_sum`, `_true_sum` and `_total_sum` so there's nothing to reset. Is removing this going to break something?

dirkgr:
Conceptually, the `reset()` method must do something. Otherwise, something is amiss. AllenNLP metrics accumulate data every time you call them (once per batch, usually), and then give you the metric when you ask for it. The metric is always a summary of the data that was given since the last call to `reset()` (or since the metrics object was constructed). So when `reset()` doesn't do anything, something must be missing.

dirkgr:
Sorry for the long delay. I have not forgotten about this.

allennlp/training/metrics/fbeta_multi_label_measure.py
```diff
@@ -0,0 +1,210 @@
...
+from overrides import overrides
+
+from allennlp.common.checks import ConfigurationError
+from allennlp.common.util import is_distributed
+from allennlp.training.metrics import FBetaMeasure
+from allennlp.training.metrics.metric import Metric
+
+
+@Metric.register("fbeta_multi_label")
+class FBetaMultiLabelMeasure(FBetaMeasure):
```
dirkgr, line 14: It might be good to have a multi-label F1 that is just this class with a hard-coded `1.0` for beta. People will be looking for that.

allennlp/training/metrics/fbeta_multi_label_measure.py
```diff
@@ -0,0 +1,210 @@
...
+            mask = torch.ones_like(gold_labels).bool()
+        gold_labels = gold_labels.float()
+
+        # If the prediction tensor is all zeros, the record is not classified to any of the labels.
+        pred_mask = (predictions.sum(dim=-1) != 0).unsqueeze(-1)
+        threshold_predictions = torch.where(
+            predictions >= self._threshold,
+            torch.ones_like(predictions),
+            torch.zeros_like(predictions),
+        )
```
dirkgr, line 152: Isn't that the same as
```suggestion
        threshold_predictions = (predictions >= self._threshold).float()
```

dirkgr (approved):
I love it, especially the detailed tests!

allennlp/training/metrics/fbeta_multi_label_measure.py
```diff
@@ -0,0 +1,210 @@
...
+        # Shape: (num_classes, )
+        self._total_sum: Union[None, torch.Tensor] = None
+        # the total number of instances under each _predicted_ class,
+        # including true positives and false positives
+        # Shape: (num_classes, )
+        self._pred_sum: Union[None, torch.Tensor] = None
+        # the total number of instances under each _true_ class,
+        # including true positives and false negatives
+        # Shape: (num_classes, )
+        self._true_sum: Union[None, torch.Tensor] = None
```
dirkgr, line 103:
```suggestion
        super().__init__(beta, average, labels)
        self._threshold = threshold
```
Any reason we can't do this?

dirkgr (changes requested):

allennlp/training/metrics/fbeta_multi_label_measure.py
JohnGiorgi: It is! I've replace the code with your suggestion 👍

allennlp/training/metrics/fbeta_multi_label_measure.py
JohnGiorgi: Much simpler, good catch! Fixed.

dirkgr (approved):
Wonderful!

Issue #4360, opened 2020-06-15
Can't get correct output using .saliency_interpret_from_json() from Interpret


koren-v:
Hi, I'm trying to get the importance of each word in a sentence while sentiment classification but always get the same result for different inputs. The Code:
```
from allennlp.predictors.predictor import Predictor
import allennlp_models.classification
from allennlp.interpret.saliency_interpreters import (
    SaliencyInterpreter,
    SimpleGradient,
    SmoothGradient,
    IntegratedGradient,
)

predictor = Predictor.from_path("https://storage.googleapis.com/allennlp-public-models/sst-roberta-large-2020.06.08.tar.gz")
simple_grad = SimpleGradient(predictor)

input_json = {'sentence' : "a very well-made, funny and entertaining picture."}

simple_grad.saliency_interpret_from_json(input_json)
```
gives:
```
{'instance_1': {'grad_input_1': [1.0,
   0.0,
   0.0,
   0.0,
   0.0,
   0.0,
   0.0,
   0.0,
   0.0,
   0.0,
   0.0]}}
```
So what can be the reason of this problem?

matt-gardner:
No idea what's causing the issue yet, but I have confirmed that this is a bug.  I get the correct output with rc3, but it's broken with rc6.

koren-v:
@matt-gardner Thanks for your answer.

koren-v:
@matt-gardner Is it possible way to load this model in rc3? Because using the same code with this version gives:
```
sst_tokens not in acceptable choices for validation_dataset_reader.type: ['conll2003', 'interleaving', 'sequence_tagging', 'sharded', 'babi', 'text_classification_json']. You should either use the --include-package flag to make sure the correct module is loaded, or use a fully qualified class name in your config file like {"model": "my_module.models.MyModel"} to have it imported automatically.
```

matt-gardner:
Yeah, just include this line in your script: `from allennlp_models import sentiment`.

matt-gardner:
Possibly related: https://github.com/allenai/allennlp-models/pull/85

matt-gardner:
I thought the issue might have been resolved with recent changes to the SST model / tokenization stuff.  But I just ran the model on master (both this and the models repo), and had the same issue.  This is still not resolved.

[...]

Issue #4059, opened 2020-04-12
"TextFields must be passed Tokens" Configuration error


elmanderx:
System:
 - OS: OSX
 - Python version: 3-7
 - AllenNLP version:v0.9.0

Hello, I am looking to predict a text with AllenNLP and a trained SciBERT model - it compiles fine when I put:
`python -m allennlp.run predict ./scibert-master/outputdir2/model.tar.gz ./scibert-master/test.json --include-package scibert` but doesn't result in what I want since I want to use the text classifier. However when I input:
`python -m allennlp.run predict ./scibert-master/outputdir2/model.tar.gz ./scibert-master/test.json --include-package scibert --predictor text_classifier` I get the following error: 

```Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/me/Downloads/scibert-master/venv/src/allennlp/allennlp/run.py", line 21, in <module>
    run()
  File "/Users/me/Downloads/scibert-master/venv/src/allennlp/allennlp/run.py", line 18, in run
    main(prog="allennlp")
  File "/Users/me/Downloads/scibert-master/venv/src/allennlp/allennlp/commands/__init__.py", line 102, in main
    args.func(args)
  File "/Users/me/Downloads/scibert-master/venv/src/allennlp/allennlp/commands/predict.py", line 227, in _predict
    manager.run()
  File "/Users/me/Downloads/scibert-master/venv/src/allennlp/allennlp/commands/predict.py", line 206, in run
    for model_input_json, result in zip(batch_json, self._predict_json(batch_json)):
  File "/Users/me/Downloads/scibert-master/venv/src/allennlp/allennlp/commands/predict.py", line 151, in _predict_json
    results = [self._predictor.predict_json(batch_data[0])]
  File "/Users/me/Downloads/scibert-master/venv/src/allennlp/allennlp/predictors/predictor.py", line 64, in predict_json
    instance = self._json_to_instance(inputs)
  File "/Users/me/Downloads/scibert-master/venv/src/allennlp/allennlp/predictors/text_classifier.py", line 34, in _json_to_instance
    return self._dataset_reader.text_to_instance(sentence)
[...]

PR #4576, opened 2020-08-18
Pass strings, not tokens, to the dataset reader in TextClassifierPredictor

matt-gardner:
I got an error when trying to diagnose #4360 with current code, and this fixes that error.  Not sure what other side effects this will have; hopefully tests will catch it.


matt-gardner:
Now looking at the blame for what caused this and looking back through github conversations, I found [this](https://github.com/allenai/allennlp/issues/4059#issuecomment-656032426), which I agree with.  So the proper fix is in the dataset reader in the models repo, not here.  This does mean that we're not testing our models well enough, though, because the predictor for the SST reader doesn't currently work.  I'll add a test for this and fix it.

Issue #4424, opened 2020-06-30
Transformer Toolkit


dirkgr:
We want a collection of modules that form the building blocks of transformers, such as
 * self-attention
 * layer norm
 * feed-forward
 * positional embeddings
 * maybe memory layers like https://arxiv.org/abs/1907.05242?
(Not all of these really require code by themselves.)

We also want some useful combinations of these, like a combination of self-attention, layer-norm, and feed-forward, which forms one standard BERT layer.

Finally, we want a way of loading these and initializing them flexibly with huggingface weights.

For example, imagine a scenario where we run the first six layers of BERT-Large, then add in a memory layer, six more layers, and then some information from another modality. How would we configure this? How can we make this easy?

jiasenlu:
do we want to focus on BERT? which transformer models in your mind we should include for now?

dirkgr:
I don't think it matters much, since the way to load weights will be quite similar between the different models. That said, let's start with BERT and RoBERTa, and go from there?

dirkgr:
Matt has a bunch of code for this in https://github.com/allenai/allennlp/pull/4423. The work for this issue is about pulling that out into its own set of reusable modules.

dirkgr:
@AkshitaB, can we close this issue and open new, more detailed ones, for what we find is left to do on the toolkit?

PR #4577, opened 2020-08-19
Transformer toolkit

AkshitaB:
Fixes https://github.com/allenai/allennlp/issues/4424

Adding a collection of transformer building blocks.

- [x] Pull stuff from https://github.com/allenai/allennlp/blob/vision/allennlp/models/vilbert.py into separate modules.
- [ ] Memory layers
- [x] Add different types of attention mechanisms
- [x] Add different types of positional embeddings
- [ ] Add documentation
- [x] Ability to initialize from huggingface weights


matt-gardner:
Quick general comment: I think we should remove the word `Bert` from basically all of these classes.

AkshitaB:
> Quick general comment: I think we should remove the word `Bert` from basically all of these classes.

@matt-gardner I kept `Bert` to distinguish it from any future transformer modules that may become the norm. 😄

dirkgr:
BERT is just the name for one set of weights though. The architecture is called "transformer". RoBERTa is not BERT, but it could be loaded with these `Bert*` classes, right?

allennlp/modules/transformer/bert_biattention.py
```diff
@@ -0,0 +1,122 @@
...
+    ):
+        super().__init__()
+        if combined_hidden_size % num_attention_heads != 0:
+            raise ValueError(
+                "The hidden size (%d) is not a multiple of the number of attention "
+                "heads (%d)" % (combined_hidden_size, num_attention_heads)
+            )
+
+        self.num_attention_heads = num_attention_heads
+        self.attention_head_size = int(combined_hidden_size / num_attention_heads)
```
dirkgr, line 26: We should throw an exception if this is not an even division.

allennlp/modules/transformer/bert_biattention.py
```diff
@@ -0,0 +1,122 @@
...
+        self.query2 = torch.nn.Linear(hidden_size2, self.all_head_size)
+        self.key2 = torch.nn.Linear(hidden_size2, self.all_head_size)
+        self.value2 = torch.nn.Linear(hidden_size2, self.all_head_size)
+
+        self.dropout2 = torch.nn.Dropout(dropout2)
+
+    def transpose_for_scores(self, x):
+        new_x_shape = x.size()[:-1] + (self.num_attention_heads, self.attention_head_size,)
+        x = x.view(*new_x_shape)
+        return x.permute(0, 2, 1, 3)
```
dirkgr, line 44: This should either be documented, or (more likely) prefixed with a `_`. I'm assuming this is an internal function, right?

allennlp/modules/transformer/bert_biattention.py
```diff
@@ -0,0 +1,122 @@
...
+        self,
+        input_tensor1,
+        attention_mask1,
+        input_tensor2,
+        attention_mask2,
+        co_attention_mask=None,
+        use_co_attention_mask=False,
+    ):
+
+        # for vision input.
```
dirkgr, line 56: In full generality, either input could be anything.

allennlp/modules/transformer/bert_embeddings.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def __init__(
+        self,
+        vocab_size: int,
+        hidden_size: int,
+        pad_token_id: int,
+        max_position_embeddings: int,
+        type_vocab_size: int,
+        dropout: float,
+    ):
```
dirkgr, line 21: Do none of these need defaults?

allennlp/modules/transformer/bert_intermediate.py
```diff
@@ -0,0 +1,26 @@
+from typing import Union
+import torch
+
+from allennlp.common import FromParams
+
+from transformers.modeling_bert import ACT2FN  # TODO: remove this dependency?
```
dirkgr, line 6: That seems like a nice feature? I'm happy to support it.

allennlp/modules/transformer/bert_self_attention.py
```diff
@@ -0,0 +1,64 @@
...
+class BertSelfAttention(torch.nn.Module, FromParams):
+    def __init__(
+        self, hidden_size: int, num_attention_heads: int, dropout: float = 0.0,
+    ):
+        super().__init__()
+        if hidden_size % num_attention_heads != 0:
+            raise ValueError(
+                "The hidden size (%d) is not a multiple of the number of attention "
+                "heads (%d)" % (hidden_size, num_attention_heads)
+            )
```
dirkgr, line 17: I knew I had seen this check somewhere else!

allennlp/modules/transformer/bert_self_attention.py
```diff
@@ -0,0 +1,64 @@
...
+        self.query = torch.nn.Linear(hidden_size, self.all_head_size)
+        self.key = torch.nn.Linear(hidden_size, self.all_head_size)
+        self.value = torch.nn.Linear(hidden_size, self.all_head_size)
+
+        self.dropout = torch.nn.Dropout(dropout)
+
+    def transpose_for_scores(self, x):
+        new_x_shape = x.size()[:-1] + (self.num_attention_heads, self.attention_head_size,)
+        x = x.view(*new_x_shape)
+        return x.permute(0, 2, 1, 3)
```
dirkgr, line 31: I guess `BiAttention` could derive from `SelfAttention` so that we can share this function. But maybe that's too much machinery to save four lines of code.

allennlp/modules/transformer/bert_self_attention.py
```diff
@@ -0,0 +1,64 @@
...
+        key_layer = self.transpose_for_scores(mixed_key_layer)
+        value_layer = self.transpose_for_scores(mixed_value_layer)
+
+        # Take the dot product between "query" and "key" to get the raw
+        # attention scores.
+        attention_scores = torch.matmul(query_layer, key_layer.transpose(-1, -2))
+        attention_scores = attention_scores / math.sqrt(self.attention_head_size)
+        # Apply the attention mask is (precomputed for all layers in
+        # BertModel forward() function)
+        attention_scores = attention_scores + attention_mask
```
dirkgr, line 50: In this context, it doesn't really make sense to talk about where the mask comes from.

allennlp/modules/transformer/bert_self_output.py
```diff
@@ -0,0 +1,17 @@
...
+        super().__init__()
+        self.dense = torch.nn.Linear(hidden_size, hidden_size)
+        self.layer_norm = torch.nn.LayerNorm(hidden_size, eps=1e-12)
+        self.dropout = torch.nn.Dropout(dropout)
+
+    def forward(self, hidden_states, input_tensor):
+        hidden_states = self.dense(hidden_states)
+        hidden_states = self.dropout(hidden_states)
+        hidden_states = self.layer_norm(hidden_states + input_tensor)
+        return hidden_states
```
dirkgr, line 17: This is exactly the same as `BertOutput`, except that it doesn't support the `intermediate_size` parameter. Looks like we can just make a default of `intermediate_size == hidden_size`, and use only one class.

AkshitaB:
@dirkg @mattg This is ready for another look. In particular, the focus is on loading pretrained weights - `allennlp.modules.transformer.transformer_module` is the base class for this. I’ve tried to make it generic and flexible. Examples of how it is being used are in `allennlp.modules.transformer.output_layer`, `allennlp.modules.transformer.transformer_layer`, `allennlp.modules.transformer.bimodal_encoder`, etc. Note that `_huggingface_mapping` is just one default mapping, but `source` and `_construct_default_mapping` can be used for any number of defaults, in addition to the ability to specify any one-off `mapping` in `from_pretrained_module`. 

@dirkgr One of the tests in `transformer_encoder_test` is `test_loading_partial_pretrained_weights`, which can be a basis for the "give-me-x-layers-of-bert" type of functionality that we discussed.

Since I started with Vilbert, it’s possible that the toolkit is too geared towards that. However, once we establish if the current way of loading pretrained weights is fine, any other modules should be quicker. 

Vilbert using the transformer toolkit is at `allennlp.models.vilbert_new`. 

Other general modules are `SinusoidalPositionalEmbedding`, and the various attention_scores.

mattg:
@AkshitaB I'm not the Matt you're looking for.

AkshitaB:
> @AkshitaB I'm not the Matt you're looking for.

Ah, sorry about that!

matt-gardner:
@AkshitaB, it looks like this diff is showing a ton of changes that aren't intended to be in this PR.  Is that because the branch isn't up to date with `vision`?

AkshitaB:
> @AkshitaB, it looks like this diff is showing a ton of changes that aren't intended to be in this PR. Is that because the branch isn't up to date with `vision`?

@matt-gardner I've updated the branch now.

allennlp/models/vilbert_new.py
```diff
@@ -0,0 +1,276 @@
...
+
+        image_embeddings = torch.nn.Linear(feature_dim, hidden_dim)
+        image_location_embeddings = torch.nn.Linear(4, hidden_dim)
+        embeddings = [image_embeddings, image_location_embeddings]
+        super().__init__(embeddings, hidden_dim, dropout)
+
+
+@Model.register("nlvr2_vilbert_new")
+@Model.register("nlvr2_vilbert_from_huggingface_new", constructor="from_huggingface_model_name")
+class Nlvr2Vilbert(Model, TransformerModule):
```
matt-gardner, line 40: Why is this in a new file, registered as "new"?  I would have expected you to just modify the file that was there.

allennlp/models/vilbert_new.py
```diff
@@ -0,0 +1,276 @@
...
+    TransformerEmbeddings,
+    BiModalEncoder,
+    ActivationLayer,
+    TransformerModule,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class ImageFeatureEmbeddings(Embeddings):
```
matt-gardner, line 25: This seems like it should go in `modules` somewhere.

allennlp/modules/transformer/bimodal_encoder.py
```diff
@@ -0,0 +1,303 @@
...
+        pretrained_module: torch.nn.Module,
+        source="huggingface",
+        mapping: Optional[Dict[str, str]] = None,
+        **kwargs,
+    ):
+        """
+        The `pretrained_module` only supplies one of the modalities.
+        """
+
+        required_kwargs = [
```
matt-gardner, line 280: I don't really understand why we need these arguments to be hidden in `kwargs`.  `kwargs` are really messy and annoying to deal with in general, and I'd strongly recommend avoiding using them except in a few specific use cases.  They make it so you have to look at the method implementation to know what arguments a function takes, instead of just looking at the function definition, and that makes IDE integrations and documentation generators not work.

Is there another solution to the problem this is solving that doesn't require `kwargs`?

tests/models/vilbert_new_test.py
```diff
@@ -0,0 +1,74 @@
...
+        #     combined_hidden_size=5,
+        #     pooled_output_dim=7,
+        #     image_intermediate_size=11,
+        #     image_attention_dropout=0.0,
+        #     image_hidden_dropout=0.0,
+        #     v_biattention_id=[0, 1],
+        #     t_biattention_id=[0, 1],
+        #     fixed_t_layer=0,
+        #     fixed_v_layer=0,
+        # )
```
dirkgr, line 51: Leftovers?

dirkgr (changes requested):
There is no documentation on any of these classes. That's fine at this stage, but we will want it before we make an official release (i.e., before we merge `vision` into `master`).

That said, it would be great if we could identify some natural entry point to the transformer toolkit, maybe `allennlp/modules/transformer/__init__.py`, where we can stick some documentation with high-level examples of how to use this. I'm not set on which examples to include, but here are some ideas for things that aren't completely outlandish:
 * Use a 4-layer transformer on top of glove embeddings (with no pre-trained weights).
 * Use the embedding plus the first 4 layers of bert-base-uncased.
 * We have two pieces of text that we want to combine meaningfully. We use the first 8 layers of bert-base-uncased on these separately, then we concatenate the representations and use the final 4 layers on the concatenated sequence. This is a real case, an architecture @HarshTrivedi has experimented with before.
 * Instantiate bert-base-uncased, but use the compressed embedding system from Albert instead of the original ones. I could imagine an experiment where we want to save some space, so we try to train these embeddings only while keeping the rest of the transformer stack fixed.

If not all of these examples come out easy, let's use that a as starting points for planning next steps on the toolkit.

allennlp/modules/transformer/transformer_module.py
```diff
@@ -0,0 +1,105 @@
...
+
+import torch
+
+
+class TransformerModule(torch.nn.Module):
+    """
+    Base class to help with generalized loading of pretrained weights.
+    """
+
+    _huggingface_mapping: Dict[str, str] = {}
```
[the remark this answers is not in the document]
AkshitaB: An example of the structure being different from the pretrained module is in `BiModalEncoder`. Right now, we can take only one pretrained_module, but perhaps it makes sense to add the ability to specify more than one?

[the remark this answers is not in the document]
dirkgr: To me, the ability to map from one to the other is an implementation detail. It is how we interact with huggingface behind the scenes, but it doesn't have to be part of the official transformer toolkit API. So in that sense, I don't care that it covers all cases, as long as it covers the cases we care about.

allennlp/modules/transformer/transformer_embeddings.py
```diff
@@ -0,0 +1,159 @@
...
+        for i, inp in enumerate(inputs):
+            outputs.append(self.embeddings[i](inp))
+
+        outputs = sum(outputs)  # type: ignore
+        outputs = self.layer_norm(outputs)
+        outputs = self.dropout(outputs)
+        return outputs
+
+
+class TransformerEmbeddings(TransformerModule, FromParams):
```
dirkgr, line 100: Also no way to construct these from huggingface weights?

allennlp/modules/transformer/transformer_embeddings.py
```diff
@@ -0,0 +1,159 @@
...
+#     """Construct the embeddings from image, spatial location (omit now) and
+#     token_type embeddings.
+#     """
+
+#     def __init__(self, feature_dim: int, hidden_dim: int, dropout: float = 0.0):
+
+#         image_embeddings = torch.nn.Linear(feature_dim, hidden_dim)
+#         image_location_embeddings = torch.nn.Linear(4, hidden_dim)
+#         embeddings = [image_embeddings, image_location_embeddings]
+#         super().__init__(embeddings, hidden_dim, dropout)
```
dirkgr, line 159: Old code?

allennlp/modules/transformer/transformer_encoder.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def forward(
+        self,
+        hidden_states,
+        attention_mask=None,
+        head_mask=None,
+        encoder_hidden_states=None,
+        encoder_attention_mask=None,
+        output_attentions=False,
+        output_hidden_states=False,
```
dirkgr, line 47: Doesn't `mypy` demand type annotations for this?

allennlp/modules/transformer/transformer_encoder.py
```diff
@@ -0,0 +1,99 @@
...
+import torch
+
+from allennlp.common import FromParams
+
+from allennlp.modules.util import replicate_layers
+from allennlp.modules.transformer.transformer_layer import TransformerLayer
+from allennlp.modules.transformer.transformer_module import TransformerModule
+
+
+class TransformerEncoder(TransformerModule, FromParams):
```
dirkgr, line 12: I'd call this `TransformerBlock`. This is really a bunch of fully-formed transformer layers stacked together, right? End-to-end BERT would be a `Embedding` plus a `TransformerBlock`?

dirkgr: When it's time to write docs for this, we should be clear about which components correspond to which BERT, and how they plug together.

allennlp/modules/transformer/transformer_module.py
```diff
@@ -0,0 +1,183 @@
...
+    `_huggingface_mapping` is an optional mapping for each class, that determines
+    any differences in the module names between the class modules and the huggingface model's
+    modules.
+
+    `_relevant_module` is an optional str which is the expected name of the module in
+    the huggingface pretrained model.
+    """
+
+    _huggingface_mapping: Dict[str, str] = {}
+    _relevant_module: str = ""
```
dirkgr, line 25: If it's optional, it should be `Optional[str] = None`.

allennlp/modules/transformer/transformer_module.py
```diff
@@ -0,0 +1,183 @@
...
+            mapping = self._huggingface_mapping
+        for name, module in self.named_modules():
+            if name != "":
+                if hasattr(module, "_construct_default_mapping"):
+                    # We handle collisions by giving priority to the outer module's mapping.
+                    mapping = dict(
+                        list(module._construct_default_mapping(source).items())
+                        + list(mapping.items())
+                    )
+        # self._default_mapping = mapping
```
dirkgr, line 69: Leftover code?

dirkgr (changes requested):
I had a ton of comments, but there isn't a lot wrong with this actually. It's mostly small stuff or questions.

If you can address a comment in less than 15 minutes, let's do it in this PR, but if it becomes something major, let's merge it as it is and we'll make an issue for it. This PR is big enough as it is, and we're already having trouble with incompatible merges, so let's not let that get worse.

PR #4578, opened 2020-08-19
Vision features cache script

sanjayss34:
Create a standalone script to produce image features and cache them.


allennlp/data/dataset_readers/nlvr2.py
```diff
@@ -52,7 +52,6 @@ def __init__(
         image_loader: ImageLoader,
         image_featurizer: GridEmbedder,
         region_detector: RegionDetector,
-        *,
```
dirkgr, line 55: Why did you take this away? It makes sure we can later re-order the optional parameters without breaking the API.

scripts/cache_vision_features.py
```diff
@@ -0,0 +1,47 @@
...
+    image_keys = [os.path.basename(filename) for filename in image_paths]
+
+    image_loader = DetectronImageLoader()
+    image_featurizer = ResnetBackbone()
+    region_detector = FasterRcnnRegionDetector()
+    if torch.cuda.is_available() and args.use_cuda:
+        backbone.cuda()
+        region_detector.cuda()
+
+    images, sizes = image_loader(image_paths)
```
dirkgr, line 36: This loads all the images at the same time. Does that fit?

Also, how much faster is it to run this than running just 2 images at a time (which is what the NLVR reader does right now)?

sanjayss34: Actually loading all images at once does give an error when using a GPU, but it's unclear to me why the image loader interacts with the GPU at all...

I found that caching the NLVR2 dev set takes 39 minutes with a batch size of 16 (batch size of 32 didn't fit on 24 GB GPU). How does this compare with what you found for doing it 2 at a time? [I changed it so that images are loaded in batches to make this work, but still wondering about why loading all at once causes a problem]

dirkgr: Doing the training set took about 16 hours I think.

dirkgr:
I am finding that data loading does in fact transfer the images to the GPU. That's not ideal. We'll fix it in #4595.

scripts/cache_vision_features.py
sanjayss34: Ok, extrapolating the dev set time from this gives 8103*16/103171 \approx 1.25 hours, so I think there is a speed up here.

OhadRubin:
I think the design choice of using a key-value based in-memory a bit lacking.
Key-value based solutions are best suited for mutable objects, this is a cache, so it won't be changing.
I suggest to move to something similar to fairseq [here](https://github.com/pytorch/fairseq/blob/master/fairseq/data/indexed_dataset.py) in MMapIndexedDataset.
There might be speedup from using this type of solution reads/writes sequentially.
It is also might be more suitable for multiprocessing, since every process can have it's own file, where all the files are merged at the end of processing.
@dirkgr, what do you think?

epwalsh:
> It is also might be more suitable for multiprocessing, since every process can have it's own file, where all the files are merged at the end of processing.

👍 that's a good point

jiasenlu:
i checked the code, the reason for oov is simply the batch_size is too large. Try to use batch_size = 1.

sanjayss34:
> * can we implement a save and resume that can start from the existing lmdb file?

I think this would

dirkgr:
@OhadRubin, are you talking about using lmdb for the cache?

It does support multiprocessing as it is already. All the processes can open the same cache files, and write and read to them as they like. You could use @sanjayss34's script to fill the cache, and at the same time start the training process that reads it. I have tried this, and it works. You can, in fact, run multiple training runs at the same time, all reading and filling the same incomplete cache, and they coordinate like they should!

I don't know if sequential writing and reading would speed it up further, but at least in the way we're using it now, the `TensorCache` class is not a bottleneck, and the flexibility of not having to finish writing before starting to read is nice.

We're also talking about using lmdb to cache instances, but that's a different story, and not decided yet. If `MMapIndexedDataset` has a solution there, we'd be quite interested. The name implies exactly what we want ...

dirkgr:
I looked at the code for `MMapIndexedDataset` a bit. It won't quite do what we need, since it only stores tensors, while our `Instances` are more than just tensors. The other issue with caching at the instance level is that the same image might appear in multiple instances, and we have to prevent copying the image data for every instance that needs it.

In fact the NLVR reader right now takes no appreciable amount of time to create instances if the cache of preprocessed images is warm. But that's only one dataset. I'm sure in other cases we'll see different results.

OhadRubin:
> We're also talking about using lmdb to cache instances, but that's a different story, and not decided yet. If `MMapIndexedDataset` has a solution there, we'd be quite interested. The name implies exactly what we want ...

Yeah, I was thinking about it in that context.

> I looked at the code for `MMapIndexedDataset` a bit. It won't quite do what we need, since it only stores tensors, while our `Instances` are more than just tensors.

I agree, for instances we could have for example: the 'FromParams Metaclass save option' solution Matt suggested, could have an option to save as bytes. But that is agnostic to the choice of cache implementation.

> The other issue with caching at the instance level is that the same image might appear in multiple instances, and we have to prevent copying the image data for every instance that needs it.

Is that so bad? For instance caching the solution here is not optimal, instances should be kept in memory contiguously.

Regarding multiprocessing, correct me if i'm wrong, but we are writing to the same (lmdb) file from multiple processes, and that can't possibly be good. A solution where each process writes to a different file sounds way faster.

dirkgr:
lmdb has the concept of a transaction built in. I don't know how it's implemented internally, but there are many possible ways. That's how it makes it safe to write to the same file from multiple processes.

It supports only one concurrent write transaction, but as long as processing the data takes a long time, and storing it takes a short time, this is not a limitation. We only start a write transaction when the data is fully processed and ready to be written. For NLVR (and the script in this PR), this has not been a problem. Check out this line: https://github.com/allenai/allennlp/blob/DetectronNLVR/allennlp/common/file_utils.py#L386 The write transaction is only held for a small amount of time.

OhadRubin:
I understand. Thanks.

dirkgr:
@sanjayss34, @jiasenlu, I changed the code a bit so that the model's preprocessing step is run with the model, not with the data loading.

I also noticed that this script does not take into account entries that might already be in the cache. Can you add that check, so it doesn't create cache entries for images that already exist?

sanjayss34:
@dirkgr I've made the change you requested; is this okay to be merged now?

sanjayss34:
@dirkgr Please take another look when you're ready and let me know if this can be merged. Thanks!

dirkgr:
@sanjayss34, for some reason I can't push into this branch anymore. I just wanted to merge the latest `DetectronNLVR` into this branch. Did you change something?

scripts/cache_vision_features.py
```diff
@@ -0,0 +1,58 @@
...
+    coordinates_cache = TensorCache(os.path.join(args.cache_dir, "coordinates"))
+    image_paths = []
+    for extension in IMG_EXTENSIONS:
+        image_paths += list(glob.iglob(os.path.join(args.image_dir, "**", "*."+extension), recursive=True))
+    filtered_image_paths = []
+    image_keys = []
+    for path in image_paths:
+        key = os.path.basename(path)
+        if key in features_cache and key in coordinates_cache:
+            continue
```
dirkgr, line 36: It's a bit of a nit, but can you delay this check until the second loop? It means you have to re-work the batching logic a bit, but it has several advantages:
1. It will start up much quicker when you have lots of images.
2. You can run multiple copies of this script at the same time, pointing at the same cache files. Also, you can run this at the same time as a training job, all pointing at the same files, and they will coordinate (almost) perfectly.

dirkgr:
Because I can't push into this branch anymore, I opened another PR with the same contents here: https://github.com/allenai/allennlp/pull/4633

I will close this one then to avoid duplication.

PR #4585, opened 2020-08-21
specify module names by a regex in predictor.capture_model_internals()

### LINKED ISSUE #4574: speedup predictor.capture_model_internals by specifying layers by a regexp
It turns out that predictor.capture_model_internals works really slow when the model is large, e.g. with BERT. The bottleneck is call to `sanitize` to the outputs of each layer, which will cause a lot of memory copies from torch device to python list. It may be an unnecessary computation in many cases, because for transformers predictor will capture outputs not only from each layers, but also from each attention, and so on. I think that we may make it faster by specifying regexp that should match module name to actually capture a module.

add a parameter to the `Predictor.capture_model_internals` like `module_regex: str`, and inside the function
```
compiled_module_regex = re.compile(module_regex)
for idx, (name, module) in enumerate(self._model.named_modules()):
    if compiled_module_regex.fullmatch(name):
        <actually copy module outputs>
```

We may also make copying itself faster, but this will not fix the memory consumption problem.

mojesty:
This fixes #4574

I did essentially the same thing I described in the issue -- added a regex which module name must satisfy for module outputs to be captured. It speeds up capture_model_internals significantly when only a subset of internals are importnat for user, saving time and memory.

It is backward compatible PR, because default regex if r'.*', so nothing will break.
Covered it with tests.


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- predictor.capture_model_internals() now accepts a regex specifying
+  modules to capture
```
epwalsh, line 12:
```suggestion
### Added

- `Predictor.capture_model_internals()` now accepts a regex specifying
  which modules to capture
```

PR #4597, opened 2020-08-25
make existing readers work with multi-process loading

epwalsh:
This makes it so the dataset readers we provide can be used efficiently with the multi-process data loader.

### DIFF
--- a/allennlp/data/dataset_readers/babi.py
+++ b/allennlp/data/dataset_readers/babi.py
@@ -127,6 +127,7 @@ class BabiReader(DatasetReader):
 
         return Instance(fields)
 
+    @overrides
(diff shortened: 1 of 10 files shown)


dirkgr:
I have a question open at https://github.com/allenai/allennlp-models/pull/122, so I'll wait for that to get resolved before looking here.

dirkgr (approved):
Hmm, I guess that's better, but not as much better as I had hopes. I was somehow hoping that there would be a single place we could call this in the base `Predictor` class, but it looks like there is not.

This will require a bunch of changes in `allennlp-models` then, because that's where most of the predictors live, right?

epwalsh:
We should only need to update `Predictors` that override these methods. Hopefully that's not too many.

Issue #4596, opened 2020-08-25
Add batched_index_fill to allennlp.nn.util


OhadRubin:
Would you guys be interested in adding the following utility method to util.py?
```
import torch
import allennlp
from allennlp.nn import util

def batched_index_fill(
    target: torch.Tensor,
    indices: torch.LongTensor,
    mask,
    flattened_indices = None,
    fill_value:int = 1 
) -> torch.Tensor:
    prev_shape = target.size()
    
    flattened_indices = util.flatten_and_batch_shift_indices(indices*mask, target.size(1))
    flattened_target = target.view(-1, 1)
    mask = mask.bool().view(-1)
    flattened_target.scatter_(0, flattened_indices[mask].unsqueeze(-1),fill_value)
    return flattened_target.reshape(prev_shape)


tensor = torch.zeros([3,5])
indices = torch.tensor([[4,2,3,-1],[0,1,-1,-1],[1,3,-1,-1]])
mask = indices>=0

batched_index_fill(tensor, indices, mask)
```
Output:
```
tensor([[0., 0., 1., 1., 1.],
        [1., 1., 0., 0., 0.],
        [0., 1., 0., 1., 0.]])
```
Also, I think this can be extended to a general scatter operation as well.

OhadRubin:
And a general scatter operation:
```
import torch
import numpy as np
import allennlp
from allennlp.nn import util

def batched_index_scatter(
    target: torch.Tensor,
    indices: torch.LongTensor,    
    mask,
    replace,
    flattened_indices = None,
) -> torch.Tensor:
    prev_shape = target.size()
    flattened_indices = util.flatten_and_batch_shift_indices(indices*mask, target.size(1))
    flattened_target = target.view(-1, target.size(-1))
    mask = mask.bool().view(-1)
    flattened_target[flattened_indices[mask]] = replace.view(-1,replace.size(-1))[mask]
    return flattened_target.reshape(prev_shape)

leaf_indices = []
batch_size = 3
seq_len = 5
emb_size = 1

tensor = torch.zeros([3,5,2])
indices = torch.tensor([[4,2,3,-1],[0,1,-1,-1],[3,1,-1,-1]])
replace = torch.arange(2*indices.numel()).reshape(indices.shape+(2,)).float()
replace = torch.arange(indices.numel()).float().unsqueeze(-1)
replace = torch.cat([replace,replace],dim=-1)
mask = indices>=0
batched_index_scatter(tensor, indices, mask, replace)
[...]

PR #4598, opened 2020-08-25
Added batched versions of scatter and fill to util.py

OhadRubin:
See #4596


matt-gardner:
@dirkgr, you seem like you have the most context for reviewing this.

allennlp/nn/util.py
```diff
@@ -1264,6 +1264,89 @@ def batched_index_select(
...
+
+    return filled_target
+
+
+def batched_index_scatter(
+    target: torch.Tensor, indices: torch.LongTensor, mask: torch.Tensor, replace: torch.Tensor,
+) -> torch.Tensor:
+    """
+    The given `indices` in `target` will be will be scattered with corresponding index
+    from the `replace` tensor given a `mask`.
```
dirkgr, line 1314: In a way, this is `target[indices] = replace[indices]`, but with a mask? Is an operation like that really called "scatter"? It sounds more like `batched_index_replace` to me.

images

I have checked the pytorch docs, and it is indeed called "scatter": https://pytorch.org/docs/stable/tensors.html#torch.Tensor.masked_scatter_

I'm surprised this function doesn't boil down to `masked_scatter()` or `index_copy()`.

OhadRubin: Yeah, they do, but they don't deal with both masks and indices, I guess better names would be masked_index_fill and masked_index_replace

allennlp/nn/util.py
```diff
@@ -1264,6 +1264,89 @@ def batched_index_select(
...
+    # Shape: (batch_size * num_indices)
+    flattened_indices = flatten_and_batch_shift_indices(indices * mask, target.size(1))
+    # Shape: (batch_size * num_indices, 1)
+    mask = mask.bool().view(-1)
+    # Shape: (batch_size * sequence_length, 1)
+    flattened_target = target.view(-1, 1)
+    # Shape: (nonzero_indices, 1)
+    unmasked_indices = flattened_indices[mask].unsqueeze(-1)
+
+    flattened_target.scatter_(0, unmasked_indices, fill_value)
```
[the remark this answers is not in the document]
OhadRubin: Yeah, it shouldn't? Anyway, i'll add that to the documentation if that wasn't clear.

allennlp/nn/util.py
dirkgr: I'm happy with whatever names you think make sense :-)

allennlp/nn/util.py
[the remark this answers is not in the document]
dirkgr: Unless absolutely necessary for performance, these functions should never modify their input.

allennlp/nn/util.py
```diff
@@ -1289,11 +1289,13 @@ def batched_index_fill(
     filled_target : `torch.Tensor`
         A tensor with shape (batch_size, sequence_length) where 'indices' are filled with `fill_value`
     """
+    target = target.clone()
```
dirkgr, line 1292: I guess that works, but I thought you could just replace `scatter_()` with `scatter()`?

OhadRubin: Sure

allennlp/nn/util.py
```diff
@@ -1307,7 +1306,7 @@ def masked_index_fill(
     # Shape: (nonzero_indices, 1)
     unmasked_indices = flattened_indices[mask].unsqueeze(-1)
 
-    flattened_target.scatter_(0, unmasked_indices, fill_value)
+    flattened_target.scatter(0, unmasked_indices, fill_value)
```
dirkgr, line 1309: I don't think that works. It has to be
```suggestion
    flattened_target = flattened_target.scatter(0, unmasked_indices, fill_value)
```
I'm hoping that this fails tests, because if it passes something is wrong with the tests ...

OhadRubin: haha ok they failed.

PR #4600, opened 2020-08-25
update image_loader and remove image-preprocessing step. 

jiasenlu:


sanjayss34:
Does this supersede the other PR now?

Also, I think the default batch size should be at least 8, if not 16. The main advantage of this script is taking advantage of a larger batch size than what is currently done in the dataset readers.

dirkgr:
> Does this supersede the other PR now?

No. I don't think this approach is going to work, because it changes the API again such that we can only use Detectron classes. I put a different version of this fix into your branch.

Issue #4422, opened 2020-06-30
Multi-task learning


dirkgr:
# Goals

* Be able to train VilBERT as in https://arxiv.org/abs/1912.02315 (12-in-1: Multi-Task Vision and Language Representation Learning)
* Be able to do a similar thing for NLP, i.e., fine-tune BERT for 12 tasks the same way

# Data Readers

We’ll write a `MultitaskReader` that takes a `Dict[str, DatasetReader]` as one of its `__init__()` arguments. This way we define the different datasets we want to read. Much like the `InterleavingDatasetReader`, this reader’s `_read()` method takes a `Dict[str, str]`, so that we can pass the right path to the right inner reader. Accordingly, the `*_data_path` keys in the config need to be dictionaries.

`MultitaskReader`’s `read()` method will return a `MultitaskDataset`. `MultitaskDataset` will act like `ChainDataset` (chaining the inner datasets into one large one), with two differences:
 * It adds an extra “task” field to each instance. This field is a `MetadataField` that contains the name of the task.
 * It gives access directly to the inner readers with two functions:
    * `tasks() -> List[str]`
    * `reader_for_task(task: str) -> DatasetReader`

# Data Loaders

We’ll write a `MultitaskDataLoader` that takes a `Dict[str, DataLoader]` as a `__init__()` argument that defines the different data loaders for each task. It also takes a `MultitaskScheduler` that determines which task gets batched next. The `__iter__()` method batches instances from the next task and returns them as a `TensorDict`.

When it does this, it makes one change to the `TensorDicts` that come from the inner data loaders: The instances will have a “task” field identifying the task, so the batch with those instances will contain a `List[str]` with the task in it. Since we know that one batch always comes from a single task, we replace that field with a single `str` containing the task name.

Note that this design makes it possible to define different batch sizes and shuffling options per task.

## Multitask Schedulers
Multitask schedulers are the only type of object that is genuinely new. As usual, we should have a base class with a a few implementations:
 * `“roundrobin”`: This should be the default scheduler that you get if you don’t specify anything. Optionally, this should take counts or a schedule, in case you want to do something more complicated. Counts are formatted as `Dict[str, int]`, specifying how often to repeat a given task. A schedule is a `List[str]`, for example `[“vqa”, “nlvr2”, “vqa”]`, to get that specific sequence of tasks.
 * `“stochastic”`, where you get a random task every time. Optionally you can specify weights for each task.

# Models
[...]

PR #4601, opened 2020-08-26
Add MultiTaskModel

matt-gardner:
Fixes the first part of #4422. 

This isn't *quite* ready for a full review yet, there are a few minor things left to implement.  But the basics are in place enough that a quick design review of what's here would be good, before I get too much farther.

This is currently based off of the `vision` branch, targeting 2.0, but there isn't really anything here that would stop us merging this directly into master, if we wanted.  I could go either way on what to do with this.


dirkgr:
If it can go into `master`, let's put it there. As long as we don't break backwards compatibility, we should put things in `master`.

allennlp/models/heads/classifier_head.py
```diff
@@ -0,0 +1,136 @@
...
+from allennlp.data import Vocabulary
+from allennlp.models.heads.head import Head
+from allennlp.models.model import Model
+from allennlp.modules import FeedForward, Seq2VecEncoder
+from allennlp.modules.seq2vec_encoders import ClsPooler
+from allennlp.training.metrics import CategoricalAccuracy
+
+
+@Head.register("classifier")
+class ClassifierHead(Model):
```
dirkgr, line 15: Usually a class gets registered under its own super type, but not here. I guess it does not matter?

allennlp/models/heads/classifier_head.py
```diff
@@ -0,0 +1,136 @@
...
+
+    vocab : `Vocabulary`
+        Used to get the number of labels, if `num_labels` is not provided, and to translate label
+        indices to strings in `make_output_human_readable`.
+    seq2vec_encoder : `Seq2VecEncoder`, optional (default = `ClsPooler`)
+        The input to this module is assumed to be a sequence of encoded vectors.  We use a
+        `Seq2VecEncoder` to compress this into a single vector on which we can perform
+        classification.  If nothing is provided, we will use a `ClsPooler`, which simply takes the
+        first element of the sequence as the single vector (and is the standard thing to do when you
+        are running a classifier on top of a transformer).
```
dirkgr, line 32: We also have a `BertPooler`, which (I think) does the same thing, but benefits from pre-trained weights.

I guess for generality, we have to have `ClsPooler` here as a default, though a lot of the time `BertPooler` would be the better option.

allennlp/models/heads/classifier_head.py
dirkgr: Now that I've read the definition of `Head`, it seems to me that this should derive from `Head`.

allennlp/models/multitask.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def get_forward_arguments(module: torch.nn.Module) -> List[str]:
+    signature = inspect.signature(module.forward)
+    arguments = []
+    for arg in signature.parameters:
+        if arg == "self":
+            continue
+        arguments.append(arg)
+    return arguments
```
dirkgr, line 29:
```suggestion
return [arg for arg in signature.parameters if arg != "self"]
```

allennlp/models/multitask.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class MultiTaskModel(Model):
+    """
+    initializer : `InitializerApplicator`, optional (default=`InitializerApplicator()`)
+        If provided, will be used to initialize the model parameters.
```
dirkgr, line 35: This will need some more documentation.

allennlp/models/heads/classifier_head.py
matt-gardner: Thanks, this was a copy-paste error.

allennlp/models/heads/classifier_head.py
matt-gardner: There's one pretrained 768x768 linear transform followed by a tanh that you're missing out on if you use a `ClsPooler` instead of a `BertPooler`.  I'd guess the effect of the transform is very small, if there is any, but that's just a guess at this point.  But you're right that we can't do that as a default, because we don't have a model name here.

allennlp/models/multitask.py
matt-gardner: Yeah, I thought I was going to need to do more checks inside the loop.  Simpler this way with only one check.

allennlp/models/multitask.py
matt-gardner: Why?  Oh, this reminds me that I need to make `self._heads` into a `ModuleDict`, so that it trains correctly.

Ok, I thought you were talking about the initializer parameter itself, then I went and looked at the code to fix the `ModuleDict` error and realized you were probably talking about the whole docstring in general.  Yes, definitely.

allennlp/models/multitask.py
matt-gardner: Ah, I remember.  We'll probably need to check for optional arguments, so that you still call a head even if you don't have the labels for it (e.g., at prediction time).  So this might need to move back into a for loop in the end.

matt-gardner: Turns out I solved this another way, so this method can stay as a list comprehension.

matt-gardner:
Ok, I think all of the main functionality here is implemented, and this is ready for review.  I still want to add more tests for the various code paths through the multitask model, but I'm not planning on adding new logic.

Question: any objection to just testing the implemented backbone and classifier head indirectly through the model test?  How much do we care about having unit tests that are specific to every class?  There isn't a lot of specific functionality in those classes that needs testing, it's more the multitask model logic that I'm worried about.

dirkgr:
> any objection to just testing the implemented backbone and classifier head indirectly through the model test?

No objection. I have thought for a long time that integration tests are the useful ones, not unit tests. Though if we really wanted to make that a philosophy, we'd need to pair it with reliable code coverage metrics.

matt-gardner:
@dirkgr, I know you've got a lot on your plate right now, but this is ready for a final review, if you didn't realize that.  I also am changing this to merge into the `vision` branch, as I will definitely need to make use of the new data code to get the multi-task data loading working correctly.

matt-gardner:
Also, I'm not blocked on getting this merged, so it's not a huge deal if you don't get to this for a few days, I was just letting you know that it's ready.

jbrry:
These changes look great Matt. Is there any use of me sketching out a config file for this to see if I can get everything integrated and make sure these components are working together as expected. Or to get this working fully are there still dependencies on some of the work in `multitask_data` which is still a work-in-progress?

matt-gardner:
@jbrry, they could be used right now in two cases:

1. With a separate mechanism for doing multi-task data loading, like what you have been working on in your repo.  We'll have one built-in eventually, but if you have your own, it _should_ work with what's here.
2. If you just want a simple way of building a multi-headed model, with uniform data loading.  Like, a model where you have a language model loss as a separate head as output along with a classification decision, and your dataset reader always gives you both supervision signals.  There's only a classification head here so far, but it shouldn't be hard to add more.

jbrry:
That's brilliant @matt-gardner. I had been working on a similar model in my repo. It's basically a `MultiTaskModel` whose `forward` method first embeds/encodes the text and then passes the encoded text to separate classifiers or `heads`, kind of like how it was done in the UDify implementation (where they use `decoders` in place of `heads`): https://github.com/Hyperparticle/udify/blob/b4c5b5e7ec8f079c005702239aca9cf916e13985/udify/models/udify_model.py#L37.

The new MTL changes scheduled for `v2.0` basically cover everything I was trying to do so it will be better for me to work with the more concrete abstractions of `MultiTaskModel`, `BackBone` and `Head` than what I had been doing previously.

Thanks for clarifying about the data loading. I should be good to go with that so!

allennlp/models/multitask.py
```diff
@@ -0,0 +1,195 @@
...
+from allennlp.models.model import Model
+from allennlp.models.heads import Head
+from allennlp.nn import InitializerApplicator
+
+
+def get_forward_arguments(module: torch.nn.Module) -> Set[str]:
+    signature = inspect.signature(module.forward)
+    return set([arg for arg in signature.parameters if arg != "self"])
+
+
```
jbrry, line 19: Does this model first need to be registered, e.g.:

```
@Model.register("multi_task_model")
```

so that it can be picked up in a config file and then all the subsequent arguments passed underneath that `model`? E.g.:

```
"model": {
  "type": "multi_task_model",
    "backbone": {
    ...
    }
```

matt-gardner: Hmm, yep, that's an oversight.

allennlp/models/multitask.py
```diff
@@ -0,0 +1,195 @@
...
+class MultiTaskModel(Model):
```
matt-gardner, line 20:
```suggestion
@Model.register("multitask")
class MultiTaskModel(Model):
```

allennlp/models/heads/head.py
```diff
@@ -0,0 +1,21 @@
+from allennlp.models.model import Model
+
+
+class Head(Model):
```
dirkgr, line 4: It would make some sense to re-write `TransformerMC` and `TransformerQA` to use this.

allennlp/models/multitask.py
```diff
@@ -0,0 +1,196 @@
...
+    # Parameters
+
+    vocab: `Vocab`
+    backbone: `Backbone`
+    heads: `Dict[str, Head]`
+    loss_weights: `Dict[str, float]`, optional (default = `equal weighting`)
+        If you want, you can specify a weight for each head, which we will multiply the loss by when
+        aggregating across heads.  This is equivalent in many cases to specifying a separate
+        learning rate per head, and just putting a weighting on the loss is much easier than
+        figuring out the right way to specify that in the optimizer.
```
dirkgr, line 47: Doing multi-task in this way doesn't break some important assumptions that optimizers make, does it?

For example, what happens to the momentum of parameters that aren't involved at all for a few batches?

allennlp/models/multitask.py
```diff
@@ -0,0 +1,196 @@
...
+        By default, the `MultiTaskModel` will try to infer which heads to run from the arguments
+        passed to `forward`.  During training, we will only run a head if we have all of its
+        arguments, including optional arguments, which typically means the argument is the
+        prediction target; if we don't have it, we typically can't compute a loss, so running during
+        training is pointless.  During evaluation, we will run all heads.
+
+        If you want to limit which heads are run during evaluation, or if the inference for which
+        task to run during training is incorrect (e.g., if your head has multiple optional
+        arguments, and only some are actually required to compute a loss), then you can use this
+        method to override our inference and force the use of whatever set of hads you want.
```
dirkgr, line 108:
```suggestion
        method to override our inference and force the use of whatever set of heads you want.
```

allennlp/models/multitask.py
```diff
@@ -0,0 +1,196 @@
...
+
+        self._allowed_arguments = allowed_arguments or {
+            "backbone": get_forward_arguments(backbone),
+            **{key: get_forward_arguments(heads[key]) for key in heads},
+        }
+        self._loss_weights = loss_weights or defaultdict(lambda: 1.0)
+        self._active_heads: List[str] = None
+        initializer(self)
+
+    def set_active_heads(self, active_heads: List[str]) -> None:
```
dirkgr, line 97: What is the use case for this?

allennlp/models/multitask.py
```diff
@@ -0,0 +1,196 @@
...
+        return metrics
+
+    @overrides
+    def make_output_human_readable(
+        self, output_dict: Dict[str, torch.Tensor]
+    ) -> Dict[str, torch.Tensor]:
+        output_dict = self._backbone.make_output_human_readable(output_dict)
+        for head_name in self._heads:
+            if self._active_heads is not None and head_name not in self._active_heads:
+                continue
```
dirkgr, line 188: If this respects `_active_heads`, then `get_metrics()` should do so too.

dirkgr (approved):
I only had some minor points. In fact, most are just thoughts, not requests for changes.

allennlp/models/multitask.py
```diff
@@ -0,0 +1,196 @@
...
+    A `MultiTaskModel` consists of a `Backbone` that encodes its inputs in some way, then a
+    collection of `Heads` that make predictions from the backbone-encoded inputs.  The predictions
+    of each `Head` are combined to compute a joint loss, which is then used for training.
+
+    This model works by taking `**kwargs` in `forward`, and passing the right arguments from that to
+    the backbone and to each head.  By default, we use `inspect` to try to figure out getting the
+    right arguments to the right modules, but we allow you to specify these arguments yourself in
+    case our inference code gets it wrong.
+
+    It the caller's responsibility to make sure that the backbone and all heads are compatible with
```
jbrry, line 32:
```suggestion
    It is the caller's responsibility to make sure that the backbone and all heads are compatible with
```

allennlp/models/heads/head.py
matt-gardner: Agreed, and I'm planning on doing something similar when I create a new baseline for ORB this quarter.

allennlp/models/multitask.py
matt-gardner: If the gradient is `None`, the momentum computation is skipped (I just checked what `Adam` does, and I assume the others do the same).  So this should do the right thing.

allennlp/models/multitask.py
matt-gardner: My breaking `e` key strikes again.  I need to get my keyboard fixed...

allennlp/models/multitask.py
matt-gardner: It's given in the docstring...  I assume you're asking if the uses specified in the docstring will ever happen in practice.  I don't know, as I don't know yet how exactly this will get used.  But, e.g., say you're doing TAPT, and you don't want the LM head to run during evaluation, to speed things up.  Or maybe you can use this in conjunction with a callback in the trainer to implement Jiasen's particular method of multi-task scheduling.

In short, I'm not totally sure yet, but I found myself hard-coding some assumptions about which heads to use in `forward`, and so I wanted to give a means of bypassing those assumptions if desired.

allennlp/models/multitask.py
matt-gardner: Yes, good catch, thanks.

CHANGELOG.md
```diff
@@ -13,10 +13,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   by adding a class-level variable called `authorized_missing_keys` to any PyTorch module that a `Model` uses.
   If defined, `authorized_missing_keys` should be a list of regex string patterns.
 - Added `FBetaMultiLabelMeasure`, a multi-label Fbeta metric. This is a subclass of the existing `FBetaMeasure`.
+- A `MultiTaskModel` and abstractions to use with it, including `Backbone` and `Head`.  The
```
matt-gardner, line 16: Note: the `vision` branch should probably have a separate `unreleased-2.0` or something, as this is currently mixing things that will get released from the master branch before 2.0 with 2.0 stuff.  I moved the 2.0 stuff I recognized up from the 1.1 release notes, but it'll just get stale again with the next release.

matt-gardner: @dirkgr, @epwalsh, I just updated the changelog to fix this; it might be good to double check the things I moved to make sure I didn't miss anything.  Feel free to change how I did this if you want to separate these things differently.

PR #4608, opened 2020-08-26
rename 'node_rank' to 'global_rank' in dataset reader 'DistributedInfo'

epwalsh:
I think `node_rank` is the wrong name to use in this context because this is supposed to represent the ID of the current **process** within the distributed group, not the ID of the node (which usually contains multiple processes). `global_rank` is more accurate.

### DIFF
--- a/allennlp/data/dataset_readers/dataset_reader.py
+++ b/allennlp/data/dataset_readers/dataset_reader.py
@@ -38,7 +38,7 @@ class WorkerInfo:
 @dataclass
 class DistributedInfo:
     """
-    Contains information about the process rank and world size when the reader is being
+    Contains information about the global process rank and total world size when the reader is being
     used within distributed training.
 
     From a `DatasetReader` this can be accessed with the [`get_distributed_info()`](#get_distributed_info) method.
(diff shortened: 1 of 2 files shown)


allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -38,20 +38,20 @@ class WorkerInfo:
 @dataclass
 class DistributedInfo:
     """
-    Contains information about the node rank and world size when the reader is being
+    Contains information about the process rank and world size when the reader is being
```
matt-gardner, line 41: You call it "process rank" here, but `global_rank` in the code.  Would `process_rank` be better?

epwalsh: I think having the word "global" is important to be clear what the rank represents. I'll update the terminology in the doc comment.

epwalsh, line 41:
```suggestion
    Contains information about the global process rank and total world size when the reader is being
```

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -38,20 +38,20 @@ class WorkerInfo:
...
     """
-    The total number of distributed nodes.
+    The total number of processes in the distributed group.
     """
 
-    node_rank: int
+    global_rank: int
     """
-    The 0-indexed ID of the current node.
+    The 0-indexed ID of the current process within the distributed group.
```
epwalsh, line 54:
```suggestion
    The 0-indexed ID of the current process within the distributed group. This will be between 0 and `world_size - 1`,
    inclusive.
```

Issue #4591, opened 2020-08-24
Better caching


OhadRubin:
Something along the lines of what fairseq has:
[Here](https://github.com/pytorch/fairseq/blob/master/fairseq/data/indexed_dataset.py
)
For example, in MMapIndexedDataset, they calculate the sizes for each field and then read exactly that size into an array.
[Here](https://github.com/allenai/allennlp/pull/4497#issuecomment-667314473), @matt-gardner mentions that this is the type of caching that you guys might implement. A database that just stores tensors directly.

matt-gardner:
There is some initial work along these lines [here](https://github.com/allenai/allennlp/blob/0a4765c4e978374a6665139ccf9f235f833c0e7f/allennlp/common/file_utils.py#L333-L404) (used in a dataset reader [here](https://github.com/allenai/allennlp/blob/0a4765c4e978374a6665139ccf9f235f833c0e7f/allennlp/data/dataset_readers/nlvr2.py#L114-L116)).  I think we definitely want to figure out how to include this in our main `Instance` creation pipeline, but that'll come a little bit later.  If you're interested in helping to design / work on this, contributions welcome.  This is a big enough piece of work that some kind of design document (or detailed github description) should come before any code is written.

@epwalsh, do we have an issue open already for this?  If not, I'll keep this open and add it to either the 2.0 or the performance milestone (maybe 2.0?).

epwalsh:
@matt-gardner I don't think we had a separate issue tracking caching until now, so adding this to 2.0 sounds good to me.

OhadRubin:
Should I create a .md page or a Google Docs? 
Anyway, most of the logic will use the to_tensor method of each field, right?
What I was thinking was:
For each instance, call to_tensor on all of its fields and then either:
1. use TensorCache.
2. call .numpy() and record the sizes, flatten the data and use logic from MMapIndexedDataset.
For fields that don't have to_tensor, like MetadataField, use jsonpickle.

epwalsh:
Hey @OhadRubin, I've already started a design document which I'll link to on this thread once it's ready to be seen.

epwalsh:
There is a good discussion here about `lmdb`: https://github.com/allenai/allennlp/pull/4578#issuecomment-680776842.

I'm working on an API design that would be agnostic to the backend or ser/deserialization method we use, so we can decide on that later.

OhadRubin:
[...]

PR #4611, opened 2020-08-27
Added MMapIndexedCache

OhadRubin:
#4591
@epwalsh,

### DIFF
--- a/allennlp/common/mmap.py
+++ b/allennlp/common/mmap.py
@@ -1,0 +1,251 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from functools import lru_cache
+import os
+import shutil
+import struct
+import numpy as np
+import torch
+from allennlp.data.fields import DataArray
+
+dtypes = {
+    1: np.uint8,
+    2: np.int8,
+    3: np.int16,
+    4: np.int32,
+    5: np.int64,
+    6: np.float,
+    7: np.double,
+    8: np.uint16,
+}
+
+
+def code(dtype):
+    for k in dtypes.keys():
+        if dtypes[k] == dtype:
+            return k
+    raise ValueError(dtype)
+
+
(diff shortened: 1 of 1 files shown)


allennlp/common/mmap.py
```diff
@@ -0,0 +1,251 @@
...
+        else:
+            self._dtype = np.int32
+
+    def add_instance(self, instance: Instance):
+        tensor_dict = instance.as_tensor_dict()
+        flattened_dict = self.flatten_dict(tensor_dict)
+        if not self._field_names:
+            self._field_names = list(sorted(flattened_dict.keys()))
+            assert self._field_names
+            # TODO: what if some instances have a different set of field names, i.e missing some, for test instances, we don't have supervision.....
```
OhadRubin, line 208: @epwalsh Do you have any ideas how to handle this?

OhadRubin: Optimally, we would like to write the names of the fields once, but the set of possible fields is different for train and test.

epwalsh: Yea, I think it's fine to write the names of each field with each instance. Should be a relatively small cost.

OhadRubin: Well, I want to solve it anyway if you don't mind haha.
I'll use a some sort  of enum (that can be updated) for each field name and write the enum index instead of the string itself.

OhadRubin:
Hey, @epwalsh, i'm in the process  of adapting the original fairseq code to our needs.
Right now the structure of the classes if kinda weird with _Writer and Index (from fairseq).
Can you review what I have here so far?

epwalsh:
Hey @OhadRubin, is there any reason why the `Index` class needs to be defined within the `MMapCacheReader` class?

OhadRubin:
No, that's one of the things from fairseq I kinda want to change...

OhadRubin:
Btw, during instance creation, is it possible to require that all fields exists, this will really help deserialization.
Where in "test set" situation, where there isn't any supervision, the user passes a empty field, or something along that.

PR #4617, opened 2020-08-31
remove duplicate padding calculations in collate fn

epwalsh:
There is no need to pass the result of `Batch.get_padding_lengths()` to `Batch.as_tensor_dict()` since, internally, `Batch.as_tensor_dict()` will call `Batch.get_padding_lengths()` anyway.


epwalsh:
I was looking at this line actually: https://github.com/allenai/allennlp/blob/vision/allennlp/data/batch.py#L123. `self.get_padding_lengths()` is called regardless.

matt-gardner:
Oh, hmm, yes, you're right.  Looks like it's been too long since I looked at that logic...  Nice catch!

Issue #4501, opened 2020-07-23
When extracting an archive fails half-way through, it corrupts the cache


dirkgr:
`cached_path()` can extract archives, and now even get files from inside the archive. But if the process gets killed during the extraction process, we are left with a directory that contains some of the extracted files, and we never catch back up.

PR #4622, opened 2020-09-01
improve robustness of cached_path when extracting archives

epwalsh:
Closes https://github.com/allenai/allennlp/issues/4501.

### DIFF
--- a/allennlp/common/file_utils.py
+++ b/allennlp/common/file_utils.py
@@ -202,7 +202,6 @@ def cached_path(
                     tar_file.close()
                 # Extraction was successful, rename temp directory to final
                 # cache directory.
-                os.makedirs(extraction_path)
                 os.replace(tmp_extraction_dir, extraction_path)
             finally:
                 shutil.rmtree(tmp_extraction_dir, ignore_errors=True)
--- a/allennlp/common/file_utils.py
+++ b/allennlp/common/file_utils.py
@@ -122,21 +122,32 @@ def cached_path(
     if isinstance(url_or_filename, PathLike):
         url_or_filename = str(url_or_filename)
 
+    file_path: str
+
     # If we're using the /a/b/foo.zip!c/d/file.txt syntax, handle it here.
     exclamation_index = url_or_filename.find("!")
     if extract_archive and exclamation_index >= 0:
         archive_path = url_or_filename[:exclamation_index]
-        archive_path = cached_path(archive_path, cache_dir, True, force_extract)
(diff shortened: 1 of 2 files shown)


allennlp/common/file_utils.py
```diff
@@ -161,29 +172,40 @@ def cached_path(
...
-                    zip_file.extractall(extraction_path)
-                    zip_file.close()
-            else:
-                tar_file = tarfile.open(file_path)
-                tar_file.extractall(extraction_path)
-                tar_file.close()
+
+            # We extract first to a temporary directory in case something goes wrong
+            # during the extraction process so we don't end up with a corrupted cache.
+            tmp_extraction_dir = tempfile.mkdtemp(dir=os.path.split(extraction_path)[0])
```
dirkgr, line 193: This is fine, but why didn't you use `with tempfile.TemporaryDirectory(...) as d:`?

allennlp/common/file_utils.py
```diff
@@ -161,29 +172,40 @@ def cached_path(
...
+                        zip_file.extractall(tmp_extraction_dir)
+                        zip_file.close()
+                else:
+                    tar_file = tarfile.open(file_path)
+                    tar_file.extractall(tmp_extraction_dir)
+                    tar_file.close()
+                # Extraction was successful, rename temp directory to final
+                # cache directory.
+                os.makedirs(extraction_path)
+                os.replace(tmp_extraction_dir, extraction_path)
```
dirkgr, line 206: One more comment: Doesn't the first line create all the directories, including the target one? If we fail between these two lines, I think it'll conclude that the archive has been extracted and won't do it again?

In summary, I think that first one has to be `os.makedirs(os.path.split(extraction_path)[0])`.

allennlp/common/file_utils.py
epwalsh: When used as a context manager, it will try to remove the temp directory when we exit the context (there is no option to disable that). But at that point - if the extraction was successful - we've already renamed the temp directory to it's final location, so the context manager will raise an exception when it tries to remove the temp directory that no longer exists.

allennlp/common/file_utils.py
epwalsh: Ah yes, good point.

epwalsh: Just fixed. After taking another look, I realized we can just remove that first line altogether because at that point the base directory should already exist.

PR #4625, opened 2020-09-04
Multitask data loading and scheduling

matt-gardner:
This is still very much in progress, and the way that I want to finish this implementation is very much dependent on some other major changes to the data pipeline.  I'm making a draft PR now, so that we have a place that's easy to see and talk about  a use case for those major data changes, to help with design choices when making those changes.

### DIFF
--- a/allennlp/data/data_loaders/multitask_data_loader.py
+++ b/allennlp/data/data_loaders/multitask_data_loader.py
@@ -16,6 +16,22 @@ from allennlp.data.vocabulary import Vocabulary
 @DataLoader.register("multitask")
 class MultiTaskDataLoader(DataLoader):
     """
+    A `DataLoader` intended for multi-task learning.  The basic idea is that you use a
+    `MultiTaskDatasetReader`, which takes a dictionary of `DatasetReaders`, keyed by some name.  You
+    use those same names for various parameters here, including the data paths that get passed to
+    each reader.  We will load each dataset and iterate over instances in them using a
+    `MultiTaskEpochSampler` and a `MultiTaskScheduler`.  The `EpochSampler` says how much to use
+    from each dataset at each epoch, and the `Scheduler` orders the instances in the epoch however
+    you want.  Both of these are designed to be used in conjunction with trainer `Callbacks`, if
+    desired, to have the sampling and/or scheduling behavior be dependent on the current state of
+    training.
+
+    While it is not necessarily required, this `DatasetReader` was designed to be used alongisde a
+    `MultiTaskModel`, which can handle instances coming from different datasets.  If your datasets
+    are similar enough (say, they are all reading comprehension datasets with the same format), or
+    your model is flexible enough, then you could feasibly use this `DataLoader` with a normal,
+    non-multitask `Model`.
+
     Registered as a `DataLoader` with name "multitask".
 
     # Parameters
--- a/allennlp/data/data_loaders/multitask_data_loader.py
+++ b/allennlp/data/data_loaders/multitask_data_loader.py
@@ -1,12 +1,13 @@
-from typing import Any, Dict, Iterator
+from typing import Any, Dict, Iterable, Iterator, List
 import itertools
 import math
 
 from allennlp.common import util
 from allennlp.data.batch import Batch
-from allennlp.data.data_loaders.data_loader import DataLoader, TensorDict, allennlp_collate
+from allennlp.data.data_loaders.data_loader import DataLoader, TensorDict
 from allennlp.data.data_loaders.multi_process_data_loader import MultiProcessDataLoader
 from allennlp.data.data_loaders.multitask_scheduler import MultiTaskScheduler
+from allennlp.data.data_loaders.multitask_epoch_sampler import MultiTaskEpochSampler
 from allennlp.data.dataset_readers.multitask import MultiTaskDatasetReader
 from allennlp.data.instance import Instance
 from allennlp.data.vocabulary import Vocabulary
@@ -99,6 +100,7 @@ class MultiTaskDataLoader(DataLoader):
         self.readers = reader.readers
         self.data_paths = data_path
         self.scheduler = scheduler
+        self.sampler = sampler
 
         self._batch_size = batch_size
         self._instances_per_epoch = instances_per_epoch
@@ -167,9 +169,9 @@ class MultiTaskDataLoader(DataLoader):
         # Finally, we take that combined list and yield `batch_size` batches from it.
         epoch_instances = self._get_instances_for_epoch()
         scheduled_instances = self.scheduler.order_epoch_instances(epoch_instances)
-        batch_instances = []
+        batch_instances: List[Instance] = []
         current_batch_size = 0
-        for dataset, instance in scheduled_instances.items():
+        for dataset, instance in scheduled_instances:
             current_batch_size += self._batch_size_multiplier.get(dataset, 1)
             if current_batch_size > self._batch_size:
                 batch = Batch(batch_instances)
@@ -204,10 +206,13 @@ class MultiTaskDataLoader(DataLoader):
         for loader in self._loaders.values():
             loader.index_with(vocab)
 
-    def _get_instances_for_epoch(self) -> Dict[str, Iterator[Instance]]:
+    def _get_instances_for_epoch(self) -> Dict[str, Iterable[Instance]]:
         if self._instances_per_epoch is None:
-            return {key: iter(loader) for key, loader in self._loaders}
-        dataset_proportions = self.scheduler.get_task_proportions()
+            return {
+                key: util.shuffle_iterable(loader.iter_instances())
+                for key, loader in self._loaders.items()
+            }
+        dataset_proportions = self.sampler.get_task_proportions(self._loaders)
         num_instances_per_dataset = {
             key: math.floor(proportion * self._instances_per_epoch)
             for key, proportion in dataset_proportions.items()
--- a/allennlp/data/data_loaders/multitask_data_loader.py
+++ b/allennlp/data/data_loaders/multitask_data_loader.py
@@ -6,7 +6,10 @@ from allennlp.common import util
 from allennlp.data.batch import Batch
 from allennlp.data.data_loaders.data_loader import DataLoader, TensorDict
 from allennlp.data.data_loaders.multi_process_data_loader import MultiProcessDataLoader
-from allennlp.data.data_loaders.multitask_scheduler import MultiTaskScheduler, RoundRobinScheduler
+from allennlp.data.data_loaders.multitask_scheduler import (
+    MultiTaskScheduler,
+    HomogeneousRoundRobinScheduler,
+)
 from allennlp.data.data_loaders.multitask_epoch_sampler import MultiTaskEpochSampler
 from allennlp.data.dataset_readers.multitask import MultiTaskDatasetReader
 from allennlp.data.instance import Instance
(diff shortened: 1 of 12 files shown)


matt-gardner:
Ok @dirkgr, I think this is ready for review.  Tests pass locally, with fake dataset readers and simple scheduling.  I haven't implemented the more complex scheduling stuff yet, but those should be really easy with these building blocks, and should wait for a subsequent PR.

I wasn't sure how to see coverage information locally, so I'm not sure if I'm missing anything big for the tests.  I think I tested enough, but if there's something you notice that isn't tested enough, let me know.

allennlp/data/data_loaders/multitask_data_loader.py
```diff
@@ -0,0 +1,264 @@
...
+    `MultiTaskDatasetReader`, which takes a dictionary of `DatasetReaders`, keyed by some name.  You
+    use those same names for various parameters here, including the data paths that get passed to
+    each reader.  We will load each dataset and iterate over instances in them using a
+    `MultiTaskEpochSampler` and a `MultiTaskScheduler`.  The `EpochSampler` says how much to use
+    from each dataset at each epoch, and the `Scheduler` orders the instances in the epoch however
+    you want.  Both of these are designed to be used in conjunction with trainer `Callbacks`, if
+    desired, to have the sampling and/or scheduling behavior be dependent on the current state of
+    training.
+
+    While it is not necessarily required, this `DatasetReader` was designed to be used alongisde a
```
dirkgr, line 29: `alongside`

allennlp/data/data_loaders/multitask_data_loader.py
```diff
@@ -0,0 +1,264 @@
...
+        If this is not `None`, it specifies how much of the batch an instance from each dataset
+        takes up.  That is, if this is 1 for every dataset (which is the default), then batch size
+        is computed as normal.  If dataset "A" has a value of 2 in this dictionary, than each
+        instance from dataset "A" counts as 2 instances for the purposes of computing batch size.
+        This option is available to you to account for the fact that some operations might be *much*
+        less costly than others (e.g., if you are multitasking a coref model with a simple document
+        classification model).  If you want finer-grained control than integer multiples, just scale
+        all values in the dictionary and the batch size accordingly (e.g., to achieve a 1:1.5 ratio,
+        double your batch size and use 2 and 3 as your multipliers).  If you use this, you're on
+        your own as far as figuring out how it interacts with optimization behavior.
```
dirkgr, line 66: What is this for?

If we need it, couldn't we make this easier to specify by making it a `Dict[str, float]`, and just saying we'll do it by the fraction of the sum of the given floats?

allennlp/data/data_loaders/multitask_data_loader.py
```diff
@@ -0,0 +1,264 @@
...
+        how batch size is computed.
+    scheduler: `MultiTaskScheduler`, optional (default = `RoundRobinScheduler`)
+        The `scheduler` determines how instances are ordered within an epoch.  By default, we'll
+        select one instance from each dataset in turn, trying to ensure as uniform a mix of datasets
+        as possible.
+    sampler: `MultiTaskEpochSampler`, optional (default = `None`)
+        Only used if `instances_per_epoch` is not `None`. If we need to select a subset of the data
+        for an epoch, this `sampler` will tell us with what proportion we should sample from each
+        dataset.  For instance, we might want to focus more on datasets that are underperforming in
+        some way, by having those datasets contribute more instances this epoch than other datasets.
```
dirkgr, line 55: Why do we need both of these?

allennlp/data/data_loaders/multitask_data_loader.py
```diff
@@ -0,0 +1,264 @@
...
+            # iterator, and it will call the lambda function each time it runs out of instances,
+            # which will produce a new shuffling of the dataset.
+            key: util.cycle_iterator_function(
+                # This default argument to the lambda function is necessary to create a new scope
+                # for the loader variable, so a _different_ loader gets saved for every iterator.
+                # Dictionary comprehensions don't create new scopes in python.  If you don't have
+                # this loader, you end up with `loader` always referring to the last loader in the
+                # iteration...  mypy also doesn't know what to do with this, for some reason I can't
+                # figure out.
+                lambda l=loader: util.shuffle_iterable(l.iter_instances())  # type: ignore
```
dirkgr, line 162: Everywhere else we make shuffling optional, because when you're debugging you want the same order every time.

allennlp/data/data_loaders/multitask_data_loader.py
```diff
@@ -0,0 +1,264 @@
...
+        num_workers: Dict[str, int] = None,
+        max_instances_in_memory: Dict[str, int] = None,
+        start_method: Dict[str, str] = None,
+        instance_queue_size: Dict[str, int] = None,
+        instance_chunk_size: Dict[str, int] = None,
+    ) -> None:
+        self.readers = reader.readers
+        self.data_paths = data_path
+        self.scheduler = scheduler
+        self.sampler = sampler
```
dirkgr, line 119: `"homogeneous_roundrobin"` should be the default here, no?

allennlp/data/data_loaders/multitask_data_loader.py
```diff
@@ -0,0 +1,264 @@
...
+        drop_last: bool = False,
+        num_workers: Dict[str, int] = None,
+        max_instances_in_memory: Dict[str, int] = None,
+        start_method: Dict[str, str] = None,
+        instance_queue_size: Dict[str, int] = None,
+        instance_chunk_size: Dict[str, int] = None,
+    ) -> None:
+        self.readers = reader.readers
+        self.data_paths = data_path
+        self.scheduler = scheduler
```
dirkgr, line 118: `"homogeneous_roundrobin"` should be the default here, no?

dirkgr (changes requested):
Do we have to have both the `*EpochSampler` and the `*Scheduler`? They seem very similar in purpose and might even have to share data, which is very hard to do if they are two different objects.

matt-gardner:
The scheduling and sampling operations are related, yes, but they are orthogonal and can be varied independently.  Making them one object in a world where we use subclasses to specify configuration options would mean having a class for every element in the cross product of the two axes, which seems suboptimal.  At some point, you really want to have the configuration done by composition, not inheritance.  It didn't seem worth it to introduce an intermediate class whose only job was to compose those two independent operations - we'd still have the separate Sampler and Scheduler classes, they would just go through some shim before getting passed to the data loader as one object.

allennlp/data/data_loaders/multitask_data_loader.py
matt-gardner: > What is this for?

I gave an example in the docstring, so I'm not sure what you're asking here.

> If we need it, couldn't we make this easier to specify by making it a `Dict[str, float]`, and just saying we'll do it by the fraction of the sum of the given floats?

Hmm, yeah, I was originally thinking that we'd have to have the batch size always be an integer, and so we'd need this to stay as integers.  But looking through the logic, it looks like having float batch sizes works alright for the internal logic here, and the floats would never get seen outside of this class, so we should be ok.  I'll change this.

allennlp/data/data_loaders/multitask_data_loader.py
matt-gardner: No, the default here should be `None`, because by default we use all instances.  This only should be set if `instances_per_epoch` is not `None`, and if you do that, it's better to make your configuration of how to sample explicit.

allennlp/data/data_loaders/multitask_data_loader.py
matt-gardner: Yep, good catch.

allennlp/data/data_loaders/multitask_data_loader.py
```diff
@@ -0,0 +1,292 @@
...
+            raise ValueError(
+                f"Mismatch between readers ({self.readers.keys()}) and data paths"
+                " ({self.data_paths.keys()})"
+            )
+        self._loaders = {key: self._make_data_loader(key) for key in self.readers}
+
+        # This stores our current iterator with each dataset, so we don't just iterate over the
+        # first k instances every epoch if we're using instances_per_epoch.  We'll grab instances
+        # from here each epoch, and refresh it when it runs out.  We only use this in the case that
+        # batchs_per_epoch is not None, but these iterators are lazy, so always creating them
```
dirkgr, line 169: You mean `instances_per_epoch`.

dirkgr (changes requested):
Doesn't the `*Scheduler` already completely determine the makeup of the batches?

I seems to me that there is a lot of machinery to make sure an epoch has a defined length. Do we lose a lot if we just say "If you're doing multi-task learning, all iterators cycle ad infinitum, and you have to specify `instances_per_epoch` all the time?

The way it is now, the shortest iterator determines the epoch length, though not exactly, since it depends on what the scheduler does. When you specify `instances_per_epoch`, the story is different though, and now it's determined by the sampler, where you can only change things at the beginning of the epoch.

What's supposed to happen if the sampler says we should use only `"task_a"`, and the scheduler only schedules instances from `"task_b"`? What will happen in this code is that the epoch ends up with 0 instances, which will be surprising. The problem is that I can specify that preference in two places, and the two places can contradict each other.

The concept of an "epoch" as the complete training set doesn't really make sense in this setting. It only works out that way in a few special cases. I don't think we have to go so far out of our way to preserve it.

matt-gardner:
You are misunderstanding what the scheduler does.  Perhaps this means we should use different names for these objects, or improve [docstrings](https://github.com/allenai/allennlp/pull/4625/files#diff-3da82e0cebfcb2509a64a331c06a506e27cbef052a912df27da981aef38446ffR16-R18), or something, or perhaps you're fixating too much on something you're familiar with.  The sampler says, "out of all of the datasets I have, which instances should I use this epoch?"  The scheduler says, "given this set of instances this epoch, how should I order them into batches?"  Those are different, independent questions.  The scheduler doesn't choose which instances to use, it only orders them.  Using the sampler, we create new, sliced iterators for each dataset at every epoch.  The scheduler's job is just to choose how to combine those iterators; it should always return everything from each one by the end of the epoch.

allennlp/data/data_loaders/multitask_data_loader.py
matt-gardner:
```suggestion
        # instances_per_epoch is not None, but these iterators are lazy, so always creating them
```

allennlp/data/data_loaders/multitask_epoch_sampler.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def get_task_proportions(self, data_loaders: Mapping[str, DataLoader]) -> Dict[str, float]:
+        """
+        Given a dictionary of `DataLoaders` for each dataset, returns what percentage of the
+        instances for the current epoch of training should come from each dataset.  The input
+        dictionary could be used to determine how many datasets there are (e.g., for uniform
+        sampling) or how big each dataset is (e.g., for sampling based on size), or it could be
+        ignored entirely.
+        """
+        raise NotImplementedError
```
dirkgr, line 25: This would be a safer API if the caller had to normalize the proportions so they sum up to 1. Right now, if this sums up to 2, your epoch will have 2x the expected number of instances.

dirkgr (changes requested):

dirkgr:
@matt-gardner, I get how it's supposed to work, but I'm not sure what the use case is that can't be addressed with just a scheduler on top of infinite iterators. More importantly, I anticipate the first issue we get about this is that someone has a situation where they really need sampler and scheduler to be the same object, or at least the two objects need to communicate a lot. This is exactly what happened with the trainer callbacks. We don't have nice solutions for that.

allennlp/data/data_loaders/multitask_epoch_sampler.py
matt-gardner: Done.

matt-gardner:
Use case: https://www.semanticscholar.org/paper/Dynamic-Sampling-Strategies-for-Multi-Task-Reading-Gottumukkala-Dua/6b34507e185ead4c482754774d9d87a974dcb169.

matt-gardner:
Note also that both the sampler and the scheduler have an `update_from_epoch_metrics` method, which allows coordination through a callback, though I currently don't really see any practical use of that method in the scheduler.

dirkgr:
So to do Stop&Go with this, I would set some low epoch size (so I get to re-evaluate quickly), have the sampler say `1` for the tasks in the `Go` state, and `0` for the tasks in `Stop` state, and then it doesn't matter much what kind of scheduler I use?

That's not how I think of it, but I guess it's workable.

dirkgr (approved):
This is a straight-up implementation of how you structured the problem in that paper. I fear that within three months we'll have someone that structured it differently and struggles to express that in terms of this API. But I can't prove that, so let's merge this as it is.

matt-gardner:
Yes, that's exactly how I would do Stop&Go with this.  This implementation additionally gives you flexibility to schedule things differently within an epoch, if you want to do that, as we did in our paper.  So it covers both papers.

Issue #4614, opened 2020-08-28
FromParams doesn't handle kwargs on base class


epwalsh:
## Checklist

- [x] I have verified that the issue exists against the `master` branch of AllenNLP.
- [x] I have read the relevant section in the [contribution guide](https://github.com/allenai/allennlp/blob/master/CONTRIBUTING.md#bug-fixes-and-new-features) on reporting bugs.
- [x] I have checked the [issues list](https://github.com/allenai/allennlp/issues) for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/allenai/allennlp/pulls) for existing proposed fixes.
- [x] I have checked the [CHANGELOG](https://github.com/allenai/allennlp/blob/master/CHANGELOG.md) and the [commit log](https://github.com/allenai/allennlp/commits/master) to find out if the bug was already fixed in the master branch.
- [x] I have included in the "Description" section below a traceback from any exceptions related to this bug.
- [x] I have included in the "Related issues or possible duplicates" section beloew all related issues and possible duplicate issues (If there are none, check this box anyway).
- [x] I have included in the "Environment" section below the name of the operating system and Python version that I was using when I discovered this bug.
- [x] I have included in the "Environment" section below the output of `pip freeze`.
- [x] I have included in the "Steps to reproduce" section below a minimally reproducible example.

## Description

If your create class that inherits from `FromParams`, and this class takes `**kwargs` in its constructor, then calling the `.from_params` method on this class will fail. See "Steps to reproduce" below.

## Related issues or possible duplicates

- This issue is similar: https://github.com/allenai/allennlp/issues/4592

## Environment

OS: OS X

Python version: Python 3.7

<details>
<summary><b>Output of <code>pip freeze</code>:</b></summary>
<p>

```
-e git+git@github.com:allenai/allennlp.git@e840a589afc4bfdac0165a8650145259a7603807#egg=allennlp
-e git+git@github.com:allenai/allennlp-models.git@2e5444997f416c112c0c6e02cc33175c24982049#egg=allennlp_models
apipkg==1.5
appdirs==1.4.4
appnope==0.1.0
attrs==20.1.0
backcall==0.2.0
black==20.8b1
bleach==3.1.5
blis==0.4.1
boto3==1.14.51
botocore==1.17.51
catalogue==1.0.0
certifi==2020.6.20
chardet==3.0.4
click==7.1.2
codecov==2.1.9
colorama==0.4.3
conllu==4.0
[...]

PR #4629, opened 2020-09-06
Improve handling of **kwargs in FromParams

eladsegal:
Continue fixing https://github.com/allenai/allennlp/issues/4614.

### DIFF
--- a/allennlp/common/from_params.py
+++ b/allennlp/common/from_params.py
@@ -126,6 +126,10 @@ def infer_params(cls: Type[T], constructor: Callable[..., T] = None) -> Dict[str
         if param.kind == param.VAR_KEYWORD:
             has_kwargs = True
 
+    args_param = parameters.get("args", None)
+    if args_param and args_param.kind == args_param.VAR_POSITIONAL:
+        del parameters["args"]
+
     if not has_kwargs:
         return parameters
 
--- a/allennlp/common/from_params.py
+++ b/allennlp/common/from_params.py
@@ -504,10 +504,10 @@ class FromParams:
     because sometimes we want non-Registrable classes to be instantiatable from_params.
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self):
         """
-        A constructor so that argument inference on super classes will stop here
-        and not on `object`'s constructor which has unwanted *args.
+        A bare constructor so that argument inference on super classes will stop here
+        and not on `object`'s constructor which has unwanted *args and **kwargs.
         """
         pass
 
--- a/allennlp/common/from_params.py
+++ b/allennlp/common/from_params.py
@@ -504,12 +504,12 @@ class FromParams:
     because sometimes we want non-Registrable classes to be instantiatable from_params.
     """
 
-    def __init__(self):
+    '''def __init__(self):
         """
         A bare constructor so that argument inference on super classes will stop here
         and not on `object`'s constructor which has unwanted *args and **kwargs.
         """
-        pass
+        pass'''
 
     @classmethod
     def from_params(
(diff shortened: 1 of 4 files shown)


tests/common/from_params_test.py
```diff
@@ -961,3 +961,16 @@ def __init__(self, a: int, b: str = None, **kwargs) -> None:
...
+    def test_from_params_child_has_kwargs_base_implicit_constructor(self):
+        class Foo(FromParams):
+            pass
+
+        class Bar(Foo):
+            def __init__(self, a: int, **kwargs) -> None:
+                self.a = a
+
+        Bar.from_params(Params({"a": 2}))
+
```
epwalsh, line 974: I think this is redundant.

```suggestion
```

epwalsh (commented):
Thanks @eladsegal! This LGTM, but I'm going to defer to @matt-gardner for the official review since he has the most context here.

allennlp/common/from_params.py
```diff
@@ -126,6 +126,10 @@ def infer_params(cls: Type[T], constructor: Callable[..., T] = None) -> Dict[str
         if param.kind == param.VAR_KEYWORD:
             has_kwargs = True
 
+    args_param = parameters.get("args", None)
```
matt-gardner, line 129: Hmm, the point of line 116 above was to handle this case.  Looks like I didn't do it well enough, because it doesn't handle intermediate classes that also have no constructor.

I was trying to decide if this is really what we want to do with a `*args` parameter, and I think it probably is; we don't really have any basis for constructing anything in `*args`, nor do we have a good way of passing in those arguments.

So, yes, I think this is a good change, but can you also do two small other changes, to handle this whole thing more cleanly?  First, remove the check for `FromParams` on line 116 above, and second, add a bare constructor to `FromParams`, with a docstring saying that this exists so that argument inference on super classes hits a bare constructor in the base case, instead of the `object` constructor, which has `*args` and `**kwargs` which we don't want.

matt-gardner:
Also, can you modify the line in the CHANGELOG about this fix, adding a mention of handling `*args`?

matt-gardner:
@epwalsh, crazy idea: would it work to have a github action that adds a dummy changelog change if the changelog check fails, so that we can modify the file ourselves on github really easily?

epwalsh:
> @epwalsh, crazy idea: would it work to have a github action that adds a dummy changelog change if the changelog check fails, so that we can modify the file ourselves on github really easily?

Hmm we'd have to create a service account to make commits from GH Actions. That's not too hard, but then it attributes the PR to the service account as well, which is not ideal IMO because it makes the commit history kind of ugly (like this https://github.com/allenai/allennlp-docs/commits/master).

It also might be tricky to make a good dummy commit. Like, what exactly would that dummy commit include?

matt-gardner:
Yeah, ok, the complexity of what you've said already is enough to make me discard the idea as too crazy 😅.

allennlp/common/from_params.py
eladsegal: So just to makes things clear:
- Adding a bare constructor to `FromParams` solves the problem of reverting to `object`'s constructor.
- Keeping the code which explicitly removes `*args` is good as well because we don't use it in `FromParams`, even if it is explicit in the class constructor. For this reason, I remove any VAR_POSITIONAL and not only ones named `args`.

eladsegal: @matt-gardner Adding a bare constructor to `FromParams` is causing failures when initializing `Optimizer`s, because it gets to the `FromParams` constructor instead of `torch.optim.*`'s.
Changing the inheritance order fixes this issue, but I am not sure what side effects it might have.
Test-wise, it only causes type check (https://github.com/allenai/allennlp/pull/4629/checks?check_run_id=1095095838) to fail. Didn't check how to fix it.

matt-gardner (approved):
Ack, sorry @eladsegal, I didn't expect that to cause other errors, and I didn't mean to have you figure that out.  Thanks for doing this!

PR #4639, opened 2020-09-14
VQAv2

dirkgr:


allennlp/data/dataset_readers/vqav2.py
```diff
@@ -435,11 +443,12 @@ class Split(NamedTuple):
         # It would be much easier to just process one image at a time, but it's faster to process
         # them in batches. So this code gathers up instances until it has enough to fill up a batch
         # that needs processing, and then processes them all.
-        question_dicts = self.shard_iterable(questions["questions"])
-        image_prefix = "COCO_%s_" %questions['data_subtype']
+        question_dicts = list(self.shard_iterable(questions["questions"]))
```
dirkgr, line 446: Why does this need to be a list?

allennlp/data/dataset_readers/vqav2.py
```diff
@@ -0,0 +1,563 @@
...
+                + "Questions_Val_abstract_v002.zip!OpenEnded_abstract_v002_val2015_questions.json",
+            ),
+            "abstract_scenes_test": Split(
+                None,
+                scene_base
+                + "Questions_Test_abstract_v002.zip!OpenEnded_abstract_v002_test2015_questions.json",
+            ),
+        }
+
+        if isinstance(split_name, list):
```
dirkgr, line 419: @matt-gardner, in Python, strings are lists 🙄. I'll change this around.

matt-gardner: Oh man, that's unexpected...

allennlp/data/dataset_readers/vqav2.py
matt-gardner: You iterate over it more than once.  The test failed to return all instances because the iterator was getting consumed (in the dataset reader test, and in the model test, I was seeing one instance instead of 3).  I didn't dig into `shard_iterable` to figure out why it was getting consumed, but consuming the whole iterator and making it a list fixed the problem.  Something funny is going on inside `shard_iterable` when you iterate over it more than once.

Issue #4642, opened 2020-09-16
Immediate removal of temp dir in load_archive


mulhod:
**Is your feature request related to a problem? Please describe.**
In the `archival.load_archive` function, the archive file (if it is not a directory, that is) is extracted to a temporary directory and that directory’s removal is postponed till the interpreter exits (via `atexit`). This is done because the extracted directory’s contents are used outside this function in some cases. I have a use case where I would like to remove this temporary directory as soon as possible (by the end of `load_archive`). I have long-running interpreter sessions involving many models being loaded (and used for inference only) and I don’t want such temporary directories to be retained if they are not needed.

Currently, I’m working around this by extracting the archive file myself and then passing in the extracted directory path, calling `load_archive` with that path, then manually cleaning up.

**Describe the solution you'd like**
Ideally I would like to be able to call `load_archive` with an archive file directly and have the temporary directory that it creates removed without postponement. For example, a `postpone_cleanup` keyword argument in `load_archive` (and in some other methods that call it) that is set to `True` by default so as not to change anything about normal operation, but can be set to `False` if a user wants this behavior.

**Describe alternatives you've considered**
The alternative is doing this manually.

**Additional context**
I will be submitting a PR to address this issue.

epwalsh:
Closing per #4643

PR #4643, opened 2020-09-16
Feature/prevent temp directory retention

mulhod:
Addresses #4642.

I have added a couple tests, but honestly I'm not very satisfied about them due to their hacky qualities. I would prefer not to resort to logging statements to get the name of the temporary directory, but I don't see another way to do it.

### DIFF
--- a/allennlp/models/archival.py
+++ b/allennlp/models/archival.py
@@ -3,7 +3,6 @@ Helper functions for archiving models and restoring archived models.
 """
 from os import PathLike
 from typing import NamedTuple, Union
-import atexit
 import logging
 import os
 import tempfile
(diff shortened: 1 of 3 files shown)


allennlp/models/model.py
```diff
@@ -293,7 +294,9 @@ def _load(
         # stored in our weights.  We don't need any pretrained weight file anymore, and we don't
         # want the code to look for it, so we remove it from the parameters here.
         remove_pretrained_embedding_params(model_params)
-        model = Model.from_params(vocab=vocab, params=model_params)
+        model = Model.from_params(
+            vocab=vocab, params=model_params, postpone_cleanup=postpone_cleanup
```
epwalsh, line 298: Why does this need to be passed to `from_params`?

epwalsh (changes requested):
Hey @mulhod, this seems reasonable.

But to be honest, having the clean up happen `atexit` is strange to me. I have no idea what use case requires it.

@matt-gardner do you know?

matt-gardner:
`atexit` was added because of a [bug](https://github.com/allenai/allennlp/issues/2180) that dealt with how `load_archive` interacted with `files_to_archive`.  `files_to_archive` has been removed, so that bug isn't applicable anymore, I believe, and so `atexit` can probably be safely removed, if it's a problem for some reason.  (I didn't look too closely at what's going on in this PR.)

epwalsh:
Okay, sounds like there's no reason to keep the `atexit` hook. @mulhod can you update your PR to always clean up the temp directories immediately?

mulhod:
Sure, no problem.

allennlp/models/model.py
mulhod: At the end of `load_archive`, `Model.load` is called, which calls `model_class._load` in `model.py`, which calls `Model.from_params`, etc. At some point it leads right back to `load_archive`.

Here's the sequence of calls I see when I was debugging via `pudb`:
```
load_archive archival.py:183
load model.py:396
_load model.py:296
from_params from_params.py:595
from_params from_params.py:628
from_archive model.py:440
load_archive archival.py:160
```

In order to ensure that the second call to `load_archive` also resulted in the same behavior, it was necessary to pass the argument through. I lack the details about why this would be necessary, however.

mulhod:
I see that a similar `atexit` usage occurs in `allennlp/tools/archive_surgery.py`.
```python
# Extract archive to temp dir
tempdir = tempfile.mkdtemp()
with tarfile.open(archive_file, "r:gz") as archive:
    archive.extractall(tempdir)
atexit.register(lambda: shutil.rmtree(tempdir))
```

It's just a utility script, but for the sake of consistency I could fix it there as well here or in a separate PR.

epwalsh (changes requested):
Thanks @mulhod!

My only concern is that the chunk of code between opening the temp directory and removing it should be in a `try` block, with the removal in a `finally` block so that it is always removed even if there's an error during the loading process. Just like we do here: https://github.com/allenai/allennlp/blob/master/allennlp/common/file_utils.py#L193-L207

epwalsh:
As for the tools script, personally I've never used it. @matt-gardner is it worth keeping that script? Text editors like `vim` already have built in support for editing archives directly.

allennlp/models/model.py
matt-gardner: Seeing this comment raised a flag for me that made me look into it.  There's no issue here, though.  This happened because the model type was "from_archive".  So, we first try to load the initial archive (the first call to `load_archive`).  When we're loading that archive, we see a path to another archive file specified (inside of the model params), which is where the nested call to `load_archive` happens.

But, I'm glad that there's a better solution that removes the need for this line.

matt-gardner:
That script just batches together a few shell commands to make it a bit easier to modify a model archive.  I don't think it's _that_ commonly used, but if I had remembered it when modifying all of our hosted models during the 1.0 upgrade, it would have made my life slightly easier.  I don't think it hurts anything to have it there, and I don't think it's necessary to remove the call to `atexit` there.  It's a simple script that doesn't interact with the rest of the code, so the `atexit` doesn't cause issues like it did here.

matt-gardner:
Oh, I notice you're saying that `vim` has built-in support for editing tar files; I know I can read them, but whenever I've tried to modify them, it tells me it's read only.  Is there some magic I need to do to get that to work?

allennlp/models/archival.py
```diff
@@ -158,43 +157,40 @@ def load_archive(
...
+            serialization_dir=serialization_dir,
+            cuda_device=cuda_device,
+        )
+
+        return Archive(model=model, config=config)
+
+    finally:
+        if tempdir is not None:
+            logger.info(f"removing temporary unarchived model dir at {tempdir}")
+            shutil.rmtree(tempdir, ignore_errors=True)
```
epwalsh, line 196: Does this work with the return statement above the `finally` block? I would think the `finally` block would be skipped if this method successfully returned.

Either way, I think it's more clear to have the return statement after the `finally` block.

mulhod: Sure, changed it.

mulhod: I think that technically whatever is in the `finally` block will be executed even when there is a `return` in the `try` block. This is according the [Python docs](https://docs.python.org/3/reference/compound_stmts.html#the-try-statement). But, in any case, it's clearer, as you mentioned.

epwalsh:
> Oh, I notice you're saying that `vim` has built-in support for editing tar files; I know I can read them, but whenever I've tried to modify them, it tells me it's read only. Is there some magic I need to do to get that to work?

@matt-gardner Hmm oh, yea you're right. It is read only with vim (https://www.vim.org/scripts/script.php?script_id=522).

epwalsh (approved):
Looks great! I'll merge once tests pass 👌

PR #4645, opened 2020-09-16
improve how cached_path extracts archives

epwalsh:
Currently, when `cached_path()` is called with `extract_archive=True` on a local archive file, the archive is extracted in the same directory. For example,

```python
>>> cached_path("~/vocab.tar.gz", extract_archive=True)
#> ~/vocab-tar-gz-extracted
```

But I don't think this is a good way to do it because we shouldn't be messing with peoples' local file system outside of the designated cache directory.

This PR changes the behavior so that local archives are extracted to unique sub-directories of the cache root. The name of the sub-directory is derived in a similar way that we derive the cached path for remote resources: with a hash of the path and a hash of the ETag. Except, instead of an ETag in this case, we use the last modified time of the local file.

This approach has the added benefit that if a local archive is modified after an initial call to `cached_path()`, subsequent calls to `cached_path()` will automatically know to re-extract.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Ignore `*args` when constructing classes with `FromParams`.
-- Ensured some consistency in the types of the values that metrics return
+- Ensured some consistency in the types of the values that metrics return.
 - Fix a PyTorch warning by explicitly providing the `as_tuple` argument (leaving
   it as its default value of `False`) to `Tensor.nonzero()`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,10 +17,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `transformers` dependency updated to version 3.1.0.
+- When `cached_path` is called on a local archive with `extract=True`, the archive is now extracted into a unique subdirectory of the cache root instead of a subdirectory of the archive's directory. The extraction directory is also unique to the modification time of the archive, so if the file changes, subsequent calls to `cached_path` will know to re-extract the archive.
 
 ### Fixed
 
-- Ignore *args when constructing classes with `FromParams`.
+- Ignore `*args` when constructing classes with `FromParams`.
 - Ensured some consistency in the types of the values that metrics return
 
 ## [v1.1.0](https://github.com/allenai/allennlp/releases/tag/v1.1.0) - 2020-09-08
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `transformers` dependency updated to version 3.1.0.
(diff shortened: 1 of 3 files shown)


epwalsh:
@dirkgr you probably have the most context to review this.

CHANGELOG.md
```diff
@@ -19,11 +19,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 - `transformers` dependency updated to version 3.1.0.
+- When `cached_path` is called on a local archive with `extract_archive=True`, the archive is now extracted into a unique subdirectory of the cache root instead of a subdirectory of the archive's directory. The extraction directory is also unique to the modification time of the archive, so if the file changes, subsequent calls to `cached_path` will know to re-extract the archive.
 
 ### Fixed
 
-- Ignore *args when constructing classes with `FromParams`.
-- Ensured some consistency in the types of the values that metrics return.
+- Ignore `*args` when constructing classes with `FromParams`.
+- Ensured some consistency in the types of the values that metrics return
```
epwalsh, line 27:
```suggestion
- Ensured some consistency in the types of the values that metrics return.
```

allennlp/common/file_utils.py
```diff
@@ -46,15 +46,15 @@
     )
 
 
-def url_to_filename(url: str, etag: str = None) -> str:
+def resource_to_filename(resource: str, etag: str = None) -> str:
```
epwalsh, line 49: I was a little hesitant to change the name of this function because it's technically not a "private" function, but I don't believe it was ever intended to be used outside of this module. Maybe we should give it a leading underscore to make it clear that it's supposed to be "private".

dirkgr: Yes, underscore is the way to go.

allennlp/common/file_utils.py
```diff
@@ -158,17 +161,21 @@ def cached_path(
             # This is the path the file should be extracted to.
             # For example ~/.allennlp/cache/234234.21341 -> ~/.allennlp/cache/234234.21341-extracted
             extraction_path = file_path + "-extracted"
+            extraction_lock_path = file_path + ".lock"
```
dirkgr, line 164: I think this means we use the same lock file for downloading and extracting. I don't think that's ideal.

dirkgr (changes requested):

allennlp/common/file_utils.py
epwalsh: Right. I thought a little bit about this, and I think it's actually the right thing to do.

If we used a different lock file for extraction, it's possible for one process to be writing to an archive file while another process starts trying to extract it, resulting in an extraction error.

Using the same lock file avoids this situation.

epwalsh: Then again... since we download the archive to a temp file first then just rename it, that situation might be impossible anyway. But I don't see any downside to using the same lock file.

epwalsh: Then again again... I guess one downside to using the same lock file is this:

Say you have one process (process "A") extracting the cached archive, and another process (process "B") that just wants the path to the cached archive, but doesn't want it extracted.

If process A is extracting the archive while process B is just trying to get the path to the archive, process B will have to wait until process A is finished extracting.

epwalsh: Ok I changed to a different lock file for extraction ✅

dirkgr (approved):

PR #4648, opened 2020-09-17
smooth out release process

epwalsh:
This PR does two things:

1. Adds a script `scripts/release_notes.py` which automatically prepares markdown release notes from the `CHANGELOG` and commit history.
2. Updates `RELEASE_PROCESS.md` to include using the new script, and also clarifies some other things.

### DIFF
--- a/scripts/release_notes.py
+++ b/scripts/release_notes.py
@@ -1,0 +1,61 @@
+# encoding: utf-8
+
+"""
+Prepares markdown release notes for GitHub releases.
+"""
+
+import os
+from typing import List
+
+from allennlp.version import VERSION
+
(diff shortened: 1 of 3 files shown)


scripts/release_notes.py
```diff
@@ -0,0 +1,61 @@
...
+
+import os
+from typing import List
+
+from allennlp.version import VERSION
+
+TAG = os.environ["TAG"]
+
+
+ADDED_HEADER = "### Added 🎉"
```
AkshitaB, line 15: Never considered the possibility of these icons in Python!

AkshitaB (approved):

PR #4655, opened 2020-09-21
add build-vocab command

epwalsh:


allennlp/commands/build_vocab.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+
+        subparser.add_argument(
+            "-o",
+            "--overrides",
+            type=str,
+            default="",
+            help=(
+                "a json(net) structure used to override the experiment configuration, e.g., "
+                "'{\"iterator.batch_size\": 16}'.  Nested parameters can be specified either"
```
matt-gardner, line 49: You might consider a more meaningful help text here.
```suggestion
                "'{\"vocabulary.min_count\": {\"labels\": 10}}'.  Nested parameters can be specified either"
```

(not sure if using dots the whole way would work...  maybe it would)

allennlp/commands/build_vocab.py
```diff
@@ -0,0 +1,100 @@
...
+
+    params = Params.from_file(args.param_path)
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+        # Serializes the vocab to 'tempdir/vocabulary'.
+        make_vocab_from_params(params, temp_dir)
+
+        # We create a temp file to archive the vocab to in case something goes wrong
+        # while creating the archive so we're not left with a corrupted file.
+        temp_archive = tempfile.NamedTemporaryFile(
```
matt-gardner, line 77: This whole "create inside a temp file" seems like a complex piece of code that happens in a few places and could maybe be shared.  Then again, you've got some vocabulary-specific stuff in here.  What happens in other places when we save the vocab?  Is there similar logic there?  Just a note to consider whether it's worth sharing this logic somewhere.  Might not be worth it.

matt-gardner (approved):

allennlp/commands/build_vocab.py
epwalsh: Ah, good point. I forgot we have the `CacheFile` context manager that already handles this logic.

allennlp/common/file_utils.py
```diff
@@ -386,14 +386,16 @@ class CacheFile:
     goes wrong while writing to the temporary file, it will be removed.
     """
 
-    def __init__(self, cache_filename: Union[Path, str], mode="w+b") -> None:
+    def __init__(
```
epwalsh, line 389: Maybe we should call this class something other than `CacheFile`.

PR #4672, opened 2020-09-25
The truncation setting doesn't do anything anymore

dirkgr:


allennlp/data/tokenizers/pretrained_transformer_tokenizer.py
```diff
@@ -230,12 +221,15 @@ def tokenize(self, text: str) -> List[Token]:
...
+        if max_length is not None and self._add_special_tokens:
+            max_length -= self.num_special_tokens_for_sequence()
+
         encoded_tokens = self.tokenizer.encode_plus(
             text=text,
             add_special_tokens=False,
-            max_length=self._max_length,
+            max_length=max_length,
             stride=self._stride,
-            truncation=self._truncation_strategy if self._max_length is not None else False,
```
epwalsh, line 238: Are you sure this doesn't work anymore? It should: https://github.com/huggingface/transformers/blob/3323146e904a1092def4e8527de9d2a7479c1c14/src/transformers/tokenization_utils_base.py#L1990

dirkgr: It only makes a difference when you pass in more than one string, which we never do.

epwalsh: Oh I see

epwalsh (approved):
LGTM! I think it's worth noting in the CHANGELOG that we've removed the `truncation_strategy` option and why

PR #4673, opened 2020-09-25
Small optimization when zero-ing gradients

epwalsh:
See https://youtu.be/9mS1fIYj1So?t=540

This is from the tweet you sent me @dirkgr

### DIFF
--- a/allennlp/predictors/predictor.py
+++ b/allennlp/predictors/predictor.py
@@ -109,7 +109,11 @@ class Predictor(Registrable):
             )
 
             loss = outputs["loss"]
-            self._model.zero_grad()
+            # Zero gradients.
+            # NOTE: this is actually more efficient than calling `self._model.zero_grad()`
+            # because it avoids a read op when the gradients are first updated below.
+            for p in self._model.parameters():
+                p.grad = None
             loss.backward()
 
         for hook in hooks:
--- a/allennlp/training/trainer.py
+++ b/allennlp/training/trainer.py
@@ -581,7 +581,12 @@ class GradientDescentTrainer(Trainer):
             self._batch_num_total += 1
             batch_num_total = self._batch_num_total
 
-            self.optimizer.zero_grad()
+            # Zero gradients.
+            # NOTE: this is actually more efficient than calling `self.optimizer.zero_grad()`
+            # because it avoids a read op when the gradients are first updated below.
+            for param_group in self.optimizer.param_groups:
+                for p in param_group["params"]:
+                    p.grad = None
 
             batch_group_outputs = []
             for batch in batch_group:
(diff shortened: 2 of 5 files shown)


allennlp/predictors/predictor.py
```diff
@@ -109,7 +109,11 @@ def get_gradients(self, instances: List[Instance]) -> Tuple[Dict[str, Any], Dict
             )
 
             loss = outputs["loss"]
-            self._model.zero_grad()
+            # Zero gradients.
+            # NOTE: this is actually more efficient than calling `self._model.zero_grad()`
+            # because it avoids a read op when the gradients are first updated below.
+            for p in self._model.parameters():
```
dirkgr, line 115: Here, the parameters come from the model. In other places they come from the optimizer. Does that matter?

allennlp/training/trainer.py
```diff
@@ -581,7 +581,12 @@ def _train_epoch(self, epoch: int) -> Dict[str, float]:
             self._batch_num_total += 1
             batch_num_total = self._batch_num_total
 
-            self.optimizer.zero_grad()
+            # Zero gradients.
+            # NOTE: this is actually more efficient than calling `self.optimizer.zero_grad()`
+            # because it avoids a read op when the gradients are first updated below.
+            for param_group in self.optimizer.param_groups:
+                for p in param_group["params"]:
+                    p.grad = None
```
dirkgr, line 589: Do the param groups necessarily cover all parameters?

dirkgr (approved):
I see that all of these places are subtly different, so maybe it doesn't make sense to do this, but I wanted to float the idea of putting this into `util` somewhere, or maybe into `Model`, some common place?

allennlp/training/trainer.py
epwalsh: All parameters that require gradients, yes: https://github.com/allenai/allennlp/blob/master/allennlp/training/trainer.py#L1128-L1131

allennlp/predictors/predictor.py
epwalsh: I don't think it matters much. The only difference here is that we may be setting `p.grad` to `None` for some parameters where `p.grad` is already `None`.

epwalsh:
> I wanted to float the idea of putting this into util somewhere, or maybe into Model, some common place?

Yea, I was thinking about overriding the `zero_grad` method on `Model`, which would work fine there, but it's tricky to override this on the `Optimizer` class because our base optimizer doesn't actually inherit from the torch `Optimizer`. So that led me to considering just having a util function in `nn.util`, but then we'd need a separate function for both the model and the optimizer, and at that point it seemed like it was more trouble than it's worth. I also wanted it to be very obvious what we are doing in the training loop. Having another indirection with a util function call would probably not help that.

Issue #4653, opened 2020-09-20
Closing the TensorBoard writer


mahnerak:
I had noticed that sometimes tensorboard fails to properly read the `tfevents` files.
They are mainly in case of runs which I decided to interrupt the training (not sure about other abortion scenarios).
Then, I wanted to make sure the tensorboard writer is properly closed. I found the `close()` is called only here, but not inside any `with` or `finally` construct:
https://github.com/allenai/allennlp/blob/fbd2ccca1dd767e919a62d344c6b6569fc38227a/allennlp/training/trainer.py#L970
Here we can see the writer is closed just by calling `.close()` so there's no guarantee that the data is going to be flushed.

Is there any reason for this implementation? Does this need to be fixed?

epwalsh:
Hi @mahnerak, that seems like an oversight on our part. Would you be interested in making a PR to fix that?

mahnerak:
Sure!

I just don't want to increase the indentation of the train-loop code by adding `try/finally` and `with` to `train()`.

One good option is to add `close()` method to `Trainer` so it can handle even multiple handles to be closed. And the trainer instance can be closed from:

https://github.com/allenai/allennlp/blob/fbd2ccca1dd767e919a62d344c6b6569fc38227a/allennlp/commands/train.py#L503-L506

epwalsh:
Cool. I think maybe `Trainer.clean_up()` is a better/more general name for the method?

PR #4674, opened 2020-09-25
Making Trainer a context manager

mahnerak:
This is a fix for #4653

Sometimes a Trainer instance may use closable resources. Right now, the tensorboard logger implementation requires to be manually closed.

The reason to implement a context manager is that in the near future we may have multiple such handles that need to be closed (say, almost any logger that works on top of filesystem or network).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added context manager functionality (`__enter__` and `__exit__`) to `Trainer`,
   which will call automatically close used resources if `with` statement is used.
+- Added a `build-vocab` subcommand that can be used to build a vocabulary from a training config file.
+- Added `tokenizer_kwargs` argument to `PretrainedTransformerMismatchedIndexer`.
+- Added `tokenizer_kwargs` and `transformer_kwargs` arguments to `PretrainedTransformerMismatchedEmbedder`.
 - Added official support for Python 3.8.
 - Added a script: `scripts/release_notes.py`, which automatically prepares markdown release notes from the
   CHANGELOG and commit history.
@@ -30,15 +33,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Now `trainer.train()` does not close the `tensorboard` writer in the end (moved to `Trainer.close()`).
-  Instead, `with` statement is used with `Trainer` to auto-close such objects properly.
+- `trainer.train()` does not close the `tensorboard` writer, it merely flushes is. This makes it possible to use the
+  `GradientDescentTrainer` object more than once. To close the writer, call `GradientDescentTrainer.close()`, or
+  use it in a `with` statement.
+- Subcommands that don't require plugins will no longer cause plugins to be loaded or have an `--include-package` flag.
+- Allow overrides to be JSON string or `dict`.
 - `transformers` dependency updated to version 3.1.0.
 - When `cached_path` is called on a local archive with `extract_archive=True`, the archive is now extracted into a unique subdirectory of the cache root instead of a subdirectory of the archive's directory. The extraction directory is also unique to the modification time of the archive, so if the file changes, subsequent calls to `cached_path` will know to re-extract the archive.
 - Removed the `truncation_strategy` parameter to `PretrainedTransformerTokenizer`. The way we're calling the tokenizer, the truncation strategy takes no effect anyways.
 
+### Removed
+
+- Removed `common.util.is_master` function.
+
 ### Fixed
 
-- Fixed a bug where sometimes tensorboard logger may be not closed properly. Making `Trainer` a context manager.
+- Class decorators now displayed in API docs.
 - Fixed up the documentation for the `allennlp.nn.beam_search` module.
 - Ignore `*args` when constructing classes with `FromParams`.
 - Ensured some consistency in the types of the values that metrics return.
@@ -51,7 +61,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed a bug where `cached_path()` would fail if passed a `cache_dir` with the user home shortcut `~/`.
 - Fixed a bug in our doc building script where markdown links did not render properly
   if the "href" part of the link (the part inside the `()`) was on a new line.
-
+- Changed how gradients are zeroed out with an optimization. See [this video from NVIDIA](https://www.youtube.com/watch?v=9mS1fIYj1So)
+  at around the 9 minute mark.
+- Fixed a bug where parameters to a `FromParams` class that are dictionaries wouldn't get logged
+  when an instance is instantiated `from_params`.
+- Fixed a bug in distributed training where the vocab would be saved from every worker, when it should have been saved by only the local master process.
 
 ## [v1.1.0](https://github.com/allenai/allennlp/releases/tag/v1.1.0) - 2020-09-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Instead, `with` statement is used with `Trainer` to auto-close such objects properly.
 - `transformers` dependency updated to version 3.1.0.
 - When `cached_path` is called on a local archive with `extract_archive=True`, the archive is now extracted into a unique subdirectory of the cache root instead of a subdirectory of the archive's directory. The extraction directory is also unique to the modification time of the archive, so if the file changes, subsequent calls to `cached_path` will know to re-extract the archive.
+- Removed the `truncation_strategy` parameter to `PretrainedTransformerTokenizer`. The way we're calling the tokenizer, the truncation strategy takes no effect anyways.
 
 ### Fixed
 
@@ -50,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed a bug where `cached_path()` would fail if passed a `cache_dir` with the user home shortcut `~/`.
 - Fixed a bug in our doc building script where markdown links did not render properly
   if the "href" part of the link (the part inside the `()`) was on a new line.
+
 
 ## [v1.1.0](https://github.com/allenai/allennlp/releases/tag/v1.1.0) - 2020-09-08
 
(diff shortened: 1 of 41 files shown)


mahnerak:
@epwalsh I decided to stick with name `close()` as it (at least for me) makes more sense for context managers.
Please feel free to change it if you still prefer `clean_up()`.

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a bug where sometimes tensorboard logger may be not closed properly. Making `Trainer` a context manager.
```
epwalsh, line 36: Just a nit-pick here but this should really be two different CHANGELOG entries.

Something like this should be in the `### Added` section: "Added a `Trainer.close()` method to close / clean up any open resources the trainer is using, which will be automatically called when the `Trainer` is used as a context manager"

And then the bug fix note in the `### Fixed` section like you have it.

epwalsh (commented):
I think this is reasonable. Though @dirkgr is usually more opinionated than me about API changes like this, so I'd like to hear his thoughts as well.

CHANGELOG.md
mahnerak: Sorry about that.
Just added to the changelog.

dirkgr (commented):
This is a neat idea, but have you tried closing the writer in the `__del__()` method of the trainer? Does that work? If it works, it would mean that users of the trainer don't have to remember that it's a context object. The downside is that it's not clear when `__del__()` gets called. I don't know if it gets called at all when we're exiting due to an exception, hence my question.

mahnerak:
The issue is not specific to the allennlp trainer.
In Python, there is no guarantee of when `__del__` will be called.

Please look at:
https://stackoverflow.com/a/1481512
> However, this is an implementation detail of CPython. The only required property of Python garbage collection is that it happens after all references have been deleted, so this might not necessary happen right after and might not happen at all.

In order to make it explicit that `with` statement should be used with `Trainer`, we may have set a flag during `__enter__` and inside the `train` we can check whether it's inside the `with` or not. 
We can always add a warning if we detect such unexcepted usage from the user's end.

epwalsh:
I prefer this over https://github.com/allenai/allennlp/pull/4676. Making the trainer a context manager makes it clear that there are resources that need to be closed.

mahnerak:
Okay. I'll add some helpful warnings.
Also, the `find-lr` does not use `Trainer.train()` but implements his own.

There are also tests referring `Trainer`, but I'll add `with` statements to them in a separate PR.

allennlp/commands/train.py
```diff
@@ -502,7 +502,8 @@ def __init__(
         self.batch_weight_key = batch_weight_key
 
     def run(self) -> Dict[str, Any]:
-        return self.trainer.train()
+        with self.trainer as trainer:
+            return trainer.train()
```
dirkgr, line 506:
```suggestion
        with self.trainer:
            return self.trainer.train()
```
I don't really care one way or another, but I wanted to suggest this since it avoids an extra variable.

dirkgr:
What about not initializing `GradientDescentTrainer._tensorboard` until `train()` is called, and wrapping it in a try-finally block there? That way the public API stays the same and unencumbered, but we could still be certain it gets closed when it has to.

Is there ever a case where we need the tensor board writer to stay alive across two calls to the trainer's public API? If there is no such case, there is no reason to make a context manager.

mahnerak:
@dirkgr That's almost the same as #4676, isn't it?

Actually, I like the idea of having an easier API, but given that:
> https://github.com/allenai/allennlp/blob/d9bdaa952b455916973c9d1fa72e321c8cab2e86/allennlp/training/trainer.py#L41-L42

I think we need to have a more flexible API.
Also, this is related to #4675, I believe that Callbacks should be context managers as well, and it's better if they can be fired even before/after the `train` command (say, to dump the model hyperparameters to some logger, or to store the results in case of `evaluate_on_test`).

dirkgr:
@mahnerak, we decided to fix this issue with #4731 instead. That one just uses a try-finally block.

Your solutions, this one and #4676, were very good, but they made us realize that the API for the trainer as a whole is a little sketchy, so we didn't want to change it without addressing the root of the problem.

FWIW, the root of the problem as we see it is that there is only one way to use the trainer, and that's to initialize it, and then call `train()` exactly once. An API like that should just be a function. Making the trainer closable or making it a context manager does not address the root problem.

We will re-visit the issue of the trainer API for version 2.0. I'd be happy to get your input when we do!

shimengfeng:
Hi, 

I want to follow up with this issue. I have try to follow the instruction online to perform my training as 
`def build_trainer(model: model,
    train_loader: DataLoader,
    dev_loader: DataLoader
) -> Trainer:
    parameters = [
        [n, p]
        for n, p in model.named_parameters() if p.requires_grad
    ]
    optimizer = AdamOptimizer(parameters, lr=lr)
    trainer = GradientDescentTrainer(
        model=model,
        data_loader=train_loader,
        validation_data_loader=dev_loader,
        num_epochs=TOTAL_EPOCHES,
        optimizer=optimizer,
        cuda_device = 0,
        patience = PATIENCE,
        serialization_dir = model_dir,
        grad_clipping =1
    )
    return trainer
trainer = build_trainer(model, train_loader, dev_loader)
trainer.train()`

But I keep getting Permission denied error, and within the error details it seems the tensorboard.close() does not close properly. I am not sure whether it relates to this problem? Could someone explain a bit more what is the final solution? Thank you!

epwalsh:
@shimengfeng could you please provide a stacktrace?

shimengfeng:
Hi @epwalsh ,

Thank you for such a quick response. Will this be helpful?

```PermissionError                           Traceback (most recent call last)
<command-11051376> in <module>
     21     return trainer
     22 trainer = build_trainer(model, train_loader, dev_loader)
---> 23 trainer.train()
     24 #save vocab
     25 # vocab.save_to_files(vocab_dir)

/databricks/python/lib/python3.7/site-packages/allennlp/training/trainer.py in train(self)
    968 
    969         # make sure pending events are flushed to disk and files are closed properly
--> 970         self._tensorboard.close()
    971 
    972         # Load the best model state before returning

/databricks/python/lib/python3.7/site-packages/allennlp/training/tensorboard_writer.py in close(self)
    329         """
    330         if self._train_log is not None:
--> 331             self._train_log.close()
    332         if self._validation_log is not None:
    333             self._validation_log.close()

/databricks/python/lib/python3.7/site-packages/tensorboardX/writer.py in close(self)
   1050         for writer in self.all_writers.values():
   1051             writer.flush()
-> 1052             writer.close()
   1053         self.file_writer = self.all_writers = None
   1054 

/databricks/python/lib/python3.7/site-packages/tensorboardX/writer.py in close(self)
    168         Call this method when you do not need the summary writer anymore.
    169         """
--> 170         self.event_writer.close()
    171 
    172     def reopen(self):

/databricks/python/lib/python3.7/site-packages/tensorboardX/event_file_writer.py in close(self)
    157             self.flush()
    158             self._worker.stop()
--> 159             self._ev_writer.close()
    160             self._closed = True
    161 

/databricks/python/lib/python3.7/site-packages/tensorboardX/event_file_writer.py in close(self)
     74         return_value = self.flush()
     75         with self._lock:
---> 76             self._py_recordio_writer.close()
     77         return return_value
     78 

/databricks/python/lib/python3.7/site-packages/tensorboardX/record_writer.py in close(self)
    125 
    126     def close(self):
--> 127         self._writer.close()
    128 
    129 

PermissionError: [Errno 13] Permission denied```

I am using version 1.1.0

**Comment by epwalsh** [comment]
Thanks @shimengfeng. I think this is a separate issue. What are you passing for the `serialization_dir`? Is it a temp directory by any chance?

As of right now there is not a way to disable tensorboard completely, although that would probably be a good feature to have.

**Comment by shimengfeng** [comment]
@epwalsh currently I am using a directory that I created before model train and then I move the trained model to a unique s3 location. Right after I move the model, I will delete this created directory.  I will run the model multiple times with the same created directory (create, delete, and recreate, delete, etc.) 

Do you think that is what caused the issue? If that is the case, I can create unique locations each time locally before moving the file to s3.

**Comment by epwalsh** [comment]
It's possible. I would definitely try using a unique location each time and see if that fixes the issue.

**Comment by shimengfeng** [comment]
Thank you! Will try it. Please let me know if in the future if you want to implement the functionality of disable tensorboard, more than happy to contribute, considering I use this package quite often.

**Comment by epwalsh** [comment]
This is definitely something that we want. I was just trying to find a clean way to do this without breaking backwards compatibility, but in doing I realized the way we handle `Lazy` optional argument in `FromParams` constructors is not ideal. I'm working on a solution to that right now which should allow us to make the `TensorboardWriter` truly optional.

**Comment by dirkgr** [comment]
If possible, Tensorboard should be implemented as a `TrainerCallback`, just like the other logging backends.

**Comment by epwalsh** [comment]
@dirkgr I agree that makes more sense. But in the meantime @shimengfeng it would now be an easy change to make `TensorboardWriter` optional. This line:

https://github.com/allenai/allennlp/blob/b7cec515448f46daec9a76551d0a3342ef51b27b/allennlp/training/trainer.py#L1230

Could be changed to

```python
tensorboard_writer: Optional[Lazy[TensorboardWriter]] = Lazy(TensorboardWriter),
```

Which would allow you to set `tensorboard_writer: null` in your config to disable it. @shimengfeng are you still interested in making a PR for that?

shimengfeng:
Yup, totally! Do I need to fill in a feature request and then change it?

mahnerak:
AFAIK As of now, if there's no `tensorboard_writer` specified in the config, a default `TensorboardWriter` is used.
If I'm not wrong, this change will disable tensorboard logging for such config files, right? Do we want to change the default behaviour?

epwalsh:
@mahnerak, no, we will keep the default behavior the same, i.e. the default writer is used unless it is explicity set to `null`.

epwalsh:
@shimengfeng don't worry about making a feature request since we're already discussing it here.

mahnerak:
@epwalsh If the tensorboard writer callback being absent in the list of callbacks is the default behaviour, then what do we have to specify in the config to disable the callback?
Do you mean that tensorboard callback should never be passed through `trainer_callbacks`, although it should be implemented as a callback?

epwalsh:
The `TensorboardWriter` is not currently implemented as a callback. But if it were implemented as a callback, I suppose we would have a set of default callbacks of which the `TensorboardWriter` would be part of. So these default callbacks would always be used unless overridden by the config.

PR #4676, opened 2020-09-26
Make `Trainer.train()` to auto-close all the resources

### LINKED ISSUE #4653: Closing the TensorBoard writer

mahnerak:
This is a much clear alternative to #4674 supposed to fix  #4653.

Taking into account the fact, that `train()` is intended to be called once per `Trainer` (per run), also that only `train()` uses resources that need to be closed, we can just auto-close them right before leaving the method.

This eliminates the need to manually `close()` the `Trainer` or to use a `with` statement.
However, this approach has few drawbacks: it limits the usage of resources to the `train` method.
Either 1) the other methods are not guaranteed to have access to such resources or 2) had to manually close the resources (not a trivial thing to do, if the method may be called multiple times).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `Trainer` now has a `close()` method to free the resources. Defaults to no-op.
 - Added official support for Python 3.8.
 - Added a script: `scripts/release_notes.py`, which automatically prepares markdown release notes from the
   CHANGELOG and commit history.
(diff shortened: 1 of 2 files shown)


mahnerak:
@dirkgr @epwalsh 
Anyway, I still believe that it's better for users to explicitly state when the resources can be released. I see #4674 as a better and more permanent way to fix the issue.

dirkgr:
Closing, see comment in #4674.

Issue #4656, opened 2020-09-22
Allow overrides to be JSON string or dict


mulhod:
**Is your feature request related to a problem? Please describe.**
Currently, `overrides` expects a JSON string representing a dictionary. It would be nice to be able to pass in a Python `dict` or a JSON string. This would make it slightly easier if using Python directly to pass in a value to `overrides`.

**Describe the solution you'd like**
Enable `overrides` to accept either a JSON string or a Python `dict`.

**Describe alternatives you've considered**
The alternative is to use `json.dumps` to dump the `dict` to a `str` and pass that value in via `overrides`.

**Additional context**
Mentioned in #4651

epwalsh:
Closed via https://github.com/allenai/allennlp/pull/4680

PR #4680, opened 2020-09-28
Allow overrides to be JSON string or dict

mulhod:
Addresses #4656

### DIFF
--- a/allennlp/common/params.py
+++ b/allennlp/common/params.py
@@ -457,7 +457,10 @@ class Params(MutableMapping):
 
     @classmethod
     def from_file(
-        cls, params_file: Union[str, PathLike], params_overrides: str = "", ext_vars: dict = None
+        cls,
+        params_file: Union[str, PathLike],
+        params_overrides: Union[str, Dict[str, Any]] = "",
+        ext_vars: dict = None,
     ) -> "Params":
         """
         Load a `Params` object from a configuration file.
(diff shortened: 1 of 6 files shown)


mulhod:
@epwalsh Just tagging you here since I don't think I can technically "assign" anybody to be a reviewer.

allennlp/common/params.py
```diff
@@ -490,6 +493,8 @@ def from_file(
 
         file_dict = json.loads(evaluate_file(params_file, ext_vars=ext_vars))
 
+        if isinstance(params_overrides, dict):
+            params_overrides = json.dumps(params_overrides)
         overrides_dict = parse_overrides(params_overrides)
```
epwalsh, line 498: How about just:

```suggestion
        if isinstance(params_overrides, dict):
            overrides_dict = params_overrides
        else:
            overrides_dict = parse_overrides(params_overrides)
```

mulhod: Yeah, I actually tried this first. But there is some logic to `parse_overrides` that converts the value.

For example, let's say we have the following as the input:
```python
input_ = {
    'train_data_path': 'FOO',
    'model': {'type': 'BAR'},
    'model.text_field_embedder.tokens.type': 'BAZ',
    'data_loader.batch_sampler.sorting_keys.0': 'question'
}
```

This happens if we pass it into `parse_overrides` after `dump`ing:
```python
parse_overrides(json.dumps(input_))
# prints {'data_loader': {'batch_sampler': {'sorting_keys': {'0': 'question'}}},
#         'model': {'type': 'BAR', 'text_field_embedder': {'tokens': {'type': 'BAZ'}}},
#         'train_data_path': 'FOO'}
```

I want `parse_overrides` to do the same kind of transformation of the `dict` as it would do if it were represented as a JSON string. Does this make sense?

epwalsh: Oh right, that's a good point.

epwalsh (approved):

Issue #4634, opened 2020-09-12
[Contribution] DeepSpeed Integration


jacobdanovitch:
**DeepSpeed background**

[DeepSpeed](https://www.deepspeed.ai/) is a distributed training engine for PyTorch, primarily for training very large language models with significantly less memory. For example, the [17.7 billion parameter Turing-NLG](https://www.microsoft.com/en-us/research/blog/turing-nlg-a-17-billion-parameter-language-model-by-microsoft/) was trained with DeepSpeed's ZeRO optimizer.

**Proposal**

It seems like a natural fit to have a way to use this with AllenNLP for use with large, distributed experiments. It also shouldn't require any _major_ changes to integrate. Their training loop looks like:

```python
# https://www.deepspeed.ai/getting-started/#training
model_engine, optimizer, _, _ = deepspeed.initialize(args=cmd_args,
                                                     model=model,
                                                     model_parameters=params)
for step, batch in enumerate(data_loader):
    #forward() method
    loss = model_engine(batch)

    #runs backpropagation
    model_engine.backward(loss)

    #weight update
    model_engine.step()
```

In terms of where it would fit into the library, I think a standalone `DeepSpeedTrainer(Trainer)` subclass would make sense. It should be fairly similar to `GradientDescentTrainer` (minus stuff that DeepSpeed handles itself, like gradient accumulation). It could then be initialized from a config file by the user as per usual.

I know not having dependencies on other libraries is a point of emphasis. It should be possible to include without adding deepspeed as a dependency and allowing the user to install it independently by doing something like:

```
# allennlp.training.__init__.py
# ...
try:
  from allennlp.training.deepspeed_trainer import DeepSpeedTrainer
except ImportError:
  pass # maybe a warning here or something
```

**Initial results**

I was able to get a prototype up and running pretty easily. I didn't subclass `GradientDescentTrainer` (I had a lot of trouble doing that, for whatever reason), but I just copied and pasted the code and started ripping stuff out as I went.

I setup a training experiment for a basic classifier on the first 10k instances of SST using RoBERTA-base across two GPUs. The `GradientDescentTrainer` completed an epoch in 20.40s, using 8936MB / 10202MB of GPU memory. The `DeepSpeedTrainer` prototype completed an epoch in 46.91s, using just **4184MB / 4348MB** of GPU memory (less than half!). I don't know why it took so much longer but I strongly assume it's something I implemented wrong myself.

[...]

Issue #4623, opened 2020-09-03
Cannot train language_model: RuntimeError: Tensors must be CUDA and dense


tpanza:
## Checklist

- [x] I have verified that the issue exists against the `master` branch of AllenNLP.
- [x] I have read the relevant section in the [contribution guide](https://github.com/allenai/allennlp/blob/master/CONTRIBUTING.md#bug-fixes-and-new-features) on reporting bugs.
- [x] I have checked the [issues list](https://github.com/allenai/allennlp/issues) for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/allenai/allennlp/pulls) for existing proposed fixes.
- [x] I have checked the [CHANGELOG](https://github.com/allenai/allennlp/blob/master/CHANGELOG.md) and the [commit log](https://github.com/allenai/allennlp/commits/master) to find out if the bug was already fixed in the master branch.
- [x] I have included in the "Description" section below a traceback from any exceptions related to this bug.
- [x] I have included in the "Related issues or possible duplicates" section beloew all related issues and possible duplicate issues (If there are none, check this box anyway).
- [x] I have included in the "Environment" section below the name of the operating system and Python version that I was using when I discovered this bug.
- [x] I have included in the "Environment" section below the output of `pip freeze`.
- [x] I have included in the "Steps to reproduce" section below a minimally reproducible example.

## Description

<details>
<summary><b>Python traceback:</b></summary>
<p>

```
0 | 2020-09-03 12:39:41,240 - INFO - allennlp.training.trainer - Beginning training.
0 | 2020-09-03 12:39:41,240 - INFO - allennlp.training.trainer - Epoch 0/9
0 | 2020-09-03 12:39:41,241 - INFO - allennlp.training.trainer - Worker 0 memory usage MB: 2984.032
0 | 2020-09-03 12:39:41,241 - INFO - allennlp.training.trainer - Worker 1 memory usage MB: 2985.344
0 | 2020-09-03 12:39:41,241 - INFO - allennlp.training.trainer - Worker 2 memory usage MB: 2988.028
0 | 2020-09-03 12:39:41,241 - INFO - allennlp.training.trainer - Worker 3 memory usage MB: 2984.496
0 | 2020-09-03 12:39:41,241 - INFO - allennlp.training.trainer - Worker 4 memory usage MB: 2989.064
0 | 2020-09-03 12:39:41,241 - INFO - allennlp.training.trainer - Worker 5 memory usage MB: 2987.996
[...]

PR #4693, opened 2020-10-02
Deepspeed integration

jacobdanovitch:
Draft for #4634 . Small change to `allennlp.training.metrics.categorical_accuracy` addresses my comment in #4623 . Still very rough, but functional.

Example config: [gist](https://gist.github.com/jacobdanovitch/d3b8d3ca476d645234f920b4ff228954)

### DIFF
--- a/allennlp/training/__init__.py
+++ b/allennlp/training/__init__.py
@@ -1,6 +1,7 @@
 from allennlp.training.checkpointer import Checkpointer
 from allennlp.training.tensorboard_writer import TensorboardWriter
 from allennlp.training.no_op_trainer import NoOpTrainer
+from allennlp.training.deepspeed_trainer import DeepspeedTrainer
 from allennlp.training.trainer import (
     Trainer,
     GradientDescentTrainer,
--- a/allennlp/training/deepspeed_trainer.py
+++ b/allennlp/training/deepspeed_trainer.py
@@ -1,0 +1,856 @@
+import logging
+from deepspeed.utils import logger as ds_logger
+ds_logger.setLevel(logging.WARNING)
+ds_logger.propagate = False
+
+import datetime
+import logging
+import os
+import re
+import math
+import json
+import tempfile
+import time
+import traceback
+from copy import deepcopy
+from contextlib import contextmanager
+from typing import Any, Dict, Iterator, List, Optional, Tuple, Union
+
+from allennlp.common.util import int_to_device
+
+import torch
+import torch.distributed as dist
+from torch.cuda import amp
+from torch.nn.parallel import DistributedDataParallel
+from torch.nn.utils import clip_grad_norm_
+
+import deepspeed
+
+from allennlp.common import Lazy, Registrable, Tqdm, Params, FromParams
+from allennlp.common import util as common_util
+from allennlp.common.checks import ConfigurationError, check_for_gpu
+from allennlp.data import DataLoader
+from allennlp.data.dataloader import TensorDict
+from allennlp.models.model import Model
+from allennlp.nn import util as nn_util
+from allennlp.training import util as training_util
+from allennlp.training.checkpointer import Checkpointer
+from allennlp.training.metric_tracker import MetricTracker
+from allennlp.training.moving_average import MovingAverage
+from allennlp.training.optimizers import Optimizer
+from allennlp.training.tensorboard_writer import TensorboardWriter
+from allennlp.training.trainer import Trainer, BatchCallback, EpochCallback
+
+logger = logging.getLogger(__name__)
+
+JsonDict = Dict[str, Any]
+
+class DeepspeedConfig(FromParams):
+    def __init__(
+        self,
+        optimizer: JsonDict,
+        fp16: JsonDict = {'enabled': False},
+        amp:  JsonDict = {'enabled': False},
+        zero_optimization: Union[bool, Dict] = False,
+        zero_allow_untested_optimizer: bool = True,
+        wall_clock_breakdown: bool = False
+    ):
+        self.optimizer = optimizer
+        self.fp16 = fp16
+        self.amp = amp
+        self.zero_optimization = zero_optimization
+        self.zero_allow_untested_optimizer = zero_allow_untested_optimizer
+        self.wall_clock_breakdown = wall_clock_breakdown
+
+    @staticmethod
+    def build_deepspeed_args(deepspeed_config_path: str, local_rank: int = 0):
+        from argparse import ArgumentParser, Namespace
+        parser = ArgumentParser()
+        parser.add_argument('--local_rank', type=int, default=local_rank)
+        parser = deepspeed.add_config_arguments(parser)
+
+        args, _ = parser.parse_known_args()
+        arg_dict = vars(args)
+
+        arg_dict.update(dict(deepspeed_config=deepspeed_config_path, deepspeed=True, local_rank=local_rank))
+        return Namespace(**arg_dict)
+
+    @property
+    def config(self):
+        # return {
+        #     'fp16': self.fp16,
+        #     'amp': self.amp,
+        #     'zero_optimization': self.zero_optimization,
+        #     'zero_allow_untested_optimizer': self.zero_allow_untested_optimizer
+        # }
+        return vars(self)
+
+    def _to_temp_file(self, serialization_dir, **kwargs):
+        fd, path = tempfile.mkstemp(dir=serialization_dir)
+        
+        config = {**self.config, **kwargs}
+        with os.fdopen(fd, 'w') as f:
+            f.write(json.dumps(config))
+
+        return path
+
+    def launch(
+        self,
+        model: torch.nn.Module,
+        optimizer: Union[str, torch.optim.Optimizer],
+        local_rank: int, 
+        serialization_dir: str,
+        batch_size: int, 
+        gradient_accumulation_steps: int,
+        **kwargs
+    ):
+        path = self._to_temp_file(serialization_dir, train_batch_size=batch_size, gradient_accumulation_steps=gradient_accumulation_steps)
+        ds = deepspeed.initialize(
+            args=self.build_deepspeed_args(path, local_rank),
+            model=model,
+            model_parameters=model.parameters(),
+            dist_init_required=False,
+            **kwargs
+        )
+
+        os.remove(path)
+        return ds
+
+@Trainer.register("deepspeed", constructor="from_partial_objects")
+class DeepspeedTrainer(Trainer):
+    def __init__(
+        self,
+        model: Model,
+        optimizer: torch.optim.Optimizer,
+        data_loader: DataLoader,
+        deepspeed_config: DeepspeedConfig,
+        patience: Optional[int] = None,
+        validation_metric: str = "-loss",
+        validation_data_loader: DataLoader = None,
+        num_epochs: int = 20,
+        serialization_dir: Optional[str] = None,
+        checkpointer: Checkpointer = None,
+        cuda_device: Optional[Union[int, torch.device]] = None,
+        grad_norm: Optional[float] = None,
+        grad_clipping: Optional[float] = None,
+        tensorboard_writer: TensorboardWriter = None,
+        moving_average: Optional[MovingAverage] = None,
+        batch_callbacks: List[BatchCallback] = None,
+        epoch_callbacks: List[EpochCallback] = None,
+        distributed: bool = False,
+        local_rank: int = 0,
+        world_size: int = 1,
+        num_gradient_accumulation_steps: int = 1,
+        use_amp: bool = False,
+    ) -> None:
+        super().__init__(serialization_dir, cuda_device, distributed, local_rank, world_size)
+
+        # I am not calling move_to_gpu here, because if the model is
+        # not already on the GPU then the optimizer is going to be wrong.
+        self.model = model
+
+        self.data_loader = data_loader
+        self._validation_data_loader = validation_data_loader
+        self.optimizer = optimizer
+
+        if patience is None:  # no early stopping
+            if validation_data_loader is not None:
+                logger.warning(
+                    "You provided a validation dataset but patience was set to None, "
+                    "meaning that early stopping is disabled"
+                )
+        elif (not isinstance(patience, int)) or patience <= 0:
+            raise ConfigurationError(
+                '{} is an invalid value for "patience": it must be a positive integer '
+                "or None (if you want to disable early stopping)".format(patience)
+            )
+
+        # For tracking is_best_so_far and should_stop_early
+        self._metric_tracker = MetricTracker(patience, validation_metric)
+        # Get rid of + or -
+        self._validation_metric = validation_metric[1:]
+
+        self._num_epochs = num_epochs
+
+        if checkpointer is not None:
+            self._checkpointer = checkpointer
+        else:
+            self._checkpointer = Checkpointer(serialization_dir)
+
+        self._grad_norm = grad_norm
+        self._grad_clipping = grad_clipping
+
+        self._moving_average = moving_average
+        self._batch_callbacks = batch_callbacks or []
+        self._epoch_callbacks = epoch_callbacks or []
+
+        # We keep the total batch number as an instance variable because it
+        # is used inside a closure for the hook which logs activations in
+        # `_enable_activation_logging`.
+        self._batch_num_total = 0
+
+        self._tensorboard = tensorboard_writer or TensorboardWriter(serialization_dir)
+        self._tensorboard.get_batch_num_total = lambda: self._batch_num_total
+        self._tensorboard.enable_activation_logging(self.model)
+
+        self._last_log = 0.0  # time of last logging
+
+        self._num_gradient_accumulation_steps = num_gradient_accumulation_steps
+
+        # Enable automatic mixed precision training.
+        self._scaler: Optional[amp.GradScaler] = None
+        self._use_amp = use_amp
+        if self._use_amp:
+            if self.cuda_device == torch.device("cpu"):
+                raise ValueError("Using AMP requires a cuda device")
+            self._scaler = amp.GradScaler()
+
+        self._pytorch_model = self.model
+        
+        self._ds_config = deepspeed_config
+        self.model_engine, self.ds_optimizer, _, _ = self._ds_config.launch(
+            self.model,
+            None, # self.optimizer,
+            local_rank,
+            serialization_dir,
+            self.data_loader.batch_size,
+            num_gradient_accumulation_steps
+        )
+
+        def mute_log(*args, **kwargs):
+            pass
+
+        if hasattr(self.model_engine, 'timers'):
+            self.model_engine.timers.log = mute_log
+
+    def batch_outputs(self, batch: TensorDict, for_training: bool) -> Dict[str, torch.Tensor]:
+        """
+        Does a forward pass on the given batch and returns the output dictionary that the model
+        returns, after adding any specified regularization penalty to the loss (if training).
+        """
+        # batch = nn_util.move_to_device(batch, self.cuda_device)
+        batch = nn_util.move_to_device(batch, self.model_engine.device)
+        output_dict = self.model_engine(**batch)
+
+        if for_training:
+            try:
+                assert "loss" in output_dict
+                regularization_penalty = self.model.get_regularization_penalty()
+
+                if regularization_penalty is not None:
+                    output_dict["reg_loss"] = regularization_penalty
+                    output_dict["loss"] += regularization_penalty
+
+            except AssertionError:
+                if for_training:
+                    raise RuntimeError(
+                        "The model you are trying to optimize does not contain a"
+                        " 'loss' key in the output of model.forward(inputs)."
+                    )
+
+        return output_dict
+
+    def _train_epoch(self, epoch: int) -> Dict[str, float]:
+        """
+        Trains one epoch and returns metrics.
+        """
+        logger.info("Epoch %d/%d", epoch, self._num_epochs - 1)
+        cpu_memory_usage = []
+        for worker, memory in common_util.peak_memory_mb().items():
+            cpu_memory_usage.append((worker, memory))
+            logger.info(f"Worker {worker} memory usage MB: {memory}")
+        gpu_memory_usage = []
+        for gpu, memory in common_util.gpu_memory_mb().items():
+            gpu_memory_usage.append((gpu, memory))
+            logger.info(f"GPU {gpu} memory usage MB: {memory}")
+
+        regularization_penalty = self.model.get_regularization_penalty()
+
+        train_loss = 0.0
+        batch_loss = 0.0
+
+        if regularization_penalty is not None:
+            train_reg_loss = 0.0
+            batch_reg_loss = 0.0
+        else:
+            train_reg_loss = None
+            batch_reg_loss = None
+        # Set the model to "train" mode.
+        self.model_engine.train()
+
+        # Get tqdm for the training batches
+        batch_generator = iter(self.data_loader)
+        batch_group_generator = common_util.lazy_groups_of(
+            batch_generator, self._num_gradient_accumulation_steps
+        )
+
+        logger.info("Training")
+
+        num_training_batches: Union[int, float]
+        try:
+            len_data_loader = len(self.data_loader)
+            num_training_batches = math.ceil(
+                len_data_loader / self._num_gradient_accumulation_steps
+            )
+        except TypeError:
+            num_training_batches = float("inf")
+
+        # Having multiple tqdm bars in case of distributed training will be a mess. Hence only the master's
+        # progress is shown
+        batch_group_generator_tqdm = batch_group_generator
+        if self._master:
+            batch_group_generator_tqdm = Tqdm.tqdm(
+                batch_group_generator, total=num_training_batches
+            )
+
+        self._last_log = time.time()
+
+        batches_this_epoch = 0
+        if self._batch_num_total is None:
+            self._batch_num_total = 0
+
+        done_early = False
+        for batch_group in batch_group_generator_tqdm:
+            batches_this_epoch += 1
+            self._batch_num_total += 1
+            batch_num_total = self._batch_num_total
+
+            self.optimizer.zero_grad()
+
+            batch_group_outputs = []
+            for batch in batch_group:
+                with amp.autocast(self._use_amp):
+                    batch_outputs = self.batch_outputs(batch, for_training=True)
+                    batch_group_outputs.append(batch_outputs)
+                    loss = batch_outputs.get("loss")
+                    reg_loss = batch_outputs.get("reg_loss")
+                    if torch.isnan(loss):
+                        raise ValueError("nan loss encountered")
+                    loss = loss / len(batch_group)
+
+                    batch_loss = loss.item()
+                    train_loss += batch_loss
+                    if reg_loss is not None:
+                        reg_loss = reg_loss / len(batch_group)
+                        batch_reg_loss = reg_loss.item()
+                        train_reg_loss += batch_reg_loss
+
+                
+                self.model_engine.backward(loss)
+                self.model_engine.step()
+
+            param_updates = None
+            if self._tensorboard.should_log_histograms_this_batch() and self._master:
+                # Get the magnitude of parameter updates for logging.  We need to do some
+                # computation before and after the optimizer step, and it's expensive because of
+                # GPU/CPU copies (necessary for large models, and for shipping to tensorboard), so
+                # we don't do this every batch, only when it's requested.
+                param_updates = {
+                    name: param.detach().cpu().clone()
+                    for name, param in self.model.named_parameters()
+                }
+
+                if self._scaler is not None:
+                    self._scaler.step(self.optimizer)
+                    self._scaler.update()
+                else:
+                    self.optimizer.step()
+
+                for name, param in self.model.named_parameters():
+                    param_updates[name].sub_(param.detach().cpu())
+            else:
+                if self._scaler is not None:
+                    self._scaler.step(self.optimizer)
+                    self._scaler.update()
+                else:
+                    self.optimizer.step()
+
+            # Update moving averages
+            if self._moving_average is not None:
+                self._moving_average.apply(batch_num_total)
+
+            # Update the description with the latest metrics
+            metrics = training_util.get_metrics(
+                self.model,
+                train_loss,
+                train_reg_loss,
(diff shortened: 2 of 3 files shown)


dirkgr:
I did not carefully examine the difference between the existing trainer and the deep speed one, but it looks like they are almost the same?

jacobdanovitch:
> I did not carefully examine the difference between the existing trainer and the deep speed one, but it looks like they are almost the same?

Yes, they are very similar. The main differences are that deepspeed's model engine handles things like gradient accumulation, norm, clipping, schedulers, so I removed a lot of that functionality, and modified the backprop step. I also may need to adjust the checkpointing. 

This is an MWE, but like I was talking about in the issue thread, I think it could be further optimized by avoiding the direct use of their model engine, which I'll take a look at next.

allennlp/training/deepspeed_trainer.py
```diff
@@ -0,0 +1,856 @@
...
+        for batch_group in batch_group_generator_tqdm:
+            batches_this_epoch += 1
+            self._batch_num_total += 1
+            batch_num_total = self._batch_num_total
+
+            self.optimizer.zero_grad()
+
+            batch_group_outputs = []
+            for batch in batch_group:
+                with amp.autocast(self._use_amp):
```
dirkgr, line 322: DeepSpeed doesn't autocast itself?

allennlp/training/deepspeed_trainer.py
```diff
@@ -0,0 +1,856 @@
...
+            )
+
+        self._last_log = time.time()
+
+        batches_this_epoch = 0
+        if self._batch_num_total is None:
+            self._batch_num_total = 0
+
+        done_early = False
+        for batch_group in batch_group_generator_tqdm:
```
dirkgr, line 313: This whole `batch_group` business is for gradient accumulation. Does DeepSpeed have its own solution for this?

allennlp/training/deepspeed_trainer.py
```diff
@@ -0,0 +1,856 @@
...
+
+
+            if self._master:
+                self._checkpointer.save_checkpoint(
+                    epoch, self, is_best_so_far=self._metric_tracker.is_best_so_far()
+                )
+
+            # Wait for the master to finish saving the checkpoint
+            if self._distributed:
+                dist.barrier()
```
dirkgr, line 656: These kinds of distributed checks could go away, I assume, because DeepSpeed has its own way of handling distribution.

dirkgr (commented):
You said it's slow because of a bunch of logging and metrics stuff. Do you think the issue is tensor board and the AllenNLP metrics code? Or are there other things that I didn't see?

I put in some comments about stuff that might have to change. They are mostly there to confirm or deny the half-truths I know about DeepSpeed.

jacobdanovitch:
See my comments in the issue thread for more detail. The slowdown seems to be related to gradient accumulation. The next steps are (1) seeing if the slowdown is reproducible on other machines and (2) confirming the right place in the library for this to go. Once those are both resolved I'm going to simplify the model engine using `Registrable`s and then this should be ~reviewable.

allennlp/training/deepspeed/__init__.py
```diff
@@ -1 +1,11 @@
-from allennlp.training.deepspeed.deepspeed_trainer import DeepspeedTrainer
\ No newline at end of file
+from allennlp.training.deepspeed.trainer import DeepspeedTrainer
```
jacobdanovitch, line 1: Todo before review:

1. Catch up to master
2. Implement checkpointing
3. Add LR scheduler, activation checkpointing, communication options to config

Stretch goals:
* Get sparse pretrained transformers working
* Pipeline parallelism

jacobdanovitch:
@dirkgr I think this is ready to take a look at. Some notes thus far:

* Deepspeed is heavily config-based and it's hard to avoid, so rather than fighting it, I just tried to make the config itself registrable. So everything can be handled from within one config file.
* That said, some things (like optimizers) can be instantiated directly, so I've made it possible to either pass an optimizer to the trainer as normal, or pass an optimizer config to deepspeed. If it's a bit confusing to have both ways, it'll be easy to pick one and stick with it.
* There's a lot of code overlap with the `GradientDescent` trainer, still. I don't know if this is avoidable without breaking the trainer into various hooks like Lightning does. 
* I've only worked on the optimization/scaling-related features, I haven't looked at Sparse Attention / the transformer kernel yet (in allennlp or in general). I can try to add these things or just wait until the existing stuff is done with.

allennlp/training/__init__.py
```diff
@@ -9,3 +9,10 @@
     TrainerCallback,
     TrackEpochCallback,
 )
+from allennlp.training.deepspeed import DeepspeedTrainer
+
+# import warnings
+# try:
+#     from allennlp.training.deepspeed import DeepspeedTrainer
+# except ImportError:
+#     warnings.warn('Deepspeed plugin not installed. Ignoring.')
```
dirkgr, line 18: Is this leftover debug code?

allennlp/training/deepspeed/checkpointer.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class DeepspeedCheckpointer(Checkpointer):
+    # def maybe_save_checkpoint(
+    #     self, 
+    #     trainer: _DeepspeedTrainer, 
+    #     epoch: int, 
+    #     batches_this_epoch: int
+    # ) -> None:
+    #     0/0
```
dirkgr, line 28: More debug code

allennlp/training/deepspeed/checkpointer.py
```diff
@@ -0,0 +1,131 @@
...
+            
+            checkpoint_id = "deepspeed_epoch_{}".format(epoch)
+            model_path = os.path.join(self._serialization_dir, "model_state_epoch_{}".format(epoch))
+            model_engine.save_checkpoint(self._serialization_dir, checkpoint_id)
+
+            # TODO
+            # Model will need a weight file to load; 
+            # not sure if ZeRO stage 2 will mess this up
+            if not os.path.isfile(model_path):
+                    torch.save(model_state, model_path)
```
dirkgr, line 51: This would be good to know. Have you tried the checkpointing logic? Does it work?

allennlp/training/deepspeed/checkpointer.py
```diff
@@ -0,0 +1,131 @@
...
+
+import allennlp
+from allennlp.nn import util as nn_util
+from allennlp.training import util as training_util, Checkpointer
+
+logger = logging.getLogger(__name__)
+_DeepspeedTrainer = "allennlp.training.deepspeed.trainer.DeepspeedTrainer"
+
+
+class DeepspeedCheckpointer(Checkpointer):
```
dirkgr, line 21: Does this differ much from the regular checkpointer?

allennlp/training/deepspeed/checkpointer.py
```diff
@@ -0,0 +1,131 @@
...
+
+        checkpoints = (self._serialization_dir and Path(self._serialization_dir).glob('deepspeed_epoch_*')) or []
+        checkpoints = sorted(c for c in checkpoints if c.is_dir())
+        if not checkpoints:
+            return None
+
+        engine_path = checkpoints[-1]
+        return engine_path, model_path, training_state_path
+
+    def restore_checkpoint(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
```
dirkgr, line 117: I didn't make sure, but a lot of these functions look identical to the ones from the regular checkpointer. Can you derive from that one and just override the methods that have differences?

allennlp/training/deepspeed/config.py
```diff
@@ -0,0 +1,68 @@
...
+    loss_scale: float = 0.
+    initial_scale_power: int = 32
+    loss_scale_window: int = 1000
+    hysteresis: int = 2
+    min_loss_scale: float = 1.
+
+@dataclass
+class DeepspeedAMPConfig(FromParams):
+    enabled: bool = False
+    opt_level: str = "O1"
```
dirkgr, line 19: I thought AMP was dead and we now use things built directly into PyTorch?

allennlp/training/deepspeed/optimizers.py
```diff
@@ -0,0 +1,87 @@
...
+            betas=betas,
+            eps=eps,
+            weight_decay=weight_decay,
+            amsgrad=amsgrad,
+            bias_correction=bias_correction,
+            adam_w_mode=adam_w_mode,
+            set_grad_none=set_grad_none,
+        )
+
+# This does not currently work
```
dirkgr, line 39: Why not? If it doesn't work and is not necessary, can we remove it?

allennlp/training/deepspeed/sparse_transformer_embedder.py
```diff
@@ -0,0 +1,11 @@
+from allennlp.modules.token_embedders.token_embedder import TokenEmbedder
+from allennlp.modules.token_embedders.pretrained_transformer_embedder import PretrainedTransformerEmbedder
+
+from deepspeed.ops.sparse_attention.sparse_attention_utils import SparseAttentionUtils
+
+# Doesn't work yet
```
dirkgr, line 6: same

allennlp/training/deepspeed/trainer.py
```diff
@@ -0,0 +1,703 @@
...
+        self._tensorboard.get_batch_num_total = lambda: self._batch_num_total
+        self._tensorboard.enable_activation_logging(self.model)
+
+        self._last_log = 0.0  # time of last logging
+
+        self._num_gradient_accumulation_steps = num_gradient_accumulation_steps
+
+        self._pytorch_model = self.model
+
+    def batch_outputs(self, batch: TensorDict, for_training: bool) -> Dict[str, torch.Tensor]:
```
dirkgr, line 124: How many methods in here actually differ from `GradientDescentTrainer`. Can you subclass that one and only override the methods that are actually different?

dirkgr: To be sure, I don't want to compromise `GradientDescentTrainer` to reduce code duplication here, but if we can reduce it without compromising, let's do it.

allennlp/training/deepspeed/trainer.py
```diff
@@ -0,0 +1,703 @@
...
+
+        done_early = False
+        for batch in batch_generator_tqdm:
+            batches_this_epoch += 1
+            self._batch_num_total += 1
+            batch_num_total = self._batch_num_total
+
+            # with LineProfiler(self.batch_outputs, target_gpu=self.model_engine.device) as prof:
+            batch_outputs = self.batch_outputs(batch, for_training=True)
+            # prof.print_stats() # display()
```
dirkgr, line 209: debugging code?

allennlp/training/deepspeed/trainer.py
```diff
@@ -0,0 +1,703 @@
...
+                estimated_time_remaining = training_elapsed_time * (
+                    (self._num_epochs - epoch_counter) / float(epoch - epoch_counter + 1) - 1
+                )
+                formatted_time = str(datetime.timedelta(seconds=int(estimated_time_remaining)))
+                logger.info("Estimated training time remaining: %s", formatted_time)
+
+            epochs_trained += 1
+
+        # make sure pending events are flushed to disk and files are closed properly
+        self._tensorboard.close()
```
dirkgr, line 540: I'm afraid this has moved somewhere else in the meantime. So we already have things diverging. Can you try to handle closing tensor board the same way here?

allennlp/training/deepspeed/trainer.py
```diff
@@ -0,0 +1,703 @@
...
+        checkpoint_id, model_state, training_state = self._checkpointer.restore_checkpoint()
+
+        if not training_state:
+            # No checkpoint to restore, start at 0
+            return 0
+
+        self.model.load_state_dict(model_state)
+        self.model_engine.load_checkpoint(self._serialization_dir, checkpoint_id)
+        # self.optimizer.load_state_dict(training_state["optimizer"])
+        # training_util.move_optimizer_to_cuda(self.optimizer)
```
dirkgr, line 595: What happened here?

allennlp/training/deepspeed/utils.py
```diff
@@ -0,0 +1,39 @@
...
+        optimizer=optimizer,
+        model_parameters=model.parameters(),
+        dist_init_required=False,
+        config_params=config
+    )
+    if hasattr(ds, 'timers'):
+        def mute_log(*args, **kwargs):
+            pass
+        ds.timers.log = mute_log
+    return ds, ds.optimizer
```
dirkgr, line 39: Does this have to be in a separate function? It's only called in one place, so it could live there?

If we put it here, we are sort of creating a new public API. I'd like to avoid that if we don't have to.

allennlp/training/deepspeed/trainer.py
```diff
@@ -0,0 +1,703 @@
...
+        validation_metric: str = "-loss",
+        num_epochs: int = 20,
+        cuda_device: Optional[Union[int, torch.device]] = None,
+        distributed: bool = None,
+        world_size: int = 1,
+        num_gradient_accumulation_steps: int = 1,
+        no_grad: List[str] = None,
+        optimizer: Lazy[Optimizer] = None,
+        deepspeed_optimizer: Dict[str, Any] = None,
+        deepspeed_args: Lazy[DeepspeedArgs] = Lazy(DeepspeedArgs),
```
dirkgr, line 641: Do these have to be lazy?

dirkgr (changes requested):
This is a great start! Some things are missing before we can consider merging it.
 * passing tests
 * tests for the new functionality (maybe the same tests we have, just running with a different trainer?)
 * code formatting and all that stuff. Are you familiar with `black` and `flake8`?
 * left-over debug code
 * Maybe we can do something about the massive code duplication? It doesn't have to be a primary goal, but there might be low-hanging fruit. In particular, could you highlight for me where the differences are between `GradentDescentTrainer` and `DeepspeedTrainer`? If they are all in the configuration and initialization, maybe we can come up with a much lighter-weight approach to getting this done.

dirkgr:
> I haven't looked at Sparse Attention / the transformer kernel yet (in allennlp or in general).

These are special `nn.Module`s that work particularly well with DeepSpeed?

allennlp/training/deepspeed/checkpointer.py
jacobdanovitch: The checkpointing works for saving; it's able to go through the training process E2E, doing the checkpointing and so on. I'm just not sure how model-parallel affects this part, if it's saving the entire model state or just the state local to that device. I imagine that this could be validated in a test case.

allennlp/training/deepspeed/checkpointer.py
jacobdanovitch: Yes, it delegates as much as possible to deepspeed's checkpointer, which is fairly complicated for the reasons mentioned above (mostly model-parallel).

allennlp/training/deepspeed/checkpointer.py
jacobdanovitch: It is derived from the regular checkpointer. I might be able to clean more of this up depending on the above points; if I didn't have to re-load the torch weights and could delegate almost entirely to deepspeed, it would simplify things quite a lot.

allennlp/training/deepspeed/config.py
jacobdanovitch: Yeah but it's a required install for deepspeed and you can use it there, so I thought I would keep it in for compatibility. It can be removed if need be.

allennlp/training/deepspeed/sparse_transformer_embedder.py
jacobdanovitch: This is just a question of scope. I think it would be a very neat little addition for not much cost, but I've been having trouble getting it installed properly (doing all of this on a SLURM cluster is leading to most of these problems, esp. not being able to use Docker). Fwiw, someone in the issue thread was expressing interest in this.

So overall it's just a question of if you'd like me to give it a shot to get it working now, or leave it for the future. Happy either way.

allennlp/training/deepspeed/trainer.py
jacobdanovitch: Ah, just leftover. Deepspeed handles both of those things in `load_checkpoint`. Removed.

allennlp/training/deepspeed/utils.py
jacobdanovitch: Yep, I'll just move this to the constructor.

allennlp/training/deepspeed/trainer.py
jacobdanovitch: Yeah, they need the batch size and number of gradient accumulation steps (which I get from this constructor and the dataloader), so Lazy felt right here. I could set the attributes manually if we're specifically avoiding Lazy objects.

allennlp/training/__init__.py
jacobdanovitch: Depends on how we want to include this. Based on my experience, I wouldn't recommend making deepspeed a required dependency. If we're doing the `pip install allennlp[deepspeed]` thing, this could be replaced/updated (not sure offhand how that gets handled but I can look for some examples).

jacobdanovitch:
Thanks for looking it over! I'll start linting everything and getting the tests up and running (we can probably re-use the existing Trainer tests, yeah).

As for the code duplication, the biggest overlaps are: 
* Checkpointing (I don't think this will be too duplicated when all is said and done)
* Certain parts of `_train_epoch`
* Almost all of `_validation_loss`

With Deepspeed, their training looks like:

```python
for batch in data_loader:
    loss = model_engine(**batch) # <- same as GDTrainer, but different attribute
    model_engine.backward(loss) # <- different
    model_engine.step() # <- different
```

The `model_engine` handles gradient accumulation, learning rate scheduling, optimization, etc, so these specific parts end up different but then the rest is all the same. That includes stuff in `_train_epoch` like:
* CPU/GPU memory usage
* Regularization penalty
* Tensorboard logging
* Metrics / progress bar stuff

And then almost all of `_validation_loss` is duplicated, I'm pretty sure the only change I made was `model_engine.backwards(loss)`. 

I think I'll be able to inherit `train` directly as well if I inherit from the GD trainer, but I wouldn't be able to use its constructor, so I'd have to do something like:

```python
@Trainer.register("deepspeed", constructor="from_partial_objects")
class DeepspeedTrainer(GradientDescentTrainer):
    Trainer.__init__(self, serialization_dir, cuda_device, distributed, local_rank, world_size)
```

I don't immediately see any problems with that, so if that sounds good that should help reduce duplication too. Overall, the problematic overlap  is where I just have to call one thing differently, like `model_engine.backwards(loss)`. Lightning's [hooks](https://pytorch-lightning.readthedocs.io/en/stable/lightning_module.html?highlight=hooks#hooks) pattern helps reduce this a bit, but that would be a bit more of a refactor than just changing a few lines.

allennlp/training/deepspeed/optimizers.py
jacobdanovitch: > > I haven't looked at Sparse Attention / the transformer kernel yet (in allennlp or in general).
> 
> These are special `nn.Module`s that work particularly well with DeepSpeed?

More or less, as far as I understand they're heavily optimized CUDA kernels that help for things like long sequences / are more efficient in general.

jacobdanovitch:
> These are special `nn.Module`s that work particularly well with DeepSpeed?

More or less, as far as I understand they're heavily optimized CUDA kernels that help for things like long sequences / are more efficient in general.

dirkgr:
```Python
class DeepspeedTrainer(GradientDescentTrainer):
    Trainer.__init__(self, serialization_dir, cuda_device, distributed, local_rank, world_size)
```
I don't understand this. Do you mean you would not be able to do this normal thing?
```Python
class DeepspeedTrainer(GradientDescentTrainer):
    def __init__(...):
       super().__init__(...)
```

allennlp/training/__init__.py
dirkgr: If you don't mind doing the work making it optional, then let's make it optional.

allennlp/training/deepspeed/checkpointer.py
dirkgr: That seems important. I don't want to make a release claiming that this works and then it doesn't in a fairly common use case.

allennlp/training/deepspeed/checkpointer.py
dirkgr: How do you find out whether you can do this?

I suspect that deep speed will work best the more stuff we delegate to it.

allennlp/training/deepspeed/config.py
dirkgr: Hmm. Surely in the next DeepSpeed version they will make it use PyTorch-native AMP. But if we need it for now, that's cool.

allennlp/training/deepspeed/sparse_transformer_embedder.py
dirkgr: I think it would be a great addition, but let's not make this PR bigger than it already is. It would be a nice follow-up though, especially if those classes end up being the secret to those massive promises that they have on their website.

You can make PRs that branch off of other PRs, just in case you didn't know :-P

allennlp/training/deepspeed/trainer.py
dirkgr: We don't go out of our way to avoid using them, but if we don't have to use them, we don't, because it's hard to debug what's going on. I'll leave it to your judgement. I thought these were just user-specified data classes, in which case it would be a bit weird to make them lazy, but if they are filled automatically, it makes more sense.

dirkgr:
If it's too difficult to not duplicate code, let's not do it. I looked at the code for `_validation_loss`, and while it would be a shame to have all those lines twice, it's pretty tightly integrated there, and I wouldn't want to compromise the regular trainer for this too much.

jacobdanovitch:
Still working on deduplicating code (and linting). I was able to get a lot reduced almost the entire constructor) by lying to the `super().__init__()` and passing `distributed=False` so that it wouldn't try to setup `DDP`.  Also gave up on the sparse attention embedder, too much stuff needed to be installed with root and I couldn't get around it.

Just so I know explicitly how far I should go, should I just not touch anything at all inside of the `GradientDescentTrainer`? The stuff inside of `for batch_group in batch_group_generator_tqdm:` in `_try_train` could potentially be trainer-dependent while everything outside of and around it should be common to all trainers,  so that could be a useful thing to put in a hook-like method to prevent a bunch of repetition. But if you'd rather I just not touch anything at all I'll leave it.

Some nice news is that deepspeed is now pip installable, so it's a lot easier to get everything configured.

dirkgr:
I'm fine with small modifications to the regular trainer, but what you're proposing sounds like a bigger deal, so let's hold off on that. We may want to re-visit when and if deep speed proves permanently useful.

jacobdanovitch:
Got all the typechecks out the way, phew. I've also managed to cut out a lot of duplicated code, I think! The remainder is almost entirely checkpointing related. For loading/saving, there's a bit of duplication here and there but nothing overwhelming, and the rest is delegated to deepspeed. The last thing that would be really, really nice to get around would be the `dist.barrier()` calls in `GradientDescentTrainer`:

Lines [1006-1008](https://github.com/allenai/allennlp/blob/master/allennlp/training/trainer.py#L1006), [1091-1093](https://github.com/allenai/allennlp/blob/master/allennlp/training/trainer.py#L1091)
```
# Wait for the master to finish saving the model checkpoint
            if self._distributed:
                dist.barrier()
```

I could be wrong with my limited understanding of DDP, but as far as I can tell, this causes a fatal hang for deepspeed, which also calls `dist.barrier()` within checkpointing. This seems like it could be a fairly common situation for anyone trying to override the trainer for any distributed-related stuff (like Fairscale or maybe the upcoming huggingface model parallelism stuff). 

Do you think there's a clean solution to this? That's about ~150 LOC duplicated for the removal of 4 lines, which isn't great. Is there a way that this could be delegated to the checkpointer itself, perhaps? Once that's settled, it should just be tests / dependencies left to do.

dirkgr:
Is there a way to detect whether we are in a deepspeed context? If so, I'd be OK with some sort of `if not in_deepspeed:`. Otherwise, let's just duplicate it.

allennlp/training/__init__.py
jacobdanovitch: > Is there a way to detect whether we are in a deepspeed context? If so, I'd be OK with some sort of `if not in_deepspeed:`. Otherwise, let's just duplicate it.

I mean, one easy way would just be to set an environment variable `DEEPSPEED=1` :stuck_out_tongue: Would that work, or would you rather more a more general solution?

jacobdanovitch:
> Is there a way to detect whether we are in a deepspeed context? If so, I'd be OK with some sort of `if not in_deepspeed:`. Otherwise, let's just duplicate it.

I mean, one easy way would just be to set an environment variable `DEEPSPEED=1` :stuck_out_tongue: Would that work, or would you rather more a more general solution?

dirkgr:
If deepspeed doesn't have some sort of global context (like `dist` does), then let's duplicate the code. I'm not that comfortable with inventing our own global flags, but if they are already there, defined and documented by major libraries, I'm OK using them.

jacobdanovitch:
Sounds good. I think Deepspeed might set some environment variables itself, similarly to torch, so I'll poke around to see if we can use one of those. If not we can just duplicate for now and I'll proceed onto testing.

jacobdanovitch:
Took a holiday break from this while our cluster was down for maintenance for a bit. Turns out that checkpointing/barrier issue might be more complicated than I thought, but not sure if it's something to do with our cluster (the networking seems buggy, `all_reduce` and such often hang outside of allennlp for me). It's freezing while collecting memory usage now, which is odd, so still trying to figure that out. 

Outside of that, I have a basic test working (well, when the above works), but it's more complicated than the existing trainer tests because all it's really doing is testing distributed training, which requires `dist.init_process_group` and the whole works. I don't see that being done for the gradient descent trainer, so not sure if I should be doing that.

dirkgr:
Collecting memory usage often defers to shelling out to `nvidia-smi`, which holds a lock. If you have a lot of processes calling `nvidia-smi`, you can see deadlocks. We’ve had this happen on our clusters as well.

Distributed training has tests. It’s tested from the test for the training command, not on the trainer directly: https://github.com/allenai/allennlp/blob/main/tests/commands/train_test.py#L191

You could do the same thing for the deep speed trainer. Just test it end-to-end, not individually.

jacobdanovitch:
Ah I think I see the real issue here. It's not the logging itself hanging.

1. (All ranks) My trainer tells my checkpointer to save **if** it's the master process
2. (Rank 0) My checkpointer delegates to `DeepspeedEngine.save_checkpoint`.
3. (Rank 0) `DeepspeedEngine.save_checkpoint` calls `dist.barrier()`
4. (Ranks 1-n) The other workers never get a chance to make the above barrier call, which leads to a lock as reported in pytorch [here](https://github.com/pytorch/pytorch/issues/20630#issuecomment-730184591).

Any subsequent distributed operation, including the `all_reduce` in memory logging, then hangs (not sure how they make it through the barriers honestly). So, in fact, the real culprits were not the `dist.barrier()` calls I mentioned above, but rather the `if self._master and self._checkpointer is not None:` checks. Everything works perfectly when removing the first half. 

So circling back to the code duplication issue, it's not so much being in a deepspeed context that I need to check for, it's something like:

```python

# trainer.py => _try_train

if self._checkpointer is not None and self._checkpointer.call_on_rank(self._rank):

# checkpointer

class Checkpointer(Registrable):
    def call_on_rank(self, rank: int) -> bool:
        return rank == 0

# deepspeed checkpointer

class DeepspeedCheckpointer(Checkpointer):
    @overrides
    def call_on_rank(self, rank: int) -> bool:
        return True
```

So if you're open to adding some sort of flag like that to the base checkpointer, that would solve the issue. Or, we could check if it's a deepspeed checkpointer:

```python
if self._checkpointer is not None and (self._master or isinstance(self._checkpointer, DeepspeedCheckpointer)):
```

But that might cause some circular import issues/issues for those who want to install without deepspeed. Either one would let me completely eliminate my override of `_try_train`; as long as I have a way to force `self._checkpointer.save_checkpoint(epoch, self)` to get called in every process.

<hr/>

> You could do the same thing for the deep speed trainer. Just test it end-to-end, not individually.

Yep this worked perfectly, thanks. Exact same as `test_train_model_distributed` but different config. Should I decorate it with `@requires_multi_gpu`?

dirkgr:
Why isn't the checkpointing thing a problem outside of AllenNLP? This should be an issue with DeepSpeed all the time, right?

> Should I decorate it with @requires_multi_gpu?
It makes sense to me even abstractly that something like Deepspeed can only be accurately tested on a multi-GPU box.

jacobdanovitch:
> Why isn't the checkpointing thing a problem outside of AllenNLP? This should be an issue with DeepSpeed all the time, right?

Their typical training loop is something like ([source](https://www.deepspeed.ai/getting-started/#model-checkpointing)):

```python
# load checkpoint
for step, batch in enumerate(data_loader):
    loss = model_engine(batch)
    model_engine.backward(loss)
    model_engine.step()

    if step % args.save_interval:
        ckpt_id = loss.item()
        model_engine.save_checkpoint(args.save_dir, ckpt_id)
```

Note that all of these calls are made on every worker, so every worker enters the `model_engine.save_checkpoint` function and hit the `dist.barrier` call(s) contained within. I believe that if you did `if step % args.save_interval and dist.get_rank() == 0` then you'd have the same problem. It's sort of a case of Deepspeed and AllenNLP both trying to help the user and running into each other as a result.

dirkgr:
I see. We can always determine our rank, right? So we could just move the rank check into the checkpointer. The regular checkpointer will say `if rank() != 0: return`, and the DeepSpeed one will proceed with all the ranks. Does that work?

jacobdanovitch:
Yeah that should work perfectly, I'll give it a try.

yanaiela:
Hey, are there any news regarding this PR?
I'd be interested in using Deepspeed with Allennlp.

dirkgr:
@epwalsh is testing it as we speak!

jeffra45:
@dirkgr @yanaiela @jacobdanovitch any blockers on this PR? Happy to help answer any deepspeed related questions that might be causing issues here.

jeffra:
@dirkgr @yanaiela @jacobdanovitch any blockers on this PR? Happy to help answer any deepspeed related questions that might be causing issues here.

dirkgr:
We recently integrated FairScale to get the ZeRO optimizer into AllenNLP. It would be interesting to have DeepSpeed as well, since it has more features, but it's no longer quite so pressing. If anyone wants to pick up @jacobdanovitch's work and bring it over the finish line, I'd be happy to work with you.

Issue #4666, opened 2020-09-23
When running multiple distributed runs on the same box, AllenNLP says "RuntimeError address already in use”


dirkgr:
The workaround is to set `"distributed" { "master_port": 29501 }` in the training config, but we can do better and automatically find a free port.

PR #4696, opened 2020-10-02
automatically find local open port in distributed training

epwalsh:
Closes https://github.com/allenai/allennlp/issues/4666

### DIFF
--- a/allennlp/commands/train.py
+++ b/allennlp/commands/train.py
@@ -257,9 +257,14 @@ def train_model(
         check_for_gpu(device_ids)
 
         master_addr = distributed_params.pop("master_address", "127.0.0.1")
-        master_port = distributed_params.pop(
-            "master_port", common_util.find_open_port(host=master_addr)
-        )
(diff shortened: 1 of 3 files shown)


allennlp/commands/train.py
```diff
@@ -257,7 +257,13 @@ def train_model(
         check_for_gpu(device_ids)
 
         master_addr = distributed_params.pop("master_address", "127.0.0.1")
-        master_port = distributed_params.pop("master_port", 29500)
+        if master_addr in ("127.0.0.1", "0.0.0.0", "localhost"):
+            # If running locally, we can automatically find an open port if one is not specified.
+            master_port = distributed_params.pop("master_port", common_util.find_open_port())
```
dirkgr, line 262: This will run `find_open_port()` before it tries to read the port from the config. I think we should not run port discovery at all if we don't have to.

dirkgr (changes requested):

allennlp/commands/train.py
epwalsh: Good point

dirkgr (approved):

PR #4697, opened 2020-10-02
ensure parameters that are actual dictionaries get logged

epwalsh:


matt-gardner (approved):
I don't remember why we turned off logging for this; I don't see any reason to keep it off.  We also are in the position of calling `.as_dict()` much less often now, I think.  Just curious - where did you run into this?

epwalsh:
I noticed when I was training a transformer model on SQuAD. The model took a dictionary of `tokenizer_kwargs`, but I noticed that wasn't showing in the logs.

PR #4702, opened 2020-10-05
Developer-friendly changes

eladsegal:
This pull request fixes a few things that I found missing/problematic during development with AllenNLP.
Some of the changes can be achieved without changing the library, however it isn't as clean.
The changes are small and shouldn't interfere with anything (they are also pretty separated by commits, but let me know if you prefer that I split it to multiple pull requests):
- Pass serialization_dir to Model and DatasetReader: Serves as a working directory for either classes (https://github.com/allenai/allennlp/pull/4702/commits/459c6809ee8175164a41c1eb17f18aa125ff501b)
- Define paths in serialization_dir that should be archived, in addition to config/best.th/vocab: Complements using serialization_dir as a working directory. (https://github.com/allenai/allennlp/pull/4702/commits/e5e6dcd8cb02ac0c2d576159b58d88b048bd38ab)
- When running in distributed mode, log also the stuff that happens before multiprocessing: This change logs the first DatasetReader run, and also logs an error in case it happens in of the multiprocess workers. (https://github.com/allenai/allennlp/pull/4702/commits/8e938c1ea729beabdfc31a4d3425b19095a55cf7)
- When using PretrainedModelInitializer, load weights to the cpu explicitly and save usage of expensive GPU memory. (https://github.com/allenai/allennlp/pull/4702/commits/ee580de81e40a0eddddd1716439c9f4891e734b5)
- Pretrained embeddings are not loaded when loading a model, but pretrained weights from PretrainedModelInitializer unnecessarily are: With this change they are both treated the same. https://github.com/allenai/allennlp/pull/4702/commits/bb23e9b9ac14fcc12322956b626fba1c599d0a22
- Allow usage of .tar.gz with PretrainedModelInitializer. (https://github.com/allenai/allennlp/pull/4702/commits/e6a0c076cc6a1fbece083c392283739745a92b8b)

I also snuck in a small bug fix - moved to https://github.com/allenai/allennlp/pull/4705:
common.util.is_master returned wrong result when there was only a single node. It was only used when saving vocabulary, so no harm done.

---
There's another developer-friendly change I wanted to make:
- Save training and model checkpoints *before* evaluation: Sometimes an error occurs when evaluating on the dev set. This change prevents the loss of the model weights of the epoch, and allows for easier debugging to find exactly what has caused the error.

I saw it is more complex than the other changes since it requires separating checkpointing, so I did it an another pull request: https://github.com/allenai/allennlp/pull/4704

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   be used from the `cached-path` command with `allennlp cached-path --inspect`.
 - Added a function `remove_cache_entries` to `common.file_utils` that removes any cache entries matching the given
   glob patterns. This can used from the `cached-path` command with `allennlp cached-path --remove some-files-*`.
+- Developer-friendly changes
 
 ### Changed
 
--- a/allennlp/commands/evaluate.py
+++ b/allennlp/commands/evaluate.py
@@ -135,9 +135,9 @@ def evaluate_from_args(args: argparse.Namespace) -> Dict[str, Any]:
     # to the default dataset_reader used for both training and validation.
     validation_dataset_reader_params = config.pop("validation_dataset_reader", None)
     if validation_dataset_reader_params is not None:
-        dataset_reader = DatasetReader.from_params(validation_dataset_reader_params)
+        dataset_reader = DatasetReader.from_params(validation_dataset_reader_params, serialization_dir=archive.serialization_dir)
     else:
-        dataset_reader = DatasetReader.from_params(config.pop("dataset_reader"))
+        dataset_reader = DatasetReader.from_params(config.pop("dataset_reader"), serialization_dir=archive.serialization_dir)
     evaluation_data_path = args.input_file
     logger.info("Reading evaluation data from %s", evaluation_data_path)
     instances = dataset_reader.read(evaluation_data_path)
--- a/allennlp/commands/evaluate.py
+++ b/allennlp/commands/evaluate.py
@@ -135,9 +135,13 @@ def evaluate_from_args(args: argparse.Namespace) -> Dict[str, Any]:
     # to the default dataset_reader used for both training and validation.
     validation_dataset_reader_params = config.pop("validation_dataset_reader", None)
     if validation_dataset_reader_params is not None:
-        dataset_reader = DatasetReader.from_params(validation_dataset_reader_params, serialization_dir=archive.serialization_dir)
+        dataset_reader = DatasetReader.from_params(
+            validation_dataset_reader_params, serialization_dir=archive.serialization_dir
+        )
     else:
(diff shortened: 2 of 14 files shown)


eladsegal:
Hi @matt-gardner @epwalsh, can you please take a look and tell me what you think?

epwalsh:
Hey @eladsegal, the bug you found with `is_master()` is actually pretty substantial IMO, and I think it should be it's own PR. Do you mind making a separate PR with just that?

eladsegal:
> Hey @eladsegal, the bug you found with `is_master()` is actually pretty substantial IMO, and I think it should be it's own PR. Do you mind making a separate PR with just that?

Done: https://github.com/allenai/allennlp/pull/4705

epwalsh:
Hey @eladsegal, these should all really be in separate PRs. Also, I'm curious what use case you have where the model and dataset reader need to know where the serialization directory is?

eladsegal:
> Hey @eladsegal, these should all really be in separate PRs. Also, I'm curious what use case you have where the model and dataset reader need to know where the serialization directory is?

Hey @epwalsh,
I've created these new pull requests:
- Pass serialization_dir to Model and DatasetReader, and support `additional_archival_targets` (https://github.com/allenai/allennlp/pull/4713, combined the two as I realized the first isn't helpful without the second)
- Added logging for the main process when running in distributed mode (https://github.com/allenai/allennlp/pull/4710)
- Load weights to cpu with PretrainedModelInitializer (https://github.com/allenai/allennlp/pull/4712)
- Don't use PretrainedModelInitializer when loading a model (https://github.com/allenai/allennlp/pull/4711 - tests are failing due to a name change affecting allennlp-models, created also https://github.com/allenai/allennlp-models/pull/141)
- Allow usage of .tar.gz with PretrainedModelInitializer (https://github.com/allenai/allennlp/pull/4709)

Regarding your question -
When using huggingface's transformers with allennlp, I sometimes want to add special tokens to the vocabulary.
While it is simple to add tokens in the dataset reader, there are two problems with it:
- Each time the dataset reader is instantiated, these tokens will have to be added. It is preferable to load an already-created tokenizer with all of its settings than creating one for `dataset_reader` and then again for `validation_dataset_reader`.
-  The model requires the same tokenizer as the dataset readers, as it needs to have the right vocabulary size in order to update the token embeddings matrix accordingly, and it also needs the new special tokens and their mapping from the tokenizer to use for decoding. So there has to be a way to pass this information to the model.

The simplest solution I found was to save the tokenizer using huggingface's `save_pretrained`, and load it with `from_pretrained`. The best place to do it is in the serialization dir, and also add the tokenizer to the archive so the trained model and its dataset reader will always have access to the same tokenizer the training was based on.

While it is not a common usage, I believe that it can be helpful for people using allennlp in more advanced ways.

dirkgr:
Thank you for doing this!

PR #4705, opened 2020-10-05
Fix a bug in saving vocab during distributed training

eladsegal:
common.util.is_master returned wrong result when there was only a single node. It was only used when saving vocabulary, so I think no harm done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `transformers` dependency updated to version 3.1.0.
 - When `cached_path` is called on a local archive with `extract_archive=True`, the archive is now extracted into a unique subdirectory of the cache root instead of a subdirectory of the archive's directory. The extraction directory is also unique to the modification time of the archive, so if the file changes, subsequent calls to `cached_path` will know to re-extract the archive.
 - Removed the `truncation_strategy` parameter to `PretrainedTransformerTokenizer`. The way we're calling the tokenizer, the truncation strategy takes no effect anyways.
+- Removed common.util.is_master, use `local_rank == 0` instead.
 
 ### Fixed
 
@@ -55,7 +56,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   at around the 9 minute mark.
 - Fixed a bug where parameters to a `FromParams` class that are dictionaries wouldn't get logged
   when an instance is instantiated `from_params`.
-- Fix a bug in common.util.is_master that considered every worker as master when there was only a single node. 
 
 
 ## [v1.1.0](https://github.com/allenai/allennlp/releases/tag/v1.1.0) - 2020-09-08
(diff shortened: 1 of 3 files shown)


epwalsh:
@eladsegal, I didn't realize this function was only used in a single place, and I think it's actually not a good function anyway, because it's not clear if `is_master()` means "local master" or "global master".

I think we should actually just get rid of this function, and instead just check if `local_rank == 0`. What do you think?

eladsegal:
@epwalsh If the purpose is to identify the master process in a node, this function effectively checks whether `local_rank == 0`. I am not familiar with a more direct way to do it.
However, I am not sure when one would even like to know this information, so I agree that we can get rid of it.
I don't think that `is_master`, in its single usage, is even used correctly - https://github.com/allenai/allennlp/blob/3506e3fd2d4a4bc205123f41767b2d005f3c79dc/allennlp/commands/train.py#L655-L657
I believe the intention here was to check for the global master (i.e. global rank 0), because I don't see a reason to save the vocab again for each node.

epwalsh:
> I believe the intention here was to check for the global master (i.e. global rank 0), because I don't see a reason to save the vocab again for each node.

Actually each node needs a local copy of the vocab, so the intent is to check if local rank is 0.

eladsegal:
> > I believe the intention here was to check for the global master (i.e. global rank 0), because I don't see a reason to save the vocab again for each node.
> 
> Actually each node needs a local copy of the vocab, so the intent is to check if local rank is 0.

Oh, now I get it. I thought of it differently because I use a network directory, so the serialization_dir is the same for all nodes.

So given the fact the local rank is indeed required, I don't really see an alternative to `is_master`.

epwalsh:
What I'm saying is that we could just replace the `if common_util.is_master():` line with `if local_rank == 0:` here:

https://github.com/allenai/allennlp/blob/3506e3fd2d4a4bc205123f41767b2d005f3c79dc/allennlp/commands/train.py#L655-L657

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `transformers` dependency updated to version 3.1.0.
 - When `cached_path` is called on a local archive with `extract_archive=True`, the archive is now extracted into a unique subdirectory of the cache root instead of a subdirectory of the archive's directory. The extraction directory is also unique to the modification time of the archive, so if the file changes, subsequent calls to `cached_path` will know to re-extract the archive.
 - Removed the `truncation_strategy` parameter to `PretrainedTransformerTokenizer`. The way we're calling the tokenizer, the truncation strategy takes no effect anyways.
+- Removed common.util.is_master, use `local_rank == 0` instead.
```
epwalsh, line 38:
```suggestion

### Removed

- Removed `common.util.is_master` function.
```

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ### Fixed
```
epwalsh, line 41:
```suggestion

- Fixed a bug in distributed training where the vocab would be saved from every worker, when it should have been saved by only the local master process.
```

eladsegal:
OK, I agree. It also aligns with https://github.com/allenai/allennlp/blob/3506e3fd2d4a4bc205123f41767b2d005f3c79dc/allennlp/commands/train.py#L378

I updated the pull request and removed `is_master`.

epwalsh (changes requested):
Thanks @eladsegal! I just have a little nit-pick to clarify the CHANGELOG and then we can get this merged 👍

epwalsh (approved):

PR #4706, opened 2020-10-05
fix reported batch_loss with gradient accumulation

epwalsh:
Currently, when training with gradient accumulation (i.e. with `num_gradient_accumulation_steps > 1`), the reported `batch_loss` metric is always just the loss from last batch in the batch group divided by the number of batches in the group. By itself, this is arbitrary and kind of meaningless. Instead, `batch_loss` should report the cumulative loss for the entire group.

By the way, this came up in https://stackoverflow.com/questions/64181843/calculate-loss-in-train-epoch-function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


dirkgr (approved):
Nice find!

PR #4713, opened 2020-10-07
Pass serialization_dir to Model, DatasetReader, and support `include_in_archive`

eladsegal:


dirkgr:
This depends on #4709.

dirkgr (changes requested):
Let's call this parameter "`include_in_archive`" instead. Sorry to be a pain in the ass about names, but these could stick with us for a long time.

What do you actually need this for? What is the functionality for?

eladsegal:
> Let's call this parameter "`include_in_archive`" instead. Sorry to be a pain in the ass about names, but these could stick with us for a long time.
> 
> What do you actually need this for? What is the functionality for?

Sure, no problem.

Regarding my usage -
When using huggingface's transformers with allennlp, I sometimes want to add special tokens to the vocabulary.
While it is simple to add tokens in the dataset reader, there are two problems with it:
- Each time the dataset reader is instantiated, these tokens will have to be added. It is preferable to load an already-created tokenizer with all of its settings than creating one for dataset_reader and then again for validation_dataset_reader.
- The model requires the same tokenizer as the dataset readers, as it needs to have the right vocabulary size in order to update the token embeddings matrix accordingly, and it also needs the new special tokens and their mapping from the tokenizer to use for decoding. So there has to be a way to pass this information to the model.

The simplest solution I found was to save the tokenizer using huggingface's `save_pretrained`, and load it with `from_pretrained`. The best place to do it is in the serialization dir, and also add the tokenizer to the archive so that the trained model and its dataset reader will always have access to the same tokenizer the training was based on.

While it is not a common usage, I believe that it can be helpful for people using allennlp in more advanced ways.

dirkgr:
Interesting. We have long been considering an option to let the `DatasetReader` return a `Vocabulary` object (instead of discovering the vocabulary automatically). So in your case, you would use the huggingface tokenizer in the reader, and in your `DatasetReader.get_vocab()`, you would do something like `return Vocabulary.from_huggingface(self._tokenizer)`. This is the object that gets passed to the model.

Would this solve your problem?

eladsegal:
> Interesting. We have long been considering an option to let the `DatasetReader` return a `Vocabulary` object (instead of discovering the vocabulary automatically). So in your case, you would use the huggingface tokenizer in the reader, and in your `DatasetReader.get_vocab()`, you would do something like `return Vocabulary.from_huggingface(self._tokenizer)`. This is the object that gets passed to the model.
> 
> Would this solve your problem?

Yes, sounds like it would solve the problem.
However, I still like the idea of being able to have files in the serialization_dir that both the dataset reader and the model can use and can be archived too.

eladsegal:
Hey @epwalsh, as Dirk is on vacation, can you please finish the review for this?

allennlp/commands/train.py
```diff
@@ -371,6 +376,8 @@ def _train_worker(
     file_friendly_logging : `bool`, optional (default=`False`)
         If `True`, we add newlines to tqdm output, even on an interactive terminal, and we slow
         down tqdm's output to only once every 10 seconds.
+    include_in_archive : `List[str]`, optional
+        Relative paths to serialization_dir that should be archived in addition to the default ones.
```
epwalsh, line 380:
```suggestion
        Paths relative to `serialization_dir` that should be archived in addition to the default ones.
```

allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -163,6 +164,7 @@ def __init__(
             os.makedirs(self._cache_directory, exist_ok=True)
         self.manual_distributed_sharding = manual_distributed_sharding
         self.manual_multi_process_sharding = manual_multi_process_sharding
+        self._serialization_dir = serialization_dir
```
epwalsh, line 167: Let's make this consistent with the other attributes:

```suggestion
        self.serialization_dir = serialization_dir
```

allennlp/models/archival.py
```diff
@@ -92,10 +92,22 @@ def extract_module(self, path: str, freeze: bool = True) -> Module:
...
+            raise ConfigurationError(
+                f"{', '.join(saved_names)} are saved names and cannot be used for include_in_archive."
+            )
+
+
 def archive_model(
     serialization_dir: Union[str, PathLike],
     weights: str = _DEFAULT_WEIGHTS,
     archive_path: Union[str, PathLike] = None,
+    include_in_archive: Optional[List[str]] = None,
```
epwalsh, line 110: Missing a description of this in the docstsring.

allennlp/models/archival.py
```diff
@@ -132,6 +144,12 @@ def archive_model(
         archive.add(weights_file, arcname=_WEIGHTS_NAME)
         archive.add(os.path.join(serialization_dir, "vocabulary"), arcname="vocabulary")
 
+        if include_in_archive is not None:
+            for archival_target in include_in_archive:
+                archival_target_path = os.path.join(serialization_dir, archival_target)
+                if os.path.exists(archival_target_path):
+                    archive.add(archival_target_path, arcname=archival_target)
```
epwalsh, line 151: I think it would be good to allow wildcard paths as well. You could use the `glob` module from the standard library.

allennlp/models/model.py
```diff
@@ -71,10 +71,16 @@ class Model(torch.nn.Module, Registrable):
     _warn_for_unseparable_batches: Set[str] = set()
     default_predictor: Optional[str] = None
 
-    def __init__(self, vocab: Vocabulary, regularizer: RegularizerApplicator = None) -> None:
+    def __init__(
+        self,
+        vocab: Vocabulary,
+        regularizer: RegularizerApplicator = None,
+        serialization_dir: Optional[str] = None,
```
epwalsh, line 78: Missing a description of this in the docstring.

epwalsh (changes requested):
Hey @eladsegal, just a few more minor comments.

allennlp/commands/train.py
```diff
@@ -225,6 +225,9 @@ def train_model(
     training_util.create_serialization_dir(params, serialization_dir, recover, force)
     params.to_file(os.path.join(serialization_dir, CONFIG_NAME))
 
+    include_in_archive = params["trainer"].get("include_in_archive", None)
```
epwalsh, line 228: Hey @eladsegal, I'm having trouble understanding this.

You have made `include_in_archive` a parameter of the `Trainer` class here, but it doesn't look like you actually added it to the `Trainer` class, nor does it seem like it should be a parameter of the `Trainer` class since the `Trainer` class doesn't seem to need it. This will probably cause `allennlp train` to fail when `include_in_archive` is used because it will be passed to the `Trainer`, resulting in `TypeError: Unexpected argument` or something.

Instead, it either needs to be a top-level field in `params` or a command-line flag `--include-in-archive`.

We should also add a test to `tests/commands/train_test.py`.

epwalsh (changes requested):

allennlp/commands/train.py
eladsegal: Instead of adding a test in `train_test.py`, I changed the test in `archival_test.py` to be like the rest of the tests which test archival via calling `train_model`.
It is not calling `archive_model` directly to use `include_in_archive`, but adds it to the params.

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   sampling indices from log probabilities
 - Made `BeamSearch` registrable.
 - Added `top_k_sampling` and `type_p_sampling` `BeamSearch` implementations.
+- Pass `serialization_dir` to `Model` and `DatasetReader`, and support `include_in_archive`.
```
epwalsh, line 40:
```suggestion
- Pass `serialization_dir` to `Model` and `DatasetReader`.
- Added an optional `include_in_archive` parameter to the top-level of configuration files. When specified, `include_in_archive` should be a list of paths relative to the serialization directory which will be bundled up with the final archived model from a training run.
```

epwalsh (commented):
Thank you so much for sticking with this @eladsegal! And I'm sorry to keep annoying you with requests 😉 

Just one more thing: could you please add a test to `tests/command/train_test.py` for training a model with `include_in_archive` specified?

epwalsh:
Ope, I just saw your previous comment about the test. Disregard my request.

epwalsh (approved):
Looks great, thanks @eladsegal!

Issue #4716, opened 2020-10-07
Docstring for AllenNlpTestCase is out-of-date


nicolasahar:
https://github.com/allenai/allennlp/blob/ae7cf85b8f755f086341a087aeca1ba7c1df3769/allennlp/common/testing/test_case.py#L14

This docstring suggests that the TestCase class should subclass `unittest.TestCase` but it does not do so. Should this be updated or is this intentional?

epwalsh:
The docstring is out-of-date. Would you be interested in making a PR to fix it?

nicolasahar:
> The docstring is out-of-date. Would you be interested in making a PR to fix it?

Done. PR #4722 [here](https://github.com/allenai/allennlp/pull/4722)

PR #4722, opened 2020-10-11
Updated AllenNlpTestCase docstring

nicolasahar:
Updated AllenNlpTestCase test case to remove reference to `unittest.TestCase` as per issue #4716


epwalsh (approved):
LGTM!

PR #4724, opened 2020-10-12
Adding stochastic beam search

jvstokes:
## Description
Adding stochastic beam search to `beam_search.py`

## Changes
`allennlp/nn/samplers/samplers.py`: added `GumbelMaxSampler`
`allennlp/nn/beam_search.py`: added `stochastic_beam_search`

## Tests
`tests/nn/beam_search_tests.py`: added basic tests for stochastic beam search

### DIFF
--- a/allennlp/nn/beam_search.py
+++ b/allennlp/nn/beam_search.py
@@ -242,7 +242,7 @@ class BeamSearch(Registrable):
             )
             if last_true_log_probabilities is not None:
                 return start_predicted_classes.unsqueeze(-1), last_true_log_probabilities
-                
+
             return start_predicted_classes.unsqueeze(-1), start_top_log_probabilities
 
         # The log probabilities for the last time step.
--- a/allennlp/nn/beam_search.py
+++ b/allennlp/nn/beam_search.py
@@ -494,6 +494,7 @@ class BeamSearch(Registrable):
         beam_size: int = 10,
         k: int = 1,
         temperature: float = 1.0,
+        with_replacement: bool = False,
     ) -> "BeamSearch":
         """
         Given an index of the end token in target vocabulary, return a `BeamSearch` object
@@ -506,7 +507,7 @@ class BeamSearch(Registrable):
                 f'{"value of selection threshold `k` invalid."}'
                 f'{"`k` must be a positive `int`."}'
             )
-        sampler_k = TopKSampler(k, temperature)
+        sampler_k = TopKSampler(k, temperature, with_replacement=with_replacement)
         return cls(
             end_index=end_index,
             max_steps=max_steps,
@@ -523,6 +524,7 @@ class BeamSearch(Registrable):
         beam_size: int = 10,
         p: float = 0.9,
         temperature: float = 1.0,
+        with_replacement: bool = False,
     ) -> "BeamSearch":
         """
         Given an index of the end token in target vocabulary, return a `BeamSearch` object
@@ -537,7 +539,7 @@ class BeamSearch(Registrable):
                 f'{"`p` must be a float between `0.0` and `1.0`"}'
             )
 
-        sampler_p = TopPSampler(p, temperature)
+        sampler_p = TopPSampler(p, temperature, with_replacement=with_replacement)
         return cls(
             end_index=end_index,
             max_steps=max_steps,
@@ -547,7 +549,7 @@ class BeamSearch(Registrable):
         )
 
     @classmethod
-    def stochastic_beam_search(
+    def stochastic(
         cls,
         end_index: int,
         max_steps: int = 50,
@@ -576,4 +578,4 @@ class BeamSearch(Registrable):
 BeamSearch.register("without_sampling", constructor="without_sampling")(BeamSearch)
 BeamSearch.register("top_p_sampling", constructor="top_p_sampling")(BeamSearch)
 BeamSearch.register("top_k_sampling", constructor="top_k_sampling")(BeamSearch)
-BeamSearch.register("stochastic_beam_search", constructor="stochastic_beam_search")(BeamSearch)
+BeamSearch.register("stochastic", constructor="stochastic")(BeamSearch)
--- a/allennlp/nn/beam_search.py
+++ b/allennlp/nn/beam_search.py
@@ -215,7 +215,9 @@ class BeamSearch(Registrable):
         # shape: (batch_size, beam_size), (batch_size, beam_size)
         if self.sampler is not None:
             start_top_log_probabilities, start_predicted_classes = self.sampler(
-                start_class_log_probabilities, start_class_log_probabilities, num_samples=self.beam_size
+                start_class_log_probabilities,
+                start_class_log_probabilities,
+                num_samples=self.beam_size,
             )
         else:
             (
@@ -324,13 +326,16 @@ class BeamSearch(Registrable):
 
             if last_true_log_probabilities is not None:
                 # add the last true probs to previous sequence probs
-                summed_true_log_probabilities = torch.add(last_true_log_probabilities, cleaned_log_probabilities)
-                
+                summed_true_log_probabilities = torch.add(
+                    last_true_log_probabilities, cleaned_log_probabilities
+                )
 
             # shape (both): (batch_size * beam_size, per_node_beam_size)
             if self.sampler is not None:
                 top_log_probabilities, predicted_classes = self.sampler(
-                    summed_true_log_probabilities, summed_log_probabilities, num_samples=self.per_node_beam_size
+                    summed_true_log_probabilities,
+                    summed_log_probabilities,
+                    num_samples=self.per_node_beam_size,
                 )
             else:
                 top_log_probabilities, predicted_classes = summed_log_probabilities.topk(
@@ -338,7 +343,7 @@ class BeamSearch(Registrable):
                 )
 
             summed_top_log_probabilities = top_log_probabilities
-            
+
             # shape: (batch_size, beam_size * per_node_beam_size)
             reshaped_summed = summed_top_log_probabilities.reshape(
                 batch_size, self.beam_size * self.per_node_beam_size
--- a/allennlp/nn/beam_search.py
+++ b/allennlp/nn/beam_search.py
@@ -1,5 +1,5 @@
 from inspect import signature
-from typing import List, Callable, Tuple, Dict, cast, TypeVar
+from typing import List, Callable, Tuple, Dict, cast, TypeVar, Optional
 import warnings
 
 import torch
@@ -214,9 +214,18 @@ class BeamSearch(Registrable):
         # Else, select the top `beam_size` tokens.
         # shape: (batch_size, beam_size), (batch_size, beam_size)
         if self.sampler is not None:
-            start_top_log_probabilities, start_predicted_classes = self.sampler(
-                start_class_log_probabilities, num_samples=self.beam_size
-            )
+            if isinstance(self.sampler, GumbelMaxSampler):
+                # phi represents the summed log probabilities of the sequence at hand
+                phi = start_class_log_probabilities
+
+                # Our sampler gets the perturbed log probabilities for each token
+                start_top_log_probabilities, start_predicted_classes = self.sampler(
+                    phi, num_samples=self.beam_size
+                )
+            else:
+                start_top_log_probabilities, start_predicted_classes = self.sampler(
+                    start_class_log_probabilities, num_samples=self.beam_size
+                )
         else:
             (
                 start_top_log_probabilities,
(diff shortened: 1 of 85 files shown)


allennlp/nn/beam_search.py
```diff
@@ -222,6 +223,11 @@ def _search(
                 start_predicted_classes,
             ) = start_class_log_probabilities.topk(self.beam_size)
 
+        if isinstance(self.sampler, GumbelMaxSampler):
+            last_true_probabilities = torch.gather(
+                start_class_log_probabilities, 1, start_predicted_classes
+            )
```
epwalsh, line 229: I actually think we should always gather `last_true_log_probabilities` since there is no guarantee with other samplers that they won't perturb the log probabilities.

```suggestion
        # Some samplers add perturb the log probabilities with random noise, such
        # as the GumbelMaxSampler, so we need to recover and keep track of the
        # true log probabilities here.
        last_true_log_probabilities = torch.gather(
            start_class_log_probabilities, 1, start_predicted_classes
        )
```

Would need a similar change below as well.

allennlp/nn/beam_search.py
```diff
@@ -327,6 +333,26 @@ def _search(
...
+                summed_true_log_probabilities = (
+                    cleaned_log_probabilities.gather(1, predicted_classes)
+                    + expanded_true_log_probabilities
+                )
+
+                # shape: (batch_size, beam_size * per_node_beam_size)
+                reshaped_summed_true_probabilities = summed_true_log_probabilities.reshape(
+                    batch_size, self.beam_size * self.per_node_beam_size
+                )
+
```
epwalsh, line 355: Same comment as above.

allennlp/nn/beam_search.py
```diff
@@ -344,6 +370,15 @@ def _search(
...
 
+            # track the true probabilities for the selected beams
+            # shape: (batch_size, beam_size)
+            if isinstance(self.sampler, GumbelMaxSampler):
+                restricted_true_probabilities = reshaped_summed_true_probabilities.gather(
+                    1, restricted_beam_indices
+                )
+
+                last_true_probabilities = restricted_true_probabilities
+
```
epwalsh, line 381: Same comment here as well.

allennlp/nn/beam_search.py
```diff
@@ -408,6 +443,10 @@ def _search(
         # shape: (batch_size, beam_size, max_steps)
         all_predictions = torch.cat(list(reversed(reconstructed_predictions)), 2)
 
+        # If stochastic beam search, return the true probabilities instead of perturbed
+        if isinstance(self.sampler, GumbelMaxSampler):
+            return all_predictions, last_true_probabilities
+
         return all_predictions, last_log_probabilities
```
epwalsh, line 450: With the above changes, we'd just always return `last_true_log_probabilities`.

```suggestion
        return all_predictions, last_true_log_probabilities
```

allennlp/nn/beam_search.py
```diff
@@ -489,7 +528,35 @@ def top_p_sampling(
...
+        per_node_beam_size: int = None,
+        temperature: float = 1.0,
+    ) -> "BeamSearch":
+        """
+        Given an index of the end token in target vocabulary, return a `BeamSearch` object
+        that can be used to find `beam_size` candidate sequences, found by using Stochastic
+        Beam Search, which leverages Gumbel-TopK sampling. See
+        [*Stochastic Beams and Where to Find Them: The Gumbel-Top-k Trick for Sampling
+        Sequences Without Replacement*, W Kool, H Van Hoof and M Welling, 2010]
+        (https://arxiv.org/abs/1903.06059).
```
epwalsh, line 546: I don't think this will render properly in the documentation. It might have to look like:

```suggestion
        [*Stochastic Beams and Where to Find Them: The Gumbel-Top-k Trick for Sampling Sequences Without Replacement*, W Kool, H Van Hoof and M Welling, 2010]
        (https://arxiv.org/abs/1903.06059).
```

You can check this by building and serving the docs locally with `make serve-docs` and then opening [http://127.0.0.1:8000/api/nn/beam_search/](http://127.0.0.1:8000/api/nn/beam_search/) in your browser.

allennlp/nn/beam_search.py
jvstokes: Should I still check `if self.sampler is not None:` ? Right now only the gumbel sampler will return perturbed log probabilities, and deterministic beam search will always keep track of the true probabilities

epwalsh: Ah true, yes let's do that

allennlp/nn/beam_search.py
jvstokes: This looks fine when viewing the docs, and the linter complains otherwise so I left it!

allennlp/nn/beam_search.py
```diff
@@ -222,6 +223,14 @@ def _search(
                 start_predicted_classes,
             ) = start_class_log_probabilities.topk(self.beam_size)
 
+        # Some samplers add perturb the log probabilities with random noise, such
+        # as the GumbelMaxSampler, so we need to recover and keep track of the
+        # true log probabilities here.
+        if self.sampler is not None:
+            last_true_probabilities = torch.gather(
+                start_class_log_probabilities, 1, start_predicted_classes
+            )
```
epwalsh, line 232: I think it's better to define this variable outside of this branch. I'm surprised Python allows this kind of thing, but then again... it's Python so I guess anything goes 😝

```suggestion
        last_true_probabilities: Optional[torch.Tensor] = None
        if self.sampler is not None:
            last_true_probabilities = torch.gather(
                start_class_log_probabilities, 1, start_predicted_classes
            )
```

allennlp/nn/beam_search.py
```diff
@@ -327,6 +336,27 @@ def _search(
                 batch_size, self.beam_size * self.per_node_beam_size
             )
 
+            # Track the true probabilities in addition to perturbed probabilities
+            # for use in samplers that may perturb probabilities
+            if self.sampler is not None:
```
epwalsh, line 341: With the above change, might make sense to check against null value of `last_true_probabilities` here to make mypy happy:

```suggestion
            if last_true_probabilities is not None:
```

allennlp/nn/beam_search.py
```diff
@@ -344,6 +374,15 @@ def _search(
                 1, restricted_beam_indices
             )
 
+            # track the true probabilities for the selected beams
+            # shape: (batch_size, beam_size)
+            if self.sampler is not None:
```
epwalsh, line 379:
```suggestion
            if last_true_probabilities is not None:
```

allennlp/nn/beam_search.py
```diff
@@ -408,6 +447,10 @@ def _search(
         # shape: (batch_size, beam_size, max_steps)
         all_predictions = torch.cat(list(reversed(reconstructed_predictions)), 2)
 
+        # If a sampler is used, return the true probabilities instead of perturbed probabilities
+        if self.sampler is not None:
```
epwalsh, line 451: Same here

```suggestion
        if last_true_probabilities is not None:
```

allennlp/nn/beam_search.py
epwalsh: Also, should this be `last_true_log_probabilities`, not `last_true_probabilities`?

jvstokes: Yup I'll change that variable name!

tests/nn/beam_search_test.py
```diff
@@ -277,6 +277,26 @@ def test_top_k_search(self):
...
+        beam_size = beam_size or 1
+        batch_size = 5
+
+        # top_p should be shape `(batch_size, beam_size, max_predicted_length)`.
+        assert list(top_k.size())[:-1] == [batch_size, beam_size]
+
+        assert ((0 <= top_k) & (top_k <= 5)).all()
+
+        # log_probs should be shape `(batch_size, beam_size, max_predicted_length)`.
+        assert list(log_probs.size()) == [batch_size, beam_size]
```
epwalsh, line 298: We should be able to verify exactly what `log_probs` should be given `top_k`, which would be good to do in this test. If that's too difficult though, we could have a similar test using a dummy sampler that does the same thing as deterministic beam search. Then we'd just have to check that the `log_probs` result is the same when using the dummy sampler vs using no sampler.

allennlp/nn/samplers/samplers.py
```diff
@@ -123,3 +130,42 @@ def __call__(
...
+@Sampler.register("gumbel-max")
+class GumbelMaxSampler(Sampler):
+    """
+    Represents a `Sampler` which uses the Gumbel-Max trick to sample `num_samples`
+    instances without replacement
+    `logits` is a tensor of log-probabilities to be selected from.
+    `temperature` modules the probabilitis of the selected tokens. A `temperature` below 1.0 produces a
+    sharper probability distribution and a `temperature` above 1.0 produces a flatter probability
+    distribution.
+    `num_samples` is the number of instances to sample
```
epwalsh, line 144: Maybe we should mention here that it returns perturbed log probs?

allennlp/nn/samplers/samplers.py
```diff
@@ -17,22 +17,26 @@ def __init__(self, temperature: float = 1.0, filter_val: float = -float("inf"))
...
     def __call__(
-        self, logits: torch.Tensor, num_samples: int = 1, with_replacement: bool = True
+        self,
+        log_probs: torch.Tensor,
+        perturbed_log_probs: torch.Tensor = None,
+        num_samples: int = 1,
+        with_replacement: bool = True,
     ) -> torch.Tensor:
-        probabilities = torch.nn.functional.softmax(logits, dim=-1)
+        probabilities = torch.nn.functional.softmax(log_probs, dim=-1)
```
epwalsh, line 26: `log_probs` are already normalized, so we can just do `log_probs.exp()` to get `probabilities`.

allennlp/nn/beam_search.py
```diff
@@ -213,15 +214,33 @@ def _search(
...
+                phi = start_class_log_probabilities
+
+                # Our sampler gets the perturbed log probabilities for each token
+                start_top_log_probabilities, start_predicted_classes = self.sampler(
+                    phi, num_samples=self.beam_size
+                )
+            else:
+                start_top_log_probabilities, start_predicted_classes = self.sampler(
+                    start_class_log_probabilities, num_samples=self.beam_size
+                )
```
epwalsh, line 228: Is this special case handling for `GumbelMaxSampler` still necessary?

And it looks like the way we call `self.sampler()` in the `else` branch is incorrect / inconsistent with the signature of `Sampler`.

allennlp/nn/beam_search.py
```diff
@@ -489,7 +546,34 @@ def top_p_sampling(
             sampler=sampler_p,
         )
 
+    @classmethod
+    def stochastic_beam_search(
```
epwalsh, line 550: Maybe just call this ⬇️ ?

```suggestion
    def stochastic(
```

allennlp/nn/beam_search.py
```diff
@@ -489,7 +546,34 @@ def top_p_sampling(
...
+            beam_size=beam_size,
+            per_node_beam_size=per_node_beam_size or beam_size,
+            sampler=gumbel_sampler,
+        )
+
 
 BeamSearch.register("without_sampling", constructor="without_sampling")(BeamSearch)
 BeamSearch.register("top_p_sampling", constructor="top_p_sampling")(BeamSearch)
 BeamSearch.register("top_k_sampling", constructor="top_k_sampling")(BeamSearch)
+BeamSearch.register("stochastic_beam_search", constructor="stochastic_beam_search")(BeamSearch)
```
epwalsh, line 579: Same suggestion here ⬇️

```suggestion
BeamSearch.register("stochastic", constructor="stochastic")(BeamSearch)
```

allennlp/nn/samplers/sampler.py
```diff
@@ -9,6 +9,10 @@ class Sampler(Registrable):
     """
 
     def __call__(
-        self, logits: torch.Tensor, num_samples: int = 1, with_replacement: bool = True
+        self,
+        log_probs: torch.Tensor,
+        perturbed_log_probs: torch.Tensor = None,
+        num_samples: int = 1,
+        with_replacement: bool = True,
```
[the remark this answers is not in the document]
epwalsh: Would need to be changed for all the subclasses too.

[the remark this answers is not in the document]
epwalsh: But sampling with replacement doesn't even make sense for Gumbel, right? So maybe we should just remove this argument entirely. It could be a parameter in the `__init__` method for samplers that it makes sense for.

epwalsh:
In `samplers/__init__.py` we are missing Gumbel.

allennlp/nn/samplers/sampler.py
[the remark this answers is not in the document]
jvstokes: You can change the with/without replacement for the other samplers, but stochastic is always without replacement. I'll switch it to be in the Init method.

allennlp/nn/samplers/samplers.py
```diff
@@ -109,17 +135,80 @@ def __call__(
...
+    `num_samples` is the number of instances to sample
+
+    Returns a tensor of the top `num_samples` perturbed log probabilities and their
+    indices within `log_probs`.
+
+    Registered as a `Sampler` with name "gumbel-max".
+    """
+
+    def __init__(self):
+        self.num_samples = 1
```
epwalsh, line 175: I don't think we need this.
```suggestion
```

epwalsh:
We should have a check here that `last_true_log_probabilities` is also finite: https://github.com/allenai/allennlp/pull/4724/files#diff-ccf1daaf744838c9a6be618bb929927a249b6c6dfeb6dc37fada88e660a3c5a1R451

epwalsh:
Closed via #4768

PR #4728, opened 2020-10-13
Transformer toolkit: BiModalEncoder now has separate `num_attention_heads` for both modalities

AkshitaB:


allennlp/modules/transformer/__init__.py
```diff
@@ -8,147 +8,117 @@
...
+
+        self.transformer = TransformerBlock.from_pretrained_module(
+            pretrained, num_hidden_layers=4
+        )
+
+    @overrides
+    def forward(self, token_ids: torch.LongTensor):
+        x = self.embeddings(token_ids)
+        x = self.transformer(x)
+        return x
```
dirkgr, line 56: I don't think this is completely correct, because it neglects token type IDs. That's not a thing to fix in this PR though.

allennlp/modules/transformer/__init__.py
```diff
@@ -8,147 +8,117 @@
...
+
+        left = self.embeddings(left_token_ids)
+        left = self.separate_transformer(left)
+
+        right = self.embeddings(right_token_ids)
+        right = self.separate_transformer(right)
+
+        # combine the sequences in some meaningful way. here, we just add them.
+        # combined = combine_masked_sequences(left, left_mask, right, right_mask)
+        combined = left + right
```
dirkgr, line 94: Maybe make it clear that we keep this simpler for clarity, but in a real world scenario you'd probably want to concatenate them. This example, as it is, will only run if the two sequences are the same length, which is a little weird.

allennlp/modules/transformer/transformer_module.py
```diff
@@ -35,16 +35,19 @@ def _get_mapped_submodules(
...
+        combined_mapping = {}
+        if "huggingface" in source:
+            combined_mapping.update(cls._huggingface_mapping)
+        if mapping is not None:
+            combined_mapping.update(mapping)
+        # if mapping is None:
+        #     if "huggingface" in source:
+        #         mapping = cls._huggingface_mapping
+        #     else:
+        #         mapping = {}
```
dirkgr, line 47: ?

allennlp/modules/transformer/transformer_module.py
```diff
@@ -73,29 +76,37 @@ def _load_from_pretrained_module(
...
         Optionally, a `mapping` is specified for any differences in parameter names
         between `pretrained_module` and the instance.
         """
-        if mapping is None:
-            mapping = self._construct_default_mapping(source)
-
-        inverse_mapping = {val: key for key, val in mapping.items()}
+        ignore_absent_parameters = ignore_absent_parameters or []
+        # if mapping is None:
+        #    mapping = self._construct_default_mapping(source)
```
dirkgr, line 88: ?

dirkgr (approved):
Except for the left-overs of commented out code, this looks good!

PR #4730, opened 2020-10-15
separating TransformerPooler as a new module

AkshitaB:


allennlp/modules/transformer/transformer_pooler.py
```diff
@@ -0,0 +1,11 @@
+from allennlp.common import FromParams
+from allennlp.modules.transformer.activation_layer import ActivationLayer
+
+
+class TransformerPooler(ActivationLayer, FromParams):
```
epwalsh, line 5: I think it would be good to document this and `ActivationLayer` to at least say what the input and output shapes look like and where this class is typically used.

tests/modules/transformer/transformer_pooler_test.py
```diff
@@ -0,0 +1,29 @@
...
+        self.pooler = TransformerPooler.from_params(params)
+
+    def test_can_construct_from_params(self):
+
+        assert self.pooler.dense.in_features == self.params_dict["hidden_size"]
+        assert self.pooler.dense.out_features == self.params_dict["intermediate_size"]
+
+    def test_forward_runs(self):
+
+        self.pooler.forward(torch.randn(2, 7, 5))
```
epwalsh, line 29: Might be worth having an assertion here for the expected shape:

```suggestion
        out = self.pooler.forward(torch.randn(2, 7, 5))
        assert out.size() == (2, 5)
```

epwalsh (approved):
LGTM!

allennlp/modules/transformer/transformer_pooler.py
AkshitaB: Yes, it's on my list to add documentation to all classes.

Issue #4649, opened 2020-09-18
Problem with PretrainedTransformerEmbedder and models such as T5


ianupright:
In  PretrainedTransformerEmbedder and PretrainedTransformerTokenizer, there is the line:

        self._num_added_end_tokens = len(tokenizer.single_sequence_end_tokens)

In the case of models such as T5, the _num_added_tokens results in zero.

but then there is code like:

        embeddings = embeddings[
            :, :, self._num_added_start_tokens : -(self._num_added_end_tokens), :
        ]  # truncate segment-level start/end tokens

which results in an empty tensor, because I think it should be a -1 to get the last element in the array, instead of -(self._num_added_end_tokens), which results in 0?

Nevertheless, a T5 model doesn't seem to work with the PretrainedTransformerEmbedder/PretrainedTransformerTokenizer

PR #4732, opened 2020-10-15
Fixes pretrained embeddings for transformers that don't have end tokens

dirkgr:
Fixes https://github.com/allenai/allennlp/issues/4649.


AkshitaB (approved):

dirkgr:
Unfortunately this turns out not to be so easy, because the Huggingface tokenizer for T5 is a little broken. I'll convert this to a draft and tackle it in 1.3.

dirkgr:
This is the issue on the huggingface side: https://github.com/huggingface/transformers/issues/7840

Issue #4355, opened 2020-06-12
hotflip api not working for some models


aimichal:
Example request to NAQANET resulting in HTTP 500:

```
curl -sv 'https://demo.allennlp.org/api/naqanet/attack/hotflip' -H 'Content-Type: application/json' --data-raw '{"inputs":{"model":"naqanet","passage":"Robotics is an interdisciplinary branch of engineering and science that includes mechanical engineering, electrical engineering, computer science, and others. Robotics deals with the design, construction, operation, and use of robots, as well as computer systems for their control, sensory feedback, and information processing. These technologies are used to develop machines that can substitute for humans. Robots can be used in any situation and for any purpose, but today many are used in dangerous environments (including bomb detection and de-activation), manufacturing processes, or where humans cannot survive. Robots can take on any form but some are made to resemble humans in appearance. This is said to help in the acceptance of a robot in certain replicative behaviors usually performed by people. Such robots attempt to replicate walking, lifting, speech, cognition, and basically anything a human can do.","question":"What do robots that resemble humans attempt to do?"},"input_field_to_attack":"question","grad_input_field":"grad_input_2"}'
```

The logs show this stack trace:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/app/allennlp_demo/common/http.py", line 210, in attack_handler
    with_cache_hit_response_headers(self.attack_with_cache, attacker_id, request.data)
  File "/app/allennlp_demo/common/http.py", line 40, in with_cache_hit_response_headers
    r = fn(*args)
  File "/app/allennlp_demo/common/http.py", line 103, in attack_with_cache
[...]

PR #4733, opened 2020-10-16
Made Interpret code handle mismatched cases better

### LINKED ISSUE #4360: Can't get correct output using .saliency_interpret_from_json() from Interpret

matt-gardner:
Fixes #4360.  The basic problem is that mismatched indexing led to gradients being computed at the top of a transformer, instead of at the end.  In order to solve that problem, we need to get gradients at the embedding layer, but aggregate them into token-level gradients.  The piece that knows about the mapping from wordpieces to tokens is an `offsets` dictionary in the `TextFieldTensors`.  We need to pull that out during a forward pass, so that when we get gradients during a backward pass we can aggregate things correctly.

This PR adds a forward hook in places where there was a backward hook, to check for mismatched indexing and return the offsets, if present.  It also updates the logic where gradients were processed to combine things in the right way.  I would have liked to isolate the changes a little bit more than this, but I didn't see a way to do it; at least the only things that needed to change are the base `Predictor` and the `SaliencyInterpreters`, and not any individual model, or anything.

I did _not_ touch any `Attackers`.  Those still might not work with mismatched indexing.  But, we have #4355 tracking attacker problems, which I suspect is related, and I'll be looking into that one next, maybe next week.

I think this needs a couple of small tests, so it's not _quite_ ready for review, but I'm running it through CI to see what's missing.


matt-gardner:
I don't know why the GPU check failed, but it looks like it didn't even run.  This should be ready for review now.  The coverage numbers look lower than they actually are, because some of these things get tested in the models repo.  Let me know if you'd like me to add more tests in this PR.

dirkgr (approved):
I'm not happy about the built-in pitfalls this has, but it's better than what was there before, and I don't want to start on yet another API redesign.

PR #4743, opened 2020-10-21
stricter typing for Optional[T] types, improve handling of Lazy params

### LINKED ISSUE #4742: Our type checking should be stricter
Currently we run `mypy` with the `--no-strict-optional` flag which is kind of crazy because it ends up ignoring a TON of typing errors.

epwalsh:
Closes #4742 and improves the [`Lazy` API](https://github.com/allenai/allennlp/blob/01644caf4de31710a6614fb96a9d4ea291353486/allennlp/common/lazy.py) (see below for details). Also fixes a bug where the serialization directory wouldn't get passed to custom `Checkpointers` and `TensorboardWriters`.

This PR seems to do a lot of different things, but it all has to do with the conventions we've adopted and assumptions we make about `Optional[T]` types, which, to be precise, are types which are `Union[NoneType, T]`. I will refer to these types loosely as "optionals".

We currently have two conventions around the use and type-checking of optionals:

1. First, in a class, method, or function signature, a parameter annotation such as `x: int = None` implies that `x` is actually optional, i.e. that `x: Optional[int] = None`. So these two annotations are equivalent. MyPy doesn't complain about you writing `x: int = None` because we don't use the [`--no-implicit-optional`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-no_implicit_optional) flag.

    Personally, I am completely fine with this as it cuts down on a lot of verbosity. Writing out `Optional`s everywhere gets very annoying.

2. The second convention we currently have is to pass the [`--no-strict-optional`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-strict_optional) flag to MyPy. This is CRAZY, as I brought up in https://github.com/allenai/allennlp/issues/4742, because it means MyPy will never check whether an optional type is `None` or non-`None` in a logical branch. That opens up the possibility for a ton of errors that can only be found at runtime.

Now, related to these conventions is our `Lazy` API. Currently, when we write a custom `from_params` classmethod constructor for a class, and do something like `foo: Lazy[Foo] = None`, **we are violating convention 1** because `foo` will actually never be `None` due to the logic in `from_params.py`. But MyPy doesn't care either way due to convention 2.

So, if we want to be more strict about our type checking by removing convention 2, we will also have to change the way we handle `Lazy`-ness.

This got me thinking more about the `Lazy` API, and I realized there are some other things wrong with it too. Maybe "wrong" is too strong of a word, but the issue is that because `foo: Lazy[Foo] = None` is actually never `None`, there is no way to make this `foo` `None` when you actually want `None`. For context, this issue came up around the use of our `TensorboardWriter`, which is created `Lazy`-ly, so there is no way to disable it (see [this thread](https://github.com/allenai/allennlp/pull/4674#issuecomment-713871415)).

So this PR makes the `Lazy` API a little more flexible in that you *can* in fact force a `Lazy` object to be `None` (by setting that field to `null` in your config) and at the same time it removes the violation of convention 1, i.e. `foo: Lazy[Foo] = None` is equivalent to `foo: Optional[Lazy[Foo]] = None`.

You can now do things like this:

```python
@classmethod
def my_constructor(
    cls,
    some_object: Lazy[MyObject],
    optional_object: Lazy[MyObject] = None,
    required_object_with_default: Lazy[MyObject] = Lazy(MyObjectDefault),
) -> MyClass:
    obj1 = some_object.construct()
    obj2 = None if optional_object is None else optional_object.construct()
    obj3 = required_object_with_default.construct()
```

For examples of this, see the changes to the `from_params` test: https://github.com/allenai/allennlp/pull/4743/files#diff-d2f18035610a7ffeac715f1e63c1856d6e0a64be671142972b4ea10e11144800

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Make sure weights are first loaded to the cpu when using PretrainedModelInitializer, preventing wasted GPU memory.
 - Load dataset readers in `load_archive`.
 - Updated `AllenNlpTestCase` docstring to remove reference to `unittest.TestCase`
+- Enforced stricter typing requirements around the use of `Optional[T]` types.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- Enforced stricter typing requirements around the use of `Optional[T]` types.
+- Changed the behavior of `Lazy` types in `from_params` methods. Previously, if you defined a `Lazy` parameter like
+  `foo: Lazy[Foo] = None` in a custom `from_params` classmethod, then `foo` would actually never be `None`.
+  This behavior is now different. If no params were given for `foo`, it will be `None`.
+  You can also now set default values for foo like `foo: Lazy[Foo] = Lazy(Foo)`.
+  Or, if you want you want a default value but also want to allow for `None` values, you can
+  write it like this: `foo: Optional[Lazy[Foo]] = Lazy(Foo)`.
+
 ### Fixed
 
 - Made it possible to instantiate `TrainerCallback` from config files.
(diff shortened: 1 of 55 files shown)


epwalsh:
@matt-gardner I would love to get your opinion on this if you're not too busy because you've probably done the most work on `from_params`.

matt-gardner (approved):
I only looked closely at the changes to Lazy.  They look good to me.  I didn't really like how that worked originally, but I didn't try hard enough to make it work better.  Looks like I should have tried a little harder, as this doesn't look like that much of a code change to get a substantial improvement in usability.

Just curious on the `--no-strict-optional` bit - did you actually find any bugs?  Scanning through the rest of the changes, it looks like most of the lines are `# type: ignore`s.

epwalsh:
> I only looked closely at the changes to Lazy. They look good to me. I didn't really like how that worked originally, but I didn't try hard enough to make it work better. Looks like I should have tried a little harder, as this doesn't look like that much of a code change to get a substantial improvement in usability.

That's where the only substantial changes were. Everything else was pretty much just fixing type hints.

> Just curious on the `--no-strict-optional` bit - did you actually find any bugs? Scanning through the rest of the changes, it looks like most of the lines are `# type: ignore`s.

I don't think I fixed any bugs, but I did fix a lot of erroneous type hints. For examples, there were places where a return type of a method had an `Optional[T]`, but should just be `T`, or vice versa. MyPy never picked up on these due to the `--no-strict-optional` flag.

PR #4570, opened 2020-08-18
Bug fix in distributed metrics

AkshitaB:
Fixes https://github.com/allenai/allennlp/issues/4564
Remaining metrics

- [ ] Covariance

PR #4744, opened 2020-10-22
Fix device mismatch bug for categorical accuracy metric in distributed training

wlhgtc:
I got same error( `RuntimeError: Tensors must be CUDA and dense` )  in #4623 when I used distributed training with **allennlp 1.1.0** and **pytorch 1.6**.
I read your PR in #4570 , seem you only fix it in `Average` Class. When I use `CategoricalAccuracy `, the same error occur. 
So I try to fix it by move these tensors to GPU (same in `Average`), it works well.
Maybe you could help me check my solution, then I could update all related Metrics Classes ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,7 +63,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed `common.util.is_master` function.
 
 ### Fixed
-
+- Fix distributed metrics error.
 - Fixed a bug where the reported `batch_loss` metric was incorrect when training with gradient accumulation.
 - Class decorators now displayed in API docs.
 - Fixed up the documentation for the `allennlp.nn.beam_search` module.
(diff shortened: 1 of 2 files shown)


wlhgtc:
@dirkgr  @matt-gardner

CHANGELOG.md
```diff
@@ -63,7 +63,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed `common.util.is_master` function.
 
 ### Fixed
-
+- Fix distributed metrics error.
```
epwalsh, line 66:
```suggestion

- Fix CUDA/CPU device mismatch bug during distributed training for categorical accuracy metric.
```

epwalsh (changes requested):
Thanks for catching this @wlhgtc!

CHANGELOG.md
wlhgtc: After re-check other metrics, seem the operation is missed only in `CategoricalAccuracy `.
But this operation is done by different method in different classes.
Do you think we need a universal method for all tensors ?

epwalsh (approved):

PR #4756, opened 2020-10-28
Generalizing self attention

AkshitaB:
* `SelfAttention` - Tests for checking output of forward
* `SelfAttention` - Tests for loading different HF models
* Minor refactoring of `TransformerModule`
* `SelfAttention` - Docstrings


allennlp/modules/transformer/self_attention.py
```diff
@@ -13,14 +13,27 @@ class SelfAttention(TransformerModule, FromParams):
...
+
+    hidden_size: `int`
+    num_attention_heads: `int`
+    dropout: `float` (default = `0.0`)
+    scoring_func: `str` (default = `scaled_dot_product`)
+        The name of the attention-calculating function to be used.
+        Eg. `additive`, `linear`, etc. For a complete list, please check :mod:`allennlp.modules.attention`.
     """
 
+    _relevant_module = ["encoder.layers.0.attention.self", "encoder.layers.0.attention"]
```
dirkgr, line 27: It occurs to me that if we pass in a BART model, which has a separate encoder and decoder, then this mapping wouldn't work. Am I wrong about that?

allennlp/modules/transformer/self_attention.py
```diff
@@ -45,6 +58,10 @@ def __init__(
         else:
             self.attn = Attention.by_name(self.scoring_func)()
 
+        # out linear layer for distilbert.
+        if output_linear:
+            self.output = torch.nn.Linear(hidden_size, self.all_head_size)
```
dirkgr, line 63: What is this? Does it get initialized as well?

tests/modules/transformer/self_attention_test.py
```diff
@@ -1,20 +1,84 @@
...
+from transformers.modeling_roberta import RobertaSelfAttention
+from transformers.configuration_electra import ElectraConfig
+from transformers.modeling_electra import ElectraSelfAttention
+from transformers.configuration_distilbert import DistilBertConfig
+from transformers.modeling_distilbert import MultiHeadSelfAttention
+
+# from transformers.configuration_mobilebert import MobileBertConfig
+# from transformers.modeling_mobilebert import MobileBertSelfAttention
+# from transformers.configuration_t5 import T5Config
+# from transformers.modeling_t5 import T5LayerSelfAttention
```
dirkgr, line 24: What are these for?

tests/modules/transformer/self_attention_test.py
```diff
@@ -1,20 +1,84 @@
...
+    # hf_module = MobileBertSelfAttention(MobileBertConfig(**params))
+    # modules["mobile_bert"] = hf_module
+
+    # torch.manual_seed(1234)
+    # t5params = copy.deepcopy(params_dict)
+    # t5params["num_heads"] = t5params.pop("num_attention_heads")
+    # t5params["d_model"] = t5params.pop("hidden_size")
+    # t5params["dropout_rate"] = t5params.pop("dropout")
+    # hf_module = T5LayerSelfAttention(T5Config(**t5params))
+    # modules["t5"] = hf_module
```
dirkgr, line 72: Leftovers?

tests/modules/transformer/self_attention_test.py
```diff
@@ -1,20 +1,84 @@
...
+    "dropout": 0.0,
+}
+
+
+def get_modules(params_dict):
+    modules = {}
+    params = copy.deepcopy(params_dict)
+    params["attention_probs_dropout_prob"] = params.pop("dropout")
+
+    # bert, roberta, electra, layoutlm self attentions have the same code.
```
dirkgr, line 38: `layoutlm`? That's not mentioned again later.

tests/modules/transformer/self_attention_test.py
```diff
@@ -37,5 +101,84 @@ def test_can_construct_from_params(self):
...
+            )[0]
+        else:
+            hf_output = pretrained_module.forward(hidden_states, attention_mask=attention_mask)[0]
+
+        # FIX: look into the reason for mismatch.
+        # Update: The discrepancy comes from torch.nn.Dropout layer, despite setting random seeds.
+        # Have also tried setting random seeds right before the actual call to dropout in both modules.
+        # assert torch.allclose(output, hf_output)
+        print(output)
+        print(hf_output)
```
dirkgr, line 184: This looks like debugging code.

dirkgr (changes requested):

allennlp/modules/transformer/self_attention.py
AkshitaB: Yes. If the `_relevant_module` isn't found, the code will return the module that was passed in, which will eventually fail. The goal is to ensure that we don't require too many one-off specifications for different variants, but some will be inevitable.

allennlp/modules/transformer/self_attention.py
AkshitaB: Distilbert has this extra linear layer at the end. Trying to ensure that the module works for different variants.

tests/modules/transformer/self_attention_test.py
AkshitaB: I wanted to test with different versions, but T5 requires some more changes. I am planning to add tests for them once I finish the first pass for all modules (i.e. all modules should at least generate the same output as vanilla bert).

tests/modules/transformer/self_attention_test.py
AkshitaB: See previous comment on T5.

tests/modules/transformer/self_attention_test.py
AkshitaB: Yes, I didn't add it because it's the exact same code as bert. Technically, tests for roberta and electra aren't needed either (at least for the self-attention module). I just added them to demonstrate that the toolkit modules are expected to work for different models.

tests/modules/transformer/self_attention_test.py
AkshitaB: Right, so the commented line before the print statements is supposed to check that the output is the same. However, for some reason, a discrepancy occurs in the dropout layer (this is after I set the random seeds). I want to see if the problem also occurs in other modules. I will remove the print statements, however.

AkshitaB: Ah, I remember now. I wanted to keep the rest of the code, so I could later debug the issue I mentioned, but lint would complain about not using the defined variables `output` and `hf_output`. The print statements were just a way to get around that.

While I want to create smaller PRs, the modules are somewhat tied to each other, and hence the development is incremental. Thus, the code in the initial PRs may not be completely polished (eg. the debug prints/comments).

dirkgr: I guess that's fine this time, since you're immediately following up, but in general PRs should be 1. self contained, and 2. small, in that order. If you have to compromise one of these goals, compromise goal 2 first.

dirkgr (approved):
... assuming you follow up with another PR that makes the commented code useful (or disappear).

Issue #4757, opened 2020-10-28
PretrainedTransformerTokenizer fails when disabling "fast" tokenizer


mklimasz:
## Checklist

- [X] I have verified that the issue exists against the `master` branch of AllenNLP.
- [X] I have read the relevant section in the [contribution guide](https://github.com/allenai/allennlp/blob/master/CONTRIBUTING.md#bug-fixes-and-new-features) on reporting bugs.
- [X] I have checked the [issues list](https://github.com/allenai/allennlp/issues) for similar or identical bug reports.
- [X] I have checked the [pull requests list](https://github.com/allenai/allennlp/pulls) for existing proposed fixes.
- [X] I have checked the [CHANGELOG](https://github.com/allenai/allennlp/blob/master/CHANGELOG.md) and the [commit log](https://github.com/allenai/allennlp/commits/master) to find out if the bug was already fixed in the master branch.
- [X] I have included in the "Description" section below a traceback from any exceptions related to this bug.
- [X] I have included in the "Related issues or possible duplicates" section beloew all related issues and possible duplicate issues (If there are none, check this box anyway).
- [X] I have included in the "Environment" section below the name of the operating system and Python version that I was using when I discovered this bug.
- [X] I have included in the "Environment" section below the output of `pip freeze`.
- [X] I have included in the "Steps to reproduce" section below a minimally reproducible example.

## Description
**Tested on 1.2.0rc1 and master**
Intra work tokenizer doesn't work when we deliberately set use fast tokenizer to false (not sure if it's new transformers change).
I think that setting return_token_type_ids to None instead of False is solution here. 

<details>
<summary><b>Python traceback:</b></summary>
<p>

```
Traceback (most recent call last):
  File "bug_example.py", line 4, in <module>
    tokenizer_kwargs={"use_fast": False}).intra_word_tokenize(["My", "text", "will"])
  File "X/venv/lib/python3.6/site-packages/allennlp-1.2.0rc1-py3.6.egg/allennlp/data/tokenizers/pretrained_transformer_tokenizer.py", line 387, in intra_word_tokenize
    tokens, offsets = self._intra_word_tokenize(string_tokens)
  File "X/venv/lib/python3.6/site-packages/allennlp-1.2.0rc1-py3.6.egg/allennlp/data/tokenizers/pretrained_transformer_tokenizer.py", line 354, in _intra_word_tokenize
    return_token_type_ids=False,
[...]

PR #4758, opened 2020-10-28
Make returning token type id default in transformers intra word tokenization.

mklimasz:
Addressing #4757 bug - check description there.

I can duplicate intra tokenizer test with use_fast False flag, but not sure if you want to keep such tests (there can be multiple of such tests with different flags).


epwalsh (changes requested):
Thanks @mklimasz! Could you just add a note about this to the "Fixed" section of the CHANGELOG?

epwalsh (approved):
LGTM!

Issue #4554, opened 2020-08-12
Training ELMo Transformer, and using it as an embedder


lianna1016:
Hello! I am trying to train transformer ELMo using my own dataset. I am trying to use [this tutorial](https://github.com/allenai/allennlp/blob/v0.9.0/tutorials/how_to/training_transformer_elmo.md) but it is proving a little difficult for me because of the older version. In particular I had a question regarding [the example config file](https://github.com/allenai/allennlp/blob/v0.9.0/training_config/bidirectional_language_model.jsonnet). It has the line ```  "allow_unmatched_keys": true, ``` which is a feature that is now deprecated, and I was wondering how to correct for this as when I try to train the model I'm now getting the error ```allennlp.common.checks.ConfigurationError: Mismatched token keys: dict_keys(['token_characters']) and dict_keys(['token_characters', 'tokens'])``` Thank you!

dirkgr:
The model still exists, but it was moved to the [allennlp-models](https://github.com/allenai/allennlp-models) repository. There is an updated config here: https://github.com/allenai/allennlp-models/blob/master/training_config/lm/bidirectional_language_model.jsonnet

Can you try that?

I don't know for sure that this config will train properly. It takes a long time to train this model, so we can't run tests for it all the time. But if you find that it no longer works in that state, we'll be happy to help you fix it.

lianna1016:
Thank you! 

I see that these lines were able to fix this problem
``` go
"tokens": {
    "type": "empty"
},
```

And I was able to do a quick training using the config file you pointed me too. However, now when trying to use these elmo embeddings in a different model, I run into the following error.

```
allennlp.common.checks.ConfigurationError: LM from tmp/elmo_test1 trained with multiple embedders!
```

and trying to solve this problem, I am looking at [language_model_token_embedder.py](https://github.com/allenai/allennlp/blob/v0.9.0-3971/allennlp/modules/token_embedders/language_model_token_embedder.py) that states

```           
 # We don't currently support embedding with language models trained with multiple
 # embedded indices.
```

Does this have anything to do with the tokens lines of code I referenced above, and if so, do you have any recommendations? Thanks!

dirkgr:
@nelson-liu, do you know how this is supposed to be used? It says quite clearly in `LanguageModelTokenEmbedder` that you should use it with only one indexer, but isn't ELMo always trained with two indexers, one for tokens, and one for characters?

nelson-liu:
[...]

PR #4761, opened 2020-10-29
allow TextFieldEmbedder to have EmptyEmbedder that may not be in input

tpanza:
Additional patch for #4554

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Made it possible to instantiate `TrainerCallback` from config files.
 - Fixed the remaining broken internal links in the API docs.
 - Fixed a bug where `BeamSearch` would fail if `max_steps` was equal to 1.
+- Fixed `BasicTextFieldEmbedder` to not raise ConfigurationError if it has embedders that are empty and not in input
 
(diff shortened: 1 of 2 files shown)


tpanza:
@dirkgr can you please take a look

epwalsh:
Hey @tpanza, Dirk is currently out but will be back next week, so this may have to wait until then.

dirkgr (approved):
I could have sworn I made this exact change already.

Anyways, I'm not wild about giving `EmptyEmbedder` special treatment all over, but it's the lesser of the evils, compared to re-jiggering the API to allow for the empty case more cleanly.

dirkgr:
@tpanza, is this the last of the issues? Can you now re-train ELMo and use it as an embedder?

tpanza:
> 
> 
> @tpanza, is this the last of the issues? Can you now re-train ELMo and use it as an embedder?

Hi @dirkgr. Yes! What would be the best place to document steps with a dummy/toy example? In other words, what would be the best place to make an updated version of [this tutorial](https://github.com/allenai/allennlp/blob/v0.9.0/tutorials/how_to/training_transformer_elmo.md)?

dirkgr:
The new place for that kind of stuff is at https://github.com/allenai/allennlp-guide. That ultimately gets published here: https://guide.allennlp.org. You still write Markdown, but the layout engine makes that fancy website out of it. I'd love it if you could add a chapter on this!

PR #4764, opened 2020-10-31
Pass batch metrics to `BatchCallback`

eladsegal:


dirkgr (changes requested):
This is great! I can't remember if we discussed this somewhere or if you just read my mind.

Can you add a changelog entry?

eladsegal:
> This is great! I can't remember if we discussed this somewhere or if you just read my mind.
> 
> Can you add a changelog entry?

Just a coincidence :)
Changelog entry was added

dirkgr (approved):

PR #4768, opened 2020-11-03
Add sampling strategies to beam search

epwalsh:
TODO (@jvstokes)
- [x] implement multinomial sampler
- [x] implement top-k sampler
- [x] implement top-p sampler
- [x] test multinomial sampler
- [x] test top-k sampler
- [x] test top-p sampler
- [x] test gumbel sampler, especially with edge cases like when the end token is predicted
- [x] update CHANGELOG
- [x] finalize documentation


allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,365 @@
...
+    Default is `True`.
+    """
+
+    def __init__(
+        self,
+        temperature: float = 1.0,
+        with_replacement: bool = False,
+    ) -> None:
+        self.temperature = temperature
+        self.filter_val = min_value_of_dtype(torch.float)
```
epwalsh, line 118: It doesn't look like this is used.

```suggestion
        self.filter_val = min_value_of_dtype(torch.float)
```

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,365 @@
...
+class MultinomialSampler(Sampler):
+    """
+    Represents a sampler to choose values from a multinomial distribution.
+    Registered as a `Sampler` with name "multinomial".
+    `log_probs` is a tensor of normalized log-probabilities to be selected from.
+    `temperature` modules the probabilities of the selected tokens. A `temperature` below
+    1.0 produces a sharper probability distribution and a `temperature` above 1.0 produces
+    a flatter probability distribution.
+    `with_replacement` is `False` if sampled instances must be repeated in the sampled set.
+    Default is `True`.
```
epwalsh, line 109:
```suggestion
    A `Sampler` which samples nodes from the given multinomial distribution. Beams are sampled
    in the default, non-deterministic way.
    
    # Parameters
    
    temperature : `float`, optional (default = `1.0`)
        A `temperature` below 1.0 produces a sharper probability distribution and a `temperature` above 1.0
        produces a flatter probability distribution.
    with_replacement : `bool`, optional (default = `False`)
        Whether to sample with replacement.
```

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,365 @@
...
+        self.k = k
+        self.temperature = temperature or 1.0
+        self.filter_val = min_value_of_dtype(torch.float)
+        self.with_replacement = with_replacement
+
+    @overrides
+    def sample_nodes(
+        self, log_probs: torch.Tensor, per_node_beam_size: int, state: StateType
+    ) -> Tuple[torch.Tensor, torch.Tensor, StateType]:
+        assert self.k <= len(log_probs[-1])
```
epwalsh, line 167:
```suggestion
        assert per_node_beam_size <= self.k <= log_probs.size()[1]
```

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,365 @@
...
+
+    def __init__(
+        self,
+        k: int = 1,
+        temperature: float = 1.0,
+        filter_val: float = -float("inf"),
+        with_replacement: bool = False,
+    ):
+        if not isinstance(k, int) or k < 1:
+            raise ValueError("k must be a positive integer")
```
epwalsh, line 157: Just a minor nit. We generally don't check types since we have type hints. I think checking the value is sufficient.

```suggestion
        if k < 1:
            raise ValueError("k must be a positive integer")
```

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,365 @@
...
+    sharper probability distribution and a `temperature` above 1.0 produces a flatter probability
+    distribution.
+    Registered as a `Sampler` with name "top-k".
+    """
+
+    def __init__(
+        self,
+        k: int = 1,
+        temperature: float = 1.0,
+        filter_val: float = -float("inf"),
```
epwalsh, line 153: It's better to create the `filter_val` on the fly with the `sample_nodes()` method based on the value of `log_probs`, since we could be training on float32 or float16, depending on whether or not AMP is enabled.

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,365 @@
...
+        filtered_probabilities = torch.nn.functional.softmax(_log_probs, dim=-1)
+
+        # Shape: (batch_size*beam_size, per_node_beam_size)
+        selected_indices = torch.multinomial(
+            filtered_probabilities, per_node_beam_size, replacement=self.with_replacement
+        )
+
+        # Return (selected log probabilities, selected classes)
+        # shape: (batch_size*beam_size, num_samples), (batch_size*beam_size, num_samples)
+        return torch.gather(_log_probs, 1, selected_indices), selected_indices, state
```
epwalsh, line 189: I think we could avoid an extra call to softmax by doing something like this instead:

```suggestion
        # shape (both): (batch_size, k)
        top_k_log_probs, top_k_indices = log_probs.topk(self.k, dim=-1)

        # Apply temperature if necessary.
        # shape: (batch_size, k)
        if self.temperature != 1.0:
            top_k_log_probs = top_k_log_probs / self.temperature

        # Re-normalize the subset.
        # shape: (batch_size, k)
        normalized_top_k_probs = torch.nn.functional.softmax(top_k_log_probs, dim=-1)

        # Sample from the re-normalized subset.
        # NOTE: These indices are not indices into `log_probs`, they are indices into `top_k_log_probs`.
        # shape: (batch_size, per_node_beam_size)
        sampled_indices = torch.multinomial(
            normalized_top_k_probs, per_node_beam_size, replacement=self.replacement
        )

        # Convert `sampled_indices` back to indices in the original `log_probs` tensor.
        # shape: (batch_size, per_node_beam_size)
        indices = top_k_indices.gather(-1, sampled_indices)

        return log_probs.gather(1, indices), indices, state
```

allennlp/nn/beam_search.py
epwalsh: Actually, with the suggested change below, I don't think we need this.

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,365 @@
...
+@Sampler.register("top-p")
+class TopPSampler(Sampler):
+    """
+    Represents a `Sampler` which redistributes the probability mass function among
+    the top choices with a cumulative probability of at least `p` then selects from that subset
+    `p` if minimum cumulative probability of highest-probability options that the returned
+    choice will be selected from `temperature` modules the probabilitis of the selected tokens.
+    A `temperature` below 1.0 produces a sharper probability distribution and a `temperature`
+    above 1.0 produces a flatter probability distribution.
+    Registered as a `Sampler` with name "top-p".
```
epwalsh, line 201:
```suggestion
    A `Sampler` which redistributes the probability mass function for nodes among
    the top choices with a cumulative probability of at least `p`, then samples from that subset
    after re-normalizing the probabilities.
    
    Beams are sampled in the default, deterministic way.
    
    # Parameters
    
    p : `float`, optional (default = `0.9`)
        The cumulative probability cutoff threshold. A higher value of `p` will result in more possible
        examples to sample from.
    temperature : `float`, optional (default = `1.0`)
        A `temperature` below 1.0 produces a sharper probability distribution and a `temperature`
        above 1.0 produces a flatter probability distribution.
```

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,365 @@
...
+
+    def __init__(
+        self,
+        p: float = 0.9,
+        temperature: float = 1.0,
+        filter_val: float = -float("inf"),
+        with_replacement: bool = False,
+    ):
+        if not isinstance(p, float) or p < 0.0 or p > 1.0:
+            raise ValueError("p must be a non-negative float no greater than 1.0")
```
epwalsh, line 212: Same as above, I think it's okay to omit the type check:

```suggestion
        if p < 0.0 or p > 1.0:
            raise ValueError("p must be a non-negative float no greater than 1.0")
```

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,365 @@
...
+        self.p = p
+        self.temperature = temperature or 1.0
+        self.filter_val = min_value_of_dtype(torch.float)
```
epwalsh, line 215: Same suggestion as above, we should create this value on the fly.

```suggestion
        self.filter_val = min_value_of_dtype(torch.float)
```

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,365 @@
...
+        self, log_probs: torch.Tensor, per_node_beam_size: int, state: StateType
+    ) -> Tuple[torch.Tensor, torch.Tensor, StateType]:
+
+        # First apply temperature coefficient:
+        if self.temperature != 1.0:
+            _log_probs = torch.nn.functional.log_softmax(log_probs / self.temperature, dim=-1)
+        else:
+            _log_probs = log_probs
+
+        # Sort the probabilities to hi ghest-first, then find the cumulative sum accross those
```
epwalsh, line 229:
```suggestion
        # Sort the probabilities to highest-first, then find the cumulative sum across those.
        # shape (both): (batch_size, num_classes)
```

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,365 @@
...
+        log_probs_descending, sorting_indices = torch.sort(_log_probs, descending=True)
+        probabilities_descending = torch.nn.functional.softmax(log_probs_descending, dim=-1)
```
epwalsh, line 231: No need for softmax, could just use `exp`:

```suggestion
        # shape: (batch_size, num_classes)
        probabilities_descending = log_probs_descending.exp()
        
```

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,365 @@
...
+            filter_idx = sort_indices[filter_vals]
+            row[filter_idx] = self.filter_val
+
+        filtered_probabilities = torch.nn.functional.softmax(_log_probs, dim=-1)
+        print(filtered_probabilities)
+
+        # Here we sample from the filtered distribution
+        selected_indices = torch.multinomial(
+            filtered_probabilities, per_node_beam_size, replacement=self.with_replacement
+        )
```
epwalsh, line 251: I think we could avoid a loop by doing something like this. This also avoids modifying `_log_probs` in place (which would modify `log_probs` if temperature was 1.0).

```suggestion
        # Create a mask for filtering out probabilities that don't make the top `p`.
        # shape: (batch_size, num_classes)
        exclusion_mask = probabilities_summed > self.p

        # Now re-normalized the included log probs.
        # shape: (batch_size, num_classes)
        log_probs_descending[exclusion_mask] = min_value_of_dtype(log_probs.dtype)
        # shape: (batch_size, num_classes)
        filtered_probabilities = torch.nn.functional.softmax(log_probs_descending, dim=-1)

        # Sample from the re-normalized subset.
        # NOTE: These indices are not indices into `log_probs`, they are indices into `log_probs_descending`.
        # shape: (batch_size, per_node_beam_size)
        sampled_indices = torch.multinomial(filtered_probabilities, per_node_beam_size, replacement=False)

        # Convert `sampled_indices` back to indices in the original `log_probs` tensor.
        # shape: (batch_size, per_node_beam_size)
        selected_indices = sorting_indices.gather(-1, sampled_indices)
```

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,365 @@
...
+
+        # Return (selected log probabilities, selected classes)
+        # shape: (len(log_probs),1) , (len(log_probs), 1)
+        return torch.gather(_log_probs, 1, selected_indices), selected_indices, state
```
epwalsh, line 255: We want to use the original log probs here, not the temperature-adjusted log probs.

```suggestion
        return torch.gather(log_probs, 1, selected_indices), selected_indices, state
```

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,365 @@
...
+@Sampler.register("gumbel")
+class GumbelSampler(Sampler):
+    """
+    Represents a `Sampler` which uses the Gumbel-TopK trick to sample without replacement. See
+    [*Stochastic Beams and Where to Find Them: The Gumbel-Top-k Trick for Sampling
+    Sequences Without Replacement*, W Kool, H Van Hoof and M Welling, 2010]
+    (https://arxiv.org/abs/1903.06059).
+    A `temperature` below 1.0 produces a sharper probability distribution and a `temperature`
+    above 1.0 produces a flatter probability distribution.
+    Registered as a `Sampler` with name "gumbel".
```
epwalsh, line 267:
```suggestion
    A `Sampler` which uses the Gumbel-Top-K trick to sample without replacement. See
    [*Stochastic Beams and Where to Find Them: The Gumbel-Top-k Trick for Sampling
    Sequences Without Replacement*, W Kool, H Van Hoof and M Welling, 2010]
    (https://arxiv.org/abs/1903.06059).
    
    # Parameters
    
    temperature : `float`, optional (default = `1.0`)
        A `temperature` below 1.0 produces a sharper probability distribution and a `temperature`
        above 1.0 produces a flatter probability distribution.
```

epwalsh (commented):
Almost there 😬! My suggestions are mostly cosmetic. I also think I found a way to avoid the loop within top-p.

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,389 @@
...
+        probabilities_descending = log_probs_descending.exp()
+        probabilities_summed = torch.cumsum(probabilities_descending, dim=-1)
+
+        # Create a mask for filtering out probabilities that don't make the top `p`.
+        # shape: (batch_size, num_classes)
+        exclusion_mask = probabilities_summed >= self.p
+
+        # We want to include the firt index where probabilities_summes >= p, so we shift over one.
+        exclusion_mask[..., 1:] = exclusion_mask[..., :-1].clone()
+        exclusion_mask[..., 0] = False
```
epwalsh, line 254: How about this:

```suggestion
        # Create a mask for filtering out probabilities that don't make the top `p`.
        # shape: (batch_size, num_classes)
        exclusion_mask = probabilities_summed > self.p
        # Make sure there's at least `per_node_beam_size` options.
        exclusion_mask[..., :per_node_beam_size] = False
```

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,389 @@
...
+        if p < 0.0 or p > 1.0:
+            raise ValueError("p must be a non-negative float no greater than 1.0")
+        self.p = p
+        self.temperature = temperature or 1.0
+        self.with_replacement = with_replacement
+
+    @overrides
+    def sample_nodes(
+        self, log_probs: torch.Tensor, per_node_beam_size: int, state: StateType
+    ) -> Tuple[torch.Tensor, torch.Tensor, StateType]:
```
epwalsh, line 231:
```suggestion
    ) -> Tuple[torch.Tensor, torch.Tensor, StateType]:
            if not per_node_beam_size <= log_probs.size()[1]:
                raise ValueError("per_node_beam_size cannot be greater than vocabulary size")
```

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,389 @@
...
+        self.k = k
+        self.temperature = temperature or 1.0
+        self.with_replacement = with_replacement
+
+    @overrides
+    def sample_nodes(
+        self, log_probs: torch.Tensor, per_node_beam_size: int, state: StateType
+    ) -> Tuple[torch.Tensor, torch.Tensor, StateType]:
+        if not per_node_beam_size <= self.k <= log_probs.size()[1]:
+            raise ValueError("k must be a postive integer no greater than vocabulary size")
```
epwalsh, line 172:
```suggestion
        if not per_node_beam_size <= self.k <= log_probs.size()[1]:
            raise ValueError("k must be a postive integer no less than per_node_beam_size and no greater than vocabulary size")
```

allennlp/nn/beam_search.py
```diff
@@ -25,6 +27,391 @@
...
+    """
+
+    def __init__(
+        self,
+        p: float = 0.9,
+        temperature: float = 1.0,
+        with_replacement: bool = False,
+    ):
+        if p < 0.0 or p > 1.0:
+            raise ValueError("p must be a non-negative float no greater than 1.0")
```
epwalsh, line 223:
```suggestion
            raise ValueError("p must be a positive float no greater than 1.0")
```

allennlp/nn/beam_search.py
jvstokes: I think we actually need both. If we don't shift the mask then then the cumulative sum is less than p, i.e.
probs = [0.5, 0.31, 0.19] , p = 0.8
cumsum = [0.5, 0.81, 1.0]
mask: [False, True, True]
softmax: [1.0, 0.0, 0.0]

epwalsh: Good point. How about this then:

```suggestion
        # Create a mask for filtering out probabilities that don't make the top `p`.
        # shape: (batch_size, num_classes)
        exclusion_mask = probabilities_summed >= self.p

        # We want to include the first index where probabilities_summed >= p, so we shift over one.
        exclusion_mask[..., 1:] = exclusion_mask[..., :-1].clone()

        # We also want to make sure we have at least `per_node_beam_size` options.
        exclusion_mask[..., :per_node_beam_size] = False
```

jvstokes: Do we want to allow the client to sample  with replacement from < per_node_beam_size options?

jvstokes: i.e. maybe they sample the 10 examples from the top 8 cumulative probabilities

epwalsh: Possibly. So maybe only fill up to `per_node_beam_size` as `False` if `with_replacement` is `False`, otherwise only fill the first as `False`.

jvstokes: If you don't think we should build that functionality in though I can clean it up!

epwalsh: I think it's okay to have that.

jvstokes: oops sorry I forgot to refresh and missed your comment! And yes that's the change I ended up making, so it should be all good

jvstokes (approved):

PR #4770, opened 2020-11-05
Change tensorboard usage optional

shimengfeng:
Provide users a choice for tensorboard usage. As a temporary solution by adding Optional to the code. Please refer to #4674 for more reference.

* note: this is my first contribution to AllenNlp, it might be a bit messy.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -247,7 +247,7 @@ jobs:
     - name: Install core package
       run: |
         pip install $(ls dist/*.whl)
-        # In PyTorch 1.7, dataclasses is an unconditional dependency, when it should
+        # TODO(epwalsh): In PyTorch 1.7, dataclasses is an unconditional dependency, when it should
         # only be a conditional dependency for Python < 3.7.
         # This has been fixed on PyTorch master branch, so we should be able to
         # remove this check with the next PyTorch release.
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -147,8 +147,7 @@ jobs:
     - name: Clean up
       if: always()
       run: |
-        pip uninstall --yes allennlp
-        pip uninstall --yes allennlp_models
+        pip uninstall --yes allennlp allennlp-models
 
   # Builds package distribution files for PyPI.
   build_package:
@@ -248,6 +247,11 @@ jobs:
     - name: Install core package
       run: |
         pip install $(ls dist/*.whl)
+        # In PyTorch 1.7, dataclasses is an unconditional dependency, when it should
+        # only be a conditional dependency for Python < 3.7.
+        # This has been fixed on PyTorch master branch, so we should be able to
+        # remove this check with the next PyTorch release.
+        pip uninstall -y dataclasses
 
     - name: Pip freeze
       run: |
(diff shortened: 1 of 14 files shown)


shimengfeng:
Hi @epwalsh 
By changing it to optional, the tensorboard construct failed, so I tried to add 

`if tensorboard_writer is not None:
       tensorboard_writer_ = tensorboard_writer.construct(serialization_dir=serialization_dir)
else:
       tensorboard_writer_  = None`

But then I will have a type issue since the default is set up as TensorboardWriter. Please correct me if I am wrong.

epwalsh:
@shimengfeng you could do it like this:

```python
tensorboard_writer_: Optional[TensorboardWriter] = None
if tensorboard_writer is not None:
    tensorboard_writer_ = tensorboard_writer.construct(...)
```

epwalsh:
@shimengfeng you will also need to make some minor changes elsewhere in the `GradientDescentTrainer` class anywhere that `self._tensorboard_writer` is used.

shimengfeng:
That make sense, let me take another poke later this week.

mahnerak:
@shimengfeng Can you also make sure that the trainer logs metrics to the console.
For example, this piece of log comes from `allennlp.training.tensorboard_writer`.
```
2020-12-15 22:53:35,349 - INFO - allennlp.training.tensorboard_writer -                        Training |  Validation
2020-12-15 22:53:35,349 - INFO - allennlp.training.tensorboard_writer - accuracy           |     0.834  |     0.843
2020-12-15 22:53:35,350 - INFO - allennlp.training.tensorboard_writer - gpu_0_memory_MB    |  4449.827  |       N/A
2020-12-15 22:53:35,350 - INFO - allennlp.training.tensorboard_writer - loss               |     0.440  |     0.437
2020-12-15 22:53:35,351 - INFO - allennlp.training.tensorboard_writer - worker_0_memory_MB |  8806.191  |       N/A
```
If the `tensorboard_writer` is disabled we still want to have these number after each epoch/validation loop, right?

shimengfeng:
Sorry for the delays, let me take a look again tonight. We had some reorg happened at work, was taking time to respond to those changes. Apologizes for the delays.

shimengfeng:
@mahnerak I take a look at the code yesterday and I want to make sure I understand what will be the right thing to do before I change the code. When we set the tensorboard writer to optional, we still want to output the loss, accuracy, and GPU performance, but just not train and validation information right?

If I understand properly, for the train.py where logging the training information: 
`if self._master:
                self._tensorboard.log_metrics(
                    train_metrics, val_metrics=val_metrics, log_to_console=True, epoch=epoch + 1
                )  # +1 because tensorboard doesn't like 0
`

I just need to change this to only log_metrics if tensorboard is given, right?

mahnerak:
@shimengfeng First of all, you need to check `self._tensorboard` not being None before using it, as previously mentioned by @epwalsh.
Second, I believe making the `tensorboard_writer` optional makes sense only if it disables only actual tensorboard features. However, right now the metrics are logged to the console from the `tensorboard_writer` class. We need to move them into the `GradientDescentTrainer`, just before the `self._tensorboard.log_metrics`.

epwalsh:
Closing in favor of https://github.com/allenai/allennlp/pull/4913.

PR #4774, opened 2020-11-07
Set the device for an auto-created mask

dirkgr:
When we create completely new tensors outside of a dataset reader, we almost always have to be careful where we put them.


dirkgr:
I found this issue in https://github.com/allenai/allennlp-models/pull/163, though that PR has nothing to do with this error ... 😔

epwalsh:
https://github.com/allenai/allennlp/pull/4774/files#diff-dc2fdf0ed8bb2faa7593db783e3ae12142da63ccee04731a67ce535af1dd1d87L134 this might also need the device as @MichalMalyska points out here: https://github.com/allenai/allennlp/pull/4746#issuecomment-723335446

jvstokes (approved):

PR #4776, opened 2020-11-09
Generalizing transformer layers

AkshitaB:
* Added HF tests.
* Added docstrings with expected input shapes.


allennlp/modules/transformer/transformer_block.py
```diff
@@ -10,6 +10,24 @@
 
 
 class TransformerBlock(TransformerModule, FromParams):
+    """
+    This module is the basic transformer block, which acts as an encoder.
```
dirkgr, line 14: It's possible I have this terminology wrong, but why do you call this an "encoder"? What does it encode?

I would understand the terminology if this took word piece ids in its input, because those get encoded, but this takes vectors and returns vectors.

Is that what it's called in Devlin et al?

dirkgr (approved):

dirkgr:
Let me know when this should get another review?

AkshitaB:
> Let me know when this should get another review?

@dirkgr The tests are finally passing. The main changes from your last review are reducing the size of tensors in the test so that there are no memory issues, and a bunch of merge updates. I'll merge once I get another green signal from you.

tests/modules/transformer/toolkit_test.py
```diff
@@ -11,6 +12,7 @@
 from allennlp.common.testing import AllenNlpTestCase
 
 
+# @pytest.mark.transformer_toolkit
```
dirkgr, line 15: ?

tests/modules/transformer/transformer_block_test.py
```diff
@@ -78,3 +107,67 @@ def test_loading_partial_pretrained_weights(self):
...
+        output = module.forward(hidden_states, attention_mask=attention_mask.squeeze())[0]
+        torch.manual_seed(1234)
+        hf_output = pretrained_module.forward(hidden_states, attention_mask=attention_mask_hf)[0]
+
+        # FIX: look into the reason for mismatch.
+        # Update: The discrepancy comes from torch.nn.Dropout layer, despite setting random seeds.
+        # Have also tried setting random seeds right before the actual call to dropout in both modules.
+        # assert torch.allclose(output, hf_output)
+        print(output)
+        print(hf_output)
```
dirkgr, line 173: Is this fixed?

dirkgr (approved):
Looks like there is a bit of debug code left. Looks good otherwise!

Issue #4775, opened 2020-11-07
Ask for further integration with Optuna


himkt:
Hello, I'm a member of Optuna dev and the author of the allennlp-guide chapter on hyperparameter optimization.

Recently, I created [allennlp-optuna](https://github.com/himkt/allennlp-optuna), a prototype of a wrapper for Optuna to enable users to optimize hyperparameter of AllenNLP model. It provides a way to use Optuna by `allennlp` subcommands.

For explaining `allennlp-optuna`, I made the quick tutorial of `allennlp-optuna` in [readthedocs](https://allennlp-optuna.readthedocs.io/en/latest/tutorial/index.html). With `allennlp-optuna`, users can run multi-node distributed optimization by simply executing the same command on multiple machines. I put a simple example on [README](https://github.com/himkt/allennlp-optuna#5-hyperparameter-optimization-at-scale).

And recently, I also wrote [the post](https://techlife.cookpad.com/entry/2020/11/06/110000) (sorry in Japanese...) about the NER system using AllenNLP+Optuna.

Can we have an opportunity to add `allennlp-optuna` to default plugins of AllenNLP?  If it helps the decision, I'm willing to transfer `allennlp-optuna` to AllenAI. I'm really happy if it helps NLP practitioners to tune the hyperparameter of their model with ease. 

Thank you, as always.

epwalsh:
Hi @himkt, I think what we'd like to do is create a place to prominently display third-party plugins like `allennlp-optuna`. For now, the best place for this is probably in our README. We should also do better with documenting how to use third-party plugins. Would that suffice?

himkt:
Hi @epwalsh, thank you for your consideration.

> create a place to prominently display third-party plugins

It feels really nice!

> We should also do better with documenting how to use third-party plugins

I agree. For now, it's not easy to know we can register subcommand to `allennlp` as I understand. If how to use third-party plugins is documented (in guide.allennlp.org [chapter 2](https://guide.allennlp.org/writing-your-own-script)?), it would help people who want to expand/utilize AllenNLP a lot!

PR #4779, opened 2020-11-10
support global plugins file, improve plugins docs

epwalsh:
Adds support for declaring a global plugins file at `~/.allennlp/plugins` and improves the documentation for plugins. This also closes https://github.com/allenai/allennlp/issues/4775 by added a place to display third-party plugins.

### DIFF
--- a/allennlp/common/plugins.py
+++ b/allennlp/common/plugins.py
@@ -7,8 +7,7 @@ provides custom registered classes or additional `allennlp` subcommands.
 In order for AllenNLP to find your plugins, you have to create either a local plugins
 file named `.allennlp_plugins` in the directory where the `allennlp` command is run, or a global
 plugins file at `~/.allennlp/plugins`. The file should list the plugin modules that you want to
-be loaded, one per line. A plugins file in the current directory will take precedence
-over a global plugins file in `~/.allennlp/`.
+be loaded, one per line.
 """
 
 import importlib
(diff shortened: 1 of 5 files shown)


allennlp/common/plugins.py
```diff
@@ -1,46 +1,68 @@
...
     """
     Returns an iterable of the plugins found.
     """
-    with push_python_path("."):
-        yield from discover_file_plugins()
+    if os.path.isfile(LOCAL_PLUGINS_FILENAME):
+        with push_python_path("."):
+            yield from discover_file_plugins(LOCAL_PLUGINS_FILENAME)
+    elif os.path.isfile(GLOBAL_PLUGINS_FILENAME):
+        yield from discover_file_plugins(GLOBAL_PLUGINS_FILENAME)
```
dirkgr, line 63: This means only one of the files will be respected. Wouldn't it make more sense to load from both files if they are both present?

allennlp/common/plugins.py
```diff
@@ -1,46 +1,68 @@
...
+    else:
+        yield from []
```
dirkgr, line 65: I don't think you need this. You can just let the function end.

dirkgr (changes requested):

allennlp/common/plugins.py
epwalsh: I think either way would make sense. I don't feel strongly about it, so I'm happy to make the change if you do.

allennlp/common/plugins.py
epwalsh: Yeup, you're right.

allennlp/common/plugins.py
```diff
@@ -56,13 +55,20 @@ def discover_plugins() -> Iterable[str]:
     """
     Returns an iterable of the plugins found.
     """
+    plugins: Set[str] = set()
     if os.path.isfile(LOCAL_PLUGINS_FILENAME):
         with push_python_path("."):
```
dirkgr, line 60: The global one doesn't need this?

dirkgr (approved):

allennlp/common/plugins.py
epwalsh: I guess the assumption made with global plugins is that they should be installed in your virtual environment, rather than just existing in the current directly.

PR #4793, opened 2020-11-13
improve independence of vision components

epwalsh:
This this does a couple things.

First, I added an optional "extra" to the `setup.py` called "vision", where we can specify vision-only requirements like `torchvision` and `detectron2`, except not `detectron2` because that has to be installed separately and prior to installing AllenNLP since it requires that `torch` is installed first 🙃

The other thing the PR does is sort of de-couple all of the vision-specific submodules so that AllenNLP won't break when the vision dependencies are missing, i.e. when someone just installs AllenNLP with `pip install allennlp`, as opposed to `pip install allennlp[vision]`.

This seems a little hacky with `try-except` statements around imports, but I don't know a better way aside from rewriting the entire library in Rust where this could be trivially solved with optional package features.


Dockerfile
```diff
@@ -20,7 +20,11 @@ WORKDIR /stage/allennlp
 
 # Install the wheel of AllenNLP.
 COPY dist dist/
-RUN pip install $(ls dist/*.whl)
+RUN pip install $(ls dist/*.whl)[vision]
+
+# Install detectron2.
+RUN pip install detectron2 -f https://dl.fbaipublicfiles.com/detectron2/wheels/cu102/torch1.7/index.html
```
dirkgr, line 26: This line has the CUDA version and the torch version built in. Is there a way to not be dependent on that?

tests/training/metrics/auc_test.py
```diff
@@ -100,6 +100,7 @@ def test_distributed_auc(self):
         labels = torch.randint(3, 5, (8,), dtype=torch.long)
         # We make sure that the positive label is always present.
         labels[0] = 4
+        labels[4] = 4
```
dirkgr, line 103: This is unrelated?

dirkgr (approved):

AkshitaB (approved):

tests/training/metrics/auc_test.py
epwalsh: Unrelated. Might need this on master too actually.

Dockerfile
epwalsh: Yea, good point. We could (and should) make that a build variable.

epwalsh: Once https://github.com/allenai/allennlp/pull/4796 goes through I'll add a build ARG for detectron.

PR #4799, opened 2020-11-17
SNLI_VE dataset reader

AkshitaB:
* Extract common vision dataset reader functions from VQA under base class `VisionReader`.
* Use `VisionReader` to implement `VisualEntailmentReader`.


allennlp/data/dataset_readers/visual_entailment.py
```diff
@@ -0,0 +1,102 @@
...
+            else:
+                features, coords = image
+
+            fields["box_features"] = ArrayField(features)
+            fields["box_coordinates"] = ArrayField(coords)
+
+        if label:
+            if label == "-":
+                # No gold label could be decided.
+                return None
```
dirkgr, line 94: This can never be `True`.

test_fixtures/data/visual_entailment/sample_pairs.jsonl
```diff
@@ -0,0 +1,16 @@
+{"Flickr30K_ID": "2248275918", "annotator_labels": ["contradiction"], "captionID": "2248275918.jpg#2", "gold_label": "contradiction", "pairID": "2248275918.jpg#2r1c", "sentence1": "A toddler poses in front of a computer at a business office.", "sentence1_binary_parse": "( ( A toddler ) ( ( poses ( in ( front ( of ( ( a computer ) ( at ( a ( business office ) ) ) ) ) ) ) ) . ) )", "sentence1_parse": "(ROOT (S (NP (DT A) (NN toddler)) (VP (VBZ poses) (PP (IN in) (NP (NP (NN front)) (PP (IN of) (NP (NP (DT a) (NN computer)) (PP (IN at) (NP (DT a) (NN business) (NN office)))))))) (. .)))", "sentence2": "A toddler sleeps outside.", "sentence2_binary_parse": "( ( A toddler ) ( ( sleeps outside ) . ) )", "sentence2_parse": "(ROOT (S (NP (DT A) (NN toddler)) (VP (VBZ sleeps) (ADVP (RB outside))) (. .)))"}
```
dirkgr, line 1: They come with parse trees? Woha.

tests/modules/transformer/bimodal_attention_test.py
```diff
@@ -49,7 +49,7 @@ def test_forward_runs(self):
 
         self.biattention.forward(
             torch.randn(2, 3, 6),
-            torch.randn(2, 2, 3, 3),
             torch.randn(2, 3, 4),
             torch.randn(2, 2, 3, 3),
+            torch.randn(2, 2, 3, 3),
         )
```
dirkgr, line 55: Why did this change?

dirkgr (changes requested):
This is great, just one confusion about a failure case.

tests/modules/transformer/bimodal_attention_test.py
AkshitaB: I changed the order of the arguments in bimodal_attention, to allow for masks to have default `None` values.

allennlp/data/dataset_readers/visual_entailment.py
AkshitaB: Ah, I did so based on this sentence in the paper: `

> Specifically, each (Ptext, Htext) pair is
assigned a gold label, indicating the label is agreed by a
majority of crowdsourcing workers (at least 3 out of 5). If
such a consensus is not reached, the gold label is marked as
“-”.

It looks like those aren't actually part of the final dataset, though.

AkshitaB: Removed.

allennlp/data/dataset_readers/visual_entailment.py
```diff
@@ -0,0 +1,98 @@
...
+                failed_instances_count += 1
+            else:
+                yield instance
+
+            if attempted_instances_count % 2000 == 0:
+                failed_instances_fraction = failed_instances_count / attempted_instances_count
+                if failed_instances_fraction > 0.1:
+                    logger.warning(
+                        f"{failed_instances_fraction*100:.0f}% of instances have no answers."
+                    )
```
dirkgr, line 63: Now, `text_to_instance()` will never return `None` anymore, so we don't need this check either.

allennlp/data/dataset_readers/visual_entailment.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @overrides
+    def text_to_instance(
+        self,  # type: ignore
+        image: Union[str, Tuple[Tensor, Tensor]],
+        hypothesis: str,
+        label: Optional[str] = None,
+        *,
+        use_cache: bool = True,
+    ) -> Optional[Instance]:
```
dirkgr, line 73: Also, the return type can be just `Instance`.

dirkgr (approved):
There are some more simplifications you can do now that `text_to_instance()` never fails.

Issue #4810, opened 2020-11-22
GradientDescentTrainer breaks when constructed with validation_data_loader==None and learning_rate_scheduler!=None


IINemo:
## Checklist

- [x] I have verified that the issue exists against the `master` branch of AllenNLP.
- [x] I have read the relevant section in the [contribution guide](https://github.com/allenai/allennlp/blob/master/CONTRIBUTING.md#bug-fixes-and-new-features) on reporting bugs.
- [x] I have checked the [issues list](https://github.com/allenai/allennlp/issues) for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/allenai/allennlp/pulls) for existing proposed fixes.
- [x] I have checked the [CHANGELOG](https://github.com/allenai/allennlp/blob/master/CHANGELOG.md) and the [commit log](https://github.com/allenai/allennlp/commits/master) to find out if the bug was already fixed in the master branch.
- [x] I have included in the "Description" section below a traceback from any exceptions related to this bug.
- [x] I have included in the "Related issues or possible duplicates" section beloew all related issues and possible duplicate issues (If there are none, check this box anyway).
- [x] I have included in the "Environment" section below the name of the operating system and Python version that I was using when I discovered this bug.
- [x] I have included in the "Environment" section below the output of `pip freeze`.
- [x] I have included in the "Steps to reproduce" section below a minimally reproducible example.

## Description

When you construct GradientDescentTrainer with validation_data_loader==None and learning_rate_scheduler!=None, the code
breaks when an update step is performed for learning_rate_scheduler.  This is a typical case for training a Transformer model. 

This issue appeared since version 1.2.0 and is present now in the master branch.

<details>
<summary><b>Python traceback:</b></summary>
<p>

```
UnboundLocalErrorTraceback (most recent call last)
<ipython-input-9-513339770b41> in <module>
     46 
     47 try:
---> 48     metrics = trainer.train()
     49 except KeyboardInterrupt:
     50     pass

/opt/.pyenv/versions/3.7.4/lib/python3.7/site-packages/allennlp/training/trainer.py in train(self)
    964         """
    965         try:
--> 966             return self._try_train()
    967         finally:
    968             # make sure pending events are flushed to disk and files are closed properly

/opt/.pyenv/versions/3.7.4/lib/python3.7/site-packages/allennlp/training/trainer.py in _try_train(self)
   1080             # if it doesn't, the validation metric passed here is ignored.
[...]

PR #4811, opened 2020-11-22
Fix issue in GradientDescentTrainer, when validation data is absent

IINemo:
* Prevent an error when GradientDescentTrainer is constructed with
validation_data_loader=None and learning_rate_scheduler!=None

This pull request fixes issue: https://github.com/allenai/allennlp/issues/4810


epwalsh (approved):
Good catch, thanks @IINemo!

PR #4815, opened 2020-11-23
Bring back NLVR

dirkgr:
This reverts commit 824f83223b2c74341e172f9f158085924e57d474.

NLVR doesn't work like this, but it could be made to work relatively easily.


dirkgr:
There is little benefit at this point in bringing this back.

PR #4821, opened 2020-11-26
Switch to torchvision for vision components 👀

epwalsh:
Still a few things to do:

- [x] Implement an image loader using `torchvision`
- [x] Implement the `ResnetBackbone` `GridEmbedder` using `torchvision`
- [x] Implement the `FasterRcnnRegionDetector` using `torchvision`
- [x] Make sure the pipeline reconstructed using our wrappers gives the same output as the full pipeline in torchvision.
- [x] Get the various vision dataset readers up-to-date with the new API
  - [x] Add field to VQAv2 reader instances that keeps track of the number of boxes. That way we can construct a mask for the boxes / box features downstream.
- [x] Get the VilBERT model up-to-date with the new API (there might not be anything else we need to do there)
- [ ] Re-train VilBERT


allennlp/models/vision_text_model.py
```diff
@@ -196,7 +193,7 @@ def forward(
         else:
             extended_attention_mask = None
 
-        extended_image_attention_mask = image_attention_mask.unsqueeze(1).unsqueeze(2).float().log()
+        extended_image_attention_mask = box_mask.unsqueeze(1).unsqueeze(2).float().log()
```
dirkgr, line 196: You confirmed that this actually does something useful?

allennlp/modules/vision/grid_embedder.py
```diff
@@ -1,127 +1,69 @@
...
-
-    For every image, it embeds a patch of the image, and returns the embedding
-    of the patch. The size of the image might change during this operation.
+    A `GridEmbedder` takes a batch of images as a tensor with shape
+    `(batch_size, color_channels, height, width)`, and returns an ordered dictionary
+    of tensors with shape `(batch_size, *)`, each representing a specific feature.
     """
 
-    def forward(self, images: FloatTensor, sizes: IntTensor) -> FloatTensor:
+    def forward(self, images: FloatTensor, sizes: IntTensor) -> "OrderedDict[str, FloatTensor]":
```
dirkgr, line 17: `OrderedDict`?

allennlp/modules/vision/region_detector.py
```diff
@@ -1,31 +1,46 @@
...
-    keys are "coordinates", "features", and "class_probs".
+
+    - `features` (`List[Tensor]`): A feature vector for each region, which is a tensor of shape
+      `(num_boxes, feature_dim)`.
+    - `boxes` (`List[Tensor]`): The coordinates of each region within the original image, with shape
+      `(num_boxes, 4)`.
+    - `class_probs` (`Optional[List[Tensor]]`): Class probabilities from some object
+      detector that was used to find the regions of interest, with shape `(num_predicted_classes,)`.
+    - `class_labels` (`Optional[List[Tensor]]`): The labels corresponding to `class_probs`.
+      Each tensor in this list has the same shape as the corresponding tensor in `class_probs`.
```
dirkgr, line 34: How would we get labels?

allennlp/modules/vision/region_detector.py
```diff
@@ -37,194 +52,238 @@ class RandomRegionDetector(RegionDetector):
...
-        return {"features": features, "coordinates": coordinates}
+            boxes[image_num][0, 2] = sizes[image_num, 0]
+            boxes[image_num][0, 3] = sizes[image_num, 1]
+            boxes[image_num][1, 2] = sizes[image_num, 0]
+            boxes[image_num][1, 3] = sizes[image_num, 1]
+        return RegionDetectorOutput(features, boxes)
 
 
 @RegionDetector.register("faster_rcnn")
 class FasterRcnnRegionDetector(RegionDetector):
```
dirkgr, line 76: Not in this PR, but it would be good if we cached these somewhere/somehow. In a multi-task setting, multiple readers will be featurizing images in the same way, and they don't all need to instantiate their own frozen models to do so.

dirkgr (changes requested):
I have only small requests. I really hope this does at least as well as the detectron stuff, because the code is much better.

allennlp/modules/vision/region_detector.py
epwalsh: They are in the torchvision docs: https://pytorch.org/docs/stable/torchvision/models.html#object-detection-instance-segmentation-and-person-keypoint-detection

allennlp/modules/vision/grid_embedder.py
epwalsh: See https://mypy.readthedocs.io/en/latest/common_issues.html#using-classes-that-are-generic-in-stubs-but-not-at-runtime.

allennlp/models/vision_text_model.py
epwalsh: The mask is necessary.

allennlp/data/image_loader.py
```diff
@@ -1,69 +1,195 @@
...
+            scale_factor = self.min_size / min_size
+            if max_size * scale_factor > self.max_size:
+                scale_factor = self.max_size / max_size
+            image = torch.nn.functional.interpolate(
+                image[None],
+                scale_factor=scale_factor,
+                mode="bilinear",
+                recompute_scale_factor=True,
+                align_corners=False,
+            )[0]
```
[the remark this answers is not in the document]
epwalsh: In order to use these particular transforms directly from torchvision we will need to load a batch of images into an `ImageList`  type of object, instead of one big tensor.

allennlp/data/image_loader.py
```diff
@@ -1,69 +1,195 @@
...
-            raise ValueError("Unknown type of `config`")
-
-        self.mapper = pipeline.mapper
-
-    def load(self, filenames: ManyPaths) -> ImagesWithSize:
-        images = [{"file_name": str(f)} for f in filenames]
-        images = [self.mapper(i) for i in images]
-        images = ImageList.from_tensors([image["image"] for image in images])
-        return (images.tensor.float() / 256, torch.tensor(images.image_sizes, dtype=torch.int32))  # type: ignore
+            image = (image - mean) / std
```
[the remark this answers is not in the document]
epwalsh: Correct.

allennlp/data/image_loader.py
```diff
@@ -1,69 +1,195 @@
...
-        if config is None:
-            pipeline = detectron.get_pipeline_from_flat_parameters(
-                make_copy=False, fp=DetectronFlatParameters()
+        image_backend: str = None,
+        resize: bool = True,
+        normalize: bool = True,
+        min_size: int = 800,
+        max_size: int = 1333,
+        pixel_mean: Tuple[float, float, float] = (0.485, 0.456, 0.406),
+        pixel_std: Tuple[float, float, float] = (0.229, 0.224, 0.225),
```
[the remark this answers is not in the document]
epwalsh: Actually this is what all torchvision pretrained models use:

image

(https://pytorch.org/docs/stable/torchvision/models.html)

allennlp/data/dataset_readers/vision_reader.py
```diff
@@ -134,6 +140,20 @@ def __init__(
 
             self.image_processing_batch_size = image_processing_batch_size
 
+    @property
+    def image_featurizer(self) -> GridEmbedder:
+        if self._image_featurizer is None:
+            self._image_featurizer = self._lazy_image_featurizer.construct().to(self.cuda_device)
+            self._image_featurizer.eval()  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
epwalsh: This happened here before: https://github.com/allenai/allennlp/blob/vision/allennlp/common/detectron.py#L121

allennlp/data/image_loader.py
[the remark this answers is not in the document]
dirkgr: Hmm. Alright then!

allennlp/modules/vision/grid_embedder.py
dirkgr: I'm not confused about the string form of the type. I'm confused why it has to be an `OrderedDict` instead of a `Dict`.

allennlp/modules/vision/region_detector.py
dirkgr: I don't understand. Are these "labels" not gold annotated data from humans?

allennlp/modules/vision/grid_embedder.py
epwalsh: This is consistent with `torchvision`. I'm guessing it's because there is a natural order to the features, as each feature comes from a different layer of the model.

epwalsh:
I'm getting this warning while training:

image

Does that mean the BERT weights are not getting applied at initialization?

epwalsh:
I'll be training ViLBERT on VQA on Beaker here: https://beaker.org/ws/ai2/ViLBERT-VQA/experiments?archived=false&sort=created:descending

epwalsh:
I discovered multiple issues with multi-process data loading that stemmed from playing with tensors in workers and also passing said tensors around between processes. See https://github.com/allenai/allennlp/issues/4848 and https://github.com/allenai/allennlp/issues/4847. I spent a lot of effort figuring out what was going on here, and in the process decided to drastically simplify the `MultiProcessDataLoader` while making it more robust.

These changes should really be in a separate PR, and I may pull them out later, but right now I need them on this branch so I can train ViLBERT faster.

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -94,39 +94,37 @@ class MultiProcessDataLoader(DataLoader):
...
-        The default value should be reasonable for most use-cases.
+        The first work-around is to disable parallelism for these low-level libraries.
+        For example, setting the environment variables `OMP_NUM_THREADS=1` and `TOKENIZERS_PARALLELISM=0`
+        will do so for PyTorch and Numpy (for CPU operations) and HugggingFace Tokenizers, respectively.
 
-    batch_queue_size: `int`, optional (default = `1000`)
-        This determines the size of the of the
-        [multiprocessing queue](https://docs.python.org/3.7/library/multiprocessing.html#multiprocessing.Queue)
-        used to pass `TensorDict` batches to the consuming process when `num_workers > 0`.
+        Alternatively, changing the `start_method` to "spawn" (when available, depending on your OS)
```
dirkgr, line 112: Can we do this automatically on the right OSs?

dirkgr: I guess it has a performance penalty if the alternative is "fork".

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -340,24 +358,67 @@ def _instance_worker(self, worker_id: int, queue: mp.JoinableQueue) -> None:
...
+        # Wait until this process can safely exit.
         queue.join()
 
+    def _gather_instances(self, queue: mp.JoinableQueue) -> Iterable[Instance]:
+        done_count: int = 0
+        while done_count < self.num_workers:
+            for instance, worker_error in iter(queue.get, (None, None)):
+                if worker_error is not None:
+                    e, tb = worker_error
+                    sys.stderr.write("".join(tb))
```
dirkgr, line 392: Why not log it?

allennlp/models/vision_text_model.py
```diff
@@ -168,10 +168,30 @@ def forward(
...
+            Shape: `(batch_size, num_boxes, 4)`
+
+        box_mask : `Tensor`
+            A bool and 0-1 tensor of shape `(batch_size, num_boxes)`.
+
+        text : `TextFieldTensors`
+
+        label : `Optional[Tensor]`
+
+        label_weights : `Optional[Tensor]`
```
dirkgr, line 192: Better than no documentation at all ...

allennlp/modules/transformer/self_attention.py
```diff
@@ -116,7 +117,7 @@ def forward(
             mask_reshp = (batch_size, 1, 1, k_length)
             attention_mask = (attention_mask == 0).view(mask_reshp).expand_as(
                 attention_scores
-            ) * -10e5
+            ) * nn_util.min_value_of_dtype(attention_scores.dtype)
```
dirkgr, line 120: Is that really what we use? Let's use whatever huggingface uses. `min_value` seems risky, because of numeric weirdnesses at the edge of floating point precision.

dirkgr: The old value is a bit weird too though. `-10e5`? Why wouldn't you write `-1e6` then? 🤷

dirkgr: https://github.com/huggingface/transformers/blob/master/src/transformers/modeling_utils.py#L272

tests/training/trainer_test.py
```diff
@@ -150,7 +150,7 @@ def test_data_loader_lazy_epoch_size_correct(self):
             self.model,
             self.optimizer,
             self.data_loader_lazy,
-            validation_data_loader=self.validation_data_loader,
+            #  validation_data_loader=self.validation_data_loader,
```
dirkgr, line 153: Debug code?

dirkgr (changes requested):
I reviewed this, but I don't know if you thought it was in a state to be reviewed. I skimmed the actual data loading guts a bit. I think I learned something about how these queues are supposed to work though :-)

allennlp/data/dataset_readers/vqav2.py
```diff
@@ -533,3 +530,35 @@ def text_to_instance(
...
     def apply_token_indexers(self, instance: Instance) -> None:
         instance["question"].token_indexers = self._token_indexers  # type: ignore
+
+    def _get_answers_by_question_id(self, split):
+        answers_by_question_id = {}
+        if split.annotations is not None:
+            # Pre-processing the annotations is time-consuming, so we don't want to
+            # have to re-do it each time we call read(). So we cache this result.
+            annotations_path = cached_path(split.annotations, extract_archive=True)
+            with LocalCacheResource(split.annotations + "-cache", annotations_path) as cache:
```
dirkgr, line 540: What's this? I missed this the first time around. This calculation used to take seconds. Now building the cache takes minutes. Why did you cache this?

dirkgr (changes requested):

allennlp/data/dataset_readers/vqav2.py
dirkgr: It takes 2 minutes on my machine now (without writing the cache). That's certainly odd. Not sure why this started.

dirkgr: With a4baae6ed4431c019e27c2a1719cd308669fabda, it takes 15 seconds on my machine to calculate them all. I think that's preferable to caching to disk.

allennlp/data/dataset_readers/vqav2.py
```diff
@@ -534,31 +533,17 @@ def apply_token_indexers(self, instance: Instance) -> None:
     def _get_answers_by_question_id(self, split):
```
epwalsh, line 533: Did you mean to decorate this with `lru_cache` as well?

dirkgr: No, though I could see it. In a normal run, that function only gets called twice, and with different arguments. But with our way of using the validation set twice, we could save 15 seconds, I guess.

dirkgr: 15 seconds per epoch, that is

epwalsh: This gets called each epoch (since we're doing lazy loading). And takes about 3 minutes to run each time. It's definitely worth caching.

dirkgr: Why does it take so much time? It's 15 seconds on my Mac.

epwalsh: Whaaat? How is it so quick for you? It consistently takes 2.5 - 3 minutes on every computer I've ran it on. See these beaker logs for instance:

image

dirkgr: Are you using a different config? Different dataset?

epwalsh: This is the config: https://data.beaker.org/datasets/01erwyaezkf7csjmyj8hkavm0q/files/config.json

allennlp/data/data_loaders/multi_process_data_loader.py
epwalsh: Logging might be disabled

allennlp/data/dataset_readers/vision_reader.py
```diff
@@ -134,6 +140,20 @@ def __init__(
 
             self.image_processing_batch_size = image_processing_batch_size
 
+    @property
+    def image_featurizer(self) -> GridEmbedder:
+        if self._image_featurizer is None:
+            self._image_featurizer = self._lazy_image_featurizer.construct().to(self.cuda_device)
```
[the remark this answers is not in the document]
epwalsh: I like that. I don't know if it's worth adding another dep or just implementing it ourselves, because it's pretty simple.

allennlp/common/file_utils.py
```diff
@@ -474,6 +493,8 @@ def __init__(
             sync=True,
             readahead=False,
             meminit=False,
+            readonly=read_only,
+            lock=not read_only,
```
dirkgr, line 497: We still need the lock. A file that's opened read-only by one process might be opened read-write by another.

allennlp/common/file_utils.py
```diff
@@ -464,8 +470,21 @@ def __init__(
...
+                if not os.access(filename, os.W_OK):
+                    if not read_only:
+                        warnings.warn(
+                            f"File '{filename}' is read-only, so cache will be read-only",
+                            UserWarning,
+                        )
+                    read_only = True
+            else:
+                # If it's not a file, raise an error.
+                raise ValueError("Expect a file, found a directory instead")
```
dirkgr, line 485: Sweet!

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -94,39 +94,37 @@ class MultiProcessDataLoader(DataLoader):
...
-        at a time when `num_workers > 0`.
+    !!! Warning
+        Multiprocessing code in Python is complicated! Especially code that involves lower-level libraries
+        which may be spawning their own threads / processes. If you run into dead-locks while
+        using `num_workers > 0`, luckily there are two simple work-arounds which usually fix the issue.
 
-        The default value should be reasonable for most use-cases.
+        The first work-around is to disable parallelism for these low-level libraries.
+        For example, setting the environment variables `OMP_NUM_THREADS=1` and `TOKENIZERS_PARALLELISM=0`
+        will do so for PyTorch and Numpy (for CPU operations) and HugggingFace Tokenizers, respectively.
```
dirkgr, line 110: `ggg`

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -94,39 +94,37 @@ class MultiProcessDataLoader(DataLoader):
...
+        Another issue besides dead-locks that you could run into when using `num_workers > 0`
+        is running out of shared memory, since tensors are passed between processes
+        using shared memory, and some systems impose strict limits on the allowed size of shared
+        memory.
 
-        The default value should be reasonable for most use-cases.
+        Luckily there is also a simple work-around for this. Either decrease `max_instances_in_memory`
+        or increase your system's `ulimit`.
 
+        See [issue #4847](https://github.com/allenai/allennlp/issues/4847) for more info.
```
dirkgr, line 126: Is there a way we can catch errors like this and make this message part of the error message?

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -293,42 +336,43 @@ def _start_instance_workers(self, queue: mp.JoinableQueue, ctx) -> List[BaseProc
...
+    def _join_workers(self, workers: List[BaseProcess], queue) -> None:
+        # Each worker will be blocking on a call to `queue.join()`,
+        # calling `queue.task_done()` times the number of workers will
+        # call the `queue.join()` to return, and each worker should exit on its own.
+        for _ in range(len(workers)):
+            try:
+                queue.task_done()
+            except ValueError:
+                # This happens if a worker died early.
+                break
```
dirkgr, line 358: Don't we have to do something about that? How are we propagating the error from the failed worker?

allennlp/data/data_loaders/multi_process_data_loader.py
dirkgr: Then your process silently exits with an error code, as expected, since you disabled logging.

Do we do this anywhere else like this?

dirkgr: We re-throw the exception below, so don't we get a stack trace anyways if the exception is uncaught?

allennlp/data/dataset_readers/vqav2.py
dirkgr: I don't understand this, but it's not important. Maybe you're reading from a slow drive? Anyways, if you think we should cache, let's cache.

allennlp/models/vision_text_model.py
```diff
@@ -231,7 +287,10 @@ def forward(
         else:
             raise ValueError(f"Fusion method '{self.fusion_method}' not supported")
 
+        # Shape: (batch_size, num_labels)
         logits = self.classifier(pooled_output)
+
+        # Shape: (batch_size, num_labels)
```
dirkgr, line 293: We should find some time to experiment with named tensor dimensions. I think a lot of our code could be much more readable if we used them.

dirkgr (changes requested):
The only real requested change is the one for `TensorCache`. The rest looks fine.

allennlp/common/file_utils.py
epwalsh: You need write permissions to use the lock though.

allennlp/data/data_loaders/multi_process_data_loader.py
epwalsh: Not that I'm aware of. In some cases the process just hangs.

allennlp/data/data_loaders/multi_process_data_loader.py
epwalsh: This is executed in a `finally` cause. The error is still propagated.

allennlp/data/data_loaders/multi_process_data_loader.py
epwalsh: ?

allennlp/data/data_loaders/multi_process_data_loader.py
epwalsh: Silently exit? I don't think that's ever expected.

If I'm just running a script that uses AllenNLP, I might not enable logging, and so `logging.error` messages would never show up. I'd still expect to see error messages in `stderr` though.

epwalsh: This just reminded me I have to make a fix here. There is no guarantee that exceptions will be pickle-able, and so we can't send an exception from a worker process back to the main process. So instead, we just send the traceback as a string and a string representation of the exception back to the main process. So the best we can is print out the traceback to stderr from the main process, since the traceback that the main process creates will not include the traceback from the worker.

epwalsh: I don't know if I did a good job explaining that.

It's definitely not ideal. There may be a better way that I'm not aware of yet.

dirkgr: I looked up how I solved this problem in [bettermap](http://github.com/allenai/bettermap), and I actually came to the same conclusion you did: https://github.com/allenai/bettermap/blob/master/bettermap.py#L108

I don't remember why, but if both you and past me think it's a good idea, I think it's probably fine :-)

That code though has never had a problem with pickling exceptions. Maybe because it uses `dill` instead of `pickle`?

allennlp/common/file_utils.py
dirkgr: Hmm. We should at least make it safe to do this if you have permission to use the lock though. You would definitely expect safety if you yourself have one process that writes to this and another one that reads.

If someone else tries to read it at the same time, and their process crashes because they couldn't lock, then we did the best we could.

allennlp/data/data_loaders/multi_process_data_loader.py
dirkgr: just a typo in "hugggingface"

allennlp/data/data_loaders/multi_process_data_loader.py
dirkgr: 😩

allennlp/data/data_loaders/multi_process_data_loader.py
dirkgr: :+1:

allennlp/data/data_loaders/multi_process_data_loader.py
epwalsh: Well I believe most exceptions are pickled just fine, so it's very likely you just never ran into an exception that was not pickle-able.

allennlp/common/file_utils.py
epwalsh: Good point.

epwalsh: Just updated.

dirkgr (approved):

Issue #4823, opened 2020-11-26
Add the Gaussian Error Linear Unit as an Activation option


tomsherborne:
The [Gaussian Error Linear Unit](https://arxiv.org/pdf/1606.08415.pdf) activation is currently not a possible option from the set of registered Activations. Since this class just directly called the PyTorch classes - adding this in is a 1 line addition. Motivation is that models like BART/BERT use this activation in many places and elegant consistency of activation function across models that are "something pretrained" + "more weights trained on AllenNLP" would be nice. 

**Describe the solution you'd like**
Add the following snippet to the end of the [Activations class](https://github.com/allenai/allennlp/blob/master/allennlp/nn/activations.py) class
```
    "gelu": (torch.nn.GELU, None),
```

**Describe alternatives you've considered**
Manually hardcoding the activation. This isn't very robust and modules such as FeedForward complain since Gelu isnt a registered activation to insert between layers (as far as I can tell). 

Thanks - happy to submit a tiny PR for this

epwalsh:
Hi @tomsherborne, yes please go ahead a submit a PR for this. Thanks!

PR #4828, opened 2020-11-30
Add GELU activation

tomsherborne:
Following up from Issue #4823. This PR adds the Gaussian Error Linear Unit activation to the Activations class in AllenNLP.


epwalsh (approved):
LGTM!

PR #4829, opened 2020-12-01
fix: removes all handlers from the root logger

tshu-w:
We should not mutate (insert/remove elements) the list you're currently iterating on. In fact, this causes handlers to not be completely deleted on my side.

For more discussion, see https://stackoverflow.com/questions/7484454/removing-handlers-from-pythons-logging-loggers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added links to source code in docs.
+
+### Fixed
+
 - Fixed issue with GradientDescentTrainer when constructed with validation_data_loader==None and learning_rate_scheduler!=None.
+- Fixed a bug when removing all handlers in root logger.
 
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added links to source code in docs.
 - Fixed issue with GradientDescentTrainer when constructed with validation_data_loader==None and learning_rate_scheduler!=None.
+- Fixed issue with removing all handlers in root logger.
```
epwalsh, line 15: This should be under the `### Fixed` section (you may need to create this section if it doesn't exist under the `## Unreleased` section.

epwalsh (changes requested):
Thanks @tshu-w, great catch! Just need one small tweak to the CHANGELOG and then we can get this merged.

CHANGELOG.md
tshu-w: @epwalsh Does it look okay now? BTW, I've also moved the above fixed to this new section.

epwalsh (approved):
LGTM! Thanks!

Issue #4825, opened 2020-11-30
ShardedDatasetReader doesn't inherit lazy from base reader


aleSuglia:
## Checklist

- [x] I have verified that the issue exists against the `master` branch of AllenNLP.
- [x] I have read the relevant section in the [contribution guide](https://github.com/allenai/allennlp/blob/master/CONTRIBUTING.md#bug-fixes-and-new-features) on reporting bugs.
- [x] I have checked the [issues list](https://github.com/allenai/allennlp/issues) for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/allenai/allennlp/pulls) for existing proposed fixes.
- [x] I have checked the [CHANGELOG](https://github.com/allenai/allennlp/blob/master/CHANGELOG.md) and the [commit log](https://github.com/allenai/allennlp/commits/master) to find out if the bug was already fixed in the master branch.
- [x] I have included in the "Description" section below a traceback from any exceptions related to this bug.
- [x] I have included in the "Related issues or possible duplicates" section beloew all related issues and possible duplicate issues (If there are none, check this box anyway).
- [x] I have included in the "Environment" section below the name of the operating system and Python version that I was using when I discovered this bug.
- [x] I have included in the "Environment" section below the output of `pip freeze`.
- [x] I have included in the "Steps to reproduce" section below a minimally reproducible example.

## Description

Currently the `ShardedDatasetReader` doesn't inherit the `lazy` property from the `base_reader`. Therefore, the dataset reader becomes not `lazy` anymore. 

<details>
<summary><b>Python traceback:</b></summary>
<p>

```
Processed killed for insufficient RAM
```

</p>
</details>

## Related issues or possible duplicates

- None

## Environment

OS: OS X

Python version: Python 3.8.5

<details>
<summary><b>Output of <code>pip freeze</code>:</b></summary>
<p>

```
allennlp==1.2.2
attrs==20.3.0
beautifulsoup4 @ file:///tmp/build/80754af9/beautifulsoup4_1601924105527/work
blis==0.7.3
boto3==1.16.25
botocore==1.19.25
catalogue==1.0.0
certifi==2020.11.8
chardet==3.0.4
click==7.1.2
cymem==2.0.4
en-core-web-sm @ https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-2.3.1/en_core_web_sm-2.3.1.tar.gz
filelock==3.0.12
future==0.18.2
h5py==3.1.0
idna==2.10
[...]

PR #4830, opened 2020-12-01
Bugfix for attribute inheritance in ShardedDatasetReader

aleSuglia:
Implemented automatic parameters inheritance from `base_reader` in `ShardedDatasetReader`. This allows the user to specify parameters like `lazy` or `max_instances` either for the `base_reader` or the `ShardedDatasetReader` itself. 

Fixes the issue reported in https://github.com/allenai/allennlp/issues/4825

### DIFF
--- a/allennlp/data/dataset_readers/sharded_dataset_reader.py
+++ b/allennlp/data/dataset_readers/sharded_dataset_reader.py
@@ -13,6 +13,8 @@ from allennlp.data.instance import Instance
 
 logger = logging.getLogger(__name__)
 
+_INHERITED_DATASET_PARAMS = ("lazy",)
+
 
 @DatasetReader.register("sharded")
 class ShardedDatasetReader(DatasetReader):
@@ -67,7 +69,8 @@ class ShardedDatasetReader(DatasetReader):
         # can be safely inherited. However, ShardedDatasetReader is a class instance of a DatasetReader as well.
         # So we give priority to the parameters for the current instance stored in 'kwargs'.
         # If not present, we check the ones in the base reader
-        for attr_name, attr_val in self.reader.__dict__.items():
+        for attr_name in _INHERITED_DATASET_PARAMS:
+            attr_val = getattr(self.reader, attr_name)
             # copy over only shared attributes between the two classes
             if attr_name in self.__dict__:
                 setattr(self, attr_name, kwargs.get(attr_name, attr_val))
--- a/allennlp/data/dataset_readers/sharded_dataset_reader.py
+++ b/allennlp/data/dataset_readers/sharded_dataset_reader.py
@@ -13,8 +13,6 @@ from allennlp.data.instance import Instance
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


allennlp/data/dataset_readers/sharded_dataset_reader.py
```diff
@@ -59,6 +63,15 @@ def __init__(self, base_reader: DatasetReader, **kwargs) -> None:
...
         self.reader.manual_distributed_sharding = True
 
+        # ShardedDatasetReader is a wrapper for the original base_reader so some of the parameters like 'lazy'
+        # can be safely inherited. However, ShardedDatasetReader is a class instance of a DatasetReader as well.
+        # So we give priority to the parameters for the current instance stored in 'kwargs'.
+        # If not present, we check the ones in the base reader
+        for attr_name, attr_val in self.reader.__dict__.items():
+            # copy over only shared attributes between the two classes
+            if attr_name in self.__dict__:
+                setattr(self, attr_name, kwargs.get(attr_name, attr_val))
```
epwalsh, line 73: This makes me a little nervous. I'd rather just explicitly set only the attributes we care about (`lazy` and mabye `max_instances`). Although, I'm not sure about `max_instances`...

Setting `max_instances` in the `base_reader` but not in the `ShardedDatasetReader` itself actually has clear and well-defined behavior: at most `max_instances` will be read from *each* shard.

epwalsh (changes requested):

allennlp/data/dataset_readers/sharded_dataset_reader.py
aleSuglia: I see. That makes a lot of sense. I will make the change.

allennlp/data/dataset_readers/sharded_dataset_reader.py
```diff
@@ -59,6 +65,16 @@ def __init__(self, base_reader: DatasetReader, **kwargs) -> None:
...
 
+        # ShardedDatasetReader is a wrapper for the original base_reader so some of the parameters like 'lazy'
+        # can be safely inherited. However, ShardedDatasetReader is a class instance of a DatasetReader as well.
+        # So we give priority to the parameters for the current instance stored in 'kwargs'.
+        # If not present, we check the ones in the base reader
+        for attr_name in _INHERITED_DATASET_PARAMS:
+            attr_val = getattr(self.reader, attr_name)
+            # copy over only shared attributes between the two classes
+            if attr_name in self.__dict__:
+                setattr(self, attr_name, kwargs.get(attr_name, attr_val))
```
epwalsh, line 76: I think this is more complicated than it needs to be. I also think it would be more robust to set `lazy` on the call to `super().__init__(...)` on line 46 above, since the base class `DatasetReader` might have special logic in the `__init__` method that depends on the value of `lazy` (well, that's not the case right now, but it could be in the future).

So something like (right above line 46):

```python
if "lazy" not in kwargs:
    kwargs["lazy"] = base_reader.lazy
```

epwalsh (approved):
Looks great! Thanks!

PR #4832, opened 2020-12-01
Added GQA reader

jvstokes:
Adds / Tests GQA dataset reader


allennlp/data/dataset_readers/gqa.py
```diff
@@ -0,0 +1,161 @@
...
+            )
+
+            processed_images = self._process_image_paths(
+                self.images[f"{question_dict['imageId']}.jpg"] for question_dict in question_dicts
+            )
+
+            for question_dict, processed_image in zip(question_dicts, processed_images):
+                answers = {
+                    "answer": question_dict["answer"],
+                    "fullAnswer": question_dict["fullAnswer"],
```
AkshitaB, line 128: We don't actually return `fullAnswer` in the `text_to_instance` function, right? Any reason to pass it to the function?

tests/data/dataset_readers/gqa_test.py
```diff
@@ -0,0 +1,107 @@
...
+from allennlp.data import Batch, Vocabulary
+from allennlp.data.dataset_readers import GQAReader
+from allennlp.data.image_loader import DetectronImageLoader
+from allennlp.data.tokenizers import WhitespaceTokenizer
+from allennlp.data.token_indexers import SingleIdTokenIndexer
+from allennlp.modules.vision.grid_embedder import NullGridEmbedder
+from allennlp.modules.vision.region_detector import RegionDetector
+
+
+class FakeRegionDetector(RegionDetector):
```
AkshitaB, line 16: I think you can use the `RandomRegionDetector` instead of having to specify a fake one here. It's basically the same code logic.

AkshitaB (changes requested):

allennlp/data/dataset_readers/gqa.py
jvstokes: Ahh I wasn't sure if fullAnswer would be used later but I'll remove it!

AkshitaB (approved):

PR #4833, opened 2020-12-02
Several micro optimizations

epwalsh:
Mainly replacing calls where a tensor is created and then sent to a device with `.to(device)` with calls that create the tensor directly on the device, which is over twice as fast. You can run the benchmarks yourself with:

```bash
pytest -c benchmarks/pytest.ini benchmarks/nn/util_bench.py -k 'create_tensor'
```

These are the results I got:

image


allennlp/nn/util.py
```diff
@@ -1548,7 +1548,6 @@ def add_sentence_boundary_token_ids(
         The new mask for the tensor, taking into account the appended tokens
         marking the beginning and end of the sentence.
     """
-    # TODO: matthewp, profile this transfer
```
epwalsh, line 1551: I benchmarked this. This function is actually faster this way than with keeping `sequence_lengths` on GPU with:

```
sequence_lengths = mask.sum(dim=1).detach()
```

AkshitaB (approved):

dirkgr (approved):
Heh, cool!

Issue #4790, opened 2020-11-13
Interpret code should be usable for non-AllenNLP models


AkshitaB:
* Currently, saliency interpreters assume that the input is an AllenNLP model and rely on the presence of `TextFieldEmbedder` to find the correct embedding layer to register hooks. 
* We want to add the ability for the user to specify the right embedding layer for the cases when the model does not have a `TextFieldEmbedder`.

jianliu-ml:
Hi, looking forward for this function!
Could you share ideas for how to impelement a interpret function for a pytroch-based model?

PR #4836, opened 2020-12-02
Add ability to specify the embedding layer if the model does not use `TextFieldEmbedder`

AkshitaB:
Fixes https://github.com/allenai/allennlp/issues/4790


dirkgr (commented):
Controversial opinion: Why don't we put `get_embedding_layer()` on the model instead of on the predictor, with the default implementation we already have? It would be pretty trivial to return the correct layer from the model most of the time, and it gives us the option of keeping predictors more general.

What do you think of changing the name at the same time? I struggle to come up with something more [wieldy](https://english.stackexchange.com/questions/129391/how-did-wieldy-the-positive-form-of-unwieldy-come-to-be-a-non-existent-wor) than `get_interpretable_input_layer()`.

AkshitaB:
> Controversial opinion: Why don't we put `get_embedding_layer()` on the model instead of on the predictor, with the default implementation we already have? It would be pretty trivial to return the correct layer from the model most of the time, and it gives us the option of keeping predictors more general.

The advantage with keeping it in Predictor is that you can then wrap a predictor around a non-AllenNLP model and use the interpreters. That was the original use case.

> 
> What do you think of changing the name at the same time? I struggle to come up with something more [wieldy](https://english.stackexchange.com/questions/129391/how-did-wieldy-the-positive-form-of-unwieldy-come-to-be-a-non-existent-wor) than `get_interpretable_input_layer()`.

Can't think of a better name off the top of my head. Perhaps it makes sense to add the need for the method in the docstring? That this is required for interpret code?

dirkgr:
> The advantage with keeping it in Predictor is that you can then wrap a predictor around a non-AllenNLP model and use the interpreters. That was the original use case.

Fair enough. I think I'd love to have both, where `Predictor` by default defers to the `Model`, but it can do something else if needed. But let's not overload this PR.

dirkgr (approved):
Looks good, but I'd really like to change those names. Can we make them `get_interpretable_layer()` and `get_interpretable_text_field_embedder()`? I know those names are long, but for the vision models, the layer you would want is already not really an embedding layer.

AkshitaB:
> Looks good, but I'd really like to change those names. Can we make them `get_interpretable_layer()` and `get_interpretable_text_field_embedder()`? I know those names are long, but for the vision models, the layer you would want is already not really an embedding layer.

@dirkgr Done. The names are changed.

PR #4841, opened 2020-12-04
fix FromParams bug

epwalsh:
This was an interesting one. I found this after digging into the test that seemed to fail out of nowhere in this (unrelated) PR: https://github.com/allenai/allennlp-models/pull/175/checks?check_run_id=1499712028. This bug only manifested itself because of a different bug fix regarding the `ShardedDatasetReader` (https://github.com/allenai/allennlp/pull/4830).

Basically, the `ShardedDatasetReader` needs to inspect its `**kwargs` for the "lazy" parameter. The behavior depends on whether or not "lazy" is present. But the current `FromParams` logic removes the "lazy" param from `kwargs` if it's the same as the default value in the superclass.

So this fixes the behavior of `FromParams` so that parameters in the `Params` object will ALWAYS be passed to the constructor.


dirkgr (approved):
I don't have the full context, so I trust that you've thought through this more than I have. Is there any chance this will break other expected behavior?

epwalsh:
> Is there any chance this will break other expected behavior?

I can't be completely certain, but we do have very high test coverage here that covers every edge case I can think of so far.

PR #4843, opened 2020-12-04
Adding multilabel option to `VisionTextModel`

AkshitaB:


dirkgr (approved):
Looks good. Thanks for jumping on this. I assume there is no change necessary for losses because those are computed in the individual models anyways.

AkshitaB:
> Looks good. Thanks for jumping on this. I assume there is no change necessary for losses because those are computed in the individual models anyways.

Yes, that's right.

PR #4846, opened 2020-12-06
Improve the band-aid solution for seg faults and the static TLS error

eladsegal:
I kept getting "ImportError: dlopen: cannot load any more object with static TLS" and segmentation faults.
At first, it seemed the cause is the import of `sklearn` in the AUC metric. After adding `sklearn` import before everything else, I didn't get the static TLS error anymore, but continued getting seg faults which came from imports from `transformers`.
Adding a `transformers` import before all other imports helped with both errors.


epwalsh (approved):
I don't understand why this happens but I'm glad we have a work-around. Thanks!

PR #4851, opened 2020-12-08
gqa reader fixes during vilbert training

jvstokes:
Adds updates to the GQA reader to allow it to be used to train the vilbert_vqa model.

- Modifies the instance dict to include a ListField for labels

- Checks that answers are present in vocabulary to be included in instance labels


dirkgr (approved):
Looks pretty straightforward, and it looks like I'll need this to make multitask work.

Issue #4839, opened 2020-12-04
Superfluous warning when extending the vocab in the `Embedding`


dcfidalgo:
## Checklist

- [x] I have verified that the issue exists against the `master` branch of AllenNLP.
- [x] I have read the relevant section in the [contribution guide](https://github.com/allenai/allennlp/blob/master/CONTRIBUTING.md#bug-fixes-and-new-features) on reporting bugs.
- [x] I have checked the [issues list](https://github.com/allenai/allennlp/issues) for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/allenai/allennlp/pulls) for existing proposed fixes.
- [x] I have checked the [CHANGELOG](https://github.com/allenai/allennlp/blob/master/CHANGELOG.md) and the [commit log](https://github.com/allenai/allennlp/commits/master) to find out if the bug was already fixed in the master branch.
- [x] I have included in the "Description" section below a traceback from any exceptions related to this bug.
- [x] I have included in the "Related issues or possible duplicates" section beloew all related issues and possible duplicate issues (If there are none, check this box anyway).
- [x] I have included in the "Environment" section below the name of the operating system and Python version that I was using when I discovered this bug.
- [x] I have included in the "Environment" section below the output of `pip freeze`.
- [x] I have included in the "Steps to reproduce" section below a minimally reproducible example.

## Description

If one creates an `allennlp.modules.token_embedders.embedding.Embedding` without a `pretrained_file`, you still get a warning when extending the vocab that no `pretrained_file` is found. I would expect the warning only to trigger if one specified a `pretrained_file` when creating the `Embedding` or when an `extension_pretrained_file` is passed on to `Embedding.extend_vocab`. 

I would be more than happy to provide a PR if you think this is actually and issue and should be addressed.

<details>
<summary><b>Python traceback:</b></summary>
<p>

```
WARNING:root:Embedding at model_path, None cannot locate the pretrained_file.  If you are fine-tuning and want to use using pretrained_file for embedding extension, please pass the mapping by --embedding-sources argument.

```

</p>
</details>

## Related issues or possible duplicates

- None

## Environment

OS: Ubuntu 20.04

Python version: 3.8.0

<details>
<summary><b>Output of <code>pip freeze</code>:</b></summary>
<p>

```
attrs==20.3.0
blis==0.7.3
boto3==1.16.29
botocore==1.19.29
catalogue==1.0.0
certifi==2020.11.8
chardet==3.0.4
[...]

PR #4854, opened 2020-12-09
Turn superfluous warning to info when extending the vocab in the embedding matrix

dcfidalgo:
This PR addresses #4839: one gets a warning about a missing pretrained_file, although no pretrained_file was specified in the first place.

I basically exchanged the warning with an info, and made use of the defined module loggger instead of using the root logger.


dirkgr (approved):

dirkgr:
Thank you for the contribution!

PR #4859, opened 2020-12-11
Toolkit: Adding documentation and small changes for `BiModalAttention`

AkshitaB:
@dirkgr This includes the `_apply_mask` function that you added in `torchvision`, with minor changes. I shifted the reshaping part to the module (though it will still work if the reshaping happens outside) to make it consistent with `SelfAttention`.

### DIFF
--- a/allennlp/modules/transformer/bimodal_attention.py
+++ b/allennlp/modules/transformer/bimodal_attention.py
@@ -2,10 +2,8 @@ import torch
 
 from allennlp.common import FromParams
 from allennlp.modules.attention import Attention
-
-# from allennlp.modules.transformer.attention_scores import ATTN_MAP
-
 from allennlp.modules.transformer.transformer_module import TransformerModule
+from allennlp.nn import util as nn_util
 
 
 class BiModalAttention(TransformerModule, FromParams):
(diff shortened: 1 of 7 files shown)


allennlp/modules/transformer/bimodal_attention.py
```diff
@@ -74,6 +112,17 @@ def _transpose_for_scores(self, x):
...
 
+    @staticmethod
+    def _apply_mask(values: torch.FloatTensor, mask: torch.BoolTensor) -> torch.FloatTensor:
+        if len(mask.shape) == 2:
+            # We create a 4D attention mask from a 2D tensor mask.
+            # The shape is `batch_size x 1 x 1 x target_seq_len` which is broadcast
+            # to `batch_size x num_attention_heads x source_seq_len x target_seq_len`
+            mask = mask.unsqueeze(1).unsqueeze(2)
+        # `mask==1` to convert float tensors.
+        mask = (~(mask == 1)) * nn_util.min_value_of_dtype(values.dtype)
```
dirkgr, line 123: I'm not sure it makes sense to use the min value for the type. For one thing, that tensor might get cast to float16 later, and for another, it might be weird to have a value there that goes to `-inf` at the drop of a hat.

AkshitaB: Fixed!

allennlp/modules/transformer/bimodal_attention.py
```diff
@@ -74,6 +111,17 @@ def _transpose_for_scores(self, x):
...
 
+    @staticmethod
+    def _apply_mask(values: torch.FloatTensor, mask: torch.BoolTensor) -> torch.FloatTensor:
+        if len(mask.shape) == 2:
+            # We create a 4D attention mask from a 2D tensor mask.
+            # The shape is `batch_size x 1 x 1 x target_seq_len` which is broadcast
+            # to `batch_size x num_attention_heads x source_seq_len x target_seq_len`
+            mask = mask.unsqueeze(1).unsqueeze(2)
+        # `mask==1` to convert float tensors.
+        mask = (~(mask == 1)) * -10e5  # to ensure that the model also works in half-precision mode.
```
dirkgr, line 122: Can you make clear in the comment which part of the code makes sure it works in half-precision?
```suggestion
        mask = (~(mask == 1)) * -10e5  # -10e5 to ensure that the model also works in half-precision mode.
```

allennlp/modules/transformer/self_attention.py
```diff
@@ -72,6 +72,17 @@ def _transpose_for_scores(self, x):
...
+    @staticmethod
+    def _apply_mask(values: torch.FloatTensor, mask: torch.BoolTensor) -> torch.FloatTensor:
+        if len(mask.shape) == 2:
+            # We create a 4D attention mask from a 2D tensor mask.
+            # The shape is `batch_size x 1 x 1 x seq_len` which is broadcast
+            # to `batch_size x num_attention_heads x seq_len x seq_len`
+            mask = mask.unsqueeze(1).unsqueeze(2)
+        # `mask==1` to convert float tensors.
+        mask = (~(mask == 1)) * -10e5  # to ensure that the model also works in half-precision mode.
+        return values + mask
```
dirkgr, line 84: If we need this in many places, can we put it into a single place? This is always the same for all transformers, right? So some general place for the toolkit?

dirkgr (approved):
It's good as it is. I have two minor suggestions.

We don't expect this would change any scores, right?

allennlp/modules/transformer/self_attention.py
AkshitaB: Done.

PR #4863, opened 2020-12-14
Adding gqa training config

jvstokes:


training_configs/vilbert_gqa_from_huggingface.jsonnet
```diff
@@ -0,0 +1,92 @@
...
+local construct_vocab = false;
+
+local vocabulary = if construct_vocab then {
+      // read the files to construct the vocab
+      "min_count": {"answers": 1}
+    } else {
+      // read the constructed vocab
+      "type": "from_files",
+      // CHANGE LOC
+      "directory": "/Users/jacksons/Projects/gqa-train/vocabulary/vilbert_gqa_train.vocab.tar.gz"
```
dirkgr, line 15: Can you upload this to GCloud? If you don't know how or where, ping me.

training_configs/vilbert_gqa_from_huggingface.jsonnet
```diff
@@ -0,0 +1,92 @@
...
+      "tokens": {
+        "type": "pretrained_transformer",
+        "model_name": model_name
+      }
+    },
+    // Change max instances
+    "max_instances": 1000,
+    "image_processing_batch_size": 16,
+    "answer_vocab": if construct_vocab then null else vocabulary,
+    "keep_unanswerable_questions": false
```
dirkgr, line 41: This must depend on whether we're constructing vocab or not, doesn't it?

training_configs/vilbert_gqa_from_huggingface.jsonnet
```diff
@@ -0,0 +1,92 @@
...
+  },
+  "trainer": {
+    "optimizer": {
+        "type": "huggingface_adamw",
+        "lr": 4e-5
+    },
+    "learning_rate_scheduler": {
+      "type": "linear_with_warmup",
+      "warmup_steps": 300000 / 30,
+      "num_steps_per_epoch": std.ceil(644401 / $["data_loader"]["batch_size"] / $["trainer"]["num_gradient_accumulation_steps"])
```
dirkgr, line 85: Did you verify that this number makes sense? Is that the correct number in this case?

Is the GQA dataset even large enough that we need to worry about lazy loading? If we don't lazy load, than it figures out this number for us.

dirkgr (changes requested):

training_configs/vilbert_gqa_from_huggingface.jsonnet
jvstokes: Should I keep the min_count size of 9 for this vocab? generating on beaker now

dirkgr: Yes, let's stick with that number.

Generating vocab takes only a few minutes on my machine.

dirkgr:
Closed in favor of #4884.

PR #4865, opened 2020-12-15
Make tests work again

dirkgr:


allennlp/nn/activations.py
```diff
@@ -99,5 +99,4 @@ def _get_name(self):
     "softsign": (torch.nn.Softsign, None),
     "tanhshrink": (torch.nn.Tanhshrink, None),
     "selu": (torch.nn.SELU, None),
-    "gelu": (torch.nn.GELU, None),
```
epwalsh, line 102: Why remove this?

dirkgr: It's a duplicate. It's already there a few lines up.

epwalsh: Ah, good catch

epwalsh (approved):

PR #4866, opened 2020-12-15
log import errors for default plugins

epwalsh:


allennlp/common/plugins.py
```diff
@@ -86,8 +86,9 @@ def import_plugins() -> None:
             # For default plugins we recursively import everything.
             import_module_and_submodules(module_name)
             logger.info("Plugin %s available", module_name)
-        except ModuleNotFoundError:
-            pass
+        except ModuleNotFoundError as e:
+            if e.name != module_name:
```
AkshitaB, line 90: This is to check if the failure is in a submodule?

epwalsh: Yea

AkshitaB (approved):

PR #4867, opened 2020-12-15
issue warning instead of failing when lock can't be acquired on a resource that exists in a read-only file system

### LINKED ISSUE #4838: Can't load a model from a readonly file system because of Filelock
Basically see here https://github.com/allenai/allennlp/blob/b7cec515448f46daec9a76551d0a3342ef51b27b/allennlp/data/vocabulary.py#L341 the `FileLock` tries to lock the file in the same directory as the model. This means that if the model is not on a writable file system, it will not be able to get a lock and fails. Existing code
```
        # We use a lock file to avoid race conditions where multiple processes
        # might be reading/writing from/to the same vocab files at once.
        with FileLock(os.path.join(directory, ".lock")):
```

This seems like an unnecessary restriction, why should we not be able to load the model from a readonly file system. More specifically, a docker image can be run on another machine where assumptions about the machine cannot be made, (e.g AWS Lambda)

Ideally, given that we have a cache directory, why not hold the `.lock` file in the cache directory. It is a fair assumption that the cache dir is writable.

epwalsh:
Closes #4838


allennlp/common/file_utils.py
```diff
@@ -63,6 +65,38 @@
...
+        self, lock_file: Union[str, PathLike], timeout=-1, read_only_ok: bool = False
+    ) -> None:
+        super().__init__(str(lock_file), timeout=timeout)
+        self._read_only_ok = read_only_ok
+
+    @overrides
+    def acquire(self, timeout=None, poll_interval=0.05):
+        try:
+            super().acquire(timeout=timeout, poll_intervall=poll_interval)
+        except PermissionError:
```
vikigenius, line 89: @epwalsh  I don't think it catches all types of errors. For instance when running from a Docker I get
`OSError: [Errno 30] Read-only file system: '/opt/ml/model/vocabulary/.lock'`

Here is the traceback

```
  File "/opt/program/predictor.py", line 43, in get_model
    cls.model.initialize(model_path)
  File "/opt/program/SiameseEncoder.py", line 45, in initialize
    "model.transformer_model": siamese_base_path,
  File "/usr/local/lib/python3.6/site-packages/allennlp/models/archival.py", line 208, in load_archive
    model = _load_model(config.duplicate(), weights_path, serialization_dir, cuda_device)
  File "/usr/local/lib/python3.6/site-packages/allennlp/models/archival.py", line 246, in _load_model
    cuda_device=cuda_device,
  File "/usr/local/lib/python3.6/site-packages/allennlp/models/model.py", line 406, in load
    return model_class._load(config, serialization_dir, weights_file, cuda_device)
  File "/usr/local/lib/python3.6/site-packages/allennlp/models/model.py", line 294, in _load
    vocab_dir, vocab_params.get("padding_token"), vocab_params.get("oov_token")
  File "/usr/local/lib/python3.6/site-packages/allennlp/data/vocabulary.py", line 339, in from_files
    with FileLock(os.path.join(directory, ".lock"), read_only_ok=True):
  File "/usr/local/lib/python3.6/site-packages/filelock.py", line 323, in __enter__
    self.acquire()
  File "/usr/local/lib/python3.6/site-packages/allennlp/common/file_utils.py", line 88, in acquire
    super().acquire(timeout=timeout, poll_intervall=poll_interval)
  File "/usr/local/lib/python3.6/site-packages/filelock.py", line 271, in acquire
    self._acquire()
  File "/usr/local/lib/python3.6/site-packages/filelock.py", line 384, in _acquire
    fd = os.open(self._lock_file, open_mode)
```

epwalsh: Ah, good catch, thank you! Just updated to catch `OSError` as well.

allennlp/common/file_utils.py
```diff
@@ -63,6 +65,38 @@
...
+        self, lock_file: Union[str, PathLike], timeout=-1, read_only_ok: bool = False
+    ) -> None:
+        super().__init__(str(lock_file), timeout=timeout)
+        self._read_only_ok = read_only_ok
+
+    @overrides
+    def acquire(self, timeout=None, poll_interval=0.05):
+        try:
+            super().acquire(timeout=timeout, poll_intervall=poll_interval)
+        except (PermissionError, OSError):
```
dirkgr, line 89: Is there a way to be more specific? This could capture all sorts of things.

epwalsh: Yeup, good point. Just updated.

allennlp/common/file_utils.py
```diff
@@ -86,7 +86,12 @@ def __init__(
     def acquire(self, timeout=None, poll_interval=0.05):
         try:
             super().acquire(timeout=timeout, poll_intervall=poll_interval)
-        except (PermissionError, OSError):
+        except (PermissionError, OSError) as err:
+            if isinstance(err, OSError):
+                # OSError could be a lot of different things, but what we're looking
+                # for in particular is Linux errno 30 - EROFS - "Read-only file system".
```
dirkgr, line 92: Well, this is interesting. When the whole file system is read-only, we're guaranteed to not have concurrent writers, so we don't even need to print the warning. I don't know if you want to get that granular though ...

epwalsh: > When the whole file system is read-only, we're guaranteed to not have concurrent writers

That's not necessarily true. As @vikigenius mentioned above, this error can occur from within a Docker container while trying to access an attached volume. So while it is true that you couldn't have concurrent writers within the same Docker container, you could have concurrent writers on separate Docker containers or on the host OS itself.

dirkgr: Ah, yes.

allennlp/common/file_utils.py
epwalsh: Oops, `PermissionError` is actually a subclass of `OSError`. Just fixed.

allennlp/common/file_utils.py
```diff
@@ -86,12 +86,15 @@ def __init__(
...
-                # for in particular is Linux errno 30 - EROFS - "Read-only file system".
-                if err.errno != 30:
-                    raise
+        except OSError as err:
+            # OSError could be a lot of different things, but what we're looking
+            # for in particular are permission errors, such as:
+            #  - errno 1  - EPERM  - "Operation not permitted"
+            #  - errno 13 - EACCES - "Permission denied"
+            #  - errno 30 - EROFS  - "Read-only file system"
+            if err.errno not in (1, 13, 30):
```
dirkgr, line 95: In principle, this should be a set, not a tuple, i.e., `{1, 13, 30}`. In practice, it doesn't matter.

dirkgr (approved):

PR #4869, opened 2020-12-16
Fix cache volume

epwalsh:
Same fix as https://github.com/allenai/allennlp-demo/pull/659


epwalsh:
I need to merge this before anything else runs or the cache will break again.

PR #4870, opened 2020-12-17
Only cache, no featurizing

dirkgr:
Adds the ability to just read from a cache instead of featurizing images.

There are some differences in the way that the vision readers are configured now. To turn off featurizing, you just don't supply any featurizers. To turn off the cache, don't supply a cache. We don't need extra flags for that.


allennlp/modules/vision/region_detector.py
```diff
@@ -74,6 +76,19 @@ class RandomRegionDetector(RegionDetector):
     the proposal are a random 10-dimensional vector, and the coordinates are the size of the image.
     """
 
+    def __init__(self, seed: Optional[int] = None):
+        super().__init__()
+        self.random = random.Random(seed)
+
+    def _seeded_random_tensor(self, *shape: int, device) -> torch.FloatTensor:
```
AkshitaB, line 83: Cool! This _may_ be helpful for the non-deterministic dropout layers that were causing (currently commented) test failures in the transformer toolkit. I'll try it out.

tests/data/dataset_readers/vqav2_test.py
```diff
@@ -53,3 +53,14 @@ def test_read(self):
         # Nothing should be masked out since the number of fake boxes is the same
         # for each item in the batch.
         assert tensors["box_mask"].all()
+
+    def test_read_without_images(self):
```
AkshitaB, line 57: What do we gain by allowing the ability to read without images?

allennlp/modules/vision/region_detector.py
dirkgr: Cool indeed if this fixes an unrelated problem! Totally uncool from pytorch that we have to do it this way though...

tests/data/dataset_readers/vqav2_test.py
dirkgr: We use that to make vocab.

allennlp/data/dataset_readers/vision_reader.py
```diff
@@ -183,14 +218,35 @@ def _coordinates_cache(self) -> MutableMapping[str, Tensor]:
...
+        image_paths: `Iterable[str]`
+            the image paths to process
+        use_cache: `bool`, default = `True`
+            Usually the cache behavior is governed by the `write_to_cache` parameter given to
+            `__init__()`. But sometimes we want to override this behavior and turn off the
+            cache completely. This parameter lets you do that. This is useful for the
+            `Predictor`, so we can make predictions without having to touch a cache,
+            even if the model was trained with a cache.
+        """
+        assert self.produce_featurized_images, (
```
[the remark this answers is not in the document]
dirkgr: If we have neither feature cache nor featurization components, the reader outputs no image features at all. That's a legitimate configuration. It's just `_process_image_paths()` that cannot be called when you're configured that way.

tests/data/dataset_readers/vqav2_test.py
AkshitaB: Ah, ok. So the configuration of not calling `_process_image_paths` when reading serves this purpose.

AkshitaB (approved):

training_configs/vilbert_vqa_pretrained.jsonnet
```diff
@@ -54,7 +53,7 @@ local vocabulary = if construct_vocab then {
   "model": {
     "type": "vqa_vilbert_from_huggingface",
     "model_name": model_name,
-    "image_feature_dim": 1024,
+    "image_feature_dim": 2048,
```
dirkgr, line 56:
```suggestion
    "image_feature_dim": 1024,
```
This wasn't supposed to be changed.

PR #4871, opened 2020-12-18
add CombinedLearningRateScheduler

### LINKED ISSUE #4856: Make configuring learning rate schedulers more flexible
In particular, it would be great if you do something like this:

```jsonnet
"learing_rate_scheduler": [
  // use this scheduler for the first two epochs
  {"type": "linear_with_warmup", "num_epochs": 2},
  // use this scheduler for the remaining epochs
  {"type": "cosine"},
]
```

epwalsh:
Closes #4856. This will allow me to try some other things with VilBERT VQA.

### DIFF
--- a/allennlp/training/learning_rate_schedulers/combined.py
+++ b/allennlp/training/learning_rate_schedulers/combined.py
@@ -103,9 +103,10 @@ class CombinedLearningRateScheduler(LearningRateScheduler):
 
     @overrides
     def get_values(self):
-        if self.current_scheduler is not None:
-            return self.current_scheduler.get_values()
-        return self.base_values
+        """
+        This should never be called directly.
+        """
+        raise NotImplementedError
 
     @overrides
     def step_batch(self, batch_num_total: int = None) -> None:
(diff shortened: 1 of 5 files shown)


allennlp/training/learning_rate_schedulers/combined.py
```diff
@@ -0,0 +1,121 @@
...
+@LearningRateScheduler.register("combined")
+class CombinedLearningRateScheduler(LearningRateScheduler):
+    """
+    This `LearningRateScheduler` can be used to apply an arbitrary number of other schedulers
+    one after the other.
+
+    These schedulers are defined though the `schedulers` parameter, which takes a list
+    of `Tuple[int, Lazy[LearningRateScheduler]]`. The first field of the tuple, the `int`,
+    specifies how many epochs the corresponding scheduler will be used before the next
+    scheduler takes its place.
```
dirkgr, line 19: It would be neat if we could schedule these in a finer-grained way than per epoch, but it'll do. In fact, since we can just define epochs to be arbitrary numbers of steps, this might be all we ever need.

dirkgr (approved):

PR #4879, opened 2020-12-21
Data loading cuda device

### LINKED ISSUE #4670: Add memory pinning option to new data loader
See 
- https://pytorch.org/docs/stable/data.html#memory-pinning
- https://github.com/pytorch/pytorch/blob/master/torch/utils/data/_utils/pin_memory.py#L45
- https://github.com/pytorch/pytorch/blob/master/torch/utils/data/dataloader.py#L410

epwalsh:
- ~~Adds memory pinning option to new data loader.~~
- Data loaders load tensors directly to target CUDA device. This leads to a slight speed up with the `MultiProcessDataLoader` when `num_workers > 0` and `start_method` is "spawn" (or "forkserver").
- Improves the `nn.util.move_to_device` function.


dirkgr (approved):
This is cool. Did you benchmark it in any way, or is it just "a good idea"?

I guess nothing is using the new flags yet, but it would be good to know how important proper pinning is to performance.

dirkgr (commented):
The reason this doesn't run out of memory is because we're only moving batches to the GPU, not instances.

What happens when we're training with multiple GPUs? Will it do the right thing?

Do you have before/after numbers for how this impacts training time?

Where did this happen before? What's the intuition why it's faster now?

epwalsh:
> What happens when we're training with multiple GPUs? Will it do the right thing?

The data loader is assigned the same GPU as the trainer, so it will work fine when training in distributed mode.

> Do you have before/after numbers for how this impacts training time?

It only impacts lazy, multi-process data loading / training. I tested this with TransformerQA and saw a speed-up of ~3%. Not large, but not trivial either. And I suspect we'd see a bigger improvement on datasets that have larger tensors like VQA.

> Where did this happen before? What's the intuition why it's faster now?

It happened in the main process in the trainer. It's faster (when using multiple data loader workers) now because the CPU to GPU transfer happens in a worker thread, so it doesn't block the main thread.

dirkgr (approved):

PR #4885, opened 2020-12-23
add path to MultiTaskDataLoader to __init__.py

jbrry:
The `MultiTaskDataLoader` currently cannot be found by my system without the relative import. This change means that the class should be able to be found.


epwalsh (approved):
Thanks!

jbrry:
I updated the branch on my end. Apologies if this caused any problems! This was my first time editing a file from a repository on GitHub. Let me know if there is anything I need to do.

epwalsh:
All good!

PR #4898, opened 2021-01-04
Multitask example

dirkgr:
This has gotten fairly big. Not only does it add an example, it also fixed everything that broke while building the example.


dirkgr:
900 lines of code 😂

allennlp/commands/train.py
```diff
@@ -557,16 +554,16 @@ def from_partial_objects(
         serialization_dir: str,
         local_rank: int,
         dataset_reader: DatasetReader,
-        train_data_path: DataPath,
+        train_data_path: Any,
```
dirkgr, line 557: Inputs for multitask are dicts. Some other readers do lists.

allennlp/data/data_loaders/multitask_data_loader.py
```diff
@@ -126,13 +104,10 @@ def __init__(
         self,
         reader: MultiTaskDatasetReader,
         data_path: Dict[str, str],
-        batch_size: int,
+        scheduler: MultiTaskScheduler,
```
dirkgr, line 107: The biggest change here is that now the `MultiTaskScheduler` is responsible for making batches, not the data loader. When the data loader makes batches, but the scheduler sorts the instances, we can't guarantee things like homogeneous batches, so I changed it.

dirkgr: This also means that the scheduler is no longer optional. You _have_ to specify one. Otherwise we don't even know a batch size.

allennlp/data/data_loaders/multitask_data_loader.py
```diff
@@ -206,56 +173,27 @@ def __init__(
...
             # This will raise a TypeError if any of the underlying loaders doesn't have a length,
-            # which is actually what we want.  If the loader has a length, we set batch_size = 1, so
-            # this will give us the right number of instances.
-            total_instances += self._batch_size_multiplier.get(key, 1.0) * len(loader)
-        if self._drop_last or total_instances % self._batch_size == 0:
-            return int(total_instances) // self._batch_size
+            # which is actually what we want.
+            return self.scheduler.count_batches(
+                {dataset: len(loader) for dataset, loader in self._loaders.items()}
+            )
```
dirkgr, line 181: Because schedulers now make batches, only schedulers know how many batches they are making. So the responsibility to count is devolved to them.

allennlp/models/multitask.py
```diff
@@ -84,58 +84,59 @@ def __init__(
...
-        task to run during training is incorrect (e.g., if your head has multiple optional
-        arguments, and only some are actually required to compute a loss), then you can use this
-        method to override our inference and force the use of whatever set of heads you want.
-
-        To get back to the default mode of operation, call this method with `None` as an argument.
-        """
-        self._active_heads = active_heads
-
     def forward(self, **kwargs) -> Dict[str, torch.Tensor]:  # type: ignore
+        if "task" not in kwargs:
```
dirkgr, line 98: The change here is that the old code automatically detected which heads to run based on the input parameters. If the signature fits, it would run. In the case of VQA/GQA though, we are running exactly the same head twice, once for each dataset, so this approach couldn't work. I changed it so now the instance keeps the task name in a `MetadataField`, and uses that to determine which task to run.

allennlp/models/multitask.py
```diff
@@ -84,58 +84,59 @@ def __init__(
...
+        task_indices: Dict[str, torch.LongTensor] = {
+            task: torch.LongTensor(indices) for task, indices in task_indices_just_for_mypy.items()
+        }
+
+        def make_inputs_for_task(task: str, whole_batch_input: Union[torch.Tensor, List]):
+            if isinstance(whole_batch_input, torch.Tensor):
+                task_indices[task] = task_indices[task].to(whole_batch_input.device)
+                return torch.index_select(whole_batch_input, 0, task_indices[task])
+            else:
+                return [whole_batch_input[i] for i in task_indices[task]]
```
dirkgr, line 117: This machinery splits up the batch by head, and runs the heads separately. I wonder if in this way we can support heterogeneous batches more or less transparently and for all models, though in the interest of time I did not try this.

allennlp/models/vision_text_model.py
```diff
@@ -57,17 +56,19 @@ def __init__(
...
-        self.v_pooler = TransformerPooler(encoder.hidden_size2, pooled_output_dim)
+        self.backbone = VilbertBackbone(
+            vocab,
+            text_embeddings,
+            image_embeddings,
+            encoder,
+            pooled_output_dim,
+            fusion_method,
+            dropout,
+        )
```
dirkgr, line 71: This changed so we don't need to duplicate the backbone. We still duplicate the heads, but they are tiny.

allennlp/modules/backbones/vilbert_backbone.py
```diff
@@ -0,0 +1,234 @@
...
+from allennlp.data.vocabulary import Vocabulary
+from allennlp.modules.backbones.backbone import Backbone
+from allennlp.modules.transformer import BiModalEncoder, ImageFeatureEmbeddings, Embeddings
+
+logger = logging.getLogger(__name__)
+
+
+@Backbone.register("vilbert")
+@Backbone.register("vilbert_from_huggingface", constructor="from_huggingface_model_name")
+class VilbertBackbone(Backbone):
```
dirkgr, line 17: Most of the code in here is cut and paste from `VisionTextModel`.

allennlp/training/metric_tracker.py
```diff
@@ -10,27 +10,26 @@ class MetricTracker:
...
+    `metric_name` that starts with "+" to indicate an increasing metric, or "-"
+    to indicate a decreasing metric.
 
     # Parameters
 
+    metric_name : `Union[str, List[str]]`
+        Specifies the metric or metrics to track. Metric names have to start with
+        "+" for increasing metrics or "-" for decreasing ones. If you specify more
+        than one, it tracks the sum of the increasing metrics metrics minus the sum
+        of the decreasing metrics.
```
dirkgr, line 23: For multitask, it does not make sense to track only one metric for early stopping, so I gave this thing the ability to track the sum of multiple metrics. If you want something fancier you need to write your own derivation of `MultitaskModel` that produces a metric you can track. I'm hoping that this will capture the 90% case though.

allennlp/data/data_loaders/multitask_data_loader.py
```diff
@@ -206,56 +173,27 @@ def __init__(
...
-        # Based on how we yield batches above, we are guaranteed to always have leftover instances,
-        # so we don't need a check for that here.
-        if not self._drop_last or current_batch_size == self._batch_size:
-            batch = Batch(batch_instances)
-            tensor_dict = batch.as_tensor_dict()
-            if self.cuda_device is not None:
-                tensor_dict == nn_util.move_to_device(tensor_dict, self.cuda_device)
-            yield tensor_dict
+        return (
+            nn_util.move_to_device(
```
epwalsh, line 191: Calling `move_to_device` here isn't necessary as long as we pass the target device through to `self._loaders` in `set_target_device()`.

epwalsh (commented):
I haven't finished a full review yet, just have one comment so far.

allennlp/data/data_loaders/multitask_data_loader.py
dirkgr: That sounds correct. I will check.

allennlp/data/data_loaders/multitask_data_loader.py
```diff
@@ -309,9 +247,9 @@ def _get_instances_for_epoch(self) -> Dict[str, Iterable[Instance]]:
 
     def _make_data_loader(self, key: str) -> MultiProcessDataLoader:
```
epwalsh, line 248: Line 259-262 can be removed. Those params to the data loader don't exist anymore.

allennlp/data/data_loaders/multitask_data_loader.py
epwalsh: Oh, never mind. This is definitely necessary.

allennlp/data/data_loaders/multitask_data_loader.py
```diff
@@ -309,9 +247,9 @@ def _get_instances_for_epoch(self) -> Dict[str, Iterable[Instance]]:
...
         kwargs: Dict[str, Any] = {}
-        kwargs["reader"] = self.readers[key]
+        kwargs["reader"] = _MultitaskDatasetReaderShim(self.readers[key], key)
         kwargs["data_path"] = self.data_paths[key]
-        kwargs["batch_size"] = 1
+        kwargs["batch_size"] = 1  # So that the loader gives us one instance at a time.
```
epwalsh, line 252: We only use it to call `iter_instances()`, right? So `batch_size` won't matter.

allennlp/data/data_loaders/multitask_data_loader.py
dirkgr: No, wait. We read `Instance`s from the child loaders, not batches. Instances don't have devices. This is the place where we make batches, and only those have devices, so this is where we have to set the device.

epwalsh (commented):
One concern I have is that the `MultiProcessDataLoader` is only being used to load `Instances`, and so we lose out on one of it's advantages: that indexing, padding, and batching can be done in worker processes, not the main process.

But by only using the `MultiProcessDataLoader` to call `iter_instances()`, `MultiTaskDataLoader` becomes responsible for indexing, padding, and batching, which then have to be done in the main process unless we add some extra multi-processing logic to `MultiTaskDataLoader`, which I think is a bad idea because that stuff is complicated.

Here's an alternative suggestion:

`MultiTaskDataLoader` becomes a very thin wrapper around `MultiProcessDataLoader` that just uses a special `BatchSampler` to handle the scheduling and make sure only instances from the same task are batched together.

I'm pretty sure this would work, but I haven't thought through every detail. I also realize it's probably unrealistic to make big changes this late in the game. I don't think this would actually require any API changes though, so we might be able to make these changes later on if it's worth it without breaking backwards compatibility.

allennlp/data/data_loaders/multitask_data_loader.py
dirkgr: done

allennlp/data/data_loaders/multitask_data_loader.py
dirkgr: True, but we have to set something. I clarified the comment.

dirkgr:
> `MultiTaskDataLoader` becomes a very thin wrapper around `MultiProcessDataLoader` that just uses a special `BatchSampler` to handle the scheduling and make sure only instances from the same task are batched together.

You are right that this is probably better, but you are also right that it's too late now. I don't think it will be quite as easy as you say, since we'd still have to manage the child data loaders.

epwalsh (approved):

mahnerak:
In `metric_tracker.py` it implements `state_dict()`/`load_state_dict()` for `self._patience` but not for `self.tracked_metrics`.
I suppose either both of them should be serialized/deserialized, or none of them.

I mean, it depends on whether these values are going to be changed during the training. Basically, we should serialize/deserialize the same states that we see in the `clear()` method.

dirkgr:
You are absolutely correct: https://github.com/allenai/allennlp/pull/4928

PR #4900, opened 2021-01-06
Toolkit: Cleaning up TransformerEmbeddings

AkshitaB:

- `TransformerEmbeddings` and `ImageFeatureEmbeddings` inherit from the general `Embeddings` class.
- Added docstrings and expected input shapes.
- Fixed the issue of non-deterministic dropout in tests.


allennlp/models/vision_text_model.py
```diff
@@ -106,42 +106,33 @@ def from_huggingface_model_name(
...
 
+            text_embeddings = TransformerEmbeddings.from_pretrained_module(
+                transformer.embeddings, output_size=config.hidden_dim
+            )
+
             from transformers.models.albert.modeling_albert import AlbertModel
 
             if isinstance(transformer, AlbertModel):
-                linear_transform = deepcopy(transformer.encoder.embedding_hidden_mapping_in)
+                text_embeddings.linear_transform = deepcopy(
```
dirkgr, line 122: Usually I don't pay a lot of attention to this, but Codecov tells me that this line is not covered by tests. It seems somewhat important that the Albert case be covered in tests, since it is so different from the other transformers.

allennlp/modules/transformer/transformer_embeddings.py
```diff
@@ -7,132 +7,153 @@
...
-        token_type_embeddings = torch.nn.Embedding(type_vocab_size, hidden_size)
-
-        embeddings = [word_embeddings, position_embeddings, token_type_embeddings]
-
-        self.embeddings = Embeddings(embeddings, hidden_size, dropout)
-
-    def forward(
+        word_embeddings = torch.nn.Embedding(vocab_size, embedding_size, padding_idx=pad_token_id)
+        position_embeddings = torch.nn.Embedding(max_position_embeddings, embedding_size)
+        token_type_embeddings = torch.nn.Embedding(type_vocab_size, embedding_size)
```
dirkgr, line 124: This works OK if `type_vocab_size` is 0? Not all transformers have types.

tests/modules/transformer/transformer_embeddings_test.py
```diff
@@ -1,50 +1,167 @@
...
+
+    torch.manual_seed(1234)
+    hf_module = BertEmbeddings(BertConfig(**params))
+    modules["bert"] = hf_module
+
+    albertparams = copy.deepcopy(params_dict)
+    albertparams["hidden_dropout_prob"] = albertparams.pop("dropout")
+
+    torch.manual_seed(1234)
+    hf_module = AlbertEmbeddings(AlbertConfig(**albertparams))
```
dirkgr, line 49: Looks like it's tested after all, so Codecov is just wrong?

If that's the case we should probably turn it off.

dirkgr (approved):
Looks good, just one concern about empty type embeddings, and a likely spurious issue with code coverage.

dirkgr:
That's a lot of tests!

allennlp/modules/transformer/transformer_embeddings.py
AkshitaB: It works now.

Issue #4904, opened 2021-01-08
Prepare data loading for 2.0


epwalsh:
- [x] Remove `PyTorchDataLoader`. I wouldn't recommend anyone use this, and it doesn't provide any benefit over `MultiProcessDataLoader`. (https://github.com/allenai/allennlp/pull/4907)
- [x] Make submodule names consistent. Currently we have `multi_process_data_loader.py` and `multitask_*.py`. We should either rename `multi_process_*` to `multiprocess_*` or rename `multitask_*` to `multi_task_*`. (https://github.com/allenai/allennlp/pull/4906)
- [x] `data_path` param should accept `PathLike`. (https://github.com/allenai/allennlp/pull/4908)
- [x] Do more extensive testing on `MultiProcessDataLoader`. Potentially improve error handling. (https://github.com/allenai/allennlp/pull/4912)
- [x] Document best practices for `MultiProcessDataLoader`. (https://github.com/allenai/allennlp/pull/4909)
- [ ] ~~❓Improve where `TokenIndexers` fit it. Maybe they are part of `DatasetReader`. This would actual require minimal changes.~~
- [x] Update `allennlp-models@vision` for these changes. (https://github.com/allenai/allennlp-models/pull/192)

dirkgr:
@epwalsh, is this done?

PR #4906, opened 2021-01-12
rename 'multi_process_*' -> 'multiprocess' for consistency

epwalsh:
See #4904.

```
find ./ -type f -exec sed -i '' 's/multi_process/multiprocess/g' {} \;
```

Requires a small change in `allennlp-models` for CI to pass.


dirkgr (approved):
This is definitely something we want before the 2.0 release :-)

PR #4907, opened 2021-01-12
remove PyTorchDataLoader, add SimpleDataLoader for testing

epwalsh:
See #4904.

- Removes `PyTorchDataLoader`
- Adds a very simple `SimpleDataLoader` which is convenient for testing

Requires a small fix in `allennlp-models` for CI to pass.


allennlp/data/data_loaders/simple_data_loader.py
```diff
@@ -0,0 +1,91 @@
...
+        self.instances = instances
+        self.batch_size = batch_size
+        self.shuffle = shuffle
+        self.batches_per_epoch = batches_per_epoch
+        self.vocab = vocab
+        self.cuda_device: Optional[torch.device] = None
+        self._batch_generator: Optional[Iterator[TensorDict]] = None
+
+    def __len__(self) -> int:
+        return len(list(lazy_groups_of(self.instances, self.batch_size)))
```
dirkgr, line 39: Ouch. So inefficient. Why not `ceil(len(self.instances) / self.batch_size)`?

dirkgr (approved):

allennlp/data/data_loaders/simple_data_loader.py
epwalsh: That's better. I was being lazy 🙃

PR #4908, opened 2021-01-12
MultiProcessDataLoader takes PathLike data_path

epwalsh:
See #4904.

- Changes `data_path` type to `Union[str, os.PathLike]` in `MultiProcessDataLoader`.


dirkgr (approved):
This is an improvement, but we have started putting all sorts of stuff into `data_path` lately, including dicts and lists. We'll have to think about this seriously at some point.

PR #4909, opened 2021-01-12
improve data loading docs

epwalsh:
See  #4904.

- Small fixes to data loading docs.
- Adding "Best practices" and "Common issues" sections to `MultiProcessDataLoader` docs.
- Added a method `get_batch_size(self) -> Optional[int]` to `BatchSampler` class. This helps `MultiProcessDataLoader` make a better guess for the queue size.


allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -75,7 +75,7 @@ class MultiProcessDataLoader(DataLoader):
         to `__iter__()`.
 
     num_workers: `int`, optional (default = `0`)
-        The number of workers to use to read `Instance`s in parallel.
+        The number of workers to use to read `Instances` in parallel.
```
dirkgr, line 78: That was correct before, no? `Instance` is the type, so `Instance`s is many of them?

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -84,13 +84,20 @@ class MultiProcessDataLoader(DataLoader):
...
+            [Common issues](#multiprocessdataloader.common_issues) sections below before using this option.
+
     max_instances_in_memory: `int`, optional (default = `None`)
         If not specified, all instances will be read and cached in memory for the duration
         of the data loader's life. This is generally ideal when your data can fit in memory
         during training. However, when your datasets are too big, using this option
         will turn on lazy loading, where only `max_instances_in_memory` instances are processed
         at a time.
 
+        When using this option you should generally set it to a multiple of `batch_size`.
```
dirkgr, line 99: Does that really matter? Isn't it just the size of the `pool` during shuffling?

allennlp/data/data_loaders/multi_process_data_loader.py
```diff
@@ -123,14 +168,23 @@ class MultiProcessDataLoader(DataLoader):
...
+        Luckily this is simple to debug and simple to fix.
+
+        First, to verify that this is your issue just watch your shared memory as your data loader runs.
+        For example, run `watch -n 0.3 'df -h | grep shm'`.
+
+        If you're seeing your shared memory blow up until it maxes-out, then you either need to decrease
+        `max_instances_in_memory` or increase your system's `ulimit`.
+
+        If you're using Docker, you can increase the shared memory available on a container by running
+        it with the option `--ipc=host` or by setting `--shm-size`.
```
dirkgr, line 187: This is some heavy stuff to keep in your head :-/

dirkgr (approved):

allennlp/data/data_loaders/multi_process_data_loader.py
epwalsh: It matters when you're using a `BatchSampler`. Otherwise you're right, it's just the size of the pool used in `shuffle_iterable`. I should clarify.

allennlp/data/data_loaders/multi_process_data_loader.py
epwalsh: That's why I wrote it down 😜

allennlp/data/data_loaders/multi_process_data_loader.py
dirkgr: PpV0evc

allennlp/data/data_loaders/multi_process_data_loader.py
dirkgr: Well, that's better than not having it written down.

But our lofty ideal should be that data loading is something a researcher never even thinks about. It just does the right thing automatically. 

I'm not sure we can get there with the tools we have right now (`TOKENIZERS_PARALLELISM` what?), and this is definitely a step up. I just don't like that data loading is so complicated in general.

allennlp/data/data_loaders/multi_process_data_loader.py
dirkgr: So it matters for how well shuffled your data is. If you set this to 1x batch size, you get no shuffle at all.

PR #4913, opened 2021-01-14
Log to TensorBoard through a TrainerCallback in GradientDescentTrainer

epwalsh:
Finishes the work started in #4770.

- You can now completely disable tensorboard logging with the `GradientDescentTrainer` by setting `"tensorboard_writer": null` in your config.


dirkgr (commented):
I guess that's better, but the real solution is to make the tensor board writer a callback. Any chance we can do that instead?

epwalsh:
@dirkgr I'm not sure we could easily duplicate this logic with a callback:

https://github.com/allenai/allennlp/blob/8b95316bc8d7c717dccc2d72fb19d27870ee3fcf/allennlp/training/trainer.py#L646-L669

dirkgr:
It could be done. The callback needs to work across batches. If we want to do this every 10 batches, then on batch 9 we make a copy of all the parameters, and on batch 10 we compare to the old parameters to the new ones.

Basically, we need the existing `should_log_histograms_this_batch()`, and also a `should_log_histograms_next_batch()`, so we can get ready.

epwalsh:
@dirkgr this is ready for another review. The models CI will fail until I make one fix there.

allennlp/training/trainer.py
```diff
@@ -191,15 +213,83 @@ def on_batch(
...
+            # GPU/CPU copies (necessary for large models, and for shipping to tensorboard), so
+            # we don't do this every batch, only when it's requested.
+            self._param_updates = {
+                name: param.detach().cpu().clone()
+                for name, param in trainer.model.named_parameters()
+            }
+        elif self._tensorboard.should_log_histograms_this_batch():
+            assert self._param_updates is not None
+            for name, param in trainer.model.named_parameters():
+                self._param_updates[name].sub_(param.detach().cpu())
```
dirkgr, line 240: These conditions could both be true at the same time, if we log histograms every batch. So `this_batch` needs to be processed first, and the `if` statements need to be independent.

dirkgr (changes requested):
This is good stuff, a much cleaner API for the trainer, and I love the number of lines reduction as well.

allennlp/training/trainer.py
epwalsh: Good catch. Fixed

dirkgr (approved):

PR #4914, opened 2021-01-14
Toolkit decoder

AkshitaB:

- `TransformerBlock` --> `TransformerStack` (to avoid issues with models like distilbert, which use `TransformerBlock` for something else.
- Adding cross-attention option to `TransformerLayer`.
- `TransformerStack` can be initialized with pre-built `TransformerLayer` object.


allennlp/modules/transformer/transformer_layer.py
```diff
@@ -102,19 +107,25 @@ class TransformerLayer(TransformerModule, FromParams):
...
         Dropout probability for the `SelfAttention` layer.
-    hidden_dropout: `float` (default = `0.0`)
+    hidden_dropout : `float` (default = `0.0`)
         Dropout probability for the `OutputLayer`.
-    activation: `Union[str, torch.nn.Module]`
-
+    activation : `Union[str, torch.nn.Module]`
+    add_cross_attention : `bool` (default = `False`)
+        If True, an extra `AttentionLayer` is added for cross-attention.
+        This is helpful when using the layer in a decoder.
```
dirkgr, line 120: Where is this added? What does it do?

dirkgr (approved):
Code seems fine, but I'm not sure what the extra cross attention is for. Can you add an explanation?

allennlp/modules/transformer/transformer_layer.py
AkshitaB: It's added in `__init__()`. The idea is to be able to use the same layer for encoders and decoders (where you want self attention as well as source attention).

dirkgr: I don't have the background to understand this. What I meant was, where in the model architecture is it added? How does it help in an encoder/decoder situation?

AkshitaB: Well, you set it to `True` when you need to use the `TransformerStack` as a decoder, and it will add a cross-attention layer. The `forward` will know what to do with it. Following what is done in the original transformer paper - 

> "In addition to the two sub-layers in each encoder layer, the decoder inserts a third sub-layer, which performs multi-head attention over the output of the encoder stack".

Issue #4826, opened 2020-11-30
TrackEpochCallback is not an EpochCallback


mahnerak:
Is there any reason why `TrackEpochCallback` should not inherit from `EpochCallback`?
https://github.com/allenai/allennlp/blob/5b30658514a00e11000e648fec23be11a998bd92/allennlp/training/trainer.py#L179-L188

epwalsh:
No, this was an oversight. Thanks for catching it. Would you like to make a PR to fix?

mahnerak:
Sure! I'll make a PR.

epwalsh:
Hey @mahnerak are you still planning on making a PR for this? If not I can do it.

mahnerak:
Yes, I do. I was just preparing a couple of other notes on Callbacks and wanted to submit them as a single PR.

mahnerak:
@epwalsh The `end` callbacks are called within the `try` block.
https://github.com/allenai/allennlp/blob/1ca478ab674dd412447b2f11c820ca5664737cc2/allennlp/training/trainer.py#L1111
However, I was expecting `end` callbacks to be called even in case of an error. Or, IDK, maybe there should be another way to inform the callback about the end of the training (so it can do a cleanup).

mahnerak:
As there are no many third-party callbacks implemented for AllenNLP (yet) and the backwards-compatibility is not an issue right now, I think it's better to revisit the callback API.
For example, we may need `on_start()` callback instead of firing `on_epoch(epoch=-1)`. 

Also, I don't get why there should be separate `BatchCallback`, `EndCallback`, `EpochCallback`, and a metaclass `TrainCallback` to rule them all. I believe you had your reasons to implement it this way. But why not to have a single Callback class with `__init__(self, serialization_dir, trainer)`, `on_start()`, `on_batch_start()`, `on_batch_end()`, `on_epoch_start()`, `on_epoch_end()`, `on_end()`, `on_error()`, `on_after_backward()`, etc...?
If someone wants to have a callback just for epochs, they can implement only a single method `on_epoch_end()`.

I just feel the API of callbacks is the only non-intuitive abstraction in AllenNLP.
Pytorch-lightning has much straightforward callback interface.

epwalsh:
[...]

PR #4917, opened 2021-01-19
Generic Callbacks

mahnerak:
This is a draft for generic callbacks to handle multiple types of events.
The volume of code changed is big, so I fully understand you not rushing to merge this into master.
However, this change introduces several breaking changes that may be best to have for v2.0 (discussed at #4826).

- Removed `EpochCallback`, `BatchCallback` in favour of `TrainerCallback`. The metaclass-wrapping implementation is removed as well.
- The `TrainerCallack.on_epoch()` does not fire with `epoch=-1` at the start of the training. Instead, method `TrainerCallback.on_start()` should be used for these cases.
- Now Trainer always fires `TrainerCallback.on_end()` so all the resources can be cleaned up properly.

- The `TrainerCallback` constructor accepts `serialization_dir` provided by `Trainer`. This can be useful for `Logger` callbacks those need to store files in the run directory.
- The `TrainerCallback` event methods now accept `**kwargs`. This may be useful to maintain backwards-compatibility of callbacks easier in the future. E.g. we may decide to pass the exception/traceback object in case of failure to `on_end()` and the older callbacks may simply ignore the argument instead of raising a `TypeError`.
- `TensorBoardBatchMemoryUsage` is converted from `BatchCallback` into `TrainerCallback`. `TrackEpochCallback` is converted from `EpochCallback` into `TrainerCallback`.
- `Trainer` can accept callbacks simply with the name `callbacks` instead of `trainer_callbacks`.
- Fixed the misspelling, changed `TensoboardBatchMemoryUsage` to `TensorBoardBatchMemoryUsage`.
- We set a value to `epoch` so in case of firing `TrainerCallback.on_end()` the variable is bound.
  This could have lead to an error in case of trying to recover a run after it was finished training, and at least one end callback is enabled.


epwalsh:
Hey @mahnerak, thanks for drafting this so quickly! Could you actually change the base branch to `vision` (instead of `main`)? `vision` is our target branch for 2.0 changes until we merge it into `main`, and there have been some commits on that branch that might slightly affect your PR.

mahnerak:
This is what we have in the YerevaNN's internal version of `allennlp` now.
It looks like I have to invest some time modifying/adding tests.

dirkgr:
> Fixed the misspelling, changed `TensoboardBatchMemoryUsage` to `TensorBoardBatchMemoryUsage`.

🙏

CHANGELOG.md
```diff
@@ -74,18 +74,35 @@ dataset at every epoch) and a `MultiTaskScheduler` (for ordering the instances w
   This makes it possible to use the `FileLock` class on a read-only file system.
 - Added a new learning rate scheduler: `CombinedLearningRateScheduler`. This can be used to combine different LR schedulers, using one after the other.
 - Moving `ModelCard` and `TaskCard` abstractions into the main repository.
+- The `TrainerCallback` constructor accepts `serialization_dir` provided by `Trainer`. This can be useeful for `Logger` callbacks those need to store files in the run directory.
```
dirkgr, line 77: typo, "useful"

dirkgr (commented):
This is quite good. Can we trivially make this backwards compatible by writing `BatchCallback` and `EpochCallback` base classes that are sub-classes of `TrainerCallback`?

dirkgr:
> Can we trivially make this backwards compatible by writing `BatchCallback` and `EpochCallback` base classes that are sub-classes of `TrainerCallback`?

I tried to do this, and decided it's not worth it. The fact that we have a name conflict with old `TrainerCallback` and new `TrainerCallback` convinced me that version 2.0 is a great opportunity to break old APIs.

dirkgr:
We couldn't wait any longer to release `v2.0.0rc1`, but I'll still try to get it into `v2.0.0`.

mahnerak:
@dirkgr  Sorry, I was super busy with ACL deadlines.
Trying to finalize the PR right now.

allennlp/training/trainer.py
```diff
@@ -1105,9 +1014,8 @@ def _try_train(self) -> Dict[str, Any]:
                 logger.info("Estimated training time remaining: %s", formatted_time)
 
             epochs_trained += 1
-
-        for callback in self._end_callbacks:
-            callback(self, metrics=metrics, epoch=epoch, is_primary=self._primary)
+        else:
+            epoch = self._num_epochs - 1
```
dirkgr, line 1018: `for ... else` ... a real rarity. This fixes the situation where the training had already finished, was checkpointed, but something failed during validation/test, and you want to run again, I assume? Looks quite reasonable, but I wanted to make sure it's on purpose.

dirkgr (approved):
I love it, thanks for bringing this in! I'll leave this open for a few hours so you have a chance to see my question about the `for ... else` construct, but it looks ready to go to me!

allennlp/training/trainer.py
mahnerak: Consider you want to `--recover` the training, but it's already done, so you basically run `for epoch in range(epoch_counter, self._num_epochs):` where `epoch_counter == self._num_epochs`. This means that the loop will not run at all.
The `epoch` parameter, in that case, will be unbound.

Minimal example will be:
```python
In [1]: for i in range(10, 10):
   ...:     pass
   ...:
   ...: print(i)
```
This will output:
```
NameError                                 Traceback (most recent call last)
<ipython-input-1-58b7096fed5d> in <module>
      2     pass
      3
----> 4 print(i)

NameError: name 'i' is not defined
```

dirkgr: Yep, I get it. I just wanted to make sure it's intentional, because most people don't even know that `for ... else` exists :-)

PR #4928, opened 2021-01-22
Consistent metric tracker

dirkgr:
Makes the metric tracker's `state_dict` handling consistent.

Neither `patience` nor `tracked_metrics` should be in the `state_dict`, because those come from the constructor. Also, `best_epoch_metrics` should be saved and restored.

@mahnerak

### DIFF
--- a/allennlp/training/metric_tracker.py
+++ b/allennlp/training/metric_tracker.py
@@ -83,7 +83,10 @@ class MetricTracker:
         self._is_best_so_far = state_dict["is_best_so_far"]
         self._epoch_number = state_dict["epoch_number"]
         self.best_epoch = state_dict["best_epoch"]
-        self.best_epoch_metrics = state_dict["best_epoch_metrics"]
+        
+        # Even though we don't promise backwards compatibility for the --recover flag,
+        # it's particularly easy and harmless to provide it here, so we do it.
+        self.best_epoch_metrics = state_dict.get("best_epoch_metrics", {})
 
     def add_metrics(self, metrics: Dict[str, float]) -> None:
         """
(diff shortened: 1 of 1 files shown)


allennlp/training/metric_tracker.py
```diff
@@ -59,32 +59,31 @@ def clear(self) -> None:
...
         A `Trainer` can use this to hydrate a metric tracker from a serialized state.
         """
         self._best_so_far = state_dict["best_so_far"]
-        self._patience = state_dict["patience"]
         self._epochs_with_no_improvement = state_dict["epochs_with_no_improvement"]
         self._is_best_so_far = state_dict["is_best_so_far"]
-        self.best_epoch_metrics = state_dict["best_epoch_metrics"]
         self._epoch_number = state_dict["epoch_number"]
         self.best_epoch = state_dict["best_epoch"]
+        self.best_epoch_metrics = state_dict["best_epoch_metrics"]
```
mahnerak, line 86: What if we try to `--recover` a run originally trained with previous versions? Do we need some kind of check here?

dirkgr: That's not a supported scenario, though in this particular case it would be totally harmless to add a check.

mahnerak: This is something that will be difficult to fix for a user. Even if it will only appear in v2.0, we may want backwards-compatibility here.

dirkgr: Added in 764ce61.

mahnerak: One more question: Is metric_tracker loaded when we run `allennlp evaluate`?

dirkgr: No, only the trainer uses it.

epwalsh (approved):
LGTM

Issue #4868, opened 2020-12-16
Make sure tensors have dimensions multiple of 8


mahnerak:
The `torch.amp` automatically enabled mixed-precision training but on the latest GPUs, it's faster (according to the documentation) because of Tensor Cores. But latter is enabled (almost) for fp16 calculations if the tensor dimensions are multiple of 8.

The embedding size and the batch size is easy to maintain as multiples of 8 manually.
However, we may need to modify `TokenIndexer` to pad it automatically, similar to: https://github.com/YerevaNN/allennlp/commit/374acec5e62d6d74586b18a3d5bb2f9b5a169da4

We also may need to modify `max_tokens`-based samplers to have 8x batch_sizes.

epwalsh:
Thanks for pointing that out @mahnerak! This is pretty important so I'm putting it on our 1.4 release milestone. We're all tied up with other stuff at the moment though, so we might not be able to get to this until the new year, but I'd be happy to review a PR if you or anyone else is interesting.

For anyone else coming upon this, NVIDIA provides some more info here: https://developer.nvidia.com/blog/optimizing-gpu-performance-tensor-cores/

dirkgr:
We don't really need the issue as well as the draft PR.

PR #4929, opened 2021-01-22
Pad tokens to make number of tokens 8x

mahnerak:
This addresses issue #4868.

The torch.amp automatically enabled mixed-precision training but on the latest GPUs, it's faster (according to the documentation) because of Tensor Cores. But latter is enabled (almost) for fp16 calculations if the tensor dimensions are multiple of 8.
Reference[1]: https://developer.nvidia.com/blog/optimizing-gpu-performance-tensor-cores/

The embedding size and the batch size is easy to maintain as multiples of 8 manually. This PR makes the number of tokens multiple of any given number.
The only thing to do is to provide extra `{"pad_to_multiple_of": 8}` to the `token_indexer` config.

P.S. I did not observe a significant boost in performance. According to Figure 3 of [1] it sometimes can work fast even for tensors with axis sizes non-divisible by 8, depends on the architecture.


epwalsh:
@mahnerak, what kind of GPU did you benchmark this on?

Even if we don't observe a speed up on the GPUs we have at hand, I think it's a good addition to the library given that there should be an improvement on *some* GPUs, in theory at least.

epwalsh:
And as you mentioned in https://github.com/allenai/allennlp/issues/4868, we should also have this feature for the `MaxTokensBatchSampler`.

mahnerak:
@epwalsh I've been testing it on Titan V gpus.

I'll make a comparison table before making this draft ready for reviews, on Titan V, RTX3080 and GTX1080Ti (the latter doesn't have Tensor Cores).

epwalsh:
Cool. Well we also have access to A100s so I could potentially run benchmarks on one of those.

epwalsh:
What type of model are you using to benchmark?

The [NVIDIA post](https://developer.nvidia.com/blog/optimizing-gpu-performance-tensor-cores/) suggests that it could be important for the vocab size to be multiple of 8 as well 😒 That would be harder to control.

dirkgr:
Any progress on these? I'd be interested in having this if it makes a measurable difference.

JohnGiorgi:
This would also be useful for certain transformer models that use block sparse attention, like BigBird, as they require the sequence length to be divisible by the block size (see https://huggingface.co/transformers/model_doc/bigbird.html).

PR #4945, opened 2021-01-29
Sanitize set

epwalsh:

- `common.util.sanitize()` didn't handle `set` objects before. Now it will.


dirkgr (approved):
I'm a bit unhappy about the fact that we make sets into lists, but we already do that with tuples, so I guess that's how this function works.

Issue #4690, opened 2020-10-01
Train a model with transformer embeddings and additional_special_tokens


pvcastro:
## Checklist

- [x] I have verified that the issue exists against the `master` branch of AllenNLP.
- [x] I have read the relevant section in the [contribution guide](https://github.com/allenai/allennlp/blob/master/CONTRIBUTING.md#bug-fixes-and-new-features) on reporting bugs.
- [x] I have checked the [issues list](https://github.com/allenai/allennlp/issues) for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/allenai/allennlp/pulls) for existing proposed fixes.
- [x] I have checked the [CHANGELOG](https://github.com/allenai/allennlp/blob/master/CHANGELOG.md) and the [commit log](https://github.com/allenai/allennlp/commits/master) to find out if the bug was already fixed in the master branch.
- [x] I have included in the "Description" section below a traceback from any exceptions related to this bug.
- [x] I have included in the "Related issues or possible duplicates" section beloew all related issues and possible duplicate issues (If there are none, check this box anyway).
- [x] I have included in the "Environment" section below the name of the operating system and Python version that I was using when I discovered this bug.
- [x] I have included in the "Environment" section below the output of `pip freeze`.
- [x] I have included in the "Steps to reproduce" section below a minimally reproducible example.

## Description

Hi there! I'm trying to train a transformer-based text classifier model in AllenNLP, but I need to add 5 additional special tokens, in a way compatible with tokenizers lib. I tried adding them to the jsonnet AllenNLP config file and then to the transformer's model path, but neither worked, with each approach having a different problem, which will be described below.

<details>
<summary><b>Python traceback:</b></summary>
<p>

```
2020-09-30 23:56:17,398 - INFO - allennlp.training.trainer - Epoch 0/9
2020-09-30 23:56:17,398 - INFO - allennlp.training.trainer - Worker 0 memory usage MB: 10065.304
2020-09-30 23:56:17,484 - WARNING - allennlp.common.util - unable to check gpu_memory_mb() due to occasional failure, continuing
Traceback (most recent call last):
  File "/media/discoD/repositorios/allennlp/allennlp/common/util.py", line 415, in gpu_memory_mb
    encoding="utf-8",
  File "/media/discoD/anaconda3/envs/allennlp/lib/python3.7/subprocess.py", line 411, in check_output
    **kwargs).stdout
[...]

PR #4946, opened 2021-01-29
Resize transformers word embeddings layer for additional_special_tokens

pvcastro:
Fixes part of #4690 

- I opened issue #4690 a while ago for AllenNLP to provide `additional_special_tokens` parameter for transformers models. @dirkgr already provided a fix for the parameter to be added and worked very well, but we were missing some way for the model to adapt it's word embeddings layer to the additional tokens provided. I'm not sure if this is the best way to do it, but this PR seems to solve this.

### DIFF
--- a/allennlp/modules/token_embedders/pretrained_transformer_embedder.py
+++ b/allennlp/modules/token_embedders/pretrained_transformer_embedder.py
@@ -1,3 +1,4 @@
+import logging
 import math
 from typing import Optional, Tuple, Dict, Any
 
@@ -11,6 +12,8 @@ from allennlp.data.tokenizers import PretrainedTransformerTokenizer
 from allennlp.modules.scalar_mix import ScalarMix
 from allennlp.modules.token_embedders.token_embedder import TokenEmbedder
 from allennlp.nn.util import batched_index_select
+
+logger = logging.getLogger(__name__)
 
 
 @TokenEmbedder.register("pretrained_transformer")
@@ -103,10 +106,12 @@ class PretrainedTransformerEmbedder(TokenEmbedder):
         )
 
         try:
-            self.transformer_model.resize_token_embeddings(len(tokenizer.tokenizer))
+            if self.transformer_model.get_input_embeddings().num_embeddings != len(tokenizer.tokenizer):
+                self.transformer_model.resize_token_embeddings(len(tokenizer.tokenizer))
         except NotImplementedError:
             # Can't resize for transformers models that don't implement base_model.get_input_embeddings()
-            pass
+            logger.warning("Could not resize the token embedding matrix of the transformer model. "
+                           "This model does not support resizing.")
 
         self._num_added_start_tokens = len(tokenizer.single_sequence_start_tokens)
         self._num_added_end_tokens = len(tokenizer.single_sequence_end_tokens)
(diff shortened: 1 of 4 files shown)


allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -101,6 +101,13 @@ def __init__(
             model_name,
             tokenizer_kwargs=tokenizer_kwargs,
         )
+
+        try:
+            self.transformer_model.resize_token_embeddings(len(tokenizer.tokenizer))
```
dirkgr, line 106: I'm assuming this is a no-op if the size already matches?

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
```diff
@@ -101,6 +101,13 @@ def __init__(
...
+        except NotImplementedError:
+            # Can't resize for transformers models that don't implement base_model.get_input_embeddings()
+            pass
```
dirkgr, line 109: Is this completely safe? Maybe we should at least print a warning, because you have a good chance of running into trouble later, right?

dirkgr (commented):
I like this; I have just one question.

dirkgr:
Out of curiosity, is there an API to increase the token type id vocab as well?

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
pvcastro: Yes, nothing would happen if no additional tokens are added.

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
pvcastro: One of the unit tests in `pretrained_transformer_embedder_test.py` failed (`test_encoder_decoder_model`). Transformers lib call a `base_model.get_input_embeddings()`as part of the resize, but the actual base_model in an Encoder-Decoder model is the actual Encoder (BartEncoder, in the mentioned unit test), and they don't implement this:
https://github.com/huggingface/transformers/blob/1420b5ff675ccdc3296c6776b339a08a22d2e941/src/transformers/modeling_utils.py#L451. 
We could add a warning here, but I'm not sure which exact condition could be used here to trigger the warning. I _think_ that this exception could happen for other Encoder-Decoder models, but I'm not sure about all of them (nor I know them exaustively). I wouldn't be aware of this if it wasn't for the unit test that failed. 
What do you think, what do you propose?

pvcastro:
> Out of curiosity, is there an API to increase the token type id vocab as well?

I don't know about token type ids, I don't know the exact logic these models have for setting them, do you @dirkgr ?

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
dirkgr: I would propose just a `logger.warning("Could not resize the token embedding matrix of the transformer model. This model does not support resizing.")`. But make sure to only print that message if the sizes actually differ!

If we're lucky, it will save someone hours down the line when they try to use this with BART and fail.

dirkgr:
You set token type IDs by specifying another tensor in the input. Annoyingly, not all transformers have token type IDs, so AllenNLP gets up to some shenanigans to make it work anyways for most cases.

dirkgr (approved):
Thank you!

allennlp/modules/token_embedders/pretrained_transformer_embedder.py
pvcastro: OK, got it here. I thought there wouldn't be a single method for checking this, but there is...only the resizing that it seems to not being possible for encoder-decoders.

dirkgr (approved):

PR #4956, opened 2021-02-02
fix bug with MultiProcessDataLoader

epwalsh:
Fixes a bug in `MultiProcessDataLoader` that causes a crash when using lazy loader with workers when `start_method = "spawn"` and `batches_per_epoch` is set.

The bug happens because generator objects cannot be pickled. And with those settings in the `MultiProcessDataLoader`, a generator would be created and assigned to `self._batch_generator` before spawning workers, and since spawning uses pickle, this would fail.

### DIFF
--- a/allennlp/data/data_loaders/multiprocess_data_loader.py
+++ b/allennlp/data/data_loaders/multiprocess_data_loader.py
@@ -322,14 +322,19 @@ class MultiProcessDataLoader(DataLoader):
         if self.batches_per_epoch is None:
             yield from self._iter_batches()
         else:
-            if self._batch_generator is None:
-                self._batch_generator = self._iter_batches()
+            if self._batch_generator is not None:
+                batch_generator = self._batch_generator
+                # Can't have a pointer to this in `self` when we try to spawn workers.
+                self._batch_generator = None
+            else:
+                batch_generator = self._iter_batches()
             for i in range(self.batches_per_epoch):
                 try:
-                    yield next(self._batch_generator)
-                except StopIteration:  # data_generator is exhausted
-                    self._batch_generator = self._iter_batches()  # so refresh it
-                    yield next(self._batch_generator)
+                    yield next(batch_generator)
+                except StopIteration:  # batch_generator is exhausted
+                    batch_generator = self._iter_batches()  # so refresh it
+                    yield next(batch_generator)
+            self._batch_generator = batch_generator
 
     @overrides
     def iter_instances(self) -> Iterator[Instance]:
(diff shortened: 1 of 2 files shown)


allennlp/data/data_loaders/multiprocess_data_loader.py
```diff
@@ -322,14 +322,19 @@ def __iter__(self) -> Iterator[TensorDict]:
...
                 try:
-                    yield next(self._batch_generator)
-                except StopIteration:  # data_generator is exhausted
-                    self._batch_generator = self._iter_batches()  # so refresh it
-                    yield next(self._batch_generator)
+                    yield next(batch_generator)
+                except StopIteration:  # batch_generator is exhausted
+                    batch_generator = self._iter_batches()  # so refresh it
+                    yield next(batch_generator)
+            self._batch_generator = batch_generator
```
dirkgr, line 337: That's quite an ugly hack. But I think it's OK if you wrap this in a try/finally that will set `self._batch_generator` even if we bail out with an exception.

dirkgr (changes requested):

allennlp/data/data_loaders/multiprocess_data_loader.py
epwalsh: If there's an exception from `_batch_generator`, then `_batch_generator` will be broken in that it will always raise a `StopIteration` on subsequent calls to `next()`. So there's no reason to keep it around.

dirkgr: That's if the error is something you foresee now. What if it's something weird like `self.batches_per_epoch` isn't an `int`? You're still screwed, but it would be extra unexpected if then `self._batch_generator` is suddenly gone as well.

epwalsh: `batches_per_epoch` not being an `int` is literally the only possible error that could happen which would **not** result in a broken `batch_generator`, because it's the only error that could arise outside of `batch_generator`.

But if that error did occur, then `_batch_generator` would be an "uninitialized" generator at that point since `next()` hasn't been called yet. It's also a private attribute that users shouldn't know about, and is only used internally in this one method.

So I don't see the point of keeping the reference to it in the advent of this one possible error that could only happen if someone gave a non int value to `batches_per_epoch`, which, by the way, is something we validate anyway in `__init__()` (https://github.com/allenai/allennlp/pull/4956/files#diff-7afd5388fe687e8ec4b9e85a4d5eaa9ca122c78c8a70b457ae887f1b208ce81fL223)?

I don't think the extra indentation is worth it.

dirkgr: I'm more worried about the shape the code could take in the future. The next person to refactor this might not know what's going on and do something unsafe.

But either way, it's just a recommendation.

dirkgr (approved):

PR #4958, opened 2021-02-03
Fix empty vocab bug when multi-process dataloder

### LINKED ISSUE #4937: Empty vocab when use multiprocess dataloader
I use a [BART](https://github.com/allenai/allennlp-models/blob/main/training_config/generation/bart_cnn_dm.jsonnet) like config. And I got a Key Error when convert idx to token, cause I got an empty vocab in [here]( https://github.com/allenai/allennlp-models/blob/main/allennlp_models/generation/models/bart.py#L369).
```
ipdb> self.vocab
Vocabulary with namespaces:  Non Padded Namespaces: {'*tags', '*labels'}
```
But when I set `num_workers=0` , all things go well.
```
ipdb> self.vocab
Vocabulary with namespaces:  tokens, Size: 21130 || Non Padded Namespaces: {'*tags', '*labels'}
```

wlhgtc:
Fix #4937 
I don't know where should I put the `copy_transformer_vocab` function ,`tools` maybe not  a good choice...
And more importantly, I use just dict rather than `Vocabulary` to pass result to avoid cycle import.
@epwalsh


allennlp/data/vocabulary.py
```diff
@@ -259,6 +260,19 @@ def __init__(
...
+    @classmethod
+    def from_pretrained_transformer(
+        cls, model_name: str, namespace: str = "tokens"
+    ) -> "Vocabulary":
+        vocab = cls.empty()
+        from allennlp.common import cached_transformers
+
+        tokenizer = cached_transformers.get_tokenizer(model_name)
+        result = copy_transformer_vocab(tokenizer)
+        vocab._token_to_index[namespace] = result["token_to_index"]
```
wlhgtc, line 272: @epwalsh  I don't know whether  it's safe to replace the whole vocab with a `dict`

epwalsh (changes requested):
Hey @wlhgtc, thanks for this and sorry I've been slow to respond.

Instead of having the `copy_transformer_vocab` function, I think it would be better to just have a method on the `Vocabulary` class like this:

```python
class Vocabulary:
    # ...
    def add_transformer_vocab(self, model_name: str, namespace: str) -> None:
        # ... basically the same logic that's in `copy_transformer_vocab`
```

Then the `PretrainedTransformerIndexer` can just call this method. And `Vocabulary.from_pretrained_transformer` could just look like this:

```python
@classmethod
def from_pretrained_transformer(cls, model_name: str, namespace: str = "tokens") -> "Vocabulary":
    vocab = cls.empty()
    vocab.add_transformer_vocab(model_name, namespace)
    return vocab
```

wlhgtc:
@epwalsh , this do seems better, but small question about parameters in `add_transformer_vocab`:

Cause we run this function in `PretrainedTransformerIndexer`, it already has a `tokenizer`. It's wired to initialize a new one by `model_name`, you could see code in [here](https://github.com/allenai/allennlp/blob/main/allennlp/data/token_indexers/pretrained_transformer_indexer.py#L59). Maybe I should keep the same parameter like:
```python
    @staticmethod
    def add_transformer_vocab(tokenizer: PreTrainedTokenizer) -> Dict[str, Dict]:

```

epwalsh:
> @epwalsh , this do seems better, but small question about parameters in `add_transformer_vocab`:
> 
> Cause we run this function in `PretrainedTransformerIndexer`, it already has a `tokenizer`. It's wired to initialize a new one by `model_name`, you could see code in [here](https://github.com/allenai/allennlp/blob/main/allennlp/data/token_indexers/pretrained_transformer_indexer.py#L59). Maybe I should keep the same parameter like:
> 
> ```python
>     @staticmethod
>     def add_transformer_vocab(tokenizer: PreTrainedTokenizer) -> Dict[str, Dict]:
> ```

Sure, that sounds okay.

allennlp/data/vocabulary.py
```diff
@@ -417,6 +430,26 @@ def empty(cls) -> "Vocabulary":
         """
         return cls()
 
+    @staticmethod
+    def add_transformer_vocab(tokenizer: PreTrainedTokenizer) -> Dict[str, Dict]:
```
epwalsh, line 434: I meant that this should be an instance method:

```suggestion
    def add_transformer_vocab(self, tokenizer: PreTrainedTokenizer, namespace: str = "tokens") -> None:
```

allennlp/data/vocabulary.py
```diff
@@ -259,6 +259,19 @@ def __init__(
...
+        cls, model_name: str, namespace: str = "tokens"
+    ) -> "Vocabulary":
+        vocab = cls.empty()
+        from allennlp.common import cached_transformers
+
+        tokenizer = cached_transformers.get_tokenizer(model_name)
+        result = vocab.add_transformer_vocab(tokenizer)
+        vocab._token_to_index[namespace] = result["token_to_index"]
+        vocab._index_to_token[namespace] = result["index_to_token"]
+        return vocab
```
epwalsh, line 273:
```suggestion
        vocab.add_transformer_vocab(tokenizer)
        return vocab
```

epwalsh (changes requested):

allennlp/data/vocabulary.py
wlhgtc: done!

allennlp/data/vocabulary.py
wlhgtc: done!

wlhgtc:
@epwalsh  I have finished  all changes, ready for merge~

allennlp/data/vocabulary.py
```diff
@@ -259,6 +259,18 @@ def __init__(
             min_pretrained_embeddings,
         )
 
+    def from_pretrained_transformer(
```
epwalsh, line 262:
```suggestion
    @classmethod
    def from_pretrained_transformer(
```

allennlp/data/vocabulary.py
```diff
@@ -259,6 +259,19 @@ def __init__(
...
+        cls, model_name: str, namespace: str = "tokens", padding_token="[PAD]", oov_token="[UNK]"
+    ) -> "Vocabulary":
+        vocab = cls.empty()
+        vocab._oov_token = oov_token
+        vocab._padding_token = padding_token
+        from allennlp.common import cached_transformers
+
+        tokenizer = cached_transformers.get_tokenizer(model_name)
+        vocab.add_transformer_vocab(tokenizer, namespace)
+        return vocab
```
epwalsh, line 273:
```suggestion
    def from_pretrained_transformer(
        cls, model_name: str, namespace: str = "tokens", oov_token: Optional[str] = None
    ) -> "Vocabulary":
        """
        Initialize a vocabulary from the vocabulary of a pretrained transformer model.
        If `oov_token` is not given, we will try to infer it from the transformer tokenizer.
        """
        from allennlp.common import cached_transformers

        tokenizer = cached_transformers.get_tokenizer(model_name)
        if oov_token is None:
            if hasattr(tokenizer, "_unk_token"):
                oov_token = tokenizer._unk_token
            elif hasattr(tokenizer, "special_tokens_map"):
                oov_token = tokenizer.special_tokens_map.get("unk_token")
        vocab = cls(non_padded_namespaces=[namespace], oov_token=oov_token)
        vocab.add_transformer_vocab(tokenizer, namespace)
        return vocab
```

allennlp/data/vocabulary.py
```diff
@@ -417,6 +430,23 @@ def empty(cls) -> "Vocabulary":
         """
         return cls()
 
+    def add_transformer_vocab(
+        self, tokenizer: PreTrainedTokenizer, namespace: str = "tokens"
+    ) -> None:
+        """
+        Copies tokens from ```transformers``` model's vocab
```
epwalsh, line 437:
```suggestion
        Copies tokens from a transformer tokenizer's vocab into the given namespace.
```

allennlp/data/vocabulary.py
```diff
@@ -417,6 +430,23 @@ def empty(cls) -> "Vocabulary":
...
+        try:
+            vocab_items = tokenizer.get_vocab().items()
+        except NotImplementedError:
+            vocab_items = (
+                (tokenizer.convert_ids_to_tokens(idx), idx) for idx in range(tokenizer.vocab_size)
+            )
+
+        for word, idx in vocab_items:
+            self._token_to_index[namespace][word] = idx
+            self._index_to_token[namespace][idx] = word
```
epwalsh, line 448:
```suggestion
            self._index_to_token[namespace][idx] = word
            
        self._non_padded_namespaces.add(namespace)
```

epwalsh (changes requested):
Thanks for sticking with this @wlhgtc!

There's just one thing that needs to be fixed around the use of the padding and oov tokens. AllenNLP `Vocabulary` objects make some assumptions about how padding is done that is not compatible with `transformers`. So the work-around is just make sure the transformer vocab namespace is considered a non padded namespace so that AllenNLP doesn't mess with it.

I also added a suggestion that we try to infer the OOV token from the tokenizer when we can.

And finally, could you add a test to `tests/data/vocabulary_test.py` that just makes sure we can save and load the vocab correctly. Something like this:

```python
class TestVocabularyFromPretrainedTransformer(AllenNlpTestCase):
    @pytest.mark.parametrize("model_name", ["bert-base-cased", "roberta-base"])
    def test_from_pretrained_transformer(self, model_name):
        namespace = "tokens"
        tokenizer = cached_transformers.get_tokenizer(model_name)

        vocab = Vocabulary.from_pretrained_transformer(model_name, namespace=namespace)
        assert vocab._token_to_index[namespace] == tokenizer.get_vocab()
        vocab.save_to_files(self.TEST_DIR / "vocab")

        vocab1 = Vocabulary.from_files(self.TEST_DIR / "vocab")
        assert vocab1._token_to_index[namespace] == tokenizer.get_vocab()
```

wlhgtc:
@epwalsh  Excellent work !
When I didn't add the OOV in this method, I count an error when I use multi-GPU training. 
Seems it reads the vocab by   `set_from_file `    method so I miss the OOV (`@@UNKNOWN@@`).
Then I added the OOV and PAD for convenient, anyway your method is better!

CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `ModelUsage` to `ModelCard` class.
 - Added a way to specify extra parameters to the predictor in an `allennlp predict` call.
-
+- Added a way to initialize a `Vocabulary` from transformers models.
```
epwalsh, line 18:
```suggestion
- Added a way to initialize a `Vocabulary` from transformers models.

```

epwalsh (approved):
@wlhgtc LGTM! Thanks for taking this!

PR #4959, opened 2021-02-03
Fixes recording validation metrics for learning rate schedulers that rely on it

dirkgr:
@bratao, @wlhgtc, it was a screw-up after all.


bratao:
@dirkgr  You rock!

dirkgr:
I'm still making a test for this. Don't want this to get missed a second time.

AkshitaB (approved):

PR #4961, opened 2021-02-05
Update mean_absolute_error.py

willfrey:
I had a `TypeError`stating that `torch.Tensor` objects are not JSON-serializable, which seems perfectly reasonable. The culprit was `MeanAbsoluteError.get_metric(...)` because it was returning a tensor and not a float.

No issue is currently open because it was a quick edit. I can open one if you like.

- Explicitly return a `float` in `MeanAbsoluteError.get_metric`

### DIFF
--- a/allennlp/training/metrics/mean_absolute_error.py
+++ b/allennlp/training/metrics/mean_absolute_error.py
@@ -66,7 +66,7 @@ class MeanAbsoluteError(Metric):
         mean_absolute_error = self._absolute_error / self._total_count
         if reset:
             self.reset()
-        return {"mae": mean_absolute_error}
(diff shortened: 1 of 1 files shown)


willfrey:
Looks like there are two tests where there's a strict `==` comparison for floating point values. I can fix the tests if the powers-that-be think this is indeed a problem and this is the best way to resolve it.

I'm always nervous when changing existing tests that I didn't write 😅

epwalsh:
Hey @willfrey, yes thank you! Please go ahead and fix the test when you get a chance 🙂

willfrey:
Closing and re-opening because I deleted my original fork 🙃

PR #4968, opened 2021-02-10
re-write docs commit history on releases

epwalsh:
We ran into a GitHub repo size limit with https://github.com/allenai/allennlp-docs, so GitHub refused to continue to deploy to GitHub Pages.

To fix this, I rewrote the commit history according to this guide: https://medium.com/@sangeethkumar.tvm.kpm/cleaning-up-a-git-repo-for-reducing-the-repository-size-d11fa496ba48. But we'll hit this limit again in the future unless we periodically rewrite the commit history. This PR proposes we automatically do that on every release.


schmmd:
Makes sense to me--the version history is not too useful here.  We could also consider publishing our docs via another mechanism.

dirkgr (approved):
This is a complete abuse of GitHub and we should not do this. But I also think there is no need to spend a bunch of time right now on re-engineering this once again. So let's ship this while being aware that at some point this will blow up again.

AkshitaB (approved):

PR #4969, opened 2021-02-10
T5

dirkgr:
This adds a transformer toolkit version of T5. The code is mostly stolen from huggingface and then given type annotations and some minor improvements.


epwalsh:
Currently I'm debugging. The loss doesn't match what you get from HF.

dirkgr:
Holy fixes Batman. Did I screw up that much in my port?

epwalsh:
Ready for review now. Not necessary ready to merge, because I definitely want feedback on the API.

allennlp/data/dataset_readers/t5.py
```diff
@@ -0,0 +1,43 @@
...
+        )
+        self.tokenizer = PretrainedTransformerTokenizer(model_name)
+        self.token_indexers = {"tokens": PretrainedTransformerIndexer(model_name)}
+
+    @overrides
+    def _read(self, file_path):
+        with open(cached_path(file_path)) as data_file:
+            for line in self.shard_iterable(data_file):
+                source, target = line.strip().split("\t")
+                yield self.text_to_instance(source, target)
```
epwalsh, line 27: As of right now, this reader is really designed for fine-tuning tasks, not LM-ing. Should we also have a reader for LM-ing? Or just have a flag that makes it work for LM-ing?

epwalsh: The main difference would be that we'd just have source text/tokens, no target.

allennlp/modules/transformer/t5.py
```diff
@@ -56,14 +70,16 @@ def forward(self, hidden_states) -> FloatT:
...
 class T5DenseGatedGeluDense(TransformerModule, FromParams):
-    def __init__(self, hidden_size: int = 1024, ff_size: int = 4096, dropout: float = 0.1):
+    def __init__(self, hidden_size: int = 512, ff_size: int = 2048, dropout: float = 0.1):
         super().__init__()
-        self.wi_0 = nn.Linear(hidden_size, ff_size, bias=False)
-        self.wi_0.weight.data.normal_(mean=0.0, std=hidden_size ** -0.5)
-        self.wi_1 = nn.Linear(hidden_size, ff_size, bias=False)
-        self.wi_1.weight.data.normal_(mean=0.0, std=hidden_size ** -0.5)
-        self.wo = nn.Linear(ff_size, hidden_size, bias=False)
+        self.hidden_size = hidden_size
```
dirkgr, line 77: Do we have to store the hidden size? Can't we just read it out of `self.wi_0` whenever we need it?

dirkgr: Same question for `T5DenseReluDense`

allennlp/modules/transformer/t5.py
```diff
@@ -109,14 +125,12 @@ class T5AttentionOutput:
...
-    _huggingface_mapping = {"d_model": "hidden_size", "n_heads": "num_heads"}
-
     def __init__(
         self,
         is_decoder: bool = False,
-        hidden_size: int = 1024,
-        key_value_proj_dim: int = 64,
-        num_heads: int = 16,
+        hidden_size: int = 512,
+        key_value_proj_dim: int = 64,  # d_kv
```
dirkgr, line 132: `d_kv`?

allennlp/modules/transformer/t5.py
```diff
@@ -326,17 +340,16 @@ def project(hidden_states, proj_layer, key_value_states, past_key_value) -> Floa
                 position_bias = position_bias[:, :, -seq_length:, :]
 
             if mask is not None:
-                position_bias = apply_mask(
-                    position_bias, mask
-                )  # (batch_size, n_heads, seq_length, key_length)
+                # Shape: (batch_size, num_heads, seq_length, key_length)
+                position_bias = position_bias + mask
```
dirkgr, line 344: I changed this to use our `apply_mask()`, and you changed it back. I assume you looked into this and you know what's going on?

I seem to remember that with masking, HF likes to make a float mask early on and then use that everywhere, whereas I was trying to pass around boolean masks and make them into floats at the last possible moment (if ever). I think my approach is better, because you can use the boolean mask in different ways in different places, for example, apply in an fp16 way in one place, but apply in a fp32 way in another place, or use `sum()` to find the number of non-masked items, etc.

allennlp/modules/transformer/t5.py
```diff
@@ -693,24 +712,31 @@ def forward(
...
+        extended_attention_mask = get_extended_attention_mask(
+            attention_mask, input_shape, inputs_embeds.dtype, is_decoder=self.is_decoder
+        )
+
+        if self.is_decoder and encoder_attention_mask is not None:
+            encoder_extended_attention_mask = invert_attention_mask(
+                encoder_attention_mask, inputs_embeds.dtype
+            )
+        else:
+            encoder_extended_attention_mask = None
```
dirkgr, line 733: More masking code that I removed and you brought back ...

allennlp/modules/transformer/t5.py
```diff
@@ -773,187 +801,473 @@ def forward(
...
+                    self_attention=block_self_attention.construct(
+                        is_decoder=False, has_relative_attention_bias=(i == 0)
+                    )
+                ),
+                cross_attention=None,
+                ff=block_ff.construct(),
+            )
+            for i in range(num_blocks)
+        ]
+        return cls(token_embeddings, blocks, final_layer_norm=final_layer_norm, dropout=dropout)
```
dirkgr, line 841: This approach is my current best idea for how to support the different pre-trained dimensions without resorting to HF's config mechanism.

allennlp/modules/transformer/t5.py
```diff
@@ -773,187 +801,473 @@ def forward(
...
 
 
 @dataclass
-class T5ForConditionalGenerationOutput:
-    logits: FloatT
-    loss: Optional[FloatT] = None
-    past_key_values: Optional[List[KeyValueStates]] = None
-    decoder_hidden_states: Optional[List[FloatT]] = None
+class T5Output:
```
dirkgr, line 894: Even if we do nothing else, this version of T5, with the types and the docs and everything, is already a superior starting point if you want to hack on top of T5. I wonder if we should see if HF would want some if this ported back? Let's talk about that once this PR is done.

allennlp/modules/transformer/t5.py
```diff
@@ -773,187 +801,473 @@ def forward(
...
 
         return shifted_input_ids
 
+    def _get_lm_logits(self, decoder_last_hidden_state: FloatT) -> FloatT:
+        # Shape: (batch_size, target_length, model_dim)
+        sequence_output = decoder_last_hidden_state
+        # Rescale output before projecting on vocab
+        # TODO: HF only does this when does this when embeddings are tied.
+        # Currently tied embeddings is the only option we have, but if make
+        # that configurable then we should put this in an 'if' block.
```
dirkgr, line 1094: The T5 paper does a bunch of experiments tying things here and there, but ultimately concludes that almost none of them work. The HF code supports all the weight tying options, but I removed that stuff because I just wanted to run T5-Large at the time. I suppose it would be good to have for someone to do future experiments, but not very critical.

allennlp/modules/transformer/t5.py
```diff
@@ -773,187 +801,473 @@ def forward(
...
+            if decoder_attention_mask is None:
+                decoder_attention_mask = ~(labels == self.pad_token_id)
+
+            # Get decoder inputs from shifting lm labels to the right and pre-pending
+            # the decoder start token ID.
+            # Shape (both): (batch_size, target_length)
+            decoder_input_ids = self._shift_right(labels, self.decoder_start_token_id)
+
+            # Replace possible -100 values in labels by `pad_token_id`
+            decoder_input_ids.masked_fill_(decoder_input_ids == -100, self.pad_token_id)
```
dirkgr, line 1139: A backwards compatibility hack we carried over I assume?

allennlp/modules/transformer/transformer_module.py
```diff
@@ -32,23 +32,26 @@ def __init__(self, *args, **kwargs):
...
+        source: str = "huggingface",
         mapping: Optional[Dict[str, str]] = None,
     ):
         """
         Returns the mapping to be used, based on the optional `pretrained_module`.
         If `pretrained_module` is not given, the default module-level mapping is returned.
         """
         combined_mapping = {}
-        if "huggingface" in source:
+        if "huggingface" == source:
```
dirkgr, line 43: @AkshitaB, do you sign off on this change? Is it safe? Was there a good reason you were doing a substring lookup before?

tests/modules/transformer/t5.py
```diff
@@ -1,13 +1,118 @@
...
+@pytest.fixture(scope="module")
+def hf_model(model_name):
+    return hf_t5.T5ForConditionalGeneration.from_pretrained(model_name).eval()
+
+
+def test_t5_forward_loss(
+    model: T5,
+    hf_model: hf_t5.T5ForConditionalGeneration,
+    tokenizer: hf_t5.T5Tokenizer,
+):
```
dirkgr, line 44: Tests 😍

dirkgr (commented):
I can only comment since it's my one PR :-)

epwalsh:
See https://github.com/allenai/allennlp-models/pull/241

allennlp/modules/transformer/t5.py
epwalsh: Good catch, didn't need this after all.

allennlp/modules/transformer/t5.py
epwalsh: Carried over, but I don't know why 😏

allennlp/modules/transformer/t5.py
epwalsh: Done.

dirkgr:
Is this ready for another round of reviews?

epwalsh:
@dirkgr yes it's ready

allennlp/predictors/predictor.py
```diff
@@ -294,17 +294,8 @@ def predict_batch_json(self, inputs: List[JsonDict]) -> List[JsonDict]:
     def predict_batch_instance(self, instances: List[Instance]) -> List[JsonDict]:
         for instance in instances:
             self._dataset_reader.apply_token_indexers(instance)
-        return [
-            sanitize(self._finalize_output(o)) for o in self._model.forward_on_instances(instances)
-        ]
-
-    def _finalize_output(self, output: JsonDict) -> JsonDict:
```
dirkgr: Why are we not doing this anymore?

dirkgr (commented):
Changes look good, just one question.

I can't hit "Approve" because I started this PR, but I'm fine with merging this.

allennlp/predictors/predictor.py
epwalsh: I added this initially, then after doing some other refactoring I realized I really didn't need it, and instead the logic I put there should live in `Model.make_output_human_readable()`.

allennlp/modules/transformer/transformer_module.py
dirkgr: @AkshitaB, ping?

dirkgr (commented):
No concerns, though I didn't read the big file a second time. In fact, I'm not sure why GitHub thinks it changed so much since the last time I looked at it.

allennlp/modules/transformer/transformer_module.py
AkshitaB: Yes, this is fine. Initially, the thought was that the source could be different huggingface configs, with some minor differences that could be addressed. But we haven't used it like that anywhere, so this should work fine.

PR #4970, opened 2021-02-11
Add options to log inputs in trainer

AkshitaB:

- Log batch inputs to tensorboard.
- Log batch inputs to console.
- Added `LogWriter` class. `TensorBoardWriter` and `ConsoleWriter` inherit from this. 
- Added `LogCallback` class. `TensorBoardCallback` and `ConsoleLogCallback` inherit from this.


AkshitaB:
The test case is not super useful in showcasing what's logged, unfortunately.

epwalsh (commented):
I think the console logging functionality should be separate from the TensorBoard writer. I would definitely love to see that in the logs, but I just don't want to have to use the TensorBoard writer in order to get that. Maybe we could just have a separate TrainerCallback for that?

AkshitaB:
> I think the console logging functionality should be separate from the TensorBoard writer. I would definitely love to see that in the logs, but I just don't want to have to use the TensorBoard writer in order to get that. Maybe we could just have a separate TrainerCallback for that?

@epwalsh That's a good point. Should we also move some of the console logging stuff in `log_metrics`?

epwalsh:
> Should we also move some of the console logging stuff in log_metrics?

That would be great. I was actually wondering where that functionality had gone the other day. I thought it had been removed because I don't normally use the `TensorBoardWriter`.

epwalsh (commented):
I'm not sure how I feel about the `LogWriter` abstraction. It seems like `ConsoleWriter` doesn't really fit in with that API (for example, `ConsoleWriter` doesn't actually log histograms), and doesn't really use or need most of the functionality that is implemented in the `LogWriter` base class, right? I mean, all we really want `ConsoleWriter` to do is to nicely display metrics and mem usage at the end of each epoch, right? I think that could just be implemented entirely within a `TrainerCallback`, and then we wouldn't need the `LogWriter` base class.

AkshitaB:
> It seems like `ConsoleWriter` doesn't really fit in with that API (for example, `ConsoleWriter` doesn't actually log histograms), and doesn't really use or need most of the functionality that is implemented in the `LogWriter` base class, right? 

It does log those tensors; that functionality is used in `log_batch`, etc.. perhaps we can rename the method to `add_train_tensor` instead of `add_train_histogram`? 

> I mean, all we really want `ConsoleWriter` to do is to nicely display metrics and mem usage at the end of each epoch, right? I think that could just be implemented entirely within a `TrainerCallback`, and then we wouldn't need the `LogWriter` base class. 

We log learning rate, model inputs (which was the primary goal here), metrics, gradient updates, etc. in both, console writer and tensorboard writer (common functionality that's implemented in log writer). The only thing that is currently not logged on the console is the activation histograms, because I wasn't sure if they'll be particularly useful. But it will be good to have the option to implement console options for that; maybe some concise metrics like mean etc. instead of printing out the histogram?

epwalsh:
Okay, fair enough. `add_train_tensor` instead of `add_train_histogram` sounds good.

This might also be a good way to officially incorporate Weights & Biases into AllenNLP, i.e. by adding a `class WandBWriter(LogWriter)` (we can worry about that later).

I do want to make sure we keep things backwards compatible though. So if you rename `add_train_histogram -> add_train_tensor`, I think we should still have a `add_train_histogram` method on the `TensorBoardWriter` than can just call `add_train_tensor`.

allennlp/training/console_writer.py
```diff
@@ -0,0 +1,85 @@
...
+            elif val_metric is not None:
+                logger.info(no_train_message_template, name.ljust(name_length), "N/A", val_metric)
+            elif train_metric is not None:
+                logger.info(no_val_message_template, name.ljust(name_length), train_metric, "N/A")
+
+    def enable_activation_logging(self, model: Model) -> None:
+        if self._distribution_interval is not None:
+            logger.info("Activation logging is not available for ConsoleWriter.")
+
+    def log_activation_histogram(self, outputs, log_prefix: str) -> None:
```
epwalsh, line 84: Should this be

```suggestion
    def log_activation_distribution(self, outputs, log_prefix: str) -> None:
```

allennlp/training/log_writer.py
```diff
@@ -0,0 +1,269 @@
...
+        be a closure around an instance variable in your `Trainer` class.  Because of circular
+        dependencies in constructing this object and the `Trainer`, this is typically `None` when
+        you construct the object, but it gets set inside the constructor of our `Trainer`.
+    """
+
+    def __init__(
+        self,
+        serialization_dir: Optional[str] = None,
+        summary_interval: int = 100,
+        distribution_interval: int = None,
```
epwalsh, line 57:
```suggestion
        distribution_interval: Optional[int] = None,
```

allennlp/training/console_writer.py
```diff
@@ -0,0 +1,85 @@
...
+
+import torch
+from allennlp.models.model import Model
+
+from allennlp.training.log_writer import LogWriter
+
+logger = logging.getLogger(__name__)
+
+
+class ConsoleWriter(LogWriter):
```
epwalsh, line 13: It would be good to have a descriptive docstring here

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `ModelUsage` to `ModelCard` class.
 - Added a way to specify extra parameters to the predictor in an `allennlp predict` call.
 - Added a way to initialize a `Vocabulary` from transformers models.
+- Added `LogWriter` class. `TensorBoardWriter` and `ConsoleWriter` inherit from this.
```
epwalsh, line 23: I think these notes should make it clear that `ConsoleWriter` is an addition as well

allennlp/training/log_writer.py
```diff
@@ -0,0 +1,269 @@
...
+
+    # Parameters
+
+    summary_interval : `int`, optional (default = `100`)
+        Most statistics will be written out only every this many batches.
+    distribution_interval : `int`, optional (default = `None`)
+        If provided, activation distributions will be written out every this many batches.
+        If None, activation distributions will not be written out.
+        When this parameter is specified, the following additional logging is enabled:
+            * distributions of model parameters
```
epwalsh, line 28:
```suggestion
            * Distributions of model parameters
```

epwalsh (commented):
I just have few minor comments after skimming over the changes. I'll take a closer look later and run some training jobs to make sure everything looks okay 👍

allennlp/training/tensorboard_writer.py
```diff
@@ -61,210 +64,94 @@ def __init__(
...
-        if serialization_dir is not None:
+
+        if distribution_interval is not None and histogram_interval is not None:
+            logger.warning(
+                f"`histogram_interval`=`{histogram_interval}` will be ignored "
+                f"in favor of `distribution_interval`=`{distribution_interval}`"
+            )
+
+        if distribution_interval is None:
+            distribution_interval = histogram_interval
```
epwalsh, line 83: I think we should issue a deprecation warning instead. And instead of listing `histogram_interval` as an actual parameter, we could just check for it in `**kwargs`.

Like what we did here: https://github.com/allenai/allennlp/blob/1cff6ad97fe8304d03923ddfaedb5e05bb67bee5/allennlp/data/dataset_readers/conll2003.py#L99-L104

allennlp/training/tensorboard_writer.py
```diff
@@ -1,23 +1,21 @@
...
+from allennlp.training.log_writer import LogWriter
+
 logger = logging.getLogger(__name__)
 
 
-class TensorBoardWriter(FromParams):
+class TensorBoardWriter(LogWriter):
     """
-    Class that handles TensorBoard (and other) logging.
+    Class that handles TensorBoard logging.
```
epwalsh, line 18: The API docs here are not building correctly, and I think it's because there's a blank line on line 24 (not your fault).

It's actually okay to have a blank line there, but we need to have a blank line in between every different parameter description. In other words, we need add a blank line after lines 26, 28, 43, etc...

allennlp/training/log_writer.py
```diff
@@ -0,0 +1,269 @@
...
+
+    summary_interval : `int`, optional (default = `100`)
+        Most statistics will be written out only every this many batches.
+    distribution_interval : `int`, optional (default = `None`)
+        If provided, activation distributions will be written out every this many batches.
+        If None, activation distributions will not be written out.
+        When this parameter is specified, the following additional logging is enabled:
+            * Distributions of model parameters
+            * The ratio of parameter update norm to parameter norm
+            * Distribution of layer activations
```
epwalsh, line 30: This isn't rendering correctly in the docs. May need blank lines before and after this list.
```suggestion

            * Distributions of model parameters
            * The ratio of parameter update norm to parameter norm
            * Distribution of layer activations
            
```

allennlp/training/log_writer.py
```diff
@@ -0,0 +1,269 @@
...
+        self,
+        serialization_dir: Optional[str] = None,
+        summary_interval: int = 100,
+        distribution_interval: Optional[int] = None,
+        batch_size_interval: Optional[int] = None,
+        should_log_parameter_statistics: bool = True,
+        should_log_learning_rate: bool = False,
+        should_log_inputs: bool = False,
+        get_batch_num_total: Callable[[], int] = None,
+        **kwargs,
```
epwalsh, line 63: Probably don't need `**kwargs`

```suggestion
```

epwalsh (commented):
After trying out `ConsoleWriter`, I was a overwhelmed with the output. Some of the output could be useful for debugging but the default settings just resulted in the Mem and CPU / GPU usage being logged after every batch, which is not particularly useful.

Also, when I set `should_log_inputs` to `True`, I didn't actually see that in the logs.

Personally I think that the logging of the first batch is the only feature of `ConsoleLogger` that I would want to use, and so I feel that a simpler approach may be better. For example, we could just implement that directly in the `GradientDescentTrainer` and get rid of the `ConsoleLogger`.

Let me know what you think.

AkshitaB:
@epwalsh This is ready for another look. I've removed `ConsoleWriter`, but I think it's a good idea to keep `LogWriter` and `LogCallback` abstractions for Weights and Biases as was discussed.

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `ModelUsage` to `ModelCard` class.
 - Added a way to specify extra parameters to the predictor in an `allennlp predict` call.
 - Added a way to initialize a `Vocabulary` from transformers models.
+- Added `LogWriter` classe. `TensorBoardWriter` now inherits from `LogWriter`.
```
epwalsh, line 20:
```suggestion
- Added `LogWriter` class. `TensorBoardWriter` now inherits from `LogWriter`.
```

allennlp/training/trainer.py
```diff
@@ -30,10 +30,28 @@
 from allennlp.training.moving_average import MovingAverage
 from allennlp.training.optimizers import Optimizer
 from allennlp.training.tensorboard_writer import TensorBoardWriter
+from allennlp.training.log_writer import LogWriter
 
 logger = logging.getLogger(__name__)
 
 
+def get_train_and_validation_metrics(metrics: Dict):
```
epwalsh, line 38: Nit: should probably be a private function:
```suggestion
def _get_train_and_validation_metrics(metrics: Dict):
```

allennlp/training/trainer.py
```diff
@@ -290,8 +297,130 @@ def on_end(
...
     ) -> None:
         if is_primary:
-            assert self._tensorboard is not None
-            self._tensorboard.close()
+            assert self._log_writer is not None
+            self._log_writer.close()
+
+
+@TrainerCallback.register("console")
+class ConsoleLogCallback(LogCallback):
```
epwalsh, line 305: How about?
```suggestion
@TrainerCallback.register("console-logger")
class ConsoleLoggerCallback(LogCallback):
```

epwalsh, line 305: Should this really inherit from `LogCallback`?

CHANGELOG.md
epwalsh, line 20: We should also mentioned that the `histogram_interval` parameter is deprecated.

allennlp/training/trainer.py
```diff
@@ -290,8 +297,130 @@ def on_end(
...
+    def _log_fields(self, fields: Dict, log_prefix: str = ""):
+        for key, val in fields.items():
+            key = log_prefix + "/" + key
+            if isinstance(val, dict):
+                self._log_fields(val, key)
+            elif isinstance(val, torch.Tensor):
+                logger.info(f"\"{key}\" (Shape : {' x '.join([str(x) for x in val.shape])})")
+                torch.set_printoptions(threshold=2)
+                logger.info(f"{val}")
+                torch.set_printoptions(threshold=1000)
```
epwalsh, line 364: This looks a little weird in the logs because of how the representation of the tensor is split over multiple lines:

image

Maybe something like this:

```suggestion
                torch.set_printoptions(threshold=2)
                logger.info("%s (Shape: %s)\n%s", key, " x ".join([str(x) for x in val.shape]), val)
                torch.set_printoptions(threshold=1000)
```

Which gives you:

image

Feel free to implement this differently (see my comment about my preference for avoiding f-strings below).

allennlp/training/trainer.py
```diff
@@ -290,8 +297,130 @@ def on_end(
...
+            elif isinstance(val, List):
+                logger.info(f'Field : "{key}" : (Length : {len(val)} of type "{type(val[0])}")')
```
epwalsh, line 366: This is a minor nit pick, and only a suggestion, so feel free to ignore...

I think it's better to avoid f-strings with logging, because with f-strings the expressions in `{ ... }` always have to be evaluated, regardless of the log level. By doing this instead:

```suggestion
                logger.info("Field : \"%s\" : (Length %d of type \"%\")", key, len(val), type(val[0]))
```

The expression is only evaluated if the log level is it or below INFO.

allennlp/training/trainer.py
```diff
@@ -290,8 +297,130 @@ def on_end(
...
+            elif isinstance(val, str):
+                logger.info(f'Field : "{key}"')
+                logger.info("{:20.20} ...".format(val))
+            else:
+                logger.info(f'"{key}" : "{val}"')
```
epwalsh, line 371: Same comment about f-strings here.

allennlp/training/trainer.py
```diff
@@ -290,8 +297,130 @@ def on_end(
...
+        batch_number: int,
+        is_training: bool,
+        is_primary: bool = True,
+        batch_grad_norm: Optional[float] = None,
+        **kwargs,
+    ) -> None:
+        # In the distributed case we need to call this from every worker, since every
+        # worker reports its own memory usage.
+        cpu_memory_usage = common_util.peak_cpu_memory()
+        gpu_memory_usage = common_util.peak_gpu_memory()
```
epwalsh, line 330: We probably don't want to call this on every batch, especially if we're only logging it on the first batch

allennlp/training/trainer.py
```diff
@@ -226,17 +243,17 @@ def on_batch(
...
+        if self._log_writer.should_log_distributions_this_batch():
             assert self._param_updates is not None
             for name, param in trainer.model.named_parameters():
                 self._param_updates[name].sub_(param.detach().cpu())
         else:
             self._param_updates = None
 
-        self._tensorboard.log_memory_usage(cpu_memory_usage, gpu_memory_usage)
-        self._tensorboard.log_batch(
+        self._log_writer.log_memory_usage(cpu_memory_usage, gpu_memory_usage)
```
epwalsh, line 255: I think this was my fault, but we might as well fix it in this PR. We shouldn't log mem and GPU usage on every batch because that's rather costly. We could either remove this entirely (as well as lines 241 and 242 above) or hide it behind a flag. I'm in favor of removing entirely.

allennlp/training/trainer.py
epwalsh: See comment below as well.

allennlp/training/trainer.py
```diff
@@ -290,8 +292,129 @@ def on_end(
...
+            # Log memory usage
+            cpu_memory_usage_total = 0.0
+            for worker, mem_bytes in cpu_memory_usage.items():
+                memory = mem_bytes / (1024 * 1024)
+                logger.info(f"memory_usage/worker_{worker}_cpu - {memory}")
+                cpu_memory_usage_total += memory
+            logger.info(f"memory_usage/cpu - {cpu_memory_usage_total}")
+            for gpu, mem_bytes in gpu_memory_usage.items():
+                memory = mem_bytes / (1024 * 1024)
+                logger.info(f"memory_usage/gpu_{gpu} - {memory}")
```
epwalsh, line 343: I don't think we really need this. We get these statistics logged at the end of each epoch anyway.

epwalsh (commented):
Looks great! Thanks for sticking with this. Just two final comments 👇

epwalsh (approved):

PR #4976, opened 2021-02-12
for torch.distributed.is_available()==False

KoichiYasuoka:
`import allennlp.nn.util` always fails when `torch.distributed.is_available()==False`. See [here](https://pytorch.org/docs/stable/distributed.html#torch.distributed.is_available) details.

- check `torch.distributed.is_available()` before `import ReduceOp`
- if unavailable, dummy `ReduceOp` created

### DIFF
--- a/allennlp/nn/util.py
+++ b/allennlp/nn/util.py
@@ -12,7 +12,12 @@ import math
 import numpy
 import torch
 import torch.distributed as dist
-from torch.distributed import ReduceOp
+
+if dist.is_available():
+    from torch.distributed import ReduceOp
+else:
+    class ReduceOp(object):
+        SUM = None
 
 from allennlp.common.checks import ConfigurationError
 from allennlp.common.util import int_to_device, is_distributed


allennlp/nn/util.py
```diff
@@ -12,7 +12,12 @@
 import numpy
 import torch
 import torch.distributed as dist
-from torch.distributed import ReduceOp
+
+if dist.is_available():
+    from torch.distributed import ReduceOp
+else:
+    class ReduceOp(object):
+        SUM = None
```
epwalsh, line 20: `ReduceOp` is actually only imported for use a type hint. So we could do this instead:

```suggestion
if TYPE_CHECKING:
    from torch.distributed import ReduceOp
```

The `TYPE_CHECKING` variable can be imported on line 9 above from the `typing` module.

epwalsh (changes requested):
Thanks for catching this @KoichiYasuoka! I left one suggestion below.

allennlp/nn/util.py
KoichiYasuoka: Sorry, but how do we rewrite `def dist_reduce(value: _V, reduce_op: ReduceOp, **kwargs)` when `TYPE_CHECKING` is False?

KoichiYasuoka: And I found that [average.py](https://github.com/allenai/allennlp/blob/main/allennlp/training/metrics/average.py#L4) imported `ReduceOp`. Is it really "type hint" use?

epwalsh: In `def dist_reduce(value: _V, reduce_op: ReduceOp, **kwargs)`, `ReduceOp` is simply a type hint. The Python interpreter doesn't care about type hints. And since the `TYPE_CHECKING` flag is `False` when the interpreter is running, you can use that flag to exclude imports that are only used as type hints.

The `TYPE_CHECKING` flag is only true when you're running `mypy` or some other type-checking program.

KoichiYasuoka: I've tried your suggestion above, but it causes another error:

```
>>> import allennlp.training.metrics.average
  File "/home/yasuoka/.local/lib/python3.7/site-packages/allennlp/training/__init__.py", line 1, in <module>
    from allennlp.training.checkpointer import Checkpointer
  File "/home/yasuoka/.local/lib/python3.7/site-packages/allennlp/training/checkpointer.py", line 13, in <module>
  File "/home/yasuoka/.local/lib/python3.7/site-packages/allennlp/nn/util.py", line 2024, in <module> 
    def dist_reduce(value: _V, reduce_op: ReduceOp, **kwargs) -> _V:
NameError: name 'ReduceOp' is not defined
```

The use `TYPE_CHECKING` flag causes another error, since `allennlp.training.metrics.average` imports `ReduceOp` as a class object.

KoichiYasuoka: Also, the dummy `ReduceOp` must have a dummy attribute `SUM`. It is used in `allennlp.training.metrics.average`.

epwalsh: Ah, you're right, I was wrong about how the interpreter handles type hints. I apologize.

I still want to avoid making a dummy `ReduceOp`. So instead let's just remove the type hint from `dist_reduce()` and then in `allennlp.training.metrics.average` we could avoid directly importing `ReduceOp` and only use it when `is_distributed()` is `True`.

epwalsh:
Hey @KoichiYasuoka, any update on this? If you don't have time I could take it.

allennlp/nn/util.py
KoichiYasuoka: I cannot rewrite `allennlp.training.metrics.average` since it is called from so many libraries. How about you @epwalsh?

KoichiYasuoka:
I still stick to the dummy `ReduceOp`, but you may rewrite it if you can really safely,@epwalsh.

epwalsh:
Closing in favor of https://github.com/allenai/allennlp/pull/5011.

Issue #4911, opened 2021-01-13
Separate Optimizers for each Head in the MultiTaskModel


jbrry:
**Is your feature request related to a problem? Please describe.**
I'm using the `multitask` model which consists of multiple `heads`. At the moment, the loss from each head is summed together and is then jointly optimized using a single optimizer. I currently have a head where the input is only a word-level feature and another head where the input is only characters. The heads have comparable performance when I use a pre-trained word embedding or a randomly initialized embedding, e.g. the word and character heads achieve `0.837` and `0.819` accuracy respectively. When I changed the word-level feature to using the output of a pre-trained transformer, the scores of the heads change dramatically to `0.902` and `0.210` for word and character models respectively. In this case, the head which uses an embedding from the pretrained transformer starts to dominate.

**Describe the solution you'd like**
There currently is an option `loss_weights` which controls the weight to apply to the loss from each head. However, the [work I am replicating](https://arxiv.org/pdf/1805.08237.pdf) uses separate optimizers for each head and they found this to be most effective: 

> Separate optimization outperformed joint optimization by up to 2.1 percent absolute [...] We hypothesize that separately training the models forces each sub model (word and character) to be strong enough to make high accuracy predictions and in some sense serves as a regularizer in the same way that dropout does for individual neurons.

This approach might be better than manipulating the `loss_weights` values, which might differ between tasks and would thus require some level of intuition to properly control. Additionally, separate optimizers would enable each head to minimize its own loss independently which could result in individually stronger heads.

Admittedly, I do not have that much experience with the `Trainer` code but I imagine a crude solution would be to create my own `Trainer`, e.g. `gradient_descent_separate_optimizers` and change the type of the optimizer from `optimizer : torch.nn.Optimizer` to `optimizers : Dict[str, torch.nn.Optimizer]`, where the optimizer is keyed on the head name. Then I could mostly duplicate some of the logic in the `gradient_descent` `Trainer` but instead of calling the individual optimizer, e.g. `self.optimizer.step()`, the head-specific optimizer would be called, optimizing the loss for each head.

I'm just wondering would this be the best way of going about this? I'd be happy to hear if you have any suggestions or advice for implementing this or if you would think it would be a good idea to have as part of the multitask code. Thanks!

dirkgr:
The concerns
--

[...]

PR #4981, opened 2021-02-15
Regex optimizer

jbrry:
Provides a solution for the feature request in #4911 (submitting for review).

- Adds a [`RegexOptimizer`](https://github.com/jbrry/allennlp/blob/fe463590d088f9c59654110bf09b65939caf75b0/allennlp/training/optimizers.py#L208) to `training/optimizers.py`.
- This is an `Optimizer` which supports storing multiple optimizers, such that separate optimizers can be assigned to certain model parameters.
- It uses [`make_parameter_groups`](https://github.com/jbrry/allennlp/blob/fe463590d088f9c59654110bf09b65939caf75b0/allennlp/training/optimizers.py#L34) to obtain matches between certain regexes and model parameters, which will be assigned their own optimizer (you can also override optimizer defaults for groups as well).
- Any parameters which did not match the regexes are assigned to a `default` group.
- The `RegexOptimizer` overrides the methods in a `torch.optim.Optimizer`: `step()`, `state_dict()`, `load_state_dict()`, and `zero_grad()` such that these methods are called recursively for each optimizer in the object `self._grouped_optimizers`.
- The class is initialized with its own parameter groups in the same order as the groups returned by `make_parameter_groups`. The result is that when the optimizer is initialized in the `Trainer`, it has an object `optimizer.param_groups` which is necessary for a `torch.optim.Optimizer` and also means the parameters can be modified by tools such as `grad_scaler`, which scales and unscales the gradients of the parameters in `optimizer.param_groups` in the `Trainer`. These parameters are also modified by the `gradient_descent` `Trainer` [to set grads to None](https://github.com/jbrry/allennlp/blob/fe463590d088f9c59654110bf09b65939caf75b0/allennlp/training/trainer.py#L690), i.e. we don't actually call `optimizer.zero_grad()` in the `Trainer` but uses the current logic there.
- Each parameter group has its own optimizer `kwargs` such as the `"lr"` so it can be used with `learning_rate_scheduler`.
- Added tests to [`tests/training/optimizer_test.py`](https://github.com/jbrry/allennlp/blob/fe463590d088f9c59654110bf09b65939caf75b0/tests/training/optimizer_test.py#L80) which test optimizer group `kwargs` and that modifications to parameters in `optimizer.param_groups` and `optimizer._grouped_optimizers` are consistent.


allennlp/training/optimizers.py
```diff
@@ -196,6 +205,159 @@ def default(model_parameters: List) -> "Optimizer":
...
+
+    # Parameters
+
+    optimizers: `List[Dict[str, Any]]`
+        A list of optimizers to use. Each entry in the list is a dictionary of keyword arguments. A 'name' 
+        keyword argument should be given which will serve as the key to match the optimizer with a 
+        specific parameter group. You should also supply an entry for the default parameter group, 
+        e.g. 'name': 'default'.
+
+    parameter_groups:  `List[Tuple[List[str], Dict[str, Any]]`, optional (default = `None`)
```
jbrry, line 232: I had originally allowed this parameter to be `None` where everything would be appended to the `default` group but really you only want to use this class if you have different parameter groups and it required more code to handle so it is now a required parameter.

allennlp/training/optimizers.py
```diff
@@ -75,6 +80,10 @@ def make_parameter_groups(
     The dictionary's return type is labeled as `Any`, because it can be a `List[torch.nn.Parameter]`
     (for the "params" key), or anything else (typically a float) for the other keys.
     """
+    # The RegexOptimizer passes in the `groups` argument as False so this function is a no-op.
+    if groups is False:
```
jbrry, line 84: I'm not super happy about altering the type of `groups` (e.g. can now be `Any`) but it seemed to be the least invasive way of not touching any of the code when the individual optimizers are initialized (where they call this function with the `model_parameters` and `parameter_groups`)

allennlp/training/optimizers.py
```diff
@@ -196,6 +205,159 @@ def default(model_parameters: List) -> "Optimizer":
...
+        optimizer_names = [optimizer["name"] for optimizer in self.optimizers]
+        assert "default" in optimizer_names, (
+           "No optimizer was provided for the 'default' group."
+           " Please provide a 'name': 'default' parameter for one of the `optimizers`."
+           )
+
+        parameter_groups = make_parameter_groups(self.model_parameters, self.parameter_groups)
+        # Manually set optimizer kwargs for the default group, so certain schedulers are able
+        # to know the options for this group.
+        if len(list(parameter_groups[-1].keys())) == 1:
```
jbrry, line 260: By default, `make_parameter_groups` does not assign any optimizer options to the last (default) group. These options are set when the optimizer is initialized. But as this class is initialized with these `parameter_groups` I added these manually for the default group so they were visible to a `learning_rate_scheduler`.

allennlp/training/optimizers.py
```diff
@@ -196,6 +205,159 @@ def default(model_parameters: List) -> "Optimizer":
...
+                    " Alternatively, you can remove this optimizer from the provided `optimizers`"
+                    " if it is not relevant to a particular parameter group."
+                    )
+
+        # We have already created our parameter groups so we set `parameter_groups` to False, which is a no-op in `make_parameter_groups`.
+        # This is so that we don't have to change any of the code when initializing individual optimizers.
+        for optimizer_name, (params, optimizer_kwargs) in optimizer_groups.items():
+            self._grouped_optimizers[optimizer_name] = Optimizer.from_params(model_parameters=params, parameter_groups=False, params=Params(optimizer_kwargs))
+        
+        super().__init__(parameter_groups, optimizer_groups["default"][1])
```
jbrry, line 285: This is how the `RegexOptimizer`'s `parameter_groups` are accessed in the `Trainer` where they can be modified. The main `kwargs` (`optimizer_groups["default"][1]`) for the `RegexOptimizer` class are never used but it is a requirement for initializing this class. I think I should add another test for this to be sure.

allennlp/training/optimizers.py
```diff
@@ -196,6 +205,159 @@ def default(model_parameters: List) -> "Optimizer":
...
+        appropriate optimizer or to the 'default' optimizer.
+        """
+        optimizer_groups = {}
+
+        for parameter_group in parameter_groups:
+            # Check to see what optimizer this group should be assigned to.
+            if "name" in list(parameter_group.keys()):
+                optimizer_key = parameter_group["name"]
+                if not optimizer_key in optimizer_groups:
+                    optimizer_groups[optimizer_key] = ([], {})
```
jbrry, line 299: Each group contains a tuple where the first element is an iterable which stores model parameters or groups of parameters and the second element is a dictionary for the optimizer's default options. For reference: https://pytorch.org/docs/stable/optim.html#constructing-it

allennlp/training/optimizers.py
```diff
@@ -196,6 +205,159 @@ def default(model_parameters: List) -> "Optimizer":
...
+    @overrides
+    def load_state_dict(self, training_state: Dict[str, Any]):
+        """
+        Loads each optimizer's `state_dict`.
+        """
+        for optimizer_key, optimizer in self._grouped_optimizers.items():
+            optimizer.load_state_dict(training_state[f"{optimizer_key}_optimizer"])
+
+    @overrides
+    def zero_grad(self, set_to_none: bool = False):
```
jbrry, line 353: `set_to_none` enables setting gradients to `None` or zero, see [here for more info](https://pytorch.org/docs/stable/optim.html#torch.optim.Optimizer.zero_grad). I don't like hard-coding the Boolean here but perhaps I could add a parameter `set_gradients_to_none` to `RegexOptimizer` to change this. But the `Trainer` doesn't call `optimizer.zero_grad()` so I didn't implement this.

tests/training/optimizer_test.py
```diff
@@ -77,6 +77,116 @@ def test_optimizer_parameter_groups(self):
...
+                params = param_group["params"]
+                for param in params:
+                    regex_optimizer_grouped_optimizer_params.append(id(param))
+                    assert param.sum() == 0, "Param has non-zero values."
+
+        # As the optimizers are created in accordance with the order of groups in `parameter_groups`, the order of parameters should be deterministic.
+        assert regex_optimizer_params == regex_optimizer_grouped_optimizer_params
+
+
 class TestDenseSparseAdam(AllenNlpTestCase):
```
jbrry, line 190: There is a test failing here for a device type. It is not related to the `RegexOptimizer` tests but I didn't manage to fix it.

jbrry:
This should be ready for an initial review @dirkgr, so it would be great to get your opinion on this whenever you get the chance. I didn't get around to fixing the CI checks yet as it is just a draft but I can do those first if you prefer.

dirkgr:
Thanks! I'll get to this today!

dirkgr:
Sorry, this got more complicated than expected. I hope to send some suggestions on Monday, but I needed to make sure first that my suggestions work 😅.

allennlp/training/optimizers.py
```diff
@@ -151,6 +156,9 @@ def make_parameter_groups(
                     regex,
                 )
 
+        # If the default group is empty, remove it.
+        if len(parameter_groups[-1]["params"]) <= 0:
+            del parameter_groups[-1]
```
dirkgr, line 161: I also got rid of the change you made to `make_parameter_groups()`. That didn't feel very elegant to me. But then I discovered I had to make this other change instead.

Do you know why this function creates a default parameter group at all? PyTorch optimizers already handle the default case just fine, so I don't get why this is here. But I also didn't want to change code that's been fine for years for no reason, so I left it.

allennlp/training/optimizers.py
dirkgr: It still says `optional` though?

allennlp/training/optimizers.py
```diff
@@ -196,6 +205,149 @@ def default(model_parameters: List) -> "Optimizer":
...
+        # optimizer.params when optimizer is a RegexOptimizer.
+        parameter_groups = copy.deepcopy(parameter_groups)
+        for parameter_group in parameter_groups:
+            regexes, pg_overrides = parameter_group
+            optimizer_name = pg_overrides.get("optimizer_name", "default")
+            optimizer = self.optimizers[optimizer_name]
+            for key, value in optimizer.defaults.items():
+                if key not in pg_overrides:
+                    pg_overrides[key] = value
+        super().__init__(make_parameter_groups(model_parameters, parameter_groups), {})
```
dirkgr, line 304: I added this to capture the intent you had where you could still see the correct parameters in the outer optimizer, even if that's not the optimizer that's doing the actual work. I'm not entirely sure we need this though. Thoughts?

allennlp/training/optimizers.py
dirkgr: What's hardcoded? Just the default value?

tests/training/optimizer_test.py
```diff
@@ -77,6 +80,171 @@ def test_optimizer_parameter_groups(self):
...
+                        {
+                            "optimizer_name": "embedder",
+                            "betas": (0.9, 0.98),
+                            "lr": 2,
+                            "weight_decay": 0.01,
+                        },
+                    ],
+                    [["^encoder.*bias"], {"optimizer_name": "encoder", "lr": 0.001}],
+                    [["^encoder.*weight"], {"optimizer_name": "encoder", "lr": 0.002}],
+                    [["^tag_projection_layer.*.weight$"], {"lr": 5}],
```
dirkgr, line 122: I changed the format this expects a little bit. I didn't go quite as far as @epwalsh suggested, but far enough to make `Lazy` work. I think it's neat that the `parameter_groups` syntax stays the same compared to other optimizers (though I think that format is a little insane).

tests/training/optimizer_test.py
```diff
@@ -77,6 +80,171 @@ def test_optimizer_parameter_groups(self):
...
+            assert (
+                len(optimizer_names_for_parameter) == 1
+            ), f"Parameter {parameter_name} is assigned to the same optimizer more than once."
+
+            # make sure the parameter is really assigned to that optimizer
+            inner_optimizer = optimizer.optimizers[optimizer_names_for_parameter[0]]
+            parameters_in_optimizer = [
+                id(p) for pg in inner_optimizer.param_groups for p in pg["params"]
+            ]
+            assert id(parameter) in parameters_in_optimizer
```
dirkgr, line 157: This added test is new. The other tests only have superficial changes.

tests/training/optimizer_test.py
```diff
@@ -77,6 +80,171 @@ def test_optimizer_parameter_groups(self):
...
+        # These parameter groups are assigned to their own optimizer by the RegexOptimizer.
+        # Check that changes to the parameters in optimizer.param_groups affect the parameters in
+        # optimizer._grouped_optimizers.
+        regex_optimizer_params = set()
+        regex_optimizer_grouped_optimizer_params = set()
+
+        for param_group in optimizer.param_groups:
+            # Each param_group should have optimizer options visible so they can be used by schedulers.
+            lr = param_group["lr"]
+            assert lr > 0
```
dirkgr, line 231: This test still fails. On my machine, it's the only one that fails. The problem is that parameters that are not caught by any regex go into a default `param_group` with no parameters. This makes sense, because in that case the base parameters from the optimizer are used. It would also make sense if we didn't have that extra default `param_group` at all, but it looks like we've always had it, so I kept it.

My question is, @jbrry, why did this work in your code, and why do you think it should work? I don't think we guarantee that every `param_group` has an `lr` parameter, `RegexOptimizer` or not. That's not how `make_parameter_groups()` handles things. So I'm inclined to remove this test. But maybe you had it here for a reason?

tests/training/optimizer_test.py
dirkgr: Don't worry about it if it's clearly not related. I'll take a look if it comes back.

dirkgr (commented):
I wanted to use `Lazy` to create the inner optimizers. It's a problem to call into the `from_params` logic from a constructor, but it's also not obvious how to use `Lazy` to accomplish the same thing. Then things spiraled out of control a bit and I spent two days getting this to work the same way you had it already ...

@jbrry, can you review my changes? GitHub won't allow you to press "approve", but I'd still like your opinion on it.

dirkgr:
@epwalsh, this changes a lot of code that you touched before. Can you take a look too?

allennlp/training/optimizers.py
jbrry: Yes my bad, I only noticed I kept that there when reviewing.

allennlp/training/optimizers.py
jbrry: Glad that you found a better way to do this!

allennlp/training/optimizers.py
jbrry: I like your change where the optimizer options are now initialized with an empty `Dict` which makes it less likely for the outer optimizer to do anything unexpected with some placeholder options.

allennlp/training/optimizers.py
jbrry: Yes, just the default value of `set_to_none`, i.e. there is nowhere in the code where you can change this to `True`

allennlp/training/optimizers.py
jbrry: Thanks for fixing that! I didn't like the solution I had and am glad you found another way to do it.

> Do you know why this function creates a default parameter group at all? PyTorch optimizers already handle the default case just fine, so I don't get why this is here.

Yes, I'm also noticing in the output of the tests, every time an optimizer is constructed, it will create an extra group (like below), which in our case will be empty as all of our `model_parameters` for this optimizer will be assigned to groups already. I can look into this a bit further to try and see if there is a case where it's not necessary.

```
16:16:34 - INFO - allennlp.training.optimizers - Group 0: ['encoder._module.bias_ih_l0', 'encoder._module.bias_hh_l1', 'encoder._module.bias_hh_l0', 'encoder._module.bias_ih_l1'], {'optimizer_name': 'encoder', 'lr': 0.001}
16:16:34 - INFO - allennlp.training.optimizers - Group 1: ['encoder._module.weight_ih_l0', 'encoder._module.weight_hh_l0', 'encoder._module.weight_ih_l1', 'encoder._module.weight_hh_l1'], {'optimizer_name': 'encoder', 'lr': 0.002}
16:16:34 - INFO - allennlp.training.optimizers - Group 2: [], {}
```

allennlp/training/optimizers.py
jbrry: Yes, I don't think it's completely necessary to be able to view the optimizer options in the outer optimizer groups. The one place I found this to be necessary was when a `scheduler` wants to know the `param_group_field` such as the `"lr"`:

https://github.com/allenai/allennlp/blob/a02f67da036f8f7df4c9bdbd59870686d55bf4c2/allennlp/training/scheduler.py#L36-L39

So when accessing `self.optimizer.param_groups`, the default group only has a 'params' key: 
```
0 dict_keys(['params', 'betas', 'optimizer_name', 'lr', 'eps', 'weight_decay', 'amsgrad'])
1 dict_keys(['params', 'lr', 'optimizer_name', 'betas', 'eps', 'weight_decay', 'amsgrad'])
2 dict_keys(['params', 'optimizer_name', 'lr', 'betas', 'eps', 'weight_decay', 'amsgrad'])
3 dict_keys(['params']) # default group
```

which causes the following error if using a learning rate scheduler (it's line 41 instead of 38 for me as I have some debugging print statements): 
```
# Include a learning_rate_scheduler in the training config:
"learning_rate_scheduler": {
  "type": "linear_with_warmup",
  "warmup_steps": 5000,
},

File "/home/jbarry/allennlp-fork/allennlp/allennlp/training/scheduler.py", line 41, in __init__
    raise KeyError(f"{param_group_field} missing from param_groups[{i}]")
KeyError: 'lr missing from param_groups[3]'
```


The way the code is now, even when copying these optimizer options to each parameter group, the options from the `default` optimizer won't be copied over because the loop operates on [`parameter_groups`](https://github.com/jbrry/allennlp/blob/77ea6a63ca15c92940e1817fa83a28d24c3a11ca/allennlp/training/optimizers.py#L297) and the main `default` group is usually just catching what's left over from `parameter_groups` (unless you want to include some parameters in the `default` optimizer with some non-default options).

tests/training/optimizer_test.py
jbrry: > I wanted to use Lazy to create the inner optimizers. It's a problem to call into the `from_params` logic from a constructor, but it's also not obvious how to use `Lazy` to accomplish the same thing. Then things spiraled out of control a bit and I spent two days getting this to work the same way you had it already ...

That's cool that you got it to work with `Lazy`, I had initially used a `List` of `optimizers` as otherwise keying the optimizers on some name would throw an error that I hadn't registered that class if I recall correctly. By using `Lazy` it means you can construct the inner optimizers when you are ready which I think is a much better way of doing it. Thanks so much for updating all of the code, it's great to see it being done in a way I didn't think of! and sorry if I left you with a template that might have not been that flexible!

allennlp/training/optimizers.py
dirkgr: I don't think it's necessary to look into it too deeply. The only thing I'm curious about is why we create the default group at all. But arguably I'm in a better position to find out why than you are, since I know the people that wrote that code originally :-)

tests/training/optimizer_test.py
dirkgr: Your template was great, saved me a ton of time. Especially the tests! I had to learn about `Lazy` first too.

It's way easier to fill in some function signatures than to come up with the whole structure yourself :-)

allennlp/training/optimizers.py
dirkgr: We don't usually count default parameters as "hardcoded". They are just defaults.

Also, they become part of the function signature. Python will let you change them, but you really shouldn't for maximum compatibility.

allennlp/training/optimizers.py
dirkgr: This seems pretty important. Learning rate schedulers have to work. But how did this work before? I don't think we ever populated the default group this way?

tests/training/optimizer_test.py
jbrry: This test was related to the point about `learning_rate_schedulers` above, in that it needed to access the `"lr"` key from `self.optimizer.param_groups`.

It only worked in my code after [manually updating the `default` group's kwargs here](https://github.com/jbrry/allennlp/blob/fe463590d088f9c59654110bf09b65939caf75b0/allennlp/training/optimizers.py#L258-L263). I thought it should work so that the value would be visible to schedulers.

You're right, `make_parameter_groups()` only returns specified options in the dictionary (or empty for the `default` group) but after the Optimizer is initialized, the options make their way into the `self.optimizer.param_groups` so this must occur in the PyTorch source code (which is also relevant to the point above). So it might be worth trying to keep all relevant fields available to the outer optimizer, e.g. `self.optimizer.param_groups` so they can be accessed directly from there.

```
# Optimizer options for groups 0 and 1
group 0: {'lr': 0.002}
group 1: {}

# extra fields are then added to self.optimizer.param_groups after initialization
group 0: dict_keys(['params', 'lr', 'betas', 'eps', 'weight_decay', 'amsgrad', 'initial_lr'])
group 1: dict_keys(['params', 'lr', 'betas', 'eps', 'weight_decay', 'amsgrad', 'initial_lr'])

AdamOptimizer (
Parameter Group 0
    amsgrad: False
    betas: (0.9, 0.9)
    eps: 1e-08
    initial_lr: 0.002
    lr: 0.0
    weight_decay: 0.0

Parameter Group 1
    amsgrad: False
    betas: (0.9, 0.9)
    eps: 1e-08
    initial_lr: 0.00123
    lr: 0.0
    weight_decay: 0.0
)
```

allennlp/training/optimizers.py
dirkgr: I made a change in 63ec4f3817491d748de58b0a2a42116bfd8d2b61 that makes this work again. But I think this scenario would be broken with any other optimizer when using parameter groups. Can you confirm?

dirkgr:
Everything is called "parameter" in this thing. So confusing.

allennlp/training/optimizers.py
jbrry: This now works for me with a `learning_rate_scheduler` and group `3` now looks like (and seems to match the traditional `Optimizer` case):

```
0 dict_keys(['params', 'betas', 'optimizer_name', 'lr', 'eps', 'weight_decay', 'amsgrad'])
1 dict_keys(['params', 'lr', 'optimizer_name', 'betas', 'eps', 'weight_decay', 'amsgrad'])
2 dict_keys(['params', 'optimizer_name', 'lr', 'betas', 'eps', 'weight_decay', 'amsgrad'])
3 dict_keys(['params', 'lr', 'betas', 'eps', 'weight_decay', 'correct_bias'])

Parameter Group 3
    betas: (0.9, 0.999)
    correct_bias: True
    eps: 1e-08
    initial_lr: 0.0004
    lr: 0.0
    weight_decay: 0.01
)

```

tests/training/optimizer_test.py
jbrry: This test is now passing since https://github.com/allenai/allennlp/commit/63ec4f3817491d748de58b0a2a42116bfd8d2b61.

epwalsh (commented):
Overall this looks good! I didn't look too deeply into the logic, but I'm okay with the API. I just have some minor comments. Biggest thing is I'm not crazy about the "regex" name.

dirkgr (approved):

PR #4984, opened 2021-02-15
raise on HTTP errors in cached_path

### LINKED ISSUE #4978: We cache 404 responses
When you do `cached_path("https://raw.githubusercontent.com/allenai/allennlp-models/main/training_config/snli_roberta.jsonnet")`, we get a 404 response with a body. We cache the body as if nothing had happened. We should throw an exception in this case.

epwalsh:
Fixes #4978.

- When the HTTP response we get from remote resources has an error code, we raise an `HTTPError` now by simply calling `.raise_for_status()` on the response object.

### DIFF
--- a/tests/common/file_utils_test.py
+++ b/tests/common/file_utils_test.py
@@ -9,7 +9,7 @@ from filelock import Timeout
 import pytest
 import responses
 import torch
-from requests.exceptions import ConnectionError
+from requests.exceptions import ConnectionError, HTTPError
 
 from allennlp.common import file_utils
 from allennlp.common.file_utils import (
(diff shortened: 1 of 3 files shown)


tests/common/file_utils_test.py
```diff
@@ -306,6 +306,22 @@ def test_cached_path(self):
         with open(filename, "r") as f:
             assert f.read().startswith("I mean, ")
 
+    @responses.activate
+    def test_cached_path_http_err_handling(self):
+        url_404 = "http://fake.datastore.com/does-not-exist"
```
dirkgr, line 311: Can we use a URL that 404s more reliably? Maybe point to our own site, or to GitHub? I'm a little worried that someone at datastore.com can cause our tests to fail.

dirkgr (approved):
Thanks for jumping on this so quickly!

tests/common/file_utils_test.py
epwalsh: With the `@responses.activate` decorator, external requests are never made. Every URL you hit has to be mocked.

dirkgr: Oh, sweet!

PR #4986, opened 2021-02-15
Add methods for human readable representation of fields and instances

leo-liuzy:
Fixes # .

- in some cases, we might want to save instances to a JSON file; for example, in the example-based explanation method, the user might want to inspect what training examples most influence the model's prediction on a test example. So, it would be nice to add `to_json` to `Field` and `Instance` so that users don't have to convert to JSON themselves.
-


epwalsh (commented):
If all we care about is having a readable representation of an `Instance` / `Field`, then there is already a method for that `__str__(self) -> str`. So I think the `Instance` and `Field` classes should just override that method instead of adding this new method.

leo-liuzy:
I had that idea, but then realize we might want to save a structured data (json has dictionary, list, etc) rather than raw string.

epwalsh:
@leo-liuzy hmm right. I can definitely see why this would be useful, I'm just not 100% sure this is the best way to do it. A method name like `.to_json()` sounds to me like it is serializing the object to JSON, which isn't really the intended purpose here. Maybe we call it something else like `.json_repr()`? @AkshitaB, @dirkgr do you have any ideas?

dirkgr:
Both `to_json()` and `json_repr()` imply that all the information is captured. You'd expect there to be a `from_json()` method. Can we call this `human_readable_json()`, and remove the flag? @AkshitaB, is there some overlap with what you did for visualizing batches?

Also, `ListField` should be covered by this.

AkshitaB:
>  @AkshitaB, is there some overlap with what you did for visualizing batches?

Initially, I did want to utilize the `Field` info and have different visualizations for each type, but since we need to visualize the inputs right before they go into the model (so after the dataloader), we lose the `Field` abstraction. 

Does it make sense to name the function `to_dict` instead of `to_json`?

dirkgr:
Yeah, `human_readable_dict()` seems like a better name if the output isn't an actual JSON string (which it should not be).

epwalsh:
Thanks for your input @AkshitaB and @dirkgr. I like Dirk's suggestion of `human_readable_dict()` because it makes it clear what the return type is and that the function does not necessarily have an inverse.

leo-liuzy:
Got it! I didn't realize the naming of functions is such important!
BTW, should this be a method for `Field`?

epwalsh:
Yea I think this should be a method on the base class that should be overridden / implemented by each Field subclass

leo-liuzy:
But what about field like `TensorField`?

epwalsh:
For `TensorField` we have a couple of options:
1. Output the tensor as a (nested) list.
2. Display it's size, data type, mean / std dev.
3. Both 1 and 2. Or always do 2, and also do 1 if the tensor is not too big.

leo-liuzy:
Sounds legit! Will do so!

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+-  New functionality `to_json()` to support outputing instances to files
+
```
epwalsh, line 12: We already have an "Added" section below.

epwalsh (changes requested):

allennlp/data/fields/field.py
```diff
@@ -53,6 +53,28 @@ def count_vocab_items(self, counter: Dict[str, Dict[str, int]]):
         """
         pass
 
+    def human_readable_dict(self):
```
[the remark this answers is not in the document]
leo-liuzy: Got it. I will also make changes to the instance class!

CHANGELOG.md
leo-liuzy: Oh, now I see the structure of the file(using vim). Previously, I thought I need to have "###Added" for every feature. Thanks for pointing it out!

allennlp/data/fields/tensor_field.py
```diff
@@ -76,3 +76,11 @@ def __eq__(self, other) -> bool:
...
         """This is a compatibility method that returns the underlying tensor as a numpy array."""
         return self.tensor.numpy()
+
+    @overrides
+    def human_readable_dict(self):
+        size = list(self.tensor.shape)
+        std = torch.std(self.tensor.float()).item()
+        ave = torch.mean(self.tensor.float()).item()
+
+        return {"tensor": {"size": size, "element_std": std, "element_ave": ave}}
```
epwalsh, line 86: I recommend just keeping the "shape" terminology:

```suggestion
        shape = list(self.tensor.shape)
        std = torch.std(self.tensor.float()).item()
        ave = torch.mean(self.tensor.float()).item()
        return {"tensor": {"shape": shape, "element_std": std, "element_ave": ave}}
```

allennlp/data/instance.py
```diff
@@ -111,3 +113,34 @@ def duplicate(self) -> "Instance":
...
+        for key, field in self.fields.items():
+            human_readable_dict = field.human_readable_dict()
+            assert len(human_readable_dict) == 1  # each field (as dict) has its default key
+            key_in_dict = list(human_readable_dict.keys())[0]
+            if key != key_in_dict:
+                human_readable_dict[key] = deepcopy(human_readable_dict[key_in_dict])
+                del human_readable_dict[key_in_dict]
+            ret.update(human_readable_dict)
+
+        return ret
```
epwalsh, line 146: I think it would be better to keep it simple and nested by the field name:

```suggestion
        return {key: field.human_readable_dict() for key, field in self.fields.items()}
```

epwalsh (commented):
A few comments 👇

allennlp/data/instance.py
leo-liuzy: However, wouldn't writing methods this way will make us end up with something like `{"tokens": {"tokens": ["I", "like", "apple"]}}`?

leo-liuzy: How about we rename the function to be `human_readable_repr`?

epwalsh: > How about we rename the function to be human_readable_repr?

That's a good idea. That way `TextField.human_readable_repr()` could just return `List[str]`, right? Is that what you're thinking?

leo-liuzy: Exactly. But now, what's the expected type from the base class? `object`? hhh

allennlp/data/fields/field.py
```diff
@@ -54,15 +54,15 @@ def count_vocab_items(self, counter: Dict[str, Dict[str, int]]):
         """
         pass
 
-    def human_readable_dict(self) -> JsonDict:
+    def human_readable_repr(self) -> object:
```
epwalsh, line 57: Let's just make the return type `Any`. I thinks that's the canonical way to type it.

```suggestion
    def human_readable_repr(self) -> Any:
```

allennlp/data/instance.py
```diff
@@ -111,3 +112,16 @@ def duplicate(self) -> "Instance":
...
+    def human_readable_dict(self) -> JsonDict:
+        """
+        This function help to output instances to json files or print for human readability.
+        Use case includes example-based explanation, where it's better to have a output file or
+        rather than printing or logging.
+
+        For example, - if the field is LabelField, then we just output,`field.label` (string)
+                     - if the field is TextField, then we just output,
+                     field.tokens` (preferrably un-numericalized tokens)
+                     - the instances output then is {"label": `field.label`, "tokens": `field.tokens`}
```
epwalsh, line 125: Could probably just remove this part.

```suggestion
```

allennlp/data/fields/tensor_field.py
```diff
@@ -76,3 +77,10 @@ def __eq__(self, other) -> bool:
     def array(self):
         """This is a compatibility method that returns the underlying tensor as a numpy array."""
         return self.tensor.numpy()
+
+    @overrides
+    def human_readable_repr(self) -> JsonDict:
+        shape = list(self.tensor.shape)
+        std = torch.std(self.tensor.float()).item()
+        ave = torch.mean(self.tensor.float()).item()
+        return {"shape": shape, "element_std": std, "element_ave": ave}
```
epwalsh, line 86: I think we should include the data type as well.

allennlp/data/fields/field.py
```diff
@@ -53,6 +54,19 @@ def count_vocab_items(self, counter: Dict[str, Dict[str, int]]):
         """
         pass
 
+    def human_readable_repr(self) -> Any:
+        """
+        This method should be implemented by subclasses to return a structured, yet human-readable
+        representation of the field.
+
+        The return type is `JsonDict`, which means it's just a Python dictionary that is JSON-serializable.
```
epwalsh, line 62:
```suggestion
```

epwalsh (commented):
This is looking good Leo!

Can you finish up the implementation of this method for the other fields? Then we can get this merged.

CHANGELOG.md
```diff
@@ -28,7 +28,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `ModelUsage` to `ModelCard` class.
 - Added a way to specify extra parameters to the predictor in an `allennlp predict` call.
-- Added a way to initialize a `Vocabulary` from transformers models. 
+- Added a way to initialize a `Vocabulary` from transformers models.
+- Add new method on `Field` class: `.human_readable_repr() -> object`
```
epwalsh, line 32:
```suggestion
- Add new method on `Field` class: `.human_readable_repr() -> Any`
```

allennlp/data/fields/tensor_field.py
```diff
@@ -76,3 +77,15 @@ def __eq__(self, other) -> bool:
...
+    @overrides
+    def human_readable_repr(self) -> JsonDict:
+        shape = list(self.tensor.shape)
+        std = torch.std(self.tensor.float()).item()
+        ave = torch.mean(self.tensor.float()).item()
+        return {
+            "shape": shape,
+            "element_std": std,
+            "element_ave": ave,
+            "type": str(type(self.tensor)),
```
epwalsh, line 90: How about just this:

```suggestion
            "type": str(self.tensor.dtype),
```

allennlp/data/fields/tensor_field.py
```diff
@@ -76,3 +77,15 @@ def __eq__(self, other) -> bool:
...
+
+    @overrides
+    def human_readable_repr(self) -> JsonDict:
+        shape = list(self.tensor.shape)
+        std = torch.std(self.tensor.float()).item()
+        ave = torch.mean(self.tensor.float()).item()
+        return {
+            "shape": shape,
+            "element_std": std,
+            "element_ave": ave,
```
epwalsh, line 89: As a former statistician, I prefer to keep the "mean" terminology:

```suggestion
        mean = torch.mean(self.tensor.float()).item()
        return {
            "shape": shape,
            "element_std": std,
            "element_mean": mean,
```

epwalsh (commented):
Almost there 😬 Just a couple more suggestions

CHANGELOG.md
leo-liuzy: Didn't make the change before. Thanks for the scrutiny!!!

allennlp/data/fields/tensor_field.py
leo-liuzy: Good idea!

allennlp/data/fields/tensor_field.py
leo-liuzy: Sounds good! Follow the expert :D

CHANGELOG.md
```diff
@@ -28,7 +28,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `ModelUsage` to `ModelCard` class.
 - Added a way to specify extra parameters to the predictor in an `allennlp predict` call.
-- Added a way to initialize a `Vocabulary` from transformers models. 
+- Added a way to initialize a `Vocabulary` from transformers models.
+- Add new method on `Field` class: `.human_readable_repr() -> Any`
+- Add new method on `Instance` class: `.human_readable_dict() -> JsonDict`.
```
epwalsh, line 33: This actually needs to be under the "Unreleased > Added" section above (line 15).

tests/data/instance_test.py
```diff
@@ -40,3 +41,23 @@ def test_duplicate(self):
...
+        assert type(instance1.human_readable_dict()) is dict
+        assert instance1.human_readable_dict() == {"words": ["hello"], "labels": 1}
+
+        instance1_human_readable_dict = instance1.human_readable_dict()
+        array = TensorField(numpy.asarray([1.0, 1, 1]))
+        array_human_readable_dict = {
+            "shape": [3],
+            "element_mean": 1.0,
+            "element_std": 0,
+            "type": "torch.float64",
```
epwalsh, line 59: Oh, I didn't realize we'd get the "torch." in the dtype. I think it would be better if we just saw:

```suggestion
            "type": "float64",
```

So maybe in `TensorField.human_readable_repr()` we do `str(tensor.dtype).replace("torch.", "")` or something

allennlp/data/fields/field.py
```diff
@@ -53,6 +54,17 @@ def count_vocab_items(self, counter: Dict[str, Dict[str, int]]):
...
         pass
 
+    def human_readable_repr(self) -> Any:
+        """
+        This method should be implemented by subclasses to return a structured, yet human-readable
+        representation of the field.
+
+        !!! Note
+            `human_readable_repr()` is not meant to be used as a method to serialize a `Field` since the return
+            value does not necessarily contain all of the attributes of the `Field` instance.
```
epwalsh, line 64:
```suggestion
            value does not necessarily contain all of the attributes of the `Field` instance. But the object
            returned should be JSON-serializable.
```

epwalsh (approved):
Looks great @leo-liuzy!

leo-liuzy:
Hooray!!! Thanks for the scrutiny. I didn't realize producing high-quality code is so non-trivial

PR #4987, opened 2021-02-16
Add ListField example to apply token indexers

JohnGiorgi:
Addresses a [comment in the migration guide](https://github.com/allenai/allennlp/discussions/4933#discussioncomment-371349).

- Added an example for fields of type `ListField[TextField]` to `apply_token_indexers` API docs.

The documentation for `apply_token_indexers` now reads like this:

image

Let me know if there's anything you would like me to change! @epwalsh

### DIFF
--- a/allennlp/data/dataset_readers/dataset_reader.py
+++ b/allennlp/data/dataset_readers/dataset_reader.py
@@ -233,6 +233,14 @@ class DatasetReader(Registrable):
         """
         If `Instance`s created by this reader contain `TextField`s without `token_indexers`,
         this method can be overriden to set the `token_indexers` of those fields.
+
+        If your `TextField`s are wrapped in a `ListField`, you can access them via `field_list`.
+        E.g. if you had a `"source"` field of `ListField[TextField]` objects, you could:
+
+        ```python
+        for text_field in instance["source"].field_list:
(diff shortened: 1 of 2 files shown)


allennlp/data/dataset_readers/dataset_reader.py
```diff
@@ -233,6 +233,14 @@ def apply_token_indexers(self, instance: Instance) -> None:
         """
         If `Instance`s created by this reader contain `TextField`s without `token_indexers`,
         this method can be overriden to set the `token_indexers` of those fields.
+
```
epwalsh, line 236:
```suggestion

        E.g. if you have you have `"source"` `TextField`, you could implement this method like this:

        ```python
        def apply_token_indexers(self, instance: Instance) -> None:
            instance["source"].token_indexers = self._token_indexers
        ```

```

epwalsh (commented):
Thanks @JohnGiorgi! I just left a suggestion to add an example of the case where you have `TextField` by itself.

JohnGiorgi:
> Thanks @JohnGiorgi! I just left a suggestion to add an example of the case where you have `TextField` by itself.

Good call! Committed your suggestion

epwalsh (approved):
LGTM!

PR #4988, opened 2021-02-17
Add influence functions to interpret module

leo-liuzy:
Feature enrichment.

- Adding new function for example-based interpretation methods --- `SimpleInfluence`. This function will score training examples for each test example; the scores are indicators of how the training example affects the model's (learning and thus) predictions on the test example. 
-


allennlp/common/testing/interpret_test.py
```diff
@@ -43,3 +43,14 @@ def get_interpretable_layer(self):
...
+
+class DummyBilinearModelForTestingIF(Model):
+    def __init__(self, vocab, params):
+        super().__init__(vocab)
+        self.x = torch.nn.Parameter(params.float(), requires_grad=True)
+
+    def forward(self, tensors):
+        A = tensors  # (batch_size, ..., ...)
+        output_dict = {"loss": 1 / 2 * (A @ self.x @ self.x)}
+        return output_dict
```
epwalsh, line 56: I think this should probably be in `tests/interpret/simple_influence_test.py` instead of here.

epwalsh:
Looks like you have an empty module: `influence_utils` (https://github.com/allenai/allennlp/blob/8489f74fd452d759bdab6f80eb69e4e9751f046e/allennlp/interpret/influence_interpreters/influence_utils/__init__.py). Can you remove that?

leo-liuzy:
Yeah sure; I am not sure why I kept letting files/folders lurking into my commit :(

allennlp/common/testing/interpret_test.py
leo-liuzy: Sure!

allennlp/interpret/influence_interpreters/influence_interpreter.py
```diff
@@ -0,0 +1,111 @@
...
+from allennlp.models.model import Model
+from allennlp.predictors import Predictor
+from allennlp.data.dataset_readers import DatasetReader
+from allennlp.data.data_loaders import MultiProcessDataLoader
+
+
+class InfluenceInterpreter(Registrable):
+    """
+    A `SaliencyInterpreter` interprets an AllenNLP Predictor's outputs by assigning an influence
+    score to each training instance with respect to each input.
```
epwalsh, line 16:
```suggestion
    score to each training instance with respect to each test input.
```

... right?

allennlp/interpret/influence_interpreters/influence_interpreter.py
```diff
@@ -0,0 +1,111 @@
...
+        train_dataset_reader: DatasetReader,
+        test_dataset_reader: Optional[DatasetReader] = None,
+        params_to_freeze: List[str] = None,
+        k: int = 20,
+        device: int = -1,
+    ) -> None:
+        if type(device) is not int:
+            raise ValueError("'device' has to be int, -1 for cpu")
+        if type(device) is not int:
+            raise ValueError("'k' (i.e. number of supporting examples) has to be int")
```
epwalsh, line 33: The canonical way to do this would be

```python
        if not isinstance(device, int):
            raise ValueError("'device' has to be int, -1 for cpu")
        if not isinstance(k, int):
            raise ValueError("'k' (i.e. number of supporting examples) has to be int")
```

That said, I don't think having these checks are necessary since we have type hints.

```suggestion
```

allennlp/interpret/influence_interpreters/influence_interpreter.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+        self._train_loader.set_target_device(self._device)
+        self._train_loader.index_with(self.vocab)
+        self._test_loader = MultiProcessDataLoader(
+            self.test_dataset_reader, test_filepath, batch_size=1
+        )
+        self._test_loader.set_target_device(self._device)
+        self._test_loader.index_with(self.vocab)
+
+        # Number of supporting training instances has to be lass than the size of train set
```
epwalsh, line 56:
```suggestion
        # Number of supporting training instances has to be less than the size of train set
```

allennlp/interpret/influence_interpreters/influence_interpreter.py
```diff
@@ -0,0 +1,111 @@
...
+        self._k = min(k, len(self._train_loader))
+
+        self.model.to(self._device)
+
+        # so far, we assume all parameters are tuned during training
+        # we use freeze, because when model is loaded from archive, the requires_grad
+        # flag will be re-initialized to be true
+        if params_to_freeze is not None:
+            self.freeze_model(self.model, params_to_freeze, verbose=True)
+        # self._used_name2params = None  # some parameters might not be used.
```
epwalsh, line 66:
```suggestion
```
?

allennlp/interpret/influence_interpreters/influence_interpreter.py
```diff
@@ -0,0 +1,111 @@
...
+        # this is not set until we actually run the calculation, because some parameters might not be used.
+        self._used_params = None
+        self._used_params_name = None
+
+    def influence_interpret_from_json(self, inputs: JsonDict) -> JsonDict:
```
epwalsh, line 71: Why do we need this method if we pass in a `test_data_path` in `__init__`? And it seems that `SimpleInfluence` doesn't implement this method.

allennlp/interpret/influence_interpreters/simple_influence.py
```diff
@@ -0,0 +1,357 @@
...
+        `train_dataset_reader`
+    params_to_freeze: Optional[List[str]] = None
+        Optional. This is a provided list of string that for freezeing the parameters. Expectedly, each string
+        is a substring within the paramter name you intend to freeze.
+    k: int = 20
+        Optional. To demonstrate each test data, we found it most informative to just provide `k` examples with the
+        highest and lowest influence score. If not provided, we set to 20.
+    device: int = -1,
+        Optional. The index of GPU device we want to calculate scores on. If not provided, we uses -1
+        which correspond to using CPU.
```
epwalsh, line 49: Parameter descriptions for these parameters that the base class uses should go in the docstring for the base class, and then can be omitted here.

epwalsh (changes requested):

allennlp/interpret/influence_interpreters/influence_interpreter.py
leo-liuzy: Yes! Thanks for pointing that out!

allennlp/interpret/influence_interpreters/influence_interpreter.py
leo-liuzy: [facepalm]

allennlp/interpret/influence_interpreters/simple_influence.py
leo-liuzy: Make sense!

allennlp/interpret/influence_interpreters/influence_interpreter.py
leo-liuzy: Deprecated variable; deleted.

allennlp/interpret/influence_interpreters/influence_interpreter.py
leo-liuzy: No we don't. I started this method by mimicking the `attack_from_json` as in `Attackers`, but later deprecated. Again, forget to delete...

allennlp/interpret/influence_interpreters/influence_interpreter.py
```diff
@@ -0,0 +1,109 @@
...
+
+from allennlp.common import Registrable
+from allennlp.common.util import JsonDict
+from allennlp.models.model import Model
+from allennlp.predictors import Predictor
+from allennlp.data.dataset_readers import DatasetReader
+from allennlp.data.data_loaders import MultiProcessDataLoader
+
+
+class InfluenceInterpreter(Registrable):
```
epwalsh, line 13: So, how does one actually use this class since it has no methods?

I think there should be a method like this:

```python
def interpret(self, train_data_path, test_data_path):
    pass
```

And then there's no need to pass `train_filepath` and `test_filepath` into `__init__`.

This makes it clear how the class should be used, and how subclasses should implement it. And also allows an instance of this class to be used multiple times, on different test data paths.

leo-liuzy: I think that makes sense. I will make the change. However, I think it's necessary to specify the `train_file` in __init__ right? The train source should be invariant; giving freedom to change test examples does sound useful

epwalsh: Yea, that sounds reasonable

allennlp/interpret/influence_interpreters/simple_influence.py
```diff
@@ -0,0 +1,352 @@
...
+                train_grads = autograd.grad(train_loss, self._used_params)
+                # dL_test * d theta as in 2.2 of [https://arxiv.org/pdf/2005.06676.pdf]
+                influences[train_idx] = torch.dot(inv_hvp, self.flatten_tensors(train_grads)).item()
+
+            # For each test example, we output top-k and training instances
+            # Then, we record the top-k training instances
+            _, indices = torch.topk(torch.tensor(influences), self._k)
+            top_k_train_instances: List[Instance] = []
+            for idx in indices:
+                train_instance = self._train_loader._instances[idx]
```
leo-liuzy, line 197: Due to the declaration [here](https://github.com/allenai/allennlp/blob/59b92106443ff6ce26776beef5259dee3a422b40/allennlp/data/data_loaders/multiprocess_data_loader.py#L273), mypy will give 
```
allennlp/interpret/influence_interpreters/simple_influence.py:197: error: Value of type "Optional[List[Instance]]" is not indexable
```
Do you know how to fix it?

epwalsh: I think it would be better to load the training instances into a list, and then you don't need to keep `_train_loader` around.

leo-liuzy: Interesting point. Will do

epwalsh:
@leo-liuzy looks like you already lost access to Slack 😭

This is looking great! I'll give it a final look on Monday. It was great having you as an intern, and good luck next quarter!

leo-liuzy:
Thanks Pete for being such a great mentor!

epwalsh:
TODO:
- [ ] Add more tests to improve coverage
- [ ] Add a real-world example using one of our pretrained models, verify that the results make sense.
- [ ] Get feedback on the API

epwalsh (approved):

allennlp/interpret/influence_interpreters/influence_interpreter.py
```diff
@@ -0,0 +1,361 @@
...
+    """
+    The loss corresponding to the `test_instance`.
+    """
+
+    top_k: List[TrainInstanceInfluence]
+    """
+    The top `k` most influential training instances along with their influence score.
+    """
+
+    bottom_k: List[TrainInstanceInfluence]
```
matt-gardner, line 41: I would expect the bottom k to be basically noise.  Not that it's a problem to have this, but you might want to at least warn people in the docstring that this is only questionably useful for most methods.

allennlp/interpret/influence_interpreters/influence_interpreter.py
```diff
@@ -0,0 +1,361 @@
...
+    """
+    The bottom `k` least influential training instances along with their influence score.
+    """
+
+
+class TrainInstance(NamedTuple):
```
matt-gardner, line 47: This name is a bit confusing to me.  It doesn't immediately say why it's different from `Instance`.  Maybe instead use `InstanceWithGrads`?

allennlp/interpret/influence_interpreters/influence_interpreter.py
```diff
@@ -0,0 +1,361 @@
...
+from allennlp.common import Registrable, Lazy, plugins
+from allennlp.common.tqdm import Tqdm
+from allennlp.common.util import int_to_device
+from allennlp.data import Instance, DatasetReader, DatasetReaderInput, Batch
+from allennlp.data.data_loaders import DataLoader, SimpleDataLoader
+from allennlp.models import Model, Archive, load_archive
+from allennlp.nn.util import move_to_device
+
+
+class TrainInstanceInfluence(NamedTuple):
```
matt-gardner, line 17: Not sure that `Train` really adds much to this name.  `InstanceInfluence` is likely sufficient, and already implies that it's a training instance.

allennlp/interpret/influence_interpreters/influence_interpreter.py
```diff
@@ -0,0 +1,361 @@
...
+
+    instance: Instance
+    loss: float
+    grads: Sequence[torch.Tensor]
+
+
+class InfluenceInterpreter(Registrable):
+    """
+    A `SaliencyInterpreter` interprets an AllenNLP Predictor's outputs by assigning an influence
+    score to each training instance with respect to each test input.
```
matt-gardner, line 62:
```suggestion
    An `InfluenceInterpreter` interprets an AllenNLP Predictor's outputs by finding the training
    instances that had the most influence on the prediction for each test input.
```

allennlp/interpret/influence_interpreters/influence_interpreter.py
```diff
@@ -0,0 +1,361 @@
...
+                    param.requires_grad = False
+
+        # This is not set until we actually run the calculation since some parameters might not be used.
+        self._used_params: List[torch.Tensor] = []
+        self._used_param_names: List[str] = []
+
+        # Load training instances, compute loss and gradients.
+        self.train_instances: List[TrainInstance] = []
+        self._model.train()
+        for instance in Tqdm.tqdm(
```
matt-gardner, line 144: Having all of this computation in a constructor makes me really nervous.  Maybe it's fine, I just try to avoid this design.  We handled this kind of initialization in hotflip by having [a method](https://github.com/allenai/allennlp/blob/0c7d60bcf008ef963a61ee6829420aec9ecf4372/allennlp/interpret/attackers/hotflip.py#L76-L83) that [gets called](https://github.com/allenai/allennlp/blob/0c7d60bcf008ef963a61ee6829420aec9ecf4372/allennlp/interpret/attackers/hotflip.py#L248-L249) the first time it's needed.

allennlp/interpret/influence_interpreters/influence_interpreter.py
```diff
@@ -0,0 +1,361 @@
...
+class InfluenceInterpreter(Registrable):
+    """
+    A `SaliencyInterpreter` interprets an AllenNLP Predictor's outputs by assigning an influence
+    score to each training instance with respect to each test input.
+
+    Subclasses are required to implement the `calculate_influence_scores()` method.
+
+    # Parameters
+
+    model : `Model`, required
```
matt-gardner, line 68: Up above you said that this was for predictors, but here's it's a model.

allennlp/interpret/influence_interpreters/influence_interpreter.py
```diff
@@ -0,0 +1,361 @@
...
+    def _gather_instances(
+        self, scores: torch.Tensor, indices: torch.Tensor
+    ) -> List[TrainInstanceInfluence]:
+        outputs: List[TrainInstanceInfluence] = []
+        for score, idx in zip(scores, indices):
+            instance, loss, _ = self.train_instances[idx]
+            outputs.append(TrainInstanceInfluence(instance=instance, loss=loss, score=score.item()))
+        return outputs
+
+    def calculate_influence_scores(
```
matt-gardner, line 352: This looks like it should be private?  I get that you want people writing new `InfluenceInterpreters` to see this and know they have to implement this method, but you don't want people _using_ an `InfluenceInterpreter` to call this method, or even have to think about it.  Having it here with a public name makes me as a user spend time wondering if I should be using this.

matt-gardner (commented):
The basic outline seems fine to me, there wasn't anything too controversial or unexpected in here.  I had some small suggestions on naming and things, but nothing major.  If you have any specific follow-up questions, let me know.

allennlp/interpret/influence_interpreters/influence_interpreter.py
epwalsh: Yea, that's a good point. I don't see the value of keeping it, so I'm going to remove.

allennlp/interpret/influence_interpreters/influence_interpreter.py
epwalsh: Yes good point, I'll remove mention of predictors.

allennlp/interpret/influence_interpreters/influence_interpreter.py
epwalsh: Yea, I mainly wanted to make sure this showed up in the docs, but I forgot we actually have a way of allowing certain private methods to appear in the docs (like we do with `DatasetReader._read()`). I'll update to make this private and include in the docs.

PR #4990, opened 2021-02-18
Adding batchnorm verification

AkshitaB:

- `BatchNormVerification` added as a sanity check.
- Updated `ModelTestCase` to run the sanity checks for all allennlp models.

### DIFF
--- a/allennlp/sanity_checks/batch_norm_verification.py
+++ b/allennlp/sanity_checks/batch_norm_verification.py
@@ -1,0 +1,85 @@
+"""
+Code based almost entirely on
+https://github.com/awaelchli/pytorch-lightning-snippets/commit/7db53f774715d635c59ef56f21a17634d246b2c5
+"""
+
+import torch
+from torch import nn as nn
+from typing import Tuple, List, Callable
+from allennlp.sanity_checks.verification_base import VerificationBase
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class BatchNormVerification(VerificationBase):
+
+    normalization_layer = (
+        nn.BatchNorm1d,
+        nn.BatchNorm2d,
+        nn.BatchNorm3d,
+        nn.SyncBatchNorm,
+        nn.InstanceNorm1d,
+        nn.InstanceNorm2d,
+        nn.InstanceNorm3d,
+        nn.GroupNorm,
+        nn.LayerNorm,
+    )
+
+    def __init__(self, model: nn.Module):
+        super().__init__(model)
+        self._hook_handles: List[Callable] = []
+        self._module_sequence: List[Tuple[str, nn.Module]] = []
+        self._detected_pairs: List[Tuple] = []
+
+    @property
+    def detected_pairs(self) -> List[Tuple]:
+        return self._detected_pairs
+
+    def check(self, inputs) -> bool:  # type: ignore
+        inputs = self._get_inputs_copy(inputs)
+        self.register_hooks()
+        # trigger the hooks and collect sequence of layers
+        self._model_forward(inputs)
+        self.destroy_hooks()
+        self.collect_detections()
+        return not self._detected_pairs
+
+    def collect_detections(self):
+        detected_pairs = []
+        for (name0, mod0), (name1, mod1) in zip(
+            self._module_sequence[:-1], self._module_sequence[1:]
+        ):
+            bias = getattr(mod0, "bias", None)
+            detected = (
+                isinstance(mod1, self.normalization_layer)
+                and mod1.training  # TODO: do we want/need this check?
+                and isinstance(bias, torch.Tensor)
+                and bias.requires_grad
+            )
+            if detected:
+                detected_pairs.append((name0, name1))
+        self._detected_pairs = detected_pairs
+        if detected_pairs:
+            logger.warning(
+                f"The following invalid BatchNorm-bias layer pairings were detected: {detected_pairs}"
+            )
+        return detected_pairs
+
+    def register_hooks(self):
+        hook_handles = []
+        for name, module in self.model.named_modules():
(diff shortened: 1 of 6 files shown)


AkshitaB:
The verification check found invalid layer combinations in our QANet models!

dirkgr:
> The verification check found invalid layer combinations in our QANet models!

@AkshitaB, what exactly did it find?

dirkgr:
Looks like it found a `BatchNorm` layer after a layer with bias. @matt-gardner, is that an oversight in QANet?

allennlp/sanity_checks/batch_norm_verification.py
```diff
@@ -0,0 +1,85 @@
...
+    normalization_layer = (
+        nn.BatchNorm1d,
+        nn.BatchNorm2d,
+        nn.BatchNorm3d,
+        nn.SyncBatchNorm,
+        nn.InstanceNorm1d,
+        nn.InstanceNorm2d,
+        nn.InstanceNorm3d,
+        nn.GroupNorm,
+        nn.LayerNorm,
```
dirkgr, line 26: Given that it detects all of these, can we call it something else? `NormalizationVerification` or something?

allennlp/training/trainer.py
```diff
@@ -177,6 +178,50 @@ def on_end(
...
+    ) -> None:
+        if not is_primary:
+            return None
+
+        # We destroy the hooks after the first batch, since we only want to
+        # perform this check once.
+        if epoch == 0 and batch_number == 1 and is_training:
+            self._verification.destroy_hooks()
+            detected_pairs = self._verification.collect_detections()
+            # TODO: Should we actual fail with an error instead?
```
dirkgr, line 217: Yes, but we should give a way to not run this check.

allennlp/training/trainer.py
```diff
@@ -177,6 +178,50 @@ def on_end(
...
+            if detected_pairs:
+                logger.warning(
+                    "The model failed the BatchNormVerification check. "
+                    f"It contains {len(detected_pairs)} invalid layer combination(s)."
+                )
```
dirkgr, line 222: Can we have a more descriptive error message, so people working on models can immediately find where the problem is?

tests/training/trainer_test.py
```diff
@@ -755,6 +757,31 @@ def test_trainer_can_log_learning_rates_tensorboard(self):
 
         trainer.train()
 
+    def test_sanity_check_callback(self):
+        class BiasBatchNormModel(Model):
+            def __init__(self, use_bias=True):
```
dirkgr, line 762: This has the `use_bias` parameter, but it's always `True`. Did you mean to parameterize this test to test both cases?

dirkgr: Wait, this class is here twice. Can you re-use it from the other file?

dirkgr (commented):
This is very cool. I wish we could think of 10 more checks like it.

Can you think of a way to run this all the time by default, but give a way to turn it off sometimes, without blowing up the config? I can't think of a way to do this without plumbing another parameter into `GradientDescentTrainer.from_params`. But maybe that's worth it?

matt-gardner:
> Looks like it found a `BatchNorm` layer after a layer with bias. @matt-gardner, is that an oversight in QANet?

Quite possibly; I don't know the answer to this question, as I'm not super familiar with the internals of QANet, or with the batchnorm issue being checked here.  I wasn't the one who implemented this.

allennlp/sanity_checks/batch_norm_verification.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def collect_detections(self):
+        detected_pairs = []
+        for (name0, mod0), (name1, mod1) in zip(
+            self._module_sequence[:-1], self._module_sequence[1:]
+        ):
+            bias = getattr(mod0, "bias", None)
+            detected = (
+                isinstance(mod1, self.normalization_layer)
+                and mod1.training  # TODO: do we want/need this check?
```
dirkgr, line 56: I think the whole thing only makes sense if we're training. When evaluating, you might want to run all kinds of crazy stuff, because you're making baselines of somebody else's code.

dirkgr:
@AkshitaB, can we fix it in QANet and re-train? It trains pretty quickly.

AkshitaB:
@dirkgr Can you take a look at this? This PR will have to be merged because the parameter for ignoring the sanity checks currently failing is also added as part of this PR.

allennlp/training/trainer.py
```diff
@@ -195,6 +196,47 @@ def on_end(
 TrainerCallback.register("null")(TrainerCallback)
 
 
+@TrainerCallback.register("sanity-checks")
+class SanityCheckCallback(TrainerCallback):
+    """
+    Performs model sanity checks.
+    """
```
dirkgr, line 203: It would be nice if somewhere in this class it explained that even though we have only one sanity check right now, this is the place to add additional ones.

allennlp/sanity_checks/normalization_bias_verification.py
```diff
@@ -0,0 +1,104 @@
...
+                isinstance(mod1, self.normalization_layers)
+                and mod1.training
+                and isinstance(bias, torch.Tensor)
+                and bias.requires_grad
+            )
+            if detected:
+                detected_pairs.append((name0, name1))
+        self._detected_pairs = detected_pairs
+        if detected_pairs:
+            logger.warning(self._verification_message())
```
dirkgr, line 69: I thought we wanted to throw an exception if this fails? Otherwise it will get lost in a sea of messages during training.

AkshitaB: We throw an exception in the trainer callback.

